PR #14305, opened 2025-12-03
fix: recompute IDP role mapping on every login when enabled

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-9362

## Description

When "**Computed during each user authentication**" was selected in the IDP configuration, admin-assigned roles continued to override the IDP role mapping. As a result, roles were not recomputed during each authentication as expected.

Now, IDP role mapping is always executed on every login when the option is enabled.

Fix:

https://github.com/user-attachments/assets/f6cd546d-3119-405f-b712-a58bd379fbc5

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
@@ -2139,10 +2139,9 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
         List<Membership> overrideUserMemberships = new ArrayList<>();
         // Delete existing memberships
         userMemberships.forEach(membership -> {
-            // Consider only membership "created by" the identity provider
-            if (identityProviderId.equals(membership.getSource())) {
-                // if there is no mapping configured on the social idp, we do not remove / reset it
-                if (hasMapping) {
+            if (hasMapping) {
+                // Consider only membership "created by" the identity provider
+                if (identityProviderId.equals(membership.getSource())) {
                     membershipService.deleteReferenceMemberBySource(
                         executionContext,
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -2153,7 +2153,17 @@ private void refreshUserMemberships(
...
+                        MembershipReferenceType.valueOf(membership.getReferenceType().name()),
+                        membership.getReferenceId(),
+                        MembershipMemberType.USER,
+                        userId
+                    );
+                } else {
+                    overrideUserMemberships.add(membership);
+                }
             }
         });
```
phiz71, line 2166: I think it would be clearer like this. WDYT ?
```java
        // Delete existing memberships
        userMemberships.forEach(membership -> {
            if (hasMapping) {
                // Consider only membership "created by" the identity provider
                if (identityProviderId.equals(membership.getSource())) {
                    // if there is no mapping configured on the social idp, we do not remove / reset it
                    membershipService.deleteReferenceMemberBySource(
                        executionContext,
                        MembershipReferenceType.valueOf(membership.getReferenceType().name()),
                        membership.getReferenceId(),
                        MembershipMemberType.USER,
                        userId,
                        membership.getSource()
                    );
                } else {
                    membershipService.deleteReferenceMember(
                        executionContext,
                        MembershipReferenceType.valueOf(membership.getReferenceType().name()),
                        membership.getReferenceId(),
                        MembershipMemberType.USER,
                        userId
                    );
                }
            } else {
                overrideUserMemberships.add(membership);
            }
        });
```

vikrantgravitee: Done.

phiz71 (approved):

PR #14324, opened 2025-12-03
feat(portal): implement create application form

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-12056

## Description

Implements the create application form in the portal with dynamic field generation based on selected application type. 

### What's missing

The following items are not included in this PR and will be addressed in separate subtasks:

1. Additional metadata field (requires new component for key-value pairs)
2. Application type descriptions alignment with Figma design (requires backend changes)
3. Route guards for permission checks

## Additional context 

Demo videos (both desktop and mobile res): 

https://github.com/user-attachments/assets/7a258fd0-073c-4bc9-bdf5-136956c0f793

https://github.com/user-attachments/assets/dac8660d-d862-4c8b-b570-2c16dc968566

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
+++ b/gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
@@ -1,13 +1,13 @@
 <!--
 
     Copyright (C) 2025 The Gravitee team (http://gravitee.io)
-    
+
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
     You may obtain a copy of the License at
-    
+
             http://www.apache.org/licenses/LICENSE-2.0
-    
+
     Unless required by applicable law or agreed to in writing, software
     distributed under the License is distributed on an "AS IS" BASIS,
     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
```diff
@@ -20,5 +20,163 @@
...
+      <div class="cards-container" [appIsMobile]="'cards-container--mobile'">
+        <mat-card appearance="outlined">
+          <mat-card-content>
+            <h3 class="section-title" i18n="@@createApplicationSecuritySetupTitle">Set up security</h3>
+            @if (isMobile()) {
+              <mat-form-field appearance="outline" class="type-select-mobile">
+                <mat-label i18n="@@createApplicationTypeTitleMobile">Type</mat-label>
+                <mat-select [formControl]="typeIdControl">
+                  @for (type of types; track type.id ?? type.name) {
+                    <mat-option [value]="type.id ?? type.name">{{ type.name }}</mat-option>
```
jourdiw: The application types don't always have an id?

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
```diff
@@ -20,5 +20,163 @@
...
+    @if (enabledApplicationTypes(); as types) {
+      <div class="cards-container" [appIsMobile]="'cards-container--mobile'">
+        <mat-card appearance="outlined">
+          <mat-card-content>
+            <h3 class="section-title" i18n="@@createApplicationSecuritySetupTitle">Set up security</h3>
+            @if (isMobile()) {
+              <mat-form-field appearance="outline" class="type-select-mobile">
+                <mat-label i18n="@@createApplicationTypeTitleMobile">Type</mat-label>
+                <mat-select [formControl]="typeIdControl">
+                  @for (type of types; track type.id ?? type.name) {
```
jourdiw: If you want to avoid the check, you can choose one of the internal values:
```suggestion
                  @for (type of types; track $index) {
```

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
```diff
@@ -20,5 +20,163 @@
...
+                      >The client_id of the application. This field is required to subscribe to certain types of API Plan (OAuth2, JWT)
+                    </mat-hint>
+                  </mat-form-field>
+                }
+
+                @if (grantTypesList().length > 0) {
+                  <div class="grant-types-container">
+                    <div>
+                      <div class="m3-title-medium">
+                        <span i18n="@@createApplicationGrantTypes">Select the grant types your application will use*</span>
```
jourdiw: I think we can leave off the "*" for now, this isn't necessarily a design pattern that we have in the current portal 🤔

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
```diff
@@ -20,5 +20,163 @@
...
+                      <div class="grant-type-toggle">
+                        <mat-slide-toggle
+                          [checked]="isGrantTypeSelected(grantType.type)"
+                          [disabled]="grantType.isDisabled"
+                          (change)="toggleGrantType(grantType.type, $event.checked)">
+                          <span class="grant-type-label">{{ grantType.name }}</span>
+                        </mat-slide-toggle>
+                      </div>
+                    }
+                    @if (grantTypesControl && grantTypesControl.touched && grantTypesControl.hasError('mandatoryMissing')) {
```
jourdiw: How can a user be missing mandatory grant types? I would imagine that they would be true and disabled if they are mandatory? 🤔

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.scss
jourdiw, line 1: If you can, try to avoid font attributes in the scss for the NG Portal. You can try to find the closest match in https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-portal-webui-next/src/scss/m3-adapter.scss

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.ts
```diff
@@ -13,15 +13,318 @@
...
+    const formValue = this.form.getRawValue();
+    const selectedType = this.selectedType();
+
+    if (!selectedType) {
+      return;
+    }
+
+    const applicationInput = this.buildApplicationInput(formValue, selectedType);
+
+    this.applicationService.create(applicationInput).subscribe({
```
jourdiw: You can include `takeUntilDestroyed` to handle closing the pipe once the navigation has started 👍

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.ts
```diff
@@ -13,15 +13,318 @@
...
+    if (!typeId) return null;
+    const types = this.enabledApplicationTypes();
+    if (!types) return null;
+    return types.find(type => type.id === typeId || type.name === typeId) ?? null;
+  });
+
+  readonly isSimpleType = computed(() => this.selectedType()?.id === 'simple');
+  readonly requiresRedirectUris = computed(() => this.selectedType()?.requires_redirect_uris ?? false);
+
+  readonly grantTypesList = computed<GrantTypeVM[]>(() => {
```
jourdiw, line 114: WDYT about adding logic in this computed that replaces the need for `isGrantTypeSelected` in the template?

gravitee-apim-portal-webui-next/src/entities/application/application.ts
```diff
@@ -70,16 +70,20 @@ export interface ApplicationSettingsApp {
 }
 
 export interface ApplicationSettingsOAuth {
```
jourdiw, line 72: Why did they need to be all marked as optional?

jourdiw (commented):
Will you have a separate PR for styling the application cards when selected to match Figma?

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
```diff
@@ -20,5 +20,163 @@
...
+    @if (hasApplicationError) {
+      <div class="m3-title-medium error-message" i18n="@@createApplicationGeneralErrorMessage">
+        There was an error creating your application and it could not be processed. Try again, and if the issue persists, contact your
+        portal administrator.
+      </div>
+    }
+
+    <div class="actions">
+      <button
+        mat-button
```
jourdiw: You can use a `mat-filled-button` here to match the Figma 👍

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.scss
JedrzejJanasiak: Sure, no problemo. Done :)

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.ts
JedrzejJanasiak: Makes sense. I also think it would be a nice cleanup for the template.
But when I started changing it, it turned out that because grant types are stored in a dynamically added reactive FormControl, keeping an isSelected flag reactive inside the computed is not that elegant. To do it properly, you need an observable to signal bridge plus some extra effect/subscription/cleanup logic.
For the current scope I’d keep the simple isGrantTypeSelected() method, so we avoid extra plumbing for a small template win, but I’d definitely simplify this once Signal Forms finally become stable 🚀
That said, if you still think it’s worth pushing in that direction, I can refactor it this way.

gravitee-apim-portal-webui-next/src/entities/application/application.ts
JedrzejJanasiak: I followed the Portal API and, from what I see in other apps, we stick to the same scheme too (basically all fields optional):
- [Console](https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-console-webui/src/entities/application/Application.ts)
- [Old Portal](https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-portal-webui/projects/portal-webclient-sdk/src/lib/model/oAuthClientSettings.ts)

`client_id` and `client_secret` are backend-generated, so they have to stay nullable on creation. For some other fields we could make them required (like `application_type`), but then we’d lose consistency between the entity model and the API contract.

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
JedrzejJanasiak: sure 👍

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.ts
JedrzejJanasiak: done

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
JedrzejJanasiak: Yeah, you're right. It's basically impossible, removed that

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
JedrzejJanasiak: 👍

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
JedrzejJanasiak: sure, good call, thanks

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
JedrzejJanasiak: I don't think so, fixed

JedrzejJanasiak:
> Will you have a separate PR for styling the application cards when selected to match Figma?

Done sth like this: 

https://github.com/user-attachments/assets/7ece69fa-6213-40a9-8b34-6fc983102b28

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.ts
jourdiw: That works for me to keep the method in the template 👍

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
```diff
@@ -20,5 +20,162 @@
...
+                </mat-select>
+              </mat-form-field>
+            } @else {
+              <span class="radio-group-label m3-title-medium" id="application-type-radio-group-label" i18n="@@createApplicationTypeLabel"
+                >Type</span
+              >
+              <mat-radio-group aria-labelledby="application-type-radio-group-label" class="types-group" [formControl]="typeIdControl">
+                @for (type of types; track type.id ?? type.name) {
+                  <mat-radio-button class="type-radio" [value]="type.id ?? type.name">
+                    <div class="application-type-label">
```
jourdiw: Can this match like what you have on line 33?

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
```diff
@@ -20,5 +20,162 @@
...
+                      <input
+                        [matChipInputFor]="redirectURIChipGrid"
+                        [matChipInputAddOnBlur]="true"
+                        (matChipInputTokenEnd)="addRedirectUri($event)" />
+                    </mat-chip-grid>
+                    <mat-hint i18n="@@applicationRedirectUrisHint">
+                      The URI where the authorization server sends OAuth responses.
+                    </mat-hint>
+                    @if (redirectUrisCtrl.touched && (redirectUrisCtrl.hasError('required') || redirectUrisCtrl.hasError('minlength'))) {
+                      <mat-error i18n="@@redirectUrisRequired">At least one redirect URI is required</mat-error>
```
jourdiw: If you are using the same error for both, WDYT:

```suggestion
                      <mat-error i18n="@@redirectUrisRequired">At least one valid redirect URI is required</mat-error>
```

gravitee-apim-portal-webui-next/src/entities/application/application.ts
jourdiw: Perhaps that's the difference between the object `ApplicationInput` for creating an application vs the object that's returned. 🤔 

The only downside is now we will have a lot of undefined checks if we want to use a field in the object.

I would encourage to base upon what the backend is expecting than what is in Console/Old Portal with the objects, only because the objects aren't always up to date. 👍

gravitee-apim-portal-webui-next/src/app/applications/create-application/create-application.component.html
JedrzejJanasiak: Sure, I've missed that 😅

jourdiw (approved):
Nice work!

gravitee-apim-portal-webui-next/src/entities/application/application.ts
jourdiw: As discussed, let's just keep it as all optional fields as you have added in the PR 👍

PR #14346, opened 2025-12-05
feat(helm): add distributed sync for gateway

mukul-tyagi08:
## Issue

https://gravitee.atlassian.net/browse/APIM-11038
## Description

Add helm config for distributed sync with redis.

### DIFF
--- a/helm/templates/gateway/gateway-configmap.yaml
+++ b/helm/templates/gateway/gateway-configmap.yaml
@@ -24,6 +24,89 @@ metadata:
 data:
   {{- if not (include "gateway.externalConfig" .)}}
   gravitee.yml: |
+    # Distributed Sync Setup
+    {{- if .Values.gateway.cluster }}
+    cluster:
+      type: {{ .Values.gateway.cluster.type }}
+      {{- if eq .Values.gateway.cluster.type "hazelcast" }}
+      hazelcast:
+        config-path: {{ .Values.gateway.cluster.hazelcast.configPath }}
+      {{- end }}
+    {{- end }}
+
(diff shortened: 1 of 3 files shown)


helm/templates/gateway/gateway-configmap.yaml
```diff
@@ -24,6 +24,25 @@ metadata:
...
+      {{- end }}
+    {{- end }}
+
+    {{- if .Values.gateway.distributedSync }}
+    distributed-sync:
+      type: {{ .Values.gateway.distributedSync.type }}
+      {{- if eq .Values.gateway.distributedSync.type "redis" }}
+      redis:
+        host: {{ .Values.gateway.distributedSync.redis.host }}
+        port: {{ .Values.gateway.distributedSync.redis.port }}
```
remibaptistegio, line 43: We should be able to configure redis with all the properties available here like ssl etc ... Just like we do for rate limit for instance.

helm/values.yaml
```diff
@@ -915,6 +915,22 @@ gateway:
...
+#    hazelcast:
+#      configPath: /opt/graviteeio-gateway/config/hazelcast.xml
+
+  # Uncomment below to enable Redis-based distributed sync
+#  distributedSync:
+#    enabled: true
+#    type: redis
+#    redis:
+#      host: redis
+#      port: 6379
```
remibaptistegio, line 932: Same comment here.

remibaptistegio (changes requested):
I have added comments for the redis configuration. I think you should also add a unit test to validate the changes.

remibaptistegio (approved):

phiz71 (approved):

PR #14357, opened 2025-12-05
fix: allow llm proxy entrypoint configuration

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-12062

## Description

When we create an LLM proxy entrypoint we are able to configure token tracking. 
This PR makes it editable after creation, in _APIs_ -> _Entrypoints_.

Here an exemple with the LLM proxy option : 

Capture d’écran 2025-12-05 à 11 05 42

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/entrypoints-v4/api-entrypoints-v4-general.component.html
+++ b/gravitee-apim-console-webui/src/management/api/entrypoints-v4/api-entrypoints-v4-general.component.html
@@ -162,6 +162,49 @@
       </div>
     </mat-card-content>
   </mat-card>
+  <mat-card *ngIf="api?.type === 'LLM_PROXY'">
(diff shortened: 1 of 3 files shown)


jgiovaresco (changes requested):
The configuration of the entrypoint come from the json-schema inside the plugin.

We should follow the UX from Message APIs

Image

Having a list beween context-path & exposed entrypoint that will contain only one entrypoint (LLM Proxy) we can edit to update the configuration. In LLM Proxy, we can get rid of the QUALITY OF SERVICE column in that Entrypoints types table

thomasballivet:
With the changes :

Capture d’écran 2025-12-05 à 16 18 37


Capture d’écran 2025-12-05 à 16 18 52

jgiovaresco (approved):

PR #14365, opened 2025-12-05
feat(portal): implement permission and require-auth guard

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-12172

## Description

Fixed security issue where unauthenticated users could access `/applications` and `/applications/create` routes when "Force authentication to access portal" setting was disabled.

To do so, I implemented two new guards:

### Permission guard

Based on the permission guard we have in Console, this one checks for correct permissions.

It uses `anyOf` to provide a list of permissions required to access the path as well as the `unauthorizedFallbackTo` parameter.

### Require auth

This is a twin implementation of the auth guard but with one difference. It doesn't look into Console -> Settings -> Authentication -> "Force authentication to access portal" parameter, so it always requires being logged in to enter.

This way we don't have a scenario like we have on master env when this parameter is turned off and unlogged user can access both `/applications` and `/applications/create`.

## Additional context

Demo of Permission Guard on `/applications/create` route with fallback when user has no required permissions:

https://github.com/user-attachments/assets/f89eccd9-cdfd-4734-85dd-161af01ed2c6

Demo of Require Auth Guard on `/applications` and `/applications/create` routes with unlogged user:

https://github.com/user-attachments/assets/ffb5cad9-b353-4e56-91d7-2188c66f52fe

### DIFF
--- a/gravitee-apim-portal-webui-next/src/guards/permission.guard.ts
+++ b/gravitee-apim-portal-webui-next/src/guards/permission.guard.ts
@@ -1,0 +1,128 @@
+/*
+ * Copyright (C) 2025 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { inject } from '@angular/core';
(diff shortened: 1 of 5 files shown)


gravitee-apim-portal-webui-next/src/guards/permission.guard.ts
```diff
@@ -0,0 +1,128 @@
...
+    resource: resource as PermissionResource,
+    action: action as PermissionAction,
+  };
+}
+
+function isValidAction(action: string): action is PermissionAction {
+  return ['C', 'R', 'U', 'D'].includes(action);
+}
+
+function hasUserPermission(userPermissions: UserEnvironmentPermissions, parsed: ParsedPermission): boolean {
```
graviteeio, line 114: ### :warning: Aqua detected sensitive data in your code  
**Category:** BaseAuth   
**Description:** BaseAuth Standalone Credentials   
**Severity:** HIGH   
**Match:** erPermission(userPermissions: **************************, parsed: ParsedPerm  
[This comment was created by Aqua Pipeline]

PR #14371, opened 2025-12-07
Move APIM Reporters dependency in the monorepo

phiz71:
## Description

Move reporter-common, reporter-file & reporter-elasticsearch inside APIM.
A new module has been created: gravitee-apim-reporters


gravitee-apim-reporter/pom.xml
```diff
@@ -0,0 +1,78 @@
+<!--
```
jhaeyaert, line 1: Wondering if all these new modules should be added to the apim-bom? I suppose it will allow us to implement other reporters (ex: TCP)

ThibaudAV (approved):

jgiovaresco (approved):

PR #14382, opened 2025-12-08
apim-11964-delete-portal-section-be

dev-marek:
## Issue

https://gravitee.atlassian.net/browse/APIM-11964

## Description

Added service to delete navigation items.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/exception/ItemHasChildrenException.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/exception/ItemHasChildrenException.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal_page.exception;
+
+import io.gravitee.apim.core.exception.ValidationDomainException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ItemHasChildrenException extends ValidationDomainException {
+
+    public ItemHasChildrenException(String message, Map<String, String> parameters) {
+        super(message, parameters);
+    }
+
+    public static ItemHasChildrenException forId(String id) {
+        Map<String, String> params = new HashMap<>();
+        params.put("id", id);
+        return new ItemHasChildrenException("Item has children", params);
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationFolder.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationFolder.java
@@ -16,7 +16,9 @@
 package io.gravitee.apim.core.portal_page.model;
 
 import jakarta.annotation.Nonnull;
+import lombok.experimental.SuperBuilder;
 
+@SuperBuilder
 public final class PortalNavigationFolder extends PortalNavigationItem {
 
     private static final PortalNavigationItemType TYPE = PortalNavigationItemType.FOLDER;
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationLink.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationLink.java
@@ -18,8 +18,10 @@ package io.gravitee.apim.core.portal_page.model;
 import jakarta.annotation.Nonnull;
 import lombok.Getter;
 import lombok.Setter;
+import lombok.experimental.SuperBuilder;
 
 @Getter
+@SuperBuilder
 public final class PortalNavigationLink extends PortalNavigationItem {
 
     private static final PortalNavigationItemType TYPE = PortalNavigationItemType.LINK;
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationPage.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationPage.java
@@ -18,8 +18,10 @@ package io.gravitee.apim.core.portal_page.model;
 import jakarta.annotation.Nonnull;
 import lombok.Getter;
 import lombok.Setter;
+import lombok.experimental.SuperBuilder;
 
 @Getter
+@SuperBuilder(toBuilder = true)
 public final class PortalNavigationPage extends PortalNavigationItem {
 
     private static final PortalNavigationItemType TYPE = PortalNavigationItemType.PAGE;
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/DeletePortalNavigationItemUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/DeletePortalNavigationItemUseCase.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal_page.use_case;
+
+import io.gravitee.apim.core.UseCase;
+import io.gravitee.apim.core.portal_page.crud_service.PortalNavigationItemCrudService;
+import io.gravitee.apim.core.portal_page.crud_service.PortalPageContentCrudService;
+import io.gravitee.apim.core.portal_page.exception.PortalNavigationItemNotFoundException;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemComparator;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationPage;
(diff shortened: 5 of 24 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/DeletePortalNavigationItemUseCase.java
```diff
@@ -0,0 +1,92 @@
...
+            var processingId = toProcess.poll();
+            var children = portalNavigationItemsQueryService.findByParentIdAndEnvironmentId(input.environmentId(), processingId);
+            children.forEach(item -> idsToDelete.add(item.getId()));
+            children.forEach(child -> toProcess.add(child.getId()));
+            children
+                .stream()
+                .filter(item -> item instanceof PortalNavigationPage)
+                .map(PortalNavigationPage.class::cast)
+                .map(PortalNavigationPage::getPortalPageContentId)
+                .forEach(contentIdsToDelete::add);
```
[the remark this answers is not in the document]
jourdiw: I was going to suggest something like this as well! 
It would be nice to avoid looping multiple times over the same list.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/DeletePortalNavigationItemUseCase.java
jourdiw, line 1: WDYT about moving the actual deleting act to the `PortalNavigationItemDomainService`? That way it can be more easily reused in the future 🤔

dev-marek: Reused where? I'm ok with refactor/optimization, but I'd avoid doing that without a real need for that.

jourdiw: With CRD and GKO needs 🤔

dev-marek: I think it will be easier for them to extract logic they need from this use case than doing a guesswork what would they need.

jourdiw: Ok I understand. We currently have create and update there, so I was guessing that the delete would be in there too 🤷‍♀️

phiz71:
I thought we said we only delete a folder when it is empty. I'm not sure this is a behaviour we would like to have to delete everything inside a folder.
In the current documentation feature, we block the deletion if a folder is not empty.

To check with produyct, but I think that it would be simpler for everyone if we keep the delete resource basic.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationFolder.java
```diff
@@ -16,7 +16,9 @@
 package io.gravitee.apim.core.portal_page.model;
 
 import jakarta.annotation.Nonnull;
+import lombok.experimental.SuperBuilder;
 
+@SuperBuilder
 public final class PortalNavigationFolder extends PortalNavigationItem {
 
     private static final PortalNavigationItemType TYPE = PortalNavigationItemType.FOLDER;
```
phiz71, line 24: Should we add `@Builder.Default` here ?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/exception/ItemHasChildrenException.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal_page.exception;
+
+import io.gravitee.apim.core.exception.ValidationDomainException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ItemHasChildrenException extends ValidationDomainException {
```
jourdiw: This seems to be a bit different from how the other exceptions are named 🤔 

WDYT about:
- FolderNotEmptyException
- PortalNavigationItemHasChildrenException
? Other ideas?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/DeletePortalNavigationItemUseCase.java
```diff
@@ -0,0 +1,80 @@
...
+        var parentId = existing.getParentId();
+        var deletedOrder = existing.getOrder();
+        var siblings = portalNavigationItemsQueryService.findByParentIdAndEnvironmentId(input.environmentId(), parentId);
+        var siblingsToUpdate = siblings
+            .stream()
+            .sorted(PortalNavigationItemComparator.byOrder())
+            .filter(sibling -> !sibling.getId().equals(existing.getId()))
+            .filter(sibling -> sibling.getOrder() > deletedOrder)
+            .toList();
+        siblingsToUpdate.forEach(sibling -> sibling.setOrder(sibling.getOrder() - 1));
```
jourdiw: This seems to be a very similar action to what we have in the `PortalNavigationItemDomainService` 🤔 

```
this.retrieveSiblingItems(
                portalNavigationItem.getParentId(),
                portalNavigationItem.getEnvironmentId(),
                createPortalNavigationItem.getArea()
            )
            .stream()
            .filter(item -> !Objects.equals(item.getId(), portalNavigationItem.getId()))
            .filter(sibling -> sibling.getOrder() >= portalNavigationItem.getOrder())
            .forEach(followingSibling -> {
                followingSibling.setOrder(followingSibling.getOrder() + 1);
                this.crudService.update(followingSibling);
            });
```

What do you think about mutualizing the code somehow?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationLink.java
```diff
@@ -18,8 +18,10 @@
 import jakarta.annotation.Nonnull;
 import lombok.Getter;
 import lombok.Setter;
+import lombok.experimental.SuperBuilder;
 
 @Getter
+@SuperBuilder
 public final class PortalNavigationLink extends PortalNavigationItem {
 
     private static final PortalNavigationItemType TYPE = PortalNavigationItemType.LINK;
```
phiz71, line 27: Should we add @Builder.Default here?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationPage.java
```diff
@@ -18,8 +18,10 @@
 import jakarta.annotation.Nonnull;
 import lombok.Getter;
 import lombok.Setter;
+import lombok.experimental.SuperBuilder;
 
 @Getter
+@SuperBuilder(toBuilder = true)
 public final class PortalNavigationPage extends PortalNavigationItem {
 
     private static final PortalNavigationItemType TYPE = PortalNavigationItemType.PAGE;
```
phiz71, line 27: Should we add @Builder.Default here?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/DeletePortalNavigationItemUseCase.java
```diff
@@ -0,0 +1,80 @@
...
+        var existing = portalNavigationItemsQueryService.findByIdAndEnvironmentId(input.environmentId(), input.navigationItemId());
+        if (existing == null) {
+            throw new PortalNavigationItemNotFoundException(input.navigationItemId().json());
+        }
+
+        // If the item has direct children, deletion is not allowed
+        var directChildren = portalNavigationItemsQueryService.findByParentIdAndEnvironmentId(input.environmentId(), existing.getId());
+        if (!directChildren.isEmpty()) {
+            throw ItemHasChildrenException.forId(existing.getId().toString());
+        }
```
phiz71: Should we do it only for folders?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/use_case/DeletePortalNavigationItemUseCaseTest.java
```diff
@@ -0,0 +1,197 @@
...
+        portalNavigationItemsCrudService.reset();
+        portalNavigationItemsQueryService.reset();
+    }
+
+    @Test
+    void should_delete() {
+        // Given
+        var toDelete = PortalNavigationItemFixtures.aPage(PortalNavigationItemFixtures.PAGE11_ID, "page11", null);
+        portalNavigationItemsCrudService.initWith(List.of(toDelete));
+        portalNavigationItemsQueryService.initWith(List.of(toDelete));
```
phiz71, line 72: When you init a query service with a list, does it init also the inmemory storage used ? If yes, then you probably don't need to init both with the same list. WDYT?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/use_case/DeletePortalNavigationItemUseCaseTest.java
```diff
@@ -0,0 +1,197 @@
...
+            )
+        );
+
+        // Then
+        assertThat(portalNavigationItemsCrudService.storage()).isEmpty();
+        assertThat(portalPageContentCrudService.storage()).isEmpty();
+    }
+
+    @Test
+    void should_delete_cascade() {
```
phiz71:
```suggestion
    void should_not_delete_cascade() {
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationFolder.java
dev-marek: Not for a static field.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationLink.java
dev-marek: For static field not.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationPage.java
dev-marek: Static field no-no.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/use_case/DeletePortalNavigationItemUseCaseTest.java
dev-marek: It is not true for all test cases.

phiz71 (approved):

PR #14390, opened 2025-12-08
fix: api migration parsing

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-11978

## Description

When trying to import an API V4 that has been migrated from V2, the generated schema is invalid.

```
"endpoints" : [ {
      "name" : "default",
      "type" : "http-proxy",
      "weight" : 1,
      "inheritConfiguration" : true,
      "configuration" : {
        "target" : "http://localhost:8080"
      },
      "services" : {
        "healthCheck" : {
          "overrideConfiguration" : false,
          "configuration" : {
            "schedule" : null,
            "failureThreshold" : 2,
            "successThreshold" : 2
          },
          "enabled" : true,
          "type" : "http-health-check"
        }
      },
```

`target` is expected in the service configuration, and `schedule` should not be null.
When inherited, we should stop serializing the health check service configuration at the endpoint level. This will prevent the schema invalidation and eliminate redundancy.


thomasballivet:
To avoid others schema validation issues I added some conditions to make sure we do not migrate `ScheduledService`s that are not valid (schedule is null), and also added some conditions to not set fields with null values in the endpoint group (that would also lead to schema invalidation).

jgiovaresco (approved):

PR #14396, opened 2025-12-08
fix: remove explicit error messages leaking user existence

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-11945
https://gravitee.atlassian.net/browse/APIM-12183

## Description

Replaced explicit user-identifying error messages with generic responses to prevent revealing whether an account or email exists.

### DIFF
--- a/gravitee-apim-console-webui/src/user/my-accout/user.controller.ts
+++ b/gravitee-apim-console-webui/src/user/my-accout/user.controller.ts
@@ -64,10 +64,14 @@ class UserController {
   }
 
   save() {
-    this.UserService.save(this.user).then(() => {
-      this.NotificationService.show('User has been updated successfully');
-      this.onSaved();
-    });
+    this.UserService.save(this.user)
+      .then(() => {
+        this.NotificationService.show('User has been updated successfully');
+        this.onSaved();
+      })
+      .catch((error) => {
+        this.NotificationService.showError(error?.data ? error : { data: 'Failed to update user' });
+      });
   }
 
   displayDangerZone(): boolean {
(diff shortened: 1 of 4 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/exceptions/UserCannotBeCreatedException.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public String getMessage() {
+        return "User cannot be created.";
+    }
+
+    @Override
+    public String getTechnicalCode() {
+        return "user.cannot.be.created";
```
jourdiw: I think we usually have something like an object here. 

```suggestion
        return "user.invalid.creation";
```

jourdiw (commented):
WDYT of using an InvalidUserException in the BE that takes in a message to show? 

I'm wondering if it's necessary to have two separate exceptions 🤔

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/exceptions/UserCannotBeCreatedException.java
vikrantgravitee: @jourdiw 
I added user.invalid to keep generic and created 1 exception class having both cases (it can be modified in future for multiple scenarios).

jourdiw (approved):

PR #14436, opened 2025-12-09
feat: unify app type descriptions from Figma across apps

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-12096

## Description

Implement the Figma app type description in the type selection field 

All application type texts are compile-time constants, so Angular i18n can extract and handle them properly.
We also keep a backend fallback for unknown types or missing translations.

Also, modify the Console to fetch the value from the backend, and in the Old portal, update the eng version as well

## Additional context

Next Gen Portal: 
Zrzut ekranu 2025-12-9 o 16 09 28

Old Portal: 
Zrzut ekranu 2025-12-9 o 16 07 10

Console: 
Zrzut ekranu 2025-12-9 o 16 25 18

### DIFF
--- a/gravitee-apim-portal-webui-next/src/pipe/application-type-translate.pipe.ts
+++ b/gravitee-apim-portal-webui-next/src/pipe/application-type-translate.pipe.ts
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) 2025 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


gravitee-apim-portal-webui-next/src/pipe/application-type-translate.pipe.ts
```diff
@@ -0,0 +1,77 @@
...
+})
+export class ApplicationTypeTranslatePipe implements PipeTransform {
+  transform(type: ApplicationType | null | undefined, field: ApplicationTypeField): string {
+    if (!type?.id) {
+      return '';
+    }
+
+    const backendValue = field === 'name' ? type.name || type.id : type.description || '';
+
+    if (typeof $localize === 'undefined') {
```
jourdiw: Why undefined as a string here?

jourdiw (approved):

gravitee-apim-portal-webui-next/src/pipe/application-type-translate.pipe.ts
JedrzejJanasiak: good point, fixed

PR #14443, opened 2025-12-09
feat(console): Add a Link and Folder item to a folder nav item

aqibmohammadkhan:
## Issue

https://gravitee.atlassian.net/browse/APIM-11766

## Description

- Added the ability to create a link and folder directly from the menu, available only when the navigation item is of type folder.
- If the item is not a folder, the Add link and folder option is hidden.
- When the Add action is triggered, the existing “create” dialog is opened; cancelling closes it without changes, and confirming creates a new link and folder under the same folder where the action was initiated.
- added test cases.

https://github.com/user-attachments/assets/7d460f91-9e2d-4ff5-8e72-a59ca3d2d711

### DIFF
--- a/gravitee-apim-console-webui/src/portal/components/tree-component/tree-node.component.spec.ts
+++ b/gravitee-apim-console-webui/src/portal/components/tree-component/tree-node.component.spec.ts
@@ -16,18 +16,27 @@
 
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
-import { By } from '@angular/platform-browser';
+import { By, ɵSharedStylesHost as SharedStylesHost } from '@angular/platform-browser';
 import { MatIconTestingModule } from '@angular/material/icon/testing';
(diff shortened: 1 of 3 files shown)


gravitee-apim-console-webui/src/portal/components/tree-component/tree-node.component.spec.ts
```diff
@@ -45,11 +54,27 @@ describe('TreeNodeComponent', () => {
             hasAnyMatching: jest.fn().mockReturnValue(true),
           },
         },
+        // Prevent Angular from injecting style tags (which jsdom can't parse due to `@layer`).
```
jourdiw, line 57: I think we have this need in a lot of tests. What do you think of adding it to the GioTestingModule?

jourdiw (approved):

gravitee-apim-console-webui/src/portal/components/tree-component/tree-node.component.spec.ts
aqibmohammadkhan: let me do a separate PR for it. that way this could be tested now.

PR #14472, opened 2025-12-10
fix: update ownership of nginx files in console web ui Dockerfile for IPv4 only

jdesnoes:
## Issue

This issue concerns the Console Web UI image.

Currently, enabling IPv4-only mode via the `IPV4_ONLY=true` environment variable does not work.  
The problem comes from incorrect permissions on the Nginx configuration files: they are owned by root, preventing the nginx user from executing the command
`mv /etc/nginx/conf.d/default.no-ipv6.conf /etc/nginx/conf.d/default.conf`
inside the `check_ip_config.sh` script.  
As a result, the IPv6 configuration cannot be replaced.

This regression seems to have appeared in version 4.9, following this commit:
https://github.com/gravitee-io/gravitee-api-management/commit/4a2a84ea75c7c2c43fc84079ffd3deca4e2c6dc8

## Description

This fix aligns the file ownership and permissions with those applied in the Portal Web UI image, ensuring the script can modify the Nginx configuration as expected.

### DIFF
--- a/gravitee-apim-console-webui/docker/Dockerfile
+++ b/gravitee-apim-console-webui/docker/Dockerfile
@@ -26,7 +26,9 @@ ADD docker/config/constants.json /usr/share/nginx/html/constants.json
 ADD docker/config/default.conf /etc/nginx/conf.d/default.conf
 COPY docker/config/default.no-ipv6.conf /etc/nginx/conf.d/default.no-ipv6.conf
 
-RUN chmod -R g=u /usr/share/nginx/ /etc/nginx/
+RUN chown -R nginx:root /usr/share/nginx \
+    && chown -R nginx:root /etc/nginx \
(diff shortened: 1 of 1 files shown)


jgiovaresco:
@jdesnoes Thanks for your contribution.
It will need some additional checks on our end to ensure OpenShift compatibility is still ok.

jdesnoes:
> @jdesnoes Thanks for your contribution. It will need some additional checks on our end to ensure OpenShift compatibility is still ok.

Hi @jgiovaresco 
Any updates on the review of this PR?
The fix is similar to what was done on the portal, so the impact should be minimal.
Thanks!

jgiovaresco:
@jdesnoes I will close this PR has we have to fix older version. I've made you co author in https://github.com/gravitee-io/gravitee-api-management/pull/15159 
Thanks again for your contribution 🫡 and sorry for the delay

PR #14511, opened 2025-12-11
fix(console): Edit GMD page Save button should be primary

aqibmohammadkhan:
## Issue

https://gravitee.atlassian.net/browse/APIM-11761

## Description

- changed mat-stroked to mat-flat and color = primary for save button while gmd page edit

https://github.com/user-attachments/assets/9b16040f-7ac3-4e88-8cb7-cf11dcfce72e


jourdiw (approved):
🚀

PR #14514, opened 2025-12-11
refactor(console): move layout from global styling to route root styling

jourdiw:
## Issue

https://gravitee.atlassian.net/browse/APIM-11813

## Description

Create layout mixins to be used throughout the NG Portal instead of global settings

### DIFF
--- a/gravitee-apim-portal-webui-next/src/scss/layout.scss
+++ b/gravitee-apim-portal-webui-next/src/scss/layout.scss
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 The Gravitee team (http://gravitee.io)
+ * Copyright (C) 2025 The Gravitee team (http://gravitee.io)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


gravitee-apim-portal-webui-next/src/scss/layout.scss
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@mixin full-width-container {
+  @include -container(100%);
+}
+
+@mixin large-container {
+  @include -container(1280px);
```
kovaren, line 21: Shouldn't it be 1400px?

kovaren (changes requested):

gravitee-apim-portal-webui-next/src/scss/layout.scss
jourdiw: No this is for the legacy views that have a max width of 1280

jourdiw: Renamed to have "-legacy" in the name 👍

kovaren (approved):

PR #14518, opened 2025-12-11
fix(console): display group roles immediately after adding to user

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-12276

## Description

Roles were not visible in the groups table until page refresh. Now roles are immediately displayed after adding a group

### Before Fix
https://github.com/user-attachments/assets/03761f72-b557-47ba-93cd-cb9ef68cd8a2

### After Fix
https://github.com/user-attachments/assets/82fb8e0a-a264-4298-8c14-baf587abf47a


ytvnr (approved):
it looks good to me

PR #14544, opened 2025-12-15
fix: remove null values from CRD yaml file

kamiiiel:
see https://gravitee.atlassian.net/browse/GKO-1949


a-cordier (commented):
Maybe we should update `ApiDefinitionResourceTest` accordingly ?

PR #14549, opened 2025-12-15
Update `Interval` handling and schema in analytics

carlos-andres-osorio:
## Issue

https://gravitee.atlassian.net/browse/GKO-1955

## Description

- Updated OpenAPI analytics schema to support `Interval` representation as either a string or an integer.
- Implemented custom serializer and deserializer for the `Interval` class to ensure proper parsing and handling.

Supported string interval suffixes are:
- `s`: seconds
- `m`: minutes
- `h`: hours

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/model/analytics/engine/CustomInterval.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/model/analytics/engine/CustomInterval.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.management.v2.rest.model.analytics.engine;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.time.Duration;
+
+/**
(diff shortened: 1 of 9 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/model/analytics/engine/CustomInterval.java
```diff
@@ -0,0 +1,60 @@
...
+        if (duration == null || duration.isEmpty()) {
+            return Duration.ZERO;
+        }
+
+        duration = duration.toUpperCase();
+        if (duration.endsWith("D")) {
+            return Duration.parse("P" + duration);
+        }
+        return Duration.parse("PT" + duration);
+    }
```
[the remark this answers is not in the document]
carlos-andres-osorio: This case is handled in the deserializer

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-analytics.yaml
```diff
@@ -1191,9 +1191,8 @@ components:
         Interval:
           description: |
             A fixed time interval for time series analytics queries.
-            Intervals are be expressed in milliseconds.
-          type: integer
-          format: int64
+            Intervals are be expressed in milliseconds or duration strings (e.g., 10s, 1m, 5h).
+          type: interval
```
a-cordier: is interval an actual type in open api specs ? or should we define it with a oneOf ?

a-cordier: so it appears not if you pass the open api file in that linter

https://pb33f.io/doctor/?view=explore&node=root

(score was 96%) before the PR

otherwise I tested it and it works 👍 

I think we can dig up the commit were we disabled the oneOf and restore the definition type for interval.

a-cordier (changes requested):
(see open api comment)

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-analytics.yaml
carlos-andres-osorio: Added the initial `Interval` definition and overrode the class that handles deserialization from the generated one to the one I implemented.

I could not have two classes `Interval` in the same package so my class was renamed to `CustomInterval`

a-cordier (approved):
Great!

PR #14562, opened 2025-12-16
fix(security): prevent info disclosure via freemarker

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-11971

## Description

We want to make sure that sensitive information is not accessible from freemarker templates.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/domain_service/documentation/FreemarkerTemplateResolverTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/domain_service/documentation/FreemarkerTemplateResolverTest.java
@@ -15,7 +15,7 @@
  */
 package io.gravitee.apim.infra.domain_service.documentation;
 
-import static org.assertj.core.api.Assertions.*;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.catchThrowable;
 
 import io.gravitee.apim.core.api.model.Api;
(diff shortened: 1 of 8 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/domain_service/documentation/FreemarkerTemplateResolverTest.java
```diff
@@ -153,9 +153,35 @@ public static Stream<Arguments> provideParameters() {
             Arguments.of("picture", "a lovely picture"),
             Arguments.of("state", "STARTED"),
             Arguments.of("visibility", "PUBLIC"),
-            Arguments.of("tags[0]", "first tag"),
+            Arguments.of("tags?first", "first tag"),
```
michel-barret, line 156: I’m a bit affraid. `tags[0]` don’t work anymore? If yes it’s a blocking change

michel-barret (commented):
You apply it only on master, I imagine it’s should be in at least 4.10

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/domain_service/documentation/FreemarkerTemplateResolverTest.java
thomasballivet: I think it was working before because of the default freemarker object wrapper. `tags[0]` is not working anymore but it sounds normal to me as tags is a `Set` and has no index. Shouldn't we use a `List` if we expect an index?

thomasballivet: I managed to avoid breaking changes

thomasballivet:
Reopened as PR #14604

PR #14406, opened 2025-12-09
feat: disable NODE alert triggers in cloud-hosted mode

gcusnieux:
## Issue

https://gravitee.atlassian.net/browse/CJ-3195

## Description

A small description of what you did in that PR.

PR #8954, opened 2024-09-11
feat: update hide gateways view for cloud flag

josephlomasney:
## Issue

https://gravitee.atlassian.net/browse/CJ-1835

## Description

Rather than hiding this based on the installation type being multi-tenant which was introduced here https://github.com/gravitee-io/gravitee-api-management/pull/8876 we have decided it is better to have a separate cloud flag which controls this being hidden as there may be use cases where APIM is run in multi-tenant mode but would want the gateway view still available

Also made change to routing, removing uneccessary and problematic canActivateChild (fixes cypress tests)

Cloud disabled:
Screenshot 2024-09-11 at 15 32 50

Cloud enabled:
Screenshot 2024-09-11 at 15 34 36

---

📚&nbsp;&nbsp;View the storybook of this branch [here](https://612657caa8e859003a8a6430-ryyxjimniu.chromatic.com)

PR #14564, opened 2025-12-16
fix: show gateways view when cloud hosted

josephlomasney:
## Issue

https://gravitee.atlassian.net/browse/CJ-3968

## Description

Product have decided that they want the gateways view to be visible in cloud hosted instances. 

Therefore this undoes the gateways view hiding added here: https://github.com/gravitee-io/gravitee-api-management/pull/8954

Cloud hosted flag is left as that is now used elsewhere (https://github.com/gravitee-io/gravitee-api-management/pull/14406)


jgiovaresco (approved):
LGTM

PR #14658, opened 2025-12-18
Apim 12282 portal filtering problem

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-12282

## Description
Extract response-time range patterns from query strings and convert them to explicit Elasticsearch range queries instead of using query_string, which doesn't handle numeric ranges correctly.

### After fix
https://github.com/user-attachments/assets/e241b995-9826-4dee-8f86-0449e046d6ef


phiz71 (changes requested):
The changes have to be made on portal side. Check this PR: https://github.com/gravitee-io/gravitee-api-management/pull/14762

PR #14694, opened 2025-12-19
fix(security): media validation before upload

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-11973

## Description

This PR introduces a file validation service to ensure uploaded files match expected MIME types. It leverages Apache Tika for content-based type detection and Apache PDFBox to analyze PDF complexity.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/MediaValidationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/MediaValidationServiceImpl.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.service.impl;
+
+import io.gravitee.rest.api.model.MediaEntity;
+import io.gravitee.rest.api.service.MediaValidationService;
+import io.gravitee.rest.api.service.exceptions.UploadUnauthorized;
+import java.io.IOException;
+import java.util.Set;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.pdfbox.Loader;
+import org.apache.pdfbox.pdmodel.PDDocument;
+import org.apache.pdfbox.pdmodel.PDDocumentCatalog;
+import org.apache.pdfbox.pdmodel.common.PDDestinationOrAction;
+import org.apache.pdfbox.pdmodel.interactive.action.PDActionImportData;
+import org.apache.pdfbox.pdmodel.interactive.action.PDActionJavaScript;
+import org.apache.pdfbox.pdmodel.interactive.action.PDActionLaunch;
+import org.apache.pdfbox.pdmodel.interactive.action.PDActionSubmitForm;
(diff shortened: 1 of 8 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/MediaValidationServiceImpl.java
```diff
@@ -0,0 +1,123 @@
...
+
+            PDDestinationOrAction openAction = catalog.getOpenAction();
+            return !(
+                openAction instanceof PDActionJavaScript ||
+                openAction instanceof PDActionLaunch ||
+                openAction instanceof PDActionImportData ||
+                openAction instanceof PDActionSubmitForm
+            );
+        } catch (IOException e) {
+            return false;
```
phiz71, line 134: It's maybe worth adding a log here, at least debug or warn.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/MediaValidationServiceImpl.java
```diff
@@ -0,0 +1,123 @@
...
+            return true;
+        }
+
+        return FORBIDDEN_TYPES.contains(mediaType.toLowerCase());
+    }
+
+    private boolean isAcceptablePDF(byte[] pdf) {
+        try (PDDocument document = Loader.loadPDF(pdf)) {
+            if (document.getNumberOfPages() == 0) {
+                return false;
```
phiz71, line 107: It's maybe worth adding a debug log here.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/MediaValidationServiceImpl.java
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            if (document.isEncrypted()) {
+                return false;
```
phiz71, line 112: It's maybe worth adding a debug log here.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/MediaValidationServiceImpl.java
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            PDDocumentCatalog catalog = document.getDocumentCatalog();
+            if (catalog.getNames() != null && catalog.getNames().getJavaScript() != null) {
+                return false;
```
phiz71, line 118: It's maybe worth adding a debug log here.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/MediaValidationServiceImpl.java
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            PDDestinationOrAction openAction = catalog.getOpenAction();
+            return !(
+                openAction instanceof PDActionJavaScript ||
+                openAction instanceof PDActionLaunch ||
+                openAction instanceof PDActionImportData ||
+                openAction instanceof PDActionSubmitForm
+            );
```
phiz71: It's maybe worth adding a debug log here.

phiz71 (commented):
LGTM.
I would add more debug logs, especially in the `isAcceptablePDF` method.
Another point, could you check the size of the dependencies you have added, and more over, if their dependencies are not conflicting with ours (like Jackson or others)

thomasballivet:
`pdfbox` was conflicting with `swagger-parser` on `commons-logging` so I forced 1.3.5 in `dependencyManagement`.
This was the only conflict related to pdfbox and tika.
Tika is very light, less than 1M and doesn't need dependencies.
pdfbox needs some dependencies and is around 3.7Mo with its dependencies.

I also added some logs

phiz71 (approved):

phiz71:
Good Job 👍🏻

PR #14696, opened 2025-12-19
chore: polish analytics renderer implementation 2/2

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/GKO-1953
## Description

- [x] →improve time-series context (x axe) 
- [x] →change color default order (blue > green >  yellow > orange > red) to avoid having status 200 in red and 500 in green
- [x] →add bar chart
- [x] →add url injection instead of input

   - What exactly is meant by this? Read baseURL from URL query params? Use Angular dependency injection? Something else?
- [ ] →ask backend to improve experience when no data available (facet and measure)
   - Will not do for now. See [thread](https://graviteeio.slack.com/archives/C06F602EWBD/p1766066983877439?thread_ts=1766052007.879799&cid=C06F602EWBD) in Slack for more context.

## Additional context

I'm not sure if the intent is to replace any of the charts in the current analytics/dashboard-v4 view with the bar chart, but I did replace Response Statuses so it's easy to see what it looks like. Please advise on which charts should be bar charts or if you just want it to be an option. 

Screenshot 2025-12-19 at 9 59 08 AM

### DIFF
--- a/gravitee-apim-console-webui/src/management/analytics/analytics-viewer/analytics-viewer.component.html
+++ b/gravitee-apim-console-webui/src/management/analytics/analytics-viewer/analytics-viewer.component.html
@@ -22,4 +22,4 @@
   </div>
 </div>
 
-<gd-dashboard [widgetConfigs]="widgets" [baseURL]="baseURL" [filters]="filters"></gd-dashboard>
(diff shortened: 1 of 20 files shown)


paulatulis:
Closing and will create a separate PR for each of the items.

PR #14741, opened 2026-01-05
fix(security): add security headers

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-11975

## Description

This PR strengthens our security posture by implementing a set of standard HTTP security headers. Through a unified SecureHeadersConfigurer across all Spring Security configurations, this change provides defense against common web vulnerabilities like XSS, clickjacking...

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-security/src/test/java/io/gravitee/rest/api/management/security/config/BasicSecurityConfigurerAdapterTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-security/src/test/java/io/gravitee/rest/api/management/security/config/BasicSecurityConfigurerAdapterTest.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.management.security.config;
+
+import io.gravitee.rest.api.security.config.AbstractSecurityConfigurerAdapterTest;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.servlet.config.annotation.EnableWebMvc;
+
+@ContextConfiguration(
+    classes = {
+        BasicSecurityConfigurerAdapter.class,
+        AbstractSecurityConfigurerAdapterTest.TestConfig.class,
+        BasicSecurityConfigurerAdapterTest.DummyController.class,
+    }
+)
+class BasicSecurityConfigurerAdapterTest extends AbstractSecurityConfigurerAdapterTest {
+
+    @Override
+    protected String getPath() {
+        return "/openapi.json";
+    }
+
+    @Configuration
+    @EnableWebMvc
+    @RestController
+    static class DummyController {
+
+        @GetMapping(path = { "/openapi.json" })
+        public void handle() {}
+    }
+}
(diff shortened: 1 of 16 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-security/src/test/java/io/gravitee/rest/api/management/security/config/BasicSecurityConfigurerAdapterTest.java
```diff
@@ -0,0 +1,193 @@
...
+    void setUp() {
+        mockMvc = MockMvcBuilders.webAppContextSetup(wac).apply(springSecurity()).build();
+    }
+
+    @Test
+    void should_configure_secure_headers() throws Exception {
+        mockMvc
+            .perform(get("/openapi.json").secure(true))
+            .andExpect(header().string("X-Content-Type-Options", "nosniff"))
+            .andExpect(header().string("X-Frame-Options", "DENY"))
```
jgiovaresco: In https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-console-webui/docker/config/default.conf#L6 we use a different value. 
I would use the same here 
```suggestion
            .andExpect(header().string("X-Frame-Options", "SAMEORIGIN"))
```

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-security/src/test/java/io/gravitee/rest/api/management/security/config/BasicSecurityConfigurerAdapterTest.java
```diff
@@ -0,0 +1,193 @@
...
+            .andExpect(header().string("Referrer-Policy", "strict-origin-when-cross-origin"))
+            .andExpect(header().string("Permissions-Policy", "camera=(), microphone=(), geolocation=()"))
+            .andExpect(header().string("Strict-Transport-Security", "max-age=31536000 ; includeSubDomains"))
```
jgiovaresco: I'm worried about that one. 
>The Strict-Transport-Security (HSTS) header is a powerful security instruction that tells a browser to only communicate with your website using HTTPS.

If we "hardcode" this it means we won't be able to use http access with localhost 🤔 

Also Gemini raises an important warning

> Important Warning: Don't Lock Yourself Out
> HSTS is very hard to undo. If you set a max-age of one year and your SSL certificate expires or you need to move back to HTTP for some reason, users who have visited your site will be blocked until the year is up or you get a new certificate.
> 
> Best Practice for Implementation:
> 1. Start with a short time: max-age=300 (5 minutes).
> 2. Test everything (including subdomains).
> 3. Increase to a week, then a month.
> 4. Finally, set it to one year (31536000) once you are certain your HTTPS setup is stable.

I'm not sure we should set this header here but let our users to set it using through a proxy 🤔

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-security/src/test/java/io/gravitee/rest/api/management/security/config/BasicSecurityConfigurerAdapterTest.java
```diff
@@ -0,0 +1,193 @@
...
+            .andExpect(header().string("Content-Security-Policy", "default-src 'self'; frame-ancestors 'none'; object-src 'none'"));
```
jgiovaresco: In https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-console-webui/docker/config/default.conf#L7 we use a different value.
I would use the same here
```suggestion
            .andExpect(header().string("Content-Security-Policy", "frame-ancestors 'self';"));
```

jgiovaresco (commented):
Can you also take a look at the nginx configuration file for the Console and the Portal to ensure the same headers are also set.
I noticed inconsistencies between all the files and saw that the header Referrer-Policy and Permissions-Policy are not set

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-security/src/test/java/io/gravitee/rest/api/management/security/config/BasicSecurityConfigurerAdapterTest.java
michel-barret: I’m agree with you it must not be automatically set and it should be manage by the TLS terminaison (API management or a reverse proxy).

michel-barret (changes requested):

thomasballivet:
I adapted to fit remarks on hsts and to use gravitee.yml to set default values.

michel-barret (approved):
LGTM

PR #14756, opened 2026-01-06
fix: hide the default resource filtering policy for a new plan when API is MCP

schehade-gravitee:


gravitee-apim-console-webui/src/management/api/component/plan/3-restriction-step/plan-edit-restriction-step.component.ts
```diff
@@ -95,4 +99,8 @@ export class PlanEditRestrictionStepComponent implements OnInit, OnDestroy {
     this.unsubscribe$.next(true);
     this.unsubscribe$.unsubscribe();
   }
+
+  get isMcpApi(): boolean {
```
ThibaudAV: can you change it for something link `isAIApi` ? because we have the same issue for LLM Proxy API

schehade-gravitee: done

ThibaudAV (approved):

PR #14773, opened 2026-01-07
fix: analytics for not found errors

nive1980:
## Issue

https://gravitee.atlassian.net/browse/APIM-12149

## Description

Fixed the Not found api issue in V4 NotFoundProcessor.

Also fixed the Log index for the same.

Since the analytics were not appearing for 404 errors on UI, needed to add a hack to add the apiId=1(the id for not found) to list of api ids while filtering

https://github.com/user-attachments/assets/0853ce44-3eca-4f95-979d-2fe886e43dae

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/reporter/ReporterProcessor.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/reporter/ReporterProcessor.java
@@ -58,7 +58,9 @@ public class ReporterProcessor implements Processor {
             Metrics metrics = ctx.metrics();
             if (metrics != null && metrics.isEnabled()) {
                 metrics.setRequestEnded(true);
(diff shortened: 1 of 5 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/reporter/ReporterProcessor.java
```diff
@@ -94,6 +96,13 @@ public Completable execute(final HttpExecutionContextInternal ctx) {
                 } else {
                     // No api found report only metrics
                     reporterService.report(metrics);
+                    Log log = metrics.getLog();
+                    if (log != null) {
+                        log.setApiId(metrics.getApiId());
+                        log.setApiName(metrics.getApiName());
+                        log.setRequestEnded(metrics.isRequestEnded());
+                        reporterService.report(log);
+                    }
```
phiz71, line 105: too dangerous, in case the gateway is spammed with 404 error.
It could have a huge impact on the gateway memory

phiz71 (changes requested):

PR #14778, opened 2026-01-07
feat(dashboard-lib): change color default order

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/GKO-1953
## Description

change color default order (blue > green >  yellow > orange > red) to avoid having status 200 in red and 500 in green

Screenshot 2026-01-07 at 11 30 54 AM

## Additional context
I believe @samirtechlab said the rounded edges will go away, so take the screenshot with a grain of salt.


a-cordier (approved):

SamHardi (approved):
LGTM 😎

PR #14789, opened 2026-01-08
feat: enrich logging thanks to NodeAwareLogger

ytvnr:
## Issue

https://gravitee.atlassian.net/browse/APIM-12306

## Description

This PR introduces the integration of the new Gravitee Node Logging system. While most of the commits consist of migrating existing log factory calls, the core changes focus on standardizing how we handle logging across the application to improve observability.

### Key Changes
- Enriched Logs: Updated logback.xml configurations to include [nodeId:] in the log pattern. This ensures better traceability in distributed environments.
- New Logging Patterns: Developers should now use @CustomLog (Lombok) or the NodeLoggerFactory to obtain logger instances. This ensures that all logs are properly intercepted and enriched by the Gravitee Node framework.
- Guardrails with ArchUnit: Added a new ArchUnit rule to enforce the use of the correct logger factory. Any usage of standard SLF4J LoggerFactory instead of NodeLoggerFactory will now cause test failures. Note: This is a first iteration and the enforcement mechanism might be further refined in future updates.

## How to use
Instead of:
```java
@Slf4j
public class MyClass { ... }
// OR
private static final Logger proc = LoggerFactory.getLogger(MyClass.class);
```

You should now use:
```java
@CustomLog
public class MyClass { ... }
// OR
private static final Logger proc = NodeLoggerFactory.getLogger(MyClass.class);
```


ytvnr:
> Fails
> 🚫	Gravitee node plugins defined in Helm values.yaml needs to be consistent with the version defined in the pom.xml.
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [e8cb652](https://github.com/gravitee-io/gravitee-api-management/commit/e8cb652fddb122603bd939dd56664c7e82132b9b)

@gravitee-io/platform team, I added a commit here https://github.com/gravitee-io/gravitee-api-management/pull/14789/commits/712fc0f06fbebbd5a2bd129d462cb9e2a8e400ee to comply with the current version of gravitee-node used in apim.
It will allow to have the new "rich" logging also for those plugins.

helm/values.yaml
```diff
@@ -473,8 +473,8 @@ cloud:
 
 cluster:
   plugins:
-    - https://download.gravitee.io/plugins/node-cache/gravitee-node-cache-plugin-hazelcast/gravitee-node-cache-plugin-hazelcast-7.18.3.zip
-    - https://download.gravitee.io/plugins/node-cluster/gravitee-node-cluster-plugin-hazelcast/gravitee-node-cluster-plugin-hazelcast-7.18.3.zip
+    - https://download.gravitee.io/plugins/node-cache/gravitee-node-cache-plugin-hazelcast/gravitee-node-cache-plugin-hazelcast-8.0.0-alpha.2.zip
```
ThibaudAV, line 476: do we need to add changelog here : https://github.com/gravitee-io/gravitee-api-management/blob/8ef5cebf1e5989f0312cc3fa203e8a8334f366f4/helm/Chart.yaml#L18 ?

ThibaudAV (approved):

helm/values.yaml
ytvnr: we do not neet to do it as I only modify the values.yml and not the chart.yaml

PR #14804, opened 2026-01-09
fix(automation): should not configure secure headers

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-11975

## Description

This PR follows #14741. As we don't need secure headers for automation.


gravitee-apim-rest-api/gravitee-apim-rest-api-security/src/main/java/io/gravitee/rest/api/security/config/SecureHeadersConfigurer.java
```diff
@@ -68,10 +78,48 @@ private void referrerPolicy(HttpSecurity security, ConfigurableEnvironment envir
         security.headers().referrerPolicy(policy);
```
benoitgravitee, line 78:
```suggestion
var policy = Arrays.stream(ReferrerPolicyHeaderWriter.ReferrerPolicy.values())
.filter(p -> p.getPolicy().equalsIgnoreCase(policyString))
.findFirst();
if (policy.isPresent()) {
  security.headers().referrerPolicy(policy.get());
}
```

benoitgravitee (changes requested):
Nice refactoring. I'm asking for a change as merging this will not solve the issue. Mgmt API won't start with Helm charts if referer policy is unset.
I suggested a fix, but it is up to you of course

michel-barret (approved):
LGTM

benoitgravitee (approved):

PR #14822, opened 2026-01-13
fix: add bulk retrieval methods to fix N+1 problem

dev-marek:
## Issue

https://gravitee.atlassian.net/browse/APIM-12281

## Description

Improve response time of plans endpoint.


gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcFlowRepository.java
phiz71, line 1: What is that 😲?

PR #14825, opened 2026-01-13
Introduce `ApiTypePreProcessor` for analytics filtering

carlos-andres-osorio:
## Issue

https://gravitee.atlassian.net/browse/gko-1973

## Description

Added support for filtering analytics requests by API type. Filter supports the `EQ` and `IN` operators and takes one or more of the following values:
- `HTTP_PROXY`
- `MESSAGE`
- `KAFKA`
- `LLM`
- `MCP`

Example measures request with a filter:
```json
{
  "timeRange": {
    "from": "2026-01-08T21:24:00Z",
    "to": "2026-01-08T21:30:00Z"
  },
  "filters": [
    {
      "name": "API_NAME",
      "operator": "EQ",
      "value": "HTTP_PROXY"
    }
  ],
  "metrics": [
    {
      "name": "HTTP_REQUESTS",
      "measures": [
        "COUNT"
      ]
    }
  ]
}
```

Note: Filter name `API_NAME` was chosen because it's consistent with the naming used in other analytics endpoints.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/analytics/engine/api/query/Filter.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/analytics/engine/api/query/Filter.java
@@ -22,6 +22,7 @@ package io.gravitee.repository.analytics.engine.api.query;
 public record Filter(Filter.Name name, Operator operator, Object value) {
     public enum Name {
         API,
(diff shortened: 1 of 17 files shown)


a-cordier (changes requested):
I don't think the API_NAME filter will ever be used from a client perspective, how about we infer the api type from the metrics passed in the request ? we should be able to do so by finding back the type from the metrics.

PR #14837, opened 2026-01-14
fix: fallback to default roles for INTEGRATION scope

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-12391

## Description
When a Group Owner (non-admin) attempts to add a member to a group, an error is displayed in the UI.

Root Cause:
For the INTEGRATION role scope, no roles were defined at the group level. As a result, the system failed to assign a role during member addition.

Fix Implemented:
The existing logic has been updated to apply default roles when the group does not define any roles for the requested role scope (INTEGRATION). Since default roles already exist for this scenario, the code now correctly falls back to using them.

The fix has been implemented, but confirmation or feedback from the team is still awaited.


phiz71 (approved):
Your fix seems good. But I'm wondering if we should add the possibility to define integration (and cluster) default roles like for API and application:
Capture d’écran 2026-01-21 à 14 14 19

Anyway, that would be another PR.

phiz71:
@pragsty you have applied the 'apply on 4.9.x' label. But your PR is targeting the 4.9.x branch.
So you can remove this label.
However, do you know if this PR should be applied on 4.8.x or 4.7.x?

pragsty:
> @pragsty you have applied the 'apply on 4.9.x' label. But your PR is targeting the 4.9.x branch. So you can remove this label. However, do you know if this PR should be applied on 4.8.x or 4.7.x?

Updated the label and no the Issue didn't get reproduced on 4.6.x till 4.8.x

PR #14900, opened 2026-01-23
feat: allow api in api product

gaurav-gravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-12363

## Description

Toggle “Allow API Products” added to V4 HTTP Proxy APIs

Default:

1. OFF for existing APIs
2. ON for new APIs

**Key Items:**
- APIs with toggle OFF cannot be added to products
- Toggle disabled if API already belongs to a product
- API Product search only returns APIs with toggle ON

- Add `allowInApiProduct` field to ApiEntity and V4 definition
- Index new field in Elasticsearch and expose it in API Search criteria
- Add validation to prevent disabling flag if API is used in a Product
- Update Management API import logic to respect product associations
- Add UI toggle in General Info section for V4 Proxy APIs

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/api/ApiEntity.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/api/ApiEntity.java
@@ -18,6 +18,7 @@ package io.gravitee.rest.api.model.api;
 import com.fasterxml.jackson.annotation.JsonFilter;
 import com.fasterxml.jackson.annotation.JsonGetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonSetter;
 import io.gravitee.common.component.Lifecycle;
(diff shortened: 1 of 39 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/exceptionMapper/domain/ApisNotLinkableExceptionMapper.java
```diff
@@ -0,0 +1,41 @@
...
+
+import io.gravitee.apim.core.api_product.exception.ApisNotLinkableException;
+import io.gravitee.rest.api.management.v2.rest.model.Error;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.ext.ExceptionMapper;
+import jakarta.ws.rs.ext.Provider;
+
+@Provider
+public class ApisNotLinkableExceptionMapper implements ExceptionMapper<ApisNotLinkableException> {
```
gaurav-gravitee: This ensures the API returns a stable error code (apis_not_linkable) for the UI to consume, without risking a massive refactor of the entire exception hierarchy.

ValidationDomainExceptionMapper:

- Emits only a generic error
- Or has no hook for custom technical codes

Other option I had: Update ValidationDomainExceptionMapper to use getTechnicalCode()

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/api/ApiEntity.java
```diff
@@ -237,6 +238,24 @@ public class ApiEntity implements GenericApiEntity {
...
+        example = "false"
+    )
+    private Boolean allowInApiProduct;
+
+    /**
+     * Custom getter that returns null for non-V4-Proxy APIs to ensure the key is omitted from JSON.
+     * Jackson's @JsonInclude(NON_NULL) will skip serializing this field when null.
+     * This is the V2 ApiEntity class, so it will always return null (V2 APIs don't support allowInApiProduct).
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
```
gaurav-gravitee, line 247: Custom getter that returns null for non-V4-Proxy APIs to ensure the key is omitted from JSON.
Jackson's @JsonInclude(NON_NULL) will skip serializing this field when null.
This is the V2 ApiEntity class, so it will always return null (V2 APIs don't support allowInApiProduct).

gaurav-gravitee:
This PR is cancelled because another pull request: https://github.com/gravitee-io/gravitee-api-management/pull/14962 is created to merge this feature code.

PR #14903, opened 2026-01-23
feat(mcp-tool-server): support outputSchema & annotations

jgiovaresco:
## Issue

https://gravitee.atlassian.net/browse/APIM-12478

## Additional context

Linked to https://github.com/gravitee-io/gravitee-entrypoint-mcp/pull/15

### DIFF
--- a/gravitee-apim-console-webui/src/entities/entrypoint/mcp.ts
+++ b/gravitee-apim-console-webui/src/entities/entrypoint/mcp.ts
@@ -23,10 +23,20 @@ export interface MCPTool {
   gatewayMapping: MCPToolGatewayMapping;
 }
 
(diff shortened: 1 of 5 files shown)


ThibaudAV (approved):
Lol, I had fun doing exactly the same thing on the Java side. 😁 I wonder, and I was talking about it with @michel-barret  
I wonder if we could make a public “openapi to mpc” library. There isn't one in Java. There are only two in TypeScript, one in Go, and one in Python. 🤷‍♂️ 

The idea would be to clean up this code and pass it on to the Java side, but also to do other things with it.
Well, nothing to do with this PR, of course.

michel-barret (approved):

PR #14904, opened 2026-01-23
fix(console): Permissions not working for nav item drop down

aqibmohammadkhan:
## Issue

https://gravitee.atlassian.net/browse/APIM-12393

## Description

https://github.com/user-attachments/assets/0f4587c9-4c1c-4d89-acfa-f42508f7fad3

### DIFF
--- a/gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.ts
+++ b/gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.ts
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { Component, computed, effect, input, output, viewChild } from '@angular/core';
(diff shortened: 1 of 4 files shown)


gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.ts
```diff
@@ -104,6 +107,25 @@ export class FlatTreeComponent {
...
+    return this.permissionService.hasAnyMatching(['environment-documentation-c']);
+  }
+
+  canUpdate(): boolean {
+    return this.permissionService.hasAnyMatching(['environment-documentation-u']);
+  }
+
+  canDelete(): boolean {
+    return this.permissionService.hasAnyMatching(['environment-documentation-d']);
+  }
```
jourdiw: These will all be unchanging values. WDYT about having them declared directly to a variable instead of a function?

jourdiw (approved):
😎

PR #14952, opened 2026-01-27
CI: do a git diff based on common commit

phiz71:
## Description

instead of comparing the HEAD of the current branch with the HEAD of the target branch, we try to compare the HEAD of the current branch with the commit in common between current and target branches.

This way, the git diff contains only the modified files on the current branch.

⚠️ Once merged, we will have to force all PRs to be up to date with target branch so they can be merged.


jhaeyaert (approved):
🤩🤩🤩 💵💵💵

PR #14958, opened 2026-01-28
fix(console): Separator shown when create and either update or delete…

aqibmohammadkhan:
… permission

## Issue

https://gravitee.atlassian.net/browse/APIM-12393

## Description

separator will be shown only when it has any permissions other than create.

Screenshot 2026-01-28 at 13 39 12


dev-marek (approved):
I find this change acceptable.

PR #14959, opened 2026-01-28
fix: Added gateway instance details to api metrics detail

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-12400

## Description
Added gateway instance details to api metrics detail instead making another getInstanceById API call which explicitly requires INSTANCE_READ permission

Result: API owners can now access API logs along with instance details without any error
After Fix: 

https://github.com/user-attachments/assets/8c622334-74bd-4b0b-9aa7-a2fb05456402

Before Fix: 

https://github.com/user-attachments/assets/21931f48-ba51-48b4-a2a2-1e91ecd1aabb

### DIFF
--- a/gravitee-apim-console-webui/src/entities/management-api-v2/analytics/apiMetricsDetailResponse.fixture.ts
+++ b/gravitee-apim-console-webui/src/entities/management-api-v2/analytics/apiMetricsDetailResponse.fixture.ts
@@ -39,6 +39,8 @@ export const fakeApiMetricResponse = (modifier?: Partial<ApiMetricsDetailRespons
       apiKeyMode: 'UNSPECIFIED',
     },
     gateway: 'b504bb7b-8b6e-426f-84bb-7b8b6e626f3f',
+    gatewayHostname: 'hostname.example.com',
+    gatewayIp: 'ip.example',
     uri: '/v4/echo',
     status: 202,
     requestContentLength: 0,
--- a/gravitee-apim-console-webui/src/entities/management-api-v2/analytics/apiMetricsDetailResponse.ts
+++ b/gravitee-apim-console-webui/src/entities/management-api-v2/analytics/apiMetricsDetailResponse.ts
@@ -33,6 +33,8 @@ export interface ApiMetricsDetailResponse {
   plan: BasePlan;
   application: BaseApplication;
   gateway: string;
+  gatewayHostname?: string;
+  gatewayIp?: string;
   uri: string;
   status: number;
   requestContentLength: number;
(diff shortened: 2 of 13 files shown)


phiz71 (changes requested):
I'm not a huge fan of adding a query parameter "apiId" to the /instances route.
This /instances is meant to be accessed at env level to retrieve information about gateways.
We should find another way to retrieve the information we want to display in the API logs. Can't we modify the `/apis/${apiId}/analytics/${requestId}` route so that we return the gateway hostname and ip ?

gravitee-apim-console-webui/src/services-ngx/instance.service.ts
```diff
@@ -62,8 +62,9 @@ export class InstanceService {
     return this.http.get<Instance>(`${this.constants.env.baseURL}/instances/${id}`);
   }
 
-  getByGatewayId(id: string): Observable<BaseInstance> {
-    return this.http.get<Instance>(`${this.constants.env.v2BaseURL}/instances/${id}`);
+  getByGatewayId(id: string, apiId?: string): Observable<BaseInstance> {
+    const params = apiId ? new HttpParams().set('apiId', apiId) : undefined;
+    return this.http.get<Instance>(`${this.constants.env.v2BaseURL}/instances/${id}`, { params });
```
dev-marek: I can't see usage of apiId in controller. Also API_LOG is a global permission, not per-API.

dev-marek (changes requested):

pragsty:
> I'm not a huge fan of adding a query parameter "apiId" to the /instances route. This /instances is meant to be accessed at env level to retrieve information about gateways. We should find another way to retrieve the information we want to display in the API logs. Can't we modify the `/apis/${apiId}/analytics/${requestId}` route so that we return the gateway hostname and ip ?

Let me check if we can modify.

gravitee-apim-console-webui/src/services-ngx/instance.service.ts
pragsty: A filter is executed before that which needs apiId for API_LOG permission
anyways checking for alternate solution

pragsty:
> I'm not a huge fan of adding a query parameter "apiId" to the /instances route. This /instances is meant to be accessed at env level to retrieve information about gateways. We should find another way to retrieve the information we want to display in the API logs. Can't we modify the `/apis/${apiId}/analytics/${requestId}` route so that we return the gateway hostname and ip ?

This is done

phiz71:
it looks better 👍🏻 . Don't you have an openapi file to adapt to add gateway hostname and ip ?

pragsty:
> it looks better 👍🏻 . Don't you have an openapi file to adapt to add gateway hostname and ip ?

Added to openapi.

phiz71 (approved):

dev-marek (approved):
Great job! I see you’ve adapted well to our codebase :) Before merging, please check if this fix should be applied to newer versions. If so, you'll need to tag this PR with `apply-on-version`.

PR #14966, opened 2026-01-28
fix: improve message when unable to find crd member role

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-2165


paulatulis (approved):
LGTM!

PR #14976, opened 2026-01-29
feat: upgrade to vertx 5

remibaptistegio:
## Issue

https://gravitee.atlassian.net/browse/ARCHI-601

## Description

Prepare Vertx 5 upgrade


gravitee-apim-bom/pom.xml
```diff
@@ -33,15 +33,14 @@
...
+            Vertx RxJava3 bindings must match the vertx-core version.
+            Previous note: There was a bug in 4.5.x related to SSE events (https://github.com/vert-x3/vertx-web/issues/2725)
+            but that is resolved in Vert.x 5.x. Use ${vertx.version} from gravitee-bom.
             -->
             <dependency>
                 <groupId>io.vertx</groupId>
                 <artifactId>vertx-rx-java3</artifactId>
-                <version>4.5.1</version>
+                <version>${vertx.version}</version>
             </dependency>
```
phiz71: if it's the same as the one in gravitee-bom, maybe e could simply remove this dependency.
I think the purpose was to fix the version used.

phiz71 (approved):

PR #14991, opened 2026-01-29
Filter logs based on user permissions.

carlos-andres-osorio:
## Issue

https://gravitee.atlassian.net/browse/GKO-2164

## Description

🎯 What Changed
- Moves API filtering logic from ManagementFilterPreProcessor to a new MetricsContextLoader
- Relocates MetricsContext to a dedicated core.metrics package
- Integrates the new loader into all analytics compute use cases and the logs search use case
- Added tests for the request handler

Executed manual tests verifying that roles affect which logs a user can see.


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCaseTest.java
```diff
@@ -119,35 +138,34 @@ void should_map_basic_string_and_array_filters() {
...
             );
 
-            when_searching(request);
-
-            var filtersCaptor = ArgumentCaptor.forClass(SearchLogsFilters.class);
-            verify(connectionLogsCrudService).searchApiConnectionLogs(any(), filtersCaptor.capture(), any(), any());
+            var response = when_searching(request);
 
-            assertThat(filtersCaptor.getValue().apiIds()).isEmpty();
+            assertThat(response.response().pagination().totalCount()).isEqualTo(0);
```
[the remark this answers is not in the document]
carlos-andres-osorio: Assertion is correct. In filters should take a list and not a string

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
```diff
@@ -65,9 +70,15 @@ public record Output(SearchLogsResponse response) {}
...
         var pageable = buildPageable(input.request);
-        var searchFilters = buildFilters(input.request);
-        io.gravitee.rest.api.model.v4.log.SearchLogsResponse<BaseConnectionLog> searchLogResponse =
-            connectionLogsCrudService.searchApiConnectionLogs(executionContext, searchFilters, pageable, List.of(DefinitionVersion.V4));
+
+        var metricsContext = metricsContextLoader.loadApis(new MetricsContext(input.auditInfo));
+
+        var searchFilters = buildFilters(metricsContext, input.request);
+
+        io.gravitee.rest.api.model.v4.log.SearchLogsResponse<BaseConnectionLog> searchLogResponse = searchFilters.apiIds().isEmpty()
```
a-cordier: It's not a common pattern to have the rest model invading the use case, instead there is more of a mapping layer involved at the resource level. I'm even surprised this is not caught by archunit.

a-cordier: this way we can get to

```java
        var searchLogResponse = searchFilters.apiIds().isEmpty()
            ? new SearchLogsResponse<BaseConnectionLog>(0, List.of())
            : connectionLogsCrudService.searchApiConnectionLogs(executionContext, searchFilters, pageable, List.of(DefinitionVersion.V4));
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/metrics/model/MetricsContext.java
```diff
@@ -13,37 +13,43 @@
...
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import lombok.Getter;
 
 /**
  * @author Antoine CORDIER (antoine.cordier at graviteesource.com)
  * @author GraviteeSource Team
  */
 public record MetricsContext(
```
a-cordier: Should we have more of a LogsContext so that we don't clutter this class too much ?

a-cordier (changes requested):
Requested change is mostly about not having the REST model invading the useCase, rest is open questions

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/metrics/model/MetricsContext.java
carlos-andres-osorio: This context was simplified by removing the `Filters` (which is unused anyway) and renaming it to UserContext, which better explains what the context is for.

Another option would be to rename it to `AnalyticsContext` and put it in the analytics package.  Logs and Metrics are both analytics, so this should be OK.  Let me know if it makes more sense, and I can do this.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
carlos-andres-osorio: Fixed

carlos-andres-osorio:
All comments have been addressed, and the PR is ready, @a-cordier.
By the way, I refactored the context, but I'm open to discussing its name or functionality.

a-cordier (approved):
👍 @carlos-andres-osorio Do you think we could squash a bit the history now that it's reviewed ? There's a typo in one of the commits as well.

PR #15001, opened 2026-01-30
Feat/re design/developers portal catalog view

adam-rosowicz:
Link to jira ticket: https://gravitee.atlassian.net/browse/EXT-65

https://github.com/user-attachments/assets/14a2b600-59d7-477b-8ceb-70dcac975233


gravitee-apim-portal-webui-next/src/app/catalog/components/apis-list/apis-list.component.html
```diff
@@ -16,50 +16,107 @@
...
+            [picture]="api.picture"
+            [isEnabledMcpServer]="api.isEnabledMcpServer" />
+        }
+      </div>
+    } @else {
+      <div class="api-list__table">
+        <div class="api-list__table-header">
+          <div class="api-list__table-header__cell api-list__table-header__cell--name">API Name</div>
+          <div class="api-list__table-header__cell api-list__table-header__cell--labels">Labels</div>
+          <div class="api-list__table-header__cell api-list__table-header__cell--version">Version</div>
```
jourdiw: You can include the `@@xxx` i18n labels here 👍

gravitee-apim-portal-webui-next/src/app/catalog/components/apis-list/apis-list.component.html
```diff
@@ -16,50 +16,107 @@
...
+            <div class="api-list__table-row__cell api-list__table-row__cell--name">
+              <span class="api-list__table-row__name-text">{{ api.title }}</span>
+              @if (api.content) {
+                <button class="api-list__table-row__info-button" [matTooltip]="api.content">
+                  <span class="api-list__table-row__info-button__text">i</span>
+                </button>
+              }
+            </div>
+            <div class="api-list__table-row__cell api-list__table-row__cell--labels">
+              @if (api.labels && api.labels.length > 0) {
```
jourdiw: Is there a need to check if the label length is greater than 0?

gravitee-apim-portal-webui-next/src/app/catalog/components/apis-list/apis-list.component.html
```diff
@@ -16,50 +16,107 @@
...
-          } @else {
-            <p i18n="@@noApisAvailable">Sorry, there are no APIs listed yet.</p>
-            <p i18n="@@moreApisComingSoon">Our API catalog is currently being updated. More APIs will be available soon.</p>
-          }
-        </div>
-      </mat-card-content>
-    </mat-card>
+    <div class="api-list__empty-state">
+      <span class="material-icons icon-with-background m3-headline-large" aria-hidden="true">search</span>
+      @if (query() || currentCategory().id) {
```
jourdiw: There will no longer be any category filtering, so this can be simplified as well as the message below.

gravitee-apim-portal-webui-next/src/app/catalog/categories-view/categories-view.component.ts
jourdiw, line 1: I think all components in `categories-view` can be deleted 🤔

gravitee-apim-portal-webui-next/src/app/catalog/components/apis-list/apis-list.component.html
jourdiw, line 1: This is now the principal view for the catalog. WDYT of renaming this component to `catalog.component.ts` and removing categories-view + tabs-view?

gravitee-apim-portal-webui-next/src/app/catalog/components/apis-list/apis-list.component.spec.ts
```diff
@@ -127,94 +127,34 @@ describe('ApisListComponent', () => {
...
 
-    it('should call second page after scrolled event', async () => {
+    it('should call second page when pagination changes', async () => {
       const apiCard = await harnessLoader.getAllHarnesses(ApiCardHarness);
       expect(apiCard).toBeDefined();
       expect(apiCard.length).toEqual(2);
       expect(await apiCard[0].getTitle()).toEqual('Test title');
 
-      document.getElementsByClassName('api-list__container')[0].dispatchEvent(new Event('scrolled'));
+      fixture.componentInstance.onPageChange(2);
```
jourdiw: I would recommend to use harnesses instead of the methods on the component for more reliable tests 👍

gravitee-apim-portal-webui-next/src/app/catalog/components/apis-list/apis-list.component.spec.ts
```diff
@@ -127,94 +127,34 @@ describe('ApisListComponent', () => {
...
-      fixture.detectChanges();
-
-      const allHarnesses = await harnessLoader.getAllHarnesses(ApiCardHarness);
-      expect(allHarnesses.length).toEqual(3);
-
-      const secondPageApiTitle = await allHarnesses[2].getTitle();
-      expect(secondPageApiTitle).toEqual('second page api');
-    });
-
-    it('should call API list with search query', async () => {
```
jourdiw, line 162: Is there a reason you deleted this test?

gravitee-apim-portal-webui-next/src/app/catalog/components/apis-list/apis-list.component.ts
```diff
@@ -60,13 +81,18 @@ export class ApisListComponent {
 
   apiPaginator$: Observable<ApiPaginatorVM> = of();
   loadingPage: boolean = true;
+  pageSize = 20;
+  pageSizeOptions = [8, 20, 40, 80];
+  viewMode = signal<'grid' | 'list'>('grid');
 
   apiListContainerClasses = computed(() => ({
     'api-list__container--mobile': this.isMobile(),
+    'api-list__container--list': this.viewMode() === 'list',
```
jourdiw: This is rather used to handle our mobile views. 
Can you add this list-specific class to the template rather than in the typescript component?

gravitee-apim-portal-webui-next/src/app/catalog/components/apis-list/apis-list.component.ts
```diff
@@ -60,13 +81,18 @@ export class ApisListComponent {
...
   }));
   protected readonly isMobile = inject(ObservabilityBreakpointService).isMobile;
 
   private readonly page$ = new BehaviorSubject<number>(1);
+  private readonly pageSize$ = new BehaviorSubject<number>(20);
```
jourdiw: Does `pageSize` and `pageSize$` need to be two different variables? Can they be the same source of truth?

gravitee-apim-portal-webui-next/src/app/catalog/components/apis-list/apis-list.component.ts
```diff
@@ -95,22 +123,18 @@ export class ApisListComponent {
...
 
+  toggleViewMode() {
+    this.viewMode.set(this.viewMode() === 'grid' ? 'list' : 'grid');
+  }
+
   private loadApis$(): Observable<ApiPaginatorVM> {
     return this.page$.pipe(
-      map(currentPage => ({ currentPage, category: this.categoryId(), query: this.query() })),
+      switchMap(currentPage =>
+        this.pageSize$.pipe(map(pageSize => ({ currentPage, pageSize, category: this.categoryId(), query: this.query() }))),
```
jourdiw: Maybe categoryId can be 🗑️ ?

gravitee-apim-portal-webui-next/src/app/app.routes.ts
```diff
@@ -142,48 +138,25 @@ export const routes: Routes = [
...
+  {
+    path: 'categories',
+    redirectTo: 'catalog',
+    pathMatch: 'full',
+  },
+  {
+    path: 'categories/catalog',
+    redirectTo: 'catalog',
+    pathMatch: 'full',
+  },
```
jourdiw, line 151: Maybe add a comment here that this is only for backward compatibility for users that already have the categories url saved someplace?

gravitee-apim-portal-webui-next/src/components/api-card/api-card.component.html
```diff
@@ -15,18 +15,14 @@
     limitations under the License.
 
 -->
-<mat-card class="api-card" appearance="outlined">
-  <mat-card-content class="api-card__content">
+<gmd-card class="api-card" [routerLink]="['.', 'api', id]">
```
jourdiw: Instead of adding a router link here, I think it might be best to have an output event `(select)` and let the parent component handle the navigation.

This will respect more of the smart/dummy architecture and allow this component to be more reusable 👍

gravitee-apim-portal-webui-next/src/guards/catalog-tabs-view.guard.ts
jourdiw, line 1: I think this can be 🗑️

jourdiw (commented):
We will discuss the CSS in a dedicated call 👍

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.ts
```diff
@@ -36,18 +42,33 @@ interface ApiVM {
   content: string;
   isEnabledMcpServer: boolean;
   picture?: string;
+  labels?: string[];
```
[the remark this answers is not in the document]
adam-rosowicz: The ApiCardComponent doesn't have a labels input - labels are only rendered in the table view inside catalog.component.html, not inside app-api-card. The ApiCardHarness reflects what api-card actually exposes, so there's nothing to add there.

gravitee-apim-portal-webui-next/src/app/app.routes.ts
adam-rosowicz: Done.

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.html
```diff
@@ -0,0 +1,134 @@
...
+        </button>
+        <button
+          class="api-list__view-toggle__button"
+          [class.api-list__view-toggle__button--active]="viewMode() === 'list'"
+          (click)="viewMode.set('list')"
+          aria-label="List view">
+          <mat-icon>view_list</mat-icon>
+          <span>List</span>
+        </button>
+      </div>
```
jourdiw: If you can make these buttons Material Buttons, then you can avoid a lot of the CSS design that you have in the stylesheet and guarantee matching behavior with other buttons in the application. 👍

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.scss
```diff
@@ -0,0 +1,347 @@
...
+    align-items: center;
+    justify-content: space-between;
+    gap: 16px;
+
+    .m3-headline-medium {
+      font-size: 48px;
+      font-weight: 400;
+      line-height: 1.2;
+      color: theme.$default-text-color;
+    }
```
jourdiw: Instead of overriding the font class, can you just reference a class from the html? You can choose a class from the `typography.scss` file 👍

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.scss
```diff
@@ -0,0 +1,347 @@
...
+    &--narrow {
+      grid-template-columns: repeat(2, 1fr);
+    }
+
+    &--mobile {
+      grid-template-columns: 1fr !important;
+    }
+
+    &--list {
+      grid-template-columns: 1fr !important;
```
jourdiw: Are the "!important" necessary?

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.scss
```diff
@@ -0,0 +1,347 @@
...
+    overflow: hidden;
+    width: 100%;
+
+    &-header {
+      display: grid;
+      grid-template-columns: 1fr 2fr 0.5fr 100px;
+      background-color: theme.$table-header-background-color;
+      padding: 16px;
+      font-weight: 600;
+      font-size: 14px;
```
jourdiw: Instead of using font attributes in the scss, use a class on the html tag from the `typography.scss` file 👍

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.html
```diff
@@ -0,0 +1,134 @@
...
+              [title]="api.title"
+              [version]="api.version"
+              [content]="api.content"
+              [picture]="api.picture"
+              [isEnabledMcpServer]="api.isEnabledMcpServer"
+              (select)="navigateToApi(api.id)" />
+          }
+        </div>
+      } @else {
+        <div class="api-list__table">
```
jourdiw: You should be able to use the table in the `paginated-table.component.ts` here which will greatly simplify the css (there shouldn't be any css for the table really except for the column widths).

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.ts
```diff
@@ -0,0 +1,159 @@
...
+  private readonly breakpointService = inject(ObservabilityBreakpointService);
+  private readonly router = inject(Router);
+  private readonly route = inject(ActivatedRoute);
+  protected readonly isMobile = this.breakpointService.isMobile;
+  protected readonly isNarrow = this.breakpointService.isNarrow;
+
+  private readonly page$ = new BehaviorSubject<number>(1);
+
+  constructor(private readonly apiService: ApiService) {
+    this.query = toSignal(this.route.queryParams.pipe(map(p => p['query'] ?? '')), { initialValue: '' });
```
jourdiw: These can be injected like your other services so that the code is just a tad more coherent:


```suggestion
  private readonly apiService = inject(ApiService);
  protected readonly query = toSignal(this.route.queryParams.pipe(map(p => p['query'] ?? '')), { initialValue: '' });
```

gravitee-apim-portal-webui-next/src/app/app.routes.ts
```diff
@@ -143,48 +138,25 @@ export const routes: Routes = [
...
   {
     path: 'catalog',
     canActivateChild: [redirectGuard, authGuard],
-    data: { breadcrumb: 'Catalog' },
+    data: { breadcrumb: { skip: true } },
```
jourdiw: You can delete this if you like, we will not be using the breadcrumb at all I think for Catalog in the 4.11 release.

gravitee-apim-portal-webui-next/src/components/api-card/api-card.component.ts
```diff
@@ -33,12 +28,12 @@ export class ApiCardComponent {
...
-  @Input({ required: true })
-  id!: string;
   @Input()
   picture: string | undefined;
   @Input()
   isEnabledMcpServer: boolean = false;
   @Input()
   content?: string;
+
+  select = output<void>();
```
jourdiw: Perhaps this can emit the id of the api at least so that the parent can just use the `$event` for routing the request?

gravitee-apim-portal-webui-next/src/components/search-bar/search-bar.component.scss
```diff
@@ -1,43 +1,68 @@
...
   --mat-form-field-subscript-text-size: 0;
+  width: 100%;
+
+  @include mat.form-field-overrides(
+    (
+      outlined-outline-color: theme.$search-bar-border-color,
+      outlined-hover-outline-color: theme.$search-bar-border-color,
+      outlined-container-shape: 4px,
+    )
+  );
```
jourdiw: You should be able to use the `form-field-overrides` for all of the tokens 👍

gravitee-apim-portal-webui-next/src/components/search-bar/search-bar.component.scss
jourdiw, line 1: - Could you replace the license in the file?
- Instead of using hard-coded font sizes, you can use a font class in the `typography.scss` file

gravitee-apim-portal-webui-next/src/scss/theme/variables.scss
```diff
@@ -120,3 +121,15 @@ $chip-background: var(--gio-app-chip-container-color, $card-background-color);
...
 $calendar-background-color: color-mix(in srgb, $banner-background-color 50%, $card-background-color);
 $primary-highlight-color: color-mix(in srgb, $primary-main-color 8%, white 92%);
+
+// -- Catalog / View Toggle --
+$view-toggle-active-background-color: var(--gio-app-view-toggle-active-container-color, $primary-highlight-color);
+$view-toggle-active-text-color: var(--gio-app-view-toggle-active-label-text-color, $primary-main-color);
+
+// -- MCP chip --
+$mcp-chip-background-color: var(--gio-app-mcp-chip-container-color, color-mix(in srgb, $primary-main-color 17%, transparent));
+$mcp-chip-text-color: var(--gio-app-mcp-chip-label-text-color, $primary-main-color);
```
jourdiw: This is a bit too specific for theme-level variables.
I think that you can create a new component: `badge` which looks like the MCP badge in the Figma.
Then the tokens can be `--gio-app-badge-**` and stay generic.

gravitee-apim-portal-webui-next/src/scss/theme/variables.scss
```diff
@@ -120,3 +121,15 @@ $chip-background: var(--gio-app-chip-container-color, $card-background-color);
...
+
+// -- Keyboard shortcut --
+$keyboard-shortcut-text-color: var(--gio-app-keyboard-shortcut-label-text-color, $paragraph-color);
+$keyboard-shortcut-background-color: var(--gio-app-keyboard-shortcut-container-color, #e1e5ed);
```
jourdiw: I think these can be scoped by the search bar (lines 107-110) and rely on chip colors for fallback values:

```suggestion
$search-bar-shortcut-text-color: var(--gio-app-search-bar-shortcut-label-text-color, $paragraph-color);
$search-bar-shortcut-background-color: var(--gio-app-search-bar-shortcut-container-color, $card-background-color);
```

gravitee-apim-portal-webui-next/src/scss/theme/variables.scss
```diff
@@ -120,3 +121,15 @@ $chip-background: var(--gio-app-chip-container-color, $card-background-color);
 $select-panel-background-color: color-mix(in srgb, $banner-background-color 12%, $card-background-color);
 $calendar-background-color: color-mix(in srgb, $banner-background-color 50%, $card-background-color);
 $primary-highlight-color: color-mix(in srgb, $primary-main-color 8%, white 92%);
+
+// -- Catalog / View Toggle --
+$view-toggle-active-background-color: var(--gio-app-view-toggle-active-container-color, $primary-highlight-color);
+$view-toggle-active-text-color: var(--gio-app-view-toggle-active-label-text-color, $primary-main-color);
```
jourdiw: I think that we could have a separate component for the button toggle as it is a common design system component.

WDYT:
- create a dedicated component in `/components` for `button-toggle-group` and `button-toggle-option`
- use https://v20.material.angular.dev/components/button-toggle/examples as inspiration for how to organize the component and what kind of inputs make sense
- scope the css tokens --> `--gio-app-button-toggle-*`

gravitee-apim-portal-webui-next/src/scss/helper.scss
```diff
@@ -43,3 +43,18 @@ code {
...
+  height: 27px;
+  padding: 0 10px;
+  border-radius: 4px;
+  background-color: #{theme.$mcp-chip-background-color};
+  color: #{theme.$mcp-chip-text-color};
+  font-family: Helvetica, Arial, sans-serif;
+  font-weight: 400;
+  font-size: 14px;
+  line-height: 27px;
+}
```
jourdiw: This can be a separate component (`badge`) and not use any font overrides.

gravitee-apim-portal-webui-next/src/scss/material-design-overrides.scss
```diff
@@ -19,7 +19,57 @@
...
+
+      @media (max-width: 880px) {
+        padding: 0 8px !important;
+      }
+    }
+
+    .mdc-evolution-chip__action--primary::before {
+      display: none !important;
+    }
+  }
```
jourdiw: This is not best practice by using `!important`. Can you use the chip overrides mixin from angular? 
This also seems like a mix between the chips and mcp chips. I'm not sure why this is in the global level of the app 🤔

gravitee-apim-portal-webui-next/src/services/observability-breakpoint.service.ts
```diff
@@ -20,14 +20,21 @@ import { map, shareReplay } from 'rxjs/operators';
...
-      shareReplay({ refCount: true, bufferSize: 1 }),
-    );
+  readonly isMobile$ = this.breakpointObserver.observe([Breakpoints.XSmall]).pipe(
+    map(state => state.matches),
+    shareReplay({ refCount: true, bufferSize: 1 }),
+  );
   readonly isMobile = toSignal(this.isMobile$);
+
+  // Narrow screens (max-width: 1199.98px) — used for 2-column catalog layout
+  readonly isNarrow$ = this.breakpointObserver.observe(['(max-width: 1199.98px)']).pipe(
```
jourdiw: Can you find an equivalent breakpoint in the Breakpoints enum rather than hard-coded?

yarn.lock
jourdiw, line 1: I think all these changes can be discarded.

gravitee-apim-portal-webui-next/src/app/catalog/components/apis-list/apis-list.component.spec.ts
adam-rosowicz: Have to be by mistake. It's restored.

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.html
```diff
@@ -0,0 +1,129 @@
...
+            <span class="m3-body-medium" i18n="@@showPerPage">Show per page:</span>
+            <mat-form-field appearance="outline" subscriptSizing="dynamic">
+              <mat-select [value]="pageSize" (selectionChange)="onPageSizeChange($event.value)">
+                @for (size of pageSizeOptions; track size) {
+                  <mat-option [value]="size">{{ size }}</mat-option>
+                }
+              </mat-select>
+            </mat-form-field>
+          </div>
+        </div>
```
jourdiw: You can reuse what exists in this PR: #15464 
That way you are not re-coding the drop down menu and layout

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.scss
```diff
@@ -0,0 +1,193 @@
...
+
+.api-list {
+  &__header {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    gap: 16px;
+
+    .next-gen-h1 {
+      color: theme.$default-text-color;
```
jourdiw: Isn't this the color by default 🤔 Why are you overriding it here?

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.scss
```diff
@@ -0,0 +1,193 @@
...
+    overflow-x: auto;
+  }
+
+  &__table {
+    width: 100%;
+
+    .mat-column-name {
+      width: 25%;
+      font-weight: 500;
+      color: theme.$default-text-color;
```
jourdiw: Best not to override the theming here 🤔

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.html
```diff
@@ -0,0 +1,129 @@
...
+          <table mat-table [dataSource]="apis.data" class="api-list__table">
+
+            <ng-container matColumnDef="name">
+              <th mat-header-cell *matHeaderCellDef i18n="@@tableHeaderApiName">API Name</th>
+              <td mat-cell *matCellDef="let api">
+                <span class="api-list__table__name-text">{{ api.title }}</span>
+                @if (api.content) {
+                  <button class="api-list__table__info-button" [matTooltip]="api.content">
+                    <span class="api-list__table__info-button__text next-gen-caption">i</span>
+                  </button>
```
jourdiw: You can use the material icon instead of making a button: https://fonts.google.com/icons?selected=Material+Symbols+Outlined:info:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=info&icon.size=24&icon.color=%23e8eaed

This can then greatly simplify the css that you have in the `catalog.component.scss`

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.scss
```diff
@@ -0,0 +1,193 @@
...
+        font-family: 'Times New Roman', serif;
+        font-style: italic;
+        color: theme.$default-text-color;
+        line-height: 1;
+      }
+
+      &:hover {
+        background-color: var(--mat-sys-surface-container-low);
+      }
+    }
```
jourdiw: I think this can all be deleted once using the material icon 🤔

gravitee-apim-portal-webui-next/src/components/badge/badge.component.ts
```diff
@@ -13,28 +13,15 @@
...
-  }
+@Component({
+  selector: 'app-badge',
+  standalone: true,
+  templateUrl: './badge.component.html',
+  styleUrl: './badge.component.scss',
+})
+export class BadgeComponent {
+  @Input({ required: true })
+  label!: string;
```
jourdiw: You can use signal inputs for new components 👍

gravitee-apim-portal-webui-next/src/components/button-toggle-group/button-toggle-option.component.html
```diff
@@ -0,0 +1,25 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<button
+  class="button-toggle-option"
+  [class.button-toggle-option--active]="isActive"
+  mat-button
+  (click)="select()">
+  <mat-icon>{{ icon }}</mat-icon>
```
jourdiw: You can simplify the scss by using: `<span class="material-icons">{{ icon }}</span>`

gravitee-apim-portal-webui-next/src/components/button-toggle-group/button-toggle-option.component.scss
```diff
@@ -13,35 +13,36 @@
...
 
-.api-list {
-  &__header {
-    display: flex;
-    flex-wrap: wrap;
-    align-items: center;
-    justify-content: space-between;
-    gap: 20px;
+.button-toggle-option {
+  --mdc-text-button-container-height: 40px;
```
jourdiw: I would recommend to use the angular material mixins rather than the hard-coded token 👍

gravitee-apim-portal-webui-next/src/components/button-toggle-group/button-toggle-option.component.scss
```diff
@@ -13,35 +13,36 @@
...
+    line-height: 20px;
   }
 
-  &__container {
-    display: grid;
-    gap: 32px;
-    grid-template-columns: repeat(3, minmax(0, 1fr));
+  &--active {
+    background-color: theme.$button-toggle-active-background-color;
+    color: theme.$button-toggle-active-text-color;
```
jourdiw: I would recommend to use the angular material mixins rather than overriding the colors here 👍

gravitee-apim-portal-webui-next/src/components/button-toggle-group/button-toggle-option.component.scss
```diff
@@ -13,35 +13,36 @@
...
-    display: flex;
-    flex-wrap: wrap;
-    align-items: center;
-    justify-content: space-between;
-    gap: 20px;
+.button-toggle-option {
+  --mdc-text-button-container-height: 40px;
+  border-radius: 0;
+  min-width: unset;
+  color: theme.$paragraph-color;
```
jourdiw: Both border-radius and color should be covered by the angular material button mixins

gravitee-apim-portal-webui-next/src/components/button-toggle-group/button-toggle-option.component.ts
```diff
@@ -0,0 +1,48 @@
...
+})
+export class ButtonToggleOptionComponent {
+  @Input({ required: true })
+  value!: string;
+
+  @Input({ required: true })
+  icon!: string;
+
+  @Input({ required: true })
+  label!: string;
```
jourdiw: Use input signals for all new components 👍

gravitee-apim-portal-webui-next/src/components/search-bar/search-bar.component.scss
jourdiw: Still missing these changes ⤴️

gravitee-apim-portal-webui-next/src/scss/material-design-overrides.scss
```diff
@@ -19,7 +19,21 @@
...
+    (
+      elevated-container-color: #{theme.$label-chip-container-color},
+      label-text-color: #{theme.$label-chip-label-text-color},
+      container-shape-radius: #{theme.$label-chip-container-shape-radius},
+    )
+  );
+
+  .mdc-evolution-chip__action--primary::before {
+    display: none;
+  }
```
jourdiw: Is this still necessary after refactoring?

adam-rosowicz: Yes, these overrides are still necessary. Without them the chip appearance reverts to the default Material Design styles - the container color, label color, and border radius all differ from the design spec.

gravitee-apim-portal-webui-next/src/components/button-toggle-group/button-toggle-option.component.scss
```diff
@@ -0,0 +1,58 @@
...
+    )
+  );
+  min-width: unset;
+
+  .material-icons {
+    font-size: 20px;
+    width: 20px;
+    height: 20px;
+    line-height: 20px;
+  }
```
jourdiw: This should be handled by the global `.material-icons` 🤔 I don't think you need to add the class here.

gravitee-apim-portal-webui-next/src/components/button-toggle-group/button-toggle-option.component.ts
```diff
@@ -0,0 +1,57 @@
...
+export class ButtonToggleOptionComponent {
+  readonly value = input.required<string>();
+  readonly icon = input.required<string>();
+  readonly label = input.required<string>();
+
+  protected readonly group = inject(ButtonToggleGroupComponent);
+
+  get isActive(): boolean {
+    return this.group.value() === this.value();
+  }
```
jourdiw: Should be able to use a computed here to minimize complexity.
```suggestion
  isActive(): signal<boolean> = computed(() => this.group.value() === this.value());
```

gravitee-apim-portal-webui-next/src/scss/theme/variables.scss
```diff
@@ -111,12 +111,27 @@ $search-bar-input-text-color: var(--gio-app-search-bar-input-text-color);
...
-$table-header-background-color: var(--gio-app-table-header-container-color, none);
+$table-header-background-color: var(--gio-app-table-header-container-color, var(--mat-sys-surface-container-low));
 $table-body-background-color: var(--gio-app-table-body-container-color, $card-background-color);
 $table-header-headline-color: var(--gio-app-table-header-headline-color, $default-text-color);
 
 // -- Other Components --
 $chip-background: var(--gio-app-chip-container-color, $card-background-color);
+$label-chip-container-color: var(--gio-app-label-chip-container-color, #e1e5ed);
+$label-chip-label-text-color: var(--gio-app-label-chip-label-text-color, #5f6368);
+$label-chip-container-shape-radius: var(--gio-app-label-chip-container-shape-radius, 16px);
```
jourdiw: What's the functional difference between the label chip and the chip? 
My guess is that chips should now look like what you are designating "label chip". WDYT?

jourdiw (approved):
🎊

PR #15006, opened 2026-01-30
fix(gateway): strip double quotes from Forwarded header host

prateektiwari-gravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-12533

## Description

This pull request addresses an issue where the gateway was unable to correctly parse the host value from the Forwarded header when it was enclosed in double quotes. The changes ensure that the host is extracted accurately, aligning with RFC 7239 specifications, which improves the reliability of host resolution in proxied environments.

## Additional context

Forwarded Header Parsing Fix: Corrected the parsing logic for the host parameter within the Forwarded HTTP header to properly handle values enclosed in double quotes, as permitted by RFC 7239.
Quote Removal: Implemented a mechanism to remove double quotes from the extracted host string if they are present, ensuring accurate host resolution.
New Test Cases: Added comprehensive test cases to validate the updated Forwarded header parsing, covering scenarios with quoted hosts, hosts with ports, and multiple hosts within the header.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/forward/XForwardProcessor.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/forward/XForwardProcessor.java
@@ -93,6 +93,9 @@ public class XForwardProcessor implements Processor {
                 }
                 if (normalizedElement.startsWith("host=")) {
                     String host = element.trim().substring(5);
+                    if (host.indexOf('"') != -1) {
+                        host = host.replace("\"", "");
+                    }
                     int commaIndex = host.indexOf(',');
                     originalHost = commaIndex == -1 ? host : host.substring(0, commaIndex).trim();
                 }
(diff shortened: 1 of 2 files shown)


prateektiwari-gravitee:
## Existing Functionality

https://github.com/user-attachments/assets/ce40541c-2f7b-4502-914f-b1146429ce48



## With Fixes
https://github.com/user-attachments/assets/fb8a19d8-36c8-4864-91b8-6a4698cc94d3

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/forward/XForwardProcessor.java
```diff
@@ -93,6 +93,9 @@ static String generateOriginalUrl(HttpBaseRequest request) {
                 }
                 if (normalizedElement.startsWith("host=")) {
                     String host = element.trim().substring(5);
+                    if (host.indexOf('"') != -1) {
+                        host = host.replaceAll("\"", "");
```
arpzilla-dev: syntactical review : 

- replace could be use to avoid replaceAll regex overhead

phiz71:
Should it be also applied on 4.8.x and 4.7.x ?

schehade-gravitee (approved):

prateektiwari-gravitee:
> Should it be also applied on 4.8.x and 4.7.x ?

This issue is only valid for newer versions. The issue was in [this newly added](https://github.com/gravitee-io/gravitee-api-management/blob/d638ecea3c05bfd22144ef5c8ef08724b8e98719/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/forward/XForwardProcessor.java#L74) function for 4.10

PR #15026, opened 2026-02-03
fix: Added gateway instance details to api metrics detail

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-12400

## Description
Added gateway instance details to api metrics detail instead making another getInstanceById API call which explicitly requires INSTANCE_READ permission

Result: API owners can now access API logs along with instance details without any error
After Fix: 

https://github.com/user-attachments/assets/8c622334-74bd-4b0b-9aa7-a2fb05456402

Before Fix: 

https://github.com/user-attachments/assets/21931f48-ba51-48b4-a2a2-1e91ecd1aabb


pragsty:
Created by mistake. Already Existing PR https://github.com/gravitee-io/gravitee-api-management/pull/15010

PR #15027, opened 2026-02-03
Some warnings displayed after upgrade to v 4.7.x

juned-gravitee:
## Issue

## Description

Introduced a centralized repositories configuration block to standardize how plugins (Management, Analytics, Rate Limit) are defined. kwhile maintaining backward compatibility.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/common/AbstractJdbcRepositoryConfiguration.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/common/AbstractJdbcRepositoryConfiguration.java
@@ -60,10 +60,8 @@ public abstract class AbstractJdbcRepositoryConfiguration implements Application
     @Autowired
     private Environment env;
 
-    @Value("${management.jdbc.prefix:}")
     private String prefix;
 
-    @Value("${ratelimit.jdbc.prefix:}")
     private String rateLimitPrefix;
 
     // Default escape char for reserved keywords
@@ -139,7 +137,8 @@ public abstract class AbstractJdbcRepositoryConfiguration implements Application
         dsConfig.setJdbcUrl(jdbcUrl);
         dsConfig.setUsername(readPropertyValue("jdbc.username"));
         dsConfig.setPassword(readPropertyValue("jdbc.password", false));
-        final String schema = env.getProperty(getScope() + "." + "jdbc.schema", String.class);
+        final String schema = readPropertyValue("jdbc.schema", String.class, null);
+
         if (schema != null) {
             dsConfig.setSchema(schema);
         }
@@ -155,7 +154,8 @@ public abstract class AbstractJdbcRepositoryConfiguration implements Application
 
         final DataSource dataSource = new HikariDataSource(dsConfig);
 
-        // Upgrader framework may also set this value in the runtime to skip running Liquibase scripts in upgrade mode
+        // Upgrader framework may also set this value in the runtime to skip running
+        // Liquibase scripts in upgrade mode
         Boolean liquibase = readPropertyValue("jdbc.liquibase", Boolean.class, LIQUIBASE_ENABLED);
         if (liquibase) {
             runLiquibase(dataSource);
@@ -201,6 +201,9 @@ public abstract class AbstractJdbcRepositoryConfiguration implements Application
     private void runLiquibase(DataSource dataSource) {
         LOGGER.debug("Running Liquibase on {}", dataSource);
 
+        this.prefix = readPropertyValue("jdbc.prefix", String.class, "");
+        this.rateLimitPrefix = readPropertyValue("ratelimit.jdbc.prefix", String.class, "");
+
         System.setProperty("liquibase.databaseChangeLogTableName", prefix + "databasechangelog");
         System.setProperty("liquibase.databaseChangeLogLockTableName", prefix + "databasechangeloglock");
         System.setProperty("gravitee_prefix", prefix);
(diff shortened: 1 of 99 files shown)


gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/common/AbstractJdbcRepositoryConfiguration.java
```diff
@@ -232,7 +233,26 @@ private <T> T readPropertyValue(String propertyName, Class<T> propertyType, T de
...
+        if (value != null) {
+            return value;
+        }
+
+        String oldKey = scope + "." + propertyName;
+        value = env.getProperty(oldKey, propertyType, defaultValue);
+
+        if (value != null && propertyName.equals("jdbc.url")) {
+            LOGGER.warn(
+                "The JDBC configuration for scope {} is loaded from discouraged section '{}'. Please use '{}'.",
```
wbabyte: why do you want to duplicate this message?

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RateLimitRepositoryConfiguration.java
```diff
@@ -20,35 +20,62 @@
...
 import org.springframework.beans.factory.annotation.Qualifier;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.core.env.Environment;
 
 /**
  * @author David BRASSELY (david.brassely at graviteesource.com)
  * @author GraviteeSource Team
  */
+@Slf4j
```
wbabyte: use @CustomLog

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RateLimitRepositoryConfiguration.java
```diff
@@ -20,35 +20,62 @@
...
+
+        // Determine prefix by checking if a core property exists in the new path
+        String resolvedPrefix = (environment.containsProperty(newPrefix + ".redis.host") ||
+                environment.containsProperty(newPrefix + ".redis.sentinel.nodes[0].host"))
+            ? newPrefix
+            : oldPrefix;
+
+        if (resolvedPrefix.equals(oldPrefix) && environment.containsProperty(oldPrefix + ".redis.host")) {
+            log.warn(
+                "Redis configuration for scope {} is loaded from discouraged section '{}.redis'. Please use '{}.redis'.",
```
wbabyte: why duplicate?

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RateLimitRepositoryConfiguration.java
```diff
@@ -20,35 +20,62 @@
...
 
     @Bean("redisRateLimitClient")
     public RedisClient redisRedisClient(Environment environment, Vertx vertx) {
+        String scope = Scope.RATE_LIMIT.getName(); // "ratelimit"
+        String oldPrefix = scope;
+        String newPrefix = "repositories." + scope;
+
+        // Determine prefix by checking if a core property exists in the new path
+        String resolvedPrefix = (environment.containsProperty(newPrefix + ".redis.host") ||
+                environment.containsProperty(newPrefix + ".redis.sentinel.nodes[0].host"))
```
wbabyte: why you check it like this?

wbabyte (changes requested):
you are missing tests. This has to be tested very well to not introduce a breaking change 
remember about helm tests as well. 
when using agents could you instruct it to not add comments when not required? Could you remove non required comments?

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/common/AbstractJdbcRepositoryConfiguration.java
juned-gravitee: When RepositoryPluginHandler or RepositoryTypeReader initializes plugins, they check the environment and log a WARN if the legacy ratelimit scope is used instead of repositories.ratelimit

you are correct , will revert this part

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RateLimitRepositoryConfiguration.java
juned-gravitee: you are correct , will revert this part

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RateLimitRepositoryConfiguration.java
juned-gravitee: Checking for these specific "leaf" properties is more robust than just checking if the key exists. In Spring, a key might exist but be empty or contain irrelevant data. By checking for the core connection parameters, we ensure that we only switch to the new prefix if there is actual connection data there to read. repositories.ratelimit

PR #15031, opened 2026-02-03
feat: use ClientCertificates instead of application metadata

benoitgravitee:
## Issue

https://gravitee.atlassian.net/browse/GKO-1997

## Description

* Replace usage of App metadata by client certificate service
* Added a new repository method to streamline performances
* package multiple certificate in PKCS7 format in subscription
* let Gateway read PEM and PCKS7 when updating truststore 

Backward compatibility is maintained, single certificate subscription => still using base64 encoded PEM. For multiple => check if it is a PKCS7 and fallback on PEM.


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ApplicationClientCertificateMigrationUpgrader.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Autowired
+    private ClientCertificateCrudService clientCertificateCrudService;
+
+    @Override
+    public boolean upgrade() throws UpgraderException {
+        return this.wrapException(() -> {
+            log.info("Starting migration of application client certificates to ClientCertificate entity");
+
+            List<Application> applicationsWithCertificate = applicationRepository
```
carlos-andres-osorio: applicationRepository.findAll() loads every application into memory. For large installations this risks OOM. Is it possible to use pagination/streaming?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ApplicationClientCertificateMigrationUpgrader.java
```diff
@@ -0,0 +1,133 @@
...
+                        app.getMetadata().containsKey(METADATA_CLIENT_CERTIFICATE) &&
+                        app.getMetadata().get(METADATA_CLIENT_CERTIFICATE) != null &&
+                        !app.getMetadata().get(METADATA_CLIENT_CERTIFICATE).isBlank()
+                )
+                .toList();
+
+            log.info("Found {} applications with client certificates to migrate", applicationsWithCertificate.size());
+
+            for (Application application : applicationsWithCertificate) {
+                migrateApplicationCertificate(application);
```
carlos-andres-osorio: If `migrateApplicationCertificate(application)` fails for one application (e.g., duplicate fingerprint, invalid cert), migration is halted for all remaining applications. Should catch/log per-application errors and continue?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/application_certificate/use_case/CreateClientCertificateUseCase.java
```diff
@@ -26,9 +27,11 @@
 public class CreateClientCertificateUseCase {
 
     private final ClientCertificateCrudService clientCertificateCrudService;
+    private final ApplicationCertificatesUpdateDomainService applicationCertificatesUpdateDomainService;
 
     public Output execute(Input input) {
         ClientCertificate certificate = clientCertificateCrudService.create(input.applicationId(), input.createClientCertificate());
+        applicationCertificatesUpdateDomainService.updateActiveMTLSSubscriptions(input.applicationId());
```
carlos-andres-osorio, line 33: Do requests run DB updates in a transaction?

If not, for create and update operations, If PKCS7 bundle creation throws an exception, the certificate will be left in the DB.

Delete will delete the certificate and leave the subscriptions in a bad state.

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoader.java
```diff
@@ -39,20 +41,36 @@ public class SubscriptionTrustStoreLoader extends AbstractKeyStoreLoader<Subscri
...
+                },
+                false
+            ).orElseGet(() -> {
+                aliases.add(subscription.getId());
+                return KeyStoreUtils.initFromPemCertificate(new String(decodedData), getPassword(), subscription.getId());
+            });
+
+            // compute digests from all certificates to be able to find them later
+            for (String alias : aliases) {
+                digests.add(DigestUtils.md5DigestAsHex(keystore.getCertificate(alias).getEncoded()));
```
carlos-andres-osorio: MD5 has been deprecated for some time due to the posibility of having collitions.  Would it be better to use a hash like SHA-256?

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoaderManager.java
```diff
@@ -61,18 +61,24 @@ public void registerSubscription(Subscription subscription, Set<String> deployOn
             .map(s -> (VertxServer<?, ?>) s)
             .forEach(server -> server.trustStoreLoaderManager().registerLoader(loader));
         subscriptionTrustStoreLoaders.put(subscription.getId(), loader);
```
carlos-andres-osorio: The check (`containsKey`) and the update (`put`) operations are not atomic, so it's possible to have concurrent operations add the same ID twice.

Would it be better to use `putIfAbsent`?

```suggestion
subscriptionTrustStoreLoaders.putIfAbsent(subscription.getId(), loader);
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
```diff
@@ -496,89 +480,60 @@ public ApplicationEntity create(final ExecutionContext executionContext, NewAppl
...
-        Application application
+        Application application,
+        String clientCertificate
     ) {
         try {
             application.setEnvironmentId(executionContext.getEnvironmentId());
 
+            // Create client certificate if provided
+            if (clientCertificate != null) {
+                clientCertificateCrudService.create(
```
carlos-andres-osorio, line 522: clientCertificateCrudService.create() is called before applicationRepository.create(). 

If application creation fails, the certificate is orphaned. Is this OK?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
```diff
@@ -784,80 +714,190 @@ public ApplicationEntity update(
...
+        Optional<String> clientCertificate
+    ) {
+        // Set correct client_id for all active subscriptions
+        SubscriptionQuery subQuery = new SubscriptionQuery();
+        subQuery.setApplication(applicationId);
+        subQuery.setStatuses(Set.of(SubscriptionStatus.ACCEPTED, SubscriptionStatus.PAUSED, SubscriptionStatus.PENDING));
+
+        String clientId = application.getMetadata().get(METADATA_CLIENT_ID);
+
+        final String subscriptionClientCertificate = clientCertificate
```
carlos-andres-osorio, line 769: Is it OK to use a base64-encoded single PEM here, or should it also use PKCS7 bundling?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ApplicationClientCertificateMigrationUpgrader.java
```diff
@@ -0,0 +1,133 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * Upgrader that migrates client certificates from application metadata to the ClientCertificate entity.
+ *
+ * @author GraviteeSource Team
+ */
+@Slf4j
+@Component
+public class ApplicationClientCertificateMigrationUpgrader implements Upgrader {
```
carlos-andres-osorio, line 47: Is it possible to add tests for this class?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/crud_service/application_certificates/ClientCertificateCrudServiceImplTest.java
```diff
@@ -0,0 +1,535 @@
...
+
+        assertThatThrownBy(() -> clientCertificateCrudService.findById(CERTIFICATE_ID))
+            .isInstanceOf(TechnicalManagementException.class)
+            .hasMessageContaining(CERTIFICATE_ID);
+    }
+
+    @Test
+    void should_create_certificate_with_active_status_when_no_dates() throws TechnicalException {
+        CreateClientCertificate createDto = new CreateClientCertificate("New Certificate", null, null, PEM_CERTIFICATE);
+        when(clientCertificateRepository.existsByFingerprintAndActiveApplication(any(), eq(ENVIRONMENT_ID))).thenReturn(false);
```
carlos-andres-osorio, line 161: All tests set `existsByFingerprintAndActiveApplication` to return `false`.  It would be good to test the case where the certificate already exists.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ApplicationClientCertificateMigrationUpgrader.java
carlos-andres-osorio: It really depends on how large our bigger customers are, so it may not be necessary right now, but if it's simple, it would be a good improvement.

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoader.java
carlos-andres-osorio: I see that it was like this before, so I'm not sure of the impact if it's changed now.

Also it may be unlikelly for a collition to happen, so that may impact the decision to stick to MD5 hashes.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ApplicationClientCertificateMigrationUpgrader.java
benoitgravitee: It should not fail, but I will make a test with duplicated certificates on an active and archived application.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/application_certificate/use_case/CreateClientCertificateUseCase.java
benoitgravitee: Answers is yes it is transactional for JDBC. But Mongo does not support transactions... Chances are low since PKCS7 use certificates that where already validated.

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoader.java
benoitgravitee: 100% agree... It implies I create a BC in mtls policy I will do it

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
benoitgravitee: Yes... there are more chances that the application will not fail. it was the other way around before... a bad cert leaves in a bad state.
I need to make adjustements around validation anyways

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
benoitgravitee: Gateway backward compatibility... It is meant to allow "lower to eq to 4.10" gateway to consume subscriptions created with a more recent control plane.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_11_0/02_add_client_certificates_table.yml
```diff
@@ -13,14 +13,13 @@ databaseChangeLog:
               - column: { name: starts_at, type: timestamp }
```
graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/test/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoaderTest.java
```diff
@@ -94,6 +98,102 @@ void should_not_start_a_loader_already_started() {
...
+            if (keyStore.isCertificateEntry(alias)) {
+                certCount++;
+            }
+        }
+        assertThat(certCount).isEqualTo(2);
+    }
+
+    @SneakyThrows
+    @Test
+    void should_compute_digest_from_first_certificate_in_pkcs7_bundle() {
```
carlos-andres-osorio: When I read the test name I was expecting to find only one digest, and the assertions were not specific about how many digests should be so I ran the tests and see that (as expected) there are two.

Could the test name be more clear and assertions check this?

```suggestion
    should_compute_digests_from_all_certificates_in_pkcs7_bundle() {
```

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_11_0/02_add_client_certificates_table.yml
graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/crud_service/application_certificates/ClientCertificateCrudServiceImplTest.java
```diff
@@ -0,0 +1,546 @@
+/*
```
carlos-andres-osorio, line 1: There are a few tests that may be worth adding:
- certificate with start and end in the future is added with status scheduled
- when there are multiple active certificates, should `findMostRecentActiveByApplicationId` return the most recent?
- `findMostRecentActiveByApplicationId` returns empty when there are no active certificates

carlos-andres-osorio (approved):
Except for a few suggested unit tests, PR looks good

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
```diff
@@ -889,22 +929,18 @@ public ApplicationEntity updateApiKeyMode(final ExecutionContext executionContex
...
+                updatedApplication.getUpdatedAt(),
+                applicationToUpdate,
+                updatedApplication
             );
 
             return convertApplication(executionContext, Collections.singleton(updatedApplication)).iterator().next();
         } catch (TechnicalException ex) {
             String error = String.format(
-                "An error occurs while trying to update application {} with apiKeyMode {}",
+                "An error occurs while trying to update application % with apiKeyMode %s",
```
a-cordier:
```suggestion
                "An error occurs while trying to update application %s with apiKeyMode %s",
```

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_11_0/02_add_client_certificates_table.yml
```diff
@@ -13,14 +13,13 @@ databaseChangeLog:
...
-              - column: { name: updated_at, type: timestamp(6), constraints: { nullable: false }, defaultValueComputed: CURRENT_TIMESTAMP(6) }
+              - column: { name: updated_at, type: timestamp(6)}
               - column: { name: certificate, type: nclob, constraints: { nullable: false } }
               - column: { name: certificate_expiration, type: timestamp }
               - column: { name: subject, type: nvarchar(1024) }
               - column: { name: issuer, type: nvarchar(1024) }
-              - column: { name: fingerprint, type: nvarchar(256) }
+              - column: { name: fingerprint, type: nvarchar(256), constraints: { nullable: false } }
               - column: { name: environment_id, type: nvarchar(64), constraints: { nullable: false } }
-              - column: { name: organization_id, type: nvarchar(64), constraints: { nullable: false } }
```
a-cordier, line 23: This changeset is only on master and was introduced for the feature under development right ?

a-cordier (changes requested):
The change request is only about this string formatting, the liquibase changeset comment is just a question and can be resolved directly if the answer is yes :)

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_11_0/02_add_client_certificates_table.yml
benoitgravitee: yes. checked with Flo that updating this, won't cause harm as we don't have JDBC envs where master is redeployed

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_11_0/02_add_client_certificates_table.yml
graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

a-cordier (approved):

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_11_0/02_add_client_certificates_table.yml
graviteeio, line 13: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

PR #15035, opened 2026-02-03
fix(portal-next): regularize user registration routes with classic portal

jourdiw:
## Issue

https://gravitee.atlassian.net/browse/APIM-12613

## Description

When the user was clicking the confirm link in the email, it was sent to the same path for user registration confirmation that exists in the Classic Portal. However, this was creating a 404.

This fix regularizes the routing to the user registration confirmation page.

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/app.routes.ts
+++ b/gravitee-apim-portal-webui-next/src/app/app.routes.ts
@@ -280,14 +280,23 @@ export const routes: Routes = [
     ],
   },
   {
-    path: 'registration',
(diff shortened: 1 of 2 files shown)


kovaren (commented):
Could you check if this link still works with the current changes? https://github.com/gravitee-io/gravitee-api-management/blob/67d08683075e2bdc73f4b60e9297f90e335179df/gravitee-apim-portal-webui-next/src/app/log-in/log-in.component.html#L78

kovaren (approved):

PR #15037, opened 2026-02-03
fix: default referrer policy  if not explicitly configured

prateektiwari-gravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-12476

The application strictly requires the presence of the referrerPolicy configuration. If missing, the Spring security filter chain fails to instantiate, causing the entire Management API to crash with the error: Invalid referrer policy provided: 'null'.

## Description

 - Added fall back for `http.secureHeaders.referrerPolicy.policy` 
 - When the referer policy is not provided, it is set as 'strict-origin-when-cross-origin' 
 
## Additional context

| Scenario                                       |  Action                                                                     | Referer Policy                                  | 
| ------------------------------- | ----------------------------------------------- | --------------------------------- | 
| Valid policy                                  |  System applies config referer policy                    |   As configured                                | 
| Invalid policy value                     |  System rejects invalid config with error               |    NA                                                  | 
| No value chosen                         |  System set the default value for referer policy    |  strict-origin-when-cross-origin. |

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-security/src/main/java/io/gravitee/rest/api/security/config/SecureHeadersConfigurer.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-security/src/main/java/io/gravitee/rest/api/security/config/SecureHeadersConfigurer.java
@@ -69,8 +69,10 @@ public interface SecureHeadersConfigurer {
     }
 
     static void referrerPolicy(HttpSecurity security, ConfigurableEnvironment environment) throws Exception {
-        String policyString = environment.getProperty("http.secureHeaders.referrerPolicy.policy");
-
+        String policyString = environment.getProperty(
+            "http.secureHeaders.referrerPolicy.policy",
+            ReferrerPolicyHeaderWriter.ReferrerPolicy.STRICT_ORIGIN_WHEN_CROSS_ORIGIN.getPolicy()
+        );
         var policy = Arrays.stream(ReferrerPolicyHeaderWriter.ReferrerPolicy.values())
             .filter(p -> p.getPolicy().equalsIgnoreCase(policyString))
             .findFirst()


gravitee-apim-rest-api/gravitee-apim-rest-api-security/src/main/java/io/gravitee/rest/api/security/config/SecureHeadersConfigurer.java
```diff
@@ -69,8 +70,10 @@ static void xContentTypeOptions(HttpSecurity security, ConfigurableEnvironment e
     }
 
     static void referrerPolicy(HttpSecurity security, ConfigurableEnvironment environment) throws Exception {
-        String policyString = environment.getProperty("http.secureHeaders.referrerPolicy.policy");
-
+        String policyStringEnv = environment.getProperty("http.secureHeaders.referrerPolicy.policy");
```
thomasballivet:
```suggestion
        String policyString = environment.getProperty("http.secureHeaders.referrerPolicy.policy", ReferrerPolicyHeaderWriter.ReferrerPolicy.STRICT_ORIGIN_WHEN_CROSS_ORIGIN.getPolicy());
```

thomasballivet (approved):

prateektiwari-gravitee:
## Pre fix behaviour: 
 
The video below highlights that the error comes up when the policy is not defined in the config yaml

https://github.com/user-attachments/assets/75f5e699-8de5-4152-b233-e233e81cdd64

## Post fix behaviour: 

 The video below highlights that the error has been fixed when the policy is not defined in the config yaml. The fallback for the referrer policy is `strict-origin-when-cross-origin`


https://github.com/user-attachments/assets/8504455b-cb96-4b40-8e5d-c1cc68bf3f48

PR #15044, opened 2026-02-04
fix: 503 error because thred blocked with vertx for weighted round-robin endpoint

schehade-gravitee:
https://gravitee.atlassian.net/browse/APIM-12668


gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/endpoint/loadbalancer/WeightedRoundRobinLoadBalancer.java
```diff
@@ -36,6 +38,8 @@
  */
 public class WeightedRoundRobinLoadBalancer extends WeightedLoadBalancer {
 
+    private static final Logger log = LoggerFactory.getLogger(WeightedRoundRobinLoadBalancer.class);
+
```
phiz71: you should use `@CustomLog`

phiz71 (approved):
Be careful.
`@CustomLog` will probably not be available on support branches

PR #15058, opened 2026-02-05
gko-2016 add missing llm metrics, facets and filters

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-2016


gravitee-apim-rest-api/gravitee-apim-rest-api-standalone/gravitee-apim-rest-api-standalone-distribution/src/main/resources/config/logback.xml
```diff
@@ -66,6 +66,7 @@
         <appender-ref ref="FILE-UPGRADERS" />
     </logger>
     <logger name="io.gravitee" level="INFO" />
+    <logger name="io.gravitee.repository.elasticsearch.v4.analytics" level="DEBUG" />
```
carlos-andres-osorio: should this be reverted?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
```diff
@@ -64,6 +66,8 @@ spec:
               - GEO_IP_CITY
               - GEO_IP_CONTINENT
               - CONSUMER_IP
+              - LLM_PROXY_MODEL
+              - LLM_PROXY_PROVIDER
```
carlos-andres-osorio, line 70: Should the `HTTP_ERRORS` metric below be updated to add facets `LLM_PROXY_MODEL` and `LLM_PROXY_PROVIDER`?

carlos-andres-osorio: And maybe applies also for other HTTP metrics like  HTTP_REQUEST_CONTENT_LENGTH, HTTP_RESPONSE_CONTENT_LENGTH, HTTP_ENDPOINT_RESPONSE_TIME, HTTP_GATEWAY_RESPONSE_TIME, HTTP_GATEWAY_LATENCY?

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/HTTPMeasuresQueryAdapter.java
```diff
@@ -84,21 +89,54 @@ JsonObject adaptMetrics(List<MetricMeasuresQuery> metrics) {
...
+            case LLM_PROMPT_TOTAL_TOKEN -> aggregateLLMTotalToken(aggName, measure);
+            case LLM_PROMPT_TOKEN_COST -> aggregateLLMTotalCost(aggName, measure);
+            case HTTP_ERRORS -> aggregateHTTPErrors(aggName, field, metric, measure);
+            default -> aggregateByMeasure(aggName, field, metric, measure);
+        };
+    }
+
+    private Optional<Map<String, JsonObject>> aggregateLLMTotalToken(String aggName, Measure measure) {
+        return switch (measure) {
+            case COUNT -> Optional.of(llmTotalTokenBuilder.buildSum(aggName));
```
carlos-andres-osorio, line 107: is the `buildSum` function right for calculating count?

Or maybe using `SUM` as the measure would be more clear?

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/HTTPMeasuresQueryAdapter.java
```diff
@@ -84,21 +89,54 @@ JsonObject adaptMetrics(List<MetricMeasuresQuery> metrics) {
...
+            case AVG -> Optional.of(llmTotalTokenBuilder.buildAvg(aggName));
+            default -> Optional.empty();
+        };
+    }
+
+    private Optional<Map<String, JsonObject>> aggregateLLMTotalCost(String aggName, Measure measure) {
+        return switch (measure) {
+            case COUNT -> Optional.of(llmTotalCostBuilder.buildSum(aggName));
```
carlos-andres-osorio, line 115: ditto

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/HTTPFieldResolver.java
carlos-andres-osorio, line 1: `LLM_PROMPT_TOTAL_TOKEN` and `LLM_PROMPT_TOKEN_COST` are not in `fromMetric()` but these new values were added to the `Metric` enum. Is that OK?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
a-cordier: Good idea, added 🚀

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/HTTPMeasuresQueryAdapter.java
a-cordier: The metric is indeed TOTAL_TOKEN with a count, but to build it, from an ES perspective, we need to perform a sum on TOKEN_SENT and a sum on TOKEN_RECEIVED, then sum this again in a bucket script. There is no 1-1 relation ship between documents and token, one request carries a batch of tokens

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/HTTPMeasuresQueryAdapter.java
a-cordier: ☝️

gravitee-apim-rest-api/gravitee-apim-rest-api-standalone/gravitee-apim-rest-api-standalone-distribution/src/main/resources/config/logback.xml
a-cordier: 🙈 done

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/HTTPFieldResolver.java
a-cordier: Yes that's OK, because this fromMetric is only used when we have a 1-1 relationship between a metric and an indexed field. 

Because those metrics require computation, they cannot be inferred with the field resolver.

This would/will be clearer if/when we have custom mapping on metrics, because we will be able to perform those computation in the service layer. meaning that we could have only fields resolved directly from the index for the metrics.

carlos-andres-osorio (approved):

PR #15079, opened 2026-02-07
gko-1984: REST model and repository layer for custom dashboards

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-1984
specifically https://gravitee.atlassian.net/browse/GKO-2317

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/CustomDashboardWidget.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/CustomDashboardWidget.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.management.model;
+
+import java.util.List;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * @author Antoine CORDIER (antoine.cordier at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+@Builder(toBuilder = true)
+@NoArgsConstructor
+@AllArgsConstructor
+@Data
+public class CustomDashboardWidget {
+
+    private String id;
+    private String title;
+    private String type;
+    private Layout layout;
+    private Request request;
+
+    @Builder
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @Data
+    public static class Layout {
+
+        private int cols;
+        private int rows;
+        private int x;
+        private int y;
+    }
+
+    /**
+     * Time range for widget requests (ISO 8601 date-time strings).
+     * Matches frontend TimeRange interface: { from: string; to: string }.
+     */
+    @Builder
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @Data
+    public static class TimeRange {
+
+        private String from;
+        private String to;
+    }
+
+    @Builder
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @Data
+    public static class Request {
+
+        private String type;
+        private TimeRange timeRange;
+        private List<String> metrics;
+        private List<String> by;
+        private Integer limit;
+    }
+}
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcCustomDashboardRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcCustomDashboardRepository.java
@@ -1,0 +1,223 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.jdbc.management;
+
+import static io.gravitee.repository.jdbc.common.AbstractJdbcRepositoryConfiguration.escapeReservedWord;
+import static io.gravitee.repository.jdbc.orm.JdbcColumn.getDBName;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.gravitee.repository.exceptions.TechnicalException;
+import io.gravitee.repository.jdbc.orm.JdbcColumn;
+import io.gravitee.repository.jdbc.orm.JdbcObjectMapper;
+import io.gravitee.repository.management.api.CustomDashboardRepository;
+import io.gravitee.repository.management.model.CustomDashboard;
+import io.gravitee.repository.management.model.CustomDashboardWidget;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Types;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import lombok.CustomLog;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.jdbc.core.PreparedStatementCreator;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.stereotype.Repository;
+
+/**
+ * @author Antoine CORDIER (antoine.cordier at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+@CustomLog
+@Repository
+public class JdbcCustomDashboardRepository
+    extends JdbcAbstractCrudRepository<CustomDashboard, String>
+    implements CustomDashboardRepository {
+
+    private static final ObjectMapper JSON_MAPPER = new ObjectMapper();
+
+    private final String INSERT_SQL;
+    private final String UPDATE_SQL;
+
+    JdbcCustomDashboardRepository(@Value("${management.jdbc.prefix:}") String tablePrefix) {
+        super(tablePrefix, "custom_dashboards");
+        INSERT_SQL = buildInsertSql();
+        UPDATE_SQL = buildUpdateSql();
+    }
+
+    @Override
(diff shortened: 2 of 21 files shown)


gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_11_0/06_create_custom_dashboards_table.yml
```diff
@@ -0,0 +1,27 @@
...
+            constraintName: pk_${gravitee_prefix}custom_dashboards
+            columnNames: id
+            tableName: ${gravitee_prefix}custom_dashboards
+        - createIndex:
+            tableName: ${gravitee_prefix}custom_dashboards
+            indexName: idx_${gravitee_prefix}custom_dashboards_organization_id
+            columns:
+              - column:
+                  name: organization_id
+                  type: nvarchar(64)
```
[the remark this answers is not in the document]
a-cordier: Done

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoCustomDashboardRepository.java
```diff
@@ -0,0 +1,131 @@
...
+        var res = mapper.map(createdDashboardMongo);
+        log.debug("Create custom dashboard [{}] - Done", customDashboard.getName());
+        return res;
+    }
+
+    @Override
+    public CustomDashboard update(CustomDashboard customDashboard) throws TechnicalException {
+        if (customDashboard == null || customDashboard.getName() == null) {
+            throw new IllegalStateException("Custom dashboard to update must have a name");
+        }
```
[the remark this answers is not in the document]
a-cordier: Service layer is coming in another PR but 👍

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-analytics.yaml
```diff
@@ -1230,6 +1377,269 @@ components:
...
+                    type: integer
+                    minimum: 0
+                    description: Vertical position (row index) in the grid
+                    example: 0
+            example:
+                cols: 1
+                rows: 1
+                x: 0
+                "y": 0
+
```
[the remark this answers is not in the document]
a-cordier: Done

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/CustomDashboardWidget.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Builder
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @Data
+    public static class Request {
+
+        private String type;
+        private String timeRange;
```
carlos-andres-osorio: just out of curiosity, how will the time range from the request (which has fields `to` and `from`) be mapped to a string?

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-analytics.yaml
```diff
@@ -1230,11 +1371,274 @@ components:
...
+            required:
+                - type
+                - timeRange
+                - metrics
+            description: |
+                Defines the analytics request associated with a widget. This determines what data the widget displays.
+
+                The `type` field selects the analytics request type:
+                - **measures**: Returns aggregated values. Uses `MetricRequest` for metrics.
+                - **facets**: Groups data by facet values. Uses `FacetMetricRequest` for metrics, and supports `by` and `limit` fields.
```
carlos-andres-osorio: will support for time-series be added in a future PR?

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcCustomDashboardRepository.java
```diff
@@ -0,0 +1,225 @@
...
+            }
+            first = false;
+            builder.append(escapeReservedWord(getDBName(column.name)));
+            builder.append(" = ?");
+        }
+        builder.append(", labels = ?, widgets = ? where id = ?");
+        return builder.toString();
+    }
+
+    private <T> T deserialize(String json, TypeReference<T> typeRef) {
```
carlos-andres-osorio, line 173: If data can't be deserialized, the dashboard will be returned with empty widgets/labels. Would it be better (more explicit to the user) to return an error?

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcCustomDashboardRepository.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+        try {
+            return JSON_MAPPER.readValue(json, typeRef);
+        } catch (Exception e) {
+            log.error("Failed to deserialize JSON", e);
+            return null;
+        }
+    }
+
+    private String serialize(Object value) {
```
carlos-andres-osorio, line 184: If data can't be serialized, the dashboard will be saved with empty widgets/labels. Would it be better to return an error?

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcCustomDashboardRepository.java
```diff
@@ -0,0 +1,225 @@
...
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.stereotype.Repository;
+
+/**
+ * @author Antoine CORDIER (antoine.cordier at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+@CustomLog
+@Repository
+public class JdbcCustomDashboardRepository
```
carlos-andres-osorio, line 49: Based on the current scope, do you know if method `findAll()` will be used at any point?

If yes, shouldn't it be overridden here, or dashboards won't be loaded correctly.

gravitee-apim-repository/gravitee-apim-repository-test/src/test/java/io/gravitee/repository/management/CustomDashboardRepositoryTest.java
```diff
@@ -0,0 +1,246 @@
...
+import static io.gravitee.repository.utils.DateUtils.compareDate;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.fail;
+
+import io.gravitee.repository.management.model.CustomDashboard;
+import io.gravitee.repository.management.model.CustomDashboardWidget;
+import java.util.*;
+import org.junit.Test;
+
+public class CustomDashboardRepositoryTest extends AbstractManagementRepositoryTest {
```
carlos-andres-osorio, line 27: Not an issue in these tests, but tests in this package don't follow a consistent naming convention. 

Is there a preference for using camelCase vs. snake_case with   `DisplayNameGeneration`?

gravitee-apim-repository/gravitee-apim-repository-test/src/test/java/io/gravitee/repository/management/CustomDashboardRepositoryTest.java
```diff
@@ -0,0 +1,246 @@
...
+            .build();
+
+        var dashboard = CustomDashboard.builder()
+            .id("new-cd")
+            .organizationId("DEFAULT")
+            .name("New Dashboard")
+            .createdBy("user-new")
+            .createdAt(new Date(1000000000000L))
+            .lastModified(new Date(1111111111111L))
+            .labels(Map.of("env", "prod"))
```
carlos-andres-osorio, line 115: are empty labels/widgets treated in a way that should be tested explicitly?

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/CustomDashboardWidget.java
a-cordier: That's a good point, this was taken from the ts file maybe something is wrong here thought @samirtechlab ?

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcCustomDashboardRepository.java
a-cordier: I think the only method we will use will be findByOrganizationId

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-analytics.yaml
a-cordier: good point, it's already supported, I'll remove this entirely, it should just be an internal link or something

gravitee-apim-repository/gravitee-apim-repository-test/src/test/java/io/gravitee/repository/management/CustomDashboardRepositoryTest.java
a-cordier: all the tests in the repository being junit 4 tests, I followed the convention here, we use DisplayNameGenerator with junit 5

gravitee-apim-repository/gravitee-apim-repository-test/src/test/java/io/gravitee/repository/management/CustomDashboardRepositoryTest.java
a-cordier: Adding a test case for labels, empty/null widgets will be rejected by the service layer

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/CustomDashboardWidget.java
SamHardi: Timerange is an object as you can see [here](https://github.com/gravitee-io/gravitee-api-management/blob/97d4d90f5632df3cc6400cf75d29d9f2b1298677/gravitee-apim-portal-webui-next/projects/gravitee-dashboard/src/lib/components/widget/model/request/request.ts).

`export interface GlobalRequest<E extends RequestType> {
  type: E;
  timeRange?: TimeRange;
  filters?: RequestFilter[];
  metrics: MetricRequest[];
}`

`export interface TimeRange {
  from: string;
  to: string;
}`

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcCustomDashboardRepository.java
a-cordier: It was the pattern in place everywhere, but fair enough, let's throw an illegal state exception :)

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/CustomDashboardWidget.java
a-cordier: done

carlos-andres-osorio (approved):

PR #15100, opened 2026-02-09
Make subscription available in templateEngine for messages

phiz71:
## Issue

https://gravitee.atlassian.net/browse/APIM-12561

## Description

Use latest version of EL lib to improve template engine duplication for messages. By doing this, the subscription is now available in the template engine context of the publish and subscribe phase of message APIs.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/AbstractExecutionContext.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/AbstractExecutionContext.java
@@ -226,11 +226,7 @@ public abstract class AbstractExecutionContext<RQ extends MutableRequest, RS ext
 
     @Override
     public TemplateEngine getTemplateEngine(Message message) {
-        final TemplateEngine engine = TemplateEngine.templateEngine();
-        prepareTemplateEngine(engine);
-        if (templateVariableProviders != null) {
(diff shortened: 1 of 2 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/AbstractExecutionContext.java
```diff
@@ -226,11 +226,7 @@ public TemplateEngine getTemplateEngine() {
 
     @Override
     public TemplateEngine getTemplateEngine(Message message) {
-        final TemplateEngine engine = TemplateEngine.templateEngine();
-        prepareTemplateEngine(engine);
-        if (templateVariableProviders != null) {
-            templateVariableProviders.forEach(templateVariableProvider -> templateVariableProvider.provide(engine.getTemplateContext()));
-        }
+        final TemplateEngine engine = TemplateEngine.fromTemplateEngine(this.templateEngine);
```
jhaeyaert: Should it be `TemplateEngine.fromTemplateEngine(this.getTemplateEngine());` instead to avoid NPE (unlikely improbable but I just saw the TU and wondering if it could be done like this)?

phiz71: hum yes, that would be safer. I'll do that

jhaeyaert (approved):

PR #15105, opened 2026-02-10
feat(portal): Update Subscription Table and Filters

aqibmohammadkhan:
## Issue

https://gravitee.atlassian.net/browse/APIM-12676

## Description

Description

This PR introduces several enhancements to the Subscriptions management in the Portal, including support for multi-select filters, URL synchronization for deep linking, and a more robust subscription table experience.

1. Backend Changes (Rest API)

- Multiple ID Filtering: Updated SubscriptionsResource to support filtering by multiple apiId and applicationId parameters.
- Enhanced Permission Checks: Added verification to ensure the user has READ permissions for all specified application IDs in the filter.
- 
- Improved Metadata Handling: Refined the buildSubscriptionMetadata method to provide necessary API, Application, and Plan details for the frontend.

2. Frontend Changes (Portal WebUI Next)

- Multi-Select Filters: Replaced standard single-select dropdowns with a new DropdownSearchComponent that supports searching and multi-selection for APIs and Applications.
- 
- URL Synchronization: Filter states (APIs, Applications, Statuses) and pagination parameters are now synchronized with the URL query parameters.
- This allows for:
- Better deep linking to specific filtered views.
- Consistent state when using the browser's back/forward buttons.
- 
- Improved Subscription Table:
- Displays more descriptive names for APIs, Plans, and Applications using metadata.
- Added support for dynamic page size and current page navigation.
- 
- Shared Components:
- DropdownSearchComponent: A reusable component for paginated, searchable multi-select lists.
- DropdownSearchOverlayComponent: Handles the overlay logic and search interaction for the dropdown.

3. Verification Results

- Automated Tests: 
- Updated SubscriptionsComponent tests to verify filter initialization from URL and state updates.
- Added comprehensive unit tests for DropdownSearchComponent to ensure search, selection, and pagination work correctly.

- Excluded from this PR
  - Pagination component does not have 'prev' and 'next' as per figma
  - subsciption resource discussion on 401 for no read permissions
  - fliter selected icon - more round than oval
  - gap between search and list in the dropdown.

https://github.com/user-attachments/assets/74082fd2-3ea0-4b48-b020-d3f8d6313144

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/api/subscribe-to-api/subscribe-to-api.component.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/app/api/subscribe-to-api/subscribe-to-api.component.spec.ts
@@ -20,7 +20,7 @@ import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { MatButtonHarness } from '@angular/material/button/testing';
 import { MatChipHarness } from '@angular/material/chips/testing';
 import { By } from '@angular/platform-browser';
-import { of } from 'rxjs/internal/observable/of';
+import { of } from 'rxjs';
 
 import { TermsAndConditionsDialogHarness } from './components/terms-and-conditions-dialog/terms-and-conditions-dialog.harness';
 import { SubscribeToApiCheckoutHarness } from './subscribe-to-api-checkout/subscribe-to-api-checkout.harness';
@@ -37,9 +37,13 @@ import { fakeApplication, fakeApplicationsResponse } from '../../../entities/app
 import { Page } from '../../../entities/page/page';
 import { fakePage } from '../../../entities/page/page.fixtures';
 import { fakePlan } from '../../../entities/plan/plan.fixture';
-import { CreateSubscription, Subscription } from '../../../entities/subscription/subscription';
-import { fakeSubscription, fakeSubscriptionResponse } from '../../../entities/subscription/subscription.fixture';
-import { SubscriptionsResponse } from '../../../entities/subscription/subscriptions-response';
+import {
+  CreateSubscription,
+  fakeSubscription,
+  fakeSubscriptionResponse,
+  Subscription,
+  SubscriptionsResponse,
+} from '../../../entities/subscription';
 import { ConfigService } from '../../../services/config.service';
 import { AppTestingModule, TESTING_BASE_URL } from '../../../testing/app-testing.module';
 
@@ -104,8 +108,17 @@ describe('SubscribeToApiComponent', () => {
       data: [
         fakePlan({ id: KEYLESS_PLAN_ID, security: 'KEY_LESS' }),
         fakePlan({ id: API_KEY_PLAN_ID, security: 'API_KEY', comment_required: false, general_conditions: undefined }),
-        fakePlan({ id: API_KEY_PLAN_ID_COMMENT_REQUIRED, security: 'API_KEY', comment_required: true, general_conditions: undefined }),
-        fakePlan({ id: API_KEY_PLAN_ID_GENERAL_CONDITIONS, security: 'API_KEY', general_conditions: GENERAL_CONDITIONS_ID }),
+        fakePlan({
+          id: API_KEY_PLAN_ID_COMMENT_REQUIRED,
+          security: 'API_KEY',
+          comment_required: true,
+          general_conditions: undefined,
+        }),
+        fakePlan({
+          id: API_KEY_PLAN_ID_GENERAL_CONDITIONS,
+          security: 'API_KEY',
+          general_conditions: GENERAL_CONDITIONS_ID,
+        }),
         fakePlan({ id: OAUTH2_PLAN_ID, security: 'OAUTH2', general_conditions: undefined }),
         fakePlan({ id: JWT_PLAN_ID, security: 'JWT', general_conditions: undefined }),
         fakePlan({ id: MTLS_PLAN_ID, security: 'MTLS', general_conditions: undefined }),
@@ -120,7 +133,6 @@ describe('SubscribeToApiComponent', () => {
   });
 
   describe('User subscribes to Push plan', () => {
-    beforeEach(async () => {});
     describe('selects push plan', () => {
       it('should be able to go to consumer configuration form', async () => {
         await init(true);
@@ -196,7 +208,6 @@ describe('SubscribeToApiComponent', () => {
   });
 
   describe('User subscribes to Keyless plan', () => {
-    beforeEach(async () => {});
     describe('Step 1 -- Choose a plan', () => {
       it('should be able to go to step 3 once plan chosen', async () => {
         await init(true);
@@ -217,8 +228,6 @@ describe('SubscribeToApiComponent', () => {
       });
     });
     describe('Step 3 -- Checkout', () => {
-      beforeEach(async () => {});
-
       describe('V4 Proxy', () => {
         beforeEach(async () => {
           await init(true, fakeApi({ id: API_ID, type: 'PROXY', definitionVersion: 'V4', entrypoints: [ENTRYPOINT] }));
@@ -243,7 +252,15 @@ describe('SubscribeToApiComponent', () => {
       describe('V4 Native', () => {
         const KAFKA_ENTRYPOINT = 'my.kafka.entrypoint:9092';
         beforeEach(async () => {
-          await init(true, fakeApi({ id: API_ID, type: 'NATIVE', definitionVersion: 'V4', entrypoints: [KAFKA_ENTRYPOINT] }));
+          await init(
+            true,
+            fakeApi({
+              id: API_ID,
+              type: 'NATIVE',
+              definitionVersion: 'V4',
+              entrypoints: [KAFKA_ENTRYPOINT],
+            }),
+          );
 
           const step1 = await harnessLoader.getHarness(SubscribeToApiChoosePlanHarness);
           await step1.selectPlanByPlanId(KEYLESS_PLAN_ID);
@@ -474,7 +491,11 @@ describe('SubscribeToApiComponent', () => {
           expect(await subscribeButton?.isDisabled()).toEqual(false);
           await subscribeButton?.click();
 
-          expectPostCreateSubscription({ plan: API_KEY_PLAN_ID_COMMENT_REQUIRED, application: 'app-id', request: 'My new message' });
+          expectPostCreateSubscription({
+            plan: API_KEY_PLAN_ID_COMMENT_REQUIRED,
+            application: 'app-id',
+            request: 'My new message',
+          });
         });
       });
       describe('When terms and conditions need to be accepted', () => {
@@ -531,7 +552,13 @@ describe('SubscribeToApiComponent', () => {
             await selectPlan(API_KEY_PLAN_ID);
             await selectApplication(APP_ID_NO_SUBSCRIPTIONS);
 
-            expectGetSubscriptionsForApplication(APP_ID_NO_SUBSCRIPTIONS, fakeSubscriptionResponse({ data: [], metadata: {} }));
+            expectGetSubscriptionsForApplication(
+              APP_ID_NO_SUBSCRIPTIONS,
+              fakeSubscriptionResponse({
+                data: [],
+                metadata: {},
+              }),
+            );
             fixture.detectChanges();
           });
           it('should NOT show api key mode choice', async () => {
@@ -1144,7 +1171,15 @@ describe('SubscribeToApiComponent', () => {
 
   function expectGetSubscriptionsForApi(apiId: string, subscriptions: SubscriptionsResponse = fakeSubscriptionResponse({ data: [] })) {
     httpTestingController
-      .expectOne(`${TESTING_BASE_URL}/subscriptions?apiId=${apiId}&statuses=PENDING&statuses=ACCEPTED&size=-1`)
+      .expectOne(req => {
+        return (
+          req.url === `${TESTING_BASE_URL}/subscriptions` &&
+          req.params.get('apiId') === apiId &&
+          req.params.getAll('statuses')?.includes('PENDING') === true &&
+          req.params.getAll('statuses')?.includes('ACCEPTED') === true &&
+          req.params.get('size') === '-1'
+        );
+      })
       .flush(subscriptions);
   }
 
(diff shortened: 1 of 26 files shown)


gravitee-apim-portal-webui-next/projects/gravitee-dashboard/src/lib/components/filter/dropdown-search/dropdown-search.component.ts
jourdiw, line 1: Since we are creating a new component, these changes can be 🌬️

gravitee-apim-portal-webui-next/src/app/api/subscribe-to-api/subscribe-to-api.component.spec.ts
jourdiw, line 1: How are the tests passing using `applicationId` and `apiId` when the component is using the `applicationIds` and `apiIds` query parameters?

gravitee-apim-portal-webui-next/src/app/dashboard/subscriptions/subscriptions.component.harness.ts
```diff
@@ -57,39 +53,7 @@ export class SubscriptionsComponentHarness extends ComponentHarness {
...
-        const text = await option.getText();
-        if (values.includes(text)) {
-          await option.click();
-        }
-      }
-    }
-  }
+  // TODO: Implement filter harness methods for DropdownSearchComponent
+  // The filters now use app-dropdown-search (CDK overlay-based) instead of MatSelect.
+  // These methods need to be reimplemented to interact with the new overlay component.
```
jourdiw: If you have created the component and implemented it in a parent component, I would have expected the harness to be used as well by the parent in tests 🤔 WDYT about creating the harness in this PR?

gravitee-apim-portal-webui-next/src/app/dashboard/subscriptions/subscriptions.component.spec.ts
jourdiw, line 1: Best practices for testing is to use harnesses instead of directly testing components 👍

gravitee-apim-portal-webui-next/src/app/dashboard/subscriptions/subscriptions.component.ts
jourdiw, line 1: It seems that you are initializing the filters initially from the url, however the url is not getting updated when you are changing filters or pages.

I would suggest using the router as the source of truth: the form values are updated from the url, when the user changes the filter value, page, or page size, the router is then updated with the new values.

This will prevent buggy behavior --> when a user has applied filters, clicks a subscription, and then clicks back, the state will not be reset to the last view but rather to the initial view.

gravitee-apim-portal-webui-next/src/app/dashboard/dashboard.component.spec.ts
jourdiw, line 1: In tests, I strongly recommend using `httpTestingController` instead of mocking services. It will guarantee that the backend is called as expected with the expected number of times.

gravitee-apim-portal-webui-next/src/app/dashboard/dashboard.component.ts
```diff
@@ -44,7 +44,8 @@ export class DashboardComponent {
     this.router.events.pipe(
       filter(event => event instanceof NavigationEnd),
       map(() => {
-        const segments = this.router.url.split('/').filter(Boolean);
+        const urlTree = this.router.parseUrl(this.router.url);
+        const segments = urlTree.root.children['primary']?.segments.map(s => s.path) ?? [];
```
jourdiw: What problem were you encountering that you applied this fix?

gravitee-apim-portal-webui-next/src/components/dropdown-search/dropdown-search.component.scss
```diff
@@ -0,0 +1,80 @@
...
+      margin-left: 8px;
+      color: app-theme.$paragraph-color;
+      transition:
+        color 0.2s ease,
+        transform 0.2s ease;
+
+      // &--opened {
+      //  color: app-theme.$primary-main-color;
+      //  transform: rotate(180deg);
+      // }
```
jourdiw: 🗑️

gravitee-apim-portal-webui-next/src/components/dropdown-search/dropdown-search.component.scss
```diff
@@ -0,0 +1,80 @@
...
+      outline: none;
+    }
+
+    &--disabled {
+      cursor: not-allowed;
+      opacity: 0.5;
+    }
+
+    &--opened {
+      padding: 7px 11px;
```
jourdiw: Best practice is to avoid odd-numbered pixels. Is this odd-numbered padding absolutely necessary?

gravitee-apim-portal-webui-next/src/components/dropdown-search/dropdown-search.component.scss
```diff
@@ -0,0 +1,80 @@
...
+      min-width: 24px;
+      height: 24px;
+      align-items: center;
+      justify-content: center;
+      padding: 0 6px;
+      border-radius: 12px;
+      background-color: app-theme.$primary-main-color;
+      color: app-theme.$default-text-contrast-color;
+      font-size: 0.875rem;
+      font-weight: 500;
```
jourdiw: Can you use a theme font instead of setting the weight here?

gravitee-apim-portal-webui-next/src/components/dropdown-search/dropdown-search.component.spec.ts
jourdiw, line 1: Since this is part of a reactive form, I would strongly suggest including tests with the component integrated in a form (such as a fake TestComponent) and that the states are handled as expected.
The harness would be helpful here, too 😉

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionsResource.java
```diff
@@ -153,25 +153,31 @@ public Response createSubscription(@Valid @NotNull(message = "Input must not be
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     public Response getSubscriptions(
-        @QueryParam("apiId") String apiId,
-        @QueryParam("applicationId") String applicationId,
+        @QueryParam("apiId") List<String> apiIds,
+        @QueryParam("applicationId") List<String> applicationIds,
```
jourdiw: As discussed, this will need to be a team decision to choose this solution 👍

jourdiw (changes requested):

gravitee-apim-portal-webui-next/src/app/dashboard/dashboard.component.ts
aqibmohammadkhan: as discussed, updated the fix.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionsResource.java
aqibmohammadkhan: as discussed we will be going with 

- Add new plural filters to the available query params & deprecate old parameters:

gravitee-apim-portal-webui-next/src/app/dashboard/subscriptions/subscriptions.component.html
jourdiw, line 1: The search placeholder can all be plural 🐍

gravitee-apim-portal-webui-next/src/app/dashboard/subscriptions/subscriptions.component.ts
```diff
@@ -58,123 +67,250 @@ export default class SubscriptionsComponent {
...
-        applicationId?: string;
-        statuses: SubscriptionStatusEnum[] | null;
-        size?: number;
-        page?: number;
-      };
-    }) => this.subscriptionService.list(params.params),
+    };
   });
 
+  filters = computed(() => this.urlDerivedFilters() ?? this.formDerivedFilters());
```
jourdiw: Use a single source of truth: the filters should be directly linked to the router --> the router params should be parsed in order to give the value to the filters.
This way, there are no race conditions for the value of the filters.

gravitee-apim-portal-webui-next/src/app/dashboard/subscriptions/subscriptions.component.ts
```diff
@@ -58,123 +67,250 @@ export default class SubscriptionsComponent {
...
-    const response = this.subscriptionsResource.value() as SubscriptionsResponse | undefined;
+    const response = this.subscriptionsList();
     return response?.metadata?.['paginateMetaData']?.totalElements ?? response?.data?.length ?? 0;
   });
 
+  hasSubscriptions = computed(() => {
+    const response = this.subscriptionsList();
+    const apiIds = this.apiFilterSignal();
+    const applicationIds = this.applicationFilterSignal();
+    const statuses = this.statusFilterSignal();
```
jourdiw: If `filters` becomes the source of truth (because it is only using the url as a source of truth), then we can rather rely on `filters` instead of the form controls for value checks. 👍

gravitee-apim-portal-webui-next/src/app/dashboard/subscriptions/subscriptions.component.ts
```diff
@@ -58,123 +67,250 @@ export default class SubscriptionsComponent {
...
+            size: filters.size,
+          },
+          queryParamsHandling: 'merge',
+          replaceUrl: true,
+        });
+      });
+
+    this.activatedRoute.queryParams.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(params => {
+      this.applyQueryParamsToForm(params);
+    });
```
jourdiw:
```suggestion
    this.filters: Signal<Filters> = toSignal(this.activatedRoute.queryParams.pipe(
        map((params) => this.mapQueryParamsToFilters(params))
    ));
    
    function mapQueryParamsToFilters(params: Record<string, unknown>) {
        const apiIds = parseArrayParam(params['apiIds']);
    
        const applicationIds = parseArrayParam(params['applicationIds']);
    
        const statuses = this.parseStatusParam(params['statuses']);
    
        const page = parsePageParam(params['page'], SubscriptionsComponent.DEFAULT_PAGE);
        const size = parseSizeParam(params['size'], SubscriptionsComponent.DEFAULT_PAGE_SIZE, SubscriptionsComponent.MAX_PAGE_SIZE);
    
        return {
          apiIds: apiIds.length ? apiIds : undefined,
          applicationIds: applicationIds.length ? applicationIds : undefined,
          statuses: statuses.length ? statuses : null,
          page,
          size,
        };
  }
  
  constructor() {
    effect(() => {
        // When `this.filters()` changes in value, then the apiFilters, applicationFilter, statusFilter, page, and sie are updated
    });
  }
```

jourdiw: Example blocks... please feel free to take and throw away what you want. The goal is to not set variables directly from local changes, but rather have the url drive all value changes everywhere, page and size included 👍

gravitee-apim-portal-webui-next/src/components/dropdown-search/dropdown-search-overlay/dropdown-search-overlay.harness.ts
jourdiw, line 1: This can be combined with the parent to help with developer experience 👍

gravitee-apim-portal-webui-next/src/utils/common.utils.ts
```diff
@@ -0,0 +1,81 @@
...
+ */
+export function toTitleCase(value: string): string {
+  return value.charAt(0).toUpperCase() + value.slice(1).toLowerCase();
+}
+
+/**
+ * Compares two filter objects for equality.
+ * Compares arrays and primitive values within the filter objects.
+ */
+export function areFiltersEqual<T extends Record<string, unknown>>(prev: T, curr: T): boolean {
```
jourdiw, line 69: WDYT about having this be `areParamsEqual`?

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionsResource.java
```diff
@@ -153,25 +153,36 @@ public Response createSubscription(@Valid @NotNull(message = "Input must not be
...
             final Set<ApplicationListItem> applications = applicationService.findByUser(executionContext, getAuthenticatedUser());
             if (applications == null || applications.isEmpty()) {
                 return createListResponse(executionContext, emptyList(), paginationParam, paginationParam.hasPagination());
             }
             query.setApplications(applications.stream().map(ApplicationListItem::getId).collect(toSet()));
-        } else if (!hasPermission(executionContext, RolePermission.APPLICATION_SUBSCRIPTION, applicationId, RolePermissionAction.READ)) {
-            throw new ForbiddenAccessException();
+        } else {
+            for (String appId : effectiveApplicationIds) {
+                if (!hasPermission(executionContext, RolePermission.APPLICATION_SUBSCRIPTION, appId, RolePermissionAction.READ)) {
```
jourdiw, line 185: WDYT that since this is querying, if a user does not have Read permissions on an application, that the app id is just filtered from the list?
Throwing an error might be a bit harsh 🤔 
Let's ask the team for their input 👍

jourdiw (commented):
Some tests for the resource would be great, especially for edge cases that we won't necessarily run into in the UI, but users could run into when using the backend directly (aka. permissions, applications not found, etc).

gravitee-apim-portal-webui-next/src/utils/common.utils.ts
aqibmohammadkhan: for later PR.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionsResource.java
aqibmohammadkhan: for later PR

jourdiw (approved):

PR #15126, opened 2026-02-10
fix: check env permissions on integration endpoints

michel-barret:
## Issue

https://gravitee.atlassian.net/browse/APIM-12697


gravitee-apim-repository/gravitee-apim-repository-mongodb/src/test/java/io/gravitee/repository/mongodb/management/internal/integration/IntegrationMongoRepositoryImplTest.java
```diff
@@ -85,6 +85,39 @@ void findAllByEnvironmentIdAndGroups() {
         );
     }
 
+    @Test
+    @SneakyThrows
+    void findAllByEnvironmentIdAndGroups2() {
```
jgiovaresco: It would be better to add a test in the repository test to have this case cover by Mongo & JDBC

michel-barret: added here https://github.com/gravitee-io/gravitee-api-management/pull/15126/changes#diff-9905a310b92f19bd6cd3ca67d3d7fc9bf925a5608d23ffdb2f0fb0ab5fb69a60R125

jgiovaresco (approved):

PR #15134, opened 2026-02-10
feat: add apiId to environment logs response

carlos-andres-osorio:
## Issue

https://gravitee.atlassian.net/browse/GKO-2328

## Description

Add the `apiId` field to the environment-level logs response so consumers can identify which API each log entry belongs to.

Changes:
- **Inline `ApiLog` schema** in `openapi-logs.yaml` — replaced the `$ref` to `openapi-apis.yaml` with a local schema definition, adding the new `apiId` string property
- **Add `String apiId`** to the `ApiLog` Java record (core model)
- **Update `mapApiLog()`** in `SearchEnvironmentLogsUseCase` to map `BaseConnectionLog.apiId` to the response
- **Update tests** to verify the `apiId` mapping

## Additional context

The source data model (`BaseConnectionLog`) already had the `apiId` field — this change surfaces it through the OpenAPI contract and the mapping layer.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-logs.yaml
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-logs.yaml
@@ -154,11 +154,91 @@ components:
         data:
           type: array
           items:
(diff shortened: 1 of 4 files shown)


paulatulis (commented):
This correctly adds what I will need in order to connect the dots and present the log detail view. I'd like someone else to check over to make sure everything looks good from a purely technical standpoint. Thanks, Andres!

a-cordier (commented):
commit should be rephrased to refactor IMO

a-cordier (approved):

PR #15154, opened 2026-02-11
Compute Original URL in debug mode

phiz71:
## Issue

https://gravitee.atlassian.net/browse/APIM-12705

## Description

Since 4.10.x and MCP Proxy, oauth2 policy implements a new method "onWellKnown".
This method requires that the attribute `gravitee.attribute.request.original-url` is set.

In the default reactor, it's the XForwardedProcessor which is responsible for computing this URL. However, the `DebugPlatformProcessorChainFactory` is created by default without this processor.

This fix adds an init debug processor to compute a simple original URL

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-debug/src/main/java/io/gravitee/gateway/reactive/debug/reactor/processor/DebugInitProcessor.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-debug/src/main/java/io/gravitee/gateway/reactive/debug/reactor/processor/DebugInitProcessor.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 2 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-debug/src/main/java/io/gravitee/gateway/reactive/debug/reactor/processor/DebugInitProcessor.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(DebugInitProcessor.class);
+
+    @Override
+    public String getId() {
+        return "processor-debug-init";
+    }
+
+    @Override
+    public Completable execute(final HttpExecutionContextInternal ctx) {
+        return Completable.defer(() -> {
```
jhaeyaert: You can use Completable.fromRunnable I think.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-debug/src/main/java/io/gravitee/gateway/reactive/debug/reactor/processor/DebugInitProcessor.java
```diff
@@ -0,0 +1,54 @@
...
+            String scheme = request.scheme();
+            String originalHost = request.originalHost();
+            String uri = request.uri();
+
+            String originalUrl = scheme + "://" + originalHost + uri;
+            LOGGER.debug("Original URL: {}", originalUrl);
+
+            ctx.setAttribute(ContextAttributes.ATTR_REQUEST_ORIGINAL_URL, originalUrl);
+            return Completable.complete();
+        }).subscribeOn(Schedulers.io());
```
jhaeyaert: I don't think it is necessary to switch to Schedulers.io() here as it is just about setting an attribute.

jhaeyaert (changes requested):

aditya-goyal01 (approved):

jhaeyaert (approved):

PR #15161, opened 2026-02-12
fix: replace hardcoded SVG icons with a dynamic icon retrieval function and remove root API create

dev-marek:
## Issue

https://gravitee.atlassian.net/browse/APIM-12541

## Description

1. Removed possibility to create root item of API type
2. Updated API icons on list.

### DIFF
--- a/gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.html
+++ b/gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.html
@@ -54,7 +54,7 @@
       [class.selected]="isSelected(node)"
     >
       <div *cdkDragPreview class="tree__drag-preview">
-        <mat-icon [svgIcon]="'gio:' + (node.type | lowercase)"></mat-icon>
+        <mat-icon [svgIcon]="node.type | portalNavigationItemIcon"></mat-icon>
         {{ node.label }}
       </div>
       <div *cdkDragPlaceholder class="tree__drag-placeholder"></div>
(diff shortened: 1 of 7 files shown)


gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.html
```diff
@@ -149,24 +149,22 @@
         }
       }
 
-      @if (canUpdate) {
+      @if (canUpdate && node.type !== 'API') {
```
jourdiw, line 152: Shouldn't we be able to edit an API? At least for the visibility of Public/Private?
Or perhaps that will be covered in a different ticket?

gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.html
jourdiw, line 1: Seeing as this function is used multiple times for the mat-icon, WDYT about creating a dedicated, simple component (or ngTemplate) that maps the node.type to the correct gio icon? 
This could help centralize the logic rather than having a separate ts file with the mapping and using a function in the template.

kovaren: Or maybe create a pipe, it could be a drop-in replacement for `lowercase`

gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.html
dev-marek: Yup, next ticket.

kovaren (approved):

PR #15183, opened 2026-02-13
fix: reference fields updates for subscription

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-12490

## Description

fix reference id and type in subscription

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
@@ -54,7 +54,9 @@ import io.gravitee.repository.management.api.SubscriptionRepository;
 import io.gravitee.repository.management.api.search.Order;
 import io.gravitee.repository.management.api.search.SubscriptionCriteria;
 import io.gravitee.repository.management.api.search.builder.PageableBuilder;
+import io.gravitee.repository.management.apiproducts.ApiProductsRepository;
 import io.gravitee.repository.management.model.ApiKey;
+import io.gravitee.repository.management.model.ApiProduct;
 import io.gravitee.repository.management.model.ApplicationStatus;
 import io.gravitee.repository.management.model.ApplicationType;
 import io.gravitee.repository.management.model.Audit;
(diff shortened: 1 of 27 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
```diff
@@ -264,6 +272,73 @@ public Set<SubscriptionEntity> findByIdIn(Collection<String> subscriptionIds) {
...
+
+        final Map<String, GenericApiEntity> apisById = apiIds.isEmpty()
+            ? Collections.emptyMap()
+            : apiSearchService
+                .findGenericByEnvironmentAndIdIn(executionContext, apiIds)
+                .stream()
+                .collect(toMap(GenericApiEntity::getId, Function.identity()));
+        final Map<String, ApiProduct> apiProductsById = apiProductIds
+            .stream()
+            .flatMap(id -> {
```
phiz71:
```suggestion
        final Map<String, ApiProduct> apiProductsById = apiProductIds.isEmpty()
            ? Collection.emptyMap()
            : apiProductIds
                  .stream()
                  .flatMap(id -> {
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
```diff
@@ -264,6 +272,73 @@ public Set<SubscriptionEntity> findByIdIn(Collection<String> subscriptionIds) {
...
+                try {
+                    return apiProductsRepository.findById(id).stream();
```
phiz71: Don't you have a findByreferenceIdsAndReferenceTypeAndEnvironment ? that could be used here ?

phiz71: It's ok, I mixed with PlanRepository

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
vikrantgravitee: Updated!

phiz71 (approved):

PR #15213, opened 2026-02-16
fix(api-product): populate subscription cache before product deploy event

arpzilla-dev:
## Issue

https://gravitee.atlassian.net/browse/APIM-12801

## Description

Fixes 401 when calling a newly added API in an API Product with an existing subscription key. Subscriptions and API keys are now refreshed before the product event is emitted, so the security chain finds them in cache.

Also unregisters subscriptions and API keys when an API is removed from a product or when the product is deleted, so removed APIs no longer accept the product's keys.

## Additional context

working video :
via script : 

https://github.com/user-attachments/assets/ede70ec6-802b-4d8b-92c3-0f299b8bc0b5

via postman :

https://github.com/user-attachments/assets/74f0e448-e71d-4e5a-8955-298d05f96d44

remove api from api product :

https://github.com/user-attachments/assets/9e666a9e-47a8-4224-b205-fce5a24f5d47

### Why this bug arose

`ApiProductChangedEvent` was emitted before the subscription/API key cache was populated. The security chain reacted to the event and refreshed before the cache, so lookups failed and returned 401.

### What we changed

- **ApiProductManager**: Added `register(product, beforeEmit)` so a callback runs after the product is in the registry but before the event is emitted.
- **ApiProductDeployer**: Uses the callback to run `ApiProductSubscriptionRefresher` (load subscriptions, explode to APIs, deploy to `SubscriptionService` and `ApiKeyService`) before the event.
- **ApiProductSubscriptionRefresher**: Loads subscriptions for product plans, explodes them via `SubscriptionMapper`, and registers them in both caches.
- **DeployerFactory**: Injects `ApiProductSubscriptionRefresher`; `createApiProductDeployer()` no longer takes parameters.

image

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiProductManagerImpl.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiProductManagerImpl.java
@@ -22,6 +22,7 @@ import io.gravitee.gateway.handlers.api.event.ApiProductEventType;
 import io.gravitee.gateway.handlers.api.manager.ApiProductManager;
 import io.gravitee.gateway.handlers.api.registry.ApiProductRegistry;
 import io.gravitee.node.api.license.LicenseManager;
+import io.reactivex.rxjava3.core.Completable;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Objects;
@@ -48,7 +49,12 @@ public class ApiProductManagerImpl implements ApiProductManager {
 
     @Override
     public void register(ReactableApiProduct apiProduct) {
-        register(apiProduct, false);
+        register(apiProduct, Completable.complete()).blockingAwait();
+    }
+
+    @Override
+    public Completable register(ReactableApiProduct apiProduct, Completable doBeforeEmit) {
+        return Completable.defer(() -> register(apiProduct, false, doBeforeEmit));
     }
 
     @Override
@@ -66,7 +72,7 @@ public class ApiProductManagerImpl implements ApiProductManager {
         return apiProducts.values();
     }
 
-    private boolean register(ReactableApiProduct apiProduct, boolean force) {
+    private Completable register(ReactableApiProduct apiProduct, boolean force, Completable doBeforeEmit) {
         // License check: API Products require universe tier
         var license = licenseManager.getOrganizationLicenseOrPlatform(apiProduct.getOrganizationId());
         if (!Objects.equals(license.getTier(), "universe")) {
@@ -74,7 +80,7 @@ public class ApiProductManagerImpl implements ApiProductManager {
                 "The API Product [{}] can not be deployed because it is not allowed by the current license (universe tier)",
                 apiProduct.getName()
             );
-            return false;
+            return Completable.complete();
         }
 
         // Get currently deployed API Product
@@ -89,41 +95,41 @@ public class ApiProductManagerImpl implements ApiProductManager {
 
         // Deploy new API Product
         if (apiProductToDeploy) {
-            deploy(apiProduct);
-            return true;
+            return deploy(apiProduct, doBeforeEmit);
         }
 
         // Update existing API Product
         if (apiProductToUpdate) {
-            update(apiProduct);
-            return true;
+            return update(apiProduct, doBeforeEmit);
         }
 
-        return false;
+        return Completable.complete();
     }
 
-    private void deploy(ReactableApiProduct apiProduct) {
+    private Completable deploy(ReactableApiProduct apiProduct, Completable doBeforeEmit) {
         log.debug("Deploying API Product [{}]", apiProduct.getId());
 
         apiProducts.put(apiProduct.getId(), apiProduct);
         apiProductRegistry.register(apiProduct);
 
-        // Emit DEPLOY event after successful registration
-        emitApiProductChangedEvent(ApiProductEventType.DEPLOY, apiProduct);
-
-        log.info("API Product [{}] has been deployed", apiProduct.getId());
+        Completable emit = Completable.fromRunnable(() -> {
+            emitApiProductChangedEvent(ApiProductEventType.DEPLOY, apiProduct);
+            log.info("API Product [{}] has been deployed", apiProduct.getId());
+        });
+        return (doBeforeEmit != null ? doBeforeEmit : Completable.complete()).andThen(emit);
     }
 
-    private void update(ReactableApiProduct apiProduct) {
+    private Completable update(ReactableApiProduct apiProduct, Completable doBeforeEmit) {
         log.debug("Updating API Product [{}]", apiProduct.getId());
 
         apiProducts.put(apiProduct.getId(), apiProduct);
         apiProductRegistry.register(apiProduct);
 
-        // Emit UPDATE event after successful registration
-        emitApiProductChangedEvent(ApiProductEventType.UPDATE, apiProduct);
-
-        log.info("API Product [{}] has been updated", apiProduct.getId());
+        Completable emit = Completable.fromRunnable(() -> {
+            emitApiProductChangedEvent(ApiProductEventType.UPDATE, apiProduct);
+            log.info("API Product [{}] has been updated", apiProduct.getId());
+        });
+        return (doBeforeEmit != null ? doBeforeEmit : Completable.complete()).andThen(emit);
     }
 
     private void undeploy(String apiProductId) {
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
@@ -23,6 +23,8 @@ import io.gravitee.gateway.services.sync.process.distributed.service.Distributed
 import io.gravitee.gateway.services.sync.process.repository.service.PlanService;
 import io.gravitee.gateway.services.sync.process.repository.synchronizer.apiproduct.ApiProductReactorDeployable;
 import io.reactivex.rxjava3.core.Completable;
+import java.util.Set;
+import java.util.stream.Collectors;
 import lombok.CustomLog;
 import lombok.RequiredArgsConstructor;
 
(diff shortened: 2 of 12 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductSubscriptionRefresher.java
```diff
@@ -0,0 +1,164 @@
...
+                // Load and deploy API keys
+                List<ApiKey> apiKeys = loadApiKeys(subscriptions, environments);
+                log.debug("Loaded {} API keys for subscriptions", apiKeys.size());
+
+                apiKeys.forEach(apiKey -> {
+                    try {
+                        apiKeyService.register(apiKey);
+                        log.debug("Deployed API key [{}] for api [{}]", apiKey.getKey(), apiKey.getApi());
+                    } catch (Exception e) {
+                        log.warn("Failed to deploy API key [{}]", apiKey.getKey(), e);
```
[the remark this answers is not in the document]
arpzilla-dev: removed, done

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductSubscriptionRefresher.java
```diff
@@ -0,0 +1,164 @@
...
+                });
+
+                // Load and deploy API keys
+                List<ApiKey> apiKeys = loadApiKeys(subscriptions, environments);
+                log.debug("Loaded {} API keys for subscriptions", apiKeys.size());
+
+                apiKeys.forEach(apiKey -> {
+                    try {
+                        apiKeyService.register(apiKey);
+                        log.debug("Deployed API key [{}] for api [{}]", apiKey.getKey(), apiKey.getApi());
```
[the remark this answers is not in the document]
arpzilla-dev: removed, done

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiProductManagerImpl.java
```diff
@@ -48,7 +48,12 @@ public ApiProductManagerImpl(ApiProductRegistry apiProductRegistry, EventManager
 
     @Override
     public void register(ReactableApiProduct apiProduct) {
-        register(apiProduct, false);
+        register(apiProduct, null);
```
ytvnr: maybe we can just pass an No Op Runnable ? `() -> {}`
(I try to avoid more and more nullity, but it is not a blocker for the review)

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiProductManagerImpl.java
```diff
@@ -89,37 +94,45 @@ private boolean register(ReactableApiProduct apiProduct, boolean force) {
...
-            update(apiProduct);
+            update(apiProduct, beforeEmit);
             return true;
         }
 
         return false;
     }
 
-    private void deploy(ReactableApiProduct apiProduct) {
+    private void deploy(ReactableApiProduct apiProduct, Runnable beforeEmit) {
```
ytvnr: I am not totally convinced about the name, but at the same time, I have no good idea
- beforeEmitRunnable ?
- doBeforeEmit ?

```suggestion
    private void deploy(ReactableApiProduct apiProduct, Runnable beforeEmitRunnable) {
```

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
```diff
@@ -47,10 +49,13 @@ public Completable deploy(ApiProductReactorDeployable deployable) {
...
-                apiProductManager.register(reactableApiProduct);
-
-                // Register API Product in PlanService for subscription validation
-                registerApiProductPlans(deployable);
+                // Use beforeEmit callback so subscription refresh runs before APIs react to the event.
+                apiProductManager.register(reactableApiProduct, () -> {
+                    registerApiProductPlans(deployable);
+                    subscriptionRefresher
+                        .refresh(deployable.subscribablePlans(), Set.of(reactableApiProduct.getEnvironmentId()))
+                        .blockingAwait();
```
ytvnr: I think we have an issue here. You are blocking the thread to do the refresh and it will create problems.

Consider passing a `Completable` instead of a `Runnable` as a parameter in order to properly chain things

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiProductManagerImpl.java
ytvnr: ‼️ See my comment on `ApiProductDeployer`, I think Runnable is not the good choice

ytvnr (changes requested):

ytvnr (commented):
On a more general note, this implementation looks a bit like a patch.
The way you manage cache implies a bit too much coupling between subscription and APIs in a ApiProduct.

Consider a subscription `sub-1` and 3 apis (`api-1`, `api-2` and `api-3`), you keep a state for `sub-1 api-1`, `sub-1 api-2` and `sub-1 api-3`.

Now, you change de ApiProduct to not have `api-1` anymore, but `api-4`, `api-2` and `api-3`, you have a complicated thing to do to completely realign the state.

It would be interesting to challenge this part and evaluate impacts and efforts

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiProductManagerImpl.java
arpzilla-dev: yes, implemented the same

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiProductManagerImpl.java
arpzilla-dev: changed to "doBeforeEmit"

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
arpzilla-dev: Yes, changed it

arpzilla-dev:
> On a more general note, this implementation looks a bit like a patch. The way you manage cache implies a bit too much coupling between subscription and APIs in a ApiProduct.
> 
> Consider a subscription `sub-1` and 3 apis (`api-1`, `api-2` and `api-3`), you keep a state for `sub-1 api-1`, `sub-1 api-2` and `sub-1 api-3`.
> 
> Now, you change de ApiProduct to not have `api-1` anymore, but `api-4`, `api-2` and `api-3`, you have a complicated thing to do to completely realign the state.
> 
> It would be interesting to challenge this part and evaluate impacts and efforts

Hi @ytvnr 

Thanks for raising this. Here’s the context behind the approach:

The security chain looks up subscriptions by (apiId, clientId, plan) or (apiId, subscriptionId); the API is always known at request time, so the lookup is fully **_API-driven_**. 
 getByApiAndClientIdAndPlan() and getByApiAndId().
 
I also considered an "API-product-centric" lookup, but that would require changes to SubscriptionService, SubscriptionCacheService, and all security handlers (JWT, OAuth2, API key) to support product-level lookup. 
The refactor scope felt _too large for this fix_, 

so I implemented the expansion logic instead: one subscription is expanded into one cache entry per API for the api- product.

And Yes, When the API product changes from [api-1, api-2, api-3] to [api-4, api-2, api-3], we 
(1) unregister removed APIs (e.g. sub-1→api-1), and (2) run a full refresh: load subscriptions and expand them with the new product state. We re-register api-2 and api-3 even though they didn’t change; that’s idempotent and keeps the logic simple.

:point_right: I’d be happy to discuss alternative architectures, OR
 I after this PR, I could explore another approach in a follow-up and share it with you separately.

ytvnr:
> > On a more general note, this implementation looks a bit like a patch. The way you manage cache implies a bit too much coupling between subscription and APIs in a ApiProduct.
> > Consider a subscription `sub-1` and 3 apis (`api-1`, `api-2` and `api-3`), you keep a state for `sub-1 api-1`, `sub-1 api-2` and `sub-1 api-3`.
> > Now, you change de ApiProduct to not have `api-1` anymore, but `api-4`, `api-2` and `api-3`, you have a complicated thing to do to completely realign the state.
> > It would be interesting to challenge this part and evaluate impacts and efforts
> 
> Hi @ytvnr
> 
> Thanks for raising this. Here’s the context behind the approach:
> 
> The security chain looks up subscriptions by (apiId, clientId, plan) or (apiId, subscriptionId); the API is always known at request time, so the lookup is fully **_API-driven_**. getByApiAndClientIdAndPlan() and getByApiAndId().
> 
> I also considered an "API-product-centric" lookup, but that would require changes to SubscriptionService, SubscriptionCacheService, and all security handlers (JWT, OAuth2, API key) to support product-level lookup. The refactor scope felt _too large for this fix_,
> 
> so I implemented the expansion logic instead: one subscription is expanded into one cache entry per API for the api- product.
> 
> And Yes, When the API product changes from [api-1, api-2, api-3] to [api-4, api-2, api-3], we (1) unregister removed APIs (e.g. sub-1→api-1), and (2) run a full refresh: load subscriptions and expand them with the new product state. We re-register api-2 and api-3 even though they didn’t change; that’s idempotent and keeps the logic simple.
> 
> 👉 I’d be happy to discuss alternative architectures, OR I after this PR, I could explore another approach in a follow-up and share it with you separately.

Hey @arpzilla-dev ,

The problem with this approach is it will not scale properly. Some of our customer may have a lot of APIs x a lot of subscription, adding a layer with ApiProduct will be an opportunity for those clients to have ApiProducts containing maybe 300 APIs, so it is a new cartesian product that you will refresh on every change.

I don't get what you mean by "that would require changes to SubscriptionService, SubscriptionCacheService, and all security handlers (JWT, OAuth2, API key) to support product-level lookup." From memory, subscription is retrieved at one place, why impacting all of those classes ?

I guess we will need to organize a workshop to discuss that issue ;)

ytvnr (approved):

arpzilla-dev:
> > > On a more general note, this implementation looks a bit like a patch. The way you manage cache implies a bit too much coupling between subscription and APIs in a ApiProduct.
> > > Consider a subscription `sub-1` and 3 apis (`api-1`, `api-2` and `api-3`), you keep a state for `sub-1 api-1`, `sub-1 api-2` and `sub-1 api-3`.
> > > Now, you change de ApiProduct to not have `api-1` anymore, but `api-4`, `api-2` and `api-3`, you have a complicated thing to do to completely realign the state.
> > > It would be interesting to challenge this part and evaluate impacts and efforts
> > 
> > 
> > Hi @ytvnr
> > Thanks for raising this. Here’s the context behind the approach:
> > The security chain looks up subscriptions by (apiId, clientId, plan) or (apiId, subscriptionId); the API is always known at request time, so the lookup is fully **_API-driven_**. getByApiAndClientIdAndPlan() and getByApiAndId().
> > I also considered an "API-product-centric" lookup, but that would require changes to SubscriptionService, SubscriptionCacheService, and all security handlers (JWT, OAuth2, API key) to support product-level lookup. The refactor scope felt _too large for this fix_,
> > so I implemented the expansion logic instead: one subscription is expanded into one cache entry per API for the api- product.
> > And Yes, When the API product changes from [api-1, api-2, api-3] to [api-4, api-2, api-3], we (1) unregister removed APIs (e.g. sub-1→api-1), and (2) run a full refresh: load subscriptions and expand them with the new product state. We re-register api-2 and api-3 even though they didn’t change; that’s idempotent and keeps the logic simple.
> > 👉 I’d be happy to discuss alternative architectures, OR I after this PR, I could explore another approach in a follow-up and share it with you separately.
> 
> Hey @arpzilla-dev ,
> 
> The problem with this approach is it will not scale properly. Some of our customer may have a lot of APIs x a lot of subscription, adding a layer with ApiProduct will be an opportunity for those clients to have ApiProducts containing maybe 300 APIs, so it is a new cartesian product that you will refresh on every change.
> 
> I don't get what you mean by "that would require changes to SubscriptionService, SubscriptionCacheService, and all security handlers (JWT, OAuth2, API key) to support product-level lookup." From memory, subscription is retrieved at one place, why impacting all of those classes ?
> 
> I guess we will need to organize a workshop to discuss that issue ;)

Hi @ytvnr 

Yes, rightly said, it will cause issue with scaling of api product, and needed to be fixed. 
I will look into this and improve my approach.
created this ticket for reference : 
https://gravitee.atlassian.net/browse/APIM-12874

PR #15231, opened 2026-02-17
fix: report not found at platform level when enabled

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-2336

The reported metric was written to the v4-metrics index only and the query layer was not searching for unknown id "1", making the 404 not reflected in the platform analytics even with the parameter enabled at the reporter level

This fixes the issue, but only for admin:

Screenshot 2026-02-17 at 12 26 25

There is also pr in reporter-common, because a ftl error occurred when the api was unknown

https://github.com/gravitee-io/gravitee-reporter-common/pull/110

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/EnvironmentAnalyticsResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/EnvironmentAnalyticsResource.java
@@ -71,6 +71,7 @@ import org.jetbrains.annotations.NotNull;
 public class EnvironmentAnalyticsResource extends AbstractResource {
 
     public static final String API_FIELD = "api";
+    public static final String UNKNOWN_ID = "1";
     public static final String APPLICATION_FIELD = "application";
     public static final String STATE_FIELD = "state";
     public static final String LIFECYCLE_STATE_FIELD = "lifecycle_state";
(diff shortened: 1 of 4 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/EnvironmentAnalyticsResource.java
```diff
@@ -71,6 +71,7 @@
 public class EnvironmentAnalyticsResource extends AbstractResource {
 
     public static final String API_FIELD = "api";
+    public static final String UNKNOWN_ID = "1";
```
wbabyte, line 74: can you document it please to explain why it's needed

wbabyte (commented):
tests are missing

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/EnvironmentAnalyticsResource.java
```diff
@@ -242,9 +243,11 @@ private Map<String, Set<String>> buildTerms(ExecutionContext executionContext, A
                 applicationQuery.setStatus(ApplicationStatus.ACTIVE.name());
                 applicationQuery.setExcludeFilters(List.of(ApplicationExcludeFilter.OWNER));
                 ids = applicationService.searchIds(executionContext, applicationQuery, null);
+                ids.add(UNKNOWN_ID);
```
wbabyte, line 246: would fail if ids is immutable

a-cordier:
@wbabyte i'll close that one following our discussion

PR #15244, opened 2026-02-17
fix(console): load envs and current env when navigating to org settings

jourdiw:
## Issue

https://gravitee.atlassian.net/browse/APIM-12616

## Description

When loading the organization settings > Policies view and not having an environment with an id of "DEFAULT", the user was seeing an error snackbar that Something went wrong! and the view was not loading.

This error was specifically reproducible when a user would refresh a view in the Organization settings part of Console.

The fix loads the environments if not already loaded and chooses the first one as the default organization.

**More info:**

Ideally, we would not call the backend for information scoped by the "currentEnv" environment when in the organizations section. However the work to address that change would be expansive. Especially since multiple components in this section call the backend to an environment-scoped endpoint.

The MAPI-v2 now has policies at the organization level. So there is promise that we have more and more resources that will be scoped by organization rather than environment in the future.

### DIFF
--- a/gravitee-apim-console-webui/src/organization/configuration/organization-environment.guard.spec.ts
+++ b/gravitee-apim-console-webui/src/organization/configuration/organization-environment.guard.spec.ts
@@ -1,0 +1,113 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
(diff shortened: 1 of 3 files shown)


aqibmohammadkhan (approved):
LGTM

PR #15247, opened 2026-02-17
fix(failover): reset response headers between failover retries

jgiovaresco:
## Issue

https://gravitee.atlassian.net/browse/APIM-12807

## Description

Restore response headers to their pre-failover baseline before each retry attempt in FailoverInvoker to prevent cross-attempt header leakage (e.g. duplicated Content-Length) while preserving gateway-managed headers.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/failover/FailoverInvoker.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/failover/FailoverInvoker.java
@@ -24,6 +24,7 @@ import io.github.resilience4j.circuitbreaker.CircuitBreakerConfig;
 import io.github.resilience4j.circuitbreaker.CircuitBreakerRegistry;
 import io.github.resilience4j.rxjava3.circuitbreaker.operator.CircuitBreakerOperator;
 import io.gravitee.definition.model.v4.failover.Failover;
+import io.gravitee.gateway.api.http.HttpHeaders;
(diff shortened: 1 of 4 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/failover/FailoverInvoker.java
```diff
@@ -82,11 +88,17 @@ public Completable invoke(ExecutionContext executionContext) {
     @Override
     public Completable invoke(HttpExecutionContext ctx) {
         final String originalEndpoint = ctx.getAttribute(ATTR_REQUEST_ENDPOINT);
+        final Map<String, List<String>> baselineResponseHeaders = snapshotHeaders(ctx.response().headers());
```
jhaeyaert: You can use `HttpHeaders.create(ctx.request().headers())`

gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/failover/FailoverInvoker.java
```diff
@@ -82,11 +88,17 @@ public Completable invoke(ExecutionContext executionContext) {
...
+        final AtomicInteger attempts = new AtomicInteger();
```
jhaeyaert: AtomicBoolean instead?

jhaeyaert (approved):
Minor comments but LGTM 👍

ytvnr (approved):
agree with Jeoffrey, appart from that, LGTM

PR #15248, opened 2026-02-17
feat: close subscription in portal - create confirmation popup and service method.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-12271

## Description

This PR is the first in the series to introduce subscription closure feature in subscription details in the portal.

### Included in this PR:
- new service method to close a subscription (uses existing back end endpoint)
- new confirmation popup component

### Excluded from this PR:
- button "Close subscription"
- updates to subscription-details and subscription-info components (button handler, output)
- generic dialog component
- styles updates
Screenshot 2026-02-18 at 09 14 11

### DIFF
--- a/gravitee-apim-portal-webui-next/src/services/subscription.service.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/services/subscription.service.spec.ts
@@ -50,4 +50,15 @@ describe('SubscriptionService', () => {
 
     req.flush(subscriptionResponse);
   });
+
+  it('should close subscription', done => {
+    service.close('subscriptionId').subscribe(() => {
+      done();
+    });
+
+    const req = httpTestingController.expectOne(`${TESTING_BASE_URL}/subscriptions/subscriptionId/_close`);
+    expect(req.request.method).toEqual('POST');
+
+    req.flush(null);
+  });
 });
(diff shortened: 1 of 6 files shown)


gravitee-apim-portal-webui-next/src/services/subscription.service.ts
```diff
@@ -68,6 +68,10 @@ export class SubscriptionService {
     return this.http.put<Subscription>(`${this.configService.baseURL}/subscriptions/${subscriptionId}`, updatedSubscription);
   }
 
+  close(subscriptionId: string) {
+    return this.http.post<void>(`${this.configService.baseURL}/subscriptions/${subscriptionId}/_close`, null);
```
aqibmohammadkhan, line 72: how about using a Subscription object instead of void?

wdyt about updating the closeSubscription endpoint and service method to return the updated 
Subscription object instead of an empty 204 No Content response.

gravitee-apim-portal-webui-next/src/services/subscription.service.spec.ts
aqibmohammadkhan, line 1: wdyt about adding an additional test with a mock response for close subscription call?

gravitee-apim-portal-webui-next/src/services/subscription.service.ts
kovaren: As discussed together, the endpoint is also used in the classic developer portal and the impact on it is not justified here, we will keep the current version.

gravitee-apim-portal-webui-next/src/services/subscription.service.spec.ts
kovaren: As discussed together, we decided not to change the endpoint so there is no response to mock.

aqibmohammadkhan (approved):
LGTM

PR #15264, opened 2026-02-18
fix: move timestamps to base trigger in alert api

josephlomasney:
## Issue

https://gravitee.atlassian.net/browse/CJ-4041

## Description

Remove the timestamp fields in trigger superclass as they are added in trigger base class to be sent through to alert engine. 

Alert api change here - https://github.com/gravitee-io/gravitee-alert-api/pull/76

I made this a breaking change on alert api side since apim can only use the new alert-api version with the changes in this PR (as otherwise will fail to compile with duplicate fields).

(Version will be updated from snapshot once alert api released and before merging this PR but wanted to check this PR before doing that release) 

Applying on master and 4.10.x as timestamps will be used to fix alert engine consistent schedule which needs to be delivered before 4.11

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/alert/AlertTriggerEntityWrapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/alert/AlertTriggerEntityWrapper.java
@@ -21,6 +21,7 @@ import io.gravitee.alert.api.trigger.Dampening;
 import io.gravitee.alert.api.trigger.Trigger;
 import io.gravitee.notifier.api.Notification;
 import io.gravitee.notifier.api.Period;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/alert/AlertTriggerEntityWrapper.java
```diff
@@ -157,6 +157,26 @@ public void setNotificationPeriods(List<Period> notificationPeriods) {
         trigger.setNotificationPeriods(notificationPeriods);
     }
 
+    @Override
+    public java.util.Date getCreatedAt() {
```
mariano-fernandez: The suggestion below applies to the other usages of the complete path
```suggestion
    public Date getCreatedAt() {
```

phiz71 (approved):

jgiovaresco (approved):
As I said in the api's PR, we could create an alpha version depending on how confident you are that more change will come or not 🤷

PR #15268, opened 2026-02-18
refactor(gravitee-markdown): allow parent to provide form state store

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-12634

## Description
This PR is the first of two and updates the gravitee-markdown library so that the form state store can be provided by the parent component instead of being created inside `gmd-form-host`. This allows consumers (e.g. the subscription form screen) to own the store and use it for validation (e.g. disabling Save when there are config errors).

### Included in this PR
- add `provideGmdFormStore()` so the store can be provided at the call site
- remove store creation from `gmd-form-host`; host only consumes and resets store on content change
- update tests to provide the store via `provideGmdFormStore()`

### Excluded from this PR (will be done in subsequent PRs)
- console subscription form screen and REST integration

### DIFF
--- a/gravitee-apim-webui-libs/gravitee-markdown/src/lib/services/gmd-form-state.store.ts
+++ b/gravitee-apim-webui-libs/gravitee-markdown/src/lib/services/gmd-form-state.store.ts
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { computed, Injectable, InjectionToken, signal } from '@angular/core';
+import { computed, Injectable, InjectionToken, Provider, signal } from '@angular/core';
 import { isEqual } from 'lodash';
 
 import { GmdConfigError, GmdFieldState } from '../models/formField';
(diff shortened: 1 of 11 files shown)


gravitee-apim-webui-libs/gravitee-markdown/src/lib/services/gmd-form-state.store.ts
```diff
@@ -210,3 +210,13 @@ export class GmdFormStateStore {
...
+
+/**
+ * Provide a scoped GMD form state store instance.
+ */
+export const provideGmdFormStore = (): Provider[] => [
+  {
+    provide: GMD_FORM_STATE_STORE,
+    useFactory: () => new GmdFormStateStore(),
+  },
+];
```
kovaren: WDYT about returning one provider instead of an array?

```suggestion
export const provideGmdFormStore = (): Provider => {
  provide: GMD_FORM_STATE_STORE,
  useFactory: () => new GmdFormStateStore(),
};

kovaren (changes requested):

jourdiw (approved):
Looks good to me after implementing the change asked by @kovaren 👍

kovaren (approved):

PR #15273, opened 2026-02-18
feat(env-logs): add Plan, Endpoint reached, Issues, Preview columns to table

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/GKO-2313
## Description

Table now resembles updated design: 
Screenshot 2026-02-18 at 9 40 35 AM


gravitee-apim-console-webui/src/management/analytics/env-logs/models/env-log.model.ts
carlos-andres-osorio: `plan` and `endpoint` are declared as required here, but the template uses `log.plan?.name` (optional chaining) and the fixture sets them to `undefined` for log-2 — so the actual data contract is that these can be absent.

As-is, TypeScript won't warn if future code accesses `log.plan.name` directly (no `?.`), which will throw a `TypeError` at runtime when the backend omits the field.

```ts
plan?: { name: string };
endpoint?: string;
```

carlos-andres-osorio (commented):
Automated review from Claude Code

carlos-andres-osorio:
image

Description for response time doesn't match the operator, which is greater than or equal.

Also, response time could be lower than or equal, so this field should allow specifying a range, but this can be done in another PR.

carlos-andres-osorio (approved):
I did some manual tests and filter works fine.  There is a comment about time range, but it's not a blocker and can be implemented in a follow-up PR if wanted.

There is also a comment about required fields. I'm approving, and will let @paulatulis decide if this is a bug that needs to be fixed now.

gravitee-apim-console-webui/src/management/analytics/env-logs/models/env-log.model.ts
paulatulis: Thank you! Addressed.

PR #15282, opened 2026-02-18
feat: close subscription in portal - create a generic confirmation dialog.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-12271

## Description

This PR follows the changes made by [this PR](https://github.com/gravitee-io/gravitee-api-management/pull/15272) to introduce subscription closure feature in subscription details in the portal.

### Included in this PR:
- generic dialog component

### Excluded from this PR:
- misc styles updates


gravitee-apim-portal-webui-next/src/app/applications/application/application-tab-settings/application-tab-settings.component.ts
```diff
@@ -62,14 +62,21 @@ export class ApplicationTabSettingsComponent implements OnInit {
   }
 
   deleteApplication(): void {
+    const dialogData: ConfirmDialogData = {
+      title: $localize`:@@titleDeleteApplicationDialog:Delete application`,
+      content: $localize`:@@contentDeleteApplicationDialog:All your subscriptions will be closed. Are you sure you want to delete this application?`,
+      confirmLabel: $localize`:@@confirmDeleteApplicationDialog:Delete`,
+      cancelLabel: $localize`:@@cancelDeleteApplicationDialog:Cancel`,
+    };
```
jourdiw, line 70: Is there a fancy way for the parent component to submit all of these values without knowing how to do localize?

Such as:

```suggestion
    const dialogData: ConfirmDialogData = {
      title: 'Delete application',
      content: 'All your subscriptions will be closed. Are you sure you want to delete this application?',
      ...
      localizeTag: 'DeleteApplicationDialog'
    };
```

and then the key in ConfirmDialogData is transformed as the localize key like `@@<field name, ex. "content">DeleteApplicationDialog`?

gravitee-apim-portal-webui-next/src/components/confirm-dialog/confirm-dialog.component.ts
```diff
@@ -13,13 +13,26 @@
...
-  selector: 'app-close-subscription-dialog',
-  templateUrl: './close-subscription-dialog.component.html',
+  selector: 'app-confirm-dialog',
+  templateUrl: './confirm-dialog.component.html',
   imports: [MatDialogModule, MatButtonModule],
+  standalone: true,
 })
-export class CloseSubscriptionDialogComponent {}
+export class ConfirmDialogComponent {
+  public readonly defaultConfirmLabel = $localize`:@@confirmDialogConfirmAction:OK`;
```
jourdiw: For UX best practices, I would argue that the confirm button should always be pertinent to the dialog box. WDYT about making the confirm label required?

gravitee-apim-portal-webui-next/src/app/applications/application/application-tab-settings/application-tab-settings.component.ts
kovaren: I gave it a try but unfortunately `localize` expects the key to be a static string at compile time.

For example, this result in an error:
```
$localize`:@@title${customLocalizeTag}:${defaultValue}`
```
There doesn't seem to be a way around it apart using another translation library.

gravitee-apim-portal-webui-next/src/components/confirm-dialog/confirm-dialog.component.ts
kovaren: Good point, I'd even argue cancel should also be required

kovaren: Fixed!

gravitee-apim-portal-webui-next/src/app/applications/application/application-tab-settings/application-tab-settings.component.ts
jourdiw: WDYT about adding a comment to the Data explaining the expected format or including "Localize" in the name of the data interface?
Since this isn't a very standard pattern in our code base to pass "$localize...", perhaps we can leave some hints to the devs and agents?

jourdiw (approved):

gravitee-apim-portal-webui-next/src/app/applications/application/application-tab-settings/application-tab-settings.component.ts
kovaren: Added a doc comment to ConfirmDialogData

PR #15287, opened 2026-02-18
feat(env logs details): connect backend to frontend 

paulatulis:
## Issue
https://gravitee.atlassian.net/browse/GKO-2013
https://gravitee.atlassian.net/browse/GKO-2390
## Description

Hooks up backend to frontend for env logging log detals view. 
Screenshot 2026-02-18 at 4 35 39 PM

host, remoteAddress, endpointResponseTime, gatewayLatency, responseContentLength don't exist in either backend endpoint's response schema.

### DIFF
--- a/gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-details/env-logs-details.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-details/env-logs-details.component.spec.ts
@@ -13,41 +13,68 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { TestBed } from '@angular/core/testing';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { ActivatedRoute, provideRouter } from '@angular/router';
-import { provideHttpClient } from '@angular/common/http';
-import { provideHttpClientTesting } from '@angular/common/http/testing';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
+import { HttpTestingController } from '@angular/common/http/testing';
 
 import { EnvLogsDetailsComponent } from './env-logs-details.component';
 import { EnvLogsDetailsHarness } from './env-logs-details.harness';
 
-import { fakeEnvLogs } from '../../models/env-log.fixture';
+import { GioTestingModule } from '../../../../../shared/testing/gio-testing.module';
+import { fakeConnectionLogDetail } from '../../../../../entities/management-api-v2/log/connectionLog.fixture';
+import { fakeApiMetricResponse } from '../../../../../entities/management-api-v2/analytics/apiMetricsDetailResponse.fixture';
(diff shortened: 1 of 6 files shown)


gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-details/env-logs-details.component.spec.ts
```diff
@@ -13,120 +13,258 @@
...
+    await createComponent();
+
+    const searchReq = expectSearchRequest();
+    const body = searchReq.request.body;
+    expect(body.timeRange).toBeDefined();
+    // Verify it's a wide window (> 1 year) not just 24h
+    const from = new Date(body.timeRange.from).getTime();
+    const to = new Date(body.timeRange.to).getTime();
+    const oneYearMs = 365 * 24 * 60 * 60 * 1000;
+    expect(to - from).toBeGreaterThan(oneYearMs);
```
carlos-andres-osorio, line 239: nit: This test validates the time range but never asserts that the request body includes the `REQUEST_ID` filter. Consider adding:
```typescript
expect(body.filters).toEqual([{ name: 'REQUEST_ID', operator: 'EQ', value: logId }]);
```
This is the primary mechanism ensuring the detail page finds the correct log — without it, the search could return arbitrary results and the test wouldn't catch the regression.

paulatulis: Done! Line 234

gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-details/env-logs-details.component.ts
```diff
@@ -75,17 +102,70 @@ export class EnvLogsDetailsComponent {
...
+      switchMap(searchResponse => {
+        const overviewLog = searchResponse.data[0];
+        if (!overviewLog) {
+          return of(undefined);
+        }
+
+        return forkJoin({
+          overview: of(overviewLog),
+          detail: this.apiLogsV2Service.searchConnectionLogDetail(overviewLog.apiId, overviewLog.id).pipe(catchError(() => of(null))),
+          metrics: this.apiAnalyticsV2Service.getApiMetricsDetail(overviewLog.apiId, overviewLog.id).pipe(catchError(() => of(null))),
```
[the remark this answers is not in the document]
paulatulis: This is intentional graceful degradation — the detail and metrics calls are supplementary enrichment on top of the core log overview from 

searchLogs
. If either fails (e.g., the API was deleted, or the user lacks permissions), the page should still render with whatever data is available rather than showing a full error state.

This is covered by two dedicated tests:

should gracefully degrade when detail endpoint fails — verifies the page renders with overview + metrics data when the detail call 404s
should gracefully degrade when metrics endpoint fails — verifies the page renders with overview + detail data when the metrics call 404s
The outer catchError on the full pipeline does surface errors from the primary searchLogs call as a user-visible error banner.

carlos-andres-osorio (approved):

PR #15291, opened 2026-02-19
refactor: validate str length for dashboard name and widget titles

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-2342

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/analytics/dashboards/DashboardResourceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/analytics/dashboards/DashboardResourceTest.java
@@ -39,12 +39,15 @@ import jakarta.ws.rs.client.WebTarget;
 import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayNameGeneration;
 import org.junit.jupiter.api.DisplayNameGenerator;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 
 @DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
 class DashboardResourceTest extends AbstractResourceTest {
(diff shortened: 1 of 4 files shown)


carlos-andres-osorio (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/analytics/dashboards/DashboardsResourceTest.java
```diff
@@ -26,7 +25,6 @@
 import assertions.MAPIAssertions;
 import fixtures.DashboardFixtures;
 import inmemory.DashboardCrudServiceInMemory;
-import io.gravitee.rest.api.management.v2.rest.model.analytics.engine.CreateDashboard;
 import io.gravitee.rest.api.management.v2.rest.model.analytics.engine.CustomInterval;
```
carlos-andres-osorio, line 28: The import `io.gravitee.rest.api.management.v2.rest.model.analytics.engine.CreateDashboard` was removed in this PR, but the 403 permission test still calls `new CreateDashboard()`. Since this test lives inside `@Nested class CreateDashboard`, the reference now resolves to the **enclosing test class** rather than the API model.

The test still passes because the 403 check doesn't depend on body content.

**Suggestion:** Re-add the import or use the fully qualified class name.

carlos-andres-osorio (commented):
Minor issue spotted during review — see inline comment.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/analytics/dashboards/DashboardResourceTest.java
```diff
@@ -221,6 +224,28 @@ void should_return_400_if_missing_body() {
...
+        @MethodSource("provideInvalidName")
+        void should_return_400_if_widget_title_is_invalid(String name) {
+            var updateDashboard = DashboardFixtures.anUpdateDashboard();
+            updateDashboard.getWidgets().getFirst().setTitle(name);
+            var response = target.request().put(json(updateDashboard));
+            assertThat(response.getStatus()).isEqualTo(BAD_REQUEST_400);
+        }
+
+        private static Stream<String> provideInvalidName() {
+            return Stream.of("ab", "a".repeat(257));
```
carlos-andres-osorio, line 246: could users send null as well?

carlos-andres-osorio: maybe another test would be a string with only spaces. I guess that's a valid name but not sure if it could cause issues when being returned.

carlos-andres-osorio:
rebased by mistake :grimacing:

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/analytics/dashboards/DashboardResourceTest.java
a-cordier: they can't send null, but i will test this only spaces thing 👍

a-cordier: coming in one of the other PR :)

PR #15292, opened 2026-02-19
fix: added API product - create, update plan and plan list

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-12835

## Description
Added API product - create, update plan and plan list
For now navigating to plans directly from api product list
Later side menu will be added that will have consumers submenu from where one can navigate to plans


gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/permissions/EnvironmentPermission.java
vikrantgravitee, line 1: Seems this is not right.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/permissions/RolePermission.java
vikrantgravitee, line 1: Also it is not environment level.

pragsty:
Raised another PR https://github.com/gravitee-io/gravitee-api-management/pull/15351

PR #15297, opened 2026-02-19
feat(portal): Enhance subscription table styling and fix query parame…

aqibmohammadkhan:
…ter keys

## Issue

https://gravitee.atlassian.net/browse/APIM-12269

## Description

 - fixed filters not working
 - added max width 1400px
 - removed double border from table
 
 

https://github.com/user-attachments/assets/5045dd91-ee22-422d-a54c-e160d5dfeb79

### DIFF
--- a/gravitee-apim-portal-webui-next/src/components/paginated-table/paginated-table.component.scss
+++ b/gravitee-apim-portal-webui-next/src/components/paginated-table/paginated-table.component.scss
@@ -30,7 +30,6 @@
   }
 
   &__container {
-    border: 1px solid app-theme.$table-border-color;
     border-radius: app-theme.$table-border-radius;
     overflow-x: auto;
   }
@@ -44,7 +43,6 @@
   }
 
   &__column-expand {
-    width: 48px;
     text-align: center;
 
     mat-icon {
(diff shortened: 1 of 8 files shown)


gravitee-apim-portal-webui-next/src/components/paginated-table/paginated-table.component.scss
```diff
@@ -17,6 +17,7 @@
 
 .paginated-table {
   width: 100%;
+  max-width: app-theme.$inner-content-width;
```
jourdiw: I think it might be a bit easier if the parent controls this max-width rather than the component 🤔 That way we can let parents handle the layout and the components can be more flexible. WDYT?

aqibmohammadkhan: i think this can be removed. parent [here](https://github.com/gravitee-io/gravitee-api-management/pull/15297/changes/BASE..826c3875da2ba9745ef09f9255fafeda917f6c9b#diff-d4d9ded05713d804cd6ac458dffd3a2b57ef117463eccb0da8e3da055463349eR26)
has max width and the table has width 100% so it will fill the space the parent will give.

jourdiw (approved):

PR #15302, opened 2026-02-19
GKO-2002-manage-several-certs

benoitgravitee:
## Issue

https://gravitee.atlassian.net/browse/GKO-2002
https://gravitee.atlassian.net/browse/GKO-2003

## Description

* Add a list of certificates at the API level for GKO
* Apply the logic so that several certs can be managed
* Update subscription outside ApplicationService
* Refactor domain client cert to be a record
* Fix faulty behaviour of subscription truststore loader in multi server env
* Fix GKO-2003
   * Change cardinality in subscription trustore loader and manager to work at the certificate level
   * Adjust register/unregister order to avoid connection cuts
   * Rewrite test to be more functional 
     * Less introspection to check internal states 
     * Less mock to use lower level layer

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
@@ -162,7 +162,7 @@ public class SubscriptionCacheService implements SubscriptionService {
     private void registerFromClientCertificate(final Subscription subscription) {
         final Set<String> servers = extractApiServersId(subscription);
 
-        // register new certs and remove old one
+        // Ensure trust store is updated before cache to maintain certificate validation consistency
         subscriptionTrustStoreLoaderManager.registerSubscription(subscription, servers);
 
         // keep the old one
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
@@ -178,8 +178,12 @@ public class SubscriptionCacheService implements SubscriptionService {
             Objects.equals(subscription.getApi(), cached.getApi()) &&
             !Objects.equals(subscription.getClientCertificate(), cached.getClientCertificate())
         ) {
-            evictKeyForApi(cached.getApi(), identityCacheKey(cached));
-            evictKeyForApi(cached.getApi(), identityCacheKeyWithoutPlan(cached));
+            String cacheKey = identityCacheKey(cached);
+            String cacheKeyWithoutPlan = identityCacheKeyWithoutPlan(cached);
+            evictKeyForApi(cached.getApi(), cacheKey);
+            evictKeyForApi(cached.getApi(), cacheKeyWithoutPlan);
+            cacheByClientCertificate.remove(cacheKey);
+            cacheByClientCertificate.remove(cacheKeyWithoutPlan);
         }
     }
 
@@ -203,11 +207,13 @@ public class SubscriptionCacheService implements SubscriptionService {
 
     private void updateIdentityCache(Subscription subscription, Map<String, Subscription> cache) {
         updateCacheKeyByApiId(subscription.getApi(), subscription.getId());
-        updateCacheKeyByApiId(subscription.getApi(), identityCacheKey(subscription));
-        cache.put(identityCacheKey(subscription), subscription);
+        String cacheKey = identityCacheKey(subscription);
+        updateCacheKeyByApiId(subscription.getApi(), cacheKey);
+        cache.put(cacheKey, subscription);
         // Index the subscription without plan id to allow search without plan criteria.
-        cache.put(identityCacheKeyWithoutPlan(subscription), subscription);
-        updateCacheKeyByApiId(subscription.getApi(), identityCacheKeyWithoutPlan(subscription));
+        String cacheKeyWithoutPlan = identityCacheKeyWithoutPlan(subscription);
+        cache.put(cacheKeyWithoutPlan, subscription);
+        updateCacheKeyByApiId(subscription.getApi(), cacheKeyWithoutPlan);
     }
 
     private void updateCacheKeyByApiId(final String apiId, final String cacheKey) {
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
@@ -25,6 +25,9 @@ import io.gravitee.gateway.reactive.api.policy.SecurityToken;
 import io.gravitee.gateway.reactive.handlers.api.v4.Api;
 import io.gravitee.gateway.reactor.ReactableApi;
 import io.gravitee.gateway.security.core.SubscriptionTrustStoreLoaderManager;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
@@ -33,12 +36,11 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.BiPredicate;
-import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import lombok.CustomLog;
 import lombok.RequiredArgsConstructor;
+import lombok.SneakyThrows;
 
 @CustomLog
 @RequiredArgsConstructor
@@ -50,11 +52,10 @@ public class SubscriptionCacheService implements SubscriptionService {
 
     // Caches only contains active subscriptions
     private final Map<String, Subscription> cacheByApiClientId = new ConcurrentHashMap<>();
-    private final Map<String, Subscription> cacheByApiClientCertificate = new ConcurrentHashMap<>();
+    private final Map<String, Subscription> cacheByClientCertificate = new ConcurrentHashMap<>();
     private final Map<String, Subscription> cacheBySubscriptionId = new ConcurrentHashMap<>();
     private final Map<String, Set<Subscription>> cacheBySubscriptionIdAll = new ConcurrentHashMap<>(); // exploded subscriptions
-    private final Map<String, Set<String>> cacheByApiId = new ConcurrentHashMap<>();
-    private final Map<String, Object> locksBySubscriptionId = new ConcurrentHashMap<>();
+    private final Map<String, Set<String>> cacheKeysByApiId = new ConcurrentHashMap<>();
 
     @Override
     public Optional<Subscription> getByApiAndSecurityToken(String api, SecurityToken securityToken, String plan) {
@@ -66,14 +67,14 @@ public class SubscriptionCacheService implements SubscriptionService {
                 .getByApiAndMd5Key(api, securityToken.getTokenValue())
                 .flatMap(apiKey -> getByApiAndId(api, apiKey.getSubscription()));
             case CLIENT_ID -> getByApiAndClientIdAndPlan(api, securityToken.getTokenValue(), plan);
-            case CERTIFICATE -> subscriptionTrustStoreLoaderManager.getByCertificate(api, securityToken.getTokenValue(), plan);
+            case CERTIFICATE -> subscriptionTrustStoreLoaderManager.getByCertificate(api, plan, securityToken.getTokenValue());
             default -> Optional.empty();
         };
     }
 
     @Override
     public Optional<Subscription> getByApiAndClientIdAndPlan(String api, String clientId, String plan) {
-        return Optional.ofNullable(cacheByApiClientId.get(buildCacheKeyFromClientInfo(api, clientId, plan)));
+        return Optional.ofNullable(cacheByApiClientId.get(cacheKey(api, plan, clientId)));
     }
 
     @Override
@@ -81,27 +82,18 @@ public class SubscriptionCacheService implements SubscriptionService {
         return Optional.ofNullable(cacheBySubscriptionId.get(subscriptionId));
     }
 
-    /** Returns all subscriptions for the given ID (multiple for exploded API Product subscriptions). */
+    /**
+     * Returns all subscriptions for the given ID (multiple for exploded API Product subscriptions).
+     */
     public Collection<Subscription> getAllById(String subscriptionId) {
         Set<Subscription> subscriptions = cacheBySubscriptionIdAll.get(subscriptionId);
         return subscriptions != null ? Set.copyOf(subscriptions) : Collections.emptySet();
     }
 
-    /** Returns subscription for the given API and ID (for exploded subscriptions). */
-    public Optional<Subscription> getByApiAndId(String api, String subscriptionId) {
-        Set<Subscription> subscriptions = cacheBySubscriptionIdAll.get(subscriptionId);
-        if (subscriptions == null || subscriptions.isEmpty()) {
-            // Fallback to old cache for backward compatibility
-            return getById(subscriptionId);
-        }
-        return subscriptions
-            .stream()
-            .filter(sub -> Objects.equals(api, sub.getApi()))
-            .findFirst();
-    }
-
     @Override
     public void register(final Subscription subscription) {
+        // only once per synchronization window
+        // take all fields (including "updatedAt" in metadata) into account
         if (ACCEPTED.name().equals(subscription.getStatus())) {
             if (subscription.getClientCertificate() != null) {
                 registerFromClientCertificate(subscription);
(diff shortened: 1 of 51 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
```diff
@@ -640,37 +656,38 @@ public ApplicationEntity update(
             Map<String, String> metadata = new HashMap<>();
 
             // Handle TLS certificate changes via ClientCertificateCrudService
-            String newCertificateToCreate = null;
-            if (
-                updateApplicationEntity.getSettings().getTls() != null &&
-                !StringUtils.isBlank(updateApplicationEntity.getSettings().getTls().getClientCertificate())
-            ) {
-                String newCertificate = updateApplicationEntity.getSettings().getTls().getClientCertificate().trim();
+            List<CreateClientCertificate> newCertificates = getClientCertificates(
```
carlos-andres-osorio: When updateApplicationEntity.getSettings().getTls() is null (e.g. a PUT that omits TLS), getClientCertificates() returns List.of(), causing the reconciliation loop to expire all active
certificates.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
```diff
@@ -694,12 +711,18 @@ public ApplicationEntity update(
...
             }
 
-            // Create new certificate if certificate was changed
-            final Optional<String> clientCertificate = createNewCertificateIfNecessary(
-                applicationId,
-                updateApplicationEntity,
-                newCertificateToCreate
-            );
+            // Create new certificates
+            for (var cert : certsToCreate) {
```
carlos-andres-osorio: certs are expired before new ones are added, so failing to add a certificate could leave the app without certs.

Could creation be done before expiration?

Also, this loop  and others that call CRUD could also have a try...catch that would not abort on the first error.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/application_certificates/ApplicationCertificatesUpdateDomainServiceImpl.java
```diff
@@ -64,7 +65,7 @@ public void updateActiveMTLSSubscriptions(String applicationId) {
         List<ClientCertificate> activeCertificates = clientCertificateCrudService
             .findByApplicationIdAndStatuses(applicationId, ClientCertificateStatus.ACTIVE, ClientCertificateStatus.ACTIVE_WITH_END)
             .stream()
-            .sorted(Comparator.comparing(ClientCertificate::getCreatedAt))
+            .sorted(Comparator.comparing(ClientCertificate::createdAt))
```
carlos-andres-osorio: In line 73 below, if there are no active certificates, the subscription will be updated without certificates.  Does it make the subscription available without authentication?

If so, maybe the `log.debug` should be a warning?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
```diff
@@ -640,37 +656,38 @@ public ApplicationEntity update(
...
-                                .startsAt(existingCert.getStartsAt())
-                                .endsAt(new Date())
-                                .build()
-                        )
+            // Determine which new certs need to be created (not already present)
+            var certsToCreate = new ArrayList<CreateClientCertificate>();
+            for (var newCert : newCertificates) {
+                boolean alreadyExists = existingActiveCerts
+                    .stream()
+                    .anyMatch(existing -> newCert.certificate().equals(existing.certificate()));
```
carlos-andres-osorio: is every certificate guaranteed to have a non-null `.certificate()`?  If not it may be better to compare using `Objects.equals()`

```
.anyMatch(existing -> Objects.equals(newCert.certificate(), existing.certificate()));                                                               
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
benoitgravitee: That was one of thoughts over the weekend. It will now return a 400 error causing the reconcile loop to stop.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/application_certificates/ApplicationCertificatesUpdateDomainServiceImpl.java
benoitgravitee: As mentioned above that should be an error. If there are active subscriptions we cannot update with no certs. If user want to block API access, it should pause/remove the subscription instead.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
benoitgravitee: Ok with reversing the logic. But I think it should a fail fast thing. You don't want all errors to be ignored and have a bad sate. When we are here the content of the incoming certificate should be validated already so it is just about DB errors.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
benoitgravitee: Yes, you can't set a null cert but worth using a null safe way. I agree

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoaderManager.java
```diff
@@ -33,63 +38,123 @@
 @CustomLog
 public class SubscriptionTrustStoreLoaderManager {
 
-    private final Map<String, SubscriptionTrustStoreLoader> subscriptionTrustStoreLoaders = new ConcurrentHashMap<>();
-    private final Map<String, Subscription> subscriptionsByApi = new ConcurrentHashMap<>();
+    private final Map<String, Set<SubscriptionCertificate>> certificates = new ConcurrentHashMap<>();
+    private final Map<Key, Subscription> subscriptions = new ConcurrentHashMap<>();
+    private final Multimap<Key, SubscriptionTrustStoreLoader> truststoreLoaders = MultimapBuilder.hashKeys().hashSetValues().build();
```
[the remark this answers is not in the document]
benoitgravitee: Look closer dummy... truststoreLoaders is always in a computeIf* methods of concurrent maps

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoaderManager.java
```diff
@@ -33,63 +38,123 @@
...
-        } catch (MalformedCertificateException e) {
-            log.error(e.getMessage(), e.getCause());
-            return;
-        }
-        if (subscriptionTrustStoreLoaders.containsKey(subscription.getId())) {
-            log.debug("A TrustStoreLoader for subscription {} is already registered", subscription.getId());
-            return;
-        }
-        log.debug("Registering TrustStoreLoader for subscription {}", subscription.getId());
+        certificates.computeIfPresent(subscription.getId(), (id, currentState) -> {
```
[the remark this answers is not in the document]
benoitgravitee: It is commented enough thanks

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
```diff
@@ -115,199 +107,222 @@ public void register(final Subscription subscription) {
...
+            return cacheKey(subscription.getApi(), subscription.getClientId());
+        } else {
+            Objects.requireNonNull(subscription.getClientCertificate(), "Client certificate must not be null");
+            return cacheKey(subscription.getApi(), sha256(subscription.getClientCertificate()));
         }
     }
 
-    String buildClientCertificateCacheKey(Subscription subscription) {
-        return buildCacheKeyFromClientInfo(subscription.getApi(), subscription.getClientCertificate(), subscription.getPlan());
+    @SneakyThrows
```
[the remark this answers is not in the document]
benoitgravitee: Certs are unique, so no possible caching.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
```diff
@@ -475,14 +482,21 @@ public ApplicationEntity create(final ExecutionContext executionContext, NewAppl
...
+        }
+        var result = new ArrayList<CreateClientCertificate>();
+        if (CollectionUtils.isNotEmpty(settings.getTls().getClientCertificates())) {
+            result.addAll(settings.getTls().getClientCertificates());
+        }
+        if (StringUtils.isNotBlank(settings.getTls().getClientCertificate())) {
+            result.add(new CreateClientCertificate(appName, null, null, settings.getTls().getClientCertificate()));
         }
-        return null;
+        return result;
```
carlos-andres-osorio: This endpoint adds all certificates passed in `settings.tls.client_certificate` and `settings.tls.client_certificate` which is inconsistent with the validation for `management/v2/environments/DEFAULT/applications/_import/crd`, in method `ValidateApplicationSettingsDomainServiceImpl.validateTlsSettings`.

Shouldn't they have similar logic?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
```diff
@@ -42,16 +47,42 @@ public class SubscriptionValidationServiceImpl extends TransactionalService impl
...
+        String applicationId
+    ) {
+        validateTls(genericPlanEntity, subscription, applicationId);
         subscription.setConfiguration(validateAndSanitizeSubscriptionConfiguration(genericPlanEntity, subscription.getConfiguration()));
     }
 
+    private void validateTls(final GenericPlanEntity genericPlanEntity, final UpdateSubscriptionEntity subscription, String applicationId) {
+        if (
+            subscription.getEndingAt() != null &&
+            Objects.equals(genericPlanEntity.getPlanSecurity().getType(), PlanSecurityType.MTLS.name())
```
carlos-andres-osorio, line 78: This if statement didn't trigger because the `type()` is lowercase and `name()` is uppercase.

```suggestion
Objects.equals(genericPlanEntity.getPlanSecurity().getType().toLowerCase(), PlanSecurityType.MTLS.name().toLowerCase())
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
```diff
@@ -42,16 +47,42 @@ public class SubscriptionValidationServiceImpl extends TransactionalService impl
...
+        ) {
+            Set<ClientCertificate> byApplicationIdAndStatuses = clientCertificateCrudService.findByApplicationIdAndStatuses(
+                applicationId,
+                ClientCertificateStatus.ACTIVE_WITH_END
+            );
+            boolean subscriptionEndsAfterCertificate = byApplicationIdAndStatuses
+                .stream()
+                // all ending certificates must be before the subscription ending date
+                .allMatch(clientCertificate -> clientCertificate.endsAt().toInstant().isBefore(subscription.getEndingAt().toInstant()));
```
carlos-andres-osorio, line 92: **Bug**: `allMatch` on an empty stream returns `true`. When the application has only **ACTIVE** certificates (permanent, no end date), `findByApplicationIdAndStatuses(..., ACTIVE_WITH_END)` returns an empty set. The empty stream causes `subscriptionEndsAfterCertificate = true`, which throws `SubscriptionEndsAfterClientCertificateException` — rejecting valid subscription updates.

**Reproduced via API**: `PUT /apis/{api}/subscriptions/{sub}` with `{"ending_at": "..."}` on an application with only permanent certs returns:
```json
{"httpStatus":400,"message":"Subscription cannot end after a certificate expiration","technicalCode":"subscription.endingAt"}
```

**Impact**: Any mTLS application using only permanent certificates (the most common setup) cannot have its subscription updated with an end date.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/application_certificates/ApplicationCertificatesUpdateDomainServiceImpl.java
```diff
@@ -53,36 +55,36 @@ public void updateActiveMTLSSubscriptions(String applicationId) {
...
 
-    private String createEncodedCertificate(List<ClientCertificate> certificates) {
+    private String encodeCertificates(List<ClientCertificate> certificates) {
         if (certificates.size() == 1) {
-            String pem = certificates.getFirst().getCertificate();
+            String pem = certificates.getFirst().certificate();
             return Base64.getEncoder().encodeToString(pem.getBytes(StandardCharsets.UTF_8));
         } else {
-            byte[] pkcs7Bundle = PKCS7Utils.createBundle(certificates.stream().map(ClientCertificate::getCertificate).toList());
+            byte[] pkcs7Bundle = PKCS7Utils.createBundle(certificates.stream().map(ClientCertificate::certificate).toList());
```
carlos-andres-osorio, line 87: any corrupt cert in DB permanently breaks updateActiveMTLSSubscriptions with no cert-level diagnostic

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
benoitgravitee: Agreed. I'll make it clearer

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
benoitgravitee: I checked. DB contains the value in uppercase.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
benoitgravitee: Well spotted... Ok will add a check

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/application_certificates/ApplicationCertificatesUpdateDomainServiceImpl.java
benoitgravitee: Not really.
If you have come to call this, either the API created the cert and your are safe, or the Application, hence the cert were added, updated. Either way you cannot call this on bad certs.
But there is one case that is kinda not great, especially with Mongo, where there are no transactions:
You remove all certs from the app. You will get no errors at the application level, because we only check that there is at least one cert that works until the subscription ends.
But... calling this method will fail. DB state will => no certs, but subscriptions will remain untouched. You need to update the application to update the application in order to fix the issue. It is not ideal, but is work for the API (individual certs) and for GKO (certs in app).
IMO we should refactor the whole shazam to have a proper service that does it all, but it means we have to get rid of the Application Service as it is today, and add UseCase for /applications endpoints... that is a lot of work but it would be consistent and we would use the correct abstraction. I don't like the idea that CrudService do a lot checks for instance.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
carlos-andres-osorio: I had this issue while sending a request. I'll try to repro to understand how did I get here.

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoaderManager.java
```diff
@@ -33,63 +38,123 @@
 @CustomLog
 public class SubscriptionTrustStoreLoaderManager {
 
-    private final Map<String, SubscriptionTrustStoreLoader> subscriptionTrustStoreLoaders = new ConcurrentHashMap<>();
-    private final Map<String, Subscription> subscriptionsByApi = new ConcurrentHashMap<>();
+    private final Map<String, Set<SubscriptionCertificate>> certificates = new ConcurrentHashMap<>();
+    private final Map<CacheKey, Subscription> subscriptions = new ConcurrentHashMap<>();
+    private final Multimap<CacheKey, SubscriptionTrustStoreLoader> truststoreLoaders = MultimapBuilder.hashKeys().hashSetValues().build();
```
carlos-andres-osorio, line 44: **Concurrency issue**: `truststoreLoaders` is a Guava `HashMultimap` which is **not thread-safe**. It is accessed from within `computeIfPresent`/`computeIfAbsent` lambdas on the `certificates` ConcurrentHashMap, but those lambdas only hold the segment lock for a single key in `certificates` — concurrent registrations for *different* subscription IDs will access `truststoreLoaders` without any mutual exclusion, which can corrupt its internal state.

Consider wrapping it with `Multimaps.synchronizedMultimap()`, or switching to a `ConcurrentHashMap<CacheKey, Set<SubscriptionTrustStoreLoader>>` to match the thread-safety level of the other two maps.

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoaderManager.java
```diff
@@ -33,63 +38,123 @@
...
+                toUpdate.forEach(s -> subscriptions.put(new CacheKey(s), subscription));
+
+                // update
+                var state = new HashSet<>(currentState);
+                state.addAll(toAdd);
+                state.removeAll(toRemove);
+                return state;
+            } catch (MalformedCertificateException e) {
+                log.error(e.getMessage(), e.getCause());
+                // keep the current state if anything goes wrong
```
carlos-andres-osorio, line 81: **Partial state corruption on exception**: If `MalformedCertificateException` is thrown partway through the `toAdd` loop (line 67-69), some loaders have already been registered with servers and inserted into `truststoreLoaders` and `subscriptions`. The catch block returns `currentState` to preserve the `certificates` map entry, but the other two maps are already mutated — creating an inconsistent state.

For example, if `toAdd` has certs [A, B, C] and B throws:
- A is already in `truststoreLoaders` and `subscriptions`
- `certificates` still has the old state (no A, B, or C)
- No way to clean up A from `truststoreLoaders`/`subscriptions`

The same issue exists in `computeIfAbsent` below (lines 88-97).

Consider either:
1. Staging new loaders in a temporary collection first, then committing to the real maps only after all succeed, or
2. Adding rollback logic in the catch block that unregisters any loaders already registered from `toAdd` before the failure.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
```diff
@@ -639,40 +657,8 @@ public ApplicationEntity update(
...
-                                .endsAt(new Date())
-                                .build()
-                        )
-                    );
-                    // Mark new certificate for creation after application update
-                    newCertificateToCreate = newCertificate;
-                }
-            }
+            // Create, update, delete client certificate regarding of the new application state
+            syncClientCertificates(executionContext, applicationId, updateApplicationEntity);
```
carlos-andres-osorio, line 651: **Missing gateway sync**: After `syncClientCertificates` creates/deletes certificates, `updateActiveMTLSSubscriptions` is never called.

This causes two potential issues:
- The subscriptions associated with the app won't be updated until the next cron execution, which by default is 1 day.
- If the change to an application is to delete all certs, the cron job will not run for the app, and the subscription will never be updated.

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoaderManager.java
```diff
@@ -33,63 +38,123 @@
...
+                // update the subscription that may have changed
+                toUpdate.forEach(s -> subscriptions.put(new CacheKey(s), subscription));
+
+                // update
+                var state = new HashSet<>(currentState);
+                state.addAll(toAdd);
+                state.removeAll(toRemove);
+                return state;
+            } catch (MalformedCertificateException e) {
+                log.error(e.getMessage(), e.getCause());
```
carlos-andres-osorio: **nit**: `log.error(e.getMessage(), e.getCause())` logs only the inner cause's stack trace, losing the wrapping `MalformedCertificateException` context. Same on line 94.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
carlos-andres-osorio: This also applies to request handler `ApiSubscriptionsResource.updateApiSubscription`

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoaderManager.java
benoitgravitee: Not an issue. 
This is only for register and unregister, called by the SubscriptionCache service. What is calling the subscription cache service for register/unregister is synchronous (sync process).
I'll add comment

benoitgravitee: See commit 51729c2b2bfa7ba51dfd261213dc7c8ce16aec52

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoaderManager.java
benoitgravitee: a8f9d441418c87b799eee728928c976ca0b3820e

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
benoitgravitee: There is an onion architecture issue here. But it is too big to fix, we would need to move the application endpoint to v2 API and add usecases validators etc... and our coverage it not sufficient.

But it called by the ImportApplicationCRDUseCase, same for subscriptions (but here it is not an issue either because the cert cannot change unless it is changed by the application).

We really need to consider refactoring, after we merge, that should something to do. Be more align with onion archi espacially for the certificates endpoint. GKO/Automation are pretty much covered at the usecase level.

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoaderManager.java
benoitgravitee: https://github.com/gravitee-io/gravitee-api-management/commit/a8f9d441418c87b799eee728928c976ca0b3820e

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
benoitgravitee: I have reasons to believe this is not an issue... that would be a bug in APIM if that field can be uppercase or lower case...

carlos-andres-osorio (approved):

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
```diff
@@ -115,199 +107,223 @@ public void register(final Subscription subscription) {
...
     }
 
-    private void registerFromId(final Subscription subscription) {
-        String cacheKey = subscription.getId();
-        synchronized (lockFor(cacheKey)) {
-            cacheBySubscriptionId.put(cacheKey, subscription);
-            cacheBySubscriptionIdAll.computeIfAbsent(cacheKey, k -> ConcurrentHashMap.newKeySet()).add(subscription);
-            addKeyForApi(subscription.getApi(), cacheKey);
-        }
+    private void updateIdentityCache(Subscription subscription, Map<String, Subscription> cache) {
```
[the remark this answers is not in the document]
benoitgravitee: it is called two times that's not worth a refactoring

PR #15307, opened 2026-02-19
fix: return a bad request when parameters cannot be parsed

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-2396

This is to avoid this very weird 500 error with a "404 not found" message when parameters cannot be parsed.


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/analytics/dashboards/DashboardsResourceTest.java
```diff
@@ -26,7 +25,6 @@
 import assertions.MAPIAssertions;
 import fixtures.DashboardFixtures;
 import inmemory.DashboardCrudServiceInMemory;
-import io.gravitee.rest.api.management.v2.rest.model.analytics.engine.CreateDashboard;
 import io.gravitee.rest.api.management.v2.rest.model.analytics.engine.CustomInterval;
```
carlos-andres-osorio, line 28: The import io.gravitee.rest.api.management.v2.rest.model.analytics.engine.CreateDashboard was removed, but new CreateDashboard() on line 157 still compiles — it now resolves to the        
enclosing @Nested class CreateDashboard (the test class), not the model DTO.

carlos-andres-osorio (approved):

PR #15323, opened 2026-02-20
feat: add get subscription form endpoint for portal api

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-12845

## Description

Part of: https://gravitee.atlassian.net/browse/APIM-12636

First PR in the subscription form flow: expose an endpoint for the portal to **fetch** the subscription form from the backend. No UI nor persistence of form data yet.

### Included in this PR

- New Portal API endpoint to get subscription form (e.g. `GET /environments/{envId}/subscription-form`)
- **Endpoint description:** always 200 with a `SubscriptionForm` object. When the form is **enabled**, the response includes `id`, `gmdContent`, and `enabled: true`. When the form is **disabled**, the response is a minimal object with only `enabled: false` (no `id` or `gmdContent`), so the frontend can treat it as “no form to show” without receiving unnecessary content. **404** is returned only when no subscription form exists for the environment.
- OpenAPI spec
- Unit tests for the resource (enabled form, disabled form → 200 + minimal object, not found → 404)

### Excluded from this PR (will be done in subsequent PRs)

- Rendering the subscription form on the subscription checkout page (portal UI)
- Sending / persisting subscription form data (metadata) when creating or updating a subscription
- Metadata sanitization/validation for subscription inputs

PR #15329, opened 2026-02-20
feat: render subscription form on subscription checkout

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-12846

## Description

Part of: https://gravitee.atlassian.net/browse/APIM-12636

This PR follows the work in [PR 1 – get subscription form endpoint](https://github.com/gravitee-io/gravitee-api-management/pull/15323) and **displays** the subscription form on the subscription checkout page in the portal. Form is loaded from the new portal API; submission and persistence of form data are in the next PR.

### Included in this PR

- Portal: fetch subscription form via portal API and render it on the subscription checkout page
- Handle API contract: show form when response is 200 with `gmdContent`; when 404, treat as no form (do not render).
- Updates to `subscribe-to-api-checkout` component (template, styles, harness)
- Entity `subscription-form` and portal service changes to support the new API
### Excluded from this PR (will be done in subsequent PRs)

- Sending submitted form data as subscription metadata when creating/updating subscription
- Metadata sanitization for subscription inputs

## Additional Context

https://github.com/user-attachments/assets/9bb1141d-a6e3-40d1-b42f-1832d3bcb496


gravitee-apim-console-webui/src/management/api/component/plan/1-general-step/plan-edit-general-step.component.html
```diff
@@ -67,7 +67,7 @@ <h2>Subscriptions</h2>
     </gio-form-slide-toggle>
 
     <gio-form-slide-toggle class="commentRequired-field">
-      Consumer must provide a comment when subscribing to the plan
+      Consumer must provide a comment when subscribing to the plan (classic portal only)
```
kovaren: Just to give it a more product-y feel 😊

```suggestion
      Consumer must provide a comment when subscribing to the plan (Classic Portal only)

File: gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.html [hunk 2]
```diff
@@ -31,126 +31,122 @@ <h1 class="next-gen-h3" i18n="@@subscriptionDetailsHeader">Subscription Details<
       </app-loader>
     </div>
   } @else {
-    <div class="subscription-info__content__subsections">
```
kovaren, line 34: Do we have to remove the subsections? This component is not only used in the checkout but also in subscription details where it's useful to have these data pieces spread out horizontally in columns. Normally, the subsections are responsive so if there's not enough horizontal space they should align themselves vertically.

gravitee-apim-portal-webui-next/src/app/api/subscribe-to-api/subscribe-to-api-checkout/subscribe-to-api-checkout.component.html
```diff
@@ -16,79 +16,62 @@
...
-    </div>
+    }
+  </div>
+
+  @if (subscriptionForm) {
+    <section class="subscribe-to-api-checkout__form" aria-label="Subscription form">
+      <gmd-form-host [content]="subscriptionForm.gmdContent">
+        <gmd-viewer [content]="subscriptionForm.gmdContent" />
+      </gmd-form-host>
+    </section>
```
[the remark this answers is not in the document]
kovaren: I might be missing some context but is the `gmd-form-host` necessary in this case?

gravitee-apim-portal-webui-next/src/app/api/subscribe-to-api/subscribe-to-api.component.ts
```diff
@@ -84,19 +88,52 @@ interface CheckoutData {
...
+  private readonly matDialog = inject(MatDialog);
+  private readonly currentApplicationsPage = new BehaviorSubject(1);
+  private readonly subscriptionFormSignal = toSignal(this.portalService.getSubscriptionForm(), { initialValue: null });
+  private readonly subscriptionFormSync = effect(() => {
+    const form = this.subscriptionFormSignal();
+    this.subscriptionForm.set(form);
+    if (!form) {
+      this.store.reset();
+    }
+  });
```
kovaren: Writing to signals inside an effect is generally considered an anti-pattern. Also, we don’t need to assign the effect to a variable here.

```suggestion

File: gravitee-apim-portal-webui-next/src/app/api/subscribe-to-api/subscribe-to-api.component.ts [hunk 5]
```diff
@@ -84,19 +88,52 @@ interface CheckoutData {
...
 
   currentStep = signal(1);
   currentPlan = signal<Plan | undefined>(undefined);
   currentApplication = signal<Application | undefined>(undefined);
-
   message = signal<string>('');
   applicationApiKeyMode = signal<'EXCLUSIVE' | 'SHARED' | 'UNSPECIFIED' | null>(null);
   subscriptionInProgress = signal<boolean>(false);
   showApiKeyModeSelection = signal<boolean>(false);
+  subscriptionForm = signal<SubscriptionForm | null>(null);
```
kovaren: We can directly assign the form here

```suggestion
  subscriptionForm = toSignal(
    this.portalService.getSubscriptionForm().pipe(
      tap((form) => {
        if (!form) {
          this.store.reset();
        }
      })
    ),
    { initialValue: null }
  );

kovaren (changes requested):

kovaren:
The paddings of the form seem a bit too big in the video. Is it because the mat-card borders are invisible or is it the form itself?

gravitee-apim-portal-webui-next/src/app/api/subscribe-to-api/subscribe-to-api-checkout/subscribe-to-api-checkout.component.html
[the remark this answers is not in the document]
JedrzejJanasiak: `gmd-form-host` still serves a purpose here, though a slightly limited one than before my changes in [this PR](https://github.com/gravitee-io/gravitee-api-management/pull/15268)

It has an internal effect() that calls store.reset() whenever its content input changes. This handles the case where a user navigates back and picks a different plan, then the form state (filled-in fields) gets cleared when new form content loads. The parent component doesn't cover this scenario: it only resets the store when subscriptionForm becomes null (i.e. no form at all), not when it changes from one form to another.

gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.html
JedrzejJanasiak: I brought back the original layout for the component.

It's fine in most cases, but when there's no comment section and only the subscription form below, the two grids end up aligning side by side, which looks like this: 
image

Jourdi suggested having all details in a single column, but this component is also used in the subscription details view, and that's actually what causes this side effect. Over there, with more data visible, the two-column layout makes more sense. If you're ok with how it looks on the screenshot, I'd leave it as is.

JedrzejJanasiak:
> The paddings of the form seem a bit too big in the video. Is it because the mat-card borders are invisible or is it the form itself?

Yeah, the card outline color matches the card background, and that's why it looks like this. Here' you can see the padding looks normal: 
image

gravitee-apim-portal-webui-next/src/app/api/subscribe-to-api/subscribe-to-api-checkout/subscribe-to-api-checkout.component.html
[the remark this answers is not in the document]
kovaren: Is this a consequence of loading the form inside `subscribe-to-api.component` instead of `subscribe-to-api-checkout.component`? If yes then using `gmd-form-host` here feels more like a workaround than a necessity.

If the form is only ever displayed in the checkout, could `subscribe-to-api-checkout.component` be in charge of loading the form instead of its parent?

gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.html
kovaren: Yep, this follows the intended behavior with the subsections and I'd argue that it makes sense in this component as well, since without the comment section there's now a lot of empty space to show the props.
@jourdiw WDYT?

gravitee-apim-portal-webui-next/src/app/api/subscribe-to-api/subscribe-to-api-checkout/subscribe-to-api-checkout.component.html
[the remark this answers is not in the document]
kovaren: As discussed with @jourdiw this is not a deal breaker, let's merge and have a discussion about potential improvements later.

gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.html
kovaren: As discussed with @jourdiw it's all good.

kovaren (approved):

PR #15333, opened 2026-02-20
feat: add API Product APIs list and add API dialog

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-12361
https://gravitee.atlassian.net/browse/APIM-12365

## Description

This PR adds the **APIs** page within the API Product detail flow, allowing API Product Managers to view all associated APIs and search for/add eligible APIs. The page integrates with existing navigation and reuses the v2 APIs `_search` endpoint with V4 HTTP Proxy and `allowedInApiProducts` filtering.

---

## User Stories & Acceptance Criteria

| Criteria | Implementation |
|----------|----------------|
| APIs page lists all associated APIs | Table populated from product `apiIds` via `ApiProductV2Service` + `ApiV2Service.get()` |
| API name, version, and status shown | Columns: Name, Context Path, Definition, Version; "Deprecated" badge when applicable |
| Navigation consistent with Console UX | APIs tab added to submenu alongside Configuration; uses `gio-table-wrapper` and `gio-submenu` |

| Criteria | Implementation |
|----------|----------------|
| Search UI available when adding APIs | Autocomplete search in Add API dialog |
| Reuses existing API `_search` endpoint | `ApiV2Service.search()` → `POST /environments/{envId}/apis/_search` |
| Only V4 HTTP Proxy APIs returned | Search params: `apiTypes: ['V4_HTTP_PROXY']`, `allowedInApiProducts: true` |
| Supports filtering and selection | Client-side filtering of already-selected and existing product APIs; multi-select with chips |
| UI and backend validation aligned | UI validates duplicates; backend `ValidateApiProductService` enforces `allowedInApiProducts` |

---

### Backend (Reference – No Changes in This PR)

| Area | Path |
|------|------|
| API Products OpenAPI | `gravitee-apim-rest-api-management-v2/.../openapi-api-products.yaml` |
| APIs search OpenAPI | `gravitee-apim-rest-api-management-v2/.../openapi-apis.yaml` |
| Validate API Product | `gravitee-apim-rest-api-service/.../ValidateApiProductService.java` |

---

### `ApiProductApisComponent` (`api-product-apis.component.spec.ts`)

| Test | Description |
|------|-------------|
| `should create` | Component initializes |
| `should display empty state when no APIs in product` | Shows empty state when `apiIds` is empty |
| `should display table with API rows when product has APIs` | Table shows name, context path for each API |
| `should display Loading... then data` | Loading state, then table after API fetch |
| `should handle error when loading API product` | SnackBar error on failed load |
| `should have Add API button` | Add API button present |
| `should open Add API dialog when Add API clicked` | Dialog opens with `apiProductId`, `existingApiIds` |
| `should filter APIs by search term` | Client-side filter updates table (e.g. "Orders") |

### `ApiProductAddApiDialogComponent` (`add-api-dialog/api-product-add-api-dialog.component.spec.ts`)

| Test | Description |
|------|-------------|
| `should create` | Dialog initializes |
| `should close dialog on cancel` | Cancel closes dialog |
| `should disable submit when no APIs selected` | Submit disabled when selection is empty |
| `should call close with true on successful submit` | Success flow: GET product → PUT update → snackbar → close |
| `should show error when API already in product` | SnackBar error and no close when API already in product |

---
https://github.com/user-attachments/assets/d95556eb-ea3f-4a9f-8d82-7599631ac1d8

### DIFF
--- a/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.html
+++ b/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.html
@@ -1,13 +1,13 @@
 <!--
 
-    Copyright (C) 2026 The Gravitee team (http://gravitee.io)
-    
+    Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
     You may obtain a copy of the License at
-    
+
             http://www.apache.org/licenses/LICENSE-2.0
-    
+
     Unless required by applicable law or agreed to in writing, software
     distributed under the License is distributed on an "AS IS" BASIS,
     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.html
+++ b/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.html
@@ -1,0 +1,115 @@
+<!--
+
+    Copyright (C) 2026 The Gravitee team (http://gravitee.io)
+    
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+    
+            http://www.apache.org/licenses/LICENSE-2.0
+    
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+
+<h2 mat-dialog-title>Add API</h2>
+
+<form autocomplete="off" (ngSubmit)="submit()">
+  <mat-dialog-content class="dialog-content">
+    <gio-banner-info>
+      Can't see all your APIs?
+      <span gioBannerBody>APIs must have API products enabled before they appear in the list.</span>
+    </gio-banner-info>
+
+    @if (selectedApis.length > 0) {
+      <div class="selected-apis-section">
+        <h3 class="selected-apis-section__header">APIs added to this product</h3>
+        <mat-chip-set class="selected-apis-section__chips">
+          @for (api of selectedApis; track api.id) {
+            <mat-chip
+              class="selected-apis-section__chip"
+              [removable]="true"
+              (removed)="removeApi(api)"
+              [attr.aria-label]="'Remove ' + api.name"
+              [attr.data-testid]="'selected-api-chip'"
+            >
+              <span class="selected-apis-section__chip-name">{{ api.name }}</span>
+              @if (getContextPath(api); as contextPath) {
+                <span class="selected-apis-section__chip-path">{{ contextPath }}</span>
+              }
+              @if (api.lifecycleState === 'DEPRECATED') {
+                <span class="gio-badge-error">Deprecated</span>
+              }
+              <mat-icon matChipRemove>cancel</mat-icon>
+            </mat-chip>
+          }
+        </mat-chip-set>
+      </div>
+    }
+
+    <mat-form-field appearance="outline" class="form-field">
+      <mat-label>Add APIs</mat-label>
+      <input matInput type="search" [matAutocomplete]="auto" [formControl]="searchApiControl" [attr.data-testid]="'api-select-input'" />
+      @if (searchApiControl.value) {
+        <button matSuffix mat-icon-button aria-label="Clear input" (click)="resetSearchTerm()">
+          <mat-icon>close</mat-icon>
+        </button>
+      }
+      <mat-autocomplete #auto="matAutocomplete" (optionSelected)="selectAPI()" [displayWith]="displayFn">
+        @if (filteredOptions$ | async; as filteredOptions) {
+          @if (filteredOptions.length > 0) {
+            @for (option of filteredOptions; track option.id) {
+              <mat-option [value]="option" [attr.data-testid]="'api-select-option'">
+                <span class="autocompleteRow">
+                  <gio-avatar
+                    class="autocompleteRow__avatar"
+                    [src]="option._links?.pictureUrl"
+                    [name]="option.name + ' ' + option.apiVersion"
+                    [size]="28"
+                    [roundedBorder]="true"
+                  ></gio-avatar>
+                  <span class="autocompleteRow__content">
+                    <span class="autocompleteRow__name">{{ option.name }}</span>
+                    @if (getContextPath(option); as contextPath) {
+                      <span class="autocompleteRow__path">{{ contextPath }}</span>
+                    }
+                  </span>
+                  @if (option.lifecycleState === 'DEPRECATED') {
+                    <span class="autocompleteRow__badge gio-badge-error">Deprecated</span>
+                  }
+                </span>
+              </mat-option>
+            }
+          } @else {
+            @if (hasNonEmptySearchTerm) {
+              <mat-option disabled>No APIs matching search criteria</mat-option>
+            }
+          }
+        }
+      </mat-autocomplete>
+    </mat-form-field>
+  </mat-dialog-content>
+
+  <mat-dialog-actions class="actions">
+    <button type="button" mat-button (click)="onCancelClick()" [disabled]="isSubmitting" [attr.data-testid]="'cancel-button'">
+      Cancel
+    </button>
+    <button
+      type="submit"
+      [disabled]="selectedApis.length === 0 || isSubmitting"
+      mat-raised-button
+      color="primary"
+      [attr.data-testid]="'submit-button'"
+    >
+      @if (isSubmitting) {
+        Adding...
+      } @else {
+        Add API
+      }
+    </button>
+  </mat-dialog-actions>
+</form>
--- a/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.html
+++ b/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.html
@@ -25,21 +25,25 @@
       <span gioBannerBody>APIs must have API products enabled before they appear in the list.</span>
     </gio-banner-info>
 
-    @if (selectedApis.length > 0) {
+    @if (validationError(); as error) {
+      <gio-banner-error>{{ error }}</gio-banner-error>
+    }
+
+    @if (selectedApisWithContext().length > 0) {
       <div class="selected-apis-section">
         <h3 class="selected-apis-section__header">APIs added to this product</h3>
         <mat-chip-set class="selected-apis-section__chips">
-          @for (api of selectedApis; track api.id) {
+          @for (api of selectedApisWithContext(); track api.id) {
             <mat-chip
               class="selected-apis-section__chip"
               [removable]="true"
               (removed)="removeApi(api)"
               [attr.aria-label]="'Remove ' + api.name"
-              [attr.data-testid]="'selected-api-chip'"
+              [attr.data-testid]="'selected-api-chip-' + api.id"
             >
               <span class="selected-apis-section__chip-name">{{ api.name }}</span>
-              @if (getContextPath(api); as contextPath) {
-                <span class="selected-apis-section__chip-path">{{ contextPath }}</span>
+              @if (api.contextPath) {
+                <span class="selected-apis-section__chip-path">{{ api.contextPath }}</span>
               }
               @if (api.lifecycleState === 'DEPRECATED') {
                 <span class="gio-badge-error">Deprecated</span>
@@ -61,30 +65,28 @@
       }
       <mat-autocomplete #auto="matAutocomplete" (optionSelected)="selectAPI()" [displayWith]="displayFn">
         @if (filteredOptions$ | async; as filteredOptions) {
-          @if (filteredOptions.length > 0) {
-            @for (option of filteredOptions; track option.id) {
-              <mat-option [value]="option" [attr.data-testid]="'api-select-option'">
-                <span class="autocompleteRow">
-                  <gio-avatar
-                    class="autocompleteRow__avatar"
-                    [src]="option._links?.pictureUrl"
-                    [name]="option.name + ' ' + option.apiVersion"
-                    [size]="28"
-                    [roundedBorder]="true"
-                  ></gio-avatar>
-                  <span class="autocompleteRow__content">
-                    <span class="autocompleteRow__name">{{ option.name }}</span>
-                    @if (getContextPath(option); as contextPath) {
-                      <span class="autocompleteRow__path">{{ contextPath }}</span>
-                    }
-                  </span>
-                  @if (option.lifecycleState === 'DEPRECATED') {
-                    <span class="autocompleteRow__badge gio-badge-error">Deprecated</span>
+          @for (option of filteredOptions; track option.id) {
+            <mat-option [value]="option" [attr.data-testid]="'api-select-option-' + option.id">
+              <span class="autocompleteRow">
+                <gio-avatar
+                  class="autocompleteRow__avatar"
+                  [src]="option._links?.pictureUrl"
+                  [name]="option.name + ' ' + option.apiVersion"
+                  [size]="28"
+                  [roundedBorder]="true"
+                ></gio-avatar>
+                <span class="autocompleteRow__content">
+                  <span class="autocompleteRow__name">{{ option.name }}</span>
+                  @if (option.contextPath) {
+                    <span class="autocompleteRow__path">{{ option.contextPath }}</span>
                   }
                 </span>
-              </mat-option>
-            }
-          } @else {
+                @if (option.lifecycleState === 'DEPRECATED') {
+                  <span class="autocompleteRow__badge gio-badge-error">Deprecated</span>
+                }
+              </span>
+            </mat-option>
+          } @empty {
             @if (hasNonEmptySearchTerm) {
               <mat-option disabled>No APIs matching search criteria</mat-option>
             }
@@ -95,21 +97,9 @@
   </mat-dialog-content>
 
   <mat-dialog-actions class="actions">
-    <button type="button" mat-button (click)="onCancelClick()" [disabled]="isSubmitting" [attr.data-testid]="'cancel-button'">
-      Cancel
-    </button>
-    <button
-      type="submit"
-      [disabled]="selectedApis.length === 0 || isSubmitting"
-      mat-raised-button
-      color="primary"
-      [attr.data-testid]="'submit-button'"
-    >
-      @if (isSubmitting) {
-        Adding...
-      } @else {
-        Add API
-      }
+    <button type="button" mat-button (click)="onCancelClick()" [attr.data-testid]="'cancel-button'">Cancel</button>
+    <button type="submit" [disabled]="selectedApis().length === 0" mat-raised-button color="primary" [attr.data-testid]="'submit-button'">
+      Add API
     </button>
   </mat-dialog-actions>
 </form>
--- a/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.html
+++ b/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.html
@@ -87,7 +87,7 @@
               </span>
             </mat-option>
           } @empty {
-            @if (hasNonEmptySearchTerm) {
+            @if (hasNonEmptySearchTerm()) {
               <mat-option disabled>No APIs matching search criteria</mat-option>
             }
           }
--- a/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.scss
+++ b/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.scss
@@ -1,0 +1,115 @@
+@use 'sass:map';
+@use '@angular/material' as mat;
+@use '@gravitee/ui-particles-angular' as gio;
+
+$typography: map.get(gio.$mat-theme, typography);
+
+:host {
+  gio-banner-info {
+    margin-bottom: 24px;
+  }
+}
+
+.dialog-content {
+  overflow: visible;
+  max-height: none;
+  padding: 0 24px 8px 24px;
+}
+
+.selected-apis-section {
+  margin-bottom: 24px;
+
+  &__header {
+    @include mat.m2-typography-level($typography, subtitle-2);
+    margin: 0 0 12px 0;
+    color: mat.m2-get-color-from-palette(gio.$mat-space-palette, 'darker20');
+  }
+
+  &__chips {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    flex-wrap: wrap;
+    gap: 8px;
+  }
+
+  &__chip {
+    background-color: mat.m2-get-color-from-palette(gio.$mat-dove-palette, darker10);
+    color: mat.m2-get-color-from-palette(gio.$mat-dove-palette, darker10-contrast);
+    min-height: 1.5rem;
+    height: fit-content;
+    padding: 0 4px 0 12px;
+    align-self: center;
+    display: inline-flex;
+    align-items: center;
+    gap: 8px;
+
+    &-avatar {
+      flex-shrink: 0;
+      margin-right: 4px;
+    }
+
+    &-name {
+      @include mat.m2-typography-level($typography, body-2);
+    }
+
+    &-path {
+      @include mat.m2-typography-level($typography, caption);
+      margin-left: 2px;
+      opacity: 0.9;
+      font-style: italic;
+      white-space: nowrap;
+    }
+
+    .gio-badge-error {
+      padding: 2px 8px;
+      line-height: 1.25;
+    }
+  }
+}
+
+.form-field {
+  width: 100%;
+}
+
+.autocompleteRow {
+  display: flex;
+  align-items: center;
+  gap: 12px;
+
+  &__avatar {
+    flex-shrink: 0;
+  }
+
+  &__content {
+    flex: 1;
+    min-width: 0;
+    display: flex;
+    flex-direction: column;
+    gap: 2px;
+  }
+
+  &__name {
+    @include mat.m2-typography-level($typography, body-2);
+  }
+
+  &__path {
+    @include mat.m2-typography-level($typography, caption);
+    color: mat.m2-get-color-from-palette(gio.$mat-space-palette, 'darker20');
+  }
+
+  &__badge {
+    flex-shrink: 0;
+    margin-left: auto;
+    padding: 2px 8px;
+    line-height: 1.25;
+  }
+}
+
+.actions {
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+  gap: 8px;
+  padding: 16px 24px 24px 24px;
+}
--- a/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.scss
+++ b/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.scss
@@ -57,7 +57,6 @@ $typography: map.get(gio.$mat-theme, typography);
       @include mat.m2-typography-level($typography, caption);
       margin-left: 2px;
       opacity: 0.9;
-      font-style: italic;
       white-space: nowrap;
     }
 
--- a/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.ts
+++ b/gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2026 The Gravitee team (http://gravitee.io)
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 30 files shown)


gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.html
```diff
@@ -0,0 +1,113 @@
...
+                  </span>
+                  @if (option.lifecycleState === 'DEPRECATED') {
+                    <span class="autocompleteRow__badge gio-badge-error">Deprecated</span>
+                  }
+                </span>
+              </mat-option>
+            }
+          } @else if (hasNonEmptySearchTerm) {
+            <mat-option disabled>No APIs matching search criteria</mat-option>
+          }
```
jourdiw: I think you might be able to use the `@empty` annotation here for the message on line 88 instead of checking if there are results 🤔

gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.scss
```diff
@@ -0,0 +1,115 @@
...
+
+    &-name {
+      @include mat.m2-typography-level($typography, body-2);
+    }
+
+    &-path {
+      @include mat.m2-typography-level($typography, caption);
+      margin-left: 2px;
+      opacity: 0.9;
+      font-style: italic;
```
jourdiw: I find the italic a bit hard to read in the badge. I would say with a different color and different size, the path is already visually different so no italic would be needed 👍

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.html
```diff
@@ -0,0 +1,100 @@
...
+        <td mat-cell *matCellDef="let element">{{ element.contextPath }}</td>
+      </ng-container>
+      <ng-container matColumnDef="definition">
+        <th mat-header-cell *matHeaderCellDef>Definition</th>
+        <td mat-cell *matCellDef="let element">{{ element.definition }}</td>
+      </ng-container>
+      <ng-container matColumnDef="version">
+        <th mat-header-cell *matHeaderCellDef>Version</th>
+        <td mat-cell *matCellDef="let element">
+          v{{ element.version }}
```
jourdiw: I would remove the "v" so that users can put in whatever value they want for the versions 🤔

gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.html
```diff
@@ -0,0 +1,113 @@
...
+          @for (api of selectedApis; track api.id) {
+            <mat-chip
+              class="selected-apis-section__chip"
+              [removable]="true"
+              (removed)="removeApi(api)"
+              [attr.aria-label]="'Remove ' + api.name"
+              [attr.data-testid]="'selected-api-chip-' + api.id"
+            >
+              <span class="selected-apis-section__chip-name">{{ api.name }}</span>
+              @if (getContextPath(api); as contextPath) {
```
jourdiw: Instead of calling the function here (which is called every millisecond), could you map the selected APIs to an object used in the template that includes this mapped context path attribute?

gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.ts
```diff
@@ -0,0 +1,197 @@
...
+          map(apisResponse => {
+            const existingIds = this.data.existingApiIds || [];
+            const selectedIds = selectedApis.map(api => api.id);
+            const apis = apisResponse.data || [];
+            return apis.filter(api => !existingIds.includes(api.id) && !selectedIds.includes(api.id));
+          }),
+        );
+      }),
+      tap(() => (this.isApiSelected = false)),
+      takeUntilDestroyed(this.destroyRef),
```
jourdiw: This is not needed if it is being passed to an AsyncPipe in the template 👍

gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.ts
```diff
@@ -0,0 +1,197 @@
...
+        return this.apiService.search({ query: term, apiTypes: ['V4_HTTP_PROXY'], allowedInApiProducts: true }).pipe(
+          map(apisResponse => {
+            const existingIds = this.data.existingApiIds || [];
+            const selectedIds = selectedApis.map(api => api.id);
+            const apis = apisResponse.data || [];
+            return apis.filter(api => !existingIds.includes(api.id) && !selectedIds.includes(api.id));
+          }),
+        );
+      }),
+      tap(() => (this.isApiSelected = false)),
```
jourdiw: Does this mean that every time there is a change the search, all apis are unselected?

gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.ts
jourdiw, line 1: There seems to be a strong correlation between `selectedApis` and `selectedApis$` and `isApiSelected`. Is there a way to use signals to handle what these objects are trying to accomplish?

gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.ts
```diff
@@ -0,0 +1,197 @@
...
+          return;
+        }
+        const apiNames = this.selectedApis.map(api => api.name).join(', ');
+        this.snackBarService.success(
+          this.selectedApis.length === 1
+            ? `API "${apiNames}" has been added to the API Product`
+            : `${this.selectedApis.length} APIs have been added to the API Product`,
+        );
+        this.dialogRef.close(true);
+      });
```
jourdiw: Typically the parent component will handle the success/error with dialogs. WDYT about moving this logic to `api-product-apis`?

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.html
```diff
@@ -0,0 +1,100 @@
...
+          </button>
+        </td>
+      </ng-container>
+
+      <tr mat-header-row *matHeaderRowDef="displayedColumns"></tr>
+      <tr mat-row *matRowDef="let row; columns: displayedColumns"></tr>
+
+      <tr class="mat-mdc-row mdc-data-table__row" *matNoDataRow>
+        <td class="mat-mdc-cell mdc-data-table__cell" [attr.colspan]="displayedColumns.length">
+          @if (!isLoadingData && apisTableDS.length === 0) {
```
jourdiw: It will always be `apisTableDS.length === 0` as true here since it is the `matNoDataRow` 👍

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+    name: 'Payments API',
+    apiVersion: '1.0',
+    listeners: [
+      {
+        type: 'HTTP',
+        paths: [{ path: '/payments' }],
+        entrypoints: [{ type: 'http-proxy' }],
+      },
+    ] as any,
+  }) as Api;
```
jourdiw: I don't think you need to cast this to as `any` and `Api` 🤔

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+          provide: ActivatedRoute,
+          useValue: {
+            snapshot: { params: {}, queryParams: {} },
+            params: new BehaviorSubject({}),
+            queryParams: queryParams$.asObservable(),
+            parent: {
+              snapshot: { params: { apiProductId: API_PRODUCT_ID }, queryParams: {} },
+              params: new BehaviorSubject({ apiProductId: API_PRODUCT_ID }),
+              queryParams: queryParams$.asObservable(),
+            } as any,
```
jourdiw:
```suggestion
            },
```

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+      }),
+    );
+  });
+
+  it('should filter APIs by search term', async () => {
+    await initComponent(fakeApiProduct, [fakeApi1, fakeApi2]);
+
+    const tableWrapper = await loader.getHarness(GioTableWrapperHarness);
+    await tableWrapper.setSearchValue('Orders');
+    await new Promise(r => setTimeout(r, 450));
```
jourdiw: Is this promise necessary?

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+    const table = await loader.getHarness(MatTableHarness.with({ selector: 'table' }));
+    const rows = await table.getRows();
+    expect(rows.length).toBe(1);
+    const rowCells = await rows[0].getCellTextByIndex();
+    expect(rowCells[1]).toContain('Orders API');
+  });
+
+  async function initComponent(apiProduct: ApiProduct, apis: Api[] = []) {
+    fixture.detectChanges();
+    await new Promise(r => setTimeout(r, 150));
```
jourdiw: Is this really necessary?

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.spec.ts
jourdiw, line 1: I would try to avoid any timeouts --> they will add time to tests which cost money for the CI. If you are unable to make it work with fixture.detectChanges and fixture.whenStable, then maybe fakeAsync can help you jump the timeouts.

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.ts
```diff
@@ -0,0 +1,289 @@
...
+  ngOnInit(): void {
+    const getApiProductId = (route: ActivatedRoute): string | null => {
+      if (route.snapshot.params['apiProductId']) {
+        return route.snapshot.params['apiProductId'];
+      }
+      if (route.parent) {
+        return getApiProductId(route.parent);
+      }
+      return null;
+    };
```
jourdiw: Why not just have something like this instead of a function? Shouldn't the params always be available at the child level?
```
apiProductId = toSignal(this.activatedRoute.params.pipe(map(...))
```

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.ts
```diff
@@ -0,0 +1,289 @@
...
+            this.snackBarService.error(error.error?.message || 'An error occurred while loading APIs');
+            return of([]);
+          }),
+        );
+      }),
+    );
+  }
+
+  private processAndDisplayApis(apis: (Api | null)[]): void {
+    let filteredApis = apis.filter((api): api is Api => api !== null);
```
jourdiw, line 168: How can an api be null in the list?

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.ts
```diff
@@ -0,0 +1,289 @@
...
+
+  onAddApi(): void {
+    this.apiProductV2Service
+      .get(this.apiProductId)
+      .pipe(
+        catchError(error => {
+          this.snackBarService.error(error.error?.message || 'An error occurred while loading the API Product');
+          return of(null);
+        }),
+        filter((apiProduct): apiProduct is NonNullable<typeof apiProduct> => apiProduct != null),
```
jourdiw: This seems a bit too much casting for the simple filter of it is null, don't continue.

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.ts
```diff
@@ -0,0 +1,289 @@
...
+      .subscribe();
+  }
+
+  onAddApi(): void {
+    this.apiProductV2Service
+      .get(this.apiProductId)
+      .pipe(
+        catchError(error => {
+          this.snackBarService.error(error.error?.message || 'An error occurred while loading the API Product');
+          return of(null);
```
jourdiw: You can return `EMPTY` which ends the Observable 👍

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.ts
```diff
@@ -0,0 +1,289 @@
...
+      .pipe(
+        tap(apis => this.processAndDisplayApis(apis)),
+        takeUntilDestroyed(this.destroyRef),
+      )
+      .subscribe();
+  }
+
+  onAddApi(): void {
+    this.apiProductV2Service
+      .get(this.apiProductId)
```
jourdiw: Why do you re-request the api product?

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.ts
```diff
@@ -0,0 +1,289 @@
...
+      .afterClosed()
+      .pipe(
+        filter((result): result is true => result === true),
+        switchMap(() =>
+          this.apiProductV2Service.get(this.apiProductId).pipe(
+            catchError(error => {
+              this.snackBarService.error(error.error?.message || 'An error occurred while loading the API Product');
+              return of(null);
+            }),
+          ),
```
jourdiw: You repeat this a bit with the catchError. Maybe you can have a shared method for handling the error?

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.ts
```diff
@@ -0,0 +1,289 @@
...
+          const currentApiIds = apiProduct.apiIds || [];
+          const updatedApiIds = currentApiIds.filter(id => id !== api.id);
+          return this.apiProductV2Service.updateApiProductApis(this.apiProductId, updatedApiIds).pipe(
+            catchError(error => {
+              this.snackBarService.error(error.error?.message || 'An error occurred while removing the API');
+              return of(null);
+            }),
+          );
+        }),
+        filter((result): result is NonNullable<typeof result> => result !== null),
```
jourdiw: Since you do not want to continue if there is an error (update api product apis has an error and null is returned), you can do:


```suggestion
        switchMap(apiProduct => {
          const currentApiIds = apiProduct.apiIds || [];
          const updatedApiIds = currentApiIds.filter(id => id !== api.id);
          return this.apiProductV2Service.updateApiProductApis(this.apiProductId, updatedApiIds)
        }),
        catchError(error => {
              this.snackBarService.error(error.error?.message || 'An error occurred while removing the API');
              return of(EMPTY);
            }),
```

gravitee-apim-console-webui/src/management/api-products/configuration/danger-zone/api-product-danger-zone.component.html
jourdiw, line 1: Should any of the danger zone be shown if the user is in read only?
- Do not show it --> code can be simplified
- Show it --> make the buttons disabled and add a tooltip for the "why" (this is how it works in the api danger zone I believe)

gravitee-apim-console-webui/src/management/api-products/configuration/danger-zone/api-product-danger-zone.component.html
```diff
@@ -21,19 +21,31 @@
     <h3 class="danger-card__title">Danger Zone</h3>
     <div class="danger-card__actions">
       <div class="danger-card__actions__action">
-        <span>Remove all the APIs of this API Product.</span>
+        <span>
+          Remove all the APIs of this API Product.
+          @if (!apiProduct().apiIds?.length) {
+            <span>No APIs in this product.</span>
+          }
```
jourdiw: Can this be a tooltip instead?

gravitee-apim-console-webui/src/shared/utils/api.util.ts
```diff
@@ -14,6 +14,15 @@
  * limitations under the License.
  */
 
+import { Api, ApiV4, HttpListener } from '../../entities/management-api-v2';
+
+/** Extracts the context path from V4 HTTP Proxy APIs (HTTP listener's first path). */
+export function getApiContextPath(api: Api | null | undefined): string | null {
```
jourdiw: I think you can use `getApiAccess` in `api-access.util.ts` for this mapping 👍

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource.java
```diff
@@ -321,14 +322,14 @@ public ApisResponse searchApis(
             apiQueryBuilder.addFilter(FIELD_PORTAL_STATUS, apiSearchQuery.getPublished());
         }
 
-        if (apiSearchQuery.getPublished() != null && !apiSearchQuery.getPublished().isEmpty()) {
-            apiQueryBuilder.addFilter(FIELD_PORTAL_STATUS, apiSearchQuery.getPublished());
-        }
-
```
jourdiw, line 327: Why did you remove this?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/ApiSearchServiceImpl.java
```diff
@@ -296,6 +299,13 @@ public Page<GenericApiEntity> search(
             return new Page<>(List.of(), 0, 0, 0);
         }
 
+        // Filter by allowedInApiProducts when requested (source of truth from DB, handles index staleness)
+        apis = filterApisByAllowedInApiProducts(apis, apiQuery);
+
+        if (apis.isEmpty()) {
+            return new Page<>(List.of(), 0, 0, 0);
+        }
```
jourdiw: This seems a bit odd to reassign the value to the variable and do the exact same empty check as above. Can this be simplified?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/ApiSearchServiceImpl.java
```diff
@@ -324,11 +334,71 @@ public Page<GenericApiEntity> search(
             .sorted((o1, o2) -> orderingComparator.compare(o1.getId(), o2.getId()))
             .collect(
                 Collectors.collectingAndThen(Collectors.toList(), apiEntityList ->
-                    new Page<>(apiEntityList, pageable.getPageNumber(), apis.size(), apiIds.size())
+                    new Page<>(apiEntityList, pageable.getPageNumber(), apiEntityList.size(), apiIds.size())
                 )
             );
     }
 
+    private List<Api> filterApisByAllowedInApiProducts(List<Api> apis, io.gravitee.rest.api.service.search.query.Query<?> apiQuery) {
```
jourdiw: This is a lot of logic in quite a long function 😅 I'm not exactly sure if this needs to be so complicated for searching apis...

gravitee-apim-console-webui/src/management/api-products/list/api-product-list.component.ts
jourdiw, line 1: Why did you add the cdr mark for check?

jourdiw (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/ApiSearchServiceImpl.java
```diff
@@ -296,6 +299,13 @@ public Page<GenericApiEntity> search(
             return new Page<>(List.of(), 0, 0, 0);
         }
 
+        // Filter by allowedInApiProducts when requested (source of truth from DB, handles index staleness)
```
phiz71: Why did you add this ?
If the boolean is stored in Lucene, then what you did in the ApisResource (https://github.com/gravitee-io/gravitee-api-management/pull/15333/changes/80d46559ff4da458528a82bd8211cbbec162cbab#diff-a872fff7a41f0f32129e8d6f336d205f286880c2dbad29223a033a145fc74beaR329-R332) should be enough.

Have you tested it ?

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource.java
phiz71: It's a duplicate ;)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/ApiSearchServiceImpl.java
ankita-gupta21: removed it

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/ApiSearchServiceImpl.java
ankita-gupta21: removed it

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/ApiSearchServiceImpl.java
ankita-gupta21: I recently made changes to the open-api and resource sections. Initially, these changes didn’t work. Then I removed the logic  to check again and ran the script manually to run gateway and restapi. I realized that the issue was a build problem, not a code issue. Therefore, **I removed the logic.** 

Also I'll be keeping this 
https://github.com/gravitee-io/gravitee-api-management/commit/80d46559ff4da458528a82bd8211cbbec162cbab#diff-5218a149fd62967dcfe78a9ab692e28d019b4a81b5e786e495a91791479413c5R184

Using a string value ("allowedInApiProducts": "true") would hit the String branch, but that would:
- Conflict with the OpenAPI spec (type: boolean)
- Require changing all clients

gravitee-apim-console-webui/src/management/api-products/list/api-product-list.component.ts
ankita-gupta21: Before i had ChangeDetectionStrategy.OnPush in my component so i had added cdr mark
but i have now refactored it

gravitee-apim-console-webui/src/management/api-products/configuration/danger-zone/api-product-danger-zone.component.html
ankita-gupta21: I’ve removed these conditional buttons since they’re not present in Figma. I’ve kept the button disabled for simplicity.

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.ts
ankita-gupta21: done

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.ts
ankita-gupta21: When loading APIs for an API Product, we:

1. Load the API Product and get its list of API IDs.
2. For each ID, call apiV2Service.get(apiId).
3. Use forkJoin to wait for all those calls to finish.
4. Pass the resulting array to processAndDisplayApis.

Each individual get(apiId) is wrapped in catchError(() => of(null)). So when a fetch fails (API deleted, network issue, 404, etc.), we emit null instead of erroring the whole flow.

That means the array we pass in is (Api | null)[]: some entries are Api objects, others are null for failed fetches.

The apis.filter((api): api is Api => api !== null) step removes those null entries so the table only shows successfully loaded APIs, without failing the entire load because of one bad or missing API.

gravitee-apim-console-webui/src/management/api-products/configuration/danger-zone/api-product-danger-zone.component.html
ankita-gupta21: I have removed this instead and made the button disable based on the condition ` @if (!apiProduct().apiIds?.length) ` , so we don't need ` <span>No APIs in this product.</span>`

gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.ts
```diff
@@ -0,0 +1,153 @@
...
+export class ApiProductAddApiDialogComponent implements OnInit {
+  private readonly injector = inject(Injector);
+  private readonly dialogRef = inject(MatDialogRef<ApiProductAddApiDialogComponent>);
+  private readonly apiService = inject(ApiV2Service);
+  readonly data = inject<ApiProductAddApiDialogData>(MAT_DIALOG_DATA);
+
+  get hasNonEmptySearchTerm(): boolean {
+    const v = this.searchApiControl.value;
+    return typeof v === 'string' && v.trim().length > 0;
+  }
```
jourdiw: Can this be a computed signal?

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.ts
jourdiw, line 1: I noticed one architectural detail we should tweak. Right now, the table is using a local BehaviorSubject (filters$) to manage state, and then pushing those changes to the URL. This means if a user uses the browser's Back/Forward buttons, the URL will change, but the table won't update because we are only reading the route snapshot on init.

Could we refactor this to make the Router the true source of truth? Instead of UI actions updating the Subject, they should just call router.navigate(). Then, we can listen directly to activatedRoute.queryParams to trigger our data loads.

gravitee-apim-console-webui/src/management/api-products/list/api-product-list.component.ts
```diff
@@ -100,54 +102,58 @@ type ApiProductListTableWrapperFilters = GioTableWrapperFilters;
...
+          catchError(error => {
+            this.snackBarService.error(this.getErrorMessage(error, 'An error occurred while loading API Products'));
+            return of(EMPTY_API_PRODUCTS_RESPONSE);
+          }),
+          map((response: ApiProductsResponse) => ({
+            tableDS: this.toApiProductsTableDS(response.data || []),
+            totalCount: response.pagination?.totalCount || 0,
+            isLoading: false,
+          })),
+          startWith(INITIAL_API_PRODUCTS_STATE),
```
jourdiw, line 140: Is it necessary to define the initial state here if you are starting the `apiProducstState` signal with a default value?

gravitee-apim-console-webui/src/shared/utils/api.util.ts
```diff
@@ -14,6 +14,16 @@
  * limitations under the License.
  */
 
+import { getApiAccess } from './api-access.util';
+
+import { Api, ApiV2, ApiV4 } from '../../entities/management-api-v2';
+
+export function getApiContextPath(api: Api | null | undefined): string | null {
```
jourdiw: Is it necessary to create a global method for this? It seems like the output here is a bit specific to your component (returns only the first result of access) 🤔 
However if you feel that it should be global, could you add it to the `api-access` and make it complimentary to the existing method?

gravitee-apim-console-webui/src/management/api-products/list/api-product-list.component.ts
ankita-gupta21: The initial state constant is used in two places, and both are needed:
1. **`initialValue` in `toSignal`** – The signal must have a value before the observable emits. Without this, the signal would be undefined until the first emission.
2. **`startWith()` in the RxJS pipe** – When filters change (e.g. search or pagination), switchMap creates a new inner observable. startWith makes that inner observable emit the loading state immediately, so the UI shows loading instead of stale data while the new request runs.

If we only used initialValue, the loading state would appear on the first load, but when filters change we’d keep showing old data until the new response arrives.
Using a single INITIAL_API_PRODUCTS_STATE constant keeps both usages in sync and avoids duplication.

gravitee-apim-console-webui/src/shared/utils/api.util.ts
ankita-gupta21: I moved `getApiContextPath` into `api-access.util.ts` so it sits next to `getApiAccess`, which it wraps. It’s a small helper that returns the first access path for display (e.g. context path in tables and chips). It’s used in both the API product table and the add-API dialog, so I extracted it to avoid duplication and keep the logic in one place.

gravitee-apim-console-webui/src/management/api-products/apis/api-product-apis.component.ts
ankita-gupta21: done

gravitee-apim-console-webui/src/management/api-products/apis/add-api-dialog/api-product-add-api-dialog.component.ts
ankita-gupta21: done

jourdiw (approved):
Woohoo! 🎊

PR #15337, opened 2026-02-23
feat(portal): adjust the footer to the figma design

adam-rosowicz:
Link to jira ticket: https://gravitee.atlassian.net/browse/EXT-62
image
image


gravitee-apim-portal-webui-next/src/components/footer/footer.component.html
```diff
@@ -15,5 +15,4 @@
     limitations under the License.
 
 -->
-<app-company-title [title]="title" [logo]="logo()" />
-<p class="footer__copyright">© {{ currentYear }}. All rights reserved.</p>
+<p class="footer__copyright">© {{ currentYear }} Company. All rights reserved.</p>
```
jourdiw: With hard-coded text, please add a `i18n="@@...."` tag before so that it has a unique key identifier 👍

jourdiw (changes requested):
Since you are no longer using the `app-company-title` component, you can also remove it from the imports in `footer.component.ts` 👍

jourdiw (approved):

PR #15350, opened 2026-02-23
fix: map API_KEY security type for automation API

kamiiiel:
see https://gravitee.atlassian.net/browse/GKO-2408


gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/ApiResourceTest.java
```diff
@@ -130,7 +131,7 @@ void should_get_api_with_null_plan_type() {
                         .get()
                         .extracting(PlanV4::getSecurity)
                         .extracting(io.gravitee.apim.rest.api.automation.model.PlanSecurity::getType)
-                        .isNull();
+                        .isEqualTo(PlanSecurityType.API_KEY);
```
carlos-andres-osorio: Test name `should_get_api_with_null_plan_type` doesn't match the new validation.

carlos-andres-osorio: Not directly related to this change, but it would be good to have similar tests for other plan security types to prevent regressions.

carlos-andres-osorio (approved):

PR #15368, opened 2026-02-24
refactor: move certificate revocation scheduler from plugin to embedded bean

carlos-andres-osorio:
## Issue

[GKO-2413 — Refactor plugin to use the same scheduling pattern used by ScheduledCommandsRefresherServiceImpl](https://gravitee.atlassian.net/browse/GKO-2413)

## Description

Migrates `ScheduledCertificateRevocationService` from a standalone plugin to an embedded Spring `@Component`, following the same pattern as `ScheduledCommandsRefresherServiceImpl`.

- Deleted the `gravitee-apim-rest-api-services-certificate-revocation` plugin module (pom, assembly, plugin.properties)
- Moved the service implementation into `gravitee-apim-rest-api-service` under the standard `service/impl` package
- Introduced a `ScheduledCertificateRevocationService<T>` lifecycle marker interface, registered in `GraviteeApisNode`
- Added `CertificateRevocationConfiguration` Spring configuration with a dedicated `ThreadPoolTaskScheduler` (graceful shutdown, explicit error handler to prevent silent task death on uncaught `Error`)
- `doStop()` now cancels the `ScheduledFuture` to properly honour the Gravitee lifecycle contract
- Test class moved and expanded: idempotency, lifecycle stop, cron expression assertion, future cancellation

## Additional context

The scheduler uses a dedicated thread pool (separate from the application-wide scheduler) to isolate certificate revocation from other scheduled work. The pool is configured with `setWaitForTasksToCompleteOnShutdown(true)` and `setAwaitTerminationSeconds(30)` so an in-progress revocation run is never interrupted mid-execution on shutdown — this differs intentionally from `CommandRefresherConfiguration`.


benoitgravitee:
No need of the PR, first implementation was the right choice.

PR #15373, opened 2026-02-24
feat: store subscription form metadata

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-12636

## Description

This PR follows [PR 2 – render subscription form on checkout](https://github.com/gravitee-io/gravitee-api-management/pull/15329) and completes the subscription form flow by persisting subscription form data as metadata when creating or updating a subscription, with validation and sanitization to keep data safe and within limits.

### Included in this PR

- `SubscriptionMetadataSanitizer`: validation and sanitization of subscription metadata (e.g. invalid keys, excessively long values)
- Sanitization applied on subscription create and update (`SubscriptionResource`, `SubscriptionsResource`)
- Unit tests for sanitizer and for create/update with invalid or oversized metadata

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
@@ -155,8 +155,11 @@ public class SubscriptionResource extends AbstractResource {
 
         UpdateSubscriptionConfigurationEntity updateSubscriptionConfigurationEntity = new UpdateSubscriptionConfigurationEntity();
         updateSubscriptionConfigurationEntity.setSubscriptionId(subscriptionId);
-        updateSubscriptionConfigurationEntity.setMetadata(updateSubscriptionInput.getMetadata());
         SubscriptionConfigurationInput subscriptionConfigurationInput = updateSubscriptionInput.getConfiguration();
+
+        if (updateSubscriptionInput.getMetadata() != null) {
+            updateSubscriptionConfigurationEntity.setMetadata(updateSubscriptionInput.getMetadata());
+        }
         if (subscriptionConfigurationInput != null) {
             SubscriptionConfigurationEntity subscriptionConfigurationEntity = new SubscriptionConfigurationEntity();
             subscriptionConfigurationEntity.setChannel(subscriptionConfigurationInput.getChannel());
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
@@ -18,8 +18,6 @@ package io.gravitee.rest.api.portal.rest.resource;
 import static io.gravitee.rest.api.model.permissions.RolePermission.APPLICATION_SUBSCRIPTION;
 import static io.gravitee.rest.api.model.permissions.RolePermissionAction.UPDATE;
 
-import io.gravitee.apim.core.audit.model.AuditActor;
-import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.apim.core.subscription.use_case.CloseSubscriptionUseCase;
 import io.gravitee.common.http.MediaType;
 import io.gravitee.definition.jackson.datatype.GraviteeMapper;
@@ -115,24 +113,7 @@ public class SubscriptionResource extends AbstractResource {
         SubscriptionEntity subscriptionEntity = subscriptionService.findById(subscriptionId);
         final ExecutionContext executionContext = GraviteeContext.getExecutionContext();
         if (hasPermission(executionContext, APPLICATION_SUBSCRIPTION, subscriptionEntity.getApplication(), RolePermissionAction.DELETE)) {
-            final var user = getAuthenticatedUserDetails();
-
-            closeSubscriptionUsecase.execute(
-                new CloseSubscriptionUseCase.Input(
-                    subscriptionId,
-                    AuditInfo.builder()
-                        .organizationId(executionContext.getOrganizationId())
-                        .environmentId(executionContext.getEnvironmentId())
-                        .actor(
-                            AuditActor.builder()
-                                .userId(user.getUsername())
-                                .userSource(user.getSource())
-                                .userSourceId(user.getSourceId())
-                                .build()
-                        )
-                        .build()
-                )
-            );
+            closeSubscriptionUsecase.execute(new CloseSubscriptionUseCase.Input(subscriptionId, getAuditInfo()));
             return Response.noContent().build();
         }
         throw new ForbiddenAccessException();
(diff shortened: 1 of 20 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
```diff
@@ -155,8 +155,11 @@ public Response updateSubscription(
 
         UpdateSubscriptionConfigurationEntity updateSubscriptionConfigurationEntity = new UpdateSubscriptionConfigurationEntity();
         updateSubscriptionConfigurationEntity.setSubscriptionId(subscriptionId);
-        updateSubscriptionConfigurationEntity.setMetadata(updateSubscriptionInput.getMetadata());
         SubscriptionConfigurationInput subscriptionConfigurationInput = updateSubscriptionInput.getConfiguration();
+
```
dev-marek, line 140: Instead changing flow here, please use `UpdateSubscriptionUseCase`

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionsResource.java
```diff
@@ -116,7 +116,9 @@ public Response createSubscription(@Valid @NotNull(message = "Input must not be
                 );
                 newSubscriptionEntity.setGeneralConditionsContentRevision(generalConditionsContentRevision);
             }
-            newSubscriptionEntity.setMetadata(subscriptionInput.getMetadata());
+            if (subscriptionInput.getMetadata() != null) {
```
dev-marek: Instead changing flow here, use `CreateSubscriptionUseCase`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
```diff
@@ -912,7 +912,9 @@ public SubscriptionEntity update(
             );
             subscription.setFailureCause(null);
             setSubscriptionConfig(subscriptionConfigEntity.getConfiguration(), subscription);
-            subscription.setMetadata(subscriptionConfigEntity.getMetadata());
+            if (subscriptionConfigEntity.getMetadata() != null) {
```
dev-marek: It might cause unwanted side effect for MAPI

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
JedrzejJanasiak: **Why migrating the portal `PUT /subscriptions/{id}` to `UpdateSubscriptionUseCase` is non-trivial**

The portal's `SubscriptionResource#updateSubscription` and `UpdateSubscriptionUseCase` both ultimately go through `SubscriptionServiceImpl`, but they call **two different overloads** with meaningfully different business logic. A naive swap would introduce silent behavioral regressions.

---

### Current portal path — "config-only" update

```
SubscriptionResource.updateSubscription()
  └─► subscriptionService.update(ExecutionContext, UpdateSubscriptionConfigurationEntity)   ← config-only overload
```

### UseCase path — "full" update

```
UpdateSubscriptionUseCase.execute()
  └─► UpdateSubscriptionDomainServiceImpl.update()
        └─► subscriptionService.update(ExecutionContext, UpdateSubscriptionEntity)           ← full-update overload
              └─► subscriptionService.update(ExecutionContext, UpdateSubscriptionEntity, Consumer<Subscription>)
```

---

### Behavioral differences

#### 1. Allowed subscription statuses

| | Config-only (current portal) | Full update (UseCase) |
|---|---|---|
| `ACCEPTED` | ✅ | ✅ |
| `PENDING` | ✅ | ✅ |
| `PAUSED` | ✅ | ✅ |
| `REJECTED` | ✅ allowed | ❌ throws `SubscriptionNotUpdatableException` |
| `CLOSED` | ❌ throws | ❌ throws |

The full-update overload explicitly guards: `if (status == ACCEPTED || PENDING || PAUSED) { … } else throw SubscriptionNotUpdatableException`. The config-only overload only rejects `CLOSED`, so `REJECTED` subscriptions can currently be reconfigured via the portal — that would silently break.

#### 2. Auto-reset to `PENDING` for manually-validated plans

The config-only overload contains:

```java
Subscription.Status newSubscriptionStatus = planEntity.getPlanValidation() == MANUAL
    ? PENDING
    : subscription.getStatus();
```

When a subscriber updates configuration on a **MANUAL validation** plan, the subscription is automatically put back to `PENDING` so the API publisher re-reviews it before it becomes active. The full-update path has **no equivalent logic** — the status stays as-is after a config change, bypassing the re-approval gate.

#### 3. Consumer status recovery after `FAILURE`

The config-only overload explicitly handles the async consumer `FAILURE` state:

```java
subscription.setConsumerStatus(
    previousSubscription.getConsumerStatus() == FAILURE
        ? STARTED
        : previousSubscription.getConsumerStatus()
);
subscription.setFailureCause(null);
```

If a push-subscription's consumer is stuck in `FAILURE` (e.g. unreachable callback URL), a subscriber updating their configuration today gets a free recovery — the consumer is restarted and the failure cause is cleared. The full-update path does **nothing** with `consumerStatus` or `failureCause`, so those subscriptions would stay broken after the same update.

#### 4. Fields in scope

| Field | Config-only | Full update (UseCase) |
|---|---|---|
| `configuration` (entrypoint) | ✅ | ✅ |
| `metadata` | ✅ | ✅ |
| `startingAt` / `endingAt` | ❌ not touched | ✅ set + propagated to API keys |
| `daysToExpirationOnLastNotification` | ❌ not reset | ✅ reset to `null` |

The full-update path also cascades a new `endingAt` to all non-shared active API keys for API_KEY plans — the config-only path never does this, because date management is out of scope for a portal user.

---

### Conclusion

These aren't minor edge cases. Points **2** and **3** represent intentional portal-specific safety mechanisms (re-approval gate for MANUAL plans; consumer recovery) that exist in the config-only overload precisely because the portal is subscriber-facing. The UseCase was designed for the management plane (Console) where publishers perform full subscription lifecycle management, not for subscriber self-service configuration updates.

Migrating the portal UPDATE to `UpdateSubscriptionUseCase` would require either:
- back-porting the MANUAL-plan reset and consumer-recovery logic into the UseCase/domain service (expanding its scope), or
- accepting a feature regression for portal subscribers.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
```diff
@@ -914,8 +914,6 @@ public SubscriptionEntity update(
             setSubscriptionConfig(subscriptionConfigEntity.getConfiguration(), subscription);
             subscription.setMetadata(subscriptionConfigEntity.getMetadata());
 
-            subscription = subscriptionRepository.update(subscription);
```
dev-marek, line 917: are you 100% sure?

JedrzejJanasiak: 😶 brought it back

dev-marek (approved):

PR #15394, opened 2026-02-25
display metadata in console

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-12159

## Description
This PR closes the subscription flow by letting API publishers and portal admins view subscription metadata in the Subscription Details screen. They can see what consumers submitted during subscription (e.g. via subscription form) and use it when approving or reviewing subscriptions.
Included in this PR

### Subscription Details – metadata display: 
Subscription metadata is shown in the API subscription edit screen and in the application subscription screen. When metadata exists, it is displayed in a read-only Monaco editor with JSON formatting (key-value structure). When there is no metadata, or it is empty, a dash (-) is shown instead.
### Where it appears: 
Metadata is visible in the subscription details when reviewing a pending subscription (before approve/reject), and  after approval for later reference.
### Security: 
Metadata is rendered via Angular property binding and the read-only Monaco editor only; no innerHTML or raw HTML injection of user-provided values.

## Additional context

image

### 🎬 : 

https://github.com/user-attachments/assets/5ba3c5e2-cb9a-4cc9-b62c-3be196de23ec

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/subscriptions/edit/api-subscription-edit.component.html
+++ b/gravitee-apim-console-webui/src/management/api/subscriptions/edit/api-subscription-edit.component.html
@@ -124,6 +124,11 @@
           <dd data-testId="subscription-domain">
             {{ subscription.domain || '-' }}
           </dd>
+
+          <dt>Metadata</dt>
+          <dd class="subscription__metadata-cell" data-testId="subscription-metadata">
+            <subscription-metadata-viewer [metadata]="subscription.metadata" />
+          </dd>
         </dl>
       </mat-card-content>
       <mat-card-actions>
--- a/gravitee-apim-console-webui/src/management/application/details/subscriptions/subscription/application-subscription.component.html
+++ b/gravitee-apim-console-webui/src/management/application/details/subscriptions/subscription/application-subscription.component.html
@@ -101,6 +101,11 @@
         }
         <dt>Closed at</dt>
         <dd>{{ (pageVM.subscription.closed_at | date: 'medium') || '-' }}</dd>
+
+        <dt>Metadata</dt>
+        <dd class="subscription__metadata-cell" data-testid="subscription-metadata">
+          <subscription-metadata-viewer [metadata]="pageVM.subscription.metadata" />
+        </dd>
       </dl>
     </mat-card-content>
 
--- a/gravitee-apim-console-webui/src/management/application/details/subscriptions/subscription/application-subscription.component.ts
+++ b/gravitee-apim-console-webui/src/management/application/details/subscriptions/subscription/application-subscription.component.ts
@@ -42,6 +42,7 @@ import { SnackBarService } from '../../../../../services-ngx/snack-bar.service';
 import { ApplicationSubscriptionService } from '../../../../../services-ngx/application-subscription.service';
 import { SubscriptionApiKeysComponent } from '../components/subscription-api-keys/subscription-api-keys.component';
 import { SubscriptionEditPushConfigComponent } from '../../../../../components/subscription-edit-push-config/subscription-edit-push-config.component';
+import { SubscriptionMetadataViewerComponent } from '../../../../../shared/components/subscription-metadata-viewer/subscription-metadata-viewer.component';
 
 type PageVM = {
   application: Application;
@@ -65,6 +66,7 @@ type PageVM = {
     SubscriptionApiKeysComponent,
     MatTooltip,
     SubscriptionEditPushConfigComponent,
+    SubscriptionMetadataViewerComponent,
   ],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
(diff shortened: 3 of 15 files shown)


gravitee-apim-console-webui/src/management/api/subscriptions/edit/api-subscription-edit.component.ts
```diff
@@ -189,6 +213,11 @@ export class ApiSubscriptionEditComponent implements OnInit {
               metadata: subscription.metadata,
             };
 
+            this.metadataJson =
+              !this.subscription.metadata || Object.keys(this.subscription.metadata).length === 0
```
aqibmohammadkhan: are we sure `metadataJson` would be reassigned on subscription state change like approve/reject/transfer?
since this is only computed once in ngOnInit.

i would suggest either derive it as a `get` accessor from `this.subscription` or ensure the refresh path reassigns it explicitly.

gravitee-apim-console-webui/src/management/api/subscriptions/edit/api-subscription-edit.component.html
```diff
@@ -124,6 +124,23 @@
           <dd data-testId="subscription-domain">
             {{ subscription.domain || '-' }}
           </dd>
+
+          <dt>Metadata</dt>
+          <dd class="subscription__metadata-cell" data-testId="subscription-metadata">
+            @if (metadataJson) {
+              <div class="subscription__metadata-editor" gioMonacoClipboardCopy>
+                <gio-monaco-editor
+                  disabled
```
aqibmohammadkhan: I would recommend removing the disabled attribute on <gio-monaco-editor> since readOnly: true in options already handles it.

gravitee-apim-console-webui/src/management/application/details/subscriptions/subscription/application-subscription.component.html
```diff
@@ -101,6 +101,24 @@
...
         <dt>Closed at</dt>
         <dd>{{ (pageVM.subscription.closed_at | date: 'medium') || '-' }}</dd>
+
+        <dt>Metadata</dt>
+        <dd class="subscription__metadata-cell" data-testid="subscription-metadata">
+          @if (!pageVM.subscription.metadata || (pageVM.subscription.metadata | keyvalue).length === 0) {
+            -
+          } @else {
+            <div class="subscription__metadata-editor" gioMonacoClipboardCopy>
+              <gio-monaco-editor
```
aqibmohammadkhan: This null/empty check is done in the template using the keyvalue pipe, whereas in `api-subscription-edit `the equivalent check is done in the component ts and the template simply checks `@if (metadataJson)`.

The API subscription view is cleaner — the template just checks the precomputed string. The application subscription view already computes `metadataJson` in `pageVM$,` but then re-checks metadata again in the template instead of just using `@if (pageVM.metadataJson)`.

i would suggest simplifying the template condition to:

`
```
@if (pageVM.metadataJson) {
  <div ...>...</div>
} @else {
  -
} 
```
`

aqibmohammadkhan: This is consistent with the API subscription view and avoids a redundant keyvalue pipe execution on every change detection cycle.

gravitee-apim-console-webui/src/management/application/details/subscriptions/subscription/application-subscription.component.ts
```diff
@@ -89,12 +96,30 @@ export class ApplicationSubscriptionComponent {
...
+    overviewRulerBorder: false,
+    readOnly: true,
+    wordWrap: 'on',
+    scrollBeyondLastLine: false,
+    padding: { top: 0, bottom: 0 },
+    scrollbar: {
+      vertical: 'auto',
+      horizontal: 'hidden',
+      useShadows: false,
+    },
```
aqibmohammadkhan: this duplicates [this](https://github.com/gravitee-io/gravitee-api-management/pull/15394/changes#diff-ad4d7df1f51547e816af243d7e365fa6ba11cd7adfd06dd0b25e3d63c20eb791R131)

wdyt about extracting this to a shared constant?
if too much effort, ignore this?

gravitee-apim-console-webui/src/management/application/details/subscriptions/subscription/application-subscription.component.scss
```diff
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
 
+@use 'sass:map';
```
aqibmohammadkhan: is this used here? If not please remove.

aqibmohammadkhan (changes requested):

gravitee-apim-console-webui/src/management/application/details/subscriptions/subscription/application-subscription.component.ts
JedrzejJanasiak: Instead of just pulling out a constant, I went with a shared SubscriptionMetadataViewerComponent (shared/components/subscription-metadata-viewer/), which felt cleaner since the duplication wasn't just the scrollbar options but also the template logic and the border styles. Both parent components now just do <subscription-metadata-viewer [metadata]="..." /> and that's it. lmk if that works for you!

gravitee-apim-console-webui/src/management/api/subscriptions/edit/api-subscription-edit.component.ts
JedrzejJanasiak: Since I extracted the metadata display to a separate component, this whole logic is now handled by a computed() signal inside it :)

aqibmohammadkhan (approved):
LGTM

PR #15415, opened 2026-02-25
fix: handle null and empty dashboard name

kamiiiel:
see https://gravitee.atlassian.net/browse/GKO-2414

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/dashboard/domain_service/DashboardDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/dashboard/domain_service/DashboardDomainService.java
@@ -78,6 +78,7 @@ public class DashboardDomainService {
         if (dashboard.getName().isBlank()) {
             throw new ValidationDomainException("Dashboard name cannot be blank");
         }
+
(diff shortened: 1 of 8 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/dashboard/domain_service/DashboardDomainService.java
```diff
@@ -75,9 +75,12 @@ public void delete(String dashboardId) {
...
 
     private void validate(Dashboard dashboard) {
-        if (dashboard.getName().isBlank()) {
-            throw new ValidationDomainException("Dashboard name cannot be blank");
-        }
+        ofNullable(dashboard.getName()).ifPresent(name -> {
+            if (name.isBlank()) {
+                throw new ValidationDomainException("Dashboard name cannot be blank");
+            }
+        });
```
a-cordier: But can't you set a null name on create if you do this ?

kamiiiel: The code will never arrives there if you set name to null when creating a dashboard

a-cordier (approved):

PR #15423, opened 2026-02-26
fix(gateway): improve message in response for authentication

wbabyte:
## Issue

https://gravitee.atlassian.net/browse/APIM-12974

## Description


phiz71 (approved):

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -849,6 +849,8 @@ api:
   # change), this timeout will be the maximum time (in milliseconds) to wait for all pending requests to terminate
 #  pending_requests_timeout: 10000
   validateSubscription: true  # set to false if you want to skip validating the subscription, default value is true
+  security:
+    verbose401: true  # set to true to return detailed error messages in 401 responses
```
phiz71: shouldn't we set it to false by default ?

phiz71 (commented):
should we add it also into the values.yml ? (false by default)

ytvnr (approved):

PR #15433, opened 2026-02-26
fix(promotion): ignored malformed promotion

phiz71:
## Issue

https://gravitee.atlassian.net/browse/APIM-12964

## Description

when a promotion cannot be parsed, we now ignore it instead of throwing an error. To avoid breaking the /tasks resource

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/promotion/PromotionTasksServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/promotion/PromotionTasksServiceImpl.java
@@ -35,12 +35,10 @@ import io.gravitee.rest.api.model.promotion.PromotionQuery;
 import io.gravitee.rest.api.service.EnvironmentService;
 import io.gravitee.rest.api.service.PermissionService;
 import io.gravitee.rest.api.service.common.ExecutionContext;
-import io.gravitee.rest.api.service.exceptions.TechnicalManagementException;
 import io.gravitee.rest.api.service.impl.AbstractService;
 import io.gravitee.rest.api.service.promotion.PromotionService;
 import io.gravitee.rest.api.service.promotion.PromotionTasksService;
 import io.gravitee.rest.api.service.v4.ApiSearchService;
-import io.gravitee.rest.api.service.v4.ApiService;
 import java.util.*;
 import java.util.stream.Collectors;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/promotion/PromotionTasksServiceImpl.java
```diff
@@ -128,6 +127,17 @@ private List<TaskEntity> getPromotionTasksForEnvironments(List<EnvironmentEntity
...
+                JsonNode jsonNode;
+                try {
+                    jsonNode = objectMapper.readTree(promotionEntity.getApiDefinition());
+                } catch (JsonProcessingException e) {
+                    logger.warn("Problem while deserializing api definition for promotion {}", promotionEntity.getId());
+                    return false;
+                }
+                JsonNode export = jsonNode.get("export");
+                return export == null;
+            })
```
phiz71: Actually, this is not really mandatory as an API V4 promotion will fail during the convert and the result will be the same. We ignore them.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/promotion/PromotionTasksServiceImpl.java
```diff
@@ -128,6 +127,17 @@ private List<TaskEntity> getPromotionTasksForEnvironments(List<EnvironmentEntity
         return promotionsPage
             .getContent()
             .stream()
+            .filter(promotionEntity -> {
+                JsonNode jsonNode;
+                try {
+                    jsonNode = objectMapper.readTree(promotionEntity.getApiDefinition());
+                } catch (JsonProcessingException e) {
+                    logger.warn("Problem while deserializing api definition for promotion {}", promotionEntity.getId());
```
ytvnr: should it be interesting to have the error message in this warn ? at least to have a clue ?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/promotion/PromotionTasksServiceImpl.java
```diff
@@ -154,7 +166,7 @@ private TaskEntity convert(PromotionEntity promotionEntity, boolean isUpdate, Op
             apiEntity = objectMapper.readValue(promotionEntity.getApiDefinition(), ApiEntity.class);
         } catch (JsonProcessingException e) {
             logger.warn("Problem while deserializing api definition for promotion {}", promotionEntity.getId());
```
ytvnr: same question about error message ?

ytvnr (approved):

Okhelifi (approved):

PR #15455, opened 2026-02-27
feat: add confirmation dialog for updating scheduled alerts

Okhelifi:
## Issue

## Description

- Display a confirmation dialog when updating an alert that has a scheduled duration (duration + timeUnit condition), warning the user that saving will reset the evaluation schedule
- Show the "Last updated at" timestamp on the alert edit page in both AngularJS and Angular views

Demo:

https://github.com/user-attachments/assets/e7545049-91f4-4959-a062-815b307e5c38

### DIFF
--- a/gravitee-apim-console-webui/src/components/alerts/alert/alert.component.ajs.ts
+++ b/gravitee-apim-console-webui/src/components/alerts/alert/alert.component.ajs.ts
@@ -54,6 +54,7 @@ const AlertComponentAjs: ng.IComponentOptions = {
       this.tabs = ['_', 'general', 'notifications', 'history'];
       this.selectedTab = 0;
       this.currentTab = this.tabs[this.selectedTab];
+      this.isScheduledAlert = false;
 
       this.$onInit = () => {
         this.severities = ['INFO', 'WARNING', 'CRITICAL'];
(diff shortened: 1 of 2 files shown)


gravitee-apim-console-webui/src/components/alerts/alert/alert.component.ajs.ts
```diff
@@ -80,6 +81,7 @@ const AlertComponentAjs: ng.IComponentOptions = {
           this.alerts.push(this.alert);
         } else {
           this.alert = getAlertWithoutMutation(this.alerts, this.activatedRoute.snapshot.params.alertId, this.referenceType);
+          this.isScheduledAlert = this.alert.conditions?.some(c => 'duration' in c && 'timeUnit' in c) ?? false;
```
ytvnr, line 84: I wonder if a || would be enough ?

ytvnr (approved):

gravitee-apim-console-webui/src/components/alerts/alert/alert.component.ajs.ts
Okhelifi: It should not be possible technically to have one or the other you should have both or nothing I think

PR #15482, opened 2026-03-02
fix: improve query parameter handling in router navigation

prateektiwari-gravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-12956

### Root cause
When switching tabs (e.g. General → Documentation), onNavChange in AppComponent was called with a path like catalog/api/123/doc that has no query params

### Description

In app.component.ts, navigation now preserves query params when the target path has none.

### Additional context

Pre Fix behaviour: 

https://github.com/user-attachments/assets/b2891413-32f8-4b10-aab3-56f40aa17cb8

Post Fix Behaviour: 

https://github.com/user-attachments/assets/83dfe27d-3a4a-4426-8a5b-ba9c0d99d1e8

### DIFF
--- a/gravitee-apim-portal-webui/src/app/app.component.ts
+++ b/gravitee-apim-portal-webui/src/app/app.component.ts
@@ -331,7 +331,7 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
     } else {
       const urlTree = this.router.parseUrl(route.path);
       const path = urlTree.root.children[PRIMARY_OUTLET].segments.join('/');
-      this.router.navigate([path], { queryParams: urlTree.queryParams });
+      this.router.navigate([path], { queryParams: urlTree.queryParams, queryParamsHandling: 'merge' });
     }
   }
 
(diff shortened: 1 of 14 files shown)


gravitee-apim-portal-webui/src/app/app.component.ts
```diff
@@ -331,7 +331,9 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
     } else {
       const urlTree = this.router.parseUrl(route.path);
       const path = urlTree.root.children[PRIMARY_OUTLET].segments.join('/');
-      this.router.navigate([path], { queryParams: urlTree.queryParams });
+      const hasQueryParams = urlTree.queryParams && Object.keys(urlTree.queryParams).length > 0;
```
mukul-tyagi08: The moment a user clicks any nav link that happens to carry its own query param, same issue may happen again.
I think just adding queryParamsHandling: 'merge' will be better and cleaner approach.

prateektiwari-gravitee: Yes, I agree.

mukul-tyagi08 (commented):
Your commit message is incorrect.

prateektiwari-gravitee:
This PR had issues with diverged git histories. Closing this one to ensure clean history and correct commit messages.

PR #15497, opened 2026-03-04
fix: allow user attributes in mail templates

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-12958

## Description

User attributes such as ${user.displayName} were not resolved in email templates during subscription notifications. Owner attributes were working correctly.

This change ensures user attributes are properly evaluated in both placeholders and conditional expressions within templates.

## Additional context
When a user creates a subscription, notification emails are sent to the API owner and the application owner. These email templates support variable interpolation (e.g. `${owner.displayName}`, `${api.name}`) via Gravitee's `NotificationParamsBuilder`.

However, **user-specific attributes** such as `${user.displayName}` were **not being resolved** in these templates. The `user` object was never added to the notification parameters during subscription creation, causing the placeholders to silently render as empty strings or fail conditional expressions in templates.

API owner attributes (`${owner.*}`) worked correctly because they were already mapped separately. The missing piece was the subscribing user's `UserEntity`.

Documentation : https://documentation.gravitee.io/apim/configure-and-manage-the-platform/gravitee-gateway/notifications

Approach:

1. Retrieve the current `UserDetails` from the `SecurityContext`.
2. Prefer `UserDetails#getId()` as the lookup key; fall back to `UserDetails#getUsername()` for token/JWT-based auth flows where `id` may be null.
3. Call `userService.findById(executionContext, identifier)` to load the full `UserEntity`.
4. If resolution succeeds, add the user to the params builder via `paramsBuilder.user(subscribedByUser)`.
5. If the authenticated user is `null`, blank, or the lookup throws (e.g. user not found), the notification is still triggered — the `user` param is simply absent. No exception is propagated.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
@@ -870,6 +870,17 @@ public class SubscriptionServiceImpl extends AbstractService implements Subscrip
             .application(applicationEntity);
         if (subscriptionEntity != null) {
             paramsBuilder.subscription(subscriptionEntity);
+            String subscribedBy = subscriptionEntity.getSubscribedBy();
+            if (subscribedBy != null) {
+                try {
+                    UserEntity subscribedByUser = userService.findById(executionContext, subscribedBy);
+                    if (subscribedByUser != null) {
+                        paramsBuilder.user(subscribedByUser);
+                    }
+                } catch (Exception e) {
+                    log.debug("Could not resolve subscribed-by user {} for API subscription notification params", subscribedBy, e);
+                }
+            }
         }
         if (subscriptionsUrl != null) {
             paramsBuilder.subscriptionsUrl(subscriptionsUrl);
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/SubscriptionServiceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/SubscriptionServiceTest.java
@@ -93,6 +93,7 @@ import io.gravitee.rest.api.model.SubscriptionStatus;
 import io.gravitee.rest.api.model.TransferSubscriptionEntity;
 import io.gravitee.rest.api.model.UpdateSubscriptionConfigurationEntity;
 import io.gravitee.rest.api.model.UpdateSubscriptionEntity;
+import io.gravitee.rest.api.model.UserEntity;
 import io.gravitee.rest.api.model.api.ApiEntity;
 import io.gravitee.rest.api.model.application.ApplicationSettings;
 import io.gravitee.rest.api.model.application.SimpleApplicationSettings;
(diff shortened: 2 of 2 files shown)


ThibaudAV (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
```diff
@@ -824,6 +824,22 @@ public SubscriptionEntity create(
         }
     }
 
+    /**
+     * Resolves the authenticated user for notification params (e.g. email templates).
+     * Used for SUBSCRIPTION_NEW, SUBSCRIPTION_FAILED, SUBSCRIPTION_PAUSED, SUBSCRIPTION_RESUMED, SUBSCRIPTION_TRANSFERRED.
+     */
+    private Optional<UserEntity> resolveSubscribedByUser(ExecutionContext executionContext, UserDetails userDetails) {
```
ThibaudAV: Where is it used?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
```diff
@@ -2087,20 +2102,6 @@ public SubscriptionEntity transfer(
...
-            throw new TechnicalManagementException(
-                String.format("An error occurs while trying to transfer subscription %s by %s", transferSubscription.getId(), userId),
-                ex
-            );
-        }
-    }
-
-    @Override
-    public String exportAsCsv(Collection<SubscriptionEntity> subscriptions, Map<String, Map<String, Object>> metadata) {
-        final StringBuilder sb = new StringBuilder();
```
ThibaudAV, line 2103: The removal of this rating seems unintentional, doesn't it?

ThibaudAV (commented):
Please check again, it seems that part of the code is missing.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
vikrantgravitee: It got removed while resolving conflict. I will raise clean PR

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
vikrantgravitee: I will fix it in clean PR.

vikrantgravitee: It is fixed and removed.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
vikrantgravitee: It is fixed and it was revoked (deleted due to conflict file mismatch)

PR #15508, opened 2026-03-04
Improve CI test execution time for REST API modules

thomasballivet:


gravitee-apim-rest-api/gravitee-apim-rest-api-integration-controller/src/test/java/io/gravitee/integration/controller/command/hello/HelloCommandHandlerTest.java
```diff
@@ -131,7 +131,7 @@ void should_reply_succeeded_when_integration_exists() {
         commandHandler
             .handle(COMMAND)
             .test()
-            .awaitDone(10, TimeUnit.SECONDS)
+            .awaitDone(2, TimeUnit.SECONDS)
```
michel-barret: Shouldn’t change test duration, but increase the instability. This duration is like a timeout not a sleep.

michel-barret (approved):

PR #15524, opened 2026-03-05
fix(elasticsearch): escape special characters in Lucene query_string

araffass:
## Issue

https://gravitee.atlassian.net/browse/APIM-12955

## Description

Fixed the 500 error occurring when log filters contain Lucene special characters (e.g., /, (, )), found in custom metric based on the User-Agent header

## Additional context

Why not used QueryParser (.escape()) or any other alternative? 
It is too aggressive; it escapes structural characters like : and *, which breaks standard field filters (e.g., status:200) and wildcard searches.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/log/ElasticLogRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/log/ElasticLogRepository.java
@@ -81,10 +81,11 @@ public class ElasticLogRepository extends AbstractElasticsearchRepository implem
         try {
             final String logQueryString = getQuery(query.query(), true);
             if (isEmpty(logQueryString)) {
+                String safeQuery = this.createSafeElasticsearchJsonQuery(query);
                 final Single<SearchResponse> result = this.client.search(
                     this.indexNameGenerator.getIndexName(queryContext.placeholder(), Type.REQUEST, from, to, clusters),
                     !info.getVersion().canUseTypeRequests() ? null : Type.REQUEST.getType(),
-                    this.createElasticsearchJsonQuery(query)
+                    safeQuery
                 );
 
                 return this.toTabularResponse(result.blockingGet());
@@ -95,7 +96,7 @@ public class ElasticLogRepository extends AbstractElasticsearchRepository implem
                     .size(MAX_RESULT_WINDOW)
                     .query(logQueryString)
                     .build();
-                final String sQuery = this.createElasticsearchJsonQuery(logQuery);
+                final String sQuery = this.createSafeElasticsearchJsonQuery(logQuery);
 
                 Single<SearchResponse> result = this.client.search(
                     this.indexNameGenerator.getIndexName(queryContext.placeholder(), Type.LOG, from, to, clusters),
@@ -126,7 +127,7 @@ public class ElasticLogRepository extends AbstractElasticsearchRepository implem
                     result = this.client.search(
                         this.indexNameGenerator.getIndexName(queryContext.placeholder(), Type.REQUEST, from, to, clusters),
                         !info.getVersion().canUseTypeRequests() ? null : Type.REQUEST.getType(),
-                        this.createElasticsearchJsonQuery(requestQueryBuilder.build())
+                        this.createSafeElasticsearchJsonQuery(requestQueryBuilder.build())
                     );
                 }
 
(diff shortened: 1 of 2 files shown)


gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/log/ElasticLogRepository.java
```diff
@@ -81,10 +81,11 @@ public TabularResponse query(final QueryContext queryContext, final TabularQuery
         try {
             final String logQueryString = getQuery(query.query(), true);
             if (isEmpty(logQueryString)) {
+                String patchedJson = this.createSafeElasticsearchJsonQuery(query);
```
ytvnr:
```suggestion
                String safeQuery = this.createSafeElasticsearchJsonQuery(query);
```

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/log/LuceneEscapeTest.java
```diff
@@ -0,0 +1,36 @@
...
+    @Test
+    public void shouldEscapeToxicUserAgent() {
+        String toxicInput = "custom.userAgentMelodi:RMel/1.0.0 (Ubuntu)";
+
+        String escaped = QueryParser.escape(toxicInput);
+        String result = escaped.replace("\\", "\\\\");
+
+        assertThat(result).contains("RMel\\\\/1.0.0");
+        assertThat(result).contains("\\\\(Ubuntu\\\\)");
+        System.out.println("JSON String: " + result);
```
ytvnr: to remove ?

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/log/LuceneEscapeTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.apache.lucene.queryparser.classic.QueryParser;
+import org.junit.jupiter.api.Test;
+
+public class LuceneEscapeTest {
+
+    @Test
+    public void shouldEscapeToxicUserAgent() {
```
ytvnr: I am not sure this test is really testing your implementation.
Was it to validate the behavior of QeuryParser before applying your fix ?

ytvnr (approved):

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/log/ElasticLogRepository.java
```diff
@@ -249,4 +250,25 @@ private TabularResponse toTabularResponse(final SearchResponse response, final l
 
         return tabularResponse;
     }
+
+    /**
+     * Fix APIM-12955: Escapes Lucene special characters in the query string.
+     * Quadruple backslashes are required to survive both Java and JSON parsing levels
+     * so that Lucene finally receives the mandatory single backslash escape (e.g., \( ).
+     */
+    private String createSafeElasticsearchJsonQuery(final TabularQuery query) {
```
michel-barret, line 259: Do you have considere use of https://lucene.apache.org/core/4_10_2/queryparser/org/apache/lucene/queryparser/flexible/standard/parser/EscapeQuerySyntaxImpl.html ?

araffass: YEs i indeed  considered using EscapeQuerySyntaxImpl or QueryParser.escape(), but both are unfortunately too aggressive for this use case. They escape structural characters like colons (:) and wildcards (*). Since our filter string often contains field-based queries (e.g., status:200 or api:uuid), using the standard Lucene utility breaks these searches, resulting in zero hits and failing tests.

This was the only approach that specifically targets characters that cause the Lexer to crash (like slashes and parentheses in User-Agents) while preserving the functional query syntax required by our UI."

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/log/LuceneEscapeTest.java
araffass: You are correct. The initial test was a baseline to confirm that QueryParser escapes everything, including structural characters.

I have updated the test class to explicitly validate my implementation. The goal is to prove that while QueryParser.escape() fixes the crash, it introduces a functional regression by escaping the colons (:) and wildcards (*)

PR #15539, opened 2026-03-07
fix: persist widgets metric level filters

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-2444

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/CustomDashboardWidget.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/CustomDashboardWidget.java
@@ -68,10 +68,22 @@ public class CustomDashboardWidget {
     @NoArgsConstructor
     @AllArgsConstructor
     @Data
+    public static class Filter {
+
+        private String name;
+        private String operator;
+        private Object value;
+    }
+
+    @Builder
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @Data
     public static class MetricRequest {
 
         private String name;
         private List<String> measures;
+        private List<Filter> filters;
     }
 
     @Builder
@@ -86,5 +98,6 @@ public class CustomDashboardWidget {
         private Long interval;
         private List<String> by;
         private Integer limit;
+        private List<Filter> filters;
     }
 }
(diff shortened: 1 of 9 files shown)


gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/CustomDashboardWidget.java
```diff
@@ -63,6 +63,17 @@ public static class TimeRange {
...
     }
 
+    @Builder
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @Data
+    public static class Filter {
+
+        private String name;
+        private String operator;
```
SamHardi, line 74: Same as the previous PR, don't you want to use an Enum?

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/DashboardMapper.java
```diff
@@ -126,9 +137,59 @@ default MetricRequest mapToRestMetricRequest(DashboardWidget.MetricRequest metri
...
+        switch (operator) {
+            case EQ -> restFilter.setActualInstance(
+                new StringFilter().name(filterName).operator(operator).value((String) filter.getValue())
+            );
+            case LTE, GTE -> restFilter.setActualInstance(
+                new NumberFilter().name(filterName).operator(operator).value(((Number) filter.getValue()).intValue())
+            );
+            case IN -> restFilter.setActualInstance(
+                new ArrayFilter().name(filterName).operator(operator).value((List<String>) filter.getValue())
+            );
```
SamHardi: "Add a default case to this switch." ⤴

SamHardi (approved):

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/CustomDashboardWidget.java
a-cordier: Same comment :)

PR #15556, opened 2026-03-09
fix: don’t fail all token check in case of the match fail

michel-barret:
## Description

We meet an issue, when one token fail. We don’t have enought log and all tokens usages are broken.


gravitee-apim-rest-api/gravitee-apim-rest-api-integration-controller/src/main/java/io/gravitee/integration/controller/websocket/auth/IntegrationWebsocketControllerAuthentication.java
```diff
@@ -61,7 +61,7 @@ public IntegrationCommandContext authenticate(final HttpServerRequest httpServer
                     .map(entry -> new IntegrationCommandContext(true, entry.getKey(), entry.getValue()))
                     .orElse(new IntegrationCommandContext(false));
             } catch (Exception e) {
-                log.warn("Unable to authenticate incoming websocket controller request");
+                log.warn("Unable to authenticate incoming websocket controller request", e);
```
jgiovaresco: I'm wondering if we should keep the warn as is and add a debug log to have more info. I'm afraid we would have a lot of log when someone try to login without token (or with incorrect token 🤔)

michel-barret: I hope it’s better now

jgiovaresco (approved):

PR #15557, opened 2026-03-09
feat(upgrader): add PortalNavigationItemRootIdUpgrader

aqibmohammadkhan:
Introduces `PortalNavigationItemRootIdUpgrader` to populate `rootId` for portal navigation items. Includes functionality for resolving root IDs, handling nested hierarchies, and extensive test coverage.

## Issue

https://gravitee.atlassian.net/browse/APIM-12818

## Description

Implements the service-level upgrader for portal navigation item rootId data migration

Context: All portal navigation items must have a rootId so the hierarchy can be resolved easily (e.g. redirect from Catalog to a specific API doc page). Existing items created before rootId was added have rootId = null and need a one-time backfill.

Changes:

 - New service-level upgrader (PortalNavigationItemRootIdUpgrader): Runs once at startup, iterates all environments, loads all portal navigation items per environment, and backfills rootId:
      - Root items (parentId == null): set rootId = id
      - Nested items: walk up the parentId chain to the top-level ancestor and set rootId to that ancestor’s id
      - Items that already have a valid rootId are skipped (idempotent)
      - Uses repository-level read/update only (no domain side effects like reordering)

Execution order: Registered in UpgraderOrder with value 715, after default portal navigation items (712) and subscription form (714).

Tests: 
   - Unit tests for no environments, no nav items, repository errors, root vs nested items, deep hierarchies, already-migrated items, multiple environments, and orphaned children.

Note: Schema (Liquibase for JDBC, MongoDB model), domain/repository models, creation logic (markAsRoot() / updateParent()), and update/parent-change logic (including propagateRootIdToDescendants()) were already in place; this PR adds only the one-time migration upgrader and its tests.

**Before**

https://github.com/user-attachments/assets/1a0e03c4-7be5-4a59-ad3b-4f266aa47c00

**After Upgrader Run**

 - DB
 

https://github.com/user-attachments/assets/220079fa-ac29-4493-962f-4990d1fb2cb6

 - API calls

https://github.com/user-attachments/assets/991bd86f-ffb4-4e72-924f-c2200668a0f4

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.service.impl.upgrade.upgrader;
+
+import static io.gravitee.rest.api.service.impl.upgrade.upgrader.UpgraderOrder.PORTAL_NAVIGATION_ITEM_ROOT_ID_UPGRADER;
+
+import io.gravitee.node.api.upgrader.Upgrader;
+import io.gravitee.node.api.upgrader.UpgraderException;
+import io.gravitee.repository.exceptions.TechnicalException;
+import io.gravitee.repository.management.api.EnvironmentRepository;
+import io.gravitee.repository.management.api.PortalNavigationItemRepository;
+import io.gravitee.repository.management.model.PortalNavigationItem;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import lombok.CustomLog;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Component;
+
+/**
+ * One-time backfill of {@code rootId} on existing portal navigation items so the front end can
+ * resolve the root of a navigation hierarchy without traversing parent chains at query time (e.g.
+ * redirect from Catalog to a specific API doc page using only the API nav item id). Root-level
+ * items get {@code rootId = id}; nested items get the top-level ancestor's id.
+ */
+@Component
+@CustomLog
+public class PortalNavigationItemRootIdUpgrader implements Upgrader {
+
+    private static final int MAX_PARENT_CHAIN_DEPTH = 50;
+
+    /**
+     * Sentinel value for "no root" in the domain; repository may store null/empty or this UUID string.
+     */
+    private static final String ZERO_ROOT_ID = "00000000-0000-0000-0000-000000000000";
+
+    private final EnvironmentRepository environmentRepository;
+    private final PortalNavigationItemRepository portalNavigationItemRepository;
+
+    public PortalNavigationItemRootIdUpgrader(
+        @Lazy EnvironmentRepository environmentRepository,
+        @Lazy PortalNavigationItemRepository portalNavigationItemRepository
+    ) {
+        this.environmentRepository = environmentRepository;
+        this.portalNavigationItemRepository = portalNavigationItemRepository;
+    }
+
+    @Override
+    public int getOrder() {
+        return PORTAL_NAVIGATION_ITEM_ROOT_ID_UPGRADER;
+    }
+
+    @Override
+    public boolean upgrade() throws UpgraderException {
+        return this.wrapException(this::applyUpgrade);
+    }
+
+    private boolean applyUpgrade() throws TechnicalException {
+        for (final var environment : environmentRepository.findAll()) {
+            migrateEnvironment(environment.getOrganizationId(), environment.getId());
+        }
+        return true;
+    }
+
+    private void migrateEnvironment(String organizationId, String environmentId) throws TechnicalException {
+        List<PortalNavigationItem> items = portalNavigationItemRepository.findAllByOrganizationIdAndEnvironmentId(
+            organizationId,
+            environmentId
+        );
+
+        Map<String, PortalNavigationItem> itemsById = items
+            .stream()
+            .collect(Collectors.toMap(PortalNavigationItem::getId, Function.identity()));
+        Map<String, String> rootIdsByItemId = new HashMap<>();
+        int updatedCount = 0;
+
+        for (PortalNavigationItem item : items) {
+            if (!needsRootIdMigration(item)) {
+                rootIdsByItemId.put(item.getId(), item.getRootId());
+                continue;
+            }
+
+            item.setRootId(findRootId(item, itemsById, rootIdsByItemId));
+            portalNavigationItemRepository.update(item);
+            rootIdsByItemId.put(item.getId(), item.getRootId());
+            updatedCount++;
+        }
+
+        log.debug("Updated {} portal navigation items for environment {}", updatedCount, environmentId);
+    }
+
+    private static boolean needsRootIdMigration(PortalNavigationItem item) {
+        String rootId = item.getRootId();
+        return rootId == null || rootId.isBlank() || ZERO_ROOT_ID.equals(rootId);
+    }
+
+    private static void cacheRootId(List<String> visitedItemIds, String rootId, Map<String, String> rootIdsByItemId) {
+        for (String itemId : visitedItemIds) {
+            rootIdsByItemId.put(itemId, rootId);
+        }
+    }
+
+    private static String findRootId(
+        PortalNavigationItem item,
+        Map<String, PortalNavigationItem> itemsById,
+        Map<String, String> rootIdsByItemId
+    ) {
+        String cached = rootIdsByItemId.get(item.getId());
+        if (cached != null) {
+            return cached;
+        }
+
+        List<String> visitedItemIds = new ArrayList<>();
+        Set<String> visitedItemIdSet = new HashSet<>();
+        PortalNavigationItem current = item;
+        int depth = 0;
+
+        while (depth < MAX_PARENT_CHAIN_DEPTH) {
+            String currentId = current.getId();
+
+            String cachedRootId = rootIdsByItemId.get(currentId);
+            if (cachedRootId != null) {
+                cacheRootId(visitedItemIds, cachedRootId, rootIdsByItemId);
+                return cachedRootId;
+            }
+
+            if (!visitedItemIdSet.add(currentId)) {
+                cacheRootId(visitedItemIds, currentId, rootIdsByItemId);
+                rootIdsByItemId.put(currentId, currentId);
+                return currentId;
+            }
+
+            visitedItemIds.add(currentId);
+
+            if (current.getParentId() == null) {
+                cacheRootId(visitedItemIds, currentId, rootIdsByItemId);
+                return currentId;
+            }
+
+            PortalNavigationItem parent = itemsById.get(current.getParentId());
+            if (parent == null) {
+                cacheRootId(visitedItemIds, currentId, rootIdsByItemId);
+                return currentId;
+            }
+
+            current = parent;
+            depth++;
+        }
+
+        String fallbackRootId = visitedItemIds.isEmpty() ? item.getId() : visitedItemIds.get(visitedItemIds.size() - 1);
+        cacheRootId(visitedItemIds, fallbackRootId, rootIdsByItemId);
+        return fallbackRootId;
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
@@ -23,14 +23,12 @@ import io.gravitee.repository.exceptions.TechnicalException;
 import io.gravitee.repository.management.api.EnvironmentRepository;
 import io.gravitee.repository.management.api.PortalNavigationItemRepository;
 import io.gravitee.repository.management.model.PortalNavigationItem;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Deque;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 import lombok.CustomLog;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.stereotype.Component;
@@ -45,11 +43,6 @@ import org.springframework.stereotype.Component;
 @CustomLog
 public class PortalNavigationItemRootIdUpgrader implements Upgrader {
 
-    private static final int MAX_PARENT_CHAIN_DEPTH = 50;
-
-    /**
-     * Sentinel value for "no root" in the domain; repository may store null/empty or this UUID string.
-     */
     private static final String ZERO_ROOT_ID = "00000000-0000-0000-0000-000000000000";
 
     private final EnvironmentRepository environmentRepository;
@@ -70,10 +63,10 @@ public class PortalNavigationItemRootIdUpgrader implements Upgrader {
 
     @Override
     public boolean upgrade() throws UpgraderException {
-        return this.wrapException(this::applyUpgrade);
+        return this.wrapException(this::migrateAllEnvironments);
     }
 
-    private boolean applyUpgrade() throws TechnicalException {
+    private boolean migrateAllEnvironments() throws TechnicalException {
         for (final var environment : environmentRepository.findAll()) {
             migrateEnvironment(environment.getOrganizationId(), environment.getId());
         }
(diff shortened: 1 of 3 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private static String resolveRootId(PortalNavigationItem item, Map<String, PortalNavigationItem> itemsById) {
+        if (item.getParentId() == null) {
+            return item.getId();
+        }
+
+        PortalNavigationItem current = item;
+        while (current.getParentId() != null) {
+            PortalNavigationItem parent = itemsById.get(current.getParentId());
```
jaroslawlakomy-grav: Put some depth limit (like 50)  here to avoid infinite loop on broken data

jaroslawlakomy-grav (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,108 @@
...
+            if (parent == null) {
+                return current.getId();
+            }
+            current = parent;
+        }
```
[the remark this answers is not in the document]
dev-marek: Won't happen

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,162 @@
...
+                item.setRootId(rootId);
+                portalNavigationItemRepository.update(item);
+                log.debug("Set rootId={} on portal navigation item {}", rootId, item.getId());
+            }
+        }
+        return true;
+    }
+
+    private static boolean hasValidRootId(PortalNavigationItem item) {
+        return item.getRootId() != null && !item.getRootId().isBlank();
```
dev-marek: Please also check if we're not dealing with zero-id

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,162 @@
...
+
+        List<PortalNavigationItem> path = new ArrayList<>();
+        Set<String> visitedIds = new HashSet<>();
+        PortalNavigationItem current = item;
+        int depth = 0;
+
+        while (depth <= MAX_ROOT_RESOLUTION_DEPTH) {
+            String currentId = current.getId();
+            if (visitedIds.contains(currentId)) {
+                for (PortalNavigationItem pathNode : path) {
```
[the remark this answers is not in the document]
aqibmohammadkhan: New helper cacheResolvedRootIdsForPath(List<PortalNavigationItem> path, String effectiveRoot, Map<String, String> resolvedRootIds)

Iterates over path and puts (pathNode.getId(), effectiveRoot) into resolvedRootIds.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+        return true;
+    }
+
+    private static boolean hasValidRootId(PortalNavigationItem item) {
+        return item.getRootId() != null && !item.getRootId().isBlank();
+    }
+
+    private static String resolveRootId(
```
dev-marek: why static?

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
aqibmohammadkhan: It never touches `this`, `environmentRepository`, or `portalNavigationItemRepository`. So it’s a stateless helper that could live on the class rather than on a particular instance.

jaroslawlakomy-grav (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,170 @@
...
+        int updatedCount = 0;
+
+        for (PortalNavigationItem item : items) {
+            if (!needsRootIdMigration(item)) {
+                rootIdsByItemId.put(item.getId(), item.getRootId());
+                continue;
+            }
+
+            item.setRootId(findRootId(item, itemsById, rootIdsByItemId));
+            portalNavigationItemRepository.update(item);
```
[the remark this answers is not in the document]
aqibmohammadkhan: only partially true in theory. It still updates inside the loop, but PortalNavigationItemRepository does not expose a batch update API, so there isn’t an obvious cleaner repository-level optimization in scope for this upgrader.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,172 @@
...
+        return rootId == null || rootId.isBlank() || ZERO_ROOT_ID.equals(rootId);
+    }
+
+    private static void cacheRootId(List<String> visitedItemIds, String rootId, Map<String, String> rootIdsByItemId) {
+        for (String itemId : visitedItemIds) {
+            rootIdsByItemId.put(itemId, rootId);
+        }
+    }
+
+    private static String findRootId(
```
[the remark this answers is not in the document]
aqibmohammadkhan: not necessarily. For a small migration helper, keeping the traversal in one place is probably clearer than re-splitting it into more abstractions.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,172 @@
...
+            environmentId
+        );
+
+        Map<String, PortalNavigationItem> itemsById = items
+            .stream()
+            .collect(Collectors.toMap(PortalNavigationItem::getId, Function.identity()));
+        Map<String, String> rootIdsByItemId = new HashMap<>();
+        int updatedCount = 0;
+
+        for (PortalNavigationItem item : items) {
```
dev-marek: Instead of iterating over all unordered items, I'd propose working on queue:
1. Initialize queue with `item.parentId == null` (roots) and map `resolved<String, String>`
2. In loop:
3. item <- queue.pop()
4. if item has invalid root: `item.rootId = item.id` if `parentId == null` or `item.rootId = resolved.get(item.parentId)`
5. update item
6. `resolved.put(item.id, item.rootId)`
7. `queue.addAll(items where parentId == item.id)` and remove them from the list
8. if `queue` is not empty, go to 2
9. if `items` are not empty -> log them, throw exception

dev-marek (changes requested):
One simplification suggested

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
aqibmohammadkhan: I refactored the upgrader to a queue-based parent-first traversal:

- queue is seeded with roots (parentId == null)
- resolved roots are stored in a map and reused for children
- children are enqueued only after parent resolution
- Also added a strict unresolved check: if items remain after queue drain (e.g., orphan/cycle/disconnected graph), we now log unresolved IDs and fail with a technical exception instead of assigning fallback root IDs.

Tests were updated accordingly for orphan/cycle cases to assert failure and no unintended updates.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,156 @@
...
+            String resolvedRootId = item.getParentId() == null ? item.getId() : resolvedRootIdsByItemId.get(item.getParentId());
+            if (resolvedRootId == null) {
+                continue;
+            }
+
+            if (!needsRootIdMigration(item)) {
+                resolvedRootIdsByItemId.put(item.getId(), item.getRootId());
+            } else {
+                item.setRootId(resolvedRootId);
+                portalNavigationItemRepository.update(item);
```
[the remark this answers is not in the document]
aqibmohammadkhan: the current repository API in this module exposes single-item update semantics,
this migration is a one-time startup upgrader (not a hot-path request flow),
changing to bulk writes would require broader repository/API changes and additional cross-storage validation (JDBC/Mongo behavior), which is outside the scope of this PR.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,156 @@
...
+        return true;
+    }
+
+    private void migrateEnvironment(String organizationId, String environmentId) throws TechnicalException {
+        List<PortalNavigationItem> items = portalNavigationItemRepository.findAllByOrganizationIdAndEnvironmentId(
+            organizationId,
+            environmentId
+        );
+
+        Map<String, PortalNavigationItem> pendingItemsById = items
```
dev-marek: Why do you need this map?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,156 @@
...
+            .stream()
+            .filter(item -> item.getParentId() == null)
+            .forEach(queue::addLast);
+
+        Map<String, String> resolvedRootIdsByItemId = new HashMap<>();
+        int updatedCount = 0;
+
+        while (!queue.isEmpty()) {
+            PortalNavigationItem item = queue.removeFirst();
+            if (!pendingItemsById.containsKey(item.getId())) {
```
dev-marek: If item is in queue = it is pending. There are no duplicates.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,156 @@
...
+                continue;
+            }
+
+            String resolvedRootId = item.getParentId() == null ? item.getId() : resolvedRootIdsByItemId.get(item.getParentId());
+            if (resolvedRootId == null) {
```
dev-marek: Not possible because queue operates hierarchy top-down and there will always be resolved parent

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,156 @@
...
+                continue;
+            }
+
+            if (!needsRootIdMigration(item)) {
+                resolvedRootIdsByItemId.put(item.getId(), item.getRootId());
```
dev-marek: map.put should be outside condition (after line 123) and then we can replace else with if(needsRootIdMigration)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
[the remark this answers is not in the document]
dev-marek: It is what it is

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,156 @@
...
+            organizationId,
+            environmentId
+        );
+
+        Map<String, PortalNavigationItem> pendingItemsById = items
+            .stream()
+            .collect(
+                Collectors.toMap(PortalNavigationItem::getId, Function.identity(), (existing, replacement) -> existing, LinkedHashMap::new)
+            );
+        Map<String, List<PortalNavigationItem>> childrenByParentId = items
```
dev-marek: Not really necessary

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,156 @@
...
+                resolvedRootIdsByItemId.put(item.getId(), item.getRootId());
+                updatedCount++;
+            }
+
+            pendingItemsById.remove(item.getId());
+            for (PortalNavigationItem child : childrenByParentId.getOrDefault(item.getId(), List.of())) {
+                if (pendingItemsById.containsKey(child.getId())) {
+                    queue.addLast(child);
+                }
+            }
```
dev-marek: would be easier if you'd collect children from stream to list, and add all to queue (https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#addAll-java.util.Collection-) and removeAll from items (https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#removeAll-java.util.Collection-)

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
aqibmohammadkhan: this was done against the aikido bot comments. 
Removed. Using the items list directly to track unprocessed items.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
aqibmohammadkhan: this was also done against bot review.
Removed the containsKey guard.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
aqibmohammadkhan: Removed the resolvedRootId == null null check.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
aqibmohammadkhan: resolved.put(...) now runs unconditionally, then if (needsRootIdMigration) without else.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
aqibmohammadkhan: Removed. Children are filtered from items on each iteration.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
aqibmohammadkhan: Using queue.addAll(children) and items.removeAll(children).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemRootIdUpgrader.java
```diff
@@ -0,0 +1,131 @@
...
+            String rootId = item.getParentId() == null ? item.getId() : resolved.get(item.getParentId());
+            resolved.put(item.getId(), rootId);
+
+            if (needsRootIdMigration(item)) {
+                item.setRootId(rootId);
+                portalNavigationItemRepository.update(item);
+                updatedCount++;
+            }
+
+            List<PortalNavigationItem> children = items
```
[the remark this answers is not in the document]
aqibmohammadkhan: no as map iteration introduces unnecessary  complexities.

dev-marek (approved):
LGTM

PR #15588, opened 2026-03-10
fix: allow user attributes in mail templates

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-12958

## Description
User attributes such as ${user.displayName} were not resolved in email templates during subscription notifications. Owner attributes were working correctly.

This change ensures user attributes are properly evaluated in both placeholders and conditional expressions within templates.

Additional context
When a user creates a subscription, notification emails are sent to the API owner and the application owner. These email templates support variable interpolation (e.g. ${owner.displayName}, ${api.name}) via Gravitee's NotificationParamsBuilder.

However, user-specific attributes such as ${user.displayName} were not being resolved in these templates. The user object was never added to the notification parameters during subscription creation, causing the placeholders to silently render as empty strings or fail conditional expressions in templates.

API owner attributes (${owner.*}) worked correctly because they were already mapped separately. The missing piece was the subscribing user's UserEntity.

Documentation : https://documentation.gravitee.io/apim/configure-and-manage-the-platform/gravitee-gateway/notifications

Approach:

Retrieve the current UserDetails from the SecurityContext.
Prefer UserDetails#getId() as the lookup key; fall back to UserDetails#getUsername() for token/JWT-based auth flows where id may be null.
Call userService.findById(executionContext, identifier) to load the full UserEntity.
If resolution succeeds, add the user to the params builder via paramsBuilder.user(subscribedByUser).
If the authenticated user is null, blank, or the lookup throws (e.g. user not found), the notification is still triggered — the user param is simply absent. No exception is propagated.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/SubscriptionServiceImpl.java
@@ -870,6 +870,17 @@ public class SubscriptionServiceImpl extends AbstractService implements Subscrip
             .application(applicationEntity);
         if (subscriptionEntity != null) {
             paramsBuilder.subscription(subscriptionEntity);
+            String subscribedBy = subscriptionEntity.getSubscribedBy();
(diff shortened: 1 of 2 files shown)


vikrantgravitee:
Duplicate PR.

PR #15590, opened 2026-03-10
subscription form minor bugs fixes

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-13077

## Description

**Placeholder fix:** GmdTextareaComponent rendered the literal string "undefined" as placeholder attribute when no value was passed. Fixed by binding via [attr.placeholder] with a ?? null fallback, which removes the attribute entirely when the input is not set.

**Subscription metadata encoding fix:** Special characters in subscription form metadata (e.g. @ in email addresses) were being stored in the database as HTML entities (&#64;, &#43;, etc.) due to OWASP HTML Sanitizer over-encoding plain-text values. Added an unescapeSafeChars post-processing step in SubscriptionMetadataSanitizer that restores these characters after sanitization. XSS-relevant entities (&lt;, &gt;, &amp;) are intentionally left encoded.

### Placeholder fix
Before:
Zrzut ekranu 2026-03-10 o 19 12 11

After: 
image
 
**Subscription metadata encoding fix:**

Before:
image

After:
image

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/validation/SubscriptionMetadataSanitizer.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/validation/SubscriptionMetadataSanitizer.java
@@ -15,7 +15,6 @@
  */
 package io.gravitee.rest.api.service.v4.validation;
 
-import io.gravitee.rest.api.service.sanitizer.HtmlSanitizer;
 import io.gravitee.rest.api.service.v4.exception.SubscriptionMetadataInvalidException;
 import java.util.Collections;
 import java.util.HashMap;
@@ -24,7 +23,9 @@ import java.util.regex.Pattern;
 import org.springframework.stereotype.Component;
 
 /**
- * Validates and sanitizes subscription form metadata (keys, value length, HTML).
+ * Validates and sanitizes subscription form metadata (keys, value length).
+ * Values are plain text; HTML tags are stripped to prevent XSS when metadata is rendered.
+ * No HTML encoding is applied, so characters like {@code @}, {@code +}, {@code =} are stored as-is.
  *
  * @author GraviteeSource Team
  */
@@ -32,17 +33,13 @@ import org.springframework.stereotype.Component;
 public class SubscriptionMetadataSanitizer {
 
     private static final Pattern KEY_PATTERN = Pattern.compile("^[A-Za-z0-9_-]{1,100}$");
+    private static final Pattern HTML_TAG = Pattern.compile("<[^>]*>");
     private static final int MAX_VALUE_LENGTH = 1024;
     private static final int MAX_METADATA_COUNT = 25;
 
-    private final HtmlSanitizer htmlSanitizer;
-
-    public SubscriptionMetadataSanitizer(HtmlSanitizer htmlSanitizer) {
-        this.htmlSanitizer = htmlSanitizer;
-    }
-
     /**
-     * Validates metadata keys and value lengths, then sanitizes each value with HTML sanitizer.
+     * Validates metadata keys and value lengths, then strips HTML tags from each value.
+     * Any remaining characters (including {@code <}, {@code >}, non-Latin, etc.) are stored as-is.
      *
      * @param metadata raw metadata from the client
      * @return sanitized metadata, or empty map if input is null
@@ -78,7 +75,7 @@ public class SubscriptionMetadataSanitizer {
                 );
             }
 
-            String sanitizedValue = htmlSanitizer.sanitize(value);
+            String sanitizedValue = stripHtmlTags(value);
             if (sanitizedValue == null || sanitizedValue.isBlank()) {
                 continue;
             }
@@ -86,5 +83,16 @@ public class SubscriptionMetadataSanitizer {
         }
 
         return sanitizedMetadata;
+    }
+
+    /**
+     * Removes HTML tags from plain-text metadata. Used instead of OWASP HTML Sanitizer so that
+     * special characters (e.g. {@code @}, {@code +}, {@code =}) are not encoded.
+     */
+    private static String stripHtmlTags(String content) {
+        if (content == null) {
+            return null;
+        }
+        return HTML_TAG.matcher(content).replaceAll("").trim();
     }
 }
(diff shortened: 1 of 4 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/validation/SubscriptionMetadataSanitizer.java
```diff
@@ -87,4 +87,22 @@ public Map<String, String> sanitizeAndValidate(Map<String, String> metadata) {
...
+            return null;
+        }
+        return content
+            .replace("&#64;", "@")
+            .replace("&#43;", "+")
+            .replace("&#61;", "=")
+            .replace("&#39;", "'")
+            .replace("&#34;", "\"")
+            .replace("&#96;", "`");
+    }
```
aqibmohammadkhan: i feel a little less confident with this change due to security concern.
Unescaping backtick (`) and double-quote (") is risky

I see the description states "XSS-relevant entities (<, >, &) are intentionally left encoded," the threat model seems incomplete — <, >, & are not the only XSS-relevant characters. If these metadata values are ever rendered in HTML without additional escaping, quotes and backticks can be exploited.

wdyt about using `StringEscapeUtils.unescapeHtml4()` from Apache Commons Text and then re-escape only dangerous chars (<, >, &, ", ').
Or, configure the `OWASP sanitizer` with a custom policy that doesn't encode these characters in the first place, since the inputs are plain-text metadata values and not HTML content.

aqibmohammadkhan: Consider whether OWASP HTML Sanitizer is the right tool for plain-text metadata at all. If the values are never rendered as raw HTML, a simpler validation (reject or strip HTML tags) might be more appropriate and avoid the encoding dance entirely.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/validation/SubscriptionMetadataSanitizerTest.java
```diff
@@ -112,4 +112,40 @@ void should_sanitize_values_and_omit_empty_values() {
...
+        when(htmlSanitizer.sanitize("my@company.com")).thenReturn("my&#64;company.com");
+
+        var result = cut.sanitizeAndValidate(Map.of("email", "my@company.com"));
+
+        assertThat(result).containsEntry("email", "my@company.com");
+    }
+
+    @Test
+    void should_unescape_safe_special_chars_over_encoded_by_owasp() {
+        when(htmlSanitizer.sanitize("key&#61;value&#43;1")).thenReturn("key&#61;value&#43;1");
```
aqibmohammadkhan: The mock is given already-encoded input and returns the same thing. In reality, the sanitizer receives raw input like "key=value+1" and returns "`key&#61;value&#43;1`". 
wdyt about below change

````
when(htmlSanitizer.sanitize("key=value+1")).thenReturn("key&#61;value&#43;1");
var result = cut.sanitizeAndValidate(Map.of("field_a", "key=value+1"));
assertThat(result).containsEntry("field_a", "key=value+1");
````

File: gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/validation/SubscriptionMetadataSanitizerTest.java [hunk 3]
```diff
@@ -112,4 +112,40 @@ void should_sanitize_values_and_omit_empty_values() {
...
+        assertThat(result)
+            .containsEntry("field_a", "key=value+1")
+            .containsEntry("field_b", "it's a \"test\"")
+            .containsEntry("field_c", "code`snippet");
+    }
+
+    @Test
+    void should_not_unescape_xss_relevant_html_entities() {
+        when(htmlSanitizer.sanitize("&lt;b&gt;bold&lt;/b&gt;")).thenReturn("&lt;b&gt;bold&lt;/b&gt;");
+
```
aqibmohammadkhan, line 127: same as https://github.com/gravitee-io/gravitee-api-management/pull/15590/changes#r2917311472

passes HTML-encoded &lt;b&gt;bold&lt;/b&gt; as the input to sanitize(), but in production the input would be the raw `<b>bold</b>`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/validation/SubscriptionMetadataSanitizer.java
```diff
@@ -87,4 +87,22 @@ public Map<String, String> sanitizeAndValidate(Map<String, String> metadata) {
 
         return sanitizedMetadata;
     }
+
+    /**
+     * Reverses the over-encoding that OWASP HTML Sanitizer applies to characters that are safe in plain-text metadata
+     * values (e.g. {@code @} in email addresses is encoded as {@code &#64;}).
+     * We intentionally do NOT unescape {@code &lt;}, {@code &gt;}, or {@code &amp;} to preserve XSS protection.
+     */
+    private static String unescapeSafeChars(String content) {
```
aqibmohammadkhan: wdyt about making it package-private (default access) instead of `private static` so it can be unit-tested independently.

kovaren:
This is a bit unrelated to the PR, but why is the font different on the second screenshot (the first font is the correct one)
image

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/validation/SubscriptionMetadataSanitizer.java
JedrzejJanasiak: Yeah, so I went with the HTML tag strip approach :)

JedrzejJanasiak:
> This is a bit unrelated to the PR, but why is the font different on the second screenshot (the first font is the correct one) image

The font is different since the first screenshot was taken in the Portal Settings and the second one is from Storybook :)

aqibmohammadkhan (approved):
LGTM

PR #15592, opened 2026-03-10
Clear stale cert from mTLS subscriptions when all certificates expire

carlos-andres-osorio:
## Issue

[GKO-2450](https://gravitee.atlassian.net/browse/GKO-2450)

## Description

When all client certificates for an application expire or are revoked, mTLS subscriptions were left holding the stale certificate bundle. The gateway kept old trust store loaders active with the expired cert, creating inconsistent state.

**Root cause:** `ApplicationCertificatesUpdateDomainServiceImpl.updateActiveMTLSSubscriptions()` threw `ClientCertificateLastRemovalException` when no active certs remained, so the subscription was never updated.

**Changes:**

- **ApplicationCertificatesUpdateDomainServiceImpl** — Remove the exception when no active certificates remain. Instead, update mTLS subscriptions with an empty PKCS7 bundle, clearing the stale cert from the DB and triggering gateway sync.
- **SubscriptionTrustStoreLoader.readSubscriptionCertificate** — Return an empty set when the client certificate is null/blank, and pass `allowEmpty=true` to `PKCS7Utils.pkcs7ToTruststore` so empty bundles are handled gracefully.
- **SubscriptionCacheService** — Added test covering the transition to an empty certificate bundle.

## Additional context

- The empty PKCS7 bundle flows through `SubscriptionTrustStoreLoaderManager.registerSubscription()`: the update path removes all existing loaders via diff, and the new subscription path registers no loaders.
- Tests cover null cert, blank cert, empty PKCS7 bundle, and valid PKCS7 bundle scenarios.

## Manual test scenarios

I ran the following tests:

- Happy path: mTLS request with a valid client certificate succeeds (200).
- Main bug scenario: after all certs expire and the cron job runs, the gateway rejects the old certificate.
- Partial expiry: the remaining valid cert still allows mTLS access after one cert expires.
- Partial expiry: the expired cert is rejected after another cert on the same app has expired.
- Recovery: adding a new cert after full expiry restores mTLS access.
- Manual revocation: gateway rejects mTLS after certs are manually revoked and the gateway syncs.
- API Key subscription on the same API is unaffected when the mTLS app's certs expire.
- OAuth2/JWT subscription on the same API is unaffected when the mTLS app's certs expire.
- mTLS subscription on a different application is unaffected when another app's certs expire.
- Same app subscribed to mTLS plans on two different APIs: both subscriptions are cleared after full expiry.
- New clientId-based (OAuth2/JWT) subscription on the same application works normally after its mTLS certs have been cleared.

[GKO-2450]: https://gravitee.atlassian.net/browse/GKO-2450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/test/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheServiceTest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/test/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheServiceTest.java
@@ -270,6 +270,29 @@ class SubscriptionCacheServiceTest {
         }
 
         @Test
+        void should_update_trust_store_when_subscription_transitions_to_empty_certificate() {
+            // Register with a client certificate
+            Subscription subscription = buildAcceptedSubscriptionWithClientCertificate(SUB_ID, API_ID, CLIENT_CERTIFICATE, PLAN_ID);
+            subscriptionService.register(subscription);
+
+            assertThat(subscriptionService.getById(SUB_ID)).isPresent();
+            assertThat(subscriptionService.getByClientCertificate(subscription)).isPresent();
+
+            // Re-register with empty string — simulates all certs expired
+            Subscription updatedWithEmptyCert = buildAcceptedSubscriptionWithClientCertificate(SUB_ID, API_ID, "", PLAN_ID);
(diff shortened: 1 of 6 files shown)


benoitgravitee (dismissed):
I don't agree with this fix.
I know the overall architecture of this feature lacks consistency.

At the moment this service is the safe guard to not remove all cert by accident on ongoing subscriptions.

Let's discuss a plan to mitigate. I'm afraid merging this will create other bug. .

benoitgravitee (approved):

PR #15610, opened 2026-03-11
Not found-404 requests not visible in analytics

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-12831

## Description

| | |
|---|---|
| **Problem** | 404 (not found) requests to non-existent APIs do not appear in platform analytics (e.g. Status pie chart), even with `handlers.notfound.analytics.enabled: true`. |
| **Cause** | Analytics queries filter by API IDs from the DB. The gateway stores not-found requests with `api="1"` (UNKNOWN_SERVICE), which is never in that filter, so those documents are excluded. |
| **Fix** | Add the not-found sentinel `"1"` to the API terms used in analytics whenever the filter is on APIs (not applications). |

## Changes

| Area | Change |
|------|--------|
| **PlatformAnalyticsResource** (v1 `.../platform/analytics`) | In `findApiIds()`, always add `UNKNOWN_SERVICE` (`"1"`) to the set so the Analytics tab Status widget includes 404s. |
| **EnvironmentAnalyticsResource** (v1 `.../analytics`) | In `buildTerms()`, when `fieldName` is `API_FIELD`, add `UNKNOWN_SERVICE` to the ids before building the terms map. |
| **SearchEnvironmentResponseStatusRangesUseCase** (v2 `.../analytics/response-status-ranges`) | Add `UNKNOWN_SERVICE` to the api IDs list and always include `DefinitionVersion.V2` in definition versions so the v2 index is queried and 404s are counted. |


pragsty:
Not needed, raising another PR with required commit only

pragsty:
Not needed, raising another PR with required commit only

PR #15623, opened 2026-03-11
fix: deploy dynamic dictionary on specific environment

superbird63:
## Issue

https://github.com/gravitee-io/issues/issues/11236


superbird63:
Fixed by https://github.com/gravitee-io/gravitee-api-management/pull/16025
As mentioned here: https://github.com/gravitee-io/issues/issues/11236#issuecomment-4141691808

PR #15646, opened 2026-03-11
fix(portal-next): JWT profile in API subscription advanced config

aditya-goyal01:
## Issue

https://gravitee.atlassian.net/browse/APIM-12981

## Summary

This Pull Request introduces a new authentication mechanism, "JWT Profile for OAuth2", to the webhook subscription configuration module. It adds support for configuring authentication parameters required for JWT-based OAuth2 flows.

## Main Changes 
- Template Updates: Added new UI elements in the HTML template to configure JWT settings (issuer, subject, audience, signature algorithm, key source, etc.).
- Support for optional fields like custom claims, jwtId, x509CertChain, and other cryptographic keys and algorithms.
Component Updates
- Defined new interfaces for JwtProfileOauth2AuthConfiguration, including claims, keystore options, cryptography, and other related attributes.

### DIFF
--- a/gravitee-apim-portal-webui-next/src/components/subscription/webhook/consumer-configuration-authentification/consumer-configuration-authentication.component.ts
+++ b/gravitee-apim-portal-webui-next/src/components/subscription/webhook/consumer-configuration-authentification/consumer-configuration-authentication.component.ts
@@ -28,6 +28,8 @@ import {
   Validator,
   Validators,
 } from '@angular/forms';
+import { MatButtonModule } from '@angular/material/button';
+import { MatCheckboxModule } from '@angular/material/checkbox';
 import { MatChipInputEvent, MatChipsModule } from '@angular/material/chips';
 import { MatError, MatFormFieldModule, MatHint } from '@angular/material/form-field';
 import { MatIcon } from '@angular/material/icon';
@@ -38,6 +40,7 @@ import { tap } from 'rxjs';
 import { AUTHENTICATION_TYPES, AuthFormValue } from './consumer-configuration-authentication.model';
 import {
   BasicAuthConfiguration,
+  JwtProfileOauth2AuthConfiguration,
   Oauth2AuthConfiguration,
   TokenAuthConfiguration,
   WebhookSubscriptionConfigurationAuth,
@@ -59,6 +62,8 @@ import { AccordionModule } from '../../../accordion/accordion.module';
     AccordionModule,
     MatChipsModule,
     MatIcon,
+    MatCheckboxModule,
+    MatButtonModule,
   ],
   templateUrl: './consumer-configuration-authentication.component.html',
   styleUrls: ['consumer-configuration-authentication.component.scss'],
@@ -77,6 +82,8 @@ import { AccordionModule } from '../../../accordion/accordion.module';
 })
 export class ConsumerConfigurationAuthenticationComponent implements AfterViewInit, ControlValueAccessor, Validator {
   scopes = signal<string[]>([]);
+  customClaims = signal<{ name: string; value: string }[]>([]);
+  customClaimError = signal<string | null>(null);
 
   authForm = new FormGroup({
     type: new FormControl<WebhookSubscriptionConfigurationAuthType>('none', { nonNullable: true }),
@@ -93,10 +100,27 @@ export class ConsumerConfigurationAuthenticationComponent implements AfterViewIn
     clientId: new FormControl(),
     clientSecret: new FormControl(),
     scopes: new FormControl(),
+
+    // for jwtProfileOauth2 authentication
+    issuer: new FormControl(),
+    subject: new FormControl(),
+    audience: new FormControl(),
+    expirationTime: new FormControl(),
+    expirationTimeUnit: new FormControl(),
+    signatureAlgorithm: new FormControl(),
+    keySource: new FormControl(),
+    jwtId: new FormControl(),
+    secretBase64Encoded: new FormControl(),
+    x509CertChain: new FormControl(),
+    alias: new FormControl(),
+    storePassword: new FormControl(),
+    keyPassword: new FormControl(),
+    keyId: new FormControl(),
+    keyContent: new FormControl(),
+    customClaims: new FormControl(),
   });
   readonly types = AUTHENTICATION_TYPES;
   private readonly destroyRef = inject(DestroyRef);
-  private isDisabled = false;
 
   constructor() {
     this.authForm.controls.type.valueChanges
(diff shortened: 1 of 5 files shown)


gravitee-apim-portal-webui-next/src/components/subscription/webhook/consumer-configuration-authentification/consumer-configuration-authentication.component.ts
```diff
@@ -123,6 +147,11 @@ export class ConsumerConfigurationAuthenticationComponent implements AfterViewIn
     this.authForm.patchValue({});
     if (this.authForm.controls.type.value === 'oauth2' && this.authForm.controls.scopes.getRawValue()) {
       this.scopes.set(this.authForm.controls.scopes.getRawValue());
+      this.authForm.controls.scopes.setValue(this.scopes());
```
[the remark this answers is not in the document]
aditya-goyal01: Fixed in this PR — added `{ emitEvent: false }` to all `setValue` calls in `writeValue` and `ngAfterViewInit`.

gravitee-apim-portal-webui-next/src/components/subscription/webhook/consumer-configuration-authentification/consumer-configuration-authentication.component.ts
```diff
@@ -93,6 +99,24 @@ export class ConsumerConfigurationAuthenticationComponent implements AfterViewIn
     clientId: new FormControl(),
     clientSecret: new FormControl(),
     scopes: new FormControl(),
+
+    // for jwtProfileOauth2 authentication
```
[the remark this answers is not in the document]
aditya-goyal01: Fixed — comment removed.

gravitee-apim-portal-webui-next/src/components/subscription/webhook/consumer-configuration-authentification/consumer-configuration-authentication.component.ts
```diff
@@ -156,16 +193,34 @@ export class ConsumerConfigurationAuthenticationComponent implements AfterViewIn
 
   removeScope(scopeToRemove: string) {
     this.scopes.set([...this.scopes().filter(scope => scope !== scopeToRemove)]);
+    this.authForm.controls.scopes.setValue(this.scopes());
```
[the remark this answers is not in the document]
aditya-goyal01: Intentional. `addScope`/`removeScope` are user interactions — they must emit `valueChanges` so the parent `FormControl` is notified of the change. Suppressing events here would break the `ControlValueAccessor` contract: the parent form would never receive scope mutations made by the user.

gravitee-apim-portal-webui-next/src/components/subscription/webhook/consumer-configuration-authentification/consumer-configuration-authentication.component.ts
```diff
@@ -204,6 +279,30 @@ export class ConsumerConfigurationAuthenticationComponent implements AfterViewIn
...
+            expirationTimeUnit: authFormValue.expirationTimeUnit,
+            signatureAlgorithm: authFormValue.signatureAlgorithm,
+            keySource: authFormValue.keySource,
+            jwtId: authFormValue.jwtId,
+            secretBase64Encoded: authFormValue.secretBase64Encoded,
+            x509CertChain: authFormValue.x509CertChain,
+            keystoreOptions: {
+              alias: authFormValue.alias,
+              storePassword: authFormValue.storePassword,
+              keyPassword: authFormValue.keyPassword,
```
mukul-tyagi08: we always add this? I mean, we can conditionally add this?

gravitee-apim-portal-webui-next/src/components/subscription/webhook/consumer-configuration-authentification/consumer-configuration-authentication.component.ts
```diff
@@ -156,16 +193,34 @@ export class ConsumerConfigurationAuthenticationComponent implements AfterViewIn
...
     const value = (event.value || '').trim();
     if (value) {
       this.scopes.set([...this.scopes(), value]);
+      this.authForm.controls.scopes.setValue(this.scopes());
     }
     event.chipInput.clear();
   }
 
+  removeCustomClaim(claimToRemove: { name: string; value: string }) {
+    this.customClaims.set([...this.customClaims().filter(claim => claim.name !== claimToRemove.name)]);
```
mukul-tyagi08: Just to confirm, can we have 2 claims with same name but different values? If yes, we need to filter by both name and value, else, we need to prevent same names for 2 in addCustomClaim.

gravitee-apim-portal-webui-next/src/components/subscription/webhook/consumer-configuration-authentification/consumer-configuration-authentication.component.ts
```diff
@@ -232,6 +331,17 @@ export class ConsumerConfigurationAuthenticationComponent implements AfterViewIn
...
         break;
       }
+      case 'jwtProfileOauth2': {
+        this.authForm.controls.issuer.setValidators([Validators.required]);
+        this.authForm.controls.subject.setValidators([Validators.required]);
+        this.authForm.controls.audience.setValidators([Validators.required]);
+        this.authForm.controls.expirationTime.setValidators([Validators.required]);
+        this.authForm.controls.expirationTimeUnit.setValidators([Validators.required]);
+        this.authForm.controls.signatureAlgorithm.setValidators([Validators.required]);
+        this.authForm.controls.keySource.setValidators([Validators.required]);
```
mukul-tyagi08: keySource field is only visible in the template when signatureAlgorithm === 'RSA_RS256'. For others the field is hidden but still required, which might block form validity?

aditya-goyal01: Yes, right point. Change applied. Please verify and resolve conversation.

gravitee-apim-portal-webui-next/src/components/subscription/webhook/consumer-configuration-authentification/consumer-configuration-authentication.component.ts
aditya-goyal01: Ideally, users should not have one key for many values. Claims must be identified by unique keys based on my knowledge. However, `removeCustomClaim` extended to remove by name and value. Also, `addCustomClaim` is now guarded with not allowing same key to be added  and user sees error on the screen. They should be able protect the custom claims together.

gravitee-apim-portal-webui-next/src/components/subscription/webhook/consumer-configuration-authentification/consumer-configuration-authentication.component.ts
aditya-goyal01: True. It wasn't an error, however, worth saving three fields.

mukul-tyagi08 (approved):

PR #15673, opened 2026-03-12
feat: add view parameter to API search for documentation filtering

dev-marek:
## Issue

https://gravitee.atlassian.net/browse/APIM-13068

## Description

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/ApisResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/ApisResource.java
@@ -41,6 +41,7 @@ import io.gravitee.rest.api.service.common.GraviteeContext;
 import io.gravitee.rest.api.service.filtering.FilteringService;
 import io.gravitee.rest.api.service.v4.ApiCategoryService;
 import jakarta.inject.Inject;
+import jakarta.ws.rs.BadRequestException;
 import jakarta.ws.rs.BeanParam;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.POST;
@@ -172,6 +173,9 @@ public class ApisResource extends AbstractResource<Api, String> {
         try {
             final ExecutionContext executionContext = GraviteeContext.getExecutionContext();
             if ("documentation".equals(view)) {
+                if (category != null && !category.isEmpty()) {
+                    throw new BadRequestException("Parameters 'view' and 'category' are mutually exclusive.");
+                }
                 var pageable = new PageableImpl(paginationParam.getPage(), paginationParam.getSize());
                 var output = searchApisForPortalUseCase.execute(
                     new SearchApisForPortalUseCase.Input(
@@ -184,10 +188,10 @@ public class ApisResource extends AbstractResource<Api, String> {
                     )
                 );
                 List<String> ids = output.apis().getContent().stream().map(io.gravitee.apim.core.api.model.Api::getId).toList();
-                Map<String, Object> totalOnly = new HashMap<>();
-                totalOnly.put("totalElements", output.apis().getTotalElements());
+                // paginateMetaData.totalElements tells createListResponse the true total across all pages.
+                // paginateResultList then builds full pagination metadata (total_pages, first, last, size).
                 Map<String, Map<String, Object>> metadata = new HashMap<>();
-                metadata.put("paginateMetaData", totalOnly);
+                metadata.put("paginateMetaData", Map.of("totalElements", (Object) output.apis().getTotalElements()));
                 return createListResponse(executionContext, ids, paginationParam, metadata);
             }
             Collection<String> apisList;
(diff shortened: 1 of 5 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/ApisResource.java
```diff
@@ -160,10 +166,30 @@ public Response getApis(@BeanParam PaginationParam paginationParam, @BeanParam A
...
+                        pageable,
+                        null
+                    )
+                );
+                List<String> ids = output.apis().getContent().stream().map(io.gravitee.apim.core.api.model.Api::getId).toList();
+                Map<String, Object> totalOnly = new HashMap<>();
+                totalOnly.put("totalElements", output.apis().getTotalElements());
+                Map<String, Map<String, Object>> metadata = new HashMap<>();
+                metadata.put("paginateMetaData", totalOnly);
+                return createListResponse(executionContext, ids, paginationParam, metadata);
```
aqibmohammadkhan, line 195: `SearchApisForPortalUseCase` already performs server-side pagination (via pageable). Then `createListResponse` slices the resulting list again using `paginationParam`. Because the IDs list is already the size of one page (e.g. 2 items), the second pagination either:

- Returns the same items (page 1), or
- Returns an empty list (page > 1).

Suggestion: Either pass the full unpaginated list to createListResponse, or bypass createListResponse's internal pagination and only use it for response envelope building. 

Alternatively, build the response manually the way other paginated use-cases do.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/ApisResource.java
```diff
@@ -160,10 +166,30 @@ public Response getApis(@BeanParam PaginationParam paginationParam, @BeanParam A
     public Response searchApis(
         @QueryParam("q") String query,
         @QueryParam("category") String category,
+        @QueryParam("view") String view,
         @BeanParam PaginationParam paginationParam
     ) {
         try {
             final ExecutionContext executionContext = GraviteeContext.getExecutionContext();
+            if ("documentation".equals(view)) {
```
aqibmohammadkhan, line 175: what happens if a caller passes both `view=documentation&category=Guides` ?

I think, since we are branching here when view=documentation, the code branches before the category check. If a caller passes both, the category filter is silently dropped.

Suggestion: Either:

- Pass category into the use-case input (if the use case supports it), or
- Return a 400 Bad Request if both view and category are provided, or
- Document the mutual exclusivity in the OpenAPI spec.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/ApisResource.java
```diff
@@ -160,10 +166,30 @@ public Response getApis(@BeanParam PaginationParam paginationParam, @BeanParam A
...
+                        executionContext.getEnvironmentId(),
+                        executionContext.getOrganizationId(),
+                        getAuthenticatedUserOrNull(),
+                        query,
+                        pageable,
+                        null
+                    )
+                );
+                List<String> ids = output.apis().getContent().stream().map(io.gravitee.apim.core.api.model.Api::getId).toList();
+                Map<String, Object> totalOnly = new HashMap<>();
```
aqibmohammadkhan: Use Map.of() instead of new HashMap<>()
The maps created are never mutated after construction. Using Map.of(…) is more concise and makes immutability explicit

```
Map<String, Map<String, Object>> metadata = Map.of(
    "paginateMetaData", Map.of("totalElements", output.apis().getTotalElements())
);
```

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/ApisResource.java
```diff
@@ -160,10 +166,30 @@ public Response getApis(@BeanParam PaginationParam paginationParam, @BeanParam A
...
+                totalOnly.put("totalElements", output.apis().getTotalElements());
+                Map<String, Map<String, Object>> metadata = new HashMap<>();
+                metadata.put("paginateMetaData", totalOnly);
```
aqibmohammadkhan: This manually constructs pagination metadata (totalElements) and misses fields like `totalPages`, `first`, `last`, `number`, and `size` that the other code paths likely populate. Portal may rely on those fields.

Suggestion: Reuse whatever helper the existing code paths use to build full pagination metadata, or at minimum include `totalPages` and `size`.

dev-marek:
@aqibmohammadkhan /review

PR #15706, opened 2026-03-14
feat: add error kye filter for v2

bmouhaimen:
Description:

This PR is adding new Error Key filter for V2 API.

https://gravitee.atlassian.net/browse/APIM-11345

https://github.com/user-attachments/assets/675c66a1-68c0-47f5-8f44-47b4477367f5


phiz71:
Won't do

PR #15707, opened 2026-03-14
fix: restrict API product visibility based on user permissions

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-13137

## Description

Previously, the API Products section in the Gravitee APIM Console was accessible to all authenticated users regardless of their permission level. The `GetApiProductsUseCase` and `SearchApiProductsUseCase` did not take user context into account, meaning every user — including those without any API Product permissions — could list and search all API products in the environment. Additionally, the "API Products" entry in the side navigation menu was visible to all users, regardless of whether they had the `environment-api_product-r` permission.

This PR fixes the visibility and access control for API Products on both the frontend and backend:

- **Frontend**: The "API Products" side navigation item is now gated behind the `environment-api_product-r` permission. Users who do not hold this permission will no longer see the menu entry.
- - **Backend**: The `GetApiProductsUseCase` and `SearchApiProductsUseCase` now accept the authenticated user's ID and admin status as input. For non-admin users, the use cases filter results to only return API products where the user is a member. Admin users continue to see all API products as before.
## Additional context

**Before this change:**
- Any logged-in user could see and retrieve the full list of API products via the UI and the REST API, regardless of their membership or permissions.
- - The side nav "API Products" link was shown to all users.
**After this change:**
- The "API Products" menu item is only shown to users with the `environment-api_product-r` permission.
- - API product list and search endpoints now return only the products the requesting user is a member of (for non-admin users), preventing unauthorized visibility of products they do not own or belong to.
- - Admin users are unaffected and retain full visibility of all API products.

### DIFF
--- a/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
+++ b/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
@@ -46,7 +46,12 @@ describe('GioSideNavComponent', () => {
   const expirationDateInOneYear = new Date();
   expirationDateInOneYear.setFullYear(expirationDateInOneYear.getFullYear() + 1);
 
-  const init = async (licenseNotificationEnabled = true, hasLicenseMgmtPermission = true, scoringEnabled = true) => {
+  const init = async (
+    licenseNotificationEnabled = true,
+    hasLicenseMgmtPermission = true,
+    scoringEnabled = true,
+    hasApiProductPermission = true,
+  ) => {
     await TestBed.configureTestingModule({
       declarations: [GioSideNavComponent],
       imports: [NoopAnimationsModule, GioTestingModule, GioSideNavModule, MatIconTestingModule],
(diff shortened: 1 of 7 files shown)


gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
```diff
@@ -60,6 +65,10 @@ describe('GioSideNavComponent', () => {
               if (isMatchingLicenseNotificationPermission && !hasLicenseMgmtPermission) {
                 return false;
               }
+              const isMatchingApiProductPermission = permissions.length === 1 && permissions[0] === 'environment-api_product-r';
```
arpzilla-dev, line 68: This "environment-api_product-r" could be put in const variable.

arpzilla-dev: got it, its a matter format, not static variable

arpzilla-dev (approved):

PR #15711, opened 2026-03-16
fix: improve catalog styles.

kovaren:
## Issue

### Fix styles in Catalog:

- Fix the scrollable area: only cards view and table are scrollable so that the header and pagination stay in place
- Make cards adapt their width to avoid clipping on a responsive breakpoint trigger
- Fix the loading spinner's position (always in the middle of the variable area)
- Add a light overlay on the loading spinner
- Align the cards with the search field and the pagination on the right side
- Fix view mode toggle's height to align with the search field
- Remove the right padding from the search field

https://github.com/user-attachments/assets/54c27941-62cd-41ab-8821-a2afbe633bd3

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.html
+++ b/gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.html
@@ -27,93 +27,97 @@
     </div>
   </div>
 
-  @if (apiPaginator().data.length > 0) {
-    @if (viewMode() === 'grid') {
-      <app-cards-grid [cards]="apiPaginator().data">
-        <ng-template #cardTemplate let-api>
-          <app-api-card
-            [apiId]="api.id"
-            [title]="api.title"
-            [version]="api.version"
-            [content]="api.content"
-            [picture]="api.picture"
-            [isEnabledMcpServer]="api.isEnabledMcpServer"
-            (cardSelect)="navigateToApi($event)"
-          />
-        </ng-template>
-      </app-cards-grid>
-    } @else {
-      <div class="api-list__table-container">
-        <table mat-table [dataSource]="apiPaginator().data" class="api-list__table">
-          <ng-container matColumnDef="name">
-            <th mat-header-cell *matHeaderCellDef i18n="@@tableHeaderApiName">API Name</th>
-            <td mat-cell *matCellDef="let api">
-              <span class="api-list__table__name-text">{{ api.title }}</span>
-              @if (api.content) {
-                <span class="material-icons icon-small api-list__table__info-icon" [matTooltip]="api.content">info</span>
-              }
-            </td>
-          </ng-container>
-
-          <ng-container matColumnDef="labels">
-            <th mat-header-cell *matHeaderCellDef i18n="@@tableHeaderLabels">Labels</th>
-            <td mat-cell *matCellDef="let api">
-              @if (api.labels) {
-                @for (label of api.labels; track label) {
-                  <mat-chip># {{ label }}</mat-chip>
+  <div class="api-list__content">
+    <div class="api-list__scroll" [class.loading-empty-state]="loadingPage() && apiPaginator().data.length === 0">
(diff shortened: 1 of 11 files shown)


gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.html
```diff
@@ -27,93 +27,97 @@
...
+        @if (query()) {
+          <h2 class="next-gen-h5" i18n="@@noFilteredApisAvailable">No APIs found</h2>
+          <p i18n="@@noFilteredApisDescription">Your search didn't return any APIs. Please try again with different keywords.</p>
+        } @else {
+          <h2 class="next-gen-h5" i18n="@@noApisAvailable">No APIs available yet</h2>
+          <p i18n="@@noApisDescription">Our API catalog is currently being updated. More APIs will be available soon.</p>
+        }
+      </div>
+    }
+  </div>
```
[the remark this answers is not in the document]
kovaren: The empty state is not content that scrolls, it’s a replacement view when there’s nothing to show. Keeping it outside the scroll region preserves that distinction and avoids odd scrollbar and overflow behavior.

In a subsequent PR, there will be a consolidation of cards view / table experience and the empty state will probably be integrated inside those components. Having a min height is an acceptable trade-off for now.

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.html
```diff
@@ -27,93 +27,97 @@
...
-          <ng-container matColumnDef="name">
-            <th mat-header-cell *matHeaderCellDef i18n="@@tableHeaderApiName">API Name</th>
-            <td mat-cell *matCellDef="let api">
-              <span class="api-list__table__name-text">{{ api.title }}</span>
-              @if (api.content) {
-                <span class="material-icons icon-small api-list__table__info-icon" [matTooltip]="api.content">info</span>
-              }
-            </td>
-          </ng-container>
+  <div class="api-list__content" [class.api-list__scroll--loading-empty]="loadingPage() && apiPaginator().data.length === 0">
```
aqibmohammadkhan: `class.api-list__scroll--loading-empty` — applied on `api-list__content `but the CSS modifier --`loading-empty` is defined on `&__scroll`. This looks like the class binding is on the wrong element.

This means the `min-height: $empty-state-height `rule won't apply correctly since the class doesn't match the element it's on.

you can either move the `[class.api-list__scroll--loading-empty]` binding to the inner `api-list__scroll` div, or rename the CSS to `&__content--loading-empty`.

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.scss
```diff
@@ -13,15 +13,26 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@use '../../scss/layout' as layout;
 @use '../../scss/theme' as theme;
 
+$empty-state-height: 400px;
```
aqibmohammadkhan, line 18: is this needed as a local file variable?

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.html
kovaren: Fixed!

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.scss
kovaren: Yep, it's used in 2 places:
- https://github.com/gravitee-io/gravitee-api-management/pull/15711/changes#diff-a46f121f9e260c032781928a18c40fbdb527fc63bb88d782fba5ce64667ddfb1R96
- https://github.com/gravitee-io/gravitee-api-management/pull/15711/changes#diff-a46f121f9e260c032781928a18c40fbdb527fc63bb88d782fba5ce64667ddfb1R144

aqibmohammadkhan (approved):
LGTM

PR #15734, opened 2026-03-16
fix: persist env log filters

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/GKO-2601
## Description

Selected filters do not persist after viewing log details page and then navigating back to logs list view. 

Updated to behave just like api runtime logs:
- Write filters to URL query params when the user changes filters or pagination
- Read filters from query params when the component initializes (e.g., after navigating back from details)
- Preserve query params when navigating to/from the detail view

### DIFF
--- a/gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.ts
+++ b/gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.ts
@@ -18,14 +18,21 @@ import { DatePipe } from '@angular/common';
 import { HttpErrorResponse } from '@angular/common/http';
 import { Component, computed, inject, signal } from '@angular/core';
 import { toObservable, toSignal } from '@angular/core/rxjs-interop';
+import { ActivatedRoute, Router } from '@angular/router';
 import { MatCardModule } from '@angular/material/card';
 import { GioBannerModule } from '@gravitee/ui-particles-angular';
 import { EMPTY } from 'rxjs';
 import { catchError, debounceTime, switchMap, tap } from 'rxjs/operators';
+import moment from 'moment';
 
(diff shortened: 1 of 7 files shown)


gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.ts
```diff
@@ -83,10 +95,78 @@ export class EnvLogsComponent {
...
+    this.router.navigate(['.'], {
+      relativeTo: this.activatedRoute,
+      queryParams,
+      replaceUrl: true,
+    });
+  }
+
+  private parseQueryParams(): EnvLogsInitialValues | undefined {
+    const queryParams = this.activatedRoute.snapshot.queryParams;
+    const hasAnyFilter = Object.keys(queryParams).some(k => k !== 'apiId');
```
carlos-andres-osorio: The `hasAnyFilter` guard `Object.keys(queryParams).some(k => k !== 'apiId')` doesn't account for `perPage` being the only param present. When only `perPage` is in the URL (no filter keys), the guard passes and `parseQueryParams` runs — returning an `EnvLogsInitialValues` object where every field is `undefined`. That non-`undefined` return value causes `ngOnInit` in the filter bar to run and call `moreFiltersValues.set(...)` with all-default values: a spurious state mutation and an unnecessary render cycle.

Consider checking only for known filter keys (excluding pagination keys), or returning `undefined` from `parseQueryParams` when all fields would be `undefined`.

paulatulis: addressed!

carlos-andres-osorio (approved):

PR #15735, opened 2026-03-16
fix: ensure endpoint shared configuration is visible when inherited

bmouhaimen:
When an endpoint inherits configuration from a group, the GioFormJsonSchema 
internally resets the parent FormControl, resulting in an empty or greyed-out 
display in the UI even if the data exists in the API definition.

This fix:
- Restores the shared configuration display value after the schema component is ready.
- Ensures the form control state (disabled/enabled) correctly reflects the
  'inheritConfiguration' and 'isReadOnly' status without losing data visibility.
- Fixes the bug where custom headers at group level caused endpoint-specific 
  settings to appear empty or locked.
https://gravitee.atlassian.net/browse/APIM-12816

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/endpoints-v4/endpoint/api-endpoint.component.html
+++ b/gravitee-apim-console-webui/src/management/api/endpoints-v4/endpoint/api-endpoint.component.html
@@ -99,6 +99,7 @@
               <gio-form-json-schema
                 formControlName="sharedConfigurationOverride"
                 [jsonSchema]="endpointSchema?.sharedConfig"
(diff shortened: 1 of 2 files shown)


thomasballivet (commented):
I think this fix should be in `ui-particles-angular` because the problem is in the `gio-form-json-schema` component itself. This is a workaround that will work for the `api-endpoint` component but not for other components depending on `gio-form-json-schema`

araffass:
Closed since it's a duplicate.
Root fix already covered in ui-particles 17.7.3;
PR: https://github.com/gravitee-io/gravitee-api-management/pull/16347

PR #15736, opened 2026-03-16
fix: load error types based on period set in quick filter

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/GKO-2602
## Description

Make sure error type dropdown is populated based on the values set in quick filters. Otherwise, user has to set to/from date in more filters section, which is more, unnecessary work.

### DIFF
--- a/gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-filter-bar/components/env-logs-more-filters/env-logs-more-filters.component.ts
+++ b/gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-filter-bar/components/env-logs-more-filters/env-logs-more-filters.component.ts
@@ -34,6 +34,7 @@ import { DEFAULT_MORE_FILTERS, EnvLogsMoreFiltersForm } from '../../../../models
 import { DATE_TIME_FORMATS } from '../../../../../../../shared/utils/timeFrameRanges';
 import { HTTP_METHODS } from '../../../../../../../entities/management-api-v2';
 import { AnalyticsService } from '../../../../../../../services-ngx/analytics.service';
+import { periodToMs } from '../../../../../../../services-ngx/environment-logs.service';
 import { SnackBarService } from '../../../../../../../services-ngx/snack-bar.service';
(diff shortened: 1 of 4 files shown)


gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-filter-bar/components/env-logs-more-filters/env-logs-more-filters.component.ts
```diff
@@ -237,10 +239,22 @@ export class EnvLogsMoreFiltersComponent {
...
+        effectiveTo = Date.now();
+        effectiveFrom = effectiveTo - (ms ?? 24 * 60 * 60 * 1000);
+      }
+
       this.analyticsService
-        .getEnvironmentErrorKeys(from, to)
+        .getEnvironmentErrorKeys(effectiveFrom, effectiveTo)
         .pipe(
           catchError(() => of([])),
           takeUntilDestroyed(this.destroyRef),
```
carlos-andres-osorio, line 260: When the user clears either date field, this returns the current `errorKeysOptions` unchanged — no new fetch happens. The dropdown keeps showing keys from the previous time range instead of refreshing based on the active period.

Consider applying the same `periodToMs` fallback used in the `effect()` above:

```ts
if (from && to) {
  return this.analyticsService.getEnvironmentErrorKeys(from, to).pipe(catchError(() => of([])));
}
const ms = periodToMs(this.period());
const effectiveTo = Date.now();
const effectiveFrom = effectiveTo - (ms ?? 24 * 60 * 60 * 1000);
return this.analyticsService.getEnvironmentErrorKeys(effectiveFrom, effectiveTo).pipe(catchError(() => of([])));
```

paulatulis: Good call out, thank you! Addressed.

carlos-andres-osorio (approved):

PR #15737, opened 2026-03-16
fix(jdbc): change dictionary property value column type to nclob

ThibaudAV:
## Summary
- Change the `v` column type in `dictionary_property` from `nvarchar(1000)` to `nclob` to support larger property values
- `nclob` is suitable here since dictionary property values are not indexed

## Jira
https://gravitee.atlassian.net/browse/APIM-12355

## Test plan
- [ ] Verify Liquibase migration runs successfully on a fresh database
- [ ] Verify Liquibase migration runs successfully on an existing database with data
- [ ] Verify dictionary properties with values > 1000 characters can be stored and retrieved

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_9_5/01_change_dictionary_property_value_datatype.yml
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_9_5/01_change_dictionary_property_value_datatype.yml
@@ -1,0 +1,9 @@
+databaseChangeLog:
+  - changeSet:
+      id: 4.9.5_01_change_dictionary_property_value_datatype
+      author: GraviteeSource Team
+      changes:
+        - modifyDataType:
+            tableName: ${gravitee_prefix}dictionary_property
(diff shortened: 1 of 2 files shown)


gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_9_5/01_change_dictionary_property_value_datatype.yml
```diff
@@ -0,0 +1,9 @@
+databaseChangeLog:
+  - changeSet:
+      id: 4.9.5_01_change_dictionary_property_value_datatype
+      author: GraviteeSource Team
+      changes:
+        - modifyDataType:
+            tableName: ${gravitee_prefix}dictionary_property
+            columnName: v
+            newDataType: nclob
```
ThibaudAV, line 9: we can also use `nvarchar(4000)` or more But that's still a limitation. Here, you have unlimited choices—let me know if you think this is okay :)

phiz71:
Will it work for all kind of DBMS ?

aditya-goyal01 (approved):

PR #15744, opened 2026-03-17
GKO-1888 master

benoitgravitee:
## Issue

https://gravitee.atlassian.net/browse/GKO-1888

## Description

see https://github.com/gravitee-io/gravitee-api-management/pull/15701

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/helpers/HRIDHelper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/helpers/HRIDHelper.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.rest.api.automation.helpers;
+
+import java.security.SecureRandom;
+import java.util.Random;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * @author Benoit BORDIGONI (benoit.bordigoni at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class HRIDHelper {
+
+    static final String HRID_PREFIX = "hrid-";
+    private static final Random RANDOM = new SecureRandom();
+    private static final String NON_LETTERS_AND_DIGITS_START = "^[^a-zA-Z0-9]+";
+    private static final String NON_LETTERS_AND_DIGITS_END = "[^a-zA-Z0-9]+$";
+    private static final String NON_LETTERS_AND_DIGITS = "[^a-zA-Z0-9]+";
+    private static final Pattern HRID_ONLY_GROUP = Pattern.compile("-*([a-z0-9][a-z0-9-]+[a-z0-9])-*");
+    public static final String HEADER_X_GRAVITEE_SET_HRID = "X-Gravitee-Set-Hrid";
+
+    public static String nameToHRID(String name) {
+        if (name == null || name.isBlank()) {
+            return randomHRID();
+        }
+        String sanitized = name.replaceAll(NON_LETTERS_AND_DIGITS_START, "").toLowerCase();
+        sanitized = sanitized.replaceAll(NON_LETTERS_AND_DIGITS_END, "").toLowerCase();
+        sanitized = sanitized.replaceAll(NON_LETTERS_AND_DIGITS, "-").toLowerCase();
+        Matcher matcher = HRID_ONLY_GROUP.matcher(sanitized);
+        if (!matcher.matches()) {
+            return randomHRID();
+        }
+        return matcher.group(1);
+    }
+
+    private static @NotNull String randomHRID() {
+        return HRID_PREFIX + RANDOM.nextInt(1000000);
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/helpers/HRIDHelper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/helpers/HRIDHelper.java
@@ -20,6 +20,7 @@ import java.util.Random;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import lombok.AccessLevel;
+import lombok.CustomLog;
 import lombok.NoArgsConstructor;
 import org.jetbrains.annotations.NotNull;
 
@@ -28,6 +29,7 @@ import org.jetbrains.annotations.NotNull;
  * @author GraviteeSource Team
  */
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
+@CustomLog
 public class HRIDHelper {
 
     static final String HRID_PREFIX = "hrid-";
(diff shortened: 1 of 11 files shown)


carlos-andres-osorio (commented):
**NullPointerException when API has no pages or no plans**

`setPageParentAndGeneralConditionsHRIDs` calls `api.getPages().forEach(...)` and `api.getPlans().values().stream()` with no null guards. An API with no pages (common for draft/simple APIs) or no plans will throw a `NullPointerException` at runtime whenever `X-Gravitee-Set-Hrid: true` is sent.

Note that `formatHrids()` just above correctly guards both `getPages()` and `getPlans()` with `!= null` — the same pattern is missing here.

```java
private static void setPageParentAndGeneralConditionsHRIDs(ApiCRDSpec api) {
    Map<String, String> pageIDsToHrid = new HashMap<>();
    if (api.getPages() != null) {
        api.getPages().forEach((k, v) -> pageIDsToHrid.put(v.getId(), nameToHRID(v.getName())));
        api.getPages().values().stream()
            .filter(p -> p.getParentId() != null)
            .forEach(p -> p.setParentHrid(pageIDsToHrid.get(p.getParentId())));
    }
    if (api.getPlans() != null) {
        api.getPlans().values().stream()
            .filter(p -> p.getGeneralConditions() != null)
            .forEach(p -> p.setGeneralConditionsHrid(pageIDsToHrid.get(p.getGeneralConditions())));
    }
}
```

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/ApiResource.java
```diff
@@ -133,4 +166,30 @@ private static AuditInfo buildAuditInfo(ExecutionContext executionContext, UserD
...
+
+    private static void setPageParentAndGeneralConditionsHRIDs(ApiCRDSpec api) {
+        Map<String, String> pageIDsToHrid = new HashMap<>();
+        api.getPages().forEach((k, v) -> pageIDsToHrid.put(v.getId(), nameToHRID(v.getName())));
+        api
+            .getPages()
+            .values()
+            .stream()
+            .filter(p -> p.getParentId() != null)
+            .forEach(p -> p.setParentHrid(pageIDsToHrid.get(p.getParentId())));
```
carlos-andres-osorio: **Silent null assignment — can happen under normal use, not just corruption**

`pageIDsToHrid.get(p.getParentId())` returns `null` when the parent page ID is not in the map, and `setParentHrid(null)` is called silently. Same pattern on line 184 for `generalConditionsHrid`. No log, no warning — the caller gets a spec with a silently broken hierarchy.

This isn't just a corrupted-data edge case. Gravitee auto-creates an ASIDE `SYSTEM_FOLDER` page for every API. The legacy export service (`ApiImportExportServiceImpl`) queries it separately from regular pages, which suggests it may not be included in `api.getPages()` from the CRD export path. Any user page whose `parentId` points to that system folder will silently get `parentHrid = null`, flattening the page hierarchy in the exported spec with no indication anything was dropped.

**Fix**: log a warning on miss so the user has a signal:
```java
String hrid = pageIDsToHrid.get(p.getParentId());
if (hrid == null) {
    log.warn("Could not resolve HRID for parentId={}", p.getParentId());
}
p.setParentHrid(hrid);
```

carlos-andres-osorio:
**HRIDHelper.java:38-43** — `nameToHRID` silently falls back to a random `hrid-XXXXXX` when the name doesn't match the pattern (e.g. non-ASCII names like `"日本語ページ"`). The caller cannot distinguish a derived HRID from a randomly generated one. Re-exporting the same API twice will produce different HRIDs, **breaking idempotency** with no log or indication to the user.

**Fix**: Add a debug/warn log before returning `randomHRID()`.

carlos-andres-osorio:
**Missing test coverage (3 gaps)**

**ApiResourceTest.java** — No test for `setHRID=true` with an API that has `null` pages or `null` plans. This is the direct test path for the NPE in `setPageParentAndGeneralConditionsHRIDs`.

**ApiResourceTest.java** — No test for `setHRID=true` where the API already has a non-UUID hrid. The `isUUID` guard (preserve existing hrid branch) is only exercised via the null branch in the existing test.

**SharedPolicyGroupIdHelper** — `addHRID` has no dedicated unit tests. `ApiResourceTest` covers it indirectly but only for the `request` phase with a single step. The phases `response`, `subscribe`, `publish`, `interact`, and `entrypointConnect` are untested, as is the `computeIfAbsent` guard for the already-set case.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/ApiResource.java
```diff
@@ -133,4 +166,30 @@ private static AuditInfo buildAuditInfo(ExecutionContext executionContext, UserD
             )
             .build();
     }
+
+    private static void setPageParentAndGeneralConditionsHRIDs(ApiCRDSpec api) {
+        Map<String, String> pageIDsToHrid = new HashMap<>();
+        api.getPages().forEach((k, v) -> pageIDsToHrid.put(v.getId(), nameToHRID(v.getName())));
```
carlos-andres-osorio: `pageIDsToHrid.put(v.getId(), nameToHRID(v.getName()))` inserts a `null` key when `v.getId()` is null.

`ApiCRDExportDomainServiceImpl` explicitly calls `page.setId(null)` on every page for the `HRID` and `NONE` export strategies before returning the spec, so `v.getId()` is `null` for every page when those strategies are in use.

`HashMap` permits null keys, so the put succeeds silently. Any page/plan whose `parentId` also happens to be `null` (and isn't filtered out) would spuriously match the null-keyed entry and receive a wrong HRID.

**Fix**: guard the put with `if (v.getId() != null)`:
```java
api.getPages().forEach((k, v) -> {
    if (v.getId() != null) pageIDsToHrid.put(v.getId(), nameToHRID(v.getName()));
});
```

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/helpers/HRIDHelper.java
```diff
@@ -0,0 +1,60 @@
...
+
+    static final String HRID_PREFIX = "hrid-";
+    private static final Random RANDOM = new SecureRandom();
+    private static final String NON_LETTERS_AND_DIGITS_START = "^[^a-zA-Z0-9]+";
+    private static final String NON_LETTERS_AND_DIGITS_END = "[^a-zA-Z0-9]+$";
+    private static final String NON_LETTERS_AND_DIGITS = "[^a-zA-Z0-9]+";
+    private static final Pattern HRID_ONLY_GROUP = Pattern.compile("([a-z0-9](?:[a-z0-9-]*[a-z0-9])?)");
+    public static final String HEADER_X_GRAVITEE_SET_HRID = "X-Gravitee-Set-Hrid";
+
+    public static String nameToHRID(String name) {
```
carlos-andres-osorio, line 43: **HRIDHelper.java:38-43** — `nameToHRID` silently falls back to a random `hrid-XXXXXX` when the name doesn't match the pattern (e.g. non-ASCII names like `"日本語ページ"`). The caller cannot distinguish a derived HRID from a randomly generated one. Re-exporting the same API twice will produce different HRIDs, **breaking idempotency** with no log or indication to the user.

**Fix**: Add a debug/warn log before returning `randomHRID()`.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/ApiResource.java
carlos-andres-osorio: **NullPointerException when API has no pages or no plans**

`setPageParentAndGeneralConditionsHRIDs` calls `api.getPages().forEach(...)` and `api.getPlans().values().stream()` with no null guards. An API with no pages (common for draft/simple APIs) or no plans will throw a `NullPointerException` at runtime whenever `X-Gravitee-Set-Hrid: true` is sent.

Note that `formatHrids()` just above correctly guards both `getPages()` and `getPlans()` with `!= null` — the same pattern is missing here.

```java
private static void setPageParentAndGeneralConditionsHRIDs(ApiCRDSpec api) {
    Map<String, String> pageIDsToHrid = new HashMap<>();
    if (api.getPages() != null) {
        api.getPages().forEach((k, v) -> pageIDsToHrid.put(v.getId(), nameToHRID(v.getName())));
        api.getPages().values().stream()
            .filter(p -> p.getParentId() != null)
            .forEach(p -> p.setParentHrid(pageIDsToHrid.get(p.getParentId())));
    }
    if (api.getPlans() != null) {
        api.getPlans().values().stream()
            .filter(p -> p.getGeneralConditions() != null)
            .forEach(p -> p.setGeneralConditionsHrid(pageIDsToHrid.get(p.getGeneralConditions())));
    }
}
```

benoitgravitee: This is irrelevant. The NONE strategy is not used the context of AutomationAPI

carlos-andres-osorio (approved):

PR #15752, opened 2026-03-17
feat: add search endpoint for portal navigation APIs with include support

dev-marek:
## Issue

https://gravitee.atlassian.net/browse/APIM-13064

## Description

Adds a /_search endpoint for portal navigation items and refactors the surrounding domain layer:

- PortalNavigationItemsResource: New GET /_search?type=api&query=...&include=api endpoint backed by GetVisiblePortalNavigationApisUseCase. Fixed JAX-RS enum conversion for include param by accepting Set<String> and delegating to PortalNavigationSearchInclude.fromValue().
- ApiPortalSearchQueryService: Introduces a Query record as the primary abstraction (search(Query) is now the single abstract method). Two default List<Api> convenience overloads eliminate PageableImpl(1, MAX_VALUE) boilerplate at call sites.
- PortalNavigationApiVisibilityDomainService: resolveVisibleItems and isVisibleToUser are split into two non-nullable overloads (with/without userId), removing @Nullable and an intermediate private stream method.
- GetVisiblePortalNavigationApisUseCase: resolveIncludedApis extracted from execute, replacing multiple early returns with a focused private helper.
- PortalNavigationItemCriteria: query field removed — text search is handled by Elasticsearch/Lucene, not the repository layer. Corresponding blocks removed from MongoDB, JDBC, and in-memory implementations.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/PortalNavigationItemsResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/PortalNavigationItemsResource.java
@@ -17,15 +17,24 @@ package io.gravitee.rest.api.portal.rest.resource;
 
 import static io.gravitee.rest.api.service.common.GraviteeContext.getExecutionContext;
 
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItem;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemViewerContext;
+import io.gravitee.apim.core.portal_page.use_case.GetVisiblePortalNavigationApisUseCase;
 import io.gravitee.apim.core.portal_page.use_case.ListPortalNavigationItemsUseCase;
 import io.gravitee.common.http.MediaType;
+import io.gravitee.rest.api.model.common.PageableImpl;
+import io.gravitee.rest.api.portal.rest.mapper.ApiMapper;
 import io.gravitee.rest.api.portal.rest.mapper.PortalNavigationItemMapper;
+import io.gravitee.rest.api.portal.rest.model.ErrorResponse;
 import io.gravitee.rest.api.portal.rest.model.PortalArea;
+import io.gravitee.rest.api.portal.rest.model.PortalNavigationItemsSearchResponse;
+import io.gravitee.rest.api.portal.rest.model.PortalNavigationSearchInclude;
+import io.gravitee.rest.api.portal.rest.resource.param.PaginationParam;
 import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
 import jakarta.inject.Inject;
+import jakarta.ws.rs.BeanParam;
 import jakarta.ws.rs.DefaultValue;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.Path;
@@ -34,7 +43,12 @@ import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.container.ResourceContext;
 import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.Response;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 public class PortalNavigationItemsResource extends AbstractResource {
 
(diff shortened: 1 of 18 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
```diff
@@ -3196,6 +3196,64 @@ paths:
...
+                - name: query
+                  in: query
+                  required: false
+                  schema:
+                      type: string
+                - name: type
+                  in: query
+                  required: true
+                  schema:
+                      type: string
```
kovaren, line 3219: This param can only have a value of `api`, not an arbitrary string.

WDYT about:
- typing it with an enum
- removing this check in the resource?
```
if (!"api".equalsIgnoreCase(type)) {
    return Response.status(501).entity(new ErrorResponse()).build();
}
```

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
```diff
@@ -3844,6 +3902,23 @@ components:
         #####################
         # Responses Objects #
         #####################
+        PortalNavigationSearchInclude:
+            type: string
+            enum:
+                - api
+        PortalNavigationItemsSearchResponse:
```
kovaren, line 3901: The response type lacks the pagination data which is needed by the front end for the pagination in the catalog.

Could you please add the pagination, following the example of old API response in the catalog?

```
"metadata": {
    "pagination": {
        "total": 191,
        "size": 20,
        "last": 20,
        "total_pages": 10,
        "current_page": 1,
        "first": 1
    },
    "data": {
        "total": 20
    }
}
```

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
```diff
@@ -3844,6 +3902,23 @@ components:
         #####################
         # Responses Objects #
         #####################
+        PortalNavigationSearchInclude:
+            type: string
+            enum:
+                - api
```
kovaren: This location in the schema is dedicated to responses objects as indicated in the comment above. WDYT about moving this somewhere lower below?

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/PortalNavigationItemsResource.java
```diff
@@ -65,6 +83,84 @@ public Response getPortalNavigationItems(
...
+        @QueryParam("type") String type,
+        @QueryParam("include") Set<String> include,
+        @BeanParam PaginationParam paginationParam
+    ) {
+        if (type == null || type.isBlank()) {
+            return Response.status(Response.Status.BAD_REQUEST).entity(new ErrorResponse()).build();
+        }
+        if (!"api".equalsIgnoreCase(type)) {
+            return Response.status(501).entity(new ErrorResponse()).build();
+        }
```
kovaren, line 99: As per https://github.com/gravitee-io/gravitee-api-management/pull/15752/changes#r2948172146, this can be handled by the spec validation.
```suggestion

**Review by kovaren — CHANGES_REQUESTED** [review]

File: gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml (hunk 1 above)
Comment (dev-marek) [reply gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml:3219]: We can do it on spec level, but it won't generate java enum to use as query param type

File: gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/query_service/ApiPortalSearchQueryService.java [hunk 5]
```diff
@@ -20,15 +20,55 @@
...
+
+    default List<Api> search(String environmentId, String organizationId, String query, Set<String> allowedApiIds) {
+        return search(new Query(environmentId, organizationId, query, allowedApiIds)).getContent();
+    }
+
+    default List<Api> search(String environmentId, String organizationId, Set<String> allowedApiIds) {
+        return search(new Query(environmentId, organizationId, allowedApiIds)).getContent();
+    }
+
+    record Query(
```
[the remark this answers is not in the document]
dev-marek: no need: this class is used only in context of this service

kovaren (approved):

PR #15763, opened 2026-03-17
feat: add isApiPrimaryOwner to user groups endpoint

ThibaudAV:
## Summary
- Add `isApiPrimaryOwner` boolean field to the `GET /users/{userId}/groups` response, computed server-side by checking if the group holds a PRIMARY_OWNER membership on any API
- Remove the expensive frontend fetch of up to 10,000 APIs (`apiV2Service.search({}, undefined, 1, 10000)`) that was only used to build a `Set<string>` of group IDs associated with APIs
- Use the new backend field in the memberships component to disable the API role select when appropriate

## Test plan
- [ ] `GetUserGroupsUseCaseTest` — new test verifies `isApiPrimaryOwner` is `true` for a group that is PRIMARY_OWNER of an API, and `false` otherwise
- [ ] `org-settings-user-detail.component.spec.ts` — updated to remove the API search mock
- [ ] Manual: edit group roles for a user who is PRIMARY_OWNER — the API role select should still be disabled

https://github.com/user-attachments/assets/11bda20c-abe5-4ede-8d07-012250b611b9


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/UserApiEntity.java
```diff
@@ -0,0 +1,35 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@Data
+public class UserApiEntity {
```
phiz71: Are we sure about this "Entity" ?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/GetUserApisUseCase.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private Map<String, String> resolveEnvironmentNames(Set<String> environmentIds) {
+        Map<String, String> result = new HashMap<>();
+        for (String envId : environmentIds) {
+            try {
+                result.put(envId, environmentCrudService.get(envId).getName());
+            } catch (Exception e) {
+                // Skip environments that cannot be resolved
+            }
+        }
```
phiz71, line 97: Maybe we can use a Map to cache environments, or do a single call?

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/spring/ResourceContextConfiguration.java
```diff
@@ -862,4 +863,9 @@ public PortalPageQueryService portalPageQueryService() {
     public GetUserApisUseCase getUserApisUseCase() {
         return mock(GetUserApisUseCase.class);
     }
+
+    @Bean
+    public GetUserApplicationsUseCase getUserApplicationsUseCase() {
+        return mock(GetUserApplicationsUseCase.class);
+    }
```
phiz71, line 871: should we use inmemory instance instead of a mock ?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/application/ApplicationQueryServiceImpl.java
```diff
@@ -41,6 +44,23 @@ public ApplicationQueryServiceImpl(@Lazy ApplicationRepository applicationReposi
...
+            if (environmentId != null) {
+                criteriaBuilder.environmentIds(Set.of(environmentId));
+            }
+            var repoPageable = convert(pageable);
+            var page = applicationRepository.search(criteriaBuilder.build(), repoPageable);
+            var content = page.getContent().stream().map(ApplicationAdapter.INSTANCE::toEntity).toList();
+            return new Page<>(content, page.getPageNumber(), (int) page.getPageElements(), page.getTotalElements());
+        } catch (TechnicalException e) {
+            log.error("An error occurred while searching applications by ids", e);
+            throw new TechnicalManagementException("An error occurred while searching applications by ids", e);
```
phiz71: log or throw, but not both

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/spring/ResourceContextConfiguration.java
ThibaudAV: When testing resources, we often mock the call to the use case and cover the use case with in-memory instance of the repositories and services.
But we can also test the resource using in-memory versions. This feels a bit redundant 🤷‍♂️

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/GetUserApisUseCase.java
ThibaudAV: Since this is a set, we make only one call per environemntId 💭 

I left the try-catch block here. Usually, you don't include it in the code, but I think it looks better with it.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/UserApiEntity.java
ThibaudAV: The other model that already exists here is called `BaseUserEntity` 
So I wasn't surprised when Claude kept going with :
image
I can remove it (say yes and i do it). I have to admit I'm not really sure why it's here 🤷‍♂️

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/application/ApplicationQueryServiceImpl.java
ThibaudAV: Good point. They do it in lots of other places. that's probably why agent dot it.
I'd be careful. We're keeping the throw.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/GetUserGroupsUseCase.java
```diff
@@ -0,0 +1,156 @@
...
+
+        return result;
+    }
+
+    private Map<String, String> resolveEnvironmentNames(Set<String> environmentIds) {
+        Map<String, String> result = new HashMap<>();
+        for (String envId : environmentIds) {
+            try {
+                result.put(envId, environmentCrudService.get(envId).getName());
+            } catch (Exception e) {
```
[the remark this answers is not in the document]
ThibaudAV: Yeah, but we're trying to find the name of the environment, and if we don't have it, it's not a big deal. I think we'd rather not see an error, which is why this is a best-effort approach.
Honestly, it's not going to happen. Removing the catch is fine with me too 🤷‍♂️

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/GetUserApisUseCase.java
```diff
@@ -0,0 +1,104 @@
...
+            })
+            .toList();
+
+        return new Output(data, apiPage.getTotalElements());
+    }
+
+    private Map<String, String> resolveEnvironmentNames(Set<String> environmentIds) {
+        Map<String, String> result = new HashMap<>();
+        for (String envId : environmentIds) {
+            try {
```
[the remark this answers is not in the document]
ThibaudAV: lol same catch issue

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/GetUserApplicationsUseCase.java
```diff
@@ -0,0 +1,97 @@
...
+            })
+            .toList();
+
+        return new Output(data, appPage.getTotalElements());
+    }
+
+    private Map<String, String> resolveEnvironmentNames(Set<String> environmentIds) {
+        Map<String, String> result = new HashMap<>();
+        for (String envId : environmentIds) {
+            try {
```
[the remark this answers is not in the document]
ThibaudAV: lol same catch issue

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/GetUserGroupsUseCase.java
```diff
@@ -0,0 +1,156 @@
...
+
+        return new Output(data, groupPage.getTotalElements());
+    }
+
+    private Map<String, Boolean> computeApiPrimaryOwner(List<Group> groups) {
+        Map<String, Boolean> result = new HashMap<>();
+        Map<String, List<Membership>> apiMembershipsByGroupId = new HashMap<>();
+        Set<String> apiMembershipRoleIds = new HashSet<>();
+
+        for (Group group : groups) {
```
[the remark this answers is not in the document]
ThibaudAV: Well, we're after pagination, so we have one call per item on the page, which doesn't really worry me 🤷‍♂️ 
I can figure out how to do it in a batch if a human thinks it's necessary 😁

gravitee-apim-console-webui/src/organization/configuration/user/detail/org-settings-user-detail.component.ts
```diff
@@ -211,46 +151,8 @@ export class OrgSettingsUserDetailComponent implements OnInit, OnDestroy {
...
-      .subscribe(({ metadata }) => {
-        this.applicationsTableDS = Object.entries(metadata).map(([applicationId, applicationMetadata]: [string, any]) => ({
-          id: applicationId,
-          name: applicationMetadata.name,
-          environmentId: applicationMetadata.environmentId,
-        }));
-        this.initialTableDS['applicationsTableDS'] = this.applicationsTableDS;
-        this.tablesUnpaginatedLength['applicationsTableDS'] = this.applicationsTableDS.length;
+        // Pass data to memberships sub-component
+        this.membershipEnvironments = environments.map((e) => ({ id: e.id, name: e.name }));
```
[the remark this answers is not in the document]
ThibaudAV: lol removed

phiz71 (approved):

PR #15765, opened 2026-03-17
feat: add gmd-checkbox-group multiselect form component

JedrzejJanasiak:
## Issue

## Description
Adds a new `gmd-checkbox-group` GMD form component that allows users to select multiple options
from a list of checkboxes. Modeled structurally after `gmd-radio` — same `<fieldset>`/`<legend>`
layout, same comma-separated options parsing, same `GmdFormFieldBase` extension.

The selected values are serialized as a comma-separated string.

**This PR includes:**
- `gmd-checkbox-group` component (ts, html, scss, `_overrides.scss`, `public-api.scss`)
- Component harness (`GmdCheckboxGroupComponentHarness`)
- Unit tests covering: options parsing, multi-toggle behavior, preselection, required validation,
  comma-separated value serialization, disabled/readonly states, ARIA attributes, store updates
- Storybook stories: Basic, Required, Disabled, Readonly, Pre-selected values, Many options,
  WithFieldKey, WithFormEditor, InMarkdownViewer
- Monaco editor suggestions for the new component
- README with props table, value serialization, and theming tokens
- Registration in all 5 component registries:
  `componentSelector.ts`, `prefix-stripper.parser.ts`, `component-name-selectors.ts`,
  `component-attribute-selectors.ts`, `suggestions.index.ts`

## Additional context

https://github.com/user-attachments/assets/9f6f880b-47f8-40a9-bbfb-1f2fdddbae55

### DIFF
--- a/gravitee-apim-webui-libs/gravitee-markdown/src/lib/components/checkbox-group/README.md
+++ b/gravitee-apim-webui-libs/gravitee-markdown/src/lib/components/checkbox-group/README.md
@@ -1,0 +1,54 @@
+# `gmd-checkbox-group`
+
+A checkbox group component that allows users to select multiple options from a list. Structurally modeled after `gmd-radio` — same fieldset/legend layout, same `GmdFormFieldBase` extension. Options are parsed from a comma-separated string. The selected value is serialized as a sorted, comma-separated string.
+
+## Usage
+
+```html
+<gmd-checkbox-group
+  name="features"
+  label="Required Features"
+  fieldKey="features"
+  required="true"
+  options="Authentication,Rate Limiting,Analytics,Caching"
+>
+</gmd-checkbox-group>
+```
+
+## Props
+
+| Attribute  | Type    | Default | Description                                                            |
+| ---------- | ------- | ------- | ---------------------------------------------------------------------- |
+| `name`     | string  | `''`    | HTML name attribute for the checkbox inputs                            |
+| `label`    | string  | —       | Display label shown above the group (renders as `<legend>`)            |
+| `fieldKey` | string  | —       | Key used for form state tracking and data collection                   |
+| `value`    | string  | —       | Comma-separated preselected values (e.g., `"Option 1,Option 3"`)       |
+| `required` | boolean | `false` | Whether at least one option must be selected                           |
+| `options`  | string  | `''`    | Comma-separated list of options (e.g., `"Option 1,Option 2,Option 3"`) |
+| `disabled` | boolean | `false` | Disables all checkboxes and removes field from form state              |
+
+## Value serialization
+
+The field state `value` is serialized as a comma-separated string of selected items:
+
+- Single selection: `"Option 1"`
+- Multiple selections: `"Option 1,Option 3"`
+- Nothing selected: `""`
+
+## Validation
+
+- `required`: Reports a `'required'` error when no option is selected. Error message is shown after the field is touched (blurred).
+
+## Theming tokens
+
+The component exposes CSS custom properties under the `checkbox-group` namespace:
+
+| Token                                             | Default             | Description             |
+| ------------------------------------------------- | ------------------- | ----------------------- |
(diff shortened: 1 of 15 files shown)


gravitee-apim-webui-libs/gravitee-markdown/src/lib/components/checkbox-group/gmd-checkbox-group.component.html
```diff
@@ -0,0 +1,59 @@
...
+  }
+  <div class="gmd-checkbox-group__options">
+    @for (option of optionsVM(); track option) {
+      <label class="gmd-checkbox-group__option">
+        <input
+          type="checkbox"
+          [name]="name()"
+          [value]="option"
+          [checked]="internalValues().has(option)"
+          [attr.required]="required() ? 'true' : null"
```
kovaren: Are all the checkboxes within the group marked as required if the group itself is required?

gravitee-apim-webui-libs/gravitee-markdown/src/lib/components/checkbox-group/README.md
```diff
@@ -0,0 +1,55 @@
+# `gmd-checkbox-group`
+
+A checkbox group component that allows users to select multiple options from a list. Structurally modeled after `gmd-radio` — same fieldset/legend layout, same options parsing, same `GmdFormFieldBase` extension. The selected value is serialized as a comma-separated string.
```
kovaren: Let's mention that the options parsing is not the same since the radio buttons also support JSON not just CSV.

gravitee-apim-webui-libs/gravitee-markdown/src/lib/components/checkbox-group/gmd-checkbox-group.component.spec.ts
```diff
@@ -0,0 +1,391 @@
...
+      fixture.detectChanges();
+
+      await harness.toggle('option1');
+      await harness.toggle('option3');
+      fixture.detectChanges();
+
+      expect(mockStore.updateField).toHaveBeenLastCalledWith(
+        expect.objectContaining({
+          fieldKey: 'test-key',
+          value: expect.stringMatching(/option1.*option3|option3.*option1/),
```
kovaren: Minor suggestion: ideally we should sort the values to make sure that their order is deterministic, regardless of the order in which the values are selected.

gravitee-apim-webui-libs/gravitee-markdown/src/lib/components/checkbox-group/gmd-checkbox-group.component.stories.ts
```diff
@@ -0,0 +1,222 @@
...
+
+export const Readonly: StoryObj<GmdCheckboxGroupComponent> = {
+  render: () => ({
+    template: `
+      <div style="width: 400px;">
+        <gmd-checkbox-group name="readonly" label="Readonly Checkbox Group" readonly="true" value="Option 2" options="Option 1,Option 2,Option 3"></gmd-checkbox-group>
+      </div>
+    `,
+  }),
+};
```
kovaren: The `readonly` attribute on checkboxes is ignored by the browsers so this story is not really useful. Instead `disabled` is used to achieve read only behavior.

Ideally we shouldn't expose `readonly` attribute at all.

kovaren (changes requested):

gravitee-apim-webui-libs/gravitee-markdown/src/lib/components/checkbox-group/gmd-checkbox-group.component.html
JedrzejJanasiak: good catch, doesn't make senes. Remove this attribue for single checboxes. 
We validate the required correctly here: 
```
  validationErrors = computed<GmdFieldErrorCode[]>(() => {
    const errs: GmdFieldErrorCode[] = [];

    if (this.required() && this.internalValues().size === 0) errs.push('required');

    return errs;
  });
```

kovaren (approved):

PR #15766, opened 2026-03-17
fix: Application menu not showing up on first login

bmouhaimen:
Problem
When a new user logs in for the first time via an OIDC/social identity provider (e.g. Keycloak), the Applications tab is missing from the developer portal navigation. It only appears after a manual page refresh.
Root Cause
In createNewExternalUser(), the user was created with addDefaultRole=false, meaning no default role was assigned at creation time. In 4.8.x, a request-scoped cache captured the user before roles were assigned by refreshUserMemberships(), causing GET /user to return empty permissions on first login.
Fix
Default roles are assigned before the user is cached.

https://gravitee.atlassian.net/browse/APIM-11810

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/MembershipServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/MembershipServiceImpl.java
@@ -318,6 +318,12 @@ public class MembershipServiceImpl extends AbstractService implements Membership
                 membership.setCreatedAt(updateDate);
                 membership.setUpdatedAt(updateDate);
                 membershipRepository.create(membership);
+                invalidateRoleCache(
+                    membership.getReferenceType().name(),
+                    membership.getReferenceId(),
+                    membership.getMemberType().name(),
+                    membership.getMemberId()
+                );
                 createAuditLog(executionContext, MEMBERSHIP_CREATED, membership.getCreatedAt(), null, membership);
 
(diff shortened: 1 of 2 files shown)


wbabyte (commented):
No test coverage, please add them 

not sure how exactly it works now, but when using OIDC login you can have custom roles from user info (this can be configured with OIDC) 
maybe the issue here is more that creation is done async? Can you check that? I can see in jira ticket recording and roles are accessible after refresh which would indicate that roles are created async and this change would introduce regression in that case

bmouhaimen:
Role assignment via addRoleToMemberOnReference and refreshUserMemberships is fully synchronous — no @Async, no CompletableFuture, no thread executors involved in either UserServiceImpl or MembershipServiceImpl. The issue is not async creation.
The actual root cause is the request-scoped cache GraviteeContext.getCurrentUsers() in 4.8.x. When a new user logs in for the first time via OIDC, createNewExternalUser creates the user with addDefaultRole=false and returns a user object that gets captured in the cache. refreshUserMemberships then assigns roles to the DB, but the subsequent GET /user call hits the cache and returns the stale user with empty permissions. On refresh, the cache is cleared, the DB is hit fresh, and permissions appear correctly. 
Three tests have been added :

shouldAssignDefaultRoleOnFirstLoginWhenNoRoleMapping — verifies addRoleToMemberOnReference is called when default roles exist and no role mapping is configured
shouldNotAssignDefaultRoleOnFirstLoginWhenRoleMappingExists — verifies default roles are NOT assigned at creation when role mapping is configured
shouldNotFailWhenNoDefaultRolesAndNoRoleMapping — verifies graceful fallback when neither default roles nor role mapping exist

wbabyte (approved):

PR #15785, opened 2026-03-18
feat: add subscription form schema model, GMD parser, and submission validator

JedrzejJanasiak:
## Issue

## Description
**PR 1/3** — Domain layer for subscription form backend validation.

This PR adds:
- **`SubscriptionFormSchema` / `SubscriptionFormFieldSchema`** records — flat model capturing only validation-relevant attributes (`required`, `minLength`, `maxLength`, `pattern`, `options`). UI-only fields (`label`, `placeholder`, `readonly`, `disabled`) are intentionally excluded so cosmetic edits won't create new form versions in the future versioning system.
- **`SubscriptionFormSchemaGenerator`** interface (core) + **`SubscriptionFormSchemaGeneratorImpl`** (infra, Jsoup) — parses GMD HTML and extracts per-field validation rules.
- **`SubscriptionFormSubmissionValidator`** — collects all field errors in a single pass then throws `SubscriptionFormValidationException`; handles `required`, `options` (including `checkbox-group` multi-value comma-split), `minLength`, `maxLength`, and `pattern`.

**Not in this PR** (coming in subsequent PRs):
- `formSchema` column/field on `SubscriptionForm` + DB migrations (PR 2/3)
- Use-case wiring, REST resource changes, and integration tests (PR 3/3)

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormConstraintsFactory.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormConstraintsFactory.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.subscription_form.domain_service;
+
+import io.gravitee.apim.core.subscription_form.model.Constraint;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormFieldConstraints;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.CheckboxField;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.CheckboxGroupField;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.Field;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.MaxLengthAttribute;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.MinLengthAttribute;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.OptionsAttribute;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.PatternAttribute;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.ReadOnlyValueAttribute;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Builds {@link SubscriptionFormFieldConstraints} from a parsed {@link SubscriptionFormSchema}.
+ *
+ * @author Gravitee.io Team
+ */
+public final class SubscriptionFormConstraintsFactory {
+
+    private SubscriptionFormConstraintsFactory() {}
+
+    public static SubscriptionFormFieldConstraints fromSchema(SubscriptionFormSchema schema) {
+        if (schema == null || schema.isEmpty()) {
+            return SubscriptionFormFieldConstraints.empty();
+        }
+        Map<String, List<Constraint>> map = new LinkedHashMap<>();
+        for (Field field : schema.fields()) {
+            map.put(field.fieldKey(), List.copyOf(constraintsFor(field)));
+        }
+        return new SubscriptionFormFieldConstraints(map);
+    }
+
+    private static List<Constraint> constraintsFor(Field field) {
+        if (field instanceof ReadOnlyValueAttribute ro && ro.readonlyValue() != null) {
+            return List.of(new Constraint.ReadOnly(ro.readonlyValue()));
+        }
+
+        var out = new ArrayList<Constraint>();
+
+        if (field.required()) {
+            if (field instanceof CheckboxField) {
+                out.add(new Constraint.MustBeTrue());
+            } else if (field instanceof CheckboxGroupField) {
+                out.add(new Constraint.NonEmptySelection());
+            } else {
+                out.add(new Constraint.Required());
+            }
+        }
+
+        if (field instanceof MinLengthAttribute min && min.minLength() != null) {
+            out.add(new Constraint.MinLength(min.minLength()));
+        }
+        if (field instanceof MaxLengthAttribute max && max.maxLength() != null) {
+            out.add(new Constraint.MaxLength(max.maxLength()));
+        }
+        if (field instanceof PatternAttribute pat && pat.pattern() != null) {
+            out.add(new Constraint.MatchesPattern(pat.pattern()));
+        }
+
+        if (field instanceof OptionsAttribute opt && hasOptions(opt.options())) {
+            if (field instanceof CheckboxGroupField) {
+                out.add(new Constraint.EachOf(opt.options()));
+            } else {
+                out.add(new Constraint.OneOf(opt.options()));
+            }
+        }
+
+        return out;
+    }
+
+    private static boolean hasOptions(List<String> options) {
+        return options != null && !options.isEmpty();
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSchemaGenerator.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSchemaGenerator.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.subscription_form.domain_service;
+
+import io.gravitee.apim.core.gravitee_markdown.GraviteeMarkdown;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema;
+
+/**
+ * Parses GMD (Gravitee Markdown) content and extracts a {@link SubscriptionFormSchema}
+ * capturing the validation rules for each form field.
+ *
+ * <p>The implementation lives in the infrastructure layer because it depends on an HTML
+ * parsing library (Jsoup) to traverse the GMD element tree.</p>
+ *
+ * @author Gravitee.io Team
+ */
+public interface SubscriptionFormSchemaGenerator {
+    /**
+     * Parses the given GMD content and returns a schema describing all form fields
+     * and their validation constraints.
+     *
+     * @param gmdContent GMD value object whose raw content will be parsed
+     * @return a {@link SubscriptionFormSchema} with zero or more fields; never null
+     */
+    SubscriptionFormSchema generate(GraviteeMarkdown gmdContent);
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidator.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidator.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.subscription_form.domain_service;
+
+import io.gravitee.apim.core.subscription_form.exception.SubscriptionFormValidationException;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormFieldConstraints;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Validates a subscription form submission using {@link SubscriptionFormFieldConstraints}.
+ *
+ * <p>Prefer {@link #SubscriptionFormSubmissionValidator(SubscriptionFormFieldConstraints)} when constraints are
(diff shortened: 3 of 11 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidator.java
```diff
@@ -0,0 +1,135 @@
...
+public class SubscriptionFormSubmissionValidator {
+
+    /**
+     * Validates the submitted key-value pairs against the given schema.
+     *
+     * @param schema          the form schema (may be null — validation is skipped)
+     * @param submittedValues field values keyed by fieldKey
+     * @throws SubscriptionFormValidationException with all collected field errors when the submission is invalid
+     */
+    public void validate(SubscriptionFormSchema schema, Map<String, String> submittedValues) {
```
kovaren: Should we also check that `submittedValues` contains no values that are not part of the schema?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/SubscriptionFormFieldSchema.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package io.gravitee.apim.core.subscription_form.model;
+
+import java.util.List;
+
+/**
+ * Validation schema for a single field in a subscription form.
+ *
+ * <p>Only attributes relevant to backend validation are captured. UI-only attributes
+ * (label, placeholder, readonly, disabled, rows) are intentionally excluded.</p>
```
kovaren: I'm not sure we can skip the `readonly` fields.

Their values are included in form submissions, but if a request is sent directly to the backend bypassing the browser, those values can be modified.

If the backend does not validate them, it opens the door for changing their values despite them being read-only in the form.

It's a bit of an edge case but still.

kovaren (changes requested):

kovaren:
@dev-marek /review

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSchemaGenerator.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public interface SubscriptionFormSchemaGenerator {
+    /**
+     * Parses the given GMD content and returns a schema describing all form fields
+     * and their validation constraints.
+     *
+     * @param gmdContent raw GMD string (may be null or empty)
+     * @return a {@link SubscriptionFormSchema} with zero or more fields; never null
+     */
+    SubscriptionFormSchema generate(String gmdContent);
```
dev-marek: Please use `GraviteeMarkdown` and assume not null

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidator.java
```diff
@@ -0,0 +1,135 @@
...
+ * is handled externally (e.g. via {@code @Bean} in a Spring configuration class).</p>
+ *
+ * @author Gravitee.io Team
+ */
+public class SubscriptionFormSubmissionValidator {
+
+    /**
+     * Validates the submitted key-value pairs against the given schema.
+     *
+     * @param schema          the form schema (may be null — validation is skipped)
```
dev-marek: schema should not be nullable, but think about adding `isEmpty` method to the schema

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidator.java
```diff
@@ -0,0 +1,135 @@
...
+    public void validate(SubscriptionFormSchema schema, Map<String, String> submittedValues) {
+        if (schema == null || schema.fields() == null) {
+            return;
+        }
+
+        List<String> errors = new ArrayList<>();
+
+        for (SubscriptionFormFieldSchema field : schema.fields()) {
+            String fieldKey = field.fieldKey();
+            String value = submittedValues != null ? submittedValues.get(fieldKey) : null;
```
dev-marek: make `submittedValues` not null

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidator.java
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * <p>If {@code schema} is null (forms saved before schema generation was introduced),
+ * validation is skipped for backward compatibility.</p>
+ *
+ * <p>This is a pure domain service with no framework dependencies. Bean registration
+ * is handled externally (e.g. via {@code @Bean} in a Spring configuration class).</p>
+ *
+ * @author Gravitee.io Team
+ */
+public class SubscriptionFormSubmissionValidator {
```
dev-marek, line 36: Current setup gives `ExecuteFormValidationDomainService` vibes.

From validator I'd expect to have a constructor (from schema) setting up field validators first.

After constructing validator from schema we can think of persisting validator instead of schema into db.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/SubscriptionFormFieldSchema.java
```diff
@@ -0,0 +1,54 @@
...
+ * @param fieldKey   key used in the submitted values map; maps to the fieldKey attribute in GMD
+ * @param required   whether a value must be present in the submission
+ * @param minLength  minimum string length
+ * @param maxLength  maximum string length
+ * @param pattern    regex pattern the value must match
+ * @param options    allowed values
+ *
+ * @author Gravitee.io Team
+ */
+public record SubscriptionFormFieldSchema(
```
dev-marek: Instead using generic record for all component types, please create specific inner records/classes in `SubscriptionFormSchema`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormSchemaGeneratorImpl.java
```diff
@@ -0,0 +1,106 @@
...
+        "gmd-radio",
+        SubscriptionFormComponent.RADIO,
+        "gmd-checkbox",
+        SubscriptionFormComponent.CHECKBOX,
+        "gmd-checkbox-group",
+        SubscriptionFormComponent.CHECKBOX_GROUP
+    );
+
+    @Override
+    public SubscriptionFormSchema generate(String gmdContent) {
```
dev-marek: Please use `GraviteeMarkdown`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormSchemaGeneratorImpl.java
```diff
@@ -0,0 +1,106 @@
...
+            return new SubscriptionFormSchema(List.of());
+        }
+
+        List<SubscriptionFormFieldSchema> fields = Jsoup.parseBodyFragment(gmdContent)
+            .body()
+            .getAllElements()
+            .stream()
+            .filter(el -> TAG_TO_COMPONENT.containsKey(el.tagName()))
+            .map(this::toFieldSchema)
+            .toList();
```
dev-marek: This code belongs to `io.gravitee.apim.core.gravitee_markdown` package. Please create `GraviteeMarkdownParser` with method `findElements(GraviteeMarkdown content, List<String> tags)`

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidator.java
JedrzejJanasiak: I'd leave unknown field handling out of the validator. `CreateSubscriptionUseCase` is shared between papi and mapi, and metadata in a subscription request can carry all sorts of things unrelated to the subscription form, so rejecting unknown keys in the validator would cause errors for callers that have no idea about the subscription form schema.

My plan for the PR where I wire this up: the use case decides whether to call the validator at all (only when the API has a subscription form), and filters metadata down to schema-known keys before passing them in. The validator stays deliberately dumb; it only sees what it's given.

What do you think? 🤔

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormSchemaGeneratorImpl.java
JedrzejJanasiak: I see why it makes sense, but since we can't pull Jsoup into core, making this work would require a `GmdElement` wrapper record in core to avoid leaking Jsoup's `Element` type through the interface. The downside is that `GmdElement` ends up being a stripped-down version of Jsoup's `Element`, but losing things like children(), text(), or CSS selectors. So if GMD parsing needs grow in the future, the wrapper would need to grow with it too.

On top of that, the generator is currently called only once on API save, and future stories will work with the persisted schema rather than raw GMD, so I don't see a second consumer on the horizon.

For me, this feels like overabstracting right now, but if you'd like it done for cleanliness, an example implementation would look something like this:

``` java
// core
public record GmdElement(String tagName, Map<String, String> attributes) {
    public String attr(String name) { return attributes.getOrDefault(name.toLowerCase(), ""); }
    public boolean hasAttr(String name) { return attributes.containsKey(name.toLowerCase()); }
}

public interface GraviteeMarkdownParser {
    List<GmdElement> findElements(GraviteeMarkdown content, List<String> tags);
}

// infra
public class GraviteeMarkdownParserImpl implements GraviteeMarkdownParser {
    @Override
    public List<GmdElement> findElements(GraviteeMarkdown content, List<String> tags) {
        Set<String> tagSet = new HashSet<>(tags);
        return Jsoup.parseBodyFragment(content.value())
            .body().getAllElements().stream()
            .filter(el -> tagSet.contains(el.tagName()))
            .map(el -> {
                Map<String, String> attrs = new HashMap<>();
                el.attributes().forEach(a -> attrs.put(a.getKey(), a.getValue()));
                return new GmdElement(el.tagName(), attrs);
            })
            .toList();
    }
}
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidator.java
JedrzejJanasiak: Done! Went with buildValidator() as a behavior on SubscriptionFormSchema. One thought on persisting the validator instead: the schema is going to serve a few more purposes down the line (frontend rendering, form versioning, metadata filtering), so tightly coupling validation to the schema object might actually get confusing given all those responsibilities. But maybe we can chat about it later

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidator.java
dev-marek: @kovaren if we would need this check we'd add `failIfUnknown` flag to validator on Plan level

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormConstraintsFactory.java
```diff
@@ -0,0 +1,133 @@
...
+            }
+            case CheckboxGroupField f -> {
+                if (f.required()) {
+                    out.add(new Constraint.NonEmptySelection());
+                }
+                if (hasOptions(f.options())) {
+                    out.add(new Constraint.EachOf(f.options()));
+                }
+            }
+        }
```
dev-marek, line 87: Instead of cycling trough all field types, it would be more clear just to check by field attribute, eg.:

```
        if (field instanceof SubscriptionFormSchema.RequiredAttribute required && required.required()) {
            out.add(new Constraint.Required());
        }
        if (field instanceof SubscriptionFormSchema.MinLengthAttribute minLength && minLength.minLength() != null) {
            out.add(new Constraint.MinLength(minLength.minLength()));
        }
        if (field instanceof SubscriptionFormSchema.MaxLengthAttribute maxLength && maxLength.maxLength() != null) {
            out.add(new Constraint.MaxLength(maxLength.maxLength()));
        }
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidator.java
```diff
@@ -0,0 +1,75 @@
...
+
+        List<String> errors = new ArrayList<>();
+        for (var entry : fieldConstraints.byFieldKey().entrySet()) {
+            String fieldKey = entry.getKey();
+            String raw = submittedValues.get(fieldKey);
+            String value = raw != null ? raw.trim() : "";
+            for (Constraint constraint : entry.getValue()) {
+                errors.addAll(constraint.validate(fieldKey, value));
+            }
+        }
```
dev-marek: This code can be more clear if you split validate into check and format message, eg.:

```
        List<String> errors = fieldConstraints.byFieldKey().entrySet().stream().map(
                        e -> e.getValue().stream()
                                .filter(c -> !c.check(submittedValues.getOrDefault(e.getKey(), "").trim()))
                                .map(c -> c.formatErrorMessage(e.getKey(), submittedValues.getOrDefault(e.getKey(), "").trim()))
                                .toList()
                )
                .flatMap(List::stream)
                .toList();
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/SubscriptionFormSchema.java
```diff
@@ -0,0 +1,72 @@
...
+     * Each implementation captures only the validation attributes relevant to its component type.
+     */
+    public sealed interface Field permits InputField, TextareaField, SelectField, RadioField, CheckboxField, CheckboxGroupField {
+        String fieldKey();
+
+        boolean required();
+    }
+
+    /** Text input — supports required, readonly, minLength, maxLength, pattern. */
+    public record InputField(
```
dev-marek, line 72: Using interfaces to reflect set of attributes would allow handling fields by constraints

```
    public interface RequiredAttribute {
        boolean required();
    }

    public interface ReadOnlyValueAttribute {
        String readonlyValue();
    }

    public interface MinLengthAttribute {
        Integer minLength();
    }
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/Constraint.java
```diff
@@ -0,0 +1,149 @@
...
+     * @param value    the submitted value, already trimmed; empty string when absent
+     * @return an empty list if the value is valid, or one-or-more error messages otherwise
+     */
+    List<String> validate(String fieldKey, String value);
+
+    /** Text field must not be blank. */
+    record Required() implements Constraint {
+        @Override
+        public List<String> validate(String fieldKey, String value) {
+            return value.isEmpty() ? List.of("Field '" + fieldKey + "' is required") : List.of();
```
dev-marek: Using `String.format(...)` will silence sonar

dev-marek (changes requested):

dev-marek (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidator.java
kovaren: Ok good, let's keep it as is

kovaren (approved):

PR #15809, opened 2026-03-19
refactor: ui for mtls rotation m2

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-1999

https://github.com/user-attachments/assets/386a054d-64c0-4662-bb53-bd7da4e40e20

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/ApplicationClientCertificatesResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/ApplicationClientCertificatesResource.java
@@ -17,6 +17,7 @@ package io.gravitee.rest.api.management.rest.resource;
 
 import io.gravitee.apim.core.application_certificate.use_case.CreateClientCertificateUseCase;
 import io.gravitee.apim.core.application_certificate.use_case.GetClientCertificatesUseCase;
+import io.gravitee.apim.core.application_certificate.use_case.ValidateClientCertificateUseCase;
 import io.gravitee.common.data.domain.Page;
 import io.gravitee.common.http.MediaType;
 import io.gravitee.rest.api.management.rest.mapper.ClientCertificateMapper;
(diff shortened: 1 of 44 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/application_certificate/domain_service/ClientCertificateValidationDomainService.java
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Parses and validates the given PEM certificate, then returns its metadata.
+     *
+     * @param pemCertificate the PEM-encoded certificate string
+     * @return extracted certificate information
+     * @throws ClientCertificateInvalidException if the PEM cannot be parsed or fingerprint generation fails
+     * @throws ClientCertificateEmptyException if the PEM contains no certificates
+     * @throws ClientCertificateAuthorityException if the certificate is a CA certificate
+     */
+    public CertificateInfo validate(String pemCertificate) {
```
benoitgravitee: Checks are in ClientSubscriptionCrusServiceImpl.
You can leave them there, but reuse them, it is a much simpler and efficient implementation of this validation.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/ApplicationClientCertificatesResource.java
benoitgravitee, line 116: Modify this usecase so that it benefits from the validation. User of this API may not call validate at all.

benoitgravitee (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/application_certificate/domain_service/ClientCertificateValidationDomainService.java
a-cordier: I'll go with refactoring to avoid the validation from happening in a CRUD service I think

benoitgravitee (approved):

PR #15823, opened 2026-03-20
fix: api log console error

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-12623

## Description

Aspect | Details
-- | --
Problem | When Response (or other logging phases) is disabled in Reporter Settings, the API log detail page throws TypeError: can't access property "status", t.entrypointResponse is undefined and the Details section does not render.
Root cause | The backend omits entrypointRequest, endpointRequest, entrypointResponse, and endpointResponse when their logging phase is disabled. The UI always accessed these fields without checking if they existed.
Solution | Treat these fields as optional and render each section only when its data is present. Hide the "Details" heading when none of the four objects exist.
Files changed | connectionLog.ts – mark request/response fields as optional; api-proxy-request-log-overview.component.html – add conditional rendering for each section.
Behavior | With all logging enabled: unchanged. With partial logging: only sections with data are shown;.


pragsty:
Raised with incorrect target branch, I'll be creating new PR

PR #15831, opened 2026-03-20
fix(elasticsearch): idempotent Lucene value escaping for body/payload search (APIM-13030)

wbabyte:
## Summary

- **APIM-13030**: Body/payload search in Analytics → Logs returns no results — regression introduced by APIM-12955
- Root cause: `createSafeElasticsearchJsonQuery()` applied Lucene escaping to the full query string including field patterns, corrupting the `\\*.body` wildcard field prefix used for body searches
- Fix: escaping is now applied **only to values**, not field names or patterns, across all three query paths
- Escaping is **idempotent**: detects both the console's 2-backslash pre-escape convention (`\\X`) and raw-client 1-backslash convention (`\X`), normalising both to the same Lucene-safe output — prevents double-escaping when the console or API clients pre-escape before sending

## What changed

- `getQuery()` — escapes values when building body-field queries; preserves `\\*.body` wildcard field prefix intact
- `escapeValueForLuceneJson()` — new 3-char lookahead algorithm: detects `\\X` (console convention, pass-through) and `\X` (raw-client, normalise to `\\X`); `*` and `?` intentionally not escaped (Lucene wildcards)
- `createSafeElasticsearchJsonQuery()` / `escapeFilterValues()` — same idempotent logic for the non-body query path
- 24 unit tests in `LuceneEscapeTest` covering all escaping scenarios including regression guards for both APIM-12955 and APIM-13030

## Test plan

- [ ] Body search with plain word returns results
- [ ] Body search with special chars in value — `*(Ubuntu)*`, `*RMel/1.0.0*` — returns results
- [ ] Non-body filter with special chars — `custom.userAgent:RMel/1.0.0 (Ubuntu)` — works (APIM-12955 regression)
- [ ] Combined query — `status:200 AND body:term` — works
- [ ] Wildcard body search — `body:*term*` — works
- [ ] Unit tests pass (`LuceneEscapeTest` — 24 tests)

🔗 [APIM-13030](https://gravitee.atlassian.net/browse/APIM-13030)

[APIM-13030]: https://gravitee.atlassian.net/browse/APIM-13030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/log/ElasticLogRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/log/ElasticLogRepository.java
@@ -96,7 +96,7 @@ public class ElasticLogRepository extends AbstractElasticsearchRepository implem
                     .size(MAX_RESULT_WINDOW)
                     .query(logQueryString)
                     .build();
-                final String sQuery = this.createSafeElasticsearchJsonQuery(logQuery);
+                final String sQuery = this.createElasticsearchJsonQuery(logQuery);
 
                 Single<SearchResponse> result = this.client.search(
                     this.indexNameGenerator.getIndexName(queryContext.placeholder(), Type.LOG, from, to, clusters),
@@ -127,7 +127,7 @@ public class ElasticLogRepository extends AbstractElasticsearchRepository implem
                     result = this.client.search(
                         this.indexNameGenerator.getIndexName(queryContext.placeholder(), Type.REQUEST, from, to, clusters),
                         !info.getVersion().canUseTypeRequests() ? null : Type.REQUEST.getType(),
-                        this.createSafeElasticsearchJsonQuery(requestQueryBuilder.build())
+                        this.createElasticsearchJsonQuery(requestQueryBuilder.build())
                     );
                 }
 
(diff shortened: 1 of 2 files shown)


gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/log/ElasticLogRepository.java
```diff
@@ -263,12 +266,43 @@ private String createSafeElasticsearchJsonQuery(final TabularQuery query) {
...
+        final String andSeparator = " AND ";
+        return stream(filter.split(andSeparator)).map(ElasticLogRepository::escapeClauseValues).collect(joining(andSeparator));
+    }
+
+    private static String escapeClauseValues(String clause) {
+        // Handle OR-separated sub-clauses (e.g. "_id:x OR _id:y")
+        if (clause.contains(" OR ")) {
+            return stream(clause.split(" OR ")).map(ElasticLogRepository::escapeSingleClauseValue).collect(joining(" OR "));
+        }
+        return escapeSingleClauseValue(clause);
```
mukul-tyagi08, line 297: I think we can add a test for this to verify field names survive because of explicit "OR" handling

mukul-tyagi08 (approved):
A small comment for test case and also the commit has JIRA reference, not sure if that's okay.

wbabyte:
closing, since it could introduce regression

PR #15838, opened 2026-03-20
fix: forbid plan and application changes during subscription updates

benoitgravitee:
## Issue

https://gravitee.atlassian.net/browse/GKO-2430

## Description

* Check for changes in plan and application
* Tests
* Will apply it to 4.10.x as bug was found in 0.4.x

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription/SubscriptionCRDSpecDomainServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription/SubscriptionCRDSpecDomainServiceImpl.java
@@ -19,6 +19,8 @@ import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.apim.core.subscription.domain_service.AcceptSubscriptionDomainService;
 import io.gravitee.apim.core.subscription.domain_service.CloseSubscriptionDomainService;
 import io.gravitee.apim.core.subscription.domain_service.SubscriptionCRDSpecDomainService;
+import io.gravitee.apim.core.subscription.exception.SubscriptionApplicationImmutableException;
+import io.gravitee.apim.core.subscription.exception.SubscriptionPlanImmutableException;
 import io.gravitee.apim.core.subscription.model.SubscriptionEntity;
 import io.gravitee.apim.core.subscription.model.crd.SubscriptionCRDSpec;
 import io.gravitee.apim.infra.adapter.SubscriptionAdapter;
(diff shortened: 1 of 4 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription/SubscriptionCRDSpecDomainServiceImpl.java
```diff
@@ -19,6 +19,9 @@
 import io.gravitee.apim.core.subscription.domain_service.AcceptSubscriptionDomainService;
 import io.gravitee.apim.core.subscription.domain_service.CloseSubscriptionDomainService;
 import io.gravitee.apim.core.subscription.domain_service.SubscriptionCRDSpecDomainService;
+import io.gravitee.apim.core.subscription.exception.SubscriptionApiImmutableException;
```
carlos-andres-osorio: This import references `SubscriptionApiImmutableException`, a class that does not exist anywhere in the codebase — only `SubscriptionPlanImmutableException` and `SubscriptionApplicationImmutableException` were created. The build will fail to compile.

The presence of this import also suggests a check for the API reference field (`referenceId`) may have been planned but never written.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription/SubscriptionCRDSpecDomainServiceImpl.java
```diff
@@ -93,6 +96,8 @@ private SubscriptionEntity update(AuditInfo auditInfo, SubscriptionEntity existi
             return update(auditInfo, restored, spec);
```
carlos-andres-osorio, line 97: When a CLOSED subscription is updated with a changed `planId` or `applicationId`, `restoreAsAccepted()` executes here — calling `subscriptionService.restore()` and the auto-accept, both of which write to the database — before `rejectImmutableFieldChanges` is reached in the recursive call. The caller receives the correct exception, but the subscription is left in ACCEPTED state in storage: it has been reopened and is now visible to the management API, even though the update should have been rejected without any side effects.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/domain_service/subscription/SubscriptionCRDSpecDomainServiceImplTest.java
```diff
@@ -282,6 +299,38 @@ void should_restore_AsAccepted_closed_subscription_on_update() {
         });
```
carlos-andres-osorio, line 301: The existing restore test only covers the case where `planId` and `applicationId` match the spec. There is no test for a CLOSED subscription update that also changes one of those fields.

A test for this scenario must assert both that the expected exception is thrown **and** that `subscriptionService.restore()` was never called — confirming the subscription was not reopened before being rejected.

carlos-andres-osorio (approved):

PR #15841, opened 2026-03-21
fix(mAPI): support only user mode for API Product PO

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-13276

## Description

Currently, API product only supports users as PO and not other modes (Hybrid, GROUPS)


arpzilla-dev (approved):
lgtm

PR #15847, opened 2026-03-23
fix: export API categories by name

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-11792

## Description

Area | File | Change
-- | -- | --
Export | ApiExportDomainServiceImpl.java | Injected ApiCategoryQueryService; for V4 and V4_NATIVE, use apiWithCategoryKeys() to convert category IDs to keys before export so definitions are portable across environments.
Domain Service | CategoryDomainService.java | Added resolveToCategoryIds(String environmentId, Set<String> categoryIdsOrKeys).
Domain Service | CategoryDomainServiceImpl.java | Implemented resolveToCategoryIds() by delegating to CategoryMapper.toCategoryId().
Import | ImportDefinitionCreateDomainService.java | Injected CategoryDomainService; before create, resolve category keys to IDs via resolveCategoriesForImport() so imports keep category associations when using keys in the export.

After fix: 
image

image

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/domain_service/CategoryDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/domain_service/CategoryDomainService.java
@@ -16,7 +16,6 @@
 package io.gravitee.apim.core.api.domain_service;
 
 import io.gravitee.apim.core.api.model.Api;
-import io.gravitee.repository.exceptions.TechnicalException;
 import java.util.Collection;
 import java.util.Set;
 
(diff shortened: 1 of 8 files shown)


aditya-goyal01 (commented):
## 📋 Review Summary

This PR significantly improves API portability by exporting categories by name (key) instead of internal IDs, preventing broken associations across environments. It correctly updates both V4/Native export logic and the import resolution process.

## 🔍 General Feedback

- **Missing Coverage:** The portability fix is currently limited to V4 and V4 Native APIs. V2 and Federated APIs still export internal IDs, which will cause the same issue for those API types.
- **Resilience:** The export logic uses . If this returns an empty list while categories exist (e.g., due to mapping issues), the categories are lost in the export. It might be safer to fall back to the IDs if keys cannot be resolved.

aditya-goyal01 (approved):

PR #15848, opened 2026-03-23
fix: redirect to 404 from subscribe page when api is unpublished.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-13287

## Description

Fix a case when an API is unpublished and the subscribe page is accessed by URL and appears empty.

https://github.com/user-attachments/assets/137a4b61-611c-4f33-814a-ccc2b3c69a06

### DIFF
--- a/gravitee-apim-portal-webui-next/src/resolvers/api.resolver.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/resolvers/api.resolver.spec.ts
@@ -13,21 +13,55 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import { HttpErrorResponse } from '@angular/common/http';
+import { ApplicationRef, Component } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
-import { ResolveFn } from '@angular/router';
+import { NavigationEnd, provideRouter, Router } from '@angular/router';
+import { filter, firstValueFrom, take, throwError } from 'rxjs';
 
 import { apiResolver } from './api.resolver';
-import { Api } from '../entities/api/api';
+import { ApiService } from '../services/api.service';
 
 describe('apiResolver', () => {
-  const executeResolver: ResolveFn<Api> = (...resolverParameters) =>
-    TestBed.runInInjectionContext(() => apiResolver(...resolverParameters));
+  @Component({ standalone: true, template: '' })
+  class StubComponent {}
 
-  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


gravitee-apim-portal-webui-next/src/resolvers/api.resolver.spec.ts
aqibmohammadkhan, line 1: a 404 maps to router.navigate(['/404']) and returns EMPTY. The unit test subscribes directly to the resolver observable and asserts completion, but does not verify what Angular Router does when a resolver completes without emitting during an active navigation.
Risk: depending on router internals/version, this can produce a canceled/errored current navigation rather than a clean redirect, and the current test would still pass.

Could we add a router-level test (via RouterTestingModule + real navigation) to validate that returning EMPTY from the resolver during a 404 actually redirects to /404 and doesn’t leave navigation canceled? Current test only checks observable completion.

gravitee-apim-portal-webui-next/src/resolvers/api.resolver.ts
```diff
@@ -24,4 +25,14 @@ export const apiResolver = ((
   route: ActivatedRouteSnapshot,
   _: RouterStateSnapshot,
   apiService: ApiService = inject(ApiService),
-): Observable<Api> => apiService.details(route.params['apiId'])) satisfies ResolveFn<Api>;
+  router: Router = inject(Router),
+): Observable<Api> =>
+  apiService.details(route.params['apiId']).pipe(
+    catchError((err: unknown) => {
+      if (err instanceof HttpErrorResponse && err.status === 404) {
+        void router.navigate(['/404']);
```
aqibmohammadkhan, line 33: Can we use a centralized route constant/helper for the not-found path instead of hardcoding '/404' to avoid route drift?

here we are using a literal path. If route config changes, this can silently drift.

kovaren: This suggestion is valid but there are many other places in the code with a hardcoded 404.

WDYT about handling this in a separate PR to avoid unrelated changes in this PR? I can create a ticket for that.

gravitee-apim-portal-webui-next/src/resolvers/api.resolver.spec.ts
kovaren: Fixed

aqibmohammadkhan (approved):

PR #15862, opened 2026-03-23
fix:Webhook Logs toggle not updated after save/publish

bmouhaimen:
This PR addresses a UI state synchronization bug where the Webhook Logs reporting toggle (and potentially other API settings) fails to reflect the correct state immediately after a Save and Publish action.
The root cause was  data stream using shareReplay(1). This cached the initial API definition and prevented the UI from reflecting updates sent to the backend until a manual browser refresh was performed.
Changes:
The proposed fix addresses the state-synchronization lag by replacing the static, performance-cached api$ observable with a dynamic ReplaySubject<Api>(1). This architectural shift decouples the data stream from the initial HTTP request, allowing the UI to remain subscribed to a consistent observable while the underlying data is updated in the background. To manage this, a new loadApi() method was introduced to explicitly fetch the latest API definition from the apiService using a take(1) pipe for memory efficiency. This method is now integrated into two critical lifecycle points: first in ngOnInit() for the initial page load, and again within the openSettingsDialog() success callback. By triggering this re-fetch immediately after a successful save (result?.saved), the local UI state is forced to synchronize with the Management API's truth, ensuring the Webhook Logs toggle accurately reflects the enabled state without requiring a manual browser refresh.

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/api-traffic-v4/webhook-logs/webhook-logs.component.ts
+++ b/gravitee-apim-console-webui/src/management/api/api-traffic-v4/webhook-logs/webhook-logs.component.ts
@@ -16,8 +16,8 @@
 import { isNumber } from 'angular';
 
 import { Component, inject, OnInit } from '@angular/core';
-import { map, shareReplay, switchMap, take, catchError, finalize } from 'rxjs/operators';
-import { ReplaySubject, of, Observable } from 'rxjs';
+import { map, switchMap, take, catchError, finalize } from 'rxjs/operators';
+import { ReplaySubject, of, Observable, EMPTY } from 'rxjs';
 import { ActivatedRoute, Params, Router } from '@angular/router';
 import { CommonModule } from '@angular/common';
 import { MatCardModule } from '@angular/material/card';
@@ -75,7 +75,8 @@ export class WebhookLogsComponent implements OnInit {
   private readonly applicationService = inject(ApplicationService);
   private readonly dialog = inject(MatDialog);
   private readonly snackBarService = inject(SnackBarService);
-  private readonly api$ = this.apiService.get(this.activatedRoute.snapshot.params.apiId).pipe(shareReplay(1));
+  private readonly apiSubject$ = new ReplaySubject<Api>(1);
+  private readonly api$: Observable<Api> = this.apiSubject$.asObservable();
 
   isReportingDisabled$ = this.api$.pipe(
     map((api) => {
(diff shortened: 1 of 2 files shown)


gravitee-apim-console-webui/src/management/api/api-traffic-v4/webhook-logs/webhook-logs.component.ts
```diff
@@ -610,6 +613,19 @@ export class WebhookLogsComponent implements OnInit {
...
   }
 
+  private loadApi(): void {
+    this.apiService
+      .get(this.activatedRoute.snapshot.params.apiId)
+      .pipe(
+        take(1),
+        catchError(() => {
+          this.snackBarService.error('Failed to load API settings');
+          return EMPTY;
```
mukul-tyagi08, line 623: Returning EMPTY on initial-load failure leaves apiSubject$ silent, dependent streams and the settings button may hang 
We can probably emit a fallback (e.g. of(null)) and guard downstream consumers, so the UI degrades visibly instead of silently
Wdyt?

mukul-tyagi08 (changes requested):

mukul-tyagi08 (commented):
Also just to confirm, the fix is from 4.10 onwards, not on 4.7,8.9?

gravitee-apim-console-webui/src/management/api/api-traffic-v4/webhook-logs/webhook-logs.component.ts
bmouhaimen: The behavior differs depending on which call fails:                                                                                                                         
                                                                                                                                                                                
  - Re-fetch after save (the bug scenario): EMPTY is the right choice — apiSubject$ keeps its last valid value, the UI stays in the previous known state, and the user sees the snack bar notification.                                                                                                                            
  - Initial load failure: if the API call fails at page load, the entire page is broken regardless — this is better handled at the router/guard level with a redirect, not by emitting a sentinel value inside the component. Also, of(null) would require changing the type signature from Observable<Api> to Observable<Api | null> and adding null guards in every downstream pipe.
                                                                                                                                                                                
  Worth noting: before this fix, shareReplay(1) would have propagated the error to all subscribers simultaneously, crashing every pipe that depends on api$. EMPTY + snack bar is already a strict improvement over that.
Wdyt?

bmouhaimen:
The WebhookLogsComponent was introduced in 4.10

gravitee-apim-console-webui/src/management/api/api-traffic-v4/webhook-logs/webhook-logs.component.ts
mukul-tyagi08: Agree

mukul-tyagi08 (approved):

PR #15864, opened 2026-03-23
feat: persist validation constraints for subscription forms

JedrzejJanasiak:
## Issue
https://gravitee.atlassian.net/browse/APIM-12990

## Description
**PR 2/3 — Repository layer for subscription form validation constraints.**
Previous: [apim-12990-validation-domain (domain layer)](https://github.com/gravitee-io/gravitee-api-management/pull/15785)

This PR adds persistence of `SubscriptionFormFieldConstraints` derived from the GMD schema,
covering the full repository stack and a backfill upgrader for existing forms.

**What's in this PR:**
- DB migration: nullable `validation_constraints` column on `subscription_forms` (JDBC + MongoDB)
- JSON serialization/deserialization of `SubscriptionFormFieldConstraints` at the adapter boundary
- `SubscriptionFormConstraintsUpgrader`: backfills constraints for forms created before this feature

**Not in this PR** (coming in PR 3/3 — resource layer):
- REST resource changes and OpenAPI spec updates
- Constraint generation at save time (use case currently passes `null` to defer to the resource layer)
- Integration tests for the full save → persist → validate flow

## Additional context
The JSON format stored on DB is a flat field map:
`{"email":[{"type":"required"}],"company":[{"type":"required"},{"type":"minLength","min":2}]}`

Upgrader tested manually against PostgreSQL and MongoDB.

PR #15918, opened 2026-03-24
feat: wire constraint generation on form save and validate on subscri…

JedrzejJanasiak:
## Issue

## Description
**PR 3/3** — Resource layer: constraint generation on form save + submission validation.
Previous: [#15864 (PR 2/3 — repository layer, constraint persistence)](https://github.com/gravitee-io/gravitee-api-management/pull/15864)

This PR adds:
- `UpdateSubscriptionFormUseCase` now generates constraints from GMD on every save
  (`GMD → schemaGenerator.generate() → ConstraintsFactory.fromSchema() → persist`)
- `SubscriptionForm.update()` simplified — always replaces constraints, null-as-skip removed
- `SubscriptionValidationServiceImpl` validates submitted metadata against stored constraints
  at subscription creation time (only when form exists, is enabled, and has non-empty constraints)
- Tests for `UpdateSubscriptionFormUseCase` (constraint generation scenarios)
- Tests for `SubscriptionValidationServiceImpl` (all guard conditions: form absent, disabled,
  empty constraints, null metadata, valid/invalid submission)
- Portal resource test: `SubscriptionFormValidationException` → HTTP 400

**Not in this PR** (coming in following story):
- EL-based dynamic options validation
- Frontend integration

## Additional context
The validator is intentionally "dumb" — it only validates metadata keys that have constraints
defined in the form. Metadata from other sources is passed through without validation.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
@@ -18,6 +18,9 @@ package io.gravitee.rest.api.service.v4.impl.validation;
 import io.gravitee.apim.core.application_certificate.crud_service.ClientCertificateCrudService;
 import io.gravitee.apim.core.application_certificate.model.ClientCertificate;
 import io.gravitee.apim.core.application_certificate.model.ClientCertificateStatus;
+import io.gravitee.apim.core.subscription_form.domain_service.SubscriptionFormSubmissionValidator;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionForm;
+import io.gravitee.apim.core.subscription_form.query_service.SubscriptionFormQueryService;
 import io.gravitee.definition.model.v4.plan.PlanMode;
 import io.gravitee.rest.api.model.NewSubscriptionEntity;
 import io.gravitee.rest.api.model.PlanSecurityType;
@@ -32,6 +35,7 @@ import io.gravitee.rest.api.service.v4.exception.SubscriptionEntrypointIdMissing
 import io.gravitee.rest.api.service.v4.validation.SubscriptionMetadataSanitizer;
 import io.gravitee.rest.api.service.v4.validation.SubscriptionValidationService;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import lombok.CustomLog;
 import lombok.RequiredArgsConstructor;
@@ -48,6 +52,7 @@ public class SubscriptionValidationServiceImpl extends TransactionalService impl
 
     private final EntrypointConnectorPluginService entrypointService;
     private final SubscriptionMetadataSanitizer subscriptionMetadataSanitizer;
+    private final SubscriptionFormQueryService subscriptionFormQueryService;
 
     private final ClientCertificateCrudService clientCertificateCrudService;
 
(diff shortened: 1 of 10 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
```diff
@@ -57,6 +62,22 @@ public void validateAndSanitize(final GenericPlanEntity genericPlanEntity, final
         if (subscription.getMetadata() != null) {
             subscription.setMetadata(subscriptionMetadataSanitizer.sanitizeAndValidate(subscription.getMetadata()));
         }
+        validateSubscriptionFormMetadataIfApplicable(genericPlanEntity, subscription);
```
kovaren: Should we also call the validation in the other two overloads of `validateAndSanitize()`?

JedrzejJanasiak: Good point, definitely a valuable suggestion, but after digging into it a bit more it actually raised more questions than answers for me:

We have 3 overloaded `validateAndSanitize` methods in this class:

``` java
public void validateAndSanitize(final GenericPlanEntity genericPlanEntity, final NewSubscriptionEntity subscription)

public void validateAndSanitize(
    final GenericPlanEntity genericPlanEntity,
    final UpdateSubscriptionEntity subscription,
    String applicationId
)

public void validateAndSanitize(
    final GenericPlanEntity genericPlanEntity,
    final UpdateSubscriptionConfigurationEntity subscriptionConfiguration
)
```

The first validateAndSanitize() is used only for new subscriptions and validates the constraints there.
The second and third ones are triggered during subscription updates.

The issue I just realized comes from this PRD requirement:
Zrzut ekranu 2026-03-27 o 11 00 38

Based on what’s written there and how I understand it, the schema technically shouldn’t be validated when updating a subscription from the mAPI API Console.
This actually makes sense. If some constraints were defined incorrectly, the API Publisher should be able to quickly react and fix the metadata regardless of the form.

The problem is that SubscriptionService is shared between mAPI and pAPI, which leaves us with basically 3 options:

- Leave it as is. In that case, API Consumers get validation during the subscription flow when creating a subscription request, but it creates a loophole where, if they have edit permissions via pAPI, they can update the subscription without backend validation.
With this approach, creating a subscription via mAPI is still validated, but editing metadata remains unrestricted for the API Publisher.
- Add constraint validation to the other overloaded methods, but that would break the PRD assumptions.
- Refactor the subscription code and add more conditional logic to distinguish between requests coming from mAPI vs pAPI, and trigger validation only for the latter.
IMO this is the most reasonable option, but it’s also quite invasive. It would require significant changes in the subscription logic, a lot of test updates, and would add even more complexity to an already pretty complex backend flow.

Right now I’m exploring how to implement option 3 in a minimally invasive way, but honestly, I don’t think it’s going to be particularly pretty.

JedrzejJanasiak: As agreed with Varun, we decided to keep the logic, which performs constraint validation for all overloads of validateAndSanitize(). This ensures that constraint checks are applied consistently across both mAPI and pAPI, for both subscription creation and updates.

kovaren: Great, thanks!

kovaren (approved):

PR #15938, opened 2026-03-24
fix: let users pass page HRID for plan general condition

kamiiiel:
see https://gravitee.atlassian.net/browse/GKO-2703


benoitgravitee (commented):
There is no context to this change. What is the rational, usecase... is it a Terraform issue behind the scenes ?

PR #15954, opened 2026-03-25
fix: handle API product refs in key audit logs

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-13319

## Description

Pausing an API Product subscription updates API keys and writes an audit entry. ApiKeyServiceImpl.createAuditLog always stored AuditProperties.API from subscription.getApi(), which is null for API Product subscriptions. That produced a NULL in audit_properties.value and failed the JDBC insert (NOT NULL).

Change

Resolve referenceId from referenceId with fallback to api.
For referenceType == API_PRODUCT, use API_PRODUCT in audit properties and createApiProductAuditLog; otherwise keep API and createApiAuditLog.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApiKeyServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApiKeyServiceImpl.java
@@ -70,6 +70,7 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApiKeyServiceImpl.java
```diff
@@ -707,21 +708,30 @@ private void createAuditLog(
         key
             .getSubscriptions()
             .forEach(subscription -> {
+                boolean isApiProduct = isApiProductSubscription(subscription);
+                String referenceId = isApiProduct ? subscription.getReferenceId() : subscription.getApi();
```
vikrantgravitee: In future - getAPI will be deprecated, so ensure we build dependency on ReferenceID, ReferenceType.

vikrantgravitee: Simple could be:

String referenceId = subscription.getReferenceId() != null ? subscription.getReferenceId() : subscription.getApi();

pragsty: done

pragsty:
Raised a new PR: https://github.com/gravitee-io/gravitee-api-management/pull/16096/changes this is not needed anymmore.

PR #15985, opened 2026-03-25
fix: Migration to v4 engine : flows on deprecated plans not visible anymore

bmouhaimen:
Migration to v4 engine : flows on deprecated plans not visible anymore

https://gravitee.atlassian.net/browse/APIM-13302


araffass:
Issue already resolved.

PR #15996, opened 2026-03-25
fix(api): decouple group access management from Definition Update permission

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-13273

## Description

Group management was wrongly gated behind `API_DEFINITION` `UPDATE`. Adds a dedicated `PUT /apis/{apiId}/groups` endpoint with `API_MEMBER` `UPDATE` permission and updates the frontend accordingly.

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/user-group-access/members/api-general-members.component.ts
+++ b/gravitee-apim-console-webui/src/management/api/user-group-access/members/api-general-members.component.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import { Component, OnInit } from '@angular/core';
-import { combineLatest, EMPTY, forkJoin, Observable, of, Subject } from 'rxjs';
+import { combineLatest, EMPTY, forkJoin, Observable, Subject } from 'rxjs';
 import { catchError, filter, switchMap, takeUntil, tap } from 'rxjs/operators';
 import { UntypedFormBuilder, UntypedFormControl, UntypedFormGroup, Validators } from '@angular/forms';
 import { MatDialog } from '@angular/material/dialog';
@@ -244,11 +244,10 @@ export class ApiGeneralMembersComponent implements OnInit {
       .afterClosed()
       .pipe(
         filter((apiDialogResult): apiDialogResult is ApiGroupsDialogResult => !!apiDialogResult && !this.isKubernetesOrigin),
-        switchMap(apiDialogResult => {
-          return combineLatest([of(apiDialogResult), this.apiService.get(this.activatedRoute.snapshot.params.apiId)]);
-        }),
-        switchMap(([apiDialogResult, api]) => {
-          return this.apiService.update(api.id, { ...api, groups: apiDialogResult?.groups });
+        switchMap(apiDialogResult => this.apiService.updateGroups(this.apiId, apiDialogResult.groups)),
+        catchError(({ error }) => {
+          this.snackBarService.error(error.message);
+          return EMPTY;
         }),
         takeUntil(this.unsubscribe$),
       )
--- a/gravitee-apim-console-webui/src/services-ngx/api-v2.service.ts
+++ b/gravitee-apim-console-webui/src/services-ngx/api-v2.service.ts
@@ -81,6 +81,10 @@ export class ApiV2Service {
     );
   }
 
+  updateGroups(apiId: string, groups: string[]): Observable<string[]> {
+    return this.http.put<string[]>(`${this.constants.env.v2BaseURL}/apis/${apiId}/groups`, groups);
+  }
+
   delete(apiId: string, closePlans = false): Observable<void> {
     return this.http.delete<void>(`${this.constants.env.v2BaseURL}/apis/${apiId}${closePlans ? '?closePlans=true' : ''}`);
   }
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/UpdateApiGroupsUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/UpdateApiGroupsUseCase.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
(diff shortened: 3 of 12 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/UpdateApiGroupsUseCase.java
```diff
@@ -0,0 +1,75 @@
...
+    public Output execute(Input input) {
+        Api api = apiCrudService.get(input.apiId());
+
+        if (!api.getEnvironmentId().equals(input.auditInfo().environmentId())) {
+            throw new ApiNotFoundException(input.apiId());
+        }
+
+        Set<String> oldGroups = api.getGroups();
+
+        api.setGroups(input.groups());
```
aqibmohammadkhan, line 52: this usecase calls api.setGroups(input.groups()) with whatever strings the caller sends. A caller can:

- Pass a group ID that doesn't exist → stale reference in the DB
- Pass a group that has a primary owner → violation of the documented constraint

The codebase already has `ValidateGroupsDomainService` which:

- Resolves group IDs and names
- Filters out groups that have an apiPrimaryOwner (which is documented as unsupported)
- Emits warnings for non-existent group IDs
- Handles V2 vs V4 definition version differences

wdyt about injecting `ValidateGroupsDomainService` and run the input through it before persisting:

```
var validationResult = validateGroupsDomainService.validateAndSanitize(
    new ValidateGroupsDomainService.Input(
        input.auditInfo().environmentId(),
        input.groups(),
        api.getDefinitionVersion().getLabel(),
        Group.GroupEvent.API_SUBSCRIBED,
        false
    )
);
// handle warnings, use validationResult.value().groups()
api.setGroups(validationResult.value().groups());
```

gravitee-apim-console-webui/src/management/api/user-group-access/members/api-general-members.component.ts
```diff
@@ -244,12 +244,7 @@ export class ApiGeneralMembersComponent implements OnInit {
       .afterClosed()
       .pipe(
         filter((apiDialogResult): apiDialogResult is ApiGroupsDialogResult => !!apiDialogResult && !this.isKubernetesOrigin),
```
aqibmohammadkhan, line 246: The frontend correctly guards this in `updateGroups()` but a direct `PUT /apis/{apiId}/groups` call bypasses the frontend entirely. In `ApiGeneralMembersComponent.initForm()` [here](https://github.com/gravitee-io/gravitee-api-management/blob/ef033aed7056216fa1f7beb98c1859d40f417db7/gravitee-apim-console-webui/src/management/api/user-group-access/members/api-general-members.component.ts#L300)

The backend endpoint should check `api.getDefinitionContext().isOriginKubernetes()` and return a 400 before updating. Other update paths in the v1 ApiResource do this check (e.g., the lifecycle state endpoint). Without it, a Kubernetes-synced API can have its groups overwritten by an API call, only for the next sync to restore them — a potential data loss between syncs.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/UpdateApiGroupsUseCase.java
```diff
@@ -0,0 +1,75 @@
...
+        apiCrudService.update(api);
+
+        createAuditLog(oldGroups, input.groups(), input.apiId(), input.auditInfo());
+
+        return new Output(input.groups());
```
aqibmohammadkhan, line 57: Output returns input groups, not persisted state . If group validation (once added after [this](https://github.com/gravitee-io/gravitee-api-management/pull/15996/changes#r3111707384)) sanitizes the set — removing nonexistent IDs or primary-owner groups — the 200 OK response body will not reflect what was actually stored.

wdyt about 

```
Api updated = apiCrudService.update(api);
return new Output(updated.getGroups());
```

gravitee-apim-console-webui/src/management/api/user-group-access/members/api-general-members.component.ts
```diff
@@ -244,12 +244,7 @@ export class ApiGeneralMembersComponent implements OnInit {
...
-        switchMap(apiDialogResult => {
-          return combineLatest([of(apiDialogResult), this.apiService.get(this.activatedRoute.snapshot.params.apiId)]);
-        }),
-        switchMap(([apiDialogResult, api]) => {
-          return this.apiService.update(api.id, { ...api, groups: apiDialogResult?.groups });
-        }),
+        switchMap(apiDialogResult => this.apiService.updateGroups(this.apiId, apiDialogResult.groups)),
         takeUntil(this.unsubscribe$),
       )
       .subscribe(() => this.ngOnInit());
```
aqibmohammadkhan, line 250: this has no `catchError`. If the PUT returns a `4xx/5xx`, the user gets no feedback. This is a pre-existing gap but the refactor is a good opportunity to add it. Compare with `onSubmit()` which has:

```
catchError(({ error }) => {
  this.snackBarService.error(error.message);
  return EMPTY;
}),
```

Add the same catchError + snackBarService.error(...) pattern to the updateGroups() pipe.

gravitee-apim-console-webui/src/services-ngx/api-v2.service.ts
```diff
@@ -81,6 +81,10 @@ export class ApiV2Service {
     );
   }
 
+  updateGroups(apiId: string, groups: string[]): Observable<string[]> {
+    return this.http.put<string[]>(`${this.constants.env.v2BaseURL}/apis/${apiId}/groups`, groups);
+  }
```
aqibmohammadkhan, line 86: The backend returns the final group set in the response body. Typing this as `Observable<void>` means if the backend ever sanitizes the input ([here](https://github.com/gravitee-io/gravitee-api-management/pull/15996/changes#r3111707384)), the frontend can't reflect the actual state without a full page refresh. 

wdyt about Observable<string[]> and using the response to update `this.api.groups` in the subscription.

aqibmohammadkhan (approved):

PR #15999, opened 2026-03-25
feat: Add MCP tools in next gen portal.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-13306

## Description

- Add MCP entrypoints component to Next Gen portal documentation.
- Update styles of existing components to match latest typography.

https://github.com/user-attachments/assets/f19743c2-b18d-4298-9003-28b67abb174b

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-tools/api-tab-tools.component.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-tools/api-tab-tools.component.spec.ts
@@ -18,6 +18,7 @@ import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 
 import { ApiTabToolsComponent } from './api-tab-tools.component';
+import { ApiTabToolsComponentHarness } from './api-tab-tools.component.harness';
 import { McpToolHarness } from '../../../../components/mcp-tool/mcp-tool.harness';
 import { Api } from '../../../../entities/api/api';
 import { fakeApi } from '../../../../entities/api/api.fixtures';
@@ -50,9 +51,10 @@ describe('ApiTabToolsComponent', () => {
     );
     fixture.detectChanges();
 
-    const emptyMessage = fixture.nativeElement.querySelector('.api-tab-tools__empty');
+    const harness = await TestbedHarnessEnvironment.harnessForFixture(fixture, ApiTabToolsComponentHarness);
+    const emptyMessage = await harness.getEmptyToolsMessageText();
     expect(emptyMessage).not.toBeNull();
-    expect(emptyMessage.textContent.trim()).toContain('No tools available');
+    expect(emptyMessage).toContain('It seems that there are currently no tools available for this API.');
   });
 
   it('should show tools when available', async () => {
--- a/gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.html
+++ b/gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.html
@@ -16,8 +16,13 @@
 
 -->
 <app-sidenav-layout [breadcrumbs]="breadcrumbs()">
-  @if (subscribeApiId()) {
-    <div breadcrumbActions>
+  @if (apiId()) {
+    <div breadcrumbActions class="documentation-folder__breadcrumb-actions">
+      @if (apiHasMcp()) {
+        <button mat-stroked-button type="button" (click)="this.mcpDrawerOpen.set(true)" data-testid="mcp-button">
+          <span i18n="@@documentationMcpButton">MCP</span>
+        </button>
+      }
       <button mat-stroked-button (click)="onSubscribe()" [disabled]="!currentUser()" data-testid="subscribe-button">
         @if (currentUser()) {
           <span i18n="@@subscribeToApiButton">Subscribe</span>
@@ -43,3 +48,33 @@
     <app-navigation-item-content-viewer [pageContent]="folderData()?.selectedPageContent ?? null" />
   </div>
 </app-sidenav-layout>
+
+@if (mcpDrawerOpen()) {
+  <div
+    class="documentation-folder__mcp-backdrop"
+    aria-hidden="true"
+    tabindex="-1"
+    animate.enter="documentation-folder__mcp-backdrop__enter"
+    animate.leave="documentation-folder__mcp-backdrop__leave"
+    (click)="this.mcpDrawerOpen.set(false)"
+  ></div>
+  <div
+    class="documentation-folder__mcp-panel"
+    data-testid="mcp-tools-drawer"
+    cdkTrapFocus
+    cdkTrapFocusAutoCapture
+    animate.enter="documentation-folder__mcp-panel__enter"
+    animate.leave="documentation-folder__mcp-panel__leave"
+    role="dialog"
+    aria-modal="true"
+    i18n-aria-label="@@documentationMcpPanelAria"
+    aria-label="MCP tools"
+    [appIsMobile]="'documentation-folder__mcp-panel__mobile'"
+    (keydown.escape)="mcpDrawerOpen.set(false)"
+  >
+    <app-api-tab-tools [api]="api.value()!" />
+    <button class="documentation-folder__mcp-panel__close-button" mat-icon-button (click)="this.mcpDrawerOpen.set(false)">
+      <span class="material-icons" i18n="@@documentationMcpPanelCloseButton">close</span>
+    </button>
+  </div>
+}
--- a/gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.scss
+++ b/gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.scss
@@ -14,20 +14,121 @@
  * limitations under the License.
  */
 
+@use '../../../../scss/theme' as app-theme;
+
+$mcp-overlay-duration: 400ms;
+$mcp-overlay-easing: cubic-bezier(0.25, 0.8, 0.25, 1);
+
 :host {
+  position: relative;
   width: 100%;
 }
 
 app-sidenav-layout {
   height: 100%;
+}
 
-  .documentation-folder {
-    &__sidenav {
-      &__tree,
-      &__empty-state {
-        min-width: 250px;
-        height: 100%;
-      }
+.documentation-folder {
+  &__sidenav {
+    &__tree,
+    &__empty-state {
+      min-width: 250px;
+      height: 100%;
     }
+  }
+
+  &__breadcrumb-actions {
+    display: flex;
+    align-items: center;
+    gap: 8px;
+  }
+
+  &__mcp-backdrop {
+    position: absolute;
+    z-index: 10;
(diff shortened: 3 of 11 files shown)


gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.html
aqibmohammadkhan, line 1: This template uses attributes like `animate.enter="documentation-folder__mcp-backdrop__enter" `and `animate.leave="documentation-folder__mcp-panel__leave".` 
These are not standard Angular syntax. Angular's animation system uses `[@triggerName] `syntax with `BrowserAnimationsModule`, not `animate.*` bindings. These attributes will simply be rendered as static HTML attributes and will not produce any animation.

The SCSS defines `@keyframes fade-in, fade-out, slide-in, slide-out`, but there's no mechanism to trigger these animations on `enter/leave` since `@if` blocks immediately add/remove DOM elements without animation lifecycle hooks.

Recommendation: Either use Angular's `@angular/animations` with proper `trigger() / transition() / animate()`, or use CSS-only approach with a class toggle pattern instead of `@if` (e.g., always render the drawer and toggle a .visible class with CSS transitions).

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.html
```diff
@@ -43,3 +48,30 @@
...
+    data-testid="mcp-tools-drawer"
+    animate.enter="documentation-folder__mcp-panel__enter"
+    animate.leave="documentation-folder__mcp-panel__leave"
+    role="dialog"
+    aria-modal="true"
+    i18n-aria-label="@@documentationMcpPanelAria"
+    aria-label="MCP tools"
+    [appIsMobile]="'documentation-folder__mcp-panel__mobile'"
+  >
+    <app-api-tab-tools [api]="api.value()!" />
```
aqibmohammadkhan, line 75: `api.value()` from rxResource can be undefined (while loading, or if apiId is null). The `!` non-null assertion suppresses TypeScript but the component declares `api = input.required<Api>(),` so passing undefined will cause a runtime error if the drawer is opened before the API details have loaded.

wdyt about adding a guard here like:

```
@if (api.value(); as apiValue) {
  <app-api-tab-tools [api]="apiValue" />
}
```

Or disable the MCP button while api.isLoading().

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.ts
```diff
@@ -46,27 +50,39 @@ enum NavParamsChange {
...
   tree = signal<TreeNode[]>([]);
   breadcrumbs = signal<Breadcrumb[]>([]);
-  subscribeApiId = signal<string | null>(null);
-  readonly currentUser = inject(CurrentUserService).isUserAuthenticated;
+
+  apiId = signal<string | null>(null);
+  api = rxResource({
+    params: this.apiId,
+    stream: ({ params }) => (params ? this.apiService.details(params).pipe(map(api => api)) : of(undefined)),
+  });
```
aqibmohammadkhan: `.pipe(map(api => api))` is a no-op identity mapping and should be removed.

wdyt about simplifing this block to

```
api$ = toObservable(this.apiId).pipe(
  switchMap(id => id ? this.apiService.details(id) : of(undefined)),
);
api = toSignal(this.api$);
```

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.ts
```diff
@@ -46,27 +50,39 @@ enum NavParamsChange {
...
   folderData = toSignal<FolderData | undefined>(this.loadFolderData());
   tree = signal<TreeNode[]>([]);
   breadcrumbs = signal<Breadcrumb[]>([]);
-  subscribeApiId = signal<string | null>(null);
-  readonly currentUser = inject(CurrentUserService).isUserAuthenticated;
+
+  apiId = signal<string | null>(null);
+  api = rxResource({
+    params: this.apiId,
+    stream: ({ params }) => (params ? this.apiService.details(params).pipe(map(api => api)) : of(undefined)),
```
aqibmohammadkhan: The rxResource call to `this.apiService.details(params)` has no `catchError`. If the API details request fails (network error, 404, etc.), the rxResource error state is not handled in the template. The MCP button could be in an indeterminate state.

i would recommend handling api.error() in the template, or add error handling in the stream. At minimum, don't show the MCP button if the API call failed.

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.ts
```diff
@@ -46,27 +50,39 @@ enum NavParamsChange {
...
     SidenavLayoutComponent,
     TreeComponent,
     GraviteeMarkdownViewerModule,
     NavigationItemContentViewerComponent,
     AsyncPipe,
     MatButtonModule,
+    MatIconModule,
+    ApiTabToolsComponent,
   ],
   standalone: true,
```
aqibmohammadkhan: minor: Angular 19+ standalone is the default. no need to explicitly declare.

gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-tools/api-tab-tools.component.spec.ts
```diff
@@ -50,9 +50,9 @@ describe('ApiTabToolsComponent', () => {
     );
     fixture.detectChanges();
 
-    const emptyMessage = fixture.nativeElement.querySelector('.api-tab-tools__empty');
+    const emptyMessage = fixture.nativeElement.querySelector('.api-tab-tools__tools__content .next-gen-body');
```
aqibmohammadkhan: do we have harness for this? we could use harness instead of `fixture.nativeElement.querySelector`

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.scss
aqibmohammadkhan, line 1: as mentioned [here](https://github.com/gravitee-io/gravitee-api-management/pull/15999/changes#r3010835812)
`$mcp-overlay-duration `and animation keyframes won't actually be used. The CSS animation variables and keyframes are defined but the animate.enter/animate.leave binding mechanism doesn't work, making these dead CSS code.

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.html
```diff
@@ -43,3 +48,30 @@
...
+    <button class="documentation-folder__mcp-panel__close-button" mat-icon-button (click)="this.mcpDrawerOpen.set(false)">
+      <span class="material-icons" i18n="@@documentationMcpPanelCloseButton">close</span>
```
aqibmohammadkhan, line 77: The drawer panel has `aria-modal="true"` but there's no focus trap implementation. Users can tab to elements behind the backdrop. The close button uses <span class="material-icons">close</span> instead of <mat-icon>close</mat-icon>, which is inconsistent with the `MatIconModule` import.

wdyt about using Angular CDK's `cdkTrapFocus` directive on the panel, and use <mat-icon> for the close icon since MatIconModule is already imported.

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.html
```diff
@@ -43,3 +48,30 @@
...
+    </button>
```
aqibmohammadkhan, line 78: There's no keyboard handler for closing the drawer with the Escape key, which is standard UX for dialog/drawer patterns.

what if we Add `@HostListener('document:keydown.escape') `or a `(keydown.escape)` binding on the panel?

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.html
kovaren: Actually, this is the latest Angular syntax for animations https://angular.dev/guide/animations#animateenter 

[triggerName] API has been deprecated since Angular 20.

As we can see on the attached video, the animations work 😊

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.html
kovaren: The ! is used on purpose to communicate that api.value() is guaranteed. The MCP button is controlled by `subscribeApiHasMcp` which turns true only once the API has been loaded and has MCP enabled.

https://github.com/gravitee-io/gravitee-api-management/blob/51a8be6fabe2c0d88b58a3942e806fe1ebfe2d96/gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.ts#L84

I'd argue that the null check would be redundant here, WDYT?

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.ts
kovaren: Agreed on the no-op mapping, removed.

However, I'd prefer sticking with the rxResource, it's specifically designed to replace the toObservable / toSignal pair with just one signal.

WDYT?

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.ts
kovaren: Fixed!

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.ts
kovaren: Fixed

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.scss
kovaren: Animations work as shown in the video

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.html
kovaren: Good call, fixed!

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.html
kovaren: Fixed!

gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-tools/api-tab-tools.component.spec.ts
kovaren: Yep, fixed!

aqibmohammadkhan (approved):

PR #16020, opened 2026-03-26
fix: resolve CSRF token mismatch on changePassword endpoint

ThibaudAV:
https://gravitee.atlassian.net/browse/APIM-13340
https://gravitee.atlassian.net/browse/APIM-13293

## Summary
- Use single-segment wildcard (`*`) instead of multi-segment (`**`) in Spring Security path patterns for organization, environment, and user ID segments, fixing `PathPatternParser` matching issues
- Add `credentials: 'include'` to `fetch` requests in `index.ts` so CSRF cookies are sent cross-origin, preventing token regeneration on parallel requests
- Extract CSRF token storage from bootstrap response into dedicated `storeCsrfToken` function using `CsrfInterceptor`

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-security/src/main/java/io/gravitee/rest/api/management/security/config/BasicSecurityConfigurerAdapter.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-security/src/main/java/io/gravitee/rest/api/management/security/config/BasicSecurityConfigurerAdapter.java
@@ -257,8 +257,8 @@ public class BasicSecurityConfigurerAdapter implements SecureHeadersConfigurer {
     }
 
     private HttpSecurity authorizations(HttpSecurity security) throws Exception {
-        String uriOrgPrefix = "/organizations/**";
-        String uriPrefix = uriOrgPrefix + "/environments/**";
+        String uriOrgPrefix = "/organizations/*";
+        String uriPrefix = uriOrgPrefix + "/environments/*";
 
         return security
             .authorizeHttpRequests()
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-security/src/main/java/io/gravitee/rest/api/management/security/config/BasicSecurityConfigurerAdapter.java
```diff
@@ -257,8 +257,8 @@ private HttpSecurity session(HttpSecurity security) throws Exception {
     }
 
     private HttpSecurity authorizations(HttpSecurity security) throws Exception {
-        String uriOrgPrefix = "/organizations/**";
-        String uriPrefix = uriOrgPrefix + "/environments/**";
+        String uriOrgPrefix = "/organizations/*";
+        String uriPrefix = uriOrgPrefix + "/environments/*";
```
ThibaudAV, line 261: @phiz71 That's not the issue here. But while doing some research, I came across this, and when I checked the Java documentation, it seemed fine to me. Can you doubel check that this is the valid/better approach ?

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-security/src/main/java/io/gravitee/rest/api/management/security/config/BasicSecurityConfigurerAdapter.java
```diff
@@ -306,7 +306,7 @@ private HttpSecurity authorizations(HttpSecurity security) throws Exception {
             .permitAll()
             .requestMatchers(HttpMethod.POST, uriOrgPrefix + "/users/registration/**")
             .permitAll()
-            .requestMatchers(HttpMethod.POST, uriOrgPrefix + "/users/**/changePassword")
+            .requestMatchers(HttpMethod.POST, uriOrgPrefix + "/users/*/changePassword/**")
```
ThibaudAV: ** to remove

schehade-gravitee (approved):

PR #16032, opened 2026-03-27
fix(test): align webhook 5xx retry test with entrypoint-webhook 4.2.1…

mukul-tyagi08:
… behavior

## Issue

## Description


mukul-tyagi08:
Not needed, will revert the bump instead.

PR #16065, opened 2026-03-27
feat: add EL resolution for dynamic subscription form options - domai…

JedrzejJanasiak:
## Issue

## Description
**PR 1/2** — Domain & infrastructure layer for EL-based resolution of dynamic subscription form options.

This PR adds:
- `SubscriptionFormElResolverDomainService` interface with `resolveSchemaOptions` and `resolveConstraints` methods
- `SubscriptionFormElResolverDomainServiceImpl` — resolves EL expressions (e.g. `{#api.metadata['key']}:fallback`) against API and environment metadata; falls back gracefully on resolution failure
- `Constraint` model extended with `DynamicOneOf` / `DynamicEachOf` variants for EL-driven constraints
- `SubscriptionFormSchema` extended with `DynamicOptionsAttribute` carrying the raw EL expression and parsed fallback values
- `SubscriptionFormConstraintsFactory` — `constraintsFor` broken into focused private helpers to reduce cyclomatic complexity (Sonar)
- `GetSubscriptionFormForEnvironmentUseCase.Input` migrated to Lombok `@Builder`
- `ApiMetadataQueryService#findEnvironmentMetadata` — returns global/environment-level metadata without API overrides, used to populate EL template context when no API context is available
- In-memory doubles: `SubscriptionFormElResolverInMemory`, updated `ApiMetadataQueryServiceInMemory`

**Not in this PR** (coming in PR 2/2):
- REST resource changes (management v2 and portal endpoints)
- OAS spec updates
- Portal endpoint moved from env-level to API-level (`/apis/{apiId}/subscription-form`)

## Additional context
[Previous in chain: #15918 (apim-12990-validation-resource — validation constraints at resource layer)](https://github.com/gravitee-io/gravitee-api-management/pull/15918)

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormConstraintsFactory.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormConstraintsFactory.java
@@ -18,14 +18,16 @@ package io.gravitee.apim.core.subscription_form.domain_service;
 import io.gravitee.apim.core.subscription_form.model.Constraint;
 import io.gravitee.apim.core.subscription_form.model.SubscriptionFormFieldConstraints;
 import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema;
-import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.CheckboxField;
-import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.CheckboxGroupField;
 import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.Field;
 import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.InputField;
 import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.MinLengthAttribute;
-import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.OptionsAttribute;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.MultiValueOptionsField;
 import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.PatternAttribute;
 import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.ReadOnlyValueAttribute;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.RequiredSelectionAttribute;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.RequiredTrueAttribute;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.RequiredValueAttribute;
+import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.SingleValueOptionsField;
 import io.gravitee.apim.core.subscription_form.model.SubscriptionFormSchema.TextareaField;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
@@ -61,17 +63,27 @@ public final class SubscriptionFormConstraintsFactory {
         }
 
         var out = new ArrayList<Constraint>();
+        addRequiredConstraint(field, out);
+        addLengthConstraints(field, out);
+        addPatternConstraint(field, out);
+        addOptionsConstraint(field, out);
+        return out;
+    }
 
-        if (field.required()) {
-            if (field instanceof CheckboxField) {
-                out.add(new Constraint.MustBeTrue());
-            } else if (field instanceof CheckboxGroupField) {
-                out.add(new Constraint.NonEmptySelection());
-            } else {
-                out.add(new Constraint.Required());
-            }
+    private static void addRequiredConstraint(Field field, List<Constraint> out) {
+        if (!field.required()) {
+            return;
         }
+        if (field instanceof RequiredTrueAttribute) {
+            out.add(new Constraint.MustBeTrue());
+        } else if (field instanceof RequiredSelectionAttribute) {
+            out.add(new Constraint.NonEmptySelection());
+        } else if (field instanceof RequiredValueAttribute) {
+            out.add(new Constraint.Required());
+        }
+    }
 
+    private static void addLengthConstraints(Field field, List<Constraint> out) {
         if (field instanceof MinLengthAttribute min && min.minLength() != null) {
             out.add(new Constraint.MinLength(min.minLength()));
         }
@@ -82,19 +94,42 @@ public final class SubscriptionFormConstraintsFactory {
                 textarea.maxLength() != null ? Constraint.MaxLength.forTextarea(textarea.maxLength()) : Constraint.MaxLength.forTextarea()
             );
         }
+    }
+
+    private static void addPatternConstraint(Field field, List<Constraint> out) {
         if (field instanceof PatternAttribute pat && pat.pattern() != null) {
             out.add(new Constraint.MatchesPattern(pat.pattern()));
         }
+    }
 
-        if (field instanceof OptionsAttribute opt && hasOptions(opt.options())) {
-            if (field instanceof CheckboxGroupField) {
-                out.add(new Constraint.EachOf(opt.options()));
-            } else {
-                out.add(new Constraint.OneOf(opt.options()));
-            }
+    private static void addOptionsConstraint(Field field, List<Constraint> out) {
+        if (field instanceof SingleValueOptionsField single) {
+            addSingleValueOptionsConstraint(single, out);
+        } else if (field instanceof MultiValueOptionsField multi) {
+            addMultiValueOptionsConstraint(multi, out);
         }
+    }
 
-        return out;
+    private static void addSingleValueOptionsConstraint(SingleValueOptionsField field, List<Constraint> out) {
+        var dynamic = field.dynamicOptions();
+        if (dynamic != null) {
+            out.add(Constraint.OneOf.dynamic(dynamic.expression(), dynamic.fallback()));
+            return;
+        }
+        if (hasOptions(field.options())) {
+            out.add(new Constraint.OneOf(field.options()));
+        }
+    }
+
+    private static void addMultiValueOptionsConstraint(MultiValueOptionsField field, List<Constraint> out) {
+        var dynamic = field.dynamicOptions();
+        if (dynamic != null) {
+            out.add(Constraint.EachOf.dynamic(dynamic.expression(), dynamic.fallback()));
+            return;
+        }
+        if (hasOptions(field.options())) {
+            out.add(new Constraint.EachOf(field.options()));
+        }
     }
 
     private static boolean hasOptions(List<String> options) {
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/exception/SubscriptionFormDefinitionValidationException.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/exception/SubscriptionFormDefinitionValidationException.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.subscription_form.exception;
+
+import io.gravitee.apim.core.exception.ValidationDomainException;
+import java.util.List;
+import lombok.Getter;
+
+/**
+ * Thrown when a subscription form definition (GMD template itself) is invalid.
+ *
+ * <p>This is distinct from {@link SubscriptionFormValidationException}, which is used for validating
+ * submitted subscription metadata values against an already valid schema.</p>
+ *
+ * @author Gravitee.io Team
+ */
+@Getter
+public class SubscriptionFormDefinitionValidationException extends ValidationDomainException {
+
+    private final List<String> errors;
+
+    public SubscriptionFormDefinitionValidationException(List<String> errors) {
+        this(errors, null);
+    }
+
+    public SubscriptionFormDefinitionValidationException(List<String> errors, Throwable cause) {
+        super("Subscription form definition is invalid: " + String.join(", ", errors), cause);
+        this.getParameters().put("errors", String.join(", ", errors));
+        this.errors = List.copyOf(errors);
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/exception/SubscriptionFormDefinitionValidationException.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/exception/SubscriptionFormDefinitionValidationException.java
@@ -27,10 +27,14 @@ import lombok.Getter;
  *
  * @author Gravitee.io Team
  */
-@Getter
 public class SubscriptionFormDefinitionValidationException extends ValidationDomainException {
 
+    @Getter
     private final List<String> errors;
+
+    public SubscriptionFormDefinitionValidationException(String error) {
+        this(List.of(error));
+    }
 
     public SubscriptionFormDefinitionValidationException(List<String> errors) {
         this(errors, null);
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/Constraint.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/Constraint.java
@@ -21,6 +21,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.function.BiFunction;
 import java.util.regex.PatternSyntaxException;
 import java.util.stream.Stream;
 
@@ -50,6 +51,8 @@ import java.util.stream.Stream;
         @JsonSubTypes.Type(value = Constraint.MatchesPattern.class, name = "matchesPattern"),
         @JsonSubTypes.Type(value = Constraint.OneOf.class, name = "oneOf"),
         @JsonSubTypes.Type(value = Constraint.EachOf.class, name = "eachOf"),
+        @JsonSubTypes.Type(value = Constraint.OneOf.class, name = "dynamicOneOf"),
+        @JsonSubTypes.Type(value = Constraint.EachOf.class, name = "dynamicEachOf"),
     }
 )
 public sealed interface Constraint
@@ -61,8 +64,7 @@ public sealed interface Constraint
         Constraint.MinLength,
         Constraint.MaxLength,
         Constraint.MatchesPattern,
-        Constraint.OneOf,
-        Constraint.EachOf {
+        Constraint.ResolvableOptions {
     /**
      * Whether the submitted value satisfies this constraint. The value is already trimmed; an empty string means absent.
      */
@@ -211,22 +213,52 @@ public sealed interface Constraint
     }
 
     /** Single value must be one of the allowed options (skipped when empty). */
-    record OneOf(List<String> options) implements Constraint {
+    record OneOf(List<String> options, String expression, List<String> fallback) implements ResolvableOptions {
+        public OneOf(List<String> options) {
+            this(options, null, null);
+        }
+
+        public static OneOf dynamic(String expression, List<String> fallback) {
+            return new OneOf(null, expression, fallback);
+        }
+
         @Override
         public boolean check(String value) {
+            if (isDynamic() && options == null) {
+                throw unresolvedConstraintException();
+            }
             return options == null || options.isEmpty() || value.isEmpty() || options.contains(value);
         }
 
         @Override
         public String formatErrorMessage(String fieldKey, String value) {
+            if (isDynamic() && options == null) {
+                throw unresolvedConstraintException();
+            }
             return String.format("Field '%s': value '%s' is not among the allowed options", fieldKey, value);
+        }
+
+        @Override
+        public Constraint withResolvedOptions(List<String> resolvedOptions) {
+            return new OneOf(resolvedOptions);
         }
     }
 
     /** Every value in a comma-separated string must be one of the allowed options (skipped when empty). */
-    record EachOf(List<String> options) implements Constraint {
+    record EachOf(List<String> options, String expression, List<String> fallback) implements ResolvableOptions {
+        public EachOf(List<String> options) {
+            this(options, null, null);
+        }
+
+        public static EachOf dynamic(String expression, List<String> fallback) {
+            return new EachOf(null, expression, fallback);
+        }
+
         @Override
         public boolean check(String value) {
+            if (isDynamic() && options == null) {
+                throw unresolvedConstraintException();
+            }
             if (options == null || options.isEmpty() || value.isEmpty()) {
                 return true;
             }
@@ -236,6 +268,9 @@ public sealed interface Constraint
 
         @Override
         public String formatErrorMessage(String fieldKey, String value) {
+            if (isDynamic() && options == null) {
+                throw unresolvedConstraintException();
+            }
             if (options == null || options.isEmpty() || value.isEmpty()) {
                 return "";
             }
@@ -245,6 +280,42 @@ public sealed interface Constraint
                 .map(item -> String.format("Field '%s': value '%s' is not among the allowed options", fieldKey, item))
                 .toList();
             return String.join("\n", lines);
+        }
+
+        @Override
+        public Constraint withResolvedOptions(List<String> resolvedOptions) {
+            return new EachOf(resolvedOptions);
+        }
+    }
+
+    /**
+     * Contract for option constraints that are resolved at runtime.
+     *
+     * <p>Implementations may store an expression and fallback options.
+     * They must be resolved to effective options via {@link #resolve(BiFunction)} before validation.</p>
+     */
+    non-sealed interface ResolvableOptions extends Constraint {
+        String expression();
+
+        List<String> fallback();
+
+        Constraint withResolvedOptions(List<String> resolvedOptions);
+
+        default boolean isDynamic() {
+            return expression() != null;
+        }
+
+        default Constraint resolve(BiFunction<String, List<String>, List<String>> resolver) {
+            if (!isDynamic()) {
+                return this;
+            }
+            return withResolvedOptions(resolver.apply(expression(), fallback()));
+        }
+
+        default UnsupportedOperationException unresolvedConstraintException() {
+            return new UnsupportedOperationException(
+                getClass().getSimpleName() + " must be resolved before validation — call resolve() first"
+            );
         }
     }
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/Constraint.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/Constraint.java
@@ -15,13 +15,16 @@
  */
 package io.gravitee.apim.core.subscription_form.model;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import java.util.Arrays;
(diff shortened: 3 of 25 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormConstraintsFactory.java
```diff
@@ -61,17 +62,27 @@ private static List<Constraint> constraintsFor(Field field) {
...
-            } else if (field instanceof CheckboxGroupField) {
-                out.add(new Constraint.NonEmptySelection());
-            } else {
-                out.add(new Constraint.Required());
-            }
+    private static void addRequiredConstraint(Field field, List<Constraint> out) {
+        if (!field.required()) {
+            return;
         }
+        if (field instanceof CheckboxField) {
```
dev-marek: WDYT about separate interfaces (RequiredTrue, RequiredSelection, RequiredValue) with the same field required() but used differently on different field types?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormConstraintsFactory.java
```diff
@@ -82,19 +93,28 @@ private static List<Constraint> constraintsFor(Field field) {
...
+
+    private static void addPatternConstraint(Field field, List<Constraint> out) {
         if (field instanceof PatternAttribute pat && pat.pattern() != null) {
             out.add(new Constraint.MatchesPattern(pat.pattern()));
         }
+    }
 
-        if (field instanceof OptionsAttribute opt && hasOptions(opt.options())) {
+    private static void addOptionsConstraint(Field field, List<Constraint> out) {
+        if (field instanceof DynamicOptionsAttribute dyn && dyn.dynamicOptions() != null) {
```
dev-marek: Similar as above comment: https://github.com/gravitee-io/gravitee-api-management/pull/16065/changes#r3015713481

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/Constraint.java
```diff
@@ -248,6 +252,42 @@ public String formatErrorMessage(String fieldKey, String value) {
...
     }
 
+    /**
+     * Placeholder for a single-value option constraint whose allowed set is resolved at runtime from an EL expression.
+     *
+     * <p>This constraint is stored in the database and must be pre-resolved to a {@link OneOf} constraint
+     * (using the expression against the target API's metadata) before being passed to the validator.
+     * Its {@link #check(String)} throws {@link IllegalStateException} — it must never be evaluated directly.</p>
+     */
+    record DynamicOneOf(String expression, List<String> fallback) implements Constraint {
```
dev-marek: For those two I'd create a `ResolvableOptions` class implementing `Constraint` and implementations `DynamicOneOf` and `DynamicEachOf`.
`ResolvableOptions` stores expression and options (`fallback` as default ones)
Method `resolve(Function<String, List<String>> resolver)` will be used to update list.
Method `check` and `formatErrorMessage` would be real implementations.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/Constraint.java
```diff
@@ -248,6 +252,42 @@ public String formatErrorMessage(String fieldKey, String value) {
...
+        @Override
+        public boolean check(String value) {
+            throw new IllegalStateException("DynamicOneOf must be resolved before validation — call resolveConstraints() first");
```
dev-marek: Even if, `UnsupportedOperationException`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormElResolverDomainServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+            default -> constraint;
+        };
+    }
+
+    private List<String> resolve(String expression, List<String> fallback, Map<String, Object> templateParams) {
+        if (templateParams.isEmpty()) {
+            return fallback;
+        }
+        try {
+            String resolved = templateResolver.resolveTemplate("${" + expression + "}", templateParams);
```
dev-marek: I was about sure that we'll use Gravitee Expression Language, not Freemaker Template. We should document it at the same level as https://documentation.gravitee.io/apim/gravitee-expression-language

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormElResolverDomainServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+            if (resolved == null || resolved.isBlank()) {
```
dev-marek: we should check against trimmed resolved

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormSchemaGeneratorImpl.java
```diff
@@ -48,6 +51,12 @@
 @Service
 public class SubscriptionFormSchemaGeneratorImpl implements SubscriptionFormSchemaGenerator {
 
+    /**
+     * Matches options values that contain an EL expression: {@code ${expression}:fallback1,fallback2}
+     * Group 1 = the Freemarker expression (inside {@code ${ }}), group 2 = the fallback list (may be empty).
+     */
+    private static final Pattern EL_OPTIONS_PATTERN = Pattern.compile("^\\$\\{(.+)}:(.*)$", Pattern.DOTALL);
```
dev-marek: Let's put freemaker expression together with `${}` so we won't need to add it later. And actually it's weird a lot that it's in a single attribute. I'd put fallback in options attribute and EL in dynamicOptions

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormSchemaGeneratorImpl.java
JedrzejJanasiak: On this one, I’d keep the current approach.

For storing the expression with ${...} / {#...} included: today we intentionally store only the metadata reference in the database, without leaking template-engine-specific details into persistence. That gives us flexibility later. If we switch engines in the future (or support multiple syntaxes), we avoid data migrations. Stripping/adding the wrapper at runtime is a very small cost, while the long-term benefit is significant.

For the second suggestion (two separate attributes): this was already discussed earlier with Varun as a product requirement, and we agreed on the current single-attribute syntax. I would not reopen that decision in this PR. It would cost time with little value right now.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/Constraint.java
```diff
@@ -248,6 +252,60 @@ public String formatErrorMessage(String fieldKey, String value) {
...
+        String expression();
+
+        List<String> fallback();
+
+        Constraint resolve(Function<String, List<String>> resolver);
+
+        @Override
+        default boolean check(String value) {
+            throw new UnsupportedOperationException(
+                getClass().getSimpleName() + " must be resolved before validation — call resolveConstraints() first"
```
dev-marek: resolve*

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/Constraint.java
```diff
@@ -248,6 +252,60 @@ public String formatErrorMessage(String fieldKey, String value) {
...
+     *
+     * <p>Implementations store an expression and fallback options. They must be resolved to static variants via
+     * {@link #resolve(Function)} before validation.</p>
+     */
+    sealed interface ResolvableOptions extends Constraint permits DynamicOneOf, DynamicEachOf {
+        String expression();
+
+        List<String> fallback();
+
+        Constraint resolve(Function<String, List<String>> resolver);
```
dev-marek: Why resolving into different constraint instead of just updating options here in abstract class and implementing `check`s for implementations?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/use_case/GetSubscriptionFormForEnvironmentUseCase.java
```diff
@@ -41,10 +52,25 @@ public Output execute(Input input) {
             throw new SubscriptionFormNotFoundException(input.environmentId());
         }
 
-        return new Output(subscriptionForm);
+        var schema = schemaGenerator.generate(subscriptionForm.getGmdContent());
+        var resolvedOptions = input.apiId() != null
+            ? elResolver.resolveSchemaOptions(schema, input.environmentId(), input.apiId())
+            : elResolver.resolveSchemaOptions(schema);
```
dev-marek, line 59: I see this case always returns `Map.of` but I guess it would be better to evaluate options even without environment and api contexts because some EL can be evaluated without them.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/use_case/UpdateSubscriptionFormUseCase.java
```diff
@@ -62,7 +62,12 @@ public Output execute(Input input) {
             );
 
         var gmd = GraviteeMarkdown.of(input.gmdContent());
-        var schema = schemaGenerator.generate(gmd);
+        SubscriptionFormSchema schema;
+        try {
+            schema = schemaGenerator.generate(gmd);
+        } catch (IllegalArgumentException e) {
+            throw new SubscriptionFormValidationException(List.of(e.getMessage()));
```
dev-marek: I see this Exception is used for submitted form validation. For validation errors regarding form itself, I'd create a specific exception.
Also, `e.getMessage()` causes to loose stack trace from `e`. If you would ever happen to touch `SubscriptionFormValidationException`, I'd change signature to `List<Throwable>` and attach last as a cause for better error tracing.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormElResolverDomainServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+    private final ApiMetadataQueryService apiMetadataQueryService;
+
+    @Override
+    public Map<String, List<String>> resolveSchemaOptions(SubscriptionFormSchema schema, String environmentId, String apiId) {
+        return resolveSchemaOptionsWithParams(schema, buildTemplateParams(environmentId, apiId));
+    }
+
+    @Override
+    public Map<String, List<String>> resolveSchemaOptions(SubscriptionFormSchema schema) {
+        return Map.of();
```
dev-marek: if we're returning plain `Map.of()`, I'd make it at least a default implementation on interface level

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormSchemaGeneratorImpl.java
dev-marek: 2nd suggestion - ok
1st - on database layer we should store full expression because for the time being it would be an expression. in future we may add expression language type field for persistence, but still we would keep it as a fallback option and won't do migration. we should avoid strange packing/unpacking of expressions if it comes with no real benefit

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormSchemaGeneratorImpl.java
```diff
@@ -114,8 +132,31 @@ private String parseReadonlyValue(Element el, String fieldKey) {
...
+        if (!el.hasAttr("options")) {
+            return new ParsedOptions(null, null);
+        }
+        String raw = el.attr("options").trim();
+
+        // Check if the whole value starts with {# — could be EL or a misconfigured EL without fallback.
+        if (raw.startsWith("{#")) {
+            Matcher m = EL_OPTIONS_PATTERN.matcher(raw);
+            if (!m.matches()) {
+                throw new IllegalArgumentException(
```
dev-marek: Please replace this generic exception with domain one

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
```diff
@@ -75,8 +77,15 @@ private void validateSubscriptionFormMetadataIfApplicable(
             .map(SubscriptionForm::getValidationConstraints)
             .filter(constraints -> !constraints.isEmpty())
             .ifPresent(constraints -> {
+                var resolved = GenericPlanEntity.ReferenceType.API.equals(genericPlanEntity.getReferenceType())
+                    ? subscriptionFormElResolver.resolveConstraints(
+                        constraints,
+                        genericPlanEntity.getEnvironmentId(),
+                        genericPlanEntity.getReferenceId()
+                    )
+                    : subscriptionFormElResolver.resolveConstraints(constraints);
```
dev-marek: this can be extracted to a method for better clarity of what happens in this chain, eg:

    private SubscriptionFormFieldConstraints planBasedElResolver(GenericPlanEntity plan, SubscriptionFormFieldConstraints constraints) {
        if (plan.getReferenceType() == API) {
            return subscriptionFormElResolver.resolveConstraints(constraints, plan.getEnvironmentId(), plan.getReferenceId());
        }
        return subscriptionFormElResolver.resolveConstraints(constraints);
    }

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
```diff
@@ -75,8 +77,15 @@ private void validateSubscriptionFormMetadataIfApplicable(
...
                 var submitted = metadata != null ? metadata : Map.<String, String>of();
```
dev-marek: In functional approach, I'd like to see it before chain:

`var submitted = ofNullable(metadata).orElseGet(Map::of);`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImpl.java
```diff
@@ -75,8 +77,15 @@ private void validateSubscriptionFormMetadataIfApplicable(
...
-                new SubscriptionFormSubmissionValidator(constraints).validate(submitted);
+                new SubscriptionFormSubmissionValidator(resolved).validate(submitted);
```
dev-marek: It would look better chained:

```
            .map(constraints -> planBasedElResolver(genericPlanEntity, constraints))
            .map(SubscriptionFormSubmissionValidator::new)
            .ifPresent(validator -> validator.validate(submitted));
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormSchemaGeneratorImpl.java
```diff
@@ -114,8 +132,31 @@ private String parseReadonlyValue(Element el, String fieldKey) {
         return readonly ? el.attr("value").trim() : null;
     }
 
-    private List<String> parseOptions(Element el) {
-        return el.hasAttr("options") ? parseCommaList(el.attr("options")) : null;
+    private record ParsedOptions(List<String> staticOptions, DynamicOptions dynamicOptions) {}
```
dev-marek: In this case only `staticOptions` or `dynamicOptions` is set. I suggest to create two separate records so that we won't need to handle null values and bind them with interface.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormConstraintsFactoryTest.java
```diff
@@ -52,7 +53,7 @@ void should_map_readonly_input_to_single_read_only_constraint() {
 
     @Test
     void should_ignore_required_and_options_when_radio_is_readonly() {
-        var schema = new SubscriptionFormSchema(List.of(new RadioField("plan", true, "Free", List.of("Free", "Pro"))));
+        var schema = new SubscriptionFormSchema(List.of(new RadioField("plan", true, "Free", List.of("Free", "Pro"), null)));
```
dev-marek: Annoying nulls in question https://github.com/gravitee-io/gravitee-api-management/pull/16065/changes#r3021313776

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormConstraintsFactoryTest.java
```diff
@@ -131,11 +132,43 @@ void should_preserve_user_max_length_when_within_system_limit() {
...
         assertThat(constraints.byFieldKey().get("tags")).containsExactly(
             new Constraint.NonEmptySelection(),
             new Constraint.EachOf(List.of("A", "B"))
         );
     }
+
+    @Test
+    void should_emit_dynamic_one_of_for_select_with_dynamic_options() {
+        var dynamic = new DynamicOptions("api.metadata['envs']", List.of("Prod", "Test"));
+        var schema = new SubscriptionFormSchema(List.of(new SelectField("env", false, null, dynamic)));
```
dev-marek: WDYT about creating SubscriptionFormSchema fixture?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/subscription_form/domain_service/SubscriptionFormSubmissionValidatorTest.java
```diff
@@ -98,7 +98,7 @@ void should_not_throw_when_required_checkbox_is_checked() {
 
         @Test
         void should_throw_when_required_checkbox_group_has_only_separators() {
-            var schema = schema(new CheckboxGroupField("tags", true, List.of("A", "B")));
+            var schema = schema(new CheckboxGroupField("tags", true, List.of("A", "B"), null));
```
dev-marek: nulls-continued

dev-marek (changes requested):
some more mostly about styling

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/use_case/UpdateSubscriptionFormUseCase.java
```diff
@@ -75,7 +75,7 @@ public Output execute(Input input) {
 
     private void validateFieldCount(SubscriptionFormSchema schema) {
         if (schema != null && schema.fields().size() > SubscriptionFormSubmissionValidator.MAX_METADATA_COUNT) {
-            throw new SubscriptionFormValidationException(
+            throw new SubscriptionFormDefinitionValidationException(
```
dev-marek, line 77: WDYT creating separate constructor only with message?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormElResolverDomainServiceImpl.java
```diff
@@ -0,0 +1,156 @@
...
+ * and are evaluated against the target API's metadata. On resolution failure the configured fallback list is used transparently.</p>
+ *
+ * @author Gravitee.io Team
+ */
+@Service
+@RequiredArgsConstructor
+@CustomLog
+public class SubscriptionFormElResolverDomainServiceImpl implements SubscriptionFormElResolverDomainService {
+
+    private final TemplateResolverDomainService templateResolver;
```
dev-marek: Still looks like Freemarker wiring.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/Constraint.java
```diff
@@ -246,6 +281,42 @@ public String formatErrorMessage(String fieldKey, String value) {
...
+        }
+    }
+
+    /**
+     * Contract for option constraints that are resolved at runtime.
+     *
+     * <p>Implementations may store an expression and fallback options.
+     * They must be resolved to effective options via {@link #resolve(BiFunction)} before validation.</p>
+     */
+    non-sealed interface ResolvableOptions extends Constraint {
```
dev-marek: I was thinking more about something like that:

```
    record OneOf(List<String> options) implements Constraint {
        public static DynamicOneOf dynamic(String expression, List<String> fallback) {
            return new DynamicOneOf(expression, fallback);
        }

        @Override
        public boolean check(String value) {
            return options == null || options.isEmpty() || value.isEmpty() || options.contains(value);
        }

        @Override
        public String formatErrorMessage(String fieldKey, String value) {
            return String.format("Field '%s': value '%s' is not among the allowed options", fieldKey, value);
        }
    }
    record EachOf(List<String> options) implements Constraint {
        public static DynamicEachOf dynamic(String expression, List<String> fallback) {
            return new DynamicEachOf(expression, fallback);
        }

        @Override
        public boolean check(String value) {
            if (options == null || options.isEmpty() || value.isEmpty()) {
                return true;
            }
            Set<String> allowed = new HashSet<>(options);
            return splitCsv(value).allMatch(allowed::contains);
        }

        @Override
        public String formatErrorMessage(String fieldKey, String value) {
            if (options == null || options.isEmpty() || value.isEmpty()) {
                return "";
            }
            Set<String> allowed = new HashSet<>(options);
            var lines = splitCsv(value)
                .filter(item -> !allowed.contains(item))
                .map(item -> String.format("Field '%s': value '%s' is not among the allowed options", fieldKey, item))
                .toList();
            return String.join("\n", lines);
        }
    }

    final class DynamicOneOf extends ResolvableOptions {
        public DynamicOneOf(@NotNull String expression, List<String> fallback) {
            super(expression, fallback);
        }

        @Override
        public boolean check(String value) {
            return options == null || options.isEmpty() || value.isEmpty() || options.contains(value);
        }

        @Override
        public String formatErrorMessage(String fieldKey, String value) {
            return String.format("Field '%s': value '%s' is not among the allowed options", fieldKey, value);
        }
    }

    final class DynamicEachOf extends ResolvableOptions {
        public DynamicEachOf(@NotNull String expression, List<String> fallback) {
            super(expression, fallback);
        }

        @Override
        public boolean check(String value) {
            if (options == null || options.isEmpty() || value.isEmpty()) {
                return true;
            }
            Set<String> allowed = new HashSet<>(options);
            return splitCsv(value).allMatch(allowed::contains);
        }

        @Override
        public String formatErrorMessage(String fieldKey, String value) {
            if (options == null || options.isEmpty() || value.isEmpty()) {
                return "";
            }
            Set<String> allowed = new HashSet<>(options);
            var lines = splitCsv(value)
                .filter(item -> !allowed.contains(item))
                .map(item -> String.format("Field '%s': value '%s' is not among the allowed options", fieldKey, item))
                .toList();
            return String.join("\n", lines);
        }
    }


    sealed abstract class ResolvableOptions implements Constraint {
        @Getter
        private final String expression;
        protected List<String> options;
        @Getter
        private final List<String> fallback;
        private boolean resolved;

        protected ResolvableOptions(@NotNull String expression, List<String> fallback) {
            this.expression = expression;
            this.options = fallback;
            this.fallback = fallback;
            this.resolved = false;
        }

        public void resolve(Function<String, List<String>> resolver) {
            if (resolved) {
                throw new IllegalStateException("Dynamic constraint already resolved");
            }
            this.options = resolver.apply(expression);
            this.resolved = true;
        }
    }
```

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/exception/SubscriptionFormDefinitionValidationException.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Thrown when a subscription form definition (GMD template itself) is invalid.
+ *
+ * <p>This is distinct from {@link SubscriptionFormValidationException}, which is used for validating
+ * submitted subscription metadata values against an already valid schema.</p>
+ *
+ * @author Gravitee.io Team
+ */
+@Getter
```
dev-marek: I prefer setting annotations just per field, but that's ok

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/GraviteeElSubscriptionFormExpressionResolverDomainServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+@CustomLog
+public class GraviteeElSubscriptionFormExpressionResolverDomainServiceImpl implements SubscriptionFormExpressionResolverDomainService {
+
+    @Override
+    public List<String> resolveToOptions(String expression, Map<String, Object> templateParams) {
+        try {
+            var templateEngine = TemplateEngine.templateEngine();
+            templateParams.forEach((key, value) -> templateEngine.getTemplateContext().setVariable(key, value));
+
+            String resolved = templateEngine.evalNow(expression, String.class);
```
dev-marek, line 36: I'm not sure if we'll always get `String`. Dictionaries are usually maps represented as objects

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription_form/model/Constraint.java
```diff
@@ -225,27 +236,177 @@ public String formatErrorMessage(String fieldKey, String value) {
...
+        public List<String> fallback() {
+            return fallback;
+        }
+
+        @JsonProperty("options")
+        public List<String> options() {
+            return options;
+        }
+
+        public final Constraint resolve(Function<String, List<String>> resolver) {
```
dev-marek: Please just make it mutable, with serializable static fields expression and fallback and stateful fields resolved and options

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/subscription_form/SubscriptionFormElResolverDomainServiceImpl.java
```diff
@@ -0,0 +1,150 @@
...
+        for (Map.Entry<String, List<Constraint>> entry : constraints.byFieldKey().entrySet()) {
+            resolved.put(entry.getKey(), resolveConstraintList(entry.getValue(), templateParams));
+        }
+        return new SubscriptionFormFieldConstraints(resolved);
+    }
+
+    private List<Constraint> resolveConstraintList(List<Constraint> constraints, Map<String, Object> templateParams) {
+        return constraints
+            .stream()
+            .map(constraint -> resolveConstraint(constraint, templateParams))
```
dev-marek: Here I'd like to avoid replacing constraints and just resolving them

dev-marek (changes requested):

dev-marek (approved):

PR #16067, opened 2026-03-27
fix(gateway): resolve dynamic routing endpoint reference misidentified as absolute URL

thomasballivet:
## Issues

https://gravitee.atlassian.net/browse/APIM-13317
https://gravitee.atlassian.net/browse/APIM-13332

## Description

`{#endpoints['name']}/{#group[0]}` evaluates to `name://path `when the capture starts with `/`, which `URIUtils.isAbsolute()` incorrectly matches. The fix uses a known URL scheme allowlist (`http, https, ws, wss`) to disambiguate.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/invoker/HttpEndpointInvoker.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/invoker/HttpEndpointInvoker.java
@@ -21,7 +21,6 @@ import static io.gravitee.gateway.reactive.api.context.InternalContextAttributes
 
 import io.gravitee.common.http.HttpMethod;
 import io.gravitee.common.http.HttpStatusCode;
-import io.gravitee.common.util.URIUtils;
 import io.gravitee.gateway.reactive.api.ExecutionFailure;
 import io.gravitee.gateway.reactive.api.connector.endpoint.EndpointConnector;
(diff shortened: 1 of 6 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/invoker/HttpEndpointInvoker.java
```diff
@@ -106,16 +109,16 @@ private <T extends HttpEndpointConnector> T resolveConnector(final HttpExecution
...
-                if (matcher.matches()) {
-                    // Set endpoint name into the criteria.
-                    endpointCriteria.setName(matcher.group(MATCH_GROUP_ENDPOINT));
-
-                    // Replace the attribute to remove the endpoint reference part ("my-endpoint:/foo/bar' -> '/foo/bar').
+            final Matcher matcher = ENDPOINT_PATTERN.matcher(evaluatedTarget);
+
+            if (matcher.matches()) {
+                final String endpointName = matcher.group(MATCH_GROUP_ENDPOINT);
+                if (URIUtils.isAbsolute(evaluatedTarget) && KNOWN_URL_SCHEMES.contains(endpointName.toLowerCase())) {
```
araffass: The fix looks good overall.

Minor suggestion: i think the URIUtils.isAbsolute(evaluatedTarget) seems a bit redundant. Since the regex already parsed the prefix, checking whether `endpointName` is in the known URL scheme set may be enough and would simplify the logic.

araffass (approved):

PR #16082, opened 2026-03-30
Fix/gko 2780 toaster custom colors are ignored

SamHardi:
## Issue

https://gravitee.atlassian.net/browse/GKO-2780

## Description

Fix color of the toaster.

## Additional context

Screenshot 2026-03-30 at 11 15 13
Screenshot 2026-03-30 at 11 14 51


jourdiw (changes requested):
Use the [mixins from Material](https://material.angular.dev/components/snack-bar/styling) so that we do not run into this same problem again 😎 👍

jourdiw (approved):

PR #16083, opened 2026-03-30
feat: introduce a flag to enable Gamma

jgiovaresco:
## Issue

## Description

Disabled by default, it prevents loading Gamma modules plugins and
enabling Gamma rest API.

This PR also configures the CI to build the new Gamma Console.
It also updates the release pipeline to release it.


jhaeyaert (approved):

.circleci/ci/src/workflows/workflow-full-release.ts
phiz71: You should depend also on Gamma console here. Wdyt?

helm/values.yaml
```diff
@@ -950,6 +953,20 @@ api:
           secretName: apim.example.com
       annotations:
         kubernetes.io/ingress.class: nginx
+    gamma:
+      enabled: false
```
phiz71, line 962: duplicate with the one line 55 ?

jgiovaresco: It is not the same. This one is used to "disable" ingress for the GAMMA API. Like we do with MAPI and Portal API 🤔

phiz71 (approved):

PR #16087, opened 2026-03-30
fix: increase flow_selectors path limit to 2048 for JDBC

araffass:
## Issue

https://gravitee.atlassian.net/browse/APIM-12488

## Description

This PR fixes a DataIntegrityViolationException occurring during API imports (Swagger/OData) when using a PostgreSQL/JDBC backend. The issue was caused by a strict VARCHAR(256) limit on the flow_selectors.path column, which is insufficient for complex or nested resource paths.

## Additional context

Database Migration: Added a Liquibase script to increase the flow_selectors.path column size to 2048 characters.


phiz71 (changes requested):
you should use 4.12.0 only if you plan. to apply this fix on master only.
But I guess you are going to apply on lower version.
You should do this improvement on the lowest supported version and then apply on upper branches.

PR #16089, opened 2026-03-30
fix(portal-next): APIM-13336 decouple subscription details from API documentation

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/APIM-13336

## Description

This PR makes Subscription Details resilient to missing API documentation in Portal Next by decoupling core data loading from view=documentation checks.
When documentation is unavailable (404), the page still renders subscription data and the “Subscribed API” item is shown as plain text instead of a clickable link

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.html
+++ b/gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.html
@@ -30,6 +30,7 @@
         [subscription]="detailsVM.result.subscription"
         [planUsageConfiguration]="detailsVM.result.planUsageConfiguration"
         [api]="detailsVM.result.api"
+        [apiName]="detailsVM.result.apiName"
         [consumerStatus]="detailsVM.result.consumerStatus"
         [failureCause]="detailsVM.result.failureCause"
         [createdAt]="detailsVM.result.createdAt"
@@ -38,16 +39,18 @@
         (resumeConsumerStatus)="resumeConsumerStatus()"
         (closeSubscription)="closeSubscription()"
       />
-      <app-api-access
-        [planSecurity]="detailsVM.result.planSecurity"
-        [subscription]="detailsVM.result.subscription"
-        [entrypointUrls]="detailsVM.result.entrypointUrls"
-        [apiKey]="detailsVM.result.apiKey"
-        [clientId]="detailsVM.result.clientId"
-        [clientSecret]="detailsVM.result.clientSecret"
-        [apiType]="detailsVM.result.api?.type"
-        [apiKeyConfigUsername]="detailsVM.result.apiKeyConfigUsername"
-      />
+      @if (detailsVM.result.api) {
+        <app-api-access
+          [planSecurity]="detailsVM.result.planSecurity"
+          [subscription]="detailsVM.result.subscription"
+          [entrypointUrls]="detailsVM.result.entrypointUrls"
+          [apiKey]="detailsVM.result.apiKey"
+          [clientId]="detailsVM.result.clientId"
+          [clientSecret]="detailsVM.result.clientSecret"
+          [apiType]="detailsVM.result.api.type"
+          [apiKeyConfigUsername]="detailsVM.result.apiKeyConfigUsername"
+        />
+      }
     </div>
   }
 
--- a/gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.ts
+++ b/gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.ts
@@ -15,6 +15,7 @@
  */
 
 import { AsyncPipe } from '@angular/common';
+import { HttpErrorResponse } from '@angular/common/http';
 import { Component, DestroyRef, Input, OnInit } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { FormsModule, ReactiveFormsModule } from '@angular/forms';
@@ -22,7 +23,7 @@ import { MatCardModule } from '@angular/material/card';
 import { MatDialog } from '@angular/material/dialog';
 import { MatFormFieldModule } from '@angular/material/form-field';
 import { MatInputModule } from '@angular/material/input';
-import { BehaviorSubject, catchError, filter, forkJoin, map, Observable, switchMap, tap } from 'rxjs';
+import { BehaviorSubject, catchError, filter, forkJoin, map, Observable, switchMap, tap, throwError } from 'rxjs';
 import { of } from 'rxjs/internal/observable/of';
 
 import { SubscriptionConsumerConfigurationComponent } from './subscription-consumer-configuration';
@@ -32,7 +33,6 @@ import { LoaderComponent } from '../../../../../components/loader/loader.compone
 import { SubscriptionInfoComponent } from '../../../../../components/subscription-info/subscription-info.component';
 import { Api } from '../../../../../entities/api/api';
 import { Application } from '../../../../../entities/application/application';
-import { UserApiPermissions } from '../../../../../entities/permission/permission';
 import { PlanMode, PlanSecurityEnum, PlanUsageConfiguration } from '../../../../../entities/plan/plan';
 import {
   SubscriptionConsumerStatusEnum,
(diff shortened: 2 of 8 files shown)


gravitee-apim-portal-webui-next/src/services/permissions.service.ts
```diff
@@ -30,9 +30,14 @@ export class PermissionsService {
     private configService: ConfigService,
   ) {}
 
-  getApiPermissions(apiId: string): Observable<UserApiPermissions> {
+  getApiPermissions(apiId: string, documentationView = true): Observable<UserApiPermissions> {
+    let params = new HttpParams().set('apiId', apiId);
+    if (documentationView) {
+      params = params.set(PortalApiViewParam.QUERY_PARAM_NAME, PortalApiViewParam.DOCUMENTATION);
+    }
+
```
kovaren: This approach will not work if we have an API that has a published nav item (new visibility rules) but is not published in the Console (old visibility rules). The subscription details will break.

We cannot use this endpoint without `view=documentation` in the Next Gen Portal because it makes the visibility dependent on the old `published` status of the API.

I don't know if there's another solution to this. We might try to just hide the API label completely from the subscription details page if the API is unpublished, but we might need other information from the API object. But if such a solution requires too many changes to the SubscriptionsDetailsComponent, I don't think it's worth it and we should rather follow the logic of the Classic portal and show the API to the users that are subscribed to it even if it's unpublished. In this case, this will be a backend change.

kovaren (changes requested):

gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.ts
```diff
@@ -163,22 +165,25 @@ export class SubscriptionsDetailsComponent implements OnInit {
...
+      switchMap(({ subscription, permissions, hasDocumentationAccess }) =>
         forkJoin({
           subscription: of(subscription),
           plan: this.getPlanData$(subscription.plan, permissions),
-          api: this.apiService.details(this.apiId),
+          api: this.apiService.details(this.apiId, false),
           application: this.applicationService.get(subscription.application),
+          hasDocumentationAccess: of(hasDocumentationAccess),
         }),
       ),
```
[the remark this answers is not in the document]
jaroslawlakomy-grav: outdated

gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.html
```diff
@@ -48,14 +48,25 @@ <h1 class="next-gen-h3" i18n="@@subscriptionDetailsHeader">Subscription Details<
...
-            <a class="internal-link next-gen-body" [routerLink]="['/catalog/api/' + api.id]">{{ api.name }}</a>
+            @if (hasDocumentationAccess && resolvedApiName && api && api.id) {
+              <a data-testid="subscription-api-link" class="internal-link next-gen-body" [routerLink]="['/catalog/api/' + api.id]">
+                {{ resolvedApiName }}
+              </a>
+            } @else {
+              <span
+                data-testid="subscription-api-label"
+                class="next-gen-body"
+                matTooltip="The API has been unpublished from the Developer Portal"
```
kovaren, line 66: Could you make sure that the tooltip is translatable? Either with an html attribute or a `$localize` template.

gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.html
```diff
@@ -48,14 +48,25 @@ <h1 class="next-gen-h3" i18n="@@subscriptionDetailsHeader">Subscription Details<
...
-
-        @if (api) {
+        @if (resolvedApiName) {
           <div class="subscription-info__content-row">
             <span class="next-gen-body">
               <span class="material-icons icon-small">api</span>
               <span i18n="@@subscriptionDetailsHeaderApi">Subscribed API:</span>
             </span>
-            <a class="internal-link next-gen-body" [routerLink]="['/catalog/api/' + api.id]">{{ api.name }}</a>
+            @if (hasDocumentationAccess && resolvedApiName && api && api.id) {
```
kovaren: WDYT about simplifying this a little bit?

```suggestion
            @if (api && api.id) {

File: gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.ts [hunk 5]
```diff
@@ -158,35 +160,51 @@ export class SubscriptionsDetailsComponent implements OnInit {
...
-      switchMap(({ subscription, permissions }) =>
+      switchMap(({ subscription, hasDocumentationAccess }) =>
         forkJoin({
           subscription: of(subscription),
-          plan: this.getPlanData$(subscription.plan, permissions),
-          api: this.apiService.details(this.apiId),
+          plan: this.getPlanData$(subscription.plan),
+          api: this.apiService.details(this.apiId).pipe(catchError(() => of(null))),
```
kovaren: We can use `hasDocumentationAccess` upfront to avoid catching the guaranteed error.

```suggestion
          api: hasDocumentationAccess ? this.apiService.details(this.apiId) : of(null),

File: gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.html [hunk 6]
```diff
@@ -30,24 +30,28 @@
         [subscription]="detailsVM.result.subscription"
         [planUsageConfiguration]="detailsVM.result.planUsageConfiguration"
         [api]="detailsVM.result.api"
+        [apiName]="detailsVM.result.apiName"
         [consumerStatus]="detailsVM.result.consumerStatus"
+        [hasDocumentationAccess]="detailsVM.result.hasDocumentationAccess"
```
kovaren: Do we really need to pass this param down to the subscription-info or could we just check if `api` is null inside subscription-info?

gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.ts
```diff
@@ -52,6 +62,12 @@ export class SubscriptionInfoComponent implements OnInit {
   @Input()
   api?: Api;
 
+  @Input()
+  apiName?: string;
+
+  @Input()
+  hasDocumentationAccess?: boolean;
+
```
kovaren: As per https://github.com/gravitee-io/gravitee-api-management/pull/16089/changes#r3027115067

```suggestion

kovaren (changes requested):

gravitee-apim-portal-webui-next/src/services/permissions.service.ts
jaroslawlakomy-grav: refactored as suggested,  view=documentation always used, updated details template to avoid referring to `api` object

gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.html
jaroslawlakomy-grav: i18n-matTooltip added

gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.html
jaroslawlakomy-grav: simplified

gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.html
jaroslawlakomy-grav: hasDocumentationAccess removed

gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.ts
jaroslawlakomy-grav: hasDocumentationAccess removed

gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.ts
jaroslawlakomy-grav: hasDocumentationAccess used to avoid 404

kovaren (approved):

PR #16092, opened 2026-03-30
fix: remove padding and make redoc viewer full width.

kovaren:
## Issue

## Description

Fix Redoc viewer styles:
- Remove global padding and handle it on a case-by-case basis
- Stretch the redoc viewer to the full width of the screen 

https://github.com/user-attachments/assets/52a4af25-3143-45bf-84a7-fd22ff633d88


aqibmohammadkhan (commented):
LGTM.
minor issue - The value 32px appears in 3 new locations as hardcoded. The layout module already defines `$container-padding-horizontal.` 
Using the SCSS variable would be more maintainable, wdyt?

aqibmohammadkhan (approved):

PR #16111, opened 2026-03-30
fix(gateway): handle org-level flow interruptions

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-13351

## Description

Organization-level policies triggering `State.FAILURE` (e.g., Groovy) returned a generic 500 instead of the custom error code. The `InterruptionFailureException` from org response flows was not caught before `handleUnexpectedError`, which overwrote the failure details. Added a dedicated error handler for org flow chains in both `SyncApiReactor` and `DefaultApiReactor`.


mukul-tyagi08 (approved):
LGTM

PR #16141, opened 2026-03-31
fix: apply NG portal visibility rules to /apis/_search when view=documentation

JedrzejJanasiak:
## Issue
https://gravitee.atlassian.net/browse/APIM-13315

## Description
The subscriptions page API filter dropdown was using classic portal visibility rules
(`lifecycle=PUBLISHED`) instead of NG portal rules (published navigation item).

- `POST /apis/_search` now accepts `view=documentation` query parameter — when set,
  delegates to `SearchApisForPortalUseCase` which filters by published portal navigation
  items instead of API lifecycle state
- `ApiService.search()` in portal-next always sends `view=documentation` (hardcoded —
  the service lives exclusively in the NG portal context)

## Additional context

First part of the video shows that none of my APIs are published for the classic portal :) 

https://github.com/user-attachments/assets/0f81c23c-1af3-436a-a208-a51772d48cf9

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/subscriptions/subscriptions.component.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/subscriptions/subscriptions.component.spec.ts
@@ -326,6 +326,16 @@ describe('SubscriptionsComponent', () => {
     c.verify();
   });
 
+  it('should send view=documentation when loading API filter options', async () => {
+    await setup();
+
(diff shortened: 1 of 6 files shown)


dev-marek (changes requested):
1. please do not kill SearchApisForPortalUseCase with fire
2. in this use case SearchApisForPortalUseCase is correct

dev-marek (approved):

kovaren (approved):
Please add a small comment in the `SearchApisForPortalUseCase` for the visibility rules. 

Smth like this:

`This use case has to be in sync with the visibility rules in GetVisiblePortalNavigationApisUseCase`.

PR #16149, opened 2026-04-01
fix: Unable to resend or retrigger expired sign-up confirmation links

schehade-gravitee:
https://gravitee.atlassian.net/browse/APIM-13386


gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/UserEntity.java
```diff
@@ -104,30 +72,25 @@ public class UserEntity implements Indexable {
...
 
     @JsonProperty("number_of_active_tokens")
     private int nbActiveTokens;
 
     private Boolean newsletterSubscribed;
+    private Boolean isServiceAccount;
 
     private Map<String, Object> customFields;
 
+    private boolean hasPassword;
```
phiz71: is it required ?

ThibaudAV (approved):

PR #16152, opened 2026-04-01
fix(portal): "cancel" and "back to API" buttons navigation bugfix for ng portal subscribe flow

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-13346

## Description

Fixed two navigation bugs in `DocumentationSubscribeComponent`:

- **"Back to API"** was navigating to the first item in the sidenav instead of the API the user came from — `pageId` input was bound to a non-existent query param; renamed to `selectedId` and fixed the back link `queryParams` accordingly
- **"Cancel"** was always resulting in a 404 — it was navigating to `/documentation/:navId/api/:apiId` which has no matching route; now it navigates back to `/documentation/:navId?selectedId=...`

### 1️⃣🐛 "Back to API" button 

Every time you click on the "Back to API" button, you're redirected to the first element on the left navbar

https://github.com/user-attachments/assets/d5034974-c704-49cc-950a-b13967fa97ba

### 2️⃣🐛 "Cancel" button

Every time you click on the "Cancel" button, you're seeing 404 instead of being redirected to the doc items list 

https://github.com/user-attachments/assets/89e3a14a-1a00-4569-bc62-b98c4f87540d

### 🔨 Fix video

In the video below, you can see that now both bugs are not present anymore

https://github.com/user-attachments/assets/cbfca67a-f456-4e1e-b63d-84c13ba9ffe6

### Bonus bugfix

Now "Back to API" button clickable area is the same size as the button size

https://github.com/user-attachments/assets/c9e11fdb-deee-4e9e-b49f-0b42750d0110

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-subscribe/documentation-subscribe.component.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-subscribe/documentation-subscribe.component.spec.ts
@@ -29,7 +29,7 @@ describe('DocumentationSubscribeComponent', () => {
 
   const MOCK_API = fakeApi({ id: 'api-1', name: 'Test API' });
 
-  const init = async () => {
+  const init = async (selectedId?: string) => {
     routerSpy = { navigate: jest.fn().mockReturnValue(Promise.resolve(true)) } as unknown as jest.Mocked<Router>;
 
     await TestBed.configureTestingModule({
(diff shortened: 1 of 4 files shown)


aqibmohammadkhan (approved):

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-subscribe/documentation-subscribe.component.spec.ts
aqibmohammadkhan, line 1: the test asserts `queryParams: { selectedId: undefined }.` Angular often omits undefined query params during URL serialization. the test is fine for verifying the method call shape, but it can be slightly implementation-coupled. (Not a blocker.)

kovaren (changes requested):
There's a tiny bonus bug in the ticket about the button's width. Could you please handle that as well?

kovaren (approved):

PR #16210, opened 2026-04-02
fix: improve subscription styles. remove extra paddings. shrink filter fields. apply next gen typography.

kovaren:
## Issue

## Description

Subscribe to API:
- Upper border of the card is not clipped in the corners anymore
- Next and Cancel buttons are aligned with the card

Subscription list:
- Apply next gen typography
- Fix extra horizontal paddings
- Shrink filter fields to align with general fields sizes

https://github.com/user-attachments/assets/531cea7b-150e-4c94-910c-b5b337515793

### DIFF
--- a/gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.scss
+++ b/gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.scss
@@ -15,6 +15,8 @@
  */
 @use '../../scss/theme' as theme;
 
+$info-label-width: 140px;
+
 .subscription-info {
   height: 100%;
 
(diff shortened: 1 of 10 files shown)


aqibmohammadkhan (approved):

JedrzejJanasiak (approved):

gravitee-apim-portal-webui-next/src/components/subscription-info/subscription-info.component.scss
```diff
@@ -48,15 +50,15 @@
...
+      flex-wrap: wrap;
+      gap: 15px 150px;
     }
 
     &__grid {
       display: grid;
       grid-auto-rows: 22px;
       grid-gap: 15px 30px;
-      grid-template-columns: min-content 1fr;
+      grid-template-columns: $info-label-width 1fr;
```
jaroslawlakomy-grav, line 61: Have you considered CSS Grid functions minmax(...) , fit-content(...) ? to avoid fixed width

kovaren: Yep, that's a good point. I initially used minmax but it had an impact on the grid behavior during the wrapping so I decided to go with a fixed width, just to stay scoped to the visual fixes and avoid changing grid behavior.

I do have a plan to revisit this component in the next version.

PR #16216, opened 2026-04-02
fix(repository): support repositories.* prefix for ES cross-cluster and proxy config

thomasballivet:
## Issue

## Description

Fix repository config resolution for `repositories.*` prefix. Adds fallback for ES cross-cluster and proxy config that bypasses `PropertySource`. Adds tests for new prefix support across MongoDB and Elasticsearch modules.


araffass (approved):
The fix LGTM.
Just a side note: users who override ES config via environment variables using the old prefix (e.g., GRAVITEE_ANALYTICS_ELASTICSEARCH_ENDPOINTS_0) will still see WARN logs after upgrading, since the Helm-generated config is now correct but their env overrides still use the legacy format. Worth a mention in the upgrade/release notes

PR #16263, opened 2026-04-06
feat: include LLM Proxy and MCP Proxy APIs in environment log search

carlos-andres-osorio:
## Issue

[APIM-13097](https://gravitee.atlassian.net/browse/APIM-13097)

## Why

The environment-level log search only returned v4 HTTP Proxy API logs. LLM Proxy and MCP Proxy APIs produce connection log data that belongs in the same unified view, but a single type predicate in `SearchEnvironmentLogsUseCase` was silently dropping them before they reached the Elasticsearch query. This change also consolidates APIM-13098 (MCP Proxy), which was a separate story for the same fix.

## What changed

- **`SearchEnvironmentLogsUseCase`** — widened the API type predicate from `PROXY`-only to `PROXY || LLM_PROXY || MCP_PROXY` via the new `ApiType.isProxyLike()` helper.
- **`ApiType`** — added `isProxyLike()` instance method that encapsulates which proxy types produce connection logs. Centralising this on the enum prevents future additions from being missed at individual call sites.
- **`SearchEnvironmentLogsUseCaseTest`** — removed `aLLMProxyApiV4()` and `aMCPProxyApiV4()` from the "excluded types" parameterized provider; added `should_include_llm_and_mcp_proxy_api_ids_in_filters` to assert both types now appear in the filter passed to the CRUD service.

No changes to `ConnectionLogsCrudService`, `UserContextLoader`, or the repository layer — they are type-agnostic once given the correct API IDs.

## How to test

No behaviour change visible without data. Confirm existing tests pass:

```
mvn test -pl gravitee-apim-rest-api/gravitee-apim-rest-api-service
```

To verify end-to-end: create an LLM Proxy or MCP Proxy API, generate traffic, then query the environment-level logs endpoint — the API's requests should now appear in results.

## Gotchas / Follow-up

`A2A_PROXY` is not included in `isProxyLike()` — it is absent from the analytics type mapping (`ApiTypeFilterTransformer`) and has no confirmed connection-log support yet. It can be added to `isProxyLike()` in a follow-up once its log pipeline is validated.

[APIM-13097]: https://gravitee.atlassian.net/browse/APIM-13097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
@@ -201,7 +201,7 @@ public class SearchEnvironmentLogsUseCase {
             .apis()
             .orElseGet(Collections::emptyList)
             .stream()
-            .filter(api -> api.getType() == ApiType.PROXY)
+            .filter(api -> isWantedHttpApi(api.getType()))
             .map(Api::getId)
             .collect(Collectors.toSet());
         filterContext.limitByApiIds(apiIds);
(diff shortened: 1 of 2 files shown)


paulatulis (approved):
Looks great!

PR #16264, opened 2026-04-06
feat: add apiType field to environment log response

carlos-andres-osorio:
## Issue

[APIM-13092](https://gravitee.atlassian.net/browse/APIM-13092)

## Why

The Console UI needs to display an API Type column (HTTP_PROXY, LLM_PROXY, MCP_PROXY) in the environment-level logs view. The current `EnvironmentApiLog` schema has no API type field, so the UI has no data to render for that column.

## What changed

- **`openapi-logs.yaml`** — added `apiType` inline enum (`HTTP_PROXY`, `LLM_PROXY`, `MCP_PROXY`) to `EnvironmentApiLog`. The enum is a subset of the full `ApiType`: KAFKA and MESSAGE APIs don't produce HTTP connection logs, and `PROXY` is renamed `HTTP_PROXY` for clarity at the API surface.
- **`ApiLog.java`** — added `ApiType apiType` as the 3rd record field, with a `withApiType()` immutable-update helper. All existing `with*` methods updated to pass the new field through.
- **`LogNamesPostProcessorImpl`** — builds an `apiTypeById` map once from `UserContext.apis()` (already loaded) in `mapLogNames()`, then resolves per log via `Map.get()` in `enrichLog()`. O(apis) once, O(1) per log — consistent with the existing enrichment pattern for `apiName`, `planName`, etc.
- **`LogsEngineMapper`** — added `mapApiType()` with an exhaustive switch: `PROXY → HTTP_PROXY`, `LLM_PROXY → LLM_PROXY`, `MCP_PROXY → MCP_PROXY`, `A2A_PROXY / MESSAGE / NATIVE → null`. Using an exhaustive switch (no `default`) gives a compile-time error if a new `ApiType` constant is added without updating the mapper.
- **Tests** — added `LogsEngineMapperTest` (7 parameterised cases covering all domain enum values + null), extended `LogNamesPostProcessorImplTest` with `WhenApiTypeResolution` (3 cases) and field-preservation assertion, and added `WhenApiTypeMapping` to `LogsSearchResourceTest`.

## User-facing impact

- The `GET /environments/{envId}/logs/search` response now includes an `apiType` field on each log entry (`HTTP_PROXY`, `LLM_PROXY`, or `MCP_PROXY`). The field is omitted when the API type cannot be determined.
- **Docs**: `openapi-logs.yaml` is updated in this PR.

Sample response:
```json
{
  "data": [
    {
      "apiId": "5ddfa5e0-c88e-4297-9fa5-e0c88e8297cd",
      "apiName": "gko-2785",
      "apiType": "HTTP_PROXY",
      "timestamp": "2026-04-03T10:16:56.878-04:00",
      "id": "100ab934-1a2f-41da-8ab9-341a2fd1daf5",
      "requestId": "100ab934-1a2f-41da-8ab9-341a2fd1daf5",
      "method": "PUT",
      "clientIdentifier": "12ca17b49af2289436f303e0166030a21e525d266e209267433801a8fd4071a0",
      "plan": {
        "id": "ba29872f-17bc-46a9-a987-2f17bcf6a941",
        "name": "keyless"
      },
      "application": {
        "id": "1",
        "name": "Unknown"
      },
      "transactionId": "100ab934-1a2f-41da-8ab9-341a2fd1daf5",
      "status": 405,
      "requestEnded": true,
      "gatewayResponseTime": 127,
      "gateway": "Andress-MacBook-Pro.local",
      "uri": "/gko-2785/?statusCode=374",
      "endpoint": "https://api.gravitee.io/echo/?statusCode=374",
      "warnings": [],
      "additionalMetrics": {}
    }
  ]
}
```

## How to test

POST `/environments/{envId}/logs/search` with a time range that returns results. Verify the response contains `"apiType": "HTTP_PROXY"` on each log entry.

## Known Issues

- MINOR: `LogNamesPostProcessorImpl` does not emit a debug log when `apiType` resolves to `null` (API absent from context). The null case is handled gracefully; logging was deferred as it is not observable by users.

[APIM-13092]: https://gravitee.atlassian.net/browse/APIM-13092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/LogsEngineMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/LogsEngineMapper.java
@@ -71,4 +71,16 @@ public interface LogsEngineMapper {
     default Operator mapOperator(io.gravitee.rest.api.management.v2.rest.model.logs.engine.Operator operator) {
         return Operator.valueOf(operator.name());
     }
+
+    default io.gravitee.rest.api.management.v2.rest.model.logs.engine.EnvironmentApiLog.ApiTypeEnum mapApiType(
+        io.gravitee.definition.model.v4.ApiType apiType
+    ) {
+        if (apiType == null) return null;
+        return switch (apiType) {
+            case PROXY -> io.gravitee.rest.api.management.v2.rest.model.logs.engine.EnvironmentApiLog.ApiTypeEnum.HTTP_PROXY;
+            case LLM_PROXY -> io.gravitee.rest.api.management.v2.rest.model.logs.engine.EnvironmentApiLog.ApiTypeEnum.LLM_PROXY;
+            case MCP_PROXY -> io.gravitee.rest.api.management.v2.rest.model.logs.engine.EnvironmentApiLog.ApiTypeEnum.MCP_PROXY;
+            case A2A_PROXY, MESSAGE, NATIVE -> null;
+        };
+    }
 }
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-logs.yaml
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-logs.yaml
@@ -315,6 +315,13 @@ components:
           type: string
           description: Name of the API.
           example: My API
+        apiType:
+          type: string
+          description: The type of the API.
+          enum:
+            - HTTP_PROXY
+            - LLM_PROXY
+            - MCP_PROXY
         timestamp:
           type: string
           format: date-time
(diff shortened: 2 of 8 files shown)


paulatulis (approved):
Approved with a few non-blockers: 

1. `ApiLog` record boilerplate is becoming unsustainable. The `ApiLog` record now has 23 fields and 5 near-identical with* methods, each manually copying all fields. Adding `apiType` required touching every single with* method and adding a new one. This is a maintenance risk — any new field requires updating 5+ methods and every test that constructs an `ApiLog` positionally.

- Suggestion: This isn't something to block on in this PR, but it would be worth considering refactoring `ApiLog` to use a builder pattern (e.g., Lombok @Builder on a record companion, or switching to a @Builder-annotated class) in a follow-up. The positional constructor with 23 nulls in tests is fragile.

2. Consider simplifying `LogsEngineMapperTest` single-value `@ParameterizedTests` to plain `@Tests`

carlos-andres-osorio:
> Approved with a few non-blockers:
> 
>     1. `ApiLog` record boilerplate is becoming unsustainable. The `ApiLog` record now has 23 fields and 5 near-identical with* methods, each manually copying all fields. Adding `apiType` required touching every single with* method and adding a new one. This is a maintenance risk — any new field requires updating 5+ methods and every test that constructs an `ApiLog` positionally.
> 
> 
>     * Suggestion: This isn't something to block on in this PR, but it would be worth considering refactoring `ApiLog` to use a builder pattern (e.g., Lombok @builder on a record companion, or switching to a @Builder-annotated class) in a follow-up. The positional constructor with 23 nulls in tests is fragile.
> 
> 
>     2. Consider simplifying `LogsEngineMapperTest` single-value `@ParameterizedTests` to plain `@Tests`

Good points. Fixed both.

PR #16265, opened 2026-04-06
feat: env logs tabs

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-13199
## Description
Adds tabbed view for env level logging: 

Screenshot 2026-04-06 at 12 25 40 PM


carlos-andres-osorio:
nit: I know the story asks for it, but `Non-message APIs` could include Kafka.  Will it be clear that this is HTTP+LL+MCP?

carlos-andres-osorio (approved):
LGTM :ship:

paulatulis:
> nit: I know the story asks for it, but `Non-message APIs` could include Kafka. Will it be clear that this is HTTP+LL+MCP?

There will be a type column, but otherwise I don't think there is a clear note specifically about Kafka.

PR #16267, opened 2026-04-06
add mcpMethod field to environment log response

carlos-andres-osorio:
## Summary

Issue [APIM-13093](https://gravitee.atlassian.net/browse/APIM-13093)

- Adds `mcpMethod` string field to the `EnvironmentApiLog` REST response, surfacing the MCP method (e.g. `tools/list`, `resources/read`) for MCP Proxy APIs
- The value was already stored in Elasticsearch (`additional-metrics.keyword_mcp-proxy_method`) and filterable via `MCP_METHOD` — this change extracts it into a dedicated response field
- Non-MCP APIs return `null` for this field

## Changes

- **OpenAPI spec** (`openapi-logs.yaml`): added `mcpMethod` property to `EnvironmentApiLog`
- **Domain model** (`ApiLog.java`): added `String mcpMethod` field
- **Use case** (`SearchEnvironmentLogsUseCase.java`): extracts value from `additionalMetrics` map
- **Tests**: happy-path extraction, null-safety, and `toBuilder()` field-preservation

## Test plan

- [x] `SearchEnvironmentLogsUseCaseTest` — verifies `mcpMethod` is extracted from `additionalMetrics` and null when map is absent
- [x] `LogNamesPostProcessorImplTest` — verifies field survives `toBuilder()` round-trip
- [x] `LogsEngineMapperTest` — passes (MapStruct auto-maps the String field)

[APIM-13093]: https://gravitee.atlassian.net/browse/APIM-13093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
@@ -378,6 +378,7 @@ public class SearchEnvironmentLogsUseCase {
             .errorComponentType(item.getErrorComponentType())
             .warnings(mapWarnings(item.getWarnings()))
             .additionalMetrics(item.getAdditionalMetrics() != null ? item.getAdditionalMetrics() : Map.of())
+            .mcpMethod(item.getAdditionalMetrics() != null ? (String) item.getAdditionalMetrics().get("keyword_mcp-proxy_method") : null)
             .build();
     }
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
@@ -378,7 +378,7 @@ public class SearchEnvironmentLogsUseCase {
             .errorComponentType(item.getErrorComponentType())
             .warnings(mapWarnings(item.getWarnings()))
             .additionalMetrics(item.getAdditionalMetrics() != null ? item.getAdditionalMetrics() : Map.of())
-            .mcpMethod(item.getAdditionalMetrics() != null ? (String) item.getAdditionalMetrics().get("keyword_mcp-proxy_method") : null)
+            .mcpMethod(item.getMcpMethod())
             .build();
     }
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/ConnectionLogAdapter.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/ConnectionLogAdapter.java
@@ -55,6 +55,10 @@ public interface ConnectionLogAdapter {
         if (metrics == null) {
             return;
         }
-        target.setMcpMethod((String) metrics.get(MCP_PROXY_METHOD_KEY));
+
+        Object mcpMethodObj = metrics.get(MCP_PROXY_METHOD_KEY);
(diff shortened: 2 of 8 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
```diff
@@ -378,6 +378,7 @@ private ApiLog mapApiLog(BaseConnectionLog item) {
             .errorComponentType(item.getErrorComponentType())
             .warnings(mapWarnings(item.getWarnings()))
             .additionalMetrics(item.getAdditionalMetrics() != null ? item.getAdditionalMetrics() : Map.of())
+            .mcpMethod(item.getAdditionalMetrics() != null ? (String) item.getAdditionalMetrics().get("keyword_mcp-proxy_method") : null)
```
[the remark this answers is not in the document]
carlos-andres-osorio: Addressed in a follow-up commit. The key is now defined as a constant in `ConnectionLogAdapter` (infra layer), and the extraction logic was moved there via an `@AfterMapping` method. `BaseConnectionLog` gains a typed `mcpMethod` field populated by the adapter, so the use case now reads `item.getMcpMethod()` with no knowledge of the underlying storage key.

The key is intentionally left in `additionalMetrics` to preserve backward compatibility — the API logs endpoint (`/apis/{apiId}/logs`) has existing consumers that read `additionalMetrics['keyword_mcp-proxy_method']` directly from that map.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/ConnectionLogAdapter.java
```diff
@@ -44,4 +48,13 @@ public interface ConnectionLogAdapter {
     List<MetricsQuery.Filter.ResponseTimeRange> convert(List<Range> range);
 
     ConnectionDiagnosticModel convert(ConnectionDiagnostic connectionDiagnostic);
+
+    @AfterMapping
```
carlos-andres-osorio: The `extractMcpMethod` logic has no coverage in `ConnectionLogAdapterTest`. No existing test fixture includes `additionalMetrics` with the MCP key, so the happy-path extraction is never exercised. The `SearchEnvironmentLogsUseCaseTest` stubs `mcpMethod` directly on `BaseConnectionLog`, bypassing the adapter entirely.

A regression here (e.g. a wrong key name) would not be caught by the test suite — removing this method entirely leaves all tests green.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/ConnectionLogAdapter.java
```diff
@@ -44,4 +48,13 @@ public interface ConnectionLogAdapter {
...
+    default void extractMcpMethod(@MappingTarget BaseConnectionLog target) {
+        Map<String, Object> metrics = target.getAdditionalMetrics();
+        if (metrics == null) {
+            return;
+        }
+        target.setMcpMethod((String) metrics.get(MCP_PROXY_METHOD_KEY));
```
paulatulis: While `keyword_` prefixed Elasticsearch parameters are virtually always guaranteed to be stored as Strings, casting this dynamically typed Map value blindly to (String) introduces a runtime risk.

If `additionalMetrics` inadvertently stores a boolean, integer, or array due to a transient mapping collision or malformed upstream proxy telemetry, (String) will immediately result in a ClassCastException. This fail-state will unexpectedly return a 500 Internal Server Error rendering the entire page of logs unviewable for the user during that timeframe.

Recommendation: Wrap it with a safe `instanceof` check to guarantee type-safety 

```
Object mcpMethodObj = metrics.get(MCP_PROXY_METHOD_KEY);
if (mcpMethodObj instanceof String mcpMethod) {
    target.setMcpMethod(mcpMethod);
}

```

paulatulis (approved):
Review suggestion is not a blocker. LGTM!

PR #16271, opened 2026-04-07
Fix cockpit idp run e2e

ThibaudAV:
## Description

Store the CSRF token from all organization fetch requests during bootstrap, not just the `/v2/ui/bootstrap` call.

Previously, `storeCsrfToken` was only applied to the initial bootstrap request. The three subsequent fetch calls to org endpoints (`/console`, `/ui/customization`, `/social-identities`) could also return a CSRF token via the `X-Xsrf-Token` header, but it was being ignored. This could lead to CSRF token mismatch errors, especially in multi-pod environments where a different pod might handle the bootstrap vs the org requests.

## Changes

- Added `storeCsrfToken` call on the response of each org fetch request in `gravitee-apim-console-webui/src/index.ts`

## How CSRF works here

The CSRF protection uses signed JWTs (HMAC-SHA256) stored in a cookie. It is stateless: any pod with the same `jwt.secret` can validate tokens generated by another pod. The frontend stores the token in `localStorage` and sends it as the `X-Xsrf-Token` header on every request.

### DIFF
--- a/gravitee-apim-console-webui/src/index.ts
+++ b/gravitee-apim-console-webui/src/index.ts
@@ -27,9 +27,11 @@ import { Build, Constants, DefaultPortal } from './entities/Constants';
 import { getFeatureInfoData } from './shared/components/gio-license/gio-license-data';
 import { ConsoleCustomization } from './entities/management-api-v2/consoleCustomization';
 import { environment } from './environments/environment';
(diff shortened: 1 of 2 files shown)


Okhelifi (approved):
Tested locally lgtm 👍🏼 
More tests incoming on QA ( here https://org-oki-1763488041215.qa.eu.823ed3.console.gravitee.dev/#!/_login ) after  the merge

PR #16280, opened 2026-04-07
feat: asset type column

paulatulis:
## Issue
https://gravitee.atlassian.net/browse/APIM-13207

## Description
Adds asset type column to logs table

Screenshot 2026-04-07 at 12 59 41 PM

### DIFF
--- a/gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-table/env-logs-table.component.html
+++ b/gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-table/env-logs-table.component.html
@@ -34,6 +34,10 @@
     }
   </ng-template>
 
+  <ng-template #assetTypeTemplate let-log>
+    {{ log.apiType || '—' }}
+  </ng-template>
+
   <ng-template #methodTemplate let-log>
     <span [class]="'gio-method-badge-' + log.method.toLocaleLowerCase()">{{ log.method }}</span>
   </ng-template>
--- a/gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-table/env-logs-table.component.ts
+++ b/gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-table/env-logs-table.component.ts
@@ -38,10 +38,12 @@ import { EnvLog } from '../../models/env-log.model';
 export class EnvLogsTableComponent {
   logs = input.required<EnvLog[]>();
   pagination = input.required<Pagination>();
+  showAssetTypeColumn = input<boolean>(false);
 
   paginationUpdated = output<GioTableWrapperPagination>();
 
   readonly timestampTemplate = viewChild.required<TemplateRef<EnvLog>>('timestampTemplate');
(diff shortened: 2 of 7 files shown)


gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-table/env-logs-table.component.ts
```diff
@@ -56,6 +57,7 @@ export class EnvLogsTableComponent {
 
   readonly columns = computed<LogsListColumnDef[]>(() => [
     { id: 'timestamp', label: 'Timestamp', template: this.timestampTemplate() },
+    { id: 'assetType', label: 'Asset Type', template: this.assetTypeTemplate() },
```
carlos-andres-osorio: The `assetType` column is hardcoded into the `columns()` computed signal, which is shared by both tab instances. The Jira story specifies this column for the non-message tab only. Currently harmless since the Messages tab always renders empty data, but when APIM-13200 wires up message logs the column will appear there too without any further change.

gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-table/env-logs-table.component.html
```diff
@@ -34,6 +34,10 @@
     }
   </ng-template>
 
+  <ng-template #assetTypeTemplate let-log>
+    {{ log.apiType || '—' }}
```
carlos-andres-osorio, line 38: The template renders the asset type as plain text only. The Jira story says "Consider icon + label display (matches Figma)" — no icon is implemented. If label-only was agreed with the team for this story, please close that item in Jira.

gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.ts
```diff
@@ -47,6 +47,12 @@ const TAB_QUERY_PARAM = 'tab';
 const MESSAGES_TAB_VALUE = 'messages';
 const DEFAULT_PER_PAGE = 10;
 
+const API_TYPE_LABELS: Record<string, string> = {
```
carlos-andres-osorio: `API_TYPE_LABELS` is typed as `Record<string, string>`. `LogApiType` is already exported from the service — using `Record<LogApiType, string>` would give a compile-time error if a new variant is added to `LogApiType` without a corresponding entry here.

```suggestion
const API_TYPE_LABELS: Record<LogApiType, string> = {
```

Also add `LogApiType` to the existing import on line 39.

gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-table/env-logs-table.component.html
paulatulis: In Figma, the label is in plain text. I don't see any icons for the api type column, and this implementation matches the designs presented: https://www.figma.com/design/nF1okCaBiG912gAsj0WsUs/Logs?node-id=4437-16639&t=mda1YBNeHNgs8EEE-0

carlos-andres-osorio (approved):

PR #16282, opened 2026-04-07
fix(v4-logs): persist filter state and period on navigation

araffass:
## Issue

https://gravitee.atlassian.net/browse/APIM-13403

## Description

Fixed filter loss when navigating v4 runtime logs by ensuring state is correctly synced with the URL and browser history.

Sync the period filter with URL query parameters and replace detail view routerLinks with browser history navigation to prevent state loss.

## Additional context

**_Before fix:_**

https://github.com/user-attachments/assets/c3c35b93-a248-4879-8693-6e1e50f9883a

**_After fix:_**

https://github.com/user-attachments/assets/989d1480-c14a-4726-b2a5-fcd7cba7794e

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs-details/components/runtime-logs-messages/api-runtime-logs-messages.component.html
+++ b/gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs-details/components/runtime-logs-messages/api-runtime-logs-messages.component.html
@@ -22,7 +22,9 @@
         <div class="card__header__title">Details</div>
         <div class="card__header__subtitle">Analyse the connection logs and messages for a specific connection</div>
       </div>
-      <a mat-button mat-raised-button [routerLink]="'..'" aria-label="Back to Connections">Back to Connections</a>
+      <button mat-button mat-raised-button routerLink=".." queryParamsHandling="preserve" aria-label="Back to Connections">
+        Back to Connections
+      </button>
     </div>
 
     <mat-tab-group (selectedTabChange)="onTabChange($event)">
--- a/gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs-details/components/runtime-logs-proxy/api-runtime-logs-proxy.component.html
+++ b/gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs-details/components/runtime-logs-proxy/api-runtime-logs-proxy.component.html
@@ -15,7 +15,9 @@
     limitations under the License.
 
 -->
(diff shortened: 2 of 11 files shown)


mukul-tyagi08 (commented):
The location.back() approach has an edge case: if a user opens a log-detail URL directly, location.back() may navigate out of the app or does nothing and also I guess this navigation pattern isn't used elsewhere in the codebase. 

I think we can use queryParamsHandling. 

Adding queryParamsHandling="preserve" (or "merge") on the routerLink in api-runtime-logs-list.component.html (list -> detail)
Keeping routerLink=".." on the detail pages but adding queryParamsHandling="preserve" (detail -> list)

The period-to-URL serialization (the store changes) would still be needed and is correct as-is.

Wdyt?

gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs-details/components/runtime-logs-messages/api-runtime-logs-messages.component.html
```diff
@@ -22,7 +22,9 @@
         <div class="card__header__title">Details</div>
         <div class="card__header__subtitle">Analyse the connection logs and messages for a specific connection</div>
       </div>
-      <a mat-button mat-raised-button [routerLink]="'..'" aria-label="Back to Connections">Back to Connections</a>
+      <button mat-button mat-raised-button routerLink=".." queryParamsHandling="preserve" aria-label="Back to Connections">
+        Back to Connections
+      </button>
```
mukul-tyagi08, line 27: <a tag is semantically more appropriate for navigation and preserves native link behaviour. So I think, we can keep <a> as it was.

gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs-details/components/runtime-logs-proxy/api-runtime-logs-proxy.component.html
```diff
@@ -15,7 +15,9 @@
     limitations under the License.
 
 -->
-<a mat-flat-button routerLink=".." aria-label="Back to logs"><mat-icon svgIcon="gio:arrow-left"></mat-icon>Back to logs</a>
+<button mat-flat-button routerLink=".." queryParamsHandling="preserve" aria-label="Back to logs">
+  <mat-icon svgIcon="gio:arrow-left"></mat-icon>Back to logs
+</button>
```
mukul-tyagi08, line 20: Same here. We can keep <a as it was.

mukul-tyagi08 (approved):

PR #16290, opened 2026-04-08
chore: add global agent rules and nested agents.md.

kovaren:
## Summary

Introduce a repo-wide layered structure for AI agent guidance.

Give coding agents **one predictable pattern** (global → stack rules → module rules) and avoid loading irrelevant guidance when editing a specific area.

## Rule structure

```
.
├── AGENTS.md                           # Global entrypoint
├── .agent-rules/
│   ├── angular.md                      # Shared Angular / frontend standards
│   └── java.md                         # Shared Java / Maven standards
│
├── gravitee-apim-gateway/              # Module-specific (Java)
│   └── AGENTS.md                                
├── gravitee-apim-rest-api/
│   └── AGENTS.md
├── gravitee-apim-rest-*/
│   └── AGENTS.md
|   ...
│
├── gravitee-apim-console-webui/        # App-specific (Angular)
│   └── AGENTS.md                       
├── gravitee-apim-portal-webui-next/
│   └── AGENTS.md
│
└── gravitee-apim-webui-libs/           # Library-specific (Angular)
    ├── gravitee-dashboard/
    │   └── AGENTS.md
    ├── gravitee-kafka-explorer/
    │   └── AGENTS.md
    └── gravitee-markdown/
        └── AGENTS.md
```

## Changes

- update root **`AGENTS.md`** to describe when to load common and nested rules and to list all **Java** and **Angular** module roots.
- create **`.agent-rules/`** with shared guidelines for Java (`java.md`) and Angular (`angular.md`).
- create nested **`AGENTS.md`**  across Maven top-level modules and Angular apps/libs.
- rename **`CLAUDE.md`** → **`AGENTS.md`** in **gravitee-kafka-explorer**.


gravitee-apim-portal-webui-next/AGENTS.md
```diff
@@ -0,0 +1,75 @@
+# Coding Guidelines for gravitee-apim-portal-webui-next
+
+Design system rules for the **main app**. Use together with the library-specific rule files when working in those libraries.
+
+## Glue rule: when to apply these rules
+
+- **Apply this file** when editing code in the **main portal-next application** (app shell, pages, app-level SCSS, layout, theme).
+- **Also use** [portal-next-gravitee-markdown.mdc](portal-next-gravitee-markdown.mdc) when editing inside the **gravitee-markdown** library (GMD components, suggestions, stories).
+- **Also use** [portal-next-gravitee-dashboard.mdc](portal-next-gravitee-dashboard.mdc) when editing inside the **gravitee-dashboard** library (charts, gd-* components).
```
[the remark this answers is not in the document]
kovaren: Fixed

jaroslawlakomy-grav (approved):
I find it as a good starting point. I can see some improvements like description what is the "onion" and "spaghetti" architecture in services, probably whole ruleset will be evolving in upcoming months

gravitee-apim-rest-api/AGENTS.md
JedrzejJanasiak, line 1: @kovaren @jaroslawlakomy-grav, wdyt of this rule? I had a sessions where the agent went straight to DTOs instead of the spec. Also happy to drop the “What happens” block if you’d rather keep it minimal.

## API-first: OpenAPI before JAX-RS

Use this when you add or change **HTTP endpoints** (paths, operations, request/response bodies) in any of these modules:

| Area | Maven module (path suffix) | OpenAPI spec(s) |
| --- | --- | --- |
| **Management API v2** | `gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest` | `src/main/resources/openapi/openapi-*.yaml` — several specs by area (e.g. `openapi-apis.yaml`, `openapi-api-products.yaml`, `openapi-environments.yaml`) |
| **Kafka Explorer** | `gravitee-apim-rest-api-kafka-explorer` | `src/main/resources/openapi/openapi-kafka-explorer.yaml` |
| **Portal API** | `gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest` | `src/main/resources/portal-openapi.yaml` |

**What happens**

1. **Edit OpenAPI** — paths, operations, `components.schemas`, etc. For **v2**, change the YAML that already holds similar routes/schemas (e.g. product APIs → `openapi-api-products.yaml`). **Portal** and **Kafka Explorer** each have a single spec file (see table).
2. **Compile the module** — e.g. `mvn -pl gravitee-apim-rest-api/<module-path> compile`. The `openapi-generator-maven-plugin` runs with `generateModels=true` and `generateApis=false`, so **Java model/DTO classes** are generated into `target/` (not hand-written in `src/main/java`).
3. **Implement JAX-RS** — resource classes and mappers import those generated types (e.g. `io.gravitee.rest.api.management.v2.rest.model.*`, `io.gravitee.rest.api.kafkaexplorer.rest.model.*`, `io.gravitee.rest.api.portal.rest.model.*`; v2 also uses subpackages such as `...model.analytics.engine` / `...model.logs.engine` where the POM overrides `modelPackage`).

**Rule for new endpoints:** always **API-first** — update the right OpenAPI file, run Maven on that module so generation runs, then add or change resources. Starting from the resource skips step 2, so the model classes do not exist yet and tooling breaks.

JedrzejJanasiak (approved):

gravitee-apim-rest-api/AGENTS.md
kovaren: Yes, if that's something that you encountered let's keep it. The highest-value rules are those created after an actual wrong behavior by an agent.

JedrzejJanasiak: done 👍

jgiovaresco (approved):

.agent-rules/java.md
```diff
@@ -0,0 +1,157 @@
...
+var agg = new Aggregation();
+agg.setBuckets(...);
+assertThat(result).hasValueSatisfying(agg -> { ... });  // 'agg' shadows outer
+
+// ✅ GOOD
+var aggregation = new Aggregation();
+assertThat(result).hasValueSatisfying(g -> { ... });
+```
+
+Use distinct names: `aggregation` / `agg` for the object, `g` / `result` for lambda params.
```
michel-barret: This is java not specific to anything. If an agent need it to handle correctly the java syntaxe it’s should be set only for this agent.

.agent-rules/java.md
```diff
@@ -0,0 +1,157 @@
...
+- **Var**: Use `var` where type is obvious.
+- **Architecture**: Hexagonal / Clean where applicable.
+- **BOM**: Align with `gravitee-apim-bom`.
+
+## 2. Exception Handling and Logging
+
+### Logger Injection
+
+1. Use `@CustomLog` to inject the appropriate logger into your classes.
+2. **Reactive Logging (Gateway & Plugins):** In the gateway and plugins (reactive code), prefer using `ctx.withLogger(log)` whenever possible to ensure proper logging context within the reactive stream.
```
ytvnr: Not sure Reactive is the good word (I know it is a copy paste :D)
```suggestion
2. **Contextualized Logging (Gateway & Plugins):** In the gateway and plugins (reactive code), prefer using `ctx.withLogger(log)` whenever possible to ensure proper logging context within the reactive stream.
```

ytvnr: this section should be moved in the gateway and apim-plugins file, wdyt ?

.agent-rules/java.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Formatting Before Tests
+
+Run the Maven Prettier plugin on modified modules before tests — the build fails on formatting checks:
+
+```bash
+mvn prettier:write -pl <module>
+```
+
+## 4. Testing
```
ytvnr, line 90: I would add something to avoid mock when possible, even more in clean archi where we should try to use real production code to instanitate test data

ytvnr (approved):

.agent-rules/java.md
kovaren: Good point, the yield and lambda shadowing are indeed generic Java, I'll trim those.

I'd still like to keep the JsonNode vs ObjectNode since it references our Aggregation.setBuckets() API.

The main idea for this PR is to have the same rules for all agents, so I'd like to avoid adding agent specific rules. But it's definitely a topic to discuss in the future.

WDYT?

michel-barret: I’m agree with that 👍

.agent-rules/java.md
kovaren: Fixed!

.agent-rules/java.md
kovaren: Done!

PR #16312, opened 2026-04-09
feat(portal-next): define application members read/search contract

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/APIM-12280

https://gravitee.atlassian.net/browse/APIM-13464

## Description

Updated the Portal OpenAPI contract for APIM-12280 on GET /applications/{applicationId}/members: added member search support (q, minLength: 1), clarified pagination behavior, and documented 400 for invalid query/pagination inputs.
Also fixed a pre-existing spec inconsistency by adding the missing components.responses.BadRequestError definition, which was already referenced and causing IDE/validation errors.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
@@ -1173,14 +1173,17 @@ paths:
                 List application members.
 
                 User must have the APPLICATION_MEMBER[READ] permission.
+                Pagination controls are always provided in the `links` object.
             operationId: getMembersByApplicationId
             responses:
                 200:
-                    description: List of members
+                    description: List of members with pagination links
                     content:
                         application/json:
                             schema:
                                 $ref: "#/components/schemas/MembersResponse"
+                400:
+                    $ref: "#/components/responses/BadRequestError"
                 403:
                     $ref: "#/components/responses/PermissionError"
                 404:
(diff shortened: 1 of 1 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
```diff
@@ -1168,19 +1168,28 @@ paths:
...
                     content:
                         application/json:
                             schema:
                                 $ref: "#/components/schemas/MembersResponse"
+                400:
+                    description: Invalid pagination or search query parameter.
+                    content:
+                        application/json:
+                            schema:
+                                $ref: "#/components/schemas/ErrorResponse"
```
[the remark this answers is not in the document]
jaroslawlakomy-grav: common response used

dev-marek (changes requested):
This approach would be correct if we would perform filtering by some indexed value or id (get by API id, get by membership type, etc.) but in case of full text search (arbitrary query string, pattern) we should go for `/_search` endpoint

also be careful about backward compatibility/breaking changes in PAPI - you've introduced a query param with `minLength: 1` - which will probably break previous usages

kovaren (commented):
This change is a departure from how the Console handles filtering.

In the Console, all members are loaded in memory and paginated on the front end. This is definitely inefficient but it allows to filter by other fields that the name.

Let's first check if we want the feature parity with the Console here and then decide on the implementation.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
```diff
@@ -4161,6 +4204,19 @@ components:
                     $ref: "#/components/schemas/MetadataMap"
                 links:
                     $ref: "#/components/schemas/Links"
+        ApplicationMembersSearchInput:
+            type: object
+            description: Search input for application members. The filters object is extensible for future search attributes.
+            properties:
+                filters:
```
dev-marek, line 4209: directly into request

dev-marek (approved):
lgtm

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
jaroslawlakomy-grav: done

PR #16323, opened 2026-04-09
feat: add api product name to env log table

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-13539
## Description

Adds apiProductName to env log list.


carlos-andres-osorio (approved):

gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-table/env-logs-table.component.scss
```diff
@@ -23,3 +23,14 @@
     text-decoration: none;
   }
 }
+
+.api-cell {
+  display: flex;
+  flex-direction: column;
+
+  &__product {
+    font-size: 12px;
```
carlos-andres-osorio: Hard-coded `font-size: 12px` — per the Particles design system styling rules:

> Tokens only: no hex/RGB or arbitrary font-size px. Use design tokens, CSS/SCSS variables.

If a caption/small-text token is available in `@gravitee/ui-particles-angular`, it should be used here instead.

PR #16326, opened 2026-04-10
fix: remove subscription end-date check against client certificate

benoitgravitee:
## Issue

* https://gravitee.atlassian.net/browse/GKO-2788
* An exception was raised when one subscription ends after the certificate, it is not required, it is the Application manager to manage this, we can't deny this or the user will be block managing it's certificates the way he want it.

## Description

- Removes the validation that threw `SubscriptionEndsAfterClientCertificateException` when a subscription end date fell after the client certificate expiry.
- Both throw sites are removed: `SubscriptionValidationServiceImpl.validateTls()` and `ApplicationServiceImpl.syncClientCertificates()`.
- The `SubscriptionEndsAfterClientCertificateException` class itself is deleted (no more references).


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImplTest.java
```diff
@@ -224,104 +214,6 @@ void should_sanitize_configuration_when_valid_configuration() {
         }
     }
 
-    @Nested
```
carlos-andres-osorio, line 227: The entire `MTLSType` nested class was deleted rather than updated. AC #2 requires asserting that the exception is *not* thrown where it previously was — meaning the scenario that previously triggered `SubscriptionEndsAfterClientCertificateException` (MTLS plan + `endingAt` in the far future) should now be tested as a passing case.

With this deletion there is zero test coverage for the MTLS subscription update path in this class. If the check were accidentally re-introduced, nothing would catch it.

carlos-andres-osorio (approved):
Approving with a minor comment

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/impl/validation/SubscriptionValidationServiceImplTest.java
benoitgravitee: I get it but the whole point was to remove any kind of check... so checking that it is working has zero value and will be confusing

PR #16316, opened 2026-04-09
feat: implement backend members read path for application

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/APIM-13465

## Description

Add portal members listing for applications with pagination and permission checks, map membership data to the Member contract (including member.id), and cover empty/populated responses in resource tests (including links = null for empty list).

PR #16332, opened 2026-04-10
feat: implement members search by displayName in portal API

jaroslawlakomy-grav:
## Issue

## Description

- add q filtering in GET /applications/{applicationId}/members (displayName only, case-insensitive)
- update OpenAPI param description to reflect displayName scope
- extend ApplicationMembersResourceTest with search scenarios, including non-matching email case


jaroslawlakomy-grav:
duplicate with https://github.com/gravitee-io/gravitee-api-management/pull/16316

PR #16339, opened 2026-04-10
fix: let users pass page HRID for plan general condition

kamiiiel:
see https://gravitee.atlassian.net/browse/GKO-2803

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/ImportApiCRDUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/ImportApiCRDUseCase.java
@@ -199,6 +199,8 @@ public class ImportApiCRDUseCase {
                 oneShotIndexation(input.auditInfo)
             );
 
+            createOrUpdatePages(input.spec.getPages(), createdApi.getId(), input.auditInfo);
+
             var planNameIdMapping = input.spec
                 .getPlans()
                 .entrySet()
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/ImportApiCRDUseCase.java
```diff
@@ -252,6 +252,9 @@ private ApiCRDStatus create(Input input) {
 
     private ApiCRDStatus update(Input input, Api existingApi) {
         try {
+            // Persist pages first so plan general-condition references are resolvable during API update.
+            createOrUpdatePages(input.spec.getPages(), existingApi.getId(), input.auditInfo);
```
benoitgravitee, line 256: I'm ok with that, but It may not be enough, what happens at creation time... looks plans are created first. Pages should be created before plans. no ?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/ImportApiCRDUseCase.java
benoitgravitee: If plan uses general conditions it may not work... I'm pretty there is a case with Terraform where we do it all at once, maybe the underlying service does not check on creation ...

kamiiiel: In here we can't create pages first, because there is no API in the DB. And we also don't have any issues with the creation, at least with the code you put there for validating general conditions

a-cordier (commented):
can we do a quick test of the fix with TF ?

benoitgravitee (approved):

PR #16355, opened 2026-04-13
fix: handle DASHBOARD audit reference type to prevent NPE in audit log UI

moerradi:
## Issue

https://github.com/gravitee-io/issues/issues/11338.

## Description

DASHBOARD was added to the repository-layer Audit.AuditReferenceType enum but was never propagated to the service-layer AuditReferenceType enum or the conversion map in AuditServiceImpl. This caused Map.get() to return null for any DASHBOARD audit record, which then triggered a NPE in getMetadata() where getReferenceType().name() was called without a null guard, making the entire audit log page fail to load.

## Additional context

the issue has all the context


vikrantgravitee:
Hi @moerradi 

Thank you again for your contribution - we really appreciate your effort here.

Following our internal testing and validation, we incorporated the changes through a separate PR, which has now been released as part of version 4.11.5.

To keep things consistent, I’ll go ahead and close this PR since the changes are already available in the repository.

We truly value your contributions, and we’d be happy to see more from you going forward.

PR #16357, opened 2026-04-13
fix: trigger subscription form backend validation only for portal endpoints

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-13609

### Summary
This change fixes subscription creation from the Console (Management API) when a plan has a subscription form with required fields. Backend validation for that form was running for every subscription request, so Console users—who never see the form—could not satisfy required fields and requests failed.

We now treat Portal and Management (Console / mAPI) subscription flows differently: subscription form backend validation runs only when the request is known to come from the Portal. Management-originated requests skip that validation so publishers and admins can still create or manage subscriptions as before.

Sanitization of subscription metadata (keys, stripping unsafe markup, etc.) remains global for all origins so we do not weaken stored-data safety for Console-created subscriptions.

### Problem
Subscription form backend validation was applied regardless of who created the subscription.
The Console does not expose the subscription form UI for that flow, so required fields from the form could never be submitted → validation errors and blocked subscriptions.
There was no reliable signal in the subscription pipeline to mean “the user actually went through the Portal form.”
### Approach (this PR)
Introduce an internal flag on the subscription-related payloads (e.g. create / update paths) that indicates whether subscription form backend validation should run.
Portal REST resources set this flag when building the subscription so validation runs when the form is shown to the consumer.
Management API / Console paths set it so validation is not run for that origin.
Keep metadata sanitization separate: it continues to run for all requests; only the form constraint validation is conditional on the flag.

## Additional context

**Important limitation — shared Portal API (pAPI)**

The Portal API is shared between the **Classic Portal** and the **Next Gen Portal**, so the backend **cannot tell** which portal a request came from.
**Implication**: if Subscription Form is enabled globally and includes fields with required constraints, **subscription-form backend validation also runs for Classic Portal flows**. Classic Portal does not surface that form in the subscription journey, so users cannot supply those values — in practice this blocks subscription creation on the old portal.
This PR does not add classic-vs-next detection on pAPI; that remains a known constraint of the current architecture while both portals use the same endpoints.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/NewSubscriptionEntity.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/NewSubscriptionEntity.java
@@ -15,9 +15,6 @@
  */
 package io.gravitee.rest.api.model;
 
-import com.fasterxml.jackson.annotation.JsonRawValue;
-import com.fasterxml.jackson.annotation.JsonSetter;
-import com.fasterxml.jackson.databind.JsonNode;
 import io.gravitee.rest.api.model.context.OriginContext;
 import java.util.Map;
 import lombok.Getter;
@@ -46,6 +43,13 @@ public class NewSubscriptionEntity {
     private OriginContext.Origin origin = OriginContext.Origin.MANAGEMENT;
 
     private Map<String, String> metadata;
+
+    /**
+     * <strong>Temporary:</strong> remove once the Console supports the subscription form flow.
+     * Until then, {@code true} runs the subscription-form metadata validation (portal); {@code false}
+     * skips it for mAPI/Console requests.
+     */
+    private Boolean subscriptionFormMetadataValidationRequired = false;
 
     public NewSubscriptionEntity(String plan, String application) {
         this.application = application;
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription/use_case/CreateSubscriptionUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription/use_case/CreateSubscriptionUseCase.java
@@ -71,7 +71,8 @@ public class CreateSubscriptionUseCase {
             input.metadata,
             input.apiKeyMode,
             input.generalConditionsAccepted,
-            input.generalConditionsContentRevision
+            input.generalConditionsContentRevision,
+            input.subscriptionFormMetadataValidationRequired
         );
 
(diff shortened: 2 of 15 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/NewSubscriptionEntity.java
```diff
@@ -47,6 +44,13 @@ public class NewSubscriptionEntity {
 
     private Map<String, String> metadata;
 
+    /**
+     * <strong>Temporary:</strong> remove once the Console supports the subscription form flow.
+     * Until then, {@code true} runs the subscription-form metadata validation (portal); {@code false}
+     * skips it for mAPI/Console requests.
+     */
+    private boolean subscriptionFormMetadataValidationRequired = false;
```
kovaren: I'd suggest to use the wrapper class everywhere to stay consistent with the existing patterns

```suggestion
    private Boolean subscriptionFormMetadataValidationRequired = false;

**Review by kovaren — CHANGES_REQUESTED** [review]

**Review by dev-marek — CHANGES_REQUESTED** [review]
It is a very bad idea to toggle form validation by entity already used deep down in legacy code. Instead please run form validation directly before passing new entity to domain service in use case. Also I'm not a huge fan of Builders for Input records in use cases, but since we've got them, please use Bulder default value instead of replacing that everywhere

**Review by dev-marek — APPROVED** [review]

File: gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription/use_case/CreateSubscriptionUseCase.java [hunk 2]
```diff
@@ -115,7 +116,8 @@ public record Input(
         io.gravitee.rest.api.model.ApiKeyMode apiKeyMode,
         Boolean generalConditionsAccepted,
         io.gravitee.rest.api.model.PageEntity.PageRevisionId generalConditionsContentRevision,
-        AuditInfo auditInfo
+        AuditInfo auditInfo,
+        boolean subscriptionFormMetadataValidationRequired
```
kovaren: Still a primitive boolean left here 😋

kovaren (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription/use_case/CreateSubscriptionUseCase.java
JedrzejJanasiak: 🙈

kovaren (approved):

PR #16363, opened 2026-04-13
fix(healthcheck): use configured Host header when sending health check request

dev-marek:
## Issue

https://gravitee.atlassian.net/browse/APIM-13512

## Description

- Fixed custom `Host` header configured in the health check HTTP headers section not being forwarded to the backend when using an `http-proxy` endpoint
- Root cause 1: `buildRequestOptions()` unconditionally removed the `Host` header when `request.host()` is null (health check context — `HttpHealthCheckRequest` never sets the `host` field), stripping the custom value before it reached the connector
- Root cause 2: `setAbsoluteURI()` in `connect()` internally overwrites `options.host` with the endpoint target hostname, which Vert.x uses as the HTTP `Host` header, ignoring anything in the headers map
- Fix: skip `Host` removal in `buildRequestOptions()` when `currentRequestHost` is null; in `connect()`, detect a custom `Host` in the headers map after `setAbsoluteURI()` and use `setServer()` to pin the TCP connection to the actual backend while setting the HTTP `Host` header to the custom value

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-healthcheck/src/main/java/io/gravitee/gateway/services/healthcheck/rule/EndpointRuleHandler.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-healthcheck/src/main/java/io/gravitee/gateway/services/healthcheck/rule/EndpointRuleHandler.java
@@ -19,11 +19,9 @@ import static java.lang.System.currentTimeMillis;
 
 import io.gravitee.alert.api.event.Event;
 import io.gravitee.common.http.HttpStatusCode;
-import io.gravitee.common.utils.UUID;
 import io.gravitee.definition.model.Endpoint;
 import io.gravitee.definition.model.services.healthcheck.HealthCheckStep;
 import io.gravitee.el.TemplateEngine;
-import io.gravitee.el.exceptions.ExpressionEvaluationException;
 import io.gravitee.gateway.api.http.HttpHeaders;
 import io.gravitee.gateway.http.vertx.VertxHttpHeaders;
 import io.gravitee.gateway.services.healthcheck.EndpointRule;
@@ -32,7 +30,6 @@ import io.gravitee.gateway.services.healthcheck.eval.EvaluationException;
 import io.gravitee.gateway.services.healthcheck.eval.assertion.AssertionEvaluation;
 import io.gravitee.gateway.services.healthcheck.http.el.EvaluableHttpResponse;
 import io.gravitee.node.api.Node;
-import io.gravitee.node.api.utils.NodeUtils;
 import io.gravitee.plugin.alert.AlertEventProducer;
 import io.gravitee.reporter.api.Reportable;
 import io.gravitee.reporter.api.common.Request;
@@ -152,45 +149,7 @@ public abstract class EndpointRuleHandler<T extends Endpoint> implements Handler
     }
 
     protected RequestOptions prepareHttpClientRequest(URL request, HealthCheckStep step) {
-        final int port = request.getPort() != -1 ? request.getPort() : (HTTPS_SCHEME.equals(request.getProtocol()) ? 443 : 80);
-
-        String relativeUrl = (request.getQuery() == null) ? request.getPath() : request.getPath() + '?' + request.getQuery();
-
-        RequestOptions options = new RequestOptions()
-            .setURI(relativeUrl)
-            .setPort(port)
-            .setHost(request.getHost())
-            .setMethod(HttpMethod.valueOf(step.getRequest().getMethod().name().toUpperCase()))
-            .putHeader(io.vertx.rxjava3.core.http.HttpHeaders.USER_AGENT, NodeUtils.userAgent(node))
-            .putHeader("X-Gravitee-Request-Id", UUID.toString(UUID.random()));
-
-        if (step.getRequest().getHeaders() != null) {
-            step.getRequest().getHeaders().forEach(h -> options.putHeader(h.getName(), resolveHeaderValue(h)));
-            applyCustomHostIfPresent(options, port, request);
-        }
-
-        return options;
-    }
-
-    private String resolveHeaderValue(io.gravitee.common.http.HttpHeader httpHeader) {
-        try {
-            String value = templateEngine.getValue(httpHeader.getValue(), String.class);
-            return value != null ? value : "";
-        } catch (ExpressionEvaluationException e) {
-            logger.warn("Expression {} cannot be evaluated for healthcheck of API {}", httpHeader.getValue(), rule.api().getId());
-            return "";
-        }
-    }
-
-    private void applyCustomHostIfPresent(RequestOptions options, int port, URL request) {
-        String customHost = options.getHeaders() != null ? options.getHeaders().get(io.vertx.core.http.HttpHeaders.HOST) : null;
-        if (customHost == null || customHost.isBlank()) {
-            return;
-        }
-        // Pin the TCP connection to the actual backend so that the custom Host value
-        // only affects the HTTP Host header and not the connection address.
-        options.setServer(io.vertx.core.net.SocketAddress.inetSocketAddress(port, request.getHost()));
-        options.setHost(customHost);
+        return new RequestOptionsBuilder(node, templateEngine, rule.api().getId()).build(request, step);
     }
 
     protected URL createRequest(T endpoint, HealthCheckStep step) throws MalformedURLException {
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-healthcheck/src/main/java/io/gravitee/gateway/services/healthcheck/rule/EndpointRuleHandler.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-healthcheck/src/main/java/io/gravitee/gateway/services/healthcheck/rule/EndpointRuleHandler.java
@@ -183,6 +183,14 @@ public abstract class EndpointRuleHandler<T extends Endpoint> implements Handler
 
                     options.putHeader(httpHeader.getName(), resolvedHeader == null ? "" : resolvedHeader);
                 });
+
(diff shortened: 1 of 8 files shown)


gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
```diff
@@ -127,6 +127,17 @@ public Completable connect(final HttpExecutionContext ctx) {
...
+
+            // setAbsoluteURI sets options.host from the endpoint target URL, which Vert.x uses as the
+            // HTTP Host header — overriding any custom Host header configured on the endpoint or HC.
+            // If a custom Host was explicitly set, pin the TCP connection to the actual backend via
+            // setServer() so the custom value only affects the HTTP Host header, not the connection address.
+            final String customHost = options.getHeaders() != null ? options.getHeaders().get(io.vertx.core.http.HttpHeaders.HOST) : null;
+            if (customHost != null && !customHost.isBlank()) {
+                options.setServer(io.vertx.core.net.SocketAddress.inetSocketAddress(options.getPort(), defaultHost));
+                options.setHost(customHost);
+            }
```
aqibmohammadkhan, line 139: i suggest we capture the parsed values before overwriting

```
String absoluteUri = VertxHttpClientFactory.toAbsoluteUri(options, defaultHost, defaultPort);
options.setAbsoluteURI(absoluteUri);

final String customHost = options.getHeaders() != null
    ? options.getHeaders().get(HttpHeaders.HOST)
    : null;
if (customHost != null && !customHost.isBlank()) {
    final String actualBackendHost = options.getHost();
    final int actualBackendPort = options.getPort();
    options.setServer(SocketAddress.inetSocketAddress(actualBackendPort, actualBackendHost));
    options.setHost(customHost);
}
```

This mirrors what the legacy `EndpointRuleHandler` and the v4 `HttpHealthCheckService` changes do (they both pin the server to the target URL host, not a constructor‑time default).

aqibmohammadkhan: A regression test that exercises the `ATTR_REQUEST_ENDPOINT` override + custom Host combination would catch this.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
```diff
@@ -238,16 +249,16 @@ protected RequestOptions buildRequestOptions(HttpExecutionContext ctx) {
...
-            // 'Host' header must be removed unless it has been set during the request flow (non-null and different from original request's host).
-            // If PropagateClientHost is enabled, we set the 'Host' header to the current request's host.
-            requestHeaders.set(HOST, currentRequestHost);
-        } else {
-            requestHeaders.remove(HOST);
+        if (currentRequestHost != null) {
+            if (sharedConfiguration.getHttpOptions().isPropagateClientHost() || !Objects.equals(originalHost, currentRequestHost)) {
+                // 'Host' header must be removed unless it has been set during the request flow (non-null and different from original request's host).
+                // If PropagateClientHost is enabled, we set the 'Host' header to the current request's host.
+                requestHeaders.set(HOST, currentRequestHost);
```
aqibmohammadkhan, line 256: are we sure that health check is the only caller that can hit `host() == null`, and if not, consider gating on an explicit marker (e.g., an execution attribute) rather than on `null`. At minimum, an explicit comment/assertion about expected callers would help.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-healthcheck/src/main/java/io/gravitee/gateway/services/healthcheck/rule/EndpointRuleHandler.java
```diff
@@ -183,6 +183,14 @@ protected RequestOptions prepareHttpClientRequest(URL request, HealthCheckStep s
 
                     options.putHeader(httpHeader.getName(), resolvedHeader == null ? "" : resolvedHeader);
                 });
+
+            final String customHost = options.getHeaders() != null ? options.getHeaders().get(io.vertx.core.http.HttpHeaders.HOST) : null;
```
aqibmohammadkhan, line 187: I think we already have an import [here](https://github.com/gravitee-io/gravitee-api-management/blob/dbcd6f441a29817789b55309d8cb9576d34dfe35/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-healthcheck/src/main/java/io/gravitee/gateway/services/healthcheck/rule/EndpointRuleHandler.java#L46)

you can drop the long package name and use the short class name.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
```diff
@@ -127,6 +127,17 @@ public Completable connect(final HttpExecutionContext ctx) {
             final RequestOptions options = buildRequestOptions(ctx);
             String absoluteUri = VertxHttpClientFactory.toAbsoluteUri(options, defaultHost, defaultPort);
             options.setAbsoluteURI(absoluteUri);
+
+            // setAbsoluteURI sets options.host from the endpoint target URL, which Vert.x uses as the
+            // HTTP Host header — overriding any custom Host header configured on the endpoint or HC.
+            // If a custom Host was explicitly set, pin the TCP connection to the actual backend via
+            // setServer() so the custom value only affects the HTTP Host header, not the connection address.
+            final String customHost = options.getHeaders() != null ? options.getHeaders().get(io.vertx.core.http.HttpHeaders.HOST) : null;
+            if (customHost != null && !customHost.isBlank()) {
```
aqibmohammadkhan, line 136: drop the long package name for HttpHeaders as the import already exist [here](https://github.com/gravitee-io/gravitee-api-management/blob/dbcd6f441a29817789b55309d8cb9576d34dfe35/gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java#L57)

use short class name.

gravitee-apim-plugin/gravitee-apim-plugin-apiservice/gravitee-apim-plugin-apiservice-healthcheck-http/src/test/java/io/gravitee/apim/plugin/apiservice/healthcheck/http/HttpHealthCheckServiceTest.java
```diff
@@ -496,6 +542,70 @@ public void should_call_target_using_instance_of_httpClient() throws Exception {
             );
         }
 
+        @Test
+        public void should_send_custom_host_header_when_configured() throws Exception {
```
aqibmohammadkhan, line 546: this test waits up to 10 s on a `CountDownLatch` driven by `reporterService.report(any())` with `schedule = "* * * * * *"`. That’s potentially slow on CI and susceptible to the second boundary. Since the public `buildRequestOptions` assertion test already covers the unit behaviour, the e2e test could use a shorter timeout or switch to an awaitility-style poll on `wiremock.findAll(...)`.

gravitee-apim-plugin/gravitee-apim-plugin-apiservice/gravitee-apim-plugin-apiservice-healthcheck-http/src/main/java/io/gravitee/apim/plugin/apiservice/healthcheck/http/HttpHealthCheckService.java
```diff
@@ -331,12 +331,23 @@ private RequestOptions buildRequestOptions(final ExecutionContext ctx, final Htt
...
         requestOptions
             .setMethod(HttpMethod.valueOf(request.method().name()))
             .setURI(target.getQuery() == null ? target.getPath() : target.getPath() + "?" + target.getQuery())
-            .setPort(VertxHttpClientFactory.getPort(target, isSsl))
+            .setPort(targetPort)
             .setSsl(isSsl)
-            .setHost(target.getHost());
+            .setHost(effectiveHost);
+
+        if (customHost != null && !customHost.isBlank()) {
```
aqibmohammadkhan, line 346: this condition `customHost != null && !customHost.isBlank()` appears multiple times, can be extracted.

```
final boolean hasCustomHost = customHost != null && !customHost.isBlank();
final String effectiveHost = hasCustomHost ? customHost : target.getHost();
...
if (hasCustomHost) {
    requestOptions.setServer(SocketAddress.inetSocketAddress(targetPort, target.getHost()));
}
```

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
dev-marek: nope, those are different headers

aqibmohammadkhan (approved):
LGTM. wdyt about resolving the sonar comments too.

PR #16369, opened 2026-04-13
Apim 13615 api product name detail

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-13615
## Description
Screenshot 2026-04-13 at 10 26 49 AM


gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-details/env-logs-details.component.html
```diff
@@ -79,6 +79,11 @@ <h3>Overview</h3>
                 <dl class="card__section__kv">
                   <env-logs-details-row label="Application">{{ log.application ?? '—' }}</env-logs-details-row>
                   <env-logs-details-row label="Plan">{{ log.plan?.name ?? '—' }}</env-logs-details-row>
+                  @if (log.apiProductName) {
```
carlos-andres-osorio: The "API Product" row is hidden when `log.apiProductName` is absent (`@if (log.apiProductName)`). The row should always be visible — logs not associated with an API Product should display `Standalone API` as the value, so operators can clearly distinguish product-routed logs from non-product-routed ones.

carlos-andres-osorio: As discussed, if the field is missing in the back-end response, the UI will show `-`, similar to the other fields.

carlos-andres-osorio (approved):
Approved with a minor comment.

PR #16387, opened 2026-04-14
fix(rest-api): return all groups for group admins in paginated search

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-13497

## Description

Fix Group Admin users only seeing their administered groups in paginated search. `search()` now returns all groups and sets `manageable` flag per group, matching `findAll()` behavior.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/GroupServiceImplTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/GroupServiceImplTest.java
@@ -83,10 +83,10 @@ public class GroupServiceImplTest {
     private EventManager eventManager;
 
     @Test
-    public void search_hasNotPermissionCreateUpdateDelete() {
+    public void search_should_return_all_groups_and_mark_administered_as_manageable_when_user_is_group_admin() {
         ExecutionContext executionContext = GraviteeContext.getExecutionContext();
         int pageNumber = 1;
-        int pageSize = 2;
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/GroupServiceImplTest.java
```diff
@@ -121,27 +118,76 @@ public void search_hasNotPermissionCreateUpdateDelete() {
...
+            () -> assertThat(searchResult.getContent()).hasSize(3),
+            () -> assertThat(searchResult.getContent().stream().map(GroupEntity::getId).toList()).containsExactly("gr1", "gr2", "gr3"),
+            () -> assertThat(searchResult.getContent().get(0).isManageable()).isTrue(),
+            () -> assertThat(searchResult.getContent().get(1).isManageable()).isFalse(),
+            () -> assertThat(searchResult.getContent().get(2).isManageable()).isFalse()
+        );
+    }
+
+    @Test
+    public void search_should_return_all_groups_with_manageable_false_when_user_is_not_group_admin() {
```
aqibmohammadkhan, line 142: this actually tests the case where the system role exists but the user holds no memberships under it. The case where the system role doesn't exist at all is untested.
wdyt about adding that case?

thomasballivet: yes good catch. Done

aqibmohammadkhan (approved):
LGTM

PR #16393, opened 2026-04-14
fix: avoid npe on certificate name equality test

a-cordier:


kovaren (approved):

kovaren:
@dev-marek @JedrzejJanasiak could you please check if this change is enough for a fix

PR #16408, opened 2026-04-16
refactor(redis): migrate rate-limit and distributed-sync to VertxRedisClientFactory

wbabyte:
## Summary

- Migrates `RedisConnectionFactory` to build `RedisClientOptions` (from `gravitee-plugin-common-configurations`) instead of Vert.x `RedisOptions` directly.
- `RedisClient` now takes `VertxRedisClientFactory` + `RedisClientOptions`, using the factory for both initial connection and reconnection — SSL/sentinel/keystore option-building routes through the factory's shared plumbing.
- Removes duplicated SSL/truststore/keystore code (PEM/PKCS12/JKS handling moved to plugin-common-configurations types, applied by factory via `SslOptionsMapper` + node SSL classes).
- No config changes — same `ratelimit.redis.*` / `distributed-sync.redis.*` YAML properties (including `tlsProtocols`, `tlsCiphers`, `alpn`, `openssl`, `hostnameVerificationAlgorithm`).

## Depends on / tracks

- Fix for: [APIM-13779](https://gravitee.atlassian.net/browse/APIM-13779)
- Epic: [APIM-13553](https://gravitee.atlassian.net/browse/APIM-13553)
- Requires upstream releases:
  - [gravitee-io/gravitee-node#551](https://github.com/gravitee-io/gravitee-node/pull/551) — fixes `VertxRedisClientFactory.configureSsl` hostname algorithm + adds `tlsProtocols`/`tlsCiphers`/`alpn` on node-vertx `SslOptions`. Needs release after merge (will be > 9.0.0-alpha.4).
  - [gravitee-io/gravitee-plugin-common-configurations#19](https://github.com/gravitee-io/gravitee-plugin-common-configurations/pull/19) — surfaces the 4 new SSL fields on `common-configurations.SslOptions`. Needs release after merge (will be > 1.2.1).
- After both releases: bump `gravitee-node.version` and `gravitee-plugin-common-configurations.version` in this PR's `pom.xml` to the new releases.

## Integration test matrix

Verified locally on a real Redis (standalone+password :6401, standalone+password+TLS/PEM :6402, host-based sentinel master:6501 + replica:6502 + sentinels :26501-26503) with patched SNAPSHOTs of the two upstream PRs installed:

| Scenario | Result |
|---|---|
| standalone + password | PASS — 2 ready clients, keys present |
| standalone + TLS + PEM + `trustAll=false` + `hostnameVerificationAlgorithm=HTTPS` + `tlsProtocols=TLSv1.2,TLSv1.3` | PASS — no `Missing hostname verification algorithm` error |
| sentinel + master/sentinel auth | PASS — sentinel cluster resolves master, keys replicated to replica |

Reference docker-compose setups: `docker/quick-setup/distributed-sync/`, `docker/quick-setup/redis-rate-limit/`.

## Known CI state

CI will be red on this PR until the two upstream PRs merge and release:

- PR compile references `SslOptions.setTlsProtocols()` / `setTlsCiphers()` / `setAlpn()` / `setOpenSsl()`, which don't exist in stock `common-configurations 1.2.1`. Compiles against the patched SNAPSHOT installed locally.
- Stock `node-vertx 9.0.0-alpha.4` has the `VertxRedisClientFactory.configureSsl` bug (no `HTTPS` branch for `hostnameVerifier=true`), so runtime TLS with non-trustAll fails until the fix ships.

## Test plan

- [x] `RedisConnectionFactoryTest` — 18/18 pass (standalone, sentinel, SSL with PEM/PKCS12/JKS, timeouts, username/password, error cases)
- [x] Integration: gateway with `ratelimit.type: redis` + `distributed-sync.type: redis` — rate-limit repo healthy, 1 connection per gateway per repo, `Repository [RATE_LIMIT] loaded by redis` and `Repository [DISTRIBUTED_SYNC] loaded by redis`, keys `distributed_sync_state:*`, `distributed_event::NODE_METADATA:*` present.
- [x] Integration with TLS (PEM truststore, `hostnameVerificationAlgorithm=HTTPS`) — previously regressed, now works after upstream fix.
- [x] Integration with sentinel (master auth + sentinel auth) — works; factory log `sentinel=mymaster(3 nodes)`.
- [ ] CI green — blocked on upstream PR merges + releases + version bumps here.

[APIM-13779]: https://gravitee.atlassian.net/browse/APIM-13779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APIM-13553]: https://gravitee.atlassian.net/browse/APIM-13553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/common/RedisConnectionFactory.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/common/RedisConnectionFactory.java
@@ -15,21 +15,26 @@
  */
 package io.gravitee.repository.redis.common;
 
+import io.gravitee.node.vertx.client.redis.VertxRedisClientFactory;
+import io.gravitee.plugin.configurations.redis.HostAndPort;
+import io.gravitee.plugin.configurations.redis.RedisClientOptions;
+import io.gravitee.plugin.configurations.redis.RedisSentinelOptions;
+import io.gravitee.plugin.configurations.ssl.KeyStore;
+import io.gravitee.plugin.configurations.ssl.SslOptions;
+import io.gravitee.plugin.configurations.ssl.TrustStore;
+import io.gravitee.plugin.configurations.ssl.jks.JKSKeyStore;
+import io.gravitee.plugin.configurations.ssl.jks.JKSTrustStore;
+import io.gravitee.plugin.configurations.ssl.none.NoneKeyStore;
+import io.gravitee.plugin.configurations.ssl.none.NoneTrustStore;
+import io.gravitee.plugin.configurations.ssl.pem.PEMKeyStore;
+import io.gravitee.plugin.configurations.ssl.pem.PEMTrustStore;
+import io.gravitee.plugin.configurations.ssl.pkcs12.PKCS12KeyStore;
+import io.gravitee.plugin.configurations.ssl.pkcs12.PKCS12TrustStore;
 import io.gravitee.repository.redis.vertx.RedisClient;
 import io.vertx.core.Vertx;
-import io.vertx.core.net.JksOptions;
-import io.vertx.core.net.NetClientOptions;
-import io.vertx.core.net.OpenSSLEngineOptions;
-import io.vertx.core.net.PemKeyCertOptions;
-import io.vertx.core.net.PemTrustOptions;
-import io.vertx.core.net.PfxOptions;
-import io.vertx.redis.client.RedisClientType;
-import io.vertx.redis.client.RedisOptions;
-import io.vertx.redis.client.RedisRole;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.TimeUnit;
 import lombok.CustomLog;
 import org.springframework.core.env.Environment;
 import org.springframework.util.StringUtils;
@@ -47,7 +52,6 @@ public class RedisConnectionFactory {
     private final Map<String, String> scripts;
 
     private static final String SENTINEL_PARAMETER_PREFIX = "sentinel.";
-
     private static final String PASSWORD_PARAMETER = "password";
 
     private static final String STORE_FORMAT_JKS = "JKS";
@@ -67,27 +71,26 @@ public class RedisConnectionFactory {
     }
 
     public RedisClient createRedisClient() {
-        return new RedisClient(vertx, buildRedisOptions(), scripts);
+        RedisClientOptions options = buildRedisClientOptions();
+        VertxRedisClientFactory factory = new VertxRedisClientFactory(vertx);
+        return new RedisClient(vertx, factory, options, scripts);
     }
 
-    protected RedisOptions buildRedisOptions() {
-        final RedisOptions options = new RedisOptions();
+    protected RedisClientOptions buildRedisClientOptions() {
+        RedisClientOptions.RedisClientOptionsBuilder builder = RedisClientOptions.builder();
 
         boolean ssl = readPropertyValue(propertyPrefix + "ssl", boolean.class, false);
         String username = readPropertyValue(propertyPrefix + "username", String.class, null);
+        String password = readPropertyValue(propertyPrefix + PASSWORD_PARAMETER, String.class);
+
+        builder.host(readPropertyValue(propertyPrefix + "host", String.class, "localhost"));
+        builder.port(readPropertyValue(propertyPrefix + "port", int.class, 6379));
+        builder.username(username);
+        builder.password(password);
+        builder.useSsl(ssl);
 
         if (isSentinelEnabled()) {
-            // Sentinels + Redis master / replicas
             log.debug("Redis repository configured to use Sentinel connection");
-
-            options.setType(RedisClientType.SENTINEL);
-            List<HostAndPort> sentinelNodes = getSentinelNodes();
-
-            // Redis Password
-            String password = readPropertyValue(propertyPrefix + PASSWORD_PARAMETER, String.class);
-            sentinelNodes.forEach(hostAndPort ->
-                options.addConnectionString(hostAndPort.withPassword(password).withSsl(ssl).toConnectionString())
-            );
 
             String redisMaster = readPropertyValue(propertyPrefix + SENTINEL_PARAMETER_PREFIX + "master", String.class);
             if (!StringUtils.hasText(redisMaster)) {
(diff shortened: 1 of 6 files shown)


gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/common/RedisConnectionFactory.java
```diff
@@ -67,241 +71,186 @@ public RedisConnectionFactory(
...
+        String tlsCiphers = readPropertyValue(propertyPrefix + "tlsCiphers", String.class, "");
+        if (StringUtils.hasText(tlsCiphers)) {
+            sslOptions.setTlsCiphers(new java.util.ArrayList<>(StringUtils.commaDelimitedListToSet(tlsCiphers)));
+        }
 
-            if (!trustAll) {
-                // Client truststore configuration (trust server certificate).
-                configureTrustStore(options.getNetClientOptions());
-            } else {
-                log.warn("Redis repository configured with ssl and trustAll which is not a good practice for security");
```
paulatulis, line 167: Did you mean to get rid of the warning?

wbabyte: yes, not a big deal honestly.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/common/RedisConnectionFactory.java
```diff
@@ -67,241 +71,186 @@ public RedisConnectionFactory(
...
-                    break;
-            }
+        if (!StringUtils.hasText(keystoreType)) {
+            return new NoneKeyStore();
         }
+
+        return switch (keystoreType.toUpperCase()) {
+            case STORE_FORMAT_PEM -> {
+                PEMKeyStore pemKeyStore = new PEMKeyStore();
+                String certPath = readPropertyValue(keystorePropertyPrefix + "certificates[0].cert", String.class);
```
carlos-andres-osorio: The old `buildKeyStore()` iterated `certificates[0..n]`, calling `addCertPath`/`addKeyPath` for each entry. This reads only `certificates[0]`. An operator with two or more PEM cert/key pairs configured in YAML will silently lose all but the first after upgrading — no error or warning is emitted.

If `PEMKeyStore` is architecturally single-cert, that's a breaking change from the previous behaviour and should be documented explicitly (e.g. a startup warning when `certificates[1+]` is present in config, or a migration note).

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/common/RedisConnectionFactory.java
```diff
@@ -67,241 +71,186 @@ public RedisConnectionFactory(
...
+        SslOptions sslOptions = new SslOptions();
+        boolean trustAll = readPropertyValue(propertyPrefix + "trustAll", boolean.class, true);
+        sslOptions.setTrustAll(trustAll);
 
-            // TLS Ciphers
-            StringUtils.commaDelimitedListToSet(readPropertyValue(propertyPrefix + "tlsCiphers", String.class, "")).forEach(cipherSuite ->
-                options.getNetClientOptions().addEnabledCipherSuite(cipherSuite.strip())
-            );
+        String hostnameVerificationAlgorithm = readPropertyValue(propertyPrefix + "hostnameVerificationAlgorithm", String.class, "NONE");
+        sslOptions.setHostnameVerifier(!"NONE".equals(hostnameVerificationAlgorithm));
```
carlos-andres-osorio: The previous code forwarded the algorithm string (`HTTPS`, `LDAPS`, …) directly to `netClientOptions.setHostnameVerificationAlgorithm()`. This collapses it to a boolean: any value other than `"NONE"` becomes `hostnameVerifier=true`, and the factory then applies `HTTPS` as the algorithm.

An operator who had `hostnameVerificationAlgorithm: LDAPS` configured will silently get HTTPS verification after the upgrade with no warning or migration note.

gravitee-apim-repository/gravitee-apim-repository-redis/src/test/java/io/gravitee/repository/redis/common/RedisConnectionFactoryTest.java
```diff
@@ -51,64 +51,67 @@ void setUp() {
...
+        assertThat(options.getSentinel()).isNotNull();
+        assertThat(options.getSentinel().getMasterId()).isEqualTo("redis-master");
+        assertThat(options.getSentinel().getNodes()).hasSize(2);
+        assertThat(options.getSentinel().getNodes().get(0).getHost()).isEqualTo("sent1");
+        assertThat(options.getSentinel().getNodes().get(1).getHost()).isEqualTo("sent2");
     }
 
     @Test
-    void shouldReturnRedisOptionsWithSentinelsSecuredEndpoints() {
+    void shouldReturnRedisClientOptionsWithSentinelAndSsl() {
```
carlos-andres-osorio, line 99: The old test explicitly asserted `hostnameVerificationAlgorithm` was `""` (disabled) on the sentinel+SSL path — a regression guard for the case where SSL is enabled without setting an explicit algorithm. That guard is gone here: no `hostnameVerificationAlgorithm` property is set, so the default is `"NONE"` → `hostnameVerifier=false`, but `isHostnameVerifier()` is never asserted. A change that accidentally flips the default to `true` on this path would not be caught.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/common/RedisConnectionFactory.java
wbabyte: Good catch — and you're right to flag it as a breaking change for documented config. Resolved by surfacing multi-cert support on the shared `PEMKeyStore` upstream rather than warning and proceeding.

Chain:
- [gravitee-io/gravitee-plugin-common-configurations#20](https://github.com/gravitee-io/gravitee-plugin-common-configurations/pull/20) — adds `certPaths: List<String>` + `keyPaths: List<String>` to `PEMKeyStore` (additive; single-cert path unchanged when lists are null).
- [gravitee-io/gravitee-node#556](https://github.com/gravitee-io/gravitee-node/pull/556) — `node-vertx.PEMKeyStore.keyCertOptions()` iterates the lists when populated, with a size-mismatch guard.
- This PR (commit `f302ca8962`) — `RedisConnectionFactory.buildKeyStore()` restores the original `for(idx = 0; ...; idx++)` loop and routes 2+ entries through `setCertPaths`/`setKeyPaths`, single-entry through the original `setCertPath`/`setKeyPath` for byte-identical behavior in that case.

Test added: `shouldReturnRedisClientOptionsWithMultiCertPemKeystore` — asserts both lists arrive on the resulting `PEMKeyStore`.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/common/RedisConnectionFactory.java
wbabyte: Right call — algorithm string is now preserved end-to-end instead of collapsed.

Chain:
- [gravitee-io/gravitee-plugin-common-configurations#20](https://github.com/gravitee-io/gravitee-plugin-common-configurations/pull/20) — adds `hostnameVerificationAlgorithm: String` to shared `SslOptions`. Null / `"NONE"` keeps the existing boolean as the source of truth (no behavior change for current callers).
- [gravitee-io/gravitee-node#556](https://github.com/gravitee-io/gravitee-node/pull/556) — `VertxRedisClientFactory.configureSsl` now calls `resolveHostnameVerificationAlgorithm(...)`. Precedence: explicit non-`NONE` algorithm string applied verbatim; otherwise falls back to the boolean (`true` → `HTTPS`, `false` → `""`).
- This PR (commit `f302ca8962`) — `buildSslOptions()` passes the raw `hostnameVerificationAlgorithm` value to `SslOptions.setHostnameVerificationAlgorithm(...)`. `LDAPS` survives unchanged.

Test added: `shouldPreserveLdapsHostnameVerificationAlgorithm` — asserts `LDAPS` arrives on the resulting `SslOptions`.

gravitee-apim-repository/gravitee-apim-repository-redis/src/test/java/io/gravitee/repository/redis/common/RedisConnectionFactoryTest.java
wbabyte: Restored, plus extended:

- Commit `e392fc796e` — added back `assertThat(options.getSsl().isHostnameVerifier()).isFalse()` on the sentinel+SSL path so an accidental flip of the default would fail the test.
- Commit `f302ca8962` — added `assertThat(options.getSsl().getHostnameVerificationAlgorithm()).isEqualTo("NONE")` alongside it, now that the algorithm string is plumbed through. Both assertions in the same test cover the boolean-default and string-default surfaces.

carlos-andres-osorio (approved):
Incremental review — all three prior findings resolved:

- **PEM multi-cert truncation** — loop now iterates all `certificates[n]` entries; routes to `setCertPaths`/`setKeyPaths` for multi-cert.
- **`hostnameVerificationAlgorithm` collapsed to boolean** — both `setHostnameVerificationAlgorithm(rawString)` and `setHostnameVerifier(bool)` now called; `LDAPS` survives the round-trip.
- **Sentinel+SSL test missing default assertions** — `isHostnameVerifier()` and `getHostnameVerificationAlgorithm()` now asserted.

LGTM.

vikrantgravitee (approved):

PR #16414, opened 2026-04-16
feat: add observability endpoint for filter values

a-cordier:
see https://gravitee.atlassian.net/browse/APIM-12090

# feat: add observability endpoint for filter values

**Jira:** [APIM-12118](https://gravitee.atlassian.net/browse/APIM-12118) / [APIM-12090](https://gravitee.atlassian.net/browse/APIM-12090)

## Summary

Adds a new observability endpoint that returns the available values for a given analytics filter, so the UI can populate filter dropdowns dynamically instead of relying on free-text input.

- New REST resource `ObservabilityFilterValuesResource` exposing `GET .../observability/filters/{filterName}/values` (paginated, searchable).
- New use case `GetFilterValuesUseCase` + `FilterValuesQueryService` / `FilterValueNameResolver` domain services to resolve raw values (e.g. API IDs, application IDs, plan IDs) into human-readable labels.
- Repository layer: `AnalyticsRepository#searchFilterValues` with an Elasticsearch implementation (`FilterValuesQueryAdapter` / `FilterValuesResponseAdapter`) and a no-op fallback.
- OpenAPI spec updated (`openapi-analytics.yaml`) and `AnalyticsDefinitionMapper` extended to expose filter metadata.

## Test plan

- [ ] Unit tests pass (`GetFilterValuesUseCaseTest`, `FilterValuesQueryServiceImplTest`, `FilterValueNameResolverImplTest`, adapter tests)
- [ ] Integration test `ObservabilityFilterValuesResourceTest` passes
- [ ] Manual check: call the endpoint against an env with ES analytics and verify values + pagination + search

[APIM-12118]: https://gravitee.atlassian.net/browse/APIM-12118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APIM-12090]: https://gravitee.atlassian.net/browse/APIM-12090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/FilterValuesQueryAdapter.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/FilterValuesQueryAdapter.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.elasticsearch.v4.analytics.engine.adapter;
+
+import io.gravitee.repository.log.v4.model.analytics.FilterValuesQuery;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import java.util.ArrayList;
+
+public class FilterValuesQueryAdapter {
+
+    private static final String API_ID_FIELD = "api-id";
+
+    public String adapt(FilterValuesQuery query) {
+        var root = new JsonObject();
+        root.put("size", 0);
+
+        var boolFilter = new JsonArray();
+        if (query.from() != null && query.to() != null) {
+            boolFilter.add(JsonObject.of("range", JsonObject.of("@timestamp", JsonObject.of("gte", query.from(), "lte", query.to()))));
+        }
+
+        if (query.searchPattern() != null && !query.searchPattern().isBlank()) {
+            // Exact match on the keyword field (case-insensitive); avoids costly leading/trailing wildcards.
+            var termValue = JsonObject.of("value", query.searchPattern(), "case_insensitive", true);
+            boolFilter.add(JsonObject.of("term", JsonObject.of(query.esFieldName(), termValue)));
+        }
+
+        if (query.apiIds() != null) {
+            if (query.apiIds().isEmpty()) {
+                boolFilter.add(JsonObject.of("match_none", JsonObject.of()));
+            } else {
+                boolFilter.add(JsonObject.of("terms", JsonObject.of(API_ID_FIELD, new JsonArray(new ArrayList<>(query.apiIds())))));
+            }
+        }
+
+        if (!boolFilter.isEmpty()) {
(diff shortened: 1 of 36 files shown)


a-cordier:
> ## Code Review
> This pull request introduces a new observability feature to retrieve distinct values for various filters (such as APIs, applications, plans, and gateway IDs) through a new REST endpoint. The implementation spans from the repository layer, using Elasticsearch composite aggregations, to the REST API layer with a dedicated use case for handling ENUM, KEYWORD, and ID-based filters. Feedback highlights that the current pagination for Elasticsearch-backed filters is inefficient due to sequential requests and lacks strict upper bounds. Additionally, fetching all applications in an environment for in-memory filtering is identified as a scalability risk, and the use of exact-match term queries for search patterns is noted as inconsistent with other filter types.

Exact match is a performance trade off for ES so that we don't introduce a performance risk that would be huge (full index scan on each key stroke from the user)

Applications fetch is a trade of for the sake of keeping the PR as tight as it can be / will be addressed later on

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/FilterValuesQueryAdapter.java
```diff
@@ -0,0 +1,74 @@
...
+import io.gravitee.repository.log.v4.model.analytics.FilterValuesQuery;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import java.util.ArrayList;
+
+public class FilterValuesQueryAdapter {
+
+    private static final String API_ID_FIELD = "api-id";
+
+    public String adapt(FilterValuesQuery query) {
```
SamHardi, line 27: Could you split this method into little specialized ones to improve readability?
Each "if" blocks look really specific

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/FilterValuesResponseAdapter.java
```diff
@@ -0,0 +1,93 @@
...
+        var afterKey = new HashMap<String, Object>();
+        afterKeyNode.fields().forEachRemaining(entry -> afterKey.put(entry.getKey(), extractJsonValue(entry.getValue())));
+        return afterKey;
+    }
+
+    private Object extractJsonValue(JsonNode node) {
+        if (node.isTextual()) return node.asText();
+        if (node.isLong()) return node.asLong();
+        if (node.isInt()) return node.asInt();
+        if (node.isDouble()) return node.asDouble();
```
SamHardi, line 89:
```suggestion
        if (node.isNumber()) return node.numberValue();
```
WDYT?

OR 


```suggestion
        return switch (node.getNodeType()) {
        case STRING -> node.asText();
        case NUMBER -> node.numberValue();
        case BOOLEAN -> node.asBoolean();
        case NULL -> null;
        default -> node.asText(); 
    };
```

SamHardi (approved):

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/FilterValuesQueryAdapter.java
a-cordier: Addressed in an internal issue later

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/FilterValuesResponseAdapter.java
a-cordier: Addressed as part of an internal issue later

PR #16424, opened 2026-04-16
feat: expose apiProduct part one

paulatulis:
## Issue

Part one of two https://gravitee.atlassian.net/browse/APIM-13545

## Description

This PR adds the data foundation: all model and repository changes needed to carry `apiProductId` and `apiProductName` through the stack. Business logic and REST API exposure will come in part two.


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/log/model/BaseConnectionLog.java
```diff
@@ -50,4 +50,12 @@ public class BaseConnectionLog {
     private String errorComponentName;
     private String errorComponentType;
     private List<ConnectionDiagnosticModel> warnings;
+    /** Read from the v4 metrics index (ES field: {@code api-product-id}). Null for v2 and when no product is associated. */
+    private String apiProductId;
+
+    /**
```
carlos-andres-osorio, line 56: nit: This references a jira issue and if this file is not in scope for that change this temporary reference will be left there. Not sure it's necessary givne that this value will be validated during testing..

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/log/adapter/connection/SearchMetricsResponseAdapterTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
+class SearchMetricsResponseAdapterTest extends AbstractAdapterTest {
+
+    private final ObjectMapper objectMapper = new ObjectMapper();
+
+    /**
+     * Builds a search hit with a custom index name, allowing testing of the v2 request-index path
+     * (which {@link AbstractAdapterTest#buildSearchHit} cannot do because it hardcodes "gravitee-v4-log").
+     */
```
carlos-andres-osorio: nit: The Javadoc cross-references `AbstractAdapterTest#buildSearchHit` and explains why it can't be reused here. That reference will rot if `AbstractAdapterTest` changes — the comment becomes misleading the moment its implementation assumptions shift. The method and parameter names already convey what this helper does; consider dropping the block so the test file isn't coupled to another class's internals.

carlos-andres-osorio (approved):
Left a couple minor comments, but implementation is good.

PR #16426, opened 2026-04-16
[gamma] - graphene-core eslint configuration

Okhelifi:
## Description

In this PR we:
- replace gravitee graphene by the new core package. It has simply be renamed so no major change for the moment on it
- extend eslint core configuration to ensure that as a consumer I respect the same code rules than the graphene project
- fix the existing lint error raised by the new configuration shared


gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/lib/buildLinearBreadcrumbs.ts
```diff
@@ -13,12 +13,12 @@
...
  * limitations under the License.
  */
-import type { BreadcrumbEntry } from '@gravitee/graphene';
+import type { BreadcrumbEntry } from '@gravitee/graphene-core';
 import type { NavigateFunction } from 'react-router-dom';
 
 /**
  * Duplicate of gravitee-gamma-control-plane-webui/src/shared/breadcrumbs/buildLinearBreadcrumbs.ts —
- * TODO(graphene): replace this file with `import { buildLinearBreadcrumbs } from '@gravitee/graphene'` when available.
+ * TODO(graphene): replace this file with `import { buildLinearBreadcrumbs } from '@gravitee/graphene-core'` when available.
```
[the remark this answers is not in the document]
jourdiw: Will address in this issue: https://github.com/gravitee-io/gravitee-ui-graphene/issues/115

gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/auth/components/LoginPage.tsx
```diff
@@ -101,7 +101,6 @@ export function LoginPage() {
                                 required
                                 autoComplete="username"
                                 placeholder="Enter your username"
-                                autoFocus
                             />
```
gcusnieux, line 104: A login page is one of the legitimate cases for `autoFocus`. Removing it means users have to click into the field every time. Could we keep it with an eslint-disable?

```suggestion
                                autoComplete="username"
                                placeholder="Enter your username"
                                // eslint-disable-next-line jsx-a11y/no-autofocus
                                autoFocus
```

gcusnieux:
Both `.eslintrc.js` files are now byte-identical, and the next gamma module will copy-paste a third one. Could we extract a small shared factory so each module becomes a one-liner?

```js
// gravitee-gamma/eslint-gamma-base.js
const grapheneConfig = require('@gravitee/graphene-core/eslint-legacy');

module.exports = projectDir => ({
    ...grapheneConfig,
    settings: {
        ...grapheneConfig.settings,
        'import/resolver': {
            typescript: {
                project: [projectDir + '/tsconfig.json'],
                alwaysTryTypes: true,
            },
        },
    },
    rules: {
        ...grapheneConfig.rules,
        'import/order': [
            'error',
            {
                groups: ['builtin', 'external', 'internal', ['sibling', 'parent'], 'index', 'unknown'],
                'newlines-between': 'always',
                alphabetize: { order: 'asc', caseInsensitive: true },
            },
        ],
    },
});
```

Then each module's `.eslintrc.js`:

```js
module.exports = require('../eslint-gamma-base')(__dirname);
```

Happy to do it as a follow-up if easier.

gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/auth/components/LoginPage.tsx
jourdiw: Done!

gcusnieux (approved):
LGMT

jourdiw (approved):

PR #16437, opened 2026-04-16
feat(rest-api): native kafka connection logs

aditya-goyal01:
## Issue

https://gravitee.atlassian.net/browse/APIM-13589 · https://gravitee.atlassian.net/browse/APIM-13592

### Schema (absorbs APIM-13592)

- `NativeAnalytics.reporterMetricsEnabled` — flat boolean on the Native v4 analytics block, default `true`, independent of `analytics.enabled`
- Validator does null-safety only; no cross-flag rejection
- Exposed on the shared `Analytics` OpenAPI schema; silently discarded on HTTP v4 APIs

### Read-side query

- Flag-agnostic `GET /environments/{env}/apis/{id}/logs`: bare `apiCrudService.get()` for 404 propagation, then pass-through.
- New query params:
  - `nativeKafkaClientIds` — terms filter on `additional-metrics.keyword_native-kafka_client-id`
  - `connectionStatuses` — terms filter on `additional-metrics.keyword_native-kafka_connection-status`, OpenAPI enum schema (`CONNECTED`/`SESSION_ERROR`/`CONNECTION_ERROR`/`INTERNAL_ERROR`)
- `errorKeys` semantics updated to match the reactor: values are Kafka protocol codes from `org.apache.kafka.common.protocol.Errors.name()` (e.g. `SASL_AUTHENTICATION_FAILED`, `UNKNOWN_SERVER_ERROR`) — not the former `GATEWAY_KAFKA_*` buckets.

### Filtering guidance

- Lifecycle phase → `connectionStatuses`.
- Specific protocol failure → `errorKeys`.

## Not in scope

Gateway emission (APIM-13587, reactor repo) · `nativeKafkaConsumerGroupIds` filter (backing field deferred to APIM-13677) · Console UI.

## Testing

Covered across the 5 filter layers (ES field → repo query → REST filter record → REST↔repo bridge → REST endpoint) plus OpenAPI schema. Adapter-level null/empty/present tests, ES repo integration test with real Kafka fixture data, MapStruct param→filter mapping, CRUD impl forwarding, use-case level filtering, and full-stack REST resource integration.

## Co-dependent PR

APIM-13587 in `gravitee-reactor-native-kafka` (PR [#301](https://github.com/gravitee-io/gravitee-reactor-native-kafka/pull/301)) emits the `connection-status` keyword and consumes `NativeAnalytics.isReporterMetricsEnabled()` to gate reporter writes.


gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/v4/nativeapi/NativeAnalytics.java
```diff
@@ -17,21 +17,27 @@
...
-@Getter
-@Setter
-@ToString
-@EqualsAndHashCode
+@Data
 public class NativeAnalytics {
 
     @Builder.Default
-    protected boolean enabled = true;
+    private boolean enabled = true;
```
ThibaudAV, line 30: For me, the Booleans here determine what analytics the gateway sends to ES or other reporting provider.  The Booleans are therefore targeted at the type of report

So the current boolean could be called `eventMetricsEnabled`(maybe just add comment for now) , as it triggers data to be sent to this metric. 
And the new one could be called `metricsEnabled`

this make sens 🤔 ?

I don't think we need a debug mode; in fact, I find it rather confusing. If we need more filtering, I think we'll do it another way, no ?

aditya-goyal01: > Done. 

- Added a `TODO` to rename variable from `NativeAnalytics.enabled` to `NativeAnalytics.eventMetricsEnabled`.
- Renamed variable from `NativeAnalytics.connectionMetricsEanbled` to `NativeAnalytics.reporterMetricsEanbled`.

> Note:

- Kafka reactor will send documents to Elasticsearch based on `NativeAnalytics.reporterMetricsEanbled`. 
- When `NativeAnalytics.reporterMetricsEanbled` is `true`, reactor will send documents on both the events, i.e. when connection succeeds and if connection fails.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
```diff
@@ -8903,6 +8911,30 @@ components:
                 type: array
                 items:
                     type: integer
+        errorKeys:
```
ThibaudAV, line 8951: why not status ? instead errorKeys or in addition to errorKeys ? To filter by log status type, isn't it?

phiz71 (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
aditya-goyal01: Done. A separate [commit](https://github.com/gravitee-io/gravitee-api-management/pull/16437/changes/ac4bd0e509c8aa9ddcbda9c12c4044da89c2ef3c) has been created.

phiz71 (approved):

PR #16438, opened 2026-04-16
feat: expose apiProductId part two

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-13545
## Description

Exposes `apiProductId` and `apiProductName` in the v4 connection logs REST API and adds the ability to filter logs by API Product.

Changes

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
@@ -2434,6 +2434,16 @@ paths:
                 - $ref: "#/components/parameters/applicationIds"
                 - $ref: "#/components/parameters/planIds"
                 - $ref: "#/components/parameters/methods"
+                - name: apiProductIds
+                  in: query
+                  description: Filter by API Product IDs. Only supported for v4 APIs.
+                  explode: false
+                  schema:
+                      type: array
+                      items:
+                          type: string
+                      example:
+                          - f5e6a5a0-1234-4b3a-9c1e-aabbccddeeff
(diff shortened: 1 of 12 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
```diff
@@ -7024,6 +7034,14 @@ components:
                   type: object
                   additionalProperties: true
                   description: A map of string keys to values
+                apiProductId:
+                  type: string
+                  description: The ID of the API Product associated with this log entry. Absent when the request is not associated with any API Product.
+                  example: product-abc-123
```
carlos-andres-osorio: product ID should be an UUID

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/log/model/BaseConnectionLog.java
```diff
@@ -53,9 +53,6 @@ public class BaseConnectionLog {
     /** The ID of the API Product associated with this connection log entry. Null when no product is associated. */
     private String apiProductId;
 
-    /**
-     * Resolved from {@code apiProductId} by {@code SearchApiV4ConnectionLogsUseCase} via {@code ApiProductQueryService}.
-     * This field is intentionally null in the data foundation layer and will be populated in APIM-13545 ST-3.
-     */
+    /** The human-readable name of the API Product. Resolved at query time from {@code apiProductId}. Null when no API Product is associated. */
```
carlos-andres-osorio: nit: not sure if this comment is necessary. the name makes the field purpose explicit and the explanation that it's based on apiProductId is easy to deduct.

carlos-andres-osorio (approved):

PR #16456, opened 2026-04-20
fix: refactor breadcrumbs in portal dashboard.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-13521

## Summary

Refactor dashboard breadcrumbs in **gravitee-apim-portal-webui-next** so they are managed by a shared `BreadcrumbService` and set explicitly from feature components, instead of being inferred from the router URL in `DashboardComponent`.

## Motivation

The previous implementation derived breadcrumb labels and links from URL segments after each `NavigationEnd`. That was brittle (special cases like subscription IDs, fallbacks to raw segments) and duplicated routing knowledge in the shell. Child routes can now declare accurate, localized breadcrumbs where the data lives.

## What changed

- **New `BreadcrumbService`** (`providedIn: 'root'`): holds breadcrumbs in a `signal`, with `setBreadcrumbs()` and `clear()`.
- **`DashboardComponent`**: binds `[breadcrumbs]="breadcrumbService.breadcrumbs()"` and clears the service on destroy; removes router-event / URL-parsing breadcrumb logic.
- **Shared builders**: `application-breadcrumbs.ts` and `subscription-breadcrumbs.ts` centralize list-level crumbs and optional `url` when a segment should be clickable (e.g. on detail pages).
- **Components set their own trail**:
  - Applications list, create application (Applications → Create), application details (Applications → app name).
  - Subscriptions list, subscription details (Subscriptions → “Subscription &lt;id&gt;”).

https://github.com/user-attachments/assets/cbb5e2f3-0c6f-40cb-9c32-9203c4afd202

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application.component.ts
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application.component.ts
@@ -13,11 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { Component, input } from '@angular/core';
+import { Component, effect, inject, input } from '@angular/core';
 import { MatTabLink, MatTabNav, MatTabNavPanel } from '@angular/material/tabs';
 import { RouterLink, RouterLinkActive, RouterOutlet } from '@angular/router';
 
 import { Application } from '../../../entities/application/application';
+import { BreadcrumbService } from '../../../services/breadcrumb.service';
+import { applicationListBreadcrumb } from '../applications/application-breadcrumbs';
 
 @Component({
   selector: 'app-application',
@@ -26,5 +28,13 @@ import { Application } from '../../../entities/application/application';
   styleUrl: './application.component.scss',
 })
 export default class ApplicationComponent {
-  application = input.required<Application>();
+  private readonly breadcrumbService = inject(BreadcrumbService);
+  readonly application = input.required<Application>();
+
+  constructor() {
+    effect(() => {
+      const application = this.application();
+      this.breadcrumbService.set([applicationListBreadcrumb(true), { id: `application-${application.id}`, label: application.name }]);
+    });
+  }
 }
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/applications/create-application/create-application.component.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/applications/create-application/create-application.component.spec.ts
@@ -34,8 +34,10 @@ import {
   fakeBackendToBackendApplicationType,
   fakeBrowserApplicationType,
 } from '../../../../entities/application/application.fixture';
+import { BreadcrumbService } from '../../../../services/breadcrumb.service';
 import { ObservabilityBreakpointService } from '../../../../services/observability-breakpoint.service';
 import { AppTestingModule, TESTING_BASE_URL } from '../../../../testing/app-testing.module';
+import { applicationListBreadcrumb } from '../application-breadcrumbs';
 
 describe('CreateApplicationComponent', () => {
   let fixture: ComponentFixture<CreateApplicationComponent>;
@@ -52,6 +54,10 @@ describe('CreateApplicationComponent', () => {
     fakeBackendToBackendApplicationType(),
     fakeBrowserApplicationType(),
   ];
+
+  function flushInitialApplicationTypes(): void {
+    httpTestingController.expectOne(`${TESTING_BASE_URL}/configuration/applications/types`).flush({ data: mockApplicationTypes });
+  }
 
   beforeEach(async () => {
     isMobileSignal = signal(false);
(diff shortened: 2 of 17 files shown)


gravitee-apim-portal-webui-next/src/services/breadcrumb.service.ts
```diff
@@ -0,0 +1,31 @@
...
+export class BreadcrumbService {
+  readonly breadcrumbs = signal<Breadcrumb[]>([]);
+
+  setBreadcrumbs(breadcrumbs: Breadcrumb[]): void {
+    this.breadcrumbs.set(breadcrumbs);
+  }
+
+  clear(): void {
+    this.breadcrumbs.set([]);
+  }
```
aqibmohammadkhan: Exposing the raw `WritableSignal` `(readonly breadcrumbs = signal<...>([]))` means any consumer can call `breadcrumbs.set(...)` and bypass the service. Since only the service's `setBreadcrumbs` / `clear` should mutate the state, consider narrowing the surface:

```
private readonly _breadcrumbs = signal<Breadcrumb[]>([]);
readonly breadcrumbs = this._breadcrumbs.asReadonly();
setBreadcrumbs(b: Breadcrumb[]): void { this._breadcrumbs.set(b); }
clear(): void { this._breadcrumbs.set([]); }
```

gravitee-apim-portal-webui-next/src/app/dashboard/applications/create-application/create-application.component.spec.ts
```diff
@@ -87,6 +89,15 @@ describe('CreateApplicationComponent', () => {
   });
 
   describe('Component initialization', () => {
+    it('should set breadcrumbs for create application', () => {
+      const breadcrumbService = TestBed.inject(BreadcrumbService);
+      expect(breadcrumbService.breadcrumbs()).toEqual([
+        applicationListBreadcrumb(true),
+        { id: 'create', label: $localize`:@@createApplicationBreadcrumb:Create` },
+      ]);
+      httpTestingController.expectOne(`${TESTING_BASE_URL}/configuration/applications/types`).flush({ data: mockApplicationTypes });
```
aqibmohammadkhan: The `expectOne(...)` here exists only to satisfy httpTestingController.verify() in afterEach. Consider either 
(a) placing the HTTP flush first (the assertion doesn't depend on it) or 
(b) extracting a small flushInitialTypes() helper so the test reads as purely about breadcrumbs

gravitee-apim-portal-webui-next/src/app/dashboard/dashboard.component.ts
```diff
@@ -39,43 +37,11 @@ const MENU_ITEMS: MenuItem[] = [
...
-            label = segment; // Fallback
-          }
-          breadcrumbs.push({ id: segment, label, url });
-        });
-        return breadcrumbs;
-      }),
-    ),
-  );
+  constructor() {
+    this.destroyRef.onDestroy(() => this.breadcrumbService.clear());
```
aqibmohammadkhan, line 45: `DashboardComponent.clear()` isn't covered by a test. There's no assertion that destroying `DashboardComponent` triggers `breadcrumbService.clear()`. 
This matters because when a user navigates out of /dashboard to e.g. /documentation, that behavior is the only thing preventing stale crumbs from flashing.

gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application.component.ts
```diff
@@ -26,5 +29,14 @@ import { Application } from '../../../entities/application/application';
...
+  private readonly breadcrumbService = inject(BreadcrumbService);
+  readonly application = input.required<Application>();
+  readonly breadcrumbs = computed<Breadcrumb[]>(() => {
+    const application = this.application();
+    return [applicationListBreadcrumb(true), { id: `application-${application.id}`, label: application.name }];
+  });
+
+  constructor() {
+    effect(() => this.breadcrumbService.setBreadcrumbs(this.breadcrumbs()));
+  }
```
aqibmohammadkhan: `computed` is unused outside the `effect`. `breadcrumbs` is a public readonly field that nothing references (template is RouterOutlet / tabs only). You can inline it and drop the computed:

```
constructor() {
  effect(() => {
    const app = this.application();
    this.breadcrumbService.setBreadcrumbs([
      applicationListBreadcrumb(true),
      { id: `application-${app.id}`, label: app.name },
    ]);
  });
}
```

gravitee-apim-portal-webui-next/src/app/dashboard/dashboard.component.ts
aqibmohammadkhan, line 1: `/dashboard/applications/:applicationId/logs/:logId` no longer gets a breadcrumb tail. Previously the URL-based logic produced something like `Applications / <appId> / logs / <logId>`. After this PR, every tab under `applications/:applicationId` — including the `logs/:logId detail` — shows only `Applications → <AppName>`. For the l`ogs/settings` tabs themselves this is fine (the MatTabNav already conveys the active tab), but the log detail page is a drill-down from the log table without a corresponding tab, so users lose a way back to the log list via breadcrumbs. Is this intended? 

if not, should we add a crumb contribution from ApplicationLogComponent (and optionally an applicationLogsBreadcrumb(...) builder that points back to .../logs).

kovaren: That's the intended behavior, we don't want the breadcrumbs to repeat the tabs. The subroutes are still there so the access by URL is not lost.

gravitee-apim-portal-webui-next/src/services/breadcrumb.service.ts
kovaren: Fixed!

gravitee-apim-portal-webui-next/src/app/dashboard/applications/create-application/create-application.component.spec.ts
kovaren: Fixed!

gravitee-apim-portal-webui-next/src/app/dashboard/dashboard.component.ts
kovaren: Fixed!

gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application.component.ts
kovaren: Fixed!

aqibmohammadkhan (approved):

PR #16464, opened 2026-04-20
fix: prevent forbidden error on API for users without API Products permission

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-13692

## Description

Fixes a 403 error on the API General Info page for users whose role does not include the ENVIRONMENT_API_PRODUCT[READ] permission. The root cause was that GET /apis/{apiId}/api-products was guarded by an environment-scoped permission that has no relationship to the API General Info page, which is access-controlled at the API level. The fix corrects the permission to API_DEFINITION[READ] (consistent with all other ApiResource sub-endpoints) and introduces a lightweight ApiProductInfo projection to avoid over-fetching full API Product data just to display the "Allow in API Product" toggle and "Included in" chip.

> Decoupled the “Allow in API Product” toggle from API Product permissions on the API page. The ability to allow an API to be added to an API Product should be determined by the API owner, so the permission-based condition has been removed.

> The toggle now explicitly empowers the API owner to control whether their API is exposed and available for other API Product owners to include in API Products.

https://github.com/user-attachments/assets/e20e0451-dedb-479f-8dfd-41e7a0af6d02


arpzilla-dev (approved):
lgtm

PR #16469, opened 2026-04-20
feat(portal): implement context loading for analytics query in portal

aqibmohammadkhan:
Introduce `PortalContextLoader` to provide analytics query context. Update `RestPortalConfiguration` with required beans and add comprehensive tests for context loading scenarios.

## Issue

https://gravitee.atlassian.net/browse/APIM-13612

## Description

Summary

- Introduces PortalContextLoader to build the analytics query context in the Portal.
- Wires the required beans in RestPortalConfiguration.
- Adds unit tests covering admin, authenticated non-admin, and anonymous (no security context) scenarios.
- Optimizes API lookups during context loading by excluding heavy fields (definition/picture) from repository searches.

Why

- Analytics queries need a consistent, correctly-scoped context (authorized API IDs, API names, app names) depending on the caller’s visibility/permissions.
- Avoids loading large API payloads when only metadata is required, improving performance and memory usage in large environments.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/analytics_engine/PortalContextLoader.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/analytics_engine/PortalContextLoader.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.infra.domain_service.analytics_engine;
+
+import io.gravitee.apim.core.analytics_engine.domain_service.AnalyticsQueryContextLoader;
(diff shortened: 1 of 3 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/analytics_engine/PortalContextLoader.java
```diff
@@ -0,0 +1,124 @@
...
+        var executionContext = new ExecutionContext(organizationId, environmentId);
+
+        List<Api> apis;
+        Map<String, String> applicationNamesById;
+
+        if (isAdmin()) {
+            apis = ApiMapper.INSTANCE.map(
+                apiRepository.search(new ApiCriteria.Builder().environmentId(environmentId).build(), API_CONTEXT_FIELDS)
+            );
+            applicationNamesById = Map.of();
```
kovaren, line 83: I might be missing something, but why are we not loading the applications for the admin? An admin can normally have applications

kovaren (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/analytics_engine/PortalContextLoader.java
aqibmohammadkhan: `applicationNamesById` is only a pre-warmed display-name cache used by `BucketNamesPostProcessor` to resolve APPLICATION bucket keys. It is not used for authorization scoping — admins already see everything via `authorizedApiIds`. When the map is empty, `BucketNamesPostProcessorImpl.populateNames` falls back to loading names for exactly the application IDs that appeared in the response buckets (see `loadApplicationNames`), so admins still get human-readable app names in the UI. Pre-loading every application in the environment for admins would be wasteful on large tenants; we mirror `ManagementContextLoader`, which also returns Map.of() for every case.

kovaren (approved):

PR #16477, opened 2026-04-20
fix(repository): support repositories.* prefix with fixed plugin and correct indentation

thomasballivet:
## Issue

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
@@ -271,66 +271,68 @@ secrets:
 
 # Management repository is used to store global configuration such as APIs, applications, apikeys, ...
 # If you use a JDBC repository, we recommend disabling liquibase scripts execution by the Gateway. Let the Management API do it.
-# management:
-#   type: jdbc
-#   jdbc:
-#     liquibase: false
+# repositories:
+#   management:
+#     type: jdbc
+#     jdbc:
+#       liquibase: false
 
 # This is the default configuration using MongoDB (single server)
 # For more information about MongoDB configuration, please have a look to:
 # - http://mongodb.github.io/mongo-java-driver/4.1/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html
-management:
-  type: mongodb                  # repository type
-  mongodb:                       # mongodb repository
-#    prefix:                      # collections prefix
-    dbname: ${ds.mongodb.dbname} # mongodb name (default gravitee)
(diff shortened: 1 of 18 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-standalone/gravitee-apim-rest-api-standalone-distribution/src/main/resources/config/gravitee.yml
jgiovaresco: I've understood that we add a new "level" with `repositories` line 160. We should also update the commented section because if you uncomment them the indentation seems incorrect 🤔

thomasballivet: thank you for pointing this. Done

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
phiz71: should we also format this ?

helm/templates/api/api-configmap.yaml
```diff
@@ -45,50 +45,51 @@ data:
...
+          password: {{ .Values.jdbc.password }}
+          {{- end }}
+          schema: {{ .Values.jdbc.schema | default "public" }}
+          liquibase: {{ .Values.jdbc.liquibase }}
+          {{- if .Values.jdbc.pool }}
+          pool:
+            {{ toYaml .Values.jdbc.pool | nindent 12 | trim -}}
+          {{- end }}
         {{- end }}
-      {{- end }}
```
jgiovaresco, line 91: are we sure? I don't understand what block it closes?

Also we should move the analytics here as it needs to be under `repositories` I'm not sure the chart works

jgiovaresco (changes requested):
We should test the chart before merging.

Also we should add a change in Chart.yml and Changelog to inform users about this change it should have no impact for them but it would be helpful to know that the chart has been updated if we find an issue later

helm/templates/api/api-configmap.yaml
thomasballivet: Done. Here is the output of `helm template test` :
```
repositories:
      management:
        type: mongodb
        mongodb:
          sslEnabled: false
          socketKeepAlive: false
          uri: mongodb://graviteeio-apim-mongodb-replicaset-headless:27017/gravitee?&replicaSet=rs0&connectTimeoutMS=30000
      analytics:
        type: elasticsearch
        elasticsearch:
          endpoints:
            - http://graviteeio-apim-elasticsearch-ingest-hl:9200
          index_mode: daily
          index: gravitee
```
The end block line 91 closes the if block line 88, and the end block line 92 closes the if block 51 now

jgiovaresco (approved):

PR #16482, opened 2026-04-20
fix(portal): add worker-src blob: directive to CSP for Redoc worker support

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-13798

## Description

Add `worker-src 'self' blob:` to default CSP in portal nginx configs so Redoc's blob workers aren't blocked.


jourdiw (approved):
LGTM

PR #16484, opened 2026-04-20
fix(helm): render upgrader additional plugins

gutek9:
## Issue

https://gravitee.atlassian.net/browse/APIM-13721

## Description

Fixes Helm rendering for the API upgrader job when `api.upgrader=true` and `api.additionalPlugins` are configured.

The upgrader job now passes the Helm context to the shared plugin init container helper, matching the regular API deployment behavior. This prevents the OpenShift security context helper from failing with a nil `.context.Values` reference.

Added regression tests for:
- rendering additional plugin init containers in the upgrader job
- OpenShift security context adaptation
- disabling OpenShift security context adaptation

## Additional context

Reproduction before the fix:

```sh
helm template test helm \
  --set api.upgrader=true \
  --set api.additionalPlugins[0]=https://example.com/foo.jar \
  --set openshift.enabled=false \
  --set openshift.adaptSecurityContext=disabled \
  --include-crds

### DIFF
--- a/helm/CHANGELOG.md
+++ b/helm/CHANGELOG.md
@@ -3,6 +3,9 @@
 
 This file documents all notable changes to [Gravitee.io API Management 3.x](https://github.com/gravitee-io/helm-charts/tree/master/apim/3.x) Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+### 4.11.3
+- Fix API upgrader job rendering when additional plugins are configured.
+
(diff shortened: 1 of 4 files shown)


wbabyte (approved):

jgiovaresco (commented):
We need to update the Chart.yaml `artifacthub.io/changes` section to have the changelog pushed to Artifactory

PR #16501, opened 2026-04-21
fix: api group permissions in portal.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-13728

## Description

Private APIs in the portal catalog now take into account **group-based access** via the API **`groups`** field: a user can see an API when their groups **intersect** the API’s groups (alongside direct `USER → API` membership and unchanged subscription behavior).

## Implementation

**`ApiPortalMembershipDomainService`**: loads group access through `ApiQueryService.search` using **`ApiSearchCriteria`** with both **`ids`** (candidates) and **`groups`** (the user’s group ids), so storage applies the intersection in the query.

https://github.com/user-attachments/assets/e6c242ec-1a8c-449f-b04a-233347d642d4

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/membership/domain_service/ApiPortalMembershipDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/membership/domain_service/ApiPortalMembershipDomainService.java
@@ -18,11 +18,16 @@ package io.gravitee.apim.core.membership.domain_service;
 import static java.util.stream.Collectors.toSet;
 
 import io.gravitee.apim.core.DomainService;
+import io.gravitee.apim.core.api.model.Api;
+import io.gravitee.apim.core.api.model.ApiFieldFilter;
+import io.gravitee.apim.core.api.model.ApiSearchCriteria;
+import io.gravitee.apim.core.api.query_service.ApiQueryService;
 import io.gravitee.apim.core.membership.model.Membership;
 import io.gravitee.apim.core.membership.query_service.MembershipQueryService;
 import io.gravitee.apim.core.subscription.model.SubscriptionEntity;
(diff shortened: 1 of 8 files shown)


gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_11_0/06_create_custom_dashboards_table.yml
```diff
@@ -7,7 +7,7 @@ databaseChangeLog:
             tableName: ${gravitee_prefix}custom_dashboards
```
graviteeio, line 7: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

kovaren: No changes in the liquibase script.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/membership/domain_service/ApiPortalMembershipDomainService.java
```diff
@@ -50,12 +57,19 @@ public Set<String> filterApiIdsByUserMembership(String userId, Set<String> candi
...
-                .map(Membership::getReferenceId)
+            : apiQueryService
+                .search(
+                    ApiSearchCriteria.builder().ids(List.copyOf(candidateApiIds)).build(),
+                    null,
+                    ApiFieldFilter.builder().pictureExcluded(true).definitionExcluded(true).build()
+                )
+                .filter(api -> api.getGroups() != null && !Collections.disjoint(api.getGroups(), userGroupIds))
+                .map(Api::getId)
                 .collect(toSet());
```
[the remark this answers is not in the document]
kovaren: Fixed

phiz71 (approved):

PR #16521, opened 2026-04-22
feat(portal): APIM-13726 add application member user lookup via use case

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/APIM-13726

## Description

Implements backend support for application member user lookup and aligns the new portal resource flow with Onion Architecture.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/UsersResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/UsersResource.java
@@ -16,7 +16,10 @@
 package io.gravitee.rest.api.portal.rest.resource;
 
 import static io.gravitee.rest.api.model.permissions.RolePermissionAction.READ;
+import static java.util.function.Predicate.not;
 
+import io.gravitee.apim.core.user.model.UserSearchQuery;
+import io.gravitee.apim.core.user.use_case.SearchUsersUseCase;
 import io.gravitee.common.http.MediaType;
 import io.gravitee.rest.api.model.InlinePictureEntity;
 import io.gravitee.rest.api.model.PictureEntity;
@@ -24,15 +27,16 @@ import io.gravitee.rest.api.model.UrlPictureEntity;
 import io.gravitee.rest.api.model.UserEntity;
 import io.gravitee.rest.api.model.permissions.RolePermission;
 import io.gravitee.rest.api.portal.rest.mapper.UserMapper;
+import io.gravitee.rest.api.portal.rest.mapper.UsersSearchQueryMapper;
 import io.gravitee.rest.api.portal.rest.model.*;
 import io.gravitee.rest.api.portal.rest.resource.param.PaginationParam;
 import io.gravitee.rest.api.portal.rest.utils.PortalApiLinkHelper;
 import io.gravitee.rest.api.rest.annotation.Permission;
 import io.gravitee.rest.api.rest.annotation.Permissions;
-import io.gravitee.rest.api.service.IdentityService;
 import io.gravitee.rest.api.service.UserService;
 import io.gravitee.rest.api.service.common.GraviteeContext;
 import io.gravitee.rest.api.service.exceptions.AbstractManagementException;
+import io.gravitee.rest.api.service.exceptions.ForbiddenAccessException;
 import io.gravitee.rest.api.service.exceptions.PasswordAlreadyResetException;
 import jakarta.inject.Inject;
 import jakarta.validation.Valid;
@@ -42,9 +46,10 @@ import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.Request;
 import jakarta.ws.rs.core.Response;
 import java.net.URI;
-import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
-import java.util.stream.Collectors;
+import java.util.Map;
+import java.util.Optional;
 
 /**
  * @author Florent CHAMFROY (florent.chamfroy at graviteesource.com)
@@ -59,7 +64,9 @@ public class UsersResource extends AbstractResource {
     private UserService userService;
 
     @Inject
-    private IdentityService identityService;
+    private SearchUsersUseCase searchUsersUseCase;
+
+    private final UsersSearchQueryMapper usersSearchQueryMapper = UsersSearchQueryMapper.INSTANCE;
 
     /**
      * Register a new user. Generate a token and send it in an email to allow a user
@@ -136,30 +143,59 @@ public class UsersResource extends AbstractResource {
         return Response.serverError().build();
     }
 
+    /**
+     * Keeps reading pagination parameters and letting createListResponse expose them for backward compatibility with the
+     * legacy endpoint contract. IdentityManager-backed user search does not apply pagination at the source, so passing
+     * pageable input to the use case would incorrectly suggest that backend pagination is supported.
+     */
     @POST
     @Path("_search")
+    @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     @Permissions({ @Permission(value = RolePermission.ORGANIZATION_USERS, acls = READ) })
-    public Response getUsers(@QueryParam("q") String query, @BeanParam PaginationParam paginationParam) {
-        String q = query;
-        if (q == null) {
-            q = "*";
-        }
-        List<User> users = identityService
-            .search(q)
-            .stream()
-            .map(searchableUser ->
-                userMapper
-                    .convert(searchableUser)
-                    .links(
-                        userMapper.computeUserLinks(PortalApiLinkHelper.usersURL(uriInfo.getBaseUriBuilder(), searchableUser.getId()), null)
-                    )
-            )
-            .sorted((o1, o2) -> CASE_INSENSITIVE_ORDER.compare(o1.getLastName(), o2.getLastName()))
-            .collect(Collectors.toList());
+    public Response getUsers(@QueryParam("q") String query, @BeanParam PaginationParam paginationParam, UsersSearchInput input) {
+        var executionContext = GraviteeContext.getExecutionContext();
+        var searchQuery = resolveSearchQuery(query, input);
+        var applicationMembership = resolveApplicationMembership(input);
+        ensureCanReadApplicationMembership(executionContext, applicationMembership);
+        var output = searchUsersUseCase.execute(new SearchUsersUseCase.Input(executionContext, searchQuery, applicationMembership));
 
-        // No pagination, because userService did it already
-        return createListResponse(GraviteeContext.getExecutionContext(), users, paginationParam, false);
+        List<User> users = output
+            .data()
+            .stream()
+            .map(searchUser ->
+                userMapper
+                    .convert(searchUser)
+                    .links(userMapper.computeUserLinks(PortalApiLinkHelper.usersURL(uriInfo.getBaseUriBuilder(), searchUser.id()), null))
+            )
+            .toList();
+
+        var metadata = createUsersSearchMetadata(output);
+        return createListResponse(executionContext, users, paginationParam, metadata, false);
+    }
+
+    private UserSearchQuery resolveSearchQuery(String query, UsersSearchInput input) {
+        if (query != null && input != null) {
+            throw new BadRequestException("Query parameter 'q' cannot be used together with request body.");
+        }
+
+        if (query != null) {
+            return new UserSearchQuery(query);
+        }
+
+        if (input != null) {
+            return usersSearchQueryMapper.toSearchQuery(input);
+        }
+
+        return new UserSearchQuery(null);
+    }
+
+    private Optional<String> resolveApplicationMembership(UsersSearchInput input) {
+        if (input == null || input.getIncludes() == null) {
+            return Optional.empty();
+        }
+
+        return Optional.ofNullable(input.getIncludes().getApplicationMembership()).filter(not(String::isBlank));
     }
 
     @GET
@@ -178,43 +214,24 @@ public class UsersResource extends AbstractResource {
         return createPictureResponse(request, (InlinePictureEntity) picture);
     }
 
-    private static final Comparator<String> CASE_INSENSITIVE_ORDER = new CaseInsensitiveComparator();
+    private Map<String, Map<String, Object>> createUsersSearchMetadata(SearchUsersUseCase.Output output) {
+        if (output.applicationMembership() == null) {
+            return null;
+        }
 
-    private static class CaseInsensitiveComparator implements Comparator<String>, java.io.Serializable {
+        Map<String, Map<String, Object>> metadata = new HashMap<>();
+        metadata.put("applicationMembership", new HashMap<>(output.applicationMembership()));
+        return metadata;
+    }
 
-        // use serialVersionUID from JDK 1.2.2 for interoperability
-        private static final long serialVersionUID = 8575799808933029326L;
-
-        @Override
-        public int compare(String s1, String s2) {
-            if (s1 == null) return 1;
-            if (s2 == null) return -1;
-
-            int n1 = s1.length();
-            int n2 = s2.length();
-            int min = Math.min(n1, n2);
-            for (int i = 0; i < min; i++) {
-                char c1 = s1.charAt(i);
-                char c2 = s2.charAt(i);
-                if (c1 != c2) {
-                    c1 = Character.toUpperCase(c1);
-                    c2 = Character.toUpperCase(c2);
-                    if (c1 != c2) {
-                        c1 = Character.toLowerCase(c1);
-                        c2 = Character.toLowerCase(c2);
-                        if (c1 != c2) {
-                            // No overflow because of numeric promotion
-                            return c1 - c2;
-                        }
-                    }
-                }
+    private void ensureCanReadApplicationMembership(
+        io.gravitee.rest.api.service.common.ExecutionContext executionContext,
+        Optional<String> applicationMembership
+    ) {
+        applicationMembership.ifPresent(applicationId -> {
+            if (!hasPermission(executionContext, RolePermission.APPLICATION_MEMBER, applicationId, READ)) {
+                throw new ForbiddenAccessException();
             }
-            return n1 - n2;
-        }
-
-        /** Replaces the de-serialized object. */
-        private Object readResolve() {
-            return CASE_INSENSITIVE_ORDER;
-        }
+        });
     }
 }
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/User.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/User.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.user.model;
+
+import java.util.List;
+import java.util.Map;
+import lombok.Builder;
+
+@Builder
+public record User(
+    String id,
+    String reference,
+    String firstName,
+    String lastName,
+    String displayName,
+    String email,
+    Boolean editableProfile,
+    Map<String, List<String>> permissions
+) {}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/query_service/UserQueryService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/query_service/UserQueryService.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.user.query_service;
+
+import io.gravitee.apim.core.user.model.User;
+import io.gravitee.apim.core.user.model.UserSearchQuery;
+import java.util.List;
+
+public interface UserQueryService {
+    List<User> search(UserSearchQuery searchQuery);
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.user.use_case;
+
+import static java.util.Objects.requireNonNull;
+import static java.util.function.Predicate.not;
+
+import io.gravitee.apim.core.UseCase;
+import io.gravitee.apim.core.application.crud_service.ApplicationCrudService;
+import io.gravitee.apim.core.membership.model.Membership;
+import io.gravitee.apim.core.membership.query_service.MembershipQueryService;
+import io.gravitee.apim.core.user.model.User;
+import io.gravitee.apim.core.user.model.UserSearchQuery;
+import io.gravitee.apim.core.user.query_service.UserQueryService;
+import io.gravitee.rest.api.service.common.ExecutionContext;
+import jakarta.validation.constraints.NotNull;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.RequiredArgsConstructor;
+
+@UseCase
+@RequiredArgsConstructor
+public class SearchUsersUseCase {
+
+    private final UserQueryService userQueryService;
+    private final MembershipQueryService membershipQueryService;
+    private final ApplicationCrudService applicationCrudService;
+
+    public Output execute(Input input) {
+        input
+            .applicationMembership()
+            .ifPresent(applicationId -> applicationCrudService.findById(applicationId, input.executionContext().getEnvironmentId()));
+
+        var searchedUsers = userQueryService.search(input.searchQuery());
+        var sortedUsers = sortUsers(searchedUsers);
+
+        return new Output(
+            sortedUsers,
+            sortedUsers.size(),
+            input
+                .applicationMembership()
+                .map(applicationId ->
+                    sortedUsers.isEmpty() ? Map.<String, Boolean>of() : buildApplicationMembership(applicationId, sortedUsers)
+                )
+                .orElse(null)
+        );
+    }
+
+    private List<User> sortUsers(List<User> users) {
+        return users.stream().sorted(Comparator.comparing(User::lastName, Comparator.nullsLast(String.CASE_INSENSITIVE_ORDER))).toList();
+    }
+
+    private Map<String, Boolean> buildApplicationMembership(String applicationId, List<User> users) {
+        var userIds = users.stream().map(User::id).toList();
+        Set<String> alreadyAddedUserIds = membershipQueryService
+            .findByMemberIdsAndMemberTypeAndReferenceType(userIds, Membership.Type.USER, Membership.ReferenceType.APPLICATION)
+            .stream()
+            .filter(membership -> applicationId.equals(membership.getReferenceId()))
+            .map(Membership::getMemberId)
+            .collect(Collectors.toSet());
+
+        return users.stream().collect(Collectors.toMap(User::id, user -> alreadyAddedUserIds.contains(user.id())));
+    }
+
+    public record Input(
+        @NotNull ExecutionContext executionContext,
+        @NotNull UserSearchQuery searchQuery,
+        @NotNull Optional<String> applicationMembership
+    ) {
+        public Input {
+            applicationMembership = requireNonNull(applicationMembership, "applicationMembership must not be null").filter(
+                not(String::isBlank)
+            );
+        }
+    }
+
+    public record Output(List<User> data, long totalCount, Map<String, Boolean> applicationMembership) {}
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/SearchUserAdapter.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/SearchUserAdapter.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.infra.adapter;
+
+import io.gravitee.apim.core.user.model.User;
+import io.gravitee.rest.api.idp.api.identity.SearchableUser;
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface SearchUserAdapter {
+    SearchUserAdapter INSTANCE = Mappers.getMapper(SearchUserAdapter.class);
+
+    @Mapping(target = "firstName", source = "firstname")
+    @Mapping(target = "lastName", source = "lastname")
+    @Mapping(target = "editableProfile", ignore = true)
+    @Mapping(target = "permissions", ignore = true)
+    User toUser(SearchableUser user);
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCaseTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCaseTest.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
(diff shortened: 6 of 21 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/application/use_case/SearchApplicationMemberUsersUseCase.java
```diff
@@ -0,0 +1,103 @@
...
+        if (userIds.isEmpty()) {
+            return Set.of();
+        }
+
+        return membershipQueryService
+            .findByMemberIdsAndMemberTypeAndReferenceType(userIds, Membership.Type.USER, Membership.ReferenceType.APPLICATION)
+            .stream()
+            .filter(membership -> applicationId.equals(membership.getReferenceId()))
+            .map(Membership::getMemberId)
+            .collect(Collectors.toSet());
```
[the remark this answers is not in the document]
jaroslawlakomy-grav: Good point in principle but... it required introducing a new query contract across MembershipQueryService and MembershipRepository. With the current available contracts, the existing lookup by paged userIds is the narrowest option

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/UsersResource.java
```diff
@@ -138,28 +142,32 @@ public Response changeUserPassword(
...
     @Produces(MediaType.APPLICATION_JSON)
     @Permissions({ @Permission(value = RolePermission.ORGANIZATION_USERS, acls = READ) })
-    public Response getUsers(@QueryParam("q") String query, @BeanParam PaginationParam paginationParam) {
-        String q = query;
-        if (q == null) {
-            q = "*";
-        }
-        List<User> users = identityService
-            .search(q)
+    public Response getUsers(@QueryParam("q") String query, @BeanParam PaginationParam paginationParam, UsersSearchInput input) {
```
aqibmohammadkhan, line 156: this endpoint now exposes `application-membership` bits without the application-level permission check. In `UsersResource`, the endpoint is still guarded only by `RolePermission.ORGANIZATION_USERS[READ]`, but when `includes.applicationMembership` is present it calls `SearchUsersUseCase`, which returns a per-user membership map for the requested application. Existing reads of application members in `ApplicationMembersResource` require `RolePermission.APPLICATION_MEMBER[READ]`, so this change creates a permission bypass: any caller who can search org users can now infer membership for an arbitrary application ID. 

wdyt about gating the enrichment path behind `APPLICATION_MEMBER[READ]` or move it under the application-scoped resource instead

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
```diff
@@ -0,0 +1,143 @@
...
+        var pageable = Optional.ofNullable(input.pageable()).orElseGet(() -> new PageableImpl(1, Integer.MAX_VALUE));
+        var pagedUsers = paginate(filteredUsers, pageable);
+
+        return new Output(pagedUsers, filteredUsers.size(), buildApplicationMembership(requestedApplicationId, pagedUsers));
+    }
+
+    private String resolveQuery(String legacyQuery, SearchUsersCriteria criteria) {
+        var query = criteria.query() != null ? criteria.query() : legacyQuery;
+        return query == null || query.isBlank() ? DEFAULT_QUERY : query;
+    }
```
aqibmohammadkhan: the new body filters can force a full identity-provider scan. `SearchUsersUseCase.resolveQuery()` falls back to `"*"` whenever the body query is null or blank, and both displayName filtering and pagination happen afterward in memory. Since filters.displayName is optional, a request with only displayName (or an empty filters.query) now becomes identityService.search("*") plus local filtering, which is likely to be expensive on large directories. 

wdyt if we either require a nonblank search term for this mode or push the extra filtering into the query service?

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/UsersResource.java
dev-marek: good catch

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/SearchUser.java
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.user.model;
+
+import lombok.Builder;
+
+@Builder
+public record SearchUser(String id, String reference, String firstName, String lastName, String displayName, String email) {}
```
dev-marek: Why not `User`?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/SearchUsersCriteria.java
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.user.model;
+
+import lombok.Builder;
+
+@Builder
+public record SearchUsersCriteria(String query, String displayName, String applicationMembership) {}
```
dev-marek: WDYT `UserSearchCriteria`?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
```diff
@@ -0,0 +1,143 @@
...
+@UseCase
+public class SearchUsersUseCase {
+
+    private static final String DEFAULT_QUERY = "*";
+
+    private final UserQueryService userQueryService;
+    private final MembershipQueryService membershipQueryService;
+    private final ApplicationCrudService applicationCrudService;
+
+    public SearchUsersUseCase(
```
dev-marek: `@RequiredArgsConstructor`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/user/UserQueryServiceLegacyWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+import io.gravitee.rest.api.idp.api.identity.SearchableUser;
+import io.gravitee.rest.api.service.IdentityService;
+import io.gravitee.rest.api.service.common.ExecutionContext;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import org.springframework.stereotype.Service;
+
+@Service
+@AllArgsConstructor
+public class UserQueryServiceLegacyWrapper implements UserQueryService {
```
dev-marek: Instead of creating wrapper, I'd migrate IdentityManager logic here

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/UsersResource.java
jaroslawlakomy-grav: fixed, added permision check

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: original version of this endpoint have fallback to "*", I don't want to introduce any breaking changes, let's discuss it

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/SearchUser.java
jaroslawlakomy-grav: What do you think about io.gravitee.rest.api.idp.api.identity.SearchableUser ? same as IdentityManager provides?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/SearchUsersCriteria.java
jaroslawlakomy-grav: For me it's more like "Searching for users" if you insist on UserSearchCriteria hat do you think about UsersSearchCriteria because we are searching for multiple users not a single one?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: added

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/user/UserQueryServiceLegacyWrapper.java
jaroslawlakomy-grav: renamed to UserQueryServiceImpl and  IdentityManager is used directly

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/SearchUser.java
jaroslawlakomy-grav: renamed to User

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/SearchUsersCriteria.java
jaroslawlakomy-grav: renamed to UserSearchCriteria

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/UsersResource.java
```diff
@@ -138,28 +144,34 @@ public Response changeUserPassword(
...
+        var executionContext = GraviteeContext.getExecutionContext();
+        var criteria = usersSearchCriteriaMapper.toSearchCriteria(input);
+        ensureCanReadApplicationMembership(executionContext, criteria);
+        var output = searchUsersUseCase.execute(
+            new SearchUsersUseCase.Input(
+                executionContext,
+                query,
```
dev-marek: Instead passing a raw string query:

If query != null -> create UserSearchCriteria.ofQuery(query)
If query == null and input != null -> map UsersSearchInput to UserSearchCriteria
If query == null and input == null -> fallback to UserSearchCriteria.default() (query = "*")
If query != null and input != null - > HTTP 400

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
```diff
@@ -0,0 +1,135 @@
...
+        var filteredUsers = filterAndSortUsers(searchedUsers, criteria.displayName());
+        var pageable = Optional.ofNullable(input.pageable()).orElseGet(() -> new PageableImpl(1, Integer.MAX_VALUE));
+        var pagedUsers = paginate(filteredUsers, pageable);
+
+        return new Output(pagedUsers, filteredUsers.size(), buildApplicationMembership(requestedApplicationId, pagedUsers));
+    }
+
+    private String resolveQuery(String legacyQuery, UserSearchCriteria criteria) {
+        var query = criteria.query() != null ? criteria.query() : legacyQuery;
+        return query == null || query.isBlank() ? DEFAULT_QUERY : query;
```
dev-marek: based on criteria, query string would be query string or displayName criteria

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
```diff
@@ -0,0 +1,135 @@
...
+        var criteria = Optional.ofNullable(input.criteria()).orElseGet(() -> UserSearchCriteria.builder().build());
+        var effectiveQuery = resolveQuery(input.legacyQuery(), criteria);
+        var requestedApplicationId = requestedApplicationId(criteria);
+
+        if (requestedApplicationId != null) {
+            applicationCrudService.findById(requestedApplicationId, input.executionContext().getEnvironmentId());
+        }
+
+        var searchedUsers = userQueryService.search(input.executionContext(), effectiveQuery);
+        var filteredUsers = filterAndSortUsers(searchedUsers, criteria.displayName());
```
dev-marek: searched users should already be filtered and sorted

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: I checked the IDP search path. `IdentityManager.search(query)` does not provide a uniform search behavior or a stable ordering across providers.

- `CompositeIdentityManager` merges results into a `HashSet`, so the final order is not guaranteed.
- Repository-backed lookup does search by email, but only by the local part before `@`, and it also searches `displayName`, `lastname+firstname`, `reference` and custom fields.
- LDAP default lookup does not search by email; by default it searches `cn` / `uid` unless `lookup.user.filter` is customized.
- In-memory lookup searches only by username.

So the extra `displayName` filtering in the use case is intentional: it enforces the explicit `displayName` criterion consistently regardless of the active IDP.

I checked the full search path and sorting actually happens in multiple places.

For repository-backed users, `UserService.search(...)` already asks Lucene to sort by `lastname_firstname` ascending, and then restores that order after loading `UserEntity` by IDs.

However, this ordering is not preserved end-to-end once we go through `IdentityManager.search(query)`: `CompositeIdentityManager` merges provider results into a `HashSet`, so the final order is not guaranteed anymore across providers.

That means the sort in `SearchUsersUseCase` is currently the only deterministic ordering applied to the final API result. So the question is not whether we need post-sort at all, but whether sorting only by `lastName` is the right comparator. If we want to stay closer to the repository search behavior, a stronger comparator like `lastName`, then `firstName`, then `displayName`, then `id` would be safer.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: resolveQuery is now based on UserSearchCriteria

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/UsersResource.java
jaroslawlakomy-grav: implemented in resolveSearchCriteria

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: displayName removed, only sorting left exactly as previous implementation was doing that (CompositeIdentityManager merges results into a HashSet, so the final order is not guaranteed)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/UserSearchCriteria.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.user.model;
+
+import lombok.Builder;
+
+@Builder
```
dev-marek: we shouldn't expose public builder when we use factory methods at the same time

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/UserSearchCriteria.java
```diff
@@ -0,0 +1,31 @@
...
+public record UserSearchCriteria(String query, String applicationMembership) {
```
dev-marek: 1. Because applicationMembership is an additional metadata, it does not make sense to include it in UserSearchCriteria
2. Because query is the only thing we'd like to wrap, I'd name it UserSearchQuery
3. Constructor setting query would do the job: `query = ofNullable(query).filter(not(String::isBlank)).orElse(DEFAULT_QUERY)`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
```diff
@@ -0,0 +1,119 @@
...
+            .collect(Collectors.toSet());
+
+        return users
+            .stream()
+            .collect(
+                Collectors.toMap(User::id, user -> alreadyAddedUserIds.contains(user.id()), (left, right) -> left, LinkedHashMap::new)
+            );
+    }
+
+    public record Input(ExecutionContext executionContext, UserSearchCriteria criteria, Pageable pageable) {}
```
dev-marek: please make executionContext, criteria and pageable `@NotNull` and add `Optional<String> applicationMambership`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
```diff
@@ -0,0 +1,119 @@
...
+        var effectiveQuery = resolveQuery(criteria);
+        var requestedApplicationId = requestedApplicationId(criteria);
+
+        if (requestedApplicationId != null) {
+            applicationCrudService.findById(requestedApplicationId, input.executionContext().getEnvironmentId());
+        }
+
+        var searchedUsers = userQueryService.search(input.executionContext(), effectiveQuery);
+        var sortedUsers = sortUsers(searchedUsers);
+        var pageable = Optional.ofNullable(input.pageable()).orElseGet(() -> new PageableImpl(1, Integer.MAX_VALUE));
```
dev-marek: the pagination is a lie -> identity lookups return only first 20 results

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
```diff
@@ -0,0 +1,119 @@
...
+            .findByMemberIdsAndMemberTypeAndReferenceType(userIds, Membership.Type.USER, Membership.ReferenceType.APPLICATION)
+            .stream()
+            .filter(membership -> applicationId.equals(membership.getReferenceId()))
+            .map(Membership::getMemberId)
+            .collect(Collectors.toSet());
+
+        return users
+            .stream()
+            .collect(
+                Collectors.toMap(User::id, user -> alreadyAddedUserIds.contains(user.id()), (left, right) -> left, LinkedHashMap::new)
```
dev-marek: Why not just two-argument `toMap`?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/User.java
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.user.model;
+
+import lombok.Builder;
+
+@Builder
+public record User(String id, String reference, String firstName, String lastName, String displayName, String email) {}
```
dev-marek: Endpoint output also reads `editableProfile` and `permissions`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCaseTest.java
```diff
@@ -0,0 +1,161 @@
...
+                )
+            )
+        );
+
+        assertThat(throwable).isInstanceOf(ApplicationNotFoundException.class).hasMessageContaining(APPLICATION_ID);
+        verify(userQueryService, never()).search(any(), any());
+    }
+
+    private User searchableUser(String id, String reference, String firstname, String lastname, String displayName, String email) {
+        return User.builder()
```
dev-marek: would be better create a fixture

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/query_service/UserQueryService.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.user.query_service;
+
+import io.gravitee.apim.core.user.model.User;
+import io.gravitee.rest.api.service.common.ExecutionContext;
+import java.util.List;
+
+public interface UserQueryService {
+    List<User> search(ExecutionContext executionContext, String query);
```
dev-marek: if the only parameter we want from `executionContext` is `environmentId` and `organizationId`, we should pass those values directly

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: add javadoc with IdP info

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/query_service/UserQueryService.java
jaroslawlakomy-grav: ExecutionContext removed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/UserSearchCriteria.java
jaroslawlakomy-grav: removed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/UserSearchCriteria.java
jaroslawlakomy-grav: renamed to UserSearchQuery  only query left

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: simplified

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCaseTest.java
jaroslawlakomy-grav: done

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: refactored

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/User.java
jaroslawlakomy-grav: Good point. The response model contains editableProfile and permissions because it reuses the generic User schema. However, the legacy /users/_search implementation was based on SearchableUser, which does not expose source or roles/permissions, and the previous mapper did not populate those fields either. For this endpoint, editableProfile remained the model default and permissions were not set. The new use case preserves that behavior.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: removed pagination from usecase, added javadoc to resource

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/User.java
jaroslawlakomy-grav: fields  `editableProfile` and `permissions`  added they are not used in search but domain model is ready for keeping them and mapping to response

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
```diff
@@ -0,0 +1,98 @@
...
+        return users.stream().collect(Collectors.toMap(User::id, user -> alreadyAddedUserIds.contains(user.id())));
+    }
+
+    public record Input(
+        @NotNull ExecutionContext executionContext,
+        @NotNull UserSearchQuery searchQuery,
+        @NotNull Optional<String> applicationMembership
+    ) {
+        public Input {
+            requireNonNull(executionContext, "executionContext must not be null");
```
dev-marek: `@NotNull` / `@NotBlank` annotations would be enough. Intention is to make static code analysis aware that we expect those values to be not null/blank.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
```diff
@@ -0,0 +1,98 @@
...
+public class SearchUsersUseCase {
+
+    private final UserQueryService userQueryService;
+    private final MembershipQueryService membershipQueryService;
+    private final ApplicationCrudService applicationCrudService;
+
+    public Output execute(Input input) {
+        var requestedApplicationId = input.applicationMembership().orElse(null);
+
+        if (requestedApplicationId != null) {
```
dev-marek: Please use `ifPresent` instead of nullchecking

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
```diff
@@ -0,0 +1,98 @@
...
+            applicationCrudService.findById(requestedApplicationId, input.executionContext().getEnvironmentId());
+        }
+
+        var searchedUsers = userQueryService.search(input.searchQuery().query());
+        var sortedUsers = sortUsers(searchedUsers);
+
+        return new Output(sortedUsers, sortedUsers.size(), buildApplicationMembership(requestedApplicationId, sortedUsers));
```
dev-marek: 3rd parameter can be improved:
`input.applicationMembership().filter(not(sortedUsers::isEmpty)).map(applicationId -> buildApplicationMembership(applicationId, sortedUsers)).orElseGet(Collections::emptyMap)`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
```diff
@@ -0,0 +1,98 @@
...
+    private List<User> sortUsers(List<User> users) {
+        return users.stream().sorted(Comparator.comparing(User::lastName, Comparator.nullsLast(String.CASE_INSENSITIVE_ORDER))).toList();
+    }
+
+    private Map<String, Boolean> buildApplicationMembership(String applicationId, List<User> users) {
+        if (applicationId == null) {
+            return null;
+        }
+
+        if (users.isEmpty()) {
```
dev-marek: this method should really be called only when we know we have application id and list of users

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/SearchUserAdapter.java
```diff
@@ -0,0 +1,33 @@
...
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface SearchUserAdapter {
+    SearchUserAdapter INSTANCE = Mappers.getMapper(SearchUserAdapter.class);
+
+    @Mapping(target = "firstName", source = "firstname")
+    @Mapping(target = "lastName", source = "lastname")
+    @Mapping(target = "editableProfile", ignore = true)
```
dev-marek, line 30: this logic should be used to determine this field value: `userItem.setEditableProfile(IDP_SOURCE_GRAVITEE.equals(user.getSource()) || IDP_SOURCE_MEMORY.equalsIgnoreCase(user.getSource()));`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/SearchUserAdapter.java
```diff
@@ -0,0 +1,33 @@
...
+    @Mapping(target = "permissions", ignore = true)
```
dev-marek, line 31: this part of logic is missing:

```
        if (user.getRoles() != null) {
            Map<String, List<String>> userPermissions = user
                .getRoles()
                .stream()
                .filter(role -> RoleScope.ENVIRONMENT.equals(role.getScope()) || RoleScope.ORGANIZATION.equals(role.getScope()))
                .map(UserRoleEntity::getPermissions)
                .map(rolePermissions ->
                    rolePermissions
                        .entrySet()
                        .stream()
                        .collect(
                            Collectors.toMap(Map.Entry::getKey, entry ->
                                new String(entry.getValue())
                                    .chars()
                                    .mapToObj(c -> (char) c)
                                    .map(String::valueOf)
                                    .collect(Collectors.toList())
                            )
                        )
                )
                .reduce(new HashMap<>(), (acc, rolePermissions) -> {
                    acc.putAll(rolePermissions);
                    return acc;
                });
            userItem.setPermissions(objectMapper.convertValue(userPermissions, UserPermissions.class));
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/User.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.user.model;
+
+import java.util.List;
+import java.util.Map;
+import lombok.Builder;
+
+@Builder
+public record User(
```
dev-marek, line 23: custom fields are missing: `userItem.setCustomFields(user.getCustomFields());`

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/query_service/UserQueryService.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.user.query_service;
+
+import io.gravitee.apim.core.user.model.User;
+import java.util.List;
+
+public interface UserQueryService {
+    List<User> search(String query);
```
dev-marek: WDYT about using `SearchUserQuery` type here?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/SearchUserAdapter.java
dev-marek: invalid

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/SearchUserAdapter.java
dev-marek: invalid

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/model/User.java
dev-marek: invalid

dev-marek (approved):
to be merged after cosmetic changes

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: removed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: ifPresent used

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: done

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/use_case/SearchUsersUseCase.java
jaroslawlakomy-grav: removed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/user/query_service/UserQueryService.java
jaroslawlakomy-grav: SearchUserQuery used in method call

PR #16528, opened 2026-04-22
feat(helm): Add Kubernetes Gateway API support to the helm chart

oussamaelkhayali-collab:
## Issue

https://gravitee.atlassian.net/browse/TT-8547

## Description

This change adds HTTPRoute support to the Helm chart for the APIM components that were already exposed through ingress, while preserving the existing ingress behavior. When both ingress and HTTPRoute are enabled, the chart keeps the current ingress-based logic so both can coexist during migration and support zero-downtime cutovers. Once ingress is disabled, the chart automatically switches to the HTTPRoute configuration for generated URLs, paths, and component wiring.


helm/values.yaml
```diff
@@ -1002,6 +1022,39 @@ api:
       #    secretName: apim.example.com
       annotations:
         kubernetes.io/ingress.class: nginx
+  httpRoute:
+    management:
```
jgiovaresco, line 1040: we should add one for gamma a new ingress have been added

helm/values.yaml
jgiovaresco, line 2576: we should add a route for this too

helm/values.yaml
oussamaelkhayali-collab: HTTPRoute added for gamma api

helm/values.yaml
oussamaelkhayali-collab: HTTPRoute added for gamma UI

helm/values.yaml
```diff
@@ -811,6 +811,16 @@ api:
...
+      httpRoute:
+        enabled: false
+        gatewayName: ""
+        gatewayNamespace: ""
+        sectionName: ""
+        hostnames: []
+        pathMatchType: ""
+        path: ""
+        annotations: {}
+        filters: []
```
jgiovaresco: I would put this in commit in the values to be used as documentation but as it is disabled by default it is not used
I would do the same for the other component

jgiovaresco (approved):
LGTM

piequi (approved):

helm/values.yaml
oussamaelkhayali-collab: You mean commenting the code block ?

oussamaelkhayali-collab: Adding adding some example values ?

jgiovaresco: I would comment everything but `enabled: false`

oussamaelkhayali-collab: Done

PR #16543, opened 2026-04-22
feat: add port-based routing fields to NativePlan [APIM-12493]

callaertanthony:
## Summary

- Add `bootstrapPort`, `brokerRangeStart`, and `brokerRangeEnd` fields to `NativePlan.java` for the upcoming Kafka port-based routing mode (`kafka.routingMode=port`).
- Fields are `Integer` (nullable) — `null` in host/SNI routing mode, populated only when publishers configure port-based plans.
- Scoped to `NativePlan` only — `Plan.java`, `HttpPlan.java`, and other plan types are not modified.

**Jira:** [APIM-12493](https://gravitee.atlassian.net/browse/APIM-12493)

This is the first PR in the port-based routing implementation series. The reactor (`gravitee-reactor-native-kafka`) will consume these fields in a follow-up PR.

## Test plan

- [x] `gravitee-apim-definition-model` compiles
- [x] All existing tests pass (no behavioral change — fields are additive and nullable)
- [x] Prettier formatting verified

[APIM-12493]: https://gravitee.atlassian.net/browse/APIM-12493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/v4/nativeapi/NativePlan.java
+++ b/gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/v4/nativeapi/NativePlan.java
@@ -22,7 +22,6 @@ import io.gravitee.definition.model.v4.plan.AbstractPlan;
 import io.swagger.v3.oas.annotations.media.Schema;
 import java.util.List;
 import java.util.Optional;
-import lombok.AllArgsConstructor;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
@@ -35,7 +34,6 @@ import lombok.experimental.SuperBuilder;
  * @author GraviteeSource Team
  */
 @NoArgsConstructor
-@AllArgsConstructor
 @SuperBuilder(toBuilder = true)
 @Getter
 @Setter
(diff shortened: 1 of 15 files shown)


gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/v4/nativeapi/NativePlan.java
```diff
@@ -46,6 +46,15 @@ public class NativePlan extends AbstractPlan {
 
     private List<NativeFlow> flows;
 
+    /**
+     * Port-based routing fields — only used when {@code kafka.routingMode=port}.
+     * Null in host (SNI) routing mode.
+     */
+    private Integer bootstrapPort;
```
ThibaudAV, line 51: I'm not sure if we can add that to a NativePlan, as it's normally Kafka "agnostic" 🤔 
Don't we have a KafkaNativePlan or something like that?

callaertanthony: Something I had actually thought about, yes

phiz71 (approved):

PR #16563, opened 2026-04-23
Normalize allowedInApiProducts null to false in v4 ApiMapper

carlos-andres-osorio:
## Issue

https://gravitee.atlassian.net/browse/APIM-13737

## Why

A v4 HTTP Proxy API can end up with `allowedInApiProducts = null` in its stored definition. Two separate paths were affected:

1. **`ApiMapper.toEntity`** — copied null verbatim into `ApiEntity`, so PUT/GET responses returned null instead of false.
2. **`IndexableApiDocumentTransformer`** — skipped writing `FIELD_ALLOW_IN_API_PRODUCTS` entirely when the value was null. Because the indexer receives its `ApiEntity` from the create/update use case flow (which does not go through `ApiMapper.toEntity`), the first fix alone left null-stored APIs invisible to `POST /apis/_search {"allowedInApiProducts": false}`.

## What changed

- **`ApiMapper.toEntity` (PROXY branch)** — replaced verbatim copy of `apiDefinition.getAllowedInApiProducts()` with `Boolean.TRUE.equals(...)`, collapsing absent-key, explicit-null, and explicit-false into `false`. The redundant `readTree` parse that only existed to check key presence was removed.
- **`IndexableApiDocumentTransformer`** — applied the same `Boolean.TRUE.equals(...)` normalization in `transformV4Api`. The field is now always written for V4 HTTP proxy APIs, so null-stored APIs are indexed as `false` and appear in `allowedInApiProducts=false` search results. Native, V2, and definition-less APIs are unaffected.
- No changes to the write path (tracked separately).

## How to test

1. Store a v4 HTTP Proxy API with `"allowedInApiProducts": null` in its definition.
2. `GET /management/v2/environments/{envId}/apis/{apiId}` — response must include `"allowedInApiProducts": false`.
3. `PUT /management/v2/environments/{envId}/apis/{apiId}` — response must include `"allowedInApiProducts": false`.
4. Re-index the API, then call `POST /apis/_search {"allowedInApiProducts": false}` — the API must appear in results.

Automated coverage: `ApiMapperTest`, `IndexableApiDocumentTransformerTest`, `ApiResource_getApiByIdTest`, `ApiResource_UpdateApiTest`.

## Gotchas / Follow-up

- Pre-existing Lucene documents indexed before this fix are unaffected until a full re-index is triggered. Release notes must flag this for operators who rely on the `allowedInApiProducts` search filter for legacy data.
- The write path (`toApiDefinition(UpdateApiEntity)`) is intentionally untouched — a null `allowedInApiProducts` on update still omits the key. Persisting a concrete Boolean on write is tracked in APIM-13738.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_UpdateApiTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_UpdateApiTest.java
@@ -207,6 +207,44 @@ public class ApiResource_UpdateApiTest extends ApiResourceTest {
     }
 
     @Test
+    public void should_return_false_for_allowedInApiProducts_when_entity_has_false() {
+        ApiEntity apiEntity = ApiFixtures.aModelHttpApiV4().toBuilder().id(API).allowedInApiProducts(null).build();
+        UpdateApiV4 updateApiV4 = ApiFixtures.anUpdateApiV4();
+
+        when(apiSearchServiceV4.findGenericById(GraviteeContext.getExecutionContext(), API, false, false, false)).thenReturn(apiEntity);
+        when(
+            apiServiceV4.update(eq(GraviteeContext.getExecutionContext()), eq(API), any(UpdateApiEntity.class), eq(false), eq(USER_NAME))
+        ).thenReturn(apiEntity.toBuilder().allowedInApiProducts(false).build());
+        when(apiStateServiceV4.isSynchronized(eq(GraviteeContext.getExecutionContext()), any())).thenReturn(true);
+
+        final Response response = rootTarget(API).request().put(Entity.json(updateApiV4));
+        assertEquals(OK_200, response.getStatus());
+
+        final ApiV4 apiV4 = response.readEntity(ApiV4.class);
+        assertNotNull(apiV4);
+        assertEquals(Boolean.FALSE, apiV4.getAllowedInApiProducts());
+    }
+
+    @Test
+    public void should_preserve_true_for_allowedInApiProducts_when_entity_has_true() {
+        ApiEntity apiEntity = ApiFixtures.aModelHttpApiV4().toBuilder().id(API).allowedInApiProducts(true).build();
+        UpdateApiV4 updateApiV4 = ApiFixtures.anUpdateApiV4();
+
+        when(apiSearchServiceV4.findGenericById(GraviteeContext.getExecutionContext(), API, false, false, false)).thenReturn(apiEntity);
+        when(
+            apiServiceV4.update(eq(GraviteeContext.getExecutionContext()), eq(API), any(UpdateApiEntity.class), eq(false), eq(USER_NAME))
+        ).thenReturn(apiEntity);
+        when(apiStateServiceV4.isSynchronized(eq(GraviteeContext.getExecutionContext()), any())).thenReturn(true);
+
+        final Response response = rootTarget(API).request().put(Entity.json(updateApiV4));
+        assertEquals(OK_200, response.getStatus());
+
+        final ApiV4 apiV4 = response.readEntity(ApiV4.class);
+        assertNotNull(apiV4);
+        assertEquals(Boolean.TRUE, apiV4.getAllowedInApiProducts());
+    }
+
+    @Test
     public void should_return_bad_request_when_updating_v4_api_with_v2_defition() {
         final ApiEntity apiEntity = ApiFixtures.aModelHttpApiV4().toBuilder().id(API).build();
         final UpdateApiV2 updateApiV2 = ApiFixtures.anUpdateApiV2();
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_UpdateApiTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_UpdateApiTest.java
@@ -207,7 +207,7 @@ public class ApiResource_UpdateApiTest extends ApiResourceTest {
     }
 
     @Test
-    public void should_return_false_for_allowedInApiProducts_when_entity_has_false() {
+    public void should_return_false_for_allowedInApiProducts_when_entity_has_null() {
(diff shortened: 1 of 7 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_UpdateApiTest.java
```diff
@@ -206,6 +206,44 @@ public void should_update_v2_api() {
         );
     }
 
+    @Test
+    public void should_return_false_for_allowedInApiProducts_when_entity_has_false() {
```
paulatulis: Rename to `should_return_false_for_allowedInApiProducts_when_entity_has_null`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/mapper/ApiMapperTest.java
```diff
@@ -140,6 +140,63 @@ public void shouldMapAllowedInApiProductsFromRepositoryDefinition_true_false() t
...
+        repo.setDefinition(objectMapper.writeValueAsString(def));
+        repo.setType(ApiType.PROXY);
+
+        var entity = apiMapper.toEntity(repo, new PrimaryOwnerEntity());
+
+        assertThat(entity.getAllowedInApiProducts()).isFalse();
+    }
+
+    @Test
+    public void shouldDefaultAllowedInApiProductsToFalseWhenKeyIsAbsentInDefinition() throws Exception {
```
paulatulis, line 163: This is fragile JSON manipulation via string surgery. It works today because the plain new ObjectMapper() in the test setup serializes null fields by default (no @JsonInclude(NON_NULL) on Api.java). But:

If @JsonInclude(NON_NULL) is added to Api.java in the future (a routine tidy-up), all three replaces become no-ops silently. rawJson == defJson, and this test becomes an exact duplicate of shouldDefaultAllowedInApiProductsToFalseWhenStoredValueIsNull. Both tests still pass green — the absent-key case is completely untested with no visible failure.

Fix: Remove the key structurally using Jackson:

```
ObjectNode node = (ObjectNode) objectMapper.readTree(objectMapper.writeValueAsString(def));
node.remove("allowedInApiProducts");
Api repo = new Api();
repo.setId("idAbsent");
repo.setDefinition(node.toString());
repo.setType(ApiType.PROXY);
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/search/lucene/transformer/IndexableApiDocumentTransformerTest.java
```diff
@@ -178,8 +178,8 @@ void should_transform_api_info() {
             // origin
             softly.assertThat(result.getField(FIELD_ORIGIN).stringValue()).isEqualTo("management");
 
-            // allow_in_api_products should not be indexed when missing in definition
-            softly.assertThat(result.getField(FIELD_ALLOW_IN_API_PRODUCTS)).isNull();
+            // null allowedInApiProducts is normalized to false so the API remains findable by allowedInApiProducts=false
```
paulatulis, line 181: This correctly updates the null case assertion. However, the test suite for IndexableApiDocumentTransformer now has no coverage for:

Explicit false stored → indexed as "false" ✗
Explicit true stored → indexed as "true" ✗
The ApiDocumentTransformerTest test additions (see below) cover those values for ApiDocumentTransformer, but that is a different transformer. The indexing logic is now different enough between the two that each deserves its own matrix.

Fix: Add two targeted tests to IndexableApiDocumentTransformerTest: one for allowedInApiProducts = false and one for true, each asserting the corresponding string value is indexed.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_getApiByIdTest.java
```diff
@@ -506,6 +507,36 @@ public void should_get_filtered_api_V2() {
         assertNull(responseApi.getProxy().getVirtualHosts().get(0).getHost());
     }
 
+    @Test
+    public void should_return_false_for_allowedInApiProducts_when_entity_has_false() {
```
paulatulis, line 511: Every existing test in this file uses camelCase (shouldGetApi, shouldNotGetApiBecauseNotFound, should_get_filtered_api_V2). The two new methods use lowercase underscore-separated style.

paulatulis (commented):
A few things but overall looks good!

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/mapper/ApiMapperTest.java
carlos-andres-osorio: Replaced the string-surgery approach with `ObjectNode.remove("allowedInApiProducts")` — the key is now removed structurally so the test remains valid regardless of any future `@JsonInclude` annotation changes on `Api.java`.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_getApiByIdTest.java
carlos-andres-osorio: The reviewer's claim turned out to be the inverse of what the file actually contains: 8 of the 9 existing methods already use underscore-separated style (`should_get_api_V4`, `should_get_filtered_api_V4`, etc.) — `shouldNotGetApiBecauseNotFound` was the sole outlier. Renamed it to `should_not_get_api_because_not_found` to align with the dominant convention in this class.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/search/lucene/transformer/IndexableApiDocumentTransformerTest.java
carlos-andres-osorio: Added `should_index_explicit_false_allowedInApiProducts` asserting the field is indexed as `"false"`. The `true` case is already covered by an existing test at line 351 which exercises `allowedInApiProducts(true)` through the full transform path — no additional test needed there.

paulatulis (approved):
🤖 approved

PR #16569, opened 2026-04-23
fix: adjust caption size and color in members. remove unnecessary custom link style.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-13466

## Description

- Use supporting color for member captions
- Change the avatar styles to stay consistent with existing components
- Adjust caption size according to design
- Remove unnecessary custom link style

Screenshot 2026-04-24 at 11 25 55

### DIFF
--- a/gravitee-apim-portal-webui-next/src/components/paginated-table/paginated-table.component.scss
+++ b/gravitee-apim-portal-webui-next/src/components/paginated-table/paginated-table.component.scss
@@ -54,10 +54,6 @@
 
   &__column-expand {
     text-align: center;
-
-    mat-icon {
-      color: app-theme.$link-icon-color;
-    }
   }
 
   &__pagination {
--- a/gravitee-apim-portal-webui-next/src/components/user-cell/user-cell.component.scss
+++ b/gravitee-apim-portal-webui-next/src/components/user-cell/user-cell.component.scss
@@ -34,9 +34,9 @@
     justify-content: center;
     width: 32px;
     height: 32px;
-    border-radius: 50%;
-    color: app-theme.$default-text-contrast-color;
-    background-color: app-theme.$user-avatar-text-color;
+    border-radius: app-theme.$user-avatar-container-shape;
+    color: app-theme.$user-avatar-text-color;
+    border: 1px solid app-theme.$user-avatar-border-color;
     flex-shrink: 0;
   }
 
@@ -45,12 +45,11 @@
     flex-direction: column;
   }
 
-  &__secondary {
-    color: app-theme.$secondary-main-color;
+  &__caption {
+    color: app-theme.$paragraph-color;
   }
 
   &__you {
-    margin-left: 4px;
-    color: app-theme.$secondary-main-color;
+    color: app-theme.$paragraph-color;
   }
 }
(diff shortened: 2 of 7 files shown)


gravitee-apim-portal-webui-next/src/components/user-cell/user-cell.component.scss
```diff
@@ -36,7 +36,7 @@
     height: 32px;
     border-radius: 50%;
     color: app-theme.$default-text-contrast-color;
-    background-color: app-theme.$user-avatar-text-color;
+    background-color: app-theme.$primary-main-color;
```
aqibmohammadkhan: this change bypasses the `--gio-app-user-avatar-text-color` CSS custom property that consumers may have set for theming.

Also note that `user-avatar.component.scss` uses `$user-avatar-text-color` for the larger user avatar button's text color — that component remains on the theming token.

just want to be sure, this is intented here.

gravitee-apim-portal-webui-next/src/components/paginated-table/paginated-table.component.scss
```diff
@@ -56,7 +56,7 @@
     text-align: center;
 
     mat-icon {
-      color: app-theme.$link-icon-color;
+      color: app-theme.$paragraph-color;
```
aqibmohammadkhan: `$paragraph-color` is a typography token (75% of the text color), not an icon/outline token. Using it for the table's expand icon is semantically off. The two values may appear close visually in the default theme, but in a dark or custom theme they could diverge noticeably. A better fit would be a dedicated icon-color token, or at least keeping a variable that maps to `--gio-app-outline-color` for icons.

wdyt?

gravitee-apim-portal-webui-next/src/components/user-cell/user-cell.component.scss
kovaren: The use of `$user-avatar-text-color` wasn't appropriate because it's a text color and we wanted a background color here.

But actually I'm realizing that there's no good reason to have a background here, the mockups just deviated from the existing styles of the avatar, we should rather stay consistent with them.

WDYT?

```scss
    border-radius: app-theme.$user-avatar-container-shape;
    color: app-theme.$user-avatar-text-color;
    border: 1px solid app-theme.$user-avatar-border-color;
```

Screenshot 2026-04-24 at 10 03 35

gravitee-apim-portal-webui-next/src/components/paginated-table/paginated-table.component.scss
kovaren: Honestly, after taking a second look, I don't think this particular icon should even have a color that's different from the text, it doesn't have any accessible utility here. I'd go for removing this style altogether, WDYT?

Speaking about `$paragraph-color`, I agree that the name is a bit off, it should rather be called smth like $supporting-text-color because it's not really used for paragraphs but rather for supplementary information. I'd keep it as is for now though just to avoid too many unrelated changes in the PR.

Also, we are not using this variable for all icons but just for a subset of icons that are most often accompanying text, so that doesn't seem too unreasonable, their styles are semantically linked.

Using `--gio-app-outline-color` for icons is definitely not right, it should only be used for borders.

aqibmohammadkhan (approved):

PR #16575, opened 2026-04-23
fix(console): sync group add-members default roles before submit

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-13783

## Description
Rebuild memberships from selected users and current form when non-API default roles change and before closing the dialog, so API_PRODUCT (and other scopes) match the dropdowns. Fix chip removal splice index.

## Additional context

https://github.com/user-attachments/assets/dcccce05-ffda-4d14-824a-b4d7b3327ea4

### DIFF
--- a/gravitee-apim-console-webui/src/management/settings/groups/group/add-members-dialog/add-members-dialog.component.html
+++ b/gravitee-apim-console-webui/src/management/settings/groups/group/add-members-dialog/add-members-dialog.component.html
@@ -20,7 +20,7 @@
   <form [formGroup]="addMemberForm" class="add-member__form">
     <mat-form-field aria-hidden="false" aria-label="Select default API role">
       <mat-label>Default API Role</mat-label>
-      <mat-select formControlName="defaultAPIRole" (selectionChange)="onAPIRoleChange()">
+      <mat-select formControlName="defaultAPIRole" (selectionChange)="onDefaultRoleChange()">
         @for (role of defaultAPIRoles; track role.id) {
           <mat-option [value]="role.name" [disabled]="disabledAPIRoles.has(role.id)">{{ role.name }}</mat-option>
         }
@@ -29,7 +29,7 @@
 
     <mat-form-field aria-hidden="false" aria-label="Select default API product role">
       <mat-label>Default API Product Role</mat-label>
-      <mat-select formControlName="defaultAPIProductRole">
+      <mat-select formControlName="defaultAPIProductRole" (selectionChange)="onDefaultRoleChange()">
         @for (role of defaultAPIProductRoles; track role.id) {
           <mat-option [value]="role.name" [disabled]="role.system">{{ role.name }}</mat-option>
         }
@@ -38,7 +38,7 @@
 
     <mat-form-field aria-hidden="false" aria-label="Select default application role">
       <mat-label>Default Application Role</mat-label>
-      <mat-select formControlName="defaultApplicationRole">
+      <mat-select formControlName="defaultApplicationRole" (selectionChange)="onDefaultRoleChange()">
         @for (role of defaultApplicationRoles; track role.id) {
           <mat-option [value]="role.name" [disabled]="role.system">{{ role.name }}</mat-option>
         }
@@ -47,7 +47,7 @@
 
     <mat-form-field aria-hidden="false" aria-label="Select default integration role">
       <mat-label>Default Integration Role</mat-label>
-      <mat-select formControlName="defaultIntegrationRole">
+      <mat-select formControlName="defaultIntegrationRole" (selectionChange)="onDefaultRoleChange()">
         @for (role of defaultIntegrationRoles; track role.id) {
           <mat-option [value]="role.name" [disabled]="role.system">{{ role.name }}</mat-option>
         }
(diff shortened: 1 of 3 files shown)


gravitee-apim-console-webui/src/management/settings/groups/group/add-members-dialog/add-members-dialog.component.html
```diff
@@ -38,7 +38,7 @@ <h2 mat-dialog-title>Add Members</h2>
 
     <mat-form-field aria-hidden="false" aria-label="Select default application role">
       <mat-label>Default Application Role</mat-label>
-      <mat-select formControlName="defaultApplicationRole">
+      <mat-select formControlName="defaultApplicationRole" (selectionChange)="onDefaultRoleChange()">
```
vikrantgravitee, line 41: Was this & integration also impacted?
Ensure it should not cause any breaking change

gravitee-apim-console-webui/src/management/settings/groups/group/add-members-dialog/add-members-dialog.component.html
```diff
@@ -20,7 +20,7 @@ <h2 mat-dialog-title>Add Members</h2>
   <form [formGroup]="addMemberForm" class="add-member__form">
     <mat-form-field aria-hidden="false" aria-label="Select default API role">
       <mat-label>Default API Role</mat-label>
-      <mat-select formControlName="defaultAPIRole" (selectionChange)="onAPIRoleChange()">
+      <mat-select formControlName="defaultAPIRole" (selectionChange)="onDefaultRoleChange()">
```
vikrantgravitee, line 23: Verify if it was pre exist that if 5 members are added and we change a default role, will it remove all users?

In short - Changing any role clears all selected users, existing or after this?

gravitee-apim-console-webui/src/management/settings/groups/group/add-members-dialog/add-members-dialog.component.ts
```diff
@@ -206,9 +206,14 @@ export class AddMembersDialogComponent implements OnInit {
   }
 
   submit() {
+    this.syncMembershipsFromForm();
     this.matDialogRef.close({ memberships: this.memberships });
   }
 
+  private syncMembershipsFromForm(): void {
```
vikrantgravitee, line 213: Is syncMembershipsFromForm required as we are cleaning roles?

In my view, we should not be adding onDefaultRoleChange for all types of roles but it should be limited to API or API Product

gravitee-apim-console-webui/src/management/settings/groups/group/add-members-dialog/add-members-dialog.component.ts
```diff
@@ -296,7 +301,8 @@ export class AddMembersDialogComponent implements OnInit {
     return this.memberships.filter(membership => membership.roles.find(role => role.scope === 'API').name === RoleName.PRIMARY_OWNER);
   }
 
-  onAPIRoleChange() {
+  /** Clears user chips and memberships when any default role changes; refreshes PRIMARY_OWNER search rules via changeFormState. */
+  onDefaultRoleChange(): void {
     this.selectedUsers = [];
     this.memberships = [];
     this.changeFormState();
```
vikrantgravitee, line 308: Did you verify this?

private changeFormState() {
    const noOfPrimaryOwners = this.filterPrimaryOwners().length;
    this.addMemberForm.controls.searchTerm.setValue('');
    this.disableSearch();

    if (this.addMemberForm.controls.defaultAPIRole.value === RoleName.PRIMARY_OWNER) {
      if (noOfPrimaryOwners > 0) {
        this.addMemberForm.controls.searchTerm.disable();
        this.addMemberForm.controls.searchTerm.removeValidators(Validators.required);
      } else if (noOfPrimaryOwners < 1) {
        this.addMemberForm.controls.searchTerm.enable();
        this.addMemberForm.controls.searchTerm.addValidators(Validators.required);
      }
    }
    this.disableSubmit = this.memberships.length === 0;
  }

  private filterPrimaryOwners() {
    return this.memberships.filter(membership => membership.roles.find(role => role.scope === 'API').name === RoleName.PRIMARY_OWNER);
  }

  /** Clears user chips and memberships when any default role changes; refreshes PRIMARY_OWNER search rules via changeFormState. */
  onDefaultRoleChange(): void {
    this.selectedUsers = [];
    this.memberships = [];
    this.changeFormState();
  }

**It has API level check only.**

_**More queries:**_
1. Is it okay and does it mean primary owners can't be on API PRODUCT level.
2. If we allow API Product PO then how it will impact API and GROUP functionality?

Now again do we need this onDefaultRoleChange for all types of roles?

Also

PRIMARY_OWNER logic is fragility:  changeFormState checks this.addMemberForm.controls.defaultAPIRole.value === RoleName.PRIMARY_OWNER to decide whether to disable search. Since onDefaultRoleChange now clears memberships, filterPrimaryOwners() always returns [], so noOfPrimaryOwners is always 0. The PRIMARY_OWNER search restriction will never activate again after a role change.

gravitee-apim-console-webui/src/management/settings/groups/group/add-members-dialog/add-members-dialog.component.html
ankita-gupta21: yes it was the issue in Application, integration, api-product

PR #16577, opened 2026-04-24
Add PATCH endpoint for v4 HTTP Proxy APIs (merge patch)

carlos-andres-osorio:
## Summary

Adds a `PATCH /environments/{envId}/apis/{apiId}` endpoint for v4 HTTP proxy APIs using **JSON Merge Patch** (RFC 7396, `application/merge-patch+json`).

- New `PatchApiUseCase` with field allow-list enforcement, dry-run support (`?dryRun=true`), and primary-owner resolution
- New `ApiPatchDomainService.applyMergePatch` backed by the `com.github.java-json-tools:json-patch` library
- `UpdateApiDomainService.validateV4` — dry-run validation path that skips persistence
- `ApiService.validate` — corresponding v1 service layer method
- `NotSupportedExceptionMapper` maps unsupported content-types to 415
- `ApiMapper` additions: `map(PrimaryOwnerEntity)`, `allowedInApiProducts`/`failover`/`responseTemplates`/`services` mappings on `mapToHttpV4`
- `ApiAdapter.toUpdateApiEntity` now maps `responseTemplates`

## Test plan

- [x] `PatchApiUseCaseTest` — 30 use-case unit tests (allow-list, merge semantics, dry-run, null handling, error paths)
- [x] `ApiPatchDomainServiceTest` — 3 merge-patch unit tests
- [x] `ApiResource_PatchApiTest` — 46 REST integration tests (200/400/403/404/415, dry-run, headers, null rejections)

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,1 +1,1 @@
-@AGENTS.md
+@../AGENTS.md
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
@@ -1,0 +1,452 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.api.use_case;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.gravitee.apim.core.UseCase;
+import io.gravitee.apim.core.api.crud_service.ApiCrudService;
+import io.gravitee.apim.core.api.domain_service.ApiPatchDomainService;
+import io.gravitee.apim.core.api.domain_service.UpdateApiDomainService;
+import io.gravitee.apim.core.api.exception.ApiInvalidDefinitionVersionException;
+import io.gravitee.apim.core.api.exception.ApiInvalidTypeException;
+import io.gravitee.apim.core.api.exception.ApiPatchNotAllowedException;
+import io.gravitee.apim.core.api.model.Api;
+import io.gravitee.apim.core.audit.model.AuditInfo;
+import io.gravitee.apim.core.exception.ValidationDomainException;
+import io.gravitee.apim.core.membership.domain_service.ApiPrimaryOwnerDomainService;
+import io.gravitee.apim.core.membership.model.PrimaryOwnerEntity;
+import io.gravitee.apim.core.workflow.model.Workflow;
+import io.gravitee.apim.core.workflow.query_service.WorkflowQueryService;
+import io.gravitee.definition.model.DefinitionVersion;
+import io.gravitee.definition.model.ResponseTemplate;
+import io.gravitee.definition.model.v4.ApiType;
+import io.gravitee.definition.model.v4.analytics.Analytics;
+import io.gravitee.definition.model.v4.failover.Failover;
+import io.gravitee.definition.model.v4.flow.execution.FlowExecution;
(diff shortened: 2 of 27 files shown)


.claude/CLAUDE.md
```diff
@@ -1 +1 @@
-@AGENTS.md
\ No newline at end of file
+@../AGENTS.md
```
paulatulis, line 1: Is this intended?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/UpdateApiDomainServiceImpl.java
```diff
@@ -58,4 +58,32 @@ public Api updateV4(Api api, AuditInfo auditInfo) {
...
+    public Api validateV4(Api api, AuditInfo auditInfo) {
+        var executionContext = new ExecutionContext(auditInfo.organizationId(), auditInfo.environmentId());
+        var updateApiEntity = ApiAdapter.INSTANCE.toUpdateApiEntity(api, api.getApiDefinitionHttpV4());
+
+        delegate.validate(executionContext, api.getId(), updateApiEntity, auditInfo.actor().userId());
+
+        return applySanitizedValues(api, updateApiEntity);
+    }
+
+    private Api applySanitizedValues(Api original, io.gravitee.rest.api.model.v4.api.UpdateApiEntity sanitized) {
```
paulatulis, line 72: `applySanitizedValues` is an incomplete mirror.
`validateAndSanitizeUpdateApi()` mutates `UpdateApiEntity` in place for tags, analytics, groups, and lifecycleState — but applySanitizedValues() only reads those four fields back. Fields like properties, labels, categories, responseTemplates, failover, flowExecution, and services are not re-read after sanitization. If the validator strips or transforms any of those, the dry-run preview diverges silently from what would actually be persisted, and the trust in dryRun=true is broken.

Fix: Re-read all patchable fields from the sanitized `UpdateApiEntity`, or refactor the validator to return a sanitized copy.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -0,0 +1,452 @@
...
+import lombok.RequiredArgsConstructor;
+
+/**
+ * @author GraviteeSource Team
+ */
+@UseCase
+@RequiredArgsConstructor
+public class PatchApiUseCase {
+
+    private static final Set<String> ALLOWED_FIELDS = Set.of(
```
paulatulis, line 58: `groups` is in the patchable allowlist with only `API_DEFINITION[UPDATE]` required. The existing PUT flow routes group changes through `UpdateApiGroupsUseCase` which has its own dedicated permission check. A user with only `API_DEFINITION[UPDATE]` can use `PATCH {"groups": ["admin-group"]}` or `PATCH [{"op":"replace","path":"/groups","value":["admin-group"]}] `to change group membership without the groups permission.

Fix: Either remove `groups` from the allowlist and require the groups endpoint, or replicate the group-mutation permission check here.

.claude/CLAUDE.md
carlos-andres-osorio: Yes, AGENTS.md is at the root, so it should be ../AGENTS.md

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/UpdateApiDomainServiceImpl.java
carlos-andres-osorio: Good catch. Verified `validateAndSanitizeUpdateApi`: it currently only mutates the four fields that `applySanitizedValues` re-read (lifecycleState, tags, groups, analytics), so no divergence exists today — the remaining patchable fields are not touched by the validator and are correctly inherited from the already-patched `Api`. That said, the forward-looking maintainability risk is real. Extended `applySanitizedValues` to re-read all patchable fields from the sanitized `UpdateApiEntity` (failover, flowExecution, services, responseTemplates, name, description, version, visibility, labels, categories, allowedInApiProducts, allowMultiJwtOauth2Subscriptions, disableMembershipNotifications) and added four unit tests covering the new behavior.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: Good catch. `PATCH /apis/{apiId}` only checks `API_DEFINITION[UPDATE]`, while group membership changes are gated behind `API_MEMBER[UPDATE]` via `PUT /apis/{apiId}/groups`. Removed `groups` from the PATCH allowlist — a patch body containing `groups` now returns 400. Group membership changes should go through the existing `PUT /apis/{apiId}/groups` endpoint.

paulatulis (approved):
No blockers!

PR #16578, opened 2026-04-24
feat(management-api): add JSON Patch support to v4 HTTP Proxy API PATCH endpoint

carlos-andres-osorio:
## Summary

Extends the `PATCH /environments/{envId}/apis/{apiId}` endpoint with **JSON Patch** (RFC 6902, `application/json-patch+json`) support alongside the existing merge-patch.

Depends on: #16577

- `JsonPatchService` (core domain) + `ApiJsonPatchServiceImpl` (infra) — applies a RFC 6902 patch array using `JsonPatch.fromJson`, wrapping `JsonPatchException`/`IOException` as `ValidationDomainException`
- `ApiPatchDomainService.applyJsonPatch` — validates the patch is a JSON array and delegates to `JsonPatchService`
- `PatchApiUseCase.PatchType.JSON_PATCH` — allow-list enforcement via top-level path segment extraction from each operation's `path` field
- `ApiResource.patchApi` — content-type detection dispatches to the correct patch strategy

## Test plan

- [x] `PatchApiUseCaseTest` — additional JSON Patch use-case tests (replace op, path-segment allow-list, null semantics, malformed body)
- [x] `ApiPatchDomainServiceTest` — JSON Patch unit tests
- [x] `ApiJsonPatchServiceImplTest` — infra-level test that runtime exceptions from the underlying library propagate without being wrapped
- [x] `ApiResource_PatchApiTest` — additional REST integration tests (200/400 for JSON Patch operations, dry-run with JSON Patch, null rejections)

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
@@ -15,6 +15,8 @@
  */
 package io.gravitee.apim.core.api.use_case;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -36,8 +38,12 @@ import io.gravitee.definition.model.DefinitionVersion;
 import io.gravitee.definition.model.ResponseTemplate;
 import io.gravitee.definition.model.v4.ApiType;
 import io.gravitee.definition.model.v4.analytics.Analytics;
+import io.gravitee.definition.model.v4.analytics.logging.Logging;
+import io.gravitee.definition.model.v4.analytics.sampling.Sampling;
+import io.gravitee.definition.model.v4.analytics.tracing.Tracing;
 import io.gravitee.definition.model.v4.failover.Failover;
 import io.gravitee.definition.model.v4.flow.execution.FlowExecution;
+import io.gravitee.definition.model.v4.flow.execution.FlowMode;
 import io.gravitee.definition.model.v4.property.Property;
 import io.gravitee.definition.model.v4.service.ApiServices;
 import java.util.ArrayList;
@@ -45,6 +51,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import lombok.Builder;
 import lombok.RequiredArgsConstructor;
 
@@ -55,10 +62,18 @@ import lombok.RequiredArgsConstructor;
 @RequiredArgsConstructor
 public class PatchApiUseCase {
 
+    private static final String FIELD_NAME = "name";
     private static final String FIELD_DESCRIPTION = "description";
     private static final String FIELD_API_VERSION = "apiVersion";
     private static final String FIELD_VISIBILITY = "visibility";
+    private static final String FIELD_LABELS = "labels";
+    private static final String FIELD_TAGS = "tags";
     private static final String FIELD_LIFECYCLE_STATE = "lifecycleState";
+    private static final String FIELD_CATEGORIES = "categories";
+    private static final String FIELD_ANALYTICS = "analytics";
+    private static final String FIELD_FAILOVER = "failover";
+    private static final String FIELD_FLOW_EXECUTION = "flowExecution";
+    private static final String FIELD_SERVICES = "services";
     private static final String FIELD_ALLOWED_IN_API_PRODUCTS = "allowedInApiProducts";
     private static final String FIELD_ALLOW_MULTI_JWT_OAUTH2_SUBSCRIPTIONS = "allowMultiJwtOauth2Subscriptions";
     private static final String FIELD_DISABLE_MEMBERSHIP_NOTIFICATIONS = "disableMembershipNotifications";
@@ -66,18 +81,18 @@ public class PatchApiUseCase {
     private static final String FIELD_RESPONSE_TEMPLATES = "responseTemplates";
 
     private static final Set<String> ALLOWED_FIELDS = Set.of(
-        "name",
+        FIELD_NAME,
         FIELD_DESCRIPTION,
         FIELD_API_VERSION,
         FIELD_VISIBILITY,
-        "labels",
-        "tags",
+        FIELD_LABELS,
+        FIELD_TAGS,
         FIELD_LIFECYCLE_STATE,
-        "categories",
-        "analytics",
-        "failover",
-        "flowExecution",
-        "services",
+        FIELD_CATEGORIES,
+        FIELD_ANALYTICS,
+        FIELD_FAILOVER,
+        FIELD_FLOW_EXECUTION,
+        FIELD_SERVICES,
         FIELD_ALLOWED_IN_API_PRODUCTS,
         FIELD_ALLOW_MULTI_JWT_OAUTH2_SUBSCRIPTIONS,
         FIELD_DISABLE_MEMBERSHIP_NOTIFICATIONS,
@@ -85,8 +100,13 @@ public class PatchApiUseCase {
         FIELD_RESPONSE_TEMPLATES
     );
 
+    private static final int MAX_PATCH_OPS = 200;
+
     private static final Set<String> BLOCKED_WITH_HINT = Set.of("state");
     private static final Set<String> BLOCKED_WITHOUT_HINT = Set.of("flows", "endpointGroups", "listeners", "resources");
+
+    private static final String NULL_NOT_ALLOWED_MESSAGE_FORMAT =
+        "'%s' cannot be null; omit the field to leave it unchanged, or send an explicit value";
 
     private final ApiCrudService apiCrudService;
     private final ApiPrimaryOwnerDomainService apiPrimaryOwnerDomainService;
@@ -108,16 +128,16 @@ public class PatchApiUseCase {
         var currentNode = toJsonNode(existingApi);
         var patchNode = parseBody(input.patchBody());
 
-        enforceAllowList(patchNode);
+        enforceAllowList(input.patchType(), patchNode);
 
-        var patchedNode = apiPatchDomainService.applyMergePatch(patchNode, currentNode);
+        var patchedNode = applyPatch(input.patchType(), patchNode, currentNode);
 
         var primaryOwner = apiPrimaryOwnerDomainService.getApiPrimaryOwner(input.auditInfo().organizationId(), input.apiId());
 
         var workflows = workflowQueryService.findAllByApiIdAndType(input.apiId(), Workflow.Type.REVIEW);
         var workflowState = workflows.isEmpty() ? null : workflows.get(0).getState();
 
-        var updatedApi = applyPatchedValues(existingApi, patchedNode, patchNode);
+        var updatedApi = applyPatchedValues(existingApi, patchedNode, input.patchType(), patchNode);
 
         if (input.dryRun()) {
             var validated = updateApiDomainService.validateV4(updatedApi, input.auditInfo());
@@ -129,33 +149,14 @@ public class PatchApiUseCase {
     }
 
     private JsonNode toJsonNode(Api api) {
-        var view = buildPatchableView(api);
-        return objectMapper.valueToTree(view);
+        return objectMapper.valueToTree(PatchableView.from(api, requireHttpV4(api)));
     }
 
-    private PatchableView buildPatchableView(Api api) {
-        if (!(api.getApiDefinitionValue() instanceof io.gravitee.definition.model.v4.Api httpV4)) {
-            throw new IllegalStateException("API definition is not an HTTP v4 definition");
+    private static io.gravitee.definition.model.v4.Api requireHttpV4(Api api) {
+        if (api.getApiDefinitionValue() instanceof io.gravitee.definition.model.v4.Api httpV4) {
+            return httpV4;
         }
-        return new PatchableView(
-            api.getName(),
-            api.getDescription(),
-            api.getVersion(),
-            api.getVisibility() != null ? api.getVisibility().name() : null,
-            api.getLabels(),
-            httpV4.getTags(),
-            api.getApiLifecycleState() != null ? api.getApiLifecycleState().name() : null,
-            api.getCategories(),
-            httpV4.getAnalytics(),
-            httpV4.getFailover(),
-            httpV4.getFlowExecution(),
-            httpV4.getServices(),
-            httpV4.getAllowedInApiProducts(),
-            api.isAllowMultiJwtOauth2Subscriptions(),
-            api.isDisableMembershipNotifications(),
-            httpV4.getProperties(),
-            httpV4.getResponseTemplates()
-        );
+        throw new IllegalStateException("API definition is not an HTTP v4 definition");
     }
 
     private JsonNode parseBody(String body) {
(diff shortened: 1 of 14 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -161,16 +161,49 @@ private JsonNode parseBody(String body) {
...
+        }
+    }
+
+    private void enforceAllowListJsonPatch(JsonNode patchArray) {
+        if (!patchArray.isArray()) {
+            throw new ValidationDomainException("JSON Patch body must be a JSON array");
+        }
+        var index = 0;
+        for (JsonNode op : patchArray) {
+            var pathNode = op.path("path");
```
paulatulis, line 190: This allowlist check only inspects the path field of each operation. RFC 6902 `move` and `copy` ops also carry a `from ` field that points to a source path. A request like `{"op":"move","from":"/listeners/0","path":"/name"}` would pass this check because only /name (the target) is validated against the allowlist. Even if harmless today (since listeners is blocked on the path side too), consider also extracting the top-level field from from and calling `validateField` on it for move and copy ops.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -288,34 +340,71 @@ private Api applyPatchedValues(Api existingApi, JsonNode patchedNode, JsonNode r
...
+    private void rejectExplicitNullOnNonNullable(PatchType patchType, JsonNode rawPatchNode, String field) {
+        switch (patchType) {
+            case MERGE_PATCH -> {
+                if (isExplicitNull(patchType, rawPatchNode, field)) {
+                    throw new ValidationDomainException(
+                        "'" + field + "' cannot be null; omit the field to leave it unchanged, or send an explicit value"
+                    );
+                }
+            }
+            case JSON_PATCH -> {
```
paulatulis: The `JSON_PATCH` branch here only guards against replace/add ops with an explicit null value. It does not guard against `remove /name`, `remove /description`, or `remove /apiVersion`. After the FGE library applies a remove op, `patchedNode` simply won't contain the key, so `textOrDefault` will silently fall back to the existing value — the mutation is invisibly dropped. Is this intentional ("treat remove as leave-unchanged")? If so, please add a test that asserts this contract. If not, this branch should also iterate the operations looking for `op=remove` on the same path.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -288,34 +340,71 @@ private Api applyPatchedValues(Api existingApi, JsonNode patchedNode, JsonNode r
...
+                        );
+                    }
+                }
+            }
         }
     }
 
-    private List<String> resolveNullableList(JsonNode rawPatchNode, JsonNode patchedNode, String field, List<String> existing) {
-        if (isExplicitNull(rawPatchNode, field)) {
+    private List<String> resolveNullableList(
```
paulatulis, line 505: This and `resolveNullableSet` call `isExplicitNull`, which always returns false for `JSON_PATCH ` (it's hardcoded to `MERGE_PATCH` only). They then call `readStringList/readStringSet`, which return `defaultValue ` (the existing value) when the node is null. This means `replace /labels null ` via JSON Patch is a silent no-op — it preserves the existing labels rather than clearing or erroring. Compare this with `resolveNullableObject` (used for analytics, failover, etc.): its inner `readObject` helper does return null when the post-patch node is null, so `replace /analytics null` correctly clears the field. The two behaviors are inconsistent. Either extend ` isExplicitNull ` to handle the JSON Patch null case for list/set fields, or add a `rejectExplicitNullOnNonNullable` guard to explicitly reject replace /labels null. Either way, a test asserting the chosen contract is needed.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
```diff
@@ -710,4 +711,205 @@ void merge_patch_null_on_responseTemplates_passes_null_to_domain_service() {
...
+                .patchBody("[{\"op\":\"replace\",\"path\":\"/responseTemplates\",\"value\":null}]")
+                .dryRun(false)
+                .auditInfo(AuditInfoFixtures.anAuditInfo(ORGANIZATION_ID, ENVIRONMENT_ID, USER_ID))
+                .build()
+        );
+
+        assertThat(output.api().getApiDefinitionHttpV4().getResponseTemplates()).isNull();
+    }
+
+    @Test
```
paulatulis: There are no test cases for the RFC 6902 test operation `({"op":"test","path":"/name","value":"expected"})`. Since the FGE library surfaces `JsonPatchException` as `ValidationDomainException`, clients would receive a 400 when a test assertion fails — but this path is unverified. Please add at minimum: (1) a test where a passing test op is included in a batch and the patch succeeds normally, and (2) a test where a failing test op throws `ValidationDomainException`.

paulatulis (commented):
Claude has much to say. Let me know what you think!

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: `enforceAllowListJsonPatch` now extracts the top-level segment from the `from` field and calls `validateField` on it for `move` and `copy` operations — so `{"op":"move","from":"/listeners/0","path":"/name"}` correctly rejects on `listeners`. Three tests were added: `move` with a blocked `from`, `copy` with a blocked `from`, and `move` with a blocked destination `path`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: addressed by https://github.com/gravitee-io/gravitee-api-management/pull/16578#discussion_r3149814962

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: `resolveNullableList` and `resolveNullableSet` now check whether the post-patch node holds an explicit JSON null for the field and return an empty collection in that case — consistent with `resolveNullableObject` (which delegates to `readObject`, which already returns `null` when the field node is null). Three tests were added: `replace /labels null`, `replace /categories null`, and `replace /tags null` each assert an empty result.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
carlos-andres-osorio: Two tests added to `PatchApiUseCaseTest`:

- `json_patch_test_op_passes_when_assertion_matches` — sends a batch with a passing `test` op (current name matches) followed by a `replace`; asserts the rename goes through.
- `json_patch_test_op_throws_when_assertion_fails` — sends a `test` op whose expected value doesn't match; asserts `ValidationDomainException` is thrown.

paulatulis (commented):
Mostly all good! One thing. Let me know if you think this is a blocker: 
There is still an issue with how `remove` operations are handled on nullable fields (labels, categories, tags, analytics, etc.). While you fixed `replace ... null` by checking `fieldNode.isNull()`, an actual `remove` operation completely drops the key from `patchedNode`. This causes `patchedNode.get("labels")` to be a standard Java `null,` which skips your explicit null check and falls all the way through to the `readStringList/readObject` fallback logic. As a result, it quietly falls back to the existing value.

To fix this, the code needs to distinguish between a field being absent because it was removed via a JSON Patch remove operation, versus a field being absent in a Merge Patch partial update. Currently, remove /labels behaves as a silent no-op.

carlos-andres-osorio:
> Mostly all good! One thing. Let me know if you think this is a blocker: 
> There is still an issue with how `remove` operations are handled on nullable fields (labels, categories, tags, analytics, etc.). While you fixed `replace ... null` by checking `fieldNode.isNull()`, an actual `remove` operation completely drops the key from `patchedNode`. This causes `patchedNode.get("labels")` to be a standard Java `null,` which skips your explicit null check and falls all the way through to the `readStringList/readObject` fallback logic. As a result, it quietly falls back to the existing value.
> 
> To fix this, the code needs to distinguish between a field being absent because it was removed via a JSON Patch remove operation, versus a field being absent in a Merge Patch partial update. Currently, remove /labels behaves as a silent no-op.

Fixed — added `isRemovedByJsonPatch` helper that iterates the raw patch array and returns true when a `remove` op targets `/<field>`. `resolveNullableList`, `resolveNullableSet`, and `resolveNullableObject` now return an empty collection / null when the helper returns true, consistent with the existing `replace ... null` behaviour. Three tests added: `remove /labels` → empty labels, `remove /tags` → empty tags, `remove /analytics` → null analytics.

paulatulis (approved):
Nice!

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -290,68 +356,200 @@ private Api applyPatchedValues(Api existingApi, JsonNode patchedNode, JsonNode r
...
-    private boolean isExplicitNull(JsonNode rawPatchNode, String field) {
-        return rawPatchNode.has(field) && rawPatchNode.get(field).isNull();
+    private List<Property> resolveProperties(PatchType patchType, JsonNode rawPatchNode, JsonNode patchedNode, List<Property> existing) {
+        if (patchType == PatchType.MERGE_PATCH && rawPatchNode.has(FIELD_PROPERTIES)) {
+            if (rawPatchNode.get(FIELD_PROPERTIES).isNull()) {
+                return null;
+            }
+            return readList(patchedNode, FIELD_PROPERTIES, Property.class, null);
+        }
+        if (patchType == PatchType.JSON_PATCH && patchedNode.has(FIELD_PROPERTIES)) {
```
paulatulis: `resolveProperties` handles `replace /properties null` correctly (via `isNullReplacedByJsonPatch`), but a `{"op":"remove","path":"/properties"}` operation is still a silent no-op. After remove, the FGE library drops the key from `patchedNode` entirely, so `patchedNode.has(FIELD_PROPERTIES)` returns false and execution falls through to return existing. This is the same remove gap that was fixed for `labels/tags/analytics` (via `isRemovedByJsonPatch`) but wasn't applied here. The fix mirrors `resolveResponseTemplates` which already correctly handles the remove case (line ~389)
A test `json_patch_remove_properties_clears_field` is also needed.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -169,16 +167,83 @@ private JsonNode parseBody(String body) {
...
+        }
+        var index = 0;
+        for (JsonNode op : patchArray) {
+            var pathNode = op.path("path");
+            if (pathNode.isMissingNode() || !pathNode.isTextual()) {
+                throw new ValidationDomainException("JSON Patch operation at index " + index + " is missing required 'path' field");
+            }
+            validateJsonPatchField(index, "path", pathNode.asText());
+            var opType = op.path("op").asText();
+            if ("move".equals(opType) || "copy".equals(opType)) {
```
paulatulis, line 196: One small observation: the error thrown for a non-textual from field (e.g. "from": 123) uses the same message as a missing from. Consider distinguishing the two: "missing required 'from' field" vs "invalid 'from' field: must be a string". Low priority.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -290,68 +356,200 @@ private Api applyPatchedValues(Api existingApi, JsonNode patchedNode, JsonNode r
...
+        JsonNode patchedNode,
+        Map<String, Map<String, ResponseTemplate>> existing
+    ) {
+        if (patchType == PatchType.MERGE_PATCH && rawPatchNode.has(FIELD_RESPONSE_TEMPLATES)) {
+            if (rawPatchNode.get(FIELD_RESPONSE_TEMPLATES).isNull()) {
+                return null;
+            }
+            return parseResponseTemplates(patchedNode);
+        }
+        if (patchType == PatchType.JSON_PATCH) {
```
paulatulis, line 403: The `remove` and `replace null ` paths look correct. One gap: `patchedNode.has(FIELD_RESPONSE_TEMPLATES)` returns `true` even when the node value is null (Jackson represents `"responseTemplates": null` as a `null NullNode`, not a missing key). But since `isNullReplacedByJsonPatch` guards that inner branch, this is correct as written. The only concern is whether there's a test covering `remove /responseTemplates` specifically — can you confirm one was added?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -290,68 +356,200 @@ private Api applyPatchedValues(Api existingApi, JsonNode patchedNode, JsonNode r
...
+            }
+        }
+        return existing;
     }
 
-    private void rejectExplicitNullOnNonNullable(JsonNode rawPatchNode, String field) {
-        if (isExplicitNull(rawPatchNode, field)) {
-            throw new ValidationDomainException(
-                "'" + field + "' cannot be null; omit the field to leave it unchanged, or send an explicit value"
+    private Map<String, Map<String, ResponseTemplate>> parseResponseTemplates(JsonNode patchedNode) {
```
paulatulis, line 421: catch `(Exception e)` here wraps any `RuntimeException` (NPE, IllegalStateException, etc.) from `fromPatchableResponseTemplates` into a `ValidationDomainException`, surfacing programming errors as 400 responses. The `ApiJsonPatchServiceImpl` was correctly narrowed to catch (IOException), but this local catch wasn't. Consider narrowing to catch (`JsonProcessingException e`) since that's what `treeToValue` actually throws. Any `RuntimeException ` from `fromPatchableResponseTemplates` should propagate as a 500.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -169,16 +167,83 @@ private JsonNode parseBody(String body) {
...
+            enforceAllowListMergePatch(patchNode);
+        }
+    }
+
+    private void enforceAllowListJsonPatch(JsonNode patchArray) {
+        if (!patchArray.isArray()) {
+            throw new ValidationDomainException("JSON Patch body must be a JSON array");
+        }
+        var index = 0;
+        for (JsonNode op : patchArray) {
```
paulatulis, line 189: No upper bound on operations in the patch array. A client could send 10,000+ ops — each triggers the allow-list check, and downstream `isRemovedByJsonPatch/isNullReplacedByJsonPatch` do O(n) linear scans over the raw patch array per nullable field. Under a large patch, this becomes O(n²) in the number of operations. Worth adding a cap at the top of `enforceAllowListJsonPatch: if (patchArray.size() > MAX_PATCH_OPS)` throw new `ValidationDomainException(...)`. Doesn't need to block this PR, but should be a follow-up ticket.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: Fixed. `resolveProperties` now checks `patchedNode.get(FIELD_PROPERTIES) == null && isRemovedByJsonPatch(patchType, rawPatchNode, FIELD_PROPERTIES)` before the `has` branch, mirroring the pattern in `resolveResponseTemplates`. A `json_patch_remove_properties_clears_field` test was added to `PatchApiUseCaseTest`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: Done. The two branches now produce distinct messages: a missing `from` throws `"missing required 'from' field"` and a non-textual `from` throws `"invalid 'from' field: must be a string"`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: Yes — two tests cover this in `PatchApiUseCaseTest`: `json_patch_remove_op_on_null_responseTemplates_sets_responseTemplates_to_null` and `json_patch_remove_op_on_existing_responseTemplates_clears_them`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: Narrowed. `parseResponseTemplates` now catches `JsonProcessingException` only — any `RuntimeException` from `fromPatchableResponseTemplates` propagates as a 500 rather than being swallowed into a 400.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: If there are more than 200 operations, an error will be thrown.

PR #16586, opened 2026-04-24
feat(portal): add analytics computation endpoints and mappers

aqibmohammadkhan:
Introduce new REST endpoints for analytics measures, facets, and time-series computations. Update OpenAPI spec, add request/response mappers, and include relevant services with tests. Handle null filters and facets defensively in core services.

## Issue

https://gravitee.atlassian.net/browse/APIM-13534

## Description

This PR introduces portal-facing analytics computation endpoints to support the Next Gen Portal dashboards experience. It adds dedicated REST endpoints to compute measures, facets, and time-series analytics, backed by the shared core analytics-engine use cases, and updates the Portal OpenAPI contract accordingly.

New endpoints under /portal/environments/{envId}/analytics:

- POST /analytics/measures
- POST /analytics/facets
- POST /analytics/time-series

Feature-flag gated via PORTAL_NEXT_ANALYTICS_ENABLED at the analytics root resource, so all analytics endpoints inherit the same access check.

Portal ↔ Core mapping layer (PortalAnalyticsMeasuresMapper) converting Portal DTOs to core models (metrics/measures/filters/facets/time range) and returning mapped responses; unknown enum values return 400 Bad Request.

Core robustness hardening:

- Validator improvements around filter validation and request constraints.
- Tests covering the new endpoints (successful mapping/delegation + 403 when the feature is disabled).

OpenAPI
Portal contract updated to document the new analytics computation operations (computeAnalyticsMeasures, computeAnalyticsFacets, computeAnalyticsTimeSeries) with request/response schemas and standard error responses.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/mapper/PortalAnalyticsMeasuresMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/mapper/PortalAnalyticsMeasuresMapper.java
@@ -1,0 +1,281 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.portal.rest.mapper;
+
+import io.gravitee.apim.core.analytics_engine.model.FacetBucketResponse;
+import io.gravitee.apim.core.analytics_engine.model.FacetMetricMeasuresRequest;
+import io.gravitee.apim.core.analytics_engine.model.FacetSpec;
+import io.gravitee.apim.core.analytics_engine.model.FacetsRequest;
+import io.gravitee.apim.core.analytics_engine.model.FacetsResponse;
+import io.gravitee.apim.core.analytics_engine.model.Filter;
+import io.gravitee.apim.core.analytics_engine.model.FilterSpec;
+import io.gravitee.apim.core.analytics_engine.model.Measure;
+import io.gravitee.apim.core.analytics_engine.model.MeasuresRequest;
+import io.gravitee.apim.core.analytics_engine.model.MeasuresResponse;
+import io.gravitee.apim.core.analytics_engine.model.MetricFacetsResponse;
+import io.gravitee.apim.core.analytics_engine.model.MetricMeasuresRequest;
+import io.gravitee.apim.core.analytics_engine.model.MetricMeasuresResponse;
+import io.gravitee.apim.core.analytics_engine.model.MetricSpec;
+import io.gravitee.apim.core.analytics_engine.model.TimeRange;
+import io.gravitee.apim.core.analytics_engine.model.TimeSeriesBucketResponse;
+import io.gravitee.apim.core.analytics_engine.model.TimeSeriesMetricResponse;
+import io.gravitee.apim.core.analytics_engine.model.TimeSeriesRequest;
+import io.gravitee.apim.core.analytics_engine.model.TimeSeriesResponse;
+import io.gravitee.apim.core.observability.model.FilterOperator;
+import io.gravitee.apim.core.observability.model.NumberRange;
+import io.gravitee.rest.api.portal.rest.model.AnalyticsArrayFilter;
+import io.gravitee.rest.api.portal.rest.model.AnalyticsComputationMetricRequest;
+import io.gravitee.rest.api.portal.rest.model.AnalyticsFacetBucket;
+import io.gravitee.rest.api.portal.rest.model.AnalyticsFacetBucketGroup;
+import io.gravitee.rest.api.portal.rest.model.AnalyticsFacetBucketLeaf;
+import io.gravitee.rest.api.portal.rest.model.AnalyticsFacetMetricRequest;
+import io.gravitee.rest.api.portal.rest.model.AnalyticsFacetName;
+import io.gravitee.rest.api.portal.rest.model.AnalyticsFacetsRequest;
+import io.gravitee.rest.api.portal.rest.model.AnalyticsFacetsResponse;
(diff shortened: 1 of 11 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/test/java/io/gravitee/rest/api/portal/rest/resource/PortalAnalyticsComputationResourceTest.java
```diff
@@ -0,0 +1,233 @@
...
+        assertThat(body.getMetrics().getFirst().getName()).isEqualTo("HTTP_REQUESTS");
+        assertThat(body.getMetrics().getFirst().getUnit()).isEqualTo("NUMBER");
+        assertThat(body.getMetrics().getFirst().getMeasures()).hasSize(1);
+        assertThat(body.getMetrics().getFirst().getMeasures().getFirst().getName()).isEqualTo("COUNT");
+        verify(computeMeasuresUseCase).execute(any());
+    }
+
+    @Test
+    @DisplayName("POST /measures — 403 when PORTAL_NEXT_ANALYTICS_ENABLED is false")
+    void measures_forbiddenWhenDisabled() {
```
dev-marek: for tests we use snake case + and replace underscores annotation

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/analytics/PortalAnalyticsResource.java
kovaren: The comment mentions "dashboards today, computation tomorrow", which doesn't seem to be relevant anymore.

WDYT about removing this whole comment section? It's not adding any value, just the agent being overly verbose 🤔

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/mapper/PortalAnalyticsMeasuresMapper.java
```diff
@@ -0,0 +1,211 @@
...
+    AnalyticsTimeSeriesResponseMetric toPortalMetric(TimeSeriesMetricResponse metric);
+
+    AnalyticsTimeSeriesBucket toPortalBucket(TimeSeriesBucketResponse bucket);
+
+    AnalyticsMeasure toPortalMeasure(Measure measure);
+
+    List<AnalyticsMeasure> toPortalMeasures(List<Measure> measures);
+
+    default TimeRange toCoreTimeRange(AnalyticsTimeRange timeRange) {
+        if (timeRange == null) {
```
dev-marek: I see a lot `null`-value-handling statements. Please consider changing some of those fields to required and check if they can/should be nullable

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
```diff
@@ -3265,6 +3265,102 @@ paths:
                 500:
                     $ref: "#/components/responses/InternalServerError"
 
+    /analytics/measures:
```
kovaren, line 3268: There are a few differences between the [Portal](https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml) and the [Management v2](https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-analytics.yaml) specs.

Is the divergence intentional? WDYT about aligning on the following points:

1. **Filter model** — mAPI `Filter` is a discriminated **`oneOf`** (`StringFilter` / `NumberFilter` / `ArrayFilter`) with typed names/operators. Portal uses a flat `AnalyticsFilter` (strings + `value` as `object`). Same backend filters, very different OpenAPI strictness and generated clients.

2. **Enums vs strings** — mAPI request/response use **`$ref` enums** (e.g. `MetricName`, `MeasureName`, `FacetName`). Portal uses plain **`string`** for the same fields (relying on the mapper to turn bad values into 400). Documented client experience diverges a lot.

3. **Facets / time-series constraints** — mAPI encodes **limits in schema** (e.g. `by` **maxItems: 3** for facets, **maxItems: 2** for time series, `limit` **minimum: 1**, `metrics` **minItems**). Portal’s `by` / `limit` are mostly unconstrained in YAML even though the engine applies the same rules (e.g. `AnalyticsQueryValidator` / core).

4. **Bucket / response shape** — mAPI uses **discriminated unions** for buckets (`type: LEAF | GROUP`, `oneOf`). Portal `AnalyticsFacetBucket` / `AnalyticsTimeSeriesBucket` are **single** recursive objects with optional `measures` vs `buckets` and no discriminator in the spec.

5. **Time-series `interval`** — mAPI `Interval` allows **oneOf** (e.g. shorthand string + number ms). Portal `interval` is **int64 ms**.

6. **Documentation depth** — mAPI operations include **long request/response `examples`**. Portal’s new operations are minimal (no examples on the `POST` bodies / 200 responses). Same for **operationId** naming (`query*` vs `compute*`) and **tags** (`computation` vs `Analytics`).

kovaren (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
aqibmohammadkhan: We started with a flatter Portal contract to ship computation quickly and keep the first client simple (stringly-typed fields + a single mapper doing enum validation). That trades strict OpenAPI / codegen quality for speed, which isn’t ideal now that these endpoints are part of the “real” product surface.

i have aligned Portal with the Management v2 analytics shapes you called out: discriminated filters, $ref enums where mAPI uses them, schema-level constraints (minItems / maxItems / minimum, etc.), discriminated bucket types, AnalyticsInterval as oneOf (ms + duration shorthand) with the same semantics as mAPI, richer examples, and query* / computation tagging for parity.

kovaren: Yep, we should try and avoid trading quality for speed, especially if it requires diverging from the existing spec.

kovaren (approved):

dev-marek (approved):

PR #16593, opened 2026-04-24
feat: expose group resources in the automation API

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-2551

Expose groups through the Automation API (`/automation/organizations/{orgId}/environments/{envId}/groups`), allowing the operator to manage groups without relying on the legacy v1/v2 endpoints.

**Endpoints:**
- `PUT /groups` — create or update a group from a `GroupSpec` (supports `dryRun` and `hridContainsUUID`)
- `GET /groups/{hrid}` — get a group by HRID (includes members)
- `DELETE /groups/{hrid}` — delete a group by HRID

Group IDs are derived deterministically from the group name via `HRIDToUUID.group()`. The `hridContainsUUID` query param supports backward compatibility for groups created before migration.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/GroupsResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/GroupsResource.java
@@ -76,7 +76,8 @@ public class GroupsResource extends AbstractResource {
         var groupCRDSpec = GroupMapper.INSTANCE.groupSpecToGroupCRDSpec(spec);
 
         if (hridContainsUUID) {
-            groupCRDSpec.setId(spec.getName());
+            groupCRDSpec.setId(spec.getHrid());
+            groupCRDSpec.setHrid(null);
         } else {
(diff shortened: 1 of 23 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/GroupsResource.java
```diff
@@ -0,0 +1,95 @@
...
+            CrdIdHelper.generateGroupId(groupCRDSpec, auditInfo);
+        }
+
+        if (dryRun) {
+            var status = validateGroupCRDUseCase.execute(new ImportGroupCRDUseCase.Input(auditInfo, groupCRDSpec)).status();
+
+            return Response.ok(GroupMapper.INSTANCE.groupSpecAndStatusToGroupState(spec, status)).build();
+        }
+
+        var status = importGroupCRDUseCase.execute(new ImportGroupCRDUseCase.Input(auditInfo, groupCRDSpec)).status();
```
benoitgravitee, line 91: I know this is not what we have done so far. But I wonder if should start returning what GET would return (not a redirection).
It would probably help having less overlays, and it plays better with Terraform in general.

benoitgravitee (approved):

PR #16600, opened 2026-04-24
fix: propagate allowedInApiProducts through v4 HTTP API create path

carlos-andres-osorio:
## Issue

[APIM-13786](https://gravitee.atlassian.net/browse/APIM-13786) — v4 API create path: propagate `allowedInApiProducts` through `NewHttpApi` to domain

Parent: [APIM-13738](https://gravitee.atlassian.net/browse/APIM-13738)

## Why

On the v4 HTTP Proxy API create path, `allowedInApiProducts` was silently dropped between the REST DTO and the domain model. The persisted value was always `null`; the subsequent GET only returned `false` because the read path coalesces `null → false`. Two problems: (1) the OpenAPI spec advertises a create default of `true`, but clients always observe `false` on GET — spec and behavior disagree; (2) explicit values from callers (`true` or `false`) were silently lost.

This change wires the field through the HTTP create bridge so the caller's value (or the OpenAPI-layer default) reaches the domain unchanged. Values survive verbatim: `true`, `false`, and `null` all pass through without coercion.

## What changed

- `NewHttpApi.java` — added `private Boolean allowedInApiProducts;` and chained `.allowedInApiProducts(allowedInApiProducts)` into `toApiDefinitionBuilder()`. No `@Builder.Default` — the OpenAPI-layer default on `CreateApiV4` remains the single source of truth.
- `ApiMapper.java` — no source edits. MapStruct auto-maps the field on `mapToNewHttpApi` via name-matching now that source (`CreateApiV4`) and target (`NewHttpApi`) share the field. `mapToNewNativeApi` stays silent — MapStruct's default policy does not warn on unmapped source properties, so no `@BeanMapping` suppression was needed. Generated `ApiMapperImpl` was inspected to confirm: HTTP mapper sets the field, native mapper does not.
- Tests — added behavioral coverage at all three layers (bridge builder, use case round-trip, REST DTO mapper). Native mapper is guarded at compile time via the type system (`NewNativeApi` has no accessor for the field).

## User-facing impact

⚠️ **Behavior change**: APIs created through `POST /management/v2/environments/{envId}/apis` without an explicit `allowedInApiProducts` value will now persist `true` (the OpenAPI default on `CreateApiV4`) instead of the previously-masked `false`. Clients that assumed the old masked default will see the stored value change after merge. This aligns stored state with the documented OpenAPI contract.

No API signature or config key changes. No docs update required — the OpenAPI spec already documents `true` as the default.

## How to test

1. Create a v4 HTTP Proxy API without `allowedInApiProducts` in the body; GET it — expect `"allowedInApiProducts": true`.
2. Create with `"allowedInApiProducts": false`; GET — expect `false`.
3. Create with `"allowedInApiProducts": true`; GET — expect `true`.

Automated coverage:

```
mvnd -pl gravitee-apim-rest-api/gravitee-apim-rest-api-service,gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest -am -Dtest='NewHttpApiTest,CreateHttpApiUseCaseTest,ApiMapperTest' -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test
```

## Gotchas / Follow-up

- `ApiCRDSpec` has the same gap on a different entry point — explicitly out of scope here; no follow-up ticket filed per the action plan.
- `NewNativeApi` intentionally does not gain the field — `allowedInApiProducts` is HTTP-Proxy-only.

[APIM-13786]: https://gravitee.atlassian.net/browse/APIM-13786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APIM-13738]: https://gravitee.atlassian.net/browse/APIM-13738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapperTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapperTest.java
@@ -44,6 +44,8 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
 import org.assertj.core.api.SoftAssertions;
+import org.junit.jupiter.api.DisplayNameGeneration;
+import org.junit.jupiter.api.DisplayNameGenerator;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
@@ -52,6 +54,7 @@ import org.junit.jupiter.params.provider.EnumSource;
 import org.junit.jupiter.params.provider.MethodSource;
 import org.mapstruct.factory.Mappers;
 
+@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
 public class ApiMapperTest {
 
     private final ApiMapper apiMapper = Mappers.getMapper(ApiMapper.class);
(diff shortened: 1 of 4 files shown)


paulatulis (approved):
Excellent

wbabyte (approved):

PR #16617, opened 2026-04-27
feat(APIM-13621/13622): capture OTel Logs in gateway & add Console UI toggle

araffass:
## Issue

https://gravitee.atlassian.net/browse/APIM-13621
https://gravitee.atlassian.net/browse/APIM-13622

## Description

Implements per-API OTel log capture and the Console UI toggle to enable it

## Additional context

 _**Gateway (APIM-13621):**_
 - Adds OtelLogs model and analytics.otelLogs field to the v4 API definition
 - Extends LoggingContext and AnalyticsUtils to propagate otelLogsEnabled
 - Injects traceId/spanId into the 4 log capture classes (LogEntrypointRequest, LogEndpointRequest, LogEndpointResponse, LogEntrypointResponse) when OTel logs are enabled
- Wires the OtelReporter in DefaultApiReactorFactory
- Adds OpenAPI spec OtelLogsV4 and updates openapi-apis.yaml

**_Console UI (APIM-13622):_**
- Adds OTel Logs toggle in Reporter Settings under the OpenTelemetry section
- Toggle is enabled only when Analytics + Tracing are both on
- Saving triggers API update and shows the Deploy banner

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/reporter-settings/reporter-settings-message/reporter-settings-message.component.html
+++ b/gravitee-apim-console-webui/src/management/api/reporter-settings/reporter-settings-message/reporter-settings-message.component.html
@@ -235,6 +235,12 @@
             Enable only for deep debugging - increases trace size significantly.
             <mat-slide-toggle gioFormSlideToggle formControlName="tracingVerbose"></mat-slide-toggle>
           </gio-form-slide-toggle>
+          <gio-form-slide-toggle>
+            <gio-form-label>OTel Logs</gio-form-label>
+            Emit message payloads as OpenTelemetry log records correlated to the active trace. Enables log-to-trace linking in Grafana and
+            other OTel-compatible backends. Capture is subject to the sampling strategy configured above.
+            <mat-slide-toggle gioFormSlideToggle formControlName="otelLogsEnabled"></mat-slide-toggle>
+          </gio-form-slide-toggle>
         </div>
       </mat-card-content>
     </mat-card>
--- a/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/analytics/AnalyticsUtils.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/analytics/AnalyticsUtils.java
@@ -18,6 +18,7 @@ package io.gravitee.gateway.reactive.core.v4.analytics;
 import io.gravitee.definition.model.v4.analytics.Analytics;
 import io.gravitee.definition.model.v4.analytics.logging.Logging;
 import io.gravitee.definition.model.v4.analytics.logging.LoggingMode;
+import io.gravitee.definition.model.v4.analytics.logging.OtelLogs;
 import io.gravitee.definition.model.v4.analytics.tracing.Tracing;
 import io.gravitee.gateway.opentelemetry.TracingContext;
 import io.gravitee.gateway.report.guard.LogGuardService;
@@ -80,8 +81,19 @@ public class AnalyticsUtils {
             Logging logging = analytics.getLogging();
             if (logging != null) {
                 final LoggingMode loggingMode = logging.getMode();
-                return loggingMode != null && loggingMode.isEnabled();
+                if (loggingMode != null && loggingMode.isEnabled()) {
+                    return true;
+                }
             }
+            return isOtelLogsEnabled(analytics);
+        }
+        return false;
+    }
+
+    public static boolean isOtelLogsEnabled(final Analytics analytics) {
+        if (analytics != null && analytics.isEnabled()) {
+            OtelLogs otelLogs = analytics.getOtelLogs();
+            return otelLogs != null && otelLogs.isEnabled();
         }
         return false;
     }
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
@@ -424,7 +424,7 @@ public class DefaultApiReactorFactory extends AbstractReactorFactory<Api> {
         if (tracingEnabled) {
             Tracer tracer = openTelemetryFactory.createTracer(
                 api.getId(),
-                api.getName(),
+                node.application(),
                 serviceNameSpace,
                 api.getApiVersion(),
                 instrumenterTracerFactories,
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
@@ -16,6 +16,7 @@
 package io.gravitee.gateway.reactive.handlers.api.v4;
 
 import io.gravitee.common.event.EventManager;
+import io.gravitee.common.util.Version;
 import io.gravitee.definition.model.DefinitionVersion;
 import io.gravitee.definition.model.v4.ApiType;
 import io.gravitee.definition.model.v4.listener.ListenerType;
@@ -70,6 +71,7 @@ import io.gravitee.plugin.entrypoint.EntrypointConnectorPluginManager;
 import io.gravitee.plugin.policy.PolicyClassLoaderFactory;
 import io.gravitee.plugin.policy.PolicyPlugin;
 import java.util.List;
+import java.util.Map;
 import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.core.ResolvableType;
@@ -423,11 +425,12 @@ public class DefaultApiReactorFactory extends AbstractReactorFactory<Api> {
         boolean tracingEnabled = isApiTracingEnabled(api);
         if (tracingEnabled) {
             Tracer tracer = openTelemetryFactory.createTracer(
-                api.getId(),
+                node.id(),
                 node.application(),
                 serviceNameSpace,
-                api.getApiVersion(),
+                Version.RUNTIME_VERSION.MAJOR_VERSION,
                 instrumenterTracerFactories,
+                Map.of("gravitee.api.id", api.getId(), "gravitee.api.name", api.getName()),
                 TracingRedactionMapper.toRedactionConfig(api)
             );
             return new TracingContext(tracer, true, isApiTracingVerboseEnabled(api));
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/analytics/logging/request/LogEndpointRequest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/analytics/logging/request/LogEndpointRequest.java
@@ -39,6 +39,12 @@ public class LogEndpointRequest extends LogRequest {
     }
 
     public void setupCapture(HttpExecutionContextInternal ctx) {
+        if (loggingContext.isOtelLogsEnabled()) {
+            var tracer = ctx.getTracer();
+            this.setTraceId(tracer != null ? tracer.traceId() : null);
+            this.setSpanId(tracer != null ? tracer.spanId() : null);
+        }
+
         final Buffer buffer = Buffer.buffer();
 
         if (isLogPayload() && loggingContext.isContentTypeLoggable(request.headers().get(HttpHeaderNames.CONTENT_TYPE), ctx)) {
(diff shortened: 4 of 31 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
```diff
@@ -423,7 +423,7 @@ protected TracingContext createTracingContext(final Api api, final String servic
         if (isApiTracingEnabled(api)) {
             Tracer tracer = openTelemetryFactory.createTracer(
                 api.getId(),
-                api.getName(),
+                node.application(),
```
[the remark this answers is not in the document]
araffass: This is intentional and an actual fix, using api.getName() caused each API to appear as a separate service in Tempo/Grafana which is wrong

gravitee-apim-console-webui/src/management/api/reporter-settings/reporter-settings-message/reporter-settings-message.component.html
```diff
@@ -235,6 +235,12 @@
             Enable only for deep debugging - increases trace size significantly.
             <mat-slide-toggle gioFormSlideToggle formControlName="tracingVerbose"></mat-slide-toggle>
           </gio-form-slide-toggle>
+          <gio-form-slide-toggle>
+            <gio-form-label>OTel Logs</gio-form-label>
+            Emit message payloads as OpenTelemetry log records correlated to the active trace. Enables log-to-trace linking in Grafana and
+            other OTel-compatible backends.
```
remibaptistegio: The capture of message payload will depend on the sampling strategy used no ? Maybe worth mentioning it if that's the case, no ?

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/analytics/logging/request/LogEndpointRequest.java
```diff
@@ -39,6 +39,12 @@ public LogEndpointRequest(LoggingContext loggingContext, HttpRequestInternal req
     }
 
     public void setupCapture(HttpExecutionContextInternal ctx) {
+        if (loggingContext.isOtelLogsEnabled()) {
+            var tracer = ctx.getTracer();
+            this.setTraceId(tracer != null ? tracer.traceId() : null);
+            this.setSpanId(tracer != null ? tracer.spanId() : null);
```
remibaptistegio: I'm not sure tracer can be null, either we have a NoOpTracer or an OpenTelemetryTracer no ?

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
```diff
@@ -424,7 +424,7 @@ protected TracingContext createTracingContext(final Api api, final String servic
         if (tracingEnabled) {
             Tracer tracer = openTelemetryFactory.createTracer(
                 api.getId(),
-                api.getName(),
+                node.application(),
```
remibaptistegio: I'm not sure to understand why we change the service.name but not the service.instance.id (still api id) and the service.version (still api version).
I think we need to align all these attributes to the node info (id, name and version) but add additional resource attributes to still have the api id and name available in the trace.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/test/java/io/gravitee/gateway/reactive/handlers/api/v4/analytics/logging/request/LogEndpointRequestTest.java
```diff
@@ -221,6 +222,37 @@ void should_log_body_not_captured_when_body_is_not_loggable() {
...
+
+        cut.setupCapture(ctx);
+        triggerRequestToBackend(null, false);
+
+        assertThat(cut.getTraceId()).isEqualTo("abc123traceId00000000000000000000");
+        assertThat(cut.getSpanId()).isEqualTo("def456spanId0000");
+    }
+
+    @Test
+    void should_set_null_traceId_and_spanId_when_tracer_is_null() {
```
remibaptistegio: Maybe add a test with a NoOpTracer to check the behavior, no ?

gravitee-apim-integration-tests/src/test/java/io/gravitee/apim/integration/tests/http/logging/LoggingV4IntegrationTest.java
```diff
@@ -552,6 +553,57 @@ void should_not_log_video(HttpClient httpClient, VertxTestContext context) throw
...
+            analytics.setEnabled(true);
+            analytics.setOtelLogs(otelLogs);
+
+            if (api.getDefinition() instanceof Api apiDefinition) {
+                apiDefinition.setAnalytics(analytics);
+            }
+        }
+
+        @Test
+        void should_report_log_when_otel_logs_enabled_without_es_logging(HttpClient httpClient, VertxTestContext context) throws Exception {
```
remibaptistegio: Maybe instead of `es` use `analytics` since it's can be other backend and not only elasticsearch. It depends on the reporter used by the customer.

gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/analytics/AnalyticsUtils.java
```diff
@@ -80,8 +81,19 @@ public static boolean isLoggingEnabled(final Analytics analytics) {
             Logging logging = analytics.getLogging();
             if (logging != null) {
                 final LoggingMode loggingMode = logging.getMode();
-                return loggingMode != null && loggingMode.isEnabled();
+                if (loggingMode != null && loggingMode.isEnabled()) {
+                    return true;
+                }
             }
+            return isOtelLogsEnabled(analytics);
```
remibaptistegio, line 88: There is one thing that I'm still not sure about. If we activate tracing and want payloads, we are activating the otel logs. But by doing that, my reporter (tcp, elastic ...) will also produce those logs which means that I will store the same data in two different backends (tempo and elasticsearch for instance). Is it ok for the product to have this behavior ? Should we try to found a solution to avoid this like maybe add a new target variable inside Reportable (tracing, analytics ...) that could be use by reporters to decide if they want to process this Reportable ?

remibaptistegio (changes requested):

gravitee-apim-console-webui/src/management/api/reporter-settings/reporter-settings-message/reporter-settings-message.component.html
araffass: Yes it's worth mentioning, i added a note to the description in the message

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/analytics/logging/request/LogEndpointRequest.java
araffass: Yes, it was just an (extra) precaution i've took, since no need i simplified that by removing that null check.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
araffass: I've indeed been thinking about that since i only patched service.name but left service.instance.id and service.version pointing at API-level values. 
I fixed both DefaultApiReactorFactory and TcpApiReactorFactory to be same as  what the global gateway tracer does: node id/name/version as the OTel service identity, with gravitee.api.id and gravitee.api.name kept as additional resource attributes.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/test/java/io/gravitee/gateway/reactive/handlers/api/v4/analytics/logging/request/LogEndpointRequestTest.java
araffass: Done. Replaced the  null tracer test with a NoOpTracer test

gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/analytics/AnalyticsUtils.java
araffass: This indeed is intentional for now: OTel Logs adds a new export destination without replacing existing reporters. 

Users who want OTel-Only logging can simply leave the ES logging mode off (entrypoint/endpoint toggles unchecked). In that case, the Log record is still created for the OTel pipeline, but the ES reporter receives a log with no logging mode configured, so it effectively no-ops.

The target variable you suggested is a solid long term idea and worth a follow-up, but it would require a change to gravitee-reporter-api and more tests which is guess is out of scope for now. I propose to keep that change for a V2 and keep it in mind as a follow up ticket

gravitee-apim-integration-tests/src/test/java/io/gravitee/apim/integration/tests/http/logging/LoggingV4IntegrationTest.java
araffass: Done

gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/analytics/AnalyticsUtils.java
remibaptistegio: Hum I'm not sure to follow when you say:

> In that case, the Log record is still created for the OTel pipeline, but the ES reporter receives a log with no logging mode configured, so it effectively no-ops.

For me we are using the same Reportable in both case and the reporter does not know about the configuration, it just checks that the class of the Reportable is supported and that's all. So even if the configuration deactivate the logging for the analytics part but we capture it for the tracing part, we will still produce a reportable that will be processed by the analytics reporter (tcp or elastic) and the otel reporter no ?

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
```diff
@@ -423,11 +425,12 @@ protected TracingContext createTracingContext(final Api api, final String servic
...
             Tracer tracer = openTelemetryFactory.createTracer(
-                api.getId(),
-                api.getName(),
+                node.id(),
+                node.application(),
                 serviceNameSpace,
-                api.getApiVersion(),
+                Version.RUNTIME_VERSION.MAJOR_VERSION,
                 instrumenterTracerFactories,
+                Map.of("gravitee.api.id", api.getId(), "gravitee.api.name", api.getName()),
```
remibaptistegio, line 433: After thinking a bit more about this, I wonder if we should not instead use classic span attributes on the root span instead of resource attribute. Wdyt ?

remibaptistegio (changes requested):

gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/analytics/AnalyticsUtils.java
araffass: You're right, I double checked. The ES reporter doesn't check the logging config, it just handles whatever Log object it receives. So when otelLogs is on but ES logging mode is off, the Log still ends up in Elasticsearch. That's a real concern. I still think the target field on Reportable is the clean fix but as it needs a
  gravitee-reporter-api change, and we need a fist version to have a clear vision to the Epic, I propose to keep that as a follow-up ticket if that's ok with you

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
araffass: Good question. I went with resource attributes because each API creates its own SdkTracerProvider with its own Resource, so gravitee.api.id and gravitee.api.name are scoped per-API anyway, so i think there's no risk of leaking across APIs. The bonus here using that approach is they propagate automatically to every span from that tracer (including child spans created by policies for example) without having to explicitly set them anywhere.
The span attribute you proposed is interesting too but with it we'll have to add them manually on every root span. In Tempo they also show up in the resource section which feels like the right place for this kind of service level metadata. Does that make sense ? or you you think the span approach is still better here ?

gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/analytics/AnalyticsUtils.java
remibaptistegio: Ok fine by me but we must not forget this follow up ticket 👍

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
remibaptistegio: Yes I understand your way of thinking on this. Our way of creating one Tracer per API is not an usual implementation (it's not a best practice to create lots of Tracer for one process). But for now I guess we can go like this and rethink this in the future.

remibaptistegio (approved):

vikrantgravitee (approved):

PR #16622, opened 2026-04-27
fix: response schema for automation group diverges

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-2839

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/GroupMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/GroupMapper.java
@@ -23,7 +23,6 @@ import io.gravitee.apim.rest.api.automation.model.Errors;
 import io.gravitee.apim.rest.api.automation.model.GroupMember;
 import io.gravitee.apim.rest.api.automation.model.GroupSpec;
 import io.gravitee.apim.rest.api.automation.model.GroupState;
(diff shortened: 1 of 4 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/GroupMapper.java
```diff
@@ -53,31 +52,28 @@ public interface GroupMapper {
     @Mapping(target = "roles", qualifiedByName = "roleScopeMapToStringMap")
     GroupMember memberToGroupMember(GroupCRDSpec.Member member);
 
-    GroupStatus toGroupStatus(GroupCRDStatus status);
-
     Errors toErrors(GroupCRDStatus.Errors errors);
 
     default GroupState groupSpecAndStatusToGroupState(GroupSpec spec, GroupCRDStatus status) {
```
benoitgravitee, line 57: for what comes from status: 
 `@Mapping(target = "id", source = "status.id")`
The rest should be mapped automatically

benoitgravitee (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/GroupMapper.java
a-cordier: As discuss offline / this is indeed to be adressed, but will come in a further refactor

PR #16626, opened 2026-04-27
fix: only display mcp button for mcp tool server.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-13808

## Description

The documentation showed the MCP tools control whenever the API payload included `mcp`. For `MCP_PROXY` APIs that is misleading: the API is already an MCP server, not an API that exposes an MCP tool server alongside normal traffic.

## Changes

- Extend the `ApiType` with `MCP_PROXY`, `A2A_PROXY`, `LLM_PROXY`.
- Avoid displaying MCP button for APIs of type MCP_PROXY.

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.spec.ts
@@ -23,6 +23,7 @@ import { of } from 'rxjs/internal/observable/of';
 
 import { DocumentationFolderComponent } from './documentation-folder.component';
 import { DocumentationFolderComponentHarness } from './documentation-folder.component.harness';
+import { ApiType } from '../../../../entities/api/api';
 import { PortalNavigationItem } from '../../../../entities/portal-navigation/portal-navigation-item';
 import { makeItem, MOCK_ITEMS } from '../../../../mocks/portal-navigation-item.mocks';
 import { ApiService } from '../../../../services/api.service';
@@ -58,6 +59,7 @@ describe('DocumentationFolderComponent', () => {
       content: string;
       isAuthenticated: boolean;
       apiHasMcp: boolean;
+      apiType: ApiType;
     }> = {
       queryParams: { selectedId: 'p1' },
       items: MOCK_CHILDREN,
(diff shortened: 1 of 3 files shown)


gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.spec.ts
```diff
@@ -58,6 +58,7 @@ describe('DocumentationFolderComponent', () => {
       content: string;
       isAuthenticated: boolean;
       apiHasMcp: boolean;
+      apiType: string;
```
aqibmohammadkhan: `apiType` is typed as string; consider `apiType?: ApiType` for better type-safety.

gravitee-apim-portal-webui-next/src/app/documentation/components/documentation-folder/documentation-folder.component.ts
```diff
@@ -90,7 +90,7 @@ export class DocumentationFolderComponent {
     params: this.apiId,
     stream: ({ params }) => (params ? this.apiService.details(params) : of(null)),
   });
-  apiHasMcp = computed(() => !this.api.error() && !!this.api.value()?.mcp);
+  apiHasMcp = computed(() => !this.api.error() && !!this.api.value()?.mcp && this.api.value()?.type !== 'MCP_PROXY');
```
aqibmohammadkhan, line 93: This fixes the documentation page, but other portal screens still gate MCP UI on `!!api.mcp` only (e.g. API details chip + “MCP Tools” tab, catalog “`isEnabledMcpServer`”). If MCP_PROXY APIs can appear there too, we may want to align the predicate across the portal (ideally a shared helper like `isMcpToolServer(api)`), otherwise the MCP UI may still show up in other places.

kovaren: That's a very good point, actually the mcp field should not even be returned for the MCP Proxy APIs. This is a regression introduced by this https://github.com/gravitee-io/gravitee-api-management/pull/16648. So the fix of the current PR will be fully covered by a revert of the aforementioned PR.

kovaren:
This fix is a regression due to https://github.com/gravitee-io/gravitee-api-management/pull/16648 and will be covered by its revert.

PR #16627, opened 2026-04-27
feat(console): explicitly send allowedInApiProducts=false in v4 API creation

carlos-andres-osorio:
## Issue

[APIM-13787](https://gravitee.atlassian.net/browse/APIM-13787)

## Why

The console's v4 API creation wizard currently omits `allowedInApiProducts` from the POST body. Once the back-end fix in the parent epic ships, the OpenAPI-declared default of `true` will start reaching the persisted API, silently flipping the GET response for every console-created v4 API from `false` to `true`. Explicitly sending `false` preserves the current observable behaviour regardless of when the back-end change lands.

## What changed

- Added `allowedInApiProducts?: boolean` to the `CreateApiV4` TypeScript interface (on the V4 type, not the shared base — the field is documented as V4 HTTP Proxy only).
- Hard-coded `allowedInApiProducts: false` as a fixed literal property in the `CreateApiV4` payload assembled inside `createApi$` in `api-creation-v4.component.ts`. No form control or wizard step is introduced.
- Extended `api-creation-v4.http-proxy.component.spec.ts` with a test that drives the wizard to completion, captures the outgoing `POST /apis` request via `HttpTestingController`, and asserts `allowedInApiProducts: false` in the body.

## User-facing impact

No observable change until the back-end fix in APIM-13738 ships. After it lands, console-created v4 APIs will continue returning `allowedInApiProducts: false` on GET — the same as today — rather than flipping to `true` from the OpenAPI default.

## How to test

Create a v4 HTTP Proxy API via the console wizard and inspect the outgoing `POST /apis` request body — it should contain `"allowedInApiProducts": false`. The new spec in `api-creation-v4.http-proxy.component.spec.ts` covers this automatically.

## Gotchas / Follow-up

This change is a no-op in production until APIM-13738 merges. Safe to land first; the explicit `false` is harmless even if the back-end change is later reverted, because the current back-end coalesces `null → false` on GET.

[APIM-13787]: https://gravitee.atlassian.net/browse/APIM-13787?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-console-webui/src/entities/management-api-v2/createApi/createApiV4.ts
+++ b/gravitee-apim-console-webui/src/entities/management-api-v2/createApi/createApiV4.ts
@@ -33,4 +33,5 @@ export interface CreateApiV4 extends CreateBaseApi {
   analytics?: Analytics;
   flowExecution?: FlowExecution;
   flows?: FlowV4[];
+  allowedInApiProducts?: boolean;
 }
--- a/gravitee-apim-console-webui/src/management/api/creation-v4/api-creation-v4.component.ts
+++ b/gravitee-apim-console-webui/src/management/api/creation-v4/api-creation-v4.component.ts
@@ -205,6 +205,7 @@ export class ApiCreationV4Component implements OnInit, OnDestroy {
       description: apiCreationPayload.description ?? '',
       listeners: listeners,
       type: apiCreationPayload.type,
+      allowedInApiProducts: false,
       endpointGroups: apiCreationPayload.selectedEndpoints.map(
         endpoint =>
           ({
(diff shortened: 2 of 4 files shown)


paulatulis (approved):
One thing just as an FYI: since this component is used for *all* V4 API creations, we will now be sending `allowedInApiProducts: false` for Message and TCP Proxy APIs as well (even though it's technically an HTTP-Proxy-only field). However, this is perfectly harmless because the backend Jackson deserializer will safely parse it and ignore/coalesce it as `false`.

Looks great!

carlos-andres-osorio:
> One thing just as an FYI: since this component is used for *all* V4 API creations, we will now be sending `allowedInApiProducts: false` for Message and TCP Proxy APIs as well (even though it's technically an HTTP-Proxy-only field). However, this is perfectly harmless because the backend Jackson deserializer will safely parse it and ignore/coalesce it as `false`.
> 
> Looks great!

Good catch — fixed. The conditional spread now only sends `allowedInApiProducts: false` when `apiCreationPayload.type === 'PROXY'`, so Message and other v4 API types omit the field entirely. Extended the message spec with a test that asserts it's absent from the POST body.

PR #16643, opened 2026-04-28
fix(reporter): write security token instead of API ID in CSV column

carlos-andres-osorio:
## Issue

https://gravitee.atlassian.net/browse/APIM-13678

## Why

The CSV reporter exposes a security-token column intended to carry the raw token value for downstream correlation. In both the v2 and v4 `MetricsFormatter`, the column was incorrectly wired to the API ID when a security token was present, and empty otherwise — making the column useless for identifying which token authenticated a given request.

## What changed

- **v2 `MetricsFormatter`** — replaced `metrics.getSecurityToken() != null ? metrics.getApi() : null` with `metrics.getSecurityToken()` directly. `SingleValueFormatter.appendString` already emits an empty quoted field for `null`, so no conditional is needed.
- **v4 `MetricsFormatter`** — same fix: replaced `metrics.getSecurityToken() != null ? metrics.getApiId() : null` with `metrics.getSecurityToken()`.
- **Three existing v2 expected CSV fixtures** (`metrics.csv`, `metrics-with-additional.csv`, `metrics-with-invalid-remote-address.csv`) were corrected — all three had a non-null `securityToken` in their given fixtures and had been encoding the buggy API-ID value in the security-token column.
- **Two new fixture pairs** added to cover the previously untested paths: v2 null-token (empty column) and v4 non-null-token (raw token value).
- **`CsvFormatterTest`** extended with two new parameterized entries to wire the new fixture pairs.

## How to test

```
mvnd test -pl gravitee-apim-reporter/gravitee-apim-reporter-common -Dskip.validation=true
```

All 103 tests pass. The corrected fixture values can be verified by diffing the expected CSV files: the security-token column now contains `21b560b2-59b8-4a4b-921a-32b3731fdec4` (raw token) instead of `ff3c6c48-...` (API ID).

## Gotchas / Follow-up

- APIM-13679 (CSV column documentation) is a related follow-on ticket and is not blocked by this change.
- The `last` flag on the v2 `appendString` call was preserved unchanged — it controls the trailing separator and is independent of the value fix.


paulatulis (approved):
Optional nit before merge: Add a brief comment to `metrics-without-security-token.json` explaining that `customMetrics: {}` is intentional (to test the `last=true` branch on the `securityToken` field), so future maintainers don't assume it's an incomplete fixture.

carlos-andres-osorio:
> Optional nit before merge: Add a brief comment to `metrics-without-security-token.json` explaining that `customMetrics: {}` is intentional (to test the `last=true` branch on the `securityToken` field), so future maintainers don't assume it's an incomplete fixture.

Thanks for the nit. `customMetrics: {}` appears in every `given/` metrics fixture in the directory — it's the standard baseline, not something specific to this file. In `MetricsFormatter`, the `last` flag on the `securityToken` field resolves to `customMetrics.isEmpty() && additionalMetrics.isEmpty()`, so an empty map is simply the default state. The fixture that exercises the `last=false` path is `metrics-with-additional.json`. The filename already communicates the fixture's purpose precisely.

PR #16645, opened 2026-04-28
fix(helm): fallback to default to value instead of empty value

oussamaelkhayali-collab:
Use this shorter version:

## Issue

Some derived URLs in the Helm chart are used in generated config even when no route resource is rendered.

Before the HTTPRoute refactor, those URLs were still implicitly getting `apim.example.com` from the default ingress values in values.yaml, even when ingress was disabled.

After introducing the effective route helpers, we didnt want to have the same bug, so when both `ingress.enabled=false` and `httpRoute.enabled=false`, the helper returned an empty host list. That made some generated config values empty, which can break the resulting application configuration.

## Fix

Restore the previous fallback behavior explicitly in the effective route helpers.

When neither ingress nor HTTPRoute is enabled, use `apim.example.com` as the fallback host instead of returning an empty value.

This keeps generated config consistent with the previous ingress-only behavior, while leaving strict HTTPRoute validation unchanged.

## Validation

Updated tests to verify the fallback is now applied for generated UI and Portal URLs.


piequi (approved):

jgiovaresco (dismissed):

jgiovaresco (approved):
Use a conventional commit message: fix(helm):.....

PR #16646, opened 2026-04-28
fix(e2e): increase sed line range for ApplicationApi SDK post-processing

jaroslawlakomy-grav:
## What

This PR fixes e2e / portal SDK generation issues caused by two recent `portal-openapi.yaml` changes.

## Root cause analysis

Two separate commits exposed two different weaknesses in `gravitee-apim-e2e/scripts/update-portal-sdk.sh`:

- `7468f6f1f1` by `Jaroslaw Lakomy`
  `feat(portal-openapi): APIM-13760 define application invitation management contract`

  This change added new invitation-related endpoints and schemas, and also extended the portal contract around user/member search. As a result, the generated import section in `ApplicationApi.ts` grew and the existing `sed` workaround in `update-portal-sdk.sh` no longer matched because it was limited to the first 50 lines.

  What had to be fixed:
  the `ApplicationApi.ts` patch range needed to be extended so the existing workaround still applies after the generated code shift.

- `3aa2a4b8c6` by `aqibmohammadkhan`
  `feat(portal): add analytics computation endpoints and mappers`

  This change introduced new analytics schemas using patterns that are not handled correctly by the old `typescript-fetch` generator version used in e2e (`@openapitools/openapi-generator-cli@1.0.18-4.3.1`).

  Problematic cases:
  - `AnalyticsInterval` uses `oneOf(string, integer)`, which generated invalid helper calls such as `numberFromJSONTyped` / `stringFromJSONTyped`
  - `AnalyticsFilter`, `AnalyticsFacetBucket`, and `AnalyticsTimeSeriesBucket` use `oneOf + discriminator`, which generated TypeScript branches later resolved to `never`

  What had to be fixed:
  post-generation patches were added for the generated analytics model files so the SDK compiles correctly with the current generator.

## Fix

This PR updates `gravitee-apim-e2e/scripts/update-portal-sdk.sh` to handle both regressions:

- keep the `ApplicationApi.ts` workaround effective after the invitation/member contract changes
- patch broken generated analytics model code after SDK generation

## Validation

Validated locally with:

- `yarn --cwd gravitee-apim-e2e update:sdk:portal`
- `yarn --cwd gravitee-apim-e2e tsc --noEmit`
- `yarn --cwd gravitee-apim-e2e build`

All commands pass after this change.

### DIFF
--- a/gravitee-apim-e2e/scripts/update-portal-sdk.sh
+++ b/gravitee-apim-e2e/scripts/update-portal-sdk.sh
@@ -27,14 +27,26 @@ yarn dlx @openapitools/openapi-generator-cli@1.0.18-4.3.1 generate \
   --reserved-words-mappings=configuration=configuration
 
 sed -i.bak "s|export \* from './AnalyticsApi';||" lib/portal-webclient-sdk/src/lib/apis/index.ts
-sed -i.bak "1,30 s|DateHistoAnalytics \| GroupByAnalytics \| CountAnalytics|CountAnalytics, DateHistoAnalytics, GroupByAnalytics|" lib/portal-webclient-sdk/src/lib/apis/AnalyticsApi.ts
-sed -i.bak "1,30 s|CountAnalytics, DateHistoAnalytics, GroupByAnalyticsFromJSON,|GroupByAnalyticsFromJSON, CountAnalyticsFromJSON,|" lib/portal-webclient-sdk/src/lib/apis/AnalyticsApi.ts
-sed -i.bak "1,30 s|CountAnalytics, DateHistoAnalytics, GroupByAnalyticsToJSON,|GroupByAnalyticsToJSON,|" lib/portal-webclient-sdk/src/lib/apis/AnalyticsApi.ts
+sed -i.bak "1,50 s|DateHistoAnalytics \| GroupByAnalytics \| CountAnalytics|CountAnalytics, DateHistoAnalytics, GroupByAnalytics|" lib/portal-webclient-sdk/src/lib/apis/AnalyticsApi.ts
+sed -i.bak "1,50 s|CountAnalytics, DateHistoAnalytics, GroupByAnalyticsFromJSON,|GroupByAnalyticsFromJSON, CountAnalyticsFromJSON,|" lib/portal-webclient-sdk/src/lib/apis/AnalyticsApi.ts
+sed -i.bak "1,50 s|CountAnalytics, DateHistoAnalytics, GroupByAnalyticsToJSON,|GroupByAnalyticsToJSON,|" lib/portal-webclient-sdk/src/lib/apis/AnalyticsApi.ts
 sed -i.bak "s|DateHistoAnalytics \| GroupByAnalytics \| CountAnalyticsFromJSON(|CountAnalyticsFromJSON(|" lib/portal-webclient-sdk/src/lib/apis/AnalyticsApi.ts
-sed -i.bak "1,50 s|DateHistoAnalytics \| GroupByAnalytics \| CountAnalytics|CountAnalytics, DateHistoAnalytics, GroupByAnalytics|" lib/portal-webclient-sdk/src/lib/apis/ApplicationApi.ts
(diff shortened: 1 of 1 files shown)


gravitee-apim-e2e/scripts/update-portal-sdk.sh
```diff
@@ -27,15 +27,27 @@ yarn dlx @openapitools/openapi-generator-cli@1.0.18-4.3.1 generate \
...
+sed -i.bak "s|return {...AnalyticsFacetBucketLeafFromJSONTyped(json, true), type: 'LEAF'};|return {...AnalyticsFacetBucketLeafFromJSONTyped(json, true), type: 'LEAF'} as AnalyticsFacetBucket;|" lib/portal-webclient-sdk/src/lib/models/AnalyticsFacetBucket.ts
+sed -i.bak "s|return {...AnalyticsFacetBucketGroupFromJSONTyped(json, true), type: 'GROUP'};|return {...AnalyticsFacetBucketGroupFromJSONTyped(json, true), type: 'GROUP'} as AnalyticsFacetBucket;|" lib/portal-webclient-sdk/src/lib/models/AnalyticsFacetBucket.ts
+sed -i.bak "s|return {...AnalyticsTimeSeriesBucketLeafFromJSONTyped(json, true), type: 'LEAF'};|return {...AnalyticsTimeSeriesBucketLeafFromJSONTyped(json, true), type: 'LEAF'} as AnalyticsTimeSeriesBucket;|" lib/portal-webclient-sdk/src/lib/models/AnalyticsTimeSeriesBucket.ts
+sed -i.bak "s|return {...AnalyticsTimeSeriesBucketGroupFromJSONTyped(json, true), type: 'GROUP'};|return {...AnalyticsTimeSeriesBucketGroupFromJSONTyped(json, true), type: 'GROUP'} as AnalyticsTimeSeriesBucket;|" lib/portal-webclient-sdk/src/lib/models/AnalyticsTimeSeriesBucket.ts
```
kovaren, line 49: These fixes are masking an issue that our customers may run into when generating an OpenAPI client from our spec.

Is there a way to fix it directly in the spec? https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml

kovaren (changes requested):

gravitee-apim-e2e/scripts/update-portal-sdk.sh
jaroslawlakomy-grav: Zrzut ekranu 2026-04-28 o 21 20 29

I think that our customers are required to do some workarounds since 2022 :) but yes, section which I had to add to make it working makes the whole thing even harder for them

@aqibmohammadkhan will take a look on that as it's related to Analitics part. 

I agree that the best solution will be updating portal-openapi.yaml to avoid patterns leading to code generation issues

jaroslawlakomy-grav: Maybe we should consider reverting portal-openapi changes related to analytics? in this case fix to  update-portal-sdk.sh would be really simple - just more lines to be scanned by sed. What do you think?

aqibmohammadkhan: The root cause is OpenAPI Generator 4.3.1 is buggy when handling `oneOf` with `discriminator`. Even with perfect schema, it generates redundant intersections.

kovaren: Ok, let's keep the sed fix for now and track the generator upgrade as a technical task.

kovaren (approved):

gravitee-apim-e2e/scripts/update-portal-sdk.sh
aqibmohammadkhan: raised a tech tesk for upgrade https://gravitee.atlassian.net/browse/APIM-13834

PR #16655, opened 2026-04-28
feat (console): open context menu with right click in portal navigation items tree.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-13827

## Description

The action menu was previously only accessible through the "more" button. This change wires a Material context menu to the right-click event so users can reach the same actions via a context menu.

## Changes

- Added a `(contextmenu)` binding on both `mat-tree-node` rows (leaf and nested) that calls a new `onContextMenu(event, node)` handler.
- Extracted the existing menu items (Add Page/API/Folder/Link, Edit, Publish/Unpublish, Delete, with their permission, tooltip and disabled-state logic) into a single `#menuItems` ng-template.
- A hidden `button` with `matMenuTriggerFor` acts as the anchor for the right-click menu.

## Screenshots / video

https://github.com/user-attachments/assets/5f6511c2-697e-4072-908d-21fb91b2d18c


aqibmohammadkhan (approved):
LGTM

PR #16667, opened 2026-04-29
feat: add revoke API key to subscriptions

MateuszJozwiak-Incubly:
## Issue

https://gravitee.atlassian.net/browse/EXT-120

## Description

- changed the way of displaying API Keys for subscriptions: table of all API Keys instead of latest one
- added button to revoke the active api key next to each active API key on the list
- added a modal to be displayed after clicking the button to confirm the action
- made the API access section hidden when no active API key is present

## Additional context

- when revoking an API key the page should be updated with the current status of the keys

## Testing scenarios

- create API several API Keys in the console and ensure all of them are visible in the new portal, only the active key has the revoke button, all inactive keys are displayed with revoked/expired at date value
- ensure revoking the API key works and the API key status is updated in console and new portal
- when there are more than one active API Keys, ensure all of them have option to be revoked

image

- ensure that when no API key is active, the 'Calling the API' section' is invisible 
image

- ensure that the 'Calling to API' section uses only the latest active API key:
image

Video of the functionality:

https://github.com/user-attachments/assets/09d2b2b1-4883-4502-833f-303181ed404d


gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -66,16 +92,118 @@ export class ApiAccessComponent {
...
+    this.isRevokingApiKey = true;
+    this.subscriptionKeysService
+      .revoke(subscriptionId, apiKey)
+      .pipe(finalize(() => (this.isRevokingApiKey = false)))
+      .subscribe({
+        next: () => this.revokeApiKey.emit(),
+        error: () => {
+          // no-op: keep UI stable and release loading state in finalize
+        },
+      });
```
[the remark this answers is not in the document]
aqibmohammadkhan: if the backend rejects revoke (e.g. 403, 409, 5xx), the user sees the dialog close, the spinner clear, and… nothing. They have no idea whether the key is gone. The rest of the portal-next codebase uses MatSnackBar for this kind of feedback — emit a localized error message, e.g.:

```
error: () => this.snackBar.open(
  $localize`:@@subscriptionDetailsApiAccessRevokeError:Failed to revoke API key. Please try again.`,
  $localize`:@@close:Close`,
  { duration: 5000 },
),
```

A success snackbar ("API key revoked") would also be a nice touch.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.html
```diff
@@ -32,38 +34,82 @@
...
+                      mat-icon-button
+                      type="button"
+                      color="warn"
+                      [disabled]="isRevokingApiKey"
+                      matTooltip="Revoke"
+                      [attr.aria-label]="'Revoke API key ' + apiKey.key"
+                      [attr.data-api-key]="apiKey.key"
+                      data-testid="api-key-revoke-button"
+                      (click)="revokeApiKeyRow(apiKey)"
+                    >
```
aqibmohammadkhan, line 78: Every other user-facing string in this template uses i18n="@@…"/$localize. The tooltip and aria-label here are bare English. Should be:

```
<button
  ...
  i18n-matTooltip="@@subscriptionDetailsApiAccessRevokeTooltip"
  matTooltip="Revoke"
  [attr.aria-label]="revokeAriaLabel(apiKey.key)"
  ...
>
```

with a method or a `$localize `template that parameterises the key. Same applies to the empty/No API keys available.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -66,16 +126,144 @@ export class ApiAccessComponent {
...
+
+  protected onApiKeysPageSizeChange(pageSize: number): void {
+    this.apiKeysPageSize = pageSize;
+    this.apiKeysCurrentPage = 1;
+    this.updateDisplayedApiKeyRows();
+  }
+
+  protected get primaryApiKey(): string {
+    return this.activeApiKey?.key ?? this._apiKeys[0]?.key ?? '';
+  }
```
aqibmohammadkhan: If there are only revoked/expired keys, this returns one of them. Today it's masked because `shouldShowApiAccessContent` hides the curl section when there's no active key, and the kafka access component is fed `primaryApiKey` only when ACCEPTED — but if any future caller (or kafka access on an expired-but-pending state) reads `primaryApiKey`, they get a dead key. Make the contract explicit:

```
protected get primaryApiKey(): string {
  return this.activeApiKey?.key ?? '';
}
```

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -66,16 +126,144 @@ export class ApiAccessComponent {
...
+
+  protected get resolvedApiKeyConfigUsername(): string {
+    return this.apiKeyConfigUsername ?? this.activeApiKey?.hash ?? this._apiKeys[0]?.hash ?? '';
+  }
```
aqibmohammadkhan: The same (just above comment) applies here `resolvedApiKeyConfigUsername`, which currently falls back to `_apiKeys[0]?.hash` even when that key is dead.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -66,16 +126,144 @@ export class ApiAccessComponent {
...
+
+  private registerApiKeyStatusIcons(): void {
+    this.matIconRegistry.addSvgIconLiteralInNamespace(
+      'gio',
+      'check-circled-outline',
+      this.sanitizer.bypassSecurityTrustHtml(GIO_CHECK_CIRCLED_OUTLINE_ICON), //NOSONAR
+    );
+    this.matIconRegistry.addSvgIconLiteralInNamespace('gio', 'x-circle', this.sanitizer.bypassSecurityTrustHtml(GIO_X_CIRCLE_ICON)); //NOSONAR
+  }
```
aqibmohammadkhan: Two problems:

- The  global icon registry already exists in the portal-next app, you're duplicating SVG payloads in this component bundle and shadowing/overwriting global registrations. If not, the right place is an `APP_INITIALIZER` / `provideAppInitializer` so it's registered once at bootstrap.

- Inlining the SVG markup as TypeScript template strings hurts readability and makes designers unable to tweak the assets. Prefer `addSvgIconInNamespace('gio', 'x-circle', sanitizer.bypassSecurityTrustResourceUrl('assets/icons/gio/x-circle.svg'))` with the SVG checked in as an asset, matching the convention already exists.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -66,16 +126,144 @@ export class ApiAccessComponent {
...
+  curlCmd = computed(() => this.formatCurlCommandLine(this.selectedEntrypointUrl(), this.planSecurity, this.primaryApiKey));
+
+  constructor(
+    private readonly configService: ConfigService,
+    private readonly subscriptionKeysService: SubscriptionKeysService,
+    private readonly dialog: MatDialog,
+    private readonly matIconRegistry: MatIconRegistry,
+    private readonly sanitizer: DomSanitizer,
+  ) {
+    this.registerApiKeyStatusIcons();
```
aqibmohammadkhan: Per-instance icon registration is the wrong place - check [here](https://github.com/gravitee-io/gravitee-api-management/pull/16667/changes#r3187610620)

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -49,7 +90,26 @@ export class ApiAccessComponent {
...
+          : $localize`:@@subscriptionDetailsApiAccessApiKeysStatusInactive:Inactive API key`,
+        isActive,
+        key: apiKey.key ?? '',
+        createdAt: apiKey.created_at,
+        revokedAt: apiKey.revoked_at ?? apiKey.expire_at,
+      };
+    });
+    this.hasActiveApiKey = !!this.activeApiKey;
+    this.totalApiKeyElements = this.apiKeyRows.length;
+    this.apiKeysCurrentPage = 1;
```
aqibmohammadkhan: `apiKeysCurrentPage` reset is a UX trap.
Because revoke success calls `revokeApiKey.emit() → parent reload → new apiKeys input → setter runs → page resets to 1`. If the operator was on page 3 of 10 keys revoking an old one, they get teleported back to page 1. Either keep the page (clamping to the new total) or reset only when the length shrinks past the current page boundary:

```
const newTotal = this.apiKeyRows.length;
const lastValidPage = Math.max(1, Math.ceil(newTotal / this.apiKeysPageSize));
this.apiKeysCurrentPage = Math.min(this.apiKeysCurrentPage, lastValidPage);
```

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -66,16 +126,144 @@ export class ApiAccessComponent {
...
+  ];
+  protected readonly apiKeysPageSizeOptions = [5, 10, 25];
+
+  protected apiKeyRows: ApiKeyTableRow[] = [];
+  protected displayedApiKeyRows: ApiKeyTableRow[] = [];
+  protected totalApiKeyElements = 0;
+  protected apiKeysCurrentPage = 1;
+  protected apiKeysPageSize = 5;
+  protected isRevokingApiKey = false;
+  protected hasActiveApiKey = false;
```
aqibmohammadkhan: `hasActiveApiKey` is a stored field updated in the setter, while `primaryApiKey`, `resolvedApiKeyConfigUsername`, and `shouldShowApiAccessContent` are getters. There's no reason `hasActiveApiKey` shouldn't also be a getter (`return !!this.activeApiKey`). Also consider replacing the setter-with-side-effects pattern with a computed() chain — the rest of the file already mixes signals (model, computed) with classic inputs, so a signal<SubscriptionDataKeys[]> plus computed derivations would be more idiomatic.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.spec.ts
```diff
@@ -245,48 +610,82 @@ describe('ApiAccessComponent', () => {
...
+  }
+  async function getRevokeApiKeyButton() {
+    return await harnessLoader.getHarnessOrNull(MatButtonHarness.with({ selector: '[data-testid="api-key-revoke-button"]' }));
+  }
+  async function revokeApiKeyButtonShown() {
+    return !!(await getRevokeApiKeyButton());
+  }
+  async function headerRevokeApiKeyButtonShown() {
+    return !!(await harnessLoader.getHarnessOrNull(MatButtonHarness.with({ selector: '[data-testid="revoke-api-key-button"]' })));
+  }
```
aqibmohammadkhan: `data-testid="revoke-api-key-button"` doesn't exist in the template; only `api-key-revoke-button` does. The helper and its single assertion (`expect(await headerRevokeApiKeyButtonShown()).toBeFalsy();`) can be removed.


Same story for the `api-access__header-row flex container` — it has one child, so `display:flex; justify-content:space-between` does nothing. Either drop the wrapper or keep it intentionally for a future header action.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.spec.ts
```diff
@@ -84,23 +103,364 @@ describe('ApiAccessComponent', () => {
...
+          component.planSecurity = 'API_KEY';
+          component.subscription = { id: 'subscription-id', status: 'ACCEPTED' } as Subscription;
+          component.apiKeys = [{ key: 'api-key', application: { id: 'app-id', name: 'app-name' } }];
+          component.canManageSubscription = false;
+
+          fixture.detectChanges();
+          (component as unknown as { revokeApiKeyRow: (apiKey: { isActive: boolean; key: string }) => void }).revokeApiKeyRow({
+            isActive: true,
+            key: 'api-key',
+          });
```
aqibmohammadkhan: This casts away protected to test a guard. The two tests that do this (`should not call revoke service when canManageSubscription is false`, `should not call revoke service when api key is missing`) can be exercised through the UI: the button is hidden when `!canManageSubscription`, and is hidden when apiKeys is empty — both are already asserted. The cast-based tests duplicate weaker checks. Either drop them or assert behaviour through public API only.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -49,7 +90,26 @@ export class ApiAccessComponent {
...
+      const isActive = this.isActiveApiKey(apiKey);
+      return {
+        statusIcon: isActive ? 'gio:check-circled-outline' : 'gio:x-circle',
+        statusLabel: isActive
+          ? $localize`:@@subscriptionDetailsApiAccessApiKeysStatusActive:Active API key`
+          : $localize`:@@subscriptionDetailsApiAccessApiKeysStatusInactive:Inactive API key`,
+        isActive,
+        key: apiKey.key ?? '',
+        createdAt: apiKey.created_at,
+        revokedAt: apiKey.revoked_at ?? apiKey.expire_at,
```
aqibmohammadkhan: Column label is "Revoked/Expired At". Renaming the row field to `closedAt` (matching the column intent) avoids future surprises where someone reads `row.revokedAt` and assumes it's strictly a revoke timestamp.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.html
```diff
@@ -32,38 +34,82 @@
...
+                    data-testid="api-key-status-icon"
+                  />
+                </ng-template>
+                <ng-template appTableCell="revoke" let-apiKey>
+                  @if (canManageSubscription && apiKey.isActive) {
+                    <button
+                      mat-icon-button
+                      type="button"
+                      color="warn"
+                      [disabled]="isRevokingApiKey"
```
aqibmohammadkhan: `isRevokingApiKey` only flips after the user confirms the dialog. While a dialog is open but unconfirmed, the in-table buttons are still enabled. A user could open multiple confirm dialogs by clicking different rows quickly. Low-impact (Material renders them stacked), but a one-liner fix is to flip a `dialogOpen` flag (or include the dialog-open phase in `isRevokingApiKey`).

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.spec.ts
```diff
@@ -84,23 +103,364 @@ describe('ApiAccessComponent', () => {
...
+      expect(apiAccessContentShown()).toBeTruthy();
       expect(await commandLineShown()).toBeTruthy();
     });
   });
 
   describe('Accepted', () => {
     describe('HTTP API', () => {
       describe('API Key', () => {
-        it('should show api key, base url and command line', async () => {
+        it('should show api keys and hide calling api content when all api keys are inactive', async () => {
```
aqibmohammadkhan, line 111: `apiAccessContentShown()` queries `.api-access__copy-code-content`, which contains the keys table, so it's truthy by construction. The intent ("calling-api hidden") is actually verified by `baseUrlShown`/`commandLineShown` being `falsy`. 

Consider renaming the helper to something like `apiAccessShellShown` to avoid confusion with the calling-api block.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -66,16 +126,144 @@ export class ApiAccessComponent {
   @Input()
   clientSecret?: string;
 
+  @Input()
+  canManageSubscription = false;
+
+  @Output()
+  revokeApiKey = new EventEmitter<void>();
```
aqibmohammadkhan: It's an "after revoke succeeded" notification, not a request. `apiKeyRevoked` reads more accurately for an EventEmitter<void>

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.spec.ts
```diff
@@ -84,23 +103,364 @@ describe('ApiAccessComponent', () => {
...
+
+          await revokeButton!.click();
+          const confirmDialog = await rootLoader.getHarness(ConfirmDialogHarness);
+          await confirmDialog.confirm();
+          fixture.detectChanges();
+
+          const revokeButtonWhileLoading = await getRevokeApiKeyButton();
+          expect(await revokeButtonWhileLoading!.isDisabled()).toBeTruthy();
+
+          revokeSubject.next();
```
aqibmohammadkhan: this test relies on Subject + manual next/complete. Good test, but revokeSubject.complete() alone is enough — the explicit next() is unused.

gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.ts
```diff
@@ -181,17 +188,18 @@ export class SubscriptionsDetailsComponent implements OnInit {
...
         forkJoin({
           subscription: of(subscription),
           plan: this.getPlanData$(subscription.plan),
-          api: hasDocumentationAccess ? this.apiService.details(this.apiId).pipe(catchError(() => of(null))) : of(null),
+          api: apiPermissions ? this.apiService.details(this.apiId).pipe(catchError(() => of(null))) : of(null),
           application: this.applicationService.get(subscription.application),
-          hasDocumentationAccess: of(hasDocumentationAccess),
-          documentationNavigationTarget: hasDocumentationAccess
+          hasDocumentationAccess: of(!!apiPermissions),
+          canManageSubscription: of(!!(apiPermissions?.SUBSCRIPTION?.includes('U') || apiPermissions?.SUBSCRIPTION?.includes('D'))),
```
aqibmohammadkhan: if you check `SubscriptionKeysResource.revokeKeySubscription` (gravitee-apim-rest-api-portal-rest),  the gateway accepts `UPDATE` on either `APPLICATION_SUBSCRIPTION` or `API_SUBSCRIPTION`. `DELETE` is not accepted (that's what `_close` uses). And the same pattern is already applied to `_renew` right above it, so it's a stable convention.

This change here
```
canManageSubscription: of(!!(apiPermissions?.SUBSCRIPTION?.includes('U') || apiPermissions?.SUBSCRIPTION?.includes('D'))),
```
where `apiPermissions` comes only from `permissionsService.getApiPermissions(this.apiId)`. that diverges from the backend.

the developer portal is consumer-facing, and an application owner (who has `APPLICATION_SUBSCRIPTION:U` on their own app and typically no `API_SUBSCRIPTION` permission at all) is precisely the user we'd expect to revoke their own keys. With this PR they will never see the button — `getApiPermissions(apiId)` returns 404 for them and the existing catchError collapses it to undefined, making `canManageSubscription` permanently false. `permissionsService.getApplicationPermissions(applicationId)` already exists and returns `UserApplicationPermissions.SUBSCRIPTION`, so wiring it in is a small change.


Suggestions

- Drop 'D'.
- OR application permissions in. Something like:
```
private getRevokePermissionFlags$(applicationId: string, apiId: string) {
  return forkJoin({
    api: this.permissionsService.getApiPermissions(apiId).pipe(
      catchError(err => (err instanceof HttpErrorResponse && err.status === 404 ? of(undefined) : throwError(() => err))),
    ),
    application: this.permissionsService.getApplicationPermissions(applicationId).pipe(
      catchError(err => (err instanceof HttpErrorResponse && err.status === 404 ? of(undefined) : throwError(() => err))),
    ),
  }).pipe(
    map(({ api, application }) => ({
      apiPermissions: api,
      canRevokeApiKey:
        !!api?.SUBSCRIPTION?.includes('U') ||
        !!application?.SUBSCRIPTION?.includes('U'),
    })),
  );
}
```

gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.html
```diff
@@ -40,16 +40,18 @@
         (resumeConsumerStatus)="resumeConsumerStatus()"
         (closeSubscription)="closeSubscription()"
       />
-      @if (detailsVM.result.api) {
+      @if (detailsVM.result.api || detailsVM.result.canManageSubscription) {
```
aqibmohammadkhan: For an application owner with no API documentation/API permission, api is undefined but `canManageSubscription` is true. That means `ApiAccessComponent` renders with `entrypointUrls: []`.

For an active API key subscription, `shouldShowApiAccessContent` is still true, so the UI can show a Calling the API section with an empty curl command. For OAuth/JWT, the broad `canManageSubscription` condition can also expose the API access card even though revoke is only relevant to API keys.

Suggested fix: only use the permission fallback for API-key subscriptions, and/or gate “Calling the API” on having entrypoints:

`@if (detailsVM.result.api || (detailsVM.result.planSecurity === 'API_KEY' && detailsVM.result.canManageSubscription)) {`

Also rename `canManageSubscription` to `canRevokeApiKey`, the current broad name contributed to this issue.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -13,30 +13,57 @@
...
+import { MatDialog } from '@angular/material/dialog';
 import { MatFormFieldModule } from '@angular/material/form-field';
+import { MatIcon } from '@angular/material/icon';
 import { MatSelectModule } from '@angular/material/select';
+import { MatTooltip } from '@angular/material/tooltip';
+import { filter, tap } from 'rxjs';
+import { finalize } from 'rxjs/operators';
 
 import { NativeKafkaApiAccessComponent } from './native-kafka-api-access/native-kafka-api-access.component';
+import { SnackBarService } from '../../../../gravitee-apim-console-webui/src/services-ngx/snack-bar.service';
```
aqibmohammadkhan: I’d avoid cross-app source imports in `portal-webui-next`,  use `MatSnackBar` directly or add a local portal-next snackbar wrapper.

gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.html
MateuszJozwiak-Incubly: Renamed to canManageApiKey to have in mind the renew key option in the next PR.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -13,70 +13,312 @@
...
 
-  constructor(private configService: ConfigService) {
     effect(() => {
-      this.selectedEntrypointUrl.set(this.entrypointUrls?.length ? this.entrypointUrls[0] : this.selectedEntrypointUrl());
+      const lastValidPage = Math.max(1, Math.ceil(this.totalApiKeyElements() / this.apiKeysPageSize()));
+      if (this.apiKeysCurrentPage() > lastValidPage) {
+        this.apiKeysCurrentPage.set(lastValidPage);
+      }
     });
   }
```
aqibmohammadkhan: writing to signals inside `effect()` throws a runtime error in dev mode by default. The second effect is the critical one. Both need `{ allowSignalWrites: true }` passed as the second argument, or the page-boundary logic should be restructured as a `computed()`.

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -13,70 +13,312 @@
...
 
-  curlCmd = computed(() => this.formatCurlCommandLine(this.selectedEntrypointUrl(), this.planSecurity, this.apiKey));
+  curlCmd = computed(() => this.formatCurlCommandLine(this.selectedEntrypointUrl(), this.planSecurityInput(), this.primaryApiKey()));
+
+  constructor(
+    private readonly configService: ConfigService,
+    private readonly subscriptionKeysService: SubscriptionKeysService,
+    private readonly snackBar: MatSnackBar,
+    private readonly dialog: MatDialog,
+  ) {
```
aqibmohammadkhan: Avoid constructor injection. Use something like

```
private readonly configService = inject(ConfigService);
private readonly subscriptionKeysService = inject(SubscriptionKeysService);
// etc.
```

gravitee-apim-portal-webui-next/src/services/subscription-keys.service.ts
```diff
@@ -0,0 +1,34 @@
...
+import { ConfigService } from './config.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SubscriptionKeysService {
+  constructor(
+    private readonly http: HttpClient,
+    private readonly configService: ConfigService,
+  ) {}
```
aqibmohammadkhan: same as [this](https://github.com/gravitee-io/gravitee-api-management/pull/16667/changes#r3225597556)

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.ts
```diff
@@ -13,70 +13,312 @@
...
+    this.canManageApiKeyInput.set(value);
+  }
+
+  get canManageApiKey(): boolean {
+    return this.canManageApiKeyInput();
+  }
+
+  @Output()
+  apiKeyRevoked = new EventEmitter<void>();
+
```
aqibmohammadkhan: output + event emitter should be output<void>

`apiKeyRevoked = output<void>();`

gravitee-apim-portal-webui-next/src/components/api-access/api-access.component.spec.ts
```diff
@@ -51,26 +60,37 @@ describe('ApiAccessComponent', () => {
...
+      imports: [ApiAccessComponent, AppTestingModule, MatIconTestingModule],
       providers: [
         {
           provide: ConfigService,
           useClass: CustomConfigurationServiceStub,
         },
+        {
+          provide: SubscriptionKeysService,
+          useValue: subscriptionKeysServiceMock,
+        },
```
aqibmohammadkhan: When a component uses services that call the backend, use `HttpTestingController` rather than mocking them." The `SubscriptionKeysService.revoke()` makes an HTTP POST. The spec for the service itself (`subscription-keys.service.spec.ts`) does use `HttpTestingController` correctly — but the component spec should also go through the real service + HTTP layer instead of mocking the service entirely.

gravitee-apim-portal-webui-next/src/app/api/api-details/api-tab-subscriptions/subscriptions-details/subscriptions-details.component.spec.ts
```diff
@@ -225,6 +250,84 @@ describe('SubscriptionsDetailsComponent', () => {
     });
   });
 
+  describe('API key revoke permissions', () => {
```
aqibmohammadkhan, line 274: this `describe` block tests `API_KEY` plan scenarios thoroughly. However, there is no test for: `OAUTH2/JWT plan + user has application permissions but no API permissions` (so api is null). In that scenario, the existing template guard `@if (detailsVM.result.api || (detailsVM.result.planSecurity === 'API_KEY' && detailsVM.result.canManageApiKey))` resolves to `false`, and the user would not see their `clientId/clientSecret`. Worth adding a test to at least document the current behavior.

aqibmohammadkhan (approved):

PR #16681, opened 2026-04-30
feat(gamma): scope module routes to environment

gcusnieux:
Add `/environments/{envId}` to the gamma module path so `GraviteeContext.getCurrentEnvironment()` is populated for module resources (via existing `GraviteeContextFilter`).

Path: `/gamma/organizations/{orgId}/modules/...` → `/gamma/organizations/{orgId}/environments/{envId}/modules/...`

Draft for workshop discussion.


gcusnieux:
Closing — wrong layer. Per discussion with Julien (and prior alignment with Jeoffrey), env scoping should live inside each module's root resource (`/modules/{pluginId}/environments/{envId}/...`), not at the GammaModulesResource level. This keeps module discovery and module-level org settings under `/organizations/{orgId}/modules` and lets each module mix org-scoped and env-scoped resources cleanly. Will apply the change in the AIM module instead.

PR #16697, opened 2026-05-01
fix: jwt clientIdClaim

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-13867
## Description

PROBLEM

When a V3 API has multiple JWT plans and `clientIdClaim` is configured as a dot-notation path (e.g. act.repository), plan routing silently fails.

`JwtPlanBasedAuthenticationHandler.getClientId()` was using a plain `Map.get(name)` — a flat lookup that looks for a key literally named "act.repository". When the JWT payload is:

  `{ "act": { "repository": "my-org/my-repo" } }`

the flat lookup returns `null`, `preCheckSubscription` returns `false,` and the gateway cannot route the request to the correct plan.

Single-plan V3 APIs and all V4 APIs are unaffected — single-plan V3 bypasses the pre-check entirely, and V4 uses a different code path.

FIX

Replace the flat `claims.get()` call with a `resolveNestedClaim()` private helper using the same flat-first / nested-walk algorithm introduced in the JWT policy (ClaimPathResolver) as part of APIM-13714:

  1. Flat lookup first — if a key with that exact name exists, return it.
     Preserves backward compatibility for claim names that literally contain a dot.
  2. Nested walk — if the flat lookup misses and the name contains a dot,
     split on '.' and traverse successive Map values.

No other code paths are changed. The azp / aud / client_id fallback chain is untouched.

TESTS

4 new unit tests added to `JwtPlanBasedAuthenticationHandlerTest`

NOTES

The class already contains two "FIXME: This is duplicated from JWT policy" comments.
Extracting this logic to a shared gateway-api module is the right long-term fix and is tracked separately — it is out of scope here.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/test/java/io/gravitee/gateway/handlers/api/security/JwtPlanBasedAuthenticationHandlerTest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/test/java/io/gravitee/gateway/handlers/api/security/JwtPlanBasedAuthenticationHandlerTest.java
@@ -29,6 +29,7 @@ import io.gravitee.reporter.api.http.Metrics;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import net.minidev.json.JSONObject;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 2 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/test/java/io/gravitee/gateway/handlers/api/security/JwtPlanBasedAuthenticationHandlerTest.java
```diff
@@ -123,6 +124,56 @@ public void getClientId_should_return_null_cause_no_clientId_found() {
...
+    }
+
+    @Test
+    public void getClientId_should_return_null_when_intermediate_is_not_a_map() {
+        when(plan.getSecurityDefinition()).thenReturn("{\"clientIdClaim\": \"sub.nested\"}");
+
+        // "sub" is a String, not a nested object — walk must abort gracefully
+        Map<String, Object> claims = Map.of("sub", "plain-string-value");
+
+        assertNull(authenticationHandler.getClientId(claims));
```
carlos-andres-osorio, line 174: The four new tests cover nested hit, flat-with-literal-dot, missing segment, and non-Map segment — but the AC also requires: _"Given a null or empty `clientIdClaim`, `getClientId()` returns null without throwing an exception."_

The null case is handled by the pre-existing `getClientId_should_return_null_cause_no_clientId_found` test. The empty-string case has no coverage.

When `clientIdClaim` is configured as `""`, `getCustomClientIdClaimName()` returns `""` (non-null — `textValue()` on an empty JSON string is `""`, not `null`). Execution enters the custom-claim branch, calls `resolveNestedClaim(claims, "")`, and returns `null` — but the `azp`/`aud`/`client_id` fallback chain is bypassed entirely. The behavior is correct per the AC, but it is untested.

paulatulis: Addressed with     
```
public void getClientId_should_return_null_when_clientIdClaim_is_empty_string() {

```

carlos-andres-osorio (approved):
Incremental review — prior finding resolved (empty-string clientIdClaim now tested). LGTM.

PR #16703, opened 2026-05-01
feat: add span attribute redaction for APIs

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-13568

## Description

Screenshot 2026-05-01 at 12 51 02 PM
Screenshot 2026-05-01 at 12 51 10 PM
Screenshot 2026-05-01 at 12 51 15 PM
Screenshot 2026-05-01 at 12 53 10 PM

Test:

https://github.com/user-attachments/assets/1523ff7d-7c9d-422a-a1fc-d07dfa5d951d

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/reporter-settings/reporter-settings-proxy/reporter-settings-proxy.component.ts
+++ b/gravitee-apim-console-webui/src/management/api/reporter-settings/reporter-settings-proxy/reporter-settings-proxy.component.ts
@@ -13,10 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { Component, DestroyRef, inject, input, InputSignal, OnInit } from '@angular/core';
+import { Component, computed, DestroyRef, inject, input, InputSignal, OnInit, signal } from '@angular/core';
 import { FormsModule, ReactiveFormsModule, UntypedFormControl, UntypedFormGroup } from '@angular/forms';
 import { EMPTY, merge } from 'rxjs';
-import { catchError, switchMap, tap } from 'rxjs/operators';
+import { catchError, map, switchMap, tap } from 'rxjs/operators';
 import { ActivatedRoute } from '@angular/router';
 import { CommonModule } from '@angular/common';
 import { MatButtonModule } from '@angular/material/button';
@@ -29,8 +29,10 @@ import { GioBannerModule, GioFormSlideToggleModule, GioIconsModule, GioSaveBarMo
 import { MatSlideToggle } from '@angular/material/slide-toggle';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 
+import { ApiRedactionRulesComponent } from './api-redaction-rules/api-redaction-rules.component';
+
 import { ApiV2Service } from '../../../../services-ngx/api-v2.service';
-import { ApiV4 } from '../../../../entities/management-api-v2';
+import { ApiV4, RedactionRule } from '../../../../entities/management-api-v2';
 import { SnackBarService } from '../../../../services-ngx/snack-bar.service';
 
 type DefaultConfiguration = {
@@ -65,6 +67,7 @@ type DefaultConfiguration = {
     GioSaveBarModule,
     GioFormSlideToggleModule,
     MatSlideToggle,
+    ApiRedactionRulesComponent,
   ],
 })
 export class ReporterSettingsProxyComponent implements OnInit {
@@ -72,6 +75,11 @@ export class ReporterSettingsProxyComponent implements OnInit {
   form: UntypedFormGroup;
   defaultConfiguration: DefaultConfiguration;
   private destroyRef = inject(DestroyRef);
+
+  protected pendingRedactionRules = signal<RedactionRule[] | null>(null);
+  protected resetTriggerCounter = signal(0);
+  protected savedApiState = signal<ApiV4 | null>(null);
+  protected effectiveApi = computed(() => this.savedApiState() ?? this.api());
 
   constructor(
     private readonly activatedRoute: ActivatedRoute,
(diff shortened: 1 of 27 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/test/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactoryTest.java
```diff
@@ -399,6 +409,155 @@ private ConfigurablePluginManager<?> registerPolicyPluginManager() {
         }
     }
 
+    @Nested
```
araffass, line 413: The `BuildRedactionConfig` nested class here and the one in `TcpApiReactorFactoryTest` are like identical, same helpers, same assertions, both testing the same static method on `TracingRedactionMapper` which has no dependency on either factory. Would it make sense to move all of this into a dedicated `TracingRedactionMapperTest`? The factory tests could then just check that the result gets passed through to `createTracer` correctly.

gravitee-apim-console-webui/src/management/api/reporter-settings/reporter-settings-proxy/reporter-settings-proxy.component.ts
```diff
@@ -199,6 +230,7 @@ export class ReporterSettingsProxyComponent implements OnInit {
             this.form.get('headers').disable();
             this.form.get('payload').disable();
             this.form.get('condition').disable();
+            this.form.get('tracingEnabled').setValue(false);
```
araffass: This `setValue(false)` changes the existing behavior a bit. Before, disabling the control preserved its value, so toggling analytics off then back on would restore the tracing toggle to where it was. Now it gets reset to `false`, meaning if tracing was enabled and the user toggles analytics off then on without saving, tracing ends up silently disabled. Not sure if that's intentional? Also don't see a test covering that specific toggle flow.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
```diff
@@ -4938,6 +4938,65 @@ components:
...
+                valuePattern:
+                    type: string
+                    description: >
+                        Optional Java regex (partial match). Rule only fires when the attribute
+                        value matches.
+                    example: "^Bearer "
+        TracingMaskingStrategy:
+            type: object
+            properties:
+                type:
```
araffass, line 4980: Looks like `type` is not marked as required here. A client could send a `maskingStrategy` block without it, pass schema validation, and silently fall back to `MaskingStrategy.DEFAULT` in the mapper. Given this is specifically about masking sensitive data, that silent fallback feels a bit risky since the user would think their rule is active when it's not. Worth making `type` required?

gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/v4/analytics/tracing/TracingRedactionConfig.java
```diff
@@ -0,0 +1,41 @@
...
+@ToString
+@EqualsAndHashCode
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class TracingRedactionConfig implements Serializable {
+
+    private String defaultReplacement;
+
+    @Valid
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private List<TracingRedactionRule> rules;
```
araffass: The field has `@Valid` and `@JsonInclude(NON_EMPTY)` but nothing prevents it from being `null`. A PUT with `"redaction": {}` would deserialize with `rules = null`, pass validation, and get stored. The mapper handles it fine but we'd be persisting a state that doesn't really make sense. Maybe worth adding `@NotNull` here?

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/test/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactoryTest.java
vikrantgravitee: Done

gravitee-apim-console-webui/src/management/api/reporter-settings/reporter-settings-proxy/reporter-settings-proxy.component.ts
vikrantgravitee: Done

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
vikrantgravitee: Done

gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/v4/analytics/tracing/TracingRedactionConfig.java
vikrantgravitee: Done

araffass (approved):
LGTM

PR #16721, opened 2026-05-04
fix(http-proxy): use HC target host in setServer instead of endpoint defaultHost

dev-marek:
## Issue
https://gravitee.atlassian.net/browse/APIM-13512

## Description
- Fixed `HttpConnector.connect()` using `defaultHost` (endpoint config host) instead of `options.getHost()` (actual HC target host) in `setServer()` when a custom `Host` header is configured
- The bug caused TCP connections to go to the endpoint's configured host instead of the absolute HC target host, so health checks always hit the wrong server
- Added a regression test: endpoint target set to an unresolvable host (`endpoint-host.invalid`), HC target set to `127.0.0.1` (WireMock) — with the bug the test fails with `NXDOMAIN`, with the fix it passes
- V2 and the apiservice-healthcheck-http `RequestOptionsBuilder` both use `target.getHost()` correctly — the bug is isolated to `HttpConnector`

## Additional context
Root cause: after `setAbsoluteURI()` resolves the HC target, `options.getHost()` reflects the HC target IP, but the `setServer()` call one line below still used the stale `defaultHost` field from the constructor.

Reproduction: v4 HTTP-proxy endpoint with endpoint target `http://api.gravitee.io/`, health check target `http://178.63.67.153/` and a configured `Host: api.gravitee.io` header — the TCP connection goes to `api.gravitee.io` instead of `178.63.67.153`.


kovaren (commented):
Looks fine but I don't have full context, not sure about the risks of this change.

@phiz71 could you please take a look?

phiz71 (approved):

PR #16722, opened 2026-05-04
feat(portal): add pin/unpin feature to analytics dashboard cards

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-13853

## Description

This PR aims to handle the pin and unpin feature for dashboard cards, which is part of [APIM-13614](https://gravitee.atlassian.net/browse/APIM-13614)

## Additional context

Adds pin/unpin for analytics dashboard cards on the Portal analytics list. Users can pin up to 4 dashboards; pinned items appear in a dedicated row above the main grid, persist in localStorage, and stay visible across sessions.

Description

- User-facing behavior
- Pin button on each analytics dashboard card (grid and pinned row).
- Pinned row at the top of the analytics page when at least one dashboard is pinned, with a primary accent on pinned cards.
- Limit of 4 pinned dashboards; when the limit is reached, the pin control is disabled, shows a “Pin limit reached” tooltip, and does not navigate to the dashboard on click.
- Persistence via localStorage key analytics-pinned-dashboards.

Implementation highlights

- AnalyticsComponent: pin state (pinnedIds), rxResource to load pinned dashboards by id, togglePin() with localStorage sync.
- Stable pinned row during reloads using linkedSignal (keeps the last loaded list while pinnedResource is loading).
- AnalyticsDashboardCardComponent: isPinned, canPin, showAccent inputs; pin click uses stopPropagation and guards when disabled.
- Harness and unit tests for pin/unpin, limit, persistence, and invalid localStorage data.

https://github.com/user-attachments/assets/e3d9c93f-2803-43ba-a21a-fc83d22727e7

[APIM-13614]: https://gravitee.atlassian.net/browse/APIM-13614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
@@ -53,6 +53,10 @@
     }
   }
 
+  &__pinned-row {
+    display: block;
+  }
+
   &__content {
     display: flex;
     flex-direction: column;
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
@@ -16,8 +16,6 @@
 @use '../../../scss/theme' as app-theme;
 
 :host {
-  --cards-grid-row-height: 120px;
-
   display: flex;
   flex-direction: column;
   max-width: app-theme.$inner-content-width;
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
@@ -43,6 +43,20 @@
     }
   }
 
+  &__pinned {
+    display: flex;
+    flex-direction: column;
+    gap: 12px;
+
+    h2 {
+      margin: 0;
+    }
+  }
+
+  &__pinned-row {
+    display: block;
+  }
+
   &__content {
     display: flex;
     flex-direction: column;
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
@@ -16,6 +16,8 @@
 @use '../../../scss/theme' as app-theme;
 
 :host {
+  --cards-grid-row-height: 120px;
+
   display: flex;
   flex-direction: column;
   max-width: app-theme.$inner-content-width;
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
@@ -53,10 +53,6 @@
     }
   }
 
-  &__pinned-row {
-    display: block;
-  }
-
   &__content {
     display: flex;
     flex-direction: column;
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.spec.ts
@@ -202,6 +202,21 @@ describe('AnalyticsComponent', () => {
       ]);
       const pinned = await harness.getPinnedDashboards();
       expect(pinned).toHaveLength(3);
+      await fixture.whenStable();
+      fixture.detectChanges();
+      const pinned = await harness.getPinnedDashboards();
+      expect(pinned).toHaveLength(4);
+    });
+
+    it('should_allow_pinning_when_under_limit', async () => {
+      await pinAndFlush(0, 'dash-1', 'Dashboard 1');
+      await pinAndFlush(1, 'dash-2', 'Dashboard 2', [{ id: 'dash-1', name: 'Dashboard 1' }]);
+      await pinAndFlush(2, 'dash-3', 'Dashboard 3', [
+        { id: 'dash-1', name: 'Dashboard 1' },
+        { id: 'dash-2', name: 'Dashboard 2' },
+      ]);
+      const pinned = await harness.getPinnedDashboards();
+      expect(pinned).toHaveLength(3);
     });
   });
 
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.spec.ts
@@ -18,13 +18,13 @@ import { provideHttpClient } from '@angular/common/http';
 import { HttpTestingController, provideHttpClientTesting } from '@angular/common/http/testing';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { provideNoopAnimations } from '@angular/platform-browser/animations';
-import { Router, provideRouter } from '@angular/router';
+import { provideRouter, Router } from '@angular/router';
 
 import { analyticsListBreadcrumb } from './analytics-breadcrumbs';
 import AnalyticsComponent from './analytics.component';
 import { AnalyticsComponentHarness } from './analytics.harness';
 import { AnalyticsDashboardsResponse } from '../../../entities/analytics-dashboard/analytics-dashboard';
-import { fakeDashboard, fakeAnalyticsDashboardsResponse } from '../../../entities/analytics-dashboard/analytics-dashboard.fixtures';
+import { fakeAnalyticsDashboardsResponse, fakeDashboard } from '../../../entities/analytics-dashboard/analytics-dashboard.fixtures';
 import { BreadcrumbService } from '../../../services/breadcrumb.service';
 import { ConfigService } from '../../../services/config.service';
 import { TESTING_BASE_URL } from '../../../testing/app-testing.module';
@@ -131,7 +131,15 @@ describe('AnalyticsComponent', () => {
       await cards[cardIndex].clickPinButtonWithoutStabilizing();
     }
 
-    async function pinAndFlush(cardIndex: number, id: string, name: string, allPinned: { id: string; name: string }[] = []): Promise<void> {
+    async function pinAndFlush(
+      cardIndex: number,
+      id: string,
+      name: string,
+      allPinned: {
+        id: string;
+        name: string;
+      }[] = [],
+    ): Promise<void> {
       await clickPinButtonForCard(cardIndex);
       fixture.detectChanges();
       const pinned = [...allPinned, { id, name }];
@@ -162,7 +170,6 @@ describe('AnalyticsComponent', () => {
       fixture.detectChanges();
       const pinned = await harness.getPinnedDashboards();
       expect(pinned).toHaveLength(0);
-      expect(JSON.parse(localStorage.getItem('analytics-pinned-dashboards')!)).toEqual([]);
     });
 
     it('should_show_pinned_dashboards_in_pinned_row', async () => {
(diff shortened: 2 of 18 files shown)


gravitee-apim-portal-webui-next/src/components/cards-grid/cards-grid.component.scss
```diff
@@ -23,7 +23,7 @@
   display: grid;
   gap: 16px;
   grid-template-columns: repeat(4, 1fr);
-  grid-auto-rows: 138px;
+  grid-auto-rows: var(--cards-grid-row-height, 138px);
```
kovaren: 1. We should avoid using the raw CSS properties because they pierce through Angular style encapsulation. If we have a good reason to customize it, the value should rather come as an input to the component.

2. In this particular case, I'd argue that the card height should stay constant to maintain consistency everywhere. WDYT?

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
```diff
@@ -16,6 +16,8 @@
 @use '../../../scss/theme' as app-theme;
 
 :host {
+  --cards-grid-row-height: 120px;
```
kovaren: As per https://github.com/gravitee-io/gravitee-api-management/pull/16722/changes#r3184230995

```suggestion

File: gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.harness.ts [hunk 3]
```diff
@@ -48,4 +49,8 @@ export class AnalyticsComponentHarness extends ComponentHarness {
     const element = await this.locateTitle();
     return element ? element.text() : null;
   }
+
+  public async getPinnedRow(): Promise<unknown> {
+    return this.locatePinnedRow();
+  }
```
kovaren, line 59: This method acts just like a query selector. Instead, we should be returning an array of pinned dashboards of type AnalyticsDashboardCardHarness to be able to check their contents.

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.spec.ts
```diff
@@ -101,6 +101,71 @@ describe('AnalyticsComponent', () => {
...
+      );
+    });
+
+    afterEach(() => {
+      localStorage.clear();
+    });
+
+    it('should_toggle_pin_and_persist_to_local_storage', () => {
+      fixture.componentInstance.togglePin('dash-1');
+      expect(fixture.componentInstance.pinnedIds()).toEqual(['dash-1']);
```
kovaren: We should avoid directly calling methods on the components because this is not how users interact with the them.

Instead, the harness should provide methods to pin / unpin a dashboard by some param (e.g. by name) which will simulate a real interaction with a user who does is not aware of the internal working of the component.

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.spec.ts
```diff
@@ -101,6 +101,71 @@ describe('AnalyticsComponent', () => {
...
+      const pinnedRow = await harness.getPinnedRow();
+      expect(pinnedRow).toBeTruthy();
+    });
+
+    it('should_not_allow_pinning_more_than_4', () => {
+      fixture.componentInstance.togglePin('dash-1');
+      fixture.componentInstance.togglePin('dash-2');
+      fixture.componentInstance.togglePin('dash-3');
+      fixture.componentInstance.togglePin('dash-4');
+      expect(fixture.componentInstance.canPinMore()).toBe(false);
```
kovaren: Instead of calling the component method directly, consider checking that the behavior. For example, we can try pinning more than 4 dashboards and check that it has no effect (there are still only 4 pinned dashboards).

kovaren (changes requested):

kovaren:
There are a couple of visual quirks that would be nice to fix:
- the whole pinned row flashes each time a dashboard is pinned / unpinned, this is probably due to the `pinnedResource` value becoming undefined during the loading (we could solve this by keeping the current pinned array in a separate linked signal while the loading is in progress to avoid this effect)
- when no more dashboards can be pinned, a click on the pin icon acts like a click on the dashboard itself, this is a bit misleading, instead it should not have any effect. We could also add a small tooltip explaining why the pin icon is disabled.

https://github.com/user-attachments/assets/03b9ca38-0cdf-49ae-bf61-88ce3dcbf221


The pin icon has a bigger padding than the label, ideally they should be aligned:

image

gravitee-apim-portal-webui-next/src/components/cards-grid/cards-grid.component.scss
thomasballivet: Yes it makes more sense for it to be constant. Done

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
thomasballivet: done

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.harness.ts
thomasballivet: Done

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.spec.ts
thomasballivet: I got the point and it's done but I've been struggling a bit to allow the pin and unpin from the harness. It seems that we have a deadlock because of the `rxResource`. The harness would wait for the flush but the flush would be after the `await` on the harness method, which never resolves because `whenStable()` blocks until the HTTP request triggered by the rxResource is flushed.
I'm looking forward to find a solution. A mock on dashboardService would be an option but is against the rules. But it seems to be a good option. WDYT?

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.spec.ts
thomasballivet: Yes it makes more sense. Done

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.spec.ts
thomasballivet: For now I'll push with a DOM fallback but let me know what do you think

kovaren: Oh yes, that's a known pain point with rxResource.

Let's use a native click inside the harness for now. I added a small commit

[03edeaf5718bbd5a46c7e4cd1b2639b2a069480d](https://github.com/gravitee-io/gravitee-api-management/pull/16722/changes/03edeaf5718bbd5a46c7e4cd1b2639b2a069480d)

thomasballivet: thanks!

thomasballivet:
Fixed the flash issue. Pin button when no more card can be pinned has now no effect and display a tooltip saying "Pin limit reached".

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.ts
```diff
@@ -55,6 +59,10 @@ export default class AnalyticsComponent {
   readonly pageSize = 20;
   private readonly currentPage = signal(1);
 
+  readonly pinnedIds = signal<string[]>(this.loadPinnedIds());
+  readonly pinnedIdsSet = computed(() => new Set(this.pinnedIds()));
+  readonly canPinMore = computed(() => this.pinnedIds().length < AnalyticsComponent.MAX_PINNED);
```
aqibmohammadkhan, line 64: `pinnedIds` is readonly but public, which means any external consumer can call `.set()` on it and mutate state bypassing `togglePin()`. It should be `private` since the template only needs `pinnedIdsSet` and `canPinMore`.

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.ts
```diff
@@ -71,8 +79,28 @@ export default class AnalyticsComponent {
...
+
   constructor() {
     this.breadcrumbService.set([analyticsListBreadcrumb()]);
+    // Cache last loaded set to keep the pinned row stable during rxResource reloads.
+    effect(() => {
+      const value = this.pinnedResource.value();
+      if (value !== undefined) {
+        this.cachedPinnedDashboards.set(value);
+      }
+    });
```
aqibmohammadkhan, line 103: The flash fix works correctly, but the approach uses an `effect()` in the constructor to manually keep `cachedPinnedDashboards` in sync — precisely the pattern Ivan originally suggested solving with a `"linked signal"`. Angular 19's linkedSignal would express this more idiomatically and without needing the constructor:

```
// current
private readonly cachedPinnedDashboards = signal<Dashboard[]>([]);
readonly pinnedDashboards = computed(() => this.pinnedResource.value() ?? this.cachedPinnedDashboards());
constructor() {
  effect(() => {
    const value = this.pinnedResource.value();
    if (value !== undefined) this.cachedPinnedDashboards.set(value);
  });
}

// more idiomatic alternative
readonly pinnedDashboards = linkedSignal<Dashboard[] | undefined, Dashboard[]>({
  source: () => this.pinnedResource.value(),
  computation: (value, previous) => value ?? previous?.value ?? [],
});
```

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.spec.ts
```diff
@@ -101,6 +101,109 @@ describe('AnalyticsComponent', () => {
...
+
+    it('should_toggle_pin_and_persist_to_local_storage', async () => {
+      await pinAndFlush(0, 'dash-1', 'Dashboard 1');
+      const pinned = await harness.getPinnedDashboards();
+      expect(pinned).toHaveLength(1);
+      expect(await pinned[0].getTitle()).toBe('Dashboard 1');
+      expect(JSON.parse(localStorage.getItem('analytics-pinned-dashboards')!)).toEqual(['dash-1']);
+    });
+
+    it('should_unpin_when_already_pinned', async () => {
```
aqibmohammadkhan, line 165: The test confirms the visual state changes to 0 pinned cards, but doesn't check `localStorage.getItem('analytics-pinned-dashboards')` is updated to `'[]'`. The `should_toggle_pin_and_persist_to_local_storage` test verifies persistence on pin but there's an asymmetry with unpin.

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.ts
```diff
@@ -82,4 +110,25 @@ export default class AnalyticsComponent {
...
+    try {
+      const stored = localStorage.getItem(AnalyticsComponent.PINNED_KEY);
+      if (!stored) return [];
+      const parsed: unknown = JSON.parse(stored);
+      return Array.isArray(parsed) && parsed.every(value => typeof value === 'string') ? parsed : [];
+    } catch (error) {
+      console.warn('Failed to read pinned analytics dashboards from localStorage', error);
+      return [];
+    }
+  }
```
aqibmohammadkhan, line 133: The method guards against malformed JSON and wrong types (non-string array) with a try/catch — but there's no unit test exercising those paths. At a minimum, a test with `localStorage.setItem('analytics-pinned-dashboards', '[1,2,3]')` (numbers instead of strings) would be valuable.

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.scss
```diff
@@ -43,6 +43,20 @@
...
+    gap: 12px;
+
+    h2 {
+      margin: 0;
+    }
+  }
+
+  &__pinned-row {
+    display: block;
+  }
```
aqibmohammadkhan, line 58: `app-cards-grid` is a block element by default. this seems redundant and can be removed.

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.harness.ts
```diff
@@ -25,7 +25,9 @@ export class AnalyticsComponentHarness extends ComponentHarness {
   private readonly locateError = this.locatorForOptional('[data-testid="analytics-list-error"]');
   private readonly locateEmptyState = this.locatorForOptional('.cards-grid__empty-state');
   private readonly locateCards = this.locatorForAll(AnalyticsDashboardCardHarness);
+  private readonly locateGridCards = this.locatorForAll(AnalyticsDashboardCardHarness.with({ ancestor: '.analytics-list__grid' }));
```
aqibmohammadkhan, line 27: `AnalyticsComponentHarness` has both `locateCards` (all cards — both grid and pinned row) and `locateGridCards` (grid only). If a test calls `getCards()` it would return cards from both sections, which can lead to confusing test failures as the pinned row grows. Now that `getGridCards()` and `getPinnedDashboards()` exist as the precise accessors, `getCards()` / `locateCards` is a footgun and could be deprecated or removed to avoid confusion.

dev-marek (approved):
lgtm

aqibmohammadkhan (approved):

PR #16732, opened 2026-05-04
fix: remove lag on manage groups button for api-product

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-13831

## Description

The action buttons (Transfer ownership, Manage groups, Add members) were rendered immediately in a disabled state while the page forkJoin was in flight, because LOADING_STATE initialised groups as []. This made "Manage groups" appear blocked until all three requests completed.

Wrapped card__header__actions in @if (!state.isLoading) so buttons only render after data is ready, matching the *ngIf="form" guard used in api-general-members. Simplified [disabled] binding to state.groups.length === 0 now that the expression is only evaluated post-load.

### DIFF
--- a/gravitee-apim-console-webui/src/management/api-products/members/api-product-members.component.html
+++ b/gravitee-apim-console-webui/src/management/api-products/members/api-product-members.component.html
@@ -26,43 +26,45 @@
               Manage who can interact with your API Product through the API Management console
             </p>
           </div>
-          <div class="card__header__actions">
-            @if (canTransferOwnership) {
+          @if (!state.isLoading) {
+            <div class="card__header__actions">
+              @if (canTransferOwnership) {
+                <button
+                  class="card__header__actions__button"
+                  mat-raised-button
+                  type="button"
+                  (click)="transferOwnership()"
+                  data-testid="api_product_members_transfer_ownership_button"
+                >
+                  Transfer ownership
+                </button>
+              }
               <button
+                *gioPermission="{ anyOf: ['api_product-member-u'] }"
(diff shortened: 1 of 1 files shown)


gravitee-apim-console-webui/src/management/api-products/members/api-product-members.component.html
```diff
@@ -26,43 +26,45 @@ <h3 data-testid="api_product_members_title">Members</h3>
...
               <button
+                *gioPermission="{ anyOf: ['api_product-member-u'] }"
                 class="card__header__actions__button"
                 mat-raised-button
                 type="button"
-                (click)="transferOwnership()"
-                data-testid="api_product_members_transfer_ownership_button"
+                aria-label="Manage groups"
+                data-testid="api_product_members_manage_groups_button"
+                [disabled]="state.groups.length === 0"
```
araffass, line 49: Note: 
"**Add members"** is also hidden during loading, but it doesn't depend on groups data at all, so if my understanding is right i think it didn't need to be wrapped.
-> The @if (!state.isLoading) could be scoped to just the Manage groups button, using [disabled]="state.isLoading || state.groups.length === 0".

araffass (approved):
Fix looks good to me

gravitee-apim-console-webui/src/management/api-products/members/api-product-members.component.html
vikrantgravitee: It will still lag as isLoading will be true.

PR #16733, opened 2026-05-04
Implement If-Match handling in PATCH API

carlos-andres-osorio:
## Issue

https://gravitee.atlassian.net/browse/APIM-13694

## Description

Implemented support for the `If-Match` HTTP header in the PATCH v4 API method. This ensures conditional updates based on ETag validation, improving concurrency control and preventing unintentional overwrites.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
@@ -21,6 +21,8 @@ import static io.gravitee.rest.api.model.permissions.SystemRole.PRIMARY_OWNER;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 
+import io.gravitee.apim.core.api.exception.ApiInvalidDefinitionVersionException;
+import io.gravitee.apim.core.api.exception.ApiInvalidTypeException;
 import io.gravitee.apim.core.api.model.UpdateNativeApi;
 import io.gravitee.apim.core.api.model.crd.IDExportStrategy;
 import io.gravitee.apim.core.api.model.import_definition.ImportDefinition;
@@ -46,8 +48,10 @@ import io.gravitee.apim.infra.adapter.ApiAdapter;
 import io.gravitee.common.component.Lifecycle;
 import io.gravitee.common.data.domain.Page;
 import io.gravitee.common.http.MediaType;
+import io.gravitee.definition.model.DefinitionVersion;
 import io.gravitee.definition.model.Proxy;
 import io.gravitee.definition.model.VirtualHost;
+import io.gravitee.definition.model.v4.ApiType;
 import io.gravitee.definition.model.v4.listener.Listener;
 import io.gravitee.definition.model.v4.listener.ListenerType;
 import io.gravitee.definition.model.v4.listener.http.HttpListener;
@@ -128,6 +132,7 @@ import io.gravitee.rest.api.service.exceptions.ApiNotFoundException;
 import io.gravitee.rest.api.service.exceptions.ForbiddenAccessException;
 import io.gravitee.rest.api.service.exceptions.ForbiddenFeatureException;
 import io.gravitee.rest.api.service.exceptions.InvalidLicenseException;
+import io.gravitee.rest.api.service.exceptions.PreconditionFailedException;
 import io.gravitee.rest.api.service.exceptions.TechnicalManagementException;
 import io.gravitee.rest.api.service.exceptions.TransferOwnershipNotAllowedException;
 import io.gravitee.rest.api.service.v4.ApiDuplicateService;
@@ -171,6 +176,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import lombok.CustomLog;
+import org.glassfish.jersey.message.internal.HttpHeaderReader;
 
 /**
  * Defines the REST resources to manage API v4.
@@ -467,6 +473,19 @@ public class ApiResource extends AbstractResource {
         @Context HttpHeaders headers,
         String body
     ) {
+        final var currentEntity = getGenericApiEntityById(apiId, false, false, false, false);
+
+        if (!DefinitionVersion.V4.equals(currentEntity.getDefinitionVersion())) {
+            throw new ApiInvalidDefinitionVersionException(apiId);
+        }
+        if (!(currentEntity instanceof ApiEntity httpApi) || !ApiType.PROXY.equals(httpApi.getType())) {
+            throw new ApiInvalidTypeException(apiId, ApiType.PROXY);
+        }
+
+        if (currentEntity.getUpdatedAt() != null) {
+            evaluateIfMatchStrictly(headers, Long.toString(currentEntity.getUpdatedAt().getTime()));
+        }
+
         var contentType = headers.getMediaType();
         var patchType = contentType != null && JSON_PATCH_MEDIA_TYPE.isCompatible(contentType)
             ? PatchApiUseCase.PatchType.JSON_PATCH
@@ -1245,6 +1264,22 @@ public class ApiResource extends AbstractResource {
         baos.write(image.getContent(), 0, image.getContent().length);
 
         return Response.ok(baos).cacheControl(cc).tag(etag).type(image.getType()).build();
+    }
+
+    private void evaluateIfMatchStrictly(final HttpHeaders headers, final String etagValue) {
+        String ifMatch = headers.getHeaderString(HttpHeaders.IF_MATCH);
+
+        if (Objects.nonNull(ifMatch) && !ifMatch.isEmpty()) {
+            if ("*".equals(ifMatch)) {
+                return;
+            }
+            try {
+                HttpHeaderReader.readMatchingEntityTag(ifMatch.replace("-gzip", ""));
+            } catch (java.text.ParseException e) {
+                throw new PreconditionFailedException();
+            }
+            evaluateIfMatch(headers, etagValue);
+        }
     }
 
     private Response apiResponse(GenericApiEntity apiEntity) {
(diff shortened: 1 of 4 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
```diff
@@ -1247,6 +1266,22 @@ private Response imageResponse(final Request request, InlinePictureEntity image)
...
 
+    private void evaluateIfMatchStrictly(final HttpHeaders headers, final String etagValue) {
+        String ifMatch = headers.getHeaderString(HttpHeaders.IF_MATCH);
+
+        if (Objects.nonNull(ifMatch) && !ifMatch.isEmpty()) {
+            if ("*".equals(ifMatch)) {
+                return;
+            }
+            try {
+                HttpHeaderReader.readMatchingEntityTag(ifMatch.replaceAll("-gzip", ""));
```
paulatulis: String.replaceAll(String, String) interprets its first argument as a regex pattern, not a literal string. "-gzip" happens to be a valid (and harmless) regex, but it is semantically incorrect and triggers SonarQube rule S5361 (Replace replaceAll() by replace()). This is a CI quality gate failure on this PR right now. It should be replace() since no regex features are needed.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
```diff
@@ -1247,6 +1266,22 @@ private Response imageResponse(final Request request, InlinePictureEntity image)
         return Response.ok(baos).cacheControl(cc).tag(etag).type(image.getType()).build();
     }
 
+    private void evaluateIfMatchStrictly(final HttpHeaders headers, final String etagValue) {
```
paulatulis, line 1269: `evaluateIfMatchStrictly` calls the parent `evaluateIfMatch` after verifying the tag parses. The parent helper accepts W/"..." (weak ETags) as matching strong ones — this is a violation of RFC 7232 §3.1, which mandates strong comparison for If-Match.

For a GET response this is low-risk. For a conditional write (PATCH), this matters: a client could receive a W/"1000" ETag from some proxy cache layer, use it in an If-Match, and bypass the intended comparison. The Jira ticket explicitly documents this as a known pre-existing defect ("weak-vs-strong ETag comparison violation"), but given this is a new write endpoint, a comment in `evaluateIfMatchStrictly` documenting this risk would prevent future maintainers from extending the helper to other write operations unknowingly.

Recommendation: Add an inline // TODO + ticket reference noting that `evaluateIfMatch` does not do strong-only comparison and should not be used on write paths without this caveat. Better still, perform a full strong-comparison directly in `evaluateIfMatchStrictly` rather than delegating.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
```diff
@@ -467,6 +473,19 @@ public Response patchApi(
         @Context HttpHeaders headers,
         String body
     ) {
+        final var currentEntity = getGenericApiEntityById(apiId, false, false, false, false);
```
paulatulis, line 476: This along with line 
```
var output = patchApiUseCase.execute(input);
```

makes it so that there are two sequential reads of the same entity. Between them, another request could modify the API. If `patchApiUseCase` re-fetches the entity, the ETag value it patched against is already stale by the time `updateApiDomainService.updateV4` runs. This makes the precondition check advisory rather than atomic — a concurrent write between the two reads can still corrupt the state even when If-Match was provided.

The Jira ticket acknowledges the same-millisecond collision, but does not explicitly acknowledge this broader TOCTOU window for all same-second or concurrent writes.

Recommendation: Document explicitly in code (or in an ADR/ticket) that the concurrency guarantee is best-effort / advisory (not serializable), referencing the window between the read in `patchApi` and the write in `patchApiUseCase`. Ideally, `PatchApiUseCase` should accept the fetched entity to avoid the double-fetch.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiTest.java
```diff
@@ -83,6 +85,15 @@ void setUpApiAndPrimaryOwner() {
         var existingApi = ApiFixtures.aProxyApiV4();
         apiCrudService.initWith(List.of(existingApi));
 
+        var apiEntity = new ApiEntity();
+        apiEntity.setId(API);
+        apiEntity.setDefinitionVersion(io.gravitee.definition.model.DefinitionVersion.V4);
+        apiEntity.setType(io.gravitee.definition.model.v4.ApiType.PROXY);
+        apiEntity.setUpdatedAt(new Date());
```
paulatulis: PatchApiTest's `@BeforeEach` sets `updatedAt` to `new Date()`, meaning the ETag value changes on every test run. This is fine as long as the test only checks HTTP 200, but it's a footgun: if any future test in this class adds an If-Match assertion against a fixed value, it will flake.

`PatchApiConcurrencyTest` correctly uses a fixed I`nstant.ofEpochMilli(1000)` — the same discipline should be applied to `PatchApiTest`.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiConcurrencyTest.java
```diff
@@ -0,0 +1,290 @@
+/*
```
paulatulis, line 1: The 412 response body is not asserted anywhere — only the status code is checked. Given that this is a new HTTP semantics feature, a test that validates the response body on 412 (or at least that it's not a stack trace) would close the feedback loop for client SDK authors.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
carlos-andres-osorio: Changed `replaceAll("-gzip", "")` to `replace("-gzip", "")` in `evaluateIfMatchStrictly`. The first argument is a literal string with no regex features needed; `replace()` is the correct method and resolves Sonar S5361.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
carlos-andres-osorio: The behavior in `evaluateIfMatch` — where the server-side `EntityTag` inherits the weakness flag from the incoming header — is a pre-existing pattern shared by all write endpoints in this class (PUT, `_start`, `_stop` at lines 430, 438, 782, 795, 953, 971, 989). Changing `evaluateIfMatchStrictly` to perform strict comparison independently would diverge from those endpoints without a clear owner for the wider refactor. The known limitation is documented in the story's risk register.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
carlos-andres-osorio: The double-fetch pattern — entity read at line 476 for ETag comparison, then use-case re-fetch on write — is consistent with all other write endpoints in this class (PUT, `_start`, `_stop`) which share the same structure. Making the concurrency guarantee serializable would require refactoring `PatchApiUseCase` to accept the pre-fetched entity, which is an architectural change beyond this PR's scope. The TOCTOU window and its advisory-only nature have been added to the story's risk documentation.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiTest.java
carlos-andres-osorio: Replaced `new Date()` with `Date.from(Instant.ofEpochMilli(1000))` in `ApiResource_PatchApiTest.@BeforeEach`, matching the fixed epoch value already used in `ApiResource_PatchApiConcurrencyTest`. The ETag derived from `updatedAt` is now deterministic across runs.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiConcurrencyTest.java
carlos-andres-osorio: Good catch. The 412 body is intentionally empty — this is consistent with the existing behavior for _start and _stop, which also return empty bodies on 412. I've
  updated the test to explicitly assert body.isEmpty() so the intent is documented rather than implicit.

paulatulis (approved):
Consider that if-match handled! LGTM.

PR #16736, opened 2026-05-05
Feat/ext122 openapi viewer frontend

adam-rosowicz:
Link to jira ticket: https://gravitee.atlassian.net/browse/EXT-122

Summary
Implements the OpenAPI viewer configuration UI for the console navigation items editor, and wires all configuration options through to both the console preview and the portal-next rendering pipeline.

Console (gravitee-apim-console-webui)
New configuration dialog (openapi-config-dialog) — allows configuring the OpenAPI viewer with all available options: viewer selection (SwaggerUI / Redoc), base URL override, and SwaggerUI-specific options (Try It Out, doc expansion, operation ID display, filtering, extensions, max displayed tags, OAuth/PKCE settings)
Redoc viewer added as a new gio-redoc Angular component, used in the split-pane OpenAPI editor preview alongside the existing gio-swagger-ui
Config options wired to preview — all SwaggerUI configuration values (tryItURL, docExpansion, displayOperationId, enableFiltering, showExtensions, showCommonExtensions, maxDisplayedTags) are now passed from the config dialog to the viewer components in real time

Configuration modal:
image

Swagger preview:
image

Redoc preview:
image

Swagger portal:
image

Redoc portal:
image

### DIFF
--- a/gravitee-apim-console-webui/project.json
+++ b/gravitee-apim-console-webui/project.json
@@ -97,6 +97,11 @@
             "glob": "**/*",
             "input": "gravitee-apim-webui-libs/gravitee-markdown/src/lib/assets/homepage",
             "output": "assets/homepage"
+          },
+          {
+            "glob": "redoc.standalone.js",
+            "input": "node_modules/redoc/bundles",
+            "output": "assets/redoc"
           }
         ],
         "styles": [
--- a/gravitee-apim-console-webui/src/components/documentation/gio-redoc/gio-redoc.component.scss
+++ b/gravitee-apim-console-webui/src/components/documentation/gio-redoc/gio-redoc.component.scss
@@ -1,11 +1,11 @@
 /*
- * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *         http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -13,17 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.gravitee.apim.core.portal_page.model;
 
-import jakarta.validation.constraints.NotNull;
-
-public record RedocConfiguration(@NotNull String tryItUrl) implements OpenApiConfiguration {
-    public RedocConfiguration() {
-        this("");
-    }
-
-    @Override
-    public Viewer viewer() {
-        return Viewer.REDOC;
-    }
+:host {
+  display: block;
+  height: 100%;
 }
--- a/gravitee-apim-console-webui/src/components/documentation/gio-redoc/gio-redoc.component.ts
+++ b/gravitee-apim-console-webui/src/components/documentation/gio-redoc/gio-redoc.component.ts
@@ -1,0 +1,118 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Component, ElementRef, OnDestroy, effect, inject, input, viewChild } from '@angular/core';
+import * as yaml from 'js-yaml';
+
+import { GioRedocService, type RedocApi } from './gio-redoc.service';
+
+const yamlSchema = yaml.DEFAULT_SCHEMA.extend([]);
+
+const parseSpec = (spec: string): object | null => {
+  try {
+    const result: unknown = JSON.parse(spec);
+    return result instanceof Object ? result : null;
+  } catch {
+    const result: unknown = yaml.load(spec, { schema: yamlSchema });
+    return result instanceof Object ? result : null;
+  }
+};
+
+@Component({
+  selector: 'gio-redoc',
+  templateUrl: './gio-redoc.component.html',
+  styleUrls: ['./gio-redoc.component.scss'],
+  standalone: true,
+})
+export class GioRedocComponent implements OnDestroy {
+  private readonly gioRedocService = inject(GioRedocService);
+
+  spec = input('');
+  tryItURL = input('');
+
+  private readonly redocNode = viewChild<ElementRef<HTMLElement>>('redoc');
+
+  private pendingTimeoutId: ReturnType<typeof setTimeout> | null = null;
+  private initCallId = 0;
+
+  private readonly initEffect = effect(() => {
+    this.spec();
+    this.tryItURL();
+    const redocNode = this.redocNode();
+    if (redocNode) {
+      this.scheduleInit();
+    }
+  });
+
+  ngOnDestroy(): void {
+    if (this.pendingTimeoutId !== null) {
+      clearTimeout(this.pendingTimeoutId);
+      this.pendingTimeoutId = null;
+    }
+  }
+
+  private scheduleInit(): void {
+    if (this.pendingTimeoutId !== null) {
+      clearTimeout(this.pendingTimeoutId);
+    }
+    this.pendingTimeoutId = setTimeout(() => {
+      this.pendingTimeoutId = null;
+      this.initRedoc();
+    }, 300);
+  }
+
+  private async initRedoc(): Promise<void> {
+    const callId = ++this.initCallId;
+    const specValue = this.spec();
+    const redocNode = this.redocNode();
+    if (!specValue || !redocNode) return;
+
+    let redoc: RedocApi;
+    try {
+      redoc = await this.gioRedocService.load();
+    } catch (err) {
+      // eslint-disable-next-line angular/log
+      console.error('[gio-redoc] Failed to load Redoc script:', err);
+      return;
+    }
+
+    if (callId !== this.initCallId) return;
+
+    const tryItURL = this.tryItURL();
+    const parsedSpec = parseSpec(specValue);
+    if (parsedSpec === null) return;
+    const spec = tryItURL ? { ...parsedSpec, servers: [{ url: tryItURL }] } : parsedSpec;
+    redocNode.nativeElement.innerHTML = '';
+    redoc.init(
+      spec,
+      {
+        hideDownloadButton: true,
+        hideLoading: true,
+        disableSearch: true,
+        onlyRequiredInSamples: true,
+        nativeScrollbars: true,
+        theme: {
+          sidebar: { width: '150px' },
+          breakpoints: { medium: '599rem', large: '600rem' },
+        },
+      },
+      redocNode.nativeElement,
+      err => {
+        // eslint-disable-next-line angular/log
+        if (err) console.error('[gio-redoc] Redoc.init failed:', err);
+      },
+    );
+  }
+}
--- a/gravitee-apim-console-webui/src/portal/components/openapi-editor/openapi-editor.component.ts
+++ b/gravitee-apim-console-webui/src/portal/components/openapi-editor/openapi-editor.component.ts
@@ -13,11 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { Component } from '@angular/core';
+import { Component, computed, input } from '@angular/core';
 import { ControlValueAccessor, FormsModule, NG_VALUE_ACCESSOR } from '@angular/forms';
 import { GioMonacoEditorModule } from '@gravitee/ui-particles-angular';
 
 import { GioSwaggerUiModule } from '../../../components/documentation/gio-swagger-ui/gio-swagger-ui.module';
+import { GioRedocComponent } from '../../../components/documentation/gio-redoc/gio-redoc.component';
+import {
+  OpenApiDocExpansion,
+  OpenApiViewer,
+  OpenApiViewerConfiguration,
+} from '../../../entities/management-api-v2/portalPageContent/openApiViewerConfiguration';
 
 @Component({
   selector: 'openapi-editor',
@@ -30,15 +36,32 @@ import { GioSwaggerUiModule } from '../../../components/documentation/gio-swagge
       multi: true,
     },
   ],
-  imports: [FormsModule, GioMonacoEditorModule, GioSwaggerUiModule],
+  imports: [FormsModule, GioMonacoEditorModule, GioSwaggerUiModule, GioRedocComponent],
 })
 export class OpenApiEditorComponent implements ControlValueAccessor {
+  configuration = input<Partial<OpenApiViewerConfiguration> | null>(null);
+
   _value = '';
   private _disabled = false;
 
   _onChange: (value: string) => void = () => ({});
 
   _onTouched: () => void = () => ({});
+
+  isRedocViewer = computed(() => this.configuration()?.viewer === OpenApiViewer.Redoc);
+
+  swaggerTryItURL = computed(() => this.configuration()?.tryItURL ?? '');
+  swaggerDocExpansion = computed(() => this.configuration()?.docExpansion ?? OpenApiDocExpansion.None);
+  swaggerDisplayOperationId = computed(() => this.configuration()?.displayOperationId ?? false);
+  swaggerFilter = computed(() => this.configuration()?.enableFiltering ?? false);
+  swaggerShowExtensions = computed(() => this.configuration()?.showExtensions ?? false);
+  swaggerShowCommonExtensions = computed(() => this.configuration()?.showCommonExtensions ?? false);
+  swaggerMaxDisplayedTags = computed(() => {
+    const raw = this.configuration()?.maxDisplayedTags;
+    if (raw == null) return undefined;
+    const v = Number(raw);
+    return Number.isFinite(v) && v >= 0 ? v : undefined;
+  });
 
   registerOnChange(fn: (value: string) => void): void {
     this._onChange = fn;
--- a/gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
+++ b/gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
@@ -1,0 +1,132 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Component, computed, inject } from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
+import { FormControl, FormGroup, ReactiveFormsModule, Validators } from '@angular/forms';
+import { MAT_DIALOG_DATA, MatDialogModule, MatDialogRef } from '@angular/material/dialog';
+import { MatButtonModule } from '@angular/material/button';
+import { MatFormFieldModule } from '@angular/material/form-field';
+import { MatInputModule } from '@angular/material/input';
+import { MatSelectModule } from '@angular/material/select';
+import { MatSlideToggleModule } from '@angular/material/slide-toggle';
+import { GioFormSlideToggleModule } from '@gravitee/ui-particles-angular';
+
+import {
+  OpenApiDocExpansion,
+  OpenApiViewer,
+  OpenApiViewerConfiguration,
+} from '../../../entities/management-api-v2/portalPageContent/openApiViewerConfiguration';
+
+export interface OpenApiConfigDialogData {
+  configuration: Partial<OpenApiViewerConfiguration>;
+}
+
+interface OpenApiFormControls {
+  viewer: FormControl<OpenApiViewer>;
+  tryItURL: FormControl<string>;
+  tryIt: FormControl<boolean>;
+  disableSyntaxHighlight: FormControl<boolean>;
+  tryItAnonymous: FormControl<boolean>;
+  showURL: FormControl<boolean>;
+  displayOperationId: FormControl<boolean>;
+  usePkce: FormControl<boolean>;
+  docExpansion: FormControl<OpenApiDocExpansion>;
+  enableFiltering: FormControl<boolean>;
+  showExtensions: FormControl<boolean>;
+  showCommonExtensions: FormControl<boolean>;
+  maxDisplayedTags: FormControl<number | null>;
+}
+
+function parseBoolean(value: string | boolean | undefined): boolean {
+  return value === 'true' || value === true;
+}
+
+function parseNumber(value: string | number | undefined): number | null {
+  if (value === undefined || value === '' || value === null) return null;
+  const parsed = Number(value);
+  return Number.isNaN(parsed) ? null : parsed;
+}
+
+@Component({
+  selector: 'openapi-config-dialog',
+  standalone: true,
+  imports: [
+    MatDialogModule,
+    ReactiveFormsModule,
+    MatButtonModule,
+    MatFormFieldModule,
+    MatInputModule,
+    MatSelectModule,
+    MatSlideToggleModule,
+    GioFormSlideToggleModule,
+  ],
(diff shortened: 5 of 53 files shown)


gravitee-apim-console-webui/src/components/documentation/gio-redoc/gio-redoc.component.scss
aqibmohammadkhan, line 1: This java file is used in other places and should not be deleted. Please create the SCSS file as a new file and keep `UpdatePortalPageContent.java`.
Force the rename detection off (git diff --no-renames locally to verify) or split the commit.

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
```diff
@@ -541,6 +550,38 @@ export class PortalNavigationItemsComponent implements HasUnsavedChanges {
...
+        filter(result => !!result),
+        switchMap(result => {
+          const configuration = Object.fromEntries(Object.entries(result).map(([k, v]) => [k, String(v)]));
+          return this.portalPageContentService
+            .updatePageContent(navItem.portalPageContentId, {
+              content: this.contentControl.value ?? '',
+              configuration,
+            })
+            .pipe(map(() => configuration));
+        }),
```
aqibmohammadkhan: The backend doesn't accept `configuration` on `PortalPageContent`.
The new console code PUTs `{ content, configuration }` and reads `configuration` back from the GET response.

the current backend contract has no such field. ( see file `gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml`)

So configuration will be dropped server-side and the user's saved viewer settings won't save after a refresh — config will only "stick" inside the in-memory `currentPageConfiguration` signal until the page is reloaded. 

This is the API-first rule in `gravitee-apim-rest-api/AGENTS.md`: spec first, then frontend. 

A companion backend PR needs to extend `PortalPageContent` and `UpdatePortalPageContent` (and the `OpenAPI` generator) with a typed configuration object, or this PR can't ship

gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  readonly form = new FormGroup<OpenApiFormControls>({
+    viewer: new FormControl(this.configuration['viewer'] ?? 'Swagger', { nonNullable: true }),
+    entrypointsAsServers: new FormControl(parseBoolean(this.configuration['entrypointsAsServers']), { nonNullable: true }),
+    entrypointAsBasePath: new FormControl(parseBoolean(this.configuration['entrypointAsBasePath']), { nonNullable: true }),
+    tryItURL: new FormControl(this.configuration['tryItURL'] ?? '', {
+      nonNullable: true,
+      validators: [Validators.pattern(/^$|^https?:\/\/.+/)],
+    }),
+    tryIt: new FormControl(parseBoolean(this.configuration['tryIt']), { nonNullable: true }),
```
aqibmohammadkhan: Console and portal use different key casing for the same field
The console writes camelCase keys into configuration:


`openapi-config-dialog.component.ts`
```
tryItURL: new FormControl(this.configuration['tryItURL'] ?? '', { ... }),
tryIt: new FormControl(parseBoolean(this.configuration['tryIt']), { nonNullable: true }),
...
displayOperationId: new FormControl(parseBoolean(this.configuration['displayOperationId']), { nonNullable: true }),
usePkce: new FormControl(parseBoolean(this.configuration['usePkce']), { nonNullable: true }),
docExpansion: new FormControl(this.configuration['docExpansion'] ?? 'none', { nonNullable: true }),
```

But on the portal side, `PortalPageContent.configuration` is typed as `PageConfiguration`, which (as auto-generated from the Portal OpenAPI spec) uses snake_case:

`page-configuration.ts`

```
export interface PageConfiguration {
  try_it?: boolean;
  disable_syntax_highlight?: boolean;
  try_it_anonymous?: boolean;
  try_it_url?: string;
  show_url?: boolean;
  display_operation_id?: boolean;
  doc_expansion?: DocExpansionEnum;
  ...
  use_pkce?: boolean;
  viewer?: ViewerEnum;
}
```

This PR papers over the mismatch with as `unknown as Record<string, unknown>` casts and "either-or" lookups throughout `page-swagger.component.ts`:


`page-swagger.component.ts`
```
const cfg = this.page.configuration as unknown as Record<string, unknown>;
const showUrl = cfg['showURL'] ?? cfg['show_url'];
const tryItUrl = (cfg['tryItURL'] ?? cfg['try_it_url']) as string | undefined;
...
config.docExpansion = (cfg['docExpansion'] ?? cfg['doc_expansion'] ?? 'none') as 'none' | 'list' | 'full';
config.displayOperationId = Boolean(cfg['displayOperationId'] ?? cfg['display_operation_id']);
```

This effectively turns off type-checking for configuration on the portal side and creates two parallel naming schemes for the same persisted record. 


Please standardize on one (the existing `snake_case` is the contract the portal API already exposes; the console can serialize to that, or the portal regeneration can move to camelCase) and remove the a`s unknown as … ` casts.

The console also uses bare 'Swagger' / 'Redoc' strings while `portal-page-configuration.ts` already defines a `ViewerEnum`. 
Reuse it (or the equivalent from the management-v2 generated models once the backend exposes it).

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
```diff
@@ -541,6 +550,38 @@ export class PortalNavigationItemsComponent implements HasUnsavedChanges {
     }
   }
 
+  onConfigure() {
```
aqibmohammadkhan, line 624: `onConfigure()` silently persists in-progress content edits and bypasses the `unsaved-changes` guard

Two problems:

- It's not wrapped in `checkUnsavedChangesAndRun(...)` like `onEdit()` and `onPublishToggle()` are. A user with pending content edits who hits "Configure" → Save will silently publish those content edits along with the new viewer config, with no confirmation.

- Coupling the viewer-config save with the content body forces this behavior even when fixed. Ideally the configuration is a separate sub-resource (`PATCH .../{id}/configuration` or similar) so "Configure → Save" only mutates configuration. Worth coordinating with the backend change in blocker [here](https://github.com/gravitee-io/gravitee-api-management/pull/16736/changes#r3295208997)

gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
```diff
@@ -0,0 +1,143 @@
...
+import { MatFormFieldModule } from '@angular/material/form-field';
+import { MatInputModule } from '@angular/material/input';
+import { MatSelectModule } from '@angular/material/select';
+import { MatSlideToggleModule } from '@angular/material/slide-toggle';
+import { GioFormSlideToggleModule } from '@gravitee/ui-particles-angular';
+
+export interface OpenApiViewerConfiguration {
+  viewer: string;
+  entrypointsAsServers: boolean;
+  entrypointAsBasePath: boolean;
```
aqibmohammadkhan: `entrypointsAsServers` and `entrypointAsBasePath` are saved but never consumed.

Either remove them from the dialog for now or wire them up (the entrypoints URL will need to come from the API context, which the dialog doesn't currently fetch).

gravitee-apim-portal-webui-next/src/components/navigation-item-content-viewer/navigation-item-content-viewer.component.html
```diff
@@ -22,7 +22,20 @@
...
+      } @else {
+        <app-page-swagger
+          [page]="{
+            id: '',
+            name: '',
+            type: 'SWAGGER',
+            order: 0,
+            content: pageContentVM.content,
+            configuration: pageContentVM.configuration,
+          }"
```
aqibmohammadkhan: `PageSwaggerComponent` reuse causes a re-init storm and a global-ID clash.

- An inline object literal is constructed on every change-detection cycle, so `ngOnChanges` fires every CD pass and `SwaggerUI({...})` is reinitialized constantly. Wrap the Page in a `computed()` derived from `pageContent()`.

- `PageSwaggerComponent` still uses a global DOM id - If two OpenAPI pages, or this viewer plus the legacy `page.component`, are ever on the same page, `getElementById('swagger')` will mount onto the wrong node. 
 Please convert to `@ViewChild / viewChild()` on a local element.

gravitee-apim-console-webui/src/index.html
```diff
@@ -189,5 +189,6 @@
       </div>
     </app-root>
     <script src="webcomponents/webcomponents-loader.js"></script>
+    <script src="assets/redoc/redoc.standalone.js"></script>
```
aqibmohammadkhan: `redoc.standalone.js` is loading it from `index.html` adds that cost to every console session, even for users who never open the OpenAPI navigation editor. 

The portal-next side already lazy-loads Redoc via `RedocService`. consider doing the same in the console (e.g. inject a <script> from `GioRedocComponent.ngOnInit` and resolve a promise on load).

gravitee-apim-console-webui/project.json
```diff
@@ -92,6 +92,11 @@
             "glob": "**/*",
             "input": "gravitee-apim-webui-libs/gravitee-markdown/src/lib/assets/homepage",
             "output": "assets/homepage"
+          },
+          {
+            "glob": "redoc.standalone.js",
+            "input": "node_modules/redoc/bundles",
+            "output": "assets/redoc"
```
aqibmohammadkhan, line 104: `redoc` missing from `gravitee-apim-console-webui/package.json`
The build only works because of workspace hoisting — that's an implicit dependency that breaks the moment anyone runs the console with a non-hoisted install. Add `"redoc": "2.4.0"` to the console's own `package.json` (matching what `gravitee-apim-portal-webui` does).

gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
```diff
@@ -0,0 +1,143 @@
...
+  styleUrls: ['./openapi-config-dialog.component.scss'],
+})
+export class OpenApiConfigDialogComponent {
+  private readonly dialogRef = inject(MatDialogRef<OpenApiConfigDialogComponent, OpenApiConfigDialogResult>);
+  private readonly data: OpenApiConfigDialogData = inject(MAT_DIALOG_DATA);
+
+  private readonly configuration = this.data.configuration ?? {};
+
+  readonly form = new FormGroup<OpenApiFormControls>({
+    viewer: new FormControl(this.configuration['viewer'] ?? 'Swagger', { nonNullable: true }),
```
aqibmohammadkhan: The dialog writes `'Swagger' | 'Redoc'` (title-case):

`openapi-config-dialog.component.ts`

`viewer: new FormControl(this.configuration['viewer'] ?? 'Swagger', { nonNullable: true }),`

Console preview compares case-sensitively (=== 'Redoc'):

`openapi-editor.component.ts`

`isRedocViewer = computed(() => this.configuration()?.['viewer'] === 'Redoc');`

Portal compares case-insensitively (`.toLowerCase() === 'redoc'`):


`navigation-item-content-viewer.component.html`

`@if (pageContentVM.configuration?.viewer?.toLowerCase() === 'redoc') {`


Three rules, one field. Pick one (preferably the existing `ViewerEnum` constant from `page-configuration.ts`) and use it on both sides, including the dialog interface (`viewer: string` should be `viewer: ViewerEnum`).

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
```diff
@@ -541,6 +550,38 @@ export class PortalNavigationItemsComponent implements HasUnsavedChanges {
...
+    this.matDialog
+      .open<OpenApiConfigDialogComponent, OpenApiConfigDialogData, OpenApiConfigDialogResult>(OpenApiConfigDialogComponent, {
+        width: GIO_DIALOG_WIDTH.MEDIUM,
+        data: { configuration: this.currentPageConfiguration() },
+      })
+      .afterClosed()
+      .pipe(
+        filter(result => !!result),
+        switchMap(result => {
+          const configuration = Object.fromEntries(Object.entries(result).map(([k, v]) => [k, String(v)]));
```
aqibmohammadkhan: Every boolean becomes `"true"/"false"`, every number becomes a string, then every consumer re-parses with `=== 'true' / Number(...)`. It also means `maxDisplayedTags` has to use `-1` as a sentinel for `"show all"`. If the backend is going to be a proper typed schema (it should — see blocker [here](https://github.com/gravitee-io/gravitee-api-management/pull/16736/changes#r3295208997) ), keep native types end-to-end and delete parseBoolean/parseNumber and the sentinel.

gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
```diff
@@ -0,0 +1,143 @@
...
+  showExtensions: boolean;
+  showCommonExtensions: boolean;
+  maxDisplayedTags: number;
+}
+
+export interface OpenApiConfigDialogData {
+  configuration: Record<string, string>;
+}
+
+export type OpenApiConfigDialogResult = OpenApiViewerConfiguration;
```
aqibmohammadkhan: `OpenApiViewerConfiguration` interface isn't actually enforced but the dialog returns `this.form.getRawValue()` (the form group's raw value), then it's spread into Object.entries. The declared `OpenApiConfigDialogResult` type isn't checked. Either remove it or build the return object explicitly with narrowed enum types.

gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  private readonly configuration = this.data.configuration ?? {};
+
+  readonly form = new FormGroup<OpenApiFormControls>({
+    viewer: new FormControl(this.configuration['viewer'] ?? 'Swagger', { nonNullable: true }),
+    entrypointsAsServers: new FormControl(parseBoolean(this.configuration['entrypointsAsServers']), { nonNullable: true }),
+    entrypointAsBasePath: new FormControl(parseBoolean(this.configuration['entrypointAsBasePath']), { nonNullable: true }),
+    tryItURL: new FormControl(this.configuration['tryItURL'] ?? '', {
+      nonNullable: true,
+      validators: [Validators.pattern(/^$|^https?:\/\/.+/)],
```
aqibmohammadkhan: Accepts `http://` (trailing whitespace allowed by .+). Also: when `entrypointsAsServers` is toggled on, the field is disabled, but `getRawValue()` still emits the previously typed URL, so the next reopen will re-populate it. 
i suggest to clear `tryItURL`  when disabling.

gravitee-apim-console-webui/src/components/documentation/gio-redoc/gio-redoc.component.ts
```diff
@@ -0,0 +1,102 @@
...
+          sidebar: { width: '150px' },
+          breakpoints: { medium: '599rem', large: '600rem' },
+        },
+      },
+      redocNode.nativeElement,
+      err => {
+        // eslint-disable-next-line angular/log
+        if (err) console.error('[gio-redoc] Redoc.init failed:', err);
+      },
+    );
```
aqibmohammadkhan: `gio-redoc.component.ts` parses with no type guard and re-renders on every keystroke.

- `parseSpec` returns unknown. If the user types a YAML scalar or null while editing, the spread throws. Add a guard: `if (typeof parsedSpec !== 'object' || parsedSpec === null) return`;.
- The `effect()` schedules a setTimeout(..., 0) per change of spec() (every keystroke in Monaco). Redoc init is expensive; debounce 200–300 ms.

gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  readonly isSwaggerViewer = computed(() => this.viewer() === 'Swagger');
+
+  private readonly tryItUrlStateEffect = effect(() => {
+    if (this.isEntrypointsAsServersEnabled()) {
+      this.form.controls.tryItURL.disable({ emitEvent: false });
+    } else {
+      this.form.controls.tryItURL.enable({ emitEvent: false });
+    }
+  });
```
aqibmohammadkhan: `OpenApiConfigDialogComponent` mixes toSignal + effect() more than the Angular rules suggest

side effects from observables should go through subscribe() + takeUntilDestroyed(destroyRef), not through a toSignal + effect() pair. Toggling enable/disable is a side effect, not state derivation — a plain subscribe is more idiomatic here. (The viewer signal is fine since it's read by a computed.)

gravitee-apim-console-webui/src/components/documentation/gio-redoc/gio-redoc.component.scss
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+:host {
+  display: block;
+  height: 100%;
+  overflow: auto;
```
aqibmohammadkhan: this here declares `:host { overflow: auto }` while `gio-swagger-ui` doesn't. If the parent flexbox cell already scrolls you'll get a double scrollbar.

gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
```diff
@@ -0,0 +1,143 @@
...
+  displayOperationId: FormControl<boolean>;
+  usePkce: FormControl<boolean>;
+  docExpansion: FormControl<string>;
+  enableFiltering: FormControl<boolean>;
+  showExtensions: FormControl<boolean>;
+  showCommonExtensions: FormControl<boolean>;
+  maxDisplayedTags: FormControl<number>;
+}
+
+function parseBoolean(value: string | boolean | undefined): boolean {
```
aqibmohammadkhan, line 53: `parseBoolean` accepts both `string` and `boolean`, but the input map is `Record<string, string>` — the boolean branch is dead. Simplify the signature.

gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
```diff
@@ -0,0 +1,143 @@
...
+  return value === 'true' || value === true;
+}
+
+function parseNumber(value: string | number | undefined): number {
+  const parsed = Number(value);
+  return isNaN(parsed) ? -1 : parsed;
```
aqibmohammadkhan: `parseNumber` returns `-1` for `empty/NaN`, so an unconfigured form opens with `-1` already in the `"max displayed tags"` field. Prefer null (and a nullable form control) so the field starts empty.

gravitee-apim-console-webui/src/portal/components/openapi-editor/openapi-editor.component.ts
```diff
@@ -30,16 +31,31 @@ import { GioSwaggerUiModule } from '../../../components/documentation/gio-swagge
...
+
+  swaggerTryItURL = computed(() => this.configuration()?.['tryItURL'] ?? '');
+  swaggerDocExpansion = computed(() => (this.configuration()?.['docExpansion'] ?? 'none') as 'none' | 'list' | 'full');
+  swaggerDisplayOperationId = computed(() => this.configuration()?.['displayOperationId'] === 'true');
+  swaggerFilter = computed(() => this.configuration()?.['enableFiltering'] === 'true');
+  swaggerShowExtensions = computed(() => this.configuration()?.['showExtensions'] === 'true');
+  swaggerShowCommonExtensions = computed(() => this.configuration()?.['showCommonExtensions'] === 'true');
+  swaggerMaxDisplayedTags = computed(() => {
+    const v = Number(this.configuration()?.['maxDisplayedTags']);
+    return !isNaN(v) && v >= 0 ? v : undefined;
```
aqibmohammadkhan: Number('') is 0, not NaN, so an unset value gets translated to 0 (hide all tags) instead of undefined. Guard with Number.isFinite(v) plus an explicit non-empty source check.

gravitee-apim-portal-webui-next/src/components/navigation-item-content-viewer/navigation-item-content-viewer.component.spec.ts
```diff
@@ -61,6 +61,7 @@ describe('NavigationItemContentViewerComponent', () => {
     const pageContent: PortalPageContent = {
       type: 'OPENAPI',
       content: 'openapi: 3.0.0\ninfo:\n  title: Test\n  version: 1.0.0',
+      configuration: { viewer: 'Redoc' as const },
```
aqibmohammadkhan: this only updates the existing `OPENAPI` case to force the Redoc branch (`configuration: { viewer: 'Redoc' as const }`). 

Add a sibling test that asserts the new `app-page-swagger` branch is taken when viewer is 'Swagger' (or unset). 

Same goes for `portal-navigation-items.component.spec.ts` — please cover the new "Configure" button opening the dialog and persisting the configuration.

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
MateuszJozwiak-Incubly: Solved in the backend PR.

gravitee-apim-console-webui/project.json
MateuszJozwiak-Incubly: The redo:2.4.0 is already added in package.json. Maybe it happened during some other changes.

gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
MateuszJozwiak-Incubly: Fixed earlier by using OpenApiViewer enum.

gravitee-apim-console-webui/src/portal/navigation-items/openapi-config-dialog/openapi-config-dialog.component.ts
MateuszJozwiak-Incubly: But for example the usage in showExtensions: new FormControl(parseBoolean(this.configuration.show_extensions), { nonNullable: true }), uses the show_extensions property which is declared as   show_extensions: boolean;, isn't it?

gravitee-apim-console-webui/src/services-ngx/portal-page-content.service.ts
```diff
@@ -40,4 +41,11 @@ export class PortalPageContentService {
   public updatePageContent(contentId: string, content: UpdatePortalPageContent): Observable<PortalPageContent> {
     return this.http.put<PortalPageContent>(`${this.constants.env.v2BaseURL}/portal-page-contents/${contentId}`, content);
   }
+
+  public updatePageContentConfiguration(contentId: string, configuration: OpenApiViewerConfiguration): Observable<PortalPageContent> {
+    return this.http.patch<PortalPageContent>(
+      `${this.constants.env.v2BaseURL}/portal-page-contents/${contentId}/configuration`,
+      configuration,
+    );
+  }
```
aqibmohammadkhan, line 50: The console now calls PATCH `${v2BaseURL}/portal-page-contents/{id}/configuration`. 

Can you confirm the backend endpoint for this is already merged on master (and cherry-picked to 4.12.x, given the apply-on-4-12-x label)? 
If the backend half of EXT-122 isn't in the same release, the Configure → Save flow will 404 in the portal console.

gravitee-apim-portal-webui-next/src/components/page/page.component.html
```diff
@@ -17,7 +17,7 @@
 -->
 @switch (page.type) {
   @case ('SWAGGER') {
-    @if (effectiveViewer.toLowerCase() === 'redoc') {
+    @if (effectiveViewer === viewerEnum.Redoc) {
```
aqibmohammadkhan, line 20: defaultType is typed as a plain string in configuration.ts, not constrained to ViewerEnum. If any environment has it stored as "redoc"/"REDOC" (legacy or hand-configured), the Redoc default will now silently fall back to Swagger — the old toLowerCase() handled that.

Please either 

(a) confirm defaultType is guaranteed to be exactly "Swagger"/"Redoc" at the source, or 
(b) normalize before comparing, e.g. compare effectiveViewer.toLowerCase() === ViewerEnum.Redoc.toLowerCase() (or coerce defaultType into ViewerEnum in effectiveViewer). A test with a lowercase defaultType would lock this down.

gravitee-apim-console-webui/src/services-ngx/portal-page-content.service.ts
MateuszJozwiak-Incubly: Master contains these changes, waiting for https://github.com/gravitee-io/gravitee-api-management/pull/17672 merge in order to have them on 4.12.x

jaroslawlakomy-grav (approved):
Approved with some reservations.
Given the size of this frontend change and the limited time available before the planned demo, I still think we should treat this area with extra caution, especially around the Redoc integration and end-to-end validation of the OpenAPI viewer configuration flow.
That said, after discussing the risk and delivery context with @kovaren , we agreed that merging this PR is needed to unblock the current milestone. I am approving it on that basis.
Follow-up recommendation: after merge, we should closely monitor the OpenAPI/Redoc flows, validate them again in the target environment, and address any regressions quickly in follow-up PRs.

PR #16743, opened 2026-05-05
fix: require UPDATE permission and enforce authenticated user on portal notification settings

araffass:
## Issue

https://gravitee.atlassian.net/browse/PEN-101

## Description

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/ApiNotificationSettingsResourceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/ApiNotificationSettingsResourceTest.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/ApiNotificationSettingsResourceTest.java
mukul-tyagi08, line 1: Class does not extend AbstractResourceTest (the established Jersey-test pattern in this module, ApiGroupsResourceTest, ApiMembersResourceTest). 
So, we can add a Jersey-style integration test extending AbstractResourceTest that PUTs to apis/{id}/notifications with permissionService.hasPermission(...UPDATE...) == false and asserts 403.
Wdyt?

araffass: Thanks Mukul,that's a fair point, and the reason i didn't go with AbstractResourceTest is that PortalNotificationConfigService isn't registered as a mock bean in ResourceContextConfiguration, so it would require touching that shared config on top of this fix.
I worked around it with a reflection test that asserts the annotation declares acls = UPDATE directly. So the test well covers the scope of the ticket.

 That said, do you think it's worth registering the service bean and adding a full Jersey 403 test here?

mukul-tyagi08: Fair point on ResourceContextConfiguration — touching it just for this fix is overkill.

mukul-tyagi08 (approved):

PR #16752, opened 2026-05-05
fix(console-ui): raise nginx request buffers for oversized auth payloads

oussamaelkhayali-collab:
## Problem

[JIRA Ticket](https://gravitee.atlassian.net/browse/TT-10981)

Some OAuth/OIDC flows can generate very large browser requests, especially when `id_token_hint` or similar values are carried in the request URI. In APIM, ingress-nginx may already be configured with larger request buffers, but the console UI image still serves traffic with its own baked nginx configuration. When that UI nginx configuration keeps smaller defaults, the request can still fail at the UI layer with `414 Request-URI Too Large` or `400` on oversized request headers.

## Solution

This PR updates the two baked console nginx templates so the UI image itself allows much larger request lines and headers:

- `gravitee-apim-console-webui/docker/config/default.conf`
- `gravitee-apim-console-webui/docker/config/default.no-ipv6.conf`

Both templates now set:

```nginx
large_client_header_buffers 4 64k;
```

That keeps the default and IPv4-only startup paths aligned and raises the UI-side ceiling well above the previous reproduction.

## Tests Done

Local validation was performed with the APIM console local stack:

1. Rebuilt the local `management_ui` image.
2. Recreated the local `management_ui` container.
3. Verified the active nginx config inside the running container contains `large_client_header_buffers 4 64k;`.
4. Sent a request with a roughly `60k` query string:
   - `GET /?id_token_hint=<~60k payload>`
   - Result: `200`
5. Sent a request with a roughly `60k` custom header:
   - `X-Long-Test: <~60k payload>`
   - Result: `200`

This confirms the local UI no longer reproduces the previous `414` / `400` behavior for the tested payload sizes.


gravitee-apim-console-webui/docker/config/default.conf
```diff
@@ -2,6 +2,7 @@ server {
     listen [::]:$HTTP_PORT ipv6only=off;
     listen [::]:$HTTPS_PORT ipv6only=off;
     server_name $SERVER_NAME;
+    large_client_header_buffers 4 64k;
```
gcusnieux, line 5: Why 64k specifically? 16k or 32k would already cover the 14k repro with margin, just curious about the value choice.

gcusnieux (approved):

gravitee-apim-console-webui/docker/config/default.conf
oussamaelkhayali-collab: I wanted to put a high value just to ensure we will not have the same issue in the future
I have seen some examples in our configurations of 32k and 64k, so I decided to put the highest value I have seen in our side

PR #16778, opened 2026-05-06
fix: stale security chain, orphan subscription for api-product

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-14025

## Description

Three separate bugs in the API Product feature, all reproducible after deleting an API Product with active subscriptions:

Bug 1 — Stale security chain (gateway)
After undeploying an API Product, member APIs continued serving 401 for Security (jwt/api-key,etc) requests. Root cause: `ApiProductManagerImpl#undeploy` emitted the `UNDEPLOY` event (which triggers a synchronous security chain refresh).

Fixed by swapping the two statements — registry removal first, event second.

Bug 2 — Orphan subscription cache entries (gateway)
After deletion, one of the member APIs would still resolve requests (false 200) because its per-API subscription entry was never evicted from`SubscriptionCacheService`. API Product subscriptions are *exploded* into one entry per member API in `cacheBySubscriptionIdAll`, but `unregister` was reading from `cacheBySubscriptionId` (single-slot, last-write-wins), so only the last-registered API's entry was cleaned up.

Fixed to use`cacheBySubscriptionIdAll` as the authoritative eviction source.

Bug 3 — Orphan subscription records in the database (rest-api)
Deleting an API Product did not cascade to its subscriptions. Those subscriptions remained in `ACCEPTED` state on the Application's  Subscriptions tab — orphan records with no corresponding API Product.

Fix: `DeleteApiProductUseCase` now does the same as API via the existing`CloseSubscriptionDomainService` and `DeleteSubscriptionDomainService`.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiProductManagerImpl.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiProductManagerImpl.java
@@ -141,10 +141,9 @@ public class ApiProductManagerImpl implements ApiProductManager {
         if (currentApiProduct != null) {
             log.debug("Undeploying API Product [{}] from environment [{}]", apiProductId, currentApiProduct.getEnvironmentId());
 
-            // Emit UNDEPLOY event before removal from registry
-            emitApiProductChangedEvent(ApiProductEventType.UNDEPLOY, currentApiProduct);
-
             apiProductRegistry.remove(apiProductId, currentApiProduct.getEnvironmentId());
+
+            emitApiProductChangedEvent(ApiProductEventType.UNDEPLOY, currentApiProduct);
 
             log.info("API Product [{}] has been undeployed", apiProductId);
         } else {
@@ -165,8 +164,12 @@ public class ApiProductManagerImpl implements ApiProductManager {
             eventManager.publishEvent(eventType, event);
             log.debug("Emitted {} event for API Product [{}]", eventType, apiProduct.getId());
         } catch (Exception e) {
-            log.warn("Failed to emit {} event for API Product [{}]", eventType, apiProduct.getId(), e);
-            // Don't fail the operation if event emission fails
+            log.error(
+                "Failed to emit {} event for API Product [{}] — security chain may be stale until next sync cycle",
+                eventType,
+                apiProduct.getId(),
+                e
+            );
         }
     }
 }
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
@@ -123,30 +123,46 @@ public class SubscriptionCacheService implements SubscriptionService {
 
     @Override
     public void unregister(final Subscription candidate) {
-        // only once per synchronization window
-        // take all fields (including "updatedAt" in metadata) into account
-        cacheBySubscriptionId.computeIfPresent(candidate.getId(), (h, existing) -> {
-            // Remove from exploded subscriptions cache
-            Set<Subscription> allSubscriptions = cacheBySubscriptionIdAll.get(existing.getId());
-            if (allSubscriptions != null) {
-                allSubscriptions.removeIf(s -> Objects.equals(existing.getApi(), s.getApi()));
-                if (allSubscriptions.isEmpty()) {
-                    cacheBySubscriptionIdAll.remove(existing.getId());
+        // Use compute() so that the isEmpty-check-then-remove is atomic with respect to
+        // concurrent register() calls on the same subscription ID. Without this, a register()
+        // interleaved between isEmpty()=true and remove() would add a new entry to the set
+        // and then have that set orphaned by the remove().
+        cacheBySubscriptionIdAll.compute(candidate.getId(), (id, allSubscriptions) -> {
+            if (allSubscriptions == null) return null;
+
+            var toEvict = allSubscriptions
+                .stream()
+                .filter(s -> Objects.equals(candidate.getApi(), s.getApi()))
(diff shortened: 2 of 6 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/use_case/DeleteApiProductUseCase.java
```diff
@@ -67,13 +74,26 @@ public void execute(Input input) {
...
     }
 
+    private void closeAndDeleteSubscriptions(String apiProductId, AuditInfo auditInfo) {
+        var subscriptions = subscriptionQueryService.findAllByReferenceIdAndReferenceType(
+            apiProductId,
+            SubscriptionReferenceType.API_PRODUCT
+        );
+        subscriptions.forEach(subscription -> {
+            closeSubscriptionDomainService.closeSubscription(subscription, auditInfo);
+            deleteSubscriptionDomainService.delete(subscription, auditInfo);
```
wbabyte: maybe add try/catch? single failure aborts the loop

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
```diff
@@ -123,31 +123,45 @@ public void register(final Subscription subscription) {
...
-                    cacheBySubscriptionIdAll.remove(existing.getId());
-                }
-            }
-            evictKeyForApi(existing.getApi(), existing.getId());
-            unregisterFromClientId(existing);
-            unregisterFromClientCertificate(existing);
+        // Use cacheBySubscriptionIdAll as the source so that each exploded per-API entry
+        // (e.g. from an API Product subscription) is cleaned up independently, regardless
+        // of which API was last written into the single-slot cacheBySubscriptionId.
+        // Collect all cached entries for this API before mutating to avoid ConcurrentModificationException.
```
wbabyte: New code does get → snapshot → mutate → isEmpty? → remove. The set is ConcurrentHashMap.newKeySet() precisely because concurrent register/unregister for sibling APIs of the same subscription ID is expected. A register(API_2) interleaved between isEmpty() and remove(id) can cause Thread A to drop the whole map entry containing B's freshly-registered subscription. Pre-PR computeIfPresent gave atomicity for free. Wrap the branch in cacheBySubscriptionIdAll.compute(...).

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiProductManagerImpl.java
```diff
@@ -141,11 +141,10 @@ private void undeploy(String apiProductId) {
         if (currentApiProduct != null) {
             log.debug("Undeploying API Product [{}] from environment [{}]", apiProductId, currentApiProduct.getEnvironmentId());
 
-            // Emit UNDEPLOY event before removal from registry
-            emitApiProductChangedEvent(ApiProductEventType.UNDEPLOY, currentApiProduct);
-
             apiProductRegistry.remove(apiProductId, currentApiProduct.getEnvironmentId());
 
+            emitApiProductChangedEvent(ApiProductEventType.UNDEPLOY, currentApiProduct);
```
wbabyte, line 146: After the reorder (registry-first, event-second), if event emission throws, the registry is already cleared but reactors never receive UNDEPLOY → stale security chain returns. Catch is Exception at warn. Narrow it, escalate to error, or compensate by re-adding to the registry.

wbabyte (changes requested):
Test coverage gaps for the documented failure modes
    - DeleteApiProductUseCaseTest.java uses unordered verify; doesn't pin close-before-delete. Add InOrder.
    - No test for closeSubscription throwing mid-loop (would lock in whatever cascade-failure semantic you choose for issue #1).
    - SubscriptionCacheServiceTest no test for clientId differing across exploded entries (the non-trivial noneMatch branch at lines 148-152).
    - No mid-state test (unregister(subApi1) then assert subApi2 still resolves)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/use_case/DeleteApiProductUseCase.java
vikrantgravitee: Let the code handle it at other places

wbabyte (commented):
The PR adds (SubscriptionCacheService.java:161-163):
  cacheBySubscriptionId.computeIfPresent(candidate.getId(), (k, existing) ->
      Objects.equals(existing.getApi(), candidate.getApi()) ? null : existing
  );
  "Only remove the single-slot entry if it points to the API we're unregistering."

  This does fix the case where you unregister the other leg (the test's chosen order):
  - unregister(sub1) → existing=sub2, APIs don't match → entry preserved ✓

  But it doesn't fix the case where you unregister the leg that happens to be in the single-slot map:
  - unregister(sub2) → existing=sub2, APIs match → entry removed
  - cacheBySubscriptionIdAll[SUB_ID] still has {sub1}
  - getById(SUB_ID) returns empty, getAllById(SUB_ID) returns [sub1]

  The two maps are now lying to each other about the same subscription.

  Why the test doesn't catch it

  should_keep_sibling_subscription_accessible_after_one_leg_unregistered does:
  register sub1 → register sub2 → unregister sub1
  That picks the favourable order — sub2 (the leg in the single slot) survives. And the assertions only check getAllById / getByApiAndClientIdAndPlan, never getById. Flip the order to unregister sub2, and getById(SUB_ID) silently returns empty even though
   sub1 is alive.

  The fix

  Don't just remove — rehydrate from the authoritative map if any leg remains:
  cacheBySubscriptionId.computeIfPresent(candidate.getId(), (k, existing) -> {
      if (!Objects.equals(existing.getApi(), candidate.getApi())) return existing;
      Set<Subscription> remaining = cacheBySubscriptionIdAll.get(candidate.getId());
      return (remaining == null || remaining.isEmpty()) ? null : remaining.iterator().next();
  });
  And add a regression test in the reverse order asserting getById(SUB_ID).isPresent().

wbabyte (approved):

PR #16696, opened 2026-04-30
fix: enable dynamic API metadata in GMD pages.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-14001

### Stacked PRs
- **master**
  - https://github.com/gravitee-io/gravitee-api-management/pull/16696 👈
    - https://github.com/gravitee-io/gravitee-api-management/pull/16779
      - https://github.com/gravitee-io/gravitee-api-management/pull/16780
        - https://github.com/gravitee-io/gravitee-api-management/pull/16781

### Portal navigation FreeMarker templating service

Introduces `PortalNavigationTemplatingService` and infra to render inline Gravitee Markdown with the same FreeMarker model pattern as legacy API docs: **`api`** when an enclosing API id is supplied, otherwise environment **`metadata`**.

### Scope

- Core port + Spring impl (`TemplateProcessor`, `ApiTemplateService`, `MetadataService`).
- Named inline template overload on `TemplateProcessor` / `FreemarkerTemplateProcessor` for clearer diagnostics.
- Exceptions: `PortalPageContentTemplateException`, `InvalidPortalPageContentTemplateException` (used by follow-up PRs).

No change yet to validators, GET flow, or console; later PRs wire this in.

PR #16779, opened 2026-05-06
feat: Resolve enclosing API for portal navigation pages.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-14001

### Stacked PRs
- **master**
  - https://github.com/gravitee-io/gravitee-api-management/pull/16696 
    - https://github.com/gravitee-io/gravitee-api-management/pull/16779 👈
      - https://github.com/gravitee-io/gravitee-api-management/pull/16780
        - https://github.com/gravitee-io/gravitee-api-management/pull/16781

###  Enclosing API resolution for portal navigation pages

Adds navigation helpers so templating can tell whether a portal page lives under an API (for `api` in FreeMarker) vs environment-only context (for `metadata`).

### Scope

- `PortalNavigationEnclosingApiDomainService`: walks ancestors to find enclosing API id.
- `PortalNavigationItemsQueryService`: default `findNavigationPageByPortalPageContentId`.
- Unit tests for domain service and query implementation.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationEnclosingApiDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationEnclosingApiDomainService.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal_page.domain_service;
+
+import io.gravitee.apim.core.DomainService;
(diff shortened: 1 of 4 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationEnclosingApiDomainService.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Resolves the API that scopes documentation templates for a navigation item by walking {@link PortalNavigationItem#getParentId()}.
+ */
+@DomainService
+@RequiredArgsConstructor
+public class PortalNavigationEnclosingApiDomainService {
+
+    private final PortalNavigationItemsQueryService queryService;
+
+    public Optional<String> findEnclosingApiId(String environmentId, PortalNavigationItem item) {
```
dev-marek, line 34: we should think about something not causing N+1 problem

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationEnclosingApiDomainService.java
kovaren: I agree, but the same parent-walking pattern already exists in PortalNavigationApiVisibilityDomainService and PortalNavigationItemDomainService.

I'd argue that we should address the N+1 as a whole across all use cases. I can create a tech task to track this as a follow-up.

Also, in practice the tree depth is pretty shallow (1-5 levels) so the impact is low for now.

WDYT?

aqibmohammadkhan (approved):

dev-marek (approved):

PR #16785, opened 2026-05-06
Guard null HttpSelector.path/pathOperator and scalar Step.configuration to return 400

carlos-andres-osorio:
## Issue

[APIM-13983](https://gravitee.atlassian.net/browse/APIM-13983)

## Why

Three unguarded null/type paths in `PUT` and `POST /management/v2/environments/{envId}/apis` caused 500 NullPointerExceptions instead of 400 validation errors:

- `HttpSelector.path = null` — null propagated to a path-param check without a prior guard
- `HttpSelector.pathOperator = null` — `FlowOperator.name()` called on null during selector persistence
- `Step.configuration` is a scalar (e.g. `"a-string"`) — the value bypassed the type check and reached a JSON schema validator that expects an object, causing an NPE

All three inputs are already forbidden by the OpenAPI spec; the fix enforces the spec at the validation layer and returns 400 with a field-level error in each case.

## What changed

- **`FlowValidationDomainService`** — `HTTP_PATH_EXTRACTOR` now filters null paths instead of forwarding them; `validatePathParameters` materialises both input streams to lists upfront so they can be iterated twice (null-path scan + param-overlap check) without an `IllegalStateException`; `validateInvalidSelectors` pipelines directly without an intermediate `.toList()`
- **`ConfigurationSerializationMapper`** — `serializeConfiguration` now validates that `String` configurations are well-formed JSON objects and rejects any non-`LinkedHashMap`, non-`String` type with `InvalidDataException` (400) instead of forwarding invalid input downstream
- **`FlowMapper`** — added import for `InvalidDataException`; replaced the inline fully-qualified reference
- **`JsonSchemaServiceImpl`** — removed dead `catch (NullPointerException e)` block; the upstream mapper fix prevents invalid input from reaching the validator

## How to test

Against a running Management API, send a `PUT` or `POST /management/v2/environments/DEFAULT/apis` payload with any of:

1. `"flows": [{"selectors": [{"type": "HTTP", "path": null, "pathOperator": "STARTS_WITH"}]}]` → expect `400`
2. `"flows": [{"selectors": [{"type": "HTTP", "path": "/", "pathOperator": null}]}]` → expect `400`
3. A step with `"configuration": "a-string"` (scalar, not an object) → expect `400`

Before this fix each of the above returned `500` with a Java NPE in the body.

[APIM-13983]: https://gravitee.atlassian.net/browse/APIM-13983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ConfigurationSerializationMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ConfigurationSerializationMapper.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.gravitee.definition.jackson.datatype.GraviteeMapper;
 import io.gravitee.node.logging.NodeLoggerFactory;
+import io.gravitee.rest.api.service.exceptions.InvalidDataException;
 import io.gravitee.rest.api.service.exceptions.TechnicalManagementException;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ConfigurationSerializationMapper.java
```diff
@@ -40,16 +41,26 @@ default String serializeConfiguration(Object configuration) {
         if (Objects.isNull(configuration)) {
             return null;
         }
+        ObjectMapper mapper = new GraviteeMapper();
```
paulatulis: ObjectMapper (and by extension GraviteeMapper) is thread-safe and expensive to construct — Jackson's own docs recommend it as a static final singleton. The pre-PR code already had this problem in the `LinkedHashMap` branch; this PR widened its scope by hoisting it above the branch, meaning the cost is now paid even for the null fast-exit path (no — null returns early, fine) but also for every valid call. 

Fix: Promote to a static final field — the interface already has a pattern for this (see JSON_MAPPER on the other FlowMapper in the REST layer):

carlos-andres-osorio: Promoted `GraviteeMapper` to a static final interface field `JSON_MAPPER` (line 37), matching the pattern already used in `FlowMapper`. Both `serializeConfiguration` and `deserializeConfiguration` now share the single instance instead of constructing one per call. The `ObjectMapper` import was removed as it's no longer referenced directly.

paulatulis (approved):
Looks great!

PR #16788, opened 2026-05-06
feat(management-api): allow PATCH flows on v4 HTTP Proxy APIs

carlos-andres-osorio:
## Issue

[APIM-13981](https://gravitee.atlassian.net/browse/APIM-13981)

## Why

`PATCH /apis/{apiId}` currently rejects any request that includes a `flows` field, returning 400 "use PUT to update structural fields". Operators need to update policies and routing flows without replacing the entire API definition. This PR lifts that restriction so `flows` is treated like any other patchable field — both JSON Merge Patch and JSON Patch semantics are supported, and the post-patch state flows through the same validator and audit path that PUT already uses.

## What changed

- **`PatchApiUseCase.java`**: removed `flows` from `BLOCKED_WITHOUT_HINT`; added `FIELD_FLOWS` to `ALLOWED_FIELDS`; added `List<Flow> flows` projection to `PatchableView`; introduced a generic `resolvePatchableList<T>` helper (five-branch resolution: merge-patch null erase, merge-patch present replace, JSON Patch remove, JSON Patch replace, untouched fall-through); enriched `readList` to emit a `ValidationDomainException` with `invalidValue` technical code and a field-path `location` on deserialization failures; wired resolved flows onto the rebuilt v4 HTTP definition via `httpV4.toBuilder().flows(...)`.
- **`openapi-apis.yaml`**: added `flows` to the patchable-fields sentence in the `PATCH /apis/{apiId}` operation description.
- **`ApiResource_PatchApiFlowsTest`** (new, 57 tests): integration tests covering array-, step-, and selector-level scenarios across both content types, validator-rejection envelope, deserialization error envelope, permission gate, audit-path delegation, and `flowExecution` preservation.
- **`PatchApiUseCaseTest`** (extended, +9 `FlowsResolution` tests): unit tests for every resolve branch; removed three stale rejection tests that expected `flows` to be blocked.

## User-facing impact

`PATCH /apis/{apiId}` now accepts `flows` for v4 HTTP Proxy APIs with both `application/json-patch+json` and `application/merge-patch+json`. Previously any PATCH request including `flows` returned 400; valid flows patches now return 200. This is purely additive — no existing behavior is changed. OpenAPI spec updated in this PR.

## How to test

```
curl -X PATCH -H "Authorization: Bearer <token>" -H "Content-Type: application/merge-patch+json" "http://localhost:8083/management/v2/organizations/DEFAULT/environments/DEFAULT/apis/<apiId>" -d '{"flows":[{"name":"my-flow","enabled":true,"request":[],"response":[],"selectors":[{"type":"HTTP","path":"/","pathOperator":"STARTS_WITH","methods":["GET"]}]}]}'
```

Expected: 200 with updated `flows` in the response body. A follow-up GET confirms the flows are persisted. Omitting `flows` from a subsequent PATCH leaves them untouched.

## Gotchas / Follow-up

- This PR is rebased on top of APIM-13983 (NPE guard for null `HttpSelector.path`/`pathOperator` and scalar `Step.configuration`). It must merge after that PR.
- `flows` is wired on `httpV4.toBuilder()` (the API definition value), not `existingApi.toBuilder()` (the domain entity) — flows are not a top-level entity field.
- `@JsonSubTypes` discriminator labels on `Selector` are lowercase (`http`, `condition`); uppercase payloads are rejected by Jackson.
- Test cases for `null HttpSelector.path`/`pathOperator` and scalar `Step.configuration` are deferred to APIM-13983.
- Dry-run sanitization copy-back for `flows` is covered by APIM-13982, not this PR.

[APIM-13981]: https://gravitee.atlassian.net/browse/APIM-13981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiFlowsTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiFlowsTest.java
@@ -1,0 +1,322 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.management.v2.rest.resource.api;
+
+import static assertions.MAPIAssertions.assertThat;
+import static io.gravitee.common.http.HttpStatusCode.BAD_REQUEST_400;
+import static io.gravitee.common.http.HttpStatusCode.FORBIDDEN_403;
+import static io.gravitee.common.http.HttpStatusCode.OK_200;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.eq;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.when;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import fixtures.core.model.ApiFixtures;
+import inmemory.InMemoryAlternative;
+import io.gravitee.apim.core.api.domain_service.UpdateApiDomainService;
+import io.gravitee.apim.core.exception.ValidationDomainException;
+import io.gravitee.apim.core.membership.model.Membership;
+import io.gravitee.apim.core.membership.model.PrimaryOwnerEntity;
+import io.gravitee.definition.model.DefinitionVersion;
+import io.gravitee.definition.model.v4.ApiType;
+import io.gravitee.definition.model.v4.flow.Flow;
+import io.gravitee.definition.model.v4.flow.step.Step;
+import io.gravitee.rest.api.management.v2.rest.model.ApiV4;
+import io.gravitee.rest.api.model.permissions.RolePermission;
+import io.gravitee.rest.api.model.permissions.RolePermissionAction;
+import io.gravitee.rest.api.model.v4.api.ApiEntity;
+import io.gravitee.rest.api.service.common.GraviteeContext;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.client.Entity;
+import java.time.Instant;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Stream;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayNameGeneration;
+import org.junit.jupiter.api.DisplayNameGenerator;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
+public class ApiResource_PatchApiFlowsTest extends ApiResourceTest {
+
+    private static final String MERGE_PATCH_TYPE = "application/merge-patch+json";
+    private static final String JSON_PATCH_TYPE = "application/json-patch+json";
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    @Inject
+    UpdateApiDomainService updateApiDomainService;
+
+    @Override
+    protected String contextPath() {
+        return "/environments/" + ENVIRONMENT + "/apis";
+    }
+
+    @BeforeEach
+    void setUpApiAndPrimaryOwner() {
+        givenApiWithFlows(List.of());
+
+        var apiEntity = new ApiEntity();
+        apiEntity.setId(API);
+        apiEntity.setDefinitionVersion(DefinitionVersion.V4);
+        apiEntity.setType(ApiType.PROXY);
+        apiEntity.setUpdatedAt(Date.from(Instant.ofEpochMilli(1000)));
+        when(apiSearchServiceV4.findGenericById(any(), eq(API), any(boolean.class), any(boolean.class), any(boolean.class))).thenReturn(
+            apiEntity
+        );
+
+        roleQueryService.resetSystemRoles(ORGANIZATION);
+        primaryOwnerDomainService.initWith(
+            List.of(Map.entry(API, PrimaryOwnerEntity.builder().id(USER_NAME).type(PrimaryOwnerEntity.Type.USER).build()))
+        );
+        membershipQueryServiceInMemory.initWith(
+            List.of(
+                Membership.builder()
+                    .memberId(USER_NAME)
+                    .referenceId(API)
+                    .roleId("api-po-id-" + ORGANIZATION)
+                    .referenceType(Membership.ReferenceType.API)
+                    .memberType(Membership.Type.USER)
+                    .build()
+            )
+        );
+
+        doAnswer(inv -> inv.getArgument(0))
+            .when(updateApiDomainService)
+            .updateV4(any(), any());
+        doAnswer(inv -> inv.getArgument(0))
+            .when(updateApiDomainService)
+            .validateV4(any(), any());
+    }
+
+    @AfterEach
+    public void tearDown() {
+        Stream.of(apiCrudService, membershipQueryServiceInMemory, primaryOwnerDomainService).forEach(InMemoryAlternative::reset);
+        reset(updateApiDomainService, apiSearchServiceV4);
+    }
+
+    private void givenApiWithFlows(List<Flow> flows) {
+        var base = ApiFixtures.aProxyApiV4();
+        var api = base.toBuilder().apiDefinitionValue(base.getApiDefinitionHttpV4().toBuilder().flows(flows).build()).build();
+        apiCrudService.initWith(List.of(api));
+    }
+
+    private static Step step(String name, String policy) {
+        return Step.builder().name(name).policy(policy).enabled(true).build();
+    }
+
+    private static Flow flow(String name, List<Step> request) {
+        return Flow.builder().name(name).enabled(true).request(request).build();
+    }
+
+    private static String mergePatchFlows(List<Map<String, Object>> flows) {
+        return "{\"flows\":" + toJson(flows) + "}";
+    }
+
+    private static String jsonPatchReplaceFlows(List<Map<String, Object>> flows) {
+        return "[{\"op\":\"replace\",\"path\":\"/flows\",\"value\":" + toJson(flows) + "}]";
+    }
+
+    private static Stream<Arguments> bothFlowsReplaceVariants(List<Map<String, Object>> flows) {
+        return Stream.of(
+            Arguments.of(mergePatchFlows(flows), MERGE_PATCH_TYPE),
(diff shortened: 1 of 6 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -453,38 +470,33 @@ private boolean isExplicitNull(PatchType patchType, JsonNode rawPatchNode, Strin
         return patchType == PatchType.MERGE_PATCH && rawPatchNode.has(field) && rawPatchNode.get(field).isNull();
     }
 
-    private boolean isRemovedByJsonPatch(PatchType patchType, JsonNode rawPatchNode, String field) {
-        if (patchType != PatchType.JSON_PATCH) {
+    private boolean isNullifiedByJsonPatch(JsonNode rawPatchNode, String field) {
```
paulatulis, line 475: The method iterates all ops and returns true on the first match for a given field path. JSON Patch is a sequential array; a valid document could have:

```
json
[
  {"op": "replace", "path": "/flows", "value": [{"name": "f1", ...}]},
  {"op": "remove", "path": "/flows"}
]
```

The current scan sees the first replace op, evaluates it, and stops (it wouldn't match remove semantics since the value is not null). But the actual result after Jackson's JSON Patch application is that /flows is removed. The `resolvePatchableList` branch then falls through to return existing rather than return null. Flows are not erased as the caller intended.

The root cause is that `isNullifiedByJsonPatch` is used as a read-ahead on the raw patch node, but the applied result (patchedNode) is already the ground truth. The correct signal for "was the field removed?" for patchedNode is simply:

```
java
// patchedNode.get(field) == null after application means the field no longer exists
```
This is partially handled in resolvePatchableList (checking patchedNode.get(field) == null), but the isNullifiedByJsonPatch guard on that branch means a non-nullifying remove that leaves the key absent (e.g., removing the entire array when it previously had a value) could still fall through to return existing.

Impact: Multi-op JSON Patch documents that first replace then remove flows may silently leave existing flows intact.

Recommendation: Trust patchedNode as the source of truth for field presence. The correct logic for resolvePatchableList when patchType == JSON_PATCH is:
```
java
if (patchedNode.get(field) == null) {
    // field was removed by the patch OR was never present; check rawPatchNode only to distinguish "patched away" vs "not touched"
    return isFieldTouchedByJsonPatch(rawPatchNode, field) ? null : existing;
}
// field exists in patchedNode — deserialize it
return readList(patchedNode, field, elementType, existing);
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -630,10 +643,33 @@ private <T> List<T> readList(JsonNode node, String field, Class<T> elementType,
             return defaultValue;
         }
         try {
-            return objectMapper.convertValue(fieldNode, objectMapper.getTypeFactory().constructCollectionType(List.class, elementType));
+            return objectMapper
```
paulatulis, line 648: FAIL_ON_INVALID_SUBTYPE flag is set after the type reference is bound — the feature may not apply

```
return objectMapper
    .<List<T>>readerFor(objectMapper.getTypeFactory().constructCollectionType(List.class, elementType))
    .with(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE)
    .readValue(fieldNode);
```
DeserializationFeature.FAIL_ON_INVALID_SUBTYPE affects type resolution during polymorphic deserialization. However, Jackson's ObjectReader.with(DeserializationFeature) returns a new reader with the feature enabled — this is correct API usage. But: by the time the feature is applied, the readerFor(...) call has already bound the root type to List<T>, not Selector. The @JsonSubTypes discriminator resolution for Selector subtypes happens inside Jackson's normal deserialization, which does respect FAIL_ON_INVALID_SUBTYPE.

This is actually functionally correct for the Flow → Selector path because Flow.selectors is List<Selector> and Jackson will recurse into it. However, the feature is redundant for the top-level collection: a List<Flow> doesn't itself involve a subtype discriminator. The real subtype guard here is that Jackson's @JsonTypeInfo / @JsonSubTypes on Selector will throw a InvalidTypeIdException (a JsonMappingException) when an unknown discriminator is seen, regardless of this feature flag.

Risk: The feature flag may give a false sense of security. If Gravitee ever adds a new Selector subtype without updating @JsonSubTypes, the behavior depends entirely on Jackson's defaultImpl configuration on the annotation. Confirm that @JsonSubTypes on Selector has no @JsonTypeInfo(defaultImpl = ...) fallback — if it does, unknown discriminators silently deserialize to the default, bypassing the 400.

Recommendation: Add an explicit test for an unknown Selector type discriminator value and verify it returns 400 with the invalidValue envelope. (The PR includes this test — confirm it passes and doesn't rely on the FAIL_ON_INVALID_SUBTYPE flag being the deciding guard.)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -367,22 +381,29 @@ private Api applyPatchedValues(Api existingApi, JsonNode patchedNode, PatchType
             .build();
     }
 
-    private List<Property> resolveProperties(PatchType patchType, JsonNode rawPatchNode, JsonNode patchedNode, List<Property> existing) {
-        if (patchType == PatchType.MERGE_PATCH && rawPatchNode.has(FIELD_PROPERTIES)) {
-            if (rawPatchNode.get(FIELD_PROPERTIES).isNull()) {
+    private <T> List<T> resolvePatchableList(
```
paulatulis, line 386: `resolvePatchableList` passes `null` as `defaultValue` to `readList` in the merge-patch branch — but existing is ignored

When patchedNode.get(field) is null (which shouldn't happen in a valid merge-patch that passed the null check above, but is possible with a deeply nested merge-patch where Jackson produces a null node), readList will return the defaultValue, which is hardcoded as null instead of existing. This means an unexpected null node would clear flows silently rather than preserving them.

The correct call should pass existing as the default so the "not found, fall through" path returns the existing value. The current code is actually a latent regression from the resolveProperties refactor.

Severity: Low-medium — requires a specific edge case to trigger, but is a behavioral inconsistency.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -630,10 +643,33 @@ private <T> List<T> readList(JsonNode node, String field, Class<T> elementType,
             return defaultValue;
         }
         try {
-            return objectMapper.convertValue(fieldNode, objectMapper.getTypeFactory().constructCollectionType(List.class, elementType));
+            return objectMapper
+                .<List<T>>readerFor(objectMapper.getTypeFactory().constructCollectionType(List.class, elementType))
+                .with(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE)
```
paulatulis, line 650: FAIL_ON_INVALID_SUBTYPE is called on a per-invocation ObjectReader but the base objectMapper is shared and mutable
The shared objectMapper is Spring-injected. The ResourceContextConfiguration.java change registers a @Bean @Primary GraviteeMapper for tests, which is correct. But in production, the objectMapper injected into PatchApiUseCase must also be GraviteeMapper (which carries the polymorphic type info for Selector, Flow, etc.).

Potential issue: If the production ObjectMapper bean is not GraviteeMapper (or a derivative), Flow.selectors deserialization will fail with a generic 400 rather than properly polymorphic deserialization. The test configuration override masks this in tests.

Recommendation: Verify the production Spring context registers GraviteeMapper as the primary ObjectMapper. Add a comment in readList() noting that GraviteeMapper with @JsonSubTypes on Selector is a prerequisite.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiFlowsTest.java
paulatulis, line 1: Low priority: non-HTTP v4 API types are not tested in ApiResource_PatchApiFlowsTest

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: Added `json_patch_replace_then_remove_flows_clears_flows` in `PatchApiUseCaseTest` — a two-op patch `[replace(non-null), remove]` that asserts the resulting flows are null. The loop in `isNullifiedByJsonPatch` does not short-circuit on the first matching op: a non-nullifying `replace` falls through without returning, so the subsequent `remove` op is reached and returns `true`. Combined with `patchedNode.get(field) == null` (the applied result as ground truth), the method correctly returns `null`. No production change needed; the test documents and guards the correct multi-op behaviour.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: `ObjectReader.with(DeserializationFeature)` is correct Jackson API — it returns a new reader with the feature enabled and the feature applies to nested deserialization including `Selector` subtypes within `Flow.selectors`. The `Selector` annotation is `@JsonTypeInfo(use=NAME)` with no `defaultImpl` fallback, so Jackson always throws `InvalidTypeIdException` on an unknown discriminator regardless of this flag. The test `patch_with_unknown_selector_type_returns_400_with_invalidValue_envelope_and_field_path_location` confirms the 400 behaviour end-to-end. No change to the flag or the test.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: Fixed: changed `readList(patchedNode, field, elementType, null)` to `readList(patchedNode, field, elementType, existing)` in the merge-patch branch. In the normal path `patchedNode` always contains the field (a non-null merge-patch value guarantees this), so the default is never reached — but if it were, returning `null` would silently erase flows rather than preserve them. Passing `existing` as the default eliminates that inconsistency.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: Verified: `ServiceConfiguration.java` registers `new GraviteeMapper(false)` with `@Primary`, so the production injected `ObjectMapper` is always `GraviteeMapper`. On the test side, `PatchApiUseCaseTest` was using `JsonMapperFactory.build()` (a plain `JsonMapper`) — switched it to `new GraviteeMapper(false)` so the unit test mapper matches the production bean and any divergence in polymorphic type handling for `Flow`/`Selector` would be caught rather than masked.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiFlowsTest.java
carlos-andres-osorio: `PatchApiUseCase` rejects any non-HTTP Proxy API before reaching the flows resolution path — `ApiType.PROXY` is checked at the entry point and throws `ApiInvalidTypeException` otherwise. This is already covered by `type_field_is_rejected` in `PatchApiUseCaseTest` and by the type-guard integration tests. No flows-specific tests for other API types are needed in `ApiResource_PatchApiFlowsTest`.

paulatulis (approved):
LGTM!

PR #16802, opened 2026-05-07
fix: remove default 10 item pagination.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-14013

### Description

Portal API search treated a missing pageable as page size 10, so Lucene-backed search could return at most ten IDs before any higher-level pagination ran.

#### ApiPortalSearchQueryServiceImpl

When `pageable` is empty, the service returns all IDs that survive the Lucene and allow-list intersection, with page number forced to 1 and no implicit cap of ten. When `pageable` is present, page size and offsets are taken from it; invalid or empty pages still yield empty content with the correct totals.


dev-marek (changes requested):
please change page calculations according to gemini's comments

dev-marek (approved):

dev-marek (approved):

aqibmohammadkhan (approved):

PR #16803, opened 2026-05-07
feat: fuzzy search in developer portal.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-14015

### Description

Developer portal API search was strict on spelling, so small typos in the catalog or navigation search often returned no useful matches. This change adds an environment-scoped setting to turn on approximate spelling (fuzzy matching) for those portal searches, wired from the management console through configuration and the portal REST layer into Lucene.

Fuzzy matching is applied only for free-text portal queries, with guards on token length, query length, and how many fuzzy clauses are added per request so short tokens and very large queries stay predictable and Lucene does not blow up on clause limits. When navigation search filters by query, results are ordered by Lucene relevance first, then by the original navigation order for a stable tie-break.

#### Management console (portal settings)

Adds a developer portal form control labeled approximate spelling for API search, bound to `portalNext.search.fuzzy`, with types and fixtures updated so the setting round-trips with the rest of portal-next settings.

#### Configuration and REST API surface

Introduces the `portal.next.search.fuzzy` parameter key (default false), a `PortalNext.Search` section on the settings model, and exposes search in the portal configuration assembly. Portal OpenAPI descriptions for API search and portal navigation search note that minor typos may match when fuzzy is enabled.

#### Portal REST and core use cases

`ApisResource` and `PortalNavigationItemsResource` read the environment boolean and pass it into `SearchApisForPortalUseCase` and `GetVisiblePortalNavigationApisUseCase`. The portal search query service and its Lucene implementation forward typo tolerance into `ApiSearchService.searchIds`. `GetVisiblePortalNavigationApisUseCase` re-sorts visible navigation APIs after a text query by search rank, then by prior navigation order.

#### Lucene search engine

`ApiDocumentSearcher` adds optional `FuzzyQuery` SHOULD clauses across boosted name, description, paths, labels, owner, and host fields for eligible tokens; explicit filter parsing keeps fuzzy off for structured clauses. `Query` / `QueryBuilder` and `ApiSearchServiceImpl` carry the typo-tolerance flag into the search engine query.

#### Tests

Updates existing portal resource, use case, config, and search tests for the new parameter and constructor inputs; adds focused fuzzy title coverage and related search-engine tests.


gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/parameters/Key.java
```diff
@@ -160,6 +160,8 @@ public enum Key {
         Boolean.FALSE.toString(),
         new HashSet<>(singletonList(ENVIRONMENT))
     ),
+    PORTAL_NEXT_INVITATIONS_ENABLED("portal.next.invitations.enabled", Boolean.FALSE.toString(), new HashSet<>(singletonList(ENVIRONMENT))),
```
kovaren: This change doesn't seem relevant to this PR.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/PortalNext.java
```diff
@@ -155,4 +158,12 @@ public static class Catalog {
...
         private String viewMode;
     }
+
+    @Data
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    public static class Search {
+
+        @ParameterKey(Key.PORTAL_NEXT_SEARCH_FUZZY)
+        private Boolean fuzzy;
+    }
```
kovaren: WDYT about nesting this setting inside `Catalog`?

dev-marek (approved):
lgtm

PR #16805, opened 2026-05-07
chore: reject null pathOperator in V4 HTTP flow selector validation

carlos-andres-osorio:
## Jira ticket

APIM-13983

## Why

A V4 HTTP flow selector with a null `pathOperator` would bypass the existing path validation and reach `FlowOperator.valueOf(null)` downstream, producing a `NullPointerException` instead of a clean 400. The `path` field already had a null guard; this extends the same pattern to `pathOperator`.

## What changed

- `FlowValidationDomainService`: added a null check for `pathOperator` inside `checkForInvalidSelectors`, throwing `InvalidDataException("flows[].selectors[].pathOperator is required")` before any further processing.
- `FlowValidationDomainServiceTest`: added a test asserting that a flow with `pathOperator: null` raises `InvalidDataException`.

## How to test

POST or PUT a V4 proxy API with a flow whose HTTP selector omits `pathOperator` (sends `null`). Expect a `400 Bad Request` response instead of a `500`.

Existing tests pass: `mvn -pl gravitee-apim-rest-api/gravitee-apim-rest-api-service test -Dskip.validation -Dtest=FlowValidationDomainServiceTest`

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/flow/domain_service/FlowValidationDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/flow/domain_service/FlowValidationDomainService.java
@@ -231,14 +231,17 @@ public class FlowValidationDomainService {
     }
 
     private static void checkForInvalidSelectors(List<Flow> apiFlowList, List<Flow> planFlowList) {
-        Stream.concat(apiFlowList.stream(), planFlowList.stream())
+        var httpSelectors = Stream.concat(apiFlowList.stream(), planFlowList.stream())
             .flatMap(flow -> flow.selectorByType(SelectorType.HTTP).stream())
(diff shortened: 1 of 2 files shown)


paulatulis (approved):
Approved.
review tool also mentioned as semi-related: `checkForInvalidSelectors` is still only called by `validatePathParameters()`, meaning any API creation routed through `ValidateApiDomainServiceLegacyWrapper` (which only calls `validateAndSanitizeHttpV4()`) will still let a `null` path or `pathOperator` slip through to the persistence layer and cause a 500.

carlos-andres-osorio:
> Approved. review tool also mentioned as semi-related: `checkForInvalidSelectors` is still only called by `validatePathParameters()`, meaning any API creation routed through `ValidateApiDomainServiceLegacyWrapper` (which only calls `validateAndSanitizeHttpV4()`) will still let a `null` path or `pathOperator` slip through to the persistence layer and cause a 500.

Good catch. Created https://gravitee.atlassian.net/browse/APIM-14022 to keep track of this

PR #16807, opened 2026-05-07
feat: improve log filtering and add API product resolution

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-13978
## Description

Takes care of several small things
---

### `GetFilterValuesUseCase.java`

**What changed:**
1. `API_PRODUCT` is now always routed through `handleIdBasedSearch` (DB query), even when there is no search text — previously it required a non-empty query string, which meant the "browse all" dropdown showed nothing.
2. `searchApiProductsByNameFromEnvironment` fixed to handle a null/blank query (the browse-all path) without NPE — previously assumed `query` was always non-null.

**Why:** API Product IDs are not indexed in the Elasticsearch analytics log records (unlike API IDs), so querying ES for them would always return empty. The DB must always be the source of truth for API Product filter values.

---

###  `ResolveFilterLabelsUseCase.java`

**What changed:**  
`API_PRODUCT` added to `RESOLVABLE_FILTERS`.

**Why:** Without this, when the dashboard resolves filter chip display names after a page reload or share-link load, it would silently skip API_PRODUCT and show the raw UUID in the chip instead of the product name.

---

### `GetFilterValuesUseCaseTest.java`

Added test coverage for:
- Browse-all path for `API_PRODUCT` (no query string) returns products from DB
- Search-by-name path for `API_PRODUCT` correctly filters by name substring
- Products with null names are excluded from results

---

###  `ResolveFilterLabelsUseCaseTest.java`

Added test coverage for:
- `API_PRODUCT` filter values are resolved to display names via `FilterValueNameResolver`

---

###  `dashboard-filters.store.ts`

**What changed:**  
`mergeResolvedLabels` now also merges the **field-level label** (the filter's display name, e.g. "API Product") from resolved conditions, not just the per-value labels (the selected values' display names).

**Why:** For ID-based filters like `API_PRODUCT`, the resolved condition carries both the field label and the value labels. Without this fix, after resolution the filter chip's header label could remain a stale default while the value labels updated correctly — visible when a condition is re-resolved after a time-range change.

---

### `observability-filters-api.service.ts`

**What changed:**  
Guard changed from `if (query.query != null && query.query !== '')` → `if (query.query)`.

**Why:** Falsy-check is simpler and correctly handles all empty/null/undefined cases. No behavioral difference; a minor code quality fix discovered during this investigation.

---

### `env-logs-table.component.ts` + `.html`

**What changed:**  
Added `isLoading = input<boolean>(false)` signal input and passed it through from the parent.

**Why:** The table uses `*matNoDataRow` to show an empty-state message. Without a loading flag, during the initial fetch the table would flash "No data to display" before results arrived, which is misleading. This wires up the loading state so the parent can control what's shown.

---

###  `env-logs.component.html`

**What changed:**  
`[isLoading]="loading()"` bound on the `<app-env-logs-table>`.

**Why:** Completes the loading propagation chain so the table knows when a fetch is in progress.

---

###  `env-logs.component.spec.ts`

Added test cases covering the `isLoading` input propagation.

---

### `filter.model.ts`

**What changed:**  
`API_PRODUCT` added to `ID_BASED_FILTER_NAMES`.

```diff
- export const ID_BASED_FILTER_NAMES = ['API', 'APPLICATION', 'PLAN'];
+ export const ID_BASED_FILTER_NAMES = ['API', 'APPLICATION', 'PLAN', 'API_PRODUCT'];
```

**Why:** `ID_BASED_FILTER_NAMES` tells the keyword-value-input component to treat selected values as opaque IDs and resolve their labels via the name-resolution API. Without this, the filter chip would show the raw UUID. With it, the component fetches the product name and shows "My Test Product" instead.

---

### `keyword-value-input.component.ts`

Three separate fixes:

**1. Empty-state hint text**  
- Before: Always showed `"No matching values"` when the list was empty  
- After: Shows `"Type to search…"` when no search text has been entered, `"No matching values"` only when there is text with no results  
- **Why:** For API_PRODUCT (and other ID-based filters), values are only loaded when the user types — showing "No matching values" before typing is incorrect and confusing.

**2. Clear search input after picking a value**  
- Before: Conditionally emitted a change event when clearing, causing a spurious re-fetch after each selection  
- After: Always uses `emitEvent: false` to suppress the re-fetch, and uses `queueMicrotask` to defer the native input clear until after Material's `_setTriggerValue` synchronously writes the selected option label into the `<input>` (which would otherwise overwrite the clear)  
- **Why:** Fixes a race condition where the typed search text would linger in the input after selection, or a redundant API call would fire.

**3. Actually calls `clearSearchInputAfterPick`**  
- Before: The method existed but was not called from `pickValue`  
- After: Called unconditionally in `pickValue` for both single- and multi-select paths  
- **Why:** The clear was never actually running for the multi-select overlay path.

---

###  `keyword-value-input.component.spec.ts`

Added test coverage for:
- Empty-state hint text shows "Type to search…" when input is empty
- Empty-state hint text shows "No matching values" when input has text
- Search input is cleared after selecting an option (both mat-autocomplete and CDK-overlay paths)

---

###  `api-traffic-v4/logs-list-base.component.ts` + `.html`

**What changed:**  
Added `isLoading = input<boolean>(false)` signal input. When `true`, shows a `<gio-loader>` spinner in the empty-row slot instead of "No data to display".

**Why:** The API-level traffic logs view has the same flash-of-empty-state problem as the environment logs view. This applies the same fix consistently using the existing `GioLoaderModule`.

---

###  `api-traffic-v4/logs-list-base.component.spec.ts`

Added test cases for `isLoading` showing loader vs empty-state message.

---

## Validation Status

| Polish item | Status |
|---|---|
| `API_PRODUCT` in `ID_BASED_FILTER_NAMES` | ✅ Done |
| `API_PRODUCT` in `RESOLVABLE_FILTERS` | ✅ Done |
| `GetFilterValuesUseCase` browse-all path for API_PRODUCT | ✅ Done |
| Dashboard filter chip label resolution (`dashboard-filters.store.ts`) | ✅ Done |
| Keyword input empty-state hint text | ✅ Done |
| Keyword input search clear after selection | ✅ Done |
| Loading state for env logs table | ✅ Done |
| Loading state for API-level logs table | ✅ Done |
| Unit tests for all Java changes | ✅ Done |
| Unit tests for all Angular changes | ✅ Done |

## Additional context
To test this out manually:
# Manual Testing Walkthrough: API Product Observability

Here are the step-by-step instructions for manually testing the end-to-end API Product Observability flow.

### 1. Setup the API Product
1. Log into the APIM Console UI.
2. Ensure you have an existing V4 or V2 API that is published and started.
3. Go to **API Products** (in the left navigation).
4. Click **Create** and set up a new API Product (give it a recognizable name like `Test API Product`).
5. Add your existing API to this API Product.
6. Publish the API Product.

1. While still in your API Product, navigate to **Plans**.
2. Click **Add Plan** and choose **API Key**.
3. Fill out the necessary details and publish the plan.

### 3. Subscribe an Application to the Product Plan
1. Navigate to **Applications** in the left menu.
2. Create a new Application (e.g., `Test Product App`) or select an existing one.
3. Go to the **Subscriptions** tab for that Application.
4. Click **Subscribe** and search for your `Test API Product`.
5. Select the API Key plan you created in step 2 and approve the subscription.
6. Copy the generated **API Key**.

### 4. Generate Traffic
1. Open your terminal or an API client (like Postman).
2. Send a few requests to your underlying API's gateway endpoint, but use the **API Key** generated from your API Product subscription:
   ```bash
   curl -H "X-Gravitee-Api-Key: YOUR_API_KEY_HERE" http://localhost:8082/your-api-context-path
   ```
3. Make sure you get successful responses (200 OK) so the traffic is properly logged by the gateway.

### 5. Validate the Observability Dashboard
1. Go back to the APIM Console UI.
2. Navigate to **Observability > Analytics** (or the Environment Logs view).
3. Click the **Add Filter** button.
4. Select the **API Product** filter.
5. In the input box:
   - **Verify Empty State:** Before typing, it should say `"Type to search…"`.
   - **Verify Search:** Type the name of your `Test API Product`. It should hit the backend and resolve the product name correctly.
   - **Verify No Lingering Text:** Select your product. The search input box should clear itself immediately.
6. Apply the filter.
7. **Verify Results:** The dashboard/table should now filter down and display the traffic you just generated with the API Product's API key.
8. **Verify UI Resolution:** Reload the page (or share the URL). Ensure that the filter chip still correctly displays the name of your API Product (e.g., "Test API Product") rather than a raw UUID.

This flow guarantees that the Gateway stamps the `api-product-id` onto the log records and that the UI successfully queries, filters, and resolves the product name.

### DIFF
--- a/gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.spec.ts
@@ -22,6 +22,7 @@ import { MatCardModule } from '@angular/material/card';
 import { MatDialog } from '@angular/material/dialog';
 import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
 import { HarnessLoader } from '@angular/cdk/testing';
+import { MatSelectHarness } from '@angular/material/select/testing';
 import { GioBannerModule } from '@gravitee/ui-particles-angular';
 import { HttpTestingController } from '@angular/common/http/testing';
 import { provideRouter } from '@angular/router';
@@ -664,6 +665,46 @@ describe('EnvLogsComponent', () => {
       // Service omits filters key entirely when filter list is empty
       expect(req.request.body.filters).toBeUndefined();
       req.flush(EMPTY_RESPONSE);
+    }));
+
+    it('should pass timeRange from store to search request when timeframe changes', fakeAsync(() => {
+      initComponent();
+
+      store().periodControl.setValue({ period: '5m', from: null, to: null });
+      store().applyCustomTimeframe();
+      fixture.detectChanges();
+
+      // Snapshot the exact timeRange the store derives BEFORE tick so the
+      // relative "now" window is frozen to the same instant as the request.
+      const expectedTimeRange = store().timeRange();
+
+      tick(1);
+
+      const req = httpTestingController.expectOne({ method: 'POST', url: SEARCH_URL });
+      expect(req.request.body.timeRange).toEqual({
+        from: expectedTimeRange.from,
+        to: expectedTimeRange.to,
+      });
+      req.flush(EMPTY_RESPONSE);
+    }));
+  });
+
+  describe('timeframe picker config', () => {
+    it('should configure timeframe selector with predefined and custom frames', fakeAsync(async () => {
+      initComponent();
+
+      // Query the rendered <mat-select> in gd-timeframe-selector and assert
+      // on the visible option labels — avoids coupling to the private `timeFrames` field.
+      // Label strings come from the timeFrames constant in timeframe-ranges.ts.
+      const select = await loader.getHarness(MatSelectHarness);
+      await select.open();
+      const options = await select.getOptions();
+      const optionTexts = await Promise.all(options.map(o => o.getText()));
+
+      expect(optionTexts.length).toBeGreaterThan(0);
+      // '1h' renders as 'Last hour', '5m' renders as 'Last 5 minutes'
+      expect(optionTexts).toContain('Last hour');
+      expect(optionTexts).toContain('Last 5 minutes');
     }));
   });
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetFilterValuesUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetFilterValuesUseCase.java
@@ -24,6 +24,7 @@ import io.gravitee.apim.core.analytics_engine.model.FilterValue;
 import io.gravitee.apim.core.analytics_engine.model.FilterValuesPage;
 import io.gravitee.apim.core.analytics_engine.query_service.AnalyticsDefinitionQueryService;
 import io.gravitee.apim.core.analytics_engine.query_service.FilterValuesQueryService;
+import io.gravitee.apim.core.api_product.model.ApiProduct;
 import io.gravitee.apim.core.api_product.query_service.ApiProductQueryService;
 import io.gravitee.apim.core.application.query_service.ApplicationQueryService;
 import io.gravitee.apim.core.audit.model.AuditInfo;
@@ -119,11 +120,14 @@ public class GetFilterValuesUseCase {
         boolean isIdBased = ID_BASED_FILTERS.contains(filterSpecName);
         boolean hasQuery = input.query() != null && !input.query().isBlank();
 
-        if (isIdBased && hasQuery) {
-            return handleIdBasedSearch(input, filterSpecName, analyticsContext);
-        }
-
         if (isIdBased) {
+            // API_PRODUCT IDs are not indexed in the analytics log records; always
+            // query the database so the browse-all path (no query) behaves identically
+            // to the with-query path. Other ID-based filters fall back to Elasticsearch
+            // when no query is present so callers see only values active in the log window.
+            if (filterSpecName == FilterSpec.Name.API_PRODUCT || hasQuery) {
+                return handleIdBasedSearch(input, filterSpecName, analyticsContext);
+            }
             return handleIdBasedNoSearch(input, filterSpecName, analyticsContext);
         }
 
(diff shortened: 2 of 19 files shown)


gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.spec.ts
```diff
@@ -665,6 +665,29 @@ describe('EnvLogsComponent', () => {
       expect(req.request.body.filters).toBeUndefined();
       req.flush(EMPTY_RESPONSE);
     }));
+
+    it('should pass timeRange from store to search request when timeframe changes', fakeAsync(() => {
```
carlos-andres-osorio, line 670: The assertion only checks presence — `expect(req.request.body.timeRange).toBeDefined()`. If the component forwarded a hardcoded constant, this test would still pass. Consider asserting the specific `from`/`to` values so the test fails if the real timeRange is not forwarded.

gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.spec.ts
```diff
@@ -665,6 +665,29 @@ describe('EnvLogsComponent', () => {
...
+      tick(1);
+
+      const req = httpTestingController.expectOne({ method: 'POST', url: SEARCH_URL });
+      expect(req.request.body.timeRange).toBeDefined();
+      req.flush(EMPTY_RESPONSE);
+    }));
+  });
+
+  describe('timeframe picker config', () => {
+    it('should configure timeframe selector with predefined and custom frames', fakeAsync(() => {
```
carlos-andres-osorio: Nit: `component['timeFrames']` accesses private internal state via bracket notation. A rename of the field breaks the test without any behavioural change. Consider querying the rendered output or a harness instead.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetFilterValuesUseCase.java
```diff
@@ -201,17 +205,21 @@ private Output searchApplicationsByNameFromEnvironment(Input input) {
 
     private Output searchApiProductsByNameFromEnvironment(Input input) {
         var environmentId = input.auditInfo().environmentId();
-        var lowerQuery = input.query().toLowerCase();
-        var matching = apiProductQueryService
+        boolean hasQuery = input.query() != null && !input.query().isBlank();
+        var lowerQuery = hasQuery ? input.query().toLowerCase() : null;
```
carlos-andres-osorio: Nit: the null guard was added here, but `searchApisByNameFromContext`, `searchApplicationsByNameFromEnvironment`, and `searchPlansByNameFromAuthorizedApis` still call `input.query().toLowerCase()` directly. The routing keeps them safe today, but the invariant is implicit — a new filter added to the unconditional path would silently NPE.

carlos-andres-osorio (changes requested):
Overall the changes look well-structured and the fixes are targeted. A few things worth looking at:

**Question — `webhook-logs-list.component.html:18`**
`<logs-list-base>` has no `[isLoading]` binding here. Was this intentional? The `isLoading` input was wired through the env-logs chain but appears to be missing from the webhook-logs chain — so the empty-state flash during loading would still occur for webhook logs.

gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.spec.ts
paulatulis: Done

gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.spec.ts
paulatulis: Done

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetFilterValuesUseCase.java
paulatulis: Done

paulatulis:
> **Question — `webhook-logs-list.component.html:18`**
> `<logs-list-base>` has no `[isLoading]` binding here. Was this intentional? The `isLoading` input was wired through the env-logs chain but appears to be missing from the webhook-logs chain — so the empty-state flash during loading would still occur for webhook logs.

I went ahead and made this consistent.

carlos-andres-osorio (approved):

PR #16817, opened 2026-05-08
feat(EXT-124): add AsyncAPI page type to portal navigation frontend

adam-rosowicz:
Adds frontend support for AsyncAPI documentation pages in portal navigation ([EXT-124](https://gravitee.atlassian.net/browse/EXT-124)).

Section editor dialog: AsyncAPI option added to the page type selector with a dedicated gio:async-api icon.
Console AsyncAPI editor: new asyncapi-editor component — Monaco editor on the left, live asyncapi-component preview on the right (mirrors the existing OpenAPI editor split-panel layout); js-yaml validation on save rejects invalid or non-AsyncAPI YAML before hitting the API
Preview styling: asyncapi-console-preview.css injected into the component's shadow DOM via cssImportPath, providing compact padding and font-size overrides so the spec renders readably in a half-width panel
Portal Next: navigation-item-content-viewer wired to render AsyncAPI pages;

Console:
https://github.com/user-attachments/assets/98d91db5-6821-4555-9625-aae5125ff0ba

Portal:

https://github.com/user-attachments/assets/f00bce46-c2c6-4668-aa29-322d9146e1d4

[EXT-124]: https://gravitee.atlassian.net/browse/EXT-124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-apim-console-webui/project.json
```diff
@@ -63,6 +63,11 @@
             "input": "node_modules/monaco-editor",
             "output": "assets/monaco-editor"
           },
+          {
+            "glob": "default.min.css",
+            "input": "node_modules/@asyncapi/web-component/node_modules/@asyncapi/react-component/styles",
```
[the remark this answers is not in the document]
adam-rosowicz: @asyncapi/react-component is not available at the top-level node_modules — it is only present as a transitive dependency of @asyncapi/web-component. The nested path is therefore the only correct path with the current dependency tree.

gravitee-apim-console-webui/src/assets/style/asyncapi-console-preview.css
```diff
@@ -0,0 +1,66 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
```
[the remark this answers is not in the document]
adam-rosowicz: asyncapi-component.css is a pre-existing file at src/assets/style/asyncapi-component.css — it has been in the repo since before this PR and is served as a static asset at assets/style/asyncapi-component.css at runtime. The relative @import 'asyncapi-component.css' is correct since both files reside in the same assets/style/ directory. This is distinct from assets/asyncapi/default.min.css which is copied from node_modules.

gravitee-apim-portal-webui-next/project.json
```diff
@@ -28,7 +28,7 @@
           },
           {
             "glob": "default.min.css",
-            "input": "node_modules/@asyncapi/react-component/styles",
+            "input": "node_modules/@asyncapi/web-component/node_modules/@asyncapi/react-component/styles",
```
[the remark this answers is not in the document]
adam-rosowicz: @asyncapi/react-component is not available at the top-level node_modules — it only exists as a transitive dependency nested under @asyncapi/web-component. The nested path is therefore the only correct path with the current dependency tree. Adding it as a direct dependency would be a separate change outside this PR's scope.

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
```diff
@@ -516,6 +519,23 @@ export class PortalNavigationItemsComponent implements HasUnsavedChanges {
...
+            return;
+          }
+        } catch (e) {
+          if (e instanceof YAMLException) {
+            this.snackBarService.error(`Invalid AsyncAPI spec: ${e.message}`);
+            return;
+          }
+          throw e;
+        }
+      }
```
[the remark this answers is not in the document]
adam-rosowicz: the fix was applied in the latest push

gravitee-apim-portal-webui-next/project.json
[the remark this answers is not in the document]
aqibmohammadkhan: I'd respectfully push back. The argument as stated treats today's `node_modules` layout as if it were a contract; it isn't.

`node_modules/@asyncapi/web-component/package.json` declares `@asyncapi/react-component` as a plain dependencies entry (no peer, no nohoist), and nothing else in the `lockfile` pulls a different version of it. There is no structural reason Yarn has to keep it nested — it just chose to. That choice can flip on any of: a Yarn upgrade, yarn dedupe, a transitive bump that ends up adding another consumer of `@asyncapi/react-component`, or a tweak to `@asyncapi/web-component` (e.g. promoting it to a peer dep). When it flips, this build asset silently disappears with no failing test — Nx will happily emit an empty `assets/asyncapi/ `directory.

Pinning the transitive isn't out of scope; it's the in-scope fix.

If you'd rather not touch dependencies at all, vendor the file.
`@asyncapi/web-component` is pinned to exact `2.6.3` in the root `package.json`, so `default.min.css` is effectively frozen. Copying it once into `src/assets/style/asyncapi-react-component.css` (or wherever) removes the entire `node_modules` traversal from the build config. Zero ongoing maintenance cost, immune to hoist outcomes. Same option, no dependency change.

**My suggestion is:**

The deciding factor is runtime coupling: this CSS is matched to the DOM/class structure that `@asyncapi/react-component` renders. If web-component is ever bumped (even a patch), the renderer changes and the CSS needs to change with it. 
Standard `node_modules/@asyncapi/react-component/styles` path is what Nx/Angular asset tooling expects — no nested traversal, no hoist coupling.
License headers and provenance stay where they belong (in node_modules), not copy-pasted into the repo.

In root package.json, add (alphabetically next to `@asyncapi/web-component`):

"`@asyncapi/react-component": "^2.6.3`",
Use the `^2.6.3 range` — same range `@asyncapi/web-component@2.6.3` already declares — so Yarn resolves both consumers to a single version and hoists it. Pinning `2.6.5` exact also dedupes today but diverges the moment web-component widens its range.

Then revert both `project.json` globs to the flat path:

`"input": "node_modules/@asyncapi/react-component/styles"`

gravitee-apim-console-webui/project.json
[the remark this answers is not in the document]
aqibmohammadkhan: same as this [comment](https://github.com/gravitee-io/gravitee-api-management/pull/16817/changes#r3209194942)

you are now reaching into a nested `node_modules`, this only works because `Yarn's` hoister chose not to hoist `@asyncapi/react-component` to the workspace root

Why this is fragile:

- Any future bump of `@asyncapi/web-component` that brings its peer/runtime range into alignment with another package may cause Yarn to hoist `@asyncapi/react-component` to the root, and the nested path stops existing. Build then silently produces a wrong/missing CSS asset.

- A yarn dedupe or `nodeLinker` change has the same effect.

gravitee-apim-console-webui/src/portal/components/asyncapi-editor/asyncapi-editor.component.ts
aqibmohammadkhan, line 1: Please write a unit test and a harness file for this component.
try to capture atleast

- creation/rendering
- `writeValue(null)` defaults to `''`
- value round-trips via `FormControl`
- `setDisabledState(true)` flips disabled and propagates to `gio-monaco-editor`

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
```diff
@@ -541,6 +548,30 @@ export class PortalNavigationItemsComponent implements HasUnsavedChanges {
     }
   }
 
+  private validateAsyncApiSpec(): boolean {
```
aqibmohammadkhan, line 562: Please add a unit test for this function.  cases to consider

- Non-ASYNCAPI content type → returns true, no snackbar.
- Empty/whitespace value → returns true, no snackbar.
- Valid YAML missing asyncapi: field → snackbar 'Invalid AsyncAPI spec: missing asyncapi version field', returns false, updatePageContent NOT called.
- Invalid YAML → YAMLException branch produces the formatted error.
- Non-YAMLException unexpected error → falls into the generic branch and calls console.error.

gravitee-apim-console-webui/src/portal/components/asyncapi-editor/asyncapi-editor.component.ts
```diff
@@ -0,0 +1,62 @@
...
+import { Component, CUSTOM_ELEMENTS_SCHEMA, ViewEncapsulation } from '@angular/core';
+import { ControlValueAccessor, FormsModule, NG_VALUE_ACCESSOR } from '@angular/forms';
+import { GioMonacoEditorModule } from '@gravitee/ui-particles-angular';
+
+@Component({
+  selector: 'asyncapi-editor',
+  standalone: true,
+  templateUrl: './asyncapi-editor.component.html',
+  styleUrl: './asyncapi-editor.component.scss',
+  encapsulation: ViewEncapsulation.None,
```
aqibmohammadkhan: `ViewEncapsulation.None` here is unnecessary and inconsistent and paired with SCSS that wraps everything under a synthetic root selector, [here](https://github.com/gravitee-io/gravitee-api-management/pull/16817/changes#diff-271ec291de4513629163d437aa552de13eabd2d4f2fb2a8f2fe3825ba1a4c13aR28)

refer `OpenApiEditorComponent` that keeps default emulated encapsulation and uses `:host`.

Shadow DOM cannot be reached from outside regardless of Angular's encapsulation, what we are styling here (`display: block; width: 100%`) is the host element of the custom element, which works perfectly fine through Angular's emulated encapsulation (it just gets an attribute selector). The CSS that actually penetrates the AsyncAPI Shadow DOM is the one passed via `cssImportPath="assets/style/asyncapi-console-preview.css"` — that has nothing to do with Angular encapsulation.

**Recommendation**: drop `ViewEncapsulation.None`, switch the SCSS back to a `:host-rooted` style like the OpenAPI editor. The generic .content__editor / .content__editor__cell selectors are leaking into the global stylesheet today.

gravitee-apim-console-webui/src/portal/navigation-items/section-editor-dialog/section-editor-dialog.component.ts
```diff
@@ -134,6 +139,7 @@ export class SectionEditorDialogComponent implements OnInit {
   public buttonTitle: string;
 
   constructor() {
+    this.iconRegistry.addSvgIconInNamespace('gio', 'async-api', this.sanitizer.bypassSecurityTrustResourceUrl('assets/logo_asyncapi.svg'));
```
aqibmohammadkhan, line 142: Two issues:

- The icon is only registered when the `section-editor dialog` opens. If any other surface ever wants to show a `gio:async-api` icon (e.g. the tree node icon for AsyncAPI pages, mirroring the existing `PortalNavigationItemIconPipe` flow), it'll silently render the `"?"` placeholder until someone opens the dialog.
- Constructor side-effects run every time the dialog is opened — the registry de-duplicates so it's harmless, but it's the wrong place semantically.

gravitee-apim-console-webui/src/portal/components/asyncapi-editor/asyncapi-editor.component.ts
```diff
@@ -0,0 +1,62 @@
...
+      multi: true,
+    },
+  ],
+  imports: [FormsModule, GioMonacoEditorModule],
+})
+export class AsyncApiEditorComponent implements ControlValueAccessor {
+  _value = '';
+  private _disabled = false;
+
+  _onChange: (value: string) => void = () => ({});
```
aqibmohammadkhan: Returning an object from a `void` callback is harmless but odd. `() => undefined / () => {}` would be clearer.

gravitee-apim-portal-webui-next/src/components/navigation-item-content-viewer/navigation-item-content-viewer.component.spec.ts
aqibmohammadkhan, line 1: please add one more unit test asserting `isShowingRedocContent()` is false

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
aqibmohammadkhan, line 562: Two UX nits on the validator:

- It only runs on Save — users don't learn the spec is invalid until they click. Consider a `FormControl` validator (or (blur) on the editor) so the Save button reflects validity in real time.

- 'asyncapi' in doc accepts asyncapi: 123 or asyncapi: null. Tightening to "non-empty string matching `^\d+\.\d+\.\d+"` would catch the common pasting-wrong-spec mistake earlier.

gravitee-apim-console-webui/src/portal/navigation-items/section-editor-dialog/section-editor-dialog.component.ts
aqibmohammadkhan: ignore this one. This is an issue but out of scope for this PR.

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
```diff
@@ -541,6 +555,46 @@ export class PortalNavigationItemsComponent implements HasUnsavedChanges {
...
+        return { message: 'Invalid AsyncAPI spec: asyncapi version field must be a non-empty semantic version string' };
+      }
+    } catch (e) {
+      if (e instanceof YAMLException) {
+        return { message: `Invalid AsyncAPI spec: ${e.message}` };
+      }
+      return { message: 'An unexpected error occurred while validating the AsyncAPI spec', cause: e };
+    }
+    return null;
+  }
```
aqibmohammadkhan, line 599: Two minor problems:

- The else branch (non-YAMLException) uses `console.error` with an eslint-disable comment. That pattern indicates the else branch is not expected — but js-yaml's load() only throws YAMLException for parse errors. The `else` clause with console.error is dead code for a YAML parse call. Remove it or narrow the catch type.
- The validation happens on every save call, but it runs `load()` on the full document string a second time (it was presumably valid enough to display in the preview). Consider a single pass or storing the last parse result.

aqibmohammadkhan (commented):
Just one more comment from my side.
Could you please take care of the sonar comments in the file. they seem logical.

aqibmohammadkhan (approved):
Good job 🚀

PR #16818, opened 2026-05-08
refactor: move JSON patch services to generic core/json_patch package

carlos-andres-osorio:
## Jira Issue
https://gravitee.atlassian.net/browse/APIM-13981

## Why

`ApiPatchDomainService` contained no API-specific logic — it only provided RFC 6902 / RFC 7396 structural validation over two library port interfaces. Keeping it in `core/api/domain_service/` implied a coupling to the API domain that doesn't exist, making the service harder to discover and reuse from other contexts.

This is the prep refactor for #16820, which extends patch support to additional resource types.

## What changed

- Renamed `ApiPatchDomainService` → `JsonPatchDomainService` and relocated it from `core/api/domain_service/` to `core/json_patch/domain_service/`.
- Moved `JsonPatchService` and `JsonMergePatchService` port interfaces to the same `core/json_patch/domain_service/` package.
- Moved infra implementations from `infra/domain_service/api/` to `infra/domain_service/json_patch/`; renamed `ApiJsonPatchServiceImpl` → `JsonPatchServiceImpl` to match the port.
- Updated all `ResourceContextConfiguration` test spring configs and `PatchApiUseCase` to reference the new class and package names.
- Moved `JsonPatchDomainServiceTest` to the core test package to match the class it tests.

## User-facing impact

No behaviour change. This is a pure rename/relocation refactor with no logic modifications.

## How to test

No behaviour change — confirm existing tests pass.

```
mvnd test -pl gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest -Dtest=PatchApiUseCaseTest,JsonPatchDomainServiceTest,JsonPatchServiceImplTest -DskipITs
```


paulatulis (approved):
A few non-blocking notes:

  1. Test placement. `JsonPatchDomainServiceTest` is for a core/ class but lives at `src/test/.../infra/domain_service/json_patch/.` The pre-PR location had the same asymmetry; since you're
  already moving files, moving the test to `src/test/.../core/json_patch/domain_service/` now is free and avoids a second blame shift later. `JsonPatchServiceImplTest` and
  `JsonMergePatchServiceImpl`-related tests are correctly in infra/ — those should stay.

  2. `@author GraviteeSource Team` header consistency. You added the header to `JsonPatchService.java, JsonPatchDomainServiceTest.java, JsonPatchServiceImplTest.java, and
  PatchApiUseCaseTest.java`, but `JsonMergePatchService.java` and `JsonMergePatchServiceImpl.java` don't have one. Either add it to those two, or drop the drive-by header changes — uniformity
  matters more than presence, and the touched-files set inflates the diff for a rename.

  3. FQN in v1-management `ResourceContextConfiguration`. The bean signatures still use the fully-qualified io.gravitee.apim.core.json_patch.domain_service.JsonPatchService rather than a
  normal import. This was likely necessary when the type briefly coexisted in two packages, but after the rename it should just be an import. Cheap to clean up; otherwise it'll read as
  deliberate-but-unexplained later.

  4. PR-trail / linkage. Worth one sentence in the description that names this as the prep refactor in front of #16820 (and that the APIM-12244 epic's allow-list change shipped separately).
   The current body reads as if APIM-13981 is fully covered here, which it isn't — a future reader running git log --follow `PatchApiUseCase.java` will appreciate the breadcrumb.

  Out of scope for this PR but worth a follow-up ticket: `JsonPatchDomainService.applyMergePatch` rejects non-object patches via `isObject()`, but RFC 7396 allows any JSON value (a scalar/array
   merge patch replaces the target outright). Pre-existing; flagging only because the new package name makes the strictness more visible.

carlos-andres-osorio:
> A few non-blocking notes:
> 
>   1. Test placement. `JsonPatchDomainServiceTest` is for a core/ class but lives at `src/test/.../infra/domain_service/json_patch/.` The pre-PR location had the same asymmetry; since you're
>   already moving files, moving the test to `src/test/.../core/json_patch/domain_service/` now is free and avoids a second blame shift later. `JsonPatchServiceImplTest` and
>   `JsonMergePatchServiceImpl`-related tests are correctly in infra/ — those should stay.
> 
>   2. `@author GraviteeSource Team` header consistency. You added the header to `JsonPatchService.java, JsonPatchDomainServiceTest.java, JsonPatchServiceImplTest.java, and
>   PatchApiUseCaseTest.java`, but `JsonMergePatchService.java` and `JsonMergePatchServiceImpl.java` don't have one. Either add it to those two, or drop the drive-by header changes — uniformity
>   matters more than presence, and the touched-files set inflates the diff for a rename.
> 
>   3. FQN in v1-management `ResourceContextConfiguration`. The bean signatures still use the fully-qualified io.gravitee.apim.core.json_patch.domain_service.JsonPatchService rather than a
>   normal import. This was likely necessary when the type briefly coexisted in two packages, but after the rename it should just be an import. Cheap to clean up; otherwise it'll read as
>   deliberate-but-unexplained later.
> 
>   4. PR-trail / linkage. Worth one sentence in the description that names this as the prep refactor in front of #16820 (and that the APIM-12244 epic's allow-list change shipped separately).
>    The current body reads as if APIM-13981 is fully covered here, which it isn't — a future reader running git log --follow `PatchApiUseCase.java` will appreciate the breadcrumb.
> 
>   Out of scope for this PR but worth a follow-up ticket: `JsonPatchDomainService.applyMergePatch` rejects non-object patches via `isObject()`, but RFC 7396 allows any JSON value (a scalar/array
>    merge patch replaces the target outright). Pre-existing; flagging only because the new package name makes the strictness more visible.

1. Done — `JsonPatchDomainServiceTest` moved to `src/test/java/io/gravitee/apim/core/json_patch/domain_service/` with the package declaration updated accordingly.

2. Done — `@author GraviteeSource Team` added to `JsonMergePatchService.java`.

3. The FQN stays. `v1-management/ResourceContextConfiguration` already imports `io.gravitee.rest.api.service.JsonPatchService` (the legacy service), so both types share the simple name `JsonPatchService`. Replacing the FQN with an import introduces an ambiguity that fails compilation. The FQN is the correct and only unambiguous way to reference the core type in that file.

4. Updated the PR description to note that this is the prep refactor preceding #16820.

PR #16820, opened 2026-05-08
feat(rest-api): reflect sanitized flows on v4 dry-run PATCH

carlos-andres-osorio:
## Issue

[APIM-13982](https://gravitee.atlassian.net/browse/APIM-13982) — Dry-run fidelity for flows.

## Why

`PATCH /apis/{apiId}?dryRun=true` on a v4 HTTP Proxy API currently echoes the caller's raw `flows` payload even when the validator would have sanitized parts of it on a real save. Callers who rely on the dry-run to preview the saved state therefore see a body the server would never actually persist. This change is the dry-run mirror of the parent epic's PATCH allow-list extension (APIM-12244 / APIM-13981) so the preview matches what a real PUT would have written.

## What changed

- **`UpdateApiDomainServiceImpl.applySanitizedValues`** — added a single `.flows(sanitized.getFlows() != null ? sanitized.getFlows() : originalDefinition.getFlows())` line to the rebuild chain (`gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/UpdateApiDomainServiceImpl.java:83`). Same null-fallback shape as the surrounding `tags` / `analytics` / `groups` / `lifecycleState` lines from PR #16453. Direct domain assignment — no mapper call, since `UpdateApiEntity.flows` is already `List<Flow>`.
- **`UpdateApiDomainServiceImplTest`** — two new unit tests using the existing `stubValidate` helper (no Spring/Jersey stack, no policy-plugin dependency):
  - `should_return_sanitized_flows_from_mutated_update_api_entity` — verifies the validator's mutated `flows` are projected onto the returned `Api`.
  - `should_preserve_original_flows_when_validator_returns_null` — explicitly seeds a non-empty `flows` list (overriding `aProxyApiV4()`'s `List.of()` default so the assertion is non-vacuous), then asserts the null-fallback to the original definition.
  - Both tests include `verify(delegate, never()).update(...)` to make the no-persistence guarantee testable.

Scope is strict: only `flows` is added to the copy-back chain. `endpointGroups`, `listeners`, and `resources` are out of scope and owned by APIM-13946 / APIM-13947 / APIM-13948.

## User-facing impact

- **Behaviour change.** A v4 HTTP Proxy `PATCH /apis/{apiId}?dryRun=true` request that includes `flows` will now return the validator-sanitized `flows` in the response body instead of the caller's raw payload. Nothing is persisted.
- **No breaking change.** Existing PATCH dry-run behaviour for `tags` / `analytics` / `groups` / `lifecycleState` / other already-copied-back fields is unchanged.
- **Documentation.** The mAPI v2 reference note about dry-run reflecting sanitized `flows` is tracked as a separate follow-up — the doc lives outside this Java repository.

## How to test

- Unit tests: `mvnd test -pl gravitee-apim-rest-api/gravitee-apim-rest-api-service -am -Dskip.validation -Dtest=UpdateApiDomainServiceImplTest -Dsurefire.failIfNoSpecifiedTests=false` — 16 tests pass, including the two new `should_*_flows_*` cases.
- Manual API check: send `PATCH /management/v2/environments/{envId}/apis/{apiId}?dryRun=true` on a v4 HTTP Proxy API with a `flows` payload that the validator normalizes (e.g. a condition expression with surrounding whitespace). Assert the response body's `flows` reflect the normalized form. Then `GET /apis/{apiId}` and confirm the persisted resource is unchanged.

## Known Issues

- AC-7 (no new DB queries / audits / logs) is satisfied by construction — the change is a single in-memory builder line in `applySanitizedValues`. No dedicated executable assertion was added; the `verify(delegate, never()).update(...)` clauses on both tests cover the persistence side implicitly.

[APIM-13982]: https://gravitee.atlassian.net/browse/APIM-13982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/UpdateApiDomainServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/UpdateApiDomainServiceImpl.java
@@ -79,34 +79,32 @@ public class UpdateApiDomainServiceImpl implements UpdateApiDomainService {
             : original.getVisibility();
         var sanitizedDefinition = originalDefinition
             .toBuilder()
-            .tags(sanitized.getTags() != null ? sanitized.getTags() : originalDefinition.getTags())
-            .flows(sanitized.getFlows() != null ? sanitized.getFlows() : originalDefinition.getFlows())
-            .analytics(sanitized.getAnalytics() != null ? sanitized.getAnalytics() : originalDefinition.getAnalytics())
-            .failover(sanitized.getFailover() != null ? sanitized.getFailover() : originalDefinition.getFailover())
-            .flowExecution(sanitized.getFlowExecution() != null ? sanitized.getFlowExecution() : originalDefinition.getFlowExecution())
-            .services(sanitized.getServices() != null ? sanitized.getServices() : originalDefinition.getServices())
-            .allowedInApiProducts(
-                sanitized.getAllowedInApiProducts() != null
-                    ? sanitized.getAllowedInApiProducts()
-                    : originalDefinition.getAllowedInApiProducts()
-            )
-            .responseTemplates(
-                sanitized.getResponseTemplates() != null ? sanitized.getResponseTemplates() : originalDefinition.getResponseTemplates()
-            )
+            .tags(coalesce(sanitized.getTags(), originalDefinition.getTags()))
+            .flows(coalesce(sanitized.getFlows(), originalDefinition.getFlows()))
+            .analytics(coalesce(sanitized.getAnalytics(), originalDefinition.getAnalytics()))
+            .failover(coalesce(sanitized.getFailover(), originalDefinition.getFailover()))
+            .flowExecution(coalesce(sanitized.getFlowExecution(), originalDefinition.getFlowExecution()))
+            .services(coalesce(sanitized.getServices(), originalDefinition.getServices()))
+            .allowedInApiProducts(coalesce(sanitized.getAllowedInApiProducts(), originalDefinition.getAllowedInApiProducts()))
+            .responseTemplates(coalesce(sanitized.getResponseTemplates(), originalDefinition.getResponseTemplates()))
             .build();
         return original
             .toBuilder()
-            .name(sanitized.getName() != null ? sanitized.getName() : original.getName())
-            .description(sanitized.getDescription() != null ? sanitized.getDescription() : original.getDescription())
-            .version(sanitized.getApiVersion() != null ? sanitized.getApiVersion() : original.getVersion())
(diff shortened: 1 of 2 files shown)


paulatulis (commented):
Erase semantics — please confirm. For application/merge-patch+json with {"flows": null}, what value does the validator return for flows? If it returns null, this coalesce restores the original list and the dry-run silently undoes the erase — which contradicts APIM-13981 AC ("Merge Patch flows: null erases all flows"). If the validator returns [] instead, we're fine, but I don't see that proven anywhere in the diff or tests. Can you either point me at the code that translates erase → [], or add a test that drives the erase scenario end-to-end on the dry-run path?

Nit:
-  PR description undersells the diff. Update the "What changed" section to say "added flows to the rebuild chain + extracted coalesce() helper across the chain" so the diff matches the description. Not worth splitting into a third PR — the helper is mechanical and contained.

carlos-andres-osorio:
> Erase semantics — please confirm. For application/merge-patch+json with {"flows": null}, what value does the validator return for flows? If it returns null, this coalesce restores the original list and the dry-run silently undoes the erase — which contradicts APIM-13981 AC ("Merge Patch flows: null erases all flows"). If the validator returns [] instead, we're fine, but I don't see that proven anywhere in the diff or tests. Can you either point me at the code that translates erase → [], or add a test that drives the erase scenario end-to-end on the dry-run path?
> 
> Nit:
> -  PR description undersells the diff. Update the "What changed" section to say "added flows to the rebuild chain + extracted coalesce() helper across the chain" so the diff matches the description. Not worth splitting into a third PR — the helper is mechanical and contained.

Erase semantics are preserved — added a test to prove it: `should_return_null_flows_when_erased_definition_has_null_flows` in `UpdateApiDomainServiceImplTest`.

The key is that the `original` parameter in `applySanitizedValues` is `updatedApi` — the **already-patched** `Api` built by `applyPatchedValues`, not the pre-patch API from the database. When `MERGE_PATCH {"flows": null}` is received, `resolvePatchableList` returns `null` (line 396 of `PatchApiUseCase`), and `applyPatchedValues` sets `definition.flows = null` in `updatedApi`. So by the time `validateV4` is called, `originalDefinition.getFlows()` is already `null`.

On the validator side, `FlowValidationServiceImpl.validateAndSanitize` short-circuits and returns its input as-is when flows is `null`, so `sanitized.getFlows()` is also `null`. Result: `coalesce(null, null) = null` — the erase is preserved, not undone.

The nit on the PR description: fair point on the "added flows to the rebuild chain" phrasing, but the coalesce extraction is an internal refactor that doesn't affect the observable contract, so I've left the description focused on the functional change.

paulatulis (approved):

PR #16824, opened 2026-05-11
fix: export allowedInApiProducts for legacy v4 APIs (APIM-13807)

gutek9:
## Issue

https://gravitee.atlassian.net/browse/APIM-13807

## Description

Ensures V4 HTTP Proxy API definition exports always include allowedInApiProducts, even for legacy APIs where the field is missing from the persisted definition. Missing legacy values are exported as false, while non-proxy APIs still omit the field.

Added regression coverage at both the export mapping layer and the Management API v2 _export/definition endpoint response.

## Additional context

Regression verified with Docker-based Maven builds using the local .m2 mount.


mukul-tyagi08 (approved):
LGTM

PR #16843, opened 2026-05-12
fix(healthcheck): report v4 health-check failures on errors

mukul-tyagi08:
## Issue

https://gravitee.atlassian.net/browse/APIM-13932

### What
- `handleConnectError` recognises timeouts (Vert.x `TimeoutException`, Netty `io.netty.handler.timeout.TimeoutException`, Netty `ConnectTimeoutException`), `InterruptionFailureException`, and generic errors — mapping each to a synthetic response so the assertion can evaluate.
- `reportStatus(false, ...)` moves to `.doOnComplete(...)` so it always runs on a real terminal, never on disposal (no phantom DOWN on shutdown / API redeploy).
- Bounded body read in the failure path via `services.healthcheck.bodyReadTimeoutMs` (default 5s, `volatile` instance field, read from `Configuration` in `start()`).
- Error logs include both API id and endpoint name.

### Why
Customer's v4 API targets a health-check backend but never sends response headers. Gateway returned **200 OK** on every request with the dashboard at **0% availability** — both reported as bugs.

### Root cause: 
`ignoreConnectionError` whitelisted only `UnknownHostException` and `SocketException`. The Vert.x `NoStackTraceTimeoutException` raised at `readTimeout` (10s default) was rethrown, swallowed by `continueOnError`, and `reportStatus(false, ...)` was never called. The endpoint stayed in its bootstrap UP state, no step data flowed to Elasticsearch, and the dashboard had nothing to display.

### DIFF
--- a/gravitee-apim-plugin/gravitee-apim-plugin-apiservice/gravitee-apim-plugin-apiservice-healthcheck-http/src/main/java/io/gravitee/apim/plugin/apiservice/healthcheck/http/HttpHealthCheckService.java
+++ b/gravitee-apim-plugin/gravitee-apim-plugin-apiservice/gravitee-apim-plugin-apiservice-healthcheck-http/src/main/java/io/gravitee/apim/plugin/apiservice/healthcheck/http/HttpHealthCheckService.java
@@ -22,6 +22,7 @@ import com.google.common.base.Strings;
 import io.gravitee.apim.plugin.apiservice.healthcheck.http.context.HttpHealthCheckExecutionContext;
 import io.gravitee.apim.plugin.apiservice.healthcheck.http.helper.HttpHealthCheckHelper;
 import io.gravitee.common.cron.CronTrigger;
+import io.gravitee.common.http.HttpStatusCode;
 import io.gravitee.common.util.URIUtils;
 import io.gravitee.definition.model.v4.endpointgroup.service.EndpointGroupServices;
 import io.gravitee.definition.model.v4.endpointgroup.service.EndpointServices;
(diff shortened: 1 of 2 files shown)


gravitee-apim-plugin/gravitee-apim-plugin-apiservice/gravitee-apim-plugin-apiservice-healthcheck-http/src/main/java/io/gravitee/apim/plugin/apiservice/healthcheck/http/HttpHealthCheckService.java
```diff
@@ -386,12 +431,14 @@ private Completable evaluateAndReport(
 
                         return response
                             .bodyOrEmpty()
-                            .doOnSuccess(body -> {
-                                reportResponse.setBody(body.toString());
+                            .timeout(bodyReadTimeoutMs, TimeUnit.MILLISECONDS, Single.just(Buffer.buffer()))
+                            .onErrorReturnItem(Buffer.buffer())
+                            .doOnSuccess(body -> reportResponse.setBody(body.toString()))
+                            .ignoreElement()
```
jhaeyaert, line 437: I don't get why, if there is a timeout, it is ignored and considered as an empty body. If a timeout occurs, it's because the server didn't answer in time, so it should be an error, right?

mukul-tyagi08: The "server didn't answer" timeout (RequestOptions.setTimeout) does become an error and a 504 — caught by handleConnectError, assertion fails, failure reported. 
The 5s bodyOrEmpty().timeout(...) is downstream of the failure decision (we're already in the assertion-failed branch) — its only job is to bound the body-attachment step. If we let it propagate as an error, it'd skip doOnComplete -> never call reportStatus(false, …) -> re-introduce the silent-swallow bug. Falling back to an empty buffer keeps the chain completing so the failure is still reported. 🙈

jhaeyaert: Ok, I missed that it was in the else block, and it was the failure case 👍

jhaeyaert (approved):

PR #16844, opened 2026-05-12
feat(apim): enable v4 API creation from remote URL (APIM-12140)

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-12140
## Description

Adds backend support for creating a v4 API by submitting a remote URL pointing to a Gravitee definition. New endpoint `POST /environments/{envId}/apis/_import/definition-url` (text/plain
  body) fetches the definition server-side, validates via the existing `UrlSanitizerUtils` whitelist + private-IP gate, then delegates to the existing import use case.

 Builds on POC `APIM-12140-enable-remote-source-for-v4`.

  Scope: APIM-12140 only — Gravitee Definition / Create flow.

  ## Out of scope (separate PRs)
  - UI: wire stepper to call the new endpoint instead of doing browser-side fetch (fixes CORS issue with raw.githubusercontent.com)
  - APIM-12142 — Update existing v4 API from URL
  - APIM-12143 / APIM-12144 — OpenAPI flows
  - Authorization-header forwarding (the UI exposes it; backend currently ignores it)

  ## Test plan
  - [x] 6 integration tests pass — happy path, permission, forbidden private URL, invalid JSON, HttpClient throws, URL not in whitelist
  - [x] `mvn prettier:write` clean
  
## Jira subtask alignment
  The existing subtasks (APIM-13647 → APIM-13658) describe UI work against the pre-refactor
  component that no longer exists. This PR implements backend support not anticipated by
  those subtasks. Recommend closing them and tracking the actual work as:
  - Backend endpoint (this PR)
  - UI rewire (follow-up PR)

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource.java
@@ -27,6 +27,8 @@ import static io.gravitee.rest.api.service.impl.search.lucene.transformer.ApiDoc
 import static io.gravitee.rest.api.service.impl.search.lucene.transformer.ApiDocumentTransformer.FIELD_TYPE_VALUE;
 import static io.gravitee.rest.api.service.impl.search.lucene.transformer.ApiDocumentTransformer.FIELD_VISIBILITY;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
 import io.gravitee.apim.core.api.domain_service.ApiStateDomainService;
 import io.gravitee.apim.core.api.exception.InvalidPathsException;
@@ -34,6 +36,7 @@ import io.gravitee.apim.core.api.model.import_definition.ImportDefinition;
 import io.gravitee.apim.core.api.use_case.CreateHttpApiUseCase;
 import io.gravitee.apim.core.api.use_case.CreateNativeApiUseCase;
 import io.gravitee.apim.core.api.use_case.ImportApiCRDUseCase;
+import io.gravitee.apim.core.api.use_case.ImportApiDefinitionFromUrlUseCase;
 import io.gravitee.apim.core.api.use_case.ImportApiDefinitionUseCase;
 import io.gravitee.apim.core.api.use_case.OAIToImportApiUseCase;
 import io.gravitee.apim.core.api.use_case.ValidateApiCRDUseCase;
@@ -75,10 +78,15 @@ import io.gravitee.rest.api.rest.annotation.Permissions;
 import io.gravitee.rest.api.security.utils.ImageUtils;
 import io.gravitee.rest.api.service.common.GraviteeContext;
 import io.gravitee.rest.api.service.search.query.QueryBuilder;
+import io.gravitee.rest.api.service.spring.ImportConfiguration;
 import io.gravitee.rest.api.service.v4.ApiStateService;
 import io.gravitee.rest.api.service.v4.exception.InvalidPathException;
 import jakarta.inject.Inject;
+import jakarta.validation.ConstraintViolation;
+import jakarta.validation.ConstraintViolationException;
 import jakarta.validation.Valid;
+import jakarta.validation.Validation;
+import jakarta.validation.Validator;
 import jakarta.validation.constraints.NotNull;
 import jakarta.ws.rs.BadRequestException;
 import jakarta.ws.rs.BeanParam;
@@ -149,7 +157,18 @@ public class ApisResource extends AbstractResource {
     private ImportApiDefinitionUseCase importApiDefinitionUseCase;
 
     @Inject
+    private ImportApiDefinitionFromUrlUseCase importApiDefinitionFromUrlUseCase;
+
+    @Inject
     private OAIToImportApiUseCase oaiToImportApiUseCase;
+
+    @Inject
+    private ImportConfiguration importConfiguration;
+
+    @Inject
+    private ObjectMapper objectMapper;
+
+    private static final Validator VALIDATOR = Validation.buildDefaultValidatorFactory().getValidator();
 
     @POST
     @Produces(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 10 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource.java
```diff
@@ -274,6 +287,31 @@ public Response createApiWithDefinition(@Valid ExportApiV4 apiToImport) {
         }
     }
 
+    @POST
+    @Path("/_import/definition-url")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.TEXT_PLAIN)
+    @Permissions({ @Permission(value = RolePermission.ENVIRONMENT_API, acls = RolePermissionAction.CREATE) })
+    public Response createApiWithDefinitionFromUrl(@NotNull String apiDefinitionUrl) {
```
wbabyte, line 301: The new endpoint parses JSON to ExportApiV4 then calls createApiWithDefinition(apiToImport) as a plain Java method. JAX-RS validation only fires at HTTP dispatch — direct invocations skip @Valid entirely. Any constraint on ExportApiV4 is enforced for
  direct-body imports but not URL-based ones. Fix: explicit Validator.validate(apiToImport) (or refactor into a shared internal method that validates first).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/FetchApiDefinitionFromUrlDomainServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Override
+    public String fetch(String url, List<String> whitelist, boolean allowPrivate) {
+        UrlSanitizerUtils.checkAllowed(url, whitelist, allowPrivate);
+
+        try {
+            // request(method, uri, headers, body, useSystemProxy) — plain GET, no extra headers/body/proxy override
+            Buffer buffer = httpClientService.request(HttpMethod.GET, url, null, null, null);
+            return buffer == null ? "" : buffer.toString();
+        } catch (RuntimeException e) {
```
wbabyte, line 52: HttpClientServiceImpl already throws TechnicalManagementException carrying the upstream status code; the catch re-wraps it and prefix-doubles the message. Any future AbstractManagementException subtype is silently downgraded to a generic 500. Fix:
  } catch (AbstractManagementException e) {
      throw e;
  } catch (RuntimeException e) {
      throw new TechnicalManagementException("Failed to fetch API definition from URL '" + url + "'", e);
  }

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource.java
```diff
@@ -274,6 +287,31 @@ public Response createApiWithDefinition(@Valid ExportApiV4 apiToImport) {
...
+
+        ExportApiV4 apiToImport = parseApiDefinition(output.apiDefinitionContent());
+        return createApiWithDefinition(apiToImport);
+    }
+
+    private ExportApiV4 parseApiDefinition(String apiDefinitionContent) {
+        try {
+            return objectMapper.readValue(apiDefinitionContent, ExportApiV4.class);
+        } catch (JsonProcessingException e) {
+            throw new BadRequestException("Invalid API definition format: " + e.getMessage(), e);
```
wbabyte: BadRequestException("Invalid API definition format: " + e.getMessage(), e) routes around the existing JsonMappingExceptionMapper whose explicit purpose is to strip internal Java class names from Jackson errors. Today this leaks io.gravitee.rest.api...
  class paths AND a slice of the fetched remote body back to the caller. Fix: drop e.getMessage() from the user-facing string; emit detail to log only.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/inmemory/FetchApiDefinitionFromUrlDomainServiceInMemory.java
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package inmemory;
+
+import io.gravitee.apim.core.api.domain_service.FetchApiDefinitionFromUrlDomainService;
+import java.util.List;
+
+public class FetchApiDefinitionFromUrlDomainServiceInMemory implements FetchApiDefinitionFromUrlDomainService {
```
wbabyte: Registered in InMemoryConfiguration but never consumed by any test. Also breaks the project's InMemoryAlternative<T> contract (every other in-memory double implements it). Fix: either consume it from the missing use-case test in #5, or delete it.

wbabyte (commented):
Zero @Slf4j / @CustomLog despite every peer infra/domain_service/api/* class using one. Operators cannot diagnose "why did this import fail at 14:32" — no record of caller, URL, upstream status, or body length.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/inmemory/FetchApiDefinitionFromUrlDomainServiceInMemory.java
paulatulis: Deleted

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/FetchApiDefinitionFromUrlDomainServiceImpl.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/inmemory/FetchApiDefinitionFromUrlDomainServiceInMemory.java
paulatulis: Tried deleting it, but turns out it's load-bearing. `ImportApiDefinitionFromUrlUseCase` (annotated @UseCase) gets auto-scanned by Spring in gravitee-apim-rest-api-management-rest's test context too (its ResourceContextConfiguration imports inmemory.spring.InMemoryConfiguration), so the bean is needed to satisfy the use case's constructor. Removing it breaks v1 tests like ApiDefinitionContextResourceTest with no qualifying bean of type 'FetchApiDefinitionFromUrlDomainService'. Kept the double but made it implement InMemoryAlternative<String> to address the contract concern.

paulatulis:
> Zero @slf4j / @CustomLog despite every peer infra/domain_service/api/* class using one. Operators cannot diagnose "why did this import fail at 14:32" — no record of caller, URL, upstream status, or body length.

Added @CustomLog to FetchApiDefinitionFromUrlDomainServiceImpl with debug logging on fetch start/end (URL, body length) and warn-level logging on both AbstractManagementException and other RuntimeException paths.

wbabyte (approved):

PR #16845, opened 2026-05-12
feat(patch-api): allow resources in v4 HTTP Proxy API PATCH allow-list

carlos-andres-osorio:
## Issue

[APIM-13948](https://gravitee.atlassian.net/browse/APIM-13948)

## Why

DevOps engineers automating API management via mAPI need to PATCH the `resources` field on a v4 HTTP Proxy API without resending the full definition. The existing PATCH allow-list blocked `resources`, forcing full-definition PUT requests for targeted resource updates. This brings v4 HTTP Proxy PATCH coverage in line with what v2 already supports.

## What changed

- **`PatchApiUseCase`**: added `resources` to `ALLOWED_FIELDS`; added a precompiled-regex guard that rejects JSON Patch operations targeting deep paths under `/resources/N/configuration/...` (configuration objects must be replaced atomically); wired `resources` into `PatchableView` and `applyPatchedValues`.
- **`UpdateApiDomainServiceImpl`**: propagates validator-sanitised resources back through the dry-run path (`applySanitizedValues`), falling back to the original resource list when the validator returns null.
- **`openapi-apis.yaml`**: updated the PATCH operation description to document the new `resources` allow-list entry, the deep-path rejection behaviour, and the dry-run sanitisation guarantee.
- Refactors (earlier commits): JSON Patch services moved to a generic `core/json_patch` package; `coalesce` helper extracted to reduce cognitive complexity.

## User-facing impact

- `PATCH /apis/{apiId}` now accepts `resources` in both JSON Patch and Merge Patch bodies for v4 HTTP Proxy APIs.
- JSON Patch operations targeting `/resources/N/configuration/<sub-field>` are rejected with an explanatory hint — callers must replace `/resources/N/configuration` atomically.
- Dry-run (`?dryRun=true`) responses now reflect validator-sanitised resource configuration and never persist changes.
- No breaking changes to existing callers; previously blocked operations now succeed.

## How to test

1. `PATCH /apis/{apiId}` with `Content-Type: application/json-patch+json` and a body containing `[{"op":"replace","path":"/resources","value":[...]}]` → `200` with updated resources.
2. Repeat with `Content-Type: application/merge-patch+json` and `{"resources":[...]}` → `200`.
3. Send a JSON Patch op targeting `/resources/0/configuration/someField` → `400` with a descriptive error.
4. Send with `?dryRun=true` → `200` with sanitised resource defaults in the body; subsequent `GET` shows resources unchanged.
5. Run existing unit tests: `mvnd test -pl gravitee-apim-rest-api/gravitee-apim-rest-api-service -Dskip.validation -Dtest="PatchApiUseCaseTest,UpdateApiDomainServiceImplTest"` — 211 tests, 0 failures.

[APIM-13948]: https://gravitee.atlassian.net/browse/APIM-13948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
@@ -298,11 +298,15 @@ paths:
                 Partially update a V4 HTTP Proxy API using JSON Patch (RFC 6902) or JSON Merge Patch (RFC 7396).
 
                 Only the following fields are patchable: name, description, apiVersion, visibility, labels, tags,
-                lifecycleState, categories, analytics, failover, flowExecution, flows, services, allowedInApiProducts,
-                allowMultiJwtOauth2Subscriptions, disableMembershipNotifications, groups, properties,
-                responseTemplates.
+                lifecycleState, categories, analytics, failover, flowExecution, flows, services, resources,
+                allowedInApiProducts, allowMultiJwtOauth2Subscriptions, disableMembershipNotifications,
+                properties, responseTemplates.
 
-                Use dryRun=true to preview the result without persisting.
+                JSON Patch operations targeting /resources/N/configuration/... (any depth under configuration) are
+                rejected with 400. Replace the full configuration object at /resources/N/configuration instead.
+
+                Use dryRun=true to preview the result without persisting. The response body reflects
+                validator-sanitised values, including plugin-default-injected resource configuration.
 
                 User must have API_DEFINITION[UPDATE] or API_GATEWAY_DEFINITION[UPDATE] permissions.
             operationId: patchApi
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
@@ -213,22 +213,24 @@ public class PatchApiUseCase {
             }
             var opType = op.path("op").asText();
             if ("move".equals(opType) || "copy".equals(opType)) {
-                var fromNode = op.path("from");
-                if (fromNode.isMissingNode()) {
-                    throw new ValidationDomainException("JSON Patch operation at index " + index + " is missing required 'from' field");
-                }
-                if (!fromNode.isTextual()) {
-                    throw new ValidationDomainException(
-                        "JSON Patch operation at index " + index + " has invalid 'from' field: must be a string"
-                    );
-                }
-                var rawFrom = fromNode.asText();
-                validateJsonPatchField(index, "from", rawFrom);
-                if (isResourceConfigurationDeepPath(rawFrom)) {
-                    throw new ApiPatchNotAllowedException(rawFrom, RESOURCE_CONFIGURATION_DEEP_PATH_MESSAGE);
-                }
+                validateMoveOrCopyFromField(index, op);
             }
             index++;
+        }
+    }
+
+    private void validateMoveOrCopyFromField(int opIndex, JsonNode op) {
+        var fromNode = op.path("from");
+        if (fromNode.isMissingNode()) {
+            throw new ValidationDomainException("JSON Patch operation at index " + opIndex + " is missing required 'from' field");
+        }
+        if (!fromNode.isTextual()) {
+            throw new ValidationDomainException("JSON Patch operation at index " + opIndex + " has invalid 'from' field: must be a string");
+        }
+        var rawFrom = fromNode.asText();
+        validateJsonPatchField(opIndex, "from", rawFrom);
+        if (isResourceConfigurationDeepPath(rawFrom)) {
+            throw new ApiPatchNotAllowedException(rawFrom, RESOURCE_CONFIGURATION_DEEP_PATH_MESSAGE);
         }
     }
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
@@ -48,12 +48,14 @@ import io.gravitee.definition.model.v4.flow.Flow;
 import io.gravitee.definition.model.v4.flow.execution.FlowExecution;
 import io.gravitee.definition.model.v4.flow.execution.FlowMode;
 import io.gravitee.definition.model.v4.property.Property;
+import io.gravitee.definition.model.v4.resource.Resource;
 import io.gravitee.definition.model.v4.service.ApiServices;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import lombok.Builder;
 import lombok.RequiredArgsConstructor;
@@ -83,8 +85,13 @@ public class PatchApiUseCase {
     private static final String FIELD_PROPERTIES = "properties";
     private static final String FIELD_RESPONSE_TEMPLATES = "responseTemplates";
     private static final String FIELD_FLOWS = "flows";
+    private static final String FIELD_RESOURCES = "resources";
 
     private static final String JSON_PATCH_PATH_PREFIX = "/";
+
+    private static final Pattern RESOURCE_CONFIGURATION_DEEP_PATH = Pattern.compile("/resources/(\\d+|-)/configuration/.+");
+    private static final String RESOURCE_CONFIGURATION_DEEP_PATH_MESSAGE =
+        "operations under /resources/N/configuration/... are not supported; replace the full configuration object at /resources/N/configuration";
 
     private static final Set<String> ALLOWED_FIELDS = Set.of(
         FIELD_NAME,
@@ -104,13 +111,14 @@ public class PatchApiUseCase {
         FIELD_DISABLE_MEMBERSHIP_NOTIFICATIONS,
         FIELD_PROPERTIES,
         FIELD_RESPONSE_TEMPLATES,
-        FIELD_FLOWS
+        FIELD_FLOWS,
+        FIELD_RESOURCES
     );
 
     private static final int MAX_PATCH_OPS = 200;
 
     private static final Set<String> BLOCKED_WITH_HINT = Set.of("state");
-    private static final Set<String> BLOCKED_WITHOUT_HINT = Set.of("endpointGroups", "listeners", "resources");
+    private static final Set<String> BLOCKED_WITHOUT_HINT = Set.of("endpointGroups", "listeners");
 
     private static final String NULL_NOT_ALLOWED_MESSAGE_FORMAT =
         "'%s' cannot be null; omit the field to leave it unchanged, or send an explicit value";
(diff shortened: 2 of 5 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
```diff
@@ -298,11 +298,15 @@ paths:
                 Partially update a V4 HTTP Proxy API using JSON Patch (RFC 6902) or JSON Merge Patch (RFC 7396).
 
                 Only the following fields are patchable: name, description, apiVersion, visibility, labels, tags,
-                lifecycleState, categories, analytics, failover, flowExecution, flows, services, allowedInApiProducts,
```
paulatulis, line 301: Groups is still in `ALLOWED_FIELDS` in code. This is a doc regression.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -198,7 +206,11 @@ private void enforceAllowListJsonPatch(JsonNode patchArray) {
             if (pathNode.isMissingNode() || !pathNode.isTextual()) {
                 throw new ValidationDomainException("JSON Patch operation at index " + index + " is missing required 'path' field");
             }
-            validateJsonPatchField(index, "path", pathNode.asText());
+            var rawPath = pathNode.asText();
+            validateJsonPatchField(index, "path", rawPath);
+            if (isResourceConfigurationDeepPath(rawPath)) {
+                throw new ApiPatchNotAllowedException(rawPath, RESOURCE_CONFIGURATION_DEEP_PATH_MESSAGE);
```
paulatulis, line 212: ApiPatchNotAllowedException reused for deep-path rejection 
  ```
throw new ApiPatchNotAllowedException(rawPath, RESOURCE_CONFIGURATION_DEEP_PATH_MESSAGE);
```
  This exception's first parameter is the field name in its existing call sites — passing a JSON Pointer (/resources/0/configuration/x) there is a type mismatch. Likely produces a malformed
   error message ("Patching field /resources/0/configuration/x is not allowed: …") and likely the wrong HTTP status code (this exception class probably maps to 403, not the 400 the PR body
  claims). The surrounding deep-path / pointer validation in validateJsonPatchField throws ValidationDomainException for the same family of failure — use that, or introduce a
  UnsupportedPatchOperationException. Confirm by checking the exception-class definition and the existing @ControllerAdvice/exception-mapper.

File: gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/UpdateApiDomainServiceImpl.java [hunk 3]
```diff
@@ -87,6 +87,7 @@ private Api applySanitizedValues(Api original, io.gravitee.rest.api.model.v4.api
             .services(coalesce(sanitized.getServices(), originalDefinition.getServices()))
             .allowedInApiProducts(coalesce(sanitized.getAllowedInApiProducts(), originalDefinition.getAllowedInApiProducts()))
             .responseTemplates(coalesce(sanitized.getResponseTemplates(), originalDefinition.getResponseTemplates()))
+            .resources(coalesce(sanitized.getResources(), originalDefinition.getResources()))
```
paulatulis, line 90: Dry-run erase not tested
The test `should_preserve_original_resources_when_validator_returns_null` asserts by design that null → restore original. But the merge-patch erase case ({"resources": null} with
  dryRun=true) reaches this exact code path. The only erase test, `merge_patch_with_null_erases_resources`, runs with dryRun=false, so it exercises applyPatchedValues, not
  applySanitizedValues. Add this test:
```
  @Test
  void dry_run_resources_null_merge_patch_reflects_erasure() {
      stubValidateV4ReturnsArgument();
      givenExistingApi(apiWithResources(List.of(
          Resource.builder().name("r1").type("cache").configuration("{}").enabled(true).build())));
      var output = execute(PatchApiUseCase.PatchType.MERGE_PATCH, mergePatch("resources", null), true);
      assertThat(httpV4Def(output.api()).getResources()).isEmpty();
  }
```
  If it fails, the production code is silently undoing erasure on dry-run.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
```diff
@@ -1639,4 +1683,183 @@ void flows_only_patch_preserves_existing_flow_execution() {
...
+                mergePatch("resources", List.of(resourceMap("new-r", "oauth2"))),
+                false
+            );
+
+            assertThat(httpV4Def(output.api()).getResources()).hasSize(1);
+            assertThat(httpV4Def(output.api()).getResources().getFirst().getName()).isEqualTo("new-r");
+        }
+
+        @Test
+        void merge_patch_with_null_erases_resources() {
```
paulatulis, line 1706: Assertion is too permissive.
  assertThat(httpV4Def(output.api()).getResources()).isNullOrEmpty();
  AC #6 says "subsequent GET returns an empty resources list" — null and [] are observably different over the wire (null may be omitted from JSON, [] is serialised). Tighten to .isEmpty()
  (and assert non-null first) or .isEqualTo(List.of()), and verify which one the GET serialiser actually emits.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
```diff
@@ -1639,4 +1683,183 @@ void flows_only_patch_preserves_existing_flow_execution() {
...
+                false
+            );
+
+            assertThat(httpV4Def(output.api()).getResources()).hasSize(1);
+            assertThat(httpV4Def(output.api()).getResources().getFirst().getName()).isEqualTo("r1");
+        }
+
+        @ParameterizedTest
+        @ValueSource(strings = { "SOME_UNKNOWN_TYPE", "CACHE" })
+        void merge_patch_accepts_resource_type_verbatim(String type) {
```
paulatulis, line 1751: may pass for the wrong reason.
  @ValueSource(strings = { "SOME_UNKNOWN_TYPE", "CACHE" })
  void merge_patch_accepts_resource_type_verbatim(String type) {
      givenExistingApi(apiWithResources(null));
      var output = execute(PatchApiUseCase.PatchType.MERGE_PATCH, mergePatch("resources", List.of(resourceMap("r1", type))), false);
  The validator isn't stubbed in this test, so it's either using the test-fixture's default validator (likely a pass-through) or whatever Spring wires in. The AC says "unknown type → 200
  verbatim" because the real validator is lax — but this test doesn't prove the real validator is lax, only that the use case doesn't reject. The behaviour under integration is the actual
  contract; this unit test gives false confidence. Add an integration test, or at minimum stub the validator with the lax behaviour explicitly.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
```diff
@@ -1639,4 +1683,183 @@ void flows_only_patch_preserves_existing_flow_execution() {
...
+            givenExistingApi(apiWithResources(List.of(r1, r2)));
+
+            var output = execute(PatchApiUseCase.PatchType.JSON_PATCH, patch("remove", "/resources/0"), false);
+
+            assertThat(httpV4Def(output.api()).getResources()).hasSize(1);
+            assertThat(httpV4Def(output.api()).getResources().getFirst().getName()).isEqualTo("r2");
+        }
+
+        @Test
+        void json_patch_replace_resource_configuration_atomically() {
```
paulatulis, line 1811: doesn't verify atomicity.
  patch("replace", "/resources/0/configuration", Map.of("ttl", 300)),
  ...
  assertThat(httpV4Def(output.api()).getResources().getFirst().getConfiguration()).contains("\"ttl\"");
  This only proves ttl is present — it would pass if the new config was merged with the old "{}" rather than replacing it atomically. To prove atomicity, start existing with
  configuration("{\"oldKey\":\"oldValue\"}") and assert the result does NOT contain oldKey. Currently the test name overstates what it verifies.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
carlos-andres-osorio: Checked `ALLOWED_FIELDS` in `PatchApiUseCase` (lines 96–116): `groups` is not there. The `.groups()` call at line 403 only forwards the existing groups unchanged through the `PatchableView` builder — it does not expose groups as patchable. The doc is accurate as written.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: Checked `ApiPatchNotAllowedException` — it extends `ValidationDomainException`, which `ValidationDomainExceptionMapper` maps to HTTP 400 `BAD_REQUEST`. Also confirmed via manual test that the endpoint returns 400 when a deep-path operation is attempted. The HTTP 403 concern doesn't apply here; the status code and message are correct as-is.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/UpdateApiDomainServiceImpl.java
carlos-andres-osorio: Added `dry_run_with_null_erases_resources` in `PatchApiUseCaseTest` — it covers exactly the described scenario: existing resources + `{"resources": null}` merge patch + `dryRun=true`, and asserts the result is `null`. The test passes, confirming the dry-run path correctly reflects erasure.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
carlos-andres-osorio: Tightened the assertion in `merge_patch_with_null_erases_resources` to `.isNull()` — a null merge-patch value produces a null list (RFC 7396 erase semantics), not an empty one. Added a separate `merge_patch_with_empty_array_clears_resources` test that asserts `.isEmpty()` for the `[]` case, keeping the two distinct wire representations separately verified.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
carlos-andres-osorio: Added `stubValidateV4ReturnsArgument()` to `merge_patch_accepts_resource_type_verbatim`. The validator is now explicitly stubbed as a pass-through, so the test actively proves the use case accepts unknown/uppercase types verbatim rather than relying on mock defaults.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
carlos-andres-osorio: Updated the test to start with `configuration("{\"oldKey\":\"oldValue\"}")` and added `assertThat(config).doesNotContain("oldKey")`. The test now proves the replace is atomic — the old key is absent in the result, ruling out a merge-instead-of-replace bug.

paulatulis (approved):

PR #16863, opened 2026-05-13
fix(env-logs): hide API Product row in details view for standalone APIs

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-12446
## Description

Screenshot 2026-05-13 at 9 26 33 AM

Screenshot 2026-05-13 at 9 26 54 AM

### DIFF
--- a/gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-details/env-logs-details.component.ts
+++ b/gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-details/env-logs-details.component.ts
@@ -28,7 +28,7 @@ import { GioBannerModule, GioClipboardModule, GioMonacoEditorModule } from '@gra
 import { forkJoin, Observable, of } from 'rxjs';
 import { catchError, map, switchMap, tap } from 'rxjs/operators';
 
-import { EnvLog } from '../../models/env-log.model';
+import { EnvLog, STANDALONE_API_PRODUCT_NAME } from '../../models/env-log.model';
 import { EnvLogsDetailsRowComponent } from '../env-logs-details-row/env-logs-details-row.component';
 import { GioHeaderComponent } from '../../../../../shared/components/gio-header/gio-header.component';
 import { EnvironmentLogsService } from '../../../../../services-ngx/environment-logs.service';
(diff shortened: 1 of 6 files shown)


gravitee-apim-console-webui/src/management/observability/env-logs/components/env-logs-details/env-logs-details.component.ts
```diff
@@ -87,6 +87,7 @@ export class EnvLogsDetailsComponent {
     const log = this.log();
     return !!log?.entrypointRequest || !!log?.endpointRequest || !!log?.entrypointResponse || !!log?.endpointResponse;
   });
+  isInApiProduct = computed(() => this.log()?.apiProductName !== 'Standalone API');
```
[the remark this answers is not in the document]
paulatulis: The suggested change (!!apiProductName && apiProductName !== 'Standalone API') collapses both cases into "hidden", which loses the dash-for-missing-data signal.

carlos-andres-osorio: `isInApiProduct` only suppresses the API Product row when the backend sends the exact sentinel `'Standalone API'`. When `apiProductName` is `null` or `undefined` — which occurs for log entries that pre-date the sentinel feature, or where the field is absent — `null !== 'Standalone API'` evaluates to `true`, so the row is rendered with a `'—'` fallback.

The existing test at `env-logs-details.component.spec.ts:324` ("should show a dash in the API Product row in the More Details panel when absent") explicitly asserts this: it flushes a response with no `apiProductName` and expects the row to show `'—'`, confirming this code path is intentionally preserved. However, from a user perspective, a `'—'` in an "API Product" row is ambiguous — it reads the same as any other missing field, making it unclear whether the API was standalone or whether product information simply wasn't captured.

carlos-andres-osorio: The sentinel string `'Standalone API'` is embedded as a string literal here and also appears verbatim in the new test at `env-logs-details.component.spec.ts:341`. These two references must stay in sync with each other and with the backend contract, but there is no compile-time enforcement linking them. A backend change to this sentinel value would break both sites silently without a build or type error.

paulatulis: Addressed.

paulatulis: Addressed.

carlos-andres-osorio (approved):

PR #16865, opened 2026-05-13
perf(api): in-memory path index for collision checks (APIM-14052)

wbabyte:
## Summary

Two changes shipped on this branch:

1. **Perf** — Replace the per-publish full catalog scan in `VerifyApiPathDomainService.unavailablePathErrors` with a per-replica in-memory `ApiPathIndex`. The previous implementation streamed the full V2/V4 API catalog from Mongo and deserialised every `ApiMongo.definition` JSON blob on every validate. The new cache turns most validates into an in-memory scan plus one indexed Mongo query bounded by the snapshot watermark.

2. **Cross-pod correctness** — The in-memory cache on its own introduces a race window: a remote pod's recently-created API is invisible to the local validator for up to one search-indexer cron interval (~5 s). The validator now combines its snapshot scan with a small indexed Mongo query (`environmentId, updatedAt ≥ refreshedAt - 10 s`) plus a per-conflict batch recheck — closing the race down to a single Mongo read latency, matching the pre-cache behaviour.

## How the validator finds collisions now

```
1. (snapshot, refreshedAt) = ApiPathIndex.snapshotOf(env, seeder)        // in-memory
2. recent  = apiQuery.search(env, updatedAt ≥ refreshedAt - 10s)         // indexed
3. merged  = snapshot ⊕ recent                                           // override wins
4. conflicts = scanPaths(merged, excludeApiId, candidatePaths)
5. for each conflicting apiId NOT in `recent`:
       batch find-by-ids → drop the conflict if the row is gone or no longer owns the path
```

Behaviour by scenario:
- **No conflict (typical)** — `recent` is empty in steady state. Just the in-memory scan + one indexed query that returns 0 rows.
- **Cross-pod create race** — `recent` catches the foreign API; conflict is reported with current paths.
- **Cross-pod delete or move race** — Snapshot says conflict; batch recheck confirms the row is gone or has moved; the conflict is dropped.

The 10 s margin is **hardcoded**. It absorbs realistic inter-pod clock skew and cron-batch processing time. It is *not* a correctness lever — the recheck path catches anything the margin misses; the margin only controls the size of the recheck cohort.

## Files

**Core (`io.gravitee.apim.core.*`, framework-free):**
- `ApiPathIndex` — per-env `ConcurrentHashMap` snapshot + per-env `AtomicReference<Instant>` watermark. Lazy seed via `computeIfAbsent`; per-row error isolation in the seeder; observer-failure invalidation; class-level Javadoc enumerates all invariants.
- `ApiPathIndexReader` / `ApiPathIndexWriter` — narrow ports. `VerifyApiPathDomainService` depends on Reader; `ApiPathIndexationObserver` depends on Writer; same bean implements both — prevents accidental mutation from the validate path.
- `ApiPathExtractor` — pure path-extraction helpers (`extractPaths(Api)`, v2/v3/v4 listener flavors) + prefix-style conflict detection. Documented as stateless / thread-safe.
- `IndexationObserver` SPI — `onIndex`/`onDelete` fired by `SearchEngineServiceImpl` before each Lucene write. Javadoc covers threading (originator runs on `indexerThreadPoolTaskExecutor`, cron runs on `searchIndexerTaskScheduler`), the failure contract (`RuntimeException` caught by dispatcher), and the **idempotency requirement** for `onDelete` (double-fire on `delete(locally=false)`).
- `VerifyApiPathDomainService` — orchestrates the snapshot scan + supplementary query + batch recheck. No Spring annotations; safety margin is a `private static final Duration`.

**Infra (Spring / repository layer):**
- `ApiPathIndexationObserver` — single observer impl. Handles `IndexableApi`, v3 `ApiEntity`, v4 `ApiEntity`. Per-event extract failure is caught and logged so one bad row can't freeze the dispatcher. Unknown api-like Indexable types log `warn` **once per class** so future api-bearing subclasses don't silently bypass the path index.
- `EnvironmentIdUpdatedAtIndexUpgrader` — Mongo compound index `{environmentId: 1, updatedAt: -1}` (`ei1ua-1`), installed at startup.
- `v4_11_1/01_add_apis_environment_id_updated_at_index.yml` — JDBC Liquibase changelog: `idx_${prefix}apis_env_updated_at` on `apis(environment_id, updated_at DESC)`.
- `ApiCriteria` (repository-api) — new `updatedAtFrom` filter; plumbed through Mongo (`ApiMongoRepositoryImpl.fillQuery`) and JDBC (`JdbcApiRepository.convert/fillPreparedStatement`).
- `ApiSearchCriteria` (core) + `ApiSearchCriteriaAdapter` — same filter at the core search-criteria layer.

## Trust model

- **Same-replica index consistency.** Every `index()` invokes `indexLocally` synchronously, so the originator's observer fires before the call returns. No drift for any of the direct `searchEngineService.index/delete` callers.
- **Same-replica delete consistency.** Originator's observer fires synchronously when broadcasting (`delete(locally=false)`). The local Lucene remove still happens later via the originator's own cron pickup — preserves Antoine Cordier's `6ce444f` 'avoid duplicated deletion' optimisation. The cron-driven `deleteLocally` invokes `onDelete` a second time with an id-only Indexable; `removeForApi` is idempotent.
- **Cross-replica.** Snapshot converges within one cron interval as before. The validator's supplementary query + batch recheck mean validate correctness no longer depends on broadcast latency.
- **Cold env.** First `findConflicts` per env runs the seeder (full env Mongo scan). Subsequent reads are in-memory. The supplementary query stays bounded by the watermark from that point on.
- **Per-row seed failure.** A single malformed API in the seed stream is caught and logged with `log.warn`. The env still installs; the snapshot is missing that one apiId; the supplementary query / recheck still find it during validates.

## Architecture

- `CoreRulesTest` / `OnionArchitectureTest` / `CycleChecksTest` / `NamingConventionTest` / `UseCaseRulesTest` / `CrudServiceRulesTest` all pass. Core (`io.gravitee.apim.core.*`) is free of any Spring annotation or import — Spring wiring lives in `infra.*` and the existing `CoreServiceSpringConfiguration` component-scan.

## Test plan

- [x] `ApiPathIndexTest` 23/23 — watermark semantics, drop-on-unseeded recovery, seed-skip on a malformed row, env isolation, host bucketing, concurrent seed.
- [x] `ApiPathIndexationObserverTest` 10/10 — extract for all three Indexable variants, idempotent delete, error isolation, unknown-type handling.
- [x] `SearchEngineServiceImplObserverTest` 11/11 — multi-observer fan-out, observer failures don't block Lucene, plus **3 new `InOrder` assertions** locking observer-before-Lucene-index, observer-before-Lucene-remove (`locally=true`), and observer-before-broadcast (`locally=false`).
- [x] `VerifyApiPathDomainServiceTest` 36/36 — 30 legacy parameterised scenarios + 6 in `WatermarkAndRecheck`: cross-pod create race detected via supplementary query; override drops stale snapshot conflict when path was moved; delete race dropped via batch recheck; real conflict survives recheck; clock-skew belt-and-suspenders dropped via recheck; **batch-recheck-fires-once verification** (Mockito `verify(...).search(argThat(criteria.ids contains both ids))`).
- [x] `ApiPathExtractorTest` (carried over from the perf-only PR).
- [x] `architecture.*Test` — all rules green.
- [x] Full module suite — `mvn -pl gravitee-apim-rest-api/gravitee-apim-rest-api-service test`: 0 failures.
- [x] Live **2-replica Kind cluster** with Helm chart (shared Mongo + Elasticsearch) — 15-scenario test script:
  - Regression: create on pod-A; both pods see the API after one cron tick.
  - Cross-pod create race: pod-A creates `/foo`, pod-B immediately tries `/foo` → **pod-B rejects** with `Path [/foo/] already exists`.
  - Override: pod-A moves X from `/foo` to `/bar`, pod-B immediately tries `/foo` → **pod-B accepts** (override drops stale snapshot).
  - Delete race: pod-A deletes X, pod-B immediately tries X's freed path → **pod-B accepts** (recheck drops snapshot conflict).
  - Real conflict survives recheck: pod-B's snapshot says conflict, `findById` confirms, **rejection holds**.
  - Same-pod immediate duplicate rejected via synchronous observer.
- [ ] CI green

## Jira
APIM-14052


carlos-andres-osorio (approved):
LGTM.

jhaeyaert (approved):

PR #16867, opened 2026-05-13
fix(management-api): encrypt and propagate properties in PATCH

carlos-andres-osorio:
## Issue

https://gravitee.atlassian.net/browse/APIM-13785

## Why

The PATCH endpoint for v4 HTTP Proxy APIs had two defects in its properties handling:

1. **Missing encryption on both legs.** `PatchApiUseCase` did not run the property encryption pipeline, so properties marked `encryptable: true` were echoed back as plaintext in the 200 response — on both real-update and dry-run. Every other write path (PUT, CRD update) encrypts via `PropertyDomainService.encryptProperties`; PATCH was the only exception.

2. **Dry-run dropped caller-supplied properties.** `UpdateApiDomainServiceImpl.applySanitizedValues` copied sanitized `tags`, `flows`, `analytics`, and other fields back onto the in-memory projection, but had no corresponding line for `properties`. Any properties supplied in the PATCH body were silently dropped from the dry-run response.

Together these caused the dry-run response to diverge from a real-update response for any API that uses properties.

## What changed

- **`PatchApiUseCase`** — added a `PatchableProperty` inner record (mirrors the existing `PatchableAnalytics` / `PatchableFlowExecution` pattern), injected `PropertyDomainService`, and added an `encryptProperties` step in `applyPatchedValues` after `resolvePatchableList` resolves the patch. Encryption is idempotent: already-encrypted properties pass through unchanged.

- **`UpdateApiDomainServiceImpl.applySanitizedValues`** — added one `coalesce` line for `properties`, matching every other field already propagated in the method.

## User-facing impact

Properties marked `encryptable: true` in a PATCH request now behave identically to a PUT request:
- The 200 response sets `encrypted: true` and returns the ciphertext in `value`.
- The dry-run (`?dryRun=true`) response reflects the same encrypted shape instead of the submitted plaintext.

No API or config changes. No data migration. Properties already stored encrypted continue to round-trip correctly (encryption is idempotent).

## How to test

```
PATCH /management/v2/environments/{envId}/apis/{apiId}
Content-Type: application/merge-patch+json

{"properties":[{"key":"secret","value":"plaintext","encryptable":true}]}
```

Expected: `200` with `properties[0].encrypted = true` and `value` set to the ciphertext — same result with and without `?dryRun=true`.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiTest.java
@@ -755,6 +755,26 @@ public class ApiResource_PatchApiTest extends ApiResourceTest {
         );
     }
 
+    @Test
+    void should_return_encrypted_value_when_property_marked_encryptable() {
+        var body = "{\"properties\":[{\"key\":\"k1\",\"value\":\"v1\",\"encryptable\":true}]}";
+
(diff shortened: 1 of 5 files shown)


wbabyte (approved):
LGTM 💪

PR #16869, opened 2026-05-14
feat(api): allow PATCH endpointGroups on v4 HTTP Proxy APIs

carlos-andres-osorio:
## Issue

[APIM-13946](https://gravitee.atlassian.net/browse/APIM-13946)

## Why

The v4 HTTP Proxy API PATCH endpoint blocked `endpointGroups` even though callers needed a way to add, remove, reorder, and edit endpoint groups without resorting to a full PUT. This change unblocks `endpointGroups` in the PATCH allow-list, giving API operators a scoped, non-destructive way to update endpoint configuration.

## What changed

- **`PatchApiUseCase`**: Moved `endpointGroups` from `BLOCKED_WITHOUT_HINT` to `ALLOWED_FIELDS` and wired it through `applyPatchedValues` and `PatchableView`. Also added deep-path guards that reject JSON Patch operations targeting opaque blobs inside endpoint groups (`configuration/...`, `sharedConfigurationOverride/...`, `sharedConfiguration/...`).
- **`UpdateApiDomainServiceImpl`**: Applied the same null-fallback coalesce pattern to `endpointGroups` in `applySanitizedValues` so the dry-run response echoes the validator-sanitised value.
- **`openapi-apis.yaml`**: Added `endpointGroups` to the patchable-fields prose in the `patch:` operation description.

## User-facing impact

API operators can now PATCH `endpointGroups` on v4 HTTP Proxy APIs (merge-patch and JSON Patch). No breaking changes to existing fields or behaviour.

## How to test

Send a PATCH request to `/v2/apis/{apiId}` with a `Content-Type: application/merge-patch+json` body containing `"endpointGroups": [...]`. The groups should be persisted and returned in the subsequent GET response. The dry-run variant (`?dryRun=true`) should echo the sanitised value without persisting.

[APIM-13946]: https://gravitee.atlassian.net/browse/APIM-13946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
@@ -45,6 +45,7 @@ import io.gravitee.definition.model.v4.analytics.Analytics;
 import io.gravitee.definition.model.v4.analytics.logging.Logging;
 import io.gravitee.definition.model.v4.analytics.sampling.Sampling;
 import io.gravitee.definition.model.v4.analytics.tracing.Tracing;
+import io.gravitee.definition.model.v4.endpointgroup.EndpointGroup;
 import io.gravitee.definition.model.v4.failover.Failover;
 import io.gravitee.definition.model.v4.flow.Flow;
 import io.gravitee.definition.model.v4.flow.execution.FlowExecution;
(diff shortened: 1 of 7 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -86,13 +87,24 @@ public class PatchApiUseCase {
...
+    private static final Pattern ENDPOINT_CONFIGURATION_DEEP_PATH = Pattern.compile(
+        "/endpointGroups/[^/]+/endpoints/[^/]+/configuration/.+"
+    );
+    private static final Pattern ENDPOINT_SHARED_CONFIG_OVERRIDE_DEEP_PATH = Pattern.compile(
+        "/endpointGroups/[^/]+/endpoints/[^/]+/sharedConfigurationOverride/.+"
+    );
+    private static final Pattern ENDPOINT_GROUP_SHARED_CONFIG_DEEP_PATH = Pattern.compile("/endpointGroups/[^/]+/sharedConfiguration/.+");
+    private static final String ENDPOINT_GROUP_OPAQUE_DEEP_PATH_MESSAGE =
+        "operations under /endpointGroups/N/endpoints/N/configuration/..., /endpointGroups/N/endpoints/N/sharedConfigurationOverride/..., or /endpointGroups/N/sharedConfiguration/... are not supported; replace the full configuration object";
+
```
wbabyte, line 117: The PR adds three regex guards (endpoint configuration, endpoint sharedConfigurationOverride, group sharedConfiguration) but misses a fourth opaque-blob path: EndpointGroupServices exposes services like healthCheck whose configuration field is @JsonRawValue String (verified at v4/service/Service.java:57-58). This is exactly the same opaque-JSON-as-String pattern the existing guards protect — a JSON Patch replace /endpointGroups/0/services/healthCheck/configuration/timeout 5000 will currently slip past isEndpointGroupOpaqueDeepPath and partially-edit a raw-value string, the very corruption mode the PR aims to prevent.

carlos-andres-osorio: Good catch. Added two more Pattern constants — ENDPOINT_GROUP_SERVICE_CONFIG_DEEP_PATH (`/endpointGroups/[^/]+/services/[^/]+/configuration/.+`) and ENDPOINT_SERVICE_CONFIG_DEEP_PATH (`/endpointGroups/[^/]+/endpoints/[^/]+/services/[^/]+/configuration/.+`) — and wired both into `isEndpointGroupOpaqueDeepPath` alongside the existing three. The error message and test cases were extended to cover `/endpointGroups/0/services/healthCheck/configuration/...` and `/endpointGroups/0/endpoints/0/services/healthCheck/configuration/...` as rejected paths.

wbabyte (approved):

PR #16873, opened 2026-05-14
feat(ci): add reusable bump-from-module workflow

FuYu3699:
## Issue

https://gravitee.atlassian.net/browse/BX-267

## Description

Adds a reusable GitHub Actions workflow that Gamma module repos call on release tag push.
It updates the module's version property in the root `pom.xml` and opens (or updates) a PR automatically.

- Triggered via `workflow_call` from module repos or `workflow_dispatch` for manual catch-up
- Fixed branch per module (`bot/bump-<module>`) — same PR is reused across releases
- Idempotent: skips commit if version already matches, refreshes PR title/body either way
- Concurrency-safe: serialized per module via `concurrency:` group

### DIFF
--- a/.github/workflows/bump-from-module.yml
+++ b/.github/workflows/bump-from-module.yml
@@ -1,0 +1,132 @@
+# Called by Gamma module repos after a release tag is pushed.
+# Creates or updates a version-bump PR in this repo for the released module.
+name: Bump Gamma module version
+
+on:
+    workflow_call:
+        inputs:
+            module:
+                description: 'Maven artifact ID (e.g. gravitee-gamma-module-aim)'
+                required: true
+                type: string
+            version:
+                description: 'New version (e.g. 1.0.0-alpha.26)'
+                required: true
+                type: string
+        secrets:
+            BOT_GITHUB_TOKEN:
+                # Needs Contents + Pull-requests write on this repo.
+                required: true
+
+    # Also runnable manually to catch up after automation gaps.
+    workflow_dispatch:
+        inputs:
+            module:
+                description: 'Maven artifact ID (e.g. gravitee-gamma-module-aim)'
+                required: true
+                type: string
+            version:
+                description: 'New version (e.g. 1.0.0-alpha.26)'
+                required: true
+                type: string
+
+jobs:
+    bump:
+        runs-on: ubuntu-latest
+        permissions:
+            contents: write
+            pull-requests: write
+        concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/bump-from-module.yml
```diff
@@ -0,0 +1,132 @@
...
+        permissions:
+            contents: write
+            pull-requests: write
+        concurrency:
+            # Hardcoded — github.repository resolves to the caller's repo in workflow_call context.
+            group: gravitee-io/gravitee-api-management-bump-${{ inputs.module }}
+            cancel-in-progress: false
+        steps:
+            - name: Checkout APIM master
+              uses: actions/checkout@v4
```
jgiovaresco: Version outdated

jgiovaresco (commented):
Why using GitHub action compare to a CircleCI job to run the same?

FuYu3699 (commented):
The trigger comes from CircleCI,  the orb fires a `curl` to `workflow_dispatch` after publish completes. But the PR-creation logic lives in GitHub Actions because:

1. GHA has native `gh` CLI and `GITHUB_TOKEN` with repo permissions, so no extra tooling to clone, configure git, or authenticate.
2. Keeping it as a `workflow_dispatch` workflow means any caller can trigger it, e.g. CircleCI, another GHA workflow, or a manual `gh workflow run` to catch up after automation gaps. It is not locked to one CI system.

.github/workflows/bump-from-module.yml
FuYu3699: updated

jgiovaresco (approved):
We need to ensure the GitHub token has the minimum privilege for this

FuYu3699:
Agreed. For the `workflow_dispatch` path (triggered by the CircleCI orb), the workflow uses the default `github.token` which is scoped by the `permissions` block to only `contents: write` and `pull-requests: write` on this repo — that's the minimum needed to push the bump branch and create/update the PR.

The `BOT_GITHUB_TOKEN` secret and `workflow_call` trigger are there as a secondary path (e.g. if another GHA workflow needs to call this directly). For the orb-driven flow, no additional token configuration is needed.

One thing to consider later: PRs created with `github.token` won't trigger CI on themselves (GitHub's infinite-loop prevention). If CI is required to merge the bump PR, we may need a GitHub App token instead. We can address that after the first end-to-end test confirms the flow works.

PR #16875, opened 2026-05-14
feat(portal): add cdkScrollable to tree container

aqibmohammadkhan:
clean up navigation item template

Introduce `cdkScrollable` to the tree container for better scroll tracking.

## Issue

https://gravitee.atlassian.net/browse/APIM-14070

## Description

Improves drag-and-drop for Developer Portal navigation items by marking the tree’s scroll container with Angular CDK’s cdkScrollable, so the CDK can detect and scroll that region during drags. Also includes minor template cleanup and stricter typing around page content loading.

Changes

- cdkScrollable on the tree panel — The mat-card that wraps portal-flat-tree-component is now a CDK scrollable (sections-panel__tree__container). CdkScrollable is imported and added to the component imports array.
- Template — Whitespace / formatting cleanup in portal-navigation-items.component.html (no behavior change beyond the scrollable directive).
- Types — loadPageContent now has an explicit return type; the selected PAGE branch uses a typed cast to PortalNavigationPage when reading portalPageContentId.

https://github.com/user-attachments/assets/aeca93b6-2ed8-43dc-b9e1-01f3c673df10

### DIFF
--- a/gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
+++ b/gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
@@ -36,6 +36,7 @@ import { MatTooltipModule } from '@angular/material/tooltip';
 import { BehaviorSubject, EMPTY, Observable, of } from 'rxjs';
 import { AsyncPipe, NgTemplateOutlet, TitleCasePipe } from '@angular/common';
 import { MatCardModule } from '@angular/material/card';
+import { CdkScrollable } from '@angular/cdk/scrolling';
 
 import {
   SectionEditorDialogComponent,
@@ -97,6 +98,7 @@ import { OpenApiEditorComponent } from '../components/openapi-editor/openapi-edi
     NgTemplateOutlet,
     TitleCasePipe,
     PortalNavigationItemIconPipe,
+    CdkScrollable,
   ],
 })
 export class PortalNavigationItemsComponent implements HasUnsavedChanges {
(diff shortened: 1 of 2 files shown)


gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
```diff
@@ -36,6 +36,7 @@ import { MatTooltipModule } from '@angular/material/tooltip';
 import { BehaviorSubject, EMPTY, Observable, of } from 'rxjs';
 import { AsyncPipe, NgTemplateOutlet, TitleCasePipe } from '@angular/common';
 import { MatCardModule } from '@angular/material/card';
+import { CdkScrollable } from '@angular/cdk/scrolling';
```
jaroslawlakomy-grav, line 39: Looks OK to import `CdkScrollable` directly here. This matches the standalone style and is also used elsewhere in portal-next for CDK scrolling.

One thing to consider: the repo is not fully consistent yet. Some Angular/Material dependencies are imported as standalone symbols, while others still use `*Module` imports. If this component only needs `cdkScrollable`, the direct `CdkScrollable` import is the narrower and clearer option. If the surrounding area is intentionally module-oriented or expected to use more scrolling CDK APIs soon, `ScrollingModule` would also be defensible. Could you double-check which convention we want to follow in this part of the portal?

jaroslawlakomy-grav (approved):

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
aqibmohammadkhan: following the same pattern as in `gio-select-search-overlay` 
I’d suggest we follow the same standalone approach as the shared overlay unless we’re about to add more scrolling CDK APIs here.

PR #16876, opened 2026-05-14
feat(authz): core module + Mongo adapters + permission and event types

podlesrafal:
End-to-end FGA authorization stack on APIM.

* Authz module (core + api): Entity and Policy domains, REST CRUD, lifecycle, audit, event publishing.
* Gateway sync chain (Order=13/14) consuming policy/entity events with cold-start atomicity.
* REST permission gating via ENVIRONMENT_AUTHORIZATION.
* Audit channel for Policy/Entity write paths.
* Broadcast Policy/Entity events to gateway sync chain.
* Entity domain + REST CRUD + lifecycle + entityId validation + listener+ schema.
* Schema generation from entities + policies with cache invalidation on every mutation.


gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/organization/RoleScopesResourceTest.java
```diff
@@ -55,6 +55,7 @@ public class RoleScopesResourceTest extends AbstractResourceTest {
             "API_PRODUCT",
             "APPLICATION",
             "AUDIT",
+            "AUTHORIZATION",
```
Okhelifi, line 58: Should we have API_PRODUCT added in this list too ?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/config/GammaEnabledCondition.java
Okhelifi, line 1: Would it make sens to move it in a shared module ? It's not specific to Authorization module no ?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/event/EventRepositoryAuthzEventPublisher.java
```diff
@@ -0,0 +1,202 @@
...
+
+    private String serialise(Map<String, Object> payload) {
+        try {
+            return objectMapper.writeValueAsString(payload);
+        } catch (JsonProcessingException e) {
+            throw new IllegalStateException("Failed to serialise authz event payload", e);
+        }
+    }
+
+    private static void requireNonBlank(String value, String name) {
```
Okhelifi: isn't it already guarantee by the annotation in the object class ?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/listener/ApiEventListener.java
```diff
@@ -0,0 +1,165 @@
...
+            handle(event.type(), coreApi);
+        }
+    }
+
+    void handle(ApiEvent type, Api coreApi) {
+        if (type == null || coreApi == null) {
+            return;
+        }
+        String envId = coreApi.getEnvironmentId();
+        if (envId == null || envId.isBlank()) {
```
Okhelifi, line 96: is it really possible ?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/EntityIdValidator.java
```diff
@@ -0,0 +1,84 @@
...
+package io.gravitee.apim.authorization.service;
+
+import io.gravitee.apim.authorization.domain.EntityKind;
+import io.gravitee.apim.authorization.domain.PolicyKind;
+import io.gravitee.apim.authorization.service.exception.EntityIdValidationCode;
+import io.gravitee.apim.authorization.service.exception.InvalidEntityIdException;
+import java.util.Objects;
+import java.util.regex.Pattern;
+
+public final class EntityIdValidator {
```
Okhelifi: Here we validate the format (regex) and the size of the entityId, but CreateOrReplaceEntityCommand already carries `@NotBlank` annotations on the same field. Since the records call Validators.validateCtor(...) in their compact constructor, any `@Size` / `@Pattern` we add there is triggered automatically at construction time, no `@Valid` needed.

Couldn't we move the size + format rules to the commands directly (CreateOrReplaceEntityCommand, UpdateEntityCommand, CreatePolicyCommand)? It would remove the duplication and make the contract visible on the type itself rather than hidden behind a service-layer call.

What would stay here is only the cross-field rule (PolicyKind.GLOBAL ↔ entityId null, RESOURCE ↔ entityId present), which isn't expressible as a field annotation. At that point the class becomes a couple of stateless helper methods — do we still need it as an injected bean, or could it be a static util. Wdyt?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,349 @@
...
+            }
+        }
+
+        int totalAffected = affectedEntities.size() + affectedPolicies.size();
+        if (totalAffected > cascadeHardLimit) {
+            throw new CascadeTooLargeException(totalAffected, cascadeHardLimit);
+        }
+
+        for (String eid : affectedEntities.keySet()) {
+            entityRepository.deleteByEntityId(environmentId, eid);
```
Okhelifi: shouldn't we have a deleteMany here to avoid the loop ?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,349 @@
...
+        }
+        for (String pid : affectedPolicies.keySet()) {
+            policyRepository.deleteById(environmentId, pid);
```
Okhelifi: same here should we have a deleteMany?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,349 @@
...
+        Objects.requireNonNull(caller, "caller must not be null");
+        requireNonBlank(entityId, "entityId");
+        entityIdValidator.validate(EntityKind.RESOURCE, entityId);
+
+        String environmentId = caller.environmentId();
+        LinkedHashMap<String, Entity> affectedEntities = computeCascadeEntities(environmentId, entityId);
+
+        LinkedHashMap<String, Policy> affectedPolicies = new LinkedHashMap<>();
+        for (String eid : affectedEntities.keySet()) {
+            for (Policy p : policyRepository.findByEntityId(environmentId, eid)) {
```
Okhelifi: can we avoid to loop on find queries and introduce a mechanism like findAllByEnvironmentIdAndEntityIdIn

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,349 @@
...
+            }
+        }
+
+        return new CascadeResult(new ArrayList<>(affectedEntities.keySet()), new ArrayList<>(affectedPolicies.keySet()));
+    }
+
+    private LinkedHashMap<String, Entity> computeCascadeEntities(String environmentId, String entityId) {
+        LinkedHashMap<String, Entity> result = new LinkedHashMap<>();
+        entityRepository.findByEntityId(environmentId, entityId).ifPresent(e -> result.put(e.entityId(), e));
+        for (Entity e : entityRepository.findByEntityIdPrefix(environmentId, entityId + ".")) {
```
Okhelifi: same here for the find in the loop and all the next one please

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,349 @@
...
+
+    private static Map<String, Object> nonNullMap(Map<String, Object> in) {
+        return in == null ? Map.of() : in;
+    }
+
+    private static List<String> nonNullList(List<String> in) {
+        return in == null ? List.of() : in;
+    }
+
+    private static String requireNonBlank(String value, String name) {
```
Okhelifi: I think those methods are duplicated no ? Could we find a way to share it somewhere ?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/AuthzSchemaServiceImpl.java
Okhelifi, line 1: Quick question on the caching strategy here. 
Since cacheByEnv lives in-process, I'm wondering how this behaves in a multi-pod K8s setup. If an upsert lands on Pod A, it invalidates locally, but wouldn't Pod B keep serving the old schema until a mutation happens to hit it directly? Could be a problem if this ends up feeding a policy validator that needs to see fresh data.

Also noticed there's no production caller for currentGaplSchema yet outside the tests. Is this meant for an upcoming PR? If it's not used in this PR I would simply remove it tbh.

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/listener/ApiEventListener.java
Okhelifi, line 1: Same here can we avoid the loops on queries in this file ?

Okhelifi (commented):
The review is still in progress but here are some comments

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/organization/RoleScopesResourceTest.java
podlesrafal: I dont understand. API_PRODUCT is already in this list

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/config/GammaEnabledCondition.java
podlesrafal: Fair point - it really is generic (just reads gamma.enabled). I checked the gamma module tree and right now authorization is the only conditional consumer (GammaModulePluginHandler uses @Value, not a @Conditional). Creating a new gravitee-gamma-common module for one 30-line class felt like overkill; I'd rather move it the moment a second consumer needs it. Filing a follow-up note. Let me know if you'd prefer to do it now.

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/listener/ApiEventListener.java
podlesrafal: Api.builder() doesn't enforce non-blank environmentId. I would keep it

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/EntityIdValidator.java
podlesrafal: Good call. Moved size + format to the records:

New EntityIdConstants in gravitee-gamma-authorization-api exposes `MAX_ENTITY_ID_LENGTH = 255` and a single regex `^[a-z0-9_-]+(?:\.[a-z0-9_-]+)*$` that folds the old length / charset / leading-dot / trailing-dot / consecutive-dot checks into one.

Entity.entityId, Policy.entityId, CreateOrReplaceEntityCommand.entityId and CreatePolicyCommand.entityId all get @Size(max = EntityIdConstants.MAX_ENTITY_ID_LENGTH) @Pattern(regexp = EntityIdConstants.FORMAT_REGEX). Since the records call Validators.validateCtor(...) in their compact constructors, this is enforced at construction time.


What stayed in EntityIdValidator:

- validate(PolicyKind, String) - cross-field only (GLOBAL ↔ entityId == null, RESOURCE ↔ entityId non-blank); not expressible as a field annotation.

- validate(EntityKind, String) - required + format check for path-param entry points (update, delete take entityId as a path param, not via a command record).

I left it as an injected bean since both services use it and it makes mocking trivial in tests — happy to flip to a static util in a follow-up if you'd rather.

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/EntityServiceImpl.java
podlesrafal: Moved both to Validators in gravitee-gamma-authorization-api

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/AuthzSchemaServiceImpl.java
podlesrafal: Two separate questions, taking them in order:

Multi-pod cache. Intentional, but only safe because there's no cross-pod consumer yet. SchemaAdminApi is consumed by EntityServiceImpl/PolicyServiceImpl running in the same mgmt-api process; the gateway sync chain reads schema-relevant state from EventRepository, not from this cache. Once the upcoming policy validator lands (it is not expected for june release). In the future the schema will validate Policies in Policy Engine before reload. For June release it is just read only schema created from Policies and Entities.

About removing currentGaplSchema. I will keep it intentionally as it's the contract the validator PR consumes, and re-introducing it later is a wash compared to leaving the (tested) skeleton in place. Happy to delete it if you'd rather see this PR strictly minimal; my preference is to keep it since you can review the schema contract here and the validator wiring there.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/organization/RoleScopesResourceTest.java
Okhelifi: 🙈 I missread the file sorry you can ignore it.

gravitee-apim-rest-api/gravitee-apim-rest-api-standalone/gravitee-apim-rest-api-standalone-distribution/pom.xml
```diff
@@ -335,6 +335,13 @@
                 </exclusion>
             </exclusions>
         </dependency>
+
+        <!-- Gamma Repositories (jar lib on parent classpath, not a Gravitee plugin) -->
+        <dependency>
```
Okhelifi, line 340: Why is this dep treated differently from every other `*-repository-mongodb` in the project? The comment says "jar lib on parent classpath, not a Gravitee plugin" but doesn't explain why.

.circleci/ci/src/workflows/workflow-pull-requests.ts
```diff
@@ -862,6 +893,14 @@ function shouldTestRepository(changedFiles: string[]): boolean {
   );
 }
 
+function shouldTestGammaRepository(changedFiles: string[]): boolean {
```
Okhelifi: If I understand it well, tomorrow if someone change something on APIM, as we have this list here
```
const mavenProjectsIdentifiers = [
    'pom.xml',
    'gravitee-apim-bom',
    'gravitee-apim-common',
    'gravitee-apim-definition',
    'gravitee-apim-parent',
    'gravitee-apim-repository',
  ];
````

the repository tests of gamma will all run that's right ?

I'm asking just to understand if it's intended. On one hand it's defensive but on the other hand it cost money and time. Wdyt ?

File: gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-api/src/main/java/io/gravitee/gamma/authorization/api/EntityRepository.java [hunk 12]
```diff
@@ -0,0 +1,116 @@
...
+import io.gravitee.gamma.authorization.domain.EntityKind;
+import io.gravitee.gamma.authorization.service.EntityFilter;
+import io.gravitee.gamma.repository.paging.Pageable;
+import io.gravitee.gamma.repository.paging.PagedResult;
+import java.util.Collection;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Optional;
+
+public interface EntityRepository {
```
Okhelifi: Naming question

I may be wrong so tell me if it's the case. 😅 The Mongo collection is authz_entities, the storage model is AuthorizationEntity, but the domain port is EntityRepository and the domain model is Entity.

From outside the package, "Entity" is one of the most overloaded names in the JVM world. So should we rename this?

gravitee-apim-rest-api/gravitee-apim-rest-api-standalone/gravitee-apim-rest-api-standalone-distribution/pom.xml
podlesrafal: The other `*-repository-mongodb` deps right above are `<type>zip</type>` with `*:*` exclusions, packaged into `plugins/` and loaded by Gravitee's isolated plugin classloader. Gamma repositories are plain jars on the rest-api parent classpath because Gamma modules wire their repository bean via standard Spring autoconfiguration / DI, which doesn't traverse the plugin classloader boundary — so the adapter has to sit next to the Spring context that consumes it.

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/AuthzSchemaServiceImpl.java
Okhelifi: Just to make sure we're aligned. The scenario I'm worried about isn't the gateway sync (agreed that reads from EventRepository) or the in-process service-to-service path (writes only invalidate). It's the REST path: once currentGaplSchema has a caller (even just a UI endpoint or a pre-deploy policy check), each REST call gets load-balanced to a random pod. If User1 mutates on Pod A and User2 reads on Pod B, Pod B serves stale until something happens to invalidate it locally.

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-api/src/main/java/io/gravitee/gamma/authorization/api/AuthzValidators.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public static <T> void validate(T obj) {
+        Set<ConstraintViolation<T>> violations = VALIDATOR.validate(obj);
+        if (!violations.isEmpty()) {
+            throw new IllegalArgumentException(message(violations));
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    public static <T> void validateCtor(Class<T> recordClass, Object... values) {
```
Okhelifi, line 43: The `@SuppressWarnings("unchecked")` comes from getDeclaredConstructors() returning Constructor<?>[]. We can avoid both the cast and the warning by resolving the canonical constructor via getRecordComponents() + getDeclaredConstructor(paramTypes), which returns Constructor<T> directly ?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-api/src/main/java/io/gravitee/gamma/authorization/service/exception/EntityNotFoundException.java
```diff
@@ -13,20 +13,24 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.gravitee.apim.core.invitation.model;
+package io.gravitee.gamma.authorization.service.exception;
 
-import io.gravitee.common.utils.TimeProvider;
-import java.time.ZonedDateTime;
+public class EntityNotFoundException extends RuntimeException {
```
Okhelifi: I have same naming concern as on EntityRepository here for example. Wdyt ?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/infra/repository/EntityRepositoryAdapter.java
Okhelifi, line 1: Should we ensure some kind of consistency between the adpater method name and the repository. We could have findAllWhatever names when we call findAll in the method and find when we call a find. Wdyt ?

Same question for the Policy Adapter

Okhelifi (approved):

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/AuthzEntityServiceImpl.java
```diff
@@ -0,0 +1,378 @@
...
+        this.eventPublisher = eventPublisher;
+        this.auditPort = auditPort;
+        if (cascadeHardLimit < 1) {
+            throw new IllegalArgumentException("cascadeHardLimit must be >= 1, got " + cascadeHardLimit);
+        }
+        this.cascadeHardLimit = cascadeHardLimit;
+    }
+
+    @Override
+    @Transactional
```
Okhelifi: What is the purpose of this `@Transactionnal` here ? Tbh I don't really know if we manage transactions on Mongo today 🤔 I only see this NoTransactionManager.java in the code. So I don't really know how it's supposed to work.

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/AuthzEntityServiceImpl.java
```diff
@@ -0,0 +1,378 @@
...
+        List<String> prefixes = new ArrayList<>(3);
+        prefixes.add(entityId + ".");
+        if (entityId.startsWith(API_PREFIX)) {
+            String apiId = entityId.substring(API_PREFIX.length());
+            if (!apiId.isEmpty() && !apiId.contains(".")) {
+                prefixes.add(MCP_PREFIX + apiId + ".");
+                prefixes.add(AGENT_PREFIX + apiId + ".");
+            }
+        }
+        for (AuthzEntity e : entityRepository.findByAnyEntityIdPrefix(environmentId, prefixes)) {
```
Okhelifi, line 373: Just an example here but can double check in this file if we avoid this kind of loops of queries ?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/AuthzEntityServiceImpl.java
```diff
@@ -0,0 +1,378 @@
...
+            .findByEntityIds(environmentId, entityIds)
+            .stream()
+            .collect(java.util.stream.Collectors.toMap(AuthzEntity::entityId, e -> e, (a, b) -> a, LinkedHashMap::new));
+
+        List<AuthzUpsertResult> results = new ArrayList<>(commands.size());
+        List<UpsertOutcome> outcomes = new ArrayList<>(commands.size());
+        for (CreateOrReplaceAuthzEntityCommand command : commands) {
+            UpsertOutcome outcome;
+            try {
+                outcome = doUpsert(command, existingByEntityId.get(command.entityId()));
```
Okhelifi: Quick question on the write loop. N save() calls = N Mongo round-trips for N commands.  Could the write be done in a single query using a bulk write ?

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/AuthzEntityServiceImpl.java
podlesrafal: Right — APIM uses `NoTransactionManager`, so `@Transactional` was a no-op annotation that misled readers into thinking atomic transactions were in play. Dropped all 9 sites (4 in `AuthzEntityServiceImpl` + 5 in `AuthzPolicyServiceImpl`) and the now-unused `TRANSACTION_TIMEOUT_SECONDS` constant in 2b2bd69e89.

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/AuthzEntityServiceImpl.java
podlesrafal: Good catch. Addressed in 2b2bd69e89:

- Added `AuthzEntityRepository.saveAll(Collection<AuthzEntity>)` on the port — default impl loops over `save()` for simple adapters, Mongo adapter overrides it with `BulkOperations.replaceOne(upsert)` to issue a single round-trip.
- `bulkUpsert` now preloads existing entities once via `findByEntityIds`, builds all `toSave` eagerly (no extra per-command `findByEntityId` calls), and commits the batch in one network call instead of N saves.
- Race retry on `DuplicateKeyException` re-runs the whole batch once (matches the single-`upsert` retry pattern).

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-core/src/main/java/io/gravitee/gamma/authorization/service/AuthzEntityServiceImpl.java
podlesrafal: Audited the file. This specific spot (`computeCascadeEntities`) was already 1 Mongo round-trip — `MongoAuthzEntityRepository` overrides `findByAnyEntityIdPrefix` with a single `$or` of regex prefixes. Same goes for the other batch ops (`findByEntityIds`, `deleteByEntityIds`).

Found one genuine N-query spot during the audit: `findApiAliases` was issuing 4 sequential queries (1 exact + 3 prefix). Collapsed to 2 (1 `findByEntityId` exact + 1 `findByAnyEntityIdPrefix` on all three prefixes via `$or`). Commit 2b2bd69e89.

PR #16878, opened 2026-05-14
refactor(portal): migrate analytics card labels to overflow-labels

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-13862

## Description

Migrate `AnalyticsDashboardCardComponent` to use the new `OverflowLabelsComponent`.

This replaces the previous fixed limit (2 labels) with a dynamic layout that automatically adapts to the card's width, improving labels visibility and simplifying the component's logic.

## Additional context

Capture d’écran 2026-05-14 à 14 30 28

### DIFF
--- a/gravitee-apim-portal-webui-next/src/components/analytics-dashboard-card/analytics-dashboard-card.component.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/components/analytics-dashboard-card/analytics-dashboard-card.component.spec.ts
@@ -15,11 +15,13 @@
  */
 import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
 import { provideNoopAnimations } from '@angular/platform-browser/animations';
 
 import { AnalyticsDashboardCardComponent } from './analytics-dashboard-card.component';
 import { AnalyticsDashboardCardHarness } from './analytics-dashboard-card.harness';
 import { fakeDashboard } from '../../entities/analytics-dashboard/analytics-dashboard.fixtures';
+import { OverflowLabelsComponent } from '../overflow-labels/overflow-labels.component';
 
 describe('AnalyticsDashboardCardComponent', () => {
   let fixture: ComponentFixture<AnalyticsDashboardCardComponent>;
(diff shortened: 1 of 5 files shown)


gravitee-apim-portal-webui-next/src/components/analytics-dashboard-card/analytics-dashboard-card.component.spec.ts
```diff
@@ -50,21 +50,8 @@ describe('AnalyticsDashboardCardComponent', () => {
...
-      fakeDashboard({
-        labels: { type: 'http', scope: 'proxy', env: 'dev', team: 'core' },
-      }),
-    );
-    fixture.detectChanges();
-    expect(await harness.getLabelCount()).toBe(3);
-    expect(await harness.getOverflowCounter()).toContain('+2');
+  it('should_render_overflow_labels_when_labels_present', async () => {
+    expect(await harness.hasOverflowLabels()).toBe(true);
   });
```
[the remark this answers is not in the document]
thomasballivet: Done, test now asserts on the +2 counter from OverflowLabelsHarness to verify the labels are actually passed through

gravitee-apim-portal-webui-next/src/components/analytics-dashboard-card/analytics-dashboard-card.component.spec.ts
```diff
@@ -50,21 +50,11 @@ describe('AnalyticsDashboardCardComponent', () => {
...
-    fixture.componentRef.setInput(
-      'dashboard',
-      fakeDashboard({
-        labels: { type: 'http', scope: 'proxy', env: 'dev', team: 'core' },
-      }),
-    );
-    fixture.detectChanges();
-    expect(await harness.getLabelCount()).toBe(3);
-    expect(await harness.getOverflowCounter()).toContain('+2');
+  it('should_pass_dashboard_labels_to_overflow_labels', async () => {
```
aqibmohammadkhan, line 55: Thia new test `should_pass_dashboard_labels_to_overflow_labels` only checks that the overflow counter exists and reads `+2`. It does not assert that the correct label values `(type:http, scope:proxy)` are wired through. This matters because `labels()` is now the only connection between the card and the overflow component — it's the main invariant to guard.

Suggested addition (if `OverflowLabelsHarness` exposes a way to read labels, or even reading them via `getLabels()`):

```
it('should_pass_correct_label_values_to_overflow_labels', async () => {
  const overflowLabels = await harness.getOverflowLabelsHarness();
  // JSDOM width=0 so all labels are hidden, but we can read them from
  // the harness or check via component directly
  expect(component['labels']()).toEqual(['type:http', 'scope:proxy']);
});
```

aqibmohammadkhan: Even a direct `expect(component['labels']()).toEqual(...) `check in the existing test would be better than relying only on the rendered `+2` counter, which is an indirect and environment-dependent signal (JSDOM width = 0 is what forces all labels to overflow).

gravitee-apim-portal-webui-next/src/components/analytics-dashboard-card/analytics-dashboard-card.component.spec.ts
```diff
@@ -50,21 +50,11 @@ describe('AnalyticsDashboardCardComponent', () => {
...
+    const overflowLabels = await harness.getOverflowLabelsHarness();
+    expect(overflowLabels).not.toBeNull();
+    // In JSDOM container width is 0 → all 2 labels go behind the counter.
+    expect(await overflowLabels!.getOverflowCounterText()).toContain('+2');
```
aqibmohammadkhan: The comment acknowledges this. The +2 behaviour is a side-effect of JSDOM returning 0 for `getBoundingClientRect().width`, not a deliberate test of the overflow logic. That logic already has its own dedicated unit test in `computeVisibleCount`. The card test shouldn't depend on it — checking +2 couples the card's test to OverflowLabelsComponent's internal rendering details. Better to instead verify the labels signal value directly, as noted above, and leave overflow behaviour to OverflowLabelsComponent's own spec

dev-marek (approved):
lgtm

PR #16884, opened 2026-05-14
fix(rest-api): return server-generated flow IDs in PATCH v4 API response

carlos-andres-osorio:
## Issue

[APIM-14076](https://gravitee.atlassian.net/browse/APIM-14076)

## Why

The PATCH endpoint for v4 HTTP Proxy APIs was echoing the caller-supplied `flow.id` back in the response body — both on dry-run and on real persists. Flow IDs are generated deep in the persistence layer (`FlowAdapter.toRepository()`), so the domain object never carries the server-generated UUID. Clients relying on the PATCH response to learn the final `flow.id` would get the wrong value; a subsequent GET would return a different ID.

## What changed

- **`PatchApiUseCase`** — when the patch operation touches flows, caller-supplied `id` values are now stripped before the update is persisted (so they never enter the database). After `updateV4` returns, the use case reads the DB-generated flow IDs back via `FlowCrudService` and injects them into the response `Output`.
- For **dry-run**, where persistence is intentionally skipped and no server ID is ever generated, flow IDs are stripped from the response (returned as `null`).
- A `flowsWerePatched()` helper detects whether the patch operation targets the `flows` field, handling both `MERGE_PATCH` and `JSON_PATCH` content types (including nested pointer paths and `move`/`copy` operations).
- `requireHttpV4` is refactored to share logic with a new `asHttpV4` helper used on the non-throwing call sites.
- `PatchApiUseCaseTest` wires in `FlowCrudServiceInMemory` and adds four test cases: dry-run strips caller IDs, real patch returns DB-generated IDs, a non-flow patch preserves pre-existing IDs, and caller-supplied IDs are not echoed back.

## User-facing impact

- **Dry-run PATCH** for v4 HTTP Proxy APIs: `flow.id` is now `null` in the response (previously returned the caller-supplied value).
- **Real PATCH** for v4 HTTP Proxy APIs: `flow.id` in the response now matches what a subsequent GET returns (previously returned the caller-supplied value instead of the server-generated UUID).
- Both Merge-Patch and JSON-Patch content types are covered.
- No API signature changes. No configuration changes.

[APIM-14076]: https://gravitee.atlassian.net/browse/APIM-14076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
@@ -34,6 +34,7 @@ import io.gravitee.apim.core.api.model.Api;
 import io.gravitee.apim.core.api.model.property.EncryptableProperty;
 import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.apim.core.exception.ValidationDomainException;
+import io.gravitee.apim.core.flow.crud_service.FlowCrudService;
 import io.gravitee.apim.core.json_patch.domain_service.JsonPatchDomainService;
 import io.gravitee.apim.core.membership.domain_service.ApiPrimaryOwnerDomainService;
 import io.gravitee.apim.core.membership.model.PrimaryOwnerEntity;
@@ -157,6 +158,7 @@ public class PatchApiUseCase {
     private final PropertyDomainService propertyDomainService;
     private final FlowListDeserializer flowListDeserializer;
     private final FlowListSerializer flowListSerializer;
+    private final FlowCrudService flowCrudService;
 
     public Output execute(Input input) {
         var existingApi = apiCrudService.get(input.apiId());
@@ -184,11 +186,61 @@ public class PatchApiUseCase {
 
         if (input.dryRun()) {
             var validated = updateApiDomainService.validateV4(updatedApi, input.auditInfo());
-            return new Output(validated, primaryOwner, workflowState);
+            return new Output(stripFlowIds(validated), primaryOwner, workflowState);
         }
 
-        var persisted = updateApiDomainService.updateV4(updatedApi, input.auditInfo());
+        var apiToUpdate = flowsWerePatched(input.patchType(), patchNode) ? stripFlowIds(updatedApi) : updatedApi;
+        var persisted = updateApiDomainService.updateV4(apiToUpdate, input.auditInfo());
+        var httpV4 = persisted.getApiDefinitionHttpV4();
+        if (httpV4 != null && httpV4.getFlows() != null) {
+            var flows = flowCrudService.getApiV4Flows(persisted.getId());
+            return new Output(
+                persisted.toBuilder().apiDefinitionValue(httpV4.toBuilder().flows(flows).build()).build(),
+                primaryOwner,
+                workflowState
+            );
+        }
         return new Output(persisted, primaryOwner, workflowState);
+    }
+
+    private boolean flowsWerePatched(PatchType patchType, JsonNode rawPatchNode) {
+        if (patchType == PatchType.MERGE_PATCH) {
+            return rawPatchNode.has(FIELD_FLOWS);
+        }
+        if (!rawPatchNode.isArray()) {
+            return false;
+        }
+        for (JsonNode op : rawPatchNode) {
+            var opType = op.path("op").asText("");
+            if ("test".equals(opType)) {
+                continue;
+            }
+            var path = op.path("path").asText("");
+            if (referencesFlows(path)) {
+                return true;
+            }
+            if (("move".equals(opType) || "copy".equals(opType)) && referencesFlows(op.path("from").asText(""))) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static boolean referencesFlows(String pointer) {
+        return pointer.equals(JSON_PATCH_PATH_PREFIX + FIELD_FLOWS) || pointer.startsWith(JSON_PATCH_PATH_PREFIX + FIELD_FLOWS + "/");
+    }
+
+    private Api stripFlowIds(Api api) {
+        var httpV4 = asHttpV4(api);
+        if (httpV4 == null || httpV4.getFlows() == null) {
+            return api;
+        }
+        var stripped = httpV4
+            .getFlows()
+            .stream()
+            .map(f -> f == null ? null : (Flow) f.toBuilder().id(null).build())
+            .toList();
+        return api.toBuilder().apiDefinitionValue(httpV4.toBuilder().flows(stripped).build()).build();
     }
 
     private JsonNode toJsonNode(Api api) {
@@ -198,8 +250,13 @@ public class PatchApiUseCase {
         return node;
     }
 
+    private static io.gravitee.definition.model.v4.Api asHttpV4(Api api) {
+        return api != null && api.getApiDefinitionValue() instanceof io.gravitee.definition.model.v4.Api httpV4 ? httpV4 : null;
+    }
+
     private static io.gravitee.definition.model.v4.Api requireHttpV4(Api api) {
-        if (api.getApiDefinitionValue() instanceof io.gravitee.definition.model.v4.Api httpV4) {
+        var httpV4 = asHttpV4(api);
+        if (httpV4 != null) {
             return httpV4;
         }
         throw new IllegalStateException("API definition is not an HTTP v4 definition");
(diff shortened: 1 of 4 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
paulatulis, line 1: Strip-before-persist breaks legitimate "update flow by id" semantics

  FlowCrudServiceImpl.save() (apim-rest-api-service/src/main/java/io/gravitee/apim/infra/crud_service/flow/FlowCrudServiceImpl.java:181–207) uses the incoming flow.getId() as the key into
  dbFlowsById. If the caller supplies an id that matches an existing flow, save() takes the update branch and rewrites the existing row in place; the row's UUID is preserved. If the id
  doesn't match (or is null), it deletes-by-not-in-set and creates fresh.

  The FlowAdapter mapping the JIRA cites (@Mapping(target = "id", expression = "java(UuidString.generateRandom())"), FlowAdapter.java:47) only runs on the create path. The update adapter
  has @Mapping(target = "id", ignore = true) (FlowAdapter.java:57). So pre-PR, caller-supplied IDs were ignored on create (giving the buggy response the JIRA describes), but were honored on
   update — i.e. a caller who sends a real existing flow UUID renames/edits that flow without losing its identity.

  Post-PR (PatchApiUseCase.execute line 186):
  var apiToUpdate = flowsWerePatched(input.patchType(), patchNode) ? stripFlowIds(updatedApi) : updatedApi;
  …all caller IDs are wiped before updateV4 runs. In save():
  - flowIdsToSave collapses to the empty set (filter(Objects::nonNull))
  - every previously stored flow is added to flowIdsToDelete and dropped
  - every caller flow goes down the create branch with a fresh UUID

  Net effect: a flows PATCH that previously meant "update flow X in place" now means "delete flow X, insert a new flow with a new UUID". From a black-box GET perspective the data still
  looks right, but any client that tracks a flow by id (or any audit/diff tooling) sees an unrelated row each time.

  This is not justified by the JIRA — the root cause analysis says only that "the saveApiFlows return value … does not appear to be propagated back into the response object." Step (2) alone
   (DB-read + inject) is enough to fix the symptom.

  Two options worth considering:
  1. Drop the strip-on-persist step entirely. Caller IDs that don't collide are already neutralized by the create-side adapter. Caller IDs that do collide are doing legitimate update-by-id
  work. The response substitution already gives the caller the authoritative IDs.
  2. If the strip is intentional (e.g. to keep the api definition JSON column from holding bogus IDs), please call it out in the PR description and add a test asserting that "PATCH flows
  with a real existing UUID preserves that UUID on subsequent GET" — and decide whether that's the desired contract for v4 flow PATCH.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -180,19 +182,74 @@ public Output execute(Input input) {
 
         if (input.dryRun()) {
```
paulatulis, line 187: if (input.dryRun()) {
      var validated = updateApiDomainService.validateV4(updatedApi, input.auditInfo());
      return new Output(stripFlowIds(validated), primaryOwner, workflowState);
  }

  For a dry-run patch on, say, just name, the validated result carries the existing flows (sourced from httpV4.getFlows() in applyPatchedValues). The PR nulls those IDs out unconditionally.
   This is asymmetric with the real-run path, which fetches authoritative IDs from FlowCrudService and returns them even when the patch didn't touch flows (covered by the new test
  patch_not_touching_flows_returns_db_stored_flow_id_in_response).

  So today: real PATCH of {"name": "x"} returns full flow IDs in the response; dry-run of the same patch returns null for every flow id. That's surprising and is also a regression for
  callers using dry-run today.

  Suggested fix: only strip when flows were patched.
  var validated = updateApiDomainService.validateV4(updatedApi, input.auditInfo());
  var output = flowsWerePatched(input.patchType(), patchNode) ? stripFlowIds(validated) : validated;
  return new Output(output, primaryOwner, workflowState);
  That preserves the JIRA's "dry-run can't know future IDs" property for the flow-mutating case without lying about untouched flows. And please add a test for dry-run + non-flow patch so
  this asymmetry stays nailed down.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
paulatulis, line 1: Test gaps:  
 - Dry-run + non-flow patch (locks in the unconditional-strip behavior, or invalidates it).
  - JSON Patch replace/add/remove on /flows/N/... sub-pointers — the new parameterized tests only exercise whole-field merge/replace via setField.
  - JSON Patch move/copy with from: /flows/N — flowsWerePatched has a code path for this that isn't exercised.
  -  "Caller supplies an id that matches an existing flow"

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: `stripFlowIds` has been removed from the persist path. `PatchApiUseCase.execute` now passes `updatedApi` directly to `updateV4` (no ternary, no strip), so `FlowCrudServiceImpl.save()` receives caller-supplied IDs intact and uses them for the update-by-id lookup in `dbFlowsById`. The response injection at lines 191–197 (`flowCrudService.getApiV4Flows`) continues to return the authoritative persisted IDs. `flows_patch_preserves_existing_flow_id_when_passed_to_update_domain_service` (parameterized over both patch types) verifies the caller-supplied existing UUID is forwarded to `updateV4` unmodified.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: The dry-run branch now guards `stripFlowIds` with `flowsWerePatched`: line 185 reads `var result = flowsWerePatched(input.patchType(), patchNode) ? stripFlowIds(validated) : validated;`. A patch that doesn't reference `/flows` leaves existing DB flow IDs intact in the dry-run response. `dry_run_patch_not_touching_flows_preserves_flow_ids` asserts a name-only merge-patch dry-run returns the pre-existing flow ID unchanged.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
carlos-andres-osorio: Four of the five gaps are now covered:
- Gap 1 (dry-run + non-flow patch): `dry_run_patch_not_touching_flows_preserves_flow_ids` in `PatchApiUseCaseTest` — asserts a name-only merge-patch dry-run preserves the pre-existing flow ID.
- Gap 2 (sub-pointer ops): `flowsWerePatched` detects any `path` equal to `/flows` or starting with `/flows/`, so JSON Patch replace/add/remove on `/flows/N/...` sub-pointers are covered by the prefix check. No dedicated parameterized test was added specifically for `flowsWerePatched` detection via sub-pointer paths.
- Gap 3 (move/copy with `from:/flows/N`): `flowsWerePatched` inspects `path` only, not `from`. A move from `/flows/0` to a non-flows destination won't be detected as a flows patch, so dry-run won't strip IDs for that operation. No test was added for this case.
- Gap 4 (caller supplies existing flow UUID): `flows_patch_preserves_existing_flow_id_when_passed_to_update_domain_service` in `PatchApiUseCaseTest` — parameterized over MERGE_PATCH and JSON_PATCH — verifies a caller-supplied existing UUID reaches `updateV4` unmodified.
- Gap 5 (service-layer, non-matching caller ID): `FlowCrudServiceImplTest.SaveApiFlows.caller_supplied_id_not_matching_db_flow_is_replaced_with_generated_uuid` verifies a flow with a caller-supplied ID absent from the DB receives a server-generated UUID on create.

wbabyte (approved):

PR #16885, opened 2026-05-14
feat(authz): Gamma REST endpoints + wire authz module into APIM

podlesrafal:
REST surface for the authorization stack and the deployment-level wiring that activates the whole feature.

* gravitee-gamma-authz-rest: new module deployed at /gamma/authz.
  - Resources: PoliciesResource, EntitiesResource, SchemaResource, HealthResource.
  - DTOs: EntityRequest/Response, PolicyRequest/Response, UpdateEntityRequest, UpdatePolicyRequest, SchemaResponse, CascadeResponse.
  - Exception mappers translate domain exceptions (EntityNotFoundException, PolicyNotFoundException, CascadeTooLargeException, InvalidStatusTransitionException, InvalidEntityIdException, IllegalArgumentException) into RFC-friendly error responses.
  - AuthzCallerResolver builds the AuthzCallerContext from the request.
  - Permission gating via ENVIRONMENT_AUTHORIZATION (added in PR1).

Depends on PR #16876 (authz-api/core for ports and services, permissions and event types)


gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/GammaModuleApplication.java
```diff
@@ -52,6 +55,10 @@ public class GammaModuleApplication extends ResourceConfig {
     public GammaModuleApplication() {
         register(GammaRootResource.class);
         register(GammaUIResource.class);
+        register(AuthzRootResource.class);
```
Okhelifi: The GammaModule SPI already exposes restResource() for exactly this use case. Modules return their JAX-RS resource class and GammaModulesResource auto-mounts it under /.../modules/{pluginId}/.... This PR creates a dedicated gravitee-gamma-authorization-rest Maven module, registers AuthzRootResource manually in GammaModuleApplication, and hardcodes `@Path("/authz")` in GammaRootResource, bypassing the SPI entirely.

I think a cleaner fit would be :

1. Move the resources/DTOs/mappers/tests into gravitee-gamma-module-authz (drop the -rest module)
AuthorizationModule.restResource() { return AuthzRootResource.class; }
2. Drop the register(AuthzRootResource.class) line in 3. GammaModuleApplication and the `@Path("/authz")` block in GammaRootResource


It's correctly done in aim module if you need an example https://github.com/gravitee-io/gravitee-gamma-module-aim/blob/alpha/src/main/java/com/graviteesource/gamma/module/aim/AIManagementModule.java

gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/rest/resource/AuthzSchemaResource.java
```diff
@@ -0,0 +1,54 @@
...
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+import java.util.Objects;
+
+@Tag(name = "Authz Schema", description = "Current GAPL schema for the environment, computed from registered entities and policies")
+@Path("/environments/{environmentId}/schema")
+@Produces(MediaType.APPLICATION_JSON)
+public class AuthzSchemaResource {
```
Okhelifi, line 39: Sorry to insist here but can we add a TODO to ensure that as soon as a POST is added in the ressource we will handle the cache correctly for multi pod instances

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-rest/src/main/java/io/gravitee/gamma/authorization/rest/exception/IllegalArgumentExceptionMapper.java
```diff
@@ -0,0 +1,36 @@
...
+
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.ext.ExceptionMapper;
+import jakarta.ws.rs.ext.Provider;
+import lombok.CustomLog;
+
+@Provider
+@CustomLog
+public class IllegalArgumentExceptionMapper implements ExceptionMapper<IllegalArgumentException> {
```
Okhelifi: Two small upstream changes could make it redundant I think :

- Validators.validateCtor can throw ConstraintViolationException(violations) directly instead of wrapping in IllegalArgumentException(message(violations)). The existing ConstraintValidationExceptionMapper (already registered in GammaModuleApplication) catches it and produces a structured 400 with path and invalid_value per field ( better than today's concatenated string. )
 - Pageable constructor and requireNonBlank callers could throw a typed AuthzInvalidArgumentException extends AuthzApiException (404 → 400, code InvalidArgument) so the existing AuthzApiExceptionMapper handles it with the uniform ErrorBody shape.

After that, any remaining IllegalArgumentException reaching the mapper is by definition unexpected (a bug, a third-party throw) and the right behaviour is to bubble it to ThrowableMapper → 500, not silently map it to 400.

Net effect : one less mapper to maintain, structured validation errors out of the box, and IllegalArgumentException regains its "this is a programmer error" semantics.

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-api/src/main/java/io/gravitee/gamma/authorization/paging/Pageable.java
```diff
@@ -39,6 +39,13 @@ public static Pageable of(int page, int perPage) {
...
     }
 
+    public static Pageable fromQuery(Integer page, Integer perPage) {
+        if (page == null && perPage == null) {
+            return firstPage();
+        }
+        return of(page == null ? 1 : page, perPage == null ? DEFAULT_PER_PAGE : perPage);
+    }
+
     public static Pageable unbounded() {
```
Okhelifi, line 44: is it documented in the openapi documentation ?

Okhelifi, line 44: LInked to previous comment but I don't even know if this module expose an openapi file btw ?

gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/rest/resource/AuthzCallerResolver.java
```diff
@@ -0,0 +1,44 @@
...
+package io.gravitee.gamma.authorization.rest.resource;
+
+import io.gravitee.gamma.authorization.api.AuthzCallerContext;
+import io.gravitee.rest.api.service.common.GraviteeContext;
+import io.gravitee.rest.api.service.exceptions.ForbiddenAccessException;
+import io.gravitee.rest.api.service.exceptions.UnauthorizedAccessException;
+import jakarta.ws.rs.core.SecurityContext;
+import java.security.Principal;
+
+final class AuthzCallerResolver {
```
Okhelifi, line 26: To double check but if we move to the SPI mounting (/.../modules/authz/... instead of the current /authz/environments/{environmentId}/...), this whole resolver collapses :

- org/env null chec -> redundant : GraviteeContextFilter already populates and validates them from the standard path, returns 400 if missing
- path env vs context env check -> impossible : under the SPI path, the env IS what populates the context, no way for them to differ
-> principal null check —> redundant : SecurityContextFilter (registered in GammaModuleApplication) already returns 401 for unauthenticated requests

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-api/src/main/java/io/gravitee/gamma/authorization/service/CreateAuthzPolicyCommand.java
```diff
@@ -15,19 +15,16 @@
...
 import jakarta.validation.constraints.NotNull;
-import jakarta.validation.constraints.Pattern;
-import jakarta.validation.constraints.Size;
 
 public record CreateAuthzPolicyCommand(
     @NotBlank String environmentId,
     @NotBlank String name,
     @NotNull AuthzPolicyKind kind,
-    @Size(max = AuthzEntityIdConstants.MAX_ENTITY_ID_LENGTH) @Pattern(regexp = AuthzEntityIdConstants.FORMAT_REGEX) String entityId,
+    String entityId,
```
Okhelifi: Removing @Size/@Pattern here isn't just a defensive regression, it opens a hole for the policy create path. The chain has zero format or length check on entityId end to end :

- DTO AuthzPolicyRequest : entityId has no annotation, not even @NotBlank
- Command CreateAuthzPolicyCommand : @Size/@Pattern removed in this PR, only @NotBlank remains on other fields
- Validator AuthzEntityIdValidator.validate(AuthzPolicyKind, String) (lines 33-54) : only enforces the cross-field rule (GLOBAL ↔ null, RESOURCE ↔ not blank). No length, no format. Only the AuthzEntityKind overload (lines 55-77) checks those.

So today a request like :

```
POST /gamma/authz/environments/env-1/policies
{ "name": "ok", "kind": "RESOURCE", "entityId": "<50KB of arbitrary chars>", "policyText": "" }
````
is accepted, persisted, and published to the gateway via eventPublisher.publishPolicyDeployed, straight into events / events_latest broadcast to every gateway instance. Easy DOS surface (oversized payloads in the event log), plus you can create policies that reference entityIds the matching entity command would itself have rejected (broken referential integrity).

Two fixes work :

- Put @Size(max=AuthzEntityIdConstants.MAX_ENTITY_ID_LENGTH) + @Pattern(regexp=AuthzEntityIdConstants.FORMAT_REGEX) back on CreateAuthzPolicyCommand.entityId, and mirror them on AuthzPolicyRequest.entityId
- Or extend AuthzEntityIdValidator.validate(AuthzPolicyKind, …) to call the same length/format checks the AuthzEntityKind overload does, so the runtime path covers both branches

First option is cheaper and gives the client a useful Swagger contract ; second option is fine if we want validation centralized.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/GammaModuleApplication.java
podlesrafal: Done — see 610ac98ca7 refactor(authz): address PR #16885 review feedback. The -rest Maven module is dropped, all REST sources moved into gravitee-gamma-module-authz via git mv (history preserved), AuthorizationModule.restResource() returns AuthzRootResource.class, and the manual register(AuthzRootResource.class) + @Path("/authz") block are gone. URL now follows the standard /gamma/organizations/{org}/environments/{env}/modules/authz/* shape via GammaModulesResource.

gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/rest/resource/AuthzSchemaResource.java
podlesrafal: Added — TODO(authz): when a write endpoint (POST/PUT) is added here, wire schema cache invalidation across pods. Sits above the class so anyone adding a write endpoint sees it before touching the resource.

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-rest/src/main/java/io/gravitee/gamma/authorization/rest/exception/IllegalArgumentExceptionMapper.java
podlesrafal: Done:

- AuthzValidators.{validate, validateCtor} now throw ConstraintViolationException directly — the existing ConstraintValidationExceptionMapper produces the structured per-field 400 (path / invalid_value).
- Pageable, requireNonBlank, requireMatchingEnv, and AuthzEntityServiceImpl.validateFilter throw a new typed AuthzInvalidArgumentException extends IllegalArgumentException (parent kept for back-compat with isInstanceOf assertions in domain tests). A dedicated AuthzInvalidArgumentExceptionMapper produces the uniform {error: "InvalidArgument", message} 400.
- requireNonBlank(null, name) now raises the typed exception too (was NullPointerException from Objects.requireNonNull), so it routes through the mapper instead of leaking to the default 500.
- IllegalArgumentExceptionMapper removed — any remaining bare IllegalArgumentException is by contract a programmer error and bubbles to ThrowableMapper → 500.

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-api/src/main/java/io/gravitee/gamma/authorization/paging/Pageable.java
podlesrafal: Good catch — no, the module doesn't expose an OpenAPI spec today (no @Operation annotations, no openapi.yaml under src/main/resources/, unlike e.g. gravitee-gamma-module-aim). Tracking as a follow-up — https://gravitee.atlassian.net/browse/GMA-348. It is worth to do in separate PR.

gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/rest/resource/AuthzCallerResolver.java
podlesrafal: Done. Caller integrity now relies on @Permissions(ENVIRONMENT_AUTHORIZATION) being present on every endpoint — PermissionsFilter rejects (403) before the resolver runs if either the env-scoped context or the principal is missing. AuthzPermissionContractTest pins that requirement.

gravitee-gamma/gravitee-gamma-authorization/gravitee-gamma-authorization-api/src/main/java/io/gravitee/gamma/authorization/service/CreateAuthzPolicyCommand.java
podlesrafal: Right, that was a clear regression — fixed. Went with your first option:

- Restored @Size(max = AuthzEntityIdConstants.MAX_ENTITY_ID_LENGTH) + @Pattern(regexp = AuthzEntityIdConstants.FORMAT_REGEX) on CreateAuthzPolicyCommand.entityId.
- Mirrored the same constraints on AuthzPolicyRequest.entityId (REST DTO) + added @NotBlank on name / @NotNull on kind and policyText.
- Same treatment on AuthzEntityRequest.entityId so the entity write path is covered too.
- Added @Valid to all request bodies on AuthzPoliciesResource and AuthzEntitiesResource (without it the DTO annotations were no-ops).
- New REST tests pin the @Valid → ConstraintViolationException → 400 contract for both blank and uppercase-format entityId on policies and entities.

Okhelifi (approved):

PR #16887, opened 2026-05-14
fix: accept uppercase selector type discriminators in v4 API PATCH flows

carlos-andres-osorio:
## Issue

[APIM-14077](https://gravitee.atlassian.net/browse/APIM-14077)

## Why

`PATCH /management/v2/environments/{env}/apis/{apiId}` rejects flow selectors with `"type": "HTTP"` (uppercase), even though both `GET` and `PUT` produce uppercase values. This means a client cannot copy a selector from a GET or PUT response and resubmit it in a PATCH without manually lowercasing the `type` field — a round-trip that should work transparently.

The root cause is that PATCH deserialization used the definition model's `@JsonSubTypes` labels (`http`, `channel`, `condition`, `mcp`) while PUT/GET use the OpenAPI-generated enum values (`HTTP`, `CHANNEL`, `CONDITION`, `MCP`).

## What changed

- `PatchApiUseCase` — `resolveFlows` now guards against a JSON-null `flows` node and delegates deserialization to `FlowListDeserializer` instead of inline Jackson binding, so selector type matching is case-insensitive.
- `FlowListDeserializer` — injected as a Spring bean with a shared `ObjectMapper` in the v1 management, v2 management, and portal Spring contexts (`RestManagementConfiguration`, `ResourceContextConfiguration`).

## How to test

1. Create or use an existing v4 HTTP Proxy API.
2. `GET /management/v2/environments/DEFAULT/apis/{apiId}` — note `flows[0].selectors[0].type` = `"HTTP"`.
3. Submit that exact response body (unmodified) as a `PATCH` on the same endpoint.
4. Expect `200` — previously this returned `400`.

[APIM-14077]: https://gravitee.atlassian.net/browse/APIM-14077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/spring/RestAutomationConfiguration.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/spring/RestAutomationConfiguration.java
@@ -15,9 +15,14 @@
  */
 package io.gravitee.apim.rest.api.automation.spring;
 
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.gravitee.apim.core.api.use_case.PatchApiUseCase.FlowListDeserializer;
 import io.gravitee.apim.infra.spring.UsecaseSpringConfiguration;
+import io.gravitee.definition.model.v4.flow.Flow;
 import io.gravitee.el.ExpressionLanguageInitializer;
 import io.gravitee.rest.api.service.spring.ServiceConfiguration;
+import java.util.List;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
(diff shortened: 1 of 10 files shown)


paulatulis (commented):
Test gaps
  - A test asserting "http" (lowercase) explicitly returns 400 with the same invalidValue envelope as "unknown-type". The renamed
  patch_with_lowercase_http_selector_type_returns_400_with_invalidValue_envelope does this — ignore, this is covered. ✓
  - JSON Patch add /flows/0/selectors/- with an uppercase-discriminator selector value (deep sub-pointer patch into flows). The new tests use whole-array replace /flows only; sub-pointer + new-deserializer behavior isn't exercised, and that's the path most likely to surface mapper edge cases (e.g., partial selector arrays).
  - Test that confirms "channel", "condition", "mcp" (lowercase) are also rejected — the lowercaseSelectorTypeVariants only varies HTTP. The bug surface is the same shape for the other three selector subtypes; would be cheap to add.

carlos-andres-osorio:
> Test gaps
>   - A test asserting "http" (lowercase) explicitly returns 400 with the same invalidValue envelope as "unknown-type". The renamed
>   patch_with_lowercase_http_selector_type_returns_400_with_invalidValue_envelope does this — ignore, this is covered. ✓
>   - JSON Patch add /flows/0/selectors/- with an uppercase-discriminator selector value (deep sub-pointer patch into flows). The new tests use whole-array replace /flows only; sub-pointer + new-deserializer behavior isn't exercised, and that's the path most likely to surface mapper edge cases (e.g., partial selector arrays).
>   - Test that confirms "channel", "condition", "mcp" (lowercase) are also rejected — the lowercaseSelectorTypeVariants only varies HTTP. The bug surface is the same shape for the other three selector subtypes; would be cheap to add.

Extended `lowercaseSelectorTypeVariants` to also cover `channel`, `condition`, and `mcp` — it now drives 8 argument sets (4 selector subtypes × 2 content-type variants) and the existing test assertion is unchanged.

For the sub-pointer path, added two new tests: `json_patch_add_sub_pointer_selector_with_uppercase_type_returns_200` (JSON Patch `add /flows/0/selectors/-` with `"type":"HTTP"` → asserts 200) and `json_patch_add_sub_pointer_selector_with_lowercase_type_returns_400_with_invalidValue_envelope` (same with `"type":"http"` → asserts 400 with `invalidValue`).

wbabyte (approved):

PR #16907, opened 2026-05-18
fix: remove invalid invitation search SQL escape clause

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/APIM-14102

## Description

https://app.circleci.com/pipelines/github/gravitee-io/gravitee-api-management/81400/workflows/5a1f4679-0187-4b47-8594-325e2087bf94/jobs/1710695/tests#failed-test-0

fix for mariadb and mysql8  regarding 

```
Caused by: org.springframework.jdbc.BadSqlGrammarException: PreparedStatementCallback; bad SQL grammar [select * from `test_gio_invitations` where reference_id = ? and reference_type = ? and lower(email) like ? escape '\' order by case when email is null then 1 else 0 end, lower(email) desc ]

```


dev-marek (approved):
lgtm

PR #16928, opened 2026-05-18
Feat/v4 wsdl import

konrad-krol-incubly:
## Issue

https://gravitee.atlassian.net/browse/EXT-103

## Description

  These commits introduce WSDL import support for v4 APIs, enabling both creating and updating APIs from WSDL definitions. Two import modes are available — raw SOAP (passthrough) and REST-to-SOAP (protocol translation) — with optional OpenAPI documentation and OAS validation, following the   
  same functional scope as the existing v2 implementation.
                                                                                                                                                                                                                                                                                                     
  Functional                                                                                                                                                                                                                                                                                       

  - Two import modes: raw SOAP (passthrough, no flows generated) and REST-to-SOAP (protocol translation with per-operation rest-to-soap + xml-json flows)
  - Two operations: create or update a v4 API from WSDL
  - REST-to-SOAP optional features: OpenAPI documentation page and OAS validation policy — only available in REST-to-SOAP mode
  - EL expressions in SOAP envelopes — leaf values use {#xmlEscape(#jsonPath(#request.content, '$.path'))} to extract from incoming JSON with XML escaping
  - Conditional flows — skipFlows=true in raw SOAP mode (no policies); flows only built for REST-to-SOAP
  - Deferred response validation — OAS validation on response is placed last so that xml-json converts the SOAP XML to JSON first, before validation against the OpenAPI schema can apply
  - Auto-added dependency — selecting rest-to-soap automatically appends xml-json policy
  - Validate Request / Apply XML Validation not ported — these v2 policies appear dead (never added during v2 WSDL import) and were omitted to keep WSDL import aligned with other import formats

  Technical

  - New endpoints: POST /v2/apis/_import/wsdl (create) and PUT /v2/apis/{apiId}/_import/wsdl (update)
  - New use cases: WsdlToImportApiUseCase and WsdlToUpdateApiUseCase — orchestrate WSDL parsing → OAI conversion → API create/update
  - New domain service: WsdlParserDomainService — wraps WSDLToOpenAPIConverter to parse WSDL into OpenAPI YAML

  Design decisions and constraints

  - XSD → JSON Schema (XsdToJsonSchemaConverter) — maps basic XSD primitives and nested structures; advanced constraints (patterns, enumerations, unions) not supported — same scope as v2
  - maxOccurs / arrays — converter does not produce JSON array types for maxOccurs > 1; known v2 limitation carried over
  - EL expression builder (ElPlaceholderXmlUtil) — supports flat/nested hierarchies; same v2 limitations for complex XSD constructs (choices, extensions, substitution groups)
  - v4 always uses EL placeholders — SoapMessageBuilder flag useElPlaceholders is always true for v4, replacing v2's hardcoded sample values with safe #xmlEscape expressions
  - Deferred validation trigger — only when format is WSDL and policies are present, preventing OAS validation from running against unconverted XML


## Rules compliant report

[rules-compliance-report.md](https://github.com/user-attachments/files/28090716/rules-compliance-report.md)

## Demonstration

https://github.com/user-attachments/assets/8c5fb6d7-fb6f-4a35-8343-c4209bc11167

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
@@ -40,6 +40,7 @@ import io.gravitee.apim.core.api.use_case.UpdateApiDefinitionFromImportUseCase;
 import io.gravitee.apim.core.api.use_case.UpdateApiGroupsUseCase;
 import io.gravitee.apim.core.api.use_case.UpdateFederatedApiUseCase;
 import io.gravitee.apim.core.api.use_case.UpdateNativeApiUseCase;
+import io.gravitee.apim.core.api.use_case.WsdlToUpdateApiUseCase;
 import io.gravitee.apim.core.audit.model.AuditActor;
 import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.apim.core.audit.model.Excludable;
@@ -71,6 +72,7 @@ import io.gravitee.rest.api.management.v2.rest.model.DuplicateApiOptions;
 import io.gravitee.rest.api.management.v2.rest.model.Error;
 import io.gravitee.rest.api.management.v2.rest.model.ExportApiV4;
 import io.gravitee.rest.api.management.v2.rest.model.ImportSwaggerDescriptor;
+import io.gravitee.rest.api.management.v2.rest.model.ImportWsdlDescriptor;
 import io.gravitee.rest.api.management.v2.rest.model.MigrationReportResponses;
 import io.gravitee.rest.api.management.v2.rest.model.MigrationReportResponsesIssuesInner;
 import io.gravitee.rest.api.management.v2.rest.model.MigrationStateType;
@@ -278,6 +280,9 @@ public class ApiResource extends AbstractResource {
     private OAIToUpdateApiUseCase oaiToUpdateApiUseCase;
 
     @Inject
+    private WsdlToUpdateApiUseCase wsdlToUpdateApiUseCase;
+
+    @Inject
     private UpdateApiGroupsUseCase updateApiGroupsUseCase;
 
     @Context
@@ -401,6 +406,31 @@ public class ApiResource extends AbstractResource {
                 .withPolicyPaths(Boolean.TRUE.equals(descriptor.getWithPolicyPaths()))
                 .auditInfo(audit)
                 .build()
+        );
+
+        boolean isSynchronized = apiStateService.isSynchronized(
+            GraviteeContext.getExecutionContext(),
+            getGenericApiEntityById(apiId, false)
+        );
+        return Response.ok().entity(ApiMapper.INSTANCE.map(output.apiWithFlows(), uriInfo, isSynchronized)).build();
+    }
+
+    @PUT
+    @Path("/_import/wsdl")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Permissions({ @Permission(value = RolePermission.API_DEFINITION, acls = RolePermissionAction.UPDATE) })
+    public Response updateApiFromWsdl(@PathParam("apiId") String apiId, @Valid @NotNull ImportWsdlDescriptor descriptor) {
+        var audit = getAuditInfo();
+        var output = wsdlToUpdateApiUseCase.execute(
+            WsdlToUpdateApiUseCase.Input.of(
+                apiId,
+                descriptor.getPayload(),
+                ImportWsdlDescriptor.TypeEnum.URL.equals(descriptor.getType()),
+                Boolean.TRUE.equals(descriptor.getWithDocumentation()),
+                Boolean.TRUE.equals(descriptor.getWithOASValidationPolicy()),
+                audit
+            )
         );
 
         boolean isSynchronized = apiStateService.isSynchronized(
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
@@ -174,6 +174,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -429,6 +430,7 @@ public class ApiResource extends AbstractResource {
                 ImportWsdlDescriptor.TypeEnum.URL.equals(descriptor.getType()),
                 Boolean.TRUE.equals(descriptor.getWithDocumentation()),
                 Boolean.TRUE.equals(descriptor.getWithOASValidationPolicy()),
+                Optional.ofNullable(descriptor.getWithPolicies()).map(ArrayList::new).orElse(null),
                 audit
             )
         );
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/WsdlToUpdateApiUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/WsdlToUpdateApiUseCase.java
@@ -20,15 +20,19 @@ import io.gravitee.apim.core.api.domain_service.WsdlParserDomainService;
 import io.gravitee.apim.core.api.model.ApiWithFlows;
 import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.rest.api.model.ImportSwaggerDescriptorEntity;
-import io.gravitee.rest.api.service.exceptions.SwaggerDescriptorException;
+import java.util.List;
+import java.util.stream.Stream;
+import lombok.RequiredArgsConstructor;
 
 @UseCase
+@RequiredArgsConstructor
 public class WsdlToUpdateApiUseCase {
 
     public sealed interface Input permits Input.Inline, Input.Url {
         String apiId();
         boolean withDocumentation();
         boolean withOASValidationPolicy();
+        List<String> withPolicies();
         AuditInfo auditInfo();
 
         record Inline(
@@ -36,11 +40,18 @@ public class WsdlToUpdateApiUseCase {
             String payload,
             boolean withDocumentation,
             boolean withOASValidationPolicy,
+            List<String> withPolicies,
             AuditInfo auditInfo
         ) implements Input {}
 
-        record Url(String apiId, String url, boolean withDocumentation, boolean withOASValidationPolicy, AuditInfo auditInfo) implements
-            Input {}
+        record Url(
+            String apiId,
+            String url,
+            boolean withDocumentation,
+            boolean withOASValidationPolicy,
+            List<String> withPolicies,
+            AuditInfo auditInfo
+        ) implements Input {}
 
         static Input of(
             String apiId,
@@ -48,11 +59,12 @@ public class WsdlToUpdateApiUseCase {
             boolean isUrl,
             boolean withDocumentation,
             boolean withOASValidationPolicy,
+            List<String> withPolicies,
             AuditInfo auditInfo
         ) {
             return isUrl
-                ? new Url(apiId, payload, withDocumentation, withOASValidationPolicy, auditInfo)
-                : new Inline(apiId, payload, withDocumentation, withOASValidationPolicy, auditInfo);
+                ? new Url(apiId, payload, withDocumentation, withOASValidationPolicy, withPolicies, auditInfo)
+                : new Inline(apiId, payload, withDocumentation, withOASValidationPolicy, withPolicies, auditInfo);
         }
     }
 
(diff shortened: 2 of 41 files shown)


konrad-krol-incubly:
Flaky test: [ERROR]   PortalPageContentsResourceTest.should_update_portal_page_content Expected response status to be <200> but was <500>

comes from master, being fixed by @dev-marek

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
```diff
@@ -388,7 +394,11 @@ public Response updateApiFromSwagger(@PathParam("apiId") String apiId, @Valid @N
         var audit = getAuditInfo();
         var importSwaggerDescriptor = ImportSwaggerDescriptorEntity.builder()
             .payload(descriptor.getPayload())
-            .withPolicies(descriptor.getWithPolicies() != null ? new ArrayList<>(descriptor.getWithPolicies()) : null)
+            .withPolicies(
+                Optional.ofNullable(descriptor.getWithPolicies())
+                    .map(policies -> policies.stream().toList())
```
vikrantgravitee: Could you please check this.
stream().toList() returns an immutable list — guaranteed UnsupportedOperationException crash on WSDL import — ApiResource.updateApiFromSwagger was changed from new ArrayList<>(descriptor.getWithPolicies()) to descriptor.getWithPolicies().stream().toList(). The resulting immutable list is passed into SwaggerServiceImpl, which calls .add("xml-json") on it when withPolicies contains "rest-to-soap" (line 132 of SwaggerServiceImpl). Every WSDL import with REST-to-SOAP policy will crash at runtime with UnsupportedOperationException.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/WsdlToUpdateApiUseCase.java
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.api.use_case;
+
+import static java.util.Collections.emptyList;
```
vikrantgravitee: some imports seems not used

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
```diff
@@ -410,6 +420,34 @@ public Response updateApiFromSwagger(@PathParam("apiId") String apiId, @Valid @N
...
+        var audit = getAuditInfo();
+        var output = wsdlToUpdateApiUseCase.execute(
+            WsdlToUpdateApiUseCase.Input.of(
+                apiId,
+                descriptor.getPayload(),
+                ImportWsdlDescriptor.TypeEnum.URL.equals(descriptor.getType()),
+                Boolean.TRUE.equals(descriptor.getWithDocumentation()),
+                Boolean.TRUE.equals(descriptor.getWithOASValidationPolicy()),
+                Optional.ofNullable(descriptor.getWithPolicies())
+                    .map(policies -> policies.stream().toList())
```
vikrantgravitee: Same validation/test might be here

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/WsdlToUpdateApiUseCase.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        var policies = addDependentPolicies(input.withPolicies());
+
+        var importSwaggerDescriptor = ImportSwaggerDescriptorEntity.builder()
+            .payload(openApiYaml)
+            .format(ImportSwaggerDescriptorEntity.Format.WSDL)
+            .withDocumentation(input.withDocumentation())
+            .withPolicies(policies)
+            .skipFlows(policies == null || policies.isEmpty())
```
vikrantgravitee: Could we ensure that it will not remove existing flows -

When the caller omits withPolicies or passes an empty list, skipFlows=true → the update chain receives an empty flows list → FlowCrudServiceImpl deletes all existing flows on the API.

Please compare this functionality with v2.

konrad-krol-incubly: @vikrantgravitee 

Investigated the v2 path — confirmed it preserves existing flows (skips `setFlows()` when `withPolicyPaths=false`). The v4 update path doesn't have this guard, so empty flows from `skipFlows=true` would replace existing ones.

  That said, I think replacing flows may actually be correct for v4 WSDL import. The user explicitly picks the mode: `REST-to-SOAP` (flows generated) or `Raw SOAP` (no flows). If someone re-imports a WSDL switching from `REST-to-SOAP` to `Raw SOAP`, keeping the old operation-specific flows doesn't
  make sense — SOAP services typically operate on a single root path, so those per-operation flows would be invalid anyway.

  Preserving only user-customized flows sounds appealing but there's no reliable way to distinguish them from import-generated ones.

  What's the expected behavior here? Should switching to Raw SOAP on update clear the existing flows (current behavior), or should we preserve them like v2 does?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImpl.java
```diff
@@ -246,17 +255,32 @@ private ImportDefinition addOASValidationPolicy(boolean withOASValidationPolicy,
...
                 .request(List.of(step))
-                .response(List.of(step))
+                .response(deferResponseValidation ? List.of() : List.of(step))
                 .build();
 
             List<Flow> apiExportFlows = (List<Flow>) importDefinition.getApiExport().getFlows();
             apiExportFlows.addFirst(flow);
 
-            importDefinition.getApiExport().setFlows(apiExportFlows);
+            if (deferResponseValidation) {
```
vikrantgravitee, line 264: setFlows is removed. What if ApiExport.getFlows() returns a copy, both addFirst and addLast are no-ops.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/sanitizer/UrlSanitizerUtils.java
```diff
@@ -30,6 +29,15 @@
  */
 public class UrlSanitizerUtils {
 
+    public static boolean isUrl(String content) {
```
vikrantgravitee, line 34: Is it still used?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/WsdlParserDomainServiceImpl.java
```diff
@@ -0,0 +1,52 @@
...
+        if (UrlSanitizerUtils.isUrl(content)) {
+            UrlSanitizerUtils.checkAllowed(
+                content,
+                importConfiguration.getImportWhitelist(),
+                importConfiguration.isAllowImportFromPrivate()
+            );
+        }
+        OpenAPI openAPI = new WsdlParser().parse(content);
+        if (openAPI == null) {
+            return null;
```
vikrantgravitee: How are we having parity for SwaggerDescriptorException

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
konrad-krol-incubly: fixed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/WsdlToUpdateApiUseCase.java
konrad-krol-incubly: fixed

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource.java
konrad-krol-incubly: fixed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImpl.java
konrad-krol-incubly: fixed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/WsdlParserDomainServiceImpl.java
konrad-krol-incubly: moved exception throw to domain service

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/sanitizer/UrlSanitizerUtils.java
konrad-krol-incubly: `UrlSanitizerUtils` sanitization in `WsdlParserDomainServiceImpl` is necessary because `WsdlParser` accepts both inline WSDL content and URLs — when given a URL, it fetches the remote resource itself. Without this check, a user could supply an arbitrary URL (e.g. pointing to internal network resources), leading to SSRF. This is the same pattern used in the v2 `SwaggerServiceImpl` for both WSDL and OAI imports.                                                                                                                                                           
                                  
  The v4 OAI import path doesn't need this because it's architecturally split: URL-based imports go through a separate use case that sanitizes the URL before fetching. By the time content reaches the OAI parser, it's always already-fetched inline content. The WSDL path can't use    
  this approach because the parser handles fetching internally.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/WsdlToUpdateApiUseCase.java
vikrantgravitee: @konrad-krol-incubly 
Makes sense, thanks for the context. The v2 SwaggerServiceImpl precedent and the parser fetching internally justify keeping it. Could we add a small unit test pinning the SSRF guard (e.g. file://, localhost, link-local) so it isn't dropped in a future refactor? Otherwise good to resolve.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/sanitizer/UrlSanitizerUtils.java
vikrantgravitee: @konrad-krol-incubly 

Agree the mode-switch case justifies replacing flows. For v2 parity, could we treat `withPolicies == null `as "no intent" and preserve existing flows, and only replace when the list is explicitly provided or the import mode changes? 
Small tweak to the `skipFlows(...) condition` + a test for the null case should cover it.

konrad-krol-incubly: done

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/WsdlToUpdateApiUseCase.java
konrad-krol-incubly: done

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/converter/oai/OAIToImportDefinitionConverter.java
```diff
@@ -50,7 +59,7 @@ public ImportDefinition toImportDefinition(OpenAPI specification, Collection<? e
             .definitionVersion(DefinitionVersion.V4)
             .apiVersion(specification.getInfo().getVersion())
             .type(ApiType.PROXY)
-            .flows(OAIToFlowsConverter.INSTANCE.convert(specification, visitors))
+            .flows(skipFlows ? List.of() : OAIToFlowsConverter.INSTANCE.convert(specification, visitors))
```
vikrantgravitee: Could we have it

`.flows(skipFlows ? new ArrayList<>() : OAIToFlowsConverter.INSTANCE.convert(specification, visitors))`

My reason:

List.of() is immutable & if a caller passes `withPolicies:[] + withOASValidationPolicy:true `directly via the REST API, addOASValidationPolicy reaches apiExportFlows.addFirst(flow) on that immutable list and throws UnsupportedOperationException (HTTP 500).

vikrantgravitee (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/converter/oai/OAIToImportDefinitionConverter.java
konrad-krol-incubly: You're right - while this scenario is unlikely (contradictory flags), I've applied the fix for a defensive approach in the code.

PR #16946, opened 2026-05-19
feat(gateway): expose module/api identity on OTel resource

remibaptistegio:
## Issue

N/A — producer-side enrichment for the upcoming TracingRepository work.

## Description

Every per-API Tracer created by the gateway now carries the following as **OTel resource attributes** (one value per Tracer instance, automatically applied to every span the Tracer emits — including child spans):

| Attribute | Value |
|---|---|
| `gravitee.module` | `apim` |
| `gravitee.api.id` | the API id |
| `gravitee.api.name` | the API display name |
| `gravitee.api.type` | reactor type — `API_V4` / `API_V4_TCP` / `API_V2` / `API_V2_EMULATED` |
| `gravitee.org.id` | the API's organization id |
| `gravitee.env.id` | the API's environment id |

Previously these were stamped on every root span via `withAttribute(...)` in `DefaultHttpRequestDispatcher` (which both bloated the span payload and missed child spans entirely). Moving them to the resource means a single value flows once per Tracer and applies uniformly to every span.

## Design choices

**Service identity flips from API-derived to gateway-derived.** The first four `createTracer` arguments (`service.instance.id`, `service.name`, `service.namespace`, `service.version`) were previously `(api.getId(), api.getName(), reactorType, api.getApiVersion())`. They're now `(node.id(), node.application(), "gravitee", Version.RUNTIME_VERSION.MAJOR_VERSION)`. Rationale: by OTel semantic conventions the `service.*` axis identifies the *producer* (the gateway process), not the API being processed. The API identity belongs on resource attributes — which is exactly where this PR puts them.

**`service.namespace = "gravitee"`** for all four Tracer callsites — gateway-wide bean + the three per-API factories. `"GATEWAY"` (previously used by the gateway-wide tracer) and `"API_V4"` / `"API_V4_TCP"` / `"API_V2"` / `"API_V2_EMULATED"` (previously used by the per-API factories) both abused `service.namespace`, which the spec defines as a logical grouping/owner of services — not the kind of service. `"gravitee"` groups the gateway with other Gravitee services for cross-service queries.

**Reactor-type discriminator preserved as `gravitee.api.type`** rather than dropped. Useful for "show me all V4 TCP traces in env X" style queries. Routed via the existing `serviceNameSpace` / `spanNamespace` parameter on `createTracingContext` (renamed to `apiType` on v4 HTTP) so subclasses / external reactor implementations don't break.

**Gateway-wide tracer (`OpenTelemetryConfiguration.gatewayTracer`) carries no `gravitee.module`.** It's used for cross-cutting paths (no-acceptor 404s, gateway lifecycle, organization reactor) that aren't module-scoped. Tagging it `apim` would mislabel ESM / AIM traffic running through the same gateway. A future global tracing view can either query without a module filter or add a `gravitee.module = gateway` value as a follow-up if needed.

## Files touched

- `OpenTelemetryConfiguration.java` — gateway-wide tracer's `service.namespace` becomes `gravitee`.
- `DefaultApiReactorFactory.java` (v4 HTTP) — service identity to node-derived, full resource map, parameter renamed to `apiType`.
- `TcpApiReactorFactory.java` (v4 TCP) — same.
- `ApiReactorHandlerFactory.java` (v2) — same; constructor's `Node node` parameter now wired through (previously passed `null` in test).
- `DefaultHttpRequestDispatcher.java` — drop the four `withAttribute` calls on the root span (now redundant).
- Three factory tests — stub `api.getOrganizationId()` / `api.getEnvironmentId()`, drop the unused `api.getApiVersion()` stub on v4 HTTP, add a `Node` mock to the v2 test.

## Additional context

PR 2 of 5 in the tracing read-side rollout. Independent of PR 1 (#16942 TracingRepository SPI — already merged) — producer and consumer can ship in either order. Subsequent PRs:

3. ElasticsearchTracingRepository impl + OTEL_TRACES scope wiring
4. TempoTracingRepository plugin + gravitee-node bump
5. gamma-module-apim TracingResource (will filter on `gravitee.env.id` resource attribute; relies on this PR)

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/ApiReactorHandlerFactory.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/ApiReactorHandlerFactory.java
@@ -17,6 +17,7 @@ package io.gravitee.gateway.handlers.api;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.gravitee.common.event.EventManager;
+import io.gravitee.common.util.Version;
 import io.gravitee.definition.model.DefinitionVersion;
 import io.gravitee.definition.model.ExecutionMode;
 import io.gravitee.el.TemplateVariableProvider;
@@ -53,6 +54,7 @@ import io.gravitee.gateway.handlers.api.processor.RequestProcessorChainFactory;
 import io.gravitee.gateway.handlers.api.processor.ResponseProcessorChainFactory;
 import io.gravitee.gateway.handlers.api.processor.transaction.TransactionResponseProcessorConfiguration;
 import io.gravitee.gateway.handlers.api.security.PlanBasedAuthenticationHandlerEnhancer;
+import io.gravitee.gateway.opentelemetry.TracerResourceAttributes;
 import io.gravitee.gateway.opentelemetry.TracingContext;
 import io.gravitee.gateway.platform.organization.manager.OrganizationManager;
 import io.gravitee.gateway.policy.PolicyChainProviderLoader;
@@ -120,6 +122,9 @@ public class ApiReactorHandlerFactory implements ReactorFactory<Api> {
     public static final String HANDLERS_REQUEST_HEADERS_X_FORWARDED_PREFIX_PROPERTY = "handlers.request.headers.x-forwarded-prefix";
     public static final String REPORTERS_LOGGING_EXCLUDED_RESPONSE_TYPES_PROPERTY = "reporters.logging.excluded_response_types";
     public static final String PENDING_REQUESTS_TIMEOUT_PROPERTY = "api.pending_requests_timeout";
+
+    /** OTel {@code gravitee.module} resource-attribute value identifying this reactor. */
+    private static final String MODULE = "apim";
 
     protected final ContentTemplateVariableProvider contentTemplateVariableProvider;
     protected final Node node;
@@ -454,13 +459,14 @@ public class ApiReactorHandlerFactory implements ReactorFactory<Api> {
         );
     }
 
-    private TracingContext createTracingContext(final Api api, final String spanNamespace) {
+    private TracingContext createTracingContext(final Api api, final String apiType) {
         Tracer tracer = openTelemetryFactory.createTracer(
-            api.getId(),
-            api.getName(),
-            spanNamespace,
-            api.getApiVersion(),
-            instrumenterTracerFactories
+            node.id(),
+            node.application(),
+            "gravitee",
+            Version.RUNTIME_VERSION.MAJOR_VERSION,
+            instrumenterTracerFactories,
+            TracerResourceAttributes.of(MODULE, api.getId(), api.getName(), apiType, api.getOrganizationId(), api.getEnvironmentId())
         );
         return new TracingContext(tracer, openTelemetryConfiguration.isTracesEnabled(), openTelemetryConfiguration.isVerboseEnabled());
     }
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
@@ -30,6 +30,7 @@ import io.gravitee.gateway.env.GatewayConfiguration;
 import io.gravitee.gateway.env.RequestTimeoutConfiguration;
 import io.gravitee.gateway.handlers.accesspoint.manager.AccessPointManager;
 import io.gravitee.gateway.handlers.api.registry.ApiProductRegistry;
+import io.gravitee.gateway.opentelemetry.TracerResourceAttributes;
 import io.gravitee.gateway.opentelemetry.TracingContext;
 import io.gravitee.gateway.platform.organization.manager.OrganizationManager;
 import io.gravitee.gateway.policy.PolicyConfigurationFactory;
@@ -71,7 +72,6 @@ import io.gravitee.plugin.entrypoint.EntrypointConnectorPluginManager;
 import io.gravitee.plugin.policy.PolicyClassLoaderFactory;
 import io.gravitee.plugin.policy.PolicyPlugin;
 import java.util.List;
-import java.util.Map;
 import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.core.ResolvableType;
@@ -81,6 +81,9 @@ import org.springframework.core.ResolvableType;
  * @author GraviteeSource Team
  */
 public class DefaultApiReactorFactory extends AbstractReactorFactory<Api> {
+
+    /** OTel {@code gravitee.module} resource-attribute value identifying this reactor. */
+    private static final String MODULE = "apim";
 
     protected final Node node;
     protected final EntrypointConnectorPluginManager entrypointConnectorPluginManager;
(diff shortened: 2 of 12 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-opentelemetry/src/main/java/io/gravitee/gateway/opentelemetry/TracerResourceAttributes.java
```diff
@@ -0,0 +1,71 @@
...
+    public static Map<String, String> of(
+        final String module,
+        final String apiId,
+        final String apiName,
+        final String apiType,
+        final String orgId,
+        final String envId
+    ) {
+        Map<String, String> attrs = new LinkedHashMap<>();
+        attrs.put("gravitee.module", module);
```
SamHardi, line 62: `module` is the only parameter not null-guarded here — the five optional parameters go through `putIfNotNull`, but this line would silently insert `"gravitee.module" → null` into the map if a caller passes `null`. Since the Javadoc already documents it as "Required (non-null)", a defensive guard at the top of the method would enforce the contract and produce a clear NPE with context rather than a silent bad value downstream:

```java
Objects.requireNonNull(module, "module must not be null");
```

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactory.java
```diff
@@ -421,16 +421,16 @@ protected ApiProductPlanPolicyManagerFactory createApiProductPlanPolicyManagerFa
...
         if (tracingEnabled) {
             Tracer tracer = openTelemetryFactory.createTracer(
                 node.id(),
                 node.application(),
-                serviceNameSpace,
+                "gravitee",
                 Version.RUNTIME_VERSION.MAJOR_VERSION,
                 instrumenterTracerFactories,
-                Map.of("gravitee.api.id", api.getId(), "gravitee.api.name", api.getName()),
+                TracerResourceAttributes.of("apim", api.getId(), api.getName(), apiType, api.getOrganizationId(), api.getEnvironmentId()),
```
SamHardi: `"apim"` is duplicated across the three APIM factories (here, `TcpApiReactorFactory`, `ApiReactorHandlerFactory`). I would not put it as a constant inside `TracerResourceAttributes` — that would turn a generic utility class into a module registry and ESM/AIM reactors would need it to know about APIM-specific identifiers.

A better fit is a private constant owned by each factory (or a shared APIM constants class):

```java
private static final String MODULE = "apim";
// ...
TracerResourceAttributes.of(MODULE, api.getId(), ...)
```

That way `TracerResourceAttributes` stays fully module-agnostic, and each future reactor (ESM, AIM) owns its own identifier without touching this class.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/TcpApiReactorFactory.java
```diff
@@ -100,10 +100,17 @@ protected TracingContext createTracingContext(final Api api) {
             Tracer tracer = openTelemetryFactory.createTracer(
                 node.id(),
                 node.application(),
-                "API_V4_TCP",
+                "gravitee",
                 Version.RUNTIME_VERSION.MAJOR_VERSION,
                 instrumenterTracerFactories,
-                Map.of("gravitee.api.id", api.getId(), "gravitee.api.name", api.getName()),
+                TracerResourceAttributes.of(
+                    "apim",
```
SamHardi: Same as above — `"apim"` should be a private constant in this factory rather than a literal. See the comment on `DefaultApiReactorFactory`.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/ApiReactorHandlerFactory.java
```diff
@@ -454,13 +456,14 @@ protected ApiReactorHandler getApiReactorHandler(
...
-            api.getName(),
-            spanNamespace,
-            api.getApiVersion(),
-            instrumenterTracerFactories
+            node.id(),
+            node.application(),
+            "gravitee",
+            Version.RUNTIME_VERSION.MAJOR_VERSION,
+            instrumenterTracerFactories,
+            TracerResourceAttributes.of("apim", api.getId(), api.getName(), apiType, api.getOrganizationId(), api.getEnvironmentId())
```
SamHardi: Same as above — `"apim"` should be a private constant in this factory rather than a literal. See the comment on `DefaultApiReactorFactory`.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/test/java/io/gravitee/gateway/handlers/api/ApiReactorHandlerFactoryTest.java
```diff
@@ -169,7 +174,19 @@ public void shouldUseV3ExecutionMode() {
...
         when(api.getDefinition()).thenReturn(definition);
         when(definition.getExecutionMode()).thenReturn(ExecutionMode.V3);
+        stubApiIdentityForTracing();
         ReactorHandler handler = apiContextHandlerFactory.create(api);
         assertThat(handler).isInstanceOf(ApiReactorHandler.class);
     }
+
+    /**
+     * The v2 factory builds the OTel resource-attribute map unconditionally inside {@code createTracingContext},
+     * so every test that exercises {@code create(api)} needs these stubs — {@code Map.of} rejects null values.
```
SamHardi: This comment references `Map.of` rejecting null values, which was the original implementation. `TracerResourceAttributes` now uses `LinkedHashMap` + `putIfNotNull` to handle nullable fields precisely to avoid that. The real reason these stubs are needed is that the factory unconditionally calls `api.getId()`, `api.getName()`, etc. to build the attribute map — the comment is worth updating to reflect the current implementation.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/test/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactorFactoryTest.java
```diff
@@ -437,15 +438,17 @@ void should_pass_redaction_config_to_createTracer() {
...
-            ).thenReturn(mock(Tracer.class));
+            when(openTelemetryFactory.createTracer(any(), any(), any(), any(), any(), anyMap(), any(RedactionConfig.class))).thenReturn(
+                mock(Tracer.class)
+            );
 
             cut.createTracingContext(api, "test-namespace");
 
             ArgumentCaptor<RedactionConfig> captor = ArgumentCaptor.forClass(RedactionConfig.class);
-            verify(openTelemetryFactory).createTracer(any(), any(), any(), any(), any(), any(Map.class), captor.capture());
+            verify(openTelemetryFactory).createTracer(any(), any(), any(), any(), any(), anyMap(), captor.capture());
```
SamHardi, line 453: The resource-attribute argument is matched with `anyMap()`, so the test doesn't validate what's actually inside the map. Since exposing the right attributes is the core goal of this PR, it would be worth adding an `ArgumentCaptor<Map<String, String>>` to assert the expected values:

```java
ArgumentCaptor<Map<String, String>> attrsCaptor = ArgumentCaptor.forClass(Map.class);
verify(openTelemetryFactory).createTracer(any(), any(), any(), any(), any(), attrsCaptor.capture(), any());
assertThat(attrsCaptor.getValue()).containsEntry("gravitee.module", "apim");
assertThat(attrsCaptor.getValue()).containsEntry("gravitee.api.id", "api-id");
```

Same applies to `TcpApiReactorFactoryTest`.

remibaptistegio: New test added to address this concern.

SamHardi (approved):

PR #16950, opened 2026-05-19
fix(portal): scope user permissions to current environment (APIM-13995)

Uzma-Mahmood:
**Issue**

https://gravitee.atlassian.net/browse/APIM-13995

**Description**

Fixes environment role permissions leaking across Developer Portal environments.

UserMapper was building the user permissions payload by merging all environment roles from user.getRoles(). When a user had different roles per environment (e.g. create access in DEV, read-only in TEST), putAll could overwrite permissions so DEV lost APPLICATION: C after TEST was assigned.

The mapper now:

Keeps organization roles from user.getRoles()
Uses only roles from user.getEnvRoles() for GraviteeContext.getCurrentEnvironment()
Falls back to the previous flat user.getRoles() behavior when envRoles is not populated
A unit test covers the multi-environment scenario from the ticket.

**Additional context**

Reproduction (before fix):

Create two environments (e.g. DEV and TEST).
Create environment roles:
LE_GROUP — APPLICATION create enabled
UE_GROUP — APPLICATION read-only
Assign a user LE_GROUP on DEV only → they can create applications in the portal for DEV.
Also assign UE_GROUP on TEST → they can no longer create applications in DEV (UI hides create; permissions merged incorrectly).

**Expected**: Permissions in the portal reflect only the current environment (plus org roles).

**Root cause**: UserServiceImpl already exposes per-environment roles via envRoles, but UserMapper ignored that map and merged every environment role into one permission set.

**UI impact**: Portal Next Gen (canCreate from permissions.APPLICATION) and classic portal use this API response for the create-application action.

**Server-side note**: PermissionServiceImpl.hasPermission() already checks the current environment for API calls; this change aligns the returned user permissions with that behavior.

Pre-fix behaviour  : 

https://github.com/user-attachments/assets/3cc49d0d-a13e-47c8-9b75-857ae8ee2132

Post fix behaviour: 

https://github.com/user-attachments/assets/838d5c6a-f9af-48f3-8638-46428616f428

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/mapper/UserMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/mapper/UserMapper.java
@@ -21,11 +21,16 @@ import io.gravitee.rest.api.idp.api.identity.SearchableUser;
 import io.gravitee.rest.api.model.*;
 import io.gravitee.rest.api.model.permissions.RoleScope;
 import io.gravitee.rest.api.portal.rest.model.*;
+import io.gravitee.rest.api.service.common.GraviteeContext;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.springframework.stereotype.Component;
 
 /**
@@ -57,10 +62,19 @@ public class UserMapper {
         userItem.setId(user.getId());
         userItem.setEditableProfile(IDP_SOURCE_GRAVITEE.equals(user.getSource()) || IDP_SOURCE_MEMORY.equalsIgnoreCase(user.getSource()));
         if (user.getRoles() != null) {
-            Map<String, List<String>> userPermissions = user
+            Stream<UserRoleEntity> orgRoles = user
                 .getRoles()
                 .stream()
-                .filter(role -> RoleScope.ENVIRONMENT.equals(role.getScope()) || RoleScope.ORGANIZATION.equals(role.getScope()))
+                .filter(role -> RoleScope.ORGANIZATION.equals(role.getScope()));
+
+            Map<String, Set<UserRoleEntity>> envRolesMap = user.getEnvRoles();
+            String environmentId = GraviteeContext.getCurrentEnvironment();
+
+            Set<UserRoleEntity> currentEnvironmentRoles = (envRolesMap == null || environmentId == null)
+                ? Set.of()
+                : envRolesMap.getOrDefault(environmentId, Set.of());
+
+            Map<String, List<String>> userPermissions = Stream.concat(orgRoles, currentEnvironmentRoles.stream())
                 .map(UserRoleEntity::getPermissions)
                 .map(rolePermissions ->
                     rolePermissions
@@ -77,7 +91,13 @@ public class UserMapper {
                         )
                 )
                 .reduce(new HashMap<>(), (acc, rolePermissions) -> {
-                    acc.putAll(rolePermissions);
+                    rolePermissions.forEach((key, permissionChars) ->
+                        acc.merge(key, new ArrayList<>(permissionChars), (existing, incoming) -> {
+                            LinkedHashSet<String> merged = new LinkedHashSet<>(existing);
+                            merged.addAll(incoming);
+                            return new ArrayList<>(merged);
+                        })
+                    );
                     return acc;
                 });
             userItem.setPermissions(objectMapper.convertValue(userPermissions, UserPermissions.class));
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/mapper/UserMapper.java
```diff
@@ -56,34 +59,68 @@ public User convert(UserEntity user) {
...
-                    return acc;
-                });
-            userItem.setPermissions(objectMapper.convertValue(userPermissions, UserPermissions.class));
+                .filter(role -> RoleScope.ORGANIZATION.equals(role.getScope()))
+                .forEach(rolesForPermissions::add);
         }
-        userItem.setCustomFields(user.getCustomFields());
-        return userItem;
+
+        String environmentId = GraviteeContext.getCurrentEnvironment();
```
vikrantgravitee: Could we ensure that this is not breaking anything and not cause issue in next check -

Set<UserRoleEntity> currentEnvironmentRoles = user.getEnvRoles().get(environmentId);

A valid unit test or manual testing would be required to confirm this usecase.

vikrantgravitee: So in place of:

 if (user.getRoles() != null) {
            Map<String, List<String>> userPermissions = user
                .getRoles()
                .stream()
                .filter(role -> RoleScope.ENVIRONMENT.equals(role.getScope()) || RoleScope.ORGANIZATION.equals(role.getScope()))

If we can have:

if (user.getRoles() != null) {
            Stream<UserRoleEntity> orgRoles = user
                .getRoles()
                .stream()
                .filter(role -> RoleScope.ORGANIZATION.equals(role.getScope()));

            Map<String, Set<UserRoleEntity>> envRolesMap = user.getEnvRoles();
            String currentEnv = GraviteeContext.getCurrentEnvironment();
            Stream<UserRoleEntity> envRoles;
            if (envRolesMap != null && envRolesMap.containsKey(currentEnv)) {
                envRoles = envRolesMap.get(currentEnv).stream();
            } else {
                // Fallback: flat set (legacy callers that do not populate envRoles)
                envRoles = user
                    .getRoles()
                    .stream()
                    .filter(role -> RoleScope.ENVIRONMENT.equals(role.getScope()));
            }

            Map<String, List<String>> userPermissions = Stream.concat(orgRoles, envRoles)

prateektiwari-gravitee: Yes, this makes more sense. I have also added the test case.

aditya-goyal01 (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/mapper/UserMapper.java
```diff
@@ -57,10 +62,23 @@ public User convert(UserEntity user) {
         userItem.setId(user.getId());
         userItem.setEditableProfile(IDP_SOURCE_GRAVITEE.equals(user.getSource()) || IDP_SOURCE_MEMORY.equalsIgnoreCase(user.getSource()));
         if (user.getRoles() != null) {
-            Map<String, List<String>> userPermissions = user
+            Stream<UserRoleEntity> orgRoles = user
                 .getRoles()
                 .stream()
-                .filter(role -> RoleScope.ENVIRONMENT.equals(role.getScope()) || RoleScope.ORGANIZATION.equals(role.getScope()))
+                .filter(role -> RoleScope.ORGANIZATION.equals(role.getScope()));
```
vikrantgravitee, line 68: Small suggestion

public User convert(UserEntity user) {
    final User userItem = new User();
    userItem.setEmail(user.getEmail());
    userItem.setFirstName(user.getFirstname());
    userItem.setLastName(user.getLastname());
    userItem.setDisplayName(user.getDisplayName());
    userItem.setId(user.getId());
    userItem.setEditableProfile(IDP_SOURCE_GRAVITEE.equals(user.getSource()) || IDP_SOURCE_MEMORY.equalsIgnoreCase(user.getSource()));
    if (user.getRoles() != null) {
        Stream<UserRoleEntity> orgRoles = user
            .getRoles()
            .stream()
            .filter(role -> RoleScope.ORGANIZATION.equals(role.getScope()));
        Map<String, List<String>> userPermissions = Stream.concat(orgRoles, currentEnvironmentRoles(user).stream())
            .map(UserRoleEntity::getPermissions)
            .map(rolePermissions ->
                rolePermissions
                    .entrySet()
                    .stream()
                    .collect(
                        Collectors.toMap(Map.Entry::getKey, entry ->
                            new String(entry.getValue()).chars().mapToObj(c -> String.valueOf((char) c)).collect(Collectors.toList())
                        )
                    )
            )
            .reduce(new HashMap<>(), (acc, rolePermissions) -> {
                rolePermissions.forEach((key, permissionChars) ->
                    acc.merge(key, new ArrayList<>(permissionChars), (existing, incoming) -> {
                        LinkedHashSet<String> merged = new LinkedHashSet<>(existing);
                        merged.addAll(incoming);
                        return new ArrayList<>(merged);
                    })
                );
                return acc;
            });
        userItem.setPermissions(objectMapper.convertValue(userPermissions, UserPermissions.class));
    }
    userItem.setCustomFields(user.getCustomFields());
    return userItem;
}


private Set<UserRoleEntity> currentEnvironmentRoles(UserEntity user) {
    Map<String, Set<UserRoleEntity>> envRoles = user.getEnvRoles();
    if (envRoles == null) {
        return Set.of();
    }
    return envRoles.getOrDefault(GraviteeContext.getCurrentEnvironment(), Set.of());
}

PR #16957, opened 2026-05-20
fix: openapi update to use group ids in generic api schema

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-13693

## Description

Align OpenAPI docs with actual API behaviour: groups accepts group identifiers, not group display names.

image


pragsty:
Needs to be raised for 4.11, hence closing this

PR #16970, opened 2026-05-20
fix(audit): chunk retention deletes to avoid Mongo timeout

wbabyte:
Fixes [APIM-14089](https://gravitee.atlassian.net/browse/APIM-14089).

## Problem

`AuditMongoRepositoryImpl.deleteByEnvironmentIdAndAge` runs a single `mongoTemplate.remove(query)` matching `environmentId == X AND createdAt < limit`. On populated `audits` collections, this op exceeds the Mongo driver socket read timeout and surfaces as:

```
DataAccessResourceFailureException: Timeout while receiving message
Caused by: java.net.SocketTimeoutException: Read timed out
```

Retention is then silently not enforced. Every hourly tick fails the same way, the working set grows, the next tick fails on an even bigger one.

## Fix

1. **New index** `e1c1` = `{environmentId:1, createdAt:1}` on `audits` via a new `AuditsEnvironmentIdCreatedAtIndexUpgrader`. The cleaner query is now bounded by both predicates directly instead of scanning the date or environment partition. Additive — does not touch existing indexes.
2. **Chunked deletes** — `AuditMongoRepositoryImpl.deleteByEnvironmentIdAndAge` now pages through results with `find().projection(_id).limit(batchSize)` then `deleteMany({_id: $in})` per chunk. Each round-trip is bounded below the socket read timeout regardless of backlog. Configurable via `services.audit.cleanup.batchSize` (default `1000`).
3. **Per-env failure isolation** — `ScheduledAuditCleanerService.run()` wraps the per-environment call in try/catch so one environment's failure no longer skips every subsequent environment in the same tick (a defect masked by the original timeout).

## Out of scope

- **JDBC backend** has a similar unchunked `deleteByEnvironmentIdAndAge` path (`JdbcAuditRepository:336-348`). No customer report on JDBC yet; left as a separate concern.
- **TTL index** on `createdAt` (would retire the cron entirely) deferred — not opening a follow-up.
- `create-index.js` bootstrap script intentionally not updated; bootstrap/upgrader drift accepted.

## Operator notice

First startup after deploy on backlogged clusters will block while the new `e1c1` index is built — duration scales with the existing `audits` collection size. Index build runs in background mode server-side (writes proceed during build) but the upgrader caller blocks until completion, same as every other Gravitee index upgrader. Look for `Starting creation of index e1c1 on audits` in mAPI logs.

## Tests

- New `AuditMongoRepositoryImplTest` — 3 tests covering the chunking loop (pages through results, terminates on empty, deletes by `_id`).
- New test in `ScheduledAuditCleanerServiceTest` — verifies env2 still runs when env1 throws.
- Cross-backend regression tests `shouldRemoveTooOldData` / `shouldNotRemoveYoungData` continue to pass.

## Manual smoke

Verified locally with apim-worktree:

1. Seeded `gravitee_apim_14089.audits` with 2500 docs (createdAt = 2 days ago) + 5 young docs in env `DEFAULT`.
2. Configured `services.audit.cron: */15 * * * * *`, `services.audit.retention.days: 1`, `services.audit.cleanup.batchSize: 100`.
3. Started mAPI; observed cleaner ticking every 15s with no `SocketTimeoutException`.
4. Confirmed: `e1c1` index present, 2500 old docs deleted (25 chunks), 5 young docs preserved, 0 audit-related errors in `mapi.log`.

## Test plan
- [ ] CI green
- [ ] Index `e1c1` visible on `audits` collection post-deploy on a populated cluster
- [ ] `audit-cleaner-1` thread no longer emits `SocketTimeoutException`
- [ ] Existing retention behavior unchanged on small collections

[APIM-14089]: https://gravitee.atlassian.net/browse/APIM-14089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/internal/audit/AuditMongoRepositoryImpl.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/internal/audit/AuditMongoRepositoryImpl.java
@@ -17,6 +17,7 @@ package io.gravitee.repository.mongodb.management.internal.audit;
 
 import static org.springframework.data.mongodb.core.query.Criteria.where;
 
+import com.mongodb.client.result.DeleteResult;
 import io.gravitee.common.data.domain.Page;
 import io.gravitee.repository.management.api.search.AuditCriteria;
 import io.gravitee.repository.management.api.search.Pageable;
@@ -24,7 +25,9 @@ import io.gravitee.repository.mongodb.management.internal.model.AuditMongo;
 import java.time.Instant;
 import java.util.Date;
 import java.util.List;
+import lombok.CustomLog;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.mongodb.core.MongoTemplate;
@@ -35,10 +38,14 @@ import org.springframework.data.mongodb.core.query.Query;
  * @author Nicolas GERAUD (nicolas.geraud at graviteesource.com)
  * @author GraviteeSource Team
  */
+@CustomLog
 public class AuditMongoRepositoryImpl implements AuditMongoRepositoryCustom {
 
     @Autowired
     private MongoTemplate mongoTemplate;
+
+    @Value("${services.audit.cleanup.batchSize:1000}")
+    private int batchSize;
 
     @Override
     public Page<AuditMongo> search(AuditCriteria filter, Pageable pageable) {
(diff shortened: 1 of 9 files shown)


gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/internal/audit/AuditMongoRepositoryImpl.java
```diff
@@ -89,6 +95,21 @@ public Page<AuditMongo> search(AuditCriteria filter, Pageable pageable) {
 
     public void deleteByEnvironmentIdAndAge(String environmentId, Instant limit) {
         var criteria = where("environmentId").is(environmentId).andOperator(where("createdAt").lt(Date.from(limit)));
-        mongoTemplate.remove(new Query(criteria), AuditMongo.class);
+        Query chunkQuery = new Query(criteria).limit(batchSize);
+        chunkQuery.fields().include("_id");
+
+        int totalDeleted = 0;
+        while (true) {
```
phiz71: I'm always afraid of 'while(true)'. Do you think we could imagine another way to batch this deletion in an other way?
Could we use "bulkWrite" ?

phiz71 (commented):
Should we add a note in the Helm Changelog (Chart.yml)?

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/internal/audit/AuditMongoRepositoryImpl.java
wbabyte: done

phiz71 (approved):
LGTM

PR #16971, opened 2026-05-20
Enforce and configure cron schedule limits for various features

gutek9:
## Issue

https://gravitee.atlassian.net/browse/APIM-14126

## Description

Adds configurable minimum schedule limits for cron-based management API jobs and rejects user configurations that run more frequently than those limits.

This covers dynamic properties, dynamic dictionaries, documentation auto-fetch, and health checks. Runtime scheduling also keeps the slower schedule as a fallback safeguard.

## Additional context

Tested with targeted Maven tests in a Java 21 container for the updated validation paths and scheduler runtime behavior.

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/component/health-check-form/api-health-check-form.component.html
+++ b/gravitee-apim-console-webui/src/management/api/component/health-check-form/api-health-check-form.component.html
@@ -88,8 +88,13 @@
       <div class="health-check-card__trigger">
         <gio-form-cron class="health-check-card__trigger__field" formControlName="schedule">
           <gio-form-cron-label>Schedule *</gio-form-cron-label>
-          <gio-form-cron-hint>Cron expression to schedule the health check</gio-form-cron-hint>
+          <gio-form-cron-hint>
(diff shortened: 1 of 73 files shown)


aditya-goyal01:
**Issue 1:** One sample

- To decide how often a cron fires, the code looks at **one gap**: it picks a fixed starting point (midnight), finds the next two firings, and measures the time between them. One sample.
- That works for **regular** schedules ("every 2 hours" → every gap is 2h). It **breaks** for **clustered/irregular** ones.

Example: `0 0 1,23 * * *` fires at **01:00 and 23:00**.
- The real shortest gap is **2 hours** (23:00 → 01:00 next day).
- But the code, starting at midnight, samples 01:00 → 23:00 = **22 hours apart** and concludes "this is very infrequent, allow it."
- So under a 3-hour limit, a job that's *actually* 2 hours apart **sneaks through.**

**Issue 2:** Limit is itself a cron string (`cron_limit`). Two problems with how it's handled-

- It's never validated in one place — it gets re-parsed on **every single request** instead of being checked once when the system starts.
- A bad or "never-fires" cron makes the code throw an exception, instead of handling it gracefully.

PR #16974, opened 2026-05-20
feat(service): user create/update methods

dev-marek:
## Issue

https://gravitee.atlassian.net/browse/APIM-14130

## Description

- Added `create`, `update`, `updateAndSetPassword`, and `isPasswordSet` to `UserCrudServiceImpl`
- Added `UserAdapter.toUser(BaseUserEntity)` MapStruct mapping; password and user-only fields (`lastConnectionAt`, `picture`, etc.) are explicitly ignored so they are never overwritten via this path
- `updateAndSetPassword` writes user + encoded password in a single repository call without exposing the hash on `BaseUserEntity` (which lives in the Lucene search index)
- `isPasswordSet` queries the repository directly for the same reason — keeps the password off the domain entity entirely
- 14 new unit tests covering all four methods including null/blank password edge cases and error propagation

## Additional context

Password is intentionally absent from `BaseUserEntity` — adding it would expose the hash via the search index. `updateAndSetPassword` and `isPasswordSet` are the direct consequence of that constraint. Decision documented in the APIM-14130 design note.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/crud_service/user/UserCrudServiceImplTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/crud_service/user/UserCrudServiceImplTest.java
@@ -15,25 +15,21 @@
  */
 package io.gravitee.apim.infra.crud_service.user;
 
+import static fixtures.core.model.BaseUserEntityFixtures.aBaseUserEntity;
+import static fixtures.repository.UserFixtures.aRepositoryUser;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.catchThrowable;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
 
+import inmemory.UserRepositoryInMemory;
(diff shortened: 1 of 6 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/crud_service/user/UserCrudServiceImplTest.java
```diff
@@ -212,6 +214,213 @@ void should_throw_when_technical_exception_occurs() throws TechnicalException {
...
+    @Nested
+    class Create {
+
+        @Test
+        void should_create_user_and_return_adapted_entity() throws TechnicalException {
+            // Given
+            var userToCreate = BaseUserEntity.builder()
+                .id("user-id")
+                .organizationId("organization-id")
+                .source("gravitee")
```
jaroslawlakomy-grav: maybe we should have some fixture for that

jaroslawlakomy-grav (approved):
minor sugestiom, lgtm

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/crud_service/user/UserCrudServiceImplTest.java
dev-marek: done!

PR #16990, opened 2026-05-21
feat(repository,rest-api): Elasticsearch impl of TracingRepository + OTEL_TRACES scope

remibaptistegio:
## Issue

N/A — third step of the tracing read-side rollout. Builds on PR 1 (#16942 — TracingRepository SPI, merged).

## Description

Adds an Elasticsearch-backed \`TracingRepository\` reading OTel-native spans (\`mapping.mode: otel\`) written by the OTel collector's \`elasticsearch\` exporter (self-managed) or Elastic Cloud's managed OTLP endpoint. Wires the new \`OTEL_TRACES\` repository scope through the rest-api host so the plugin loads cleanly.

**What this PR does**
- New \`ElasticsearchTracingRepository\` parses snake_case OTel documents (\`trace_id\`, \`span_id\`, \`parent_span_id\`, \`@timestamp\`, \`duration\`, \`status.code\`, \`attributes.*\`, \`resource.attributes.*\`).
- \`searchTraces\` aggregates by \`trace_id\` ordered by trace start desc; sub-aggs grab the root span via \`top_hits\`, the start time via \`min(@timestamp)\`, and the per-trace error count via a filter agg on \`status.code\`.
- \`getTrace\` returns the full span tree for a single id; resource-attribute filters are emitted as resource-side \`term\` clauses (defence in depth on top of the per-tenant index).
- Multi-tenancy: the index template is resolved per call from \`QueryContext\` via the analytics convention (\`IndexNameUtils.format\`), e.g. \`traces-apim.otel-{orgId}\` → \`traces-apim.otel-default\` (lowercased).
- Plugin wiring: \`ElasticsearchRepositoryProvider\` adds \`OTEL_TRACES\` to its scopes; a separate \`TracingConfiguration\` Spring class avoids cross-scope bean-name collisions when both \`ANALYTICS\` and \`OTEL_TRACES\` are wired to this plugin.
- Host wiring: \`RestApiRepositoryScopeProvider\` registers \`OTEL_TRACES\` via \`getOptionalHandledScopes()\` so an APIM deployment without tracing configured still starts up.

**What this PR deliberately does NOT do — split into follow-up PRs**
- *Span events / logs data stream*: the OTel ES exporter writes span events as separate log documents (it forces \`data_stream.type = logs\` for events). Fetching and stitching those is the future \`LogsRepository\`'s job (\`OTEL_LOGS\` scope). This repo returns spans only.
- *Calling both TracingRepository + LogsRepository to assemble a complete trace*: lives in the consumer use case (gamma APIM, PR 6).

## Design choices

**Per-scope config block, no fallback to analytics** — \`otel-traces.elasticsearch.*\` is a fully independent property block (endpoints, security, SSL, HTTP proxy/timeout, index template). Mirrors the Redis plugin's pattern where \`ratelimit.redis.*\` and \`distributed-sync.redis.*\` are standalone. Tracing can target a different ES cluster than analytics (or none at all).

\`\`\`yaml
otel-traces:
  type: elasticsearch
  elasticsearch:
    index: traces-apim.otel-{orgId}
    endpoints:
      - http://localhost:9200
    # security, ssl, http: same shape as the analytics block
\`\`\`

**Index naming via \`IndexNameUtils.format\`** — same regex-based placeholder substitution + lowercasing convention as the analytics plugin (substituted values from \`QueryContext.placeholder()\` get \`.toLowerCase()\` applied so a caller orgId of \`DEFAULT\` resolves to \`traces-apim.otel-default\`, matching what the OTel collector writes).

**No \`IndexNameGenerator\` machinery** — the analytics generators (\`MultiType\` / \`PerType\` / \`ILMIndexNameGenerator\`) all hardcode a \`-<type>\` suffix or \`-<date>\` rotation, neither of which fits OTel data streams (\`traces-<dataset>-<namespace>\` shape where namespace = our orgId, the last segment). The substitution primitive is reused; the assembler isn't. See README "Index templates, mappings & retention (ILM)" for the operator-facing detail.

**Reader-only — operators own mappings + ILM** — the OTel collector (or Elastic Cloud managed OTLP endpoint) applies index templates and field mappings on write. ILM policies are operator-configured via \`index.lifecycle.name\` on the template. This repository neither authors nor maintains them.

## Files touched

- \`gravitee-apim-repository-elasticsearch\`:
  - New: \`tracing/ElasticsearchTracingRepository\`, \`tracing/configuration/TracingInfraConfiguration\`, \`tracing/spring/TracingConfiguration\` (per-scope Spring config with \`tracing\`-prefixed bean names to dodge cross-scope conflicts).
  - Modified: \`ElasticsearchRepositoryProvider\` (declares \`OTEL_TRACES\`), \`ElasticsearchRepositoryConfiguration\` (exposes \`buildHttpClient\` as \`public static\` for reuse).
  - Tests: 11 unit tests + a Testcontainers integration test (ES + OTel Collector on a shared docker network) extending the shared \`TracingRepositoryTest\` contract.
- \`gravitee-apim-rest-api-repository\`: \`RestApiRepositoryScopeProvider\` registers \`OTEL_TRACES\` as optional. New unit test mirrors the gateway-side equivalent.
- \`helm/\`: new \`api.otelTraces.*\` values block (defaults to \`type: none\`); configmap rendering is type-agnostic so a future \`type: tempo\` (PR 4) plugs in without restructuring the template. Six helm unittest scenarios cover default, ES-with-endpoints, security, SSL, http.timeout, and the type-agnostic future-tempo path.

## Additional context

This is **PR 3 of 6** in the tracing read-side rollout (revised plan after introducing the intermediate \`OTEL_LOGS\` scope PR):

1. ✅ TracingRepository SPI (#16942 — merged)
2. 🟡 Gateway producer enrichment (#16946 — in review)
3. ✅ This PR — ES TracingRepository + OTEL_TRACES scope wiring
4. ⬜ TempoTracingRepository plugin + gravitee-node bump
5. ⬜ \`OTEL_LOGS\` scope + LogsRepository SPI + ES impl (intermediate PR)
6. ⬜ gamma-module-apim TracingResource consuming both repos

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/spring/ElasticsearchRepositoryConfiguration.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/spring/ElasticsearchRepositoryConfiguration.java
@@ -18,12 +18,6 @@ package io.gravitee.repository.elasticsearch.spring;
 import static java.lang.String.format;
 
 import io.gravitee.elasticsearch.client.Client;
-import io.gravitee.elasticsearch.client.http.ClientSslConfiguration;
-import io.gravitee.elasticsearch.client.http.HttpClient;
-import io.gravitee.elasticsearch.client.http.HttpClientConfiguration;
-import io.gravitee.elasticsearch.client.http.HttpClientJksSslConfiguration;
-import io.gravitee.elasticsearch.client.http.HttpClientPemSslConfiguration;
-import io.gravitee.elasticsearch.client.http.HttpClientPfxSslConfiguration;
 import io.gravitee.elasticsearch.exception.ElasticsearchException;
 import io.gravitee.elasticsearch.index.ILMIndexNameGenerator;
 import io.gravitee.elasticsearch.index.IndexNameGenerator;
@@ -31,7 +25,9 @@ import io.gravitee.elasticsearch.index.MultiTypeIndexNameGenerator;
 import io.gravitee.elasticsearch.index.PerTypeIndexNameGenerator;
 import io.gravitee.elasticsearch.templating.freemarker.FreeMarkerComponent;
 import io.gravitee.elasticsearch.version.ElasticsearchInfo;
+import io.gravitee.platform.repository.api.Scope;
 import io.gravitee.repository.elasticsearch.analytics.spring.AnalyticsConfiguration;
+import io.gravitee.repository.elasticsearch.client.ElasticsearchHttpClientFactory;
 import io.gravitee.repository.elasticsearch.configuration.RepositoryConfiguration;
 import io.gravitee.repository.elasticsearch.healthcheck.spring.HealthCheckConfiguration;
 import io.gravitee.repository.elasticsearch.log.spring.LogConfiguration;
@@ -45,6 +41,7 @@ import lombok.CustomLog;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
+import org.springframework.core.env.Environment;
 
 /**
  * @author David BRASSELY (david.brassely at graviteesource.com)
@@ -83,46 +80,11 @@ public class ElasticsearchRepositoryConfiguration {
     }
 
     @Bean
-    public Client client(RepositoryConfiguration repositoryConfiguration) {
-        HttpClientConfiguration clientConfiguration = new HttpClientConfiguration();
-        clientConfiguration.setEndpoints(repositoryConfiguration.getEndpoints());
-        clientConfiguration.setUsername(repositoryConfiguration.getUsername());
-        clientConfiguration.setPassword(repositoryConfiguration.getPassword());
-        clientConfiguration.setRequestTimeout(repositoryConfiguration.getRequestTimeout());
-
-        clientConfiguration.setProxyType(repositoryConfiguration.getProxyType());
-        clientConfiguration.setProxyHttpHost(repositoryConfiguration.getProxyHttpHost());
-        clientConfiguration.setProxyHttpPort(repositoryConfiguration.getProxyHttpPort());
-        clientConfiguration.setProxyHttpUsername(repositoryConfiguration.getProxyHttpUsername());
-        clientConfiguration.setProxyHttpPassword(repositoryConfiguration.getProxyHttpPassword());
-        clientConfiguration.setProxyHttpsHost(repositoryConfiguration.getProxyHttpsHost());
-        clientConfiguration.setProxyHttpsPort(repositoryConfiguration.getProxyHttpsPort());
-        clientConfiguration.setProxyHttpsUsername(repositoryConfiguration.getProxyHttpsUsername());
-        clientConfiguration.setProxyHttpsPassword(repositoryConfiguration.getProxyHttpsPassword());
-        clientConfiguration.setProxyConfigured(repositoryConfiguration.isProxyConfigured());
-
-        if (repositoryConfiguration.getSslKeystoreType() != null) {
-            if (repositoryConfiguration.getSslKeystoreType().equalsIgnoreCase(ClientSslConfiguration.JKS_KEYSTORE_TYPE)) {
-                clientConfiguration.setSslConfig(
-                    new HttpClientJksSslConfiguration(
-                        repositoryConfiguration.getSslKeystore(),
-                        repositoryConfiguration.getSslKeystorePassword()
-                    )
-                );
-            } else if (repositoryConfiguration.getSslKeystoreType().equalsIgnoreCase(ClientSslConfiguration.PFX_KEYSTORE_TYPE)) {
-                clientConfiguration.setSslConfig(
-                    new HttpClientPfxSslConfiguration(
-                        repositoryConfiguration.getSslKeystore(),
-                        repositoryConfiguration.getSslKeystorePassword()
-                    )
-                );
-            } else if (repositoryConfiguration.getSslKeystoreType().equalsIgnoreCase(ClientSslConfiguration.PEM_KEYSTORE_TYPE)) {
-                clientConfiguration.setSslConfig(
-                    new HttpClientPemSslConfiguration(repositoryConfiguration.getSslPemCerts(), repositoryConfiguration.getSslPemKeys())
-                );
-            }
-        }
-        return new HttpClient(clientConfiguration);
+    public Client client(Environment environment) {
+        // RepositoryConfiguration still owns analytics' index-naming / ILM / retention concerns and is
+        // injected into other @Beans below; the client wiring is now scope-prefixed (analytics.elasticsearch.*)
+        // through the shared factory, mirroring how the OTEL_TRACES scope reads otel-traces.elasticsearch.*.
+        return new ElasticsearchHttpClientFactory(environment, Scope.ANALYTICS.getName()).createHttpClient();
     }
 
     @Bean
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
@@ -1,0 +1,567 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.elasticsearch.tracing;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.gravitee.elasticsearch.client.Client;
+import io.gravitee.elasticsearch.model.Aggregation;
+import io.gravitee.elasticsearch.model.SearchHit;
+import io.gravitee.elasticsearch.model.SearchResponse;
+import io.gravitee.elasticsearch.utils.IndexNameUtils;
+import io.gravitee.repository.common.query.QueryContext;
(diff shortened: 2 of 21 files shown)


gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/spring/ElasticsearchRepositoryConfiguration.java
```diff
@@ -84,6 +84,16 @@ public FreeMarkerComponent freeMarkerComponent() {
...
+        return buildHttpClient(repositoryConfiguration);
+    }
+
+    /**
+     * Builds an Elasticsearch {@link HttpClient} from a {@link RepositoryConfiguration} — endpoint, auth, proxy
+     * and SSL settings are all derived from the same property bag. Exposed as a {@code public static} helper so
+     * the OTEL_TRACES scope (which uses a separate Spring configuration class to avoid duplicate bean
+     * registration across scopes) can build its own client without copying ~30 lines of wiring.
+     */
+    public static HttpClient buildHttpClient(RepositoryConfiguration repositoryConfiguration) {
```
SamHardi: Making `buildHttpClient` `public static` on the analytics `@Configuration` class gives the tracing scope a compile-time dependency on `ElasticsearchRepositoryConfiguration` — an SRP violation that will bite if the two scopes ever diverge (different ES client versions, different SSL logic, etc.).

A dedicated utility class would be cleaner:

```java
// e.g. ElasticsearchHttpClientFactory.java
public final class ElasticsearchHttpClientFactory {
    private ElasticsearchHttpClientFactory() {}
    public static HttpClient build(RepositoryConfiguration cfg) { ... }
}
```

Both `ElasticsearchRepositoryConfiguration` and `TracingConfiguration` delegate to it, and neither scope owns the other.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
```diff
@@ -0,0 +1,560 @@
...
+    static String buildSearchTracesBody(TraceSearchCriteria criteria) {
+        ObjectNode root = JSON.objectNode();
+        root.put("size", 0);
+        root.set("query", buildQuery(criteria));
+
+        ObjectNode aggs = JSON.objectNode();
+        ObjectNode tracesAgg = JSON.objectNode();
+        ObjectNode termsAgg = JSON.objectNode();
+        termsAgg.put("field", "trace_id");
+        termsAgg.put("size", criteria.limit());
```
SamHardi: `criteria.limit()` flows straight into the ES `terms` aggregation `size` with no upper-bound guard. ES `search.max_buckets` defaults to 10 000 and a very large value can spike coordinator memory or get rejected with a `circuit_breaking_exception`. A defensive cap:

```java
private static final int MAX_SEARCH_RESULTS = 200;
// ...
termsAgg.put("size", Math.min(criteria.limit(), MAX_SEARCH_RESULTS));
```

(The constant value is up for discussion — the point is that the caller can't be trusted to self-limit.)

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
```diff
@@ -0,0 +1,560 @@
...
+        return termFilter;
+    }
+
+    /**
+     * Tag keys under {@code service.*} or {@code telemetry.*} land on the OTel resource (resource attributes are how
+     * service identity is encoded), everything else is a span attribute. Keeping this rule in one place lets callers
+     * use a flat {@code Map<String, String>} without having to know the OTel encoding.
+     */
+    private static ObjectNode buildTagTermFilter(String key, String value) {
+        String fieldName = isResourceAttribute(key) ? "resource.attributes." + key : "attributes." + key;
```
SamHardi: The `service.*` / `telemetry.*` prefix heuristic in `buildTagTermFilter` silently redirects those keys from `attributes.*` to `resource.attributes.*`. But the API already provides `TraceSearchCriteria.resourceAttributeFilters()` for explicit resource-level filtering — callers who use that know exactly what they're targeting.

The risk: a caller who puts `service.name` in `attributeFilters()` expecting a span-attribute match will silently get zero results, because the implementation sends the filter to `resource.attributes.service.name` instead. This could be very confusing to debug.

Question: is `attributeFilters()` intended as a "smart flat map" (heuristic routing) or a strict span-attribute filter? If the former, the Javadoc on `TraceSearchCriteria.attributeFilters()` should document the routing rule explicitly. If the latter, removing the heuristic and always targeting `attributes.*` is safer — callers who need resource-level matching already have `resourceAttributeFilters()`.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
```diff
@@ -0,0 +1,560 @@
...
+ * <p>
+ * Tenancy: the configured index template (default {@code traces-apim.otel-{orgId}}) is resolved against the caller's
+ * {@link QueryContext} before each search — operators are expected to point the OTel collector at the matching
+ * per-tenant index/data stream.
+ *
+ * @author GraviteeSource Team
+ */
+public class ElasticsearchTracingRepository implements TracingRepository {
+
+    private static final int MAX_SPANS_PER_TRACE = 1000;
```
SamHardi, line 60: `MAX_SPANS_PER_TRACE = 1000` is hardcoded with no config override. For APIs with fan-out patterns (gateway handling many concurrent streams), 1000 spans can be hit in production and silently truncate the trace. Worth exposing as a config property:

```yaml
otel-traces.elasticsearch.max-spans-per-trace: 1000
```

Same applies to a `MAX_SEARCH_RESULTS` cap if you add one on `criteria.limit()` above.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/configuration/TracingInfraConfiguration.java
```diff
@@ -0,0 +1,167 @@
...
+        cfg.setProxyHttpsHost(proxyHttpsHost);
+        cfg.setProxyHttpsPort(proxyHttpsPort);
+        cfg.setProxyHttpsUsername(proxyHttpsUsername);
+        cfg.setProxyHttpsPassword(proxyHttpsPassword);
+        return cfg;
+    }
+
+    private List<Endpoint> initializeEndpoints() {
+        List<Endpoint> endpoints = new ArrayList<>();
+        if (environment == null) {
```
SamHardi: `environment == null` is checked defensively here (and at line 153) because tests instantiate `TracingInfraConfiguration` directly with `new` and set `indexTemplate` via the public setter. This is a sign that the class is mixing two roles: a Spring-managed property bag (with `@Value` + `@Autowired`) and a plain data object for tests.

A cleaner split: make `TracingInfraConfiguration` a plain constructor-injectable POJO (or a `record`) for the values that tests care about, and move the `@Value`/`@Autowired` wiring into `TracingConfiguration` where it belongs. The null-guard defensive code disappears with it.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
```diff
@@ -0,0 +1,560 @@
...
+                return Instant.ofEpochMilli(Long.parseLong(text));
+            }
+            long epochMillis = Long.parseLong(text.substring(0, dot));
+            String fractionalMs = text.substring(dot + 1);
+            // Sub-millisecond digits: pad / truncate to 6 chars and interpret as the nano-suffix below the ms.
+            // "675208" → 675208 ns past the millisecond boundary; "67" → 670000 ns; "67520823" → 675208 ns.
+            String padded = (fractionalMs + "000000").substring(0, 6);
+            long extraNanos = Long.parseLong(padded);
+            return Instant.ofEpochMilli(epochMillis).plusNanos(extraNanos);
+        } catch (NumberFormatException ignored) {
```
SamHardi: The `NumberFormatException` is silently swallowed — if the `@timestamp` field arrives in an unexpected format, `parseInstant` returns `null` and the span's `startTime` is silently null. In production this would cause spans to sort incorrectly or the trace root-detection fallback to kick in without any signal.

At minimum, a `log.warn("Unexpected @timestamp format: {}", text)` before returning `null` would make this much easier to diagnose.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/spring/ElasticsearchRepositoryConfiguration.java
remibaptistegio: This part has been reworked to match the pattern used in Redis repository. This way it's cleaner and avoid coupling.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
remibaptistegio: Yes good catch, it was before we added the resourceAttributeFilters. Removed this part so now it's cleaner.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
remibaptistegio: Not sure about this one. It seems that scenarios with more than 1000 spans are not very realistic (not even sure how we could display it in the UI 🤔 ). I'll keep a note on this but leave it like this for now.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/configuration/TracingInfraConfiguration.java
remibaptistegio: This part has been rewritten.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
remibaptistegio: log added 👍

SamHardi (approved):

PR #17000, opened 2026-05-21
fix(console): client id placeholder and example in DCR config

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-14084

## Description
Updated mat-hint

Before:
image

After:
image


pragsty:
needs to be raised for 4.10, hence closing

PR #17002, opened 2026-05-21
feat(definition): add EdgeApi definition model and ApiType.EDGE

ytvnr:
## Issue

https://gravitee.atlassian.net/browse/GMA-331

## Description

Adds the API definition model for the Edge daemon configuration, following the same pattern as `NativeApi` (dedicated `AbstractApi` subclass with its own domain objects).

### Why a new `AbstractApi` subclass?

The Edge daemon configuration (proxy routing, DNS interception, shadow AI monitoring) doesn't fit the existing `Api` model which is built around listeners, endpoint groups, and plans. Like `NativeApi` was introduced for Kafka-native APIs with its own domain objects (`NativeListener`, `NativeEndpointGroup`, etc.), `EdgeApi` carries Edge-specific domain objects without inheriting irrelevant HTTP/messaging concepts.

### What's included

- **`ApiType.EDGE`** — new API type, registered in `AbstractApi`'s `@JsonSubTypes` for polymorphic Jackson deserialization.
- **`EdgeApi`** — extends `AbstractApi`. Carries `EdgeProxyDefinition` (proxy routing config) and `EdgeShadowAiDefinition` (shadow AI monitoring config). Returns empty lists for `getPlugins()` / `getListeners()` since Edge doesn't use those concepts.
- **Rich domain records** instead of primitives — `DnsDomain(name)`, `ShadowAiDomain(name)`, `EdgeRoute(pathPrefix, target)` — so the model can evolve without breaking serialized data.
- **`Duration` for `reportInterval`** instead of `int` seconds — ISO-8601 standard (`PT2M`), self-documenting.

### DIFF
--- a/gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/v4/AbstractApi.java
+++ b/gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/v4/AbstractApi.java
@@ -15,6 +15,7 @@
  */
 package io.gravitee.definition.model.v4;
 
+import static io.gravitee.definition.model.v4.AbstractApi.EDGE_LABEL;
 import static io.gravitee.definition.model.v4.AbstractApi.MESSAGE_LABEL;
 import static io.gravitee.definition.model.v4.AbstractApi.NATIVE_LABEL;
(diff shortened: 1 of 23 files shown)


michel-barret (approved):
LGTM

PR #17028, opened 2026-05-22
fix(distributed-sync): ensure initialSync is marked true on primary n…

mukul-tyagi08:
## Issue

https://gravitee.atlassian.net/browse/APIM-14159

### Why
The previous `if (nextFromTime == -1)` guard left initialSync stuck at false for the primary node on every restart, which kept the `sync-process` probe unhealthy forever and made Kubernetes startup probes fail.

### What

- Mark the initial sync as done after any successful synchronization, including the very first one on a primary node that restored its position from the distributed sync state (where nextFromTime is already != -1 at this point).
- Renames the `initialSync` flag to `syncReadyForProbe` to more accurately reflect its true purpose: signaling to Kubernetes startup/health probes that the node has successfully completed at least one synchronization cycle and is healthy.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/DefaultSyncManager.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/DefaultSyncManager.java
@@ -206,9 +206,7 @@ public class DefaultSyncManager extends AbstractService<SyncManager> implements 
                     .doOnComplete(() -> {
                         lastSyncOnError.set(false);
                         lastSyncErrorMessage.set(null);
-                        if (nextFromTime == -1) {
-                            initialSync.set(true);
-                        }
+                        initialSync.set(true);
                         nextFromTime = nextToTime;
 
                         log.debug(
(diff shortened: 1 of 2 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/DefaultSyncManager.java
```diff
@@ -206,9 +206,7 @@ private Completable synchronize() {
                     .doOnComplete(() -> {
                         lastSyncOnError.set(false);
                         lastSyncErrorMessage.set(null);
-                        if (nextFromTime == -1) {
-                            initialSync.set(true);
-                        }
+                        initialSync.set(true);
```
mukul-tyagi08: Mark the initial sync as done after any successful synchronization, including the very first one on a primary node that restored its position from the distributed sync state (where nextFromTime is already != -1 at this point).
The previous `if (nextFromTime == -1)` guard left initialSync stuck at false for the primary node on every restart, which kept the `sync-process` probe unhealthy forever and made Kubernetes startup probes fail.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/test/java/io/gravitee/gateway/services/sync/process/repository/DefaultSyncManagerTest.java
```diff
@@ -82,6 +85,26 @@ void should_synchronize_at_startup() throws Exception {
...
+        when(distributedSyncService.storeState(anyLong(), anyLong())).thenReturn(Completable.complete());
+
+        RepositorySynchronizer synchronizer = spy(new FakeSynchronizer(Completable.complete(), 1));
+        synchronizers.add(synchronizer);
+
+        cut = new DefaultSyncManager(router, node, synchronizers, null, distributedSyncService, 5, TimeUnit.SECONDS, 1);
+        cut.start();
+
+        verify(synchronizer).synchronize(eq(1_000L), any(), anySet());
+        assertThat(cut.syncDone()).isTrue();
```
mukul-tyagi08, line 107: The restored from-time is the one persisted in Redis, not -1.

phiz71 (approved):

PR #17039, opened 2026-05-22
feat: add gravitee.request.id and gravitee.transaction.id as OTel span attributes

araffass:
## Issue

https://gravitee.atlassian.net/browse/APIM-14173

## Description

Adds **_gravitee.request.id_** and **_gravitee.transaction.id_** as attributes on all OTel spans.

Sub-spans are enriched via AbstractTracingHook.spanAttributes(). 
The root span is enriched by a new **_TracingPreProcessor_** that runs right after TransactionPreProcessor, since the root span is created before the **_pre-processor_** chain runs and **_transactionId_** is null at that point.

## Additional context

https://github.com/user-attachments/assets/162d8056-e74b-4480-b07e-d408d62a6a94


mukul-tyagi08 (approved):
LGTM

PR #17046, opened 2026-05-22
feat(authz-ui): scaffold gamma-module-authz with module shell

podlesrafal:
Adds the `gravitee-gamma-module-authz` Maven plugin module and wires its UI bundle into the plugin zip following the AIM/APIM pattern. Result is a federated React module that mounts under `gamma-console` and renders a sidebar + 4 placeholder pages (MCPs / AI Models / APIs / Custom Policies); real content lands in follow-up PRs.

Maven / build
- pom.xml + assembly descriptor; exec-maven-plugin runs `yarn nx build` in `generate-resources`, the assembly fileSets bundle `target/classes/ui` into the plugin zip alongside the main JAR
- jakarta.ws.rs-api (provided) so AuthzResource compiles against the rest-api distribution
- license-check + license template + Taskfile + module-level prettier / eslint / tsconfig / vitest / .nvmrc / .sdkmanrc

Frontend tooling
- rspack + Module Federation config (`./App` exposed; react / react-dom / react-router-dom / @gravitee/graphene-core as singletons)
- vitest config + render-with-providers harness
- package.json scripts (serve/build/test/lint)

Module shell
- federation.tsx entry + EnvironmentContext + resolveEnvironmentId fallback chain (host prop → ?env URL → 'DEFAULT' with one-time WARN)
- AppRoutes + ModuleLayout (sidebar nav + breadcrumbs via buildLinearBreadcrumbs); placeholder pages for MCPs / AI Models / APIs / Custom Policies
- LocalDevRoot + LocalDevShell — standalone dev runtime when serving :3003 directly (not federated)
- ModuleErrorBoundary catches loader errors at the module boundary
- navigation / routes / environment config + route helpers

Root deps
- bump @gravitee/graphene-core to 2.12.2 across workspace (root + gamma-module-apim) so authz can use the latest Sheet/DataTable/ Toaster primitives in subsequent PRs

## Issue

## Description


gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/app/AppRoutes.tsx
```diff
@@ -0,0 +1,81 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { SidebarNavigation, Toaster, useLayoutConfig } from '@gravitee/graphene-core';
+import { useCallback, useMemo } from 'react';
+import { Navigate, Outlet, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
+import { NAV_GROUPS } from '../config/navigation';
+import { ROUTES, navigateToNavKey, resolveModulePath } from '../config/routes';
+import { buildLinearBreadcrumbs } from '../lib/buildLinearBreadcrumbs';
+import { ModuleErrorBoundary } from './ModuleErrorBoundary';
```
jourdiw: **[Score: 95/100]** 🚨 `ModuleErrorBoundary` is imported from `'./ModuleErrorBoundary'`, but this file does not exist in the diff or anywhere in the authz module. This will break the build.

Neither the APIM nor the platform sibling modules define a local `ModuleErrorBoundary` — if this is a shared component, it should come from `@gravitee/gamma-modules-sdk` or `@gravitee/graphene-core`. If it's module-specific, it needs to be included in this PR.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/app/AppRoutes.tsx
```diff
@@ -0,0 +1,81 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { SidebarNavigation, Toaster, useLayoutConfig } from '@gravitee/graphene-core';
+import { useCallback, useMemo } from 'react';
+import { Navigate, Outlet, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
+import { NAV_GROUPS } from '../config/navigation';
+import { ROUTES, navigateToNavKey, resolveModulePath } from '../config/routes';
+import { buildLinearBreadcrumbs } from '../lib/buildLinearBreadcrumbs';
```
jourdiw: **[Score: 92/100]** 🚨 `buildLinearBreadcrumbs` is imported from `'../lib/buildLinearBreadcrumbs'`, but this local file does not exist in the diff and no `lib/` directory is introduced by this PR.

Both sibling modules (`gamma-module-apim` and `gamma-module-platform`) import `buildLinearBreadcrumbs` directly from `'@gravitee/graphene-core'`:

```ts
import { buildLinearBreadcrumbs, SidebarNavigation, useLayoutConfig } from '@gravitee/graphene-core';
```

This should use the same import path to avoid a build failure and stay consistent.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/app/AppRoutes.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { CustomPoliciesPage } from './features/policy-management/pages/CustomPoliciesPage';
+import { LlmsPage } from './features/policy-management/pages/LlmsPage';
+import { McpsPage } from './features/policy-management/pages/McpsPage';
+
+function ModuleLayout() {
+    const location = useLocation();
+    const navigate = useNavigate();
+
+    const { modulePrefix, activeNavKey } = useMemo(() => resolveModulePath(location.pathname), [location.pathname]);
+
```
jourdiw: **[Score: 80/100]** ⚠️ Both the APIM and platform sibling modules use `useModuleRouting()` from `@gravitee/gamma-modules-sdk/routing` with a typed `ModuleRouteConfig`. This module manually re-implements the same host-path-resolution logic via `resolveModulePath()` and `navigateToNavKey()` in `config/routes.ts`.

Suggested approach — in `config/routes.ts`, export a `ModuleRouteConfig<RouteKey>`:

```ts
import type { ModuleRouteConfig } from '@gravitee/gamma-modules-sdk/routing';

export const AUTHZ_ROUTE_CONFIG: ModuleRouteConfig<RouteKey> = {
    routeKeys: ROUTE_KEYS,
    routes: ROUTES,
    defaultRouteKey: 'mcps',
} as const;
```

Then in `ModuleLayout`:

```ts
const { activeNavKey, navigateToKey, rootPath } = useModuleRouting(AUTHZ_ROUTE_CONFIG);
```

This avoids duplicating the environment-prefix stripping logic and stays consistent with the SDK contract that all modules share.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/config/navigation.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import type { NavGroup } from '@gravitee/graphene-core';
+import { Brain, Globe, ShieldCheck, SlidersHorizontal } from 'lucide-react';
+import { ROUTES } from './routes';
```
jourdiw, line 18: **[Score: 72/100]** ⚠️ Icons are imported from `lucide-react` directly, but both sibling modules (`gamma-module-apim` and `gamma-module-platform`) import sidebar icons from the design system subpath `@gravitee/graphene-core/icons`:

```ts
// APIM navigation.ts
import { ArchiveIcon, HomeIcon, RadioIcon } from '@gravitee/graphene-core/icons';

// Platform navigation.ts
import { AppWindowIcon, LayoutDashboardIcon } from '@gravitee/graphene-core/icons';
```

Modules should use the design system's icon re-exports to stay consistent with the shared Module Federation singleton configuration. Importing `lucide-react` directly could lead to duplicate icon bundles if the MF config doesn't share that package.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/config/environment.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * @deprecated Story G1: replaced by `useEnvironment()` hook from
+ * `app/lib/env/EnvironmentContext`. The constant survives as the resolution
+ * fallback inside `resolveEnvironmentId` only — production code paths now
+ * read the env id from the host-injected provider, not from this literal.
+ *
+ * Eslint guard (`no-restricted-syntax`) blocks any new `'DEFAULT'` literal in
+ * `app/**` so this fallback can't be re-introduced by accident.
+ */
+export const DEFAULT_ENVIRONMENT_ID = 'DEFAULT';
```
jourdiw: **[Score: 55/100]** 💡 This file exports `DEFAULT_ENVIRONMENT_ID` but its own JSDoc marks it as `@deprecated`, advising to use `useEnvironment()` from `@gravitee/gamma-modules-sdk` instead.

Neither the APIM nor platform sibling modules include a local `config/environment.ts` — they rely entirely on the SDK hook. A brand-new module shouldn't ship code that's already deprecated on day one. If the fallback constant is still needed during initial bootstrap, consider placing it in a shared package rather than duplicating it per module.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/app/AppRoutes.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <Routes>
+                <Route element={<ModuleLayout />}>
+                    <Route index element={<Navigate to="mcps" replace />} />
+                    <Route path="dashboard" element={<Navigate to="../mcps" replace />} />
+                    <Route path="mcps" element={<McpsPage />} />
+                    <Route path="llms" element={<LlmsPage />} />
+                    <Route path="apis" element={<ApisPage />} />
+                    <Route path="custom-policies" element={<CustomPoliciesPage />} />
+                </Route>
+            </Routes>
```
jourdiw: **[Score: 45/100]** 💡 `<Toaster />` is rendered inside `AppRoutes`. In the sibling modules, the Toaster is provided by the host shell (`ShellLayout` / `LocalDevShell`), not by the module itself.

When this module is federated into the Gamma console host, this could produce a duplicate Toaster overlay. Consider removing it from here and relying on the host, or gating it behind a `localDev` check so it only renders in standalone mode.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/pages/McpsPage.tsx
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * Placeholder MCPs page. The full implementation arrives in a follow-up
+ * PR — this stub keeps the module shell compilable and lets the host
+ * register the route so navigation already works end-to-end.
+ */
+export function McpsPage() {
+    return (
```
jourdiw, line 22: **[Score: 25/100]** 🎨 All four placeholder pages (`McpsPage`, `LlmsPage`, `ApisPage`, `CustomPoliciesPage`) use raw HTML elements (`<div>`, `<h1>`, `<p>`) with Tailwind utility classes. While acceptable for stubs, sibling modules tend to use Graphene components (`Card`, `Alert`, etc.) even for placeholder content.

Minor nit — these will be replaced in follow-up PRs, so no action needed now unless you want visual consistency in the interim.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/app/lib/env/resolveEnvironmentId.ts
jourdiw, line 1: The environment Id can come through the parent with the gamma-modules-sdk. You don't need to resolve it yourself.

gravitee-gamma/gravitee-gamma-module-authz/AGENTS.md
```diff
@@ -0,0 +1,181 @@
+# Coding Guidelines for gravitee-gamma-module-apim
```
jourdiw: **[Score: 75/100]** ⚠️ Copy-paste oversight: the title reads `# Coding Guidelines for gravitee-gamma-module-apim` but this file lives in `gravitee-gamma-module-authz`. Should be `# Coding Guidelines for gravitee-gamma-module-authz`.

This will confuse AI agents and contributors referencing this file for authz-specific conventions.

gravitee-gamma/gravitee-gamma-module-authz/package.json
```diff
@@ -0,0 +1,17 @@
...
+        "@fontsource/dm-sans": "5.2.8",
+        "@gravitee/gamma-modules-sdk": "1.0.0-alpha.2",
+        "@gravitee/graphene-core": "2.12.2",
+        "@monaco-editor/react": "4.7.0",
+        "@xyflow/react": "12.10.1",
+        "lucide-react": "0.487.0",
+        "monaco-editor": "0.52.2",
+        "react": "19.2.4",
+        "react-dom": "19.2.4",
+        "react-router-dom": "7.13.1"
```
jourdiw, line 10: **[Score: 85/100]** ⚠️ `@tanstack/react-query` is imported in `AppRoutes.tsx` (`QueryClient`, `QueryClientProvider`) but is **not declared** in this module's `package.json`.

The sibling `gravitee-gamma-module-apim` correctly lists it:
```json
"@tanstack/react-query": "^5.80.1"
```

This works today via Yarn workspace hoisting, but:
- The dependency is invisible to Nx's dependency graph (affecting build order)
- It will break if the module is ever published or built in isolation
- It violates the explicit-dependency principle for federated remotes

Add it to the `dependencies` block.

gravitee-gamma/gravitee-gamma-module-authz/package.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "gravitee-gamma-module-authz",
+    "private": true,
+    "dependencies": {
+        "@dagrejs/dagre": "1.1.4",
```
jourdiw: **[Score: 55/100]** 💡 Several heavy dependencies are declared but **unused anywhere in this scaffold PR**:

- `@dagrejs/dagre` (graph layout)
- `@fontsource/dm-sans` (font, never imported)
- `@monaco-editor/react` + `monaco-editor` (code editor)
- `@xyflow/react` (flow diagrams)
- `lucide-react` (icons — but navigation.ts correctly imports from `@gravitee/graphene-core/icons`)

The PR title is "scaffold gamma-module-authz with module shell" — these deps likely belong to follow-up feature PRs that actually use them. Shipping unused dependencies in the scaffold:
- Inflates the initial bundle (even tree-shaken, MF needs to resolve them)
- Makes it harder to review which deps are actually required for the shell vs. features
- Adds noise to the root `yarn.lock` diff

Consider deferring these to the PRs that introduce code importing them.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/pages/McpsPage.tsx
jourdiw, line 22: **[Score: 60/100]** 💡 Architectural divergence in folder structure. In the sibling `gravitee-gamma-module-apim`, feature domains live at:
```
src/main/ui/features/<domain>/pages/...
```

But here they're nested under `app/`:
```
src/main/ui/app/features/policy-management/pages/...
```

The APIM module keeps `app/` thin (just `AppRoutes`, `LocalDevRoot`, `LocalDevShell`) and puts features alongside it at the `src/main/ui/` level. Same for the `lib/env/` directory — in APIM, cross-cutting utilities live in `src/main/ui/shared/`.

Consider aligning the folder layout now while there's minimal code to move, so follow-up PRs build on the correct structure.

gravitee-gamma/gravitee-gamma-module-authz/rspack.config.ts
```diff
@@ -0,0 +1,49 @@
...
+import { NxAppRspackPlugin } from '@nx/rspack/app-plugin.js';
+import { NxReactRspackPlugin } from '@nx/rspack/react-plugin.js';
+import config from './module-federation.config';
+
+export default {
+    output: {
+        path: join(__dirname, './target/classes/ui'),
+        publicPath: 'auto',
+        uniqueName: 'authz',
+    },
```
jourdiw, line 27: **[Score: 30/100]** 🎨 Minor naming inconsistency: `uniqueName` is set to `'authz'` (short) while the APIM sibling uses its full module name `'gravitee-gamma-module-apim'`.

While short names work, the `uniqueName` is a Module Federation global identifier — using the full qualified name (`'gravitee-gamma-module-authz'`) avoids potential collisions if other MF hosts or remotes ever use a generic `'authz'` identifier. Also aligns with the pattern set by the sibling.

jourdiw (approved):
LGTM.
If it runs locally, then 🚀

PR #17047, opened 2026-05-22
feat(authz-ui): REST client + entity / policy / schema hooks

podlesrafal:
Adds the backend integration layer for the authz UI: the REST client talks to the canonical `/gamma/authz` endpoints with bootstrap-driven URL resolution and CSRF handling; React hooks expose entities, schema, and policies to the rest of the module. Nothing on screen yet — the list pages stay placeholders until the follow-up PR wires them in.

API
- authz-api-client: bootstrap + constants.json fetch, in-flight promise cache, CSRF header / cookie roundtrip, ApiError surface (status + validation payload)
- authz-api.types: PolicyType / Status / EntityRequest / PolicyRequest / ValidationErrorResponse and friends
- authz-api.service: typed wrappers over the canonical core endpoints (entities CRUD, schema GET, policies CRUD + deploy/disable transitions) and a paged-response normaliser that tolerates the old bare-array shape during rolling restarts

Entity adapter
- entity-adapter: parses canonical `Type::"slug"` and dotted `kind.slug` UIDs into structured { type, id } pairs and back to the canonical backend form; round-trip tested
- policy-entity-refs: scanner that pulls `Type::"id"` references out of GAPL policy text grouped by clause (principal / action / resource) — used by the Entities table later to answer "which policies reference this entity?"
- chip-option: small ChipOption interface for combobox options, shared between hooks and (later) the policy editor

Hooks (live data for downstream UI)
- useEntities: paged entity listing with reload semantics
- useEntityOptions: maps the entity collection to canonical `Type::"slug"` ChipOption ids for the policy editor; supports type filtering and surfaces an error when more than 200 rows would be needed
- useSchema: GAPL schema fetch
- usePolicies: list + create + update + delete + deploy/disable transitions
- entity-types: principal / resource enum constants

All hooks + adapter come with vitest specs.

## Issue

## Description

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/chip-option.ts
+++ b/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/chip-option.ts
@@ -1,0 +1,26 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * Option shape consumed by the chip combobox in the policy editor.
+ * Lives in `lib/` (not under `app/features/`) because hooks like
+ * `useEntityOptions` build options before any feature code is loaded.
+ */
+export interface ChipOption {
+    readonly id: string;
+    readonly label: string;
+    readonly group: string;
+    readonly description?: string;
+}
--- a/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useEntities.ts
+++ b/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useEntities.ts
@@ -1,0 +1,76 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { useCallback, useEffect, useState } from 'react';
+import { authzApiService, DEFAULT_PER_PAGE } from '../api/authz-api.service';
+import type { EntityResponse, PagedResponse } from '../api/authz-api.types';
+
+export interface UseEntitiesResult {
+    readonly data: PagedResponse<EntityResponse> | null;
+    readonly isLoading: boolean;
+    readonly error: string | undefined;
+    readonly page: number;
+    readonly perPage: number;
+    readonly setPage: (page: number) => void;
+    readonly setPerPage: (perPage: number) => void;
+    readonly reload: () => void;
+}
+
+export function useEntities(environmentId: string, initialPerPage = DEFAULT_PER_PAGE): UseEntitiesResult {
+    const [page, setPage] = useState(1);
+    // Resync perPage from prop on every change so callers can drive it as a controlled value.
+    const [perPage, setPerPage] = useState(initialPerPage);
+    useEffect(() => {
+        setPerPage(initialPerPage);
+    }, [initialPerPage]);
+    const [data, setData] = useState<PagedResponse<EntityResponse> | null>(null);
+    const [isLoading, setIsLoading] = useState(true);
+    const [error, setError] = useState<string | undefined>(undefined);
+    const [nonce, setNonce] = useState(0);
+
+    useEffect(() => {
+        let cancelled = false;
+        // Gate setState to break cascading renders when nothing actually
+        // changes (isLoading already true / error already undefined).
+        if (!isLoading) setIsLoading(true);
+        if (error !== undefined) setError(undefined);
+
+        authzApiService
+            .listEntities(environmentId, { page, perPage })
+            .then(res => {
+                if (cancelled) return;
+                setData(res);
+            })
+            .catch(e => {
+                if (cancelled) return;
+                setError(e instanceof Error ? e.message : 'Failed to load entities');
+                setData(null);
+            })
+            .finally(() => {
+                if (!cancelled) setIsLoading(false);
+            });
+
+        return () => {
+            cancelled = true;
+        };
+        // isLoading/error intentionally excluded — they are read for gating
+        // setState; including them would re-trigger the effect.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [environmentId, page, perPage, nonce]);
+
+    const reload = useCallback(() => setNonce(n => n + 1), []);
+
+    return { data, isLoading, error, page, perPage, setPage, setPerPage, reload };
+}
--- a/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useEntities.ts
+++ b/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useEntities.ts
@@ -13,9 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { useCallback, useEffect, useState } from 'react';
+import { keepPreviousData, useQuery, useQueryClient } from '@tanstack/react-query';
+import { useCallback, useState } from 'react';
 import { authzApiService, DEFAULT_PER_PAGE } from '../api/authz-api.service';
 import type { EntityResponse, PagedResponse } from '../api/authz-api.types';
+import { authzQueryKeys } from '../api/query-keys';
 
 export interface UseEntitiesResult {
     readonly data: PagedResponse<EntityResponse> | null;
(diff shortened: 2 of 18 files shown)


podlesrafal (commented):
test

podlesrafal (commented):
test

podlesrafal (commented):
test

podlesrafal (commented):
test

podlesrafal (commented):
test

podlesrafal (commented):
test

podlesrafal (commented):
test

podlesrafal (commented):
test

podlesrafal (commented):
x

podlesrafal (commented):
x

podlesrafal (commented):
x

podlesrafal (commented):
x

podlesrafal (commented):
x

podlesrafal (commented):
x

podlesrafal (commented):
## Review: authz UI REST client + hooks layer

The test coverage is thorough — stale-response, unmount-cancel, and perPage-change scenarios are all covered. The adapter logic (`authz-api.service.ts`) is solid. The **main blocker** is that all four hooks hand-roll a fetch state machine that `@tanstack/react-query` (already at `5.90.12` in the monorepo root `package.json`) solves declaratively. The sibling `gravitee-gamma-module-platform` already uses it — see `shared/hooks/useEnvironmentPermissions.ts`. Please align before wiring these hooks into UI pages.

**Findings on files not in this PR's diff** (inline comments require changed lines):
- `authz-api-client.ts` line 44: CSRF token stored in `localStorage` is readable by any XSS on the page. As an interim before the backend cookie endpoint lands, use a module-level closure variable (`let csrfToken: string | null = null`) — not persisted to disk and not accessible outside the module. Tag the TODO with a Jira ticket.
- `authz-api.service.ts` line 313: `null as unknown as string` for `updatedAt` is a type lie. Change `SchemaResponse.updatedAt` to `string | null` or use an empty-string fallback.
- `authz-api.service.ts` line 337: client-side `type` post-filter skews pagination `total` (acknowledged in comment). Track as a known UX degradation until the server endpoint accepts a `kind` filter.
- `app/features/policy-structure/entity-types.ts` (`CATEGORIES`): uses raw Tailwind palette classes (`text-blue-600`, `dark:text-teal-400`, etc.) instead of semantic graphene-core tokens — violates `graphene.md` design-system rules.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useEntities.ts
```diff
@@ -0,0 +1,76 @@
...
+        };
+        // isLoading/error intentionally excluded — they are read for gating
+        // setState; including them would re-trigger the effect.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [environmentId, page, perPage, nonce]);
+
+    const reload = useCallback(() => setNonce(n => n + 1), []);
+
+    return { data, isLoading, error, page, perPage, setPage, setPerPage, reload };
+}
```
podlesrafal, line 69: **[Score: 75/100]** ⚠️ **Architectural mismatch with sibling module pattern**: `gravitee-gamma-module-platform` uses `@tanstack/react-query` (already in the monorepo root `package.json` as `@tanstack/react-query@5.90.12`) for exactly this use case — see `shared/hooks/useEnvironmentPermissions.ts`. This hook hand-rolls ~70 lines that re-implement query caching, deduplication, and cancellation. The authz `package.json` does not declare `@tanstack/react-query` even though it is available in the workspace. Adding it and replacing this with `useQuery({ queryKey: ['entities', environmentId, page, perPage], queryFn: ... })` eliminates the cancelled-flag, the nonce reload trick, and the `isLoading`-gating guards — all solved problems in react-query.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/usePolicies.ts
```diff
@@ -0,0 +1,124 @@
...
+        async (id: string) => {
+            await authzApiService.deletePolicy(environmentId, id);
+            if (!mountedRef.current) return;
+            reload();
+        },
+        [environmentId, reload],
+    );
+
+    return { data, isLoading, error, page, perPage, setPage, setPerPage, create, update, remove, reload };
+}
```
podlesrafal, line 111: **[Score: 75/100]** ⚠️ **Same architectural concern as `useEntities`**: this hook duplicates the same hand-rolled fetch state machine (useState + useEffect + cancelled flag + nonce + mountedRef). With `@tanstack/react-query`, `usePolicies` reduces to `useQuery` for the list and `useMutation` for `create`/`update`/`remove`. The `mountedRef` guard in each mutation callback is an anti-pattern — mutations should use `onSuccess: () => queryClient.invalidateQueries(...)` instead of checking `mountedRef.current`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useSchema.ts
```diff
@@ -0,0 +1,65 @@
...
+            .finally(() => {
+                if (!cancelled) setIsLoading(false);
+            });
+
+        return () => {
+            cancelled = true;
+        };
+    }, [environmentId]);
+
+    return { schema, notFound, isLoading, error };
```
podlesrafal: **[Score: 75/100]** ⚠️ **Same architectural concern**: `useSchema` is a simple fetch-once-per-env hook. With react-query: `useQuery({ queryKey: ['schema', environmentId], queryFn: () => authzApiService.getSchema(environmentId), staleTime: Infinity })`. The 404 → `notFound` mapping moves into the queryFn throw or a `select` transform. The cancelled-flag boilerplate disappears entirely.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useEntityOptions.ts
```diff
@@ -0,0 +1,148 @@
...
+    }, [environmentId, typeFilterKey]);
+
+    const options = useMemo(() => {
+        if (!typeFilter || typeFilter.length === 0) return allOptions;
+        const allow = new Set(typeFilter);
+        return allOptions.filter(o => allow.has(o.group));
+    }, [allOptions, typeFilter, typeFilterKey]); // eslint-disable-line react-hooks/exhaustive-deps
+
+    return { options, isLoading, error };
+}
```
podlesrafal, line 87: **[Score: 75/100]** ⚠️ **Same architectural concern**: same hand-rolled state machine. Additionally, the `mountedRef` at diff position 98 is unused for cancellation — the `cancelled` flag in the effect already covers unmount safety. The ref is dead code and should be removed regardless of the react-query migration decision.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useEntityOptions.ts
```diff
@@ -0,0 +1,148 @@
...
+    // fresh array literal (`['User', 'Group']`) without retriggering the
+    // effect on every render.
+    const typeFilterKey = useMemo(() => (typeFilter ? JSON.stringify([...typeFilter].sort()) : ''), [typeFilter]);
+
+    const [allOptions, setAllOptions] = useState<readonly ChipOption[]>([]);
+    const [isLoading, setIsLoading] = useState(true);
+    const [error, setError] = useState<string | undefined>(undefined);
+
+    const mountedRef = useRef(true);
+    useEffect(() => {
```
podlesrafal: **[Score: 50/100]** 💡 **`mountedRef` is dead code**: the `cancelled` flag inside the `useEffect` already prevents setState after unmount. This `mountedRef` is never actually read to guard any setState call in this file. Remove it to reduce noise.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useEntityOptions.ts
```diff
@@ -0,0 +1,148 @@
...
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [environmentId, typeFilterKey]);
+
+    const options = useMemo(() => {
+        if (!typeFilter || typeFilter.length === 0) return allOptions;
+        const allow = new Set(typeFilter);
+        return allOptions.filter(o => allow.has(o.group));
+    }, [allOptions, typeFilter, typeFilterKey]); // eslint-disable-line react-hooks/exhaustive-deps
+
+    return { options, isLoading, error };
```
podlesrafal: **[Score: 50/100]** 💡 **Stale `typeFilterKey` in `useMemo` deps**: `typeFilterKey` is derived from `typeFilter` via `useMemo`, so including both in the `options` memo dep array (next line) means the memo recalculates on two signals that always fire together. Remove `typeFilterKey` from that dep array — `typeFilter` alone is sufficient. The `// eslint-disable-line` on line 146 masks the lint warning instead of fixing the underlying dep list.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useEntities.ts
```diff
@@ -0,0 +1,76 @@
...
+    const [error, setError] = useState<string | undefined>(undefined);
+    const [nonce, setNonce] = useState(0);
+
+    useEffect(() => {
+        let cancelled = false;
+        // Gate setState to break cascading renders when nothing actually
+        // changes (isLoading already true / error already undefined).
+        if (!isLoading) setIsLoading(true);
+        if (error !== undefined) setError(undefined);
+
```
podlesrafal, line 44: **[Score: 45/100]** 💡 **Conditional `setState` gating is fragile**: reading `isLoading` and `error` inside an effect to avoid cascading renders, then excluding them from the dep array via an eslint-disable comment, is a known footgun — if those values drift out of sync the guard silently misfires. React 19 batches synchronous `setState` calls inside effects, so unconditionally calling `setIsLoading(true)` and `setError(undefined)` at the top of the effect produces at most one extra re-render. Drop the conditional guards and the eslint-disable comment.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/usePolicies.ts
```diff
@@ -0,0 +1,124 @@
...
+    const [data, setData] = useState<PagedResponse<PolicyResponse> | null>(null);
+    const [isLoading, setIsLoading] = useState(true);
+    const [error, setError] = useState<string | undefined>(undefined);
+    const [nonce, setNonce] = useState(0);
+
+    useEffect(() => {
+        mountedRef.current = true;
+        return () => {
+            mountedRef.current = false;
+        };
```
podlesrafal: **[Score: 45/100]** 💡 **Same conditional `setState` gating pattern as `useEntities`**. Same recommendation: drop the guards and rely on React 19 batching. The `// eslint-disable-next-line` further down in this effect is a code smell that the dep array is intentionally wrong.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useEntities.ts
```diff
@@ -0,0 +1,76 @@
...
+    readonly reload: () => void;
+}
+
+export function useEntities(environmentId: string, initialPerPage = DEFAULT_PER_PAGE): UseEntitiesResult {
+    const [page, setPage] = useState(1);
+    // Resync perPage from prop on every change so callers can drive it as a controlled value.
+    const [perPage, setPerPage] = useState(initialPerPage);
+    useEffect(() => {
+        setPerPage(initialPerPage);
+    }, [initialPerPage]);
```
podlesrafal: **[Score: 30/100]** 🎨 **`initialPerPage` sync via a secondary `useEffect`**: this causes an extra render cycle (mount with stale `perPage`, then the effect fires to sync). If `initialPerPage` is truly initial-only, `useState(initialPerPage)` needs no sync effect. If the hook should support a fully controlled `perPage` driven from above (which the tests confirm), document that contract explicitly.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/usePolicies.ts
```diff
@@ -0,0 +1,124 @@
...
+    readonly create: (request: PolicyRequest) => Promise<PolicyResponse>;
+    readonly update: (id: string, request: PolicyRequest) => Promise<PolicyResponse>;
+    readonly remove: (id: string) => Promise<void>;
+    readonly reload: () => void;
+}
+
+export function usePolicies(environmentId: string, options: UsePoliciesOptions = {}): UsePoliciesResult {
+    const mountedRef = useRef(true);
+    const { type, status, initialPerPage = DEFAULT_PER_PAGE } = options;
+    const [page, setPage] = useState(1);
```
podlesrafal, line 44: **[Score: 30/100]** 🎨 **Same `initialPerPage` sync pattern as `useEntities`** — same recommendation applies.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/hooks/useEntities.ts
```diff
@@ -0,0 +1,76 @@
...
+            cancelled = true;
+        };
+        // isLoading/error intentionally excluded — they are read for gating
+        // setState; including them would re-trigger the effect.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [environmentId, page, perPage, nonce]);
+
+    const reload = useCallback(() => setNonce(n => n + 1), []);
+
+    return { data, isLoading, error, page, perPage, setPage, setPerPage, reload };
```
podlesrafal: **[Score: 25/100]** 🎨 **`nonce` counter as reload trigger obscures intent**: consider renaming the state to `reloadCount` or `fetchRevision`, and renaming the exported `reload` to `refetch` to match the react-query vocabulary the rest of the codebase borrows.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/lib/chip-option.ts
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+/**
+ * Option shape consumed by the chip combobox in the policy editor.
+ * Lives in `lib/` (not under `app/features/`) because hooks like
+ * `useEntityOptions` build options before any feature code is loaded.
+ */
+export interface ChipOption {
+    readonly id: string;
+    readonly label: string;
```
podlesrafal, line 23: **[Score: 20/100]** 🎨 **JSDoc on a self-describing interface**: per project conventions, JSDoc is discouraged unless the symbol is exported to external consumers and the name alone is ambiguous. `ChipOption` with fields `id`, `label`, `group`, `description` is self-explanatory. The multi-line block comment above restates what the field names already say — remove it.

PR #17048, opened 2026-05-22
feat(authz-ui): GAPL engine — schema + policy parser, serializer, Mon…

podlesrafal:
…aco syntax

Standalone library that turns GAPL text into structured statements and back again, plus a Monaco wrapper with GAPL syntax highlighting. The policy editor uses this in a follow-up PR to drive reactive visual ↔ code synchronisation; right now it ships unwired so it can be reviewed and fuzz-tested in isolation.

GAPL schema parser
- gapl-parser: tokenises a Gravitee policy DSL schema (entity types, attributes, action definitions) and produces an AST callers can walk to enumerate available actions / principal kinds / resource kinds
- fuzz tests + property tests across malformed / edge-case schemas

GAPL policy parser
- gapl-policy-parser: parses `permit ()` / `forbid ()` clauses with `principal == User::"alice"`, `action in [Action::"read", ...]`, optional `when {}` guards; emits a `PolicyStatement[]` matched to the serializer below — supports round-trip fidelity for the slice the visual editor can represent, and reports `null` when the source has features (advanced when-clauses etc) the visual side cannot model
- companion fuzz tests + spec coverage

GAPL serializer
- statement-to-gapl: emits canonical `Type::"slug"` UIDs from PolicyStatement records, with `permit ()` / `forbid ()` clause shapes that the parser above accepts; preserves the namespace casing the user typed so editing doesn't silently lowercase `Action` → `action` on save

Monaco syntax + wrapper
- components/gapl-language: Monarch tokens registered into Monaco for GAPL — keywords, UID literals, when-block conditions, comments
- components/MonacoEditor: thin wrapper around `@monaco-editor/react` that registers the GAPL language on mount and exposes the editor + monaco instances back to callers so they can drive `revealLineInCenter` / `setPosition` on hydration
- styles/monaco.css: layout polish so Monaco's automaticLayout fits the policy-editor's flex container without collapsing to ~20px

## Issue

## Description


sclarkamlab (commented):
Notes from a review pass (cursor + a few agents). Inline below. The PR description still has the `APIM-XXX` placeholder and template boilerplate, want to fill that in before merge.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/statement-to-gapl.ts
```diff
@@ -0,0 +1,113 @@
...
+
+const DEFAULT_ACTION_KIND = 'Action';
+const CANONICAL_UID_RE = /^[^:]+::".+"$/;
+
+function quote(value: string): string {
+    return `"${value.replace(/"/g, '\\"')}"`;
+}
+
+function emitToken(kind: string | undefined, id: string, label: string): string {
+    if (CANONICAL_UID_RE.test(id)) return id;
```
sclarkamlab, line 61: Round-trip quietly drops the surrogate `id`. `emitToken` rebuilds the UID from kind+label unless `id` is already canonical, and the parser refills `id` with `Type::"..."`, so any UI-side id is lost on parse then serialize. The round-trip tests use `shape()` which strips `id` so we don't catch it. Need to check if anything actually relies on a distinct id, if not we can leave it.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/shared/gapl-policy-parser.ts
```diff
@@ -0,0 +1,265 @@
...
+        return (body.raw ?? '').trim();
+    }
+}
+
+export function parseGaplToStatements(text: string): ParsedPolicy | null {
+    if (!text || !text.trim()) return { statements: [], warnings: [] };
+
+    const stripped = stripLineComments(text);
+    const tokens = tokenise(stripped);
+    if (tokens === null) return null;
```
sclarkamlab: Bare `null` can't tell valid GAPL the visual editor just can't model (Cedar `unless`) from actual garbage, and warnings get cleared on this path. The schema parser already returns diagnostics, would be good to match that here so the editor can tell the user what's wrong.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/MonacoEditor.tsx
```diff
@@ -0,0 +1,75 @@
...
+    options,
+    ariaLabel,
+    onMount,
+}: MonacoEditorProps) {
+    const handleMount: OnMount = (editor, monaco) => {
+        registerGaplLanguage(monaco);
+        onMount?.(editor, monaco);
+    };
+
+    const monacoTheme = theme === 'vs-dark' ? 'vs-dark' : 'vs';
```
sclarkamlab: `'system'` is in the prop type but falls through to `vs` (light), there's no `prefers-color-scheme` branch. Either wire that up or drop `'system'` from the union.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/shared/strip-line-comments.ts
```diff
@@ -0,0 +1,40 @@
...
+                        continue;
+                    }
+                    if (c === '"') inString = false;
+                    continue;
+                }
+                if (c === '"') {
+                    inString = true;
+                    continue;
+                }
+                if (c === '/' && line[i + 1] === '/') return line.slice(0, i);
```
sclarkamlab: Only strips `//`, not `/* */`. Do we allow block comments in GAPL? If so the parser will happily tokenise inside them.

podlesrafal:
Thanks for the pass @sclarkamlab — all four addressed in 4edc358c66.

1. **`'system'` theme** — dropped from the `MonacoEditorProps.theme` union. No callsite ever passed it and there was no `prefers-color-scheme` wiring behind it, so it was strictly dead API surface. Added three tests locking the light / vs-dark / default mapping in case anyone wires a system option later.

2. **Block comments** — replaced `stripLineComments` with `stripGaplComments`, a single-pass scanner that handles `//` and `/* */` while respecting string literals (so `Doc::"/* not a comment */"` stays intact) and preserving newlines so diagnostic line numbers stay aligned. New unit tests plus an integration test in the policy parser exercising block comments around and inside a statement.

3. **`id` round-trip** — agreed nothing currently relies on a distinct `id`, so I locked the invariant rather than trying to preserve it. GAPL has no slot for a surrogate id, so refs come back as canonical `Type::"label"` and `stmt.id` is regenerated as `stmt-N`. JSDoc on `PolicyStatement` / ref types now says so explicitly, and there are two tests asserting it so the visual builder in ui-5 / editor sheet in ui-6 know to keep any stable identity outside these records.

4. **Parser `null` vs diagnostics** — matched the schema parser shape. `parseGaplToStatements()` now always returns a `ParsedPolicy` with `diagnostics: string[]` (renamed from `warnings`), never null. The tokeniser surfaces `Unexpected character 'X' at position N` / `Unterminated 'when {' block` and the parser flags Cedar's `unless` with a targeted message (`'unless' clauses are not supported (Cedar feature outside the GAPL visual editor's subset)`), distinct from generic parse errors. Existing `.toBeNull()` assertions rewritten to inspect `diagnostics` length + content.

PR description with the `APIM-XXX` placeholder still needs the real ticket — will fill that in before merge.

Tests: 224 → 242 passing; `nx lint-eslint` clean.

sclarkamlab (approved):

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/MonacoEditor.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// TODO(GMA-384): replace Monaco with the Graphene editor component once it lands — https://gravitee.atlassian.net/browse/GMA-384
+export function MonacoEditor({
+    value,
+    onChange,
+    language = GAPL_LANGUAGE_ID,
+    readOnly = false,
+    height = '360px',
+    theme = 'light',
```
jourdiw: A bonus would be for this to inherit the theme of the application. 
This would be additional work and not a quick fix via the sdk...

jourdiw (approved):

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/MonacoEditor.tsx
podlesrafal: Done in d4212df — added a `'system'` theme value (now the default) that reads `prefers-color-scheme: dark` on mount and subscribes to changes, so the editor reacts when the OS switches. Explicit `'light'` / `'vs-dark'` props still override.

PR #17049, opened 2026-05-22
feat(authz-ui): per-service list pages + KPI dashboard (read-only)

podlesrafal:
Wires the placeholder pages introduced earlier into real list views. Each service (MCPs / AI Models / APIs / Custom Policies) renders the same shell: header, KPI strip, search + status filter, table of existing policies, delete-confirm dialog. Create / Edit stay disabled — the policy editor arrives in the final PR.

Shared presentational components
- StatusBadge: maps PolicyStatus → Graphene Badge variant with stable label strings (Draft / Deployed / Disabled)
- KpiTile: single dashboard tile (large numeric + label) on top of Graphene Card with `success` / `muted` / `default` tones
- ValidationErrorAlert: Graphene Alert+Title+Description that renders ApiError validation lists with a fallback for generic Errors

Service definitions
- service-defs/{mcp,llms,apis,custom}: per-service config records (icon, title, description, search placeholder, resource groups, condition snippets). Pages map config → ServicePolicyPage.
- pages/{Mcps,Llms,Apis,CustomPolicies}Page: thin route bindings replacing the placeholder stubs from PR1

ServicePolicyPage (read-only flavour)
- Header with icon + title + description + Refresh + disabled Create button (tooltip points at the follow-up PR)
- KPI strip: Policies / Deployed / Draft / Unique targets (last one hidden for Custom)
- Search + status filter (All / Draft / Deployed / Disabled)
- Empty states: distinct copy for "never authored anything" vs "filter hides everything"
- PolicyListTable below: existing list rendering with Edit no-op and Delete wired through a Graphene Dialog confirm

PolicyListTable
- Graphene DataTable in server-side mode with name + target + status + updated + actions columns (target column hidden for Custom)
- Name cell renders as a primary-color clickable button so the editor PR can plug onEdit without re-layout
- Updated cell formats UNIX-seconds floats (current backend shape) + RFC strings via a heuristic, then renders the relative time
- Pagination via Graphene DataTablePagination

## Issue

## Description

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/StatusBadge.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/StatusBadge.tsx
@@ -1,0 +1,31 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Badge } from '@gravitee/graphene-core';
+import type { ComponentProps } from 'react';
+import type { PolicyStatus } from '../shared/api/authz-api.types';
+
+type BadgeVariant = NonNullable<ComponentProps<typeof Badge>['variant']>;
+
+const STATUS_PRESENTATION: Record<PolicyStatus, { readonly label: string; readonly variant: BadgeVariant }> = {
+    DRAFT: { label: 'Draft', variant: 'outline' },
+    DEPLOYED: { label: 'Deployed', variant: 'success' },
+    DISABLED: { label: 'Disabled', variant: 'warning' },
+};
+
+export function StatusBadge({ status }: { readonly status: PolicyStatus }) {
+    const { label, variant } = STATUS_PRESENTATION[status];
+    return <Badge variant={variant}>{label}</Badge>;
+}
--- a/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyListTable.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyListTable.tsx
@@ -1,0 +1,200 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Badge, Button, DataTable, DataTablePagination } from '@gravitee/graphene-core';
+import { PencilIcon, Trash2Icon } from '@gravitee/graphene-core/icons';
+import type { ColumnDef } from '@tanstack/react-table';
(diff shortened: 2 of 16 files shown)


gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/StatusBadge.tsx
```diff
@@ -0,0 +1,39 @@
...
+import type { PolicyStatus } from '../shared/api/authz-api.types';
+
+type BadgeVariant = NonNullable<ComponentProps<typeof Badge>['variant']>;
+
+const STATUS_PRESENTATION: Record<PolicyStatus, { readonly label: string; readonly variant: BadgeVariant; readonly className?: string }> = {
+    DRAFT: { label: 'Draft', variant: 'outline' },
+    DEPLOYED: {
+        label: 'Deployed',
+        variant: 'default',
+        className: 'bg-green-100 text-green-700 hover:bg-green-100 dark:bg-green-950/50 dark:text-green-300',
```
sclarkamlab: This is off the palette, raw `green-*` plus the `dark:` overrides. We've got a `success` token for exactly this, so `bg-success/10 text-success hover:bg-success/10` and drop the `dark:` pair (the token handles dark mode). We already use it that way in `PolicyBuilder`/`PolicyStatementCard`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/ServicePolicyPage.tsx
```diff
@@ -0,0 +1,210 @@
...
+    };
+
+    const Icon = config.icon;
+    const totalCount = policies.data?.total ?? 0;
+    const hasNoPolicies = !policies.isLoading && totalCount === 0 && !deferredSearch && statusFilter === 'ALL';
+    const hasNoMatches = !policies.isLoading && filteredPolicies.length === 0 && totalCount > 0;
+
+    return (
+        <div className="flex flex-col gap-4" data-testid={`page-policies-${config.type.toLowerCase()}`}>
+            <header className="flex items-start justify-between gap-3">
```
sclarkamlab, line 109: Could this be `ContentHeader` rather than a hand-rolled `<header>`/`<h1>`? Unless the console shell already owns the page title, then ignore.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/StatusBadge.tsx
podlesrafal: Good catch — moved off the raw `green-*` palette. Went with `variant="success"` directly instead of overriding `bg-success/10 text-success`, since `Badge` already ships that variant via `badgeVariants` (with the proper `text-badge-success-foreground` token, focus ring, and dark-mode handling). Also bumped DISABLED to `variant="warning"` while I was there — the previous `secondary` blended visually with DRAFT's `outline`. Commit: 095b1e9385.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/ServicePolicyPage.tsx
podlesrafal: Checked the surface — graphene's `ContentHeader` is the breadcrumb bar (props: `breadcrumbs`, `leading`, `trailing`) and is already owned by the shell (`LocalDevShell.tsx` mounts it as the subheader). The console host itself also hand-rolls its `<h1>` (see `gravitee-gamma-control-plane-webui/src/pages/home/HomePage.tsx`), so there's no equivalent page-title primitive to swap in here. Leaving the `<header>`/`<h1>` as-is — happy to revisit if graphene adds a `PageHeader` (title/description/leading/trailing) primitive. Worth a separate issue on graphene-core?

sclarkamlab (approved):

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyListTable.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+    const columns = useMemo<ColumnDef<PolicyResponse>[]>(() => {
+        const cols: ColumnDef<PolicyResponse>[] = [
+            {
+                id: 'name',
+                header: 'Name',
+                cell: ({ row }) => {
+                    const policy = row.original;
+                    return (
+                        <button
```
jourdiw: Can you use the Graphene button?

podlesrafal: Done — replaced the raw `<button>` with Graphene `<Button variant="ghost">`. Used `h-auto`, `justify-start`, and `overflow-hidden` overrides to preserve the full-width, left-aligned, two-line layout.

jourdiw (approved):

PR #17050, opened 2026-05-22
feat(authz-ui): visual policy builder — chip combobox + statement cards

podlesrafal:
Adds the visual policy-authoring widgets — multi-select chip combobox, statement card with permit/forbid + condition snippets, and the orchestrator that stitches multiple statements together. Self-contained: the editor Sheet that hosts the builder lands in the next PR.

ChipCombobox (extracted to components/, hoist candidate for Graphene)
- Multi-select chip combobox built on Graphene Popover + Command; selected entries render as inline chips in the trigger, the unselected catalog lives in a grouped dropdown below the search
- Keeps unrecognised ids (typed in code view later) as "ghost" chips with warning styling + tooltip so round-trip parsers never silently drop a reference
- Width matches the trigger via the Radix `--radix-popper-anchor-width` CSS var; CommandList caps height ~5 rows and wires a manual onWheel because cmdk's overflow:hidden wrapper swallows native wheel propagation in some browsers
- Z-index bumped above the Sheet portal so the popover content stays clickable when the editor lands

PolicyStatementCard
- permit / forbid ToggleGroup with `success` / `destructive` semantic tints + bold weight when active; statement card border + ring picks up the same tone so the card visually reads as the chosen effect
- PRINCIPAL / ACTION / RESOURCE chip combos backed by the options prop; sync handlers preserve canonical `Type::"slug"` ids and parse unknown ids into structured refs so the visual editor never strips references
- Collapsible CONDITION textarea with one-click snippet chips ("+ Business hours", "+ Trusted device", "+ Corporate IP range", "+ Owner only") that append onto the existing condition
- Copy / delete / reorder controls on each card
- Exports `ChipOption` as a re-export from `lib/chip-option` so the existing import path (`from './PolicyStatementCard'`) stays valid

PolicyBuilder
- Renders a list of PolicyStatementCard, threads add / duplicate / remove / reorder through to the parent's onChange
- "+ Add permit statement" / "+ Add forbid statement" buttons below the list, vertically stacked, with tinted outlines matching the effect; "forbid wins" footnote underneath surfaces the GAPL precedence rule

Plus PolicyStatementCard + PolicyBuilder unit tests.

## Issue

## Description


gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyBuilder.tsx
```diff
@@ -0,0 +1,135 @@
...
+                    emptyResourcesHint={emptyResourcesHint}
+                />
+            ))}
+            <div className="flex flex-col items-start gap-1.5 pt-1">
+                <Button
+                    type="button"
+                    variant="outline"
+                    size="sm"
+                    onClick={() => addStatement('permit')}
+                    className="h-7 px-2 text-xs border-success/40 text-success hover:bg-success/10 hover:text-success"
```
sclarkamlab: These get re-sized with `h-7 px-2 text-xs` on top of `size="sm"`. `size="xs"` should cover it, then drop the h-7/text-xs (keep the `border-success`/`text-success` bits).

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyStatementCard.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+                <div className="flex items-center gap-1">
+                    {canMoveUp && (
+                        <Button
+                            type="button"
+                            variant="ghost"
+                            size="sm"
+                            onClick={onMoveUp}
+                            aria-label="Move statement up"
+                            className="h-7 w-7 p-0"
```
sclarkamlab: These are icon buttons sized by hand. `size="icon-sm"` instead of `size="sm"` + `h-7 w-7 p-0`. `PolicyListTable` already uses `size="icon"` for the same edit/delete buttons.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyStatementCard.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+            <div className="mt-2">
+                <button
+                    type="button"
+                    onClick={() => setConditionOpen(o => !o)}
+                    className="flex w-full items-center gap-1.5 rounded-md px-1 py-0.5 text-left text-xs font-medium uppercase tracking-wide text-muted-foreground hover:text-foreground"
+                >
+                    <ChevronDownIcon className={cn('size-3 transition-transform', conditionOpen ? 'rotate-0' : '-rotate-90')} aria-hidden />
+                    <span>Condition</span>
+                    <span className="text-[10px] normal-case tracking-normal text-muted-foreground/70">Optional</span>
```
sclarkamlab: `text-[10px]` is below the scale, `text-xs` is the floor.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyStatementCard.tsx
```diff
@@ -0,0 +1,286 @@
...
+                    {canMoveUp && (
+                        <Button
+                            type="button"
+                            variant="ghost"
+                            size="sm"
+                            onClick={onMoveUp}
+                            aria-label="Move statement up"
+                            className="h-7 w-7 p-0"
+                        >
+                            ↑
```
sclarkamlab: Could use `ChevronUpIcon`/`ChevronDownIcon` from `@gravitee/graphene-core/icons` rather than the raw arrow glyphs. Minor.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/ChipCombobox.tsx
```diff
@@ -0,0 +1,205 @@
...
+                    )}
+                    <ChevronDownIcon className="ml-auto size-3.5 shrink-0 text-muted-foreground" aria-hidden />
+                </button>
+            </PopoverTrigger>
+            <PopoverContent
+                className="w-[var(--radix-popper-anchor-width)] min-w-60 p-0"
+                align="start"
+                id={listboxId}
+                role="listbox"
+                style={{ zIndex: 60 }}
```
sclarkamlab: `zIndex`/`maxHeight` as inline styles, `z-*` and `max-h-*` utilities would keep it on the scale (`260px` is off it anyway). Nit.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/ChipCombobox.tsx
```diff
@@ -0,0 +1,205 @@
...
+                            <Badge
+                                key={c.id}
+                                variant={c.ghost ? 'outline' : 'secondary'}
+                                title={c.ghost ? `${c.id} — not in catalog` : undefined}
+                                className={cn('gap-1 pr-1', c.ghost && 'border-dashed border-warning bg-warning/10 text-warning')}
+                            >
+                                {c.ghost ? <TriangleAlertIcon className="size-3" aria-hidden /> : null}
+                                {c.label}
+                                <span
+                                    role="button"
```
sclarkamlab: This is a `<span role="button">` nested inside the trigger `<button>`, which is invalid html and awkward for screen readers. Known tension with the chip-in-combobox pattern, just checking it's deliberate.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyBuilder.tsx
podlesrafal: Done — switched both Buttons to `size="xs"` and dropped the `h-7 px-2 text-xs` hand-roll, kept the `border-success`/`text-success` and the matching `border-destructive`/`text-destructive` bits. Commit d64f4ac216.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyStatementCard.tsx
podlesrafal: Done — all 4 buttons (move up, move down, duplicate, delete) now use `size="icon-sm"` and the `h-7 w-7 p-0` hand-sizing is gone. Commit d64f4ac216.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyStatementCard.tsx
podlesrafal: Done — `text-[10px]` → `text-xs`. Commit d64f4ac216.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyStatementCard.tsx
podlesrafal: Done — `↑`/`↓` glyphs replaced with `ChevronUpIcon`/`ChevronDownIcon` from `@gravitee/graphene-core/icons`. Commit d64f4ac216.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/ChipCombobox.tsx
podlesrafal: `maxHeight` fixed — `style={{ maxHeight: '260px' }}` → `className="max-h-64"` (256px, on-scale).

Left `style={{ zIndex: 60 }}` for now: graphene's prebuilt CSS exposes only `z-50` (the Tailwind default), and this popover needs to render above the editor sheet/dialog which already sits at z-50. Happy to switch to `z-[60]` arbitrary or wire through a `z-popover` token if graphene adds one. Commit d64f4ac216.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/ChipCombobox.tsx
podlesrafal: Yeah, deliberate. The trigger is one `<button>` to keep "click anywhere in the field opens the picker" (same affordance as the rest of graphene's input/combobox surfaces). The per-chip ✕ then has to live inside, so it's `<span role="button">` rather than nesting a real `<button>`. Trade-offs I weighed:
- `<div role="combobox">` as the trigger — wins on HTML validity but loses native button focus/Space-Enter semantics, would have to hand-roll the keyboard layer.
- Chips outside the trigger — clean HTML but breaks the click-anywhere affordance.
- Backspace-only removal — keyboard works, but loses pointer click on the ✕.

Tracking as a follow-up for an a11y pass. For this PR I think the trade-off is OK, but happy to revisit if you have a stronger pattern in mind.

sclarkamlab (approved):

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/ChipCombobox.tsx
```diff
@@ -0,0 +1,204 @@
...
+import { useId, useMemo, useState } from 'react';
+
+export const GAPL_UID_PATTERN = /^([^:]+)::"(.+)"$/;
+
+export interface ChipComboboxOption {
+    readonly id: string;
+    readonly label: string;
+    readonly group: string;
+    readonly description?: string;
+}
```
jourdiw: **[Score: 45/100]** 💡 `ChipComboboxOption` is structurally identical to `ChipOption` defined in `shared/chip-option.ts`. Consumers like `PolicyStatementCard` already import `ChipOption` — having a second interface with the same shape creates a maintenance risk: if one evolves and the other doesn't, callers silently break. Consider importing `ChipOption` here and re-exporting it (or aliasing it) rather than declaring a second interface.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/ChipCombobox.tsx
```diff
@@ -0,0 +1,204 @@
...
+                            <Badge
+                                key={c.id}
+                                variant={c.ghost ? 'outline' : 'secondary'}
+                                title={c.ghost ? `${c.id} — not in catalog` : undefined}
+                                className={cn('gap-1 pr-1', c.ghost && 'border-dashed border-warning bg-warning/10 text-warning')}
+                            >
+                                {c.ghost ? <TriangleAlertIcon className="size-3" aria-hidden /> : null}
+                                {c.label}
+                                <span
+                                    role="button"
```
jourdiw: **[Score: 55/100]** 💡 Using `<span role="button">` for the chip-remove trigger requires manual keyboard handling (Enter + Space) and manual `tabIndex`. A native `<button>` element gives all of that for free, plus it participates correctly in the accessibility tree without ARIA role overrides. This matters particularly inside a combobox where screen-reader users need clear affordances.

Suggestion:
```tsx
<button
    type="button"
    aria-label={`Remove ${c.label}`}
    className="cursor-pointer rounded-full opacity-60 hover:opacity-100"
    onClick={e => { e.stopPropagation(); toggle(c.id); }}
>
    ×
</button>
```

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/ChipCombobox.tsx
```diff
@@ -0,0 +1,204 @@
...
+                    )}
+                    <ChevronDownIcon className="ml-auto size-3.5 shrink-0 text-muted-foreground" aria-hidden />
+                </button>
+            </PopoverTrigger>
+            <PopoverContent
+                className="w-[var(--radix-popper-anchor-width)] min-w-60 p-0"
+                align="start"
+                id={listboxId}
+                role="listbox"
+                style={{ zIndex: 60 }}
```
jourdiw: **[Score: 50/100]** 💡 Hardcoded `style={{ zIndex: 60 }}` bypasses the design system's z-index scale. The PR description explains this is needed to stay above the Sheet portal, but inline z-index values tend to accumulate and conflict over time. Consider using a Tailwind utility class (e.g. `z-[60]`) or, if graphene-core defines z-index tokens, prefer those. If this value is truly tied to the Sheet stacking context, a named CSS variable (`--z-above-sheet`) would document the intent and make it greppable for the follow-up PR that introduces the Sheet.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/ServicePolicyPage.tsx
```diff
@@ -0,0 +1,210 @@
...
+    const confirmDelete = async () => {
+        if (!pendingDelete) return;
+        setDeleting(true);
+        try {
+            await policies.remove(pendingDelete.id);
+            setPendingDelete(null);
+        } finally {
+            setDeleting(false);
+        }
+    };
```
jourdiw, line 109: **[Score: 72/100]** ⚠️ `confirmDelete` silently swallows errors. If `policies.remove()` rejects, the `try` block exits early (skipping `setPendingDelete(null)`) and the `finally` resets the spinner — but no error feedback is shown to the user. The dialog stays open in an ambiguous state.

Consider adding a `catch` block that surfaces the error, e.g. via `ValidationErrorAlert` (already available in this module) or a toast:

```ts
const confirmDelete = async () => {
    if (!pendingDelete) return;
    setDeleting(true);
    try {
        await policies.remove(pendingDelete.id);
        setPendingDelete(null);
    } catch (err) {
        // Surface the error — dialog remains open so the user can retry
        console.error('Delete failed', err);
        // TODO: show inline error or toast
    } finally {
        setDeleting(false);
    }
};
```

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/statement-to-gapl.ts
```diff
@@ -0,0 +1,121 @@
...
+    return `${lines.join('\n')};`;
+}
+
+export function statementsToGapl(name: string, statements: readonly PolicyStatement[], target?: { label: string }): string {
+    const header = target ? `// Policy: ${name}\n// Target: ${target.label}` : `// Policy: ${name}`;
+    const body = statements.map(statementToGapl).join('\n\n');
+    return `${header}\n\n${body}`;
+}
+
+let emptyStatementSeq = 0;
```
jourdiw: **[Score: 55/100]** 💡 Module-level mutable counter `emptyStatementSeq` persists across component lifecycles and test runs. In a Vitest environment with shared module state, this means IDs are non-deterministic across tests — today's tests work around it with regex matching (`/^stmt-\d+$/`), but any future test that checks for a specific ID will be flaky.

Consider making ID generation injectable or using a self-contained approach:

```ts
export function createEmptyStatement(effect: PolicyEffect = 'permit'): PolicyStatement {
    return {
        id: `stmt-${crypto.randomUUID().slice(0, 8)}`,
        effect,
        principals: [],
        actions: [],
        resources: [],
        condition: '',
    };
}
```

This also aligns better with the `Math.random()` approach used in `PolicyBuilder.duplicateStatement`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyBuilder.tsx
```diff
@@ -0,0 +1,135 @@
...
+}: PolicyBuilderProps) {
+    const addStatement = (effect: PolicyEffect) => onChange([...statements, createEmptyStatement(effect)]);
+
+    const updateStatement = (next: PolicyStatement) => onChange(statements.map(s => (s.id === next.id ? next : s)));
+
+    const duplicateStatement = (index: number) => {
+        const source = statements[index];
+        const copy: PolicyStatement = {
+            ...source,
+            id: `stmt-${Math.random().toString(36).slice(2, 9)}`,
```
jourdiw: **[Score: 30/100]** 🎨 Minor inconsistency: `duplicateStatement` uses `Math.random().toString(36).slice(2, 9)` for ID generation, while `createEmptyStatement` in `statement-to-gapl.ts` uses a module-level incrementing counter. Both are UI-only IDs that don't survive round-trips (as documented), so this isn't a correctness issue — but unifying to one strategy would reduce cognitive load.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/PolicyStatementCard.tsx
```diff
@@ -0,0 +1,265 @@
...
+                    <div className="mt-1.5 space-y-1.5">
+                        <Textarea
+                            value={statement.condition ?? ''}
+                            onChange={e => onChange({ ...statement, condition: e.target.value })}
+                            placeholder="e.g. context.time.hour >= 9 && context.time.hour < 17"
+                            className="min-h-12 font-mono text-xs leading-relaxed"
+                        />
+                        <div className="flex flex-wrap gap-1">
+                            {snippets.map(s => (
+                                <button
```
jourdiw, line 260: **[Score: 35/100]** 🎨 The condition snippet buttons use raw `<button>` with hand-rolled Tailwind classes instead of the graphene-core `Button` component used everywhere else in this PR. While the visual intent (dashed-border chip) is distinct from a standard button, graphene-core's `variant="outline" size="xs"` with custom className overrides (as done for the 'Add permit/forbid' buttons in PolicyBuilder) would keep the design system consistent and get focus-visible styles for free.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-management/ServicePolicyPage.tsx
```diff
@@ -0,0 +1,210 @@
...
+    }, [allPolicies, deferredSearch]);
+
+    const kpis = useMemo(
+        () => ({
+            total: policies.data?.total ?? allPolicies.length,
+            deployed: allPolicies.filter(p => p.status === 'DEPLOYED').length,
+            draft: allPolicies.filter(p => p.status === 'DRAFT').length,
+            uniqueTargets: new Set(allPolicies.map(p => p.target?.id).filter(Boolean)).size,
+        }),
+        [allPolicies, policies.data?.total],
```
jourdiw: **[Score: 60/100]** 💡 The KPI metrics (`deployed`, `draft`, `uniqueTargets`) are computed client-side by filtering the full `allPolicies` array. But `allPolicies` comes from a single paginated API call — when `perPage < total`, the counts will only reflect the current page, not the entire dataset. For example, a user with 50 policies viewing page 1 at 10/page will see `deployed: N` based only on the first 10 rows.

Either:
1. Fetch aggregate counts from a dedicated backend endpoint (ideal).
2. Document that KPIs reflect the current page only (acceptable for now if the backend doesn't expose aggregates yet).
3. Or fetch all pages in the background for accurate counts (costly).

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/components/ChipCombobox.tsx
jourdiw, line 1: I would recommend using the Combobox coming from graphene-core. 
I am working on a fix for the scrolling.

jourdiw (approved):

PR #17055, opened 2026-05-22
feat: bulk publish and cascade delete for portal navigation folders

adam-rosowicz:
**Issue**

https://gravitee.atlassian.net/browse/EXT-126

**Summary**

Fix publish cascade: PortalNavigationItemDomainService.update() previously only propagated changes to children when published → false or visibility → PRIVATE. This meant publishing a folder left all children unpublished. Now publish cascades symmetrically with unpublish.

Cascade delete: Deleting a folder now recursively deletes its entire subtree. The old has-children guard (PortalNavigationItemHasChildrenException) is removed. The implementation reads the full subtree in a single findAllByRootId query, collects descendant IDs in memory, then bulk-deletes via a new deleteByIds operation (MongoDB uses @Query(delete=true); JDBC batches at 500 items). Page content is cleaned up before the items are deleted. Sibling reordering at the root level is preserved via the existing delete(item) call.

Frontend: Publish confirmation dialog now shows a cascade warning for container types (FOLDER / API section). Delete confirmation includes a nested-items warning when the node has children.

Legacy items without rootId: items created before the rootId migration have rootId = ZERO. Rather than throwing, cascade delete falls back to a recursive findByParentIdAndEnvironmentId traversal for these items — same strategy used by attribute propagation.

https://github.com/user-attachments/assets/03ffbf06-99eb-4a4f-8eca-6e06b03fd79b


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationItemDomainService.java
```diff
@@ -163,13 +225,9 @@ public PortalNavigationItem update(UpdatePortalNavigationItem toUpdate, PortalNa
...
         var updatedItem = crudService.update(originalItem);
-        if (PortalVisibility.PRIVATE.equals(changedVisibility) || Boolean.FALSE.equals(changedPublished)) {
-            propagateAttributesToDescendants(
-                updatedItem.getId(),
-                updatedItem.getEnvironmentId(),
-                PortalVisibility.PRIVATE.equals(changedVisibility) ? PortalVisibility.PRIVATE : null,
-                Boolean.FALSE.equals(changedPublished) ? Boolean.FALSE : null
-            );
+
+        if (changedVisibility != null || changedPublished != null) {
```
aqibmohammadkhan: Publishing cascade also makes `private` descendants `public`. The PR changed propagation to run for any visibility change, not just `PRIVATE`. When a private folder/API is made `public`, all nested private docs/APIs become public too, which seems broader than the PR’s stated “publish cascades symmetrically” behavior and can unintentionally expose private content.

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/resources/scripts/create-index.js
```diff
@@ -168,3 +168,11 @@ db.getCollection(`${prefix}custom_domain`).dropIndexes();
...
 db.getCollection(`${prefix}custom_domain`).createIndex({ referenceId: 1, referenceType: 1 }, { name: "ri1rt1" });
 db.getCollection(`${prefix}custom_domain`).reIndex();
+
+// "portal_navigation_items" collection
+db.getCollection(`${prefix}portal_navigation_items`).dropIndexes();
+db.getCollection(`${prefix}portal_navigation_items`).createIndex({ organizationId: 1, environmentId: 1 }, { name: "oi1ei1" });
+db.getCollection(`${prefix}portal_navigation_items`).createIndex({ parentId: 1, environmentId: 1 }, { name: "pi1ei1" });
+db.getCollection(`${prefix}portal_navigation_items`).createIndex({ area: 1, environmentId: 1 }, { name: "a1ei1" });
+db.getCollection(`${prefix}portal_navigation_items`).createIndex({ rootId: 1 }, { name: "ri1" });
+db.getCollection(`${prefix}portal_navigation_items`).reIndex();
```
aqibmohammadkhan: this now drops all `portal_navigation_items` indexes but does not recreate the existing standalone environmentId index (e1) that the Mongo index upgrader maintains. Running this script against an existing DB would remove that index and may regress environment-scoped queries.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationItemDomainService.java
```diff
@@ -163,13 +225,9 @@ public PortalNavigationItem update(UpdatePortalNavigationItem toUpdate, PortalNa
...
+            propagateAttributesToDescendants(updatedItem.getId(), updatedItem.getEnvironmentId(), changedVisibility, changedPublished);
```
aqibmohammadkhan: There is no test for "folder visibility changes from PRIVATE to PUBLIC" and no UI warning for it. A user changing a parent from private to public will silently make all nested docs public.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationItemDomainService.java
aqibmohammadkhan: The minimal fix is to keep visibility one-directional:

```
if (PortalVisibility.PRIVATE.equals(changedVisibility) || changedPublished != null) {
    propagateAttributesToDescendants(
        updatedItem.getId(),
        updatedItem.getEnvironmentId(),
        PortalVisibility.PRIVATE.equals(changedVisibility) ? PortalVisibility.PRIVATE : null,
        changedPublished
    );
}
```

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/resources/scripts/create-index.js
aqibmohammadkhan: `deleteByEnvironmentId` queries WHERE environment_id = ? — none of the composite indexes have `environmentId` as the leading key, so this query will scan the whole collection without `e1`. In practice `ensureIndex` from the Java upgrader restores e1 on startup, so runtime is safe, but the script is incorrect as a reference and misleading for fresh installs from JS only. Add { environmentId: 1 }, { name: "e1" } to the block.

aqibmohammadkhan (dismissed):

aqibmohammadkhan (commented):
Please take care of the comments.

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/resources/scripts/create-index.js
dev-marek, line 1: We no longer use this file for indexes

dev-marek (changes requested):
only one thing from my side

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/resources/scripts/create-index.js
adam-rosowicz: Thanks, I have reverted the changes in the file.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationItemDomainService.java
konrad-krol-incubly: The concern is addressed in code: visibility only propagates downward when set to PRIVATE — changing a folder from PRIVATE to PUBLIC passes null as visibilityToPropagate, so no propagation fires and children retain their own visibility.

A test covering this exact case was added: `should_not_propagate_visibility_when_folder_changes_from_private_to_public` in `PortalNavigationItemDomainServiceTest`.

No UI warning is needed since the operation has no side effect on nested items.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/query_service/PortalNavigationItemsQueryService.java
```diff
@@ -34,6 +34,8 @@ public interface PortalNavigationItemsQueryService {
 
     List<PortalNavigationItem> findTopLevelItemsByEnvironmentIdAndPortalArea(String environmentId, PortalArea portalArea);
 
+    List<PortalNavigationItem> findAllByRootId(PortalNavigationItemId rootId);
```
konrad-krol-incubly: Unlike every other method in this interface, this one doesn't scope by environmentId. Worth adding for consistency and isolation.

aqibmohammadkhan: yes, it matches how portal navigation is modeled everywhere else in APIM.
please add it.

konrad-krol-incubly: done

konrad-krol-incubly (approved):

dev-marek (approved):

PR #17061, opened 2026-05-25
feat: support cors for MCP Proxy

jhaeyaert:
## Issue

https://gravitee.atlassian.net/browse/APIM-14202

## Description

This PR allows configuring CORS on an MCP Proxy. This is useful when protecting the MCP Proxy with an OAuth2 plan and using a browser-based MCP Client. In that case, the MCP Client expects proper Allow-Origin-XXX headers, which should be served by the gateway that handles the oauth-protected resource metadata itself.

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/api-navigation/api-v4-menu.service.spec.ts
+++ b/gravitee-apim-console-webui/src/management/api/api-navigation/api-v4-menu.service.spec.ts
@@ -115,6 +115,64 @@ describe('ApiV4MenuService', () => {
     expect(menu.subMenuItems.some(item => item.displayName === 'Webhooks')).toBe(false);
   });
 
+  describe('Entrypoints menu tabs', () => {
+    const httpListener = {
+      type: 'HTTP' as const,
+      paths: [{ path: '/test' }],
+      entrypoints: [{ type: 'http-proxy' }],
+    };
+
+    it('should expose Entrypoints and CORS tabs for MCP_PROXY (no MCP Entrypoint, no Response Templates)', () => {
+      TestBed.overrideProvider(GioTestingPermissionProvider, { useValue: ['api-response_templates-r'] });
+      service = TestBed.inject(ApiV4MenuService);
+
+      const api = fakeApiV4({
+        type: 'MCP_PROXY',
+        listeners: [{ ...httpListener, entrypoints: [{ type: 'mcp-proxy' }] }],
+      });
+
+      const menu = service.getMenu(api);
+      const entrypoints = menu.subMenuItems.find(item => item.displayName === 'Entrypoints');
+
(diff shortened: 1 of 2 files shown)


phiz71 (approved):

gravitee-apim-console-webui/src/management/api/api-navigation/api-v4-menu.service.ts
```diff
@@ -159,7 +159,7 @@ export class ApiV4MenuService implements ApiMenuService {
       },
     };
 
-    if (api.type === 'NATIVE' || api.type === 'MCP_PROXY' || api.type === 'LLM_PROXY') {
+    if (api.type === 'NATIVE' || api.type === 'LLM_PROXY') {
```
michel-barret, line 162: suggestion: 🙏

```suggestion
    if (api.type === 'NATIVE') {
```

michel-barret (approved):

gravitee-apim-console-webui/src/management/api/api-navigation/api-v4-menu.service.spec.ts
```diff
@@ -115,6 +115,64 @@ describe('ApiV4MenuService', () => {
...
+    it('should expose Entrypoints, MCP Entrypoint, Response Templates and CORS tabs for PROXY', () => {
+      TestBed.overrideProvider(GioTestingPermissionProvider, { useValue: ['api-response_templates-r'] });
+      service = TestBed.inject(ApiV4MenuService);
+
+      const api = fakeApiV4({ type: 'PROXY', listeners: [httpListener] });
+
+      const menu = service.getMenu(api);
+      const entrypoints = menu.subMenuItems.find(item => item.displayName === 'Entrypoints');
+
+      expect(entrypoints.tabs?.map(t => t.displayName)).toEqual(['Entrypoints', 'MCP Entrypoint', 'Response Templates', 'CORS']);
```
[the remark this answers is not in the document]
jhaeyaert: Looks irrelevant to me 🤔, MCP Entrypoint tab already takes the case into consideration
image

gravitee-apim-console-webui/src/management/api/api-navigation/api-v4-menu.service.ts
jhaeyaert: That would enable it for LLM Proxy, right? 
Maybe we should consider it out of scope for now, as I don't see any value for CORS with LLM Proxy yet (unless you have a browser-based agent in mind?).

I don't think it will cost us to add it later if necessary. It's kind of trivial imho.

michel-barret: yes, but I’m ok to do it in another PR

PR #17086, opened 2026-05-25
fix: route v4 API PATCH polymorphic deserialization through REST DTO layer

carlos-andres-osorio:
## Issue

[APIM-14083](https://gravitee.atlassian.net/browse/APIM-14083)

## Why

`PatchApiUseCase` was deserializing polymorphic fields (`listeners`, `endpointGroups`, `flows`, `resources`) from the merged JSON node directly to domain classes (e.g. `Listener.class`, `EndpointGroup.class`), bypassing the REST DTO layer that handles OpenAPI discriminator normalisation. This caused PATCH to silently reject discriminator values that GET returns and PUT accepts — for example, `"type": "HTTP"` for listeners. APIM-14077 added a per-field port for flows; this change generalises the fix by routing all polymorphic fields through the same DTO deserialization chain that PUT uses.

## What changed

- **`PatchApiUseCase`** — introduced `ApiV4Deserializer` port (injected via constructor) with `toCurrentStateNode` / `fromPatchedNode` methods. Replaced the four `resolvePatchableList` calls for `listeners`, `endpointGroups`, `flows`, and `resources` with a single `resolveApiV4Fields` delegation to the port. Added generic `resolveField` helper to remove duplication in `resolveResponseTemplates` and `resolvePatchableList`. Fixed `deserializationLocation` to avoid repeating the root field name in the JSON path. Made `PatchableView` and its `from(Api)` factory public for reuse by the adapter.
- **`PatchApiV4Deserializer`** (new, `adapter` package) — REST-layer implementation of `ApiV4Deserializer`. `toCurrentStateNode` builds the current-state JSON node by serializing `PatchableView` then overlaying the REST DTO representations of the four polymorphic fields via `ApiMapper.mapToV4`. `fromPatchedNode` deserializes the merged node back through Jackson's `GenericApi` polymorphism (resolves to `ApiV4`) and maps it to `ApiV4Fields` via `ApiMapper.mapToCoreApi`.
- **`ApiMapper`** — added `mapToCoreApi(ApiV4)` overload that maps each REST DTO collection to its domain type via the existing `ListenerMapper`, `EndpointMapper`, `FlowMapper`, and `ResourceMapper` chains; added a null guard in `computeCoreApiLinks` so `mapToV4` is callable without a `UriInfo` context.
- **Spring configs** — wired `PatchApiV4Deserializer` bean in `RestManagementConfiguration` (v2) and `RestAutomationConfiguration`; provided a fail-fast stub in the v1 `RestManagementConfiguration` (v4 PATCH is not served there).

## User-facing impact

PATCH on v4 HTTP Proxy APIs now accepts the same discriminator values for `listeners`, `endpointGroups`, `flows`, and `resources` as GET and PUT. Clients sending the uppercase values returned by GET (e.g. `"type": "HTTP"`) that were previously rejected will now be accepted.

## Gotchas / Follow-up

The `FlowListDeserializer` port introduced by APIM-14077 is now superseded by the `ApiV4Deserializer` port — flows are handled through the same path as the other polymorphic fields. That port can be removed in a follow-up once this lands.

[APIM-14083]: https://gravitee.atlassian.net/browse/APIM-14083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/adapter/PatchApiV4Deserializer.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/adapter/PatchApiV4Deserializer.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.management.v2.rest.adapter;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.gravitee.apim.core.api.model.Api;
+import io.gravitee.apim.core.api.use_case.PatchApiUseCase.ApiV4Deserializer;
+import io.gravitee.apim.core.api.use_case.PatchApiUseCase.ApiV4Fields;
+import io.gravitee.apim.core.api.use_case.PatchApiUseCase.PatchableView;
+import io.gravitee.rest.api.management.v2.rest.mapper.ApiMapper;
+import io.gravitee.rest.api.management.v2.rest.model.ApiV4;
+import io.gravitee.rest.api.management.v2.rest.model.GenericApi;
+import java.io.IOException;
+import java.util.List;
+
+public class PatchApiV4Deserializer implements ApiV4Deserializer {
+
+    private static final List<String> POLYMORPHIC_LIST_FIELDS = List.of("listeners", "endpointGroups", "flows", "resources");
+
+    private final ObjectMapper objectMapper;
+
+    public PatchApiV4Deserializer(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+    }
+
+    @Override
+    public JsonNode toCurrentStateNode(Api api) {
+        var node = (ObjectNode) objectMapper.valueToTree(PatchableView.from(api));
+        var apiV4 = ApiMapper.INSTANCE.mapToV4(api, null, null);
+        setIfNonNull(node, "listeners", apiV4.getListeners());
+        setIfNonNull(node, "endpointGroups", apiV4.getEndpointGroups());
+        setIfNonNull(node, "flows", apiV4.getFlows());
+        setIfNonNull(node, "resources", apiV4.getResources());
+        return node;
+    }
+
+    @Override
+    public ApiV4Fields fromPatchedNode(JsonNode patchedNode) throws IOException {
+        var slim = objectMapper.createObjectNode();
+        slim.put("definitionVersion", "V4");
+        for (var field : POLYMORPHIC_LIST_FIELDS) {
+            var n = patchedNode.get(field);
+            if (n != null) {
+                slim.set(field, n);
+            }
+        }
+        var apiV4 = (ApiV4) objectMapper.treeToValue(slim, GenericApi.class);
+        return ApiMapper.INSTANCE.mapToCoreApi(apiV4);
+    }
+
+    private void setIfNonNull(ObjectNode node, String field, Object value) {
+        if (value != null) {
+            node.set(field, objectMapper.valueToTree(value));
+        }
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapper.java
@@ -22,16 +22,21 @@ import io.gravitee.apim.core.api.model.NewNativeApi;
 import io.gravitee.apim.core.api.model.UpdateNativeApi;
 import io.gravitee.apim.core.api.model.crd.ApiCRDSpec;
 import io.gravitee.apim.core.api.model.import_definition.ApiExport;
+import io.gravitee.apim.core.api.use_case.PatchApiUseCase;
 import io.gravitee.apim.core.documentation.model.Page;
 import io.gravitee.apim.core.membership.model.PrimaryOwnerEntity;
 import io.gravitee.apim.core.utils.CollectionUtils;
 import io.gravitee.definition.model.v4.ApiType;
 import io.gravitee.definition.model.v4.endpointgroup.AbstractEndpoint;
 import io.gravitee.definition.model.v4.endpointgroup.AbstractEndpointGroup;
+import io.gravitee.definition.model.v4.endpointgroup.EndpointGroup;
 import io.gravitee.definition.model.v4.flow.AbstractFlow;
+import io.gravitee.definition.model.v4.flow.Flow;
 import io.gravitee.definition.model.v4.listener.AbstractListener;
+import io.gravitee.definition.model.v4.listener.Listener;
 import io.gravitee.definition.model.v4.listener.entrypoint.AbstractEntrypoint;
 import io.gravitee.definition.model.v4.nativeapi.NativeAnalytics;
+import io.gravitee.definition.model.v4.resource.Resource;
 import io.gravitee.definition.model.v4.service.AbstractApiServices;
 import io.gravitee.node.logging.NodeLoggerFactory;
 import io.gravitee.rest.api.management.v2.rest.model.Analytics;
@@ -411,6 +416,9 @@ public interface ApiMapper {
 
     @Named("computeCoreApiLinks")
     default ApiLinks computeCoreApiLinks(io.gravitee.apim.core.api.model.Api api, UriInfo uriInfo) {
+        if (uriInfo == null) {
+            return null;
+        }
         return new ApiLinks()
             .pictureUrl(ManagementApiLinkHelper.apiPictureURL(uriInfo.getBaseUriBuilder(), api))
             .backgroundUrl(ManagementApiLinkHelper.apiBackgroundURL(uriInfo.getBaseUriBuilder(), api));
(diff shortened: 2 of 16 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
```diff
@@ -510,96 +493,105 @@ private Api applyPatchedValues(Api existingApi, JsonNode patchedNode, PatchType
...
-            if (rawPatchNode.get(FIELD_RESPONSE_TEMPLATES).isNull()) {
-                return null;
-            }
-            return parseResponseTemplates(patchedNode);
+        if (patchType == PatchType.MERGE_PATCH && rawPatchNode.has(field)) {
+            return rawPatchNode.get(field).isNull() ? null : patched.get();
         }
         if (patchType == PatchType.JSON_PATCH) {
-            if (patchedNode.get(FIELD_RESPONSE_TEMPLATES) == null && isNullifiedByJsonPatch(rawPatchNode, FIELD_RESPONSE_TEMPLATES)) {
+            if (patchedNode.get(field) == null && isNullifiedByJsonPatch(rawPatchNode, field)) {
```
wbabyte, line 543: JSON_PATCH falls into this branch even when the user never touched the field — `toCurrentStateNode` always emits all four polymorphic fields, so `patchedNode.has(field)` is always true.

So a `replace /name` still runs listeners/endpointGroups/flows/resources through serialise → REST DTO → deserialise → domain. Any non-idempotent corner (mapper defaults, set ordering, null/empty coercion) silently drifts state the user didn't touch.

`preserves_every_non_patched_field` won't catch it — it uses the stub deserializer, not `PatchApiV4Deserializer`.

Either short-circuit when the raw ops don't target `/<field>` (or a sub-pointer), or add an integration test that JSON_PATCHes `/name` against an API with rich listeners/endpointGroups/flows/resources and asserts equality of those four.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapper.java
```diff
@@ -514,6 +522,22 @@ default AbstractApiServices mapApiV4Services(ApiV4 apiV4) {
         }
     }
 
+    default PatchApiUseCase.ApiV4Fields mapToCoreApi(ApiV4 apiV4) {
+        List<Listener> listeners = apiV4.getListeners() == null
+            ? null
+            : ListenerMapper.INSTANCE.mapToListenerEntityV4List(apiV4.getListeners()).stream().filter(Objects::nonNull).toList();
```
wbabyte: `.filter(Objects::nonNull)` silently drops null elements from each list. `{"listeners":[null]}` returns 200 with N-1 listeners — no error, no warning.

Worth failing the request (or at least logging the dropped indices) instead of swallowing.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiEndpointGroupsTest.java
```diff
@@ -180,6 +183,29 @@ private static String toJson(Object value) {
         }
     }
 
+    @Test
+    void patch_endpoint_group_type_http_proxy_returns_200_and_persists() {
```
wbabyte, line 187: Two gaps vs the listeners test:

1. Discriminator parameterisation — listeners covers HTTP/TCP/SUBSCRIPTION/KAFKA (`ApiResource_PatchApiListenersTest:169`). endpointGroups only exercises `"http-proxy"` (the *endpoint* type, not a group-level discriminator). The bug class this PR fixes is about discriminator round-trip — endpointGroups needs the equivalent parameterisation.

2. AC3 preservation trio missing. Listeners has merge-omit (l.208), merge-null (l.232), json-remove (l.243). Worth mirroring here.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
```diff
@@ -103,12 +103,36 @@ class PatchApiUseCaseTest {
 
     static final ObjectMapper OBJECT_MAPPER = new GraviteeMapper(false);
 
-    private static final PatchApiUseCase.FlowListDeserializer FLOW_DESERIALIZER = node ->
-        OBJECT_MAPPER.<List<Flow>>readerFor(OBJECT_MAPPER.getTypeFactory().constructCollectionType(List.class, Flow.class))
-            .with(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE)
-            .readValue(node);
-
-    private static final PatchApiUseCase.FlowListSerializer FLOW_SERIALIZER = flows -> OBJECT_MAPPER.valueToTree(flows);
+    private static final PatchApiUseCase.ApiV4Deserializer API_V4_DESERIALIZER = new PatchApiUseCase.ApiV4Deserializer() {
```
wbabyte, line 106: This stub bypasses `ApiMapper.mapToCoreApi` and the `GenericApi` envelope, so use-case tests can't catch a regression where the stub passes but the real adapter doesn't.

Worth at least one happy-path test wiring `PatchApiUseCase` to the actual `PatchApiV4Deserializer`, so the seam between nullification semantics and DTO round-trip gets exercised end-to-end.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/adapter/PatchApiV4Deserializer.java
```diff
@@ -0,0 +1,71 @@
...
+    private final ObjectMapper objectMapper;
+
+    public PatchApiV4Deserializer(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+    }
+
+    @Override
+    public JsonNode toCurrentStateNode(Api api) {
+        var node = (ObjectNode) objectMapper.valueToTree(PatchableView.from(api));
+        var apiV4 = ApiMapper.INSTANCE.mapToV4(api, null, null);
```
wbabyte, line 44: This `mapToV4` call triggers `ApiMapper.clearMetadataOnDirectMapping` (`@AfterMapping`, ApiMapper.java:217) which silently nulls `metadata` on the result. Harmless here — we only read listeners/endpointGroups/flows/resources off `apiV4`. But the `@AfterMapping` is a blanket hook firing on *every* `mapToV4(ApiV4)` callsite, and its Javadoc only mentions "create, import, single-get responses". Worth a follow-up to scope it via `@Named` qualifier on the specific mappings that should null metadata, instead of a global hook the contract enforces by docs only.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
carlos-andres-osorio: Added `isTargetedByJsonPatch` to `PatchApiUseCase.resolveField` — for JSON Patch, when no operation in the patch array targets `/<field>` or a sub-pointer of it, the method now returns the existing value immediately without going through the DTO roundtrip. Also added `json_patch_name_only_preserves_polymorphic_fields` in `ApiResource_PatchApiTest` which sends a `replace /name` patch against an API with listeners and endpoint groups and asserts both fields are identical to the original fixture after the patch.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapper.java
carlos-andres-osorio: Replaced the four `.filter(Objects::nonNull)` chains in `ApiMapper.mapToCoreApi` with a `rejectNulls` helper that throws `IllegalArgumentException` when a null element is encountered in any of the polymorphic lists. The helper identifies the field name in the exception message so the failure is unambiguous.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiEndpointGroupsTest.java
carlos-andres-osorio: Added the three AC3 preservation tests: `merge_patch_omitting_endpoint_groups_preserves_previous_endpoint_groups` (merge-omit), `merge_patch_endpoint_groups_null_clears_endpoint_groups` (merge-null), and `json_patch_remove_endpoint_groups_clears_endpoint_groups` (json-remove).

On the discriminator parameterization gap: endpoint group types are plugin IDs (`http-proxy`, `kafka`, etc.) rather than a `@JsonTypeInfo` discriminator enum — there is no uppercase/lowercase mismatch class equivalent to the listener types. The existing `http-proxy` test exercises the full round-trip for proxy APIs.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/use_case/PatchApiUseCaseTest.java
carlos-andres-osorio: `PatchApiV4Deserializer` lives in `gravitee-apim-rest-api-management-v2-rest`, which has a compile-time dependency on `gravitee-apim-rest-api-service`. Wiring it back into `PatchApiUseCaseTest` would create a circular module dependency. The seam between the use case and the real adapter is already exercised end-to-end by the `ApiResource_PatchApi*Test` suite, which runs the full stack including `PatchApiV4Deserializer`, `ApiMapper.mapToCoreApi`, and the `GenericApi` Jackson envelope.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/adapter/PatchApiV4Deserializer.java
carlos-andres-osorio: Agreed that scoping `clearMetadataOnDirectMapping` via a `@Named` qualifier would make the contract explicit rather than doc-only. Left as a follow-up — the hook is harmless in `toCurrentStateNode` since only `listeners`, `endpointGroups`, `flows`, and `resources` are read from the result and the rest is discarded.

wbabyte (approved):

PR #17087, opened 2026-05-25
fix(gamma): fix failing test on gamma console

Okhelifi:
## Issue

## Description


phiz71 (changes requested):
useless, @michel-barret already fixed it

PR #17118, opened 2026-05-26
feat(authz-ui): REST client 

podlesrafal:
## Description
  
  Adds the data layer for the Authorization UI in `gravitee-gamma-module-authz`
  (`src/main/ui/lib`): a typed REST client, a service/adapter layer over the
  canonical Gamma authz REST kernel, the entity adapter, a policy↔entity
  cross-reference scanner, and the React Query hooks the pages will consume.
  No pages/components yet — this is the foundation the feature UI builds on.

  **REST client** (`lib/api/authz-api-client.ts`)
  - `createAuthzApiClients` resolves the Gamma base URL from `/constants.json` +
    `/ui/bootstrap` and caches the in-flight promise so concurrently-mounting
    hooks share a single bootstrap round-trip.
  - Typed `get/post/put/delete`, `credentials: 'include'`, CSRF token echo via
    `X-Xsrf-Token`, `204`/empty-body handling, and an `ApiError` carrying the
    backend `ValidationErrorResponse`.

  **Service + adapter layer** (`lib/api/authz-api.service.ts`)
  - `authzApiService`: `getSchema`, `listPolicies`, `createPolicy`,
    `updatePolicy`, `deletePolicy`, `listEntities`.
  - Maps the canonical server domain (`Entity{entityId,kind,source}`,
    `Policy{kind: GLOBAL|RESOURCE, entityId}`) ↔ the richer UI shape
    (`uid`, `PolicyType: MCP|AGENT|LLM|API|EVENT|CUSTOM`, `target{id,label}`).
  - UI policy `type` is derived from the `entityId` prefix (`deriveServiceType`);
    status changes fan out to the dedicated `/deploy` and `/disable` endpoints.
  - Server-side paging (`?page/?perPage`) with a fallback that wraps a bare
    `List<T>` response into a paged envelope during rolling restarts.

  **Entity adapter** (`lib/entity-adapter.ts`)
  - Bidirectional `fromBackend` / `toBackend` between `EntityResponse` and the
    UI `EntityInstance`.
  - Canonical dotted uid (`user.alice`, SCIM 3-segment `user.okta.alice`)
    parsing/formatting that round-trips verbatim, `_`-prefixed meta-key
    separation, and parent parsing.

  **Policy ↔ entity refs** (`lib/policy-entity-refs.ts`)
  - Pure, lenient scanner that extracts `Type::"id"` references from GAPL
    an `O(1)` lookup map for "which policies reference this entity" — no backend
    cross-reference endpoint needed.
    
  **Hooks** (`lib/hooks/`, `@tanstack/react-query`)
  - `useEntities`, `usePolicies` (+ create/update/remove), `useSchema`
    (404 → `notFound`), `useEntityOptions` (entities → `ChipOption[]` for the
    policy-editor pickers). Shared cache keys in `lib/api/query-keys.ts`.


gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/shared/hooks/usePolicies.ts
```diff
@@ -0,0 +1,111 @@
...
+    });
+
+    const invalidate = useCallback(
+        () => queryClient.invalidateQueries({ queryKey: authzQueryKeys.policies.all(environmentId) }),
+        [environmentId, queryClient],
+    );
+
+    const reload = useCallback(() => void invalidate(), [invalidate]);
+
+    const create = useCallback(
```
jourdiw: **[Score: 72/100]** ⚠️ **Mutations bypass `useMutation` — architectural divergence from sibling modules.**

The `create`, `update`, and `remove` functions are plain `useCallback`-wrapped async calls that manually invalidate queries. The sibling APIM module (`usePlans.ts`) consistently uses react-query's `useMutation` hook for all write operations, which provides built-in loading/error/retry state, optimistic updates, and deduplication.

By bypassing `useMutation`, this hook:
- Loses per-mutation `isLoading` / `error` state (callers must track it themselves).
- Cannot benefit from `onMutate` / `onError` / `onSettled` for optimistic UI.
- Makes error handling inconsistent (query errors land in `query.error`, mutation errors are untracked).

**Suggestion:** Extract `create`, `update`, and `remove` into separate `useCreatePolicy`, `useUpdatePolicy`, and `useDeletePolicy` hooks using `useMutation`, mirroring the APIM module's pattern (`useCreatePlan`, `useUpdatePlan`, `useDeletePlan`).

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/shared/hooks/useEntityOptions.ts
```diff
@@ -0,0 +1,86 @@
...
+    });
+    return parts.join(', ');
+}
+
+function toChipOption(entity: EntityResponse): ChipOption {
+    const { type, id } = parseEntityUid(entity.uid);
+    return {
+        id: `${type}::"${id}"`,
+        label: id,
+        group: type,
```
jourdiw, line 57: **[Score: 55/100]** 💡 **`typeFilterKey` in the query key causes redundant network fetches for the same data.**

Including `typeFilterKey` in the query key means each distinct type filter value creates a separate cache entry and triggers a new network request — even though the `queryFn` always fetches the same `listEntities(env, { page: 1, perPage: 200 })` payload.

If you request `typeFilter: ['User']`, then later `typeFilter: ['Group']`, you'll make two identical HTTP calls to `/entities?page=1&perPage=200`.

**Suggestion:** Remove `typeFilterKey` from the query key and use react-query's `select` option to filter the cached data in-memory. This way one fetch serves all filter permutations:

```ts
const query = useQuery({
  queryKey: authzQueryKeys.entityOptions(environmentId),
  queryFn: () => authzApiService.listEntities(environmentId, { page: 1, perPage: PER_PAGE }),
  select: (data) => {
    const all = data.data.map(toChipOption);
    if (!typeFilter?.length) return all;
    const allow = new Set(typeFilter);
    return all.filter(o => allow.has(o.group));
  },
  staleTime: 30_000,
});
```

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/shared/hooks/useEntities.ts
```diff
@@ -0,0 +1,69 @@
...
+        queryKey: authzQueryKeys.entities.page(environmentId, page, perPage),
+        queryFn: () => authzApiService.listEntities(environmentId, { page, perPage }),
+        staleTime: 30_000,
+        placeholderData: keepPreviousData,
+    });
+
+    const reload = useCallback(
+        () => void queryClient.invalidateQueries({ queryKey: authzQueryKeys.entities.all(environmentId) }),
+        [environmentId, queryClient],
+    );
```
jourdiw, line 58: **[Score: 45/100]** 💡 **Missing `enabled` guard — queries fire immediately even with an empty `environmentId`.**

The sibling APIM hooks consistently guard queries with `enabled: Boolean(env)` or `enabled: Boolean(envId && ...)` to prevent fetching when the environment context hasn't loaded yet. Both `useEntities` and `usePolicies` in this PR always fire their queries, which will produce invalid API paths (e.g. `/environments//modules/authz/entities`) if `environmentId` is an empty string.

**Suggestion:** Add `enabled: Boolean(environmentId)` to the query options in both `useEntities` and `usePolicies`, matching the APIM pattern.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/shared/entity-adapter.ts
```diff
@@ -0,0 +1,182 @@
...
+ * Packs meta fields back into `attributes` under reserved underscore keys.
+ */
+export function toBackend(e: EntityInstance): EntityRequest {
+    const uid = formatEntityUid(e.uid);
+
+    const attributes: Record<string, unknown> = { ...e.attrs };
+
+    if (e.source && e.source !== 'local') {
+        attributes['_source'] = e.source;
+    }
```
jourdiw, line 138: **[Score: 40/100]** 💡 **`toBackend` does not write `_kind` back into attributes — potential round-trip data loss.**

`fromBackend` reads and consumes the `_kind` attribute (line 118), using it to resolve the entity type. But `toBackend` never writes `_kind` back into the outgoing `attributes` map. If the backend relies on `_kind` being present in PUT payloads (e.g. for re-indexing or display), a round-trip through the UI will silently drop it.

The tests confirm this: `toBackend` tests don't assert on `_kind` at all, and the `round-trips local user` test only checks `uid`, `name`, and `_displayName`.

**Suggestion:** Either write `_kind` back in `toBackend`, or add a test explicitly documenting that omitting `_kind` is intentional (with a comment explaining the backend doesn't need it on writes).

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/shared/hooks/useEntityOptions.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { useQuery } from '@tanstack/react-query';
+import { useMemo } from 'react';
+import { authzApiService } from '../api/authz-api.service';
+import type { EntityResponse } from '../api/authz-api.types';
```
jourdiw: **[Score: 50/100]** 💡 **Hard ceiling of 200 entities with a user-facing error message instead of pagination — scaling concern.**

`PER_PAGE = 200` caps the entity picker at 200 items. When exceeded, the hook sets a non-dismissible error string (`'Too many entities for chip picker…'`) rather than loading more pages or offering search/filter at the API level.

For a growing authorization system (users + groups + service accounts + MCP servers + LLM routes + agents), 200 entities is a plausible ceiling to hit in production.

**Suggestion:** Consider either:
1. Implementing cursor-based or offset pagination that loads additional pages in the background.
2. Adding server-side search/filter to the entity list endpoint so the picker can fetch matching subsets.
3. At minimum, making `PER_PAGE` configurable via the `opts` parameter so callers can tune it.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/shared/api/authz-api.service.ts
```diff
@@ -0,0 +1,299 @@
...
+    readonly policyText: string;
+}
+
+interface CanonicalUpdatePolicyRequest {
+    readonly name: string;
+    readonly policyText: string;
+}
+
+// Authz is mounted at /gamma/organizations/{org}/environments/{env}/modules/authz
+// by GammaModulesResource (SPI). The client base URL already covers the
```
jourdiw, line 93: **[Score: 48/100]** 💡 **Client-side array fallback in `adaptPagedListResponse` adds hidden complexity and inaccurate totals.**

This function accepts either a paginated response or a raw array, performing client-side slicing in the array case. This defensive coding is understandable for a transitional backend, but it introduces subtle issues:

1. **The `total` from an array-fallback is the array length**, not the server's full dataset size. If the server returns all items as an array, the UI won't know there could be more.
2. **Client-side pagination on a server response is a code smell** — it masks a missing server feature and will silently break if the server starts sending partial arrays.
3. Combined with the post-filter for `params?.type` in `listPolicies` (line 155), the `total` count becomes doubly approximate.

**Suggestion:** If the backend doesn't support pagination yet, document this explicitly in a TODO and add a runtime warning (dev-only) when the array fallback is triggered, so it doesn't become permanent.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/shared/api/authz-api.service.ts
```diff
@@ -0,0 +1,299 @@
...
+interface CanonicalEntity {
+    readonly id: string;
+    readonly entityId: string;
+    readonly kind: 'PRINCIPAL' | 'RESOURCE';
+    readonly attributes: Record<string, unknown>;
+    readonly parents: readonly string[];
+    readonly source: string;
+    readonly environmentId: string;
+    readonly createdAt: string;
+    readonly updatedAt: string;
```
jourdiw, line 54: **[Score: 30/100]** 🎨 **`deriveServiceType` hardcodes entity-prefix-to-type mapping — consider making it data-driven.**

This switch statement duplicates knowledge that also lives in `entity-adapter.ts`'s `kindToType` / `typeToKind` functions. If a new entity kind is added (e.g. `webhook`), three separate switch/if-chains need updating.

**Suggestion:** Consider extracting a single canonical `ENTITY_KIND_REGISTRY` (a `Map` or object) that both the entity adapter and the service layer reference. This keeps the mapping in one place and makes it trivial to add new kinds.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/shared/hooks/useEntityOptions.ts
```diff
@@ -0,0 +1,114 @@
...
+        for (const page of pages) {
+            if (!page) continue;
+            for (const e of page.data) merged.push(toChipOption(e));
+        }
+        return merged;
+    }, [pages]);
+
+    const tooMany = pages.some(p => (p?.total ?? 0) > PER_PAGE);
+    const error =
+        networkError ?? (tooMany ? 'Too many entities for chip picker; consider refining schema or filtering by type.' : undefined);
```
jourdiw: WDYT about something less technical? 
"Chip picker" is technical UI vocab. What are the chips truly in this case?

jourdiw (approved):
LGTM

PR #17121, opened 2026-05-26
feat: paginate ApiKey sync queries with searchAfter cursor (APIM-14203)

wbabyte:
## Summary

Mirrors the APIM-14087 cursor-pagination pattern for `ApiKey`, addressing the same memory + scale problems that hit subscriptions at customer projection (25M subs → comparable ApiKey volume on api-key plans).

- **Repository SPI**: new `ApiKeyCursor` record + abstract `searchAfter(criteria, sortable, after, pageSize)` on `ApiKeyRepository`. Sort dispatch accepts `"updatedAt"` (delta sync `(updatedAt, id)` keyset) or `"id"` (warmup id-only keyset); any other field throws `TechnicalException`. ASC enforced silently — DESC is ignored, not rejected.
- **MongoDB**: internal `ApiKeyMongoRepositoryImpl.searchAfter` uses Spring Data `Criteria` with `OR(updatedAt>marker, AND(updatedAt=marker, _id>id))` for delta, `_id>id` for warmup. New `KeysEnvironmentIdUpdatedAtIdIndexUpgrader` adds compound index `e1ua1i1 {environmentId, updatedAt, _id}` on the `keys` collection.
- **JDBC**: subquery + outer `LEFT JOIN key_subscriptions` pattern with semi-join inside the inner subquery — applies `LIMIT` to the dedup-stable key set BEFORE the M:N hydrate (regression-tested via `searchAfter_shouldReturnCompleteSubscriptionsListAndNotShortPageDueToJoin`). New Liquibase changeset `v4_11_8/02_add_keys_environment_id_updated_at_id_index.yml` adds the analogous index. All existing indexes left untouched (additive-only).
- **NoOp**: returns empty list.
- **Gateway sync**:
  - `ApiKeyFetcher.fetchLatest` migrated to `Flowable.generate(ApiKeyPage)` cursor loop with null-`updatedAt` poison-row guard. The guard is deliberate: `criteria.from/to > 0` already excludes null-updatedAt rows at the repository layer, so the guard only fires on misconfigured indexes or legacy data — and when it does, it lets the sync loop survive rather than wedge on a single bad row.
  - `ApiKeyAppender.loadApiKey` chunks `subscriptions IN (...)` at default 900, paginates per chunk via id-only cursor, dedups federated keys across chunks via `Set<String> seenKeyIds`. Subscription ids are sorted before chunking so chunk boundaries are deterministic across JVM upgrades (helps reproduce customer incidents).
  - `ApiProductSubscriptionRefresher.loadApiKeys` gets the same chunk + cursor + dedup migration (symptomatically equivalent code path; extracted to shared helper on a future third caller).
  - New config property `services.sync.subscriptions_chunk_size` (default 900); existing `services.sync.bulk_items` (default 100) reused for cursor page size.

## Risk + follow-ups

- **Bridge release coordination**: this PR ships the abstract SPI method but does NOT bump the `gravitee-apim-repository-bridge` version. The bridge repo needs a parallel PR implementing `searchAfter` before bridge 8.3.0 is released; a follow-up chore commit will bump the dep here (same shape as `ac718d3316` for APIM-14087). Window between this merge and the dep bump: HTTP-bridge gateways that take a build during that gap will hit `AbstractMethodError` on first sync tick — release-timing coordination required, same operating pattern as 14087.
- **Warmup query plan unprofiled**: only the delta-path index `e1ua1i1` is added. The warmup id-only keyset with `subscriptions IN` filter falls back to the default `_id` index with the IN as a residual predicate — same gap APIM-14087 left for the subscription warmup. Pre-emptive index additions would speculate without `$indexStats` evidence; addressing both warmup paths consistently is the right shape for a separate audit, not a partial fix here.
- **Out of scope**: no `Cursor<T>` generalization (parallel `SubscriptionCursor` / `ApiKeyCursor` records — refactor on third caller, not second); `includeFederated` semantics preserved unchanged; `TIMEFRAME_DELAY` preserved unchanged.

## Test plan

- [x] Module tests green
  - JDBC TCK `ApiKeyRepositoryTest`: 46/46 passing (9 new `searchAfter_*` tests + 37 existing, some adjusted with explicit `environments` scoping so future fixture additions don't pollute global expire-range queries)
  - Mongo TCK `ApiKeyRepositoryTest`: 46/46 passing + new `EnvironmentIdUpdatedAtIdIndexUpgraderTest` for the keys variant
  - Gateway sync module: 366/366 passing (5 new `ApiKeyFetcherTest`, 4 new `ApiKeyAppenderTest` with chunk + dedup coverage, refresher tests including chunk+dedup regression)
- [x] Smoke test via `apim-worktree APIM-14203` (real MongoDB + ES + mAPI + gateway):
  - Create v4 API with apikey plan + auto-deploy → `200`
  - Create application + subscription → ApiKey issued
  - `GET /smoke14203v2` with `X-Gravitee-Api-Key` → `200` (gateway-sync picked up the ApiKey via the new `searchAfter` cursor path on the live MongoDB)
  - `GET /smoke14203v2` without key → `401`
  - No `AbstractMethodError` / sync errors in gateway or mAPI logs
- [ ] CI pipeline green on the PR
- [ ] Bridge follow-up PR + 8.3.0 release coordinated before dep bump merges here

## Reference

APIM-14087 (PR #16968) established the cursor-pagination pattern for subscriptions; this PR mirrors it for ApiKey.

Jira: https://gravitee.atlassian.net/browse/APIM-14203

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/SyncConfiguration.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/SyncConfiguration.java
@@ -77,6 +77,10 @@ public class SyncConfiguration {
 
     public static final int POOL_SIZE = Runtime.getRuntime().availableProcessors() * 2;
     public static final int DEFAULT_BULK_ITEMS = 100;
+    // Chunk size for the `subscriptions IN (...)` filter applied when loading ApiKeys (appender
+    // + refresher). Conservative default below common DB IN-list parameter limits; tune higher via
+    // services.sync.apikey.subscriptions_chunk_size only after validating the target DB's parameter ceiling.
+    public static final int DEFAULT_APIKEY_SUBSCRIPTIONS_CHUNK_SIZE = 900;
 
     @Bean("syncFetcherExecutor")
     public ThreadPoolExecutor syncFetcherExecutor(@Value("${services.sync.fetcher:-1}") int syncFetcher) {
@@ -174,8 +178,15 @@ public class SyncConfiguration {
     }
 
     @Bean
-    public ApiKeyAppender apiKeyAppender(ApiKeyRepository apiKeyRepository, ApiKeyMapper apiKeyMapper) {
-        return new ApiKeyAppender(apiKeyRepository, apiKeyMapper);
+    public ApiKeyAppender apiKeyAppender(
+        ApiKeyRepository apiKeyRepository,
+        ApiKeyMapper apiKeyMapper,
+        @Value("${services.sync.bulk_items:" + DEFAULT_BULK_ITEMS + "}") int bulkItems,
+        @Value(
+            "${services.sync.apikey.subscriptions_chunk_size:" + DEFAULT_APIKEY_SUBSCRIPTIONS_CHUNK_SIZE + "}"
+        ) int subscriptionsChunkSize
+    ) {
+        return new ApiKeyAppender(apiKeyRepository, apiKeyMapper, bulkItems, subscriptionsChunkSize);
     }
 
     @Bean
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductSubscriptionRefresher.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductSubscriptionRefresher.java
@@ -30,20 +30,21 @@ import io.gravitee.gateway.services.sync.process.repository.mapper.ApiKeyMapper;
 import io.gravitee.repository.management.api.ApiKeyRepository;
 import io.gravitee.repository.management.api.SubscriptionRepository;
 import io.gravitee.repository.management.api.search.ApiKeyCriteria;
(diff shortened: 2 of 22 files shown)


phiz71 (approved):
LGTM and I assume it replicates what have already been approved on subscriptions

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/SyncConfiguration.java
```diff
@@ -174,8 +177,13 @@ public SubscriptionAppender subscriptionAppender(
     }
 
     @Bean
-    public ApiKeyAppender apiKeyAppender(ApiKeyRepository apiKeyRepository, ApiKeyMapper apiKeyMapper) {
-        return new ApiKeyAppender(apiKeyRepository, apiKeyMapper);
+    public ApiKeyAppender apiKeyAppender(
+        ApiKeyRepository apiKeyRepository,
+        ApiKeyMapper apiKeyMapper,
+        @Value("${services.sync.bulk_items:" + DEFAULT_BULK_ITEMS + "}") int bulkItems,
+        @Value("${services.sync.subscriptions_chunk_size:" + DEFAULT_SUBSCRIPTIONS_CHUNK_SIZE + "}") int subscriptionsChunkSize
```
phiz71: is it on purpose that you reuse the same property ? should we have a dedicated "chunk size" for apikey ?

wbabyte: good spot, renamed to services.sync.apikey.subscriptions_chunk_size

phiz71 (approved):

PR #17128, opened 2026-05-26
feat(gamma-apim): integrate policy studio page with V2 compatibility

jourdiw:
### Context sidebar adoption

Migrates both `ApiDetailLayout` and `ApiProductDetailLayout` from hand-built `<aside>`/`<main>` flex layouts to graphene-core's `ContextSidebar` + `useLayoutConfig` pattern. The shell (`ShellLayout`, `LocalDevShell`) is updated to pass through the new layout slot properties (`viewMode`, `contextExpanded`, `contextSidebar`, `contentVariant`, `leading`).

**API detail (`ApiDetailLayout`):**
- Uses `ContextSidebar` with `ApiInfoHeader` (avatar, name, state badge, description, tech chips) as the header and `ApiDetailSidebarNav` as the navigation body.
- `ContextToggleButton` in the `leading` slot allows collapsing/expanding the sidebar.
- Breadcrumbs: `API Proxies > {api name}`.

**API Products (`ApiProductDetailLayout`):**
- Same pattern: `ContextSidebar` with `ProductInfoHeader` and `ApiProductSidebarNav`.
- Removes the old `<aside>` + manual divider + `<main>` with hardcoded `height: calc(100dvh - 5rem)`.
- Breadcrumbs: `API Products > {product name}`.

### Deploy banner redesign

Replaces the bulky padded `Alert`-based deploy banner with a compact single-line status strip (~40px vs ~80px):
- Uses a `border-b` flex row with a subtle `color-mix(in oklab, var(--color-warning) 6%, transparent)` background tint.
- Text uses `text-foreground` for contrast (WCAG-compliant), with `text-warning` on the icon only.
- `TriangleAlertIcon` replaces `MessageSquareWarningIcon`.
- Copy changed from "Pending changes / This API is out of sync" to "This API has undeployed changes."
- Removed unused `Alert`, `AlertAction`, `AlertDescription`, `AlertTitle` imports.
- Test assertions updated to match the new copy; unused Alert mocks removed from the spec file.
- Filed [gravitee-ui-graphene#346](https://github.com/gravitee-io/gravitee-ui-graphene/issues/346) to add a `banner` layout slot to `AppLayout` for full-width rendering in a future follow-up.

Screenshot 2026-05-28 at 20 43 42

### Policy Studio integration

Replaces the Policy Studio placeholder route with a full integration of `@gravitee/graphene-policy-studio`:
- Service layer, React Query hooks, type definitions, and a V2-to-V4 flow adapter for backward compatibility.
- V2 APIs render read-only; V4/native APIs support full editing.
- Uses `contentVariant: 'full-bleed'` for edge-to-edge rendering.

Screenshot 2026-05-28 at 20 43 55

### Shared API client

Adds `apimFetchJsonV2Org` to the shared APIM client for org-scoped v2 endpoint calls required by the policy/plugin catalog.

### Merge fix

Resolves duplicate JSX props (`viewMode`, `contextExpanded`, `contextSidebar`, `contentVariant`, `leading`) left by a merge conflict in `ShellLayout` and `LocalDevShell`.

## Notes

- Error handling/toast feedback in the save hook (`usePolicyStudioSave`) is deferred to a follow-up.
- The deploy banner renders inside the content padding area for now. Full-width rendering will land once the `banner` slot is added to graphene-core ([#346](https://github.com/gravitee-io/gravitee-ui-graphene/issues/346)).

## Test plan

- [x] Navigate to an existing V4 API -> Policy Studio — verify the studio loads with correct flows, plans, policies, and connectors
- [x] Edit a flow and save — verify the GET-then-PUT save roundtrip works and the UI reflects the saved state
- [x] Navigate to a V2 API -> Policy Studio — verify the studio renders in read-only mode with V2 flows adapted to V4 shape
- [x] Collapse/expand the context sidebar via the toggle button on both API detail and API Product detail — verify layout behavior
- [x] Verify the deploy banner appears when an API has `deploymentState: NEED_REDEPLOY` and the user has `api-definition-u` permission
- [x] Click "Deploy API" on the banner — verify the deploy mutation fires and the banner disappears after success
- [x] Navigate to an API Product detail — verify the context sidebar renders with product info header, sync status badge, and nav links
- [x] Verify plugin catalog (policies, entrypoints, endpoints, shared policy groups) loads correctly

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/detail/ApiDetailLayout.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/detail/ApiDetailLayout.tsx
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 import { useEnvironment, useHasPermission } from '@gravitee/gamma-modules-sdk';
-import { Badge, Button, Card, CardContent, Skeleton } from '@gravitee/graphene-core';
-import { CircleCheckIcon, CircleStopIcon, CircleXIcon, MessageSquareWarningIcon, TriangleAlertIcon } from '@gravitee/graphene-core/icons';
+import { Badge, Button, ContextSidebar, ContextToggleButton, Skeleton, useLayoutConfig } from '@gravitee/graphene-core';
+import { CircleCheckIcon, CircleStopIcon, CircleXIcon, TriangleAlertIcon } from '@gravitee/graphene-core/icons';
 import { useMutation, useQueryClient } from '@tanstack/react-query';
 import { useState } from 'react';
 import { Navigate, Outlet, useParams } from 'react-router-dom';
@@ -78,19 +78,14 @@ function ApiAvatar({ api }: { api: ApiDetailDto }) {
 
 function DeployBanner({ onDeploy, isPending }: { onDeploy: () => void; isPending: boolean }) {
     return (
-        <div className="px-6 pt-4">
-            <Card>
-                <CardContent className="flex items-center justify-between gap-3 py-3">
-                    <div className="flex items-center gap-2">
-                        <MessageSquareWarningIcon className="size-4 text-warning shrink-0" />
-                        <p className="text-sm font-medium">Pending changes</p>
-                        <p className="text-sm text-muted-foreground">— This API is out of sync.</p>
-                    </div>
-                    <Button size="sm" onClick={onDeploy} disabled={isPending} className="shrink-0">
-                        {isPending ? 'Deploying…' : 'Deploy API'}
-                    </Button>
-                </CardContent>
-            </Card>
+        <div className="flex items-center justify-between border-b px-6 py-2 bg-warning/10">
+            <span className="flex items-center gap-1.5 text-sm text-foreground">
+                <TriangleAlertIcon className="size-3.5 shrink-0 text-warning" />
+                This API has undeployed changes.
+            </span>
+            <Button size="sm" variant="outline" onClick={onDeploy} disabled={isPending}>
+                {isPending ? 'Deploying…' : 'Deploy API'}
+            </Button>
         </div>
     );
 }
@@ -174,6 +169,7 @@ export function ApiDetailLayout() {
     const { permissionsReady } = useApiPermissions(apiId);
     const canDeploy = useHasPermission({ anyOf: ['api-definition-u'] });
     const queryClient = useQueryClient();
+    const [contextExpanded, setContextExpanded] = useState(true);
 
     const deployMutation = useMutation({
         mutationFn: () => {
@@ -187,6 +183,24 @@ export function ApiDetailLayout() {
         },
     });
 
+    useLayoutConfig(
+        {
+            viewMode: 'context',
+            contextExpanded,
+            contextSidebar: (
+                <ContextSidebar header={<ApiInfoHeader api={api ?? null} isLoading={isLoading} />}>
+                    <ApiDetailSidebarNav groups={API_PROXY_NAV_GROUPS} basePath={basePath} permissionsReady={permissionsReady} />
+                </ContextSidebar>
+            ),
+            leading: <ContextToggleButton expanded={contextExpanded} onToggle={() => setContextExpanded(v => !v)} />,
+            breadcrumbs: [
+                { label: 'API Proxies', href: `${basePath.slice(0, basePath.lastIndexOf('/apis/'))}${'/apis'}` },
+                { label: api?.name ?? 'Loading…' },
+            ],
+        },
+        [contextExpanded, api, isLoading, basePath, permissionsReady],
+    );
+
     if (isError) {
         return (
             <div className="flex items-center justify-center p-8">
@@ -199,16 +213,11 @@ export function ApiDetailLayout() {
 
     return (
         <ApiDetailContext.Provider value={{ api: api ?? null, isLoading, permissionsReady }}>
-            <div className="flex" style={{ height: 'calc(100dvh - 5rem)' }}>
-                <aside className="w-56 min-w-0 shrink-0 overflow-y-auto overflow-x-hidden pb-4" style={{ maxWidth: '14rem' }}>
-                    <ApiInfoHeader api={api ?? null} isLoading={isLoading} />
-                    <ApiDetailSidebarNav groups={API_PROXY_NAV_GROUPS} basePath={basePath} permissionsReady={permissionsReady} />
-                </aside>
-                <div className="w-px bg-border shrink-0" />
-                <main className="min-w-0 flex-1 overflow-y-auto">
-                    {showDeployBanner && <DeployBanner onDeploy={() => deployMutation.mutate()} isPending={deployMutation.isPending} />}
+            <div className="flex h-full min-h-0 flex-col">
+                {showDeployBanner && <DeployBanner onDeploy={() => deployMutation.mutate()} isPending={deployMutation.isPending} />}
+                <div className="min-h-0 flex-1">
                     <Outlet />
-                </main>
+                </div>
             </div>
         </ApiDetailContext.Provider>
     );
--- a/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/hooks/usePolicyStudioData.ts
+++ b/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/hooks/usePolicyStudioData.ts
@@ -1,0 +1,236 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { useEnvironment } from '@gravitee/gamma-modules-sdk';
+import type { ApiType as PSApiType, ConnectorInfo, Flow, Plan, Policy, SharedPolicyGroupPolicy } from '@gravitee/graphene-policy-studio';
+import { useQuery } from '@tanstack/react-query';
+import { useMemo } from 'react';
+
+import {
+    getFullApiDetail,
+    listEndpointPlugins,
+    listEntrypointPlugins,
+    listPolicies,
+    listPublishedPlans,
+    listSharedPolicyGroupPlugins,
+} from '../services/policyStudioService';
+import type { FlowExecution, FlowV2, PolicyStudioApiDetail } from '../types/policyStudio';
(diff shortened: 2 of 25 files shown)


gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/pages/policy-studio/PolicyStudioPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+    );
+
+    if (studioData.isLoading) {
+        return (
+            <div className="flex flex-col gap-4 p-6">
+                <Skeleton className="h-8 w-48 rounded" />
+                <Skeleton className="h-64 w-full rounded" />
+            </div>
+        );
+    }
```
jourdiw: The studio handles loading itself so this can be removed

jourdiw (commented):
Not all of the services are strict to the policy studio as well as the types. They should all be added rather to the services/ folder.

The types should then as well be in the types/ folder, not under the feature unless they are specific to the UI.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/hooks/usePolicyStudioSave.ts
```diff
@@ -0,0 +1,60 @@
...
+            const envId = env.id;
+
+            const { commonFlows, flowExecution, plansToUpdate } = output;
+
+            if (commonFlows || flowExecution) {
+                const current = await getFullApiDetail(envId, apiId);
+                const updated: Record<string, unknown> = { ...current };
+                if (commonFlows) updated.flows = commonFlows;
+                if (flowExecution) updated.flowExecution = flowExecution;
+                await updateApi(envId, apiId, updated);
```
jourdiw: **[Score: 72/100]** ⚠️ The `onSave` callback has no error handling. If `updateApi` or `getAndUpdatePlanFlows` rejects (network error, 409 conflict, validation failure), the promise propagates to `PolicyStudio`'s `onSave` — but it's unclear whether that component surfaces the error to the user or swallows it. At minimum, consider wrapping in a try/catch that shows a toast or alert on failure:

```ts
try {
    // ... existing logic
} catch (err) {
    // Surface error to user (toast, alert, etc.)
    throw err; // re-throw so PolicyStudio knows save failed
}
```

Also: spreading the full API detail (`{ ...current }`) and PUTting it back could send computed/read-only fields (`_links`, `definitionContext`, etc.) that the backend might reject or ignore. Consider filtering to only writable fields before the PUT.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/utils/v2FlowAdapter.ts
```diff
@@ -0,0 +1,103 @@
...
+    };
+}
+
+/**
+ * Converts a V2 flow to the V4 Flow shape expected by `@gravitee/graphene-policy-studio`.
+ *
+ * V2 uses `pathOperator` + `methods` + `pre`/`post`.
+ * V4 uses `selectors` + `request`/`response`.
+ */
+export function mapFlowV2ToV4(flow: FlowV2): Flow {
```
jourdiw, line 40: **[Score: 75/100]** ⚠️ The V2→V4 flow adapter is non-trivial mapping logic (path operators, condition selectors, step conversion) with **zero test coverage**. This is the kind of code that benefits most from unit tests because:

1. V2 APIs can have edge cases (empty `pathOperator`, flows with `pre` but no `post`, disabled flows).
2. The adapter will be exercised in read-only mode for legacy APIs — bugs here silently corrupt the visual display.
3. The `as Selector` type assertions mask potential shape mismatches.

A small `v2FlowAdapter.test.ts` with 5-6 cases (empty flow, single step, multi-step with condition, no pathOperator) would provide high confidence cheaply.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/detail/ApiDetailLayout.tsx
```diff
@@ -187,6 +183,24 @@ export function ApiDetailLayout() {
...
+            viewMode: 'context',
+            contextExpanded,
+            contextSidebar: (
+                <ContextSidebar header={<ApiInfoHeader api={api ?? null} isLoading={isLoading} />}>
+                    <ApiDetailSidebarNav groups={API_PROXY_NAV_GROUPS} basePath={basePath} permissionsReady={permissionsReady} />
+                </ContextSidebar>
+            ),
+            leading: <ContextToggleButton expanded={contextExpanded} onToggle={() => setContextExpanded(v => !v)} />,
+            breadcrumbs: [
+                { label: 'API Proxies', href: `${basePath.slice(0, basePath.lastIndexOf('/apis/'))}${'/apis'}` },
```
jourdiw, line 197: **[Score: 55/100]** 💡 Breadcrumb `href` is built via `basePath.slice(0, basePath.lastIndexOf('/apis/'))` — this is fragile string surgery that silently breaks if the route structure ever changes (e.g. renamed segment, extra nesting). Same pattern in `ApiProductDetailLayout.tsx`.

Consider deriving it from the router's own path resolution (React Router's `useResolvedPath('..')` or the already-available `rootPath` from `useModuleRouting`) rather than manual string slicing. Alternatively, pass the list-page href down via context or a config constant.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/services/policyStudioService.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export async function getFullApiDetail(envId: string, apiId: string): Promise<PolicyStudioApiDetail> {
+    return apimFetchJsonV2<PolicyStudioApiDetail>(envId, `/apis/${encodeURIComponent(apiId)}`);
+}
+
+export async function listPublishedPlans(envId: string, apiId: string): Promise<PolicyStudioPlanPage> {
+    return apimFetchJsonV2<PolicyStudioPlanPage>(
+        envId,
+        `/apis/${encodeURIComponent(apiId)}/plans?statuses=PUBLISHED,DEPRECATED&page=1&perPage=9999`,
```
jourdiw, line 64: **[Score: 30/100]** 🎨 `perPage=9999` is a pragmatic escape hatch for loading all published plans at once. Worth a code comment documenting why the policy studio needs all plans upfront (to render all plan flows in the visual editor) so future readers don't try to "fix" it with proper pagination.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/hooks/usePolicyStudioData.ts
```diff
@@ -0,0 +1,236 @@
...
+}
+
+export function usePolicyStudioData(apiId: string | undefined): PolicyStudioData {
+    const env = useEnvironment();
+    const envId = env?.id ?? '';
+    const enabled = Boolean(env && apiId);
+
+    const apiQuery = useQuery({
+        queryKey: policyStudioKeys.api(envId, apiId ?? ''),
+        queryFn: () => getFullApiDetail(envId, apiId!),
```
jourdiw, line 104: **[Score: 35/100]** 🎨 The `apiQuery` uses `queryFn: () => getFullApiDetail(envId, apiId!)` with a non-null assertion. The `enabled` guard ensures `apiId` is truthy before the query fires, so this is safe at runtime — but the assertion hides a type constraint from the compiler. Consider narrowing the type in the `enabled` check and using the conditional `select` pattern, or simply providing a fallback:

```ts
queryFn: () => getFullApiDetail(envId, apiId ?? ''),
```

(The function already `encodeURIComponent`s the id, so an empty string would produce a 404 rather than a crash — though the `enabled` flag prevents this case anyway.)

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/utils/v2FlowAdapter.ts
jourdiw: Yes implement test cases

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/services/policyStudioService.ts
jourdiw: We are obliged to do this for now.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/pages/policy-studio/PolicyStudioPage.tsx
jourdiw: Done — removed the page-level loading skeleton. `PolicyStudio` handles its own loading state internally.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/hooks/usePolicyStudioSave.ts
jourdiw: Added try/catch that logs and re-throws so PolicyStudio still receives the rejection. Did not add toast yet (no shared toast pattern in this module). Skipped filtering writable fields — GET-then-PUT with full body matches existing module pattern (`getAndUpdatePlanFlows` does the same) and the backend ignores read-only fields.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/utils/v2FlowAdapter.ts
jourdiw: Done — added `v2FlowAdapter.spec.ts` with 14 test cases covering pathOperator, condition, combined selectors, empty pre/post, disabled flows, mapFlowsV2ToV4, mapFlowModeToExecution, and the default connector constants.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/detail/ApiDetailLayout.tsx
jourdiw: Agree it's fragile, but `useDetailBasePath` already uses the same pathname-slicing approach and `ApiProductDetailLayout` uses the identical breadcrumb pattern. Refactoring to `useResolvedPath` or module routing would be a separate cross-layout change — happy to track as follow-up.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/services/policyStudioService.ts
jourdiw: Added an inline comment explaining Policy Studio needs all published plans in one request. `perPage=9999` is intentional until the studio supports paginated plan loading.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/hooks/usePolicyStudioData.ts
jourdiw: Keeping `apiId!` — the `enabled` guard guarantees it is defined before `queryFn` runs. This is the standard TanStack Query pattern documented in their guides. `apiId ?? ''` would be a misleading fallback that never actually executes.

jourdiw:
Re: services/types placement — they are already under `features/apis/services/` and `features/apis/types/`, consistent with other feature services (`plans.ts`, `apis.ts`, etc.). Moving the shared-capable calls (e.g. `listPolicies`, `listEntrypointPlugins`) to `shared/` is a valid follow-up but out of scope for this integration PR. Will track separately.

Okhelifi (approved):

PR #17135, opened 2026-05-26
feat(rest-api): emit ETag and Last-Modified on GET plan

carlos-andres-osorio:
## Issue

[APIM-14081](https://gravitee.atlassian.net/browse/APIM-14081)

## Why

The PATCH plan endpoint requires callers to supply an `If-Match` header for concurrency control. Without ETag emission on GET, clients have no way to obtain a valid precondition value before making mutating requests. Adding `ETag` and `Last-Modified` to the GET response unblocks the full conditional-request workflow for plan consumers.

## What changed

- `ApiPlansResource#getApiPlan`: Added `.tag(...)` and `.lastModified(...)` on the 200-path response builder, populated via a new private `resolvePlanUpdatedAt` helper. The helper pattern-matches over `BasePlanEntity` subtypes (v4 and v2) to locate `updatedAt`, failing open (omitting both headers) when the value is null.
- `openapi-apis.yaml`: Documented the `ETag` and `Last-Modified` response headers on the GET plan 200 response.

## User-facing impact

`GET /plans/{planId}` now returns two new response headers:
- `ETag`: the plan's last-update timestamp as an epoch-millis string.
- `Last-Modified`: the same timestamp in RFC 1123 format.

This is a backwards-compatible addition — existing clients that ignore unknown response headers are unaffected.

[APIM-14081]: https://gravitee.atlassian.net/browse/APIM-14081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiPlansResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiPlansResource.java
@@ -74,6 +74,7 @@ import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.core.Response;
+import java.util.Date;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -241,7 +242,22 @@ public class ApiPlansResource extends AbstractResource {
             return Response.status(Response.Status.NOT_FOUND).entity(planNotFoundError(planId)).build();
         }
 
-        return Response.ok(planMapper.mapGenericPlan(planEntity)).build();
+        Response.ResponseBuilder builder = Response.ok(planMapper.mapGenericPlan(planEntity));
+        Date updatedAt = resolvePlanUpdatedAt(planEntity);
+        if (updatedAt != null) {
+            builder.tag(Long.toString(updatedAt.getTime())).lastModified(updatedAt);
+        }
+        return builder.build();
+    }
+
+    private static Date resolvePlanUpdatedAt(GenericPlanEntity planEntity) {
+        if (planEntity instanceof BasePlanEntity v4Plan) {
+            return v4Plan.getUpdatedAt();
+        }
+        if (planEntity instanceof io.gravitee.rest.api.model.BasePlanEntity v2Plan) {
+            return v2Plan.getUpdatedAt();
+        }
+        throw new IllegalStateException("Unexpected plan entity type: " + planEntity.getClass());
     }
 
     @PUT
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiPlansResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiPlansResource.java
@@ -26,6 +26,7 @@ import io.gravitee.apim.core.subscription.model.SubscriptionEntity;
 import io.gravitee.apim.core.subscription.query_service.SubscriptionQueryService;
 import io.gravitee.common.http.MediaType;
 import io.gravitee.definition.model.v4.plan.PlanMode;
+import io.gravitee.node.logging.NodeLoggerFactory;
 import io.gravitee.repository.management.model.Group;
 import io.gravitee.rest.api.management.v2.rest.mapper.FlowMapper;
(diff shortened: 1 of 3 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiPlansResource.java
```diff
@@ -241,7 +242,22 @@ public Response getApiPlan(@PathParam("planId") String planId) {
...
+    }
+
+    private static Date resolvePlanUpdatedAt(GenericPlanEntity planEntity) {
+        if (planEntity instanceof BasePlanEntity v4Plan) {
+            return v4Plan.getUpdatedAt();
+        }
+        if (planEntity instanceof io.gravitee.rest.api.model.BasePlanEntity v2Plan) {
+            return v2Plan.getUpdatedAt();
+        }
+        throw new IllegalStateException("Unexpected plan entity type: " + planEntity.getClass());
```
paulatulis: IllegalStateException is unreachable today but is a sharp edge. 

 - `GenericPlanEntity` does not expose `getUpdatedAt()` (interface inspected — only `getNeedRedeployAt()`), so the `instanceof` switch is genuinely needed.
  - All paths through `GenericPlanMapper.toGenericPlanWithFlow` produce either v4 or v2 `BasePlanEntity` subtypes, so today the throw can't fire.
  - But io.gravitee.apim.core.plan.model.Plan also implements `GenericPlanEntity`. A future caller substituting that into the resource would turn a successful GET into a 500.
  - Headers are described as "purely additive"; killing the response because we can't compute one is harsher than the null-updatedAt branch right above it (which silently omits). Consider fail-open symmetrically: log.warn(...); return null; rather than throw. Low-impact, but the asymmetry stood out.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiPlansResource.java
```diff
@@ -241,7 +242,22 @@ public Response getApiPlan(@PathParam("planId") String planId) {
...
-        return Response.ok(planMapper.mapGenericPlan(planEntity)).build();
+        Response.ResponseBuilder builder = Response.ok(planMapper.mapGenericPlan(planEntity));
+        Date updatedAt = resolvePlanUpdatedAt(planEntity);
+        if (updatedAt != null) {
+            builder.tag(Long.toString(updatedAt.getTime())).lastModified(updatedAt);
+        }
+        return builder.build();
+    }
+
+    private static Date resolvePlanUpdatedAt(GenericPlanEntity planEntity) {
```
paulatulis, line 256: no test for the throw path in `resolvePlanUpdatedAt`. If you keep the throw, a Mockito stub returning a bare `GenericPlanEntity` mock would lock that behavior. If you switch to fail-open, update the test instead. Either way, the current code path is untested.

paulatulis (approved):
No blockers.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiPlansResource.java
carlos-andres-osorio: Switched `resolvePlanUpdatedAt` to fail-open: replaced the `throw new IllegalStateException(...)` with `log.warn(...)` + `return null`. The method now returns null for any unrecognised plan entity type, matching the existing null-`updatedAt` branch above it — both paths omit the headers rather than killing the response. A `log` field was also added to the class since it had none.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiPlansResource.java
carlos-andres-osorio: Switched to fail-open per the sibling comment (see above). A direct integration test for the unknown-entity path turned out not to be feasible here: `PlanMapper.mapGenericPlan` falls through to an unchecked cast on the v2 `PlanEntity` type, so any entity that isn't `PlanEntity`, `NativePlanEntity`, or `io.gravitee.rest.api.model.PlanEntity` throws `ClassCastException` in the mapper before `resolvePlanUpdatedAt` is reached. All three types the mapper successfully handles are also matched by the existing `instanceof` branches in `resolvePlanUpdatedAt`, so the fail-open path has no production caller today. The change is defensive; the mapper is the effective guard.

PR #17158, opened 2026-05-28
fix: allow default role mapping via api v2 group endpoint

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-13400

## Description

**Changes**
Added apiRole and applicationRole to the REST and core GroupCRDSpec models.
Validated both fields during CRD import:
reject PRIMARY_OWNER
warn and skip unknown role names
Added CRDMembersDomainService.updateGroupDefaultRoles() to create or replace default role memberships.
Updated ImportGroupCRDUseCase to call this on both create and update, before member sync.

**Behaviour**
If apiRole / applicationRole are provided and valid, the group’s default roles are set or updated.
If omitted, existing default roles are left unchanged.
Invalid roles produce warnings and are not applied; PRIMARY_OWNER is rejected

**Default Roles are updated from OWNER to USER:**
image
image
image

**Default PO role update is Rejected:**
image

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/model/GroupCRDSpec.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/model/GroupCRDSpec.java
@@ -40,6 +40,10 @@ public class GroupCRDSpec {
 
     private boolean notifyMembers;
 
+    private String apiRole;
+
+    private String applicationRole;
+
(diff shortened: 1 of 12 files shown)


arpzilla-dev:
Hi @pragsty,

overall i think PR looks great, and is to the point implementation,

its just could u add more unitTestCase covering all classes [eg : ValidateGroupCRDDomainServiceImpl]
if u have already covered, le me know.

pragsty:
> its just could u add more unitTestCase covering all classes [eg : ValidateGroupCRDDomainServiceImpl]
> if u have already covered, le me know.

It has indirect coverage, also thee's no existing testclass nfor ValidateGroupCRDDomainServiceImpl. If you want I'll add one & the testcases

arpzilla-dev (approved):
LGTM, but  @pragsty before merging, kindly verify current code freeze timeline, and follow updates on slack channels.

PR #17167, opened 2026-05-28
feat(authz-ui): import resources from AIM Context Catalog + LLM→Model rename

podlesrafal:
## Issue

Stack PR — ui-8 of authz-ui chain. Builds on #17050 (ui-7, entities list).

## Description

Three commits:

**1. `fix(gamma-rest-api): add definition-model to runtime classpath`**
Authz plugin's `EventRepositoryAuthzEventPublisher` directly instantiates `io.gravitee.gamma.definition.authz.AuthzPolicy` / `AuthzEntity`. The plugin zip doesn't bundle `gravitee-gamma-definition-model` and the rest-api standalone distribution didn't ship it either — `NoClassDefFoundError` at first authz event publish. Adds the dep with compile scope to `gravitee-gamma-rest-api/pom.xml` so it transitively reaches `standalone-container` and lands in `dist/lib/`. Verified empirically with a clean `mvn clean install` of the full APIM monorepo and a fresh start of rest-api.

**2. `refactor(authz-ui): rename LLM → Model across registry, routes, and pages`**
Aligns Authorization with the AIM Context Catalog convention where AI models use the `model.{provider}.{queryName}` entityId form (e.g. `model.openai.gpt-4o`). Drops the redundant `LLMRoute` concept.
- `entity-kind-registry`: drops `llm` canonical, renames `LLMModel` → `Model`. Keeps `llm`/`llmmodel`/`llmroute` as aliases so existing policies with `llm.*` entityIds still classify as `MODEL`.
- `PolicyType`: `LLM` → `MODEL` (UI-side enum, derived from entityId prefix — not on the wire).
- Route `/llms` → `/models`, file `LlmsPage` → `ModelsPage`, service-def `llms.ts` → `models.ts`.
- Legacy `llm.*` entityIds keep working in the resource picker via a fallback branch.

**3. `feat(authz-ui): import resources from AIM Context Catalog`**
Adds an Import dialog on the Entities → Resources tab that pulls MCP servers, AI models, and agents from the sibling AIM Gamma module and registers them as read-only `RESOURCE` entities under source `gravitee-catalog`. Imported entities carry the same canonical entityId as in the catalog so policies always refer to one identifier.

- `ImportFromCatalogDialog`: Sheet with MCP / AI Models / Agents tabs, multi-select, "Select visible" / "Clear" controls, deferred-value search. Per-tab queries gated on `open` so 3× catalog scans only fire when the user opens the dialog. Bulk import runs with concurrency limit 4 (`Promise.allSettled`), reports progress in the footer, and optimistically marks just-imported items via `markImported` so the badge updates without waiting for the refetch.
- `useAimCatalogItems`: paginated catalog client (cap 20×500); `truncated` flag when the catalog has more items than fetched.
- `useImportedCatalogIds`: own query-key namespace; paginates up to 50×1000 with a warning banner if it hits the cap. `markImported` pushes newly-imported ids into the cache for optimistic UI.
- `authz-api.service`: `createEntity` + `deleteEntity` methods.
- `entity.types`: extends `EntitySource` with `gravitee-catalog`.
- `EntitiesPage`: wires the Import button + delete-confirmation dialog; invalidates both `entities` and `importedCatalogIds` caches on delete.
- `query-keys`: new `importedCatalogIds` under authz namespace; AIM catalog keys moved into a dedicated `aimQueryKeys` export.

## Additional context

- 333 vitest specs in `gravitee-gamma-module-authz` pass green.
- Rebased on top of the rebased ui-7 (which was rebased on the force-pushed ui-5 that migrated `ChipCombobox` → graphene-core 2.28 `Combobox`).
- Verified runtime: rest-api boots cleanly, authz plugin loads, no `NoClassDefFoundError`.

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/EntitiesPage.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/EntitiesPage.tsx
@@ -19,8 +19,15 @@ import {
     AlertDescription,
     AlertTitle,
     Badge,
+    Button,
     DataTable,
     DataTablePagination,
+    Dialog,
+    DialogContent,
+    DialogDescription,
+    DialogFooter,
+    DialogHeader,
+    DialogTitle,
     Empty,
     EmptyDescription,
     EmptyHeader,
@@ -35,19 +42,23 @@ import {
     TabsContent,
     TabsList,
     TabsTrigger,
+    toast,
 } from '@gravitee/graphene-core';
-import { BoxesIcon, ShieldIcon, UsersIcon } from '@gravitee/graphene-core/icons';
+import { BoxesIcon, DownloadIcon, ShieldIcon, Trash2Icon, UsersIcon } from '@gravitee/graphene-core/icons';
+import { useQueryClient } from '@tanstack/react-query';
 import type { ColumnDef } from '@tanstack/react-table';
 import { useDeferredValue, useMemo, useState } from 'react';
 import { KpiTile } from '../../components/KpiTile';
+import { authzApiService } from '../../shared/api/authz-api.service';
+import { authzQueryKeys } from '../../shared/api/query-keys';
 import { formatEntityUid, fromBackend } from '../../shared/entity-adapter';
-import { useEntities } from '../../shared/hooks/useEntities';
-import type { EntityInstance } from './entity-types';
+import { useEntities, type UseEntitiesResult } from '../../shared/hooks/useEntities';
+import { CATEGORIES, getEntityCategoryId, type EntityInstance } from './entity-types';
+import { ImportFromCatalogDialog } from './ImportFromCatalogDialog';
 
 type SourceFilter = 'all' | string;
 type TypeFilter = 'all' | string;
 type TabKey = 'principals' | 'resources';
-type EntityKind = 'PRINCIPAL' | 'RESOURCE';
 
 const PAGE_SIZE_OPTIONS = [10, 25, 50, 100];
 const ACTION_PREFIX = 'action.';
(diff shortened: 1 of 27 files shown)


gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/ImportFromCatalogDialog.tsx
```diff
@@ -0,0 +1,451 @@
...
+                                        onCheckedChange={() => toggleSelect(item)}
+                                        className="mt-0.5"
+                                        aria-label={
+                                            isImported ? `${deriveDisplayName(item)} already imported` : `Select ${deriveDisplayName(item)}`
+                                        }
+                                    />
+                                    <div className="min-w-0 flex-1">
+                                        <div className="flex items-center gap-2">
+                                            <span className="truncate text-sm font-medium">{deriveDisplayName(item)}</span>
+                                            <Badge variant="outline" className="shrink-0 font-mono text-[10px]">
```
sclarkamlab: `text-[10px]` is off the graphene type scale, should be `text-xs`. Same on 363 and 416.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/ImportFromCatalogDialog.tsx
```diff
@@ -0,0 +1,451 @@
...
+                            const description = deriveDescription(item);
+                            return (
+                                <label
+                                    key={item.id}
+                                    className={
+                                        'flex items-start gap-3 border-b px-6 py-3 transition-colors ' +
+                                        (isImported
+                                            ? 'cursor-not-allowed opacity-60'
+                                            : 'cursor-pointer hover:bg-muted/40 ' + (isSelected ? 'bg-muted/30' : ''))
+                                    }
```
sclarkamlab: We could use `cn()` here instead of building the className with string `+`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/ImportFromCatalogDialog.tsx
```diff
@@ -0,0 +1,451 @@
...
+                            <Spinner />
+                        </div>
+                    )}
+                    {query.error && (
+                        <Alert variant="destructive" className="m-4">
+                            <AlertDescription>{query.error}</AlertDescription>
+                        </Alert>
+                    )}
+                    {!query.isLoading && !query.error && filtered.length === 0 && (
+                        <div className="flex h-full min-h-[200px] items-center justify-center p-6 text-sm text-muted-foreground">
```
sclarkamlab: `min-h-[200px]` is an arbitrary value, use a scale token like `min-h-48`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/ImportFromCatalogDialog.tsx
```diff
@@ -0,0 +1,451 @@
...
+                            : selectedCount === 0
+                              ? 'No items selected'
+                              : `${selectedCount} item${selectedCount === 1 ? '' : 's'} selected`}
+                    </span>
+                    <div className="flex gap-2">
+                        <Button variant="outline" onClick={() => resetAndClose(false)} disabled={submitting}>
+                            Cancel
+                        </Button>
+                        <Button onClick={handleImport} disabled={selectedCount === 0 || submitting}>
+                            {submitting ? <Spinner className="mr-2 size-4" /> : <DownloadIcon className="mr-2 size-4" />}
```
sclarkamlab: Icon is decorative since the button has text, needs `aria-hidden="true"`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/EntitiesPage.tsx
```diff
@@ -298,38 +280,109 @@ function EntitiesFilterBar({
...
-            </AlertDescription>
-        </Alert>
-    );
+function applyFilters(entities: readonly EntityInstance[], search: string, typeFilter: TypeFilter, sourceFilter: SourceFilter) {
+    const needle = search.trim().toLowerCase();
+    return entities.filter(e => {
+        if (typeFilter !== 'all' && e.uid.type !== typeFilter) return false;
+        if (sourceFilter !== 'all' && sourceLabelOf(e) !== sourceFilter) return false;
+        if (!needle) return true;
+        if (e.uid.id.toLowerCase().includes(needle)) return true;
```
sclarkamlab, line 286: `text-[11px]` is off the type scale, should be `text-xs`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/EntitiesPage.tsx
```diff
@@ -377,63 +430,111 @@ export function EntitiesPage() {
...
+                            typeFilter={resourceTypeFilter}
+                            onTypeFilter={setResourceTypeFilter}
+                            sourcesPresent={resourceSources}
+                            sourceFilter={resourceSourceFilter}
+                            onSourceFilter={setResourceSourceFilter}
+                            searchLabel="Search resources"
+                        />
+                        <div className="ml-auto">
+                            <Button onClick={() => setImportOpen(true)}>
+                                <DownloadIcon className="mr-2 size-4" />
```
sclarkamlab: Decorative icon, needs `aria-hidden="true"`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/entity-types.ts
```diff
@@ -14,21 +14,22 @@
...
-    { id: 'llm', label: 'LLMs' },
-    { id: 'event', label: 'Events' },
-    { id: 'custom', label: 'Custom' },
+    { id: 'principal', label: 'Principals', textColor: 'text-blue-600 dark:text-blue-400' },
+    { id: 'mcp', label: 'MCP', textColor: 'text-teal-600 dark:text-teal-400' },
+    { id: 'api', label: 'APIs', textColor: 'text-indigo-600 dark:text-indigo-400' },
+    { id: 'agent', label: 'Agents', textColor: 'text-orange-600 dark:text-orange-400' },
+    { id: 'model', label: 'AI Models', textColor: 'text-fuchsia-600 dark:text-fuchsia-400' },
+    { id: 'event', label: 'Events', textColor: 'text-cyan-600 dark:text-cyan-400' },
+    { id: 'custom', label: 'Custom', textColor: 'text-slate-600 dark:text-slate-400' },
```
sclarkamlab: These are raw tailwind colours (`text-blue-600` etc), not graphene tokens. Snag is graphene has no categorical palette so there's no clean swap, open to ideas.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/EntitiesPage.tsx
podlesrafal: Done — `text-xs`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/EntitiesPage.tsx
podlesrafal: Done — added `aria-hidden` on the icon (and a couple of others I spotted while in here).

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/ImportFromCatalogDialog.tsx
podlesrafal: Done — all three (`uiTypeBadge`, `Imported` badge, tab counter) now use `text-xs`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/ImportFromCatalogDialog.tsx
podlesrafal: Done — collapsed into `cn()` with three booleans (added `cn` to the graphene-core import).

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/ImportFromCatalogDialog.tsx
podlesrafal: Done — `min-h-48`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/ImportFromCatalogDialog.tsx
podlesrafal: Done — `aria-hidden` on both the `Spinner` and `DownloadIcon`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/entity-types.ts
podlesrafal: Switched to graphene's `chart-1`..`chart-10` categorical palette: `text-chart-1` (principals), `text-chart-2` (mcp), `text-chart-8` (apis), `text-chart-7` (agents), `text-chart-5` (models), `text-chart-10` (events). `custom` falls back to `text-muted-foreground` since it's the neutral catch-all.

sclarkamlab (approved):

PR #17173, opened 2026-05-28
fix: preserve colons in relative endpoint paths

gutek9:
**Issue**

https://gravitee.atlassian.net/browse/APIM-14220

**Description**

Fixed V4 HTTP endpoint resolution so relative endpoint paths containing colons are preserved as literal paths.

HttpEndpointInvoker no longer interprets values like /bar:analyze or /document:translate as endpointName:path references.

**Additional context**

Added regression coverage for relative paths with colons, including Azure-style endpoints such as /document:translate and /openai/deployments/gpt-4o/chat/completions:stream.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/invoker/HttpEndpointInvoker.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/invoker/HttpEndpointInvoker.java
@@ -46,7 +46,7 @@ public class HttpEndpointInvoker implements HttpInvoker, Invoker {
     private static final String MATCH_GROUP_ENDPOINT = "endpoint";
     private static final String MATCH_GROUP_PATH = "path";
     private static final Pattern ENDPOINT_PATTERN = Pattern.compile(
-        "^(?<" + MATCH_GROUP_ENDPOINT + ">[^:]+):(?<" + MATCH_GROUP_PATH + ">.*)$"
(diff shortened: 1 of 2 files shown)


michel-barret (approved):

sammeidroid:
Hi @gutek9 and @michel-barret,

I wanted to bring to your attention that this fix appears to have been lost in version 4.12.x.
While the corrected ENDPOINT_PATTERN regex is present in 4.11.16:

`"^(?<endpoint>[^/:][^:]*):(?<path>.*)$" ` ✅
https://github.com/gravitee-io/gravitee-api-management/blob/4.11.16/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/invoker/HttpEndpointInvoker.java

It has reverted to the old, broken pattern in 4.12.4:

`"^(?<endpoint>[^:]+):(?<path>.*)$"`  ❌
https://github.com/gravitee-io/gravitee-api-management/blob/4.12.4/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/invoker/HttpEndpointInvoker.java

This means the original bug from issue https://github.com/gravitee-io/issues/issues/11478 is present again in 4.12.x, breaking all use cases with colons in relative paths 

Could someone please:
* Confirm whether this was an intentional exclusion or an accidental regression?
* Ensure this fix gets properly merged/cherry-picked into 4.12.x and all future releases?

It would be unfortunate if this fix only lives in the 4.11.x branch and doesn't make it into newer versions. Thank you for your attention to this matter!

PR #16929, opened 2026-05-18
feat: wire backend to url import

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14063
## Description

- Replaces browser-side HttpClient.get(<remote URL>) + parse with a single POST /apis/_import/definition-url (text/plain, body = URL)
- Drops the now-unused Authorization header field from the Remote source panel (the backend ignores Authorization anyway; auth'd remote sources will be designed in APIM-12142).
- Disables the Remote source card in the API update dialog with a tooltip ("Updating an API from a remote URL is not yet supported") — the new endpoint is create-only, so allowing Remote in update mode would
  silently create a duplicate API.

https://github.com/user-attachments/assets/e42b7d4e-2415-44fd-9e75-8db82202ebc9

https://github.com/user-attachments/assets/335ffe46-a567-4052-99b7-ea4a7f994b18

PR #17183, opened 2026-05-28
feat: re-enable Remote source for V4 API update via URL

paulatulis:
## Issue

Subtask: https://gravitee.atlassian.net/browse/APIM-14117
Parent: https://gravitee.atlassian.net/browse/APIM-12142

## Description

Re-enables the **Remote source** card in the V4 update flow and wires it to the backend endpoint that landed in APIM-14116 (`PUT /environments/{envId}/apis/{apiId}/_import/definition-url`). APIM-14063 had temporarily disabled the card with a tooltip while the backend was unavailable — that gate now comes off.

Changes:
- `api-v2.service.ts` — new `updateFromUrl(apiId, definitionUrl)` method (PUT, `text/plain` body)
- `api-import-v4-form.component.ts` — `sources` computed signal no longer disables Remote in update mode; `resolveRemoteImport$` calls `updateFromUrl` when `updateTargetApiId` is set + format=gravitee
- Removed the comment block referencing APIM-12142 and the defense-in-depth `return null` branch that's no longer load-bearing

Tests:
- `api-v2.service.spec.ts` — new `updateFromUrl` test asserts PUT shape (`text/plain`, URL as body)
- `api-import-v4-form.component.spec.ts` — replaced the two update-mode tests:
  - "should keep both source cards enabled in update mode" (replaces "should disable the Remote source card…")
  - "should PUT the remote URL as text/plain to the update endpoint for Gravitee remote in update mode" (replaces "should refuse to call importFromUrl…")

## Test plan

- [x] `yarn jest api-import-v4-form.component.spec.ts api-v2.service.spec.ts` → 78/78 pass
- [x] `npx nx lint console` → clean, prettier ok
- [ ] Manual smoke: open an existing V4 API → Import → Remote source → enter URL of a hosted Gravitee export → Update → confirm API definition reflects the remote content

## Notes for reviewers

- Targets master; backend dependency (APIM-14116, `PUT /_import/definition-url`) merged via #16945 in `e31b627d40`
- Companion to PR #16929 (APIM-14063) which removed the legacy browser-fetch path — same shape applied to update
- The pre-existing backend UX bugs surfaced in APIM-12140 QA (NPE on wrong-schema JSON, raw Netty leak on DNS failure, empty message on 404) will manifest on this flow too since it uses the same `RemoteApiDefinitionParser` + `FetchApiDefinitionFromUrlDomainServiceImpl` chain. Worth filing as separate polish tickets if not already

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.ts
+++ b/gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.ts
@@ -109,22 +109,10 @@ export class ApiImportV4FormComponent {
     { value: 'wsdl', label: 'WSDL', icon: 'gio:language' },
   ];
 
-  /**
-   * Remote source is disabled in update mode until APIM-12142 introduces a backend update-from-URL flow;
-   * the current create endpoint would silently create a new API instead of updating the target.
-   */
-  protected readonly sources = computed(() => {
-    const updateOnlySupportsLocal = !!this.updateTargetApiId();
-    return [
-      { value: 'local', label: 'Local file', icon: 'gio:laptop', disabledReason: null as string | null },
-      {
-        value: 'remote',
-        label: 'Remote source',
-        icon: 'gio:language',
-        disabledReason: updateOnlySupportsLocal ? 'Updating an API from a remote URL is not yet supported' : null,
-      },
-    ];
-  });
+  protected readonly sources = [
+    { value: 'local', label: 'Local file', icon: 'gio:laptop', disabledReason: null as string | null },
+    { value: 'remote', label: 'Remote source', icon: 'gio:language', disabledReason: null as string | null },
+  ];
 
   private readonly importFileContent = signal<string | undefined>(undefined);
 
@@ -203,7 +191,7 @@ export class ApiImportV4FormComponent {
 
   protected readonly selectedSourceLabel = computed(() => {
     const value = this.importSourceMode();
-    return this.sources().find(s => s.value === value)?.label ?? '-';
+    return this.sources.find(s => s.value === value)?.label ?? '-';
   });
 
   protected readonly showImportOptionsStep = computed(() => {
@@ -498,13 +486,7 @@ export class ApiImportV4FormComponent {
     const url = this.configureFileSourceForm.controls.remoteUrl.value?.trim() as string;
 
     if (format === 'gravitee') {
-      // Defense-in-depth: the Remote source card is disabled in update mode (see `sources`), but if the form is
-      // forced to remote programmatically, refuse rather than silently creating a new API via the create endpoint.
-      // APIM-12142 will introduce the backend update-from-URL path.
-      if (updateId) {
-        return null;
-      }
-      return this.apiV2Service.importFromUrl(url);
+      return updateId ? this.apiV2Service.updateFromUrl(updateId, url) : this.apiV2Service.importFromUrl(url);
     }
     if (format === 'wsdl') {
       return updateId
--- a/gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.html
+++ b/gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.html
@@ -112,7 +112,7 @@
               <div class="configure-file-source__col configure-file-source__col--sources">
                 <span class="configure-file-source__col-heading">Source</span>
                 <gio-form-selection-inline formControlName="source" class="configure-file-source__source-selection">
-                  @for (source of sources(); track source.value) {
+                  @for (source of sources; track source.value) {
                     <gio-form-selection-inline-card
                       [value]="source.value"
                       [disabled]="!!source.disabledReason"
--- a/gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.spec.ts
@@ -397,33 +397,32 @@ describe('ApiImportV4FormComponent', () => {
       fixture.detectChanges();
     });
 
-    it('should disable the Remote source card with a tooltip explaining the limitation', () => {
-      // `sources` is a protected signal exposed only to the template; bracket-access keeps the component API clean.
-      const sources = (
-        fixture.componentInstance as unknown as { sources: () => { value: string; disabledReason: string | null }[] }
-      ).sources();
-      const remote = sources.find(s => s.value === 'remote');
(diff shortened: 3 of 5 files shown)


gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.ts
```diff
@@ -109,22 +109,10 @@ export class ApiImportV4FormComponent {
...
-        value: 'remote',
-        label: 'Remote source',
-        icon: 'gio:language',
-        disabledReason: updateOnlySupportsLocal ? 'Updating an API from a remote URL is not yet supported' : null,
-      },
-    ];
-  });
+  protected readonly sources = computed(() => [
+    { value: 'local', label: 'Local file', icon: 'gio:laptop', disabledReason: null as string | null },
+    { value: 'remote', label: 'Remote source', icon: 'gio:language', disabledReason: null as string | null },
```
carlos-andres-osorio, line 114: Dropping the update-mode disabled reason re-enables the Remote card for **all three** format cards in update mode, not only Gravitee. In update mode `resolveRemoteImport$` now also reaches the WSDL branch (`updateApiFromWsdl(updateId, …, 'URL')`) and the OpenAPI branch (`updateApiFromSwagger(updateId, …)`), whereas the PR description scopes the change to the Gravitee `updateFromUrl` path.

I checked and this is functionally fine — those branches hit the existing `_import/wsdl` and `_import/swagger` PUT endpoints, which already accept URL payloads (Gravitee needed a brand-new `_import/definition-url` endpoint precisely because its definition-update endpoint takes a raw body, not a URL). So no runtime regression.

Two things worth confirming: (1) is enabling remote *update* for OpenAPI and WSDL intended here, or only Gravitee? (2) the WSDL remote-update branch has no test — the existing `updateApiFromWsdl` spec only exercises the local-file (INLINE) path, while Gravitee and OpenAPI remote-update are both covered.

## Evidence

A temporary spec exercising the WSDL + Remote + update-mode branch passes, confirming the branch is reachable and routes correctly (so it's a coverage gap, not a defect):

```typescript
it('should call updateApiFromWsdl with remote URL when updateTargetApiId is set', async () => {
  const apiV2 = TestBed.inject(ApiV2Service);
  jest.spyOn(apiV2, 'importWsdlApi');
  jest.spyOn(apiV2, 'updateApiFromWsdl').mockReturnValue(of(fakeApiV4({ id: 'existing-api' })));
  fixture.componentRef.setInput('updateTargetApiId', 'existing-api');
  fixture.detectChanges();

  await harness.selectFormat('wsdl');
  fixture.detectChanges();
  await harness.clickNext();
  await harness.selectSource('remote');
  await harness.setRemoteUrl('https://example.com/calculator.wsdl');
  fixture.detectChanges();
  await harness.clickNext();
  fixture.detectChanges();
  await harness.clickNext();
  await harness.clickImport();

  expect(apiV2.updateApiFromWsdl).toHaveBeenCalledWith(
    'existing-api',
    expect.objectContaining({ payload: 'https://example.com/calculator.wsdl', type: 'URL' }),
  );
  expect(apiV2.importWsdlApi).not.toHaveBeenCalled();
});
```

Result: `Tests: 1 passed, 1 total`.

gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.ts
```diff
@@ -109,22 +109,10 @@ export class ApiImportV4FormComponent {
...
-      { value: 'local', label: 'Local file', icon: 'gio:laptop', disabledReason: null as string | null },
-      {
-        value: 'remote',
-        label: 'Remote source',
-        icon: 'gio:language',
-        disabledReason: updateOnlySupportsLocal ? 'Updating an API from a remote URL is not yet supported' : null,
-      },
-    ];
-  });
+  protected readonly sources = computed(() => [
```
carlos-andres-osorio: Now that the `updateOnlySupportsLocal` branch is gone, this `computed()` no longer reads any signal — both entries are static and `disabledReason` is unconditionally `null`, so the result never changes. It reads as reactive but isn't, and `formats` right above it is already a plain `readonly` array for the same kind of static data.

carlos-andres-osorio (approved):
LGTM. The new `updateFromUrl` service method is consistent with its siblings, the gravitee update-from-URL wiring is correct, and gravitee/openapi remote-update are well covered by tests. Two non-blocking notes left inline: the change re-enables Remote update for all three formats (not just gravitee) and the WSDL remote-update branch has no test (verified reachable + correct via a throwaway spec), and `sources` is now a `computed()` that reads no signals after the disabled-reason removal. Neither blocks merge.

gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.ts
paulatulis: On scope — yes, re-enabling Remote update for all three formats is intentional. APIM-14063's disable gated at the source-card level, which broadly hid Remote for OpenAPI and WSDL in update mode too. OpenAPI and WSDL already had per-format update-from-URL endpoints (_import/swagger, _import/wsdl) and frontend wiring — the only thing blocking them in update mode was that card-level disable. Restoring all three matches pre-APIM-14063 behavior. The PR description was Gravitee-centric because that's where the new backend endpoint + updateFromUrl service method had to be added. On the WSDL test — added in 62733e69b7, but inside the existing describe('ApiImportV4FormComponent (rest-to-soap policy installed)') block rather than the main one. When I tried to add it verbatim to the main describe, it failed: with no rest-to-soap policy mocked, syncImportOptionsFromFormat('wsdl', ...) disables all three options controls (withRestToSoap, withDocumentation, withOASValidationPolicy), and Angular treats an all-disabled FormGroup as status: 'DISABLED' / valid: false — which makes canSubmitImport() return false, the import button stays disabled, the harness click no-ops, and the spy never fires. The policy-installed block already has should call updateApiFromWsdl when updateTargetApiId is set (WSDL+local+update) and should call importWsdlApi with remote URL for WSDL (WSDL+remote+create), so the new should call updateApiFromWsdl with remote URL when updateTargetApiId is set slots in cleanly alongside them and fills the remaining quadrant (WSDL+remote+update).

Worth filing a separate ticket for the all-disabled-FormGroup behavior — today a user without rest-to-soap policy installed can navigate to the WSDL options step but the import button will be permanently disabled. Out of scope for this PR but a real UX gap.

gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.ts
paulatulis: Converted sources from computed(() => [...]) to a plain readonly array, matching formats above it. Updated the three call sites (selectedSourceLabel, the @for in the template, and the spec's bracket-cast type).

PR #17185, opened 2026-05-28
docs: add Automation API sync guidelines to coding documentation

benoitgravitee:
## Issue

https://gravitee.atlassian.net/browse/GKO-2587

## Description

I updated AGENTS.md to specify that Automation should not be forgotten when it comes to add a new fields to Management API.

```text
I want to add a new field in the Application domain object. It's called rating it is an int from 1 to 5. It is not mandatory.
I want you to add it to the database and map it all up to the management API. It can set by the user upon creation and user can also update it.
Plan the backend implementation, from API to database.  
```

I did a plan: Claude (Opus 4.6) tends to forget it, Cursor did not.
I think this is the first step, we should have a reviewing agent to ensure this is not forgotten.

### DIFF
--- a/.agent-rules/java.md
+++ b/.agent-rules/java.md
@@ -93,7 +93,20 @@ mvn prettier:write -pl <module>
 - **Method names**: Must follow **snake_case** convention (e.g., `should_return_error_when_api_not_found()`).
 - **Prefer real objects over mocks**: In clean/hexagonal architecture, instantiate real domain objects and in-memory adapters instead of mocking. Reserve mocks for external I/O boundaries (HTTP clients, databases) that are costly or impractical to instantiate.
 
-## 5. Common Pitfalls
+## 5. Automation API Sync
+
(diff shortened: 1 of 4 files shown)


test-plan-claude.md
```diff
@@ -0,0 +1,202 @@
+# Plan: Add `rating` field to the Application domain object
```
FuYu3699: are these test plan files are used for local testing only?

FuYu3699 (commented):
looks good overall. I added comments about the two local test-plan files.

One suggestion: if we want to enforce a final double-check during code review, we could consider adding a checkbox to `.github/pull_request_template.md`, for example:

```
- [ ] **Automation API parity** considered (or N/A — change does not touch Management API model / OpenAPI / CRD surface). 
```
This would help make the check more visible and consistent during reviews.

test-plan-claude.md
benoitgravitee: I'll remove them... thought I did

benoitgravitee: done

FuYu3699 (approved):

a-cordier (approved):

PR #17186, opened 2026-05-28
fix: handle message phases in shared policy groups

gutek9:
## Issue

https://gravitee.atlassian.net/browse/APIM-12903

## Description

Fixes Shared Policy Group deployment and runtime handling for message API phases.

The REST/core deployment mapping now converts `PUBLISH` and `SUBSCRIBE` to the legacy definition phases `MESSAGE_REQUEST` and `MESSAGE_RESPONSE`, keeping Management API deploy payloads compatible with older gateways. The gateway side now accepts both the new phase names and the legacy message phase names, and maps them to the correct `ExecutionPhase`.

Added unit coverage for REST/core phase mapping, gateway sync deserialization, reactor phase mapping, and reactor factory phase acceptance.

## Additional context

This keeps the wire format backward-compatible for mixed-version MAPI-to-gateway deployments while allowing gateways with this fix to consume both `PUBLISH`/`SUBSCRIBE` and `MESSAGE_REQUEST`/`MESSAGE_RESPONSE`.

### DIFF
--- a/gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/v4/sharedpolicygroup/SharedPolicyGroup.java
+++ b/gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/v4/sharedpolicygroup/SharedPolicyGroup.java
@@ -80,7 +80,17 @@ public class SharedPolicyGroup implements Serializable {
     public enum Phase {
         REQUEST,
         RESPONSE,
+        PUBLISH,
+        SUBSCRIBE,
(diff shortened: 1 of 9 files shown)


ytvnr (approved):
Looks good to me,
Two remarks:
- I saw a lot of `@SuppressWarnings('deprecation')`-> i would remove them.
- `SharedPolicyGroupMessageIntegrationTest` is an integration test verifying the behavior of shared policy groups. It is currently disabled, but it should be good to enable it and check if everything works well

gutek9:
@ytvnr changes implemented.

PR #17207, opened 2026-05-28
Am connection to core

sclarkamlab:
## Issue

https://gravitee.atlassian.net/browse/GMA-309

## Description

First phase of the AM user sync work. This lands all the AM-connection plumbing in APIM core, ahead of the actual sync mechanism which lands separately in `gravitee-gamma-module-authz`.

Three layers:

Repository layer. New `AmConnectionRepository` SPI with Mongo and JDBC adapters, a Liquibase migration for the `am_connections` table, and a TCK test. One connection per organization, with the service-account token stored encrypted.

Core service layer. `AmConnectionService` / `AmConnectionServiceImpl` in `rest-api-service` plus `AmConnectionEntity` in `rest-api-model`. The service decrypts the token via `DataEncryptor` and exposes `findByOrganizationId`, `hasToken`, `save` and `delete`. Deleting an organization now also clears its connection (wired into `DeleteOrganizationCommandHandler`).

Shared gamma-sdk port. An `AmConnection` record and `AmConnectionRepository` port in `gravitee-gamma-plugin-module-api`, with `ApimAmConnectionRepository` delegating to the core service. `findByOrg` / `requireByOrg` return an `AmConnection` carrying the decrypted token plus `baseUrl`, `defaultDomainId`, `defaultDomainHrid` and `gatewayUrl`. `requireByOrg` throws `AmNotConfiguredException` when none is configured. This is the injection point the gamma modules use, not the core service directly.

## Additional context

Follow-up phase (syncing AM users into the FGA engine as PRINCIPAL entities) builds on this port and lands in `gravitee-gamma-module-authz`.

Compile on JDK 21. JDK 25 breaks lombok in `rest-api-service`.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcAmConnectionRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcAmConnectionRepository.java
@@ -43,7 +43,7 @@ public class JdbcAmConnectionRepository extends JdbcAbstractRepository<AmConnect
         return JdbcObjectMapper.builder(AmConnection.class, this.tableName, "organization_id")
             .addColumn("organization_id", Types.NVARCHAR, String.class)
             .addColumn("base_url", Types.NVARCHAR, String.class)
-            .addColumn("service_account_access_token_encrypted", Types.NVARCHAR, String.class)
+            .addColumn("service_account_access_token_encrypted", Types.CLOB, String.class)
             .addColumn("default_domain_id", Types.NVARCHAR, String.class)
             .addColumn("default_domain_hrid", Types.NVARCHAR, String.class)
             .addColumn("gateway_url", Types.NVARCHAR, String.class)
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcAmConnectionRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcAmConnectionRepository.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.jdbc.management;
+
+import io.gravitee.repository.exceptions.TechnicalException;
+import io.gravitee.repository.jdbc.orm.JdbcObjectMapper;
+import io.gravitee.repository.management.api.AmConnectionRepository;
+import io.gravitee.repository.management.model.AmConnection;
+import java.sql.Types;
+import java.util.Date;
+import java.util.List;
+import java.util.Optional;
+import lombok.CustomLog;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Repository;
+
+/**
+ * @author GraviteeSource Team
+ */
+@CustomLog
+@Repository
+public class JdbcAmConnectionRepository extends JdbcAbstractRepository<AmConnection> implements AmConnectionRepository {
+
+    JdbcAmConnectionRepository(@Value("${management.jdbc.prefix:}") String tablePrefix) {
+        super(tablePrefix, "am_connections");
+    }
+
+    @Override
+    protected JdbcObjectMapper<AmConnection> buildOrm() {
+        return JdbcObjectMapper.builder(AmConnection.class, this.tableName, "organization_id")
+            .addColumn("organization_id", Types.NVARCHAR, String.class)
+            .addColumn("base_url", Types.NVARCHAR, String.class)
+            .addColumn("service_account_access_token_encrypted", Types.NVARCHAR, String.class)
+            .addColumn("default_domain_id", Types.NVARCHAR, String.class)
+            .addColumn("default_domain_hrid", Types.NVARCHAR, String.class)
+            .addColumn("gateway_url", Types.NVARCHAR, String.class)
+            .addColumn("updated_at", Types.TIMESTAMP, Date.class)
+            .build();
+    }
+
+    @Override
+    public Optional<AmConnection> findByOrganizationId(String organizationId) throws TechnicalException {
+        log.debug("JdbcAmConnectionRepository.findByOrganizationId({})", organizationId);
+        try {
+            List<AmConnection> items = jdbcTemplate.query(
+                getOrm().getSelectAllSql() + " where organization_id = ?",
+                getOrm().getRowMapper(),
+                organizationId
+            );
+            return items.stream().findFirst();
+        } catch (final Exception ex) {
+            log.error("Failed to find am connection by organizationId: {}", organizationId, ex);
+            throw new TechnicalException("Failed to find am connection by organizationId", ex);
+        }
+    }
+
+    @Override
+    public AmConnection create(AmConnection amConnection) throws TechnicalException {
+        log.debug("JdbcAmConnectionRepository.create({})", amConnection.getOrganizationId());
+        try {
+            jdbcTemplate.update(getOrm().buildInsertPreparedStatementCreator(amConnection));
+            return findByOrganizationId(amConnection.getOrganizationId()).orElse(null);
+        } catch (final Exception ex) {
+            log.error("Failed to create am connection:", ex);
+            throw new TechnicalException("Failed to create am connection", ex);
+        }
+    }
+
+    @Override
+    public AmConnection update(AmConnection amConnection) throws TechnicalException {
+        log.debug("JdbcAmConnectionRepository.update({})", amConnection.getOrganizationId());
+        if (amConnection == null) {
+            throw new IllegalStateException("Unable to update null am connection");
+        }
+        try {
+            int rows = jdbcTemplate.update(getOrm().buildUpdatePreparedStatementCreator(amConnection, amConnection.getOrganizationId()));
+            if (rows == 0) {
+                throw new IllegalStateException("Unable to update am connection " + amConnection.getOrganizationId());
+            }
+            return findByOrganizationId(amConnection.getOrganizationId()).orElse(null);
+        } catch (final IllegalStateException ex) {
+            throw ex;
+        } catch (final Exception ex) {
+            log.error("Failed to update am connection:", ex);
+            throw new TechnicalException("Failed to update am connection", ex);
+        }
+    }
+
+    @Override
+    public void delete(String organizationId) throws TechnicalException {
+        log.debug("JdbcAmConnectionRepository.delete({})", organizationId);
+        try {
+            jdbcTemplate.update("delete from " + this.tableName + " where organization_id = ?", organizationId);
+        } catch (final Exception ex) {
+            log.error("Failed to delete am connection:", ex);
+            throw new TechnicalException("Failed to delete am connection", ex);
+        }
+    }
+}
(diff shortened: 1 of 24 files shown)


gravitee-gamma/gravitee-gamma-plugin/gravitee-gamma-plugin-module/gravitee-gamma-plugin-module-api/src/main/java/io/gravitee/apim/plugin/gamma/api/identity/AmConnection.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Shared connection details a gamma module uses to reach an AM instance for an organization.
+ * Storage and token encryption are owned by APIM rest-api core; see {@link AmConnectionRepository}.
+ *
+ * <p>Nullable fields ({@code serviceAccountAccessToken}, {@code defaultDomainId},
+ * {@code defaultDomainHrid}, {@code gatewayUrl}) may be {@code null} when not yet configured.
+ */
+public record AmConnection(
+    String baseUrl,
+    String serviceAccountAccessToken,
```
podlesrafal, line 27: 🟠 **Major — token leak via `toString()`.** This is a `record`, so Java auto-generates a `toString()` containing all components, including `serviceAccountAccessToken` in plaintext. Any `log.debug("...{}", connection)` or putting this object into an exception message will leak the secret into the logs.

`AmConnectionEntity` is safe (`@Getter/@Setter`, no `@ToString`), and the repository model POJO too — the only risk is this record.

**Suggestion:** override `toString()` to redact the token, e.g. return `"***"` instead of the value.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/AmConnectionServiceImpl.java
```diff
@@ -0,0 +1,144 @@
...
+        try {
+            Optional<AmConnection> existing = amConnectionRepository.findByOrganizationId(organizationId);
+
+            AmConnection model = new AmConnection();
+            model.setOrganizationId(organizationId);
+            model.setBaseUrl(stripTrailingSlashes(entity.getBaseUrl()));
+            model.setDefaultDomainId(entity.getDefaultDomainId());
+            model.setDefaultDomainHrid(entity.getDefaultDomainHrid());
+            model.setGatewayUrl(entity.getGatewayUrl());
+            model.setServiceAccountAccessTokenEncrypted(encryptOrPreserve(entity.getServiceAccountAccessToken(), existing.orElse(null)));
```
podlesrafal, line 87: 🟡 **Minor — asymmetric partial-update semantics.** Only the token has "null = preserve" behavior (`encryptOrPreserve`); `baseUrl`/`defaultDomainId`/`defaultDomainHrid`/`gatewayUrl` are copied straight from `entity`, so a save with any of them `null` will wipe the stored value. The current sole caller (`ApimAmConnectionRepository.toEntity`) always sets all fields, so this is latent — but the javadoc on `AmConnectionService.save` documents only the token semantics and is silent that the rest is a full overwrite.

**Suggestion:** add one sentence to the javadoc stating that the remaining fields fully replace the stored values.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/10_add_am_connections_table.yml
```diff
@@ -0,0 +1,15 @@
...
+  - changeSet:
+      id: 4.12.0_10_add_am_connections_table
+      author: GraviteeSource Team
+      changes:
+        - createTable:
+            tableName: ${gravitee_prefix}am_connections
+            columns:
+              - column: {name: organization_id, type: nvarchar(64), constraints: { nullable: false, primaryKey: true, primaryKeyName: "pk_${gravitee_prefix}am_connections" } }
+              - column: {name: base_url, type: nvarchar(256)}
+              - column: {name: service_account_access_token_encrypted, type: nvarchar(4000)}
```
podlesrafal: 🟡 **Minor — token column size.** The stored value is `DataEncryptor.encrypt(plaintext)` (AES + base64), and an AM service-account token is a JWT. Base64 + IV inflates ~1.4×, so `nvarchar(4000)` fits ~2800 plaintext chars. A plain JWT is safe, but a JWT with embedded roles/claims may exceed the limit → silent corruption or insert error. For an encrypted blob there's no benefit to the 4000 limit.

**Suggestion:** `nvarchar(max)` / `clob` (like other encrypted-secret columns in the repo).

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcAmConnectionRepository.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public AmConnection update(AmConnection amConnection) throws TechnicalException {
+        log.debug("JdbcAmConnectionRepository.update({})", amConnection.getOrganizationId());
+        if (amConnection == null) {
+            throw new IllegalStateException("Unable to update null am connection");
+        }
+        try {
+            int rows = jdbcTemplate.update(getOrm().buildUpdatePreparedStatementCreator(amConnection, amConnection.getOrganizationId()));
+            if (rows == 0) {
```
podlesrafal, line 92: 🟡 **Minor — SPI behavior divergence: JDBC vs Mongo.** Here `update` throws `IllegalStateException` on 0 updated rows. `MongoAmConnectionRepository.update` calls `internal.save()`, which does a silent upsert (creates the row). Latent, because the service always checks existence before choosing create/update — but it's a difference in the SPI contract that the TCK doesn't catch (the test only covers updating an existing row).

**Suggestion:** a TCK test "update of a non-existent row throws" (would force both adapters into agreement) or explicit documentation of the contract.

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoAmConnectionRepository.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public AmConnection create(AmConnection amConnection) throws TechnicalException {
+        log.debug("Create am connection [{}]", amConnection.getOrganizationId());
+        try {
+            return map(internalAmConnectionRepository.insert(map(amConnection)));
+        } catch (Exception e) {
+            log.error("An error occurs when creating am connection [{}]", amConnection.getOrganizationId(), e);
+            throw new TechnicalException("An error occurs when creating am connection");
```
podlesrafal: 🟡 **Minor — dropped exception `cause`.** `new TechnicalException("An error occurs when creating am connection")` doesn't pass `e` (same in `update:62` and `delete:73`). Siblings in this module (`MongoClusterRepository`, `MongoAccessPointRepository`) chain the cause, so this deviates from the house style. The cause is logged, so impact is low, but for parity it's worth adding `e` as the second argument.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/AmConnectionServiceImpl.java
```diff
@@ -0,0 +1,144 @@
...
+            model.setUpdatedAt(new Date());
+
+            AmConnection saved = existing.isPresent() ? amConnectionRepository.update(model) : amConnectionRepository.create(model);
```
podlesrafal, line 90: 🟡 **Minor — TOCTOU on first save.** `findByOrganizationId` → create/update is non-atomic; two concurrent first saves for the same organization can both hit `create` → PK violation thrown as `TechnicalManagementException`. Acceptable for a rare admin config endpoint — just awareness / a one-line comment, not a blocker.

podlesrafal (commented):
## Review — AM connection to core

Solid, well-tested PR. Encryption reuses the existing `apiPropertiesEncryptor`, JDBC is fully parameterized (no SQLi), and AM connection deletion is correctly wired into the organization-deletion cascade. No critical blockers — one Major worth fixing before merge, the rest are Minor/Info.

**🟠 Major:** the `AmConnection` plugin record has an auto-generated `toString()` that exposes the token in plaintext (details in the inline comment).

**Info (no inline anchor):**
- `AmNotConfiguredException` → the comment says "maps to 503", but `AbstractDomainException` carries no status — the 503 depends on manual translation in a module that doesn't exist anywhere yet (`requireByOrg`/`isConfigured`/the class itself are unused in main, waiting for a consumer). Semantically 412/409 fits better than 503 (which suggests a transient state).
- `@Lazy` on the repo injection (in `AmConnectionServiceImpl` and `DeleteOrganizationCommandHandler`) — consistent with the house pattern for repo-SPI beans, OK.
- Migration is correct (PK `pk_${gravitee_prefix}am_connections`, types/nullability consistent with `09_*`); the include in `master.yml` and the entry in `JdbcTestRepositoryInitializer` are present.
- Test coverage is good. Gaps: no test for the JDBC-vs-Mongo update behavior divergence; `requireByOrg`/the 503 translation are untested (dead code until a consumer exists).

gravitee-gamma/gravitee-gamma-plugin/gravitee-gamma-plugin-module/gravitee-gamma-plugin-module-api/src/main/java/io/gravitee/apim/plugin/gamma/api/identity/AmConnection.java
sclarkamlab: Good catch. Overrode toString to print the token as `***`, with a test that the plaintext never shows up. 6079f11996

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/AmConnectionServiceImpl.java
sclarkamlab: Done in a5f2d8242c. Added a line to the save javadoc saying the other fields fully overwrite, so null clears them.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/10_add_am_connections_table.yml
sclarkamlab: Yeah, switched it to `clob` (and the JDBC ORM to `Types.CLOB`) to match the other encrypted columns. b0803e190b

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcAmConnectionRepository.java
sclarkamlab: Made Mongo match instead of relaxing this. `MongoAmConnectionRepository.update` now does an existsById check and throws like JDBC, documented the contract on the SPI and added a TCK test that pins both adapters. 7cff913eef

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoAmConnectionRepository.java
sclarkamlab: Done in 7cff913eef, passing `e` through in create/update/delete now.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/AmConnectionServiceImpl.java
sclarkamlab: Added a one line comment flagging it. Fine to leave as is for a config endpoint this rare. a5f2d8242c

sclarkamlab:
Thanks Rafal. Addressed everything across 5 commits (6079f11996..87d00c9d81).

On the 503 info point: you're right, AbstractDomainException carries no status and 503 implies transient. Fixed the comment to say the REST boundary maps it to 409. The class is still unused until the consumer lands, so I left the actual mapping for that PR.

podlesrafal (approved):

PR #17208, opened 2026-05-28
fix(GKO-2822): preserve HRID on management API updates

benoitgravitee:
## Issue

https://gravitee.atlassian.net/browse/GKO-2822

## Summary

- Preserve existing API HRID when Management API v4 updates omit it (fixes detached APIs losing HRID after console edits; Automation API then returned `hrid: null`).
- Apply the same null-coalescing pattern for plan and shared policy group updates.
- Same for groups

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/model/Plan.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/model/Plan.java
@@ -299,7 +299,7 @@ public class Plan implements GenericPlanEntity {
             .name(updated.name)
             .description(updated.description)
             .order(updated.order)
-            .hrid(updated.hrid)
+            .hrid(updated.hrid == null ? hrid : updated.hrid)
             .updatedAt(TimeProvider.now())
             .planDefinitionHttpV4(updated.planDefinitionHttpV4)
             .planDefinitionNativeV4(updated.planDefinitionNativeV4)
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/shared_policy_group/model/SharedPolicyGroup.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/shared_policy_group/model/SharedPolicyGroup.java
@@ -193,7 +193,7 @@ public class SharedPolicyGroup {
 
     public SharedPolicyGroup update(UpdateSharedPolicyGroup updateSharedPolicyGroup) {
         return this.toBuilder()
-            .hrid(updateSharedPolicyGroup.getHrid())
+            .hrid(updateSharedPolicyGroup.getHrid() == null ? this.getHrid() : updateSharedPolicyGroup.getHrid())
             .crossId(updateSharedPolicyGroup.getCrossId() == null ? this.getCrossId() : updateSharedPolicyGroup.getCrossId())
             .name(updateSharedPolicyGroup.getName() == null ? this.getName() : updateSharedPolicyGroup.getName())
             .description(
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/impl/ApiServiceImplTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/impl/ApiServiceImplTest.java
@@ -831,6 +831,17 @@ public class ApiServiceImplTest {
     }
 
     @Test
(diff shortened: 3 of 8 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/model/Plan.java
```diff
@@ -299,7 +299,7 @@ public Plan update(Plan updated) {
             .name(updated.name)
             .description(updated.description)
             .order(updated.order)
-            .hrid(updated.hrid)
+            .hrid(updated.hrid == null ? hrid : updated.hrid)
```
carlos-andres-osorio, line 302: This null-hrid preservation guard has no effective test coverage. `PlanFixtures` never sets `hrid`, so every test in `UpdatePlanDomainServiceTest` (and `PlanUpdatesTest`) exercises the `null→null` path — reverting this guard to `.hrid(updated.hrid)` would not break any existing test in the suite. A plan with an established hrid would silently have it erased on the next update with no CI signal catching the regression.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/impl/ApiServiceImplTest.java
```diff
@@ -827,6 +827,17 @@ public void shouldUpdate() throws TechnicalException {
         verify(searchEngineService, times(1)).index(eq(GraviteeContext.getExecutionContext()), any(), eq(false));
     }
 
+    @Test
+    public void shouldPreserveHridWhenUpdateEntityDoesNotProvideOne() throws TechnicalException {
```
carlos-andres-osorio: Missing positive-path counterpart: this test verifies that a `null` hrid in the update entity preserves the existing value, but there is no complementary test for when `updateApiEntity.hrid` is non-null. If the production code were changed to unconditionally preserve the existing hrid (regardless of what the update entity provides), no test would catch that regression.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/shared_policy_group/model/SharedPolicyGroup.java
```diff
@@ -193,7 +193,7 @@ public static SharedPolicyGroup from(CreateSharedPolicyGroup createSharedPolicyG
 
     public SharedPolicyGroup update(UpdateSharedPolicyGroup updateSharedPolicyGroup) {
         return this.toBuilder()
-            .hrid(updateSharedPolicyGroup.getHrid())
+            .hrid(updateSharedPolicyGroup.getHrid() == null ? this.getHrid() : updateSharedPolicyGroup.getHrid())
```
carlos-andres-osorio, line 196: This `hrid` null-coalesce branch is the only one of the four entities in this PR that ships without a test. `Plan.update()`, the v4 API path, and the Group mapping all got coverage, but neither side of this branch is exercised: `UpdateSharedPolicyGroupUseCaseTest` makes no assertion on `hrid`. Reverting this line to `.hrid(updateSharedPolicyGroup.getHrid())` leaves the suite green, so a future regression here would reintroduce the GKO-2822 bug for shared policy groups with no CI signal — the same gap that was flagged as critical on `Plan.update()`. Worth covering both the preserve (update hrid null → existing kept) and the use-provided (update hrid non-null → new value) cases, mirroring the two new Plan tests.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/impl/ApiServiceImplTest.java
```diff
@@ -830,6 +830,28 @@ public void shouldUpdate() throws TechnicalException {
...
```
carlos-andres-osorio: Minor/nit: these two new tests use `public void shouldXxx` camelCase, but this is a JUnit 5 file (`org.junit.jupiter.api.Test`) that already contains package-private snake_case methods (e.g. `should_not_set_allowedInApiProducts_for_non_proxy_v4_imports`). For consistency with the migrated-test convention in this file, consider package-private snake_case names. Non-blocking — the file currently mixes both styles.

carlos-andres-osorio (approved):
Full re-review after the force-push. HRID preservation is correct across all four update paths (Plan, SharedPolicyGroup, ApiServiceImpl, GroupServiceImpl), and every test-coverage gap from the prior reviews is now resolved — each preserve-branch has a test that would fail if the guard were reverted, and the test naming is consistent. Only two minor `@ParameterizedTest` style suggestions remain, not blocking. LGTM.

PR #17210, opened 2026-05-28
refactor(rest-api): extract cache-header helper to AbstractResource

carlos-andres-osorio:
## Issue

[APIM-14212](https://gravitee.atlassian.net/browse/APIM-14212)

## Why

The `Cache-Control` / `Last-Modified` header pattern (`builder.tag(Long.toString(date.getTime())).lastModified(date)`) was duplicated across several v2 API resource classes. Consolidating it into one place removes the duplication and guarantees consistent null-handling — `Last-Modified`/`ETag` are derived from `updatedAt` in a single helper rather than re-implemented per call site.

## What changed

- Added a protected `applyCacheHeaders(ResponseBuilder, Date updatedAt)` helper to `AbstractResource` that sets the `Last-Modified`/`ETag` headers from `updatedAt`, suppressing them when `updatedAt` is `null`.
- Migrated the cache-header call sites in `ApiResource` (review, PATCH, GET, stop endpoints) and `ApiPlansResource` to use the helper.

## How to test

No special setup — this is a behaviour-preserving refactor covered by the unit tests in `AbstractResourceCacheHeaderTest`, `ApiResource_ReviewsTest`, `ApiResource_PatchApiConcurrencyTest`, `ApiResource_StopTest`, and `ApiPlansResourceTest`. Run the rest-api module test suite to verify.

[APIM-14212]: https://gravitee.atlassian.net/browse/APIM-14212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/AbstractResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/AbstractResource.java
@@ -49,6 +49,7 @@ import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.EntityTag;
 import jakarta.ws.rs.core.HttpHeaders;
+import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.SecurityContext;
 import jakarta.ws.rs.core.UriBuilder;
 import jakarta.ws.rs.core.UriInfo;
@@ -56,12 +57,14 @@ import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import lombok.CustomLog;
 import org.glassfish.jersey.message.internal.HttpHeaderReader;
 import org.glassfish.jersey.message.internal.MatchingEntityTag;
 import org.springframework.security.core.context.SecurityContextHolder;
@@ -73,6 +76,7 @@ import org.springframework.util.CollectionUtils;
  * @author Azize ELAMRANI (azize.elamrani at graviteesource.com)
  * @author GraviteeSource Team
  */
+@CustomLog
 public abstract class AbstractResource {
 
     public static final String ORGANIZATION_ADMIN = RoleScope.ORGANIZATION.name() + ':' + SystemRole.ADMIN.name();
@@ -275,6 +279,16 @@ public abstract class AbstractResource {
             requestUriBuilder.path(path);
         }
         return requestUriBuilder.build();
+    }
+
+    protected Response.ResponseBuilder applyCacheHeaders(Response.ResponseBuilder builder, Date updatedAt) {
+        if (updatedAt != null) {
+            builder.tag(Long.toString(updatedAt.getTime())).lastModified(updatedAt);
+        } else {
+            log.warn("Record has no updatedAt date set, Last-Modified header won't be added to the response.");
+        }
+
+        return builder;
     }
 
     protected void evaluateIfMatch(final HttpHeaders headers, final String etagValue) {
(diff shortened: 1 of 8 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/AbstractResource.java
```diff
@@ -277,6 +279,13 @@ protected URI getLocationHeader(String... paths) {
         return requestUriBuilder.build();
     }
 
+    protected Response.ResponseBuilder applyCacheHeaders(Response.ResponseBuilder builder, Date updatedAt) {
+        if (updatedAt != null) {
```
paulatulis, line 285: This silent suppression is the only real behavior change in the refactor. Endpoints that previously called .getUpdatedAt().getTime() unguarded (deploy/start/stop/reviews/GET) used to NPE→500 on a null updatedAt; they now return success with the ETag/Last-Modified dropped. Good as a fix, but since the ETag is the optimistic-concurrency token clients feed back via If-Match, dropping it silently changes those endpoints' null contract. If null can't happen on a persisted entity, prefer asserting/logging over quietly omitting; if it can, this changes the concurrency contract — please make it a deliberate decision.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiConcurrencyTest.java
```diff
@@ -270,6 +270,39 @@ void should_round_trip_GET_PATCH_GET_with_consistent_ETag_chain() {
...
+    @Test
+    void should_have_Last_Modified_encoding_same_instant_as_ETag_on_successful_patch() {
+        var response = rootTarget(API)
+            .request()
+            .header(HttpHeaders.IF_MATCH, "\"1000\"")
+            .method("PATCH", Entity.entity("{\"name\":\"patched\"}", MERGE_PATCH_TYPE));
+
+        assertThat(response).hasStatus(OK_200);
+        org.assertj.core.api.Assertions.assertThat(response.getHeaderString(HttpHeaders.ETAG)).isEqualTo("\"1000\"");
+        org.assertj.core.api.Assertions.assertThat(response.getLastModified().getTime()).isEqualTo(1000L);
```
paulatulis: This exact-millis assertion only passes because Last-Modified is second-granularity and 1000 ms is a whole second; 1500L would fail. Fine as-is, but worth a comment or rounding so the second-precision dependency is explicit.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/AbstractResource.java
carlos-andres-osorio: Added a warning log when updatedAt is null.
This should be an exception since all changes to the DB include an updatedAt value.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiConcurrencyTest.java
carlos-andres-osorio: Replaces misleading string 1000 with a constant that stores a Time stamp

paulatulis: Not a blocker, but the underlying brittleness isn't fully closed: 
2026-06-01T13:45:30Z is still a whole-second value (zero millis), which is the only reason `should_have_Last_Modified_encoding_same_instant_as_ETag_on_successful_patch` (getLastModified().getTime() vs toEpochMilli()) still passes — Last-Modified is second-granularity. 
If anyone ever gives that constant sub-second precision, that test breaks for a non-obvious reason. A one-line comment on the constant (// keep whole-second: Last-Modified is second-granularity) would lock it in.

paulatulis (approved):
Other than prettier CI issue, looks good!

PR #17215, opened 2026-05-29
test(gamma-console): drop obsolete about route assertions

gcusnieux:
## Why

`routes.spec.ts` still asserted the `about` host route after commit `31c13c201c` ("refactor(gamma): remove about page from home module") removed the about page, the `AppRoutes` entry, the nav config and the `about` area from `routes.ts`. The spec was not updated, so `resolveHostRoute('.../about')` now returns `home` and the suite fails.

This turns the **Lint & test Gamma Console** CI red on master, and therefore on every open PR rebased on it (e.g. #17053).

## What

Remove the `about` references left in `routes.spec.ts`:
- drop the `hostNavPath('about', ...)` assertion
- delete the `should resolve about` test
- replace the `about` case in `buildPathnameAfterEnvironmentChange` with a `home` one

`HostNavKey` is now `'home'` only, so these assertions no longer match the code.

Suite passes locally (8/8).

### DIFF
--- a/gravitee-gamma/gravitee-gamma-control-plane-webui/src/shared/config/routes.spec.ts
+++ b/gravitee-gamma/gravitee-gamma-control-plane-webui/src/shared/config/routes.spec.ts
@@ -16,7 +16,7 @@
 import { buildPathnameAfterEnvironmentChange, hostNavPath, pathSegmentsAfterEnvironment, resolveHostRoute } from './routes';
 
 describe('hostNavPath', () => {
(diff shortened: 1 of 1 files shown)


jourdiw (approved):

Okhelifi (approved):

Okhelifi:
Thx Guillaume sorry I just checked and on my PR only the backend has build and I was in the train when it has been automerged 🙈

PR #17238, opened 2026-05-29
fix(ci): detect gamma-only changes for frontend CI jobs

jourdiw:
## Summary

- The CI `changedFiles` utility (`git.ts`) truncates file paths to the first directory segment (e.g. `gravitee-gamma/gravitee-gamma-control-plane-webui/src/...` becomes `gravitee-gamma`)
- `shouldBuildGammaUI` compared against the full nested path `gravitee-gamma/gravitee-gamma-control-plane-webui`, which never matched the truncated input
- This caused frontend lint, test, and build jobs to be silently skipped for any gamma-only branch that didn't also touch root-level files like `package.json` or `yarn.lock`
- Fix: match on `gravitee-gamma` instead of the full nested workdir path
- Updated the test to use realistic truncated input (`gravitee-gamma` instead of `gravitee-gamma/gravitee-gamma-control-plane-webui`); all 100 CI tests pass

## Test plan

- [x] All 100 existing CI config generation tests pass locally
- [ ] Verify CI runs frontend jobs on a gamma-only branch after merge

### DIFF
--- a/.circleci/ci/src/config.ts
+++ b/.circleci/ci/src/config.ts
@@ -54,6 +54,7 @@ const components = {
   gamma: {
     project: 'gravitee-gamma-control-plane-webui',
     workdir: 'gravitee-gamma/gravitee-gamma-control-plane-webui',
+    rootDir: 'gravitee-gamma',
     image: 'gamma-ui',
   },
 };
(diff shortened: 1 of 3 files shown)


phiz71 (approved):
I approve to unblock, bu I have the impression we could refactor a little bit

PR #17249, opened 2026-05-29
feat(apim-console): enable remote source URL for v4 OpenAPI import

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-12143
https://gravitee.atlassian.net/browse/APIM-12144

## Description
- importSwaggerDescriptor.ts — type?: 'INLINE' | 'URL'
- api-import-v4-form.component.ts — builder sets type; remote→URL, file→INLINE (create and update paths)
- api-import-v4-form.component.spec.ts — create assertions + 2 new update-mode tests:
- HTTP-level PUT …/apis/{id}/_import/swagger carries type: 'URL'
- error-state: backend message surfaced in snackbar on a failed update fetch
- 3 files, +62 / −5 · 44/44 tests green

https://github.com/user-attachments/assets/f7d0841e-b924-49df-8e26-534beb5f2c87

### DIFF
--- a/gravitee-apim-console-webui/src/entities/management-api-v2/api/v4/importSwaggerDescriptor.ts
+++ b/gravitee-apim-console-webui/src/entities/management-api-v2/api/v4/importSwaggerDescriptor.ts
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
 export interface ImportSwaggerDescriptor {
+  /** The raw OpenAPI/Swagger content (INLINE) or the URL to fetch it from (URL). */
   payload: string;
+  /** Whether the payload is an inline OpenAPI string or a remote URL. Defaults to INLINE. */
+  type?: 'INLINE' | 'URL';
   withDocumentation?: boolean;
   withOASValidationPolicy?: boolean;
 }
--- a/gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/api/import-v4/api-import-v4-form/api-import-v4-form.component.spec.ts
@@ -161,6 +161,7 @@ describe('ApiImportV4FormComponent', () => {
     expect(req.request.body).toEqual(
       expect.objectContaining({
         payload: yaml,
+        type: 'INLINE',
         withDocumentation: true,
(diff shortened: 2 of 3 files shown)


gravitee-apim-console-webui/src/entities/management-api-v2/api/v4/importSwaggerDescriptor.ts
```diff
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
 export interface ImportSwaggerDescriptor {
+  /** The raw OpenAPI/Swagger content (INLINE) or the URL to fetch it from (URL). */
   payload: string;
+  /** Whether the payload is an inline OpenAPI string or a remote URL. Defaults to INLINE. */
+  type?: 'INLINE' | 'URL';
```
carlos-andres-osorio, line 20: This new `type` field doesn't seem to be consumed by the backend. The v2 REST `ImportSwaggerDescriptor` schema declares only `payload`, `withDocumentation`, `withOASValidationPolicy`, `withPolicies`, and `withPolicyPaths` (`gravitee-apim-rest-api/.../resources/openapi/openapi-apis.yaml:4140`), so `type` is dropped as an unknown property on deserialization. The v4 import path reads only `getPayload()` (`OAIDomainServiceImpl.convert`, `OAIDomainServiceImpl.java:74`) and `OAIParser.parse` auto-detects URL vs inline content via `isLocationUrl(content)` -> `readLocation(url)` (`OAIParser.java:51`).

The remote-URL import works regardless because the backend recognizes a URL in the payload itself, so `type` appears to be inert on the wire today. Is it an intentional forward-compatibility hook for a planned backend change, or can it be dropped? Either is fine — just flagging so it's a deliberate choice.

paulatulis: type is intentionally inert on the backend, matching the v1/v2 pattern (the descriptor carries it for entity-shape parity per APIM-13658, but the import sniffs the payload). The real v4-vs-v2 gap was the SSRF guard, which APIM-14246 now adds. https://github.com/gravitee-io/gravitee-api-management/pull/17254

carlos-andres-osorio (approved):

PR #17253, opened 2026-05-29
feat(authz-ui): edit local entities (name, description, parents)

podlesrafal:
## Summary
Wires the existing backend `PUT /entities/{entityId}` into the UI so **local** entities can be edited in place (the previously-promised "Edit … in a follow-up PR").

- **service:** `updateEntity(env, entityId, { attributes, parents })` + `UpdateEntityRequest`.
- **EditEntityDialog:** prefilled right-side sheet — **Entity ID is read-only** (immutable key); edits **display name**, **description**, **parents**. Submits the **complete attribute map** (rebuilt via `toBackend`) so unrelated attributes (e.g. `department`, `email`, `_kind`) are preserved — the backend update is a *full replace*, not a patch, so a partial payload would wipe them.
- **EntitiesPage:** row-level **Edit (pencil)** on both Principals and Resources, shown **only for `source=local`** (catalog/APIM-sourced stay read-only); opens the dialog with the right kind and invalidates the list on success.
- **tests:** EditEntityDialog (prefill, read-only id, attribute preservation, drop-description, validation, error) + EntitiesPage edit wiring (local-only).

## Test plan
- [x] `tsc -p tsconfig.app.json --noEmit` clean
- [x] `vitest run` — full module suite green (384 tests) incl. new edit tests
- [x] Manual (Chrome, local stack): Principals/Resources show Edit for local rows only; editing `user.bob` → "Bob Edited" persists, Entity ID unchanged, entity counts stable (no duplicate/data-loss); catalog-sourced rows have no Edit

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/EditEntityDialog.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/EditEntityDialog.tsx
@@ -1,0 +1,246 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import {
+    Alert,
+    AlertDescription,
+    Badge,
+    Button,
+    Combobox,
+    ComboboxChip,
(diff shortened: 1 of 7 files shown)


sclarkamlab:
Had a read through, looks good. Nothing blocking, just a few small things if you feel like it.

The save button is wired with `onClick={handleSubmit as unknown as () => void}` and `type="button"` because it sits outside the form (the footer closes after `</form>`). It works fine and enter-to-submit still goes through the form's onSubmit, but the cast is a bit ugly. If you want to drop it you could give the footer button `form="<id>"` + `type="submit"`, or move the buttons inside the form. Totally optional.

Parent chips fall back to the raw id (like `group.developers`) if an existing parent isn't in the first 200 entities the picker fetches. Cosmetic, and tied to the same paging cap you've already got a TODO for elsewhere, so not fussed.

parentsQuery loading/error is swallowed so the picker just looks empty if that fetch fails, but that matches the create flow so leaving it is fine.

Tests are genuinely solid. Two I'd consider adding: one that save is disabled while a request is in flight, and one that reopening the dialog with a different entity re-seeds name/description/parents (that `[open, entity]` effect is the subtle bit). Everything else is well covered.

I didn't run lint/tests locally, assuming CI has it.

sclarkamlab (approved):

PR #17254, opened 2026-05-29
feat(apim-rest-api): Guard v4 OpenAPI URL import against SSRF (v2 parity)

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14246
## Description

The v4 OpenAPI import fetches remote URL payloads server-side via the swagger parser, but — unlike the v2 path (SwaggerServiceImpl) — never validated the URL. This adds UrlSanitizerUtils.checkAllowed (import whitelist + private-address blocking) in OAIDomainServiceImpl when the payload is a URL, matching v2. No new endpoint, no API change; the type field stays a frontend/entity-shape detail (APIM-13658) and the backend sniffs the payload as before.


carlos-andres-osorio (approved):
LGTM!

PR #17263, opened 2026-05-30
feat(gamma-authz): sync AM users into FGA as principals

sclarkamlab:
## Issue

https://gravitee.atlassian.net/browse/GMA-415

## Description

Adds the ability to sync AM users into FGA as PRINCIPAL entities from the gamma-authz module.

There's a single button in the new User Sync page that drives the sync off the stored AM connection. The sync runs as a persisted AsyncJob so it survives restarts and the UI can poll its status. Principals are keyed on the token sub.

The work is layered cleanly: core use cases (start/status/sync) talk to an AmUserClient port, the infra layer implements it against the AM SDK, and the REST resource exposes start + status endpoints. The plugin assembly bundles the AM SDK and definition-model jars so it loads standalone.

## Additional context

Sits on top of the AM connection storage work that already merged to master.


gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/infra/service_provider/AmSdkUserClientFactory.java
```diff
@@ -0,0 +1,71 @@
...
+                    }
+                    String text = p.getValueAsString();
+                    return text == null || text.isEmpty() ? null : OffsetDateTime.parse(text);
+                }
+            }
+        );
+        apiClient.getObjectMapper().registerModule(millisTimestamps);
+
+        return new UserApiImpl(apiClient);
+    }
```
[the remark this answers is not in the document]
sclarkamlab: Good catch. Done in `3b318ffc19`. Went with a per-run session instead of a process-wide cache, builds the client once per sync and closes the WebClient when the run finishes, so it never piles up clients per page.

gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/infra/service_provider/AmUserSyncRunnerImpl.java
```diff
@@ -0,0 +1,76 @@
...
+        this(
+            syncAmUsersUseCase,
+            asyncJobCrudService,
+            Executors.newCachedThreadPool(runnable -> {
+                Thread thread = new Thread(runnable, "am-user-sync");
+                thread.setDaemon(true);
+                return thread;
+            })
+        );
+    }
```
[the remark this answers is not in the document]
sclarkamlab: Done in `3b318ffc19`. Implemented `DisposableBean` and shut the pool down in `destroy()`.

gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/core/am/use_case/SyncAmUsersUseCase.java
```diff
@@ -0,0 +1,132 @@
...
+                batch.add(toCommand(caller.environmentId(), user));
+                if (batch.size() >= BATCH_SIZE) {
+                    entitiesUpserted += flush(caller, batch);
+                }
+            }
+            if (usersFetched >= userPage.totalCount()) {
+                break;
+            }
+            page++;
+        }
```
michel-barret: I'm always a bit affraid with infinit loop... What do you think about move it to an observable (or at least encapsule the fetching in an Iterable)? It can split the fetching logic and business logic in 2 separate part

gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/infra/service_provider/AmUserSyncRunnerImpl.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private final SyncAmUsersUseCase syncAmUsersUseCase;
+    private final AsyncJobCrudService asyncJobCrudService;
+    private final ExecutorService executor;
+
+    public AmUserSyncRunnerImpl(SyncAmUsersUseCase syncAmUsersUseCase, AsyncJobCrudService asyncJobCrudService) {
+        this(
+            syncAmUsersUseCase,
+            asyncJobCrudService,
+            Executors.newCachedThreadPool(runnable -> {
```
michel-barret: maybe not a priority but do you think that is doable with virtual thread?

gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/core/am/use_case/SyncAmUsersUseCase.java
sclarkamlab: Yeah the while(true) nagged at me too. Pulled the paging into a private Iterable<AmUser> so execute() is just a for-each over users now, fetching and the business logic are split out. Termination is still bounded the same way (empty page or fetched >= totalCount). Skipped rxjava, the iterable was enough here. Done in 5c2eecfddd.

gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/infra/service_provider/AmUserSyncRunnerImpl.java
sclarkamlab: Good shout, switched it to a virtual-thread-per-task executor. A run is one long IO-bound task so it parks cheaply while waiting on AM, and virtual threads are daemon by default so the no-pinning intent of the old setDaemon still holds. Done in 5c2eecfddd.

michel-barret (approved):

michel-barret (approved):

PR #17267, opened 2026-05-30
feat(authz-ui): entity detail panel + entities page parity

podlesrafal:
## Summary
- **Entity detail panel** (right Sheet, opened by clicking an entity name) with four tabs:
  - **Overview** — attributes table (Name/Type/Value, types inferred) + provenance (source, imported/created/updated).
  - **Relationships** — parents + "referenced by" + "contains N <type>".
  - **Policies** — policies whose target is this entity.
  - **GAPL shape** — the canonical document the PDP evaluates (`uid`/`attrs`/`parents`) + **Copy JSON**.
  - Header chips: `N attrs · N parents · N referenced by · N policies`; Edit button for local entities; copy-uid.
- **Table parity:** Relationships and Policies columns, copy-to-clipboard on Entity ID, Source badge.
- **KPI:** Policy-Linked tile.
- **Settings menu:** Open entities.json (client-side export of the collection as canonical GAPL shapes).
- No backend changes — data from `useAllEntities` + `usePolicies`. Pure builders (`entity-gapl-shape`, `entity-relationships`, `entities-json`) are unit-tested.

## Design notes
- Strong graphene compliance: graphene `Table`/`Sheet`/`Tabs`/`DropdownMenu`/`Badge`/`Button` (the only raw HTML is the inline name `<button>` cell and the `<pre>` JSON block — graphene has no primitive for either).
- Known, bounded limitations (consistent with existing patterns): the Policies/Relationships columns compute per-row over the loaded page (fine at page sizes ≤ 100); `usePolicies` fetches one page (`MAX_PER_PAGE=100`), so the Policy-Linked KPI undercounts in environments with >100 policies — same limitation the policy service already documents.

## Test plan
- [x] Pure builder unit tests (GAPL shape, relationships/policy matching, entities.json)
- [x] Tab component tests + EntityDetailSheet tests (tabs, chips, Edit gating, copy)
- [x] EntitiesPage tests (open panel, relationships column, Policy-Linked KPI, settings → entities.json)
- [x] tsc, vitest (455 tests), nx lint, nx format — green


gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/shared/entity-relationships.ts
```diff
@@ -0,0 +1,39 @@
...
+    for (const child of referencedBy(entity, all)) {
+        counts.set(child.uid.type, (counts.get(child.uid.type) ?? 0) + 1);
+    }
+    return Array.from(counts, ([type, count]) => ({ type, count })).sort((a, b) => a.type.localeCompare(b.type));
+}
+
+/** Policies whose target entityId equals this entity's canonical uid. */
+export function policiesFor(entity: EntityInstance, policies: readonly PolicyResponse[]): PolicyResponse[] {
+    const uid = formatEntityUid(entity.uid);
+    return policies.filter(p => p.entityId === uid);
```
sclarkamlab: This filters on `p.entityId`, but `PolicyResponse` doesn't have that field, it's `target.id`. `entityId` only lives on the canonical wire type and `adaptPolicyResponse` maps it into `target`. So this always returns empty, which silently kills the Policy-Linked KPI, the policies column, the header chip and the Policies tab in prod.

```ts
return policies.filter(p => p.target?.id === uid);
```

Tests don't catch it because the `policy()` mocks fabricate `entityId`/`kind` that aren't on the real type (and drop the required fields), so they validate the bug. Worth fixing those to the real shape and re-running tsc, fairly sure `p.entityId` is a TS2339 as it stands.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/entity-detail/EntityDetailSheet.tsx
```diff
@@ -0,0 +1,165 @@
...
+            refs: referencedBy(entity, allEntities).length,
+            policies: policiesFor(entity, policies).length,
+        };
+    }, [entity, allEntities, policies]);
+
+    async function copyUid() {
+        try {
+            await navigator.clipboard?.writeText(entityUid);
+            setCopied(true);
+        } catch {
```
sclarkamlab, line 76: No test for this clipboard rejection path. `EntityGaplShapeTab` has a `does not throw when clipboard unavailable` test on the same try/catch, worth mirroring it here.

gravitee-gamma/gravitee-gamma-module-authz/src/main/ui/features/policy-structure/entity-detail/EntityDetailSheet.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+export function EntityDetailSheet({ entity, allEntities, policies, open, onOpenChange, onEdit }: EntityDetailSheetProps) {
+    const [tab, setTab] = useState<DetailTab>('overview');
+    const [copied, setCopied] = useState(false);
+
+    // Land on Overview each time a different entity is opened.
+    const entityUid = entity ? formatEntityUid(entity.uid) : '';
+    useEffect(() => {
+        setTab('overview');
+        setCopied(false);
```
sclarkamlab, line 53: Nothing covers this reset. Worth a test that the copied indicator clears when you open a different entity, easy regression otherwise.

sclarkamlab (approved):

PR #17271, opened 2026-05-31
feat(portal-nav): increase navigation panel default width from 350px …

adam-rosowicz:
Increases the default width of the navigation tree panel from 350px to 400px. The extra 50px is taken equally from the editor and preview columns inside the markdown editor, giving the navigation tree more room without any other layout changes.

image


aqibmohammadkhan (approved):

aqibmohammadkhan:
Please do a rebase with master before merging

PR #17292, opened 2026-06-01
fix(GKO-2863): return 400 when primary owner is a group member

benoitgravitee:
## Issue

https://gravitee.atlassian.net/browse/GKO-2863

## Summary

- `ImportGroupCRDUseCase` now throws `ValidationDomainException` on severe validation errors (same pattern as other CRD import use cases), so the Automation API returns **HTTP 400** instead of **500** when the org primary owner is listed as a group member.
- Added a regression test reproducing the `memory` / `admin` member case from the ticket.

## Test plan

- [x] `mvn -f gravitee-apim-rest-api/gravitee-apim-rest-api-service/pom.xml test -Dtest=ImportGroupCRDUseCaseTest`
- [x] `PUT /automation/organizations/DEFAULT/environments/DEFAULT/groups` with primary owner in `members` returns **400** and a message containing `can not change the role of primary owner`

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/group/use_case/ImportGroupCRDUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/group/use_case/ImportGroupCRDUseCase.java
@@ -17,7 +17,7 @@ package io.gravitee.apim.core.group.use_case;
 
 import io.gravitee.apim.core.UseCase;
 import io.gravitee.apim.core.audit.model.AuditInfo;
-import io.gravitee.apim.core.exception.TechnicalDomainException;
+import io.gravitee.apim.core.exception.ValidationDomainException;
 import io.gravitee.apim.core.group.crud_service.GroupCrudService;
 import io.gravitee.apim.core.group.domain_service.ValidateGroupCRDDomainService;
 import io.gravitee.apim.core.group.model.crd.GroupCRDSpec;
@@ -65,11 +65,16 @@ public class ImportGroupCRDUseCase {
         validationResult
             .severe()
             .ifPresent(errors -> {
-                throw new TechnicalDomainException(errors.iterator().next().getMessage());
+                throw new ValidationDomainException(
+                    String.format(
+                        "Unable to import because of errors [%s]",
+                        String.join(",", errors.stream().map(Validator.Error::getMessage).toList())
+                    )
+                );
             });
 
         var warnings = validationResult.warning().orElseGet(List::of);
-        var sanitizedInput = validationResult.value().orElseThrow(() -> new TechnicalDomainException("Unable to sanitize CRD spec"));
+        var sanitizedInput = validationResult.value().orElseThrow(() -> new ValidationDomainException("Unable to sanitize CRD spec"));
 
         var status = queryService
             .findById(sanitizedInput.spec.getId())
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/group/use_case/ImportGroupCRDUseCaseTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/group/use_case/ImportGroupCRDUseCaseTest.java
@@ -15,7 +15,10 @@
  */
 package io.gravitee.apim.core.group.use_case;
 
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.SoftAssertions.*;
+import static org.mockito.ArgumentMatchers.any;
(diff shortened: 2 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/group/use_case/ImportGroupCRDUseCase.java
```diff
@@ -65,11 +65,16 @@ public Output execute(Input input) {
...
+                    String.format(
+                        "Unable to import because of errors [%s]",
+                        String.join(",", errors.stream().map(Validator.Error::getMessage).toList())
+                    )
+                );
             });
 
         var warnings = validationResult.warning().orElseGet(List::of);
-        var sanitizedInput = validationResult.value().orElseThrow(() -> new TechnicalDomainException("Unable to sanitize CRD spec"));
+        var sanitizedInput = validationResult.value().orElseThrow(() -> new ValidationDomainException("Unable to sanitize CRD spec"));
```
carlos-andres-osorio, line 77: The guard here was changed from `TechnicalDomainException` to `ValidationDomainException` but has no test covering this path. The path is unreachable through the real `ValidateGroupCRDDomainServiceImpl` (which always populates the value), but it is reachable via a mock — and when reached, it does throw the correct `ValidationDomainException`.

Without a test, a future refactor could silently revert this back to a 500, which was the original bug.

## Evidence

```java
@Test
void sanitize_guard_throws_ValidationDomainException_when_value_is_absent() {
    var validationService = mock(ValidateGroupCRDDomainService.class);
    when(validationService.validateAndSanitize(any())).thenReturn(Validator.Result.empty());

    var cut = new ImportGroupCRDUseCase(
        validationService,
        new GroupQueryServiceInMemory(),
        new GroupCrudServiceInMemory(),
        new CRDMembersDomainServiceInMemory()
    );

    var auditInfo = AuditInfoFixtures.anAuditInfo("org", "env", "user");
    var spec = GroupCRDSpec.builder().id("group-id").name("test").build();

    assertThatThrownBy(() -> cut.execute(new ImportGroupCRDUseCase.Input(auditInfo, spec)))
        .isInstanceOf(ValidationDomainException.class)
        .hasMessage("Unable to sanitize CRD spec");
}
```

Result: `Tests run: 1, Failures: 0, Errors: 0, Skipped: 0` — BUILD SUCCESS.

benoitgravitee: done

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/group/use_case/ImportGroupCRDUseCase.java
```diff
@@ -65,11 +65,16 @@ public Output execute(Input input) {
...
 
         var status = queryService
```
carlos-andres-osorio, line 79: The update branch here (`findById(...).map(existing -> update(...))`) — the primary success path when the group already exists — has no test coverage. Both new tests exercise the validation/rejection paths only. A regression that broke group updates (e.g. creating a duplicate instead of updating the existing group) wouldn't be caught by CI.

This is a pre-existing gap rather than something this PR introduced, so feel free to treat it as out of scope — flagging it since the test class is already being touched here.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/group/use_case/ImportGroupCRDUseCaseTest.java
```diff
@@ -15,7 +15,11 @@
  */
 package io.gravitee.apim.core.group.use_case;
 
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.SoftAssertions.*;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
```
carlos-andres-osorio, line 22: Nit: `import static org.mockito.Mockito.when;` is added here but the same import already exists on the next line, so this one is redundant. It compiles, but most spotless/checkstyle no-duplicate-import rules flag it.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/group/use_case/ImportGroupCRDUseCase.java
benoitgravitee: Won't do, out of scope

carlos-andres-osorio (approved):

PR #17296, opened 2026-06-01
fix(portal): refine ownership transfer dialog UX

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/APIM-14242

## Description

Addresses the ownership transfer dialog review comments:

1. Added spacing between the transfer method icon and label.
2. Updated the role field label to “New Role for Current Primary Owner”.

Zrzut ekranu 2026-06-1 o 16 30 06

4. Display the selected target using the user cell, so the email is visible when provided by the backend response.
Zrzut ekranu 2026-06-1 o 16 30 47

5. The dialog now searches application members independently instead of using the currently visible members table rows, so results are no longer limited by table pagination.

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.scss
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.scss
@@ -39,6 +39,10 @@
     width: 100%;
   }
 
+  &__selected-target {
+    display: flex;
+  }
+
   &__loading {
     display: flex;
     align-items: center;
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.ts
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.ts
@@ -30,6 +30,7 @@ import { ButtonToggleGroupComponent } from '../../../../components/button-toggle
 import { ButtonToggleOptionComponent } from '../../../../components/button-toggle-group/button-toggle-option.component';
 import { UserCellComponent, UserCellVM } from '../../../../components/user-cell/user-cell.component';
 import { isAssignableApplicationRole } from '../../../../entities/application/application';
+import { Member, MembersResponse } from '../../../../entities/member/member';
 import { User, UsersResponse } from '../../../../entities/user/user';
 import { ApplicationService } from '../../../../services/application.service';
 import { MembershipService } from '../../../../services/membership.service';
@@ -37,10 +38,11 @@ import { UsersService } from '../../../../services/users.service';
 
 const APPLICATION_MEMBER_METHOD = 'APPLICATION_MEMBER';
 const OTHER_USER_METHOD = 'OTHER_USER';
+const APPLICATION_MEMBER_SEARCH_PAGE_SIZE = 20;
 
 type TransferOwnershipMethod = typeof APPLICATION_MEMBER_METHOD | typeof OTHER_USER_METHOD;
 
-export interface ApplicationTransferOwnershipMemberOption {
+interface ApplicationTransferOwnershipMemberOption {
   id: string;
   reference?: string;
   displayName: string;
@@ -52,7 +54,6 @@ export interface ApplicationTransferOwnershipMemberOption {
 export interface ApplicationTransferOwnershipDialogData {
   applicationId: string;
   currentOwnerId?: string;
-  members: ApplicationTransferOwnershipMemberOption[];
 }
 
 interface TransferOwnershipTarget extends ApplicationTransferOwnershipMemberOption {
@@ -98,7 +99,11 @@ export class ApplicationTransferOwnershipDialogComponent {
   readonly selectedTarget = signal<TransferOwnershipTarget | null>(null);
 
   private readonly applicationMemberSearchValue = toSignal(
-    this.applicationMemberControl.valueChanges.pipe(startWith(this.applicationMemberControl.value)),
+    this.applicationMemberControl.valueChanges.pipe(
+      debounceTime(300),
+      distinctUntilChanged(),
+      startWith(this.applicationMemberControl.value),
+    ),
     {
       initialValue: this.applicationMemberControl.value,
     },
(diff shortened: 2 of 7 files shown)


gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.ts
```diff
@@ -30,17 +30,19 @@ import { ButtonToggleGroupComponent } from '../../../../components/button-toggle
...
 import { isAssignableApplicationRole } from '../../../../entities/application/application';
+import { Member, MembersResponse } from '../../../../entities/member/member';
 import { User, UsersResponse } from '../../../../entities/user/user';
 import { ApplicationService } from '../../../../services/application.service';
 import { MembershipService } from '../../../../services/membership.service';
 import { UsersService } from '../../../../services/users.service';
 
 const APPLICATION_MEMBER_METHOD = 'APPLICATION_MEMBER';
 const OTHER_USER_METHOD = 'OTHER_USER';
+const APPLICATION_MEMBER_SEARCH_PAGE_SIZE = 20;
```
aqibmohammadkhan, line 41: question:  is this limit set elsewhere or should we consider either pagination or infinite scroll in autocomplete

gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.scss
```diff
@@ -39,6 +39,16 @@
     width: 100%;
   }
 
+  &__method-option {
+    ::ng-deep .button-toggle-option .material-icons {
```
aqibmohammadkhan: please remove ng-deep. we should never use ng-deep.

gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.ts
aqibmohammadkhan, line 1: Dialog now triggers API call on mount (empty query "")

- Could be optimized to only search when query length > 0
- Current implementation searches all members initially

gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.ts
```diff
@@ -30,17 +30,19 @@ import { ButtonToggleGroupComponent } from '../../../../components/button-toggle
...
 
 type TransferOwnershipMethod = typeof APPLICATION_MEMBER_METHOD | typeof OTHER_USER_METHOD;
 
-export interface ApplicationTransferOwnershipMemberOption {
+interface ApplicationTransferOwnershipMemberOption {
```
aqibmohammadkhan, line 45: any specific reason for this interface change, from exported to internal?

gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.ts
jaroslawlakomy-grav: The limit is local to the Application member autocomplete in this dialog. For the “Other user” autocomplete we rely on UsersService.searchUsersForApplicationMembership, which defaults to size = 20 and is also used by the add-members dialog.
Adding infinite scroll to autocomplete field would therefore be inconsistent and larger than this visual-improvement scope. The current pattern is to return the first 20 matches and let the user narrow the query.

gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.scss
jaroslawlakomy-grav: I agree we should remove ::ng-deep.
There are three possible options:
1. Update ButtonToggleOptionComponent and handle icon/text spacing in the shared component.
The component currently renders the icon and label internally:

```
<button class="button-toggle-option" ...>
  <span class="material-icons">{{ icon() }}</span>
  {{ label() }}
</button>
```

Because the icon lives inside the shared component template, the dialog cannot style it directly without breaking style encapsulation. That is why the current implementation used ::ng-deep. The downside of changing the shared component is that it may affect all ButtonToggleOptionComponent usages.
2. Remove the local override and keep the default spacing.
3. Keep using ::ng-deep, but I agree we should avoid that.
My preference is option 2: remove the local override and keep the default spacing. If the spacing still needs visual refinement, I think it would be better to track it separately as a dedicated private bug

gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.ts
jaroslawlakomy-grav: Yes, this was intentional.
Previously ApplicationTransferOwnershipMemberOption was exported because it was part of the dialog input contract: the parent component built a members: ApplicationTransferOwnershipMemberOption[] array and passed it through ApplicationTransferOwnershipDialogData.
With the current implementation, the dialog no longer receives members from the parent. It searches application members itself through MembershipService.searchApplicationMembers(...), so the parent only passes applicationId and currentOwnerId.
Because of that, ApplicationTransferOwnershipMemberOption is now only an internal base shape used by TransferOwnershipTarget, and it no longer needs to be exported.

gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.ts
jaroslawlakomy-grav: refactored to execute query only if query is not  empty

gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-transfer-ownership-dialog/application-transfer-ownership-dialog.component.scss
jaroslawlakomy-grav: removed ::ng-deep - icon changed to "person" - looks better
Zrzut ekranu 2026-06-2 o 13 47 01

aqibmohammadkhan (approved):
LGTM

jaroslawlakomy-grav (commented):
Master is locked due to Gamma, some gamma tests are failing, needs to be rebased on latest master as soon as master is unlocked

PR #17304, opened 2026-06-01
fix(portal): normalize OAS 3.1 type arrays before passing spec to Swgger UI to fix required-field validation error

araffass:
## Issue

https://gravitee.atlassian.net/browse/APIM-14231

## Description

Swagger UI's form validator expects type to be a string (OAS 3.0 style). When it encounters an OAS 3.1 type array (e.g. "type": ["integer", "string"]), it cannot resolve the parameter type and incorrectly raises "Required field is not provided", blocking the request.                                                                                                                                                                                                                                                         

Added a normalizeTypeArrays pre-processing step in both portal Swagger components that recursively walks the parsed spec before handing it to Swagger UI:

## Additional context

**Before Fix:**


https://github.com/user-attachments/assets/63ef93a5-45b3-4b98-a0b3-62fe827f0d02


**After  Fix:**


https://github.com/user-attachments/assets/dac9c131-6955-41c8-9e54-ec9be5a7dcf8


<!-- Add any other context about the PR here -->
<!-- It can be links to other PRs or docs or drawing -->
<!-- Or reproduction steps in case of bug fix -->

### DIFF
--- a/gravitee-apim-portal-webui-next/src/components/page/page-swagger/page-swagger.component.ts
+++ b/gravitee-apim-portal-webui-next/src/components/page/page-swagger/page-swagger.component.ts
@@ -21,6 +21,10 @@ import { readYaml } from '../../../app/helpers/yaml-parser';
 import { Page } from '../../../entities/page/page';
 import { CurrentUserService } from '../../../services/current-user.service';
 
+const OAS_SCHEMA_TYPES = new Set(['null', 'boolean', 'object', 'array', 'number', 'string', 'integer']);
+// Priority order: most permissive first, so we pick the least-restrictive type from the array
+const OAS_TYPE_PRIORITY = ['string', 'number', 'integer', 'boolean', 'array', 'object'];
+
 @Component({
   selector: 'app-page-swagger',
   standalone: true,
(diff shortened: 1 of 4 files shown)


gravitee-apim-portal-webui-next/src/components/page/page-swagger/page-swagger.component.ts
```diff
@@ -94,12 +94,42 @@ export class PageSwaggerComponent implements OnChanges {
...
+    }
+    const record = obj as Record<string, unknown>;
+    const result: Record<string, unknown> = {};
+    for (const key of Object.keys(record)) {
+      if (key === 'type' && Array.isArray(record[key])) {
+        const typeArray = record[key] as string[];
+        const nonNullTypes = typeArray.filter(t => t !== 'null');
+        result[key] = nonNullTypes.length === 1 ? nonNullTypes[0] : 'string';
+        if (typeArray.includes('null')) {
+          result['nullable'] = true;
```
ankita-gupta21, line 128: What happens when schema type is ['integer','number']?
With current logic, this seems to become type: 'string' 🤔

gravitee-apim-portal-webui/src/app/components/gv-page-swaggerui/gv-page-swaggerui.component.ts
```diff
@@ -110,12 +110,40 @@ export class GvPageSwaggerUIComponent implements OnInit, OnDestroy {
...
+        const nonNullTypes = (obj[key] as string[]).filter((t: string) => t !== 'null');
+        result[key] = nonNullTypes.length === 1 ? nonNullTypes[0] : 'string';
+        if ((obj[key] as string[]).includes('null')) {
+          result['nullable'] = true;
+        }
+      } else {
+        result[key] = this.normalizeTypeArrays(obj[key]);
+      }
+    }
+    return result;
```
ankita-gupta21, line 148: We normalize any key named type recursively. Do we only want this inside OpenAPI Schema Objects?
Could this affect non-schema objects that also have a type field?
Example:
```
{
  "x-ui-config": {
    "type": ["compact", "expanded"]
  }
}
```
If this is not a schema field, current recursion may still rewrite it (e.g. to "string"), which could change custom extension behavior.

gravitee-apim-portal-webui-next/src/components/page/page-swagger/page-swagger.component.ts
araffass: I though of that before, but as you also mentioned that i fixed it with a priority list (OAS_TYPE_PRIORITY). Instead of always falling back to 'string' for any multi-type array, we now pick the most permissive OAS type that actually appears in the array which looks better to me

gravitee-apim-portal-webui/src/app/components/gv-page-swaggerui/gv-page-swaggerui.component.ts
araffass: Good point, i didn't think of that use case, i addded a  OAS_SCHEMA_TYPES guard. We now only rewrite type when every value in the array is a known OAS primitive type (null, boolean, object, array, number, string, integer). Extension values like you mentioned ("compact" or "expanded") aren't OAS types, so they pass through untouched.

ankita-gupta21 (approved):

PR #17306, opened 2026-06-01
feat(portal): add core domain, use cases and HRID-to-UUID factory

konrad-krol-incubly:
## Issue

## Description

Context: First slice of the next-gen Portal feature for the Automation API — introduces the domain layer of the Portal resource so reviewers can absorb the model, contracts, and use cases in isolation before persistence or REST wiring lands.

Scope:
  - Adds the Portal core model (entity + value-object id + not-found exception) and its CRUD service interface.
  - Introduces four use cases covering create-or-update, get, delete, and default-portal seeding, fully covered by unit tests backed by an in-memory CRUD
  service.
  - Extends the HRID-to-UUID factory with a deterministic portal() derivation used by the seed.
  - No Spring wiring yet — the use cases ship without the framework annotation so the module boots green until persistence arrives in PR 2.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/exception/PortalNotFoundException.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/exception/PortalNotFoundException.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal.exception;
+
+import io.gravitee.apim.core.exception.NotFoundDomainException;
+
+/**
+ * @author GraviteeSource Team
+ */
+public class PortalNotFoundException extends NotFoundDomainException {
+
+    public PortalNotFoundException(String portalId) {
+        super("Portal [ " + portalId + " ] not found", portalId);
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/Portal.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/Portal.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal.model;
+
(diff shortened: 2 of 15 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/exception/PortalNotFoundException.java
```diff
@@ -13,12 +13,16 @@
...
-import org.springframework.data.mongodb.repository.MongoRepository;
+import io.gravitee.apim.core.exception.NotFoundDomainException;
 
 /**
  * @author GraviteeSource Team
  */
-public interface AmConnectionMongoRepository extends MongoRepository<AmConnectionMongo, String> {}
+public class PortalNotFoundException extends NotFoundDomainException {
+
+    public PortalNotFoundException(String portalId) {
```
dev-marek, line 25: We should use `PortalId` so it won't mess with traditional `String message` constructor

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/Portal.java
```diff
@@ -13,22 +13,38 @@
...
-import lombok.Setter;
 
-/**
- * @author GraviteeSource Team
- */
 @Getter
-@Setter
-@JsonIgnoreProperties(ignoreUnknown = true)
-public class KafkaPortRouting {
+@Builder(toBuilder = true)
```
dev-marek: 1. Please avoid excessive usage of lombok in domain model. Builder shouldn't be a default way of building domain objects.
2. There should be 2 factory methods in this class: `create(String evnId, String orgId, String name)` creating a new instance with random id and `of(PortalId id, String envId, ...)` for deserialization
3. I do recommend a private/protected/package-private constructor

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/PortalId.java
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal.model;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+import jakarta.annotation.Nonnull;
+import java.util.UUID;
+
+public class PortalId {
```
dev-marek: I see we have currently different approaches in our codebase, but target template for ID value objects would be 100% a record. Please refer to following draft:

```java
package io.gravitee.apim.core.portal.model;

import com.fasterxml.jackson.annotation.JsonValue;
import jakarta.annotation.Nonnull;
import java.util.UUID;

public record PortalId(@Nonnull UUID id) {


    // 2. Factory Methods work perfectly in records
    public static PortalId random() {
        return new PortalId(UUID.randomUUID());
    }

    public static PortalId of(String value) {
        return new PortalId(UUID.fromString(value));
    }

    // 3. Jackson serialization and custom string formatting
    @JsonValue
    @Override
    public String toString() {
        return id.toString();
    }
}
```

dev-marek (changes requested):
Please do some minor changes, but overall nice job

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/exception/PortalNotFoundException.java
konrad-krol-incubly: It's not possible, because it would break `OnionArchitectureTest.should_respect_onion_architecture`, as exception package isn't a declared onion layer.

To use `PortalId`, we'd need to either change the onion test config (cross-cutting) or move the exception into ..model.. (unusual).

Same approach is used for:
- `io.gravitee.apim.core.portal_page.exception.PortalNavigationItemNotFoundException`
- `io.gravitee.apim.core.portal_page.exception.PageContentNotFoundException`

I guess here authors were dealing with the same concern.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/PortalId.java
konrad-krol-incubly: adjusted

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/Portal.java
konrad-krol-incubly: adjusted

dev-marek (approved):

PR #17311, opened 2026-06-01
feat(gamma-authz): import mcp tools from catalog and use them in policy builder

podlesrafal:
## Summary
- Import MCP tools alongside their server when importing from the AIM catalog. Each tool is stored as a `RESOURCE` entity (`mcptool.<slug>`) linked to its server via `parents`, carrying `_displayName`/`_catalogId`/`description` attributes.
- Surface imported MCP tools in the MCP policy builder, in both the **action** and **resource** selectors, as a dedicated **"MCP Tools"** group shown next to the existing generic action/resource options (nothing removed). Tools are listed env-wide (all tools, any server), not scoped to the selected target.
- Extract the action/resource option-building logic out of `ServicePolicyPage` into pure, unit-tested helpers (`action-options.ts`, `resource-options.ts`).
- Backend: add a parent-aware cascade delete so deleting an MCP server also deletes its imported tools, even though the tool's `entityId` prefix (`mcptool.`) differs from the server's.

## Implementation notes
- Tool tokens are canonical (`MCPTool::"<slug>"` for resources, `Action::"<name>"` for actions), so they round-trip through the existing GAPL emit/parse path; the "MCP Tools" group label is display-only and never enters the token.
- `useEntities` gains an optional `enabled` filter so the `mcptool.*` fetch only runs on the MCP policy page.
- `findByParent` added to `AuthzEntityRepository` (default in-memory impl + Mongo `parents` query).

## Test plan
- [x] `vitest` — 60/60 pass in `policy-management` (incl. new `action-options` / `resource-options` suites)
- [x] `tsc --noEmit` clean, `eslint` clean, `prettier --check` clean
- [x] Java `AuthzEntityServiceImplTest` — 31/31 pass, incl. new parent-aware cascade-delete test
- [x] Manual: in the MCP policy builder, imported tools appear in both the action and resource selectors under "MCP Tools"; selecting one emits `Action::"<tool>"` / `resource == MCPTool::"<tool>"`

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/api/AuthzEntityRepository.java
+++ b/gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/api/AuthzEntityRepository.java
@@ -60,6 +60,13 @@ public interface AuthzEntityRepository {
         return result;
     }
 
+    default List<AuthzEntity> findByParent(String environmentId, String parentEntityId) {
+        if (parentEntityId == null || parentEntityId.isBlank()) {
+            return List.of();
+        }
+        return findAll(environmentId).stream().filter(e -> e.parents().contains(parentEntityId)).toList();
+    }
+
     default List<AuthzEntity> findByAnyEntityIdPrefix(String environmentId, Collection<String> prefixes) {
         if (prefixes == null || prefixes.isEmpty()) {
             return List.of();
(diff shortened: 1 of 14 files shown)


gcusnieux:
LGTM. Checked the AIM side too: the `/catalog/mcp-tools?parentId=` endpoint, the `mcp-tool` kind, and the `entityId` / `definition` fields all match what this consumes.

One small non-blocking thing: a tool ends up keyed differently in the two selectors. A camelCase tool like `getTeamworkGraphContext` becomes `Action::"getTeamworkGraphContext"` (raw name) as an action but `MCPTool::"getteamworkgraphcontext"` (slug) as a resource, because `slugify` lowercases. That's actually right, the action has to match the real tool name at runtime while the resource points at the catalog entity, but it reads like a bug at first glance. A one-line comment in `action-options.ts` / `resource-options.ts` would save the next reader from "fixing" it into a mismatch.

gcusnieux (approved):

PR #17329, opened 2026-06-02
fix(gamma-platform-ui): archived apps navigation, groups field UX, metadata key, remove apps tooltip

gaurav-gravitee:
## Issue

## Description

Summary

- Make archived applications clickable in the Applications list table (navigate to application details).
- Fix Groups combobox on application creation:
Don’t show “No groups available” when groups exist.
Prevent free-typing (selection/deselection only).
- Remove the Applications header tooltip (“Consumer applications that subscribe…”).
- Hide Metadata “Key” field during metadata creation on Notification settings (key still visible in edit dialog).

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/list/ApplicationListTable.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/list/ApplicationListTable.tsx
@@ -72,14 +72,15 @@ function buildActiveColumns(navigate: ReturnType<typeof useNavigate>): DataTable
             cell: ({ row }: ColCell<ApplicationListItem>) => {
                 const name = row.original.name;
                 return (
-                    <button
+                    <Button
                         type="button"
-                        className="flex items-center gap-2 font-medium text-left hover:underline"
+                        variant="link"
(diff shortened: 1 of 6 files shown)


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/list/ApplicationListTable.tsx
```diff
@@ -136,10 +137,14 @@ function buildArchivedColumns(
             cell: ({ row }: ColCell<ApplicationListItem>) => {
                 const name = row.original.name;
                 return (
-                    <div className="flex items-center gap-2 font-medium">
+                    <button
```
jourdiw: Can you use a Graphene button instead of a native button?

jourdiw (approved):
LGTM

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/list/ApplicationListTable.tsx
gaurav-gravitee: updated the button to graphene <Button>

PR #17334, opened 2026-06-02
feat: delete pending application invitations

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/APIM-14239

## Description

Implements the first part of invitation mutations: deleting pending application invitations from the Developer Portal.
Changes:
- Adds DELETE /applications/{applicationId}/invitations/{invitationId},
- Adds DeleteApplicationInvitationUseCase and domain service for application-scoped deletion,
- Validates that the application exists in the current environment before deleting.
- Ensures the invitation exists and belongs to the requested application.
- Returns 404 when the application, invitation, or application/invitation relation is invalid.
- Extends invitation CRUD service with application invitation lookup by ID.

## Out of scope
- Invitation role update.
- Resend invitation.
- Notification handling.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/invitation/exception/ApplicationInvitationNotFoundException.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/invitation/exception/ApplicationInvitationNotFoundException.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/invitation/exception/ApplicationInvitationNotFoundException.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.invitation.exception;
+
+import io.gravitee.apim.core.exception.NotFoundDomainException;
+
+public class ApplicationInvitationNotFoundException extends NotFoundDomainException {
+
+    public ApplicationInvitationNotFoundException(String invitationId) {
```
dev-marek, line 22: What about passing InvitationId instead of String?

dev-marek (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/invitation/exception/ApplicationInvitationNotFoundException.java
jaroslawlakomy-grav: tested, will cause "Architecture Violation"

PR #17335, opened 2026-06-02
fix(GKO-2934): remove EDGE from Automation API type

benoitgravitee:
## Issue
https://gravitee.atlassian.net/browse/GKO-2934

## Summary
- Remove `EDGE` from Automation API OpenAPI `ApiType` enum.
- Fail fast if `EDGE` is encountered during ApiType mapping.

## Test plan
- [x] `mvn -pl gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest test`

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/ApiMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/ApiMapper.java
@@ -52,7 +52,9 @@ import java.util.Map;
 import java.util.Objects;
 import org.mapstruct.Mapper;
 import org.mapstruct.Mapping;
+import org.mapstruct.MappingConstants;
 import org.mapstruct.MappingTarget;
+import org.mapstruct.ValueMapping;
 import org.mapstruct.factory.Mappers;
 
 /**
@@ -62,6 +64,11 @@ import org.mapstruct.factory.Mappers;
 @Mapper(uses = { DateMapper.class, OriginContextMapper.class, ServiceMapper.class })
 public interface ApiMapper {
     ApiMapper INSTANCE = Mappers.getMapper(ApiMapper.class);
+
+    @ValueMapping(source = "EDGE", target = MappingConstants.THROW_EXCEPTION)
+    io.gravitee.apim.rest.api.automation.model.ApiType map(io.gravitee.rest.api.management.v2.rest.model.ApiType type);
+
+    io.gravitee.rest.api.management.v2.rest.model.ApiType map(io.gravitee.apim.rest.api.automation.model.ApiType type);
 
     @Mapping(target = "listeners", expression = "java(mapApiV4SpecListeners(apiV4Spec))")
     @Mapping(target = "plans", expression = "java(mapApiV4SpecPlans(apiV4Spec))")
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/SharedPolicyGroupMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/SharedPolicyGroupMapper.java
@@ -15,12 +15,15 @@
  */
 package io.gravitee.apim.rest.api.automation.mapper;
 
+import io.gravitee.apim.core.plugin.model.FlowPhase;
 import io.gravitee.apim.core.shared_policy_group.model.SharedPolicyGroup;
 import io.gravitee.apim.core.shared_policy_group.model.SharedPolicyGroupCRDStatus;
 import io.gravitee.apim.rest.api.automation.model.Errors;
 import io.gravitee.apim.rest.api.automation.model.LegacySharedPolicyGroupSpec;
+import io.gravitee.apim.rest.api.automation.model.SharedPolicyGroupApiType;
 import io.gravitee.apim.rest.api.automation.model.SharedPolicyGroupState;
 import io.gravitee.apim.rest.api.automation.model.StepV4;
+import io.gravitee.definition.model.v4.ApiType;
 import io.gravitee.definition.model.v4.flow.step.Step;
 import io.gravitee.rest.api.management.v2.rest.mapper.ConfigurationSerializationMapper;
 import io.gravitee.rest.api.management.v2.rest.mapper.DateMapper;
@@ -40,10 +43,19 @@ import org.mapstruct.factory.Mappers;
 public interface SharedPolicyGroupMapper {
     SharedPolicyGroupMapper INSTANCE = Mappers.getMapper(SharedPolicyGroupMapper.class);
 
-    @ValueMapping(source = "EDGE", target = MappingConstants.THROW_EXCEPTION)
-    io.gravitee.apim.rest.api.automation.model.ApiType map(io.gravitee.definition.model.v4.ApiType apiType);
+    ApiType mapApiType(SharedPolicyGroupApiType apiType);
 
-    io.gravitee.definition.model.v4.ApiType map(io.gravitee.apim.rest.api.automation.model.ApiType apiType);
+    FlowPhase mapFlowPhase(io.gravitee.apim.rest.api.automation.model.FlowPhase phase);
+
+    @ValueMapping(source = "PROXY", target = "PROXY")
+    @ValueMapping(source = "MESSAGE", target = "MESSAGE")
+    @ValueMapping(source = MappingConstants.ANY_REMAINING, target = MappingConstants.THROW_EXCEPTION)
+    SharedPolicyGroupApiType mapSharedPolicyGroupApiType(ApiType apiType);
+
+    @ValueMapping(source = "ENTRYPOINT_CONNECT", target = MappingConstants.THROW_EXCEPTION)
+    @ValueMapping(source = "INTERACT", target = MappingConstants.THROW_EXCEPTION)
+    @ValueMapping(source = MappingConstants.ANY_REMAINING, target = MappingConstants.THROW_EXCEPTION)
+    io.gravitee.apim.rest.api.automation.model.FlowPhase mapAutomationFlowPhase(FlowPhase phase);
 
     @Mapping(target = "sharedPolicyGroupId", ignore = true)
     @Mapping(target = "originContext", ignore = true)
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/SharedPolicyGroupMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/SharedPolicyGroupMapper.java
@@ -27,7 +27,9 @@ import io.gravitee.rest.api.management.v2.rest.mapper.DateMapper;
 import io.gravitee.rest.api.management.v2.rest.mapper.OriginContextMapper;
 import org.mapstruct.Mapper;
 import org.mapstruct.Mapping;
+import org.mapstruct.MappingConstants;
 import org.mapstruct.MappingTarget;
+import org.mapstruct.ValueMapping;
 import org.mapstruct.factory.Mappers;
 
 /**
@@ -37,6 +39,11 @@ import org.mapstruct.factory.Mappers;
 @Mapper(uses = { ConfigurationSerializationMapper.class, DateMapper.class, OriginContextMapper.class })
 public interface SharedPolicyGroupMapper {
     SharedPolicyGroupMapper INSTANCE = Mappers.getMapper(SharedPolicyGroupMapper.class);
+
+    @ValueMapping(source = "EDGE", target = MappingConstants.THROW_EXCEPTION)
+    io.gravitee.apim.rest.api.automation.model.ApiType map(io.gravitee.definition.model.v4.ApiType apiType);
+
+    io.gravitee.definition.model.v4.ApiType map(io.gravitee.apim.rest.api.automation.model.ApiType apiType);
 
     @Mapping(target = "sharedPolicyGroupId", ignore = true)
     @Mapping(target = "originContext", ignore = true)
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/ApisResourceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/ApisResourceTest.java
@@ -134,6 +134,19 @@ class ApisResourceTest extends AbstractResourceTest {
         }
 
         @Test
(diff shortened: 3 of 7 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/ApiMapper.java
```diff
@@ -63,6 +65,11 @@
 public interface ApiMapper {
     ApiMapper INSTANCE = Mappers.getMapper(ApiMapper.class);
 
+    @ValueMapping(source = "EDGE", target = MappingConstants.THROW_EXCEPTION)
```
carlos-andres-osorio, line 68: No test verifies that this `@ValueMapping` guard throws when `ApiType.EDGE` is passed. If this annotation is accidentally removed or the method is refactored, the mapping silently returns a wrong value with no test signal. The guard is only as reliable as the test that exercises it.

A unit test along the lines of `assertThrows(IllegalArgumentException.class, () -> ApiMapper.INSTANCE.map(io.gravitee.rest.api.management.v2.rest.model.ApiType.EDGE))` would make this protection verifiable.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/SharedPolicyGroupMapper.java
```diff
@@ -38,6 +40,11 @@
 public interface SharedPolicyGroupMapper {
     SharedPolicyGroupMapper INSTANCE = Mappers.getMapper(SharedPolicyGroupMapper.class);
 
+    @ValueMapping(source = "EDGE", target = MappingConstants.THROW_EXCEPTION)
```
carlos-andres-osorio: No test verifies that this `@ValueMapping` guard throws when `ApiType.EDGE` is passed. Same silent-regression risk as in `ApiMapper`: removing or misaligning this annotation would produce a wrong enum mapping with no test to catch it.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/ApiMapper.java
carlos-andres-osorio, line 68: `MappingConstants.THROW_EXCEPTION` generates `throw new IllegalArgumentException(...)` at runtime. `GraviteeAutomationApplication` registers `ValidationDomainMapper`, `HRIDNotFoundMapper`, `ManagementExceptionMapper`, `ConstraintValidationExceptionMapper`, `BadRequestExceptionMapper`, `JsonMappingExceptionMapper`, and `NotAllowedExceptionMapper` — none handle `IllegalArgumentException`. When a client GETs an EDGE-typed API through the Automation REST API, the exception escapes the resource handler unhandled and Jersey falls back to an opaque HTTP 500 with no structured error body.

## Evidence

```java
// Temporary test using real GraviteeAutomationApplication via AbstractResourceTest + JerseySpringTest
@Test
void edge_type_api_get_returns_500_because_IllegalArgumentException_is_unhandled() {
    when(exportApiCRDUseCase.execute(any(ExportApiCRDUseCase.Input.class))).thenReturn(
        new ExportApiCRDUseCase.Output(
            ApiCRDSpec.builder().id("edge-id").crossId("edge-cross-id").hrid(HRID).type("EDGE").build()
        )
    );

    try (var response = rootTarget().path(HRID).request().get()) {
        assertThat(response.getStatus()).isEqualTo(500);
    }
}
```

Jetty server log during the test run:

```
WARN org.eclipse.jetty.server.Response - writeError: status=500,
  message=java.lang.IllegalArgumentException: Unexpected enum constant: EDGE,
  response=...{500, GET .../organizations/gravitee/environments/integration/apis/edge-api-hrid HTTP/1.1}
```

`Tests run: 1, Failures: 0, Errors: 0`

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/SharedPolicyGroupMapper.java
carlos-andres-osorio: Same issue as in `ApiMapper`: `MappingConstants.THROW_EXCEPTION` generates a raw `IllegalArgumentException` which has no registered exception mapper in `GraviteeAutomationApplication`. Any GET on an EDGE-typed shared policy group via `SharedPolicyGroupResource.get` returns an opaque HTTP 500 instead of a structured 4xx error.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/ApiMapper.java
```diff
@@ -63,6 +65,11 @@
...
+    io.gravitee.apim.rest.api.automation.model.ApiType map(io.gravitee.rest.api.management.v2.rest.model.ApiType type);
```
carlos-andres-osorio, line 69: No test exercises valid (non-EDGE) values through this `map(v2.ApiType)` method. The entire happy-path mapping from the v2 REST `ApiType` to `automation.ApiType` is untested. A wrong constant (e.g. `PROXY` mapped to `MESSAGE`) would silently corrupt the `type` field returned to the GKO operator with no test to detect it.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/ApiMapper.java
```diff
@@ -63,6 +65,11 @@
...
+
+    io.gravitee.rest.api.management.v2.rest.model.ApiType map(io.gravitee.apim.rest.api.automation.model.ApiType type);
```
carlos-andres-osorio, line 71: No test exercises this reverse `map(automation.ApiType)` method. A misconfigured constant in the automation → v2 REST direction would silently produce a wrong `type` value in outbound API state, undetected.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/SharedPolicyGroupMapper.java
```diff
@@ -38,6 +40,11 @@
...
+    io.gravitee.apim.rest.api.automation.model.ApiType map(io.gravitee.definition.model.v4.ApiType apiType);
```
carlos-andres-osorio: No test exercises valid (non-EDGE) values through this `map(definition.ApiType)` method. The forward mapping from the definition-layer enum to `automation.ApiType` is entirely untested. A wrong constant would silently corrupt the `type` field on shared policy group reads.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/SharedPolicyGroupMapper.java
```diff
@@ -38,6 +40,11 @@
...
+
+    io.gravitee.definition.model.v4.ApiType map(io.gravitee.apim.rest.api.automation.model.ApiType apiType);
```
carlos-andres-osorio: No test exercises this reverse `map(automation.ApiType)` method. A misconfigured constant in the automation → definition direction would silently set an incorrect type on the shared policy group definition, undetected.

carlos-andres-osorio:
> **gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/ApiResourceTest.java:1**

No resource-level integration test asserts that reading an EDGE-typed entity through the REST endpoint returns a structured error response. Without this, the HTTP error contract (status code and body shape) is unverified end-to-end. Even after the current HTTP 500 is fixed, nothing prevents a future regression to a different wrong status or an empty body. A test in `ApiResourceTest` (and `SharedPolicyGroupResourceTest`) that stubs the use case to return an EDGE-typed entity and asserts a structured 4xx response — not 500, not empty — would lock in the correct behavior.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/ApiMapper.java
carlos-andres-osorio, line 68: The `@ValueMapping(source = "EDGE", target = THROW_EXCEPTION)` guard on this method fires on the read/export path whenever a pre-existing EDGE-typed API is retrieved (e.g. `GET /environments/{envId}/apis`). MapStruct generates code that throws `IllegalArgumentException`, but `GraviteeAutomationApplication` registers no `ExceptionMapper` for that type — Jersey's default handler catches it and returns HTTP 500 with no actionable error body to the client.

The new `should_reject_edge_api_type` tests do not exercise this path: they send `"type": "EDGE"` on a PUT, which is rejected by Jackson during deserialization (because `EDGE` was removed from the generated automation `ApiType` enum) before the mapper is ever called. The guard and the 500 scenario it creates ship without test coverage.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/SharedPolicyGroupMapper.java
carlos-andres-osorio: Same root cause as the API mapper: the `@ValueMapping(source = "EDGE", target = THROW_EXCEPTION)` guard here throws `IllegalArgumentException` when a pre-existing EDGE-typed shared policy group is fetched. `SharedPolicyGroupResource.get()` catches only `SharedPolicyGroupNotFoundException`, so the `IllegalArgumentException` propagates unhandled to Jersey → HTTP 500.

`GraviteeAutomationApplication` has no `ExceptionMapper` for `IllegalArgumentException`, so both this mapper and `ApiMapper` share the same unhandled exception path. The new `should_reject_edge_api_type` tests verify the Jackson deserialization boundary (write path) and never reach this guard.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/ApisResourceTest.java
```diff
@@ -133,6 +133,19 @@ void should_return_state_from_cross_id_no_hrid() {
             }
         }
 
+        @Test
+        void should_reject_edge_api_type() {
```
carlos-andres-osorio, line 137: The two new `should_reject_edge_api_type` tests (dryRun=false here, dryRun=true at line 286) pass because `EDGE` was removed from the generated automation `ApiType` enum in `open-api.yaml`. Jackson's `fromValue` throws `IllegalArgumentException` during deserialization of the PUT request body — the `@ValueMapping(THROW_EXCEPTION)` guard in `ApiMapper` is never reached.

That guard fires only on the read/export direction (mapping `management.v2.rest.model.ApiType` → `automation.model.ApiType`), which these write-path tests do not exercise. The mapper guard therefore ships without a test that actually drives it.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/SharedPolicyGroupResourceTest.java
```diff
@@ -263,6 +263,18 @@ void should_reject_with_cross_id_and_no_hrid() {
                     assertThat(response.getStatus()).isEqualTo(400);
                 }
             }
+
+            @Test
```
carlos-andres-osorio, line 267: Symmetric issue on the SPG side: both `should_reject_edge_api_type` tests (dryRun=false here, dryRun=true at line 308) pass via Jackson deserialization rejection of `apiType: "EDGE"` on the write path, not via the `@ValueMapping(THROW_EXCEPTION)` guard in `SharedPolicyGroupMapper`.

The guard fires on the read path when mapping `io.gravitee.definition.model.v4.ApiType.EDGE` — that domain enum value still exists — but this direction is not covered by any test.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/ApiMapper.java
benoitgravitee: Not going to fix this, you can't pass the Jackson serder, I don't the point of testing this. I need the enum to be managed because it is inherited from APIM, there is a compilation issue otherwise.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/SharedPolicyGroupMapper.java
benoitgravitee: Same as above.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/ApisResourceTest.java
benoitgravitee: same as above

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/SharedPolicyGroupResourceTest.java
benoitgravitee: same as above

carlos-andres-osorio (approved):

PR #17336, opened 2026-06-02
fix(portal): fix OIDC login icon contrast when button color is unset

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-14026

## Description

Added provider text-color computation (same logic family as Console: dark bg -> white text, light bg -> black text). Extended provider model with computed textColor.
Updated style generation so provider button styles include text color variable (--gv-button--c) alongside background variable when relevant.

1. Non-hex provider colors get forced black text — incomplete fix + narrow regression · Low · breaking-change: No

- **What:** A configured provider color that isn't 6-digit hex (#abc, darkblue, rgb(...)) is treated as "light" and gets forced black text, even on a dark background.
- **Where:** gravitee-apim-portal-webui/src/app/pages/login/login.component.ts:165 (if (!match) { return false; }) → :128 (colorIsDarkAdvanced(color) ? '#ffffff' : '#000000').
- **Why it matters:** An operator who sets a dark non-hex SSO button color sees an unreadable black-on-dark label. It's a regression for that input: the old code emitted no text-color variable, so gv-button fell back to its white default (readable); the new code forces black. Reachability is bounded — the console color picker enforces ^#[0-9a-f]{6}$, so this is only settable via direct REST API.
- **Fix:** Have getProviderTextColor return undefined (not #000000) when HEX_RGB_PATTERN doesn't match, so unparseable colors fall back to the component default instead of forced black.
- **Rule:** none.

## Additional context

https://github.com/user-attachments/assets/0de9a4eb-5b06-45cd-b233-ae284ab53e2b

### DIFF
--- a/gravitee-apim-portal-webui/src/app/pages/login/login.component.ts
+++ b/gravitee-apim-portal-webui/src/app/pages/login/login.component.ts
@@ -107,10 +107,22 @@ export class LoginComponent implements OnInit, AfterViewInit, OnDestroy {
     return this.loginForm.valid.valueOf();
   }
 
-  getProviderStyle(provider) {
-    if (provider.color) {
-      return `--gv-button-${provider.type.toLowerCase()}--bgc:${provider.color};`;
+  getProviderStyle(provider: IdentityProvider): string {
+    const color = provider.color?.trim();
+    if (!color) {
+      return '';
     }
-    return '';
+    const textColor = colorIsDark(color) ? '#ffffff' : '#000000';
+    return `--gv-button-${provider.type.toLowerCase()}--bgc:${color};--gv-button--c:${textColor};`;
   }
 }
+
+const colorIsDark = (bgColor: string): boolean => {
+  const hex = bgColor.charAt(0) === '#' ? bgColor.substring(1, 7) : bgColor;
+  const r = Number.parseInt(hex.substring(0, 2), 16);
+  const g = Number.parseInt(hex.substring(2, 4), 16);
+  const b = Number.parseInt(hex.substring(4, 6), 16);
+  const uicolors = [r / 255, g / 255, b / 255];
+  const c = uicolors.map(col => (col <= 0.03928 ? col / 12.92 : Math.pow((col + 0.055) / 1.055, 2.4)));
+  return 0.2126 * c[0] + 0.7152 * c[1] + 0.0722 * c[2] <= 0.179;
+};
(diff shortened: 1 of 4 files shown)


aditya-goyal01:
1. Apply appropriate labels to propagate the fix to all affected `APIM` versions. In this case, `apply-on-4-9-x`, `apply-on-4-10-x`, `apply-on-4-11-x` and `apply-on-4-12-x`, `apply-on-master` must be applied. WDYT? 
2. If possible, add a regression test to verify the fix.

gravitee-apim-portal-webui/src/app/pages/login/login.component.ts
```diff
@@ -107,10 +112,35 @@ export class LoginComponent implements OnInit, AfterViewInit, OnDestroy {
     return this.loginForm.valid.valueOf();
   }
 
-  getProviderStyle(provider) {
+  getProviderStyle(provider: IdentityProviderVM): string {
```
vikrantgravitee: getProviderStyle enters its branch on if (provider.color) (truthy), but getProviderTextColor returns undefined when provider.color.trim() is empty. So a whitespace-only color (e.g. " ") produces the literal string in the output: --gv-button--c:undefined;. Unlikely in practice, but worth aligning the two checks (e.g. trim/isEmpty in both, or guard textColor before appending).

gravitee-apim-portal-webui/src/app/pages/login/login.component.ts
```diff
@@ -107,10 +112,35 @@ export class LoginComponent implements OnInit, AfterViewInit, OnDestroy {
...
+  const color = provider.color?.trim();
+  if (!color) {
+    return undefined;
+  }
+
+  return colorIsDarkAdvanced(color) ? '#ffffff' : '#000000';
+};
+
+const colorIsDarkAdvanced = (bgColor: string): boolean => {
+  const color = bgColor.charAt(0) === '#' ? bgColor.substring(1, 7) : bgColor;
```
vikrantgravitee: Review this sonar blocker.

ankita-gupta21: fixed

gravitee-apim-portal-webui/src/app/pages/login/login.component.ts
ankita-gupta21: fixed

vikrantgravitee (approved):

PR #17348, opened 2026-06-02
fix(console): use v2 API to load all APIs (v2 + v4) in platform alert filter dropdown

araffass:
## Issue

https://gravitee.atlassian.net/browse/APIM-14152

## Description

The "Value" dropdown in platform-level alert filters (API + equals_to operator) was using the v1 Management API endpoint (/apis/_search/) which only returns v2 APIs and is broken in some environments. Replaced with ngApiV2Service.getAll() (v2 Management API) which fetches all API versions (v2 + v4) across all pages.

## Additional context

https://github.com/user-attachments/assets/909afabc-e01e-41c4-8536-d9a03fa8892a

### DIFF
--- a/gravitee-apim-console-webui/src/entities/alerts/api.metrics.ts
+++ b/gravitee-apim-console-webui/src/entities/alerts/api.metrics.ts
@@ -15,10 +15,12 @@
  */
 
 import { each, keyBy, keys } from 'lodash';
+import { last } from 'rxjs/operators';
 
 import { ApiService } from '../../services/api.service';
 import ApplicationService from '../../services/application.service';
 import TenantService from '../../services/tenant.service';
+import { ApiV2Service } from '../../services-ngx/api-v2.service';
 import { CompareCondition, Metrics, Scope, StringCondition, ThresholdCondition, ThresholdRangeCondition, Tuple } from '../alert';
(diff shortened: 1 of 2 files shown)


gravitee-apim-console-webui/src/entities/alerts/api.metrics.ts
```diff
@@ -95,9 +96,10 @@ export class ApiMetrics extends Metrics {
       const apis: Tuple[] = [];
 
       if (type === 2) {
-        // PLATFORM: Search for all registered APIs
-        ($injector.get('ApiService') as ApiService).searchApis().then((result) => {
-          result.data.forEach((api) => {
+        // PLATFORM: Search for all registered APIs (v2 + v4)
```
vikrantgravitee, line 100: Could it be better


// PLATFORM: Search for all registered APIs (v2 + v4)

($injector.get('ngApiV2Service') as ApiV2Service)
  .getAll()
  .pipe(last())
  .subscribe((result) => {
    apis.length = 0;
    result.forEach((api) => {
      apis.push(new Tuple(api.id, api.name));
    });
  });

araffass: Updated

vikrantgravitee (approved):

PR #17349, opened 2026-06-02
feat(rest-api): PATCH endpoint for partial v4 HTTP Proxy plan updates

carlos-andres-osorio:
## Issue

[APIM-14082](https://gravitee.atlassian.net/browse/APIM-14082)

## Why

Plans are not included in the `UpdateApiV4` schema, so `PATCH /apis/{apiId}` cannot reach plan fields. When an API publisher needs to change only a subset of plan fields (e.g. name or rate-limit config), they previously had to re-send the full plan definition via `PUT`. A dedicated PATCH endpoint makes partial updates safe and eliminates that full-object round-trip.

## What changed

- **`PatchPlanUseCase`** — new domain use case that applies JSON Patch (RFC 6902) or JSON Merge Patch (RFC 7396) to a v4 HTTP Proxy plan; enforces an allow-list of 11 patchable fields, rejects `status` changes, validates post-patch state, and supports a dry-run mode.
- **`ApiPlansResource`** — new `PATCH /apis/{apiId}/plans/{planId}` endpoint wired to `PatchPlanUseCase`; handles `application/json-patch+json`, `application/merge-patch+json`, and `application/json` (treated as merge-patch); enforces `If-Match` / ETag precondition when the header is present; returns ETag and Last-Modified on success.
- **`AbstractResource`** — precondition-check helpers extracted and reused by the PATCH endpoint.
- **`PlanForApiNotFoundException` / `PlanPatchNotAllowedException`** — new domain exceptions for cross-plan ownership violation and disallowed-field patches.
- **`UpdatePlanDomainService`** — minor adjustments to support the patched plan payload.
- **`PreconditionFailedException`** — extended to cover `412` from ETag mismatches.
- **`openapi-apis.yaml`** — documents the PATCH endpoint: patchable fields, content-type semantics, `dryRun` query param, `If-Match` / ETag, and error codes.

## User-facing impact

New `PATCH /apis/{apiId}/plans/{planId}` endpoint on the Management REST API v2:
- Accepts `application/json-patch+json` (RFC 6902), `application/merge-patch+json` (RFC 7396), or plain `application/json` (treated as merge-patch).
- `?dryRun=true` validates the patch and returns the would-be plan state without persisting.
- `If-Match` header is optional; when provided and mismatched, returns `412 Precondition Failed`.
- Patch documents touching `status` return `400`; those exceeding 200 operations return `400`.
- Scoped to v4 HTTP Proxy APIs only; other API types return `4xx`.

[APIM-14082]: https://gravitee.atlassian.net/browse/APIM-14082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/adapter/PatchPlanFlowsDeserializer.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/adapter/PatchPlanFlowsDeserializer.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.management.v2.rest.adapter;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.gravitee.apim.core.exception.ValidationDomainException;
+import io.gravitee.apim.core.plan.use_case.PatchPlanUseCase.PlanFlowsConverter;
+import io.gravitee.definition.model.v4.flow.Flow;
+import io.gravitee.rest.api.management.v2.rest.mapper.FlowMapper;
+import io.gravitee.rest.api.management.v2.rest.model.FlowV4;
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * @author GraviteeSource Team
+ */
+public class PatchPlanFlowsDeserializer implements PlanFlowsConverter {
+
+    private final ObjectMapper objectMapper;
+
+    public PatchPlanFlowsDeserializer(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+    }
+
+    @Override
+    public JsonNode toCurrentFlowsNode(List<Flow> flows) {
+        return objectMapper.valueToTree(FlowMapper.INSTANCE.mapFromHttpV4(flows));
+    }
+
+    @Override
+    public List<Flow> fromPatchedFlowsNode(JsonNode flowsNode) {
+        try {
+            var restFlows = objectMapper.readerForListOf(FlowV4.class).<List<FlowV4>>readValue(flowsNode);
+            return FlowMapper.INSTANCE.mapToHttpV4(restFlows);
+        } catch (IOException e) {
+            throw new ValidationDomainException("Invalid value for field 'flows': " + e.getMessage(), e);
+        }
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/AbstractResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/AbstractResource.java
@@ -21,6 +21,7 @@ import static io.gravitee.rest.api.model.MembershipReferenceType.GROUP;
 
 import io.gravitee.apim.core.audit.model.AuditActor;
 import io.gravitee.apim.core.audit.model.AuditInfo;
+import io.gravitee.apim.core.utils.StringUtils;
 import io.gravitee.rest.api.idp.api.authentication.UserDetails;
 import io.gravitee.rest.api.management.v2.rest.model.Links;
 import io.gravitee.rest.api.management.v2.rest.pagination.PaginationLinks;
@@ -292,25 +293,53 @@ public abstract class AbstractResource {
     }
 
     protected void evaluateIfMatch(final HttpHeaders headers, final String etagValue) {
+        evaluateIfMatch(headers, etagValue, false);
+    }
+
+    protected void evaluateIfMatchStrictly(final HttpHeaders headers, final String etagValue) {
+        evaluateIfMatch(headers, etagValue, true);
+    }
+
+    private void evaluateIfMatch(final HttpHeaders headers, final String etagValue, final boolean strict) {
         String ifMatch = headers.getHeaderString(HttpHeaders.IF_MATCH);
 
-        if (Objects.nonNull(ifMatch) && !ifMatch.isEmpty()) {
-            // Handle case for -gzip appended automatically (and sadly) by Apache
-            ifMatch = ifMatch.replaceAll("-gzip", "");
+        if (StringUtils.isEmpty(ifMatch)) {
+            return;
+        }
+        // If-Match: * matches any tag; resource existence is the caller's responsibility.
+        if (strict && "*".equals(ifMatch)) {
+            return;
+        }
 
-            Set<MatchingEntityTag> matchingTags;
-            try {
-                matchingTags = HttpHeaderReader.readMatchingEntityTag(ifMatch);
-            } catch (java.text.ParseException e) {
-                return;
+        // Handle case for -gzip appended automatically (and sadly) by Apache.
+        // Weak validators (W/"...") are accepted as-is: with epoch-millis ETags every comparison is byte-identical,
+        // so the RFC 7232 strong-comparison distinction has no practical effect here.
+        var normalizedIfMatch = ifMatch.replace("-gzip", "");
+
+        Set<MatchingEntityTag> matchingTags;
+        try {
+            matchingTags = HttpHeaderReader.readMatchingEntityTag(normalizedIfMatch);
+        } catch (java.text.ParseException e) {
+            if (strict) {
+                throw new PreconditionFailedException(e);
             }
+            log.debug("Ignoring unparseable If-Match header '{}' on non-strict precondition path", ifMatch);
+            return;
+        }
 
-            MatchingEntityTag ifMatchHeader = matchingTags.iterator().next();
-            EntityTag eTag = new EntityTag(etagValue, ifMatchHeader.isWeak());
-
-            if (matchingTags != MatchingEntityTag.ANY_MATCH && !matchingTags.contains(eTag)) {
+        if (matchingTags.isEmpty()) {
+            if (strict) {
                 throw new PreconditionFailedException();
             }
+            log.debug("Ignoring If-Match header '{}' resolving to an empty tag set on non-strict precondition path", ifMatch);
+            return;
+        }
+
+        var ifMatchHeader = matchingTags.iterator().next();
+        var eTag = new EntityTag(etagValue, ifMatchHeader.isWeak());
+
+        if (matchingTags != MatchingEntityTag.ANY_MATCH && !matchingTags.contains(eTag)) {
+            throw new PreconditionFailedException();
         }
     }
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/use_case/PatchPlanUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/use_case/PatchPlanUseCase.java
@@ -1,0 +1,486 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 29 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/use_case/PatchPlanUseCase.java
```diff
@@ -0,0 +1,486 @@
...
+
+        var explicitNulls = collectExplicitNulls(input.patchType(), patchNode);
+        rejectNullOnNonNullableFields(explicitNulls);
+
+        var patchedNode = applyPatch(input.patchType(), patchNode, currentNode, input.planId(), input.apiId());
+
+        var patched = applyPatchedValues(plan, patchedNode, explicitNulls);
+        var patchedFlows = extractFlows(patchedNode, currentFlows, explicitNulls);
+
+        if (input.dryRun()) {
```
paulatulis, line 142: dry-run returns the un-sanitized would-be state. validate(...) sanitizes flows internally (validateAndSanitizeHttpV4) but returns void, so the discarded sanitization never reaches the response — line 138 returns the raw patched/patchedFlows. The real persist path below (line 141) instead returns the sanitized result of update(...). So a dryRun=true preview can diverge from what an actual PATCH would store (flow sanitization, default coercion). The AC says dry-run returns "the would-be plan state" — that arguably means the sanitized state. Could validate(...) return the sanitized Plan/List<Flow> so the preview matches the persisted outcome?

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/AbstractResource.java
```diff
@@ -292,25 +293,53 @@ protected Response.ResponseBuilder applyCacheHeaders(Response.ResponseBuilder bu
     }
 
     protected void evaluateIfMatch(final HttpHeaders headers, final String etagValue) {
+        evaluateIfMatch(headers, etagValue, false);
+    }
+
+    protected void evaluateIfMatchStrictly(final HttpHeaders headers, final String etagValue) {
```
paulatulis, line 299: Moving evaluateIfMatchStrictly into AbstractResource and unifying strict/non-strict is reasonable, and it actually fixes a latent bug: the old non-strict path called matchingTags.iterator().next() with no empty check → a NoSuchElementException (500) was possible on a parseable-but-empty If-Match; the new code guards matchingTags.isEmpty(). Good. But because this helper is now shared with ApiResource's existing strict callers, please confirm ApiResource's ETag/If-Match tests still pass — the strict empty-set and unparseable-header branches now throw 412 where the old inline version's control flow was subtly different.

paulatulis (commented):
Every PATCH re-serializes and re-persists the plan's flows, even when the patch never touches flows. In toJsonNode, flows is always written into the current-state node (whenever currentFlows != null). So after applying any patch (even {"name":"x"}), patchedNode still contains flows, and extractFlows takes this branch:

```
var flowsNode = patchedNode.get(FIELD_FLOWS);
if (flowsNode == null || flowsNode.isNull()) return currentFlows;
return planFlowsConverter.fromPatchedFlowsNode(flowsNode);   // <-- always hit for an untouched flows field
```

fromPatchedFlowsNode runs a full definition.Flow → FlowV4 → definition.Flow MapStruct round-trip, and the non-dry-run path persists the result. So a patch that only changes the plan name rewrites all of the plan's flows through a lossy-prone mapper.

The strongest evidence this is a live hazard is in this same PR: FlowAdapter gains @Mapping(target="pathOperator", defaultValue="STARTS_WITH") — a fix for exactly this round-trip dropping a field (and it lines up with your branch's recent 43a3639da0 "coerce pathOperator on read"). That tells me the round-trip is lossy and was patched for one field. Any other field with the same null/default asymmetry will be silently mutated on every unrelated patch. Secondary symptom: checkSynchronized will see "changed" flows and can spuriously set needRedeployAt on every patch.

Recommendation: when the raw patch doesn't target /flows, skip the round-trip and pass currentFlows straight through (you already have patchNode; a has("flows") / pointer-prefix check like PatchApiUseCase.isTargetedByJsonPatch is cheap). That eliminates the risk for the common case and shrinks the fidelity surface to "patches that actually edit flows." If the team would rather keep the always-round-trip design, I'd want a round-trip test covering every flow/selector field, not just pathOperator.


Otherwise,

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/AbstractResource.java
carlos-andres-osorio: Confirmed — ApiResource's ETag/If-Match behaviour is unchanged for the normal cases, and the only difference on the two edge branches is the bug fix you flagged as good.

On "subtly different control flow": the old `ApiResource.evaluateIfMatchStrictly` parsed the header once in its own wrapper and then delegated to the (old) non-strict `evaluateIfMatch`, which parsed a second time and called `matchingTags.iterator().next()` with no empty guard. The unified helper parses once and branches on a `strict` flag. The two edge branches resolve as:

- **Unparseable header** — old: parse in the wrapper throws → 412. New: single parse throws → strict → 412. Same outcome, just no longer double-parsed.
- **Empty tag set** (header parses but yields an empty set) — old: the wrapper parse succeeded, so it fell through to the non-strict delegate, whose unguarded `iterator().next()` threw `NoSuchElementException` → 500. New: `matchingTags.isEmpty()` + strict → 412. This is the only behavioural change, and it's the latent 500→412 fix.

Normal match/mismatch is identical.

Verified three ways:
1. Diff-equivalence of the old vs new strict path (as above).
2. Existing strict-caller tests green: ApiResource_PatchApiConcurrencyTest (13), ApiResource_StartTest (8), ApiResource_StopTest (6), plus ApiPlansResource_PatchApiPlanTest (35) — 62 tests, 0 failures.
3. Live curl against the running branch:

| If-Match | ApiResource `PATCH /apis/{id}` | ApiPlansResource `PATCH …/plans/{id}` (dryRun) |
|---|---|---|
| matching | (skipped — would persist) | 200 |
| mismatch | 412 | 412 |
| wildcard `*` | (skipped) | 200 |
| malformed (unterminated quote) | 412 | 412 |
| garbage header | 412 | 412 |

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/use_case/PatchPlanUseCase.java
carlos-andres-osorio: Verified against the source: both the dry-run and persist paths return the same patchedFlows reference, so flow sanitization cannot diverge between preview and persisted state. In PatchPlanUseCase the dry-run path (line 138) and persist path (line 142) both return new PlanWithFlows(..., patchedFlows) — the persisted response's flows come from patchedFlows, not from update(...). Sanitization runs on that same list in both paths: UpdatePlanDomainService.validate(...) (dry-run) and updateHttpV4ApiPlan(...) (persist) both call validateAndSanitizeHttpV4, which returns the same list reference it receives (FlowValidationDomainService.java:118) and mutates elements in place. PatchPlanUseCaseTest.FlowSanitizationReflectedInResponse pins this with paired persist_returns_sanitized_flow_configuration_matching_stored_state and dry_run_returns_sanitized_flow_configuration_in_preview tests asserting the sanitized configuration appears in both responses.

carlos-andres-osorio:
> Every PATCH re-serializes and re-persists the plan's flows, even when the patch never touches flows. In toJsonNode, flows is always written into the current-state node (whenever currentFlows != null). So after applying any patch (even {"name":"x"}), patchedNode still contains flows, and extractFlows takes this branch:
> 
> ```
> var flowsNode = patchedNode.get(FIELD_FLOWS);
> if (flowsNode == null || flowsNode.isNull()) return currentFlows;
> return planFlowsConverter.fromPatchedFlowsNode(flowsNode);   // <-- always hit for an untouched flows field
> ```
> 
> fromPatchedFlowsNode runs a full definition.Flow → FlowV4 → definition.Flow MapStruct round-trip, and the non-dry-run path persists the result. So a patch that only changes the plan name rewrites all of the plan's flows through a lossy-prone mapper.
> 
> The strongest evidence this is a live hazard is in this same PR: FlowAdapter gains @Mapping(target="pathOperator", defaultValue="STARTS_WITH") — a fix for exactly this round-trip dropping a field (and it lines up with your branch's recent 43a3639da0 "coerce pathOperator on read"). That tells me the round-trip is lossy and was patched for one field. Any other field with the same null/default asymmetry will be silently mutated on every unrelated patch. Secondary symptom: checkSynchronized will see "changed" flows and can spuriously set needRedeployAt on every patch.
> 
> Recommendation: when the raw patch doesn't target /flows, skip the round-trip and pass currentFlows straight through (you already have patchNode; a has("flows") / pointer-prefix check like PatchApiUseCase.isTargetedByJsonPatch is cheap). That eliminates the risk for the common case and shrinks the fidelity surface to "patches that actually edit flows." If the team would rather keep the always-round-trip design, I'd want a round-trip test covering every flow/selector field, not just pathOperator.
> 
> 
> Otherwise, 

Done — extractFlows now short-circuits and returns the existing currentFlows unchanged when the raw patch doesn't target /flows, so the fromPatchedFlowsNode MapStruct round-trip only runs for patches that actually edit flows. The gate is patchTargetsFlows: for merge-patch it checks patchNode.has("flows"); for JSON Patch it scans each op's path (and a move's from) against the /flows pointer, mirroring the isTargetedByJsonPatch pattern. A name-only patch therefore passes its flows through untouched. The secondary redeploy symptom is fixed too — non-flow patches no longer trip needRedeployAt, covered by the NonFlowPatchDoesNotForceRedeploy tests.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/use_case/PatchPlanUseCase.java
```diff
@@ -0,0 +1,523 @@
...
+        var validation = existing.getValidation();
+        if (validationNode != null && !validationNode.isNull()) {
+            try {
+                validation = Plan.PlanValidationType.valueOf(validationNode.asText());
+            } catch (IllegalArgumentException e) {
+                throw new ValidationDomainException("Invalid validation value: " + validationNode.asText(), e);
+            }
+        }
+
+        var excludedGroups = stringListOrDefault(patchedNode, FIELD_EXCLUDED_GROUPS, explicitNulls, existing.getExcludedGroups());
```
paulatulis, line 370: PATCH skips excludedGroups validation that PUT enforces.
excludedGroups is patchable (PatchPlanUseCase.java:364,374), but neither the use case nor UpdatePlanDomainService.updatePreFlightChecks validates that the group IDs exist in the environment. The sibling PUT /plans/{planId} in the same resource calls validateExcludedGroups(...) for V4 (ApiPlansResource.java:293), which throws GroupNotFoundException for unknown IDs. So a client can PATCH excludedGroups to garbage IDs and they persist silently — a real divergence from PUT. Tags, by contrast, are validated on both paths (validatePlanTagsAgainstApiTags). Worth either wiring the same group check into the PATCH path or pushing it into the domain service so both entry points share it.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/adapter/PatchPlanFlowsDeserializer.java
```diff
@@ -0,0 +1,53 @@
...
+        this.objectMapper = objectMapper;
+    }
+
+    @Override
+    public JsonNode toCurrentFlowsNode(List<Flow> flows) {
+        return objectMapper.valueToTree(FlowMapper.INSTANCE.mapFromHttpV4(flows));
+    }
+
+    @Override
+    public List<Flow> fromPatchedFlowsNode(JsonNode flowsNode) {
```
paulatulis, line 45: Residual flows round-trip when a flows patch does occur — only pathOperator is regression-tested. The skip fixes the common case, but when a patch touches /flows at all (even /flows/0/name), extractFlows re-serializes the entire flow list through definition.Flow → FlowV4 → definition.Flow (PatchPlanFlowsDeserializer.java:45-52), so an untouched flow in the same plan still round-trips. The FlowAdapter.toModel pathOperator default (FlowAdapter.java:111) patches the one known null/default asymmetry, but any other selector/flow field with the same asymmetry would be silently mutated. A broader round-trip fidelity test across flow + selector fields (not just pathOperator) would close this — it's the "comprehensive round-trip test" half of my original ask, now scoped down to flows-only patches.

paulatulis (commented):
The fixes look perfect. New agent found a couple additional things that I've flagged.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/use_case/PatchPlanUseCase.java
carlos-andres-osorio: Verified against the source: both the dry-run and persist paths return the same patchedFlows reference, so flow sanitization cannot diverge between preview and persisted state. In PatchPlanUseCase the dry-run path (line 138) and persist path (line 142) both return new PlanWithFlows(..., patchedFlows) — the persisted response's flows come from patchedFlows, not from update(...). Sanitization runs on that same list in both paths: UpdatePlanDomainService.validate(...) (dry-run) and updateHttpV4ApiPlan(...) (persist) both call validateAndSanitizeHttpV4, which returns the same list reference it receives (FlowValidationDomainService.java:118) and mutates elements in place. PatchPlanUseCaseTest.FlowSanitizationReflectedInResponse pins this with paired persist_returns_sanitized_flow_configuration_matching_stored_state and dry_run_returns_sanitized_flow_configuration_in_preview tests asserting the sanitized configuration appears in both responses.

paulatulis (approved):
Nice!

PR #17351, opened 2026-06-02
fix(portal): enhance navigation context handling and improve user experience

prateektiwari-gravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-14266

## Description

This bug fixes 2 things: 
1. Handling of pagination params
2. Making sure no NaNs arrive when navigating between tabs in the portal

## Additional context

- Added computed width and height for gv-identity-picture to prevent NaN issues.
- Updated display_name binding to fallback to user email if display_name is not available.
- Implemented navigation context preservation when changing routes in AppComponent and GvDocumentationComponent.
- Added tests to ensure correct behavior of pagination parameters and navigation context in Applications and Subscriptions components.
- Refactored pagination parameter parsing for better handling of non-numeric values.

Post fix behaviour: 

https://github.com/user-attachments/assets/104c7c70-4bd3-47d3-afe2-49fcdb1fab91

### DIFF
--- a/gravitee-apim-portal-webui/src/app/pages/subscriptions/subscriptions.component.ts
+++ b/gravitee-apim-portal-webui/src/app/pages/subscriptions/subscriptions.component.ts
@@ -35,6 +35,7 @@ import {
 } from '../../../../projects/portal-webclient-sdk/src/lib';
 import { ConfigurationService } from '../../services/configuration.service';
 import { formatCurlCommandLine } from '../../utils/utils';
+import { parsePaginationParam } from '../../utils/navigation-query-params.util';
 
 const StatusEnum = Subscription.StatusEnum;
 
@@ -166,8 +167,8 @@ export class SubscriptionsComponent implements OnInit, OnDestroy {
         takeUntil(this.unsubscribe$),
         switchMap(params =>
           this.applicationService.getApplications({
-            size: Number(params.get('size') ?? this.paginationSize),
-            page: Number(params.get('page') ?? 1),
+            size: parsePaginationParam(params.get('size'), this.paginationSize),
+            page: parsePaginationParam(params.get('page'), 1),
           }),
         ),
       )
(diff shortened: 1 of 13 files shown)


gravitee-apim-portal-webui/src/app/utils/navigation-query-params.util.ts
```diff
@@ -0,0 +1,53 @@
...
+  return Number.isFinite(parsed) && parsed > 0 ? parsed : defaultValue;
+}
+
+/**
+ * Navigates to `routePath` while preserving the current navigation context query params.
+ * Query params carried by `routePath` take precedence over the preserved context.
+ */
+export function navigateWithNavigationContext(router: Router, routePath: string): Promise<boolean> {
+  const urlTree = router.parseUrl(routePath);
+  const path = urlTree.root.children[PRIMARY_OUTLET].segments.join('/');
```
aditya-goyal01: `PRIMARY_OUTLET` is absent when `routePath` resolves to `/`, means the root `UrlTree` has no primary children. It will cause a failure, `TypeError` maybe. 

Yes/No?

gravitee-apim-portal-webui/src/app/pages/subscriptions/subscriptions.component.ts
```diff
@@ -166,8 +167,8 @@ export class SubscriptionsComponent implements OnInit, OnDestroy {
         takeUntil(this.unsubscribe$),
         switchMap(params =>
           this.applicationService.getApplications({
-            size: Number(params.get('size') ?? this.paginationSize),
-            page: Number(params.get('page') ?? 1),
+            size: parsePaginationParam(params.get('size'), this.paginationSize),
+            page: parsePaginationParam(params.get('page'), 1),
```
aditya-goyal01, line 171: A test e.g. `when page is non-numeric default page as 1` must be added to support the logic. No?

prateektiwari-gravitee: Yes, I have added that now.

gravitee-apim-portal-webui/src/app/utils/navigation-query-params.util.ts
prateektiwari-gravitee: Yes — good catch. Added optional chaining so a root path like / no longer throws, plus a unit test for that case. Same approach as isHomepageUrl in AppComponent.

aditya-goyal01 (approved):

PR #17367, opened 2026-06-02
fix(gamma-authz): key events_latest by resource id so deleted authz state evicts

podlesrafal:
## Problem

`EventRepositoryAuthzEventPublisher.emit()` wrote the *same* event — with a **fresh random id** — to both the append-only `events` log and the `events_latest` snapshot. `EventLatestRepository.createOrUpdate` upserts by `event.getId()`, so a random id per emit made every `PUBLISH_AUTHZ_*` and `UNPUBLISH_AUTHZ_*` a **separate** `events_latest` row. The collection never deduplicated per resource, so an `UNPUBLISH` never superseded its `PUBLISH`.

The gateway cold-start sync (`Authz{Policy,Entity}Synchronizer`, PUBLISH-only + latest-per-id) then **resurrected deleted policies/entities on every restart**:
- a deleted/revoked policy kept being enforced (correctness + security),
- authz search (`/access/v1/search/*`) over-returned stale resources.

Verified: after deleting all authz resources and restarting the gateway, it logged `63 authz policies synchronized` while the store held `1`.

## Fix

Write the `events_latest` copy with a **stable id = resource id** (the `AUTHZ_POLICY_ID` / `AUTHZ_ENTITY_ID` property value), prefixed by the property key to avoid `_id` collisions across resource types in the shared collection — mirroring core `EventServiceImpl.createOrPatchLatestEvent`, which sets a stable `latestEventId`. The append-only `events` write keeps its unique random id.

Now `events_latest` holds one row per resource (the latest event), so the standard PUBLISH-only cold-start sync (same as `ApiSynchronizer`/`DictionarySynchronizer`) correctly skips deleted resources.

## Validation (engine-vs-gamma parity harness, 31 scenarios)

End-to-end against the reference `gravitee-authorization-engine` playground. This fix removes the snapshot accumulation that previously masked sync latency; combined with the PDP parent-UID fix:

| dimension | before | after |
|---|---|---|
| single eval | 89% | **100%** |
| batch | 89% | **100%** |
| search subject | 37% | **100%** |
| search resource | 34% | **96%** |

`events_latest` rows dropped from one-per-event to one-per-resource; deleted resources no longer resurrect across gateway restarts.

### DIFF
--- a/gravitee-apim-distribution/pom.xml
+++ b/gravitee-apim-distribution/pom.xml
@@ -2078,6 +2078,19 @@
                         </exclusion>
                     </exclusions>
                 </dependency>
+                <dependency>
+                    <groupId>com.graviteesource.gamma</groupId>
+                    <artifactId>gravitee-policy-authz-pdp-authzen</artifactId>
+                    <version>${gravitee-policy-authz-pdp-authzen.version}</version>
+                    <type>zip</type>
+                    <scope>runtime</scope>
+                    <exclusions>
+                        <exclusion>
+                            <groupId>*</groupId>
+                            <artifactId>*</artifactId>
+                        </exclusion>
+                    </exclusions>
+                </dependency>
 
                 <!-- Reactors -->
                 <dependency>
--- a/gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/event/EventRepositoryAuthzEventPublisher.java
+++ b/gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/event/EventRepositoryAuthzEventPublisher.java
@@ -143,21 +143,29 @@ public final class EventRepositoryAuthzEventPublisher implements AuthzEventPubli
     }
 
     /**
-     * Writes the event log first, then refreshes the per-id latest snapshot.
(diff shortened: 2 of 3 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-standalone/gravitee-apim-rest-api-standalone-distribution/pom.xml
```diff
@@ -363,6 +363,19 @@
...
+            <version>${gravitee-policy-authz-pdp-authzen.version}</version>
+            <type>zip</type>
+            <scope>runtime</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
```
phiz71: I don't think you need it, since it's already in the apim-disrtibution pom

phiz71: Good enough for now. We will review it later

phiz71 (approved):

phiz71 (approved):

PR #17370, opened 2026-06-02
feat(ratelimit): enable Redis Cluster for rate limiting [APIM-14146]

wbabyte:
## APIM-14146 — Enable Redis Cluster (3/3: rate limiting)

Part of [APIM-14146](https://gravitee.atlassian.net/browse/APIM-14146) (cluster-safe rate limiting, requested by Posten Bring).

Makes Redis-backed rate limiting work on a real multi-master Redis Cluster. Two distinct cluster bugs are fixed.

### Commits
1. **`fix(ratelimit): make Redis rate-limit script cluster-slot-safe`** — `EVALSHA` was sent with `numkeys="2"`, making the numeric *weight* a second KEY that hashes to a different slot → `CROSSSLOT`. Now `numkeys="1"` with the weight passed as `ARGV[1]` (Lua re-indexed). Data format unchanged → **not a breaking change**; existing counters are preserved.
2. **`feat(ratelimit): enable Redis Cluster for rate limiting`** — `RedisConnectionFactory` parses `ratelimit.redis.cluster.nodes[i]` and pins `useReplicas=NEVER` (rate limiting needs master-consistent counters). Commented `cluster:` block added to the distribution `gravitee.yml`.
3. **`fix(ratelimit): fall back to EVAL on NOSCRIPT for Redis Cluster`** — `SCRIPT LOAD` only reaches the contacted master, so `EVALSHA` routed by slot to another master returns `NOSCRIPT`. On `NOSCRIPT` we now fall back to `EVAL` with the script source (cached on that node for subsequent calls).

### Tests
- Unit: `numkeys="1"` slot-safety; `NOSCRIPT`→`EVAL` fallback; cluster config parsing.
- Integration: single-node Redis Cluster Testcontainers acceptance test (reproduces CROSSSLOT on the unfixed script). Full module green (42 tests).

### Local smoke test (real 3-master cluster)
Verified a real gateway built from all three repos against a 6-node (3-master) Redis Cluster: a v4 API with a rate-limit policy (5/60s) returned `200×5` then `429` (`X-Rate-Limit-Remaining: 0`), counter stored on the slot-owning master, **0 CROSSSLOT / 0 NOSCRIPT** errors.

### Release chain
Depends on **PR 2/3** (`gravitee-node`). ⚠️ Before merge, bump `gravitee-node.version` to the alpha released from PR 2 (currently `9.0.0-alpha.14` for local testing).

[APIM-14146]: https://gravitee.atlassian.net/browse/APIM-14146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RedisRateLimitRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RedisRateLimitRepository.java
@@ -24,6 +24,7 @@ import io.gravitee.repository.ratelimit.model.RateLimit;
 import io.gravitee.repository.redis.vertx.RedisClient;
 import io.reactivex.rxjava3.core.Single;
 import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.redis.client.Response;
 import io.vertx.rxjava3.SingleHelper;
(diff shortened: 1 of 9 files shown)


gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RedisRateLimitRepository.java
```diff
@@ -102,10 +150,12 @@ private void logOperationFailure(Throwable t) {
         }
     }
 
-    private List<String> convertToList(String scriptSha1, String key, long weight, RateLimit rate) {
+    // numkeys is "1": only the rate-limit key is a KEY, so all keys touched by the command share
+    // one Redis Cluster hash slot (avoids CROSSSLOT). The weight is passed as an ARGV value, not a key.
+    static List<String> convertToList(String scriptSha1, String key, long weight, RateLimit rate) {
```
paulatulis: Nit: rename `scriptSha1` to `script` / `scriptRef` for consistency. On the fallback path it receives the script source, not a SHA.

gravitee-apim-repository/gravitee-apim-repository-redis/src/test/java/io/gravitee/repository/redis/ratelimit/RedisClusterRateLimitRepositoryTest.java
paulatulis, line 1: This spins up a real container in @BeforeAll but is named …Test, so surefire runs it in the unit phase — that makes the unit build hard-depend on a Docker daemon and fail (not skip) anywhere one isn't available. Two options:
 - rename to …IT so it runs under failsafe / the integration phase (matches the usual convention here), or
 - keep it as a unit test but guard it: assumeTrue(DockerClientFactory.instance().isDockerAvailable()) so it skips cleanly without Docker.

Two smaller notes while we're here:
 - findFreePort() closes the socket and then binds the container to that port — a TOCTOU race that can flake under parallel CI. Low risk, but worth a comment acknowledging it.
 - Worth noting in a comment that a single-node cluster reproduces the CROSSSLOT path but not the cross-master NOSCRIPT path (one node owns all slots, so EVALSHA always hits after the local SCRIPT LOAD). So the NOSCRIPT→EVAL fallback is covered by the mock unit test + the manual 3-master smoke test only, not here — fine given the announce-IP pain, just good to make explicit so nobody assumes it's integration-verified.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RedisRateLimitRepository.java
wbabyte: Done renamed to `scriptRef` with a comment noting it's a SHA on EVALSHA and the source on the EVAL fallback.

gravitee-apim-repository/gravitee-apim-repository-redis/src/test/java/io/gravitee/repository/redis/ratelimit/RedisClusterRateLimitRepositoryTest.java
wbabyte: Done guarded with `assumeTrue(DockerClientFactory.instance().isDockerAvailable())` so it skips cleanly (no failsafe/IT wiring in this module, so kept it a unit test rather than renaming to …IT, which would never run). Added the findFreePort TOCTOU comment and a javadoc note that the single-node cluster covers CROSSSLOT but not the cross-master NOSCRIPT path (that's on the mock unit test + manual 3-master smoke). Also fixed a latent flake it surfaced: bus port = dataPort+10000 must be ≤65535, so it now scans a low free-port range.

paulatulis (approved):

PR #17425, opened 2026-06-03
feat(portal): sync spec.navigation through automation API

konrad-krol-incubly:
## Issue

## Description

**Summary**

Slice 2 of nextgen portal automation API — PUT `/portals/{hrid}` now honours `spec.navigation[]` and GET returns the current list.

  **What's in**

  - Domain layer that diffs desired vs. stored `TOP_NAVBAR` folders and applies create/update/delete actions (idempotent, deterministic IDs).
  - mkdir -p approach for implicit ancestors; cascade delete for removed folders (non-folder descendants + page content cleaned up).
  - Path validation pulled into the NavigationPath value object.
  - Automation REST mapper + resources wired to pass navigation in/out.

**Example**

Example: PUT → folder structure
  
```
  curl -X PUT \
    -u admin:admin \
    -H 'Content-Type: application/json' \
    http://localhost:8083/management/v2/organizations/DEFAULT/environments/DEFAULT/portals \
    -d '{
      "hrid": "default-portal",
      "name": "Default Portal",
      "navigation": [
        { "path": "/projects/alpha", "displayName": "Alpha" },
        { "path": "/projects/alpha/docs" },
        { "path": "/projects/beta" }
      ]
    }'
```
  
  Resulting folder tree under `TOP_NAVBAR`:

```
  projects/         (implicit — mkdir -p)
  ├── Alpha/        (displayName override)
  │   └── docs/
  └── beta/
```

**Demo**

https://github.com/user-attachments/assets/87678e8d-f980-4312-ab33-942d16882557

  ---
  **For discussion:**
  
This commit adds **glue** between the new `core/portal/` domain (portal automation) and the pre-existing `core/portal_page/` domain (navigation tree, page content). Its domain to domain mapping

- [ ] Domain-to-domain mapping has to happen somewhere - is this glue concept the right shape, or is there a cleaner pattern?
     - Response (Aqib) -> Yes, this is the right shape for this codebase. `PortalNavigationSyncDomainService` and `PortalNavigationListingDomainService` sitting in `core/portal/domain_service/` while depending on `portal_page` `query/crud` service ports is the standard cross-domain wiring pattern here — exactly how other domain services bridge sub-domains in `gravitee-apim-rest-api-service`. The `portal_page` ports are still interfaces, so the core boundary is respected. No cleaner alternative is needed.
- [ ] Mapping is hand-written Java - see `core/portal/domain_service/navigation/NavigationFolderMapper`. Should we use a mapper library (MapStruct is used in the REST layer, never in core/)?
   - Response (Aqib) --> Keep it hand-written. MapStruct is appropriate for the REST/infra layer where it is already used. The `AGENTS.md` rules and the existing codebase convention both treat core as framework-free. A few static methods in `NavigationFolderMapper` are far cheaper in cognitive overhead than adding a compile-time annotation processor dependency that would be the only MapStruct usage in core. The hand-written approach is correct here.


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/PortalNavigationTreeWalker.java
aqibmohammadkhan, line 1: `PortalNavigationListingDomainService.PathCollector` and `PortalNavigationTreeWalker.traverse` both reconstruct paths from `folder.getTitle()`. But title is set to the displayName override when one is provided — "Alpha" for the path /projects/alpha.

The child traversal in `PortalNavigationTreeWalker` propagates this wrong path as fullPath to all descendants, compounding the error down the tree.

The fix is to store the path segment separately from the display title, so the listing can reconstruct the canonical path from segments and present displayName as a separate field.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationListingDomainService.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private static final class PathCollector implements PortalNavigationVisitor {
+
+        private final List<NavigationPath> paths = new ArrayList<>();
+
+        @Override
+        public void visitFolder(PortalNavigationFolder folder, String parentPath) {
+            paths.add(new NavigationPath(parentPath + "/" + folder.getTitle(), null));
+        }
+    }
```
aqibmohammadkhan, line 61: `PathCollector.visitFolder` always creates `new NavigationPath(path, null)`. There is no way for a caller to discover what `displayName` was previously synced. This makes `GET` a lossy read of what was `PUT`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
```diff
@@ -0,0 +1,122 @@
...
+import io.gravitee.apim.core.portal.domain_service.navigation.actions.FolderActions.DesiredFolder;
+import io.gravitee.apim.core.portal.domain_service.navigation.actions.FolderActions.FolderMutation;
+import io.gravitee.apim.core.portal.domain_service.navigation.actions.FolderActions.UpdateFolder;
+import io.gravitee.apim.core.portal.domain_service.navigation.actions.NavigationAction;
+import io.gravitee.apim.core.portal.model.NavigationPath;
+import io.gravitee.apim.core.portal_page.model.PortalArea;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationFolder;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItem;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
+import io.gravitee.rest.api.service.common.HRIDToUUID;
```
aqibmohammadkhan: `io.gravitee.rest.api.service.*` is the legacy service layer, which is explicitly excluded from the allowed imports in core. The ArchUnit tests in `gravitee-apim-rest-api-service/src/test/java/architecture/` will catch this.

The right fix is to move the deterministicId logic (or an equivalent ID-generation port) out of the planner — either by passing pre-computed IDs in via the `NavigationPath` model, or by extracting ID generation into an injectable interface that `NavigationSyncPlanner` depends on abstractly and `PortalNavigationSyncDomainService` (the executor, which can import legacy code) provides concretely.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/PortalMapper.java
```diff
@@ -48,4 +51,24 @@ default PortalState toPortalState(Portal portal, String hrid) {
...
+            .stream()
+            .map(n -> new io.gravitee.apim.core.portal.model.NavigationPath(n.getPath(), n.getDisplayName()))
+            .toList();
+    }
+
+    default List<NavigationPath> toApiNavigation(List<io.gravitee.apim.core.portal.model.NavigationPath> navigation) {
+        if (navigation == null) {
+            return List.of();
+        }
+        return navigation
```
aqibmohammadkhan, line 69: Both `toCoreNavigation` and `toApiNavigation` use raw List without type parameters.
Should be

```
default List<io.gravitee.apim.core.portal.model.NavigationPath> toCoreNavigation(
    List<NavigationPath> navigation) { ... }

default List<NavigationPath> toApiNavigation(
    List<io.gravitee.apim.core.portal.model.NavigationPath> navigation) { ... }
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
```diff
@@ -0,0 +1,122 @@
...
+        final var currentById = currentFolders.stream().collect(Collectors.toMap(PortalNavigationItem::getId, f -> f));
+        final var desiredIds = desiredFolders.stream().map(FolderActions.DesiredFolder::id).collect(Collectors.toSet());
+        final var depthOrder = Comparator.comparingInt((FolderMutation m) -> PathExpander.depth(m.desired().path()));
+
+        final var mutations = desiredFolders
+            .stream()
+            .<FolderMutation>map(df ->
+                currentById.containsKey(df.id())
+                    ? new UpdateFolder((PortalNavigationFolder) currentById.get(df.id()), df)
+                    : new CreateFolder(df)
```
aqibmohammadkhan: `currentById` is `Map<PortalNavigationItemId, PortalNavigationFolder>` (inferred from `List<PortalNavigationFolder>.stream().collect(toMap(...)))`, so `.get()` already returns `PortalNavigationFolder`. The cast is redundant and can be removed.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/PortalResourceTest.java
```diff
@@ -73,10 +75,27 @@ void should_return_portal_by_hrid() {
                     soft.assertThat(state.getId()).isEqualTo(PORTAL_ID.toString());
                     soft.assertThat(state.getHrid()).isEqualTo(HRID);
                     soft.assertThat(state.getName()).isEqualTo("Default Portal");
+                    soft.assertThat(state.getNavigation()).isNullOrEmpty();
                 });
             }
         }
 
+        @Test
+        void should_return_navigation_alongside_portal() {
```
aqibmohammadkhan, line 84: this test is declared at the outer class level, not inside the Get nested class where all other GET tests live. It tests GET behaviour and should be moved inside Get.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/common/HRIDToUUID.java
```diff
@@ -60,6 +60,10 @@ public static TopLevelBuilder portal() {
         return new TopLevelBuilder();
     }
 
+    public static TopLevelBuilder navFolder() {
+        return new TopLevelBuilder();
+    }
+
```
aqibmohammadkhan: this is is identical to `HRIDToUUID.portal()`.

Both return `new TopLevelBuilder()`. The only namespace differentiation is at the call site via the TOP_NAVBAR: prefix. The new method name implies a separate mechanism that doesn't exist. Either rename it to something that makes the namespace explicit, or drop it and document that callers must namespace the HRID themselves (as portal() callers already do with "portal:" prefixes implicitly).

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/PortalMapper.java
konrad-krol-incubly: @aqibmohammadkhan - both method parameters have full type parameters today, even in the example that is linked with your comments. The method signature you have proposed is identical to the current one.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
konrad-krol-incubly: @aqibmohammadkhan The runtime values are filtered to `type=FOLDER` upstream (in `PortalNavigationSyncDomainService.sync`), but at compile time `currentById.get(...)` returns `PortalNavigationItem`. The cast to `PortalNavigationFolder` is required to construct `UpdateFolder(PortalNavigationFolder existing, ...)`.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/PortalResourceTest.java
konrad-krol-incubly: @aqibmohammadkhan The test sits between lines 83 and 97, fully inside Get. I think that you misread the diff.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/PortalNavigationTreeWalker.java
konrad-krol-incubly: Added a segment field on PortalNavigationItem (with `getEffectiveSegment()` falling back to title for legacy rows). `PortalNavigationTreeWalker.traverse` now builds the child parentPath from `node.getEffectiveSegment()` — displayName overrides no longer corrupt downstream paths.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/common/HRIDToUUID.java
konrad-krol-incubly: You're right, this shouldn't have been added. Deleted in the amended commit. As part of switching the planner to path-based diff (see reply on `NavigationSyncPlanner.java`), the planner no longer needs deterministic IDs for folders — the executor uses `PortalNavigationItemId.random()` like the console does.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
konrad-krol-incubly: Removed HRIDToUUID import from the planner in the amended commit.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationListingDomainService.java
konrad-krol-incubly: PathCollector.visitFolder now emits `displayName = title` when `title != segment`, and null when they match — so GET losslessly round-trips what was PUT.

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/PortalNavigationItem.java
```diff
@@ -59,6 +59,9 @@ public enum Visibility {
 
     private String title;
 
+    /** Canonical path segment used to reconstruct hierarchical paths; falls back to {@code title} for legacy rows. */
```
dev-marek, line 62: Instead of implementing fallback please create an upgrader to fill segment

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationListingDomainService.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private static final class PathCollector implements PortalNavigationVisitor {
+
+        private final List<NavigationPath> paths = new ArrayList<>();
+
+        @Override
+        public void visitFolder(PortalNavigationFolder folder, String parentPath) {
+            final var segment = folder.getEffectiveSegment();
+            final var path = parentPath + "/" + segment;
+            final var displayName = !segment.equals(folder.getTitle()) ? folder.getTitle() : null;
```
dev-marek, line 58: can you please explain meaning of this null?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItem.java
```diff
@@ -43,6 +43,14 @@ public abstract sealed class PortalNavigationItem
     @Nonnull
     private String title;
 
+    /**
+     * Canonical path segment used to reconstruct hierarchical paths. Optional on read for legacy rows;
+     * readers should fall back to {@link #title} when {@code segment} is null.
+     */
+    @Setter
+    @Nullable
```
dev-marek, line 51: We should create upgrader instead

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationListingDomainServiceTest.java
```diff
@@ -0,0 +1,105 @@
...
+        syncService = new PortalNavigationSyncDomainService(crud, query, pageContentCrud);
+        listingService = new PortalNavigationListingDomainService(query);
+    }
+
+    @Test
+    void list_as_navigation_paths_round_trips_input_order() {
+        syncService.sync(
+            AUDIT_INFO,
+            List.of(
+                new NavigationPath("/a", null),
```
dev-marek, line 60: For navigation path I can't see any reason why it should start from `/`

dev-marek (changes requested):
- [ ] I would like to see segment name being created using some kind of slug instead of display name, which I believe can contain `/` character

konrad-krol-incubly:
Closing draft, replaced with sequence of target PRs:
https://github.com/gravitee-io/gravitee-api-management/pull/17542
https://github.com/gravitee-io/gravitee-api-management/pull/17543
https://github.com/gravitee-io/gravitee-api-management/pull/17544

PR #17434, opened 2026-06-03
fix(rest-api): return clean 4xx  on v4 remote API import

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-12140
## What
  v4 remote-URL import (`_import/definition-url`) returned HTTP 500 with leaked exception
  details on fetch/validation failures. Now returns clean, user-facing 4xx.

  - **Fetch failures** (remote 404, unreachable host, DNS) → `400` via new
    `ApiDefinitionFetchException` (generic message, cause kept for server logs only).
  - **Valid JSON that isn't a Gravitee definition** (e.g. `{"hello":"world"}`) → `400`
    "Invalid API definition" instead of an NPE → 500 (`RemoteApiDefinitionParser` now
    rejects a null/missing `api`).

  ## OpenAPI path
  Verified the `_import/swagger` URL path already returns a clean `400` for the same
  failures (no 500/leak); added regression tests to lock it in.

  ## Tests
  Both definition-url resource suites updated (404 / unreachable / non-definition →
  4xx + no leaked internals); new OpenAPI cases in `OAIDomainServiceImplTest`. All green.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/RemoteApiDefinitionParser.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/RemoteApiDefinitionParser.java
@@ -53,7 +53,9 @@ public class RemoteApiDefinitionParser {
      * plain Java methods (not via JAX-RS dispatch), which would otherwise bypass {@code @Valid} constraints
      * on those endpoint parameters.
      *
-     * @throws BadRequestException if the content is not parseable JSON. The Jackson exception detail is
+     * @throws BadRequestException if the content is not parseable JSON, or if it is valid JSON that is not a
+     *     Gravitee export (e.g. {@code {"hello":"world"}} deserializes into an {@link ExportApiV4} with a null
+     *     {@code api}, which would otherwise NPE in the downstream mapper). The Jackson exception detail is
      *     logged at warn level but never returned to the caller, to avoid leaking internal class paths or
      *     fragments of the remote body through the error message.
      * @throws ConstraintViolationException if the parsed definition fails Bean Validation.
@@ -65,6 +67,10 @@ public class RemoteApiDefinitionParser {
         } catch (JsonProcessingException e) {
             log.warn("Failed to parse API definition fetched from URL", e);
             throw new BadRequestException("Invalid API definition format");
+        }
+        if (apiToImport == null || apiToImport.getApi() == null) {
+            log.warn("API definition fetched from URL is valid JSON but is missing the required 'api' object");
+            throw new BadRequestException("Invalid API definition");
         }
         Set<ConstraintViolation<ExportApiV4>> violations = VALIDATOR.validate(apiToImport);
         if (!violations.isEmpty()) {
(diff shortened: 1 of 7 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/FetchApiDefinitionFromUrlDomainServiceImpl.java
```diff
@@ -46,12 +46,12 @@ public String fetch(String url, List<String> whitelist, boolean allowPrivate) {
             String body = buffer == null ? "" : buffer.toString();
```
carlos-andres-osorio, line 46: The `buffer == null` branch is not covered by any test. When `httpClientService.request()` returns `null`, the null guard substitutes an empty string and hands it to `parseAndValidate`, which will produce a 400 — but no test exercises this path, so a future regression here would go undetected.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/RemoteApiDefinitionParser.java
```diff
@@ -66,6 +68,10 @@ public ExportApiV4 parseAndValidate(String apiDefinitionContent) {
             log.warn("Failed to parse API definition fetched from URL", e);
             throw new BadRequestException("Invalid API definition format");
         }
+        if (apiToImport == null || apiToImport.getApi() == null) {
+            log.warn("API definition fetched from URL is valid JSON but is missing the required 'api' object");
+            throw new BadRequestException("Invalid API definition");
+        }
         Set<ConstraintViolation<ExportApiV4>> violations = VALIDATOR.validate(apiToImport);
```
carlos-andres-osorio, line 75: The `ConstraintViolationException` catch block is not covered by any test. When the parsed `ExportApiV4` has a non-null `api` field that fails Bean Validation constraints, execution reaches this branch — but no test exercises it, so a regression here would go undetected.

carlos-andres-osorio (approved):
Two minor gaps noted as inline comments (untested `null` buffer path and untested `ConstraintViolationException` path in the parser) — no blockers. The core fix is solid: fetch failures and invalid payloads now return clean 400s instead of 500s with leaked internals. LGTM.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/FetchApiDefinitionFromUrlDomainServiceImpl.java
paulatulis: Done

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api/RemoteApiDefinitionParser.java
paulatulis: Done

PR #17442, opened 2026-06-04
feat: show access control on plan form - backend

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-14269

## Summary

This change adds **backend support for “Groups excluded” on API Product plans**, matching the behaviour that already exists for **API plans**.

An API Product plan can list user groups whose members **must not see or subscribe to that plan**. The UI and OpenAPI already expose `excludedGroups`; this PR makes the backend **enforce** that setting.

---

## Problem (before this PR)

For **API plans**, excluded groups already worked end-to-end:

- Plans are **filtered** when listed
- Subscriptions are **blocked** if the user belongs to an excluded group

For **API Product plans**, the field was stored in the database but **not enforced**:

- `SubscriptionServiceImpl` had an explicit skip: *“For API Product plans, excluded groups check is not applicable”*
- Plan search for API Products did **not** filter by excluded groups
- Create/update did **not** validate that excluded group IDs exist in the environment

So configuring **Access-Control → Groups excluded** on an API Product plan had **no effect** on who could see or subscribe to that plan.

---

## How “Groups excluded” works (API Product)

**Groups excluded** is a **per-plan** setting. It does **not** change gateway security (API Key, OAuth, etc.). It only controls **who can see and subscribe to that plan** in the Console / Management API.

A user is blocked from a plan if they belong to a listed excluded group **with an API Product role** on that group membership.

Being an admin **inside** the excluded group does **not** bypass the rule. The check looks at group membership + role scope, not the role name.

### Who is always allowed? (bypass)

| Case | Result |
|------|--------|
| Plan has **no** excluded groups | Everyone who can access the API Product sees the plan |
| User is a **direct member** of the API Product (Members tab) | Always allowed for that plan, even if they are in an excluded group |
| User is **Environment Admin** | Can still create subscriptions (same as API plans) |
| User is **Organization Admin** | Sees all plans in Management API (`isAdmin()` bypass on list) |

---

## Files changed

| File | Change |
|------|--------|
| `GroupService.java` | Add `isUserAuthorizedToAccessApiProductData(...)` |
| `GroupServiceImpl.java` | Implement API Product authorization rules |
| `SubscriptionServiceImpl.java` | Enforce excluded groups for `API_PRODUCT` plans on subscribe |
| `PlanSearchQueryServiceImpl.java` | Filter API Product plans for non-admin users; inject `ApiProductsRepository` + `GroupService` |
| `PlanSearchServiceImpl.java` | Filter in `searchForApiProductPlans()` |
| `ApiProductPlansResource.java` | Validate excluded groups on create |
| `ApiProductPlanResource.java` | Validate excluded groups on update |
| `GroupService_IsUserAuthorizedToAccessTest.java` | Unit tests for API Product authorization |
| `PlanSearchQueryServiceImplTest.java` | Unit tests for plan list filtering |
| `PlanSearchServiceImpl_SearchTest.java` | Unit tests for v4 plan search filtering |

**Not changed (already in place):**

- OpenAPI (`openapi-api-products.yaml`) — `excludedGroups` already defined
- Plan persistence (`JdbcPlanRepository`, `plan_excluded_groups` table)
- Core plan model (`Plan`, `PlanUpdates`) — field already mapped
- Console UI — Access-Control field already shown on API Product plan form

---

### Setup

1. Create API Product with at least one published API
2. Create group **`excluded-grp`** and group **`allowed-grp`**
3. Attach both groups to the API Product
4. Create user **`blocked-user`**:
   - Member of **`excluded-grp`** only
   - **API Product Role = USER** on group membership
   - **Not** a direct member under API Product → Members
5. Create two **API Key** plans (Keyless is not allowed on API Products):
   - **Plan A** — exclude **`excluded-grp`**
   - **Plan B** — no excluded groups
6. Publish both plans

### Expected results

| Action | User | Expected |
|--------|------|----------|
| Console → API Product → Subscriptions → Create | `blocked-user` | Only **Plan B** in plan dropdown |
| Same | user in `allowed-grp` only | **Plan A** and **Plan B** |
| Same | direct API Product member in `excluded-grp` | **Both plans** (bypass) |
| POST subscription to Plan A | `blocked-user` | **Fails** (plan restricted) |
| POST subscription to Plan B | `blocked-user` | **Succeeds** |

## Additional context
https://github.com/user-attachments/assets/08985cdc-0aab-495b-ba06-05c5291cdc52

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/domain_service/PlanExcludedGroupsDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/domain_service/PlanExcludedGroupsDomainService.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.plan.domain_service;
+
+import io.gravitee.apim.core.DomainService;
+import io.gravitee.apim.core.api_product.model.ApiProduct;
+import io.gravitee.apim.core.group.model.Group;
+import io.gravitee.apim.core.group.query_service.GroupQueryService;
+import io.gravitee.apim.core.membership.model.Membership;
+import io.gravitee.apim.core.membership.model.Role;
+import io.gravitee.apim.core.membership.query_service.MembershipQueryService;
+import io.gravitee.apim.core.membership.query_service.RoleQueryService;
+import io.gravitee.rest.api.service.exceptions.GroupNotFoundException;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Validates plan excluded groups and resolves whether a user can access an API Product plan
+ * based on group memberships, mirroring the rules used for API plan access control.
+ */
+@DomainService
+@RequiredArgsConstructor
+public class PlanExcludedGroupsDomainService {
+
+    private final GroupQueryService groupQueryService;
+    private final MembershipQueryService membershipQueryService;
+    private final RoleQueryService roleQueryService;
+
+    public void validateExcludedGroups(String environmentId, List<String> excludedGroups) {
+        if (excludedGroups == null || excludedGroups.isEmpty()) {
+            return;
+        }
+
+        Set<String> validGroupIds = groupQueryService
+            .findByIds(new HashSet<>(excludedGroups))
+            .stream()
+            .filter(group -> environmentId.equals(group.getEnvironmentId()))
+            .map(Group::getId)
+            .collect(Collectors.toSet());
+
+        excludedGroups.forEach(excludedGroupId -> {
+            if (!validGroupIds.contains(excludedGroupId)) {
+                throw new GroupNotFoundException(excludedGroupId);
+            }
+        });
+    }
+
+    public boolean isUserAuthorizedToAccessApiProductPlan(ApiProduct apiProduct, List<String> excludedGroups, String username) {
+        if (username == null) {
+            return excludedGroups == null || excludedGroups.isEmpty();
+        }
+
+        if (excludedGroups == null || excludedGroups.isEmpty()) {
+            return true;
+        }
+
+        if (isDirectApiProductMember(apiProduct.getId(), username)) {
+            return true;
+        }
+
+        Set<String> groupsWithApiProductRole = findUserGroupsWithApiProductRole(username);
+
+        if (apiProduct.getGroups() != null && !apiProduct.getGroups().isEmpty()) {
+            Set<String> authorizedGroups = new HashSet<>(apiProduct.getGroups());
+            authorizedGroups.removeAll(excludedGroups);
+            return authorizedGroups.stream().anyMatch(groupsWithApiProductRole::contains);
+        }
+
+        return excludedGroups.stream().noneMatch(groupsWithApiProductRole::contains);
+    }
+
+    private boolean isDirectApiProductMember(String apiProductId, String username) {
+        return membershipQueryService
+            .findByMemberIdAndMemberTypeAndReferenceType(username, Membership.Type.USER, Membership.ReferenceType.API_PRODUCT)
+            .stream()
+            .anyMatch(membership -> apiProductId.equals(membership.getReferenceId()));
+    }
+
+    private Set<String> findUserGroupsWithApiProductRole(String username) {
+        Collection<Membership> groupMemberships = membershipQueryService.findGroupsThatUserBelongsTo(username);
+        Set<String> roleIds = groupMemberships.stream().map(Membership::getRoleId).filter(Objects::nonNull).collect(Collectors.toSet());
+
+        if (roleIds.isEmpty()) {
+            return Set.of();
+        }
+
+        Set<String> apiProductRoleIds = roleQueryService
+            .findByIds(roleIds)
+            .stream()
+            .filter(role -> Role.Scope.API_PRODUCT.equals(role.getScope()))
+            .map(Role::getId)
+            .collect(Collectors.toSet());
+
+        return groupMemberships
+            .stream()
+            .filter(membership -> apiProductRoleIds.contains(membership.getRoleId()))
+            .map(Membership::getReferenceId)
+            .collect(Collectors.toSet());
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/use_case/CreateApiProductPlanUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/use_case/CreateApiProductPlanUseCase.java
@@ -22,6 +22,7 @@ import io.gravitee.apim.core.api_product.crud_service.ApiProductCrudService;
 import io.gravitee.apim.core.api_product.model.ApiProduct;
 import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.apim.core.plan.domain_service.CreatePlanDomainService;
+import io.gravitee.apim.core.plan.domain_service.PlanExcludedGroupsDomainService;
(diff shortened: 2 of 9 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api_product/ApiProductPlanResource.java
```diff
@@ -65,6 +70,9 @@ public class ApiProductPlanResource extends AbstractResource {
     @Inject
     private PlanOperationsUseCase planOperationsUseCase;
 
+    @Inject
```
vikrantgravitee: group/excluded-group validation belongs in that domain/use-case layer and not in resource

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api_product/ApiProductPlanResource.java
```diff
@@ -95,6 +103,7 @@ public Response updateApiProductPlan(@PathParam("planId") String planId, @Valid
         final ExecutionContext executionContext = GraviteeContext.getExecutionContext();
 
         var userDetails = getAuthenticatedUserDetails();
+        validateExcludedGroups(updatePlan.getExcludedGroups(), executionContext.getEnvironmentId());
```
vikrantgravitee: It is used at 2 places in this PR, could we think to put in some generic service or use case.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api_product/ApiProductPlansResource.java
```diff
@@ -70,6 +73,9 @@ public class ApiProductPlansResource extends AbstractResource {
     @Inject
     private GetPlansUseCase getPlansUseCase;
 
+    @Inject
```
vikrantgravitee: Same as above

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/plan/PlanSearchQueryServiceImpl.java
```diff
@@ -39,9 +42,17 @@
 public class PlanSearchQueryServiceImpl implements PlanSearchQueryService {
 
     private final PlanRepository planRepository;
+    private final ApiProductsRepository apiProductsRepository;
+    private final GroupService groupService;
```
vikrantgravitee: Thinking to avoid using GroupService here.
Could we do the excluded-groups authorization in GetPlansUseCase using the core MembershipQueryService instead, and keep this adapter purely for data fetching.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/plan/PlanSearchQueryServiceImpl.java
```diff
@@ -75,6 +88,13 @@ public List<Plan> searchPlans(
                 }
```
vikrantgravitee, line 75: On product-not-found, the filter treats apiProductEntity == null as allow-all (no filtering), while SubscriptionServiceImpl denies (orElse(false)) - same condition, opposite outcome, and it can surface restricted plans. 
The API path avoids this by passing an already-resolved entity (404 up-front). 

Suggestion: resolve the product once up-front and deny/empty consistently in both search and subscribe.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/GroupServiceImpl.java
```diff
@@ -967,6 +967,46 @@ public boolean isUserAuthorizedToAccessApiData(GenericApiEntity api, List<String
         return false;
     }
 
+    @Override
+    public boolean isUserAuthorizedToAccessApiProductData(ApiProductEntity apiProduct, List<String> excludedGroups, String username) {
```
vikrantgravitee: This new isUserAuthorizedToAccessApiProductData adds product auth to the legacy service and re-derives membership rules. 

Suggestion: drop it and resolve the user's membership context once in GetPlansUseCase via the core MembershipQueryService primitives that ApiProductAccessibleIdsDomainService already uses (findByMemberIdAndMemberTypeAndReferenceType + findGroupsThatUserBelongsTo), then diff each plan's excludedGroups in-memory. 

Reuse the pattern/primitives (not findAccessibleApiProductIds directly - it resolves accessibility, not exclusion). Additive, no interface change → no breaking change.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api_product/ApiProductPlanResource.java
ankita-gupta21: **Done**. Removed `validateExcludedGroups` from `ApiProductPlansResource` and `ApiProductPlanResource`. Validation now lives in `PlanExcludedGroupsDomainService`, called from `CreateApiProductPlanUseCase` and `UpdateApiProductPlanUseCase` so create/update share one implementation.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api_product/ApiProductPlanResource.java
ankita-gupta21: Done. Removed `validateExcludedGroups` from `ApiProductPlansResource` and `ApiProductPlanResource`. Validation now lives in `PlanExcludedGroupsDomainService`, called from `CreateApiProductPlanUseCase` and `UpdateApiProductPlanUseCase` so create/update share one implementation.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/plan/PlanSearchQueryServiceImpl.java
ankita-gupta21: Fixed. Both list and subscribe now resolve the API Product up-front via ApiProductCrudService.get(referenceId). If the product doesn’t exist we fail consistently (404 on get), instead of allow-all on list vs deny on subscribe.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/plan/PlanSearchQueryServiceImpl.java
ankita-gupta21: **Done**. `PlanSearchQueryServiceImpl` is fetch-only again. Excluded-group auth moved to `GetPlansUseCase` + `PlanExcludedGroupsDomainService` (core `MembershipQueryService` + `RoleQueryService`). `SubscriptionServiceImpl` reuses the same service.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api_product/ApiProductPlansResource.java
ankita-gupta21: Done

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/GroupServiceImpl.java
ankita-gupta21: removed `isUserAuthorizedToAccessApiProductData` from `GroupService`. Auth now lives in `PlanExcludedGroupsDomainService` using `MembershipQueryService` + `RoleQueryService` (same primitives as `ApiProductAccessibleIdsDomainService`). `GetPlansUseCase` resolves the product up-front and filters `excludedGroups` in memory; `SubscriptionServiceImpl` reuses the same service. Added `PlanExcludedGroupsDomainServiceTest`.

vikrantgravitee (approved):

PR #17471, opened 2026-06-04
feat: run upgraders contributed by Gamma module plugins

gcusnieux:
Gamma modules had no way to run data migrations at startup. The node's `UpgraderService` collects `Upgrader` beans from the node context, but a Gamma module's beans live in the JAX-RS context, so its upgraders were never picked up.

The handler now runs the upgraders a module contributes itself, right after the plugin context is created. Same tracking as the node (`UpgraderRepository`), so each one runs once. A failing upgrader is logged and skipped instead of breaking the module load.

First user is the AIM module, verified locally end to end (the upgrader runs on startup and the data gets migrated). Unit tests cover the handler: an upgrader runs once, is skipped when already applied, and a failing one doesn't break the module.

https://gravitee.atlassian.net/browse/GMA-537

### DIFF
--- a/gravitee-gamma/gravitee-gamma-plugin/gravitee-gamma-plugin-module/gravitee-gamma-plugin-module-handler/src/main/java/io/gravitee/plugin/gamma/internal/GammaModulePluginHandler.java
+++ b/gravitee-gamma/gravitee-gamma-plugin/gravitee-gamma-plugin-module/gravitee-gamma-plugin-module-handler/src/main/java/io/gravitee/plugin/gamma/internal/GammaModulePluginHandler.java
@@ -18,12 +18,17 @@ package io.gravitee.plugin.gamma.internal;
 import static io.gravitee.plugin.gamma.internal.GammaModulePlugin.PLUGIN_TYPE;
 
 import io.gravitee.apim.plugin.gamma.api.GammaModule;
+import io.gravitee.node.api.upgrader.UpgradeRecord;
+import io.gravitee.node.api.upgrader.Upgrader;
+import io.gravitee.node.api.upgrader.UpgraderRepository;
 import io.gravitee.plugin.core.api.AbstractPluginHandler;
 import io.gravitee.plugin.core.api.Plugin;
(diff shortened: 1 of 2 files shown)


gravitee-gamma/gravitee-gamma-plugin/gravitee-gamma-plugin-module/gravitee-gamma-plugin-module-handler/src/main/java/io/gravitee/plugin/gamma/internal/GammaModulePluginHandler.java
```diff
@@ -115,12 +120,50 @@ private void initPluginSpringContext(Plugin plugin) {
...
+            // Gamma module's upgraders are run here instead, once each, tracked by the shared UpgraderRepository.
+            runPluginUpgraders(plugin, pluginCtx);
         } catch (Exception ex) {
             logger.error("Failed to create plugin context for Gamma module '{}'", plugin.id(), ex);
             pluginContextFactory.remove(plugin);
         }
     }
 
+    // A failing upgrader is logged, never propagated, so it does not unload the module.
+    private void runPluginUpgraders(Plugin plugin, ApplicationContext pluginCtx) {
```
jhaeyaert, line 134: I don't like the fact that the module itself runs the upgraders. 
Can you check if we can adopt another strategy similar to what we do with the repository plugin: create the upgrade instance and register it as a bean in the Spring parent context?

If this works, the upgrader service should be capable of fetching all the upgrader beans, including those coming from the Gamma module.

jhaeyaert (changes requested):

gcusnieux:
You were right. I tested it: declaring the upgrader as a `@Bean` is enough. 

The handler already registers the module's beans into the parent context, and the node UpgraderService picks it up and runs it once (tracked in `upgrades`). So no change is needed here, I'm closing this. The migration now lives entirely in the AIM module.

Thanks for the pointer.

PR #17481, opened 2026-06-04
fix: keep v2 api groups as ids for non-kubernetes origin

phiz71:
## Issue

https://gravitee.atlassian.net/browse/APIM-14317

## Description

After upgrading to 4.11.9 / 4.10.16 / 4.9.21 / 4.8.28, changing the groups of a V2 API made any further save fail with `These groups [xxx, yyy] do not exist.`

The root cause is in `ValidateGroupsDomainService`: for V2 APIs the resolved groups were always converted to their **names**. That is only correct for the Kubernetes operator flow; for console/management updates the groups must stay as **ids**. Once a group had been persisted by name, the next validation could no longer resolve it and rejected the save.

The validation now maps groups to names **only** for the `kubernetes` origin. The origin is plumbed through `ValidateGroupsDomainService.Input` and its callers (`ApiValidationServiceImpl`, `ValidateApiCRDDomainService`, `UpdateApiGroupsUseCase`, `ValidatePortalNotificationDomainService`).

## Additional context

Reproduction steps from the ticket:
1. Have a V2 API with groups
2. Add or remove a group on the API
3. Perform any other change (e.g. enable logging) and save
4. Before the fix: save fails with `These groups [...] do not exist.`
5. After the fix: save succeeds

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApiValidationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApiValidationServiceImpl.java
@@ -107,6 +107,7 @@ public class ApiValidationServiceImpl extends AbstractService implements ApiVali
                     executionContext.getEnvironmentId(),
                     api.getGroups(),
                     api.getGraviteeDefinitionVersion(),
+                    api.getDefinitionContext() != null ? api.getDefinitionContext().getOrigin() : null,
                     Group.GroupEvent.API_CREATE,
                     false
                 )
@@ -118,6 +119,7 @@ public class ApiValidationServiceImpl extends AbstractService implements ApiVali
                 new ValidatePortalNotificationDomainService.Input(
                     api.getConsoleNotificationConfiguration(),
                     api.getGraviteeDefinitionVersion(),
+                    api.getDefinitionContext() != null ? api.getDefinitionContext().getOrigin() : null,
                     api.getGroups(),
                     auditInfo
                 )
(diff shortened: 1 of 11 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApiValidationServiceImpl.java
```diff
@@ -107,6 +107,7 @@ public ApiValidationResult<ApiCRDEntity> validateAndSanitizeApiDefinitionCRD(Exe
                     executionContext.getEnvironmentId(),
                     api.getGroups(),
                     api.getGraviteeDefinitionVersion(),
+                    api.getDefinitionContext().getOrigin(),
```
mukul-tyagi08: how about ? api.getDefinitionContext() != null ? api.getDefinitionContext().getOrigin() : null

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApiValidationServiceImpl.java
```diff
@@ -118,6 +119,7 @@ public ApiValidationResult<ApiCRDEntity> validateAndSanitizeApiDefinitionCRD(Exe
                 new ValidatePortalNotificationDomainService.Input(
                     api.getConsoleNotificationConfiguration(),
                     api.getGraviteeDefinitionVersion(),
+                    api.getDefinitionContext().getOrigin(),
```
mukul-tyagi08: how about ? api.getDefinitionContext() != null ? api.getDefinitionContext().getOrigin() : null

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/group/domain_service/ValidateGroupsDomainServiceTest.java
mukul-tyagi08, line 1: Maybe we can add a test for the actual regression like should_sanitize_v2_groups_as_ids_for_management_origin

mukul-tyagi08 (approved):

PR #17483, opened 2026-06-05
feat(gamma-platform-ui): Use sheets instead of modals in Gamma platform application UI

gaurav-gravitee:
## Issue

## Description

- Refactor dialogs to sheets: Replaced various application-level modals (certificates, metadata, notifications, subscriptions, user permissions, ownership transfer) with sheet/side-panel components for a more consistent and modern UX.
- Introduce new sheet components & tests: Added *Sheet components (e.g. ApplicationAddCertificateSheet, EditMetadataSheet, EditNotificationSheet, ApplicationSubscriptionCreateSheet, AddMembersSheet, EditRoleSheet, ManageGroupsSheet, TransferOwnershipSheet) along with dedicated *.spec.tsx tests and shared sheetSpecHelpers.
- Update sections and pages: Adjusted application sections and pages (certificates, metadata, notifications, subscriptions, user permissions, notification settings) to use the new sheet-based flows and extracted helpers (e.g. notificationHelpers) where appropriate.
- Cleanup legacy dialogs: Removed or renamed deprecated *Dialog components and their tests, consolidating related logic into the new sheet implementations.

## Additional context

Video: 

https://github.com/user-attachments/assets/ba592c40-c783-4ecf-abc6-46efc15164db


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/user-permissions/ManageGroupsSheet.tsx
```diff
@@ -48,15 +49,13 @@ export function ManageGroupsDialog({
...
-            setSelected(new Set(currentGroupIds));
-            setSearch('');
+    useEffect(() => {
+        if (!open) {
+            return;
         }
-    }
+        setSelected(new Set(currentGroupIds));
+        setSearch('');
+    }, [open, currentGroupIds]);
```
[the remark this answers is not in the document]
gaurav-gravitee: I've updated ManageGroupsSheet to only reset when the sheet is opened.

michel-barret (approved):

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/user-permissions/ManageGroupsSheet.tsx
```diff
@@ -48,15 +49,13 @@ export function ManageGroupsDialog({
...
-            setSelected(new Set(currentGroupIds));
-            setSearch('');
+    useEffect(() => {
+        if (!open) {
+            return;
         }
-    }
+        setSelected(new Set(currentGroupIds));
+        setSearch('');
+    }, [open]);
```
jourdiw: **[Score: 45/100]** 💡 The `useEffect` resets `selected` from `currentGroupIds` but only depends on `[open]`. If the parent updates `currentGroupIds` while the sheet is already open (e.g., a background mutation invalidates queries), the sheet won't reflect the new group set — it will keep the stale snapshot.

Consider adding `currentGroupIds` to the dependency array (or using a stable serialized value like `JSON.stringify(currentGroupIds)`):

```tsx
useEffect(() => {
    if (!open) return;
    setSelected(new Set(currentGroupIds));
    setSearch('');
}, [open, currentGroupIds]);
```

This is unlikely to be triggered in normal user flows, but it's a correctness issue if the data source can change asynchronously.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/subscriptions/ApplicationSubscriptionCreateSheet.tsx
```diff
@@ -0,0 +1,449 @@
...
+    useEffect(() => {
+        if (!selectedReference) {
+            setSelectedPlan(null);
+            return;
+        }
+        if (!selectedPlan || !plans.find(p => p.id === selectedPlan.id)) {
+            const firstEnabled = plans.find(plan => !isPlanDisabled(plan));
+            setSelectedPlan(firstEnabled ?? null);
+        }
+    }, [selectedReference, plans, selectedPlan]);
```
jourdiw: **[Score: 40/100]** 💡 `selectedPlan` appears in both the effect body (as a read + write target) and the dependency array. When `setSelectedPlan` fires, the effect re-runs because `selectedPlan` changed — it terminates only because the condition is now false on the second pass. This works but is fragile:

1. It triggers an unnecessary second evaluation of the effect on every plan change.
2. Future edits might accidentally introduce an infinite loop.

A safer alternative is to track the plan selection via the reference change only, and use a functional update or a ref to avoid listing `selectedPlan` in deps:

```tsx
useEffect(() => {
    if (!selectedReference) {
        setSelectedPlan(null);
        return;
    }
    setSelectedPlan(prev => {
        if (prev && plans.some(p => p.id === prev.id)) return prev;
        return plans.find(plan => !isPlanDisabled(plan)) ?? null;
    });
}, [selectedReference, plans]);
```

This removes the circular dependency while preserving the intended behavior.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/subscriptions/ApplicationSubscriptionCreateSheet.tsx
```diff
@@ -0,0 +1,449 @@
...
+                                                    className="h-auto w-full justify-start gap-3 rounded-none border-b px-3 py-2.5 font-normal last:border-b-0 hover:bg-accent"
+                                                    onClick={() => selectReference(item)}
+                                                >
+                                                    <div className="min-w-0 flex-1 text-left">
+                                                        <p className="truncate text-sm font-medium">
+                                                            {item.value.name}
+                                                            {version ? ` — ${version}` : ''}
+                                                            {owner ? ` (${owner})` : ''}
+                                                        </p>
+                                                        <p className="text-[11px] text-muted-foreground">{searchResultTypeLabel(item)}</p>
```
jourdiw: **[Score: 30/100]** 🎨 `text-[11px]` is an arbitrary value outside the Graphene type scale. The closest design token is `text-xs` (12px).

Per `graphene.md`, arbitrary values should be avoided to stay aligned with the system:

```tsx
<p className="text-xs text-muted-foreground">{searchResultTypeLabel(item)}</p>
```

The 1px difference is unlikely to be visually meaningful and keeping to the scale makes it easier to maintain consistent typography.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/notifications/EditNotificationSheet.tsx
```diff
@@ -0,0 +1,282 @@
...
+        },
+        [onCancel],
+    );
+
+    return (
+        <Sheet open={row !== null} onOpenChange={handleOpenChange}>
+            <SheetContent
+                side="right"
+                className="flex max-h-full flex-col"
+                style={{ width: 'min(98vw, 51.84rem)', maxWidth: 'min(98vw, 51.84rem)' }}
```
jourdiw: **[Score: 25/100]** 🎨 The width `51.84rem` is a magic number that doesn't correspond to the Graphene spacing/layout scale. All other sheets in this PR use `maxWidth: '480px'` which, while also not a token, is a common UI convention (30rem).

Consider:
- Using a well-known fixed width like `640px` / `40rem` or `768px` / `48rem` if the wider content is justified here.
- Or extracting a shared constant (e.g., `WIDE_SHEET_WIDTH`) so the intent is documented and reusable.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/user-permissions/ManageGroupsSheet.tsx
gaurav-gravitee: resolved, thanks!

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/subscriptions/ApplicationSubscriptionCreateSheet.tsx
gaurav-gravitee: resolved, thanks!

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/subscriptions/ApplicationSubscriptionCreateSheet.tsx
gaurav-gravitee: resolved, thanks!

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/notifications/EditNotificationSheet.tsx
gaurav-gravitee: resolved, thanks!

jourdiw (approved):

PR #17497, opened 2026-06-05
fix(console): fix sidebar navigation race condition for Observability and Analytics

araffass:
## Issue

https://gravitee.atlassian.net/browse/APIM-14285

## Description

Fixed a race condition in the Console sidebar where clicking Observability or Analytics would silently redirect to the Dashboard instead of navigating to the correct page. 
The issue was caused by two independent subscriptions racing to rebuild the sidebar menu on init: when environment settings resolved from cache before the current environment was set, the navigation links ended up with an invalid path. Fixed by chaining the two operations so the environment is always set first.

### DIFF
--- a/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
+++ b/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
@@ -15,8 +15,8 @@
  */
 import { Component, Inject, OnDestroy, OnInit } from '@angular/core';
 import { GioLicenseService, GioMenuSearchService, LicenseOptions, MenuSearchItem, SelectorItem } from '@gravitee/ui-particles-angular';
-import { distinctUntilChanged, map, takeUntil } from 'rxjs/operators';
-import { Observable, of, Subject } from 'rxjs';
+import { catchError, distinctUntilChanged, map, switchMap, takeUntil } from 'rxjs/operators';
+import { EMPTY, Observable, of, Subject } from 'rxjs';
 import { ActivatedRoute, Router } from '@angular/router';
 
 import { GioPermissionService } from '../../shared/components/gio-permission/gio-permission.service';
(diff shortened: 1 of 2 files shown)


aditya-goyal01:
Please add a test to prevent future regressions.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
```diff
@@ -75,29 +75,28 @@ export class GioSideNavComponent implements OnInit, OnDestroy {
...
           this.gioMenuSearchService.removeMenuSearchItems([SIDE_NAV_GROUP_ID]);
           this.gioMenuSearchService.addMenuSearchItems(this.getSideNaveMenuSearchItems());
-        },
+
+          return this.environmentSettingsService.get();
+        }),
+        takeUntil(this.unsubscribe$),
+      )
+      .subscribe(envSettings => {
+        this.mainMenuItems = this.buildMainMenuItems(envSettings);
```
mukul-tyagi08: I am just wondering that the main menu items are only built once environmentSettingsService.get() emits. This may leave the side nav without menu items until the asynchronous call resolves like if that call is slow or fails.

mukul-tyagi08: Also, this.environmentSettingsService.get() is now inside the main stream's switchMap, any error emitted by that HTTP/service call (e.g., a network hiccup or a 500 error) will propagate up and permanently kill the outer subscription. How about we catch the error inside the switchMap so it returns an empty or fallback observable instead of breaking the main stream.

mukul-tyagi08 (commented):
I am just wondering that the main menu items are only built once environmentSettingsService.get() emits. This may leave the side nav without menu items until the asynchronous call resolves like if that call is slow or fails.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
araffass: Valid,  i'll fix by adding catchError(() => EMPTY) inside the switchMap. If environmentSettingsService.get() fails, the menu stays as-is (built without settings) and the outer subscription survives for future env changes

araffass: Already handled: buildMainMenuItems() is called without settings on line 82 inside the switchMap, so the menu is immediately populated. Once settings resolve, it's rebuilt with settings-dependent items (API Score)

mukul-tyagi08 (approved):

PR #17511, opened 2026-06-05
feat: allow groups to be patched on the v4 API PATCH endpoint

carlos-andres-osorio:
## Issue

[APIM-14050](https://gravitee.atlassian.net/browse/APIM-14050)

## Why

The v4 `PATCH /api/{apiId}` endpoint applies an explicit allow-list of patchable fields, and `groups` was not on it — any patch touching `groups` was rejected with `"field is not patchable"`. Group assignments could only be changed through the dedicated `PUT /api/{apiId}/groups` endpoint. This brings the generic v4 PATCH endpoint to parity with the full-update path, so a team can change an API's group assignments through the same patch request used for every other field.

## What changed

All production changes land in `PatchApiUseCase` (`gravitee-apim-rest-api-service`):

- **Allow-list** — added the `groups` field to the patchable allow-list.
- **Current-state node** — added `groups` to the `PatchableView` record so JSON-Patch `replace`/`remove` on `/groups` has a path to target (merge-patch tolerates an absent path; JSON-Patch does not).
- **Apply patched value** — the patched `groups` value now flows through via the same `resolveNullableSet` helper used for `tags`/`categories`, instead of being hard-preserved from the existing API.

Validation and sanitization come for free: the persist (`updateV4`) and dry-run (`validateV4`) paths already run `GroupValidationService` unconditionally, so PATCH inherits the exact `PUT /api/{id}` full-update semantics (ID-based group validation, API-context-aware primary-owner-group handling). No new validation wiring, Kubernetes guard, or audit code was needed.

The OpenAPI operation description for `PATCH /api/{apiId}` was updated to list `groups` among the patchable fields.

## User-facing impact

- **Behaviour change**: `groups` is now patchable on `PATCH /api/{apiId}` (v4) for both JSON-Patch (RFC 6902) and JSON-Merge-Patch (RFC 7386) bodies. Field omitted → unchanged; explicit `null` / JSON-Patch `remove /groups` → cleared; array → that set, sanitized downstream. The response `ApiV4` reflects the sanitized `groups`.
- **Intentional semantics**: PATCH rides the full-update path (`GroupValidationService`), not the stricter dedicated `PUT /groups` path — group refs are resolved by ID, the API's own primary-owner group is retained, and a foreign primary-owner group is discarded by sanitization.
- **Not breaking**: additive change (one new allow-list entry). Previously-rejected `groups` patches simply start succeeding.
- **Documentation**: the OpenAPI spec description is updated in this PR. Permission parity and Kubernetes-managed-API handling are out of scope and tracked separately.

## How to test

Against a v4 HTTP Proxy API:

- Merge-patch set: `PATCH` body `{"groups":["<groupId>"]}` → `200`, `ApiV4.groups` contains the group.
- Merge-patch clear: `{"groups":null}` → `200`, groups cleared.
- JSON-Patch remove: `[{"op":"remove","path":"/groups"}]` → `200`, groups cleared.
- Foreign primary-owner group: patch in a group that is the primary-owner group of a different API → that group is discarded (sanitized), matching `PUT /api/{id}`.

## Known Issues

- [F17806657940319] MINOR — `gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/UpdateApiDomainServiceImpl.java:107` — The dry-run path resolves groups via `coalesce(sanitized.getGroups(), original.getGroups())`, falling back to the original groups when the sanitizer returns null. A `dryRun=true` clear-groups preview can therefore still show the pre-patch groups, diverging from what the real update persists (an empty set). The file is not edited by this change; the pattern is pre-existing and shared with `categories`/`labels`, but this change is the first to make it reachable for `groups`. Tracked separately.

[APIM-14050]: https://gravitee.atlassian.net/browse/APIM-14050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
@@ -359,7 +359,7 @@ paths:
                 Partially update a V4 HTTP Proxy API using JSON Patch (RFC 6902) or JSON Merge Patch (RFC 7396).
 
                 Only the following fields are patchable: name, description, apiVersion, visibility, labels, tags,
-                lifecycleState, categories, analytics, failover, flowExecution, flows, services, resources, endpointGroups,
+                lifecycleState, categories, groups, analytics, failover, flowExecution, flows, services, resources, endpointGroups,
                 allowedInApiProducts, allowMultiJwtOauth2Subscriptions, disableMembershipNotifications,
                 properties, responseTemplates, listeners.
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiGroupsTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApiResource_PatchApiGroupsTest.java
@@ -1,0 +1,175 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.management.v2.rest.resource.api;
+
+import static assertions.MAPIAssertions.assertThat;
+import static io.gravitee.common.http.HttpStatusCode.BAD_REQUEST_400;
+import static io.gravitee.common.http.HttpStatusCode.OK_200;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doAnswer;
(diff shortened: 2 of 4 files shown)


paulatulis (approved):
No blockers. 


The sanitization ACs (7 & 8) aren't covered by automated tests — by design, but worth noting. In both test classes, updateV4/validateV4 are mocked to return their argument, so the real primary-owner-group sanitization (AC7 "foreign discarded", AC8 "own retained") is never exercised here. patched_groups_flow_through_to_update_unsanitized deliberately asserts the unsanitized value reaches updateV4. This is a reasonable "don't re-test inherited behavior" stance, but it means AC7/AC8 rest entirely on the assumption that the PUT /api/{id} path has its own coverage. Fine to accept; just flagging that the PR's own tests don't prove those two ACs.

carlos-andres-osorio:
> No blockers. 
> 
> 
> The sanitization ACs (7 & 8) aren't covered by automated tests — by design, but worth noting. In both test classes, updateV4/validateV4 are mocked to return their argument, so the real primary-owner-group sanitization (AC7 "foreign discarded", AC8 "own retained") is never exercised here. patched_groups_flow_through_to_update_unsanitized deliberately asserts the unsanitized value reaches updateV4. This is a reasonable "don't re-test inherited behavior" stance, but it means AC7/AC8 rest entirely on the assumption that the PUT /api/{id} path has its own coverage. Fine to accept; just flagging that the PR's own tests don't prove those two ACs.

Right — mocking the sanitizer here is deliberate, and the PUT-path coverage you mention is in place for both cases. `updateV4`/`validateV4` route through `ApiValidationServiceImpl.validateAndSanitizeUpdateApi` → `GroupValidationServiceImpl.removePrimaryOwnerGroups`, and that class is directly tested in `GroupValidationServiceImplTest`: foreign primary-owner group discarded (`shouldReturnFilteredGroupsWithNoApiAndDefaultGroupWithApiPrimaryOwner`, plus the `...WithApiPrimaryOwnerAndUserPrimaryOwner`/`...AndGroupPrimaryOwner` variants) and the API's own primary-owner group retained (`shouldReturnValidatedGroupsWithApiAndDefaultGroupsWithApiPrimaryOwnerAndGroupPrimaryOwner`). The use-case test stays at the layer it owns — that `groups` flows through unsanitized to `updateV4`/`validateV4` — so AC7/AC8 are proven where the logic lives.

PR #17508, opened 2026-06-05
feat(authz): engine-parsed GAPL schema (single source of truth)

podlesrafal:
> **Stacked on #17482** (`feat/authz-editable-schema`). Review/merge that first; this PR targets its branch and should be retargeted to the default branch once #17482 lands.

## What
Make the authorization **engine's grammar the single source of truth** for parsing/validating the editable GAPL schema, replacing a hand-rolled, namespace-blind frontend parser.

**Backend** (`gravitee-gamma-module-authz`)
- `AuthzSchemaParsing` wraps the engine's `AuthzSchemaParser` + `SchemaJsonExporter`.
- Validate **on save** (`PUT /schema` → 400 before persisting) and **live** (`POST /schema/validate`).
- Serve engine-parsed JSON via `GET /schema/parsed`; `parsedSchema` returns `{}` for blank/unparseable stored data (no 500 on the read path).
- Engine dep promoted to compile scope, pinned to `1.0.0-alpha.14` (carries the namespace fix). The engine artifact resolves with no transitive deps, so `antlr4-runtime` 4.13.2 is pinned and bundled in the plugin.

**Frontend**
- `engineSchemaToParsed` maps the engine's namespace-grouped JSON to the UI model, resolving bare references **Cedar-style** (local namespace first, then global) — fixes namespaced types being mis-qualified (e.g. a global `Subject` referenced inside `myapp` no longer becomes `myapp::Subject`).
- `useParsedSchema` / `useSchemaValidation` (debounced); Save is gated on pending validation and surfaces the backend's rejection.
- `SchemaPage` / `action-options` / `ServicePolicyPage` consume the parsed model.
- Removed the namespace-blind `gapl-parser.ts` + its `schema-validation` helper (kept `gapl-policy-parser` and the Monaco language).

## Why
The hand-rolled FE parser diverged from the engine grammar (notably namespace handling), and there was no backend schema validation. One grammar now drives parsing, validation, and the UI.

## Verification
- Tests green: backend module 220, FE 509, `tsc` clean.
- AuthZEN parity (reference engine alpha.14 vs Gamma/PDP) agrees on authorization decisions: eval 29/29, batch 29/29, search-subject 29/29, search-resource 28/29. (The search-action gap and one tag scenario are parity-harness provisioning limitations — it forwards neither an action catalog nor entity tags to the Gamma side — not behaviour of this change.)
[...]

PR #17482, opened 2026-06-04
feat(authz): editable GAPL schema

podlesrafal:
## Issue

https://gravitee.atlassian.net/browse/APIM-XXX 

## Description

The per-environment GAPL schema used to be **generated on read** from entities/policies (prefix-capitalised, e.g. `entity Mcp {}`) and was **read-only**. It is now an **independent, opt-in, user-authored document** with full CRUD, living only in the authz module (it was never published to the gateway/PDP, and still isn't).

Backend:
- New `AuthzSchemaRepository` port + `authz_schemas` Mongo store (one doc per env, `_id = environmentId`) + in-memory test double.
- `AuthzSchemaServiceImpl` is now a thin CRUD wrapper (`getSchema` / `saveSchema` / `deleteSchema`); the generator (`build`/`typeNameFor`/cache) and its `AuthzEntityRepository`/`AuthzPolicyRepository` deps are deleted.
- `GET /schema` (stored or empty), `PUT /schema` (create/update), `DELETE /schema`, with `ENVIRONMENT_AUTHORIZATION` READ/UPDATE/DELETE respectively.
- Removed the now-dead `invalidate` plumbing: its 9 call sites and the `schemaService` dependency dropped from `AuthzEntityServiceImpl`/`AuthzPolicyServiceImpl` (it was injected solely for cache invalidation).

Frontend:
- `updateSchema`/`deleteSchema` API calls + `useUpdateSchema`/`useDeleteSchema` mutation hooks.
- Pure `schemaDiagnostics` helper (client-side GAPL validation; blank rejected).
- `SchemaPage` gains create / edit / delete with Monaco, Save gated on validation diagnostics.

Behaviour change (intended): domains lose the auto-generated read-only schema view and start blank until a schema is created.

## Test plan

- [x] Backend: `mvn -o test` on schema + affected suites → 96/96 green (JDK 21).
- [x] Frontend: `vitest` schema-validation / useUpdateSchema / new SchemaPage tests green; `tsc` clean.
- [ ] Reviewer note: `SchemaPage.test.tsx` has **2 pre-existing KPI failures on `master`** (`'3 entities'` vs rendered `'3Entities…'`), unrelated to this PR — no new failures introduced.

PR #17468, opened 2026-06-04
chore(authz-ui): remove unused buildPolicyEntityRefs helper

podlesrafal:
## Issue

N/A — dead-code cleanup.

## Description

`buildPolicyEntityRefs` in `shared/policy-entity-refs.ts` had no callers anywhere in the module (only its own test referenced it). This removes:

- `buildPolicyEntityRefs` function
- its `PolicyRef` interface and the `entityKey` private helper
- the now-unused imports (`PolicyResponse`, `EntityInstance`, `formatEntityUid`)
- the `buildPolicyEntityRefs` test suite + its `makeEntity`/`makePolicy` helpers

Kept (still used by the policy adapter in `authz-api.service.ts`):
- `extractEntityRefsFromPolicyText`
- `deriveTargetEntityId`

## Additional context

- `tsc` clean; `vitest` for `policy-entity-refs.test.ts` 12/12 (extract + derive suites); `nx lint` (eslint + license) clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #17513, opened 2026-06-06
feat(authz): granular entityType — write-time derive + read path

podlesrafal:
> **Stacked on #17508** (→ which stacks on #17482). Review/merge those first; retarget to the default branch as they land.

## What
Make `entityType` the granular, authoritative type for authz entities — populated at write time and read by the UI — and align parent UIDs across the stack.

**Write-time derive (P1)**
- `AuthzEntityIdConstants.engineTypeForHint` maps `_kind`/prefix → canonical engine type (`User`/`Group`/`MCPServer`/`API`/…).
- `CreateOrReplaceAuthzEntityCommand` derives granular `entityType` when none is supplied — so **every** inbound path flows through one point: REST create, and AM sync (which passes `entityType=null` + `_kind`). Regression test asserts AM sync yields `User`/`Group`/`Role`/`AgentIdentity`.
- FE create dialog sources its type list from the engine-parsed schema and sends `entityType`.

**Parent-UID normalization (P2)**
- `toEngineUid`/`normalizeParentUid` (SSOT in definition-model); the gateway sync re-quotes parents to the canonical `Type::"id"`. (The matching PDP change — dropping the `parseParentUid` workaround — is gravitee-service-authz-pdp#7.)

**Read path (P3)**
- `EntityResponse.entityType` surfaced; `resolveEntityRef`/`fromBackend`/the entity picker prefer it, falling back to `_kind`/uid-prefix only for legacy entities. The prefix/`_kind` derivation tower stays as that fallback — its removal is deferred until existing data is migrated (P4) and the in-flight FE PRs land.

## Verification
- Backend: definition-model 8, module-authz 41, AM-sync suite 21. FE: 518. `tsc` clean.
- AuthZEN parity (reference engine alpha.14 ↔ Gamma/PDP) agrees on authorization decisions incl. **group membership** (eval/batch/search-subject 29/29, search-resource 28/29) with the PDP `parseParentUid` workaround removed. (search-action / one tag scenario are parity-harness provisioning gaps, not this change.)

## Not in this PR
- **P4** — migrating existing umbrella (`Principal`/`Resource`) entities to granular (breaking, coordinated; fixes the live umbrella-vs-typed-policy mismatch).
- Full deletion of the FE derivation tower (after P4 + once #17468/#17287 land).

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzEntityMapper.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzEntityMapper.java
@@ -22,6 +22,7 @@ import io.gravitee.gamma.definition.authz.AuthzEntityKind;
 import io.gravitee.gateway.services.sync.process.common.model.SyncAction;
 import io.gravitee.repository.management.model.Event;
 import io.reactivex.rxjava3.core.Maybe;
+import java.util.List;
 import lombok.CustomLog;
 import lombok.RequiredArgsConstructor;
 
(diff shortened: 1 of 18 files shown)


podlesrafal:
Superseded — migrated to **gravitee-gamma-module-authz#10** (granular entityType, write-time derive + read path). The type/UID helpers now live in the module (no definition-model change needed) and parent normalization happens at write time, so the APIM-side split (#17573) was also closed. Closing without merging.

PR #17514, opened 2026-06-06
feat(authz): migrate umbrella entityTypes to granular (admin endpoint)

podlesrafal:
> **Stacked on #17513** (→ #17508 → #17482). Retarget down the stack as they land. **Breaking-adjacent:** changes stored entity data when applied — run the dry-run first.

## What
A one-time backfill that re-types legacy umbrella entities (`Principal`/`Resource`) to granular ones, fixing the live mismatch where typed policies (`Group::"engineering"`, `API::"api.payments"`) don't match umbrella-stored entities.

- `POST /entities/migrate-types?apply=false` (default **dry-run**) → `{scanned, migrated, applied, changes:[{entityId, from, to}]}` — the inventory to review.
- `apply=true` persists: re-derives `entityType` by clearing it and reusing `CreateOrReplaceAuthzEntityCommand`'s `_kind`/prefix derivation, only where the result differs from the stored umbrella. `entityId` is never touched. Applied in bounded batches. Idempotent. UPDATE permission.

## Why this design
Spec requires "inventory first, coordinated, not silent" — hence dry-run default + explicit apply. Only umbrella rows with a derivable hint change (legacy null-typed artifacts); already-granular and hint-less umbrellas are left alone.

## Verification
- Backend: module-authz service 36 (incl. PRINCIPAL `_kind`→User, RESOURCE prefix `api.`→API, already-granular/no-hint untouched, dry-run persists nothing, apply persists, idempotent), resource 16, permission-contract 4.

## Follow-up (not here)
- Run the migration in each env (dry-run → review → apply).
- After data is migrated, delete the FE prefix/`_kind` derivation tower (the fallback becomes dead) — also gated on #17468/#17287 landing.

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/api/AuthzEntityAdminApi.java
+++ b/gravitee-gamma/gravitee-gamma-module-authz/src/main/java/io/gravitee/gamma/authorization/api/AuthzEntityAdminApi.java
@@ -43,4 +43,10 @@ public interface AuthzEntityAdminApi {
     Set<String> findApiAliases(String environmentId, String apiId);
 
     AuthzCascadeResult delete(AuthzCallerContext caller, String entityId);
+
+    /**
+     * Backfill umbrella entityTypes to granular ones, re-deriving from {@code _kind}/prefix.
+     * {@code apply=false} reports the changes without persisting (dry-run inventory).
+     */
+    AuthzEntityTypeMigrationReport migrateEntityTypes(AuthzCallerContext caller, boolean apply);
 }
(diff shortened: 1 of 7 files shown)


podlesrafal:
Dropping this PR. The runtime migration endpoint was the wrong shape for a one-off concern: re-typing existing umbrella entities is not an ongoing product capability and shouldn't live as a permanent admin API. New and synced entities already get the granular type at write time (#17513, P1), and the rare pre-P1 umbrella row can be upgraded by re-saving it (POST upsert without an explicit entityType re-derives). Existing umbrella entities are valid and evaluate fine — they just won't match a policy written against a granular type, which is acceptable for the current data. Granular entityType work continues on #17513.

PR #17542, opened 2026-06-08
feat(portal-page): add segment column to PortalNavigationItem

konrad-krol-incubly:
## Issue

## Description

### What changed

Adds a segment field to PortalNavigationItem — a URL-safe path component stored alongside the human-readable title. This separation is necessary because titles can contain arbitrary characters (e.g. "Getting Started / Overview") that would corrupt hierarchical path reconstruction if used  directly as path segments.                                                                                                                                                                                                                                               

Slugification (`NavigationSegmentSlugifier`) normalises segments to lowercase alphanumeric + hyphens at write time, both in the application layer (from() factory, update()) and in the migrations. This makes PUT/GET idempotent: the same input always produces the same stored segment regardless of how the title is phrased.                                                                                                                                                                                                                                  

Migrations handle existing rows:
  - Liquibase changeset backfills segment from title with basic normalisation (/ and spaces → -) for JDBC databases                                                                                                                                                        
  - A Mongo upgrader applies the same slugification logic in Java for MongoDB 
  
### Why it's needed

The next PR introduces a path-based navigation sync domain service. That service reconstructs and compares folder paths using segments — it needs every row to carry a clean, slash-free segment before any sync logic runs.


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItem.java
```diff
@@ -90,6 +100,10 @@ protected PortalNavigationItem(
 
     public abstract PortalNavigationItemType getType();
 
+    public String getEffectiveSegment() {
+        return segment != null ? segment : title;
+    }
```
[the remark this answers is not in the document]
konrad-krol-incubly: Fixed.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/NavigationSegmentSlugifier.java
```diff
@@ -0,0 +1,26 @@
...
+package io.gravitee.apim.core.portal_page.model;
+
+/** Converts arbitrary text into a URL-safe path segment (lowercase alphanumeric + hyphens). */
+public final class NavigationSegmentSlugifier {
+
+    private NavigationSegmentSlugifier() {}
+
+    public static String slugify(String value) {
+        return value.toLowerCase().replaceAll("[^a-z0-9]+", "-").replaceAll("^-|-$", "");
+    }
```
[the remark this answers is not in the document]
konrad-krol-incubly: fixed

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/14_backfill_segment_from_title_on_portal_navigation_items.yml
```diff
@@ -0,0 +1,10 @@
+databaseChangeLog:
+  - changeSet:
+      id: 4.12.0_14_backfill_segment_from_title_on_portal_navigation_items
+      author: GraviteeSource Team
+      changes:
+        - sql:
+            sql: >-
+              UPDATE ${gravitee_prefix}portal_navigation_items
+              SET segment = LOWER(REPLACE(REPLACE(title, '/', '-'), ' ', '-'))
+              WHERE segment IS NULL
```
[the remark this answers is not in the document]
konrad-krol-incubly: Fixed as much as reasonably possible. The SQL handles the two most common unsafe characters (/ and spaces). A Java-based customChange would be the first of its kind in this codebase - all JDBC migrations use plain SQL - so I intentionally kept it consistent with the existing approach.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/14_backfill_segment_from_title_on_portal_navigation_items.yml
```diff
@@ -0,0 +1,10 @@
+databaseChangeLog:
+  - changeSet:
+      id: 4.12.0_14_backfill_segment_from_title_on_portal_navigation_items
+      author: GraviteeSource Team
+      changes:
+        - sql:
+            sql: >-
+              UPDATE ${gravitee_prefix}portal_navigation_items
```
dev-marek: We prefer service upgraders instead of DML in changelogs. Please see for reference gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ApplicationClientCertificateMigrationUpgrader.java

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItem.java
```diff
@@ -43,6 +45,14 @@ public abstract sealed class PortalNavigationItem
     @Nonnull
     private String title;
 
+    /**
+     * Canonical path segment used to reconstruct hierarchical paths. Optional on read for legacy rows;
+     * readers should fall back to {@link #title} when {@code segment} is null.
+     */
+    @Setter
+    @Nullable
```
dev-marek: After upgraders being run, it is no longer valid

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/NavigationSegmentSlugifier.java
```diff
@@ -13,8 +13,15 @@
...
+package io.gravitee.apim.core.portal_page.model;
 
-export function querySheetHeading(name: string | RegExp) {
-    return screen.queryByRole('heading', { name });
+/** Converts arbitrary text into a URL-safe path segment (lowercase alphanumeric + hyphens). */
+public final class NavigationSegmentSlugifier {
+
+    private NavigationSegmentSlugifier() {}
+
+    public static String slugify(String value) {
```
dev-marek: Instead of creating utility class, I'd expect PortalNavigationItem to have `slug()` method

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItem.java
```diff
@@ -151,10 +166,17 @@ public static PortalNavigationItem from(
         return newItem;
     }
 
+    private static String segmentFor(CreatePortalNavigationItem item, String titleFallback) {
+        return slugify(item.getSegment() != null ? item.getSegment() : titleFallback);
```
dev-marek: for fallback, shouldn't it be something like parent.segment + `/` + slugify(titleFallback)?

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/13_add_segment_to_portal_navigation_items.yml
```diff
@@ -0,0 +1,13 @@
+databaseChangeLog:
+  - changeSet:
+      id: 4.12.0_13_add_segment_to_portal_navigation_items
+      author: GraviteeSource Team
+      changes:
+        - addColumn:
+            tableName: ${gravitee_prefix}portal_navigation_items
+            columns:
+              - column:
+                  name: segment
```
dev-marek, line 10: should be unique for environment?

dev-marek (changes requested):

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/14_backfill_segment_from_title_on_portal_navigation_items.yml
konrad-krol-incubly: Done — replaced DML changeset with `PortalNavigationItemDefaultSegmentUpgrader`, following the `PortalNavigationItemRootIdUpgrader` pattern.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItem.java
konrad-krol-incubly: `segmentFor` returns a leaf segment, not a full path, so prepending the parent path would change the field's semantics. Leaving as-is.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/13_add_segment_to_portal_navigation_items.yml
konrad-krol-incubly: Done — added composite unique constraint on `(environment_id, parent_id, segment)`. Please note that segment represents just the leaf path, so it has to be unique against environment and its parent.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItem.java
konrad-krol-incubly: Done. The constraint is now `@Nonnull`, and `getEffectiveSegment()` returns `segment` directly.

This cascaded into test fixtures — every builder call that omits `.segment()` would throw at build time, so all shared fixtures (`PortalNavigationItemFixtures`, `PortalNavigationItemsRepositoryFixtures`) and inline test helpers now supply a slugified segment.

The production assumption is that the `PortalNavigationItemDefaultSegmentUpgrader` backfills all legacy `null` rows at startup before any domain object is instantiated from a repository record.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/NavigationSegmentSlugifier.java
konrad-krol-incubly: Kept as a shared utility rather than inlining the logic in each class.

Two places need the same algorithm: `PortalNavigationItem` uses it to derive the `segment` field when creating or updating a navigation item; `NavigationPath` uses it to normalize each path component when building or accepting a hierarchical path. If both don't produce identical output for the same input, a title slugified at creation time and the same string slugified during path normalization in the sync planner will diverge — silently breaking folder matching.

Keeping the algorithm in one place makes that contract explicit and makes it trivial to evolve (e.g. locale-aware normalization, explicit allow-list) without risking a silent split.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItem.java
konrad-krol-incubly: To be explicit: `PortalNavigationItemDefaultSegmentUpgrader` **must run successfully** before this constraint is safe in production. If it fails or is skipped, loading any unbackfilled row from the DB will throw an NPE at builder construction time. The upgrader logs a warning per environment on failure so ops can detect it.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/NavigationSegmentSlugifier.java
```diff
@@ -13,21 +13,19 @@
...
-import org.junit.jupiter.api.Test;
-
-@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
-class PlanForApiNotFoundExceptionTest {
+/**
+ * Canonical slug format for portal navigation URL segments.
+ * Shared between segment derivation on {@link PortalNavigationItem} creation/update
+ * and path normalization in NavigationPath, so both always produce consistent results.
+ */
+public final class NavigationSegmentSlugifier {
```
dev-marek: I would create `Slug` value object with `from(String title, List<Slug> used)` factory method

Please note that this algorithm may produce same slug for `Questions / answers` and `Question Answers` titles so it would be a good idea to add suffix

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/CreatePortalNavigationItem.java
```diff
@@ -29,6 +29,7 @@ public final class CreatePortalNavigationItem {
 
     private PortalNavigationItemId id;
     private String title;
+    private String segment;
```
dev-marek, line 32: Please make sure it is set in existing create flow

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/UpdatePortalNavigationItem.java
```diff
@@ -27,6 +27,7 @@
 public final class UpdatePortalNavigationItem {
 
     private String title;
+    private String segment;
```
dev-marek, line 30: Please make sure it is set in update flows

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/CreatePortalNavigationItem.java
konrad-krol-incubly: Both covered without touching the use cases.

The webconsole use case intentionally does not carry `segment` — the webconsole UI has no field for it and exposing it was out of scope. Deriving it from the title is the right place for it, not the use case layer.

- **Create** — `PortalNavigationItem.from()` calls `segmentFor(item, title)`, which uses the explicit segment if provided, otherwise falls back to `slugify(title)`.
- **Update** — `PortalNavigationItem.update()` now always calls `segmentFor(navItem)`, re-deriving from the (potentially changed) title when no explicit segment is given. This keeps segment in sync when a user renames an item via the webconsole.

Tests for both paths added to `PortalNavigationItemDomainServiceTest`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/UpdatePortalNavigationItem.java
konrad-krol-incubly: Both covered without touching the use cases.

The webconsole use case intentionally does not carry `segment` — the webconsole UI has no field for it and exposing it was out of scope. Deriving it from the title is the right place for it, not the use case layer.

- **Create** — `PortalNavigationItem.from()` calls `segmentFor(item, title)`, which uses the explicit segment if provided, otherwise falls back to `slugify(title)`.
- **Update** — `PortalNavigationItem.update()` now always calls `segmentFor(navItem)`, re-deriving from the (potentially changed) title when no explicit segment is given. This keeps segment in sync when a user renames an item via the webconsole.

Tests for both paths added to `PortalNavigationItemDomainServiceTest`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/NavigationSegmentSlugifier.java
konrad-krol-incubly: I added deduplication to upgrader and both create/update flows.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItem.java
```diff
@@ -151,10 +169,23 @@ public static PortalNavigationItem from(
         return newItem;
     }
 
+    public static Slug slugify(String value) {
+        return Slug.from(value);
+    }
+
+    private static String segmentFor(CreatePortalNavigationItem item, String titleFallback) {
+        return slugify(item.getSegment() != null ? item.getSegment() : titleFallback).value();
```
dev-marek: segment is always set for create

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/13_add_segment_to_portal_navigation_items.yml
```diff
@@ -0,0 +1,13 @@
...
+                  type: nvarchar(256)
+                  constraints:
+                    nullable: true
```
dev-marek: as discussed, please make nullable false with default value replaced by upgrader

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/14_add_unique_segment_constraint_on_portal_navigation_items.yml
```diff
@@ -0,0 +1,9 @@
+databaseChangeLog:
+  - changeSet:
+      id: 4.12.0_14_unique_segment_per_sibling
+      author: GraviteeSource Team
+      changes:
+        - addUniqueConstraint:
+            tableName: ${gravitee_prefix}portal_navigation_items
+            columnNames: environment_id, parent_id, segment
```
dev-marek: it will break on environments with existing items without initialised segment - we should remove this constraint

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItem.java
konrad-krol-incubly: -> fixed — removed null guard and titleFallback param from segmentFor, domain service always sets segment before calling from()

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/13_add_segment_to_portal_navigation_items.yml
konrad-krol-incubly: -> done — nullable: false with defaultValue: __CHANGE_ME__ (never a valid slug), upgrader detects and replaces it

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/14_add_unique_segment_constraint_on_portal_navigation_items.yml
konrad-krol-incubly: -> done — removed the unique constraint changeset and its master.yml entry

dev-marek (approved):
lgtm

PR #17543, opened 2026-06-08
feat(portal): add portal navigation sync domain layer

konrad-krol-incubly:
## Issue

## Description

### What changed

  Introduces the domain layer that powers declarative navigation sync. The core idea: callers supply a flat list of hierarchical paths (e.g. ["`/getting-started`", "`/getting-started/quickstart`", "`/apis`"]) and the domain service reconciles that desired state against the current folder tree —  creating, updating, or deleting folders as needed.                                                                                                                                                                                                                       

  Sync planner (`NavigationSyncPlanner`) is a pure, side-effect-free diff engine. It expands the input paths into all implied ancestor folders, assigns sibling order, matches against current folders by reconstructed path (using the segment field from PR 1), and produces a typed plan of
  mutations. PathExpander slugifies each path component during expansion, so the planner always works with clean segments.                                                                                                                                                 

  Sync executor (`PortalNavigationSyncDomainService`) applies the plan — creating/updating folders and cascade-deleting removed ones along with all their descendants (including associated PortalPageContent for PAGE-type children).

  Listing service (`PortalNavigationListingDomainService`) reads the current folder tree back as a flat list of NavigationPath records, preserving displayName where the title differs from the segment. This is what GET returns.

  Also fixes a dormant Spring bean conflict in 4 test configs triggered by registering the new @DomainService beans.

###  Why it's needed
  
  This layer has no REST caller yet — PR 3 wires it into the use cases and resources. Keeping it separate makes the algorithmic logic reviewable on its own.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationListingDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationListingDomainService.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal.domain_service;
+
+import io.gravitee.apim.core.DomainService;
+import io.gravitee.apim.core.portal.domain_service.navigation.PortalNavigationTreeWalker;
+import io.gravitee.apim.core.portal.domain_service.navigation.PortalNavigationVisitor;
+import io.gravitee.apim.core.portal.model.NavigationPath;
+import io.gravitee.apim.core.portal_page.model.PortalArea;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationFolder;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemQueryCriteria;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemType;
+import io.gravitee.apim.core.portal_page.query_service.PortalNavigationItemsQueryService;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import lombok.RequiredArgsConstructor;
+
+@DomainService
+@RequiredArgsConstructor
+public class PortalNavigationListingDomainService {
+
+    private final PortalNavigationItemsQueryService queryService;
+
+    public List<NavigationPath> listAsNavigationPaths(String environmentId) {
+        final var folders = queryService.search(
+            PortalNavigationItemQueryCriteria.builder()
+                .environmentId(environmentId)
+                .area(PortalArea.TOP_NAVBAR)
+                .type(PortalNavigationItemType.FOLDER)
+                .build()
+        );
+        final var collector = new PathCollector();
+        PortalNavigationTreeWalker.walk(folders, collector);
+        return collector.paths;
+    }
+
+    private static final class PathCollector implements PortalNavigationVisitor {
+
+        private final List<NavigationPath> paths = new ArrayList<>();
+
+        @Override
+        public void visitFolder(PortalNavigationFolder folder, NavigationPath parentPath) {
+            final var nodePath = parentPath.descend(folder.getEffectiveSegment());
+            final var displayName = !folder.getEffectiveSegment().equals(folder.getTitle()) ? folder.getTitle() : null;
+            paths.add(new NavigationPath(nodePath.path(), Optional.ofNullable(displayName)));
+        }
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationSyncDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationSyncDomainService.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal.domain_service;
+
+import io.gravitee.apim.core.DomainService;
+import io.gravitee.apim.core.audit.model.AuditInfo;
+import io.gravitee.apim.core.portal.domain_service.navigation.NavigationFolderMapper;
+import io.gravitee.apim.core.portal.domain_service.navigation.actions.FolderActions;
+import io.gravitee.apim.core.portal.domain_service.navigation.actions.NavigationAction;
+import io.gravitee.apim.core.portal.domain_service.navigation.plan.NavigationSyncPlan;
+import io.gravitee.apim.core.portal.domain_service.navigation.plan.NavigationSyncPlanner;
+import io.gravitee.apim.core.portal.model.NavigationPath;
+import io.gravitee.apim.core.portal.model.PortalId;
+import io.gravitee.apim.core.portal_page.crud_service.PortalNavigationItemCrudService;
+import io.gravitee.apim.core.portal_page.crud_service.PortalPageContentCrudService;
+import io.gravitee.apim.core.portal_page.model.CreatePortalNavigationItem;
+import io.gravitee.apim.core.portal_page.model.PortalArea;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationFolder;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItem;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemContainer;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemQueryCriteria;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemType;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationPage;
+import io.gravitee.apim.core.portal_page.model.PortalVisibility;
+import io.gravitee.apim.core.portal_page.query_service.PortalNavigationItemsQueryService;
+import io.gravitee.rest.api.service.common.HRIDToUUID;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.RequiredArgsConstructor;
+
+@DomainService
+@RequiredArgsConstructor
+public class PortalNavigationSyncDomainService {
+
+    private static final PortalArea AREA = PortalArea.TOP_NAVBAR;
+    private static final int MAX_CASCADE_DEPTH = 50;
+
+    private final PortalNavigationItemCrudService crudService;
+    private final PortalNavigationItemsQueryService queryService;
+    private final PortalPageContentCrudService pageContentCrudService;
+
+    public void sync(AuditInfo auditInfo, PortalId portalId, List<NavigationPath> desired) {
+        final var currentFolders = queryService.search(
+            PortalNavigationItemQueryCriteria.builder()
+                .environmentId(auditInfo.environmentId())
+                .area(AREA)
+                .type(PortalNavigationItemType.FOLDER)
+                .build()
+        );
+        final var plan = NavigationSyncPlanner.plan(desired == null ? List.of() : desired, currentFolders);
+        execute(plan, auditInfo, portalId);
+    }
+
+    private void execute(NavigationSyncPlan plan, AuditInfo auditInfo, PortalId portalId) {
+        final var byPath = new HashMap<String, PortalNavigationItemContainer>();
+        plan.actions().forEach(action -> applyAction(action, byPath, auditInfo, portalId));
+    }
+
+    private void applyAction(
+        NavigationAction action,
+        Map<String, PortalNavigationItemContainer> byPath,
+        AuditInfo auditInfo,
+        PortalId portalId
+    ) {
+        switch (action) {
+            case FolderActions.FolderMutation m -> applyMutation(m, byPath, auditInfo, portalId);
+            case FolderActions.DeleteFolder d -> cascadeDelete(d.item(), auditInfo.environmentId(), 0);
+        }
+    }
+
+    private void applyMutation(
+        FolderActions.FolderMutation mutation,
+        Map<String, PortalNavigationItemContainer> byPath,
+        AuditInfo auditInfo,
+        PortalId portalId
+    ) {
+        final var df = mutation.desired();
+        final var parent = df.parentPath() == null ? null : byPath.get(df.parentPath());
+        final PortalNavigationFolder result = switch (mutation) {
+            case FolderActions.CreateFolder c -> createFolder(c.desired(), parent, auditInfo, portalId);
+            case FolderActions.UpdateFolder u -> applyUpdate(u.existing(), u.desired(), parent);
+        };
+        byPath.put(df.path(), result);
+    }
+
+    private PortalNavigationFolder createFolder(
+        FolderActions.DesiredFolder df,
+        PortalNavigationItemContainer parent,
+        AuditInfo auditInfo,
+        PortalId portalId
+    ) {
+        final var folderId = PortalNavigationItemId.of(
+            HRIDToUUID.navigation().context(auditInfo).portal(portalId.toString()).folder(df.path()).id()
+        );
+        final var create = CreatePortalNavigationItem.builder()
+            .id(folderId)
+            .title(df.title())
+            .segment(df.segment().value())
+            .area(AREA)
+            .type(PortalNavigationItemType.FOLDER)
+            .order(df.order())
+            .visibility(PortalVisibility.PUBLIC)
+            .published(true)
+            .build();
+        return (PortalNavigationFolder) crudService.create(
+            PortalNavigationItem.from(create, auditInfo.organizationId(), auditInfo.environmentId(), parent)
+        );
+    }
+
+    private PortalNavigationFolder applyUpdate(
+        PortalNavigationFolder existing,
+        FolderActions.DesiredFolder desired,
+        PortalNavigationItemContainer parent
+    ) {
+        if (NavigationFolderMapper.matches(existing, desired, parent == null ? null : parent.getId())) return existing;
+        NavigationFolderMapper.apply(desired, parent, existing);
+        return (PortalNavigationFolder) crudService.update(existing);
+    }
+
+    private void cascadeDelete(PortalNavigationItem item, String environmentId, int depth) {
+        if (depth > MAX_CASCADE_DEPTH) throw new IllegalStateException(
+            "Maximum portal navigation nesting level of %d exceeded".formatted(MAX_CASCADE_DEPTH)
+        );
+        queryService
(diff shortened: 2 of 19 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationSyncDomainService.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private void cascadeDelete(PortalNavigationItem item, String environmentId, int depth) {
+        if (depth > MAX_CASCADE_DEPTH) throw new IllegalStateException(
+            "Maximum portal navigation nesting level of %d exceeded".formatted(MAX_CASCADE_DEPTH)
+        );
+        queryService
+            .findByParentIdAndEnvironmentId(environmentId, item.getId())
+            .forEach(child -> cascadeDelete(child, environmentId, depth + 1));
+        if (item instanceof PortalNavigationPage page) pageContentCrudService.delete(page.getPortalPageContentId());
```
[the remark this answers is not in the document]
konrad-krol-incubly: fixed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/PortalNavigationTreeWalker.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private PortalNavigationTreeWalker() {}
+
+    public static void walk(List<PortalNavigationItem> items, PortalNavigationVisitor visitor) {
+        final Comparator<PortalNavigationItem> byOrder = Comparator.comparingInt(PortalNavigationItem::getOrder);
+        final var childrenByParent = items
+            .stream()
+            .filter(i -> i.getParentId() != null)
+            .collect(Collectors.groupingBy(PortalNavigationItem::getParentId));
+        childrenByParent.values().forEach(list -> list.sort(byOrder));
```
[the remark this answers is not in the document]
konrad-krol-incubly: fixed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationSyncDomainService.java
```diff
@@ -0,0 +1,147 @@
...
+        byPath.put(df.path(), result);
+    }
+
+    private PortalNavigationFolder createFolder(
+        FolderActions.DesiredFolder df,
+        PortalNavigationItemContainer parent,
+        AuditInfo auditInfo,
+        PortalId portalId
+    ) {
+        final var folderId = PortalNavigationItemId.of(
```
konrad-krol-incubly, line 107: Folder IDs here are deterministic - derived from (orgId, envId, portalId, "folder", path) via `HRIDToUUID.navigation()`.

These IDs are not surfaced through the Automation API response (GET returns navigation paths only), so the API is idempotent at the response level regardless of how IDs are minted.

Determinism is required for the upcoming changes: Portal Documentation and Portal Listing need to populate `parentId` from a user-supplied `location`. Deterministic IDs let them compute the target folder ID directly (no folder query) and give full GitOps apply-order independence - a doc PUT applied before the portal sync still references the correct future folder ID, so the page auto-becomes visible once the folder is materialized.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/actions/FolderActions.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public sealed interface FolderMutation extends FolderAction permits CreateFolder, UpdateFolder {
+        DesiredFolder desired();
+    }
+
+    public record CreateFolder(DesiredFolder desired) implements FolderMutation {}
+
+    public record UpdateFolder(PortalNavigationFolder existing, DesiredFolder desired) implements FolderMutation {}
+
+    public record DeleteFolder(PortalNavigationItem item) implements FolderAction {}
```
dev-marek, line 36: Are we sure we want to remove all folders that are unmanaged by GKO/Terraform?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private static String reconstructPath(PortalNavigationFolder folder, Map<PortalNavigationItemId, PortalNavigationFolder> byId) {
+        return PATH_DELIMITER + String.join(PATH_DELIMITER, walkSegmentChain(folder, byId));
+    }
+
+    /** Walks the parent chain from {@code folder} up to the root, returning segments in root-first order. */
+    private static Deque<String> walkSegmentChain(PortalNavigationFolder folder, Map<PortalNavigationItemId, PortalNavigationFolder> byId) {
+        final Deque<String> chain = new ArrayDeque<>();
+        PortalNavigationFolder cursor = folder;
```
dev-marek: There is `PortalNavigationItemContainer` interface, which is probably better to use as API items can also contain child items

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
```diff
@@ -0,0 +1,142 @@
...
+        while (cursor != null) {
+            chain.push(cursor.getEffectiveSegment());
+            cursor = cursor.getParentId() != null ? byId.get(cursor.getParentId()) : null;
```
dev-marek: I would rather have `cursor.isRoot()` method that checking if `parentId` is null

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private static String reconstructPath(PortalNavigationFolder folder, Map<PortalNavigationItemId, PortalNavigationFolder> byId) {
+        return PATH_DELIMITER + String.join(PATH_DELIMITER, walkSegmentChain(folder, byId));
+    }
+
+    /** Walks the parent chain from {@code folder} up to the root, returning segments in root-first order. */
+    private static Deque<String> walkSegmentChain(PortalNavigationFolder folder, Map<PortalNavigationItemId, PortalNavigationFolder> byId) {
+        final Deque<String> chain = new ArrayDeque<>();
+        PortalNavigationFolder cursor = folder;
+        while (cursor != null) {
```
dev-marek: what about graceful streams?

```java
Stream.iterate(
            folder, 
            Objects::nonNull,
            item -> item.hasParent() ? byId.get(item.getParentId()) : null
    ).collect(Collectors.toList());
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/NavigationFolderMapper.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Objects;
+
+public final class NavigationFolderMapper {
+
+    private NavigationFolderMapper() {}
+
+    public static boolean matches(
+        PortalNavigationFolder existing,
+        FolderActions.DesiredFolder desired,
+        PortalNavigationItemContainer parent
```
dev-marek: we're using only parent id - wdyt about passing it directly?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/NavigationFolderMapper.java
```diff
@@ -0,0 +1,49 @@
...
+            existing.getOrder() == desired.order() &&
+            Objects.equals(existing.getParentId(), expectedParentId)
+        );
+    }
+
+    public static void apply(FolderActions.DesiredFolder source, PortalNavigationItemContainer parent, PortalNavigationFolder target) {
+        target.setTitle(source.title());
+        target.setSegment(source.segment());
+        target.setOrder(source.order());
+        if (parent == null) target.markAsRoot();
```
dev-marek: is really prettier ok with if-else without braces?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/NavigationPath.java
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal.model;
+
+import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
+
+public record NavigationPath(@Nonnull String path, @Nullable String displayName) {
```
dev-marek: Is display name used outside of test class?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationListingDomainService.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private static final class PathCollector implements PortalNavigationVisitor {
+
+        private final List<NavigationPath> paths = new ArrayList<>();
+
+        @Override
+        public void visitFolder(PortalNavigationFolder folder, String parentPath) {
+            final var segment = folder.getEffectiveSegment();
+            final var path = parentPath + "/" + segment;
```
dev-marek: WDYT about using `NavigationPath` value object here and implementing method like `.descend(String child)`?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/NavigationPath.java
```diff
@@ -0,0 +1,38 @@
...
+    public NavigationPath {
+        if (!isValid(path)) {
```
dev-marek: This VO is used only in domain so we can assume, that provided value is validated before object construction

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
konrad-krol-incubly: The tree currently only has folders as containers and the walk calls `getEffectiveSegment()`, which lives on `PortalNavigationItem` not the interface. Using the interface here would require adding that method to it — happy to revisit when non-folder containers are introduced.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
konrad-krol-incubly: Done — added `isRoot()` / `hasParent()` to `PortalNavigationItem`; used in the refactored `walkSegmentChain`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
konrad-krol-incubly: Done — `walkSegmentChain` now uses `Stream.iterate`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/NavigationFolderMapper.java
konrad-krol-incubly: Done — `matches()` now accepts `@Nullable PortalNavigationItemId parentId` directly.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/NavigationFolderMapper.java
konrad-krol-incubly: Done — added braces to the if-else, anyway - prettier didn't complained here.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/NavigationPath.java
konrad-krol-incubly: It is used in `NavigationSyncPlanner.explicitDisplayNames()`, `PortalNavigationListingDomainService.PathCollector`, and `PortalMapper`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalNavigationListingDomainService.java
konrad-krol-incubly: Done — added `descend(String)` to `NavigationPath` and threaded it through the visitor instead of a raw String. The constructor also normalizes path segments on construction, so `PathExpander` no longer needs its own slug logic.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/NavigationPath.java
konrad-krol-incubly: Done — removed the guard from the compact constructor; VO is domain-internal and callers validate before construction.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/actions/FolderActions.java
konrad-krol-incubly: As agreed on slack with Benoit => automation api has ownership of the resources that are managed through this API. In this case its natural that it has to overwrite the whole navigation structure that is managed through this API. I've added video presenting this in the last PR: https://github.com/gravitee-io/gravitee-api-management/pull/17544

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/NavigationPath.java
dev-marek: Please then use `Optional<String>`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
```diff
@@ -0,0 +1,143 @@
...
+            .collect(Collectors.toMap(PortalNavigationItem::getId, f -> f));
+        final var result = new HashMap<String, PortalNavigationFolder>();
+        for (PortalNavigationFolder folder : byId.values()) {
+            result.put(reconstructPath(folder, byId), folder);
+        }
+        return result;
+    }
+
+    private static String reconstructPath(PortalNavigationFolder folder, Map<PortalNavigationItemId, PortalNavigationFolder> byId) {
+        return PATH_DELIMITER + String.join(PATH_DELIMITER, walkSegmentChain(folder, byId));
```
dev-marek, line 87: Follow up: do we need to start path with PATH_DELIMITER?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
```diff
@@ -0,0 +1,143 @@
...
+            .map(PortalNavigationFolder.class::cast)
+            .collect(Collectors.toMap(PortalNavigationItem::getId, f -> f));
+        final var result = new HashMap<String, PortalNavigationFolder>();
+        for (PortalNavigationFolder folder : byId.values()) {
+            result.put(reconstructPath(folder, byId), folder);
+        }
+        return result;
+    }
+
+    private static String reconstructPath(PortalNavigationFolder folder, Map<PortalNavigationItemId, PortalNavigationFolder> byId) {
```
dev-marek, line 86: WDYT about using `NavigationPath` VO?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/NavigationPath.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package io.gravitee.apim.core.portal.model;
+
+import io.gravitee.apim.core.portal_page.model.NavigationSegmentSlugifier;
+import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+public record NavigationPath(@Nonnull String path, @Nullable String displayName) {
```
dev-marek: What about List of Slug as a path and join them in toString?

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
konrad-krol-incubly: I've already replied that in another comment - that's how design was explained in PRD - all paths started from `/`. I wouldn't like to change this design. It's better if we keep some convention.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanner.java
konrad-krol-incubly: NavigationPath currently carries `displayName` alongside the path, which comes from the input spec — not from the reconstructed current tree. Using it as the `currentByPath` map key would break lookups since record equality includes `displayName`. A raw String (or a dedicated path-only VO which might be overkill here) is the right type for `reconstructPath()`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/NavigationPath.java
konrad-krol-incubly: The intent here is actually the opposite — `NavigationPath` is responsible for normalization (including slugification) so callers don't have to think about it. Slugs are not pre-computed at call sites; they're derived inside the compact constructor via `normalizePath()`. Storing `List<Slug>` internally would mean moving that same work into a factory method with no functional gain — the external API stays identical. I'd keep the current approach where `NavigationPath` owns the normalization contract.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/model/NavigationPath.java
konrad-krol-incubly: done

dev-marek (approved):

PR #17564, opened 2026-06-08
test(json-patch): cover out-of-bounds array-index add wrapping

carlos-andres-osorio:
## Issue

[APIM-12009](https://gravitee.atlassian.net/browse/APIM-12009)

## Why

The `JsonPatchServiceImpl` wraps `JsonPatchException` into `ValidationDomainException` so callers receive a domain-level error rather than a raw library exception. An out-of-bounds array-index `add` operation (e.g. `/labels/5` on a single-element array) triggers that exception path, but it was not covered by any test — leaving a gap in the regression safety net for this wrapping contract.

## What changed

Added one test case to `JsonPatchServiceImplTest` that applies an out-of-bounds array-index `add` patch and asserts:
- the call throws `ValidationDomainException` (not a raw library or runtime exception), and
- the target node is left unmutated after the failed patch.

[APIM-12009]: https://gravitee.atlassian.net/browse/APIM-12009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


paulatulis (approved):
lgtm!

PR #17573, opened 2026-06-09
feat(authz): granular entityType — definition-model helpers + gateway sync

podlesrafal:
APIM-side half of the granular-entityType change (splitting #17513). The authz-module half (write-time `entityType` derive + FE read path) moves to the standalone **gravitee-gamma-module-authz** repo in a separate PR.

## What this adds
- **`AuthzEntityIdConstants`** (shared `gamma-definition-model` artifact):
  - `engineTypeForHint(hint)` — maps a kind/source hint to the engine entity type.
  - `toEngineUid(entityType, entityId)` — builds the engine's canonical `Type::"id"` UID.
  - `normalizeParentUid(uid)` — normalizes a parent UID into that canonical form.
- **Gateway sync** (`AuthzEntityMapper`): normalizes parent UIDs via `normalizeParentUid` so the published entities carry engine-canonical parents.

## Why split
The module consumes these helpers as a **published artifact**, so they must land + release in APIM first; the standalone module PR then bumps `gravitee-apim.version` to pick them up.

## Tests
`AuthzEntityIdConstantsTest` — 8 tests, green locally (`mvn -pl gravitee-gamma/gravitee-gamma-definition-model test`). Gateway `AuthzEntityMapperTest` validated by CI.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzEntityMapper.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzEntityMapper.java
@@ -22,6 +22,7 @@ import io.gravitee.gamma.definition.authz.AuthzEntityKind;
 import io.gravitee.gateway.services.sync.process.common.model.SyncAction;
 import io.gravitee.repository.management.model.Event;
 import io.reactivex.rxjava3.core.Maybe;
+import java.util.List;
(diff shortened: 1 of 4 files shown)


podlesrafal:
Superseded — closing without merging.

Parent UID normalization is moving **into the authz module** (gravitee-gamma-module-authz) at write time, instead of the gateway publisher:
- The 3 helpers (engineTypeForHint / toEngineUid / normalizeParentUid) now live in the module (com.graviteesource.gamma.authorization.service.AuthzEngineUid), so the module no longer needs them from gravitee-gamma-definition-model.
- REST create + AM-sync normalize parents to canonical Type::"id" on write; the gateway publishes verbatim (master already does — the only delta here was normalizeParents).
- master gateway + definition-model stay unchanged, so this PR is no longer needed.

Module work: gravitee-gamma-module-authz#10 (now self-contained — un-blocked).

PR #17576, opened 2026-06-09
feat: send application invitation notifications

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/APIM-14143

Implements backend support for sending application invitation email notifications when `notify=true`.

## Changes

- Adds application invitation notification dispatch from the invitation creation flow.
- Sends notifications asynchronously using `EmailService.sendAsyncEmailNotification`.
- Builds registration token parameters through legacy `UserService`.
- Passes `confirmation_page_url` through the backend flow as `URI` 
- Adds application and role template parameters for the invitation email.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/mapper/ApplicationInvitationsCreateInputMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/mapper/ApplicationInvitationsCreateInputMapper.java
@@ -18,7 +18,6 @@ package io.gravitee.rest.api.portal.rest.mapper;
 import io.gravitee.apim.core.invitation.model.CreateApplicationInvitations;
 import io.gravitee.rest.api.portal.rest.model.InvitationCreateInput;
 import io.gravitee.rest.api.portal.rest.model.InvitationRecipientInput;
-import java.net.URI;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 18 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/mapper/ApplicationInvitationsCreateInputMapper.java
```diff
@@ -57,7 +57,7 @@ private String normalizeRoleName(String roleName) {
         return roleName == null ? null : roleName.trim();
     }
 
-    private String normalizeConfirmationPageUrl(URI confirmationPageUrl) {
-        return confirmationPageUrl == null || confirmationPageUrl.toString().isBlank() ? null : confirmationPageUrl.toString();
+    private URI normalizeConfirmationPageUrl(URI confirmationPageUrl) {
+        return confirmationPageUrl == null || confirmationPageUrl.toString().isBlank() ? null : confirmationPageUrl;
```
dev-marek: if confirmationPageUrl is not null, toString would never be blank, so null check would be enough

dev-marek (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/mapper/ApplicationInvitationsCreateInputMapper.java
jaroslawlakomy-grav: fixed

PR #17582, opened 2026-06-09
Apim 14339 frontend invitation creation enable notifications

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/APIM-14339

## Description

Adds frontend support for sending application invitation notification data from the Portal Next invitation dialog.

Changes:
- Enables the Notify users when invitations are created option.
- Sends notify in the create invitations request.
- Sends confirmation_page_url so backend email templates can build the sign-up confirmation link.
- Adds an Enter hint icon to the email input to clarify how addresses are added.

Zrzut ekranu 2026-06-9 o 10 20 03


aqibmohammadkhan (approved):
LGTM

PR #17596, opened 2026-06-09
feat(e2e): add API Product + Group E2E tests

araffass:
## Issue

https://gravitee.atlassian.net/browse/APIM-14320

## Description

API Product + Group - E2E Test Scenarios

### DIFF
--- a/gravitee-apim-e2e/api-test/src/apis/groups/mapi-v2/api-product-group-access-control.spec.ts
+++ b/gravitee-apim-e2e/api-test/src/apis/groups/mapi-v2/api-product-group-access-control.spec.ts
@@ -1,0 +1,140 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


gravitee-apim-e2e/api-test/src/apis/groups/mapi-v2/api-product-group-role-locking.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+      apiProduct = await createApiProduct(`e2e-role-lock-d1-${Date.now()}`);
+      await updateApiProduct(apiProduct.id, {
+        name: apiProduct.name,
+        description: apiProduct.description,
+        version: apiProduct.version,
+        groups: [group.id],
+      });
+    });
+
+    test('should reject role override for locked group member', async () => {
```
vikrantgravitee: Please verify it

The test name says "should reject", but addResponse.status is never asserted, and the only expect is guarded by if (member). If the member isn't found, zero assertions run and the test passes. This test passes whether role-locking works or is completely broken — it provides no real coverage. The D1 description (lockApiProductRole=true prevents role override) is not actually verified. This should assert a concrete outcome (e.g. expect([400, 409]).toContain(addResponse.status) and/or that the effective role is unchanged unconditionally).

araffass: Fxed

vikrantgravitee (approved):

PR #17612, opened 2026-06-09
feat(portal): add application invitation acceptance flow

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-14131

## Description

Adds the screen an invited user lands on from the invitation email (route `user/invitation/confirm/:token`).

The user enters their own first/last name (the invitation only carries an email) and a password, then submits to `/_finalize`. The create-invitation dialog now points its `confirmation_page_url` to this screen.

## Additional context

Capture d’écran 2026-06-09 à 17 37 52
Capture d’écran 2026-06-09 à 17 39 06


jaroslawlakomy-grav (changes requested):
Looks really good, there is one more check which needs to be done before presenting invitation confirmation form. Frontend should call `POST /applications/{applicationId}/invitations/_search`  to check if there is any pending invitation for email given in token. This is backend under `_finalize` method will do the same and throw InvitationCancelledException. So if we now that submitting a form will fail then instead of showing a form we may present something similar to "Invalid token" saying "This invitation is not valid anymore"  

EDIT: will be considered as improvement in the separate task

jaroslawlakomy-grav (approved):
Validation before presenting the form needs to be a separate feature  (as it possibly requires separate endpoint for validation available for unauthorized user)

PR #17614, opened 2026-06-09
fix: retain primary-owner own group on v4 API groups sanitization

carlos-andres-osorio:
## Issue

[APIM-14376](https://gravitee.atlassian.net/browse/APIM-14376)

## Why

When a v4 PROXY API is PATCHed with a `groups` list that contains both the primary owner's own group and a foreign group, the groups sanitization logic drops all groups instead of retaining the primary owner's group. The bug is in the `else` branch of the USER primary-owner path: it filtered out every group whose `apiPrimaryOwner` field is non-empty, which inadvertently discards the primary owner's own group. This causes silent data loss — all group memberships are cleared after a PATCH, potentially breaking visibility and subscription rules.

## What changed

- `GroupValidationServiceImpl`: In the USER primary-owner branch, the filter condition now also passes through any group whose `apiPrimaryOwner` matches the API's primary owner member ID, so the primary owner's own group is retained while unrelated primary-owner groups are still discarded.
- `GroupValidationServiceImplTest`: Added a test covering the case where the PATCH payload includes both the primary owner's own group and a foreign group, asserting the primary owner's group is kept and the foreign group is dropped.

## User-facing impact

APIs PATCHed with a groups list that includes the primary owner's own group will now correctly retain that group in the response and on subsequent GETs. Previously those groups were silently cleared.

[APIM-14376]: https://gravitee.atlassian.net/browse/APIM-14376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/GroupValidationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/GroupValidationServiceImpl.java
@@ -112,7 +112,10 @@ public class GroupValidationServiceImpl extends TransactionalService implements 
                     group -> StringUtils.isEmpty(group.getApiPrimaryOwner()) || group.getId().equals(primaryOwner.getMemberId())
                 );
             } else {
-                groupEntityStream = groupEntityStream.filter(group -> StringUtils.isEmpty(group.getApiPrimaryOwner()));
+                groupEntityStream = groupEntityStream.filter(
+                    group ->
+                        StringUtils.isEmpty(group.getApiPrimaryOwner()) || group.getApiPrimaryOwner().equals(primaryOwner.getMemberId())
+                );
             }
         } else {
             groupEntityStream = groupEntityStream.filter(group -> StringUtils.isEmpty(group.getApiPrimaryOwner()));
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/GroupValidationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/GroupValidationServiceImpl.java
@@ -114,7 +114,8 @@ public class GroupValidationServiceImpl extends TransactionalService implements 
             } else {
                 groupEntityStream = groupEntityStream.filter(
                     group ->
-                        StringUtils.isEmpty(group.getApiPrimaryOwner()) || group.getApiPrimaryOwner().equals(primaryOwner.getMemberId())
+                        StringUtils.isEmpty(group.getApiPrimaryOwner()) ||
+                        StringUtils.equals(group.getApiPrimaryOwner(), primaryOwner.getMemberId())
                 );
             }
         } else {
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/GroupValidationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/GroupValidationServiceImpl.java
@@ -109,19 +109,21 @@ public class GroupValidationServiceImpl extends TransactionalService implements 
             if (primaryOwner.getMemberType() == MembershipMemberType.GROUP) {
                 // don't remove the primary owner group of this API.
                 groupEntityStream = groupEntityStream.filter(
-                    group -> StringUtils.isEmpty(group.getApiPrimaryOwner()) || group.getId().equals(primaryOwner.getMemberId())
+                    group -> isNotPrimaryOwnerGroup(group) || group.getId().equals(primaryOwner.getMemberId())
                 );
             } else {
                 groupEntityStream = groupEntityStream.filter(
-                    group ->
-                        StringUtils.isEmpty(group.getApiPrimaryOwner()) ||
-                        StringUtils.equals(group.getApiPrimaryOwner(), primaryOwner.getMemberId())
+                    group -> isNotPrimaryOwnerGroup(group) || StringUtils.equals(group.getApiPrimaryOwner(), primaryOwner.getMemberId())
                 );
             }
         } else {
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/GroupValidationServiceImpl.java
```diff
@@ -112,7 +112,10 @@ private Set<String> removePrimaryOwnerGroups(
                     group -> StringUtils.isEmpty(group.getApiPrimaryOwner()) || group.getId().equals(primaryOwner.getMemberId())
                 );
             } else {
-                groupEntityStream = groupEntityStream.filter(group -> StringUtils.isEmpty(group.getApiPrimaryOwner()));
+                groupEntityStream = groupEntityStream.filter(
+                    group ->
+                        StringUtils.isEmpty(group.getApiPrimaryOwner()) || group.getApiPrimaryOwner().equals(primaryOwner.getMemberId())
```
paulatulis: The code is actually null-safe because || short-circuits — if isEmpty(...) is false, the value is non-null. But Sonar's symbolic engine doesn't reliably model org.apache.commons.lang3.StringUtils.isEmpty as a null-guard, so S2259 ("possible NPE, getApiPrimaryOwner() can return null") is a realistic false-positive here. Note the existing GROUP branch above dereferences group.getId() (rarely null), so this is the first place the method dereferences getApiPrimaryOwner() after an isEmpty guard — no existing precedent to suppress it.

Cheap pre-emption that's also cleaner — use commons-lang's null-safe compare:
```
group -> StringUtils.isEmpty(group.getApiPrimaryOwner())
      || StringUtils.equals(group.getApiPrimaryOwner(), primaryOwner.getMemberId())
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/GroupValidationServiceImpl.java
paulatulis, line 1: StringUtils.isEmpty(group.getApiPrimaryOwner()) now appears three times in this one method (GROUP keep-clause, the new USER branch, and the outer else). Recommend extracting a named predicate like isNotPrimaryOwnerGroup(group) / belongsToApiPrimaryOwner(group, poMemberId).

paulatulis (commented):
Two points of clarification: 
1. Divergence from the core domain service (the main thing to flag). There's a newer parallel service, io.gravitee.apim.core.api.domain_service.GroupValidationService, used for Native and Federated v4 APIs. It unconditionally drops all groups with apiPrimaryOwner != null (line 63) — no exception even for the PO's own group. So:  - The same data-loss bug still exists for Native v4 APIs, which this PR does not touch.
  - More philosophically, that service treats a group carrying a PO member as invalid for membership ("…will be discarded because it contains an API Primary Owner member, which is not supported by the API"), while this fix now deliberately retains such a group. Worth confirming with the domain owners that retaining the PO's own PO-group as a regular member group is the intended product semantics and won't cause downstream surprises (visibility/subscription/PO double-counting). The PR rationale and ticket suggest it is, but the two services now disagree.
2. Scope note: If the intent is "PO never silently loses their own group," consider whether a follow-up is needed for the Native/Federated path (core service), or whether HTTP/PROXY is genuinely the only affected surface. Not blocking — just call it out so it's a conscious decision.

carlos-andres-osorio:
> Two points of clarification: 
> 1. Divergence from the core domain service (the main thing to flag). There's a newer parallel service, io.gravitee.apim.core.api.domain_service.GroupValidationService, used for Native and Federated v4 APIs. It unconditionally drops all groups with apiPrimaryOwner != null (line 63) — no exception even for the PO's own group. So:  - The same data-loss bug still exists for Native v4 APIs, which this PR does not touch.
>   - More philosophically, that service treats a group carrying a PO member as invalid for membership ("…will be discarded because it contains an API Primary Owner member, which is not supported by the API"), while this fix now deliberately retains such a group. Worth confirming with the domain owners that retaining the PO's own PO-group as a regular member group is the intended product semantics and won't cause downstream surprises (visibility/subscription/PO double-counting). The PR rationale and ticket suggest it is, but the two services now disagree.
> 2. Scope note: If the intent is "PO never silently loses their own group," consider whether a follow-up is needed for the Native/Federated path (core service), or whether HTTP/PROXY is genuinely the only affected surface. Not blocking — just call it out so it's a conscious decision.

The core `GroupValidationService` (used for Native/Federated v4 APIs) does retain the PO's own group — the filter at line 63 drops all `apiPrimaryOwner`-carrying groups, but lines 66–68 immediately re-add `primaryOwner.id()` when `primaryOwner.type() == GROUP`. The two services follow the same two-step pattern and produce the same outcome for the PO-group retention case. No divergence exists for the scenario this PR addresses.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/GroupValidationServiceImpl.java
carlos-andres-osorio: Adopted the null-safe compare. The USER branch filter now reads `isNotPrimaryOwnerGroup(group) || StringUtils.equals(group.getApiPrimaryOwner(), primaryOwner.getMemberId())`, so `getApiPrimaryOwner()` is only dereferenced through `StringUtils.equals`, which handles null on both sides — S2259 no longer applies.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/GroupValidationServiceImpl.java
carlos-andres-osorio: Extracted the repeated `StringUtils.isEmpty(group.getApiPrimaryOwner())` check into a private `isNotPrimaryOwnerGroup(GroupEntity group)` predicate. The GROUP keep-clause, the USER branch, and the outer else now all call it (the outer else as a method reference `this::isNotPrimaryOwnerGroup`).

paulatulis (approved):
👍

PR #16886, opened 2026-05-14
feat(authz): gateway sync chain for authz policies and entities

podlesrafal:
Gateway-side synchronization of authorization policies and entities, consuming PUBLISH/UNPUBLISH_AUTHZ_POLICY and PUBLISH/UNPUBLISH_AUTHZ_ENTITY events broadcast from the management API (PR1).

Highlights:
* AuthzPolicySynchronizer and AuthzEntitySynchronizer (Order=13/14) with cold-start atomicity guarantees.
* Repository-side appender (AuthzAppender) wires the authz registry to the existing API sync flow without breaking ApiSynchronizer order.
* Deployers: AuthzEntityDeployer, AuthzPolicyDeployer integrated via DeployerFactory.
* EventBusAuthzEnginePort fans authz changes out to the in-process authorization engine.
* Mappers (AuthzEntityMapper, AuthzPolicyMapper) translate event payloads to reactor-deployable models.
* AuthzRegistry caches authz state for repository-backed sync.

Depends on #16885

PR #17615, opened 2026-06-09
chore(authz): remove dead entity-id derivation + unused constants

podlesrafal:
Removes dead/unwired authz entity-id code from the gateway sync.

## Why
`AuthzEntityIdExtractor.extract`/`extractWithAttributes`/`extractToolName` (API→authz-entity derivation minting `api.`/`mcp.` ids gated on `tools/call`) were added in #16886 but **never wired into the sync pipeline** — zero callers, and the `authzEntityIdExtractor()` bean is never injected. The live path maps module-published entity events via `AuthzEntityMapper` (entities carry their own `entityType`). The extractor's only live use is the static `toResourceEngineUid`, which is kept.

## Removed
- `AuthzEntityIdExtractor`: `extract`/`extractWithAttributes`/`extractToolName`, `EntityFragment`, `INSTANCE`
- `SyncConfiguration`: the never-injected `authzEntityIdExtractor()` bean + import
- `AuthzEntityIdExtractorTest` (covered only the removed path)
- `AuthzEntityIdConstants`: `TOOLS_CALL_METHOD` (no consumer left), `isAutoDerived` + `AUTO_DERIVED_PREFIXES` (zero callers)

No behaviour change — only uncalled code removed.

## Related
The mixed-case entity-id fix moved to the authz module itself (gravitee-gamma-module-authz#14), which now owns `FORMAT_REGEX`/`MAX_ENTITY_ID_LENGTH` locally. Removing those two from `AuthzEntityIdConstants` here is a later follow-up, once the remaining definition-model consumer (the EntityRef-parents branch) also migrates.


podlesrafal:
### Follow-up: finish emptying `AuthzEntityIdConstants`

This PR intentionally leaves `FORMAT_REGEX`, `MAX_ENTITY_ID_LENGTH` and the `api.`/`mcp.`/`agent.` prefixes in place, because they still have consumers. They should be removed in follow-ups once those consumers are gone:

- **`FORMAT_REGEX` + `MAX_ENTITY_ID_LENGTH`** — used **only** by the authz module (zero consumers in this repo). gravitee-gamma-module-authz#14 already moved them to a module-local `AuthzEntityIdFormat`. Once the remaining branch there (EntityRef parents, #13) also migrates, **delete both from `AuthzEntityIdConstants`** — removing them before that breaks the un-migrated branch's build against the published `definition-model`.
- **`API_PREFIX` / `MCP_PREFIX` / `AGENT_PREFIX`** — still used by the authz module's prefix-based cascade (`computeCascadePlan`) and `findApiAliases`. Once that cascade is cut (it duplicates the structured `parents` graph from #13), **delete the prefixes too**.

End state: `AuthzEntityIdConstants` keeps only the genuinely cross-component engine-UID vocabulary (`ENGINE_TYPE_PRINCIPAL` / `ENGINE_TYPE_RESOURCE`), or is dropped entirely if the gateway mapper inlines those.

sclarkamlab (approved):

PR #17597, opened 2026-06-09
GMA-595 - feat(gamma-platform): Port Backend AM Connection to Platform

sclarkamlab:
## Issue

https://gravitee.atlassian.net/browse/GMA-595

## Description

This is a migration/refactor, not new functionality. It lifts the Gravitee Access Management connection settings out of the AIM module (`gravitee-gamma-module-aim`) and into the platform module, which is where AM settings now live. The use cases, REST resources, DTOs, the AM SDK client and its Spring config are all ported across from AIM as-is. Behaviour is unchanged, the code just moved home (and this is the platform module's first Java backend).

A couple of things worth calling out:

- I dropped the delete AM connection endpoint as part of the move since we don't need it here (second commit).
- The ported DTO and model records are grouped into `AmDtos` / `AmModels` rather than one file per record, to keep the PR file count down.

The AM UI is intentionally left out of this PR. It'll follow in a separate one.

PR #17617, opened 2026-06-09
feat(gamma-platform): add Access Management settings UI ported from AIM

sclarkamlab:
Second half of GMA-595. The backend PR (#17597) moved the AM connection settings into gravitee-gamma-module-platform, this one brings the UI over from AIM to match.

Adds an Access Management page to the platform module under a new "System & Security" nav group. It lets you point the org at a Gravitee AM instance, test the connection, and pick the environment + domain it targets. The service account token is sent to the backend and encrypted at rest there, so nothing sensitive lives in the UI.

Stacked on the backend branch since the page talks to those endpoints, so it targets GMA-595-migrate-am-settings-to-platform rather than master. Retarget to master once the backend merges.

I ran it end to end locally against a real AM: nav renders, test connection succeeds and discovers the env/domain/gateway, save persists and reloads, token comes back encrypted in mongo, and a bad token fails cleanly with the AM 401 surfaced inline.


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
```diff
@@ -0,0 +1,678 @@
...
+        const run = async () => {
+            try {
+                const entries = await listDomainEntrypoints(entrypointCfg, entrypointCfg.environmentId, entrypointCfg.domainId);
+                if (cancelled) return;
+                setGatewayEntrypoints(entries);
+                const def = entries.find(e => e.defaultEntrypoint) ?? entries[0];
+                if (def && entries.length === 1) {
+                    setGatewayUrl(def.url.replace(/\/$/, ''));
+                } else {
+                    setGatewayUrl(null);
```
podlesrafal: **🔴 Correctness — saved gateway lost on reload for multi-entrypoint domains.**

The connection effect restores the server's `gatewayUrl`, but this effect only keeps a value when `entries.length === 1`; otherwise it does `setGatewayUrl(null)`. So a domain with **>1 entrypoints** loses its persisted gateway on every reload — the `SelectField` shows nothing selected, and if the user hits Save without re-picking, `buildRequest()` sends `gatewayUrl: null`, silently dropping a valid saved value.

Suggestion: in the multi-entrypoint branch, preserve the current `gatewayUrl` when it matches one of `entries[].url` instead of nulling it.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
```diff
@@ -0,0 +1,678 @@
...
+                }
+            } catch {
```
podlesrafal: **🟡 Minor — inconsistent error handling.** This is the only effect with a bare `catch {}` (no `console.error`); every other effect logs. Worth aligning for debuggability.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
```diff
@@ -0,0 +1,678 @@
...
+            if (result.ok) {
+                // Persist immediately so the env/domain queries below run against the tested creds,
+                // not whatever is still saved server-side.
+                try {
+                    const updated = await saveAmConnection(cfg, buildRequest());
+                    setForm(prev => ({ ...prev, accessToken: '', hadAccessToken: updated.hasAccessToken }));
+                } catch (saveErr) {
+                    console.warn('Failed to persist tested AM connection', saveErr);
+                }
+                setScopeRefreshKey(k => k + 1);
```
podlesrafal, line 302: **🔴 Surprising side effect — `Test connection` silently persists credentials.** On a successful test the handler immediately `PUT`s the connection (and writes the token at rest) before the user ever clicks **Save**. If that save fails it's swallowed with `console.warn` and the scope queries proceed anyway against possibly-stale server creds.

The comment explains the *why*, but a button labelled "Test" mutating server state breaks least-surprise. Consider having the `_test` endpoint validate the supplied creds without requiring a prior save, or make the persistence explicit in the UI.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
```diff
@@ -0,0 +1,678 @@
...
+            setForm(prev => ({
+                ...prev,
+                accessToken: '',
+                hadAccessToken: updated.hasAccessToken,
+            }));
+            setConnectionVerified(Boolean(updated.baseUrl && updated.hasAccessToken));
+            if (canSaveSelection) {
+                saveAmConfig(cfg);
+                onSaved(cfg);
+            }
```
podlesrafal, line 335: **🟡 Minor — no feedback on connection-only save.** `submit` always saves the connection but only fires the success toast (`onSaved`) when `canSaveSelection` is also true. If the connection saved but env/domain are incomplete, the user clicks **Save** and sees nothing — no toast, no error. Add a confirmation for this path.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
```diff
@@ -0,0 +1,678 @@
...
+                        Configure how this module reaches Gravitee Access Management. The service-account access token is encrypted at rest.
+                    </CardDescription>
+                </CardHeader>
+                <CardContent className="grid gap-4">
+                    <TextField
+                        label="Organization"
+                        value={cfg.organizationId}
+                        onChange={v => set('organizationId', v)}
+                        placeholder="DEFAULT"
+                    />
```
podlesrafal, line 368: **🟡 Minor — editable Organization field.** `organizationId` is bootstrap-resolved but rendered as a free-text `Input` that feeds straight into `moduleBaseUrl()`. Editing it points requests at a bogus path. Since the console runs under a single org, this should be read-only/disabled (or dropped).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/services/amManagement.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export function getAmConnection(cfg: AmConfig): Promise<AmConnectionView> {
+    return gammaFetchJson<AmConnectionView>(`${moduleBaseUrl(cfg)}/am-config`);
+}
+
+export function saveAmConnection(cfg: AmConfig, payload: AmConnectionRequest): Promise<AmConnectionView> {
+    return gammaFetchJson<AmConnectionView>(`${moduleBaseUrl(cfg)}/am-config`, { method: 'PUT', body: JSON.stringify(payload) });
+}
+
+export function deleteAmConnection(cfg: AmConfig): Promise<void> {
```
podlesrafal: **🟡 Apparently unused.** `deleteAmConnection` isn't referenced anywhere in the PR. Drop it, or wire it up if a follow-up needs it.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/utils/amConfig.ts
```diff
@@ -0,0 +1,48 @@
...
+    } catch {
+        return EMPTY;
+    }
+}
+
+export function saveAmConfig(cfg: AmConfig): void {
+    localStorage.setItem(STORAGE_KEY, JSON.stringify(cfg));
+}
+
+export function isAmConfigured(cfg: AmConfig): boolean {
```
podlesrafal: **🟡 Apparently unused.** `isAmConfigured` isn't referenced anywhere in the PR — same call as `deleteAmConnection`: remove or use.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/AccessManagementPage.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const mockNotifySuccess = jest.fn();
+jest.mock('../shared/notify', () => ({
+    notify: {
+        success: (message: string) => mockNotifySuccess(message),
+    },
+}));
+
+// Stub the heavy connection panel; expose its onSaved so we can assert the page wires a toast.
+jest.mock('../features/access-management/components/AmConfigPanel', () => ({
```
podlesrafal, line 28: **⚠️ Test coverage gap — the main one.** This spec mocks out `AmConfigPanel` entirely, so the 678-line component (effect orchestration, gateway discovery, debounced search, hrid fallback chain) plus `amManagement.ts` and `amConfig.ts` have **zero** coverage. For logic this intricate that's the weakest part of the PR.

Minimum I'd add:
- `amConfig.ts` load/save/`moduleBaseUrl` round-trip (trivial, high value)
- `amManagement.ts` URL construction + `isAmUnavailable` (503 / `am_not_configured`)
- the gateway-reload path flagged in `AmConfigPanel`.

The complexity is also a signal — extracting the connection/scope/entrypoint orchestration into a `useAmConnection` hook would decompose it and make it unit-testable.

podlesrafal (commented):
Code review — porting the AM settings UI into gamma-platform. Solid, careful work (race/staleness handling is genuinely good), but I left a few inline notes: one correctness issue on gateway reload, a surprising side effect on **Test connection**, and the fact that the 678-line `AmConfigPanel` has no test coverage. Details inline.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
sclarkamlab: good catch. in the multi-entrypoint branch i now keep the current gatewayUrl if it still matches one of the entries (realigned to the option value) instead of nulling it, so a saved gateway survives reload.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
sclarkamlab: added a console.error here so it logs like the other effects.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
sclarkamlab: yeah this one's deliberate but i get why it's surprising. the reason it persists is that the scope queries (environments/domains/entrypoints) run against the saved server-side connection, not the supplied creds, so without saving first they'd hit stale creds. the _test endpoint itself does already validate the supplied creds without needing a prior save. making test fully side-effect-free would mean the scope endpoints have to accept transient creds, which is a backend change. leaving it as-is for now, can raise a follow-up if we want it.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
sclarkamlab: added an inline 'Connection saved' alert for the path where the connection saves but env/domain aren't complete yet, so Save isn't a silent no-op.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
sclarkamlab: made it disabled. it's bootstrap-resolved and feeds moduleBaseUrl, so editing it only ever pointed requests at a bogus path.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/services/amManagement.ts
sclarkamlab: removed.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/utils/amConfig.ts
sclarkamlab: removed.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/AccessManagementPage.spec.tsx
sclarkamlab: added specs for amConfig.ts (load/save round-trip + moduleBaseUrl) and amManagement.ts (url construction + isAmUnavailable 503/am_not_configured). left the full AmConfigPanel coverage and the useAmConnection hook extraction as a follow-up since that's a bigger refactor.

podlesrafal (approved):

PR #17619, opened 2026-06-09
fix(rest-api): ApiMapper drops resources field for HTTP and native v4 APIs

carlos-andres-osorio:
## Issue

[APIM-14375](https://gravitee.atlassian.net/browse/APIM-14375)

## Why

`ApiMapper.mapToHttpV4` (`ApiMapper.java:304`) was missing a mapping for the `resources` field. This projection is shared across eight endpoints, so the omission caused every one of them to return `"resources": []` in the response body even when resources were correctly persisted.

The most visible symptom was `PATCH /apis/{apiId}` returning an empty `resources` array. A less obvious but more damaging consequence was in the JSON-Patch engine: the current-state document built by `PatchApiV4Deserializer` goes through the same mapper, so JSON-Patch operations targeting `/resources/N/configuration/...` failed outright because the current state had no resources to match against.

The same omission existed in `mapToNativeV4` (`ApiMapper.java:319`): native (Kafka) v4 APIs inherit `resources` through `AbstractApi`, but the mapper lacked the symmetric mapping, so any resources stored on a native API were also silently dropped from responses.

## What changed

- `ApiMapper` — added `@Mapping(target = "resources", source = "source.apiDefinitionHttpV4.resources")` to `mapToHttpV4`, aligning it with the GET response projection.
- `ApiMapper` — added `@Mapping(target = "resources", source = "source.apiDefinitionNativeV4.resources")` to `mapToNativeV4`, applying the same fix symmetrically for native v4 APIs.

All eight HTTP v4 callers are fixed by the first change:

| Endpoint | Pre-fix symptom |
| --- | --- |
| `PATCH /apis/{apiId}` response | empty `resources` in response body |
| `PATCH /apis/{apiId}` JSON-Patch current state | `/resources/N/configuration/...` patch ops failed |
| `PUT /apis/{apiId}/_import/definition` | response dropped `resources` (persisted correctly) |
| `PUT /apis/{apiId}/_import/swagger` | same |
| `PUT /apis/{apiId}/_import/wsdl` | same |
| `POST /apis/_import/definition` | same |
| `POST /apis/_import/swagger` | same |
| `POST /apis/_import/wsdl` | same |

Native v4 API endpoints routing through `mapToNativeV4` are fixed by the second change.

Tests added for every caller path: `PATCH` response, JSON-Patch current-state round-trip, all six `_import` endpoints, and a native v4 round-trip using `ApiFixtures.aNativeApi()`.

## How to test

**HTTP v4 (Proxy/Message API)**

1. Create a v4 Proxy API with no resources (capture `{apiId}`).
2. `PATCH /management/v2/environments/DEFAULT/apis/{apiId}` with `Content-Type: application/merge-patch+json`:
   ```json
   {"resources":[{"name":"my-cache","type":"cache","enabled":true,"configuration":{"timeToIdleSeconds":0,"timeToLiveSeconds":0,"maxEntriesLocalHeap":1000}}]}
   ```
3. Verify the 200 response body includes the `resources` array with the cache entry (previously returned `[]`).
4. Confirm `GET /management/v2/environments/DEFAULT/apis/{apiId}` returns the same resources.

**Native v4 (Kafka API)**

1. Create a native (Kafka) v4 API with a resource defined.
2. `GET /management/v2/environments/DEFAULT/apis/{apiId}` — verify `resources` is populated in the response (previously returned `null`/`[]`).

## Gotchas / Follow-up

One related gap remains unaddressed (noted for follow-up, not a regression introduced here):

- `mapToV4(ApiWithFlows)` (`ApiMapper.java:332`, used by `POST /apis`) is missing `resources`, `properties`, `services`, `responseTemplates`, and `failover` — but `CreateApiV4` has no `resources` field, so this is a latent gap rather than an active defect.

[APIM-14375]: https://gravitee.atlassian.net/browse/APIM-14375

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapper.java
@@ -301,6 +301,7 @@ public interface ApiMapper {
         expression = "java(source != null && source.getApiDefinitionHttpV4() != null ? responseTemplateMapper.mapResponseTemplateToApiModel(source.getApiDefinitionHttpV4().getResponseTemplates()) : null)"
     )
     @Mapping(target = "properties", source = "source.apiDefinitionHttpV4.properties")
+    @Mapping(target = "resources", source = "source.apiDefinitionHttpV4.resources")
     @Mapping(target = "services", source = "source.apiDefinitionHttpV4.services")
     @Mapping(target = "state", source = "source.lifecycleState")
     ApiV4 mapToHttpV4(io.gravitee.apim.core.api.model.Api source, UriInfo uriInfo, GenericApi.DeploymentStateEnum deploymentState);
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapper.java
@@ -316,6 +316,7 @@ public interface ApiMapper {
     @Mapping(target = "listeners", source = "source.apiDefinitionNativeV4.listeners", qualifiedByName = "fromNativeListeners")
     @Mapping(target = "state", source = "source.lifecycleState")
     @Mapping(target = "analytics", source = "source.apiDefinitionNativeV4.analytics")
+    @Mapping(target = "resources", source = "source.apiDefinitionNativeV4.resources")
     ApiV4 mapToNativeV4(io.gravitee.apim.core.api.model.Api source, UriInfo uriInfo, GenericApi.DeploymentStateEnum deploymentState);
 
     @Mapping(target = "definitionContext", source = "source.originContext")
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource_CreateApiFromSwaggerTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource_CreateApiFromSwaggerTest.java
@@ -58,7 +58,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 
-public class ApisResource_CreateApiFromSwagger extends AbstractResourceTest {
+public class ApisResource_CreateApiFromSwaggerTest extends AbstractResourceTest {
 
     private static final String ENVIRONMENT_ID = "my-env";
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapperTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapperTest.java
@@ -255,6 +255,25 @@ public class ApiMapperTest {
     }
 
     @Test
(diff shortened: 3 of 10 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapper.java
```diff
@@ -301,6 +301,7 @@ default ApiV4 mapToV4(GenericApiEntity genericApiEntity) {
         expression = "java(source != null && source.getApiDefinitionHttpV4() != null ? responseTemplateMapper.mapResponseTemplateToApiModel(source.getApiDefinitionHttpV4().getResponseTemplates()) : null)"
     )
     @Mapping(target = "properties", source = "source.apiDefinitionHttpV4.properties")
+    @Mapping(target = "resources", source = "source.apiDefinitionHttpV4.resources")
```
wbabyte, line 304: 🟠 **Native (Kafka) v4 APIs still drop `resources` — same root cause, worth folding into this PR.**

This line fixes the HTTP path, but the sibling `mapToNativeV4` right below (line 319) has no equivalent. `NativeApi extends AbstractApi`, which declares `protected List<Resource> resources` (`AbstractApi.java:109`), so native APIs *can* carry resources — but `ApiType.NATIVE` routes to `mapToNativeV4` (line 261-262), which drops them across the same 8 endpoints. "Inherits via AbstractApi" only means the field exists on the model; the response mapper still omits it.

Symmetric one-liner on `mapToNativeV4`:
```java
@Mapping(target = "resources", source = "source.apiDefinitionNativeV4.resources")
```
plus a test using the existing `ApiFixtures.aNativeApi()` fixture rather than `aProxyApiV4()`.

_Message APIs need no extra work — everything except `NATIVE` routes through `mapToHttpV4` (PROXY, MESSAGE, and the A2A/LLM/MCP proxy variants), so this added line already covers them._

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource_CreateApiFromSwaggerTest.java
```diff
@@ -113,6 +118,46 @@ public void should_not_import_when_no_definition_permission() {
         assertThat(response.getStatus()).isEqualTo(FORBIDDEN_403);
     }
 
+    @Test
+    @SneakyThrows
+    public void should_return_resources_in_response_on_success() {
```
wbabyte, line 123: 🔴 **This test never runs.** The class is `ApisResource_CreateApiFromSwagger` — missing the `Test` suffix every sibling uses (`ApisResource_CreateApiFromWsdlTest`, `ApisResource_CreateApiWithDefinitionTest`). No reactor pom overrides the Surefire `<includes>`, so the default patterns apply (`Test*.java`, `*Test.java`, `*Tests.java`, `*TestCase.java`) — this class matches none and is never discovered or run.

So this new test compiles but provides zero protection, leaving POST `/apis/_import/swagger` (one of the 8 endpoints this PR claims to cover) effectively untested.

**Fix:** rename the file + class to `ApisResource_CreateApiFromSwaggerTest`, then re-run the module to confirm it actually executes and passes.

wbabyte (commented):
Code review — the one-line `mapToHttpV4` fix is correct and I verified it reaches all 8 listed callers. Two findings inline: one blocking (a new test that never runs), one I'd fold into this PR (the symmetric native gap).

wbabyte (commented):
n

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/ApiMapper.java
carlos-andres-osorio: Done — added `@Mapping(target = "resources", source = "source.apiDefinitionNativeV4.resources")` to `mapToNativeV4` in `ApiMapper.java:319`, symmetric with the `mapToHttpV4` fix. Added a test in `ApiMapperTest` using `ApiFixtures.aNativeApi()` that asserts resources survive the mapping round-trip.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/api/ApisResource_CreateApiFromSwaggerTest.java
carlos-andres-osorio: Fixed — renamed `ApisResource_CreateApiFromSwagger` → `ApisResource_CreateApiFromSwaggerTest` (file and class) so Surefire's default `*Test.java` pattern picks it up.

wbabyte (approved):

PR #17620, opened 2026-06-09
chore(authz): remove dead entity-id derivation + unused constants

podlesrafal:
## What
Remove the gateway-side `AuthzEntityIdExtractor.extract`/`extractWithAttributes` (API→authz-entity derivation that minted `api.`/`mcp.` ids gated on `tools/call`) plus unused members of `AuthzEntityIdConstants`.

## Why
`extract`/`extractWithAttributes`/`extractToolName` were added in #16886 but **never wired into the sync pipeline** — they have zero callers and the `authzEntityIdExtractor()` bean is never injected. The live sync path publishes module-side entity events that are mapped by `AuthzEntityMapper` and already carry their own `entityType`, so the API-definition derivation is dead. The extractor's only live use is the static `toResourceEngineUid`, which is kept.

`isAutoDerived`/`AUTO_DERIVED_PREFIXES` have zero callers anywhere; `TOOLS_CALL_METHOD` had no consumer left once `extract*` is removed.

## Changes
- `AuthzEntityIdExtractor`: drop `extract`, `extractWithAttributes`, `extractToolName`, `EntityFragment`, `INSTANCE`; keep static `toResourceEngineUid`.
- `SyncConfiguration`: drop the unused `authzEntityIdExtractor()` bean + import.
- `AuthzEntityIdExtractorTest`: deleted (covered only the removed `extract` path).
- `AuthzEntityIdConstants`: drop `TOOLS_CALL_METHOD`, `isAutoDerived`, `AUTO_DERIVED_PREFIXES`.

## Not in scope
No behaviour change — only uncalled code is removed. `API_PREFIX`/`MCP_PREFIX`/`AGENT_PREFIX` stay: they are still used by the authz module's prefix-based cascade. Removing them (and the cascade) is a follow-up in `gravitee-gamma-module-authz`, best done on top of the EntityRef-parents work (#13 there), since the cascade is being superseded by structured parents.


podlesrafal:
Folded into #17615 — combined into a single APIM PR.

PR #17621, opened 2026-06-10
fix(v4-api): reject empty and null listeners on PATCH and import-create

carlos-andres-osorio:
## Issue

[APIM-14374](https://gravitee.atlassian.net/browse/APIM-14374)

## Why

`PATCH /apis/{apiId}` with `{"listeners":[]}` returned 200 OK and silently cleared all listeners, leaving the API unreachable on the gateway. The `minItems:1` constraint was enforced on PUT but missed on PATCH because the listener validator only validated non-null, non-empty lists — null was treated as no-change (correct for JSON Merge Patch), but an explicit empty array was let through unchanged. A separate gap existed on the import-create path, where a null listeners array could silently yield a listener-less API.

## What changed

- **`ListenerMissingException`** (new) — a `400 BAD_REQUEST` exception with technical code `listeners.missing` and message "At least one listener is required for an API."
- **`ListenerValidationServiceImpl`** (`validateAndSanitizeHttpV4`, `validateAndSanitizeNativeV4`) — split the single `if (listeners != null && !listeners.isEmpty())` guard into two: throw `ListenerMissingException` on an explicit empty list; let null pass through unchanged (PATCH/JSON Merge Patch no-change semantics). Added comments documenting the null contract on both paths.
- **`ApiValidationServiceImpl.validateAndSanitizeImportApiForCreation`** — added an explicit null guard before delegating to the listener validator, throwing `ListenerMissingException` when the import entity carries no listeners.

## User-facing impact

- `PATCH /apis/{apiId}` with `{"listeners":[]}` now returns 400 with technical code `listeners.missing` instead of 200. ⚠️ BREAKING for any caller that relied on the previous (incorrect) empty-array behavior.
- Import-create with a null `listeners` field now returns 400 instead of silently creating a listener-less API.

## Gotchas / Follow-up

The null-pass-through contract in `validateAndSanitizeHttpV4` is intentional: null means "no change" under JSON Merge Patch semantics. Only an explicit empty array is rejected. This is documented at the guard site.

[APIM-14374]: https://gravitee.atlassian.net/browse/APIM-14374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ApiValidationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ApiValidationServiceImpl.java
@@ -45,6 +45,7 @@ import io.gravitee.rest.api.service.impl.TransactionalService;
 import io.gravitee.rest.api.service.v4.ApiServicePluginService;
 import io.gravitee.rest.api.service.v4.PlanSearchService;
 import io.gravitee.rest.api.service.v4.exception.ApiTypeException;
+import io.gravitee.rest.api.service.v4.exception.ListenerMissingException;
 import io.gravitee.rest.api.service.v4.validation.AnalyticsValidationService;
 import io.gravitee.rest.api.service.v4.validation.ApiValidationService;
 import io.gravitee.rest.api.service.v4.validation.EndpointGroupsValidationService;
@@ -240,6 +241,9 @@ public class ApiValidationServiceImpl extends TransactionalService implements Ap
             groupValidationService.validateAndSanitize(executionContext, null, apiEntity.getGroups(), null, primaryOwnerEntity, true)
         );
         // Validate and clean listeners
+        if (apiEntity.getListeners() == null) {
+            throw new ListenerMissingException();
+        }
         apiEntity.setListeners(
             listenerValidationService.validateAndSanitizeHttpV4(
                 executionContext,
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ApiValidationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ApiValidationServiceImpl.java
@@ -241,6 +241,8 @@ public class ApiValidationServiceImpl extends TransactionalService implements Ap
             groupValidationService.validateAndSanitize(executionContext, null, apiEntity.getGroups(), null, primaryOwnerEntity, true)
         );
         // Validate and clean listeners
+        // ApiEntity.getListeners() is a plain Lombok getter (returns null when unset); unlike definition-layer models
+        // whose getters coalesce null → List.of(), an explicit null guard is required here before delegating.
         if (apiEntity.getListeners() == null) {
             throw new ListenerMissingException();
         }
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ListenerValidationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ListenerValidationServiceImpl.java
@@ -96,25 +96,22 @@ public class ListenerValidationServiceImpl extends TransactionalService implemen
         if (CollectionUtils.isInitializedAndEmpty(listeners)) {
             throw new ListenerMissingException();
         }
-        // null means "no change" for PATCH (JSON Merge Patch null-removal); only an explicit empty list is invalid.
-        if (listeners != null) {
-            checkDuplicatedListeners(listeners);
-            listeners.forEach(listener -> {
-                switch (listener.getType()) {
-                    case HTTP:
-                        validateAndSanitizeHttpListener(executionContext, apiId, (HttpListener) listener, endpointGroups);
-                        break;
-                    case SUBSCRIPTION:
-                        validateAndSanitizeSubscriptionListener((SubscriptionListener) listener, endpointGroups);
-                        break;
-                    case TCP:
-                        validateAndSanitizeTcpListener(executionContext, apiId, (TcpListener) listener, endpointGroups);
-                        break;
-                    default:
-                        break;
-                }
-            });
-        }
+        checkDuplicatedListeners(listeners);
+        listeners.forEach(listener -> {
+            switch (listener.getType()) {
+                case HTTP:
+                    validateAndSanitizeHttpListener(executionContext, apiId, (HttpListener) listener, endpointGroups);
+                    break;
+                case SUBSCRIPTION:
+                    validateAndSanitizeSubscriptionListener((SubscriptionListener) listener, endpointGroups);
+                    break;
+                case TCP:
+                    validateAndSanitizeTcpListener(executionContext, apiId, (TcpListener) listener, endpointGroups);
+                    break;
+                default:
+                    break;
+            }
+        });
         return listeners;
     }
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ListenerValidationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ListenerValidationServiceImpl.java
@@ -18,6 +18,7 @@ package io.gravitee.rest.api.service.v4.impl.validation;
 import io.gravitee.apim.core.api.domain_service.VerifyApiHostsDomainService;
 import io.gravitee.apim.core.api.domain_service.VerifyApiPathDomainService;
 import io.gravitee.apim.core.api.exception.InvalidPathsException;
+import io.gravitee.apim.core.utils.CollectionUtils;
 import io.gravitee.apim.infra.adapter.PathAdapter;
 import io.gravitee.definition.model.v4.ApiType;
 import io.gravitee.definition.model.v4.ConnectorFeature;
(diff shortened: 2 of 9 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/impl/validation/ListenerValidationServiceImplTest.java
```diff
@@ -129,15 +130,23 @@ void setUp() throws Exception {
     class ValidateAndSanitizeHttpV4 {
 
         @Test
-        void should_ignore_empty_list() {
+        void should_reject_empty_list() {
```
wbabyte, line 133: **[Critical] The fix is unit-tested only; no test drives it end-to-end, and pre-existing PATCH tests assert the *opposite* behavior.**

The real PATCH chain is `PatchApiUseCase → updateApiDomainService.validateV4 → ApiServiceImpl.validate → validateAndSanitizeUpdateApi → this method`. But every PATCH-layer test mocks `validateV4`, so the real validator is never hit:
- `PatchApiUseCaseTest.java:141` (mock) + `:188` (identity stub)
- `ApiResource_PatchApiListenersTest.java:106-111`

Consequently these existing tests still pass while asserting exactly the behavior this PR removes:
- `PatchApiUseCaseTest.java:2399` `merge_patch_with_empty_array_clears_listeners` (asserts `[]` clears)
- `PatchApiUseCaseTest.java:2390` `merge_patch_with_null_listeners_erases_them`
- `ApiResource_PatchApiListenersTest.java:232` `merge_patch_listeners_null_clears_listeners` (asserts 200)
- `ApiResource_PatchApiListenersTest.java:243` `json_patch_remove_listeners_clears_listeners` (asserts 200)

Please (1) add an acceptance test that drives `{"listeners":[]}` (merge-patch) and `replace /listeners []` (json-patch) through the **real** validator and asserts 400 + technicalCode `listeners.missing`, and (2) re-baseline the four tests above to the new behavior. `ListenerMissingException` currently appears only in the two unit-test files this PR touched.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ListenerValidationServiceImpl.java
```diff
@@ -92,7 +93,11 @@ public List<Listener> validateAndSanitizeHttpV4(
         final List<Listener> listeners,
         final List<EndpointGroup> endpointGroups
     ) {
-        if (listeners != null && !listeners.isEmpty()) {
+        if (CollectionUtils.isInitializedAndEmpty(listeners)) {
+            throw new ListenerMissingException();
+        }
+        // null means "no change" for PATCH (JSON Merge Patch null-removal); only an explicit empty list is invalid.
```
wbabyte: **[Important] The documented "null = no change" contract may not survive the PATCH pipeline — please verify.**

On the update path, listeners are mapped via `ApiAdapter.toUpdateApiEntity(Api, io.gravitee.definition.model.v4.Api)` (`ApiAdapter.java:107`, no explicit `listeners` `@Mapping`), so MapStruct sources from `apiDefinitionV4.getListeners()` — and `io.gravitee.definition.model.v4.Api.getListeners()` (`Api.java:167`) coalesces `null → List.of()`. A JSON-Merge-Patch `{"listeners":null}` (removal → empty) therefore reaches this method as `[]` → 400, not the no-change pass-through this comment promises.

Normal PATCHes that omit `listeners` stay safe (the original non-empty list is retained), so it's not a broad regression — but this `null` branch (and the new `should_allow_null_list` test) describes a state the real PATCH pipeline can't actually produce. Suggest verifying the adapter mapping and either rewording this comment or adding an `ApiAdapter` test asserting null-definition → `UpdateApiEntity.getListeners()` is `null` (not `List.of()`).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ListenerValidationServiceImpl.java
```diff
@@ -120,7 +125,12 @@ public List<NativeListener> validateAndSanitizeNativeV4(
         List<NativeListener> listeners,
         List<NativeEndpointGroup> endpointGroups
     ) {
-        if (listeners != null && !listeners.isEmpty()) {
+        // NativeApi.getListeners() overrides the Lombok getter to return List.of() instead of null,
```
wbabyte, line 125: **[Important] Blast radius is wider than "PATCH + import-create."**

`NativeApi.getListeners()` (`NativeApi.java:126`) coalesces `null → List.of()`, so a native API with absent listeners reaches `validateAndSanitizeNativeV4` as an *empty* list and now throws — on **native create *and* update** via `ValidateApiDomainServiceLegacyWrapper.validateAndSanitizeNativeV4Definition` (`:206`), neither of which is PATCH or import-create. This is arguably the correct invariant, but it's an undocumented breaking surface; please call it out in the PR description / changelog. The comment here is accurate that `null` is unreachable, but the *practical* new behavior is empty-list rejection — worth stating explicitly. (No real-validator test covers this either: `ValidateApiDomainServiceLegacyWrapperTest.java:100` mocks the validator and `:159/:196` still assert empty listeners are accepted on create.)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ApiValidationServiceImpl.java
```diff
@@ -240,6 +241,9 @@ public void validateAndSanitizeImportApiForCreation(
             groupValidationService.validateAndSanitize(executionContext, null, apiEntity.getGroups(), null, primaryOwnerEntity, true)
         );
         // Validate and clean listeners
+        if (apiEntity.getListeners() == null) {
```
wbabyte, line 246: **[Important] Same widening on the HTTP side, and POST-create is untested.**

This explicit null guard is correct (REST `ApiEntity.getListeners()` is a plain Lombok getter, unlike the definition models). But the shared `validateAndSanitizeHttpV4` is also reached by `validateAndSanitizeNewApi` (`:132`, POST-create) and `validateAndSanitizeUpdateApi` (`:189`, PUT/PATCH), so empty listeners now return 400 on those paths too — and there is no test for the POST-create case.

Suggest: (a) widen the PR description to cover POST-create + PUT; (b) add a `validateAndSanitizeNewApi` empty-listeners rejection test mirroring `shouldThrowExceptionWhenImportApiForCreationHasNullListeners`; and (c) add a one-line note on *why* this path needs an explicit `null` guard while the others rely on `isInitializedAndEmpty`, so it isn't "simplified" away later.

wbabyte (commented):
Automated deep-dive review (focused on the Critical/Important findings). The core change and unit tests are sound, but (1) the headline behavior isn't verified end-to-end, (2) several existing PATCH tests now assert the opposite behavior, and (3) the breaking surface is wider than the title. Inline comments target the exact code.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/v4/impl/validation/ListenerValidationServiceImplTest.java
carlos-andres-osorio: Done both. At the REST layer, `ApiResource_PatchApiListenersTest` now wires the **real** `ListenerValidationServiceImpl` into the update path via a shared `wireRealListenerValidator()` helper, so `{"listeners":[]}` (merge-patch), `{"listeners":null}` (merge-patch), and `remove /listeners` (json-patch) all drive the live validator end-to-end and assert 400 + technicalCode `listeners.missing`. The null and remove cases reject (rather than no-op) because the adapter coalesces `null → List.of()` before the validator runs — the real REST chain can't produce a no-change pass-through. Re-baselined the use-case-layer tests in `PatchApiUseCaseTest`: `merge_patch_with_empty_array_is_rejected` wires the real validator and asserts `ListenerMissingException` / `listeners.missing`; the merge-null and json-remove use-case tests keep covering patch resolution (null = no-op preserving the existing list, remove = clears to empty), with the end-to-end 400 for those inputs pinned by the REST tests above. `ListenerMissingException` is now exercised through live validator code instead of only the two unit-test files this PR touched.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ListenerValidationServiceImpl.java
carlos-andres-osorio: Verified the adapter mapping — confirmed `io.gravitee.definition.model.v4.Api.getListeners()` coalesces `null → List.of()`, so the validator's null branch is unreachable from the PATCH pipeline. Removed the dead null guard and its misleading "null = no change for PATCH" comment from `validateAndSanitizeHttpV4`, and dropped the `should_allow_null_list` test that only exercised that dead branch. Added `ApiAdapterTest.should_coalesce_null_listeners_to_empty_list_in_UpdateApiEntity` to pin the adapter-layer coalescing contract.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ListenerValidationServiceImpl.java
carlos-andres-osorio: Added two tests wiring the real `ListenerValidationServiceImpl` into `ValidateApiDomainServiceLegacyWrapperTest` to cover empty-listeners rejection on both the native create and native update paths (both were previously mocked, leaving the new rejection path untested).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/v4/impl/validation/ApiValidationServiceImpl.java
carlos-andres-osorio: Added a `validateAndSanitizeNewApi` empty-listeners rejection test in `ApiValidationServiceImplTest`, mirroring `shouldThrowExceptionWhenImportApiForCreationHasNullListeners`, to cover the previously-untested POST-create path through the shared `validateAndSanitizeHttpV4`. Added a why-comment at `ApiValidationServiceImpl.java:244` explaining the explicit null guard is required because `ApiEntity.getListeners()` is a plain Lombok getter (returns null when unset), unlike the definition-layer models whose getters coalesce null → `List.of()`. The PR description covers the POST-create + PUT widening.

wbabyte (approved):

PR #17623, opened 2026-06-10
fix(gateway): render requestMethod MDC entry as String to avoid ClassCastException

aditya-goyal01:
## Issue

https://gravitee.atlassian.net/browse/APIM-14386

## Description

Both `DefaultApiReactor` (v4) and `SyncApiReactor` (v2) registered the `requestMethod` MDC log entry with a lambda `context -> context.getAttribute(ATTR_REQUEST_METHOD)`. Because `LogEntryFactory.refreshable` requires `Function<T, String>`, javac infers `T=String` and emits a `checkcast String` inside the lambda. The `rest-to-soap` policy sets `ATTR_REQUEST_METHOD` to `io.gravitee.common.http.HttpMethod` (an enum), so when MDC is resolved the cast throws `ClassCastException`, masking the real error with a 500.

Fix: use `Objects.toString(context.getAttribute(ATTR_REQUEST_METHOD), null)` in both reactors — handles Gravitee `HttpMethod`, Vert.x `HttpMethod`, `String`, and `null` safely.

## Additional context

Regression tests use `@ParameterizedTest` with four attribute shapes (Gravitee `HttpMethod`, Vert.x `HttpMethod`, `String`, `null`) via `ReflectionTestUtils.getField` on `DEFAULT_EXECUTION_CONTEXT_LOG_ENTRIES`, driven against a real `DefaultExecutionContext`. Red→green proven; full 896-test module suite passes.

Backport labels: `apply-on-4-12-x`, `apply-on-master`


ytvnr (approved):
Good catch
I would say that this kind of test is a bit overkill, but it ensure non-reg so ok for me !

PR #17634, opened 2026-06-10
fix: reject Equals + wildcard flow paths and reload Policy Studio on …

ankita-gupta21:
…save failure

## Issue

https://gravitee.atlassian.net/browse/APIM-14322

## Description
On v4 API flows, the **Equals** path operator does not work with wildcard paths (e.g. `/**`). Previously this could be saved from Policy Studio, then caused a **500** at runtime with a `PatternSyntaxException` in the gateway.

This PR adds validation on save and fixes the console so the UI recovers after a failed save.

## Backend changes

**What changed in code:**
- Added validation in `FlowValidationDomainService` to reject HTTP selectors when `pathOperator = EQUALS` and the path contains `*`
- Added `InvalidFlowException.equalsWithWildcardPath()` for a clear validation error message
- Applied the same check in `ValidateApiCRDDomainService` for CRD imports
- Added unit tests in `FlowValidationDomainServiceTest`

**Why:**
Invalid flow config was only failing at runtime in the gateway. We now block it when the API/plan is saved or imported, so bad config never gets deployed.

## Frontend changes

**What changed in code:**
- Updated `api-v4-policy-studio-design.component.ts` to reload API and plan flows from the server after a failed save
- Extracted `loadPolicyStudioData()` so Policy Studio can refresh its data on error
- Removed internal error swallowing in save methods so save failures are handled correctly
- Added unit tests in `api-v4-policy-studio-design.component.spec.ts` for reload-on-error behavior

**Why:**
Policy Studio kept the invalid flow in local UI state after a failed save. The error toast appeared, but the bad flow stayed in the list until a manual page refresh. Now the UI resets automatically from server data when validation fails.

### Before & After change
https://github.com/user-attachments/assets/67f381a8-3c10-46a6-871e-86ee6e6f253a

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/policy-studio-v4/design/api-v4-policy-studio-design.component.ts
+++ b/gravitee-apim-console-webui/src/management/api/policy-studio-v4/design/api-v4-policy-studio-design.component.ts
@@ -15,8 +15,8 @@
  */
 
 import { ChangeDetectorRef, Component, OnDestroy, OnInit } from '@angular/core';
-import { combineLatest, EMPTY, forkJoin, Observable, of, Subject } from 'rxjs';
-import { catchError, map, switchMap, takeUntil, tap } from 'rxjs/operators';
+import { combineLatest, EMPTY, forkJoin, Observable, Subject } from 'rxjs';
+import { catchError, finalize, map, switchMap, takeUntil, tap } from 'rxjs/operators';
 import {
   ConnectorInfo,
   Flow as PSFlow,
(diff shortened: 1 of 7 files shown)


gravitee-apim-console-webui/src/management/api/policy-studio-v4/design/api-v4-policy-studio-design.component.ts
```diff
@@ -196,12 +119,23 @@ export class ApiV4PolicyStudioDesignComponent implements OnInit, OnDestroy {
...
+
     forkJoin(updates$)
       .pipe(
         tap(() => this.snackBarService.success('Policy Studio configuration saved')),
+        tap(() => this.router.navigateByUrl(this.location.path(), { skipLocationChange: true })),
+        catchError((err) => {
+          this.snackBarService.error(err.error?.message ?? err.message);
+          this.isLoading = true;
+          this.changeDetectorRef.detectChanges();
+          return this.loadPolicyStudioData(this.activatedRoute.snapshot.params);
```
aditya-goyal01: If `loadPolicyStudioData` itself fails, the error falls through to `.subscribe()` which has no error handler. At that point `isLoading` is true and never gets reset— the component is permanently locked on a spinner.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/domain_service/ValidateApiCRDDomainService.java
```diff
@@ -193,6 +194,14 @@ private void checkApiFlowsPathOperator(List<? extends AbstractFlow> flows) {
                 if (hasNullPathOperator) {
                     throw InvalidFlowException.missingPathOperator(flow.getName());
                 }
+                flow
+                    .getSelectors()
+                    .stream()
+                    .filter(FlowValidationDomainService.HTTP_SELECTOR_WITH_EQUALS_WILDCARD_PATH)
+                    .findFirst()
+                    .ifPresent(selector -> {
```
aditya-goyal01, line 202: If the predicate is ever broadened, this becomes a `ClassCastException` at runtime with no compile-time warning.

```
.ifPresent(selector -> {
    if (selector instanceof HttpSelector httpSelector) {
        throw InvalidFlowException.equalsWithWildcardPath(flow.getName(), httpSelector.getPath());
    }
});
```

aditya-goyal01:
Add appropriate mergify labels.

gravitee-apim-console-webui/src/management/api/policy-studio-v4/design/api-v4-policy-studio-design.component.ts
ankita-gupta21: fixed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/domain_service/ValidateApiCRDDomainService.java
ankita-gupta21: fixed

aditya-goyal01 (approved):

PR #17659, opened 2026-06-10
feat(rest-api): grant default roles to users created from an invitation

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-14396

## Description

Users created when accepting an application invitation got no default `ORGANIZATION`/`ENVIRONMENT` role, so the portal returned **403** (e.g. on `GET /applications`). They could log in but couldn't use it.

Extract the default-role assignment from `UserServiceImpl.addDefaultMembership` into a shared `AssignUserDefaultRolesDomainService`, now called by both:

- classic registration
- the invitation flow (`CreateUserDomainServiceImpl`)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -816,41 +819,7 @@ private UserEntity create(
     }
 
     private void addDefaultMembership(ExecutionContext executionContext, User user) {
```
dev-marek, line 821: For you to be aware: it will wire this logic also to the management API

dev-marek (approved):

PR #17664, opened 2026-06-10
fix(rest-api): serialization failure in api export returs empty 

aditya-goyal01:
## Issue

https://gravitee.atlassian.net/browse/APIM-13453

## Description

- `ApiExportServiceImpl.exportAsJson()` caught any serialization exception, logged it, and returned an empty string. The export endpoint then shipped HTTP 200 with a 0-byte body, and `PromotionServiceImpl.create()` persisted that empty string as the promotion's `apiDefinition` (the promotion looked successful but the definition was gone).

- It now rethrows as `TechnicalManagementException` (HTTP 500 with the error message), mirroring the sibling `exportAsCustomResourceDefinition()`. Promotion creation fails fast before persisting, so no promotion record is stored with an empty definition.

## Additional context

- **Root cause:** an API referencing group IDs that no longer exist in the `groups` collection makes `ApiSerializer.serialize()` throw `Groups [...] cannot be found.`, which was silently swallowed.

**Reproduction:**
1. Create an API assigned to a group.
2. Delete the group (the API still references its ID).
3. Export the API (`GET .../export`) — before: HTTP 200 + empty body; after: HTTP 500 with the error message.

**Affected:** 
- All maintained branches (4.8.x → master).

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApiExportServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApiExportServiceImpl.java
@@ -102,8 +102,8 @@ public class ApiExportServiceImpl extends AbstractService implements ApiExportSe
             return objectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(apiEntity);
         } catch (final Exception e) {
             log.error("An error occurs while trying to JSON serialize the API {}", apiEntity, e);
+            throw new TechnicalManagementException(e);
         }
-        return "";
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


mukul-tyagi08 (approved):
LGTM
Only one subtle point to verify is that TechnicalManagementException preserves the original exception’s message (or at least the group details) so that the test’s assertion (checking for "my-group" in the exception message) holds true.

If the TechnicalManagementException does not forward the cause’s message, consider modifying its constructor call to include the cause’s message for clarity.

aditya-goyal01:
> LGTM Only one subtle point to verify is that TechnicalManagementException preserves the original exception’s message (or at least the group details) so that the test’s assertion (checking for "my-group" in the exception message) holds true.
> 
> If the TechnicalManagementException does not forward the cause’s message, consider modifying its constructor call to include the cause’s message for clarity.

Done. 👍

PR #17667, opened 2026-06-10
Feat/portal listing persistence

konrad-krol-incubly:
## Issue

https://gravitee.atlassian.net/browse/EXT-134

## Description

Part of the Next-Gen Portal Automation API initiative, Slice 3: Portal Listing. This PR lays down only the data and domain layer.

  Scope is deliberately narrow: CRUD persistence only, no portal navigation tree side effect. A `PortalListing` row carries the spec (hrid + APIs) but does not yet materialize any `PortalNavigationApi` leaves in the portal nav tree — that depends on Slice 2 (portal navigation sync) and will be addressed in a separate follow-up PR once Slice 2 merges.
  
  ### Introduced

  - New `PortalListing` entity with persistence across JDBC, Mongo, and TCK, plus a Liquibase migration; unique constraint on `(environmentId, portalId, hrid)`
  keeps deterministic re-PUT idempotent.
  - Core domain layer (model, crud service, exception, infra adapter handling JSON serialization at the repository boundary).
  - Extension of `HRIDToUUID` with a portal-scoped sub-resource builder so the listing id is derived deterministically from `(portalHrid, listingHrid)`.
  - Create / Get / Delete use cases — intentionally not Spring-wired here; the REST PR turns them on.
  - Cascade cleanup wired into the existing environment-deletion command handler so the org-deletion arch test stays green.

PR #17668, opened 2026-06-10
feat(portal-listing): expose Portal Listing through the automation RE…

konrad-krol-incubly:
## Issue

## Description

Part of the Next-Gen Portal Automation API initiative, Slice 3: Portal Listing. This PR is the REST layer on top of the persistence + domain introduced in https://github.com/gravitee-io/gravitee-api-management/pull/17667.

Scope stays narrow on purpose: PUT/GET/DELETE round-trip the spec; no portal navigation tree side effect yet. Materialization of `PortalNavigationApi` leaves at the requested `location` paths is intentionally out of scope — it depends on Slice 2 (portal navigation sync) and will land in a separate follow-up PR once that slice merges.

  ### Introduced

  - JAX-RS resources for PUT (with `dryRun`) / GET / DELETE under `/portals/{portalHrid}/listings[/{hrid}]`, hung off the existing portal route via a sub-resource
  locator.
  - MapStruct mapper bridging domain ↔ generated wire models.
  - Resource tests covering dryRun, happy-path CRUD, 400 on a missing hrid, and 404 when the parent portal or the listing itself is absent.
  - Activates the use cases as Spring beans (`@UseCase` annotations restored from PR1) and registers test mocks for them.
  - Permissions reuse `ENVIRONMENT_PORTAL` with acls per HTTP verb, aligned with the existing Slice 1 portal resources.


dev-marek (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/PortalListingResource.java
```diff
@@ -0,0 +1,70 @@
...
+    @Inject
+    private DeletePortalListingUseCase deletePortalListingUseCase;
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Permissions({ @Permission(value = RolePermission.ENVIRONMENT_PORTAL, acls = RolePermissionAction.READ) })
+    public Response getPortalListingByHRID(@PathParam("hrid") String portalHrid, @PathParam("listingHrid") String listingHrid) {
+        var auditInfo = getAuditInfo();
+        try {
+            var output = getPortalListingUseCase.execute(new GetPortalListingUseCase.Input(auditInfo, portalHrid, listingHrid));
```
benoitgravitee: Resource layer computes the HRID. You should move the logic here, UseCase are agnostic of HRID.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/PortalListingResource.java
```diff
@@ -0,0 +1,70 @@
...
+            throw new HRIDNotFoundException(listingHrid);
+        }
+    }
+
+    @DELETE
+    @Permissions({ @Permission(value = RolePermission.ENVIRONMENT_PORTAL, acls = RolePermissionAction.DELETE) })
+    public Response deletePortalListingByHrid(@PathParam("hrid") String portalHrid, @PathParam("listingHrid") String listingHrid) {
+        var auditInfo = getAuditInfo();
+        try {
+            deletePortalListingUseCase.execute(new DeletePortalListingUseCase.Input(auditInfo, portalHrid, listingHrid));
```
benoitgravitee: Same here

benoitgravitee (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/PortalListingsResource.java
```diff
@@ -0,0 +1,99 @@
...
+    public Response createOrUpdate(
+        @PathParam("hrid") String portalHrid,
+        @Valid @NotNull PortalListingSpec spec,
+        @QueryParam("dryRun") boolean dryRun
+    ) {
+        var auditInfo = getAuditInfo();
+        var apis = PortalListingMapper.INSTANCE.toDomainApis(spec.getApis());
+
+        if (dryRun) {
+            var listingId = io.gravitee.apim.core.portal_listing.model.PortalListingId.of(
```
benoitgravitee: For GKO we use the dryrun in the admission controller.
It might not be very explicit, I apologies if I forgot to mention it, it is meant to run the validation logic.
The UseCase (or Domain service) runs it too.
It is acceptable to return a 400 on the first error found (create a custom "<any thing>DomainValidationService", although you should return errors (see io.gravitee.apim.core.validation.Validator) but the API is a bit tricky. You should validate anything that could go wrong before calling the service layer (for instance the portal you want add portal listing into does not exists etc...
In GKO if the error occurs after the admission controller is called and it's a 400 error, only way to understand why it is not working it to go into the GKO's logs.

benoitgravitee (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/PortalListingResource.java
konrad-krol-incubly: Done — resource derives `listingId` via `HRIDToUUID.portalListing()…` from `@PathParam`. Use case receives `PortalListingId` directly.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/PortalListingResource.java
konrad-krol-incubly: Same fix — id derivation in the resource, use case is HRID-agnostic.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/PortalListingsResource.java
konrad-krol-incubly: Refactored to the `Validator` pattern: `ValidatePortalListingDomainService` + `ValidatePortalListingUseCase`. `dryRun=true` runs validation and returns 200 with errors in `PortalListingState.errors`; `dryRun=false` runs the validator first and raises `ValidationDomainException` → 400 on severe errors.

benoitgravitee (approved):

PR #17684, opened 2026-06-11
feat(observability): unified filter catalog & endpoints in Gamma (GMA-422)

SamHardi:
## Summary
Delivers Lot 1 of the unified observability surface (epic [GMA-420](https://gravitee.atlassian.net/browse/GMA-420)): a self-describing filter discovery API in `gravitee-gamma-rest-api`, reading from the GMA-421 shared filter engine, so `@gravitee/gamma-lib-observability` can build the filter UI dynamically and apply the same vocabulary to logs and analytics. Closes [GMA-422](https://gravitee.atlassian.net/browse/GMA-422).

## Changes
- **Unified vocabulary** — expanded the GMA-421 host catalog (`StaticFilters` + `ExtensibleFilters`) with the authoritative reconciled vocabulary: canonical names (`HTTP_GATEWAY_RESPONSE_TIME`, `MCP_PROXY_METHOD`, `URI` for path; `HTTP_PATH`/`mapped-path` dropped), labelled ENUM values, and the two discovery axes (`signals` × `apiTypes`). `StaticFilters` now supports fixed labelled ENUM values.
- **`GET /observability/filters/definition`** — two optional, multi-valued axes (`signal`, `apiType`); intersection per axis; each spec carries both axes.
- **`GET /observability/filters/{filterName}/values`** — type-driven: ENUM served locally from the catalog; KEYWORD (incl. id-based `API`/`APPLICATION`/`PLAN`/`API_PRODUCT`) delegated to the platform analytics services; `NUMBER`/`STRING` → 400.
- **`POST /observability/filters/resolve`** — bulk id → label hydration for the id-based filters.
- **Hexagonal wiring** — Gamma `ObservabilityFilterDataPort` + infra adapter delegating to the platform `GetFilterValuesUseCase` / `ResolveFilterLabelsUseCase` (reuses the proven ES + management-DB machinery; the Gamma `core` does not depend on apim `core`).
- **Authz** — coarse env-level guard (`ENVIRONMENT_DASHBOARD`/`ENVIRONMENT_API` read) → 403; mgmt-v2 filter endpoints untouched.
- **OpenAPI** — documentation spec `openapi-observability-filters.yaml` (the module has no server-side codegen; DTOs are hand-written records).

## Backend / API changes
|---|---|---|---|---|
| `/observability/filters/definition` | `GET` | New | `?signal=LOGS&apiType=LLM` | `{ "data": [ { "name": "API_TYPE", "type": "enum", "operators": ["EQ","IN"], "enumValues": [{"value":"NATIVE","label":"Kafka (native)"}], "signals": ["ANALYTICS","LOGS"], "apiTypes": ["..."] } ] }` |
| `/observability/filters/{filterName}/values` | `GET` | New | `/API_TYPE/values?query=kafka&page=1&perPage=10` | `{ "data": [ {"value":"NATIVE","label":"Kafka (native)"} ], "pagination": { "totalCount": 1, "page": 1, "pageCount": 1 } }` |
| `/observability/filters/resolve` | `POST` | New | `{ "entries": [ { "filterName": "API", "ids": ["api-1"] } ] }` | `{ "entries": [ { "filterName": "API", "labels": { "api-1": "Petstore" } } ] }` |

## Reviewer notes
- Most behaviour-bearing review surface: `StaticFilters` (the full vocabulary + axes — verify names/signals/apiTypes against the ticket tables) and `ObservabilityFilterDataPortAdapter` (the apim→Gamma value/label mapping: id-based filters put the entity id in `value` and the display name in `label`; ES cursor pages report no exact total, so `totalCount` falls back to a lower bound).
- **Known limitation (tracked in [GMA-630](https://gravitee.atlassian.net/browse/GMA-630)):** KEYWORD value listing only covers filter names the platform analytics layer maps. Gamma-only names (`ERROR_KEY`, `REQUEST_ID`, `TRANSACTION_ID`, `MESSAGE_CONNECTOR_TYPE`, `EDGE_*`, `MCP_PROXY_TOOL/RESOURCE/PROMPT`) currently return 400; serving them requires extending the shared analytics ES field mapping, deferred to keep this PR from mutating the shared closed enum.

## Test plan
- [ ] `mvn -pl gravitee-gamma/gravitee-gamma-rest-api test` — unit + REST (Jersey) + ArchUnit all green.
- [ ] `definition`: no axes → full catalog; `signal=LOGS` excludes analytics-only filters; `apiType=LLM` excludes non-LLM filters; both axes intersect independently; unknown `signal`/`apiType` → 400.
- [ ] `values`: ENUM returns `{value,label}` (substring filter on value/label, 1-based pagination); NUMBER/STRING → 400; unknown filter → 404; KEYWORD (`API`) delegates to the data port.
- [ ] `resolve`: returns labels grouped by filter; non-resolvable filters → empty map (no error).
- [ ] Authz: missing dashboard/API read permission → 403 on all three endpoints.

[GMA-420]: https://gravitee.atlassian.net/browse/GMA-420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GMA-422]: https://gravitee.atlassian.net/browse/GMA-422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GMA-630]: https://gravitee.atlassian.net/browse/GMA-630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/use_case/GetObservabilityFilterValuesUseCase.java
```diff
@@ -0,0 +1,125 @@
...
+        List<FilterValue> matching = enumValues
+            .stream()
+            .filter(v -> queryLower == null || matches(v, queryLower))
+            .map(v -> new FilterValue(v.value(), v.label()))
+            .toList();
+
+        int page = (input.page() != null && input.page() > 0) ? input.page() : 1;
+        int perPage = (input.perPage() != null && input.perPage() > 0) ? Math.min(input.perPage(), MAX_PER_PAGE) : DEFAULT_PER_PAGE;
+        int fromIndex = Math.min((page - 1) * perPage, matching.size());
+        int toIndex = Math.min(fromIndex + perPage, matching.size());
```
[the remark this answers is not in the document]
SamHardi: Fixed in 97a55b74ed. Cast to long before multiply to prevent integer overflow on extreme page values.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityFilterDataPortAdapter.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private static AuditInfo currentAuditInfo() {
+        ExecutionContext executionContext = GraviteeContext.getExecutionContext();
+        UserDetails user = (UserDetails) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
+        return AuditInfo.builder()
+            .organizationId(executionContext.getOrganizationId())
+            .environmentId(executionContext.getEnvironmentId())
+            .actor(AuditActor.builder().userId(user.getUsername()).userSource(user.getSource()).userSourceId(user.getSourceId()).build())
+            .build();
+    }
```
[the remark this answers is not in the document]
SamHardi: Fixed in 97a55b74ed. Added instanceof check before casting principal to UserDetails, with fallback to toString().

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/observability/filters/ObservabilityFiltersResource.java
```diff
@@ -0,0 +1,176 @@
...
+        if (raw == null || raw.isEmpty()) {
+            return Set.of();
+        }
+        Set<E> result = EnumSet.noneOf(type);
+        for (String token : raw) {
+            if (token == null || token.isBlank()) {
+                continue;
+            }
+            try {
+                result.add(valueOf.apply(token.trim().toUpperCase()));
```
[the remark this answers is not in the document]
SamHardi: Fixed in 97a55b74ed. Added Locale.ROOT to toUpperCase() call.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/observability/filters/dto/FilterSpecDto.java
```diff
@@ -0,0 +1,71 @@
...
+) {
+    public record EnumValueDto(String value, String label) {}
+
+    public record RangeDto(@JsonInclude(JsonInclude.Include.NON_NULL) Number min, @JsonInclude(JsonInclude.Include.NON_NULL) Number max) {}
+
+    public static FilterSpecDto from(FilterSpec spec) {
+        return new FilterSpecDto(
+            spec.name(),
+            spec.label(),
+            spec.type().name().toLowerCase(),
```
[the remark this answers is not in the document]
SamHardi: Fixed in 97a55b74ed. Added Locale.ROOT to toLowerCase() call.

jourdiw (approved):
What do you think about generating entities from your open api spec instead of written DTO classes? We use that for the `mapi-v2` and it helps keeping docs up to date 👍 

Perhaps we can also start looking into how to serve the gamma open api specs. 🤔

SamHardi:
> What do you think about generating entities from your open api spec instead of written DTO classes? We use that for the `mapi-v2` and it helps keeping docs up to date 👍
> 
> Perhaps we can also start looking into how to serve the gamma open api specs. 🤔

In that case, we could update the [AGENTS.md](https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-gamma/gravitee-gamma-rest-api/AGENTS.md) (section 9) and migrate all the Gamma endpoints. 
The current Tracing endpoints follow the same manual-DTO pattern.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/observability/filters/ObservabilityFiltersResource.java
```diff
@@ -0,0 +1,180 @@
...
+        @QueryParam("to") Long to,
+        @QueryParam("page") @DefaultValue("1") int page,
+        @QueryParam("perPage") @DefaultValue("10") int perPage
+    ) {
+        checkReadObservabilityPermission();
+        var output = getFilterValuesUseCase.execute(
+            new GetObservabilityFilterValuesUseCase.Input(filterName, query, from, to, page, perPage)
+        );
+        List<FilterValueDto> data = output.page().data().stream().map(FilterValueDto::from).toList();
+        return PaginatedResponseDto.of(data, output.page().totalElements(), page, perPage);
```
remibaptistegio: The response is using the query param values for page and perPage but in the use case, you are applying some logic to restrict these values. So if I send for instance perPage=500, the use case will in fact use 100 which could break the pageCount returned by the PaginatedResponseDto no ?

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/use_case/GetObservabilityFilterValuesUseCase.java
```diff
@@ -0,0 +1,132 @@
...
+
+        String queryLower = input.query() == null || input.query().isBlank() ? null : input.query().toLowerCase();
+        List<FilterValue> matching = enumValues
+            .stream()
+            .filter(v -> queryLower == null || matches(v, queryLower))
+            .map(v -> new FilterValue(v.value(), v.label()))
+            .toList();
+
+        int page = (input.page() != null && input.page() > 0) ? input.page() : 1;
+        int perPage = (input.perPage() != null && input.perPage() > 0) ? Math.min(input.perPage(), MAX_PER_PAGE) : DEFAULT_PER_PAGE;
```
remibaptistegio: You should reuse your helper method (resolvePage and resolvePerPage) to avoid drift here no ?

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/use_case/GetObservabilityFilterValuesUseCase.java
```diff
@@ -0,0 +1,132 @@
...
+ * (the filter is addressed via the URL path). Pagination is <b>1-based</b>; {@code perPage} is
+ * clamped to {@value #MAX_PER_PAGE}.
+ *
+ * @author GraviteeSource Team
+ */
+@UseCase
+@AllArgsConstructor
+public class GetObservabilityFilterValuesUseCase {
+
+    private static final int DEFAULT_PER_PAGE = 10;
```
remibaptistegio, line 56: The default per page value is also defined in the resource, maybe we should centralized this value. Wdyt ?

remibaptistegio (approved):
Approved with some comments that I'll let you review.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/observability/filters/ObservabilityFiltersResource.java
SamHardi: Fixed in 67a00e56f0. The use case now resolves and clamps page/perPage and exposes them on its Output; the resource builds the paginated envelope from those resolved values instead of the raw query params, so pageCount stays consistent with the clamped perPage (covered by a new resource test).

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/use_case/GetObservabilityFilterValuesUseCase.java
SamHardi: Fixed in 67a00e56f0. execute() now resolves page/perPage once via resolvePage/resolvePerPage and passes them to both the ENUM and KEYWORD branches, so the two paths can no longer drift.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/use_case/GetObservabilityFilterValuesUseCase.java
SamHardi: Fixed in 67a00e56f0. Removed the @DefaultValue on the resource; the use case is now the single source of truth for the default/clamped perPage (DEFAULT_PER_PAGE), so there is no duplicated constant to drift.

PR #17725, opened 2026-06-11
feat(am-settings): persist selected environment id on the am connection

sclarkamlab:
APIM half of GMA-595. The per-org AM connection stored baseUrl, token, default domain and gateway url but no environment id. AM addresses domains as organizations/{org}/environments/{env}/domains/{domain}/..., so a consumer that only knows the domain can't build the path. The platform settings UI already lets you pick an environment, it just wasn't being persisted.

This threads a nullable `environmentId` through the whole stack, mirroring `defaultDomainId` everywhere:

- repository model + mongo/jdbc mappings, plus a new liquibase changeset (v4_12_0/15_add_environment_id_to_am_connections) altering am_connections. Kept under v4_12_0 since 4.12 is still in development and that's where the am_connections table was created
- rest-api AmConnectionEntity + AmConnectionServiceImpl
- the shared gamma-plugin AmConnection record (placed before defaultDomainId, included in the redacting toString)
- platform module Save/Get/View/Test use cases, DTOs and AmConnectionResource
- platform UI: AmConfigPanel now sends cfg.environmentId on save and hydrates it from the loaded connection

Field-addition migration only, behaviour is otherwise unchanged. Nullable everywhere since a connection can exist before an env/domain is chosen.

This is the prerequisite the AIM side of GMA-595 is blocked on. That follow-up surfaces environmentId on AIM's GET /am-config and removes the moved endpoints, but only once a published apim build carries the field.

Tested: repository conformance (jdbc + mongo testcontainers), AmConnectionServiceImplTest (incl. a new env round-trip), gamma plugin + platform use case tests, and platform UI jest (348 specs). prettier + nx build clean.


podlesrafal (commented):
## Review

Clean, symmetric field-addition — `environmentId` is threaded through every layer in lockstep with `defaultDomainId`, nullable everywhere, with round-trip tests. The in-repo change is solid. My one substantive concern is **downstream**, and it's the same publish-coordination mechanism the description already calls out for AIM — just aimed at the wrong module.

### 🔴 Arity change on the shared positional record breaks `gravitee-gamma-module-authz`

`io.gravitee.apim.plugin.gamma.api.identity.AmConnection` is a **positional** record published as the `gamma-plugin-module-api` artifact. This PR takes it from 5 to 6 components, inserting `environmentId` in the middle.

`gravitee-gamma-module-authz` (separate repo, own CI) imports that exact record and constructs it positionally with 5 args:

- `core/am/use_case/StartAmUserSyncUseCaseTest.java` — `new AmConnection("http://am:8093", "token", "domain-1", "domain-hrid", null)`
- `infra/service_provider/AmUserSyncRunnerImplTest.java`
- `core/am/use_case/SyncAmUsersUseCaseTest.java`

Once authz builds against the published apim carrying this field, **its test compilation fails** (the 5-arg constructor no longer exists). The authz production code survives — it only reads via accessors, which are additive — so this is a test-compile break, but a red build nonetheless.

Worth noting: the description anticipates this for AIM, but **AIM is actually safe** — it has its own copy of the record (`com.graviteesource.gamma.module.aim.core.identity.model.AmConnection`), not the shared one. The module that actually breaks is `module-authz`, which isn't mentioned.

**Suggested fix** — keep source compatibility by adding a 5-arg convenience constructor instead of a hard break:

```java
public AmConnection(String baseUrl, String serviceAccountAccessToken,
                    String defaultDomainId, String defaultDomainHrid, String gatewayUrl) {
    this(baseUrl, serviceAccountAccessToken, null, defaultDomainId, defaultDomainHrid, gatewayUrl);
}
```

authz keeps compiling untouched, new code uses the 6-arg form. Alternatively, land a coordinated authz PR updating the call-sites — at minimum, list `module-authz` as a downstream consumer to bump.

### 🟡 `module-authz` still ignores `environmentId` (same root cause)

`AmSdkDirectoryClient` hardcodes `AM_DEFAULT_ENVIRONMENT = "DEFAULT"` and passes it into every AM API call (`listUsers`, `listIdentityProviders`, `findRoles`, …). That's the very problem this PR fixes for the connection test: AM addresses resources per-environment. After this PR `environmentId` is persisted, but authz user-sync will still hit `DEFAULT`. Follow-up, not a blocker — but persisting the field buys nothing for sync until authz reads it. Worth tracking alongside GMA-595.

### 🟢 Nits

- Description references changeset `v4_12_0/13_add_environment_id_to_am_connections`, but the actual file is `15_…`. The code (15, after `14_add_portal_listings`) is correct — just a stale description.
- The `15_` filename prefix may collide on merge with another in-flight 4.12 changeset (the repo already has duplicate `02_*`). The `databaseChangeLog id` is unique so Liquibase won't break; only `master.yml` ordering might conflict. Acceptable and already acknowledged.
- Column type ✓ `environment_id nvarchar(64)` matches `default_domain_id` in `10_add_am_connections_table.yml`.

### 👍 Good
Full symmetry with `defaultDomainId`, nullable end-to-end, `toString()` still redacts the token while logging `environmentId` in clear (non-sensitive), env round-trip tests in `AmConnectionServiceImplTest` / `ApimAmConnectionRepositoryTest`, and every in-PR call-site (View/Output/Response/Request) updated.

One thing to settle before merge: the 🔴 (compat constructor, or a coordinated authz PR + mention in the description). The rest is clean follow-up.

gravitee-gamma/gravitee-gamma-plugin/gravitee-gamma-plugin-module/gravitee-gamma-plugin-module-api/src/main/java/io/gravitee/apim/plugin/gamma/api/identity/AmConnection.java
```diff
@@ -19,12 +19,14 @@
...
- * <p>Nullable fields ({@code serviceAccountAccessToken}, {@code defaultDomainId},
- * {@code defaultDomainHrid}, {@code gatewayUrl}) may be {@code null} when not yet configured.
+ * <p>Nullable fields ({@code serviceAccountAccessToken}, {@code environmentId},
+ * {@code defaultDomainId}, {@code defaultDomainHrid}, {@code gatewayUrl}) may be {@code null} when
+ * not yet configured. {@code environmentId} is the AM environment the {@code defaultDomainId} lives in.
  */
 public record AmConnection(
     String baseUrl,
     String serviceAccountAccessToken,
+    String environmentId,
```
podlesrafal, line 29: 🔴 **Inserting a component into this shared positional record is a breaking change for `gravitee-gamma-module-authz`.**

This record is published as the `gamma-plugin-module-api` artifact. `module-authz` constructs it positionally with the old 5-arg signature (`StartAmUserSyncUseCaseTest`, `AmUserSyncRunnerImplTest`, `SyncAmUsersUseCaseTest` — `new AmConnection(baseUrl, token, domainId, domainHrid, gatewayUrl)`). Once authz builds against a published apim carrying this field, its **test compilation breaks**. AIM is unaffected — it has its own copy of the record (`com.graviteesource.gamma.module.aim.core.identity.model.AmConnection`), which is the module the description worries about.

Keep source compatibility by adding a 5-arg convenience constructor:

```java
public AmConnection(String baseUrl, String serviceAccountAccessToken,
                    String defaultDomainId, String defaultDomainHrid, String gatewayUrl) {
    this(baseUrl, serviceAccountAccessToken, null, defaultDomainId, defaultDomainHrid, gatewayUrl);
}
```

…or land a coordinated authz PR updating the call-sites. At minimum, list `module-authz` as a downstream consumer to bump.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/15_add_environment_id_to_am_connections.yml
```diff
@@ -0,0 +1,11 @@
+databaseChangeLog:
+  - changeSet:
+      id: 4.12.0_15_add_environment_id_to_am_connections
```
podlesrafal, line 3: 🟢 Nit: the PR description references this as changeset `13_add_environment_id_to_am_connections`, but the file is `15_…`. The file is correct (it follows `14_add_portal_listings`) — just fix the description.

Also note the `15_` filename prefix may collide on merge with another in-flight 4.12 changeset (the repo already has duplicate `02_*`). The `databaseChangeLog id` here is unique so Liquibase won't break; only `master.yml` ordering could conflict.

Column type `nvarchar(64)` ✅ matches `default_domain_id` in `10_add_am_connections_table.yml`.

podlesrafal (commented):
The inline notes pin each finding to its line. One finding can't be inlined — its file isn't part of this PR:

🟡 **`module-authz` still ignores `environmentId` (same root cause).** `AmSdkDirectoryClient` hardcodes `AM_DEFAULT_ENVIRONMENT = "DEFAULT"` and feeds it into every AM API call (`listUsers`, `listIdentityProviders`, `findRoles`, …). After this PR `environmentId` is persisted, but authz user-sync still targets `DEFAULT`. Follow-up, not a blocker — but persisting the field buys nothing for sync until authz reads it. Worth tracking alongside GMA-595.

gravitee-gamma/gravitee-gamma-plugin/gravitee-gamma-plugin-module/gravitee-gamma-plugin-module-api/src/main/java/io/gravitee/apim/plugin/gamma/api/identity/AmConnection.java
sclarkamlab: Will be opening an Authz PR once this is merged

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/15_add_environment_id_to_am_connections.yml
sclarkamlab: Updated PR description

podlesrafal (approved):

PR #17771, opened 2026-06-12
feat(EXT-130): User guidance for Next-Gen portal

MateuszJozwiak-Incubly:
## Issue

https://gravitee.atlassian.net/browse/EXT-130

## Description

Added a banner in the Danger Zone that informs user that the changes there affect only old portal and provides a button to navigate to the Next gen portal settings. Also added the navigation on the side bar. 

## Additional context

Updated the sidebar navigation menu item to have an option to open the link in a new tab. The Next-gen portal settings seem to have no navigation back to the APIM console so opening in the same tab seemed troublesome.

Zrzut ekranu 2026-06-17 o 23 15 52
Zrzut ekranu 2026-06-17 o 23 16 08
Zrzut ekranu 2026-06-17 o 23 16 24
Zrzut ekranu 2026-06-17 o 23 16 32

### DIFF
--- a/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
+++ b/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
@@ -29,6 +29,7 @@ import { EnvironmentSettingsService } from '../../services-ngx/environment-setti
 interface MenuItem {
   icon?: string;
   routerLink?: string;
+  target?: string;
   displayName: string;
   permissions?: string[];
   licenseOptions?: LicenseOptions;
@@ -38,9 +39,24 @@ interface MenuItem {
   category: string;
   items?: MenuItem[];
   routerBasePath?: string;
+  externalLink?: boolean;
 }
 
 export const SIDE_NAV_GROUP_ID = 'side-nav-items';
+
+// prettier-ignore
+export const PORTAL_SETTINGS_PERMISSIONS = [
+  'environment-settings-r',
+  'environment-settings-u',
+  'environment-theme-r',
+  'environment-theme-u',
+  'environment-category-r',
+  'environment-category-u',
+  'environment-documentation-r',
+  'environment-documentation-u',
+  'environment-metadata-r',
+  'environment-metadata-u',
+];
 
 @Component({
   selector: 'gio-side-nav',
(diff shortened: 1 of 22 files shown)


gravitee-apim-console-webui/src/shared/components/classic-portal-only-banner/classic-portal-only-banner.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  actionLabel = input('Next Gen Portal Settings');
+
+  protected readonly isPortalNextEnabled = toSignal(this.environmentSettingsService.isPortalNextEnabled(), { initialValue: false });
+  protected readonly canShowSettingsAction: boolean = this.permissionService.hasAnyMatching([
+    'environment-settings-r',
+    'environment-settings-u',
+  ]);
+  protected readonly settingsRouterLink: string[] | null = (() => {
+    const envHrid = this.activatedRoute.snapshot.params['envHrid'];
+    return envHrid ? ['/', envHrid, '_portal', 'api'] : null;
```
thomasballivet: this currently points to `/_portal/api` but we want it to points to `/_portal/navigation/:navId`

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
```diff
@@ -279,6 +294,15 @@ export class GioSideNavComponent implements OnInit, OnDestroy {
       });
     }
 
+    mainMenuItems.push({
```
thomasballivet: this should be conditioned to `isPortalNextEnabled` as we don't want this link in the sidebar in case portal next is not enabled

gravitee-apim-console-webui/src/shared/components/classic-portal-only-banner/classic-portal-only-banner.component.html
```diff
@@ -0,0 +1,30 @@
...
+-->
+@if (isPortalNextEnabled()) {
+  <gio-banner-info>
+    <span data-testid="banner-title">{{ title() }}</span>
+    @if (body()) {
+      <span gioBannerBody data-testid="banner-body">{{ body() }}</span>
+    }
+    @if (canShowSettingsAction && settingsRouterLink) {
+      <div gioBannerAction>
+        <a mat-stroked-button target="_blank" [routerLink]="settingsRouterLink" data-testid="settings-action">{{ actionLabel() }}</a>
```
thomasballivet: we can add this " `'noopener noreferrer'` " as gemini suggest and you did in gio-side-nav component for example

thomasballivet (changes requested):
Looks overall good, but we have two functional gaps : 
- the next gen portal redirection from documentation should be to navigation and not api, we may include API's `navId`
- Sidebar link should be conditioned to `isPortalNextEnabled`

The side bar link should also have a little icon to show explicitly that it opens a new tab (suggestion image in JIRA ticket).
Also the banner should have space on right and left
Image

MateuszJozwiak-Incubly:
In case of the sidebar button, I see that styling the right icon might require some more general changes  @gravitee/ui-particles-angular to expose CSS custom properties for icon size and right-offset on GioMenuItemComponent, and I was not sure that this should be a part of this PR.
Zrzut ekranu 2026-06-17 o 23 16 32

thomasballivet (approved):

PR #17788, opened 2026-06-12
feat: add portal navigation propagation query params

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/EXT-145

## Summary

Adds backend/API support for choosing whether publishing a Developer Portal navigation container should affect only the selected item or also its descendants.

The propagation choice is available only for publish actions. Unpublish and delete remain mandatory recursive actions for container navigation items.

## Changes

- Added `propagatePublishToChildren` query parameter to `PUT /portal-navigation-items/{navId}`.
- Defaults publish propagation to `false` when the query parameter is omitted.
- Publishes descendants only when `propagatePublishToChildren=true`.
- Keeps unpublish recursive for folders/APIs regardless of the query parameter value.
- Keeps delete contract unchanged: no delete propagation flag, deleting a container always deletes descendants.
- Updated OpenAPI documentation and backend/resource tests accordingly.

❗ It will be a a follow-up PR soon for the frontend updates

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
@@ -1039,6 +1039,18 @@ paths:
           required: true
           schema:
             type: string
+        - name: propagatePublishToChildren
+          in: query
+          description: >
+            Controls whether a publication change on a container navigation item is propagated to descendants.
+            If omitted or false, publishing affects only the selected item.
+            If true, publishing also affects descendants.
+            For unpublishing a FOLDER or API, this parameter is ignored and descendants are always unpublished
+            according to the business rule.
+          required: false
+          schema:
+            type: boolean
+            default: false
       requestBody:
         required: true
         content:
@@ -1067,6 +1079,16 @@ paths:
           required: true
           schema:
             type: string
+        - name: propagateDeleteToChildren
+          in: query
+          description: >
+            Controls whether deleting a container navigation item also deletes its descendants.
+            If omitted or false, only the selected item is deleted.
+            If true, descendants are also deleted.
+          required: false
+          schema:
+            type: boolean
+            default: false
       responses:
         "204":
           description: Portal navigation item deleted
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
@@ -1031,6 +1031,11 @@ paths:
         ParentId is the unique identifier of the parent item.
         * Optional parent ID. If provided, the item is set to that parent.
         * If omitted, the item is moved to the root level.
+
+        Publication propagation applies only when the `published` value changes on a container navigation item (`FOLDER` or `API`).
+        When publishing a container, descendants are published only if `propagatePublishToChildren=true`.
+        If `propagatePublishToChildren` is omitted or false, only the selected item is published.
+        When unpublishing a container, descendants are always unpublished according to the business rule; `propagatePublishToChildren` is ignored and may be omitted.
       operationId: updatePortalNavigationItem
       parameters:
         - name: navId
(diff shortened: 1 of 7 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
```diff
@@ -1067,6 +1079,16 @@ paths:
           required: true
           schema:
             type: string
+        - name: propagateDeleteToChildren
```
aqibmohammadkhan: `propagateDeleteToChildren=false` on a container — what happens to children?

The description says only the selected item is deleted, but not whether children are:

- reparented to the deleted item’s parent,
- promoted to root,
- left orphaned (invalid state), or
- rejected with 4xx.

This needs to be specified before implementation, especially since default: false is the new norm.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
```diff
@@ -1039,6 +1039,18 @@ paths:
           required: true
           schema:
             type: string
+        - name: propagatePublishToChildren
```
aqibmohammadkhan, line 1047: When does `propagatePublishToChildren` apply?

It sits on the general update endpoint, but only matters when:

- published changes, and
- the item is a container (FOLDER / API).

Recommend adding to the `PUT` description block (alongside the existing parentId notes), e.g. “Only evaluated when published changes on a container item.”

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
```diff
@@ -1039,6 +1039,18 @@ paths:
...
+          in: query
```
aqibmohammadkhan, line 1048: Publish vs unpublish asymmetry could be clearer.

Consider splitting behavior in the description:

- Publish (published: true): governed by `propagatePublishToChildren`
- Unpublish (published: false) on FOLDER/API: always cascades; param ignored

Right now that is in the param description.
putting it in the operation description would help generated docs.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
aqibmohammadkhan, line 1: are we expecting something like "this is behavior unchanged until follow-up PR”. 
Without that, API readers will assume `default: false` is already enforced.

jaroslawlakomy-grav (commented):
Decided to also include backend changes in this PR

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
jaroslawlakomy-grav: no propagateDeleteToChildren param, delete is always propagated  (as it was implemented recently)

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
jaroslawlakomy-grav: description updated

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
jaroslawlakomy-grav: description updated

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-environments.yaml
jaroslawlakomy-grav: Decided to also include backend changes in this PR

aqibmohammadkhan (approved):

PR #17807, opened 2026-06-12
fix(rest-api): preserve persisted flow IDs on JSON-Patch sub-path edits

carlos-andres-osorio:
## Why

When a JSON-Patch operation targets a sub-path of `/flows` (e.g. `/flows/1/name`), `PatchApiUseCase` was building the patch base from the in-memory API definition, which holds flows without their database-assigned IDs. After the patch was applied and flows were re-saved, the untouched sibling flows lost their persisted IDs, breaking any downstream reference to those IDs.

## What changed

- `PatchApiUseCase` now fetches the persisted flows from the database before constructing the JSON-Patch base node and merges them into the existing API definition, so all flows carry their stored IDs when the patch is applied.
- Added a regression test in `PatchApiUseCaseTest` that patches one flow's name and asserts the sibling flow's persisted ID is retained.

## User-facing impact

Flows patched via the JSON-Patch API endpoint will no longer lose their persisted IDs when only a sub-path (e.g. the name) of a sibling flow is modified.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api/use_case/PatchApiUseCase.java
@@ -171,6 +171,12 @@ public class PatchApiUseCase {
             throw new ApiInvalidTypeException(input.apiId(), ApiType.PROXY);
         }
 
+        var dbFlows = flowCrudService.getApiV4Flows(input.apiId());
+        var existingHttpV4 = existingApi.getApiDefinitionHttpV4();
+        if (!dbFlows.isEmpty() && existingHttpV4 != null) {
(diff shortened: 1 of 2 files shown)


carlos-andres-osorio:
## QA Smoke Results ✅

**Type:** Smoke · **Environment:** Local management API `http://localhost:8083` (branch `andres.osorio/fix-jsonpatch-flow-id-preservation`)

### Results

| Area | Priority | Status | Notes |
| --- | --- | --- | --- |
| JSON-Patch on flow sub-path preserves sibling flow IDs | High | **PASS** | PATCH `replace` on `/flows/1/name` left `flows[0].id` unchanged in the PATCH response and a follow-up GET. |
| JSON-Patch on flow sub-path preserves patched flow ID | High | **PASS** | Same operation left `flows[1].id` unchanged. |

**Summary:** 2 PASS / 0 FAIL.

The fix — `PatchApiUseCase` fetching persisted flows from the DB before applying a JSON-Patch sub-path operation — is verified end-to-end. Both the patched flow and its siblings retain their persisted IDs across the PATCH response and a subsequent GET.

paulatulis (approved):
Nice

PR #17820, opened 2026-06-14
fix(reporter): respect lifecycle property-name overrides in ES templates

aditya-goyal01:
## Issue

https://gravitee.atlassian.net/browse/APIM-11655

## Description

The `es8x`/`es9x` index templates hard-coded the Elasticsearch ILM rollover-alias key `index.lifecycle.rollover_alias` instead of using the configured `reporters.elasticsearch.lifecycle.rollover_alias_property_name`. OpenSearch rejects an unknown index setting outright rather than ignoring it, so an operator pointing the lifecycle keys at ISM — AWS OpenSearch in Elasticsearch-compatibility mode, where the reporter selects the `es{N}x` tree — got `unknown setting [index.lifecycle.rollover_alias]` and index preparation failed at startup. The policy-name key already honoured its override; only the rollover-alias key was broken.

- Render the configured property names in every `es7x`/`es8x`/`es9x` template.
- Guard the lifecycle block with `?has_content` instead of `??`, so a policy configured as an empty string emits no lifecycle settings rather than an empty policy id. The Helm chart emits all four policy keys unconditionally once `es.lifecycle.enabled` is true, so blanking the ones you don't want is the normal way to enable a policy on a single type.
- Treat a blank or padded property *name* as unset and fall back to the default. Spring's `${property:default}` only falls back when the property is absent, so `rollover_alias_property_name:` in YAML resolves to an empty string — which rendered an empty settings key. The cluster refuses that along with the shard, replica and refresh settings in the same template, and a failed template PUT aborts index preparation, leaving the ingest pipeline unbuilt and every subsequent report dropped.
- Escape the interpolated names and policies with `?json_string`. The JSON output format does not escape interpolations, so a value containing a quote produced a malformed request body and an opaque parse error instead of a diagnosable unknown-setting response.

## Additional context

- Targets master. The same code lives on `4.11.x` and `4.12.x` and carries the same bug; the `apply-on-4-11-x` / `apply-on-4-12-x` labels propagate this commit to both. `4.9.x` and `4.10.x` have no reporter module in the monorepo — they pin the external `gravitee-reporter-elasticsearch` plugin (6.3.7 and 7.4.6), which is affected too and is tracked separately under the same Jira.
- **event-metrics behaviour is unchanged.** Its lifecycle variable is never placed in the FreeMarker data map (`AbstractIndexPreparer.getTemplateData()`), so the block is inert whatever the configuration says. Its templates were parameterised for consistency with their siblings rather than left as the only hard-coded ILM keys, but nothing renders differently. Making that policy configurable is **APIM-14441** — rollover-alias is invalid on a data stream's backing indices, so it is a distinct change.
- The `opensearch/` tree still hard-codes its ISM keys and ignores the property-name overrides. Pre-existing, untouched here.
- Tests: `IndexTemplateTest` asserts the rendered body across all three template trees for every lifecycle-capable type — configured names when overridden, default ILM keys when blank or padded, escaped output when a name or policy is malformed — and needs no container. `IndexPreparerIntegrationTest` drives `prepare()` through the es7x/es8x/es9x preparers against a real OpenSearch container with the ISM overrides set, asserting the PUTs are accepted. Module suite is 93/93 green under `TZ=UTC`.

### DIFF
--- a/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/config/ReporterConfiguration.java
+++ b/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/config/ReporterConfiguration.java
@@ -462,7 +462,7 @@ public class ReporterConfiguration {
     }
 
     public String getIndexLifecyclePolicyPropertyName() {
-        return indexLifecyclePolicyPropertyName;
+        return blankToDefault(indexLifecyclePolicyPropertyName, DEFAULT_INDEX_LIFECYCLE_POLICY_PROPERTY_NAME);
     }
 
     public void setIndexLifecyclePolicyPropertyName(String indexLifecyclePolicyPropertyName) {
@@ -581,10 +581,29 @@ public class ReporterConfiguration {
     }
 
     public String getIndexLifecycleRolloverAliasPropertyName() {
-        return indexLifecycleRolloverAliasPropertyName;
+        return blankToDefault(indexLifecycleRolloverAliasPropertyName, DEFAULT_INDEX_LIFECYCLE_ROLLOVER_ALIAS_PROPERTY_NAME);
     }
 
     public void setIndexLifecycleRolloverAliasPropertyName(String indexLifecycleRolloverAliasPropertyName) {
         this.indexLifecycleRolloverAliasPropertyName = indexLifecycleRolloverAliasPropertyName;
+    }
+
+    /**
+     * Spring's {@code ${property:default}} only falls back when the property is absent: a property that is
+     * present but empty — {@code rollover_alias_property_name:} in YAML, an empty environment variable, or a
+     * blank Helm value — resolves to an empty string and skips the default. An empty settings key makes the
+     * cluster reject the whole index template, taking the shard, replica and refresh settings down with it,
+     * so treat blank as unset. Surrounding whitespace is stripped for the same reason: a quoted YAML value,
+     * a Helm value or an environment variable can carry it, and a padded key is rejected just like an empty
+     * one. Only the property <em>names</em> are normalised — a blank policy already means "no lifecycle for
+     * this type", and a whitespace-only policy is an ordinary typo the cluster reports through its own
+     * lifecycle explain API.
+     */
+    private static String blankToDefault(String value, String defaultValue) {
+        if (value == null) {
+            return defaultValue;
+        }
+        final String stripped = value.strip();
+        return stripped.isEmpty() ? defaultValue : stripped;
     }
 }
--- a/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/resources/freemarker/es7x/mapping/index-template-event-metrics.ftl
+++ b/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/resources/freemarker/es7x/mapping/index-template-event-metrics.ftl
@@ -4,8 +4,8 @@
     "data_stream": {},
     "template": {
         "settings": {
-            <#if indexLifecyclePolicyEventMetrics??>"index.lifecycle.name": "${indexLifecyclePolicyEventMetrics}",</#if>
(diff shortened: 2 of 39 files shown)


callaertanthony (commented):
The fix itself looks correct and complete for the stated scope: grepping master for the literal `index.lifecycle.` / `index_state_management` keys confirms the `es8x`/`es9x` non-event templates were the only ones hard-coding `rollover_alias`, and all 18 are fixed here. Both property-name variables are in the shared model (`AbstractIndexPreparer#getTemplateData`, lines 108-109) with non-null defaults in `ReporterConfiguration` (lines 56-57), so there is no undefined-variable risk in any template tree.

Three asks, all on the new test.

### 1. Ticket references in the code

`IndexPreparerIntegrationTest` carries the Jira ID in two places:

- the class javadoc — *"the AWS-OpenSearch-compat-mode scenario from APIM-11655"*
- the test property — `reporters.elasticsearch.index=gravitee-apim-11655`

Ticket IDs belong in the branch name, commit message and PR description (they are all already there); in code they rot into dead references. This was flagged on a recent PR too. Please keep the *why* in plain language and pick a neutral index name, e.g. `gravitee-ism-override`.

### 2. `@EnabledIf("isDockerAvailable")` silently skips the regression test

The module's existing container-based test (`ElasticsearchReporterTest` via `IntegrationTestConfiguration`) has no such guard, so CI does have Docker. The guard buys nothing there, and its only real effect is that the single test proving this fix can go green without ever having executed. I would drop it and stay consistent with the existing test.

### 3. The test never asserts what the fix changed

It only asserts that `prepare()` completes, which is an indirect proxy for the bug: it passes vacuously if the cluster ever tolerates the unknown setting, and it costs a 60s OpenSearch container plus ~150 lines to check a string substitution.

A direct assertion on the rendered template pins the actual regression, runs in milliseconds, and can cover `es7x`/`es8x`/`es9x`/`opensearch` against every `Type`:

```java
assertThat(preparer.generateIndexTemplate(Type.REQUEST))
    .contains("index.plugins.index_state_management.rollover_alias")
    .doesNotContain("index.lifecycle.rollover_alias");
```

My suggestion would be to make that the regression test, and keep the container test only if you want the end-to-end confirmation on top of it.

---

Two non-blocking follow-ups I noticed while reading, out of scope here:

- `index-template-event-metrics.ftl` still hard-codes `index.lifecycle.*` in all three `es{N}x` trees. It is inert today because `indexLifecyclePolicyEventMetrics` is never added to the template data map, so `reporters.elasticsearch.lifecycle.policies.event_metrics` currently does nothing at all — worth either wiring up (with the override variables) or removing.
- The `opensearch/` templates hard-code the ISM keys and ignore the property-name overrides entirely, i.e. the mirror image of the bug fixed here.

callaertanthony (approved):
All three points from my previous review are addressed — thanks.

1. Ticket references are gone from the code; the test index is now `gravitee-ism-override` and the javadoc describes the scenario instead of the ticket.
2. The Docker guard is removed, so the container test can no longer silently skip.
3. `IndexTemplateTest` now asserts what the templates actually render (es7x/8x/9x × all types, no container), and the integration test is re-scoped to "the cluster accepts the body", with both javadocs cross-linking.

The additions beyond that are good calls:

- `blankToDefault` closes a real gap — `${prop:default}` only falls back when the property is *absent*, and a present-but-empty name would render an empty settings key, which makes the cluster reject the whole template and takes the shard/replica/refresh settings down with it.
- `?json_string` on the names and policy values, which also let the guards drop the now-redundant second conjunct.

I also checked `DatabaseHydrator` in the repository module, since it renders these same `.ftl` files with its own model: it supplies no lifecycle variables, so the guards are false and the new `?json_string` calls are never evaluated. No regression there.

Non-blocking, for later:

- `"${indexLifecycleRolloverAliasPropertyName?json_string}": "${indexName}"` escapes the key but not the value next to it. `indexName` is unescaped in `index_patterns` too, so it's escape-everywhere or accept it as-is — the current mix is the part that reads oddly.
- The event-metrics templates now read the override names, but `indexLifecyclePolicyEventMetrics` is still never added to the model, so `lifecycle.policies.event_metrics` stays inert. Worth either wiring up or removing — not a one-liner, since `rollover_alias` is meaningless on the es8x data-stream template.
- The `opensearch/` templates still hardcode the ISM keys and skip the escaping — the mirror image of this bug.

PR #17824, opened 2026-06-15
feat(observability): add POST /observability/logs/search endpoint in Gamma

SamHardi:
## Summary

Add an environment-wide logs search endpoint (`POST /gamma/.../observability/logs/search`) to the Gamma REST API, implementing [GMA-423](https://gravitee.atlassian.net/browse/GMA-423). This brings paginated, RBAC-scoped, filter-aware log search backed by the `v4-metrics` index, with display-name enrichment for APIs, plans, applications, gateways, and API products.

Screenshot 2026-06-15 at 14 12 17
Screenshot 2026-06-15 at 14 12 34
Screenshot 2026-06-15 at 15 33 14

### Core layer (`core/observability/logs/`)
- **Domain models**: `LogEntry`, `LogEntryWarning`, `LogsPage`, `LogsSearchQuery` — light log row records matching the v4-metrics index fields
- **Port**: `ObservabilityLogsDataPort` — abstracts RBAC-scoped API loading and log search
- **Domain service**: `AccessibleApiScopeDomainService` — computes the effective API scope by intersecting RBAC-accessible APIs, signal-relevant API types, and optional user-supplied API filter. Reusable for analytics/traces
- **Use case**: `SearchObservabilityLogsUseCase` — validates filters against the unified filter registry (LOGS signal), validates time range, applies default entrypoint scoping, delegates search to the data port. Pagination resolved/clamped (`MAX_PER_PAGE=100`), exposed in `Output(data, page, perPage)` following the pattern established in `GetObservabilityFilterValuesUseCase`
- **Exception**: `UnsupportedObservabilityFilterException` — added `signalMismatch()` and `searchTranslationNotSupported()` static factory methods for explicit rejection of untranslatable filters

### Infrastructure layer
- **Adapter**: `ObservabilityLogsDataPortAdapter` — delegates to `ConnectionLogsCrudService`, `UserContextLoader`, enriches results with plan/app/gateway/apiProduct names in batch. Maps `io.gravitee.definition.model.v4.ApiType` → Gamma `ApiType` with exhaustive switch. **Explicitly rejects** untranslatable filter conditions (e.g. `HTTP_STATUS_CODE_GROUP`) with a descriptive 400 error instead of silently dropping them
- **Config**: `GammaLogsConfiguration` — Spring wiring with `@ComponentScan` for `@UseCase` + `@DomainService`, `@Bean` for the adapter. Imported from `StandaloneConfiguration`

### REST layer (`resources/observability/logs/`)
- **Resource**: `LogsResource` — `POST /search?page=&perPage=` with coarse env-level permission guard, delegates to use case
- **DTOs**: `SearchLogsRequestDto`, `FilterConditionDto`, `LogEntryDto` — wire shapes with `@JsonInclude(NON_NULL)`, polymorphic filter value normalization
- **Mounting**: `GammaRootResource` sub-resource at `/observability/logs`, registered in `GammaModuleApplication`

### OpenAPI documentation
- **Spec**: Added `Logs` tag and `POST /logs/search` path to `openapi-observability.yaml` with full request/response schemas (`SearchLogsRequest`, `LogEntry`, `LogEntryWarning`, `LogsSearchResponse`), enriched field descriptions, and error examples (400 for signal mismatch, search translation not supported; 403 for insufficient permissions)
- **Filter operator**: Added `NOT_IN` to `FilterOperator` enum in the spec

### Tests
- **Unit**: `SearchObservabilityLogsUseCaseTest` — 15 tests covering scoping, filter validation, time range, default entrypoint injection, pagination defaults/clamping/null handling
- **REST**: `LogsResourceTest` — 5 tests covering 200 envelope, param forwarding, filter/time forwarding, empty body, 403
- **Adapter**: `ObservabilityLogsDataPortAdapterTest` — verifies explicit rejection of untranslatable filter conditions

## Test plan

- [ ] `mvn -pl gravitee-gamma/gravitee-gamma-rest-api clean test` — 124 tests pass, 0 failures
- [ ] ArchUnit `CoreRulesTest` passes — no forbidden core→infra dependencies
- [ ] Verify `POST /gamma/.../observability/logs/search` returns paginated logs with filter conditions
- [ ] Verify RBAC scoping: non-admin user only sees logs from APIs they have access to
- [ ] Verify `perPage > 100` is clamped to 100
- [ ] Verify default entrypoint scoping matches analytics dashboard totals
- [ ] Verify untranslatable filters (e.g. `HTTP_STATUS_CODE_GROUP`) return 400 with descriptive error

Refs GMA-423

[GMA-423]: https://gravitee.atlassian.net/browse/GMA-423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/domain_service/AccessibleApiScopeDomainService.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/domain_service/AccessibleApiScopeDomainService.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gamma.rest.core.observability.logs.domain_service;
+
+import io.gravitee.apim.core.DomainService;
+import io.gravitee.gamma.rest.core.observability.filter.model.ApiType;
+import io.gravitee.gamma.rest.core.observability.logs.port.service_provider.ObservabilityLogsDataPort.AccessibleApi;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Computes the effective API scope for an observability query by intersecting the caller's
+ * RBAC-accessible APIs with the signal's supported API kinds and any user-supplied API filter.
+ *
+ * <p>This encapsulation avoids duplicating the scoping logic across logs, analytics, and future
+ * observability consumers. The service is stateless — all inputs are passed explicitly so it can
+ * be unit-tested without infrastructure.
+ *
+ * @author GraviteeSource Team
+ */
+@DomainService
+public class AccessibleApiScopeDomainService {
+
+    public record ScopedApis(Set<String> apiIds, Map<String, String> apiNamesById) {}
+
+    /**
+     * @param accessibleApis   All APIs the caller can read (from the data port).
+     * @param wantedApiTypes   API kinds relevant for the current signal (e.g. PROXY, LLM_PROXY,
+     *                         MCP_PROXY for LOGS). Extensible: adding MESSAGE or NATIVE later only
+     *                         requires widening this set.
+     * @param userApiFilter    API IDs explicitly requested by the caller (from an {@code API}
+     *                         filter condition). {@code null} or empty means "all accessible".
+     * @return The intersection of the three constraints. An empty {@code apiIds} means the caller
+     *         has no data to see (→ empty result, not 403).
+     */
+    public ScopedApis computeScope(List<AccessibleApi> accessibleApis, Set<ApiType> wantedApiTypes, Set<String> userApiFilter) {
+        var filtered = accessibleApis.stream().filter(api -> api.type() != null && wantedApiTypes.contains(api.type()));
+
+        if (userApiFilter != null && !userApiFilter.isEmpty()) {
+            filtered = filtered.filter(api -> userApiFilter.contains(api.id()));
+        }
+
+        var apiList = filtered.toList();
+        var ids = apiList.stream().map(AccessibleApi::id).collect(Collectors.toSet());
+        var names = apiList.stream().collect(Collectors.toMap(AccessibleApi::id, AccessibleApi::name, (a, b) -> a));
+
+        return new ScopedApis(ids, names);
+    }
+}
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
@@ -1,0 +1,324 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gamma.rest.infra.adapter;
+
+import io.gravitee.apim.core.api_product.model.ApiProduct;
+import io.gravitee.apim.core.api_product.query_service.ApiProductQueryService;
+import io.gravitee.apim.core.application.crud_service.ApplicationCrudService;
+import io.gravitee.apim.core.gateway.model.BaseInstance;
+import io.gravitee.apim.core.gateway.query_service.InstanceQueryService;
+import io.gravitee.apim.core.log.crud_service.ConnectionLogsCrudService;
+import io.gravitee.apim.core.plan.crud_service.PlanCrudService;
+import io.gravitee.apim.core.plan.model.Plan;
+import io.gravitee.apim.core.user.domain_service.UserContextLoader;
+import io.gravitee.apim.core.user.model.UserContext;
+import io.gravitee.common.http.HttpMethod;
+import io.gravitee.definition.model.DefinitionVersion;
+import io.gravitee.gamma.rest.core.observability.filter.exception.UnsupportedObservabilityFilterException;
+import io.gravitee.gamma.rest.core.observability.filter.model.FilterCondition;
+import io.gravitee.gamma.rest.core.observability.filter.model.FilterOperator;
+import io.gravitee.gamma.rest.core.observability.logs.model.LogEntry;
+import io.gravitee.gamma.rest.core.observability.logs.model.LogEntryWarning;
+import io.gravitee.gamma.rest.core.observability.logs.model.LogsPage;
+import io.gravitee.gamma.rest.core.observability.logs.model.LogsSearchQuery;
+import io.gravitee.gamma.rest.core.observability.logs.port.service_provider.ObservabilityLogsDataPort;
+import io.gravitee.rest.api.model.BaseApplicationEntity;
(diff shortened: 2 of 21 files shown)


gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/domain_service/AccessibleApiScopeDomainService.java
```diff
@@ -0,0 +1,64 @@
...
+     * @param wantedApiTypes   API kinds relevant for the current signal (e.g. PROXY, LLM_PROXY,
+     *                         MCP_PROXY for LOGS). Extensible: adding MESSAGE or NATIVE later only
+     *                         requires widening this set.
+     * @param userApiFilter    API IDs explicitly requested by the caller (from an {@code API}
+     *                         filter condition). {@code null} or empty means "all accessible".
+     * @return The intersection of the three constraints. An empty {@code apiIds} means the caller
+     *         has no data to see (→ empty result, not 403).
+     */
+    public ScopedApis computeScope(List<AccessibleApi> accessibleApis, Set<ApiType> wantedApiTypes, Set<String> userApiFilter) {
+        var filtered = accessibleApis.stream().filter(api -> wantedApiTypes.contains(api.type()));
```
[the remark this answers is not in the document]
SamHardi: Fixed in ed418a0713. Guard against null api.type() before querying wantedApiTypes (Set.of rejects contains(null)).

a-cordier (approved):

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
```diff
@@ -0,0 +1,324 @@
...
+        Long responseTimeFrom = null;
+        Long responseTimeTo = null;
+
+        for (FilterCondition condition : query.conditions()) {
+            var values = condition.values() != null ? condition.values() : List.<String>of();
+            switch (condition.name()) {
+                case "APPLICATION" -> applicationIds.addAll(values);
+                case "PLAN" -> planIds.addAll(values);
+                case "HTTP_METHOD" -> methods.addAll(values.stream().map(this::toHttpMethod).toList());
+                case "HTTP_STATUS" -> statuses.addAll(values.stream().map(Integer::valueOf).toList());
```
remibaptistegio, line 140: In the StaticFilters, HTTP_STATUS is declared with EQ, GTE and LTE but here we only support EQ. Not sure if it's ok since there is something elsewhere preventing to use GTE/LTE.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
```diff
@@ -0,0 +1,324 @@
...
+                case "URI" -> {
+                    if (!values.isEmpty()) uri = values.getFirst();
+                }
+                case "HTTP_GATEWAY_RESPONSE_TIME" -> {
```
remibaptistegio, line 144: Same question here with EQ operator ?

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
```diff
@@ -0,0 +1,324 @@
...
+        builder.uri(uri);
+        builder.responseTimeRanges(responseTimeRanges);
+
+        return builder.build();
+    }
+
+    private LogEntry mapToLogEntry(BaseConnectionLog log, Map<String, String> apiNamesById) {
+        return LogEntry.builder()
+            .apiId(log.getApiId())
+            .apiName(apiNamesById.getOrDefault(log.getApiId(), null))
```
remibaptistegio, line 184: We don't set the apiType, we don't need it ?

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
remibaptistegio: Another question on this one, what happens if the user send 1000 as HTTP_STATUS ? I'm not sure we have a validation on this to respond with a 400 instead of a 500 code.

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapterTest.java
```diff
@@ -0,0 +1,94 @@
...
+            userContextLoader,
+            planCrudService,
+            applicationCrudService,
+            instanceQueryService,
+            apiProductQueryService
+        );
+    }
+
+    @Test
+    void should_reject_filter_not_translatable_to_log_search() {
```
remibaptistegio, line 79: Maybe add a test to check the buildSearchFilters method with a check per operator etc, wdty ?

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
SamHardi: Good catch — addressed in follow-up PR #17841 (commit a65070bf7b), since this PR was already merged.

Root cause: the unified catalog (`StaticFilters.HTTP_STATUS`) advertises `EQ/GTE/LTE`, but the logs backend (`SearchLogsFilters.statuses`) only matches discrete codes, so `GTE`/`LTE` were silently dropped. Two fixes:
- Operator support is now validated centrally against the catalog in a new `ObservabilityFilterValidator` (core, signal-agnostic) — an operator not advertised for a filter returns 400 uniformly for every filter.
- `HTTP_STATUS` `GTE`/`LTE` are now honored by expanding over the catalog-bounded interval (100–599) into the discrete status set. We deliberately expand in the logs adapter rather than mutating the shared `SearchLogsFilters` (epic GMA-420 guardrail: 'extend, never mutate shared adapters').
- On the second point (value `1000`): malformed/out-of-range values now return 400 (`ValidationDomainException`) instead of 500, with bounds read from the catalog so they can't drift.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
SamHardi: Addressed in follow-up PR #17841 (commit a65070bf7b). `HTTP_GATEWAY_RESPONSE_TIME` now supports the `EQ` operator advertised by the catalog: `EQ` maps to an exact range (`from == to`), alongside the existing `GTE`/`LTE`. Unlike `HTTP_STATUS`, the logs backend supports response-time ranges natively (`SearchLogsFilters.responseTimeRanges`), so no expansion is needed here. Any operator outside the catalog is rejected upstream by `ObservabilityFilterValidator`.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
SamHardi: You were right, we do need it — addressed in follow-up PR #17841 (commit a65070bf7b). `apiType` is now propagated onto each log row as the canonical uppercase wire value (`HTTP_PROXY`, `LLM`, `MCP`…), matching the OpenAPI contract and the `API_TYPE` filter values. The type is resolved once while computing the scope (from the accessible APIs) and carried through; the two parallel `apiNamesById`/`apiTypesById` maps were collapsed into a single `ApiReference(name, apiType)`. While adding a test for this I also surfaced and fixed a latent NPE in row enrichment (`Map.of().get(null)` for rows with a null plan/application/gateway id).

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapterTest.java
SamHardi: Done in follow-up PR #17841 (commit a65070bf7b). `ObservabilityLogsDataPortAdapterTest` now covers `buildSearchFilters` per operator: `HTTP_STATUS` (EQ → single status, GTE/LTE → bounded expansion, plus out-of-range / non-numeric → 400), `HTTP_GATEWAY_RESPONSE_TIME` (GTE/LTE/EQ), `HTTP_METHOD` (incl. OTHER fallback), `APPLICATION`, `URI`, and `apiType` propagation. Added `ObservabilityFilterValidatorTest` (name/signal/operator validation) and `AccessibleApiScopeDomainServiceTest` (asserts the uppercase apiType wire value).

PR #17841, opened 2026-06-15
fix(observability): enforce catalog operators and honor HTTP_STATUS ranges in Gamma logs search

SamHardi:
## Summary
Follow-up to the review feedback on #17824 (GMA-423, already merged). It makes the unified observability filter catalog the single source of truth so a filter behaves identically across the logs and (future) analytics signals — same name, same operators — and fixes the contract mismatches found during the post-merge audit.

## Changes
- **Operator validation moved to core.** New `ObservabilityFilterValidator` domain service validates each condition's name, signal *and* operator against the catalog. The use case delegates to it instead of the previous partial check (name + signal only), so unsupported operators are rejected uniformly for every filter.
- **`HTTP_STATUS`**: `EQ` is translated to the discrete status set, and malformed / out-of-range values now return `400` instead of `500` (bounds read from the catalog). **Range operators (`GTE`/`LTE`) are explicitly rejected (400) for now** rather than silently mistranslated — they will be wired through an ES `range` (shared with `HTTP_STATUS_CODE_GROUP`) by **GMA-683**, deliberately *not* via integer expansion, to avoid a divergent status-translation implementation.
- **`HTTP_GATEWAY_RESPONSE_TIME` supports `EQ`** (exact range), matching the catalog. This filter has native range support in the logs backend (`responseTimeRanges`).
- **`apiType` propagated onto log rows** as the canonical uppercase wire value (`HTTP_PROXY`, …), matching the OpenAPI contract and the `API_TYPE` filter values. The two parallel scope maps (`apiNamesById` / `apiTypesById`) are collapsed into a single `ApiReference`.
- **Latent NPE fixed**: enriching a row with a null plan/application/gateway id crashed the whole search (`Map.of().get(null)`); lookups are now null-safe.
- **OpenAPI** updated to document the `400` causes (unsupported/not-yet-translated operator, invalid filter value).

## Backend / API changes
|----------|--------|--------|-----------------|------------------|
| `/gamma/.../observability/logs/search` | `POST` | `HTTP_STATUS` `EQ` translated; invalid value → 400 | `{ "filters": [{ "name": "HTTP_STATUS", "operator": "EQ", "value": "500" }] }` | `200` page of rows with `status == 500` |
| `/gamma/.../observability/logs/search` | `POST` | `HTTP_STATUS` `GTE`/`LTE` → 400 (interim, pending GMA-683) | `{ "filters": [{ "name": "HTTP_STATUS", "operator": "GTE", "value": "500" }] }` | `{ "httpStatus": 400, "technicalCode": "observability.filter.unsupported_operator" }` |
| `/gamma/.../observability/logs/search` | `POST` | unsupported operator → 400 | `{ "filters": [{ "name": "HTTP_STATUS", "operator": "CONTAINS", "value": "200" }] }` | `{ "httpStatus": 400, "technicalCode": "observability.filter.unsupported_operator" }` |
| `/gamma/.../observability/logs/search` | `POST` | `apiType` now populated (uppercase) | `{}` | `{ "data": [{ "apiType": "HTTP_PROXY", ... }] }` |

## Reviewer notes
- **Architectural split**: validation lives in **core** (catalog-driven, shared by signals) while translation stays in the **infra adapter** (logs-backend-specific). `ObservabilityFilterValidator` is signal-agnostic so the upcoming Gamma analytics endpoints (GMA-586) can reuse it with `Signal.ANALYTICS`.
- **HTTP_STATUS range is intentionally deferred to GMA-683**, which adds a shared ES-range util (`StatusCodeGroups`) used by both the analytics `FilterAdapter` and the logs `SearchMetricsQueryAdapter`. Doing integer expansion here would create the exact divergent implementation GMA-683 forbids. Per epic GMA-420: "extend, never mutate shared adapters".

## Test plan
- [ ] `mvn -pl gravitee-gamma/gravitee-gamma-rest-api test` (146 tests green)
- [ ] `HTTP_STATUS` `EQ` translates to the expected status; out-of-range / non-numeric → `400`
- [ ] `HTTP_STATUS` `GTE`/`LTE` → `400` (interim) with `observability.filter.unsupported_operator`
- [ ] An operator absent from the catalog (e.g. `CONTAINS`) → `400`
- [ ] `HTTP_GATEWAY_RESPONSE_TIME` with `EQ` produces an exact range
- [ ] Log rows expose `apiType` as the uppercase enum name
- [ ] A row with null plan/application/gateway no longer fails the search

Refs GMA-423, GMA-683

### DIFF
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
@@ -18,6 +18,7 @@ package io.gravitee.gamma.rest.infra.adapter;
 import io.gravitee.apim.core.api_product.model.ApiProduct;
 import io.gravitee.apim.core.api_product.query_service.ApiProductQueryService;
 import io.gravitee.apim.core.application.crud_service.ApplicationCrudService;
+import io.gravitee.apim.core.exception.ValidationDomainException;
 import io.gravitee.apim.core.gateway.model.BaseInstance;
 import io.gravitee.apim.core.gateway.query_service.InstanceQueryService;
 import io.gravitee.apim.core.log.crud_service.ConnectionLogsCrudService;
@@ -30,6 +31,9 @@ import io.gravitee.definition.model.DefinitionVersion;
 import io.gravitee.gamma.rest.core.observability.filter.exception.UnsupportedObservabilityFilterException;
 import io.gravitee.gamma.rest.core.observability.filter.model.FilterCondition;
 import io.gravitee.gamma.rest.core.observability.filter.model.FilterOperator;
+import io.gravitee.gamma.rest.core.observability.filter.model.FilterSpec;
+import io.gravitee.gamma.rest.core.observability.filter.model.StaticFilters;
+import io.gravitee.gamma.rest.core.observability.logs.model.ApiReference;
 import io.gravitee.gamma.rest.core.observability.logs.model.LogEntry;
 import io.gravitee.gamma.rest.core.observability.logs.model.LogEntryWarning;
 import io.gravitee.gamma.rest.core.observability.logs.model.LogsPage;
@@ -65,6 +69,15 @@ import org.springframework.security.core.context.SecurityContextHolder;
  */
 @RequiredArgsConstructor
 public class ObservabilityLogsDataPortAdapter implements ObservabilityLogsDataPort {
+
+    /**
+     * Inclusive HTTP status bounds, sourced from the unified catalog ({@link StaticFilters#HTTP_STATUS})
+     * so the adapter and the advertised filter range never drift apart. Used both to validate incoming
+     * status values and to bound {@code GTE}/{@code LTE} range expansion.
+     */
+    private static final FilterSpec.Range HTTP_STATUS_RANGE = StaticFilters.HTTP_STATUS.toSpec().range();
+    private static final int HTTP_STATUS_MIN = HTTP_STATUS_RANGE.min().intValue();
+    private static final int HTTP_STATUS_MAX = HTTP_STATUS_RANGE.max().intValue();
 
     private final ConnectionLogsCrudService connectionLogsCrudService;
     private final UserContextLoader userContextLoader;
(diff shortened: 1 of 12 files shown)


gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
```diff
@@ -137,15 +150,22 @@ private SearchLogsFilters buildSearchFilters(LogsSearchQuery query) {
...
-                        if (condition.operator() == FilterOperator.GTE) responseTimeFrom = val;
-                        else if (condition.operator() == FilterOperator.LTE) responseTimeTo = val;
+                        switch (condition.operator()) {
+                            case GTE -> responseTimeFrom = val;
+                            case LTE -> responseTimeTo = val;
+                            case EQ -> {
+                                responseTimeFrom = val;
+                                responseTimeTo = val;
+                            }
+                            default -> throw unexpectedOperator("HTTP_GATEWAY_RESPONSE_TIME", condition.operator());
```
a-cordier, line 167: should we extend this behaviour to e.g. `HTTP_ENDPOINT_RESPONSE_TIME` ?

remibaptistegio (approved):

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
SamHardi: Good idea — and it is feasible, the data is already on the index. Quick rundown:

- `HTTP_ENDPOINT_RESPONSE_TIME` (and `HTTP_GATEWAY_LATENCY`) are currently **ANALYTICS-only** in the unified catalog (`StaticFilters`, `signals = {ANALYTICS}`), so they don't reach this adapter — `ObservabilityFilterValidator` rejects them for the LOGS signal (400, `signal_mismatch`). Only `HTTP_GATEWAY_RESPONSE_TIME` is `LOGS_ANALYTICS`, which is why it's the only one wired here.
- The field exists on the `v4-metrics` index the logs search hits: the reporter writes `endpointResponseTimeMs` and analytics resolves `HTTP_ENDPOINT_RESPONSE_TIME -> "endpoint-response-time-ms"` (`HTTPFieldResolver`). The logs connection-log adapter (`RequestV2MetricsV4Fields`) just doesn't declare it yet.

So exposing it on logs is a deliberate **catalog widening** (add LOGS to the signal) + filter wiring (`SearchLogsFilters` range field + `SearchMetricsQueryAdapter` translation + a `case` here), not a bug. Tracked as low-priority follow-up: **GMA-718** — https://gravitee.atlassian.net/browse/GMA-718. Out of scope for this PR.

PR #17859, opened 2026-06-16
feat(portal): add collapsible preview panel to navigation items edito…

adam-rosowicz:
Adds a slide toggle ("Preview") to the editor panel header that shows/hides the live preview column across all page types.

**Behaviour**

Toggle defaults to ON (preview visible)
State is page-level — persists when switching between navigation items
Editor always expands to fill the full panel width when preview is hidden

image
image

### DIFF
--- a/gravitee-apim-console-webui/src/portal/components/openapi-editor/openapi-editor.component.html
+++ b/gravitee-apim-console-webui/src/portal/components/openapi-editor/openapi-editor.component.html
@@ -25,20 +25,22 @@
       [disabled]="disabled"
     ></gio-monaco-editor>
   </div>
-  <div class="content__editor__cell">
-    @if (isRedocViewer()) {
-      <gio-redoc [spec]="_value" [tryItURL]="swaggerTryItURL()"></gio-redoc>
-    } @else {
-      <gio-swagger-ui
-        [spec]="_value"
-        [tryItURL]="swaggerTryItURL()"
-        [docExpansion]="swaggerDocExpansion()"
-        [displayOperationId]="swaggerDisplayOperationId()"
-        [filter]="swaggerFilter()"
-        [showExtensions]="swaggerShowExtensions()"
-        [showCommonExtensions]="swaggerShowCommonExtensions()"
-        [maxDisplayedTags]="swaggerMaxDisplayedTags()"
-      ></gio-swagger-ui>
-    }
-  </div>
+  @if (showPreview()) {
+    <div class="content__editor__cell">
+      @if (isRedocViewer()) {
+        <gio-redoc [spec]="_value" [tryItURL]="swaggerTryItURL()"></gio-redoc>
+      } @else {
+        <gio-swagger-ui
+          [spec]="_value"
+          [tryItURL]="swaggerTryItURL()"
+          [docExpansion]="swaggerDocExpansion()"
+          [displayOperationId]="swaggerDisplayOperationId()"
+          [filter]="swaggerFilter()"
+          [showExtensions]="swaggerShowExtensions()"
+          [showCommonExtensions]="swaggerShowCommonExtensions()"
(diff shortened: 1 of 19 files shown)


gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.spec.ts
```diff
@@ -2839,4 +2839,49 @@ describe('PortalNavigationItemsComponent', () => {
...
+      title: 'Page 2',
+      portalPageContentId: 'content-2',
+    });
+
+    beforeEach(async () => {
+      await expectGetNavigationItems(fakePortalNavigationItemsResponse({ items: [page1, page2] }));
+      expectGetPageContent('content-1', 'Page 1 content');
+    });
+
+    it('should flip isPreviewVisible from true to false and back when toggle is clicked', async () => {
```
kovaren: Just testing that a boolean is flipped on a component is not very valuable. Instead, we should test the observable consequence of a toggle change.

I suggest adding a i`sPreviewVisible()` method on the harness which would check the actual DOM instead of relying on a component's field.

gravitee-apim-webui-libs/gravitee-markdown/src/lib/gravitee-markdown-editor/gravitee-markdown-editor.component.spec.ts
```diff
@@ -98,4 +99,19 @@ describe('GraviteeMarkdownEditorComponent', () => {
       expect(component.editorControl.value).toBe(newValue);
     });
   });
+
+  describe('showPreview input', () => {
```
kovaren, line 103: Similar tests are missing for `asyncapi` and `openapi`.

gravitee-apim-console-webui/src/portal/components/openapi-editor/openapi-editor.component.html
```diff
@@ -25,7 +25,9 @@
       [disabled]="disabled"
     ></gio-monaco-editor>
   </div>
-  <div class="content__editor__cell">
-    <gio-swagger-ui [spec]="_value"></gio-swagger-ui>
-  </div>
+  @if (showPreview()) {
```
kovaren, line 28: This PR is in conflict with https://github.com/gravitee-io/gravitee-api-management/pull/16736 that added support for Redoc editor.

As part of the resolution, the preview toggle should also be handled by the Redoc editor.

kovaren (changes requested):

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.spec.ts
adam-rosowicz: Done. Added isPreviewVisible() to the harness using toggle.isChecked() for the state-retention test, and switched the toggle-click tests to check rootLoader.getHarnessOrNull(GraviteeMarkdownViewerHarness) — verifying whether the markdown viewer element is actually present in the DOM rather than reading the component signal.

gravitee-apim-webui-libs/gravitee-markdown/src/lib/gravitee-markdown-editor/gravitee-markdown-editor.component.spec.ts
adam-rosowicz: Done.

gravitee-apim-console-webui/src/portal/components/openapi-editor/openapi-editor.component.html
adam-rosowicz: Resolved.

kovaren (approved):

PR #17884, opened 2026-06-16
feat(authz): Gamma multi-tenant AuthZEN PDP — gateway sync, provisioning & wire model

podlesrafal:
## What

Adds the gateway-side support for the **Gamma multi-tenant AuthZEN PDP**: the gateway now provisions and synchronizes per-tenant authorization engines keyed by `(environmentId, targetPdpId)`, instead of a single node-wide engine.

## Changes

**Gateway services-sync** (`gravitee-apim-gateway-services-sync`)
- `AuthzPdpSynchronizer` — relays PDP provision/evict over the Vert.x event bus, filtered by node sharding tag; hydrates a freshly-provisioned scope from the `events_latest` backfill.
- `AuthzPolicySynchronizer` / `AuthzEntitySynchronizer` — route policies/entities to their target scopes; declarative placement reconcile that evicts policies/entities from dropped scopes on re-target.
- `EventBusAuthzEnginePort` — scoped / bare-default / env-broadcast addressing; per-address commit resilient to a dead scope.
- `AuthzPdpMapper`, `AuthzPdpProvisionDeployable`; deployers retry a failed stage instead of dropping it.
- Removed the obsolete replay responder/consumer and dead `gatewayId` plumbing.

**Wire / model**
- `definition-model`: `ApiType.AUTHZ`.
- `gamma definition-model`: `AuthzPdp` / `AuthzPolicy` / `AuthzEntity` carry `targetPdpId(s)`.
- `repository` + `rest-api`: `PUBLISH_AUTHZ_PDP` / `UNPUBLISH_AUTHZ_PDP` event types (+ OpenAPI v2) and an `AUTHZ_PDP_ID` event property for PDP filtering.

**Tests** — synchronizer chaos / gaps / retry / scope / hydration coverage.

## Notes

- Rebased onto the latest `4.12.x`. Depends on `gravitee-gateway-api` **6.0.0-alpha.9** (the temporary `*-AUTHZ-SNAPSHOT` pin was dropped during the rebase).
- This is the APIM side of a multi-repo feature; it pairs with the `gravitee-service-authz-pdp` service plugin (hosts the per-scope engines) and the `gravitee-reactor-authz` entrypoint/reactor (AuthZEN HTTP API).
- Draft: opened for CI / review; the branch has not been locally build-verified after the rebase.

## Diff
48 commits · 69 files · +4431 / −147


gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
```diff
@@ -43,49 +50,125 @@ public class EventBusAuthzEnginePort implements AuthzEnginePort {
...
 
     @Override
     public Completable commit() {
-        return vertx
-            .eventBus()
-            .<JsonObject>rxRequest(SYNC_ADDRESS, new JsonObject().put("op", OP_COMMIT), deliveryOptions)
-            .flatMapCompletable(this::verifyCommitReply);
+        return Completable.defer(() -> {
+            List<String> addresses = new ArrayList<>(touched);
+            touched.clear();
```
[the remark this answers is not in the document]
podlesrafal: Done — switched to `touched.removeAll(addresses)`. An address armed by another thread between the snapshot copy and the removal now survives for the next commit cycle instead of being silently dropped.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzEntitySynchronizer.java
```diff
@@ -156,28 +183,104 @@ private Flowable<AuthzEntityReactorDeployable> prepareForUndeployment(Flowable<E
...
+        AuthzEntityReactorDeployable deployable,
+        AtomicBoolean touched,
+        Set<String> attemptedThisCycle
+    ) {
+        String rk = runtimeKey(deployable);
+        attemptedThisCycle.add(rk);
+        // Latest control-plane intent for this entity is UNPUBLISH: cancel any pending staging.
+        pendingDeploys.remove(rk);
         return deployer
             .undeploy(deployable)
```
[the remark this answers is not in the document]
podlesrafal: Done — `undeployEntity` now falls back to the last known `placement` for the runtime key when the UNPUBLISH carries no targetPdpIds, so the entity is evicted from the scopes it was actually deployed to. Added a regression test (`unpublish_without_targetPdpIds_falls_back_to_last_placement_for_eviction`).

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPolicySynchronizer.java
```diff
@@ -156,28 +183,104 @@ private Flowable<AuthzPolicyReactorDeployable> prepareForUndeployment(Flowable<E
...
+        AuthzPolicyReactorDeployable deployable,
+        AtomicBoolean touched,
+        Set<String> attemptedThisCycle
+    ) {
+        String rk = runtimeKey(deployable);
+        attemptedThisCycle.add(rk);
+        // Latest control-plane intent for this doc is UNPUBLISH: cancel any pending staging.
+        pendingDeploys.remove(rk);
         return deployer
             .undeploy(deployable)
```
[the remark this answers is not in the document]
podlesrafal: Done — same fix as the entity synchronizer: `undeployPolicy` falls back to the `placement` map when the UNPUBLISH has no targetPdpIds. Covered by a new test in AuthzPolicySynchronizerScopeTest.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
```diff
@@ -0,0 +1,421 @@
...
+                    return true;
+                })
+                .toList()
+        );
+    }
+
+    private Flowable<AuthzPdpProvisionDeployable> tagGate(AuthzPdpProvisionDeployable deployable, boolean initialSync) {
+        if (matchesNodeTag(deployable)) {
+            return Flowable.just(deployable);
+        }
```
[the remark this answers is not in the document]
podlesrafal: Done — the tag gate now always lets UNDEPLOY through (`syncAction() == UNDEPLOY || matchesNodeTag(...)`), so a re-tagged or deleted PDP is still evicted on a node whose tags no longer match. Eviction is idempotent. Added a test (`evict_for_a_non_matching_tag_is_still_relayed_not_dropped`).

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
```diff
@@ -0,0 +1,421 @@
...
+        Set<String> livePublishKeys = deployables
+            .stream()
+            .filter(d -> d.syncAction() == SyncAction.DEPLOY)
+            .map(AuthzPdpSynchronizer::runtimeKey)
+            .collect(Collectors.toSet());
+        return Flowable.fromIterable(
+            deployables
+                .stream()
+                .filter(d -> {
+                    if (d.syncAction() == SyncAction.UNDEPLOY && livePublishKeys.contains(runtimeKey(d))) {
```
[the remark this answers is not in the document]
podlesrafal: Keeping the current behavior here. This synchronizer reads from events_latest (keyed by pdp.id via LatestEventFetcher.fetchLatest), not the raw event stream. A publish-then-unpublish of the same PDP collapses to only the UNPUBLISH, so there is no DEPLOY+UNDEPLOY pair for that scope and the evict is not suppressed. The suppression only fires for cross-id scope reuse (delete pdp-A + create pdp-B sharing the same targetPdpId), where the engine must stay — respecting chronological order as suggested would wrongly re-evict the reused scope. I can add a code comment documenting this if it helps.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
```diff
@@ -0,0 +1,423 @@
...
+
+    private Completable backfillScope(AuthzPdpProvisionDeployable provision, Set<String> environments) {
+        String targetPdpId = provision.targetPdpId();
+        String environmentId = provision.environmentId();
+        Completable policies = eventsFetcher
+            .fetchLatest(-1L, null, Event.EventProperties.AUTHZ_POLICY_ID, environments, POLICY_BACKFILL_TYPES)
+            .flatMapIterable(events -> events)
+            .flatMapMaybe(policyMapper::toDeploy)
+            .filter(deployable -> Objects.equals(environmentId, deployable.environmentId()))
+            .filter(deployable -> matchesScope(deployable.targetPdpIds(), targetPdpId))
```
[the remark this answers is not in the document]
podlesrafal: Done (with a deliberate tweak): the backfill query is now scoped to the provision's environment via `Set.of(environmentId)`, so the DB filters by environment — the efficiency win you flagged. I kept the in-memory env check as a defense-in-depth guard for cross-environment isolation; it is now a cheap no-op on the already-env-scoped result, and AuthzPdpHydrationTest pins that isolation independently of the fetcher's behaviour.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
```diff
@@ -0,0 +1,423 @@
...
+                    Set.of(targetPdpId)
+                )
+            );
+
+        Completable entities = eventsFetcher
+            .fetchLatest(-1L, null, Event.EventProperties.AUTHZ_ENTITY_ID, environments, ENTITY_BACKFILL_TYPES)
+            .flatMapIterable(events -> events)
+            .flatMapMaybe(entityMapper::toDeploy)
+            .filter(deployable -> Objects.equals(environmentId, deployable.environmentId()))
+            .filter(deployable -> matchesScope(deployable.targetPdpIds(), targetPdpId))
```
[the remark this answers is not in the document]
podlesrafal: Done — same as the policy backfill: `Set.of(environmentId)` is passed to fetchLatest for entities so the query is env-scoped at the DB level, with the in-memory env filter retained as a cross-environment safety guard.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
```diff
@@ -43,49 +50,127 @@ public class EventBusAuthzEnginePort implements AuthzEnginePort {
...
+                } else {
+                    commits.add(
+                        vertx
+                            .eventBus()
+                            .<JsonObject>rxRequest(address, c, deliveryOptions)
+                            .flatMapCompletable(this::verifyCommitReply)
+                            .onErrorComplete(t -> {
+                                log.warn("Authz PDP commit to address [{}] failed, skipping it for this cycle", address, t);
+                                return true;
+                            })
```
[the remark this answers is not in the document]
podlesrafal: Done — the onErrorComplete now re-arms the address (`touched.add(address)`) so a transient commit failure is retried on the next cycle instead of leaving the scope's staged mutations unsealed.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzEntitySynchronizer.java
```diff
@@ -104,11 +126,16 @@ public int order() {
         return Order.AUTHZ_ENTITY.index();
     }
 
-    private Completable commitIfAny(AtomicLong processed) {
-        return Completable.defer(() -> processed.get() == 0L ? Completable.complete() : enginePort.commit());
+    private Completable commitIfAny(AtomicBoolean touched) {
+        return Completable.defer(() -> touched.get() ? enginePort.commit() : Completable.complete());
     }
```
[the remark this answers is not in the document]
podlesrafal: Done — commitIfAny now always calls `enginePort.commit()`. commit() short-circuits internally when nothing is armed, so the no-bus-round-trip optimisation is preserved, while a previous cycle's re-armed failed commit now gets retried. Updated the affected tests (no_events / retry).

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPolicySynchronizer.java
```diff
@@ -104,11 +126,16 @@ public int order() {
         return Order.AUTHZ_POLICY.index();
     }
 
-    private Completable commitIfAny(AtomicLong processed) {
-        return Completable.defer(() -> processed.get() == 0L ? Completable.complete() : enginePort.commit());
+    private Completable commitIfAny(AtomicBoolean touched) {
+        return Completable.defer(() -> touched.get() ? enginePort.commit() : Completable.complete());
     }
```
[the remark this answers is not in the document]
podlesrafal: Done — same change as AuthzEntitySynchronizer: commitIfAny always calls commit() (which short-circuits when nothing is armed), so failed commits re-armed via EventBusAuthzEnginePort are retried on a later cycle.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
```diff
@@ -0,0 +1,425 @@
...
+        String environmentId = provision.environmentId();
+        // Scope the backfill query to this provision's environment at the DB level (efficiency); the
+        // in-memory env check below stays as a defense-in-depth guard against cross-environment leakage.
+        Completable policies = eventsFetcher
+            .fetchLatest(-1L, null, Event.EventProperties.AUTHZ_POLICY_ID, Set.of(environmentId), POLICY_BACKFILL_TYPES)
+            .flatMapIterable(events -> events)
+            .flatMapMaybe(policyMapper::toDeploy)
+            .filter(deployable -> Objects.equals(environmentId, deployable.environmentId()))
+            .filter(deployable -> matchesScope(deployable.targetPdpIds(), targetPdpId))
+            .concatMapCompletable(deployable ->
```
sclarkamlab: backfillScope applies docs straight through enginePort here, but it never records the scope in the entity/policy synchronizers' `placement` map (placement is private to those two classes, this class has no reference to it). Eviction is computed as `dropped = placement - newTarget` over in AuthzEntitySynchronizer:196, so it can't see hydration-applied scopes.

Scenario: scope B gets provisioned and hydrated with doc X, then a later PUBLISH narrows X to {A}. placement(X) is still {A} (B never went through a normal sync cycle), so B never gets evicted and X lingers on B. Same gap on UNPUBLISH with empty targets at AuthzEntitySynchronizer:227, where the placement fallback is empty and route() short-circuits without removing anything.

I think placement needs to be shared state, or hydration needs to feed back into it.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
```diff
@@ -43,49 +50,130 @@ public class EventBusAuthzEnginePort implements AuthzEnginePort {
...
+                JsonObject c = new JsonObject().put("op", OP_COMMIT);
+                if (isBroadcast(address)) {
+                    commits.add(Completable.fromAction(() -> vertx.eventBus().publish(address, c)));
+                } else {
+                    commits.add(
+                        vertx
+                            .eventBus()
+                            .<JsonObject>rxRequest(address, c, deliveryOptions)
+                            .flatMapCompletable(this::verifyCommitReply)
+                            .onErrorComplete(t -> {
```
sclarkamlab: This re-arms the address on every commit failure with no attempt cap, unlike the deploy path which caps at MAX_PENDING_ATTEMPTS. commitIfAny() runs every cycle, so a permanently dead scope pays the full 5s REPLY_TIMEOUT_MS on every sync cycle, forever. Probably wants the same bounded-retry treatment the deploys get.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
```diff
@@ -43,49 +50,130 @@ public class EventBusAuthzEnginePort implements AuthzEnginePort {
...
+                            })
+                    );
+                }
+            }
+            return Completable.merge(commits);
+        });
+    }
+
+    @Override
+    public Completable commitScope(String environmentId, String targetPdpId) {
```
sclarkamlab, line 144: commitScope arms one scope then calls the global commit(), which drains and commits every armed address in the shared `touched` set. With a single EventBusAuthzEnginePort bean that means a hydration commit for scope B also retries a dead scope A's address, so per-scope timeouts get coupled across scopes and synchronizers. Combined with the uncapped retry just above, a few dead scopes can stack up real latency on every cycle.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzEntityMapper.java
```diff
@@ -74,6 +76,8 @@ public Maybe<AuthzEntityReactorDeployable> toUndeploy(Event event) {
                     .engineUid(toEngineUid(kind, wire.getEntityType(), wire.getEntityId()))
                     .kind(kind)
                     .entityType(wire.getEntityType())
+                    .environmentId(wire.getEnvironmentId())
+                    .targetPdpIds(wire.getTargetPdpIds() == null ? java.util.Set.of() : wire.getTargetPdpIds())
```
sclarkamlab: Heads up on the kind default a few lines up (the `kind != null ? ... : RESOURCE` at ~71-73): it's fine when entityType is present since toEngineUid uses it verbatim, but a legacy UNPUBLISH that omits both kind and entityType for a PRINCIPAL builds a RESOURCE-shaped engineUid, so removeEntity won't match the live principal and it leaks. Are we guaranteed the control plane always sends kind or entityType on UNPUBLISH?

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPolicySynchronizer.java
```diff
@@ -156,28 +182,106 @@ private Flowable<AuthzPolicyReactorDeployable> prepareForUndeployment(Flowable<E
...
+                .concatMapCompletable(deployable -> undeployPolicy(deployer, deployable, attemptedThisCycle).ignoreElements())
+                .andThen(
+                    Flowable.fromIterable(deploys).concatMapCompletable(deployable ->
+                        deployPolicy(deployer, deployable, attemptedThisCycle).ignoreElements()
+                    )
+                );
+        });
+    }
+
+    private List<AuthzPolicyReactorDeployable> drivable(Map<String, AuthzPolicyReactorDeployable> pending, Set<String> attemptedThisCycle) {
```
sclarkamlab: This whole retry/placement engine (drivable, deployPolicy/undeployPolicy, retryPending, the pending maps) is basically identical to AuthzEntitySynchronizer, and there's a third near-copy in AuthzPdpSynchronizer. The two synchronizers are ~90% identical line for line (288 lines each, only 28 differ once you normalize the Entity/Policy naming). A fix to the shared logic has to land in 2-3 places or it silently diverges. Worth pulling out an abstract base parameterized over the deployable type and mapper.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzEntityMapper.java
```diff
@@ -47,6 +47,8 @@ public Maybe<AuthzEntityReactorDeployable> toDeploy(Event event) {
                     .entityType(wire.getEntityType())
                     .attributes(wire.getAttributes())
                     .parents(wire.getParents())
+                    .environmentId(wire.getEnvironmentId())
+                    .targetPdpIds(wire.getTargetPdpIds() == null ? java.util.Set.of() : wire.getTargetPdpIds())
```
sclarkamlab: Minor: this null-to-empty coalescing is repeated at 4 mapper sites (here plus :80, and AuthzPolicyMapper:63,92), and then both synchronizers re-check `== null` at AuthzEntitySynchronizer:195 even though the mapper already guarantees non-null. Could collapse to one helper and drop the redundant check.

sclarkamlab (commented):
Second-opinion review of the authz PDP scope-targeting changes (base 4.12.x). Findings inline, ranked roughly by severity. The placement/hydration one is the only hard correctness bug, the rest are resilience/cleanup.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
podlesrafal: Fixed in 2857fd71. Placement is now a shared `AuthzScopePlacement` bean injected into both the entity/policy synchronizers and this hydration path; `backfillScope` records each hydrated scope (`policyPlacement.addScope` / `entityPlacement.addScope`). A later narrowing PUBLISH now sees the hydrated scope in `applied` and evicts it. Added `AuthzHydrationPlacementTest` reproducing the provision-after-publish → narrow scenario.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
podlesrafal: Fixed in 2857fd71. `commit()` now re-arms via `commitOne()` with a per-address attempt counter capped at `MAX_COMMIT_ATTEMPTS`; once exceeded the address is abandoned instead of retried forever. Covered by `commit_stops_re_arming_a_permanently_dead_address_after_the_attempt_cap`.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
podlesrafal: Fixed in 2857fd71. `commitScope` no longer arms-then-drains the shared `touched` set; it commits only its own address via `commitOne`, so a hydration commit for one scope no longer retries another scope's dead address. Covered by `commitScope_commits_only_its_own_address_and_leaves_other_armed_scopes_for_the_global_commit`.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzEntityMapper.java
podlesrafal: Verified the contract: this is a non-issue for real events, so behaviour is unchanged. `AuthzEntity.kind` is `@JsonProperty(required = true)` + `@NotNull`, and the publisher (`EventRepositoryAuthzEventPublisher.unpublishEntity`) always sets `kind` on UNPUBLISH. So a PRINCIPAL can never reach the RESOURCE default — that branch only exists defensively for malformed/legacy minimal payloads (which the mapper tests exercise). Left as-is rather than tightening it and breaking that tolerance; happy to add an assertion if you'd prefer.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPolicySynchronizer.java
podlesrafal: Fixed in 2857fd71. Extracted `AbstractAuthzReactorSynchronizer<D extends AuthzScopedDeployable>`; the entity and policy synchronizers are now thin subclasses supplying only the type-specific bindings (mapper, deployer, event types, labels). The retry/placement engine lives once, so a shared-logic fix lands in one place.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzEntityMapper.java
podlesrafal: Fixed in 2857fd71. Collapsed the four coalescing sites into `AuthzWire.targetPdpIdsOrEmpty(...)` and dropped the redundant `== null` re-check in the synchronizer (the base now relies on the mapper's non-null guarantee).

podlesrafal:
Thanks for the second-opinion pass @sclarkamlab — all six addressed in 2857fd71, threads resolved:

- **Placement/hydration eviction gap** (the hard bug) → scope placement is now a shared `AuthzScopePlacement` fed by the PDP hydration backfill, so a hydration-applied scope is evicted on a later narrowing PUBLISH. New `AuthzHydrationPlacementTest` reproduces it.
- **Uncapped commit re-arm** → bounded by `MAX_COMMIT_ATTEMPTS` in `commitOne()`.
- **commitScope cross-scope coupling** → `commitScope` commits only its own address, no longer drains the shared `touched` set.
- **Synchronizer duplication** → extracted `AbstractAuthzReactorSynchronizer<D>`; entity/policy are now thin subclasses.
- **Repeated null→empty coalescing** → single `AuthzWire` helper, redundant null re-check dropped.
- **`kind` default on UNPUBLISH** → verified non-issue: `AuthzEntity.kind` is `required=true`/`@NotNull` and the publisher always sets it, so the RESOURCE default is unreachable for real events (kept as a defensive fallback for malformed payloads). Behaviour unchanged — shout if you'd rather tighten it.

Full `gravitee-apim-gateway-services-sync` suite green (577 tests).

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
```diff
@@ -40,52 +47,146 @@ public class EventBusAuthzEnginePort implements AuthzEnginePort {
...
+
+    private Completable route(String environmentId, JsonObject command, Set<String> targetPdpIds) {
+        if (targetPdpIds == null || targetPdpIds.isEmpty()) {
+            return Completable.complete();
+        }
+        List<Completable> sends = new ArrayList<>();
+        for (String targetPdpId : targetPdpIds) {
+            String address = addressFor(environmentId, targetPdpId);
+            if (isBroadcast(address)) {
+                // Broadcast is fire-and-forget (no reply): arm eagerly so the matching OP_COMMIT fans out.
```
remibaptistegio: It's ok to have this behavior ? No way to collect the address of all the PDPs ?

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
```diff
@@ -40,52 +47,146 @@ public class EventBusAuthzEnginePort implements AuthzEnginePort {
...
+    }
+
+    private Completable route(String environmentId, JsonObject command, Set<String> targetPdpIds) {
+        if (targetPdpIds == null || targetPdpIds.isEmpty()) {
+            return Completable.complete();
+        }
+        List<Completable> sends = new ArrayList<>();
+        for (String targetPdpId : targetPdpIds) {
+            String address = addressFor(environmentId, targetPdpId);
+            if (isBroadcast(address)) {
```
remibaptistegio: It's possible to have a list with multiple "broadcast" address ? Maybe we should have something to avoid this no ? or there is already something elsewhere in the code that prevents this.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
```diff
@@ -0,0 +1,440 @@
...
+                    .environmentId(deployable.environmentId())
+                    .tag(deployable.tag())
+                    .syncAction(SyncAction.UNDEPLOY)
+                    .build()
+            );
+        }
+        return Flowable.empty();
+    }
+
+    private boolean matchesNodeTag(AuthzPdpProvisionDeployable deployable) {
```
remibaptistegio, line 248: This is not the same behavior as APIs no ? Is it intended ?
For API, if there is no tag, it will be deployed if the gateway has no tag also, but if the gateway specify some tags, the API won't be deploy.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
```diff
@@ -0,0 +1,440 @@
...
+        return policies
+            .andThen(entities)
+            .andThen(enginePort.commitScope(environmentId, targetPdpId))
+            .doOnComplete(() -> {
+                pendingHydrations.remove(runtimeKey(provision));
+                log.debug("Hydrated authz PDP scope [{}] from events_latest backfill", targetPdpId);
+            })
+            .onErrorResumeNext(t -> {
+                // Remember the scope so its hydration is re-driven next cycle instead of leaving it cold.
+                pendingHydrations.put(runtimeKey(provision), provision);
```
remibaptistegio: There is no limit here no ? If there is an error, it will be retried every cycle ?

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
```diff
@@ -258,21 +270,30 @@ private Flowable<AuthzPdpProvisionDeployable> tagGate(AuthzPdpProvisionDeployabl
...
+                    deployable.syncAction(),
+                    deployable.targetPdpId(),
+                    nodeTags
+                );
+            }
+            return matches;
         }
-        boolean matches = gatewayConfiguration.hasMatchingTags(Set.of(tag));
+        // Tagged PDP runs only on nodes carrying that exact tag — no catch-all for untagged nodes.
+        boolean matches = nodeTags.contains(tag);
```
remibaptistegio: This is not the same behavior that we have for API. Why don't you use the hasMatchingTags method ?

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
podlesrafal: Fixed by mirroring the provision retry path: added a pendingHydrationAttempts counter, abandon after MAX_PENDING_ATTEMPTS (10) with a warn, and clear the counter on hydration success and on both evict paths so it can't leak

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
podlesrafal: This is intentional. `hasMatchingTags` returns true for any object when the node has no sharding tags (untagged gateway = catch-all). For PDP engines that conflicts with our strict per-node placement model: a tagged PDP would then also be provisioned on every untagged/catch-all gateway, cloning the engine everywhere. In our model "everywhere" is the role of the single global default engine; a named/tagged engine must live only on nodes carrying its exact tag.

Example — gateways gw(untagged) + gw(tag=us):

PDP(targetPdpId=Y, tag=us) → engine Y only on the us node.
with hasMatchingTags it would also land on gw(untagged), which we explicitly don't want.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
podlesrafal: It can't happen by construction. targetPdpIds is a Set, and addressFor returns a broadcast address only for the single wildcard token *. Every other value maps to a per-scope or the default address. So a call yields at most one broadcast address; duplicates aren't possible and no extra guard is needed.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
podlesrafal: The sync layer doesn't hold the authoritative list of every PDP scope in an environment. A node only knows the scopes it actually hosts (via AuthzHostedScopes), populated by provisioning. Enumerating "all PDPs" would require pulling the full control-plane PDP registry into the gateway. Instead * publishes once to the env broadcast address and each named engine subscribed on this node receives it..

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
podlesrafal: An untagged PDP is provisioned only on untagged gateways (matches = nodeTags.isEmpty()), exactly like an untagged API which is only deployed where the gateway carries no sharding tags.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
remibaptistegio: I'm not sure to understand why you can't follow the same rule. If I want to deploy different PDP engines that manage different scope on the same gateway, I will need to update the gateway tags to add all these PDP tags and that may impact my APIs deployment if they do not match those tags.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
remibaptistegio: I'm not sure to understand, maybe I miss something. If my gateway hosts PDP A and we have declared in the CP: PDP A and PDP B. If the synchronizer fetch an entity that target *, we have to deploy it to the PDP A that the gateway host no ? So if we know what the gateway hosts, we can just rxSend to this list (in our example, only PDP A) no ?

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
```diff
@@ -0,0 +1,465 @@
...
+
+    private Completable hydrate(ConcurrentLinkedQueue<AuthzPdpProvisionDeployable> provisionedScopes) {
+        return Flowable.defer(() -> {
+            Map<String, AuthzPdpProvisionDeployable> toHydrate = new java.util.LinkedHashMap<>(pendingHydrations);
+            provisionedScopes.forEach(d -> toHydrate.put(runtimeKey(d), d));
+            return Flowable.fromIterable(toHydrate.values());
+        }).concatMapCompletable(this::backfillScope);
+    }
+
+    private Completable backfillScope(AuthzPdpProvisionDeployable provision) {
```
remibaptistegio: I understand why you add to create this method but it's going against the core principle of the sync process. We now have this duplicated pipeline that mimics part of the sync process which could lead to some drift if we do not pay attention to it. Instead of bringing this complexity inside the sync process, did you explore a way to manage this inside the PDP engine ? Instead of creating multiple engines, create one global one with some kind of multitenant approach where you expose a new address when a new reactor is deployed and use the data already available in the engine to feed this new reactor with the needed entities.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
podlesrafal: You're right that the semantics differ from hasMatchingTags, and the difference is intentional.
Why it's not hasMatchingTags today: that matcher has a catch-all and an untagged gateway runs every API, and a tagged API also runs on an untagged gateway. Applying that to a named PDP engine would mean the engine gets cloned onto every untagged (and every overlapping) node. A named scope is meant to run in a specific placement, not "everywhere by default", So matchesNodeTag is strict membership without the catch-all.

But as I understand your point is that PDP placement currently reads gravitee.tags, the same source the API synchronizer uses but PDP sync uses it differently.

I see two directions:

- Keep the shared tags and document that "a gateway hosting PDP scope X must carry tag X, which also makes it an API node for X". Simple, no new config, but PDP and API placement stay coupled.
- Decouple with a dedicated PDP placement config (e.g. authz.pdp.tags or a per-PDP node selector), so you can host us + eu engines on one gateway without affecting its API deployment.

The reason that the tagging was introduces was to be able to start separate "sidecar" next to gateway which is tagged to narrow down the scope. Entities and policies can be heavy so to not consume memory twice (on untagged and tag A gateway) I narrowed it.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
podlesrafal: Ok, agreed. Replaced the env-scoped broadcast with per-scope enumeration of what the node actually hosts.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
podlesrafal: I looked into the multitenant-engine direction in gravitee-service-authz-pdp and I see two blockers:

1. No scope dimension in the engine. gravitee-authz-engine rebuilds a full immutable snapshot (PolicySet + Entities) from the whole staged set on every commit, and knows nothing about scopes. I want to keep engine as light as possible. Introducing scopes will be difficult and make this engine heavy. 
2. "Data already in the engine" isn't there. Hydration fires when a scope is first provisioned on a node; its docs live in events_latest, not in RAM (cold start = nothing loaded; new-scope = filtered out by placement). To feed from in-engine data we'd have to push every scope to every node, which defeats sharded placement and makes each gateway a full PDP for all scopes. I would like to avoid data duplications in PDP. Just keep data in engine instance where it is needed. So within the current model (DB as source of truth + sharded placement), a DB fetch on cold start / new scope is unavoidable.

What I think actually fixes the drift concern without touching the engine: fold backfillScope into the normal synchronizer path, run the existing synchronizer in full mode (from = -1) scoped to the new targetPdpId, instead of a parallel fetch→map→apply. One implementation, no second pipeline. Would that resolve it for you?

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
```diff
@@ -0,0 +1,488 @@
...
+                deployable.syncAction(),
+                deployable.targetPdpId(),
+                tag,
+                nodeTags
+            );
+        }
+        return matches;
+    }
+
+    private Completable send(AuthzPdpProvisionDeployable deployable, ConcurrentLinkedQueue<AuthzPdpProvisionDeployable> provisionedScopes) {
```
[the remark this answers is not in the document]
podlesrafal: Done in 84fe799ff5 — renamed `send` to `relayProvisionOrEvict`.

gravitee-gamma/gravitee-gamma-definition-model/src/main/java/io/gravitee/gamma/definition/authz/AuthzPdp.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @JsonProperty("targetPdpId")
+    @NotBlank
+    private String targetPdpId;
+
+    @JsonProperty
+    private String environmentId;
+
+    /** Stable sharding tag a PDP targets. Gateways provision a PDP only when this tag is among their
+     *  configured sharding tags. Null/blank means the PDP is unsharded and provisions on all gateways. */
```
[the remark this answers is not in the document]
podlesrafal: Done in 84fe799ff5 — the javadoc now states an untagged PDP provisions only on untagged gateways, matching the placement logic in matchesNodeTag.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
```diff
@@ -40,52 +52,162 @@ public class EventBusAuthzEnginePort implements AuthzEnginePort {
...
+            touched.removeAll(addresses);
+            if (addresses.isEmpty()) {
+                return Completable.complete();
+            }
+            List<Completable> commits = new ArrayList<>(addresses.size());
+            for (String address : addresses) {
+                commits.add(commitOne(address));
+            }
+            return Completable.merge(commits);
+        });
```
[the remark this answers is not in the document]
podlesrafal: Done in 84fe799ff5 — extracted the defer body into a named `commitTouchedSnapshot` method.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/spring/RepositorySyncConfiguration.java
```diff
@@ -369,6 +373,26 @@ public SharedPolicyGroupSynchronizer sharedPolicyGroupSynchronizer(
...
+    @Bean
+    @Conditional(GammaEnabledCondition.class)
+    public AuthzScopePlacement authzEntityScopePlacement() {
+        return new AuthzScopePlacement();
+    }
+
+    @Bean
+    @Conditional(GammaEnabledCondition.class)
+    public AuthzScopePlacement authzPolicyScopePlacement() {
+        return new AuthzScopePlacement();
```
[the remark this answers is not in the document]
podlesrafal: Won't fix — the two beans intentionally back distinct placement state (entity vs policy). Merging them would share one scope map across both document types, so the duplication is deliberate.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzEntityMapper.java
```diff
@@ -47,6 +47,8 @@ public Maybe<AuthzEntityReactorDeployable> toDeploy(Event event) {
                     .entityType(wire.getEntityType())
                     .attributes(wire.getAttributes())
                     .parents(wire.getParents())
+                    .environmentId(wire.getEnvironmentId())
```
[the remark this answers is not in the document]
podlesrafal: Leaving as-is — the shared null-coalescing is tiny and the toDeploy/toUndeploy builders read clearer inline than behind a helper.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPolicyMapper.java
```diff
@@ -59,6 +59,8 @@ public Maybe<AuthzPolicyReactorDeployable> toDeploy(Event event) {
                     .policyText(wire.getPolicyText())
                     .kind(kind)
                     .entityId(wire.getEntityId())
+                    .environmentId(wire.getEnvironmentId())
```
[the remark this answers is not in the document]
podlesrafal: Leaving as-is — same as the entity mapper: the inline builders are clearer than an extracted helper for two small call sites.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
```diff
@@ -40,52 +52,162 @@ public class EventBusAuthzEnginePort implements AuthzEnginePort {
...
+    }
+
+    private Completable commitOne(String address) {
+        JsonObject c = new JsonObject().put("op", OP_COMMIT);
         return vertx
             .eventBus()
-            .<JsonObject>rxRequest(SYNC_ADDRESS, new JsonObject().put("op", OP_COMMIT), deliveryOptions)
-            .flatMapCompletable(this::verifyCommitReply);
+            .<JsonObject>rxRequest(address, c, deliveryOptions)
```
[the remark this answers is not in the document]
podlesrafal: Leaving as-is — the reply-verify plus bounded-retry chain in commitOne is cohesive; splitting it would scatter the retry bookkeeping across methods for little gain.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
```diff
@@ -40,52 +52,162 @@ public class EventBusAuthzEnginePort implements AuthzEnginePort {
...
+                    commitAttempts.remove(address);
+                    log.warn("Abandoning authz PDP commit to address [{}] after {} attempts", address, attempts - 1);
+                    return true;
+                }
+                // Re-arm the address so a transient commit failure is retried next cycle, instead of
+                // leaving the scope's staged mutations unsealed until a later mutation touches it.
+                touched.add(address);
+                log.warn("Authz PDP commit to address [{}] failed (attempt {}), will retry next cycle", address, attempts, t);
+                return true;
+            });
```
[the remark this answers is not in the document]
podlesrafal: Leaving as-is — see the commitOne thread above; the chain is intentionally kept together.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzEntityDeployer.java
```diff
@@ -31,8 +32,21 @@ public class AuthzEntityDeployer implements Deployer<AuthzEntityReactorDeployabl
...
+        return eviction
+            .andThen(
+                enginePort.addOrUpdateEntity(
+                    deployable.environmentId(),
+                    deployable.engineUid(),
+                    deployable.attributes(),
+                    deployable.parents(),
+                    deployable.targetPdpIds()
+                )
+            )
```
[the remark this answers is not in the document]
podlesrafal: Leaving as-is — the evict-then-(re)stage ordering is intentional and the reactive composition keeps it in one readable chain.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpMapper.java
```diff
@@ -0,0 +1,64 @@
...
+                    .targetPdpId(wire.getTargetPdpId())
+                    .environmentId(wire.getEnvironmentId())
+                    .tag(wire.getTag())
+                    .syncAction(syncAction)
+                    .build();
+            } catch (Exception e) {
+                log.error("Unable to extract AuthzPdp from {} event [{}]", syncAction, event.getId(), e);
+                return null;
+            }
+        });
```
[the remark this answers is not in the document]
podlesrafal: Leaving as-is — the parse/validate/log/build step is local and only used here; extracting it would not improve testability.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
remibaptistegio: Why not keep the same behavior and instead document that if you want to avoid having all pdp engines on the same gateway, we need to tag every gateways ? Just like we do for APIs.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
remibaptistegio: Ok I'd say for now at least it avoids the possible drift but maybe we need to take some time later to see if we could do better.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
podlesrafal: Done

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
podlesrafal: Implemented the fold

remibaptistegio (approved):

sclarkamlab (approved):

PR #17906, opened 2026-06-16
feat(api): add portalNavigation field to API definition entity

konrad-krol-incubly:
## Issue

https://gravitee.atlassian.net/browse/EXT-134

## Description

Adds an optional `portalNavigation` field to the API definition entity (Automation API). It declares the API's own internal doc-folder tree — persisted on the API record so a later slice can consume it during portal navigation tree sync.
  
  - Add `portalNavigation: List<NavigationPath>` to `ApiCRDSpec` and core `Api`; wire through `toApiBuilder()` (kept **out** of the gateway definition).
  - Extend `core.NavigationPath` with an optional `order`.
  - Persist via a new `portal_navigation` JSON column on `apis` (Liquibase + JDBC + Mongo).
  - `ApiAdapter` serialises/deserialises the field; `ImportApiCRDUseCase.update` carries it over.
  - `ApiCRDExportDomainServiceImpl` populates it on GET so PUT/GET round-trips.
  - Thread the field through the v2 REST ↔ automation mapper chain.

  ## Out of scope

  - Rendering / nav-tree materialisation — next slice (API documentation)

PR #17913, opened 2026-06-16
feat(portal): persist portalNavigation on Portal entity and gate sync…

konrad-krol-incubly:
## Issue

## Description

Persist the navigation array on the Portal entity instead of rebuilding it from the folder tree. Sync planner now only deletes folders it previously created.

What changes

  - Add a portalNavigation JSON column on the portals table.
  - PUT `/portals` writes the array to the Portal row; GET returns it as-is
  - Sync planner takes both the previously persisted and the new list, and only deletes folders that were in the previous one.

  Why it's better

  - Aligned with the API definition changes (PR #17906) which also keep a portalNavigation property on the entity.
  - Idempotency is simpler — no path reconstruction from the folder tree.
  - Folder delete is safer — only removes folders the Portal resource managed, not ones created through webconsole.


konrad-krol-incubly:
- [x] add mergify label

dev-marek (approved):
lgtm

PR #17914, opened 2026-06-16
feat(observability): add single log detail endpoint GET /observability/logs/{requestId}

SamHardi:
## Summary

Add a new Gamma endpoint `GET /observability/logs/{requestId}?apiId=` that merges enriched metadata from the `v4-metrics` index with HTTP payloads (headers + body) from the `v4-log` index into a single response. This allows the frontend to load the full log record in a single lazy call instead of the current dual-fetch workaround.

Closes [GMA-424](https://gravitee.atlassian.net/browse/GMA-424)

## Changes

- **Core layer**: new `LogDetail` and `HttpPayload` domain models, `GetObservabilityLogDetailUseCase`, `LogDetailNotFoundException` and `MissingLogScopeException` exceptions
- **Port**: added `getLogDetail` method to `ObservabilityLogsDataPort`
- **Infra**: implemented `getLogDetail` in `ObservabilityLogsDataPortAdapter` with plan/application/gateway name enrichment; updated `GammaLogsConfiguration` to inject `AnalyticsQueryService`
- **REST**: `GET /{requestId}?apiId=` endpoint on `LogsResource` with `API_LOG[READ]` permission check and 404 collapse (permission denied is indistinguishable from not found)
- **OpenAPI**: new endpoint definition + `LogDetail` and `HttpPayload` schemas
- **Tests**: use case unit tests, adapter constructor update, resource tests (200 merged detail, 400 missing apiId, 404 not found, 404 permission denied, input forwarding)

## Backend / API changes

|----------|--------|--------|-----------------|------------------|
| `/gamma/.../observability/logs/{requestId}` | `GET` | **New** — single merged log detail | `?apiId=67e246cb-...` | `{ "requestId": "req-123", "status": 200, "planName": "Gold", "entrypointRequest": { "method": "GET", ... } }` |

## Test plan

- [ ] `GET /observability/logs/{requestId}?apiId=<valid>` returns 200 with merged metadata + HTTP payloads
- [ ] `GET /observability/logs/{requestId}` (no apiId) returns 400
- [ ] `GET /observability/logs/{requestId}?apiId=<no-permission>` returns 404 (not 403)
- [ ] `GET /observability/logs/{requestId}?apiId=<valid>` with non-existent requestId returns 404
- [ ] Partial data: when v4-log is empty but v4-metrics exists, returns 200 with metadata only (payload fields omitted)
- [ ] Partial data: when v4-metrics is empty but v4-log exists, returns 200 with payloads only (metadata fields omitted)
- [ ] Enrichment: plan name, application name, gateway hostname/IP are resolved from their respective services

[GMA-424]: https://gravitee.atlassian.net/browse/GMA-424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
@@ -15,6 +15,7 @@
  */
 package io.gravitee.gamma.rest.infra.adapter;
 
+import io.gravitee.apim.core.analytics.query_service.AnalyticsQueryService;
 import io.gravitee.apim.core.api_product.model.ApiProduct;
 import io.gravitee.apim.core.api_product.query_service.ApiProductQueryService;
 import io.gravitee.apim.core.application.crud_service.ApplicationCrudService;
@@ -34,6 +35,8 @@ import io.gravitee.gamma.rest.core.observability.filter.model.FilterOperator;
 import io.gravitee.gamma.rest.core.observability.filter.model.FilterSpec;
 import io.gravitee.gamma.rest.core.observability.filter.model.StaticFilters;
 import io.gravitee.gamma.rest.core.observability.logs.model.ApiReference;
+import io.gravitee.gamma.rest.core.observability.logs.model.HttpPayload;
+import io.gravitee.gamma.rest.core.observability.logs.model.LogDetail;
 import io.gravitee.gamma.rest.core.observability.logs.model.LogEntry;
 import io.gravitee.gamma.rest.core.observability.logs.model.LogEntryWarning;
 import io.gravitee.gamma.rest.core.observability.logs.model.LogsPage;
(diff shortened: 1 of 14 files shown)


gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
```diff
@@ -124,6 +134,111 @@ public LogsPage searchLogs(String organizationId, String environmentId, LogsSear
...
+                var plan = planCrudService.getById(metrics.getPlanId());
+                builder.planName(plan.getName());
+            } catch (PlanNotFoundException | TechnicalManagementException ignored) {}
+        }
+
+        if (metrics.getApplicationId() != null) {
+            try {
+                var app = applicationCrudService.findById(metrics.getApplicationId(), executionContext.getEnvironmentId());
+                builder.applicationName(app.getName());
+            } catch (ApplicationNotFoundException | TechnicalManagementException ignored) {}
```
a-cordier: maybe a debug log would be nice ? (that's a general comment for the whole class)

a-cordier (approved):

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
SamHardi: Fixed in 99eacfc3b0. Added `@CustomLog` to the adapter and replaced the 3 empty catch blocks in `enrichDetailWithNames` with debug-level logs that include the failing entity id and the exception. The existing `enrichWithNames` (search flow) uses bulk lookups that don't throw, so no empty catches there.

PR #17932, opened 2026-06-17
feat(portal-automation): restrict CRUD to default-portal behind allowMultiplePortalPerEnv flag

konrad-krol-incubly:
## Issue

## Description

- Adds `automation.portal.allowMultiplePortalPerEnv` config flag (default `false`) to the Automation API
- Rejects portal / portal-documentation / portal-listing PUTs that target a non-`default-portal` HRID with HTTP 400 when the flag is off
- Even when the flag is on, skips nav-tree materialization (folder sync + documentation materialization) for non-default portals — portal navigation tree currently managed through webconsole is not multiportal ready (its future scope)
- Integrate the flag into `gravitee.yml` and the Helm chart (`values.yaml` + `api-configmap.yaml`)

## Impact

-  GKO integration tests adds randomized suffix => it will be rejected because of "default-portal" restriction => GKO integration tests to be adjusted.
- GKO uses namespace+name as hrid for portal resource, it will fail for resources outside "default" namespace because there is no way to construct "default-portal" resource in this case, this is the code fragment:
```api/model/refs/namespace.go:77:
  func (n *NamespacedName) HRID() string {
      return n.Namespace + "-" + n.Name
  }
```
=> for point 2 we can probably make the restriction as following: "restrict to have only one portal per environment", instead of "restricting by name/hrid/etc to match default-portal". However, if someone sets flag to `allowMultiplePortalPerEnv=true`, then we are not able to gate navigation item materialization and make it conditional (do it only for single portal and ignore for others).


konrad-krol-incubly:
- [x] add mergify label

helm/templates/api/api-configmap.yaml
```diff
@@ -634,6 +634,14 @@ data:
       {{- end }}
     {{- end }}
 
+    {{- if and .Values.api.automation .Values.api.automation.portal }}
+    automation:
+      portal:
+        {{- if hasKey .Values.api.automation.portal "allowMultiplePortalPerEnv" }}
+        allowMultiplePortalPerEnv: {{ .Values.api.automation.portal.allowMultiplePortalPerEnv }}
+        {{- end }}
+    {{- end }}
```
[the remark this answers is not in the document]
konrad-krol-incubly: fixed

dev-marek (approved):
lgtm

konrad-krol-incubly:
- [x] TODO: confirm both points from Impact section

benoitgravitee (approved):

benoitgravitee:
@konrad-krol-incubly it is indeed wiser to check the functional constraint that is: "no more than 1 portal per env". Indeed GKO users have no control over how HRID is created, would be a strange HRID to force Terraform user to use a given hrid as well.

benoitgravitee (changes requested):
see above

benoitgravitee:
> GKO integration tests adds randomized suffix => it will be rejected because of "default-portal" restriction => GKO integration tests to be adjusted.

@michal-incubly you need to set the new Helm Chart param to `true` in `hack/kind/apim/values.yaml` for tests to pass after this PR is merged

benoitgravitee (approved):
As agreed I approve and leave comment

helm/values.yaml
```diff
@@ -952,6 +952,12 @@ api:
   # Specify the visibility duration of a gateway in Unknown State (in seconds)
   gateway:
     unknownExpireAfter: 604800
+  automation:
```
benoitgravitee: remove this block

gravitee-apim-rest-api/gravitee-apim-rest-api-standalone/gravitee-apim-rest-api-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -686,6 +686,14 @@ documentation:
   audit:
     max-content-size: -1 # Max size of content in bytes to be stored in audit logs when importing an API. Default is -1 meaning their is no limit.
 
+# Automation API capability gates.
```
benoitgravitee: remove this block

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalAutomationScopeEnforcer.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public boolean isDefaultPortal(AuditInfo auditInfo, PortalId portalId) {
+        return defaultPortalId(auditInfo).equals(portalId);
+    }
+
+    public List<Validator.Error> validate(AuditInfo auditInfo, PortalId portalId, String fieldName) {
+        if (multiplePortalsAllowed) {
+            return List.of();
+        }
+        if (isDefaultPortal(auditInfo, portalId)) {
```
benoitgravitee: As agreed count, if == 1 and incoming id are not equal to db id => error

benoitgravitee (approved):

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/17_add_portal_navigation_to_portals.yml
```diff
@@ -1,6 +1,6 @@
 databaseChangeLog:
```
graviteeio, line 1: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

dev-marek: not yet on master

graviteeio, line 1: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

graviteeio, line 1: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/EnvironmentsDefaultPortalUpgrader.java
konrad-krol-incubly, line 1: When we get rid of that, do we still need `CreateDefaultPortalUseCase` class? Its used from `EnvironmentCommandHandler` and this still going to be problematic because you will instantiate default-portal for every environment created.

I think we should defer default portal instance creation until APIM fully supports multiportal behavior. Then we can create such instance and binds all existing navigation items to default portal instance.

Till that moment I don't see any reason to keep this use case

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/17_add_portal_navigation_to_portals.yml
graviteeio, line 1: Changing this file might cause some start up issues because of a failing checksum validation during liquibase execution.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/use_case/CreateDefaultPortalUseCase.java
```diff
@@ -27,7 +27,7 @@
 @RequiredArgsConstructor
 public class CreateDefaultPortalUseCase {
 
-    static final String DEFAULT_PORTAL_HRID = "default-portal";
+    public static final String DEFAULT_PORTAL_HRID = "default-portal";
```
konrad-krol-incubly: nit: can this go back to package-private? no external usages

helm/templates/api/api-configmap.yaml
```diff
@@ -634,6 +634,16 @@ data:
...
 
+    {{- if .Values.api.automation }}
+    {{- if .Values.api.automation.portal }}
+    automation:
+      portal:
+        {{- if hasKey .Values.api.automation.portal "allowMultiplePortalPerEnv" }}
+        allowMultiplePortalPerEnv: {{ .Values.api.automation.portal.allowMultiplePortalPerEnv }}
+        {{- end }}
+    {{- end }}
+    {{- end }}
```
konrad-krol-incubly: please remove this — flag is gone from Java config and `values.yaml`, this block is orphan

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/17_add_portal_navigation_to_portals.yml
```diff
@@ -1,6 +1,6 @@
...
   - changeSet:
-      id: 4.12.0_18_add_portal_navigation_to_portals
+      id: 4.12.0_17_add_portal_navigation_to_portals
```
konrad-krol-incubly: Why changing that?

You cannot change this, its already in master and in 4.12.x branch:
https://github.com/gravitee-io/gravitee-api-management/tree/master/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0

This wasn't new migration added.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/EnvironmentsDefaultPortalUpgrader.java
benoitgravitee: That is what we agreed on. No default portal since it is not required for navigation items/docs.
One should be created when we upgrade to the multi-portal version (4.??) if none is present for the env. If one is present (created by automation API), then the upgrader we skip creating the default one.

helm/templates/api/api-configmap.yaml
benoitgravitee: No!!!
Do not remove. 
We need it to be in the config map!
But should change it a bit.
```suggestion
    {{- if hasKey .Values.api "automation" }}
    {{- if hasKey .Values.api.automation "portal" }}
    {{- if hasKey .Values.api.automation.portal "allowMultiplePortalPerEnv" }}
    automation:
      portal:
        allowMultiplePortalPerEnv: {{ .Values.api.automation.portal.allowMultiplePortalPerEnv }}
    {{- end }}
    {{- end }}
    {{- end }}
```

helm/tests/api/configmap_automation_test.yaml
```diff
@@ -0,0 +1,40 @@
+suite: Test Management API configmap for Automation API capability gates
```
benoitgravitee, line 1: Theflag exists, it is just no longer _**documented**_.
values.yaml and gravitee.yml don't have it, but the code reads `automation.portal.allowMultiplePortalPerEnv`

PR #17842, opened 2026-06-15
feat(api-product): persist and validate API Product sharding tags

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-14341
https://gravitee.atlassian.net/browse/APIM-14342

## Description
Backend foundation for API Product sharding tags (PR 1 of 2). Persists product tags, validates plan tags, includes tags in deploy payload, and handles org tag delete cascade. Gateway routing changes are in a follow-up PR.
| Area | Change |
|------|--------|
| **Repository** | `api_product_tags` table (JDBC Liquibase), Mongo `tags` field, enrich on read |
| **Management API** | `tags` on create/update/deploy models; OpenAPI + REST validation |
| **Plan validation** | Plan tags must be ⊆ product tags; reject plan tags on tagless product |
| **Deploy state** | `NEED_REDEPLOY` when product `tags` or `apiIds` differ from last deploy |
| **Product update** | Auto-strip orphaned plan tags when product tags shrink; no auto-deploy on update |
| **Org tag delete** | `ApiProductTagDomainService` removes tag key from all products + plans |
| **Deploy payload** | `DeployApiProductDomainService` includes product `tags` in event |

The testing sheet with screenshots is present in the JIRA tickets mentioned.

PR #17943, opened 2026-06-17
feat(gateway): apply API Product sharding tags at runtime

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-14341
https://gravitee.atlassian.net/browse/APIM-14342

**Description**
Apply API Product sharding tags at gateway runtime, using the same tag-matching rules as APIs (GatewayConfiguration.hasMatchingTags).

**Depends on**: the persistence/validation PR that stores product tags and includes them in the deploy event payload. This PR only consumes tags from the sync payload; without that PR, tags will be empty and sharding will not take effect end-to-end.

**What changes**
ApiProductShardingFilter — central product/plan tag matching (plan with no tags always passes once the product matches).

ApiProductManagerImpl — skip/undeploy products whose tags do not match the gateway; on update, emit events for the union of previous and current member API IDs.

ApiProductRegistryImpl — pre-built reverse index (environmentId, apiId) → plan entries, filtered by product and plan tags; copy-on-write updates guarded by ReentrantReadWriteLock.

ApiManagerImpl — isApiShardEligible() falls back to the product registry when the API’s own tags do not match; per-API ReentrantLock for concurrent sync/product-event safety; listener undeploys member APIs when product changes remove eligibility.

Sync pipeline — ApiProductMapper maps tags from deploy payload; RepositoryApiMemberResyncTrigger re-syncs member APIs on product DEPLOY/UPDATE (fixes ordering where API sync runs before the product is indexed); 

LatestEventFetcher.fetchLatestForApiIds + ApiSynchronizer.resyncMemberApis.

Example
GW-internal (tags: internal)     GW-external (tags: external)
Product prod-abc tags={internal}, apiIds={api-pay}, plan-gold (published)
→ GW-internal: product deployed, api-pay deployed via product fallback
→ GW-external: product skipped, api-pay not deployed

Additional context
Review-only stack: can be reviewed independently of the persistence PR; merge only after that [PR](https://github.com/gravitee-io/gravitee-api-management/pull/17842).

Scope: gateway + sync + integration tests only — no DB, Management API, or Console changes.

The testing sheet with screenshots is present in the JIRA tickets mentioned.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
@@ -24,6 +24,8 @@ import io.gravitee.definition.model.DefinitionVersion;
 import io.gravitee.definition.model.Plugin;
 import io.gravitee.gateway.env.GatewayConfiguration;
 import io.gravitee.gateway.handlers.api.definition.Api;
+import io.gravitee.gateway.handlers.api.event.ApiProductChangedEvent;
+import io.gravitee.gateway.handlers.api.event.ApiProductEventType;
 import io.gravitee.gateway.handlers.api.manager.ActionOnApi;
 import io.gravitee.gateway.handlers.api.manager.ApiManager;
 import io.gravitee.gateway.handlers.api.manager.Deployer;
@@ -38,16 +40,20 @@ import io.gravitee.secrets.api.discovery.Definition;
 import io.gravitee.secrets.api.discovery.DefinitionMetadata;
 import io.gravitee.secrets.api.event.SecretDiscoveryEvent;
 import io.gravitee.secrets.api.event.SecretDiscoveryEventType;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
 import lombok.CustomLog;
 import org.slf4j.MDC;
@@ -61,10 +67,13 @@ public class ApiManagerImpl implements ApiManager {
 
     private static final int PARALLELISM = Runtime.getRuntime().availableProcessors() * 2;
 
+    private static final Set<String> SUPPORTED_SECRET_DEFINITION_KINDS = Set.of(API_V4_DEFINITION_KIND, NATIVE_API_V4_DEFINITION_KIND);
+
     private final EventManager eventManager;
     private final GatewayConfiguration gatewayConfiguration;
     private final LicenseManager licenseManager;
     private final Map<String, ReactableApi<?>> apis = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, Lock> apiLocks = new ConcurrentHashMap<>();
     private final Map<Class<? extends ReactableApi<?>>, ? extends Deployer<?>> deployers;
     private final ApiProductRegistry apiProductRegistry;
 
@@ -104,45 +113,81 @@ public class ApiManagerImpl implements ApiManager {
             )
         );
 
-        // Listen to secret discovery events to update APIs when secrets change
+        subscribeToSecretDiscoveryEvents();
+        subscribeToApiProductEventsIfNeeded();
+    }
+
+    private void subscribeToSecretDiscoveryEvents() {
         eventManager.subscribeForEvents(
             event -> {
                 if (event.content() instanceof SecretDiscoveryEvent secretDiscoveryEvent) {
-                    if (secretDiscoveryEvent.definition() instanceof Definition definition) {
-                        if (!List.of(API_V4_DEFINITION_KIND, NATIVE_API_V4_DEFINITION_KIND).contains(definition.kind())) {
-                            // We only handle API V4 and Native API definitions
-                            log.debug(
-                                "Received SecretDiscoveryEvent for definition {} with kind {}, but we only handle API V4 and Native API definitions",
-                                definition.id(),
-                                definition.kind()
-                            );
-                            return;
-                        }
-
-                        ReactableApi<?> api = apis.get(definition.id());
-                        if (api != null) {
-                            MDC.put("api", api.getId());
-
-                            log.info("Secret value changed for API {}, updating it", definition.id());
-                            eventManager.publishEvent(ReactorEvent.UPDATE, api);
-
-                            log.info("{} has been updated", api);
-                            MDC.remove("api");
-                        } else {
-                            log.trace(
-                                "Received SecretDiscoveryEvent for API {}, but it is not found in the API manager. Unable to update it",
-                                definition.id()
-                            );
-                        }
-                    }
+                    onSecretDiscoveryValueChanged(secretDiscoveryEvent);
                 }
             },
             SecretDiscoveryEventType.VALUE_CHANGED
         );
     }
 
+    private void onSecretDiscoveryValueChanged(SecretDiscoveryEvent secretDiscoveryEvent) {
+        if (!(secretDiscoveryEvent.definition() instanceof Definition definition)) {
+            return;
+        }
+        if (!SUPPORTED_SECRET_DEFINITION_KINDS.contains(definition.kind())) {
+            log.debug(
+                "Received SecretDiscoveryEvent for definition {} with kind {}, but we only handle API V4 and Native API definitions",
+                definition.id(),
+                definition.kind()
+            );
+            return;
+        }
+        updateApiOnSecretChange(definition);
+    }
+
+    private void updateApiOnSecretChange(Definition definition) {
+        ReactableApi<?> api = apis.get(definition.id());
+        if (api == null) {
+            log.trace(
+                "Received SecretDiscoveryEvent for API {}, but it is not found in the API manager. Unable to update it",
+                definition.id()
+            );
+            return;
+        }
+        MDC.put("api", api.getId());
+        try {
+            log.info("Secret value changed for API {}, updating it", definition.id());
+            eventManager.publishEvent(ReactorEvent.UPDATE, api);
+            log.info("{} has been updated", api);
+        } finally {
+            MDC.remove("api");
+        }
+    }
+
+    private void subscribeToApiProductEventsIfNeeded() {
+        if (apiProductRegistry == null) {
+            return;
+        }
+        eventManager.subscribeForEvents(
+            event -> {
+                if (event.content() instanceof ApiProductChangedEvent productEvent) {
+                    onApiProductUndeploy(productEvent);
+                }
+            },
+            ApiProductEventType.UNDEPLOY
+        );
+    }
+
+    private void onApiProductUndeploy(ApiProductChangedEvent productEvent) {
+        if (productEvent.getApiIds() == null) {
+            return;
+        }
+        reEvaluateApisAfterProductChange(productEvent.getProductId(), productEvent.getApiIds());
+    }
+
     private boolean register(ReactableApi<?> api, boolean force) {
-        // Get deployed API
+        return withPerApiLock(api.getId(), () -> doRegister(api, force));
+    }
+
+    private boolean doRegister(ReactableApi<?> api, boolean force) {
         ReactableApi<?> deployedApi = get(api.getId());
 
         List<Plugin> plugins;
@@ -165,34 +210,27 @@ public class ApiManagerImpl implements ApiManager {
             return false;
         }
 
-        // Keep the check of Sharding Tags for io.gravitee.gateway.services.localregistry.LocalApiDefinitionRegistry
-        if (gatewayConfiguration.hasMatchingTags(api.getTags())) {
+        if (isApiShardEligible(api)) {
             boolean apiToDeploy = deployedApi == null || force;
             boolean apiToUpdate =
                 !apiToDeploy &&
                 (deployedApi.getDeployedAt().before(api.getDeployedAt()) && !Objects.equals(deployedApi.getRevision(), api.getRevision()));
 
-            // if API will be deployed or updated
             if (apiToDeploy || apiToUpdate) {
                 Deployer deployer = deployers.get(api.getClass());
                 deployer.initialize(api);
             }
 
-            // API is not yet deployed, so let's do it
             if (apiToDeploy) {
                 deploy(api);
                 return true;
-            }
-            // API has to be updated, so update it
-            else if (apiToUpdate) {
+            } else if (apiToUpdate) {
                 update(api);
                 return true;
             }
         } else {
             log.debug("The API {} has been ignored because not in configured tags {}", api.getName(), api.getTags());
 
-            // Check that the API was not previously deployed with other tags
-            // In that case, we must undeploy it
             if (deployedApi != null) {
                 undeploy(api.getId());
             }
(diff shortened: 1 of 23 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
```diff
@@ -234,38 +290,45 @@ public boolean register(ReactableApi api) {
 
     @Override
     public void unregister(String apiId) {
+        apiLocks.remove(apiId);
```
araffass: unregister() methode here removes the lock before undeploy() which breaks mutual exclusion                                                                                                                                                                                                                                                                                                                                                           - apiLocks.remove(apiId) runs before undeploy(apiId). A concurrent register() that already holds a reference to the lock (via computeIfAbsent) proceeds unsynchronized with the undeploy. 
Worse, a concurrent register() arriving after the remove creates a new lock, and can re-deploy the API after unregister intended to remove it.                                                                                                                                                                                           Proposition: Acquire the lock, undeploy, then remove the lock entry, or just don't remove it at all.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
```diff
@@ -365,6 +444,17 @@ private boolean isIncludedInApiProductWithPlans(ReactableApi<?> api) {
         return apiProductPlanEntries != null && !apiProductPlanEntries.isEmpty();
     }
 
+    private void reEvaluateApisAfterProductChange(ApiProductChangedEvent productEvent) {
+        log.debug("Re-evaluating {} API(s) after API Product [{}] change", productEvent.getApiIds().size(), productEvent.getProductId());
+        for (String apiId : productEvent.getApiIds()) {
+            ReactableApi<?> deployedApi = apis.get(apiId);
+            if (deployedApi != null && !isApiShardEligible(deployedApi)) {
+                log.info("API [{}] no longer eligible after API Product [{}] change — undeploying", apiId, productEvent.getProductId());
+                undeploy(apiId);
```
araffass: here undeploy(apiId) is called without holding the per-API ReentrantLock. 
A concurrent register() (which does hold the lock) can race: apis.remove from undeploy vs apis.put from doRegister. Result is non-deterministic and API may end up deployed after an undeploy intent or inverse

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
```diff
@@ -365,6 +444,17 @@ private boolean isIncludedInApiProductWithPlans(ReactableApi<?> api) {
         return apiProductPlanEntries != null && !apiProductPlanEntries.isEmpty();
     }
 
+    private void reEvaluateApisAfterProductChange(ApiProductChangedEvent productEvent) {
```
araffass: Also, when a product is deployed/updated, only already-deployed APIs that become ineligible are undeployed. APIs that were not deployed (tag-filtered out) but now become eligible via the new product are not deployed reactively. 
The async RepositoryApiMemberResyncTrigger compensates, but creates a transient unavailability window between the synchronous undeploy and the async resync.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/api/RepositoryApiMemberResyncTrigger.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private void resyncMemberApis(String environmentId, Set<String> apiIds) {
+        if (environmentId == null || apiIds == null || apiIds.isEmpty()) {
+            return;
+        }
+        Disposable d = Completable.fromAction(() ->
+            log.debug("Scheduling resync of {} member API(s) in environment [{}]", apiIds.size(), environmentId)
+        )
+            .andThen(apiSynchronizer.resyncMemberApis(apiIds, Set.of(environmentId)))
```
araffass: the resynchMemberApus() is a regular method call, not deferred. It executes eagerly on the EventManager dispatch thread before subscribeOn(syncDeployerExecutor) takes effect. 
If resyncMemberApis internally calls eventsFetcher.fetchLatestForApiIds(...) (a blocking DB query), it stalls the event dispatch thread.                                                                                                                                              
  - Proposition: Wrap in Completable.defer(() -> apiSynchronizer.resyncMemberApis(...)).

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
```diff
@@ -104,45 +113,86 @@ public ApiManagerImpl(
...
+    private void updateApiOnSecretChange(Definition definition) {
+        ReactableApi<?> api = apis.get(definition.id());
+        if (api == null) {
+            log.trace(
+                "Received SecretDiscoveryEvent for API {}, but it is not found in the API manager. Unable to update it",
+                definition.id()
+            );
+            return;
+        }
+        MDC.put("api", api.getId());
```
araffass, line 155: The MDC.put is well set, but i cannot see any MDC.remove("api") is not in a finally block. If eventManager.publishEvent() throws, the MDC key leaks to the pooled thread, corrupting log correlation for all subsequent requests on that thread. Other methods in the same class correctly use try/finally.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/api/RepositoryApiMemberResyncTrigger.java
```diff
@@ -0,0 +1,80 @@
...
+            .subscribeOn(Schedulers.from(syncDeployerExecutor))
+            .subscribe(
+                () -> log.debug("Member API resync completed for environment [{}]", environmentId),
+                error -> log.error("Member API resync failed for environment [{}]: {}", environmentId, error.getMessage(), error)
+            );
+        disposables.add(d);
```
araffass: disposables.add method here accumulates every subscription. Completed disposables are never removed from the composite so  under sustained product activity, this is a memory leak until shutdown.
Worth flagging if you can have any improvment.

araffass (changes requested):

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
pragsty: done

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
pragsty: done

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/api/RepositoryApiMemberResyncTrigger.java
pragsty: done

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
pragsty: done

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/api/RepositoryApiMemberResyncTrigger.java
pragsty: done

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
pragsty: Thanks - agreed on the behavior. Undeploy is synchronous (API already in memory); deploy goes through async resyncMemberApis() because the product event only has IDs, not full definitions.

Below are viable paths - could you indicate which you recommend?

Option 1 — Sync resync on product DEPLOY
On product DEPLOY (and optionally UPDATE), wait for resyncMemberApis() to finish before returning — on syncDeployerExecutor, not the event thread (still use defer + subscribeOn).

Product DEPLOY event → resyncMemberApis() → blockingAwait() on worker thread → APIs deployed
Pros:
Closes the availability gap with minimal diff
Still reuses ApiSynchronizer (no duplicate deploy logic)
Scoped to sync plugin only

Cons:
Slower product deploy if gateway processing blocks completion
Worker thread blocked during DB fetch + deploy (under load, resync queue backs up)
UPDATE still async vs DEPLOY sync → two behaviors to document
Timeouts needed or product deploy can hang on DB/gateway issues
Undeploy (sync in ApiManagerImpl) + deploy (sync in trigger) can still race for same API without careful lock ordering


Option 2 — Unified coordinator
New coordinator called from one product-change handler:
ApiProductChangedEvent
    → Coordinator.evaluate(event)
        → per-api lock
        → if deployed && !eligible  → undeploy()
        → if !deployed && eligible  → syncAndRegister(apiId)
Undeploy and deploy eligibility live together; deploy still uses ApiSynchronizer under the hood.

Pros:
Single source of truth for product-driven eligibility
Symmetric mental model (evaluate → act)
Easier to test and reason about races
Can tune sync vs async per event type in one class

Cons:
Largest refactor — crosses handlers-api ↔ services-sync
More Spring wiring / new abstraction
Higher review and regression risk for this PR
Still needs executor choice (sync deploy can block workers)
Migration: remove or slim down duplicate logic in ApiManagerImpl + RepositoryApiMemberResyncTrigger

araffass: I'd go with Option 1 for this PR, and track Option 2 as a follow-up.                                                                                                                                                                                              
                                                                                                                                                                                                                                                                   This PR is already large so option 1 can close the availability gap with a much smaller diff, and it stays scoped to the sync plugin.                                                                                                                                                                                                                                                           
                                      
That said, I think Option 1 needs a few tweaks to avoid the pitfalls you mentioned:                                                                                                                                                                               

- Don't blockingAwait: instead of blocking a worker thread, sequence the work: run resyncMemberApis first, then chain reEvaluateApisAfterProductChange as a downstream step in the same reactive pipeline. That way you don't get the undeploy→redeploy bounce, and you're not holding up a thread waiting on DB + deploy.

- Acquire the per-API lock in reEvaluateApisAfterProductChange before calling undeploy. Right now it races with a concurrent register() that holds the lock, whichever writes to apis last wins, and the outcome is non-deterministic. Both options have this problem unless the lock is taken.                                                                                                                                                                                                                                 

- Wrap resyncMemberApis in Completable.defer() inside the trigger. Currently the method call evaluates eagerly on the EventManager thread before subscribeOn kicks in, so the DB fetch blocks event dispatch.
                                                                                                                                                                                                                                                                    
As for Option 2, the unified coordinator is definitely the cleaner design long term, but it crosses module boundaries and deserves its own focused review.

That's my current opinion but if you believe you could go with option 2 and test it with no regression as part of this ticket you can do so all up to you

pragsty: This is done, implemented option-1

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
```diff
@@ -104,45 +113,89 @@ public ApiManagerImpl(
...
+    private void onApiProductChanged(ApiProductChangedEvent productEvent) {
+        if (productEvent.getApiIds() == null) {
+            return;
+        }
+        reEvaluateApisAfterProductChange(productEvent);
+    }
+
     private boolean register(ReactableApi<?> api, boolean force) {
-        // Get deployed API
+        Lock lock = apiLocks.computeIfAbsent(api.getId(), k -> new ReentrantLock());
```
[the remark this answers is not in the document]
pragsty: done

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/registry/ApiProductRegistryKey.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gateway.handlers.api.registry;
+
+/**
+ * Registry key combining API Product ID and Environment ID for scoping.
+ */
```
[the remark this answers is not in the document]
pragsty: updated

araffass (approved):
LGTM - the Option 1 refinements (per-API lock in reEvaluateApisAfterProductChange, Completable.defer for resync, sequenced undeploy/redeploy) address the concurrency and availability concerns raised during review.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiProductManagerImpl.java
```diff
@@ -124,18 +147,30 @@ private Completable deploy(ReactableApiProduct apiProduct, Completable doBeforeE
...
-            if (apiProduct.getApiIds() != null && !apiProduct.getApiIds().isEmpty()) {
-                emitApiProductChangedEvent(ApiProductEventType.UPDATE, apiProduct);
-            }
-            log.info("API Product [{}] has been updated", apiProduct.getId());
-        });
+        Completable emit = Completable.fromRunnable(() -> publishApiProductUpdateEvent(apiProduct, previousApiIds));
         return (doBeforeEmit != null ? doBeforeEmit : Completable.complete()).andThen(emit);
     }
 
+    private void publishApiProductUpdateEvent(ReactableApiProduct apiProduct, Set<String> previousApiIds) {
```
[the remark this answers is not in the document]
pragsty: There is overlap, but the two methods are not identical,

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/registry/impl/ApiProductRegistryImpl.java
```diff
@@ -72,49 +82,120 @@ public void remove(String apiProductId, String environmentId) {
...
 
     @Override
     public List<ApiProductPlanEntry> getApiProductPlanEntriesForApi(String apiId, String environmentId) {
-        List<ApiProductPlanEntry> entries = new ArrayList<>();
-
         if (apiId == null || environmentId == null) {
-            return entries;
+            return List.of();
         }
+        return planEntriesByApi.getOrDefault(new ApiEnvironmentKey(environmentId, apiId), List.of());
```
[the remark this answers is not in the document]
pragsty: done

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/registry/impl/ApiProductRegistryImpl.java
```diff
@@ -72,49 +82,120 @@ public void remove(String apiProductId, String environmentId) {
...
+            log.debug(
+                "Plan [{}] of product [{}] skipped: plan tags {} do not match gateway tags",
+                plan.getId(),
+                product.getId(),
+                plan.getTags()
+            );
+            return;
+        }
+        ApiProductPlanEntry entry = new ApiProductPlanEntry(product.getId(), plan);
+        index.computeIfAbsent(new ApiEnvironmentKey(product.getEnvironmentId(), apiId), k -> new ArrayList<>()).add(entry);
```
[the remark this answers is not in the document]
pragsty: doen

PR #17954, opened 2026-06-17
feat(permissions): add EDGE_CONFIGURATION permission and EDGE_MANAGER role

ytvnr:
## Issue

https://gravitee.atlassian.net/browse/GMA-728

## Description

Adds a new `EDGE_CONFIGURATION` permission under the ENVIRONMENT scope and a new default `EDGE_MANAGER` environment role, to protect the Edge Management resources.

- New `EnvironmentPermission.EDGE_CONFIGURATION` (mask 4400) + `RolePermission.ENVIRONMENT_EDGE_CONFIGURATION`
- New default role `EDGE_MANAGER`: `EDGE_CONFIGURATION` in CRUD, plus READ on API / APPLICATION / GROUP / PLATFORM
- ADMIN (ENVIRONMENT) automatically receives `EDGE_CONFIGURATION` in CRUD via `createOrUpdateSystemRoles`
- `EdgeRolesUpgrader` creates `EDGE_MANAGER` on existing installations and refreshes ADMIN

## Test plan

- Unit tests updated/added: `RoleScopesResourceTest`, `RoleService_CreateOrUpdateSystemRolesTest`, `EdgeRolesUpgraderTest`
- Manual: `/#!/_organization/roles` → ENVIRONMENT scope shows `EDGE_CONFIGURATION`; `EDGE_MANAGER` role present with the expected matrix; ADMIN has `EDGE_CONFIGURATION` in CRUD

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/EdgeRolesUpgrader.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/EdgeRolesUpgrader.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.service.impl.upgrade.upgrader;
+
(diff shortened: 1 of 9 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/EdgeRolesUpgrader.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public boolean upgrade() throws UpgraderException {
+        return this.wrapException(() -> {
+            organizationRepository
+                .findAll()
+                .forEach(organization -> {
+                    ExecutionContext executionContext = new ExecutionContext(organization);
+                    initializeEdgeRoles(executionContext);
+                    roleService.createOrUpdateSystemRoles(executionContext, executionContext.getOrganizationId());
```
phiz71, line 52: is it needed ?

ytvnr: Good catch — it was redundant. `DefaultRolesUpgrader` (order 120, before this one) already calls `createOrUpdateSystemRoles` unconditionally for every organization on each startup, which is what grants `EDGE_CONFIGURATION` to the ADMIN system role. Removed the call here; this upgrader now only creates the `EDGE_MANAGER` role when missing. Done in 1ae26e10eb.

sclarkamlab (approved):

PR #17986, opened 2026-06-18
GMA-756: feat(permissions): add AM_CONFIGURATION and AGENT_IDENTITY permissions with GAMMA_IDENTITY_ADMIN role

sclarkamlab:
https://gravitee.atlassian.net/browse/GMA-756

Adds RBAC for the Agent Identity surface. Right now the gamma AM-config and agent-identity endpoints have no permission checks at all, so this puts two new ENVIRONMENT-scoped permissions in and protects the AM-config endpoints. Modelled on the EDGE_CONFIGURATION PR (#17954).

What's in here:
- two new env permissions, ENVIRONMENT_AM_CONFIGURATION (CRU) and ENVIRONMENT_AGENT_IDENTITY (CRUD)
- a dedicated GAMMA_IDENTITY_ADMIN default role with just those perms
- a GammaIdentityRolesUpgrader so existing installs get the role created and ADMIN refreshed with the new perms
- @Permissions on the platform module AmConnectionResource (GET=READ, PUT/test=CREATE+UPDATE)
- the AM-config UI now calls through the env-scoped path so the ENVIRONMENT check actually resolves (the old org-only path left no env in context, which would 403 everyone)
- fixtures updated (RoleScopesResourceTest + console permission.fixtures)

APIM side only. The matching @Permissions on the agent-identity endpoints lives in the gravitee-gamma-module-aim repo and is a separate follow-up, unblocked once these enums are published. The env already resolves there from the inner /environments/{envId} path so no routing change is needed in that module.

Verified: model/service/platform Java compile (JDK 21), RoleScopesResourceTest + GammaIdentityRolesUpgraderTest pass, all 348 platform UI tests pass.


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/GammaIdentityRolesUpgrader.java
```diff
@@ -0,0 +1,79 @@
...
+        this.roleService = roleService;
+        this.organizationRepository = organizationRepository;
+    }
+
+    @Override
+    public boolean upgrade() throws UpgraderException {
+        return this.wrapException(() -> {
+            organizationRepository
+                .findAll()
+                .forEach(organization -> {
```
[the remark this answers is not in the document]
sclarkamlab: @AikidoSec feedback: this is fine due to the nature of the operation, this is a role upgrader and as such this is safe to do as it follows existing patterns/process

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/permissions/EnvironmentPermission.java
```diff
@@ -52,7 +52,9 @@ public enum EnvironmentPermission implements Permission {
     CLUSTER("CLUSTER", 4000),
     API_PRODUCT("API_PRODUCT", 4100),
     AUTHORIZATION("AUTHORIZATION", 4200),
-    PORTAL("PORTAL", 4300);
+    PORTAL("PORTAL", 4300),
+    AM_CONFIGURATION("AM_CONFIGURATION", 4400),
+    AGENT_IDENTITY("AGENT_IDENTITY", 4500);
```
ytvnr: beware of the numeric values when rebasing, conflict with edge_configuration

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/UpgraderOrder.java
```diff
@@ -77,4 +77,5 @@ private UpgraderOrder() {}
     public static final int ASSIGN_DEFAULT_API_PRODUCT_ROLE_TO_GROUPS_UPGRADER = 953;
     public static final int NATIVE_API_LOG_PERMISSION_UPGRADER = 954;
     public static final int OIDC_IDP_DEFAULT_SCOPES_UPGRADER = 709;
+    public static final int GAMMA_IDENTITY_ROLES_UPGRADER = 955;
```
ytvnr: beware of conflict with EDGE

ytvnr (approved):
looks good to me

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/permissions/EnvironmentPermission.java
sclarkamlab: Cheers yeah got the conflicts resolved

PR #17991, opened 2026-06-18
ci: skip full pipeline on gamma module version bump

gcusnieux:
## What

The automated PRs that bump a `gravitee-gamma-module-*.version` property in the root `pom.xml` currently trigger the whole CI (all backend tests, frontend builds, sonar, helm...).

This happens because the change detection treats any `pom.xml` change as a global change (`shouldBuildAll`), so the per-module filtering is bypassed.

## Why

Gamma modules are external dependencies that are not built nor tested by this repository. A plain version bump of one of them doesn't need the full pipeline, so running it just wastes CI time on every alpha bump.

## How

When the only change in the root `pom.xml` is a bump of one or more `gravitee-gamma-module-*.version` properties, `pom.xml` is dropped from the list of changed files before the pipeline is generated. The full CI is no longer forced and only the always-on jobs run (fs_scan, Danger JS, validation).

The decision is based on parsing the actual diff content, not the branch name, so any other `pom.xml` change keeps the current behavior.

A unit test covers the parser (single gamma bump, multiple gamma bumps, other dependency bump, mixed change, empty diff).

### DIFF
--- a/.circleci/ci/src/index.ts
+++ b/.circleci/ci/src/index.ts
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { changedFiles, isBlank, isSupportBranchOrMaster } from './utils';
+import { changedFiles, isBlank, isRootPomOnlyGammaModuleBump, isSupportBranchOrMaster } from './utils';
 import { argv } from 'node:process';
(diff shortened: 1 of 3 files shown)


sclarkamlab (approved):

a-cordier (approved):
🎉

phiz71 (approved):

PR #18010, opened 2026-06-18
chore: bump aim module to 1.0.0-alpha.172

gcusnieux:
## Issue

https://gravitee.atlassian.net/browse/APIM-XXX

## Description

A small description of what you did in that PR.

PR #18011, opened 2026-06-18
ci: detect gamma module bump from the commit (cwd-independent)

gcusnieux:
## What

Skip the test pipeline when a PR only bumps a gamma module version, by looking at the commit itself (`CIRCLE_SHA1`) instead of the PR base.

Only the `pull_requests` action is affected. Docker image build and delivery (`release`, `full_release`, `build_docker_images`, ...) are separate actions that don't read `changedFiles`, so the bump is still built and delivered normally.

## Why

The first version diffed `pom.xml` against the PR base with a `-- pom.xml` pathspec. That pathspec is resolved from the current directory, and the generator runs from `.circleci/ci`, so the diff matched nothing and the check always returned false. The skip never fired (see #18002, #18008, #18010).

## How

A bump PR is a single commit, so `${CIRCLE_SHA1}~1..${CIRCLE_SHA1}` is exactly its change. We skip when that commit touches only `pom.xml` and every changed line is a `gravitee-gamma-module-*.version` property. No pathspec, so it works regardless of the working directory.

This also avoids a smaller issue: without an open PR (push before the PR exists, mergify backport), the base is guessed by walking up to 100 commits and can land too far back. Looking at the commit sidesteps that.

## Validation

Pushed a temp bump commit on top of this branch (also touching `.circleci`, worst case): CI ran only `fs_scan`, `generate-config` and `Run Danger JS`, no build/test/front/helm. Temp commit removed before review.

### DIFF
--- a/.circleci/ci/src/index.ts
+++ b/.circleci/ci/src/index.ts
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { changedFiles, isBlank, isRootPomOnlyGammaModuleBump, isSupportBranchOrMaster } from './utils';
+import { changedFiles, isBlank, isCommitOnlyGammaModuleBump, isSupportBranchOrMaster } from './utils';
 import { argv } from 'node:process';
 import { buildCIPipeline, CircleCIEnvironment } from './pipelines';
 import * as fs from 'fs';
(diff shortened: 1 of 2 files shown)


gcusnieux:
Closing in favor of #18172 which reverts the original skip (56cba3d). The optimization is too narrow for the gain and the mechanism is fragile (diff parsing, single-commit assumption). Going back to a clean state.

PR #18021, opened 2026-06-18
fix(debug): prefer gateway with matching sharding tag over most recently started

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14389
## Description

Root cause

DebugApiUseCase.selectTargetGateway filtered candidate gateways with EnvironmentUtils.hasMatchingTags, which treats a tagless gateway as matching every API. So both the tagless SaaS gateway and the tag-matching hybrid gateway passed the filter, and the tie-breaker (max by startedAt) picked whichever started most recently — the SaaS gateway.

Fix

Make gateway selection prefer an explicit (positive, non-exclusion) sharding-tag match, falling back to most-recently-started only when no explicit match exists:
- Added hasExplicitTagMatch(instance, apiTags) — true only when the gateway carries a non-! tag that positively matches the API's tags (tagless gateways, and gateways matching only via an exclusion !tag, are not explicit matches).
- Ordered the selection by explicit-match first, then startedAt.

Behavior is unchanged when the API has no tag, or when only one gateway matches (still most-recently-started).

Tests

DebugApiUseCaseTest.GatewaySelectionByTag:
- tagged gateway wins over a more-recently-started tagless gateway (the bug)
- no API tag → most-recently-started gateway still wins (no regression)
- multiple matching tagged gateways → most-recently-started tagged wins (tie-breaker)
- explicit tag match preferred over a gateway matching only via an exclusion !tag

All 54 tests in the class pass. Also verified end-to-end locally: with the API tagged hybrid, debug routes to the tagged gateway; with the tag removed, it routes to the tagless gateway.

## Additional context

https://github.com/user-attachments/assets/19b4b0cc-7285-428b-9266-77c9dfa22f16

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/debug/use_case/DebugApiUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/debug/use_case/DebugApiUseCase.java
@@ -52,6 +52,7 @@ import io.gravitee.rest.api.model.PlanStatus;
 import java.util.Comparator;
 import java.util.EnumMap;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import lombok.CustomLog;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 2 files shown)


ytvnr (approved):
Looks good to me

carlos-andres-osorio:
## Smoke QA — Debug Mode Gateway Tag Selection

**Date:** 2026-06-24
**Environment:** Local — management REST API + two gateway instances (ports 8082 / 8085)

### Results

| Area | Status | Notes |
| --- | --- | --- |
| Tagged gateway selected over tagless | ✅ PASS | TC-001: GW-B (`customer-internal`) selected; GW-A (tagless) skipped. |
| Exclusion-tag gateway not preferred | ✅ PASS | TC-002: GW-B (`!other-tag`) not treated as explicit match; tagless GW-A (most-recently-started) selected. |
| Untagged API falls back to most-recently-started | ✅ PASS | TC-003: No API tags; most-recently-started GW-A wins. |
| Both composite-comparator keys favour the same gateway | ✅ PASS | TC-004: GW-B (`customer-internal`, started last) selected — no false interaction between the two comparator keys. |
| Debug event serialisation — GATEWAY_ID resolves to live instance | ✅ PASS | TC-005: `properties.GATEWAY_ID` matches a live STARTED instance in the registry. |

**5 PASS / 0 FAIL / 0 INCONCLUSIVE**

### Test descriptions

- **TC-001** — Two gateways running: GW-B tagged `customer-internal` (started first), GW-A tagless (started second). API tagged `customer-internal`. Debug selects GW-B — explicit tag match wins over the more-recently-started tagless gateway.
- **TC-002** — Two gateways running: GW-B with exclusion tag `!other-tag` (started first), GW-A tagless (started second). API tagged `customer-internal`. Debug selects GW-A — the exclusion-only tag on GW-B is not treated as a positive match, so the most-recently-started gateway wins the tiebreaker.
- **TC-003** — Two gateways running: GW-A tagless (started first), GW-B tagless (started second). API has no sharding tags. Debug selects GW-B — with no explicit match possible, the most-recently-started gateway wins.
- **TC-004** — Two gateways running: GW-A tagless (started first), GW-B tagged `customer-internal` (started second). API tagged `customer-internal`. Debug selects GW-B — both composite-comparator keys (`hasExplicitTagMatch` and `startedAt`) favour the same gateway; the result is correct with no false interaction between the two criteria.
- **TC-005** — After a successful debug call, `properties.GATEWAY_ID` in the response is verified against the live instances registry. The ID resolves to a STARTED instance, confirming it is a real gateway reference.

PR #18025, opened 2026-06-18
feat(gamma-apim): add learning tour onboarding for new members

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/GMA-XXX

## Description

https://github.com/user-attachments/assets/8a838fac-b099-45e4-9e1c-da875e11826b

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/app/onboarding/CoachmarkTour.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/app/onboarding/CoachmarkTour.tsx
@@ -1,0 +1,156 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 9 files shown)


gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/app/onboarding/CoachmarkTour.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+    const handleBack = () => {
+        if (!isFirstStep) {
+            goToStep(stepIndex - 1);
+        }
+    };
+
+    const card = (
+        <div
+            role="dialog"
```
arpzilla-dev: **[Score: 45/100]** 💡 `role="dialog"` signals a blocking modal to screen readers — NVDA/JAWS may intercept virtual-cursor navigation and trap the user even though the coachmark is intentionally non-blocking. If the page should remain interactive while the tour is visible, consider `role="complementary"` or a live region (`role="status"` + `aria-live="polite"`) instead. Also, `aria-modal={false}` is the default and can be dropped to reduce noise.

gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/app/onboarding/CoachmarkTour.tsx
```diff
@@ -0,0 +1,157 @@
...
+            aria-modal={false}
+            aria-label={step.title}
+            style={{
```
arpzilla-dev, line 98: **[Score: 55/100]** 💡 Per `graphene.md`, Tailwind token-scale utilities are preferred over inline styles. All values here map cleanly — drop the `style` prop entirely and extend `className`:\n\n```tsx\nclassName="fixed bottom-6 left-6 z-[1000] w-96 max-w-[calc(100vw-48px)] rounded-xl border bg-card text-card-foreground shadow-xl"\n```\n\nMapping: `position:'fixed'`→`fixed`, `bottom:24`→`bottom-6`, `left:24`→`left-6`, `zIndex:1000`→`z-[1000]`, `width:380`→`w-96` (closest token, 384px), `maxWidth`→`max-w-[calc(100vw-48px)]`.

arpzilla-dev (approved):

PR #18032, opened 2026-06-19
feat(observability): add apiType constraint to Gamma filter values endpoint

SamHardi:
## Summary

Add an optional `apiType` query parameter to the Gamma `GET /observability/filters/{filterName}/values` endpoint, allowing callers to restrict returned values to specific API kinds (e.g. MCP, LLM). This prevents filters from surfacing irrelevant values when the UI scope is limited to a particular API type.

Closes [GMA-770](https://gravitee.atlassian.net/browse/GMA-770)

Screenshot 2026-06-19 at 08 20 36
Screenshot 2026-06-19 at 08 20 58
Screenshot 2026-06-19 at 08 20 49

## Changes

- **Gamma REST resource** (`ObservabilityFiltersResource`): accept a repeated `apiType` query param, parse and validate it (400 on unknown values)
- **Gamma use case** (`GetObservabilityFilterValuesUseCase`): filter ENUM values (e.g. `API_TYPE`) to only matching kinds; propagate the constraint to the data port for KEYWORD filters
- **Gamma data port** (`ObservabilityFilterDataPort`): add `Set<ApiType> apiTypes` to `listKeywordValues` signature
- **Gamma infra adapter** (`ObservabilityFilterDataPortAdapter`): map Gamma `ApiType` → platform `ApiType` and pass to the platform use case
- **Platform use case** (`GetFilterValuesUseCase`): new `apiTypes` field on `Input`; narrow `AnalyticsQueryContext.authorizedApiIds` via `apiIdsByType()` when set
- **OpenAPI spec**: document the new `apiType` query parameter with reference to the existing `ApiType` schema
- **Unit tests**: cover ENUM restriction, KEYWORD propagation, and REST layer validation

## Test plan

- [ ] Call `GET /gamma/.../observability/filters/API_TYPE/values?apiType=MCP&apiType=LLM` and verify only MCP and LLM are returned
- [ ] Call `GET /gamma/.../observability/filters/API/values?apiType=HTTP_PROXY` and verify only HTTP Proxy APIs are returned
- [ ] Call without `apiType` and confirm full unfiltered results (backward compatible)
- [ ] Call with an unknown `apiType` value (e.g. `NOPE`) and verify HTTP 400
- [ ] Confirm Management V2 `/environments/{envId}/analytics/filters/{filterName}/values` is unchanged

[GMA-770]: https://gravitee.atlassian.net/browse/GMA-770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetFilterValuesUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetFilterValuesUseCase.java
@@ -31,11 +31,13 @@ import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.apim.core.exception.ValidationDomainException;
 import io.gravitee.apim.core.plan.model.Plan;
 import io.gravitee.apim.core.plan.query_service.PlanQueryService;
+import io.gravitee.definition.model.v4.ApiType;
 import io.gravitee.rest.api.model.BaseApplicationEntity;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -57,6 +59,21 @@ public class GetFilterValuesUseCase {
         FilterSpec.Name.API_PRODUCT
     );
 
+    private static final Map<ApiType, String> PLATFORM_TO_WIRE_API_TYPE = Map.of(
+        ApiType.PROXY,
+        "HTTP_PROXY",
+        ApiType.LLM_PROXY,
+        "LLM",
+        ApiType.MCP_PROXY,
+        "MCP",
+        ApiType.MESSAGE,
+        "MESSAGE",
+        ApiType.NATIVE,
+        "NATIVE",
+        ApiType.EDGE,
+        "EDGE"
+    );
+
     private final AnalyticsDefinitionQueryService definitionQueryService;
     private final FilterValuesQueryService filterValuesQueryService;
     private final FilterValueNameResolver filterValueNameResolver;
@@ -65,7 +82,20 @@ public class GetFilterValuesUseCase {
     private final PlanQueryService planQueryService;
     private final ApiProductQueryService apiProductQueryService;
 
-    public record Input(AuditInfo auditInfo, String filterName, Instant from, Instant to, int page, int perPage, String query) {}
+    public record Input(
+        AuditInfo auditInfo,
+        String filterName,
+        Instant from,
+        Instant to,
+        int page,
+        int perPage,
+        String query,
+        Set<ApiType> apiTypes
+    ) {
+        public Input(AuditInfo auditInfo, String filterName, Instant from, Instant to, int page, int perPage, String query) {
+            this(auditInfo, filterName, from, to, page, perPage, query, Set.of());
+        }
+    }
 
     public record Output(FilterValuesPage valuesPage) {}
 
@@ -81,6 +111,9 @@ public class GetFilterValuesUseCase {
         var filterSpec = findFilterSpec(filterSpecName);
 
         var analyticsContext = contextLoader.load(input.auditInfo());
+        if (input.apiTypes() != null && !input.apiTypes().isEmpty()) {
+            analyticsContext = narrowByApiTypes(analyticsContext, input.apiTypes());
+        }
 
         return switch (filterSpec.type()) {
             case ENUM -> handleEnum(filterSpec, input);
@@ -99,6 +132,16 @@ public class GetFilterValuesUseCase {
         }
 
         var stream = enumValues.stream();
+
+        if (filterSpec.name() == FilterSpec.Name.API_TYPE && input.apiTypes() != null && !input.apiTypes().isEmpty()) {
+            Set<String> allowedWireValues = input
+                .apiTypes()
+                .stream()
+                .map(PLATFORM_TO_WIRE_API_TYPE::get)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
+            stream = stream.filter(allowedWireValues::contains);
+        }
 
         if (input.query() != null && !input.query().isBlank()) {
             var lowerQuery = input.query().toLowerCase();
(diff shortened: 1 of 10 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetFilterValuesUseCase.java
```diff
@@ -327,6 +370,31 @@ private FilterSpec.Name validateFilterName(String filterName) {
...
 
+    private static AnalyticsQueryContext narrowByApiTypes(AnalyticsQueryContext context, Set<ApiType> apiTypes) {
+        Set<String> allowedIds = new HashSet<>();
+        for (var type : apiTypes) {
+            var idsForType = context.apiIdsByType().get(type);
+            if (idsForType != null) {
+                allowedIds.addAll(idsForType);
+            }
+        }
+        var narrowedAuthorizedIds = context.authorizedApiIds().stream().filter(allowedIds::contains).collect(Collectors.toSet());
```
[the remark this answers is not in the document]
SamHardi: Acknowledged — skipping: `AnalyticsQueryContext` is a Java record with a compact constructor that normalizes `authorizedApiIds` to `Set.of()` when null (see line 33). The field is guaranteed non-null at this point, so the additional guard is unnecessary.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetFilterValuesUseCase.java
```diff
@@ -100,6 +133,16 @@ private Output handleEnum(FilterSpec filterSpec, Input input) {
...
+                .stream()
+                .map(PLATFORM_TO_WIRE_API_TYPE::get)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
+            stream = stream.filter(allowedWireValues::contains);
+        }
+
         if (input.query() != null && !input.query().isBlank()) {
             var lowerQuery = input.query().toLowerCase();
             stream = stream.filter(v -> v.toLowerCase().contains(lowerQuery));
```
jourdiw, line 148: What is the impact of this line from reassigning the stream value in line 143?
Does this mean that the lowerQuery will never be present in the stream if there are API types specified?

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/use_case/GetObservabilityFilterValuesUseCase.java
```diff
@@ -88,17 +95,24 @@ private static int resolvePerPage(Input input) {
...
             .findFirst()
             .orElseThrow(() -> new ObservabilityFilterNotFoundException(filterName));
     }
 
-    private static FilterValuesPage handleEnum(FilterSpec spec, Input input, int page, int perPage) {
+    private static FilterValuesPage handleEnum(FilterSpec spec, Input input, Set<ApiType> apiTypes, int page, int perPage) {
         List<FilterSpec.EnumValue> enumValues = spec.enumValues() == null ? List.of() : spec.enumValues();
+        if (!apiTypes.isEmpty() && "API_TYPE".equals(spec.name())) {
+            Set<String> allowedValues = apiTypes.stream().map(Enum::name).collect(java.util.stream.Collectors.toSet());
+            enumValues = enumValues
```
jourdiw, line 111: Same thing here, make sure that it doesn't impede the query filtering below and that they are both compatible.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/observability/filters/ObservabilityFiltersResource.java
```diff
@@ -112,11 +112,20 @@ public PaginatedResponseDto<FilterValueDto> getFilterValues(
         @QueryParam("from") Long from,
         @QueryParam("to") Long to,
         @QueryParam("page") Integer page,
-        @QueryParam("perPage") Integer perPage
+        @QueryParam("perPage") Integer perPage,
+        @QueryParam("apiType") List<String> apiTypes
```
jourdiw, line 116: Double check if it is more standard for query param lists to be plural (`apiTypes`) or if singular (`apiType`) is acceptable.

jourdiw (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetFilterValuesUseCase.java
SamHardi: Good question — the two filters are compatible and independent. They chain on the same stream sequentially:

1. **Line 136-144**: if `apiTypes` is set AND the filter is `API_TYPE`, narrow the stream to matching wire values (`HTTP_PROXY`, `LLM`, etc.)
2. **Line 146-149**: if `query` is set, further narrow by substring match

So `apiType=MCP&apiType=LLM&query=m` first restricts to `{MCP, LLM}`, then keeps only those containing 'm' → returns `MCP` and `LLM`. Both filters are additive (AND), never exclusive.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/use_case/GetObservabilityFilterValuesUseCase.java
SamHardi: Same pattern as the platform use case — they are fully compatible. Here the flow is:

1. **Line 109-114**: if `apiTypes` is set AND filter is `API_TYPE`, narrow the `enumValues` list to matching entries
2. **Line 120-125**: if `query` is set, further filter by substring on value/label

The `apiTypes` filtering produces a new `List<EnumValue>` (not a stream), and the query filtering creates a separate stream from that list. They compose cleanly — both narrow the result set (AND semantics).

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/observability/filters/ObservabilityFiltersResource.java
SamHardi: Singular `apiType` is intentional and consistent with the existing convention in this resource — the `signal` param on the filter definition endpoint (line 93) uses the same pattern: `@QueryParam("signal") List<String> signals`.

This follows the standard JAX-RS / OpenAPI convention for repeated query params: `?apiType=MCP&apiType=LLM` with a singular key name, which is the most common REST pattern (e.g. GitHub API uses `label`, not `labels`, for multi-valued params).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetFilterValuesUseCase.java
```diff
@@ -81,6 +111,9 @@ public Output execute(Input input) {
         var filterSpec = findFilterSpec(filterSpecName);
 
         var analyticsContext = contextLoader.load(input.auditInfo());
+        if (input.apiTypes() != null && !input.apiTypes().isEmpty()) {
```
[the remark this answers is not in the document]
SamHardi: Acknowledged — skipping: `contextLoader.load()` is required for **all KEYWORD filters** (it loads `authorizedApiIds`, `apiNamesById`, `apiIdsByType`), not just when `apiTypes` is present. The `apiTypes` check on line 114 only decides whether to **narrow** the already-loaded context. Moving it before the load would break every KEYWORD filter call that doesn't pass `apiTypes`.

PR #18035, opened 2026-06-19
fix(api-product): e2e api incremental growth

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-14342

## Description

Restores selective gateway sync when an API Product’s API membership (apiIds) changes, fixing the failing E2E test api-product-incremental-growth.

When API Product sharding tags were introduced, UpdateApiProductUseCase stopped auto-deploying on every update so tag changes could follow an explicit Deploy flow (save → out of sync → deploy). That also removed deploy-on-membership-change, which broke incremental growth: after adding api2/api3 via PUT, the gateway registry still only indexed api1, so product-only APIs were never deployed and returned 404.

This PR reintroduces auto-deploy only when apiIds change. Tag-only and metadata-only updates still do not trigger a deploy, preserving the sharding-tag UX on the Deployment tab.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/use_case/UpdateApiProductUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/use_case/UpdateApiProductUseCase.java
@@ -22,6 +22,7 @@ import io.gravitee.apim.core.api.domain_service.ApiStateDomainService;
 import io.gravitee.apim.core.api_product.crud_service.ApiProductCrudService;
 import io.gravitee.apim.core.api_product.domain_service.ApiProductIndexerDomainService;
 import io.gravitee.apim.core.api_product.domain_service.ApiProductTagDomainService;
+import io.gravitee.apim.core.api_product.domain_service.DeployApiProductDomainService;
 import io.gravitee.apim.core.api_product.domain_service.ValidateApiProductService;
 import io.gravitee.apim.core.api_product.exception.ApiProductNotFoundException;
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/use_case/UpdateApiProductUseCase.java
```diff
@@ -197,6 +204,21 @@ private void removeOrphanedTagsFromPlan(Plan plan, String productId, Set<String>
         planCrudService.update(plan);
     }
 
+    private void deployMembershipChangeToGateway(AuditInfo auditInfo, ApiProduct updated) {
+        try {
```
araffass, line 208: the try-catch here only catches ValidationDomainException, so any unexpected runtime exception from deploy() would still bubble up and skip the audit log + notification below. 
Rather than using the catch, i guess would it be simpler to just move deployMembershipChangeToGateway after createAuditLog and triggerNotificationDomainService? That way even if deploy fails, the housekeeping is already done.
Lemme know if that's possible otherwise all look good

araffass (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/use_case/UpdateApiProductUseCase.java
pragsty: I instead added another catch block as the sequence should be update, deploy, audit, notify.
image
Previous code had the same sequence.

PR #18038, opened 2026-06-19
fix(portal): navigation scrolling enhancements

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-14104
https://gravitee.atlassian.net/browse/APIM-14494

## Description

Two fixes for the NG Portal "Manage your navigation" tree:
  - **Auto-scroll speed**: set `cdkDropListAutoScrollStep` to 2.4 (CDK default +20%) so dragging near the edges scrolls 20% faster.
  - **Drop indentation**: add a `trackBy` keyed by `id:level` so a moved node (and its descendants) is re-rendered at the correct indentation instead of staying stale until a full page reload.

### DIFF
--- a/gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.html
+++ b/gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.html
@@ -25,6 +25,7 @@
     [trackBy]="trackByNode"
     cdkDropList
     [cdkDropListAutoScrollStep]="10"
+    [cdkDropListSortingDisabled]="true"
     (cdkDropListDropped)="onDrop($event)"
   >
     <mat-tree-node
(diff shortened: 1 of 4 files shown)


gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.ts
```diff
@@ -416,12 +420,13 @@ export class FlatTreeComponent {
...
-  private sortAndCleanTree(nodes: ProcessingNode[]): SectionNode[] {
+  private sortAndCleanTree(nodes: ProcessingNode[], level = 0): SectionNode[] {
     return nodes
       .sort((a, b) => a.__order - b.__order)
       .map(({ __order, __parentId, ...node }) => ({
         ...node,
-        children: node.children ? this.sortAndCleanTree(node.children as ProcessingNode[]) : undefined,
+        level,
+        children: node.children ? this.sortAndCleanTree(node.children as ProcessingNode[], level + 1) : undefined,
       }));
```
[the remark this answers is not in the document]
thomasballivet: done

gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.html
```diff
@@ -22,7 +22,9 @@
     [dataSource]="tree()"
     [childrenAccessor]="childrenAccessor"
     [expansionKey]="getNodeId"
+    [trackBy]="trackByNode"
     cdkDropList
+    [cdkDropListAutoScrollStep]="2.4"
```
kovaren: 20% speed-up is barely noticeable, let's make it considerably faster.

```suggestion
    [cdkDropListAutoScrollStep]="10"

kovaren (changes requested):
2 issues remain to be addressed:
- dragging an item inside a folder only works when dragging from the bottom, but not from the top (in that case the item is dropped under the folder instead of inside the folder)
- ability to cancel a drop, for example by dragging outside of the tree area (currently, when an item is dragged outside and dropped, it ends up in the position in which it was while crossing the tree boundary)

gravitee-apim-console-webui/src/portal/components/flat-tree/flat-tree.component.html
thomasballivet: done

thomasballivet:
Added a couple of commits to fix both issues. The fix of the dragging issue resolves also https://gravitee.atlassian.net/browse/APIM-14494.

See here the new behavior :


https://github.com/user-attachments/assets/d618d1f2-29ff-49cd-be56-c6862fb06810

kovaren (approved):

PR #18056, opened 2026-06-19
feat(repository): token-bucket rate-limit storage contract

wbabyte:
## What

Foundation for token-bucket (burst) rate limiting: the storage **contract**, shared by all backends.

Implements **APIM-14482** (epic **APIM-14481**). The five backend implementations build on this and are in the stacked PR (APIM-14483).

## Contents (all in `gravitee-apim-repository-api`, additive — existing `RateLimit` untouched)

- `TokenBucketRateLimitRepository<T extends TokenBucket>` — one atomic op `refillAndTryConsume(key, tokensRequested, refillRate, refillPeriodMillis, capacity, nowMillis, supplier)`.
- `TokenBucket` model — `tokens` (whole tokens), `lastRefillTime`, `subscription`.
- `TokenBucketConsumeResult` — `allowed`, `remainingTokens`, `nextAvailableAtMillis`.
- `TokenBucketCalculator` — the pure reference algorithm (refill → cap → try-consume), reused by the Java backends and re-expressed server-side by Mongo/Redis.
- `AbstractTokenBucketRateLimitRepositoryContractTest` + `InMemoryTokenBucketRateLimitRepository` — the executable spec, **published as a `test-jar`** (this module's `maven-jar-plugin` `test-jar` execution) so each backend in APIM-14483 extends it.

## Design

- The rate is **`refillRate` whole tokens per `refillPeriodMillis`** (e.g. 100 tokens / 60000 ms). Balances are **whole tokens** and all refill arithmetic is **integer** — no scaled/milli storage and no floating point, so every backend is deterministic.
- Tokens accrue in whole units: `floor(elapsed * refillRate / refillPeriodMillis)`. The refill anchor advances to `now` only once a whole token is credited, so the sub-token remainder is preserved as not-yet-elapsed time. This never over-admits; under sparse, period-misaligned traffic it can be at most one in-flight token stricter than a continuous-accrual model.
- `nowMillis` is caller-supplied (gateway clock), matching the existing rate-limit convention — keeps the contract suite deterministic across backends.
- Preconditions are validated at the boundary (`refillPeriodMillis > 0`, `tokensRequested >= 0`, `capacity >= 0` → `IllegalArgumentException`). The bucket starts full; rate/period/capacity are passed per call so config changes apply immediately.

Refs: APIM-14481, APIM-14482

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitRepository.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.ratelimit.api;
+
+import io.gravitee.repository.ratelimit.model.TokenBucket;
+import io.reactivex.rxjava3.core.Single;
+import java.util.function.Supplier;
+
+/**
+ * Storage contract for token-bucket (burst) rate limiting.
+ *
+ * <p>Separate from {@link RateLimitRepository}, whose fixed-window {@code incrementAndGet} cannot
+ * express token-bucket semantics (it always increments, even for rejected requests, and has no
+ * notion of time-based refill capped at a burst capacity).
+ *
+ * @author GraviteeSource Team
+ */
+public interface TokenBucketRateLimitRepository<T extends TokenBucket> {
+    /**
+     * Atomically refill the bucket based on elapsed time, then try to consume {@code tokensRequested}
+     * tokens. The bucket is created <em>full</em> (at {@code capacity}) on first use.
+     *
+     * @param key                the bucket key
+     * @param tokensRequested    whole tokens to consume (typically 1)
+     * @param refillRate         whole tokens added per refill period ({@code <= 0} disables refill)
+     * @param refillPeriodMillis the refill period, in milliseconds (must be {@code > 0})
+     * @param capacity           burst capacity — the maximum number of accumulated tokens
+     * @param nowMillis          caller-supplied current time, in epoch millis (the authoritative clock)
+     * @param supplier           supplies the seed bucket (key, subscription) when none exists yet
+     * @return the outcome: whether the request was allowed, remaining tokens, and when the next token is due;
+     *         or a {@code null} reference (not an empty or error {@code Single}) when rate limiting is
+     *         <em>disabled</em> (the no-op backend), which callers must treat as pass-through — mirrors
+     *         {@link RateLimitRepository#incrementAndGet}
+     * @throws IllegalArgumentException if {@code refillPeriodMillis <= 0}, {@code tokensRequested < 0} or {@code capacity < 0}
+     */
+    Single<TokenBucketConsumeResult> refillAndTryConsume(
+        String key,
+        long tokensRequested,
+        long refillRate,
+        long refillPeriodMillis,
+        long capacity,
+        long nowMillis,
+        Supplier<T> supplier
+    );
+}
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/model/TokenBucket.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/model/TokenBucket.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.ratelimit.model;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Persisted state of a single token-bucket rate-limit key.
+ *
+ * <p>Unlike {@link RateLimit} (a fixed-window counter), this models burst behaviour: tokens refill
+ * over time up to a burst capacity, and each request consumes tokens.
+ *
+ * <p>The available balance is stored in <em>whole tokens</em>. The refill rate is configured as a
+ * whole-token count per period (e.g. 100 tokens / 60s), so all refill arithmetic is integer with no
+ * floating-point drift across the heterogeneous backends.
+ *
+ * @author GraviteeSource Team
+ */
+public class TokenBucket implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private String key;
+
+    /** Available balance, in whole tokens. */
+    private long tokens;
+
+    /** Epoch millis of the last refill applied to {@link #tokens}. */
+    private long lastRefillTime;
+
+    private String subscription;
+
+    private TokenBucket() {}
+
+    public TokenBucket(String key) {
+        this.key = key;
+    }
+
+    public String getKey() {
+        return key;
+    }
+
+    public long getTokens() {
+        return tokens;
+    }
+
+    public void setTokens(long tokens) {
+        this.tokens = tokens;
+    }
+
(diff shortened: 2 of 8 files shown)


gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitRepository.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Atomically refill the bucket based on elapsed time, then try to consume {@code tokensRequested}
+     * tokens. The bucket is created <em>full</em> (at {@code capacity}) on first use.
+     *
+     * @param key                the bucket key
+     * @param tokensRequested    whole tokens to consume (typically 1)
+     * @param refillRate         whole tokens added per refill period ({@code <= 0} disables refill)
+     * @param refillPeriodMillis the refill period, in milliseconds (must be {@code > 0})
+     * @param capacity           burst capacity — the maximum number of accumulated tokens
+     * @param nowMillis          caller-supplied current time, in epoch millis (the authoritative clock)
```
paulatulis, line 41: APIM-14482 AC says "Authoritative clock is store-side (the op accepts/uses server time, not gateway now)," and the epic + APIM-14483 specify $$NOW / Redis TIME / SQL now() / Hazelcast owner-node. Here nowMillis is a caller-supplied (gateway) argument instead.
This is in tension with APIM-14483's AC that all 5 backends pass this deterministic suite: a backend genuinely using $$NOW would ignore the passed value and fail the exact-count tests (tokens_refill_over_elapsed_time, etc.). So in practice the gateway clock has to win, which reverses the store-side-clock requirement.
Net behavioural effect looks fail-closed (per-key consume is still atomic; clock skew across nodes can only under-refill, never over-admit), so I think it's defensible — but it's a real reversal of the spec. Was store-side time consciously dropped?

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/model/TokenBucket.java
```diff
@@ -0,0 +1,98 @@
...
+ * @author GraviteeSource Team
+ */
+public class TokenBucket implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private String key;
+
+    /** Available balance, in whole tokens. */
+    private long tokens;
```
paulatulis, line 40: APIM-14482 AC specifies tokens "stored as milli-tokens / scaled integer, not float." This stores whole tokens. The whole-integer rationale (determinism, no FP drift across heterogeneous backends) is sound and I'd lean toward agreeing with it — but it's a documented reversal of the AC, with the cost that sub-token rates must be expressed via a larger refillPeriodMillis and traffic can be "at most one in-flight token stricter."
Can we confirm this pivot is agreed?

gravitee-apim-repository/gravitee-apim-repository-api/src/test/java/io/gravitee/repository/ratelimit/AbstractTokenBucketRateLimitRepositoryContractTest.java
```diff
@@ -0,0 +1,264 @@
...
+ * <p>The rate is expressed as {@code refillRate} whole tokens per {@code refillPeriodMillis} (e.g.
+ * {@code (2, 1000)} = 2 tokens/s, {@code (1, 2000)} = 1 token every 2s). All timing is driven by an
+ * explicit {@code nowMillis} argument rather than a wall clock, so the contract is fully deterministic
+ * across the in-memory reference and the persistent backends. Backends provide their implementation by
+ * overriding {@link #createRepository()}.
+ *
+ * @author GraviteeSource Team
+ */
+@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
+public abstract class AbstractTokenBucketRateLimitRepositoryContractTest {
```
paulatulis, line 55: APIM-14482 AC lists "entry TTL ≥ capacity÷rate" among what this suite must assert, but there's no TTL test here. TokenBucketCalculator.ttlMillis(...) implements the rule, but it isn't part of the repository interface or the result record, so nothing in the contract forces a backend to honour it.
TTL is admittedly hard to assert against an in-memory map, but as written this AC isn't met. Can we either add a contract hook for it, or explicitly note that TTL conformance is verified per-backend in APIM-14483 so it doesn't fall through the gap between the two stories?

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitRepository.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param supplier           supplies the seed bucket (key, subscription) when none exists yet
+     * @return the outcome: whether the request was allowed, remaining tokens, and when the next token is due;
```
paulatulis, line 43: An RxJava3 Single can't emit null, so this must mean the method returns a null Single reference — which NPEs any caller doing repo.refillAndTryConsume(...).map(...) without a null guard. It does mirror the existing RateLimitRepository convention, so I'm fine keeping it for consistency, but could we (a) confirm the gateway call site null-checks, and (b) tweak the wording to "returns a null reference" rather than "returns null"?

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitRepository.java
wbabyte: Yes — deliberate reversal, and the determinism point you land on is exactly the reason.

The whole purpose of the `nowMillis` parameter is to let a *single* contract suite bind all five backends to **exact** counts (`tokens_refill_over_elapsed_time`, `each_token_takes_a_full_refill_period`, …). A backend genuinely using `$$NOW` / Redis `TIME` / SQL `now()` would ignore the passed value and those exact-count assertions become untestable — you're reduced to "roughly N" with sleeps. So the design makes the **gateway-supplied clock authoritative** and has each backend re-express the *integer math* server-side, not the *time source*.

The safety property is the one you noted: per-key consume stays atomic, and the refill anchor is forward-only (`out_of_order_timestamps_do_not_inflate_refill` covers it), so cross-node skew can only **under-refill** — fail-closed, never over-admit.

I agree it contradicts the AC as literally written, so I'd rather fix the AC than the code: I've updated APIM-14482 to record that the authoritative clock is the gateway-supplied `nowMillis` and that backends re-express the math, not the clock — flagging it to the epic rather than diverging silently.

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/model/TokenBucket.java
wbabyte: Deliberate, same determinism motivation. The AC's "scaled integer, not float" exists to avoid FP drift across heterogeneous stores; whole-token integer math achieves that *more* simply — there's no scale factor to replicate identically across Mongo's aggregation pipeline, Redis Lua, SQL and Java, and no rounding-rule divergence to chase.

The documented costs are real and accepted: sub-token-per-second rates are expressed by widening `refillPeriodMillis` (1 token / 2000 ms rather than 0.5/s), and the forward-only anchor is at most one in-flight token *stricter* — never looser. The epic's actual targets (3 r/s burst 300, 10 r/s burst 1000) are all whole-token, so no real use case regresses.

So: yes to the pivot. It needs the same kind of AC amendment as the clock thread — I don't want it to read as undocumented drift, so I'd like explicit sign-off before we lock APIM-14483 to it.

gravitee-apim-repository/gravitee-apim-repository-api/src/test/java/io/gravitee/repository/ratelimit/AbstractTokenBucketRateLimitRepositoryContractTest.java
wbabyte: You're right, and the code confirms it: `TokenBucketCalculator.ttlMillis(...)` currently has zero callers and is on neither the interface nor the result record, so nothing forces a backend to honour it — and the AC line is not met by 14482 alone.

TTL is intrinsically storage-native (Mongo TTL index, Redis `PEXPIRE`, a JDBC reaper, Hazelcast entry TTL); asserting it against an in-memory map would need a fake-clock eviction model that resembles none of them. So rather than add a hollow hook, I've made the handoff explicit: `ttlMillis` stays as the **shared rule** (so every backend computes the same expiry) and *conformance* is verified per-backend in APIM-14483.

I've added a note to that effect in this suite's class javadoc so it can't fall through the gap between the two stories. If you'd prefer an enforced seam here instead (e.g. an optional `expireAt(key)` probe backends implement), say the word — but I think the per-backend test in 14483 is the honest place for it.

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitRepository.java
wbabyte: Good catches on both.

(b) Reworded to "returns a `null` reference (not an empty or error `Single`)" — an RxJava `Single` can't emit null, so the precise statement is that the method hands back a null reference.

(a) The null is only ever produced by the no-op backend (rate limiting disabled), which lands in APIM-14483, not this PR. The consuming call site is `DefaultTokenBucketRateLimitService` in the `gravitee-gateway-services-ratelimit` plugin (the async bridge), which must null-guard before subscribing and treat null as pass-through — exactly as the existing `RateLimitService` / `incrementAndGet` path already does. I'll confirm that guard (and add a regression test) when the no-op backend + policy wiring land; for this contract PR the deliverable is the convention plus the corrected wording.

paulatulis (approved):

PR #18058, opened 2026-06-19
feat: add entity resolver interface and default implementation

a-cordier:
see https://gravitee.atlassian.net/browse/GMA-786
from https://gravitee.atlassian.net/browse/GMA-735
depends on https://github.com/gravitee-io/gravitee-gateway-api/pull/355

Sets the foundation for internal resolution of protocol facing identifier to gamma entity ids.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/DefaultEntityResolver.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/DefaultEntityResolver.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/DefaultEntityResolver.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Default implementation of {@link EntityResolver} backed by a nested concurrent map
+ * keyed by entity kind, then by name. Populated at reactor startup.
+ */
+public class DefaultEntityResolver implements EntityResolver {
+
+    private final Map<String, Map<String, String>> registry = new ConcurrentHashMap<>();
```
michel-barret: In general case ConcurrentMap of ConcurrentMap isn’t very nice, it’s better to use a composite key. Here you can create a record `Key(String key, String name)` not exposed in API.

michel-barret (approved):

PR #18082, opened 2026-06-20
feat(repository): token-bucket async-mode service facade

wbabyte:
## What

Adds the `TokenBucketRateLimitService` SPI interface to `gravitee-apim-repository-api`, mirroring the existing `RateLimitService`. It is the gateway-facing facade that selects strict vs async token-bucket enforcement via a `boolean async` flag.

This is the foundation for APIM-14499 (async / non-strict token-bucket mode, off by default). Both the policy and the gateway rate-limit service compile against this shared interface, exactly as they do for `RateLimitService`.

## Why a new interface

The token-bucket policy currently looks up `TokenBucketRateLimitRepository` directly. To support an `async` flag (selecting a local-then-reconcile repository vs the strict store) the policy must go through a facade, just like `rate-limit` goes through `RateLimitService`. The interface lives in `repository-api` so both plugins can reference it without a cycle.

## Scope

Interface only. The default implementation, the async repository, the wiring, and the policy `async` flag land in follow-up PRs.

Part of APIM-14499. Stacked on #18053 (APIM-14483 backends).

## How to test

Compile only — `mvn -pl gravitee-apim-repository/gravitee-apim-repository-api compile` (BUILD SUCCESS).

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitService.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitService.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.ratelimit.api;
+
+import io.gravitee.repository.ratelimit.model.TokenBucket;
+import io.reactivex.rxjava3.core.Single;
+import java.util.function.Supplier;
+
+/**
+ * Gateway-facing facade over the token-bucket store, mirroring {@link RateLimitService}. The
+ * {@code async} flag selects the enforcement mode for a single call:
+ *
+ * <ul>
+ *   <li>{@code async = false} (default): strict, exact. The call goes straight to the backing
+ *       {@link TokenBucketRateLimitRepository}, which atomically refills and consumes per request.</li>
+ *   <li>{@code async = true}: non-strict, eventually consistent. The node keeps a local bucket and
+ *       reconciles it to the store periodically, trading exactness for far fewer store round-trips
+ *       and higher throughput. The distributed bucket is approximate, so a backend may receive more
+ *       than the configured rate.</li>
+ * </ul>
+ *
+ * <p>The {@link TokenBucketConsumeResult} returned is the same regardless of mode, so a policy does
+ * not need to know which mode is in effect.
+ *
+ * @author GraviteeSource Team
+ */
+public interface TokenBucketRateLimitService {
+    /**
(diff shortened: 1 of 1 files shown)


gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitService.java
```diff
@@ -0,0 +1,71 @@
...
+ * <p>The {@link TokenBucketConsumeResult} returned is the same regardless of mode, so a policy does
+ * not need to know which mode is in effect.
+ *
+ * @author GraviteeSource Team
+ */
+public interface TokenBucketRateLimitService {
+    /**
+     * Refill and try to consume {@code tokensRequested} tokens for {@code key}, in the mode selected
+     * by {@code async}. Parameters carry the same meaning as
+     * {@link TokenBucketRateLimitRepository#refillAndTryConsume}.
```
paulatulis, line 44: The underlyingt okenBucketRateLimitRepository.refillAndTryConsume documents that it returns null when rate limiting is disabled (the NoOp backend), and callers must treat that as pass-through. This facade's javadoc says "the TokenBucketConsumeResult returned is the same regardless of mode" but doesn't carry over the disabled→null contract. Since the future default impl will delegate to the repo, the policy calling this facade still needs to null-check. Recommend restating the null/disabled passthrough here so the contract isn't lost at the facade layer (same footgun raised on #18056).

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitService.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param async {@code true} for the non-strict local-then-reconcile path, {@code false} for the
+     *              strict per-request path
+     */
+    Single<TokenBucketConsumeResult> refillAndTryConsume(
+        String key,
+        long tokensRequested,
+        long refillRate,
+        long refillPeriodMillis,
+        long capacity,
+        long nowMillis,
```
paulatulis, line 60: Threading the gateway-supplied nowMillis through the facade is consistent with the rest of the stack (the clock decision from #18056/#18053). Just flagging for the follow-up async-impl PR: in async mode the local bucket accrues between reconciles, so it's worth being explicit there about whether nowMillis drives the local accrual or only the reconcile — not actionable in this interface-only PR, but the semantics deserve a doc line when the impl lands.

paulatulis (approved):
nits but no blockers

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitService.java
wbabyte: Good catch — restated in `c4662dffd6`. The abstract `refillAndTryConsume` now carries an `@return` mirroring the repository contract: a `null` reference (not an empty/error `Single`) when rate limiting is disabled (no-op backend), which callers must treat as pass-through, and notes the default impl delegates to `TokenBucketRateLimitRepository#refillAndTryConsume` so the disabled→null contract is unchanged through the facade. The class-level "same result regardless of mode" line still holds — that is about strict vs async; the disabled→null case is orthogonal and now documented on the method.

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/ratelimit/api/TokenBucketRateLimitService.java
wbabyte: Agreed, and noted for the async-impl PR. When the local-then-reconcile path lands I will document explicitly that `nowMillis` drives the local bucket accrual (so a node accrues between reconciles on the gateway clock) while reconciliation against the store carries its own timestamp — keeping it out of this interface-only PR as you suggest.

PR #18092, opened 2026-06-20
fix(gamma): correct home metrics, dark-mode badge, and api product nav

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/GMA-690

## Description

- platform: use graphene success/secondary status badge so the application "active" state is visible in dark mode instead of the low-contrast bg-primary chip
- console home: scope the api management card count to v4 http proxy apis (matching the apim module) and drop the env-wide requests/24h metric
- apim: resolve the active nav key from the module-relative, outermost segment so an api product's apis tab highlights "api products" (not "api proxies")


arpzilla-dev (approved):
LGTM

PR #18107, opened 2026-06-22
fix(gamma): console ui issues — tasks & approvals navigation and consumer search

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/GMA-801

## Description

Two gamma console UI fixes:

1. **Tasks & Approvals navigation** (`gamma-console`) — subscription and review tasks only deep-linked into APIM; LLM, MCP, AI Agent and Kafka tasks fell back to the module root, so approvers landed on the wrong page. Replaced the APIM-only path logic with a single `apiType` registry that resolves the deepest valid target (consumer page → API page → module root). LLM opens the llm-router consumer page, MCP opens its consumers list, AI Agent opens the agent-runtime API page; APIM unchanged. Also unified the area/module/route maps, relabeled `llm-proxy` as its own "LLM" area, restored the APIM deep link for legacy tasks missing `apiType`, and stopped showing a "cannot be opened yet" error on tasks with no destination (e.g. user registration).

2. **Consumer search icon overlap** (`gamma-apim`) — the application search on the API and API Product consumer pages hand-rolled the leading icon, so it overlapped the input text. Switched to Graphene's `InputGroup` / `InputGroupAddon` / `InputGroupInput` primitives, which reserve space for the icon.

## Additional context

- Two commits: one for tasks & approvals, one for the consumer search fix.
- Test suites updated and passing for both areas (gamma-console + gamma-apim).
- Follow-ups (not in this PR): flip MCP consumer level to `detail` once the deployed proxy exposes a per-subscription page; add a consumer page for AI Agent (`agent-runtime`); define the Kafka (`esm`) URL convention; and a durable SDK deep-link contract so the host stops hardcoding module URL shapes.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/gravitee-gamma/gravitee-gamma-control-plane-webui/src/pages/tasks/components/TaskRow.spec.tsx
+++ b/gravitee-gamma/gravitee-gamma-control-plane-webui/src/pages/tasks/components/TaskRow.spec.tsx
@@ -1,0 +1,85 @@
+/*
+ * Copyright (C) 2026 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


arpzilla-dev (approved):
lgtm

PR #18131, opened 2026-06-22
fix(portal): persist analytics dashboard pagination

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-14442

## Description

Persist the analytics dashboard pagination in the NG Portal.

- ~~**Page number** → stored in the URL query param, so it survives a reload and browser back/forward.~~
- ~~**Page size** → stored in `localStorage`, so it survives a reload and returning via the breadcrumb.~~
- ~~Opening a dashboard now **preserves the query params**, and the breadcrumb **carries the page back**, so returning from the detail lands on the page the user came from instead of resetting to page 1.~~

**Page number and page size are both stored in the URL query params** (`?page=…&size=…`), like the subscriptions page, so they survive a reload and are restored by the browser's native back navigation when returning from a dashboard.

## Additional context

~~https://github.com/user-attachments/assets/6d0088b0-0874-432b-bd9c-71524953bff3~~

https://github.com/user-attachments/assets/2d2fc816-f061-4fa2-bacc-a7206d5d7110

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics-details/analytics-details.component.ts
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics-details/analytics-details.component.ts
@@ -18,7 +18,6 @@ import { rxResource, takeUntilDestroyed, toSignal } from '@angular/core/rxjs-int
 import { FormControl, ReactiveFormsModule } from '@angular/forms';
 import { MatDialog } from '@angular/material/dialog';
 import { MAT_FORM_FIELD_DEFAULT_OPTIONS } from '@angular/material/form-field';
-import { ActivatedRoute } from '@angular/router';
 
 import {
   AddFilterDialogComponent,
@@ -75,10 +74,6 @@ export default class AnalyticsDetailsComponent {
   private readonly dialog = inject(MatDialog);
   private readonly injector = inject(Injector);
   private readonly destroyRef = inject(DestroyRef);
-  private readonly activatedRoute = inject(ActivatedRoute);
-
-  // Page the user came from in the list, captured on entry so the breadcrumb can return there.
-  private readonly listPageParam = this.activatedRoute.snapshot.queryParamMap.get('page');
 
   readonly filtersStore = inject(DashboardFiltersStore);
   readonly dashboardId = input.required<string>();
@@ -118,8 +113,7 @@ export default class AnalyticsDetailsComponent {
     effect(() => {
       const id = this.dashboardId();
       const name = this.dashboardName();
-      const queryParams = this.listPageParam ? { page: this.listPageParam } : undefined;
-      this.breadcrumbService.set([analyticsListBreadcrumb(true, queryParams), { id: `analytics-${id}`, label: name || id }]);
+      this.breadcrumbService.set([analyticsListBreadcrumb(true), { id: `analytics-${id}`, label: name || id }]);
     });
   }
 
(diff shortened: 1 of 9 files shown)


gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.ts
```diff
@@ -140,6 +148,35 @@ export default class AnalyticsComponent {
...
+
+  private loadPageSize(): number {
+    try {
+      return parseSizeParam(
+        localStorage.getItem(AnalyticsComponent.PAGE_SIZE_KEY),
+        AnalyticsComponent.DEFAULT_PAGE_SIZE,
+        AnalyticsComponent.MAX_PAGE_SIZE,
+      );
+    } catch (error) {
+      console.warn('Failed to read analytics page size from localStorage', error);
```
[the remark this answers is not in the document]
thomasballivet: consistent

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.ts
```diff
@@ -140,6 +148,35 @@ export default class AnalyticsComponent {
...
+      queryParams,
+      queryParamsHandling: 'merge',
+    });
+  }
+
+  private persistPageSize(size: number): void {
+    try {
+      localStorage.setItem(AnalyticsComponent.PAGE_SIZE_KEY, String(size));
+    } catch (error) {
+      console.warn('Failed to persist analytics page size to localStorage', error);
```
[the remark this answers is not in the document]
thomasballivet: consistent

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.ts
```diff
@@ -140,6 +148,35 @@ export default class AnalyticsComponent {
...
+    this.router.navigate([], {
+      relativeTo: this.activatedRoute,
+      queryParams,
+      queryParamsHandling: 'merge',
+    });
+  }
+
+  private persistPageSize(size: number): void {
+    try {
+      localStorage.setItem(AnalyticsComponent.PAGE_SIZE_KEY, String(size));
```
kovaren: We have a similar pagination persistence mechanism in the subscriptions page under `/dashboard/subscriptions`.

Over there we are using the query params both for the page size and the page number, which is arguably a more idiomatic way to handle this use case.

I'd suggest implementing the same pattern here instead of storing the page size in local storage.

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics-details/analytics-details.component.ts
```diff
@@ -74,6 +75,10 @@ export default class AnalyticsDetailsComponent {
   private readonly dialog = inject(MatDialog);
   private readonly injector = inject(Injector);
   private readonly destroyRef = inject(DestroyRef);
+  private readonly activatedRoute = inject(ActivatedRoute);
+
+  // Page the user came from in the list, captured on entry so the breadcrumb can return there.
+  private readonly listPageParam = this.activatedRoute.snapshot.queryParamMap.get('page');
```
kovaren: We should avoid storing the page number of another component in this query param.

Analytics component should handle both the page size and the number on its own.

Regarding navigating back to the same page when clicking on the breadcrumbs, I'd argue that it's not a must have here, browser's native "back navigation" is enough.

WDYT?

kovaren (changes requested):

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics.component.ts
thomasballivet: Will align the pattern with subscriptions

gravitee-apim-portal-webui-next/src/app/dashboard/analytics/analytics-details/analytics-details.component.ts
thomasballivet: Agree and that is simpler. Will do

kovaren (approved):

PR #18162, opened 2026-06-23
fix(gateway): isolate distributed sync Redis keys by cluster id

mukul-tyagi08:
## Issue

Fixes cross-cluster API corruption when multiple gateway clusters (external / internal / AOG) share one Redis for distributed sync. Each primary was writing tag-filtered API payloads to global Redis keys; last writer wins caused secondaries to deploy partial APIs and log "There is no published plan associated to this API, skipping deployment...".

## Description

Scope distributed-event reads/writes by Hazelcast clusterId (derived from cluster-name). Helm renders hazelcast.xml, derives cluster-name from <release>-<sharding_tags>, and validates that distributed sync is paired with a dedicated Hazelcast cluster per Helm release.

Documented deployment constraints for distributed sync (primary must populate Redis before secondaries join):

Fresh install: replicaCount: 1, wait for primary sync, then scale up one replica at a time
Upgrade: maxUnavailable: 0, maxSurge: 1

See helm/README.md (Gateway distributed sync) and the comment block in helm/values.yaml.

### Changes

- Add clusterId to DistributedEvent model, criteria, and Redis keys/index
- Stamp clusterId on write; filter fetches by clusterManager.clusterId()
- New RediSearch index distributed-event-search-idx-v2 with TAG field on clusterId
- Runtime validation: fail if distributed sync is enabled but clusterId is null/blank
- Helm: auto-render hazelcast.xml; cluster-name defaults to <release>-<sharding_tags> with truncation that preserves tag suffix
- Helm: fail-fast if distributedSync without gateway.cluster.type=hazelcast (validated on deployment, including externalConfig installs)
- Helm: backward-compatible gateway.distributedSync — boolean true, map without enabled, and enabled: true are all supported
- Docs: distributed sync runbook in helm/README.md and helm/values.yaml

### Upgrade / breaking notes

- Redis migration required: old keys (distributed_event:API:...) are not read
- After upgrade: delete stale distributed_event:* and distributed_sync_state:* for old clusterIds; drop old search index if present; rolling restart gateways; allow resync
- DistributedEventRepository.updateAll() signature adds clusterId parameter
- Operational model: one Helm release per sharding-tag cluster; distinct Hazelcast cluster-name per release; shared Redis is OK. Changing sharding_tags changes clusterId — manual Redis cleanup required
- Rolling upgrade: when distributed sync is enabled, use maxUnavailable: 0 / maxSurge: 1 (documented; chart default remains maxUnavailable: 1)

### Out of scope (follow-up)

- Resilience: secondary handover when Redis state is empty (architect round 2)
- Clone / full API payload to Redis with consumer-side PlanAppender

## Additional context

image

### DIFF
--- a/helm/templates/gateway/gateway-configmap.yaml
+++ b/helm/templates/gateway/gateway-configmap.yaml
@@ -1,4 +1,5 @@
 {{- if and (.Values.gateway.enabled) (or (not (include "gateway.externalConfig" .)) (.Values.gateway.logging.debug) (.Values.gateway.logback.override)) -}}
+{{- include "gateway.distributedSync.validate" . -}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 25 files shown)


jhaeyaert (changes requested):
I have 1 concern regarding this statement and would like to understand if it introduces a breaking change or not (considering it is a patch version):

> Redis migration required: old keys (distributed_event:API:...) are not read.
> After upgrade: delete stale distributed_event:* and distributed_sync_state:* for
> old clusterIds; drop old search index if present; rolling restart gateways; allow resync.

I don't get if it means the rolling update will not work immediately and requires performing an additional restart to allow proper sync?

Given an existing gateway cluster using the distributed sync. When moving to the fixed version, what happens by default if I don't change anything in my Helm values?
If I have tags or not, will the Hazelcast cluster name change, ensuring the first gateway starting is primary, properly fetch the data, and push it to a specific namespace in Redis?
Other gateways will sync from Redis.

Can you confirm?

mukul-tyagi08:
> I have 1 concern regarding this statement and would like to understand if it introduces a breaking change or not (considering it is a patch version):
> 
> > Redis migration required: old keys (distributed_event:API:...) are not read.
> > After upgrade: delete stale distributed_event:* and distributed_sync_state:* for
> > old clusterIds; drop old search index if present; rolling restart gateways; allow resync.
> 
> I don't get if it means the rolling update will not work immediately and requires performing an additional restart to allow proper sync?
> 
> Given an existing gateway cluster using the distributed sync. When moving to the fixed version, what happens by default if I don't change anything in my Helm values? If I have tags or not, will the Hazelcast cluster name change, ensuring the first gateway starting is primary, properly fetch the data, and push it to a specific namespace in Redis? Other gateways will sync from Redis.
> 
> Can you confirm?

No extra manual restart is required, but the upgrade is not transparent at the Redis layer during a rolling rollout. New gateway pods do not read old Redis keys. They recover automatically once a new-version primary syncs from MongoDB and republishes into the new key format. That normally happens within the same rolling deployment — not via a second deliberate restart.

Without Helm changes, the Hazelcast cluster name (and therefore clusterId) does not change by itself. What does change is the Redis key/index format. So this is an operational migration, not a config migration — unless you adopt the new Helm hazelcast.xml rendering.

Rolling update — step by step:

Phase 1 — mixed cluster
  Old pods: write/read OLD Redis keys (old index)
  New pods (secondary): read NEW index + clusterId filter → EMPTY until republish
  New pods (secondary): may deploy NO APIs briefly

Phase 2 — new primary elected (still during same rollout)
  New primary: syncs from MongoDB (as always)
  New primary: publishes to NEW Redis keys under same clusterId
  New secondaries: pick up events → APIs deployed

Phase 3 — old pods terminated
  Only new format in Redis
  
  No separate “restart again” step — the primary’s normal sync cycle repopulates Redis. You do not need to flush Redis before upgrade for recovery to work, provided MongoDB is reachable by the primary.

Redis cleanup (distributed_event:*, old index) is hygiene, not a prerequisite for recovery.

Bottom line: Not a “restart twice” migration. It is a non-seamless Redis handoff during rolling update, with automatic recovery via MongoDB on the primary. Without Helm changes, clusterId stays the same; only the Redis encoding changes. The multi-shard fix depends on already having separate Hazelcast cluster-names per tag cluster (or adopting the new Helm config to enforce that).

helm/templates/gateway/gateway-configmap.yaml
```diff
@@ -33,11 +38,11 @@ data:
       type: {{ .Values.gateway.cluster.type }}
       {{- if eq .Values.gateway.cluster.type "hazelcast" }}
       hazelcast:
-        config-path: {{ .Values.gateway.cluster.hazelcast.configPath }}
+        config-path: {{ dig "cluster" "hazelcast" "configPath" "${gravitee.home}/config/hazelcast.xml" .Values.gateway }}
       {{- end }}
     {{- end }}
 
-    {{- if .Values.gateway.distributedSync }}
+    {{- if include "gateway.distributedSync.enabled" . }}
```
jhaeyaert, line 41: If a previous values.yaml contains `gateway.distributedSync: true`, it will no longer work and has to move to `gateway.distributedSync.enabled: true`

Can you manage to avoid this by supporting old + new (specifically on a patch)?

jhaeyaert:
> New pods (secondary): may deploy NO APIs briefly

But as the new pods join an existing Hz cluster, there are chances that they will not become primary, and they will fetch nothing until a new pod becomes the primary... with several pods, it could take time (so not very brief)

Also, once the new pod is started as secondary, it will perform the initial sync and get no data. Then it will enter the diff sync process to fetch only updates. So, even if a new pod starts and is primary, all secondaries already started will never fetch the whole set of data. A restart is necessary.

Given this, I think we should consider changing the Hz cluster name so a new pod is 100% gauranteed to become a primary node, fetch all the data, and push to Redis for the secondary to sync.

One thing remaining to clarify is if 2 new pods start at the same time, is the secondary ensured to get the whole data during the initial sync? 

The full rollout process must be run against a k8s cluster to identify potential remaining flaws.

jhaeyaert (changes requested):

jhaeyaert (approved):
Looks ok. A couple of comments:

- Can you ensure the PR description aligns with our recent discussions?
- I noticed you 👍 my comment regarding the gateway.distributedSync versus gateway.distributedSync.enabled that could cause a BC. Did you confirm and apply any safeguards?
- Don't forget that the [online documentation](https://documentation.gravitee.io/apim/configure-and-manage-the-platform/gravitee-gateway/gateway-cluster-sync-with-redis) must be adapted (docker, helm, hz config xml, ...) to explain the behavior of the distributed sync (Hz cluster naming, Redis namespace, operational constraint regarding the 1 by 1 deployment, ...)
- The release note will have to specifically mention the changes regarding the distributed sync behavior

mukul-tyagi08:
> Looks ok. A couple of comments:
> 
> * Can you ensure the PR description aligns with our recent discussions?
> * I noticed you 👍 my comment regarding the gateway.distributedSync versus gateway.distributedSync.enabled that could cause a BC. Did you confirm and apply any safeguards?
> * Don't forget that the [online documentation](https://documentation.gravitee.io/apim/configure-and-manage-the-platform/gravitee-gateway/gateway-cluster-sync-with-redis) must be adapted (docker, helm, hz config xml, ...) to explain the behavior of the distributed sync (Hz cluster naming, Redis namespace, operational constraint regarding the 1 by 1 deployment, ...)
> * The release note will have to specifically mention the changes regarding the distributed sync behavior

1. PR description updated.
2. For the comment: gateway.distributedSync versus gateway.distributedSync.enabled:
   gateway.distributedSync: true (legacy boolean) still works; migration to gateway.distributedSync.enabled: true is not required.

The gateway.distributedSync.enabled helper accepts:

distributedSync: true (boolean)
distributedSync: { type: redis, ... } (map without enabled — treated as enabled)
distributedSync: { enabled: true, type: redis, ... }

Note: with boolean true, sync flags are enabled but the inline distributed-sync: Redis block is not rendered in gravitee.yml (same as when Redis config comes from external gravitee.yml). Map form with type still renders the Redis block.

Covered by helm unittest: enables sync flags when distributedSync is boolean true without inline redis config in configmap_hazelcast_cluster_test.yaml.

PR #18163, opened 2026-06-23
feat: add path setter to HttpRequestInternal to allow entrypoints to …

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-14363

## Description

Adds a `path(String)` setter to `HttpRequestInternal` so entrypoints can override the request path
exposed to policies and Expression Language (`{#request.path}`).

Previously, only `pathInfo` could be overridden, which meant entrypoints that transform an incoming
request into a different operation (e.g., MCP tool call → `GET /foo `) were invisible to path-based
policies — they would see the original entrypoint path instead of the transformed operation path.

## Additional context
This is part 1 of 2 — the setter in core. Part 2 (usage in the MCP Tool Server entrypoint) will
follow in `gravitee-entrypoint-mcp-tool-server`.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/AbstractRequest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/AbstractRequest.java
@@ -236,6 +236,12 @@ public abstract class AbstractRequest implements MutableRequest, HttpRequestInte
     }
 
     @Override
+    public HttpRequestInternal path(String path) {
+        this.path = path;
+        return this;
+    }
+
+    @Override
     public MutableRequest pathInfo(String pathInfo) {
         this.pathInfo = pathInfo;
         return this;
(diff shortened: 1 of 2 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/AbstractRequest.java
```diff
@@ -235,6 +235,12 @@ private void updatePathInfoFromContextPath(String contextPath) {
         }
     }
 
+    @Override
+    public HttpRequestInternal path(String path) {
+        this.path = path;
+        return this;
+    }
```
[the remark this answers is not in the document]
JedrzejJanasiak: Not relevant here: contextPath remains /mcp/ after the transformation, so auto-recalculating pathInfo from it would produce incorrect results. The caller sets both path and pathInfo explicitly, keeping state consistent.

gravitee-apim-gateway/gravitee-apim-gateway-core/src/test/java/io/gravitee/gateway/reactive/core/context/AbstractRequestTest.java
```diff
@@ -46,6 +46,13 @@ void setUp() {
     @Nested
     class OverrideRequestTest {
 
+        @Test
+        void should_override_path() {
```
michel-barret: looks useless to me

michel-barret (approved):

gravitee-apim-gateway/gravitee-apim-gateway-core/src/test/java/io/gravitee/gateway/reactive/core/context/AbstractRequestTest.java
JedrzejJanasiak: yeah, agree 😅 removed it

PR #18165, opened 2026-06-23
fix: distinguish "not yet started" from "expired" subscription error message

araffass:
## Purpose

Fix the misleading error message when a subscription has a future "Start At" date - the gateway incorrectly reports the subscription has "expired" instead of "not yet started".

## Summary of changes

- **SecurityChainDiagnostic**: Added a new notYetStartedSubscriptionPlans tracking list with dedicated markPlanHasNotYetStartedSubscription() method, and corresponding entries in message() and cause() with appropriate priority (after expired, before no matching rule).
- **AbstractSecurityPlan**: When isTimeValid() returns false, now checks whether startingAt is in the future to distinguish "not yet started" from "expired" and calls the appropriate diagnostic method.
- **Tests**: Added unit tests for the new message/cause in SecurityChainDiagnosticTest (message, cause, priority) and a new test in HttpSecurityPlanTest verifying the correct diagnostic is called for not-yet-started subscriptions.

## Out of scope

- The v1 security chain (CheckSubscriptionPolicy in gravitee-apim-gateway-security-jwt and gravitee-apim-gateway-security-oauth2) — these don't use SecurityChainDiagnostic and have no verbose error messaging.
- Exposing cause in response templates via EvaluableExecutionFailure — currently only message is available.

## Testing

- Unit tests: 45 tests passing in SecurityChainDiagnosticTest and HttpSecurityPlanTest.
- Local reproduction: With verbose401=true, confirmed the wrong message ("Access has expired") before the fix and the correct message ("The subscription has not yet started") after patching the gateway jar.
- Remote reproduction (4.10): Reproduced the bug on apim-4-10-x — confirmed the gateway logs show "The subscription has expired" for a subscription with a future start date.

## Related PRs / tickets

https://gravitee.atlassian.net/browse/APIM-14490

## Demo

**Before fix:**

https://github.com/user-attachments/assets/98d4b25c-9271-43db-8a69-7fbc0dbd2f53

**After fix:**

https://github.com/user-attachments/assets/d3884360-7ee3-4dab-9f61-c58f2a994227

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/security/SecurityChainDiagnostic.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/security/SecurityChainDiagnostic.java
@@ -28,6 +28,7 @@ public class SecurityChainDiagnostic {
     private static final String MESSAGE_INVALID_TOKEN = "The provided authentication token is invalid";
     private static final String MESSAGE_NOT_AUTHORIZED = "The provided credentials are not authorized";
     private static final String MESSAGE_EXPIRED = "Access has expired for the provided credentials";
+    private static final String MESSAGE_NOT_YET_STARTED = "The subscription has not yet started for the provided credentials";
     private static final String MESSAGE_NO_PLAN_MATCHED = "No plan matched the request";
     private static final String MESSAGE_NO_TOKEN = "The request did not include an authentication token";
     private static final String MESSAGE_UNAUTHORIZED = "Unauthorized";
@@ -39,6 +40,7 @@ public class SecurityChainDiagnostic {
     List<String> noMatchingRulePlans;
     List<NoSubscriptionInfo> noSubscriptionPlans;
     List<String> expiredSubscriptionPlans;
+    List<String> notYetStartedSubscriptionPlans;
 
     private record NoSubscriptionInfo(String planName, String tokenType, String maskedToken) {}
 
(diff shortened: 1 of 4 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/security/SecurityChainDiagnostic.java
```diff
@@ -93,6 +95,13 @@ public void markPlanHasExpiredSubscription(String planName, String applicationNa
         expiredSubscriptionPlans.add(planName + " (application: " + applicationName + ")");
     }
 
+    public void markPlanHasNotYetStartedSubscription(String planName, String applicationName) {
```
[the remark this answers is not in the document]
araffass: Only three similar lines of code, not worth abstraction

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/security/plan/AbstractSecurityPlan.java
```diff
@@ -170,7 +171,12 @@ private boolean validateSubscription(BaseExecutionContext ctx, SecurityToken sec
                     ctx.setInternalAttribute(ATTR_INTERNAL_SUBSCRIPTION, subscription);
                     return true;
                 }
-                securityChainDiagnostic.markPlanHasExpiredSubscription(planContext.planName(), subscription.getApplicationName());
+                var startingAt = subscription.getStartingAt();
```
schehade-gravitee: Write `Date` rather than `var` please

araffass: Done

schehade-gravitee (approved):

PR #18214, opened 2026-06-24
fix: remove mat cards. cancel edit instead of resetting form. use signals.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/APIM-13520

### Description

Multiple fixes for Application screen to make it consistent with the rest of the dev portal:
- **Style consistency**: removed mat cards and make application screens consistent with the rest of the portal. Replaced card titles with semantic headings, adjusted spacing on the host and form sections, and tweaked application details content padding.

- **Edit cancelling**: Cancel now returns to the read view instead of resetting form values while staying in edit mode.
- **Signals**: Converted hasApplicationError from a plain boolean to a signal in CreateApplicationComponent, with matching template and spec updates.

### Screenshots / video

https://github.com/user-attachments/assets/1337912f-9d43-47d1-b9fa-7afeee1f72c4

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-tab-settings/application-tab-settings-edit/application-tab-settings-edit.component.html
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-tab-settings/application-tab-settings-edit/application-tab-settings-edit.component.html
@@ -16,154 +16,137 @@
 
 -->
 @if (application$ | async; as application) {
-  <mat-card appearance="outlined" class="settings-edit">
-    <mat-card-content>
-      <form [formGroup]="applicationSettingsForm" class="settings-edit__form">
-        <div class="settings-edit__form__general">
-          <span class="next-gen-h3" i18n="@@applicationDetailsTitleApplicationSettings">Application details</span>
-          <div class="settings-edit__form__field">
-            <p class="next-gen-body" i18n="@@applicationNameApplicationSettings">Application name *</p>
-            <mat-form-field appearance="outline">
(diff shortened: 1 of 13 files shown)


gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-tab-settings/application-tab-settings-edit/application-tab-settings-edit.component.html
```diff
@@ -16,154 +16,143 @@
...
+        aria-label="Save application"
+        data-testId="save"
+      >
+        Save
+      </button>
+      <button
+        mat-stroked-button
+        (click)="cancel()"
+        i18n="@@discardChangesApplicationSettings"
+        aria-label="Discard changes"
```
jaroslawlakomy-grav: In the edit form, the visible button label was changed to `Cancel`, but the i18n key and `aria-label` still use the old “Discard changes” meaning.

Could we align them with the new behavior, e.g. use a Cancel-specific i18n key and an aria label like `Cancel editing` or simply `Cancel`?

jaroslawlakomy-grav (approved):
one non blocking comment, but in overall LGTM

gravitee-apim-portal-webui-next/src/app/dashboard/application-details/application-tab-settings/application-tab-settings-edit/application-tab-settings-edit.component.html
kovaren: Good point, fixed.

PR #18217, opened 2026-06-24
fix(gamma): refactor API URL resolution logic and add support for Gam…

Okhelifi:
## Summary

  Adds `InstallationAccessQueryService#getGammaManagementAPIUrl(orgId)` so the Gamma Console bootstrap can expose a `managementBaseURL` that lives under the same host as the Gamma API access point
  (instead of the Console API one). This keeps the auth/XSRF cookie scope consistent across `/gamma/*` and `/management/*` calls made by Gamma modules, fixing 401s on cross-host calls in multi-tenant
  deployments.

  Also dedupes the four sibling `getXxxAPIUrl` methods in `InstallationAccessQueryServiceImpl` (Console, Portal, Gamma, Gamma Management) — they all followed the same shape (resolve a base from access
  point or standalone fallback, then append a proxy path) — into one-line delegates to a single `resolveApiUrl(scopeId, accessPointLookup, standaloneUrl, proxyPath)` helper. Adds the corresponding stub in
   `InstallationAccessQueryServiceInMemory`.

  ## Why

  The previous bootstrap response used `getConsoleAPIUrl(orgId)` for `managementBaseURL`, which resolves to the **console** access point (e.g. `…console.example.com/management`). Cookies set there are
  host-only, so Gamma module calls hitting `…gamma-api.example.com/management/...` could never reuse them and got 401s. Pointing `managementBaseURL` to the Gamma API access point fixes the cookie scope
  without requiring operators to set `jwt.cookie-domain` to a common parent.

linked to https://github.com/gravitee-io/gravitee-cockpit/pull/6811

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/installation/query_service/InstallationAccessQueryService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/installation/query_service/InstallationAccessQueryService.java
@@ -16,7 +16,7 @@
 package io.gravitee.apim.core.installation.query_service;
 
 import io.gravitee.apim.core.installation.model.RestrictedDomain;
-import io.gravitee.rest.api.service.common.GraviteeContext;
+import jakarta.annotation.Nullable;
 import java.util.List;
 
 /**
@@ -57,5 +57,9 @@ public interface InstallationAccessQueryService {
 
     String getGammaUrl(final String organizationId);
 
+    @Nullable
     String getGammaAPIUrl(final String organizationId);
+
+    @Nullable
+    String getGammaManagementAPIUrl(final String organizationId);
 }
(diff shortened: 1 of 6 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/installation/query_service/InstallationAccessQueryService.java
```diff
@@ -58,4 +58,6 @@ public interface InstallationAccessQueryService {
     String getGammaUrl(final String organizationId);
 
     String getGammaAPIUrl(final String organizationId);
+
+    String getGammaManagementAPIUrl(final String organizationId);
```
michel-barret, line 64: suggestion: a small `@Nullable` to give the contract to users?

michel-barret (approved):

gcusnieux (approved):

PR #18254, opened 2026-06-25
feat(am-connection): persist environment id and verify the AM organization

sclarkamlab:
## Issue

https://gravitee.atlassian.net/browse/GMA-860

## Description

Two related changes to the AM connection.

First, the connection now persists the AM environment id (the APIM half of GMA-595). AM addresses domains as `organizations/{org}/environments/{env}/domains/{domain}`, but we only stored the domain, so a consumer couldn't build the path. I threaded `environmentId` through everywhere the existing `defaultDomainId` already lives: the repository model, the mongo and jdbc mappings (plus a liquibase changeset for the new column), the rest-api model and service, the shared `AmConnection` in the gamma plugin api, and the platform save/read use cases and UI. It's nullable, same as the domain.

Second, a fix so verifying the connection actually checks the AM organization you entered. The `/_test` endpoint was dropping `amOrganizationId` from the request body, and the tester was probing the APIM org instead of the connection's org, so a junk org still reported "connection succeeded". Now the org is carried through the test path and the tester probes it, and Save rejects a blank org up front.

Tests: added unit coverage for the tester (probes the configured org, fails fast on a blank org, surfaces AM's status when the org is rejected), the resource wiring (`/_test` and save both forward the org), the test use case, and save validation.

## Additional context

Two gamma modules consume the updated `AmConnection` and depend on this being merged and published first:
- gravitee-io/gravitee-gamma-module-aim#455
- gravitee-io/gravitee-gamma-module-authz#49

## Demo

https://github.com/user-attachments/assets/037a5ba5-468c-4fbc-8c7b-68b70510a0ff

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/21_add_am_organization_id_to_am_connections.yml
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/21_add_am_organization_id_to_am_connections.yml
@@ -9,3 +9,12 @@ databaseChangeLog:
               - column:
                   name: am_organization_id
                   type: nvarchar(64)
+        # Backfill existing connections to AM's built-in DEFAULT org so save/test (which now require a
+        # non-blank org) keep working without relying on the UI pre-fill — protects non-UI REST consumers.
+        - update:
+            tableName: ${gravitee_prefix}am_connections
+            columns:
+              - column:
+                  name: am_organization_id
+                  value: DEFAULT
+            where: am_organization_id IS NULL
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/21_add_am_organization_id_to_am_connections.yml
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/21_add_am_organization_id_to_am_connections.yml
@@ -1,0 +1,11 @@
+databaseChangeLog:
+  - changeSet:
+      id: 4.12.0_21_add_am_organization_id_to_am_connections
+      author: GraviteeSource Team
+      changes:
+        - addColumn:
+            tableName: ${gravitee_prefix}am_connections
+            columns:
+              - column:
+                  name: am_organization_id
+                  type: nvarchar(64)
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/core/am/use_case/connection/SaveAmConnectionUseCase.java
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/core/am/use_case/connection/SaveAmConnectionUseCase.java
@@ -16,6 +16,7 @@
 package io.gravitee.gamma.module.platform.core.am.use_case.connection;
 
 import io.gravitee.apim.core.UseCase;
+import io.gravitee.apim.core.exception.ValidationDomainException;
 import io.gravitee.apim.plugin.gamma.api.identity.AmConnection;
 import io.gravitee.apim.plugin.gamma.api.identity.AmConnectionRepository;
 import io.gravitee.gamma.module.platform.core.am.domain_service.AmConnectionViewDomainService;
@@ -32,6 +33,7 @@ public class SaveAmConnectionUseCase {
         String orgId,
         String baseUrl,
         String serviceAccountAccessToken,
+        String amOrganizationId,
         String environmentId,
         String defaultDomainId,
         String defaultDomainHrid,
@@ -41,6 +43,7 @@ public class SaveAmConnectionUseCase {
     public record Output(
         String baseUrl,
         boolean hasAccessToken,
+        String amOrganizationId,
         String environmentId,
         String defaultDomainId,
         String defaultDomainHrid,
@@ -48,11 +51,18 @@ public class SaveAmConnectionUseCase {
     ) {}
 
     public Output execute(Input input) {
+        // The DTO already enforces a non-blank baseUrl; the AM organization is just as required for
+        // a usable connection, so reject it here too rather than persist a half-configured one that
+        // every later AM call would fail on.
+        if (input.amOrganizationId() == null || input.amOrganizationId().isBlank()) {
+            throw new ValidationDomainException("AM organization is required");
+        }
         amConnectionRepository.save(
             input.orgId(),
             new AmConnection(
                 input.baseUrl(),
                 input.serviceAccountAccessToken(),
+                input.amOrganizationId(),
                 input.environmentId(),
                 input.defaultDomainId(),
                 input.defaultDomainHrid(),
@@ -63,6 +73,7 @@ public class SaveAmConnectionUseCase {
         return new Output(
             view.baseUrl(),
             view.hasAccessToken(),
+            view.amOrganizationId(),
             view.environmentId(),
             view.defaultDomainId(),
             view.defaultDomainHrid(),
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/core/am/use_case/connection/SaveAmConnectionUseCase.java
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/core/am/use_case/connection/SaveAmConnectionUseCase.java
@@ -57,12 +57,14 @@ public class SaveAmConnectionUseCase {
         if (input.amOrganizationId() == null || input.amOrganizationId().isBlank()) {
             throw new ValidationDomainException("AM organization is required");
         }
+        // Store the trimmed org so the persisted value matches what TestAmConnectionUseCase verifies.
+        var amOrganizationId = input.amOrganizationId().trim();
         amConnectionRepository.save(
             input.orgId(),
             new AmConnection(
                 input.baseUrl(),
                 input.serviceAccountAccessToken(),
-                input.amOrganizationId(),
+                amOrganizationId,
                 input.environmentId(),
                 input.defaultDomainId(),
                 input.defaultDomainHrid(),
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/rest/resource/am/AmConnectionResource.java
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/rest/resource/am/AmConnectionResource.java
@@ -58,6 +58,7 @@ public class AmConnectionResource {
         return new AmConnectionResponse(
             output.baseUrl(),
             output.hasAccessToken(),
+            output.amOrganizationId(),
             output.environmentId(),
             output.defaultDomainId(),
             output.defaultDomainHrid(),
@@ -80,6 +81,7 @@ public class AmConnectionResource {
                 orgId,
                 req.baseUrl(),
                 req.serviceAccountAccessToken(),
+                req.amOrganizationId(),
                 req.environmentId(),
                 req.defaultDomainId(),
                 req.defaultDomainHrid(),
@@ -89,6 +91,7 @@ public class AmConnectionResource {
         return new AmConnectionResponse(
             output.baseUrl(),
             output.hasAccessToken(),
+            output.amOrganizationId(),
             output.environmentId(),
             output.defaultDomainId(),
             output.defaultDomainHrid(),
@@ -109,8 +112,11 @@ public class AmConnectionResource {
     public AmConnectionTestResultResponse test(@PathParam("orgId") String orgId, AmConnectionRequest req) {
         String inboundBaseUrl = req == null ? null : req.baseUrl();
         String inboundToken = req == null ? null : req.serviceAccountAccessToken();
+        String inboundAmOrganizationId = req == null ? null : req.amOrganizationId();
         return AmDtoMapper.toDto(
-            testAmConnectionUseCase.execute(new TestAmConnectionUseCase.Input(orgId, inboundBaseUrl, inboundToken)).result()
+            testAmConnectionUseCase
+                .execute(new TestAmConnectionUseCase.Input(orgId, inboundBaseUrl, inboundToken, inboundAmOrganizationId))
+                .result()
         );
     }
 }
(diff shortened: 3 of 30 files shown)


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
```diff
@@ -318,13 +331,18 @@ export function AmConfigPanel({ onSaved, onCancel }: Props) {
...
         } finally {
+            clearTimeout(timeout);
+            testAbortRef.current = null;
             setTesting(false);
         }
     };
 
+    const handleCancelTest = () => testAbortRef.current?.abort();
+
     const canSaveConnection = Boolean(form.baseUrl) && Boolean(form.accessToken || form.hadAccessToken);
```
gcusnieux: The AM organization is now required, so we can do this ?
```suggestion
    const canSaveConnection = Boolean(form.baseUrl) && Boolean(form.amOrganizationId.trim()) && Boolean(form.accessToken || form.hadAccessToken);
```

gcusnieux (commented):
LGMT

gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/core/am/use_case/connection/SaveAmConnectionUseCase.java
```diff
@@ -41,18 +43,26 @@ public record Input(
...
         String gatewayUrl
     ) {}
 
     public Output execute(Input input) {
+        // The DTO already enforces a non-blank baseUrl; the AM organization is just as required for
+        // a usable connection, so reject it here too rather than persist a half-configured one that
+        // every later AM call would fail on.
+        if (input.amOrganizationId() == null || input.amOrganizationId().isBlank()) {
+            throw new ValidationDomainException("AM organization is required");
+        }
```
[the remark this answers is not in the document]
sclarkamlab: New feature so I don't believe this applies.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
sclarkamlab: Done

gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/core/am/use_case/connection/SaveAmConnectionUseCase.java
```diff
@@ -41,18 +43,26 @@ public record Input(
...
         amConnectionRepository.save(
             input.orgId(),
             new AmConnection(
                 input.baseUrl(),
                 input.serviceAccountAccessToken(),
+                input.amOrganizationId(),
```
podlesrafal: **Important — trim asymmetry.** `TestAmConnectionUseCase` trims the inbound org (`.trim()`), but here it's validated with `isBlank()` only and persisted verbatim. So `"  am-org  "` verifies OK but is stored with whitespace, and the stored value no longer matches what was verified. Suggest `var amOrg = input.amOrganizationId().trim();` after the blank check and persist that. No test currently pins trim-on-save.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/21_add_am_organization_id_to_am_connections.yml
```diff
@@ -0,0 +1,11 @@
...
+  - changeSet:
+      id: 4.12.0_21_add_am_organization_id_to_am_connections
+      author: GraviteeSource Team
+      changes:
+        - addColumn:
+            tableName: ${gravitee_prefix}am_connections
+            columns:
+              - column:
+                  name: am_organization_id
+                  type: nvarchar(64)
```
podlesrafal, line 11: **Important — backward compatibility.** This adds a nullable column, but `SaveAmConnectionUseCase` and `AmSdkConnectionTester` now hard-fail on null/blank `amOrganizationId`. Existing rows are NULL, so existing connections only keep working because the UI pre-fills `DEFAULT` (pre-fill, not auto-persist) — any non-UI REST consumer hits `400 "AM organization is required"`. Consider a backfill so it's robust independent of the frontend:
```yaml
- update:
    tableName: ${gravitee_prefix}am_connections
    columns:
      - column: {name: am_organization_id, value: DEFAULT}
    where: am_organization_id IS NULL
```
Otherwise this warrants a release note ("existing AM connections must be re-saved").

gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/rest/resource/am/AmConnectionResource.java
```diff
@@ -109,8 +112,11 @@ public AmConnectionResponse save(@PathParam("orgId") String orgId, @Valid @NotNu
     public AmConnectionTestResultResponse test(@PathParam("orgId") String orgId, AmConnectionRequest req) {
```
podlesrafal, line 112: **Important — validation asymmetry + test blind spot.** `save` has `@Valid @NotNull` but `test` doesn't, so `@NotBlank baseUrl` is never enforced on this path. That may be intentional (partial body tolerated for fallback), but it's undocumented. More importantly, `AmConnectionResourceTest` uses `@InjectMocks` with no Jersey container, so it bypasses Jackson deserialization and Bean Validation entirely — it cannot guard the serialization/binding layer where the original `/_test` amOrganizationId drop-bug would most plausibly have lived. The wiring assertion is fine, but consider a container-level test (or at least an explicit note that the partial-body tolerance is by design).

podlesrafal (commented):
Thorough review of the persistence + verification change (#1, persisting `amOrganizationId` and probing the configured AM org, is correct and well-tested — the regression test is precise and auth/permission failures now surface AM's real status with a cleaned message). Below are follow-ups grouped by severity. None are blockers on their own, but a couple guard real behavior. Inline comments have the specifics.

**Important**
- **Save persists `amOrganizationId` un-trimmed while `_test` trims it.** `"  am-org  "` verifies OK (trimmed) but is stored verbatim, so the saved value diverges from what was verified. Latent today, but it bites the moment any consumer uses the stored org for real AM calls. Trim on save (and ideally in the UI `buildRequest`). No test pins this.
- **No backfill for existing rows.** The column is nullable and existing connections have `am_organization_id = NULL`, yet save/test now hard-fail on null/blank. The only thing keeping existing connections working is the UI pre-filling `DEFAULT` (pre-fill, not auto-persist) — any non-UI REST consumer breaks. A one-line Liquibase backfill to `DEFAULT` removes the frontend coupling; otherwise this needs a release note.
- **`/_test` has no `@Valid @NotNull` (asymmetric with `save`), and the resource test can't catch it.** The `@InjectMocks` wiring test bypasses Jackson + Bean Validation, so it structurally can't guard the serialization/binding layer where the original `_test` drop-bug lived. Worth either documenting the partial-body tolerance as intentional or adding a container-level test.
- **Zero frontend coverage for the new `AmConfigPanel` logic** (org field, save-gating on `.trim()`, `AbortController` + 31s auto-timeout, Cancel). Pre-existing gap, but this is the most logic-heavy addition the component has had, and the module already has a Jest convention.

**Minor**
- `@NotBlank` layering asymmetry: `baseUrl` validated on the DTO, `amOrganizationId` imperatively in the use case. Pick one mechanism for the two required fields.
- In-flight verify isn't aborted when a field is edited (`setField` resets `connectionVerified` but doesn't `abort()`), so a stale verify can resolve late and flip `verified` back to true.
- Dead message: connectivity failures come back from the AM SDK as `ApiException(500)` → "rejected the connection", so the dedicated "Could not reach AM. Check the base URL." branch is now mostly unreachable.
- `AmConnection.isConfigured()` ignores `amOrganizationId`, so "configured" and "saveable" now use different definitions (pre-existing).
- Duplicate-ish test names in `TestAmConnectionUseCaseTest`: `should_fall_back_to_stored_am_organization_when_inbound_blank` vs `should_fall_back_to_stored_when_inbound_blank` (the latter covers baseUrl/token).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
```diff
@@ -290,20 +296,27 @@ export function AmConfigPanel({ onSaved, onCancel }: Props) {
...
         defaultDomainHrid: domainHrid || null,
         gatewayUrl: gatewayUrl || null,
     });
 
     const handleTest = async () => {
+        const controller = new AbortController();
+        testAbortRef.current = controller;
+        // Auto-abort just past the backend's 30s ceiling so a passive user isn't stuck indefinitely;
+        // the Cancel button lets them bail sooner.
+        const timeout = setTimeout(() => controller.abort(), 31_000);
```
podlesrafal, line 314: **Important — no test coverage for the new panel logic.** The `AbortController` + hardcoded 31s auto-timeout (tied to the backend's 30s ceiling), the Cancel button, the `Cancelled` vs error branching, the org field and `canSaveConnection` gating on `.trim()` — none of it is tested, and `AccessManagementPage.spec.tsx` mocks this component out. This is the most logic-heavy code the component has carried; a focused `AmConfigPanel.spec.tsx` would match the module's Jest convention.

**Minor (same file):** an in-flight verify isn't cancelled when a field is edited — `setField` resets `connectionVerified` but doesn't call `testAbortRef.current?.abort()`, so a verify started against old creds can resolve later and flip `connectionVerified` back to true against the now-edited form.

gravitee-gamma/gravitee-gamma-plugin/gravitee-gamma-plugin-module/gravitee-gamma-plugin-module-api/src/main/java/io/gravitee/apim/plugin/gamma/api/identity/AmConnection.java
```diff
@@ -19,13 +19,16 @@
...
+ * <p>Nullable fields ({@code serviceAccountAccessToken}, {@code amOrganizationId},
+ * {@code environmentId}, {@code defaultDomainId}, {@code defaultDomainHrid}, {@code gatewayUrl}) may
+ * be {@code null} when not yet configured. {@code amOrganizationId} is the AM organization and
+ * {@code environmentId} the AM environment that the {@code defaultDomainId} lives in — these are the
+ * AM-side coordinates, distinct from the APIM organization used as the storage key.
  */
 public record AmConnection(
     String baseUrl,
     String serviceAccountAccessToken,
+    String amOrganizationId,
```
podlesrafal, line 31: **Important — downstream binary compat.** This record's positional canonical constructor is published plugin API consumed by gravitee-gamma-module-aim#455 and gravitee-gamma-module-authz#49. Inserting a component changes the constructor descriptor, so a downstream jar compiled against the old 6-arg constructor throws `NoSuchMethodError` against the new class (mid-insertion vs append makes no difference to binary compat). The merge→publish→bump plan is necessary and sufficient **only if** both downstreams bump the plugin-api version and are rebuilt (watch for stale `.m2`). Worth confirming the plugin-api artifact version is bumped in this PR so consumers can't silently resolve the old descriptor. (Source-compat is fine — in-repo callers got a compile error, which is the safe outcome.)

gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/core/am/use_case/connection/SaveAmConnectionUseCase.java
sclarkamlab: Good catch. Trim it on save now and in the UI buildRequest, with a test pinning trim-on-save. 7e69127

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/21_add_am_organization_id_to_am_connections.yml
sclarkamlab: Yeah, added the backfill to DEFAULT so existing rows don't lean on the UI pre-fill. 7e69127

gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/rest/resource/am/AmConnectionResource.java
sclarkamlab: The missing @Valid on /_test is on purpose, it tolerates a partial body so the UI can test a new baseUrl while reusing the stored token. Agree the @InjectMocks test can't guard the binding layer though, I'll raise a follow-up for a container level test rather than block this.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
sclarkamlab: Fixed the stale-verify one, setField aborts the in-flight controller now. 7e69127. The panel spec is a fair gap, will raise a follow-up for AmConfigPanel.spec.tsx rather than do it in here.

gravitee-gamma/gravitee-gamma-plugin/gravitee-gamma-plugin-module/gravitee-gamma-plugin-module-api/src/main/java/io/gravitee/apim/plugin/gamma/api/identity/AmConnection.java
sclarkamlab: Right. plugin-api version is the APIM revision (${revision}${sha1}${changelist}) so it bumps every build, and aim#455 / authz#49 bump to the new version and rebuild as part of the merge->publish->bump plan. Nothing to pin in this PR but agree we watch for stale .m2 on the consumers.

podlesrafal (approved):

gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/rest/resource/am/AmConnectionResource.java
sclarkamlab: Actually just added it. AmConnectionRequestBindingTest covers the Jackson binding (the layer the drop-bug lived in) and bean validation (@NotBlank baseUrl). Didn't stand up the full container since /_test takes the unannotated body on purpose so there's no DTO constraint to assert there, noted that in the test. ecfb104

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/access-management/components/AmConfigPanel.tsx
sclarkamlab: Added AmConfigPanel.spec.tsx too. Covers the org trim gating, the verify abort signal, Cancel, and the stale-verify-on-edit case. ecfb104

PR #18300, opened 2026-06-26
fix(console): keep manually collapsed folders collapsed on tree refresh

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-14466

## Description


jaroslawlakomy-grav:
fixed in https://github.com/gravitee-io/gravitee-api-management/pull/18302

PR #18306, opened 2026-06-27
fix(console): Add panelWidth attribute to mat-select components for UI consistency

prateektiwari-gravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-14453

## Description

Updated the panelWidth attribute to mat-select components for UI consistency

## Additional context

Pre fix behaviour: 

image

Post fix behaviour:

image

### DIFF
--- a/gravitee-apim-console-webui/src/organization/configuration/user/detail/org-settings-user-detail.component.html
+++ b/gravitee-apim-console-webui/src/organization/configuration/user/detail/org-settings-user-detail.component.html
@@ -114,8 +114,33 @@
 
       <mat-form-field *ngIf="organizationRolesControl">
         <mat-label>Roles</mat-label>
-        <mat-select [formControl]="organizationRolesControl" aria-label="Organization role" multiple>
-          <mat-option *ngFor="let role of organizationRoles$ | async" [value]="role.id">{{ role.name }}</mat-option>
+        <mat-select
+          #organizationRoleSelect="matSelect"
+          [formControl]="organizationRolesControl"
+          aria-label="Organization role"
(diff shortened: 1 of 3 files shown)


gravitee-apim-console-webui/src/organization/configuration/user/detail/org-settings-user-detail-add-group-dialog.component.html
```diff
@@ -36,7 +36,7 @@ <h2 mat-dialog-title>Add a group with roles</h2>
 
       <mat-form-field>
         <mat-label>API Role</mat-label>
-        <mat-select formControlName="apiRole" aria-label="API role">
+        <mat-select formControlName="apiRole" aria-label="API role" [panelWidth]="null">
```
vikrantgravitee: What will happen if we have long name?

prateektiwari-gravitee: Yes, it was wrongly handled. Now, the field has a max-width, if the text goes beyond the max-width, the user get a tool tip on hover.

ankita-gupta21: Can you remove `[panelWidth]="null"`

prateektiwari-gravitee: Sure!

ankita-gupta21 (approved):

PR #18308, opened 2026-06-29
fix(api-product): enforce universe tier check for API Products in frontend

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-14554

## Description


phiz71 (changes requested):
Tiers and packs are for sales and TAM when they generate licenses files for customers.
We should never use them in our code nad always rely on features instead.

ankita-gupta21:
> Tiers and packs are for sales and TAM when they generate licenses files for customers. We should never use them in our code nad always rely on features instead.

@phiz71 Target : API Product feature is accessible only with Universe tier License.
I had to make changes to gravity-node. It was previously included for galaxies because the pack was inherited.

I now have done this:
- apim-api-products removed from enterprise-features (galaxy no longer inherits it) and standard (gamma-galaxy no longer inherits it)
- New universe-features pack created with apim-api-products
- universe-features added to both universe and gamma-universe tiers

Is this fine -> https://github.com/gravitee-io/gravitee-node/pull/582/changes

ankita-gupta21:
Closing the PR, Create another one 

- apim-api-products removed from enterprise-features (galaxy no longer inherits it) and standard (gamma-galaxy no longer inherits it)
- New universe-features pack created with apim-api-products
- universe-features added to both universe and gamma-universe tiers

Is this fine -> https://github.com/gravitee-io/gravitee-node/pull/582/changes

PR #18309, opened 2026-06-29
feat(opentelemetry): allow disabling V2 API tracing independently of V4

remibaptistegio:
## Problem

V2 APIs (legacy + emulated) have **no per-API tracing opt-in**, unlike V4 (which gates on `analytics.tracing.enabled`). So `services.opentelemetry.enabled=true` traces **every** V2 API on the gateway — a large, unintended span volume in deployments that mix V2 and V4 (notably Cloud, where OTel is enabled for V4 message/native observability).

## Change

Add a gateway-side flag **`services.opentelemetry.tracing.v2.enabled`** (default `true`, back-compatible). When `false`:
- the V2 reactor builds a no-op `TracingContext` (no per-API tracer allocated), and
- the V2 policy chain is not traced.

V4 is unaffected — still gated per-API on `analytics.tracing.enabled`.

The flag is read in `ApiReactorHandlerFactory` via the gateway `Configuration`. **gravitee-node stays version-agnostic** (only answers `isTracesEnabled()`); the V2-vs-V4 decision lives in the gateway, mirroring the existing V4 `analytics.tracing` gate.

## Tests

- `ApiReactorHandlerFactoryTest#shouldNotBuildTracerWhenV2TracingDisabled` — with OTel on and the flag off, no tracer is built (`verifyNoInteractions(openTelemetryFactory)`).
- Existing identity test updated to enable OTel (the factory now short-circuits to noop when tracing is disabled, so the tracer is only built when enabled).

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/ApiReactorHandlerFactory.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/ApiReactorHandlerFactory.java
@@ -122,6 +122,7 @@ public class ApiReactorHandlerFactory implements ReactorFactory<Api> {
     public static final String HANDLERS_REQUEST_HEADERS_X_FORWARDED_PREFIX_PROPERTY = "handlers.request.headers.x-forwarded-prefix";
     public static final String REPORTERS_LOGGING_EXCLUDED_RESPONSE_TYPES_PROPERTY = "reporters.logging.excluded_response_types";
     public static final String PENDING_REQUESTS_TIMEOUT_PROPERTY = "api.pending_requests_timeout";
+    public static final String TRACING_V2_ENABLED_PROPERTY = "services.opentelemetry.tracing.v2.enabled";
 
     /** OTel {@code gravitee.module} resource-attribute value identifying this reactor. */
     private static final String MODULE = "apim";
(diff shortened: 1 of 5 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/ApiReactorHandlerFactory.java
```diff
@@ -388,12 +388,8 @@ protected SyncApiReactor createSyncApiReactor(
         );
     }
 
-    protected HttpPolicyChainFactory createPolicyChainFactory(
-        Api api,
-        io.gravitee.gateway.reactive.policy.PolicyManager policyManager,
-        OpenTelemetryConfiguration openTelemetryConfiguration
-    ) {
-        return new HttpPolicyChainFactory(api.getId(), policyManager, openTelemetryConfiguration.isTracesEnabled());
+    protected HttpPolicyChainFactory createPolicyChainFactory(Api api, io.gravitee.gateway.reactive.policy.PolicyManager policyManager) {
```
phiz71, line 391: is there an impact in the reactor-message ?

phiz71 (commented):
LGTM. Just check that reactor-message is not impacted by the signature change

phiz71 (approved):

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/ApiReactorHandlerFactory.java
remibaptistegio: No impact on reactor-message, only on V2 APIs.

PR #18318, opened 2026-06-29
feat(rest-api): add email.branded_senders config list and BrandedSenderConfig DTO (APIM-14179, APIM-14180)

paulatulis:
## What
Backend data-model foundation for the configurable branded "From" / subject feature (epic APIM-14125).

- **APIM-14179** — register `EMAIL_BRANDED_SENDERS` (`email.branded_senders`) in `Key.java`, default `"[]"`, scopes `{ENVIRONMENT, ORGANIZATION, SYSTEM}`.
- **APIM-14180** — `BrandedSenderConfig` DTO (`domains`, `from`, `subject`) + `brandedSenders` on the `Email` settings model.

## How
Option A storage: the list of configs is JSON-serialised into a single `String` parameter, so it rides the existing `ParameterService` cascade / `ConfigService` plumbing (no new repository entity). A `@JsonIgnore` raw field holds the JSON; typed `getBrandedSenders()/setBrandedSenders()` accessors expose it, with a 4000-char guard matching the JDBC `parameters.value` column.

## Tests
7/7 unit tests cover default/empty, single & multi-config round-trip, null, blank, malformed JSON, and the size guard.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenderConfig.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenderConfig.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.model.settings;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * A single branded-sender configuration: the {@code from} address and {@code subject} prefix to
+ * apply when an outbound notification email targets one of {@code domains}.
+ *
+ * <p>Persisted as part of a JSON-serialised list under the {@code email.branded_senders} parameter
+ * (see {@link Email#getBrandedSenders()}).</p>
+ *
+ * @author GraviteeSource Team
+ */
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class BrandedSenderConfig {
+
+    /** Recipient domains (e.g. {@code "graviteecustomer.com"}) this configuration applies to. */
+    private List<String> domains;
+
+    /** Branded {@code From} address used for emails sent to a matching recipient domain. */
+    private String from;
+
+    /** Subject prefix wrapper (e.g. {@code "[Gravitee Customer] %s"}) applied to matching emails. */
+    private String subject;
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.model.settings;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.gravitee.rest.api.model.parameters.Key;
+import java.util.ArrayList;
+import java.util.List;
+import lombok.CustomLog;
+
+/**
+ * Serialises / deserialises the branded-sender configuration list to and from the single
+ * {@code email.branded_senders} parameter value, keeping the storage-format concerns out of the
+ * {@link Email} settings model so that it stays a plain data holder.
+ *
+ * @author GraviteeSource Team
+ */
+@CustomLog
+final class BrandedSenders {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    /**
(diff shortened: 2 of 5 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/Email.java
```diff
@@ -120,6 +147,68 @@ public void setFrom(String from) {
...
+     * write path instead (see {@link #setBrandedSenders(List)}).
+     */
+    @JsonProperty("brandedSenders")
+    public List<BrandedSenderConfig> getBrandedSenders() {
+        if (brandedSendersRaw == null || brandedSendersRaw.isBlank()) {
+            return new ArrayList<>();
+        }
+        try {
+            return BRANDED_SENDERS_MAPPER.readValue(brandedSendersRaw, new TypeReference<List<BrandedSenderConfig>>() {});
+        } catch (JsonProcessingException e) {
```
carlos-andres-osorio: The catch block logs the failure at WARN level but omits the raw stored value from the log entry. When the `email.branded_senders` parameter contains a corrupt value (e.g. a hand-edited DB row or a partial migration), the log message names only the config key and the Jackson parse-error position — not the string that failed to parse. An operator diagnosing why branded-senders silently returned empty must run a DB query to retrieve the raw value; the log alone is insufficient for post-incident identification.

Consider including the raw value in the log message so the corrupt parameter can be identified from the log without a database query.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenderConfig.java
```diff
@@ -0,0 +1,49 @@
...
+public class BrandedSenderConfig {
+
+    /** Recipient domains (e.g. {@code "graviteecustomer.com"}) this configuration applies to. */
+    private List<String> domains;
+
+    /** Branded {@code From} address used for emails sent to a matching recipient domain. */
+    private String from;
+
+    /** Subject prefix wrapper (e.g. {@code "[Gravitee Customer] %s"}) applied to matching emails. */
+    private String subject;
```
carlos-andres-osorio, line 48: The `from` and `subject` fields accept admin-supplied free-text strings that will feed email headers downstream. There is no CR/LF validation at the write boundary in `setBrandedSenders()` or on the DTO itself. A CR (`\r`) or LF (`\n`) embedded in either field could be exploited for email header injection (OWASP A05), depending on how the SMTP/notifier layer handles these values.

This concern depends on whether the downstream SMTP consumer sanitises CR/LF in header values — that code is not in this PR. If it does not, a defence-in-depth guard at the write boundary (rejecting values containing `\r` or `\n`) would close the risk at the point where data enters persistence.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/Email.java
paulatulis: Addressed!

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenderConfig.java
paulatulis: Addressed!

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
```diff
@@ -0,0 +1,120 @@
...
+     * {@link #MAX_SERIALIZED_LENGTH}, the storage column limit.
+     *
+     * <p>Any {@code ';'} in the serialised JSON is replaced with its JSON unicode escape so the stored
+     * value contains no literal {@code ';'}. The value is persisted as a single {@code ParameterService}
+     * parameter whose read path splits on {@code ';'} (the parameter separator) and keeps only the first
+     * fragment; since JSON structure never uses {@code ';'} (only {@code , : &#123; &#125; [ ] "}), every
+     * {@code ';'} here is inside a free-text {@code from} / {@code subject}. Jackson decodes the escape
+     * back to {@code ';'} transparently on read.</p>
+     */
+    static String write(List<BrandedSenderConfig> configs) {
```
carlos-andres-osorio, line 78: `write()` does not reject a null `BrandedSenderConfig` element in the input list. When a mutable list containing a null (e.g. `Arrays.asList(null, someConfig)`) is passed, `rejectHeaderInjection(null)` returns early without validation, Jackson serialises the null element as a JSON `null`, and a subsequent `parse()` call returns a list containing a null Java reference. Any caller of `getBrandedSenders()` that iterates the result will throw `NullPointerException`.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
```diff
@@ -0,0 +1,120 @@
...
+        if (raw == null || raw.isBlank()) {
+            return new ArrayList<>();
+        }
+        try {
+            var parsed = MAPPER.readValue(raw, new TypeReference<List<BrandedSenderConfig>>() {});
+            // A stored JSON literal `null` deserialises to a Java null; coerce to empty to honour the
+            // never-null contract.
+            return parsed != null ? parsed : new ArrayList<>();
+        } catch (JsonProcessingException e) {
+            log.warn("Ignoring malformed '{}' configuration [{}]; returning an empty list", Key.EMAIL_BRANDED_SENDERS.key(), raw, e);
```
carlos-andres-osorio: `parse()` catches `JsonProcessingException` and logs at `WARN` level. Operators relying on `ERROR`-level alerting or Sentry error thresholds receive no signal while the entire branded-senders feature silently reverts to the default sender and subject for every outbound email.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/settings/EmailBrandedSendersTest.java
```diff
@@ -0,0 +1,172 @@
...
+            )
+            .toList();
+
+        assertThatThrownBy(() -> email.setBrandedSenders(oversized))
+            .isInstanceOf(IllegalArgumentException.class)
+            .hasMessageContaining("maximum supported size");
+    }
+
+    @Test
+    void should_reject_line_breaks_in_from() {
```
carlos-andres-osorio, line 147: `rejectHeaderInjection` guards three header-bound fields — `from`, `subject`, and each entry in `domains` — but the test suite only covers `from` and `subject`. The `domains` line-break rejection branch has zero test coverage; a regression removing the domain check would ship silently, allowing header injection through a domain value to reach the email transport layer.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/settings/EmailBrandedSendersTest.java
paulatulis: Addressed

carlos-andres-osorio (approved):

PR #18332, opened 2026-06-30
fix(rest-api): bump swagger-parser to fix recursive OpenAPI import

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-13054

## Description

Importing certain valid OpenAPI specs (recursive / self-referencing schemas) crashed with a `StackOverflowError` thrown by swagger-parser in `ResolverFully.resolveSchema`. This was a bug in swagger-parser itself ([#1751](https://github.com/swagger-api/swagger-parser/issues/1751)), now fixed upstream.

- Bump `swagger-parser` 2.1.25 → **2.1.45** (and `swagger-core`/`swagger-jaxrs2` 2.2.29 → **2.2.52**, the aligned versions). The fixed parser detects cycles by object identity and stops instead of recursing infinitely.
- Add a regression test importing a minimal recursive spec. Verified **red on  2.1.25 / green on 2.1.45**.

### Additionnal
A second, independent commit wraps the parse call in `try/finally` so the temp file is always cleaned up (pre-existing leak when the parser throws, it can be dropped).

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/swagger/parser/OAIParser.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/swagger/parser/OAIParser.java
@@ -64,10 +64,13 @@ public class OAIParser extends AbstractDescriptorParser<OAIDescriptor> {
             }
         }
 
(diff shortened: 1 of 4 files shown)


phiz71:
I removed the "apply on 4.8.x" label since 4.8.X is no longer supported

kovaren (approved):

PR #18336, opened 2026-06-30
feat(rest-api): validate email.branded_senders and Email.from (APIM-14181)

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14178
## Description
Field validation for the new `email.branded_senders` configuration (APIM-14181), scoped to the
branded-sender path only.

- **`BrandedSenderConfig`**
  - `from`: `@NotBlank` + `@ValidSenderAddress` — a bare address (`user@example.com`) or the
    personal-name form (`Name <user@example.com>`) the SMTP send-path supports.
  - `domains`: `@NotEmpty`, `@NoDuplicates(ignoreCase)` (so `Acme.com`/`acme.com` collide), and each
    entry `@NotBlank` + a valid RFC-1035 host name.
- **`@Valid` cascade** added on `AbstractCommonSettingsEntity.email` and `Email.brandedSenders`, so the
  constraints actually fire end-to-end through the settings save path (the resource methods already
  carry `@Valid`; the nested cascade was the missing piece).
- **`email.from` is intentionally left unvalidated.** Turning on validation for that pre-existing field
  would 400 legacy stored values on the next settings save; validation is limited to the brand-new
  branded-sender data, which has no existing data and zero blast radius.
- Domains are trimmed + lower-cased at the write boundary so send-time matching can compare directly.

## New constraints (reusable, no new deps)
- `@ValidSenderAddress` / `SenderAddressValidator` — sender-address check that also accepts the
  `Name <addr>` form.
- `@NoDuplicates(ignoreCase)` / `NoDuplicatesValidator` — generic list-dedup constraint.

## Tests
- `BrandedSenderValidationTest` — field constraints (valid, display-name, invalid/blank from,
  empty/invalid/blank/duplicate domains) and the end-to-end `@Valid` cascade
  (`settings entity → email → branded senders`).
- `EmailBrandedSendersTest` — domain normalization round-trip (plus the existing serialization seam

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NoDuplicatesValidator.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NoDuplicatesValidator.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.validator;
+
+import jakarta.validation.ConstraintValidator;
+import jakarta.validation.ConstraintValidatorContext;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+
+/**
+ * @author GraviteeSource Team
+ */
+public class NoDuplicatesValidator implements ConstraintValidator<NoDuplicates, List<String>> {
+
+    private boolean ignoreCase;
+
+    @Override
+    public void initialize(NoDuplicates constraint) {
+        this.ignoreCase = constraint.ignoreCase();
+    }
+
+    @Override
+    public boolean isValid(List<String> values, ConstraintValidatorContext context) {
+        if (values == null) {
+            return true;
+        }
+        final Set<String> seen = new HashSet<>();
+        for (final String value : values) {
+            if (value == null) {
+                continue;
+            }
+            final String trimmed = value.trim();
+            final String key;
+            if (ignoreCase) {
+                key = trimmed.toLowerCase(Locale.ROOT);
+            } else {
+                key = trimmed;
+            }
+            if (!seen.add(key)) {
+                return false;
+            }
+        }
+        return true;
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/SenderAddressValidator.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/SenderAddressValidator.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.validator;
+
+import jakarta.validation.ConstraintValidator;
+import jakarta.validation.ConstraintValidatorContext;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * @author GraviteeSource Team
+ */
+public class SenderAddressValidator implements ConstraintValidator<ValidSenderAddress, String> {
+
+    /**
+     * Pragmatic {@code local@domain.tld} check for a branded sender address. The domain is matched as
+     * RFC 1035 labels (each starts and ends alphanumeric, no empty or leading-dot labels, no consecutive
+     * dots) with a bounded label length. The label group is matched with a possessive quantifier
+     * ({@code ++}) so the engine runs iteratively — it can't backtrack super-linearly or recurse into a
+     * stack overflow on a large input — and it rejects shapes like {@code a..b.com} / {@code .a.com}.
+     */
+    private static final Pattern EMAIL = Pattern.compile(
+        "^[A-Za-z0-9._%+-]+@(?:[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?\\.)++[A-Za-z]{2,}$"
+    );
+
+    /** Optional {@code Display Name <addr>} wrapper; captures the angle-bracketed address. */
+    private static final Pattern PERSONAL_NAME = Pattern.compile("^.*<\\s*([^<>]+?)\\s*>$");
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (value == null || value.isBlank()) {
+            return true;
+        }
+        final String trimmed = value.trim();
+        final Matcher personalName = PERSONAL_NAME.matcher(trimmed);
+        final String address;
+        if (personalName.matches()) {
+            address = personalName.group(1).trim();
+        } else {
+            address = trimmed;
+        }
+        return EMAIL.matcher(address).matches();
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/settings/BrandedSenderValidationTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/settings/BrandedSenderValidationTest.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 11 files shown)


carlos-andres-osorio:
> **gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/Email.java:54**

`Email.from` has no validation annotation. The acceptance criterion for APIM-14181 states: "Add `@Email` to the `from` field in `Email.java`." The field currently carries only `@ParameterKey(Key.EMAIL_FROM)`, so any string — including `"not-an-email"` — is accepted and persisted without error.

`BrandedSenderConfig.from` was annotated with `@ValidSenderAddress` in this PR; leaving the global fallback unvalidated creates an inconsistency between the two fields and leaves the AC unmet.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/settings/BrandedSenderValidationTest.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Test
+    void should_reject_empty_domains() {
+        assertThat(validator.validate(validConfig().domains(List.of()).build())).isNotEmpty();
+    }
+
+    @Test
+    void should_reject_invalid_domain() {
+        assertThat(validator.validate(validConfig().domains(List.of("not a domain")).build())).isNotEmpty();
```
carlos-andres-osorio: The only negative test for `DOMAIN_PATTERN` uses a space (`"not a domain"`), which trivially fails any domain regex without exercising the specific RFC 1035 sub-expressions. Five structurally distinct invalid shapes have no test:

- Leading hyphen per label: `"-example.com"`
- Trailing hyphen per label: `"example-.com"`
- Single-label (no dot): `"localhost"`
- One-character TLD: `"example.a"`
- Numeric-only TLD: `"example.123"`

A bug introduced in any of these sub-expressions of `DOMAIN_PATTERN` would not be caught by CI.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/settings/EmailBrandedSendersTest.java
```diff
@@ -36,6 +36,20 @@ void should_default_to_empty_list() {
         assertThat(new Email().getBrandedSenders()).isEmpty();
     }
 
+    @Test
+    void should_normalize_domains_to_trimmed_lowercase_on_write() {
```
carlos-andres-osorio, line 40: The null-domains guard in `normalizeDomains` (`if (config.getDomains() == null) return;`) has no test. `validateEntry` skips the line-break check when `getDomains()` is null, so a `BrandedSenderConfig` with no domains list reaches `normalizeDomains`. Removing the guard would produce a `NullPointerException` inside `write()` — an HTTP 500 on settings save — and no existing test would catch it.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/settings/BrandedSenderValidationTest.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    // --- @Valid cascade through the real save graph (settings entity -> email -> branded senders) ---
+
+    @Test
+    void should_cascade_from_settings_entity_into_branded_senders() {
+        var settings = new PortalSettingsEntity();
+        settings.getEmail().setBrandedSenders(List.of(validConfig().from("not-an-email").build()));
+
+        assertThat(validator.validate(settings)).anyMatch(v -> v.getPropertyPath().toString().contains("brandedSenders"));
```
carlos-andres-osorio: The cascade assertion checks only that the violation property path starts with `"brandedSenders"`, not the full indexed path. A regression that surfaced the violation at the `Email` level (e.g. `"email.brandedSenders"`) or dropped the entry index (e.g. `"brandedSenders.from"` instead of `"brandedSenders[0].from"`) would still pass this assertion. The AC requires the offending entry to be identifiable; a weaker path would not satisfy that.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/SenderAddressValidator.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * @author GraviteeSource Team
+ */
+public class SenderAddressValidator implements ConstraintValidator<ValidSenderAddress, String> {
+
+    /** Pragmatic {@code local@domain.tld} check, sufficient for a branded sender address. */
+    private static final Pattern EMAIL = Pattern.compile("^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,}$");
```
carlos-andres-osorio: The EMAIL regex domain segment `[A-Za-z0-9.-]+` permits consecutive dots (`user@a..b.com`) and a leading dot (`user@.example.com`), both of which are syntactically invalid per RFC 5321 but pass the current validator. An invalid address accepted here could cause SMTP relay rejection or silent bounce at send time — the failure this validator is intended to prevent.

Note: SonarQube also flagged this regex for super-linear backtracking risk (check annotation on line 32).

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NoDuplicatesValidator.java
```diff
@@ -0,0 +1,59 @@
...
+    public boolean isValid(List<String> values, ConstraintValidatorContext context) {
+        if (values == null) {
+            return true;
+        }
+        final Set<String> seen = new HashSet<>();
+        for (final String value : values) {
+            if (value == null) {
+                continue;
+            }
+            final String key;
```
carlos-andres-osorio, line 48: When `ignoreCase=false`, the comparison key is the raw untrimmed value, so `" a"` and `"a"` are treated as distinct entries. The `ignoreCase=true` branch trims before lowercasing. This asymmetry is undocumented on `@NoDuplicates`. The current sole production usage (domains with `ignoreCase=true`) is unaffected, but a future caller using `ignoreCase=false` would get different whitespace-handling behaviour than they might expect.

paulatulis:
> > **gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/Email.java:54**
> 
> `Email.from` has no validation annotation. The acceptance criterion for APIM-14181 states: "Add `@Email` to the `from` field in `Email.java`." The field currently carries only `@ParameterKey(Key.EMAIL_FROM)`, so any string — including `"not-an-email"` — is accepted and persisted without error.
> 
> `BrandedSenderConfig.from` was annotated with `@ValidSenderAddress` in this PR; leaving the global fallback unvalidated creates an inconsistency between the two fields and leaves the AC unmet.

I removed it deliberately but forgot to update the JIRA AC. 

`Email.from` has never been validated, so adding enforcement turns it on for existing stored values on the next settings save — and because the settings page POSTs the whole object, an admin editing an unrelated setting could get a 400 about email.from they never touched. The blast radius is small (the permissive regex accepts noreply@localhost; only genuinely-malformed legacy from values would trip), but it's the only live-behavior change in this PR, and it's on a field that nothing in this feature consumes yet (branded senders are a separate, brand-new param). The branded-sender path itself — which is what this epic is actually about — is fully validated.

I can re-add @ValidSenderAddress to Email.from (permissive regex), which satisfies the AC; accepts the small upgrade-time enforcement on malformed legacy values. Or I can keep it unvalidated for now and amend the AC — defer Email.from validation to whenever there's a migration/audit of existing values (or to the UI story that actually surfaces the field).

Do you have a strong preference?

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/settings/BrandedSenderValidationTest.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/SenderAddressValidator.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NoDuplicatesValidator.java
paulatulis: Addressed

carlos-andres-osorio:
> > > **gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/Email.java:54**
> > 
> > 
> > `Email.from` has no validation annotation. The acceptance criterion for APIM-14181 states: "Add `@Email` to the `from` field in `Email.java`." The field currently carries only `@ParameterKey(Key.EMAIL_FROM)`, so any string — including `"not-an-email"` — is accepted and persisted without error.
> > `BrandedSenderConfig.from` was annotated with `@ValidSenderAddress` in this PR; leaving the global fallback unvalidated creates an inconsistency between the two fields and leaves the AC unmet.
> 
> I removed it deliberately but forgot to update the JIRA AC.
> 
> `Email.from` has never been validated, so adding enforcement turns it on for existing stored values on the next settings save — and because the settings page POSTs the whole object, an admin editing an unrelated setting could get a 400 about email.from they never touched. The blast radius is small (the permissive regex accepts noreply@localhost; only genuinely-malformed legacy from values would trip), but it's the only live-behavior change in this PR, and it's on a field that nothing in this feature consumes yet (branded senders are a separate, brand-new param). The branded-sender path itself — which is what this epic is actually about — is fully validated.
> 
> I can re-add @ValidSenderAddress to Email.from (permissive regex), which satisfies the AC; accepts the small upgrade-time enforcement on malformed legacy values. Or I can keep it unvalidated for now and amend the AC — defer Email.from validation to whenever there's a migration/audit of existing values (or to the UI story that actually surfaces the field).
> 
> Do you have a strong preference?

Makes sense to not validate it.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/settings/EmailBrandedSendersTest.java
paulatulis: Addressed

carlos-andres-osorio (approved):
Re-reviewed after the prior round of comments. Verified all 6 previously-raised findings are fixed in the current commit: negative RFC-1035 domain tests added, null-domains guard tested, cascade assertion now checks the full indexed path, the SenderAddressValidator regex fixes both the ReDoS backtracking and the consecutive/leading-dot bypass, NoDuplicatesValidator's ignoreCase=false branch now trims consistently, and the subject/from/domain CR/LF header-injection guard is now enforced at the write boundary (BrandedSenders.validateEntry) with dedicated tests — properly surfaced as a 400 via JsonMappingExceptionMapper. Email.from is knowingly left unvalidated per the discussed AC deferral. LGTM.

PR #18358, opened 2026-07-01
feat(repository): add IdP claim whitelist & DCR mapping storage (APIM-13956)

wbabyte:
## What

First (foundational) story of epic **APIM-13949** — *Tenant/User context propagation in DCR via IdP claims*. This PR adds the **repository-layer storage** the rest of the feature builds on:

- `IdentityProvider.persistedClaimsWhitelist` (`List<String>`) — claim names allowed to be persisted on a user at login.
- `ClientRegistrationProvider.claimMappings` (`Map<String,String>`) — persisted-claim → DCR-request-field-path mapping.

Both fields are **optional / nullable** — existing installations are unaffected until configured.

## How

- **Mongo**: fields added to the `*Mongo` models; `update()` setters wired (they are field-by-field manual, not mapper-based).
- **JDBC**: `IdentityProvider` reuses its serialized-column pattern (`persisted_claims_whitelist` nclob); `ClientRegistrationProvider` gets a serialized `claim_mappings` nclob column.
- **Liquibase**: `v4_13_0` changelogs for both columns, registered in `master.yml`.

## Tests

Repository round-trip tests (create **and update**) for both fields, green on **MongoDB and JDBC** (Testcontainers).

## Notes

- Ref: **APIM-13956** (epic **APIM-13949**).
- Base of a stacked series; subsequent PRs (User `idpClaims` storage, REST config, login persistence, DCR injection, docs) build on this one and should be reviewed/merged after it.
- No UI (per product decision on APIM-13954); configuration is via the Management REST API.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/ClientRegistrationProvider.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/ClientRegistrationProvider.java
@@ -17,6 +17,7 @@ package io.gravitee.repository.management.model;
 
 import java.util.Date;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 public class ClientRegistrationProvider {
@@ -86,6 +87,11 @@ public class ClientRegistrationProvider {
     private String renewClientSecretMethod;
 
     private String softwareId;
+
+    /**
+     * Mapping of persisted IdP claim name to the DCR request field path it should be injected into.
+     */
+    private Map<String, String> claimMappings;
 
     private String trustStoreType;
 
@@ -243,6 +249,14 @@ public class ClientRegistrationProvider {
 
     public void setSoftwareId(String softwareId) {
         this.softwareId = softwareId;
+    }
+
+    public Map<String, String> getClaimMappings() {
+        return claimMappings;
+    }
+
+    public void setClaimMappings(Map<String, String> claimMappings) {
+        this.claimMappings = claimMappings;
     }
 
     public String getTrustStoreType() {
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/IdentityProvider.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/IdentityProvider.java
@@ -16,6 +16,7 @@
 package io.gravitee.repository.management.model;
 
 import java.util.Date;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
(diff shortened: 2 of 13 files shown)


paulatulis:
🤖 Claude first-pass review — Paula will follow up

**Summary:** Clean, well-scoped repository-layer change. Both new fields (`IdentityProvider.persistedClaimsWhitelist`, `ClientRegistrationProvider.claimMappings`) are nullable/optional, and JDBC + Mongo + Liquibase are wired consistently with create/update round-trip tests on both backends. I checked a couple of things that could have bitten:
- JDBC `ClientRegistrationProvider` create **and** update both go through the ORM (`buildInsertPreparedStatementCreator` / `buildUpdatePreparedStatementCreator`), so the new ORM-mapped `claim_mappings` column persists on both paths — the diff only touched `buildOrm()`, which is sufficient here.
- JDBC `IdentityProvider` uses hand-built SQL; `persisted_claims_whitelist` is appended consistently in `mapRow`, `createPreparedStatement`, `buildInsertStatement`, and `buildUpdateStatement`, so parameter ordering stays aligned for both insert and update.
- The new `v4_13_0` Liquibase directory matches the current `4.13.0-SNAPSHOT` in the root `pom.xml`, and both changelogs are registered in `master.yml`.

No blocking issues.

**Worth a look**

1. **Inconsistent deserialize/serialize error handling (robustness)** — In `JdbcClientRegistrationProviderRepository`, `serializeClaimMappings` / `deserializeClaimMappings` *throw* `IllegalArgumentException` on bad JSON. The analogous `JdbcIdentityProviderRepository.convertList` (and the existing `convert(...)` helpers) instead *log a warning and return null*. Because deserialize runs inside the ORM row mapper, a single row with malformed `claim_mappings` would fail the entire `findById` / `findAll` / `findAllByEnvironment` call (the latter two wrap it into a `TechnicalException`) rather than degrading just that one row. Real-world risk is low — the column is new and only ever written as valid JSON by this same code — but aligning with the lenient log-and-null pattern used elsewhere in these JDBC repos would be more robust and consistent.

**Minor / optional**

2. **equals/hashCode on the domain models** — The Mongo models (`IdentityProviderMongo`, `ClientRegistrationProviderMongo`) use Lombok `@EqualsAndHashCode`, so the new fields are picked up automatically. The repository-api domain models (`IdentityProvider`, `ClientRegistrationProvider`) have hand-written `equals`/`hashCode` and the new fields weren't added there. That may well be intentional (identity-by-id), just flagging for consistency.

3. **Raw `Map` type** — `serializeClaimMappings(Map ...)` / `deserializeClaimMappings(...)` use the raw `Map` to fit the ORM converter signature. Functionally fine; a `@SuppressWarnings("rawtypes")` (or parameterizing where the functional interface allows) would silence the unchecked/raw warning.

Nothing here blocks merge — the tests look solid and the change is low-risk and self-contained. This is an automated advisory pass; a human review from Paula is still coming.

paulatulis (approved):
No blockers!

PR #18365, opened 2026-07-01
feat(console): configurable From address for notification emails

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14190
https://gravitee.atlassian.net/browse/APIM-14191
## Description

What
Adds the classic console UI (Org + Portal settings → Email) for per-tenant branded notification senders: a read-only "Default notification email" section plus repeatable "Configuration N" cards, each with recipient domains (chips), a From address, and a Subject prefix. When a recipient's domain matches a rule, that rule's From/Subject is used instead of the EMAIL_FROM/EMAIL_SUBJECT defaults.

Details
- New standalone BrandedSendersComponent (src/shared/components/branded-senders/) — CVA + Validator, OnPush, signals, typed forms, gio-form-tags-input for domains.
- Wired into Org (org-settings-general) and Portal (portal-settings) email settings; brandedSenders added inline to the ConsoleSettingsEmail / PortalSettingsEmail entities.
- Client-side validation mirrors the backend rules (RFC-1035 domains, bare or Name <addr> From, 255-char subject cap).
- Harness-based spec (14 tests) + Shared / Branded senders stories.

Notes for reviewers
- Depends on the branded-senders persistence model (14179/14180), already merged to master — so this is functional standalone.
- Stricter server-side validation (14181, PR #18336) is still in review; the component enforces the same rules client-side in the meantime.
- Classic console only (gamma out of scope).

Covers S3 (APIM-14190) + S4 (APIM-14191).
## Additional context

https://github.com/user-attachments/assets/3a00e273-2153-4502-9cac-7b29afc7b10c

Documentation summary — Branded notification senders

  New capability
  Administrators can configure per-domain "branded" senders for outbound subscription/notification emails: a custom From address and Subject prefix that apply when a recipient's email domain matches a configured rule. Without a match, emails use the existing global default sender/subject, so this is purely additive. Configurable at both Organization and Environment (Portal) scope.

  Where to configure
  - Organization scope: Organization Settings → Console → Email
  - Environment scope: Environment/Portal Settings → Email
  - In both, a new "Branded notification email" section appears below the existing "Default notification email" section (which shows the system default From / Subject prefix as read-only, locked fields for reference).

  How to configure
  Each rule is a "Configuration" card with:
  - Recipient domains — one or more domains entered as chips (e.g. example.com, eu.example.com).
  - From — the branded sender address for matching recipients.
  - Subject prefix — a template applied to the notification's title; use %s where the title should appear (e.g. [Example] %s).
  Use + Add configuration to add rules and the trash icon to remove them. Changes are saved with the page's save bar.

  How matching works at send time
  - A recipient's domain (the part after @) is matched against each configuration's domains — case-insensitive and whitespace-tolerant.
  - First match wins if a domain somehow appears in more than one rule.
  - When a single notification goes to recipients across different domains, it is split so each recipient receives an email from the correct branded sender (recipients on non-matching domains get the default sender).
  - No match → the default From / Subject are used, unchanged.

  Scope & inheritance
  Values resolve Environment → Organization → system default: an environment uses its own branded senders if set, otherwise inherits the organization's, otherwise the system default.

Values resolve Environment → Organization → system default: an environment uses its own branded senders if set, otherwise inherits the organization's, otherwise the system default.

Validation rules (enforced inline in the UI and on save)
- Domain: valid host name (RFC 1035); IDN/punycode xn-- TLDs are accepted.
- From: a bare address (noreply@example.com) or a display-name form (Example Team <noreply@example.com>).
- Subject prefix: up to 255 characters.
- No duplicate domains across configurations.
- The total configuration set has a size limit; very large sets are rejected with an inline error.

Permissions
Editing requires ORGANIZATION_SETTINGS[U] (Org scope) or ENVIRONMENT_SETTINGS[U] (Environment scope). Users without the update permission see the settings form (including Add/Delete) disabled/read-only.

Known limitation (note for docs writer — deferred to a follow-up)
At Environment scope, the UI does not yet visually indicate whether displayed configurations are the environment's own overrides or values inherited from the Organization, and deleting an environment override does not visibly restore the inherited value. Tracked separately; consider documenting the current behavior (env values shown are the effective/resolved set) and revisiting once the "Inherited from Org" indicator ships.

Internal behavior — exclude from user docs
(For reviewers, not the published doc: multi-recipient notifications are delivered via Bcc; an explicit caller-provided From bypasses branding; a self-send to the configured address is not branded; the sender's copy from "copy to sender" uses the default identity. These are implementation details, not user-facing configuration.)

### DIFF
--- a/gravitee-apim-console-webui/src/entities/consoleSettings.ts
+++ b/gravitee-apim-console-webui/src/entities/consoleSettings.ts
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import { BrandedSender } from './brandedSender';
+
 export interface ConsoleSettings {
   email?: ConsoleSettingsEmail;
   metadata?: ConsoleSettingsMetadata;
@@ -47,6 +49,7 @@ export interface ConsoleSettingsEmail extends DisableableFeature {
   protocol?: string;
   subject?: string;
   from?: string;
+  brandedSenders?: BrandedSender[];
   properties?: {
     auth?: boolean;
     startTlsEnable?: boolean;
--- a/gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.html
+++ b/gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.html
@@ -735,6 +735,12 @@
             <input formControlName="sslTrust" matInput type="text" />
           </mat-form-field>
         </div>
+
+        <branded-senders
+          formControlName="brandedSenders"
+          [defaultFrom]="portalForm.get('email.from')?.value ?? ''"
+          [defaultSubject]="portalForm.get('email.subject')?.value ?? ''"
+        ></branded-senders>
       </mat-card-content>
     </mat-card>
 
--- a/gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
@@ -646,6 +646,71 @@ describe('PortalSettingsComponent', () => {
     });
   });
 
+  describe('branded senders', () => {
+    beforeEach(async () => {
+      await init();
+    });
+
+    it('should render the branded senders control enabled when emailing is on and not read-only', async () => {
+      const settings = fakePortalSettings();
+      settings.email.enabled = true;
+      expectPortalSettingsGetRequest(settings);
+
+      const addConfigurationButton = await loader.getHarness(MatButtonHarness.with({ selector: '.branded-senders__add' }));
+      expect(await addConfigurationButton.isDisabled()).toBe(false);
+    });
+
+    it('should disable branded senders when emailing is off', async () => {
+      const settings = fakePortalSettings();
+      settings.email.enabled = false;
+      expectPortalSettingsGetRequest(settings);
+
+      const addConfigurationButton = await loader.getHarness(MatButtonHarness.with({ selector: '.branded-senders__add' }));
+      expect(await addConfigurationButton.isDisabled()).toBe(true);
+    });
+
+    it('should disable branded senders when email.branded_senders is read-only', async () => {
+      const settings = fakePortalSettings();
+      settings.email.enabled = true;
+      settings.metadata.readonly.push('email.branded_senders');
+      expectPortalSettingsGetRequest(settings);
+
+      const addConfigurationButton = await loader.getHarness(MatButtonHarness.with({ selector: '.branded-senders__add' }));
+      expect(await addConfigurationButton.isDisabled()).toBe(true);
+    });
+
+    it('should round-trip branded senders from a populated GET through save at env scope', async () => {
+      const settings = fakePortalSettings();
+      settings.email.enabled = true;
+      settings.email.brandedSenders = [{ domains: ['example.com'], from: 'noreply@example.com', subject: '[Example] %s' }];
+      expectPortalSettingsGetRequest(settings);
+
+      const domainsInput = await loader.getHarness(GioFormTagsInputHarness.with({ selector: '[formControlName="domains"]' }));
+      await domainsInput.addTag('example.org');
+
+      const saveBar = await loader.getHarness(GioSaveBarHarness);
+      await saveBar.clickSubmit();
+
+      const req = httpTestingController.expectOne(`${CONSTANTS_TESTING.env.baseURL}/settings`);
+      expect(req.request.method).toEqual('POST');
+      expect(req.request.body.email.brandedSenders).toEqual([
+        { domains: ['example.com', 'example.org'], from: 'noreply@example.com', subject: '[Example] %s' },
+      ]);
+    });
+  });
+
+  describe('branded senders permissions', () => {
+    it('should disable branded senders when the user lacks environment-settings update permission', async () => {
+      await init([]);
+      const settings = fakePortalSettings();
+      settings.email.enabled = true;
+      expectPortalSettingsGetRequest(settings);
+
+      const addConfigurationButton = await loader.getHarness(MatButtonHarness.with({ selector: '.branded-senders__add' }));
+      expect(await addConfigurationButton.isDisabled()).toBe(true);
+    });
+  });
+
   function expectPortalSettingsGetRequest(portalSettings: PortalSettings) {
     httpTestingController
       .expectOne({
--- a/gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.html
+++ b/gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.html
@@ -435,6 +435,12 @@
             </mat-form-field>
           </ng-container>
         </ng-container>
+
+        <branded-senders
+          formControlName="brandedSenders"
+          [defaultFrom]="formSettings.get('email.from')?.value ?? ''"
+          [defaultSubject]="formSettings.get('email.subject')?.value ?? ''"
+        ></branded-senders>
       </mat-card-content>
     </mat-card>
     <gio-save-bar [form]="formSettings" [formInitialValues]="formInitialValues"> </gio-save-bar>
--- a/gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.spec.ts
+++ b/gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.spec.ts
@@ -31,8 +31,10 @@ import { MatIconTestingModule } from '@angular/material/icon/testing';
 import { OrgSettingsGeneralComponent } from './org-settings-general.component';
 
 import { CONSTANTS_TESTING, GioTestingModule } from '../../../shared/testing';
+import { SnackBarService } from '../../../services-ngx/snack-bar.service';
 import { OrganizationSettingsModule } from '../organization-settings.module';
 import { ConsoleSettings } from '../../../entities/consoleSettings';
+import { GioTestingPermissionProvider } from '../../../shared/components/gio-permission/gio-permission.service';
 
 describe('ConsoleSettingsComponent', () => {
   let fixture: ComponentFixture<OrgSettingsGeneralComponent>;
@@ -44,6 +46,7 @@ describe('ConsoleSettingsComponent', () => {
   beforeEach(() => {
     TestBed.configureTestingModule({
       imports: [NoopAnimationsModule, GioTestingModule, OrganizationSettingsModule, MatIconTestingModule],
+      providers: [{ provide: GioTestingPermissionProvider, useValue: ['organization-settings-u'] }],
     })
       .overrideProvider(InteractivityChecker, {
         useValue: {
@@ -503,6 +506,126 @@ describe('ConsoleSettingsComponent', () => {
         MatSlideToggleHarness.with({ name: 'emailPropertiesStartTlsEnable' }),
       );
       expect(await propertiesStartTlsEnableSlideToggle.isDisabled()).toEqual(false);
+    });
+
+    it('should enable and disable the branded senders control together with email.enabled', async () => {
+      expectConsoleSettingsGetRequest({
+        email: { enabled: true },
+        metadata: { readonly: [] },
+      });
+
+      const addConfigurationButton = await loader.getHarness(MatButtonHarness.with({ selector: '.branded-senders__add' }));
+      expect(await addConfigurationButton.isDisabled()).toEqual(false);
+
+      const emailEnabledSlideToggle = await loader.getHarness(MatSlideToggleHarness.with({ name: 'emailEnabled' }));
+      await emailEnabledSlideToggle.uncheck();
+      expect(await addConfigurationButton.isDisabled()).toEqual(true);
+
+      await emailEnabledSlideToggle.check();
+      expect(await addConfigurationButton.isDisabled()).toEqual(false);
+    });
+
+    it('should keep branded senders disabled when email.branded_senders is read-only, even after toggling email on', async () => {
+      expectConsoleSettingsGetRequest({
+        email: { enabled: true },
+        metadata: { readonly: ['email.branded_senders'] },
+      });
+
+      const addConfigurationButton = await loader.getHarness(MatButtonHarness.with({ selector: '.branded-senders__add' }));
+      expect(await addConfigurationButton.isDisabled()).toEqual(true);
+
+      // Toggling email off then on must not re-enable a system-provided (read-only) setting: this guards the
+      // camelCase form-path vs. snake_case parameter-key divergence in the enable/disable sync.
+      const emailEnabledSlideToggle = await loader.getHarness(MatSlideToggleHarness.with({ name: 'emailEnabled' }));
+      await emailEnabledSlideToggle.uncheck();
+      await emailEnabledSlideToggle.check();
+      expect(await addConfigurationButton.isDisabled()).toEqual(true);
+    });
+
+    it('should disable branded senders on initial load when email is disabled', async () => {
+      expectConsoleSettingsGetRequest({
+        email: { enabled: false },
+        metadata: { readonly: [] },
+      });
+
+      const addConfigurationButton = await loader.getHarness(MatButtonHarness.with({ selector: '.branded-senders__add' }));
+      expect(await addConfigurationButton.isDisabled()).toEqual(true);
+    });
+
+    it('should round-trip branded senders from a populated GET through save', async () => {
+      expectConsoleSettingsGetRequest({
+        email: {
+          enabled: true,
+          brandedSenders: [{ domains: ['example.com'], from: 'noreply@example.com', subject: '[Example] %s' }],
+        },
+        metadata: { readonly: [] },
+      });
+
+      const domainsField = await loader.getHarness(MatFormFieldHarness.with({ floatingLabelText: 'Recipient domains' }));
+      await (await domainsField.getControl(GioFormTagsInputHarness))!.addTag('example.org');
+
+      const saveButton = await loader.getHarness(GioSaveBarHarness);
+      await saveButton.clickSubmit();
+
+      expectConsoleSettingsSendRequest({
+        email: {
+          brandedSenders: [{ domains: ['example.com', 'example.org'], from: 'noreply@example.com', subject: '[Example] %s' }],
+        },
+      });
+    });
+
+    it('should surface a snackbar error when saving fails', async () => {
+      const snackBarService = TestBed.inject(SnackBarService);
+      const errorSpy = jest.spyOn(snackBarService, 'error');
+
+      expectConsoleSettingsGetRequest({
+        email: {
+          enabled: true,
+          brandedSenders: [{ domains: ['example.com'], from: 'noreply@example.com', subject: '[Example] %s' }],
+        },
+        metadata: { readonly: [] },
+      });
+
+      const domainsField = await loader.getHarness(MatFormFieldHarness.with({ floatingLabelText: 'Recipient domains' }));
+      await (await domainsField.getControl(GioFormTagsInputHarness))!.addTag('example.org');
+
+      const saveButton = await loader.getHarness(GioSaveBarHarness);
+      await saveButton.clickSubmit();
+
+      const req = httpTestingController.expectOne(`${CONSTANTS_TESTING.org.baseURL}/settings`);
+      req.flush({ message: 'Payload too large' }, { status: 400, statusText: 'Bad Request' });
+
+      expect(errorSpy).toHaveBeenCalledWith('Payload too large');
+    });
+  });
+
+  describe('permissions', () => {
+    beforeEach(() => {
+      // The root beforeEach grants organization-settings-u; re-create the component without it to exercise the
+      // read-only path (the previously-created instance and its HTTP controller are discarded by the reset).
+      TestBed.resetTestingModule();
+      TestBed.configureTestingModule({
+        imports: [NoopAnimationsModule, GioTestingModule, OrganizationSettingsModule, MatIconTestingModule],
+        providers: [{ provide: GioTestingPermissionProvider, useValue: [] }],
+      })
+        .overrideProvider(InteractivityChecker, {
+          useValue: {
+            isFocusable: () => true,
+          },
+        })
+        .compileComponents();
+
+      fixture = TestBed.createComponent(OrgSettingsGeneralComponent);
+      loader = TestbedHarnessEnvironment.loader(fixture);
+      httpTestingController = TestBed.inject(HttpTestingController);
+      fixture.detectChanges();
+    });
+
+    it('should disable the form including branded senders when lacking organization-settings-u', async () => {
+      expectConsoleSettingsGetRequest({ email: { enabled: true }, metadata: { readonly: [] } });
+
+      const addConfigurationButton = await loader.getHarness(MatButtonHarness.with({ selector: '.branded-senders__add' }));
+      expect(await addConfigurationButton.isDisabled()).toEqual(true);
     });
   });
 
--- a/gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.ts
+++ b/gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.ts
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-import { Subject } from 'rxjs';
-import { takeUntil, tap } from 'rxjs/operators';
+import { EMPTY, Subject } from 'rxjs';
+import { catchError, takeUntil, tap } from 'rxjs/operators';
 import { Component, OnDestroy, OnInit } from '@angular/core';
 import { UntypedFormBuilder, UntypedFormGroup } from '@angular/forms';
 import { get } from 'lodash';
@@ -26,6 +26,7 @@ import { ConsoleSettingsService } from '../../../services-ngx/console-settings.s
 import { ConsoleSettings } from '../../../entities/consoleSettings';
 import { CorsUtil } from '../../../shared/utils';
 import { SnackBarService } from '../../../services-ngx/snack-bar.service';
+import { GioPermissionService } from '../../../shared/components/gio-permission/gio-permission.service';
 
 @Component({
   selector: 'org-settings-general',
@@ -56,6 +57,7 @@ export class OrgSettingsGeneralComponent implements OnInit, OnDestroy {
     private readonly consoleSettingsService: ConsoleSettingsService,
     private readonly matDialog: MatDialog,
     private readonly snackBarService: SnackBarService,
+    private readonly permissionService: GioPermissionService,
   ) {}
 
   ngOnInit(): void {
@@ -100,6 +102,14 @@ export class OrgSettingsGeneralComponent implements OnInit, OnDestroy {
             protocol: [toFormState(this.settings, 'email.protocol')],
             subject: [toFormState(this.settings, 'email.subject')],
             from: [toFormState(this.settings, 'email.from')],
+            // Unlike the other email fields (hidden by *ngIf when email is off), branded-senders is always rendered,
+            // so it must be disabled at init when email is disabled — not only when the setting is read-only.
+            brandedSenders: [
+              {
+                value: get(this.settings, 'email.brandedSenders', []),
+                disabled: isReadonlySetting(this.settings, 'email.branded_senders') || !get(this.settings, 'email.enabled'),
+              },
+            ],
             properties: this.fb.group({
               auth: [toFormState(this.settings, 'email.properties.auth')],
               startTlsEnable: [toFormState(this.settings, 'email.properties.startTlsEnable')],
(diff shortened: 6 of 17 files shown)


carlos-andres-osorio:
> **gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.ts:709**

`onSubmit` builds the saved `email` payload from `...this.portalForm.get('email').value`, which now depends on the new `brandedSenders` FormControl holding the correct value at submit time. No test in `portal-settings.component.spec.ts` interacts with the branded-senders component and asserts the resulting save-request body contains the entered `brandedSenders` array — every other settings field group has such an assertion, but this one doesn't (confirmed via grep: zero matches). If the CVA value stopped propagating into the form (e.g. the control ends up disabled at submit time, since `.value` excludes disabled controls), saving would silently persist an empty/stale branded-senders list with a success snackbar shown to the admin.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+    });
+
+    it('should accept domains regardless of case and surrounding whitespace', () => {
+      control().setValue([{ domains: ['  Example.COM  '], from: 'noreply@example.com', subject: '' }]);
+      fixture.detectChanges();
+      expect(control().valid).toBe(true);
+    });
+  });
+
+  describe('inline error messages', () => {
```
carlos-andres-osorio, line 331: Cross-configuration duplicate-domain chip rejection has no test AND no implementation. `domainsValidator` only checks pattern validity per-entry, never comparing against sibling `FormGroup`s in the `configurations` FormArray — per the APIM-14190 grooming decision, adding a domain chip that duplicates a domain already used in another configuration should be rejected inline. Right now an admin can add the same domain (e.g. `example.com`) to two different configurations with no error, and which configuration's From/Subject actually applies is left to whatever backend first-match-wins logic exists, undocumented and invisible to the admin.

carlos-andres-osorio:
> **gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.ts:189**

The `should save email settings` test never sets a value for `brandedSenders` and never asserts it via `expectConsoleSettingsSendRequest`, which itself uses `toMatchObject` — an assertion style that only checks the subset of keys it was given, so omitting `brandedSenders` from the expected object means the test passes regardless of whether the field is present, absent, or mutated in the actual saved payload. A broken form-to-payload mapping for `brandedSenders` would silently drop the org admin's saved configuration with no test catching it.

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
```diff
@@ -646,6 +646,40 @@ describe('PortalSettingsComponent', () => {
     });
   });
 
+  describe('branded senders', () => {
```
carlos-andres-osorio, line 649: Org-inherited configuration display with an explicit "Inherited from Org" badge, and delete-restores-inherited (APIM-14191 items 13, 14, 16), have no tests and no implementation anywhere on the branch — a grep across the branded-senders and portal-settings files for "inherit"/"badge"/"cascade" returns zero real matches. The component only handles a flat list with add/edit/delete/save, with no scope-awareness (Org vs Env) or override/inherited state machine at all. An Env admin can never tell whether displayed configurations are their own Env-level overrides or values cascaded from Org, and can't tell that deleting one would fall back to Org values.

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.html
```diff
@@ -735,6 +735,12 @@ <h3 class="portal__form__card__title">Mail properties</h3>
             <input formControlName="sslTrust" matInput type="text" />
           </mat-form-field>
         </div>
+
+        <branded-senders
```
carlos-andres-osorio, line 739: Same underlying gap as the "Inherited from Org" finding above, viewed from the template angle: the `@for` loop rendering each branded-sender card has no conditional badge/label tied to an inherited state, `branded-senders.component.ts` defines no `isInherited`/`inheritedFromOrg` input or signal, and this usage only binds `formControlName`, `[defaultFrom]`, `[defaultSubject]` — no scope/origin metadata is passed through at all.

gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.spec.ts
```diff
@@ -504,6 +504,40 @@ describe('ConsoleSettingsComponent', () => {
       );
       expect(await propertiesStartTlsEnableSlideToggle.isDisabled()).toEqual(false);
     });
+
+    it('should enable and disable the branded senders control together with email.enabled', async () => {
```
carlos-andres-osorio, line 511: Org round-trip save (GET populated with `brandedSenders` → edit → POST body contains the updated `brandedSenders`) is untested — only Add-button enable/disable is tested. Same round-trip risk as the CRITICAL onSubmit finding above, specifically for the populated-GET → edit → save path rather than the empty-state path.

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
carlos-andres-osorio, line 649: Env-scope save round-trip is untested. The `describe('branded senders', ...)` block contains exactly three test cases, all scoped to `MatButtonHarness` disabled-state assertions on `.branded-senders__add`. None interact with the save bar or assert on a save HTTP request body, so a broken form-to-payload mapping for `brandedSenders` at Env scope would silently persist an empty/stale list.

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
```diff
@@ -646,6 +646,40 @@ describe('PortalSettingsComponent', () => {
...
+    beforeEach(async () => {
+      await init();
+    });
+
+    it('should render the branded senders control enabled when emailing is on and not read-only', async () => {
```
carlos-andres-osorio, line 654: Permission gating (`ENVIRONMENT_SETTINGS[U]`/`ORGANIZATION_SETTINGS[U]` missing → form disabled including the Add button) is untested at both portal and org scope, despite `init()` already accepting a `permissions` param (default `['environment-settings-u']`) — none of the three tests here vary it. A regression that leaves the branded-senders form editable for users without write permission would go undetected.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
```diff
@@ -0,0 +1,188 @@
...
+  private _onTouched: () => void = () => undefined;
+
+  // A CVA propagates its value on every form change; this subscription is the bridge to _onChange.
+  private readonly _propagateValue = this.configurations.valueChanges
+    .pipe(takeUntilDestroyed(this.destroyRef))
+    .subscribe(() => this._onChange(this.snapshot()));
+
+  // --- ControlValueAccessor ---
+
+  writeValue(value: BrandedSender[] | null): void {
```
carlos-andres-osorio, line 173: Question about `writeValue`: it calls `configurations.clear()` and then pushes freshly-built `FormGroup`s that are always constructed enabled (no read of `isDisabled()`). Since a `FormArray`'s disabled status is derived from `_allControlsDisabled()` on its *current* children, pushing one enabled child appears to flip the whole array — and every nested domains/from/subject control — back to enabled, regardless of the array's own disabled state set at directive setup. Angular only calls `setDisabledState` once at directive setup, not on subsequent `writeValue` calls. Does a later `writeValue` while the control should stay disabled (e.g. gio-save-bar's Discard resetting the form, or a post-save `ngOnInit()` form rebuild) re-enable the fields in practice, or is there a guard elsewhere that re-applies the disabled state that I'm missing?

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
```diff
@@ -0,0 +1,188 @@
...
+type BrandedSenderForm = FormGroup<{
+  domains: FormControl<string[]>;
+  from: FormControl<string>;
+  subject: FormControl<string>;
+}>;
+
+const SUBJECT_MAX_LENGTH = 255;
+
+// Mirrors the backend validation (RFC 1035 host name, and a bare address or a "Name <addr>" sender).
+const DOMAIN_PATTERN = /^(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,}$/;
```
carlos-andres-osorio: `DOMAIN_PATTERN`/`EMAIL_PATTERN` terminate with a TLD segment `[a-zA-Z]{2,}$`, allowing only alphabetic characters. ACE-encoded IDN TLDs (punycode, `xn--` prefixed, e.g. `xn--p1ai`, `xn--90ae`) contain digits and hyphens, so they fail this pattern even though they're valid LDH hostname labels — the backend `BrandedSenders.validateEntry` only rejects null entries and CR/LF, no format validation at all, so the "mirrors backend validation" comment above overstates the relationship. A customer whose recipient domain uses an IDN TLD can't create a branded sender rule for it at all, blocked by a misleading "Invalid domain(s)" client-side error.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
```diff
@@ -0,0 +1,188 @@
...
+  subject: string;
+}
+
+type BrandedSenderForm = FormGroup<{
+  domains: FormControl<string[]>;
+  from: FormControl<string>;
+  subject: FormControl<string>;
+}>;
+
+const SUBJECT_MAX_LENGTH = 255;
```
carlos-andres-osorio, line 58: `SUBJECT_MAX_LENGTH = 255` bounds only a single entry's subject, but the backend's real invariant is an aggregate 4000-char serialized-payload cap (`BrandedSenders.MAX_SERIALIZED_LENGTH`, `gravitee-apim-rest-api-model/.../BrandedSenders.java:42`), enforced in `write()` which throws `IllegalArgumentException` once exceeded. The frontend has no equivalent aggregate check anywhere in this component, so a user can add enough configurations — each individually valid per the 255-char/domain checks — to blow the 4000-char server-side budget, pass all client validation, and only discover the problem at save time via an opaque backend error with no field to blame.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
```diff
@@ -0,0 +1,188 @@
...
+import { MatCardModule } from '@angular/material/card';
+import { MatFormFieldModule } from '@angular/material/form-field';
+import { MatInputModule } from '@angular/material/input';
+import { MatButtonModule } from '@angular/material/button';
+import { MatTooltipModule } from '@angular/material/tooltip';
+import { MatIconModule } from '@angular/material/icon';
+import { GioFormTagsInputModule } from '@gravitee/ui-particles-angular';
+
+/** A single branded-sender rule — the value model exposed by this form control. */
+export interface BrandedSender {
```
carlos-andres-osorio: The `BrandedSender` model is defined here, inside the shared UI component, instead of `entities/` (the pure-model layer), and is re-declared as anonymous inline types in three other files (`consoleSettings.ts:50`, `portalSettings.ts:228`, `portal-settings.component.ts:164`). This inverts the entities-as-pure-model-layer convention and creates four independent copies that can drift silently if the backend model gains a field.

gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.html
```diff
@@ -435,6 +435,12 @@ <h3 class="settings-general__form__card__h3">Mail properties</h3>
             </mat-form-field>
           </ng-container>
         </ng-container>
+
+        <branded-senders
```
carlos-andres-osorio, line 439: Question about the placement here: `<branded-senders>` sits outside the `*ngIf="emailEnabledSlideToggle.checked"` wrapper that hides every other email field, and the component appears to only get disabled (a) at creation if readonly, or (b) via a later `email.enabled` `valueChanges` subscription — which wouldn't fire at initial load. Both spec tests start from `email.enabled: true`, so I'm not sure the initial-load case with `email.enabled: false` is covered. Is that intentional (e.g. branded-senders should stay visible/editable regardless of the enabled toggle), or should it be inside the `*ngIf` and created disabled like the portal-settings equivalent (`portal-settings.component.ts:592`)?

gravitee-apim-console-webui/src/entities/consoleSettings.ts
```diff
@@ -47,6 +47,7 @@ export interface ConsoleSettingsEmail extends DisableableFeature {
   protocol?: string;
   subject?: string;
   from?: string;
+  brandedSenders?: { domains: string[]; from: string; subject: string }[];
```
carlos-andres-osorio: The branded-sender shape is declared as an anonymous inline type here, duplicated again in `portalSettings.ts:228` and a third time in `portal-settings.component.ts:164`, even though a named `BrandedSender` interface with the exact same shape is already exported from `branded-senders.component.ts:41`.

paulatulis:
> > **gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.ts:189**
> 
> The `should save email settings` test never sets a value for `brandedSenders` and never asserts it via `expectConsoleSettingsSendRequest`, which itself uses `toMatchObject` — an assertion style that only checks the subset of keys it was given, so omitting `brandedSenders` from the expected object means the test passes regardless of whether the field is present, absent, or mutated in the actual saved payload. A broken form-to-payload mapping for `brandedSenders` would silently drop the org admin's saved configuration with no test catching it.

Thank you for the thorough review! All comments have been addressed in a follow-up commit (will squash before merge). Bugs #1/#2, validation #6/#7, the missing round-trip + permission tests, the shared BrandedSender model move, and cross-config duplicate-domain (14190) are all done. The "Inherited from Org" indicator + delete-restores (14191) is confirmed in-scope but has a backend dependency. Will address in https://gravitee.atlassian.net/browse/APIM-14612

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
paulatulis: Confirmed bug — good catch. writeValue now re-applies the disabled state after rebuilding the array, so a later Discard/reset won't re-enable it. Added a regression test.

gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.html
paulatulis: You're right, it was untested and inconsistent with Portal. Branded-senders is now disabled at Org init when email is off (not just when readonly), and there's a test for the enabled:false initial-load case.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
paulatulis: Fixed — the final label now accepts xn-- ACE/IDN TLDs. Also corrected the misleading "mirrors backend validation" comment to state these are UX-only checks (backend only rejects null + CR/LF).

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
paulatulis: Added a FormArray-level serialized-size validator (4000) with an inline error, so oversize sets fail client-side with a message instead of an opaque save-time error.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.spec.ts
paulatulis: Implemented per the 14190 grooming decision — duplicate domains across configs are now rejected inline (case-insensitive) with a message listing the offenders, and block save. Test added.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
paulatulis: Moved BrandedSender to entities/brandedSender.ts as the single source of truth; the 4 inline/duplicated shapes now reference it.

gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.spec.ts
paulatulis: Added populated-GET → edit → POST round-trip tests at both Org and Env scope.

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
paulatulis: Added a Portal test asserting the form (incl. Add) is disabled without environment-settings-u. Note: Org has no in-component permission gating (it's route-level), so there's no equivalent Org unit test.

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
paulatulis: Confirmed in-scope for 14191, but it needs backend support the current API doesn't provide. The settings endpoint returns only resolved-cascade values + metadata.readonly — no per-field scope/origin — so the Env UI can't distinguish "inherited from Org" from an Env override, and "delete restores inherited" is impossible (saving [] writes an empty override, not a delete). APIM's analogous notification-templates feature relies on a backend overridden flag + a reset endpoint. Proposing we split this into a follow-up sub-task that adds that backend support (expose overridden + reset-to-inherited) plus the Env inheritance UI, rather than shipping a misleading frontend-only approximation here.

gravitee-apim-console-webui/src/entities/consoleSettings.ts
paulatulis: Addressed

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
paulatulis: Addressed

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.html
paulatulis: Addressed above

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
```diff
@@ -0,0 +1,223 @@
...
+    return null; // emptiness is handled by Validators.required
+  }
+  const personalName = /^.*<([^<>]+)>$/.exec(value);
+  const address = personalName ? personalName[1].trim() : value;
+  return EMAIL_PATTERN.test(address) ? null : { invalidSenderAddress: true };
+};
+
+/** Rejects the whole list when its serialized size would exceed the backend's aggregate cap. */
+const serializedSizeValidator: ValidatorFn = (control: AbstractControl): ValidationErrors | null => {
+  const serializedLength = JSON.stringify(control.value ?? []).length;
```
carlos-andres-osorio: nit: `serializedSizeValidator` measures raw `JSON.stringify(control.value ?? []).length`, but the backend (`BrandedSenders.write`) replaces every `;` with the 6-character escape `\u003b` before comparing against `MAX_SERIALIZED_LENGTH`. A configuration set with semicolons in `from`/`subject` near the 4000-char limit can pass this client-side check yet still be rejected server-side.

Rather than replicating the backend's exact escaping logic here, it might be worth letting the backend stay the sole authority on this limit and surfacing its rejection with a clear, specific error on save instead — see the related note on the save subscription's missing error handler.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
```diff
@@ -0,0 +1,223 @@
...
+
+/** Accepts a bare address (`user@example.com`) or the personal-name form (`Name <user@example.com>`). */
+const senderAddressValidator: ValidatorFn = (control: AbstractControl): ValidationErrors | null => {
+  const value: string = (control.value ?? '').trim();
+  if (value === '') {
+    return null; // emptiness is handled by Validators.required
+  }
+  const personalName = /^.*<([^<>]+)>$/.exec(value);
+  const address = personalName ? personalName[1].trim() : value;
+  return EMAIL_PATTERN.test(address) ? null : { invalidSenderAddress: true };
```
carlos-andres-osorio, line 87: `senderAddressValidator` tests the extracted address as-is, but the `TLD` pattern shared with `domainsValidator` only matches a lowercase `xn--` prefix for ACE/IDN TLDs. `domainsValidator` normalizes its input with `.toLowerCase()` before testing; `senderAddressValidator` doesn't — so a sender address like `noreply@example.XN--P1AI` is rejected even though the equivalent domain entry in the same form would be accepted. DNS hostnames, including ACE/Punycode-encoded IDN labels, are case-insensitive per RFC 4343 / RFC 1035 §2.3.3.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  registerOnChange(fn: (value: BrandedSender[]) => void): void {
+    this._onChange = fn;
+  }
+
+  registerOnTouched(fn: () => void): void {
+    this._onTouched = fn;
+  }
+
+  setDisabledState(isDisabled: boolean): void {
```
carlos-andres-osorio, line 193: `setDisabledState`'s `else` branch (re-enabling the `configurations` FormArray and clearing the `isDisabled` signal that gates the add/delete buttons) isn't exercised by any test — the `disabled state` describe block only disables and asserts, it never re-enables. A regression that leaves fields or the add button stuck disabled after re-enabling would pass the suite undetected.

carlos-andres-osorio:
> **gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.ts:287**

The save subscription (`consoleSettingsService.save(...).pipe(tap(...), takeUntil(...)).subscribe(...)`) has only a success `tap` and no `error` callback. Since the client-side validators here are explicitly UX-only (not a full mirror of backend validation), a backend rejection is a live path — and today it fails silently with no error snackbar and no log. `portal-settings.component.ts` has the same pattern.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
paulatulis: Fixed — both `DOMAIN_PATTERN`/`EMAIL_PATTERN` now carry the `i` flag, so hostnames (including `xn--`/`XN--` ACE/IDN TLDs) match case-insensitively in both validators. Added a test for `noreply@example.XN--P1AI`.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
paulatulis: Good catch — added a test that disables then re-enables the control and asserts the fields and the Add button all become editable again.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
paulatulis: Kept the fail-fast client guard but made it conservative: it now mirrors the backend escaping each `;` to a 6-char sequence before measuring, so a semicolon-heavy set near the limit will fail here rather than passing and being rejected server-side. And per your related note, the backend stays the sole authority — its rejection is now surfaced via the new save error handler.

paulatulis:
> The save subscription (`consoleSettingsService.save(...).pipe(tap(...), takeUntil(...)).subscribe(...)`) has only a success `tap` and no `error` callback… `portal-settings.component.ts` has the same pattern.

Fixed in both components — the save subscription now has a `catchError` that surfaces the backend error message via `snackBarService.error(...)` (matching the `api-portal-header` pattern) instead of failing silently. Added a test that flushes a 400 on save and asserts the error snackbar fires.

gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.ts
```diff
@@ -100,6 +100,14 @@ export class OrgSettingsGeneralComponent implements OnInit, OnDestroy {
             protocol: [toFormState(this.settings, 'email.protocol')],
             subject: [toFormState(this.settings, 'email.subject')],
             from: [toFormState(this.settings, 'email.from')],
+            // Unlike the other email fields (hidden by *ngIf when email is off), branded-senders is always rendered,
+            // so it must be disabled at init when email is disabled — not only when the setting is read-only.
+            brandedSenders: [
```
carlos-andres-osorio, line 107: This form is never disabled for a user lacking `organization-settings-u` — including the new branded-senders Add/Delete controls. The only disabling logic present here is `isReadonlySetting(...)` (system-provided readonly config) and the `email.enabled` toggle; neither reflects the current user's write permission.

`portal-settings.component.ts` has the equivalent mechanism this component lacks: it injects `GioPermissionService`, computes `hasEnvironmentSettingsUpdatePermission = this.permissionService.hasAnyMatching(['environment-settings-u'])`, and calls `this.portalForm.disable()` when that's false — disabling the whole form including Add/Delete.

A read-only org admin can currently still add, edit, and delete branded-sender configurations (and any other org email setting).

carlos-andres-osorio:
Docs label is currently missing on this PR. Items identified as potentially doc-relevant, for you to decide what's worth including:

- New capability: per-domain "branded" notification senders at both Org and Environment (Portal) scope — a `From` address and `Subject` prefix override for specific recipient domains, instead of every notification email using one global sender.
- UI location: Org Settings → Email, and Portal/Environment Settings → Email — a new "Branded notification email" section below the existing "Default notification email" (locked/system) fields.
- Behavior: domain matching picks which config applies (first-match against `Recipient domains` chips); validation rules (RFC-1035 domain format, bare-or-`Name <email>` From format, 255-char subject cap); permission requirements (`ORGANIZATION_SETTINGS[U]` / `ENVIRONMENT_SETTINGS[U]` to edit).
- Known gap: Environment-scope "inherited from Org" isn't visually indicated yet (deferred to follow-up ticket APIM-14612) — env admins can't currently tell from the UI whether displayed values are their own override or cascaded from Org.

gravitee-apim-console-webui/src/organization/configuration/console/org-settings-general.component.ts
paulatulis: Fixed — OrgSettingsGeneralComponent now injects GioPermissionService and disables the whole form (including the branded-senders Add/Delete controls) when the user lacks organization-settings-u, matching PortalSettingsComponent. Added a test that re-creates the component without the permission and asserts the Add button is disabled, plus granted the permission in the existing suite's setup.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.spec.ts
```diff
@@ -0,0 +1,340 @@
...
+    loader = TestbedHarnessEnvironment.loader(fixture);
+    fixture.detectChanges();
+  });
+
+  afterEach(() => {
+    fixture.destroy();
+  });
+
+  describe('default section', () => {
+    it('should display the default from and subject as read-only fields', async () => {
```
carlos-andres-osorio, line 72: The acceptance criteria for this component call out that "Default From / Default subject prefix render as locked with tooltip." The `should display the default from and subject as read-only fields` test asserts value and `isDisabled()` state for both fields, but never asserts the `matTooltip="Configuration provided by the system"` text on the lock icon (`branded-senders.component.html:25,31`) — so half the AC-mandated behavior stays unverified, and a regression that dropped or emptied the tooltip wouldn't be caught.

Is this something that's easy to add as a unit test here (e.g. querying the lock icon's `matTooltip` in the existing test), or would it be better verified empirically/manually instead?

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.html
```diff
@@ -0,0 +1,112 @@
...
+
+-->
+<div class="branded-senders">
+  <!-- Default notification email (read-only, for context) -->
+  <h3 class="branded-senders__title">Default notification email</h3>
+  <p class="branded-senders__subtitle">Used when no branded rule matches the recipient's email domain.</p>
+
+  <mat-form-field class="branded-senders__field">
+    <mat-label>Default From</mat-label>
+    <mat-icon matPrefix class="branded-senders__lock" matTooltip="Configuration provided by the system">lock</mat-icon>
```
carlos-andres-osorio: The read-only "Default From" and "Default subject prefix" preview fields carry `matTooltip="Configuration provided by the system"` unconditionally (also at line 31). That exact string is otherwise used only conditionally elsewhere, gated by `isReadonlySetting(...)`/`isReadonly(...)`, on fields locked because a system/environment property actually overrode them (e.g. `org-settings-general.component.html:356`).

Here, these two fields are locked unconditionally by design — they're just a read-only echo of the editable `email.from`/`email.subject` fields shown earlier in the same form, not because a system property overrode them. An admin who set `email.from` themselves will still see "Configuration provided by the system," which is factually wrong and could send them looking for a non-existent env-level override.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
```diff
@@ -0,0 +1,229 @@
...
+  readonly defaultSubject = input('');
+
+  protected readonly subjectMaxLength = SUBJECT_MAX_LENGTH;
+  protected readonly configurations = new FormArray<BrandedSenderForm>([], [serializedSizeValidator, duplicateDomainsValidator]);
+  protected readonly isDisabled = signal(false);
+
+  private _onChange: (value: BrandedSender[]) => void = () => undefined;
+  private _onTouched: () => void = () => undefined;
+
+  // A CVA propagates its value on every form change; this subscription is the bridge to _onChange.
```
carlos-andres-osorio: The CVA bridge `this.configurations.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(() => this._onChange(this.snapshot()))` has no error handler. If `snapshot()` or the Angular-forms `_onChange` callback ever throws, RxJS terminates this subscription permanently (no resubscribe) — from that point on, edits to the branded-senders configurations silently stop propagating to the parent form control, with the value shown in the sub-component still changing while the parent `FormControl`/save payload freezes at the last successfully-propagated value.

Low likelihood in the current code (`snapshot()` only trims/lowercases owned controls), but it's an unguarded failure mode with a confusing, delayed user-facing symptom if a future edit introduces a throwable path.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.spec.ts
paulatulis: Added

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.html
paulatulis: Addressed

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
paulatulis: Addressed

carlos-andres-osorio (approved):

PR #18368, opened 2026-07-02
fix(api-product): distributed sync enablement

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-14555

## Description
Enables API Products in the gateway distributed sync (Redis) pipeline, so replica gateways receive product deploy/undeploy events from the primary — the same model already used for APIs, subscriptions, API keys, dictionaries, etc.

Before this change, API Products synced from the management DB on primary nodes only. Replicas had no way to consume product state from Redis, so product plans, subscriptions, and product-scoped API keys were missing on secondaries.

**Problem**
In a multi-gateway / sharded setup:

Primary syncs API Products from JDBC (ApiProductSynchronizer — already existed).
Replicas rely on distributed sync (Redis), not direct DB sync.
API_PRODUCT was not a DistributedEventType, and there was no publish/consume path for products.
Result: products deployed on primaries did not reliably propagate to replicas.

**Solution**
Adds the full publish → Redis → consume pipeline for API Products, aligned with existing distributed synchronizers.
```
Primary (DB sync)             Redis                          Replica
_________________             _____                          _______

ApiProductSynchronizer —deploy→ ApiProductDeployer             DistributedApiProductSynchronizer
                              .doAfterDeployment()   —→      (reads Redis)
                              .distributeIfNeeded()          ApiProductDeployer
``` 

Area | Change
-- | --
DistributedEventType | Add API_PRODUCT enum value
ApiProductMapper (distributed) | Bidirectional mapping: ApiProductReactorDeployable ↔ DistributedEvent (payload = serialized ReactableApiProduct + subscribable plan IDs)
DefaultDistributedSyncService | distributeIfNeeded(ApiProductReactorDeployable) — primary writes product events to Redis
DistributedApiProductSynchronizer | Replica-side consumer; extends AbstractDistributedSynchronizer, reuses ApiProductDeployer
DistributedSyncConfiguration | Spring wiring for mapper + synchronizer
ApiProductDeployer | Publish side: calls distributedSyncService.distributeIfNeeded() after deploy/undeploySharding filter: DEPLOY only distributed when product tags match gateway tags; UNDEPLOY always distributedReplica cleanup: when a tag-ineligible DEPLOY arrives (e.g. tag removed from product), publish UNDEPLOY to Redis so replicas drop stale state
DeployerFactory | Injects GatewayConfiguration into ApiProductDeployer for sharding checks

Please note: All the test scenarios with SS are attached in the mentioned jira ticket

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/distributed/mapper/ApiProductMapper.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/distributed/mapper/ApiProductMapper.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gateway.services.sync.process.distributed.mapper;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.gravitee.definition.model.v4.plan.Plan;
+import io.gravitee.gateway.handlers.api.ReactableApiProduct;
+import io.gravitee.gateway.services.sync.process.common.model.SyncAction;
+import io.gravitee.gateway.services.sync.process.repository.synchronizer.apiproduct.ApiProductReactorDeployable;
+import io.gravitee.repository.distributedsync.model.DistributedEvent;
+import io.gravitee.repository.distributedsync.model.DistributedEventType;
+import io.reactivex.rxjava3.core.Maybe;
+import java.util.Date;
+import java.util.List;
+import java.util.stream.Collectors;
+import lombok.CustomLog;
+import lombok.RequiredArgsConstructor;
+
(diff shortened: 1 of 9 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
```diff
@@ -103,7 +107,10 @@ private void unregisterApiProductPlans(ApiProductReactorDeployable deployable) {
 
     @Override
     public Completable doAfterDeployment(ApiProductReactorDeployable deployable) {
-        return distributedSyncService.distributeIfNeeded(deployable);
+        if (deployable.syncAction() == SyncAction.DEPLOY && shouldPublishUndeployForTagMismatch(deployable)) {
```
arpzilla-dev: Gating distribution on the primary's sharding tags looks risky. Tags are per-node, so with primary {public} / replica {internal} / product {internal}, the primary publishes UNDEPLOY instead of DEPLOY and the replica that should serve the product never gets it.

ApiDeployer distributes unconditionally and lets each node filter — and ApiProductManagerImpl.register() already does that per-node (undeploys locally on tag mismatch). Suggest mirroring it: just distributeIfNeeded(deployable) and drop the sharding helpers. Correct and smaller. WDYT?

Do correct me, if i understood anything wrong also. 

(Doesn't surface with a tagless primary, but tagged primaries are normal in sharded setups — exactly what this PR targets.)

"That said, if there's a reason products need to differ from APIs here, let me know — I may be missing context." 

Happy to discuss.

pragsty: Done. ApiProductDeployer now always calls distributeIfNeeded() (like ApiDeployer); tag filtering remains in ApiProductManagerImpl.

arpzilla-dev:
@pragsty checks looks great,
just dont forget to add proper labels, [4-11 seems missing]

arpzilla-dev (approved):

PR #18383, opened 2026-07-02
feat(repository): persist idpClaims on the user record (APIM-13955)

wbabyte:
## What

Second story of epic **APIM-13949** — *Tenant/User context propagation in DCR via IdP claims*. Adds repository-layer storage for the IdP claims captured for a user at login:

- `User.idpClaims` (`Map<String,String>`) — claims persisted on the user record, later injected into DCR requests.

Optional / nullable — existing users are unaffected until claims are persisted (only social-IdP logins with a configured whitelist ever write it).

## How

- **Mongo**: field added to `UserMongo`; `MongoUserRepository.update()` setter wired (update is field-by-field manual).
- **JDBC**: `idp_claims` nclob column via the `JdbcObjectMapper` serialize/deserialize converter (covers all query paths, no N+1).
- **Liquibase**: `v4_13_0/03_add_idp_claims_to_users.yml`, registered in `master.yml`. No Mongo upgrader / backfill needed (schemaless; absent == feature-off default; all reads null-safe).

## Tests

`UserRepositoryTest` round-trip (create **and update**), green on MongoDB and JDBC (Testcontainers).

## Notes

- Ref: **APIM-13955** (epic **APIM-13949**).
- Builds on #18358 (merged). Next PRs (REST config, login persistence, DCR injection, docs) follow.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/User.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/User.java
@@ -16,6 +16,8 @@
 package io.gravitee.repository.management.model;
 
 import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
@@ -71,6 +73,11 @@ public class User {
     private Boolean newsletterSubscribed;
     private Boolean isServiceAccount;
 
+    /**
+     * IdP claims captured at login (whitelisted on the identity provider), for later use such as DCR injection.
+     */
+    private Map<String, String> idpClaims;
+
     public User(User cloned) {
         this.id = cloned.id;
         this.organizationId = cloned.organizationId;
@@ -89,6 +96,7 @@ public class User {
         this.firstConnectionAt = cloned.firstConnectionAt;
         this.newsletterSubscribed = cloned.newsletterSubscribed;
         this.isServiceAccount = cloned.isServiceAccount;
+        this.idpClaims = cloned.idpClaims != null ? new HashMap<>(cloned.idpClaims) : null;
     }
 
     @Override
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcUserRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcUserRepository.java
@@ -17,6 +17,9 @@ package io.gravitee.repository.jdbc.management;
 
 import static io.gravitee.repository.jdbc.common.AbstractJdbcRepositoryConfiguration.escapeReservedWord;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.gravitee.common.data.domain.Page;
 import io.gravitee.repository.exceptions.TechnicalException;
 import io.gravitee.repository.jdbc.orm.JdbcObjectMapper;
@@ -25,6 +28,7 @@ import io.gravitee.repository.management.api.search.Pageable;
 import io.gravitee.repository.management.api.search.UserCriteria;
 import io.gravitee.repository.management.model.User;
 import io.gravitee.repository.management.model.UserStatus;
+import java.io.IOException;
 import java.sql.PreparedStatement;
 import java.sql.Types;
 import java.util.Arrays;
@@ -33,6 +37,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -46,8 +51,37 @@ public class JdbcUserRepository extends JdbcAbstractCrudRepository<User, String>
 
     private static final String STATUS_FIELD = "status";
 
+    private static final ObjectMapper JSON = new ObjectMapper();
+
     JdbcUserRepository(@Value("${management.jdbc.prefix:}") String tablePrefix) {
         super(tablePrefix, "users");
+    }
+
+    private static String serializeIdpClaims(Map claims) {
+        if (claims == null) {
+            return null;
+        }
+        try {
+            return JSON.writeValueAsString(claims);
+        } catch (JsonProcessingException e) {
+            // The ORM's per-column writer swallows and mislabels a thrown exception ("Failed to invoke getter"),
+            // so log an accurate message here instead. Serializing a String map is effectively never expected to fail.
+            log.warn("Failed to serialize user idp_claims; storing null", e);
+            return null;
+        }
+    }
+
+    private static Map deserializeIdpClaims(String json) {
+        if (json == null || json.isEmpty()) {
+            return null;
+        }
+        try {
+            return JSON.readValue(json, new TypeReference<Map<String, String>>() {});
+        } catch (IOException e) {
+            // Let the ORM's setFromResultSet handler deal with it: it logs with the partially-built row (user id
+            // already set) and skips the column, so a corrupt/legacy value degrades to null without blocking the read.
+            throw new IllegalArgumentException("Failed to deserialize user idp_claims", e);
+        }
     }
 
     @Override
@@ -70,6 +104,13 @@ public class JdbcUserRepository extends JdbcAbstractCrudRepository<User, String>
             .addColumn("first_connection_at", Types.TIMESTAMP, Date.class)
             .addColumn("newsletter_subscribed", Types.BOOLEAN, Boolean.class)
             .addColumn("is_service_account", Types.BOOLEAN, Boolean.class)
+            .addColumn(
+                "idp_claims",
+                Types.NCLOB,
+                Map.class,
+                JdbcUserRepository::serializeIdpClaims,
+                JdbcUserRepository::deserializeIdpClaims
+            )
             .build();
     }
 
--- a/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoUserRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoUserRepository.java
@@ -24,8 +24,11 @@ import io.gravitee.repository.management.model.User;
 import io.gravitee.repository.mongodb.management.internal.model.UserMongo;
 import io.gravitee.repository.mongodb.management.internal.user.UserMongoRepository;
 import io.gravitee.repository.mongodb.management.mapper.GraviteeMapper;
+import java.util.Base64;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Pattern;
@@ -56,6 +59,51 @@ public class MongoUserRepository implements UserRepository {
         this.mapper = mapper;
     }
 
+    /**
+     * Maps a domain user to its Mongo document, Base64-encoding the {@code idpClaims} keys. IdP claim names are
+     * externally controlled and OIDC namespaced claims contain dots, which MongoDB rejects as map keys; encoding
+     * the keys mirrors how {@code MongoIdentityProviderRepository} stores its group/role mapping keys.
+     */
+    private UserMongo toMongo(User user) {
+        UserMongo userMongo = mapper.map(user);
+        if (userMongo != null) {
+            userMongo.setIdpClaims(encodeClaimKeys(user.getIdpClaims()));
+        }
+        return userMongo;
+    }
+
+    private User toModel(UserMongo userMongo) {
+        User user = mapper.map(userMongo);
+        if (user != null) {
+            try {
+                user.setIdpClaims(decodeClaimKeys(userMongo.getIdpClaims()));
+            } catch (IllegalArgumentException e) {
+                // A non-Base64 key (e.g. from an out-of-band edit/import) must not make the user unreadable.
+                log.warn("Failed to decode idp_claims keys for user {}; ignoring stored claims", userMongo.getId(), e);
+                user.setIdpClaims(null);
+            }
+        }
+        return user;
+    }
+
+    private static Map<String, String> encodeClaimKeys(Map<String, String> claims) {
+        if (claims == null) {
+            return null;
+        }
+        Map<String, String> encoded = new HashMap<>(claims.size());
+        claims.forEach((key, value) -> encoded.put(new String(Base64.getEncoder().encode(key.getBytes())), value));
+        return encoded;
+    }
+
+    private static Map<String, String> decodeClaimKeys(Map<String, String> claims) {
+        if (claims == null) {
+            return null;
+        }
+        Map<String, String> decoded = new HashMap<>(claims.size());
+        claims.forEach((key, value) -> decoded.put(new String(Base64.getDecoder().decode(key)), value));
+        return decoded;
+    }
+
     @Override
     public Optional<User> findBySource(String source, String sourceId, String organizationId) {
         log.debug("Find user by name source[{}] user[{}]", source, sourceId);
@@ -71,7 +119,7 @@ public class MongoUserRepository implements UserRepository {
             String escapedSourceId = escaper.matcher(sourceId).replaceAll("\\\\$1");
             user = internalUserRepo.findBySourceAndSourceIdIgnoreCase(source, escapedSourceId, organizationId);
         }
-        User res = mapper.map(user);
+        User res = toModel(user);
 
         return Optional.ofNullable(res);
     }
@@ -91,7 +139,7 @@ public class MongoUserRepository implements UserRepository {
             users = internalUserRepo.findByEmailIgnoreCase(email, organizationId);
         }
 
-        return users.stream().map(mapper::map).toList();
+        return users.stream().map(this::toModel).toList();
     }
 
     @Override
@@ -99,7 +147,7 @@ public class MongoUserRepository implements UserRepository {
         log.debug("Find user by identifiers user [{}]", ids);
 
         Set<UserMongo> usersMongo = internalUserRepo.findByIds(ids);
-        Set<User> users = mapper.mapUsers(usersMongo);
+        Set<User> users = usersMongo.stream().map(this::toModel).collect(Collectors.toSet());
 
         log.debug("Find user by identifiers user [{}] - Done", ids);
         return users;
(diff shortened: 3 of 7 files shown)


gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/User.java
```diff
@@ -71,6 +72,11 @@ public enum AuditEvent implements Audit.AuditEvent {
     private Boolean newsletterSubscribed;
     private Boolean isServiceAccount;
 
+    /**
+     * IdP claims captured at login (whitelisted on the identity provider), for later use such as DCR injection.
+     */
+    private Map<String, String> idpClaims;
```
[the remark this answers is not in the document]
wbabyte: fixed

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcUserRepository.java
```diff
@@ -46,10 +51,34 @@ public class JdbcUserRepository extends JdbcAbstractCrudRepository<User, String>
...
+    }
+
+    private static Map deserializeIdpClaims(String json) {
+        if (json == null || json.isEmpty()) {
+            return null;
+        }
+        try {
+            return JSON.readValue(json, new TypeReference<Map<String, String>>() {});
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Failed to deserialize user idp claims", e);
```
paulatulis: Jackson parse failure throws `IllegalArgumentException`, which propagates out of findById/read paths. A single corrupt/legacy idp_claims value would then break reading that user entirely (i.e., login). The sibling repo in the same epic, JdbcIdentityProviderRepository, handles the equivalent case with log.warn(...) + default instead of throwing. Consider matching that graceful degradation for consistency and to avoid a stored-data issue becoming a hard read failure.

wbabyte: agree, fixed!

gravitee-apim-repository/gravitee-apim-repository-test/src/test/java/io/gravitee/repository/management/UserRepositoryTest.java
```diff
@@ -74,6 +75,35 @@ public void createUserTest() throws Exception {
         assertEquals(user.getNewsletterSubscribed(), false, "Invalid saved user newsletter.");
     }
 
+    @Test
+    public void should_persist_and_read_idp_claims() throws Exception {
```
carlos-andres-osorio, line 79: This test only covers the non-null round-trip for `idpClaims` — no test asserts `getIdpClaims()` returns null for a user that was never given claims (e.g. an existing fixture user). Per the acceptance criteria the column must be nullable, but that path isn't exercised, so a regression that starts returning an empty map instead of null (or a deserializer bug on the null/empty path) would go undetected.

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/internal/model/UserMongo.java
```diff
@@ -70,4 +71,5 @@ public class UserMongo extends DeprecatedAuditable {
     private Date firstConnectionAt;
     private Boolean newsletterSubscribed;
     private Boolean isServiceAccount;
+    private Map<String, String> idpClaims;
```
carlos-andres-osorio, line 76: `idpClaims` is persisted as a raw `Map<String, String>` whose keys are externally controlled IdP claim names. Spring Data MongoDB rejects map keys containing a `.` with a `MappingException` unless `mapKeyDotReplacement` is configured, and no such configuration exists in this module. URI-namespaced OIDC custom claims (e.g. `https://example.com/roles`) contain dots, so whitelisting a claim like that will make every save of that user document throw at login time.

## Evidence
Verified with a real test against a live MongoDB Testcontainer, going through the actual repository entry point:

```java
@Test
public void should_persist_and_read_dotted_key_idp_claim() throws Exception {
    User user = new User();
    user.setId(USER_ID);
    user.setOrganizationId("DEFAULT");
    user.setCreatedAt(new Date());
    user.setUpdatedAt(user.getCreatedAt());
    user.setEmail("user-idp-claims-dotted@gravitee.io");
    user.setStatus(UserStatus.ACTIVE);
    user.setSource("oidc");
    user.setSourceId(USER_ID);
    user.setIdpClaims(Map.of("https://example.com/roles", "admin"));

    User created = userRepository.create(user);
    assertThat(created).as("create must not throw and must return the user").isNotNull();

    Optional<User> found = userRepository.findBySource("oidc", USER_ID, "DEFAULT");
    assertThat(found).as("saved user must be found").isPresent();
    assertThat(found.get().getIdpClaims())
        .as("dotted-key idp claim must round-trip intact")
        .containsEntry("https://example.com/roles", "admin");
}
```

Result: `create()` threw `TechnicalException` wrapping:
```
org.springframework.data.mapping.MappingException: Map key https://example.com/roles contains dots but no replacement was configured; Make sure map keys don't contain dots in the first place or configure an appropriate replacement
```

The save aborted before the round-trip assertion was ever reached. Verified automatically during triage.

carlos-andres-osorio, line 76: Should `idpClaims` be encrypted at rest? Every other identity field sourced from the IdP on this document (`email`, `firstname`, `lastname`, `sourceId`, `picture`) is annotated `@ExplicitEncrypted`, but `idpClaims` is stored in plaintext. Whitelisted claims can carry PII (the branch's own test uses `org_id`/`tenant`), so on deployments that rely on MongoDB field-level encryption for user PII, this new field would bypass that protection. Worth noting there's no existing precedent in this codebase for encrypting a `Map`-typed field — every current `@ExplicitEncrypted` usage is on a plain `String` field — so this may need a small design decision rather than a one-line annotation.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcUserRepository.java
```diff
@@ -46,10 +51,36 @@ public class JdbcUserRepository extends JdbcAbstractCrudRepository<User, String>
...
+            return null;
+        }
+        try {
+            return JSON.writeValueAsString(claims);
+        } catch (JsonProcessingException e) {
+            throw new IllegalArgumentException("Failed to serialize user idp claims", e);
+        }
+    }
+
+    private static Map deserializeIdpClaims(String json) {
```
carlos-andres-osorio, line 74: `deserializeIdpClaims` catches `IOException` on a corrupt/legacy `idp_claims` value and returns `null` (logging a warning) rather than propagating — deliberately, so a bad stored value doesn't make the user record unreadable. That branch isn't covered by any test though: nothing writes a malformed value and asserts the graceful-degradation path still works. If this ever regressed to rethrowing, a corrupt legacy value would turn into a login-blocking failure with no test catching it first.

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoUserRepository.java
```diff
@@ -197,6 +197,7 @@ public User update(User user) throws TechnicalException {
         userMongo.setFirstConnectionAt(user.getFirstConnectionAt());
         userMongo.setNewsletterSubscribed(user.getNewsletterSubscribed());
         userMongo.setIsServiceAccount(user.getIsServiceAccount());
+        userMongo.setIdpClaims(user.getIdpClaims());
```
carlos-andres-osorio: No test covers clearing a previously-stored `idpClaims` value by updating with `idpClaims = null` — `should_persist_and_read_idp_claims` only updates to a different non-null map. If the clear path silently kept the old value, stale claims would persist and later be read as if still current, with no error signal.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcUserRepository.java
carlos-andres-osorio, line 74: Should the corrupt/legacy `idp_claims` deserialization failure surface more than a `log.warn`? Right now `deserializeIdpClaims` catches `IOException` and returns a plain `null`, which is indistinguishable from "user never had claims." This mirrors the existing precedent in `JdbcIdentityProviderRepository` for `persisted_claims_whitelist`, so it may well be intentional — there's no downstream consumer of `getIdpClaims()` yet in this PR's scope to actually need the distinction. Just flagging in case a future consumer (e.g. DCR injection) would benefit from a stronger signal here.

carlos-andres-osorio:
> **gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/internal/model/UserMongo.java:38**

The class-level Lombok `@ToString` on `UserMongo` now includes `idpClaims`, so any log statement or exception message that prints this object will emit the raw claim values — which can carry PII — to logs. None of the other fields on this class currently use `@ToString.Exclude`, so there's no existing precedent to lean on for suppressing it.

wbabyte:
Thanks @carlos-andres-osorio — thorough review. Addressed:

**Fixed (code):**
- **Dotted map keys break Mongo save** (verified `MappingException` on `https://…` claim keys) — `MongoUserRepository` now Base64-encodes `idpClaims` keys on write and decodes on read via `toMongo`/`toModel`, mirroring how `MongoIdentityProviderRepository` stores its group/role-mapping keys. Added a round-trip test with a dotted key (`https://acme.example/org`); green on Mongo **and** JDBC (JDBC stores JSON so dots were already fine).
- **`@ToString` PII leak** — added `@ToString.Exclude` on `UserMongo.idpClaims` so raw claim values don't reach logs.
- **No null-path coverage** — added `should_return_null_idp_claims_when_user_has_none` (asserts `null`, not empty map).
- **No clear-via-null coverage** — `should_persist_and_read_idp_claims` now also updates `idpClaims → null` and asserts it's cleared.

**Design decisions (no change), happy to revisit:**
- **Encryption at rest** — intentional for this epic: the persistence whitelist is the control (admins whitelist only non-sensitive org/tenant ids), `idpClaims` is never exposed via API, and there's no precedent for `@ExplicitEncrypted` on a `Map`. Documented as a security note. If field-level encryption of PII claims is required, that's a follow-up design (likely storing as an encrypted serialized string).
- **Graceful-degradation on corrupt value / stronger-than-`log.warn` signal** — kept as `log.warn` + `null`, matching the existing `JdbcIdentityProviderRepository` precedent (and Paula's earlier request to not hard-fail reads). No consumer needs the corrupt-vs-absent distinction yet; agreed it's worth revisiting when DCR injection lands. The malformed-value path isn't covered by the shared (backend-agnostic) repo test since Mongo has no "malformed JSON" equivalent — can add a JDBC-specific test in a follow-up if you'd like.

Re-ran `UserRepositoryTest` on Mongo + JDBC → 21/21 both.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcUserRepository.java
```diff
@@ -46,10 +51,36 @@ public class JdbcUserRepository extends JdbcAbstractCrudRepository<User, String>
...
     }
 
+    private static String serializeIdpClaims(Map claims) {
+        if (claims == null) {
+            return null;
+        }
+        try {
+            return JSON.writeValueAsString(claims);
+        } catch (JsonProcessingException e) {
+            throw new IllegalArgumentException("Failed to serialize user idp claims", e);
```
carlos-andres-osorio: serializeIdpClaims throws an IllegalArgumentException when Jackson fails to serialize the idpClaims map, but this call happens inside JdbcObjectMapper.setStatementValues's per-column try block (JdbcObjectMapper.java:415), whose `catch (Exception ex)` is a catch-all that logs `"Failed to invoke getter {} on {} : "` and moves on to the next column instead of aborting the write.

The exception is swallowed, the `idp_claims` parameter is left unset on the statement, and the log message mislabels a serialization failure as a getter-invocation failure — sending anyone debugging this down the wrong path.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcUserRepository.java
```diff
@@ -46,10 +51,36 @@ public class JdbcUserRepository extends JdbcAbstractCrudRepository<User, String>
...
+
+    private static Map deserializeIdpClaims(String json) {
+        if (json == null || json.isEmpty()) {
+            return null;
+        }
+        try {
+            return JSON.readValue(json, new TypeReference<Map<String, String>>() {});
+        } catch (IOException e) {
+            // Degrade gracefully: a corrupt/legacy value must not make the user unreadable (which would block login).
+            log.warn("Failed to deserialize user idp_claims; ignoring the stored value", e);
```
carlos-andres-osorio: deserializeIdpClaims catches `IOException` on a corrupt/legacy `idp_claims` value, logs a WARN, and returns null — a deliberate design so a bad stored value doesn't block that user's login. However the WARN log carries no user/row identifying context, so an operator has no way to tell which user's record needs remediation from the log alone.

`JdbcObjectMapper.setFromResultSet`'s outer catch-all (line 322) does have access to the partially-built `User` item (with `id` already set, since `id` is processed before `idp_claims` in `buildOrm()`) and would log it alongside any exception — but that path never fires here because the `IOException` is swallowed inside `deserializeIdpClaims` first.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcUserRepository.java
```diff
@@ -46,10 +51,36 @@ public class JdbcUserRepository extends JdbcAbstractCrudRepository<User, String>
...
+            return null;
```
carlos-andres-osorio: Does the Jackson deserialization exception logged here ever embed a snippet of the corrupt `idp_claims` payload (potential PII) into the logs?

This module resolves jackson-core/jackson-databind 2.21.2, and since jackson-core 2.16.0 the default for `StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` is `false`, which should redact source content from exception messages by default — so this may already be safe, and the `ObjectMapper` here (`new ObjectMapper()`, unconfigured) would inherit that default. Worth double-checking against the actual exception message for this exact call shape just to be sure?

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoUserRepository.java
```diff
@@ -197,8 +239,9 @@ public User update(User user) throws TechnicalException {
         userMongo.setFirstConnectionAt(user.getFirstConnectionAt());
         userMongo.setNewsletterSubscribed(user.getNewsletterSubscribed());
         userMongo.setIsServiceAccount(user.getIsServiceAccount());
+        userMongo.setIdpClaims(encodeClaimKeys(user.getIdpClaims()));
         UserMongo userUpdated = internalUserRepo.save(userMongo);
-        return mapper.map(userUpdated);
+        return toModel(userUpdated);
     }
```
carlos-andres-osorio, line 252: decodeClaimKeys calls `Base64.getDecoder().decode(key)` with no try/catch, so any non-Base64 key in a stored `idpClaims` map throws `IllegalArgumentException` and propagates out of `toModel`/`find*` — unlike the JDBC sibling (`deserializeIdpClaims`), which explicitly catches the equivalent corrupt-value case and degrades gracefully so a bad value never blocks a user read.

A Mongo document whose `idpClaims` map somehow contains a non-Base64 key (e.g. edited/restored/imported outside this repository's own `encodeClaimKeys` write path) would cause every read of that user — `findById`, `findBySource`, `findByEmail`, `search`, `findAll` — to fail entirely rather than degrade gracefully like the JDBC path does for the equivalent scenario.

wbabyte:
Thanks again @carlos-andres-osorio — all four addressed, and rebased onto latest `master`.

**JDBC serialize — swallowed & mislabeled by the ORM write catch (#1):** you're right that `setStatementValues`' per-column catch-all turns a thrown exception into a misleading `"Failed to invoke getter"`. Changed `serializeIdpClaims` to log an accurate `WARN` and return `null` rather than throw, so there's no misleading ORM log. (Serializing a `Map<String,String>` is effectively never expected to fail; this just makes the unreachable case honest.)

**JDBC deserialize — WARN lacked row context (#2):** removed the context-less local `WARN` and now let it propagate (as unchecked). As you noted, `setFromResultSet`'s handler logs with the partially-built row — `id` is set before `idp_claims` in `buildOrm()` — and skips the column, so we still degrade to `null` without blocking the read, now *with* the user id in the log.

**PII in the deserialization exception (#3):** confirmed safe. jackson-core 2.16+ defaults `INCLUDE_SOURCE_IN_LOCATION=false` (this unconfigured `new ObjectMapper()` inherits it), and the ORM logs `ex.getMessage()` of our own `IllegalArgumentException` (not the Jackson cause). Also verified `User.toString()` does not include `idpClaims`, so the ORM's `item` log can't leak claim values.

**Mongo `decodeClaimKeys` — no guard, breaks all reads on a non-Base64 key (#4):** `toModel` now catches `IllegalArgumentException` from the Base64 decode, logs with the user id, and degrades to `null` — matching the JDBC graceful path, so an out-of-band-edited document can no longer make a user unreadable.

**Conflict:** rebased onto `master`; renumbered the changelog `03 → 06` (upstream took `03`/`04`/`05` in the interim) and kept all upstream includes. `UserRepositoryTest` green on Mongo + JDBC (21/21).

paulatulis (approved):
Non-blocking:
- GraviteeMapper.mapUsers() is now dead code — this PR removed its only caller. (mapUserList was already dead, so it's a pre-existing pattern.)
- GraviteeMapper.map(UserMongo) remains public and returns encoded keys. A future caller using it directly would silently get Base64 claim names. An @Mapping(target="idpClaims", ignore=true) would make toModel/toMongo the only viable path. Speculative, but a real trap for the next PR in this epic (DCR injection).
- key.getBytes() uses the platform default charset rather than UTF-8. Technically latent for non-ASCII claim names — but it copies the existing precedent exactly, and claim names are ASCII in practice.
- The corrupt-value catch blocks have no tests. Hard to test from the shared backend-agnostic UserRepositoryTest without writing malformed data out-of-band, which is likely why it was skipped.

PR #18387, opened 2026-07-02
fix(http-proxy): release pool slot when client aborts the request

wbabyte:
## Issue

https://gravitee.atlassian.net/browse/APIM-14605

## Description

On a v4 HTTP proxy API, a client abort never reached the upstream request handled by the http-proxy endpoint connector: disposing the reactive chain only drops the rx bridge's observer, so an aborted request kept holding (or was later granted) its connection-pool slot until the request timeout fired. Aborted requests shed no load — under sustained abort/retry traffic against a slow backend the endpoint pool stays collapsed even after the backend recovers.

This PR propagates disposal to the upstream request in the window between pool acquisition and the response head (post-head cancellation was already handled by the response chunks' `doOnCancel`):

- the acquired `HttpClientRequest` is tracked and `reset()` on disposal, releasing its pool slot immediately;
- acquisition goes through the client's `Future` directly, so a connection granted *after* disposal (client aborted while queued in the pool) is reset on the spot instead of being held idle until the timeout;
- at the response head the tracking is dropped and the existing chunk-level cancellation takes over (`reset()` is a no-op on a completed stream, making the disposal race benign).

`GrpcConnector` inherits the fix. `WebSocketConnector` has the same flaw and is left as a follow-up.

## Additional context

New acceptance tests (`ClientAbortPropagationV4IntegrationTest`, single-connection pool + raw TCP aborts) fail on master and pass with this change:
- a request aborted while queued must not consume the slot freed by the previous request (and must never be dispatched to the backend);
- a request aborted while in flight must release its connection immediately.

Verified no regression on the connect-vs-read timeout classification (APIM-12769): `ConnectionFailureV4IntegrationTest` and all `HttpRequestTimeout*` suites green (28 tests), plus the http-proxy plugin unit tests.

Related: gravitee-io/issues#10984 (stale keep-alive reuse — the same `keepAlive: false` workaround masks both defects).

### DIFF
--- a/gravitee-apim-integration-tests/src/test/java/io/gravitee/apim/integration/tests/http/ClientAbortPropagationV4IntegrationTest.java
+++ b/gravitee-apim-integration-tests/src/test/java/io/gravitee/apim/integration/tests/http/ClientAbortPropagationV4IntegrationTest.java
@@ -1,0 +1,159 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


paulatulis (approved):
Only question is that the ticket says 4.10 is where the issue starts, but this has a 4.9 label. Is it necessary?

wbabyte:
yes @paulatulis 4.9 is affected as well

PR #18411, opened 2026-07-03
feat(home): handle missing module license on the homepage (GMA-795)

Okhelifi:
## Issue

https://gravitee.atlassian.net/browse/GMA-795

### What
Wires the shared license contract into the Gamma host and uses it on the homepage: modules that are not covered by the current license are shown in a locked state instead of a dead card — a non-clickable card with a clear "Upgrade to access" CTA that opens an upgrade dialog.

Bumps @gravitee/gamma-modules-sdk to the pre-release that ships the license module.

### Why
A Gamma installation (cloud or self-hosted) can run with a restricted license that excludes some module packs. Those modules are absent from GET /organizations/{orgId}/modules, so their homepage cards were rendered inert with no explanation. Users now get a clear "upgrade needed" path, aligned with the classic console. Relates to GMA-795 (homepage) and lays the plumbing for GMA-901 (feature-level gating inside modules).

https://github.com/user-attachments/assets/81d0c230-9b71-4bc0-a9d6-6a6ba0e80ced

### Small fixes:
- AIM redirection is broken on home page as the route has changes recently, here we align the redirection with the new Paths
- Handle count failure gracefully on home page cards

### Linked PR
- https://github.com/gravitee-io/gravitee-gamma-modules-sdk/pull/22

### DIFF
--- a/gravitee-gamma/gravitee-gamma-control-plane-webui/src/bootstrap-initialize.spec.ts
+++ b/gravitee-gamma/gravitee-gamma-control-plane-webui/src/bootstrap-initialize.spec.ts
@@ -13,18 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import { licenseService } from '@gravitee/gamma-modules-sdk';
+
 import { runApplicationBootstrap } from './bootstrap-initialize';
 import { useAuthStore } from './features/auth/auth.store';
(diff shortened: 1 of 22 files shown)


Okhelifi:
The first gemini review is based on master not 4.12.x so ignored for now

gcusnieux (commented):
LGTM

gcusnieux (approved):

PR #18417, opened 2026-07-03
fix: redis resources cannot be disabled

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-14000

## Description
When a cache-redis resource was disabled in the Console (enabled: false) and the API was redeployed, the gateway still loaded, initialized, and used the resource. The Cache policy continued to read/write Redis as if the resource were still enabled.

**Root cause**
On v2 APIs with execution_mode: v3, resources are managed by ResourceManagerImpl, which loaded all declared resources and ignored the enabled flag. v4 APIs already filter disabled resources via DefaultResourceManager.

**Solution**
Filter resources with Resource::isEnabled before loading them in ResourceManagerImpl.initialize(), matching v4 behavior.

Also updated the unit test to assert that disabled resources are skipped and not registered in the resource manager.

**Before **Fix:****
image

**After fix:**
image

image


pragsty:
As per alignment with the Architecture team and the recent revert on APIM-12214: Revert - disabled resource has no effect [10831]
changing this behavior introduces a breaking change for enterprise customers who actively rely on the resource in a disabled state on their v2 APIs. We cannot proceed with the current solution.

PR #18441, opened 2026-07-06
feat(rest-api): apply branded sender From/Subject at email send time

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14184
## Description

What
Implements S2 of the branded-sender epic (APIM-14125): outbound subscription/notification emails use a per-tenant branded From/Subject when a recipient's domain matches a configured rule, falling back to the default EMAIL_FROM/EMAIL_SUBJECT otherwise. This is the runtime behavior that makes the S1 data model + S3/S4 config UI take effect.

How
- BrandedSenders.resolve(rawParamValue, recipientEmail) (model) — parses the stored email.branded_senders config, extracts the recipient domain (handles bare user@x.com and Name <user@x.com>), and returns the first domain-matching config (case-insensitive, whitespace-tolerant).
- EmailServiceImpl — reads EMAIL_BRANDED_SENDERS alongside the other mail params and, per the ticket's multi-recipient AC (option a), groups every recipient (to + bcc + copyToSender) by its resolved (From, Subject) and sends one MimeMessage per group.

Behavior / decisions
- Per-recipient grouping — recipients on different branded domains each get their own sender identity in separate SMTP sends. The common case (no config, or all recipients resolving to the same sender) yields a single group → single message, byte-identical to today — no regression.
- Matching covers BCC recipients too — important because EmailNotifierServiceImpl routes 2+ recipients to BCC with an empty to.
- Exact domain match (each full domain is stored separately); case-insensitive.
- Explicit caller from (rare, deliberate) → no branding (From and Subject both left as today). Branding applies only on the default-sender path — which is the notification path this epic targets.
- Self-send (DEFAULT_MAIL_TO) → unbranded, as today.
- Purely additive: no match / blank / malformed config → unchanged default behavior.

Tests
- Model (EmailBrandedSendersTest): +7 resolve cases (match, case-insensitivity, personal-name form, no-match, first-match-wins, blank/malformed, no-domain).
- Service (EmailServiceTest): branded applies From+Subject; no-match keeps defaults; explicit-from → unbranded; two different-domain recipients → two sends; multi-recipient BCC gets branded; all-same-config → single send.

Depends on: S1 (data model, merged). Related: S3/S4 config UI (PR #18365).
## Additional context

Documentation notes

Feature: domain-matched branded senders (send-time)

Outbound notification emails can now use a per-tenant From and Subject when the recipient's email domain matches a configured rule. Rules are stored in the email.branded_senders setting at the environment/organization scope (the configuration UI ships in #18365).

- Matched branding: when a recipient's domain matches a rule, that rule's from and subject are used for the message to that recipient. Example: with a rule for airtel.com → noreply@airtel.com / [Airtel] %s, a notification to alice@airtel.com is sent from noreply@airtel.com with subject [Airtel] Subscription approved.
- Multi-recipient splitting: if a single notification's recipients span multiple branded domains, it is now delivered as multiple separate emails — one per matched (From, Subject) group — instead of a single message. Operators reading SMTP/notification logs should expect more than one send for one notification in this case. When all recipients resolve to the same sender (the common case, including all "no match → default"), it remains a single send.
- Fallback (additive & safe): no match, or a blank/malformed configuration, falls through to today's default EMAIL_FROM / EMAIL_SUBJECT — behavior is unchanged when the feature is unconfigured.
- Matching rules: domain match is case-insensitive and whitespace-tolerant, uses an exact domain match (no subdomain wildcards), and is first-match-wins if more than one rule lists the same domain.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
@@ -22,6 +22,8 @@ import io.gravitee.rest.api.model.parameters.Key;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
+import java.util.Optional;
 import lombok.CustomLog;
 
 /**
@@ -32,7 +34,7 @@ import lombok.CustomLog;
  * @author GraviteeSource Team
  */
 @CustomLog
-final class BrandedSenders {
+public final class BrandedSenders {
 
     private static final ObjectMapper MAPPER = new ObjectMapper();
 
@@ -50,7 +52,7 @@ final class BrandedSenders {
      * break the whole settings response. Invalid input is rejected loudly on the write path instead
      * (see {@link #write(List)}).
      */
-    static List<BrandedSenderConfig> parse(String raw) {
+    public static List<BrandedSenderConfig> parse(String raw) {
         if (raw == null || raw.isBlank()) {
             return new ArrayList<>();
         }
@@ -63,6 +65,77 @@ final class BrandedSenders {
             log.error("Ignoring malformed '{}' configuration [{}]; returning an empty list", Key.EMAIL_BRANDED_SENDERS.key(), raw, e);
             return new ArrayList<>();
         }
+    }
+
+    /**
+     * Resolves the branded-sender configuration to apply for an outbound email by matching the recipient's
+     * domain (the part after {@code '@'}, compared case-insensitively) against each configuration's domains,
+     * returning the first match. Returns {@link Optional#empty()} when the stored value is blank/malformed,
+     * the recipient has no parseable domain, or no configuration matches — in which case the caller keeps the
+     * default {@code EMAIL_FROM} / {@code EMAIL_SUBJECT}. When more than one configuration lists the same
+     * domain, the first is used (the console UI rejects cross-configuration duplicates, but a hand-edited
+     * value could still contain them).
+     */
+    public static Optional<BrandedSenderConfig> resolve(String raw, String recipientEmail) {
+        return match(parse(raw), recipientEmail);
+    }
+
+    /**
+     * Matches an already-parsed configuration list against a recipient, so a caller sending to many recipients
+     * can {@link #parse(String)} once and match per recipient rather than re-deserialising the value each time.
+     * Same semantics as {@link #resolve(String, String)}.
+     */
+    public static Optional<BrandedSenderConfig> match(List<BrandedSenderConfig> configs, String recipientEmail) {
+        final String domain = extractDomain(recipientEmail);
+        if (domain == null) {
+            return Optional.empty();
+        }
+        return configs
+            .stream()
+            .filter(config -> matchesDomain(config, domain))
+            .findFirst();
+    }
+
+    private static boolean matchesDomain(BrandedSenderConfig config, String domain) {
+        if (config == null || config.getDomains() == null) {
+            return false;
+        }
+        return config.getDomains().stream().filter(Objects::nonNull).map(BrandedSenders::normalizeDomain).anyMatch(domain::equals);
+    }
+
+    /**
+     * Extracts the lower-cased domain from a recipient address, accepting both a bare address
+     * ({@code user@example.com}) and the personal-name form ({@code Name <user@example.com>}). Returns
+     * {@code null} when no domain can be extracted, or when the string is a multi-address list
+     * ({@code a@x.com, b@y.com}) — matching such a value on a single domain would brand every address for
+     * one tenant, so it falls through to the default sender instead.
+     */
+    private static String extractDomain(String recipientEmail) {
+        if (recipientEmail == null) {
+            return null;
+        }
+        final String trimmed = recipientEmail.trim();
+        // A single recipient carries exactly one '@'. Zero means there is no domain to match; more than one
+        // means the value is really an address list — either the bare form ("a@x.com, b@y.com") or the
+        // personal-name form ("Jane <jane@x.com>, John <john@y.com>"). Matching such a value on a single
+        // domain would brand every address for one tenant, so it falls through to the default sender instead.
+        // This count is taken on the whole value (before any bracket stripping) so a trailing "Name <addr>"
+        // pair cannot hide the earlier addresses from the check.
+        final int firstAt = trimmed.indexOf('@');
+        if (firstAt < 0 || firstAt != trimmed.lastIndexOf('@')) {
+            return null;
+        }
+        String address = trimmed;
+        final int lt = address.lastIndexOf('<');
+        final int gt = address.lastIndexOf('>');
+        if (lt >= 0 && gt > lt) {
+            address = address.substring(lt + 1, gt);
+        }
+        final int at = address.lastIndexOf('@');
+        if (at < 0 || at == address.length() - 1) {
+            return null;
+        }
+        return normalizeDomain(address.substring(at + 1));
     }
 
     /**
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/EmailServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/EmailServiceImpl.java
@@ -15,15 +15,16 @@
  */
 package io.gravitee.rest.api.service.impl;
 
+import static io.gravitee.rest.api.model.parameters.Key.EMAIL_BRANDED_SENDERS;
 import static io.gravitee.rest.api.model.parameters.Key.EMAIL_ENABLED;
 import static io.gravitee.rest.api.model.parameters.Key.EMAIL_FROM;
 import static io.gravitee.rest.api.model.parameters.Key.EMAIL_SUBJECT;
-import static java.lang.String.format;
-import static java.util.Objects.isNull;
 import static org.apache.commons.lang3.StringUtils.isEmpty;
 
 import io.gravitee.rest.api.model.parameters.Key;
 import io.gravitee.rest.api.model.parameters.ParameterReferenceType;
+import io.gravitee.rest.api.model.settings.BrandedSenderConfig;
+import io.gravitee.rest.api.model.settings.BrandedSenders;
 import io.gravitee.rest.api.service.EmailNotification;
 import io.gravitee.rest.api.service.EmailService;
 import io.gravitee.rest.api.service.ParameterService;
@@ -38,10 +39,13 @@ import jakarta.mail.internet.InternetAddress;
 import java.io.File;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -67,6 +71,8 @@ import org.springframework.stereotype.Component;
 public class EmailServiceImpl extends TransactionalService implements EmailService {
 
     private static final Duration REGEX_TIMEOUT = Duration.ofSeconds(2);
+    private static final String SUBJECT_PLACEHOLDER = "%s";
+    private static final String POSITIONAL_SUBJECT_PLACEHOLDER = "%1$s";
 
     private final GraviteeJavaMailManager mailManager;
     private final NotificationTemplateService notificationTemplateService;
@@ -118,11 +124,6 @@ public class EmailServiceImpl extends TransactionalService implements EmailServi
         if (Boolean.parseBoolean(mailParameters.get(EMAIL_ENABLED)) && emailNotification.hasRecipients()) {
             try {
                 JavaMailSender mailSender = mailManager.getOrCreateMailSender(executionContext, referenceId, referenceType);
-                final MimeMessageHelper mailMessage = new MimeMessageHelper(
-                    mailSender.createMimeMessage(),
-                    true,
-                    StandardCharsets.UTF_8.name()
-                );
 
                 String emailSubject = notificationTemplateService.resolveTemplateWithParam(
                     executionContext.getOrganizationId(),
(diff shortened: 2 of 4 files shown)


carlos-andres-osorio:
👋 This PR is missing the `docs-needed`/`no-docs` label (required on gravitee-io PRs).

Note that under the current doc process, each story is expected to carry its own doc notes rather than relying on a single consolidated documentation story — so this is worth flagging on this PR specifically rather than pointing at APIM-14184's Jira "Documentation → S8" line item.

A few things from this PR that would likely need doc coverage if `docs-needed` is applied:
- **Domain-matched branding at send time**: outbound notification emails now use a per-tenant `From`/Subject when the recipient's email domain matches a configured `email.branded_senders` rule (the config UI itself is PR #18365).
- **Multi-recipient splitting behavior**: when a single notification's recipients span multiple branded domains, it now sends as multiple separate emails (one per matched domain) instead of one — worth calling out for anyone reading SMTP/notification logs and wondering why a single notification produced more than one message.
- **Fallback rule**: no match / blank / malformed config → today's default `EMAIL_FROM`/`EMAIL_SUBJECT`, unchanged — good to state explicitly so admins know the feature is additive/safe.
- **Matching rules**: case-insensitive, whitespace-tolerant, exact domain match, first-match-wins on duplicate domains.

Could you confirm whether `docs-needed` or `no-docs` applies here, and add the label?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/EmailServiceImpl.java
```diff
@@ -248,13 +208,166 @@ private static String extractMimeType(final String encoded) {
...
+        RecipientGroup group
+    ) throws Exception {
+        final MimeMessageHelper mailMessage = new MimeMessageHelper(mailSender.createMimeMessage(), true, StandardCharsets.UTF_8.name());
+
+        final InternetAddress configuredFrom = new InternetAddress(group.from());
+        if (isEmpty(configuredFrom.getPersonal())) {
+            if (isEmpty(emailNotification.getFromName())) {
+                mailMessage.setFrom(group.from());
+            } else {
+                mailMessage.setFrom(group.from(), emailNotification.getFromName());
```
carlos-andres-osorio, line 228: `sendGroupedMessage` branches on whether the resolved branded `From` address has no personal name AND the notification supplies a `fromName`: it combines them via `setFrom(group.from(), emailNotification.getFromName())` here. No test in `EmailServiceTest` ever sets `fromName` on the notification builder, so this branch — which stamps the caller's display name onto a (possibly branded) From address — never runs.

A regression here would silently send the wrong or missing sender display name with no error and no test catching it — recipients would see a generic/broken "From" line instead of the intended branded identity.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
```diff
@@ -65,6 +67,70 @@ static List<BrandedSenderConfig> parse(String raw) {
...
+    public static Optional<BrandedSenderConfig> resolve(String raw, String recipientEmail) {
+        return match(parse(raw), recipientEmail);
+    }
+
+    /**
+     * Matches an already-parsed configuration list against a recipient, so a caller sending to many recipients
+     * can {@link #parse(String)} once and match per recipient rather than re-deserialising the value each time.
+     * Same semantics as {@link #resolve(String, String)}.
+     */
+    public static Optional<BrandedSenderConfig> match(List<BrandedSenderConfig> configs, String recipientEmail) {
```
carlos-andres-osorio, line 88: Nit: no test drives a parsed config whose `domains` list is `null` through the `match` path. `parse` bypasses `write`'s validation, so a hand-edited/legacy raw value like `[{"from":"noreply@example.com"}]` (no `domains` key) reaches `matchesDomain` here.

To be clear, this is not currently a live bug — `matchesDomain()` already guards `config.getDomains() == null` and returns `false`, so this can't NPE today. The gap is that this safety is untested (every existing test builds its raw value via `write()`, which always supplies non-null domains lists), so a future refactor touching `matchesDomain` could silently drop the guard with nothing to catch it.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
```diff
@@ -65,6 +67,70 @@ static List<BrandedSenderConfig> parse(String raw) {
...
+     * one tenant, so it falls through to the default sender instead.
+     */
+    private static String extractDomain(String recipientEmail) {
+        if (recipientEmail == null) {
+            return null;
+        }
+        String address = recipientEmail.trim();
+        final int lt = address.lastIndexOf('<');
+        final int gt = address.lastIndexOf('>');
+        if (lt >= 0 && gt > lt) {
```
carlos-andres-osorio, line 131: `extractDomain`'s multi-address guard is broken for the personal-name form. It strips everything up to the LAST `<...>` bracket pair before checking for a comma. For `"Jane <jane@example.com>, John <john@example.org>"`, this slices out just `john@example.org` — the leading `"Jane <jane@example.com>, "` (and its comma) is discarded before the comma check ever runs — so the method returns `example.org` instead of `null`, silently branding the whole recipient group by the last recipient's domain. The class-level javadoc above explicitly promises multi-address safety, so this contradicts its own documented contract.

Today's only call site (`EmailServiceImpl.groupRecipient`) passes single array elements, so it isn't reachable via the current production path — but `resolve`/`match` are `public`, and any future caller forwarding a raw multi-recipient header value would hit this.

Also worth noting: a structurally identical flaw exists in `SenderAddressValidator.PERSONAL_NAME` (added in #18336/APIM-14181, guards `BrandedSenderConfig.from`) — same "match trailing bracket, check comma second" ordering. That one is benign in practice since it validates a single admin-configured field rather than a recipient list, but it's independent confirmation this parsing pattern is fragile — might be worth a shared, tested utility at some point.

## Evidence

Wrote and ran a real test against `BrandedSenders.resolve()` to confirm this reproduces:

```java
@Test
void should_return_empty_for_a_multi_address_personal_name_recipient_string() {
    var raw = BrandedSenders.write(
        List.of(BrandedSenderConfig.builder().domains(List.of("example.org")).from("noreply@example.org").build())
    );

    // Comma-joined personal-name list where only the LAST address's domain (example.org) matches the
    // config. It must not brand the whole recipient group by the last recipient's domain.
    assertThat(BrandedSenders.resolve(raw, "Jane Developer <jane@example.com>, John Support <john@example.org>")).isEmpty();
}
```

Result: **test failed** — `Expecting an empty Optional but was containing value: BrandedSenderConfig(domains=[example.org], from=noreply@example.org, subject=null)`.

(Verified interactively during review; the test was reverted from the working tree after capturing this evidence.)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/EmailServiceImpl.java
```diff
@@ -136,53 +136,13 @@ private void sendEmailNotification(
...
-
-                log.debug(
-                    "Sending an email to {} recipient(s)\nSubject: {}\nMessage: {}",
-                    emailNotification.recipientsCount(),
-                    emailSubject,
-                    html
-                );
-
-                mailSender.send(mailMessage.getMimeMessage());
             } catch (final Exception ex) {
```
carlos-andres-osorio, line 149: `sendEmailNotification` loops over `buildRecipientGroups(...)`, calling `sendGroupedMessage(...)` once per branded group inside a single try block. The enclosing `catch (final Exception ex)` wraps the whole loop and rethrows one generic `TechnicalManagementException` regardless of which iteration failed or how many prior groups already sent successfully — no logging of partial progress, no group-identifying context (From/domain/recipient count) attached anywhere.

If group 1 of 3 sends successfully and group 2 throws, the caller/on-call engineer sees a generic "email notification failed" with no way to tell some recipients were already branded-and-sent. A naive retry duplicates those sends; a false assumption that zero recipients were notified is equally possible.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/EmailServiceImpl.java
```diff
@@ -248,13 +208,166 @@ private static String extractMimeType(final String encoded) {
...
+            if (isEmpty(emailNotification.getFromName())) {
+                mailMessage.setFrom(group.from());
+            } else {
+                mailMessage.setFrom(group.from(), emailNotification.getFromName());
+            }
+        } else {
+            mailMessage.setFrom(configuredFrom);
+        }
+
+        if (!isEmpty(emailNotification.getReplyTo())) {
```
carlos-andres-osorio, line 234: The ReplyTo branch here (`if (!isEmpty(emailNotification.getReplyTo())) mailMessage.setReplyTo(...)`) is new code in this diff. The only test that sets `replyTo` (`should_send_the_copy_to_sender_with_the_default_identity_not_a_brand`) does so purely to trigger `copyToSender`'s Bcc grouping and asserts only From/To/Bcc — the ReplyTo header outcome itself is never asserted anywhere in the suite.

A regression that breaks ReplyTo (stops being set, or set on the wrong message in a multi-group send) would go completely undetected.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/EmailServiceImpl.java
```diff
@@ -248,13 +208,166 @@ private static String extractMimeType(final String encoded) {
...
+            sender = emailNotification.getReplyTo();
+        }
+        final boolean copyToSender = emailNotification.isCopyToSender() && sender != null;
+        final List<String> bcc = new ArrayList<>();
+        if (emailNotification.getBcc() != null) {
+            bcc.addAll(Arrays.asList(emailNotification.getBcc()));
+        }
+
+        final boolean selfSend = Arrays.equals(DEFAULT_MAIL_TO, emailNotification.getTo());
```
carlos-andres-osorio, line 280: The branding short-circuit here treats `DEFAULT_MAIL_TO` as a "self-send" and emits a single unbranded message, entirely skipping the branding match logic for the bcc list. But `DEFAULT_MAIL_TO` is the sentinel `MessageServiceImpl.send()` uses for its publisher→subscribers broadcast (`.to(DEFAULT_MAIL_TO).bcc(mails)`) — a genuine customer-facing notification whose bcc recipients are real external subscribers, potentially on branded domains. The javadoc above mislabels it a "self-send", which suggests this broadcast use wasn't in view when the short-circuit was written.

Result: the same customer/domain gets one notification type branded (hook notifications via `EmailNotifierService`) and another type unbranded (the message-to-subscribers broadcast) — inconsistent sender identity for the same recipient. This also seems to run counter to the ticket's stated intent that branding "applies uniformly across all outbound notification emails (consumer- and publisher-bound)" and that "BCC recipients group by the same match logic".

paulatulis: You're right, and the ticket AC backs it up ("applies uniformly … consumer- and publisher-bound"; "BCC recipients group by the same match logic"). Fixed: DEFAULT_MAIL_TO no longer short-circuits — the broadcast's BCC subscribers are branded per domain like any other recipient, and the sentinel to is replaced with the placeholder EMAIL_FROM on the default-identity message so the no-config case stays byte-identical. Guarded against emitting a spurious self-addressed message when every subscriber resolves to a brand. Covered by should_brand_bcc_subscribers_of_a_publisher_broadcast, should_send_a_single_unbranded_broadcast_when_no_configuration_is_set, and should_not_emit_a_placeholder_only_default_message_when_every_broadcast_recipient_is_branded. Also dropped the "self-send" mislabel from the javadoc.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/EmailServiceImpl.java
paulatulis: Added should_set_the_reply_to_header_on_a_branded_message, which sets replyTo and asserts the Reply-To header directly on the sent MimeMessage. A regression that drops or misroutes it now fails a test.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/EmailServiceImpl.java
paulatulis: It's actually the broader fault-isolation question — best-effort + aggregate errors vs all-or-nothing, retry idempotency, and per-group outcome logging — and it also pre-dates this PR (the single-send path could fail mid-flight too). I've captured it as F2 on the follow-up sub-task APIM-14632 with those decisions spelled out, rather than picking a semantics here. Left as today's fail-fast for now.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
paulatulis: Confirmed and fixed. The check now runs a single-@ guard on the whole trimmed value before any bracket stripping, so a trailing Name <addr> pair can't hide the earlier addresses. Added your exact repro as should_return_empty_for_a_multi_address_personal_name_recipient_string (Jane <jane@example.com>, John <john@example.org> → empty) alongside the existing bare-list case. The structurally-identical SenderAddressValidator.PERSONAL_NAME flaw you flagged is real too — since it lives in APIM-14181's code and the fix is a shared, tested single-address helper, I've logged it as M4 on the follow-up sub-task APIM-14632 rather than widening this PR.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/settings/BrandedSenders.java
paulatulis: Added should_return_empty_when_a_parsed_configuration_has_null_domains, which drives a hand-edited raw value with no domains key ([{"from":"noreply@example.com"}]) — bypassing write() validation exactly as you described — through resolve() and asserts it stays empty rather than NPEing. The guard is now covered.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/EmailServiceImpl.java
paulatulis: added should_apply_the_notification_from_name_over_a_branded_bare_from, which sets fromName on the notification against a branded bare from and asserts the resulting From is "Gravitee Team" <noreply@example.com>. That pins the setFrom(from, fromName) branch.

carlos-andres-osorio:
> **gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/EmailServiceImpl.java:116**

sendEmailNotification: no test for the catch (Exception) path that wraps any send-time failure in a TechnicalManagementException. Every existing test drives a successful send; the error branch (the catch block a few lines down) is never exercised, so a regression that swallowed the failure instead of rethrowing would let a failed email delivery pass silently.

A test that stubs mailSender.send(any(MimeMessage.class)) to throw and asserts sendEmailNotification(...) rethrows as TechnicalManagementException would close this gap.

paulatulis:
> > **gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/EmailServiceImpl.java:116**
> 
> sendEmailNotification: no test for the catch (Exception) path that wraps any send-time failure in a TechnicalManagementException. Every existing test drives a successful send; the error branch (the catch block a few lines down) is never exercised, so a regression that swallowed the failure instead of rethrowing would let a failed email delivery pass silently.
> 
> A test that stubs mailSender.send(any(MimeMessage.class)) to throw and asserts sendEmailNotification(...) rethrows as TechnicalManagementException would close this gap.

Added should_wrap_a_send_time_failure_in_a_technical_management_exception: it stubs mailSender.send(any(MimeMessage.class)) to throw and asserts sendEmailNotification(...) rethrows as TechnicalManagementException (message Error while sending email notification). That exercises the catch (Exception) branch, so a regression that swallowed a send-time failure would now fail a test.

carlos-andres-osorio (approved):

PR #18442, opened 2026-07-06
fix(reporter): use JSON escaping for webhook jsonAdditionalMetrics

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-14557

## Description

Replace ?js_string with ?j_string when serializing jsonAdditionalMetrics in v4-message-metrics.ftl (es7x, es8x, es9x). JavaScript escaping emits invalid \' sequences in JSON, causing Elasticsearch to reject webhook callback metrics when captured headers contain apostrophes.

### Before Fix:

https://github.com/user-attachments/assets/dff61b91-6a93-4d58-9567-6980e817efe0

### After Fix:

https://github.com/user-attachments/assets/1b09f707-f587-4e4a-aa46-4c1c4193189f

### DIFF
--- a/gravitee-apim-reporter/gravitee-apim-reporter-common/src/main/resources/freemarker/es7x/index/v4-message-metrics.ftl
+++ b/gravitee-apim-reporter/gravitee-apim-reporter-common/src/main/resources/freemarker/es7x/index/v4-message-metrics.ftl
@@ -103,7 +103,7 @@
     </#if>
     <#if (metrics.jsonAdditionalMetrics())??>
       <#list metrics.jsonAdditionalMetrics() as propKey, propValue>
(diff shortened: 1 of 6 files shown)


aqibmohammadkhan:
@ankita-gupta21 
please add a test case whose value contains an apostrophe (e.g. {"header":"it's a test"}) — that input is precisely what distinguishes the two built-ins and would have caught the original bug.

aqibmohammadkhan (commented):
please take care of the below comment
https://github.com/gravitee-io/gravitee-api-management/pull/18442#issuecomment-4999492273

aqibmohammadkhan (approved):

PR #18443, opened 2026-07-06
feat(gateway): write OTel trace-id and span-id in v4-log ES documents

araffass:
## Summary                                                                                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  - Set traceId/spanId on log entries whenever OTel tracing is active, not only when OTel Logs is enabled
  - Add trace-id and span-id fields to v4-log FreeMarker index templates (es7x/es8x/es9x)                                                                                                                                                                                                                                                                                                                                        
  - Add trace-id and span-id to ES index mapping templates (es7x/es8x/es9x/opensearch)
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  ## Context 
  
  Runtime log documents (v4-log) in Elasticsearch did not include the OTel trace ID, making it impossible to correlate ES logs with traces in Grafana/Tempo. The traceId was only set on log entries when the OTel Logs toggle was enabled, and was never written to the ES document even when present.                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  After this change, whenever OTel tracing is active on an API, the gateway stamps every captured log entry with the active trace-id and span-id. The ES reporter persists them in the v4-log document so they are searchable and linkable to traces.                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                 
https://gravitee.atlassian.net/browse/APIM-14630 

Local Demo:

https://github.com/user-attachments/assets/ec81f615-ca6d-4786-a535-cab8f7bfe9da


pragsty:
Files: ElasticsearchFormatterTest, given/v4/log.json, expected/elasticsearch/v4/log.json
Gateway unit tests prove in-memory capture, but this commit’s reporter changes are untested. Formatter fixtures still have no traceId/spanId, so regressions in v4-log.ftl would not be caught.

Add a fixture with trace/span on a nested request object and assert the formatted JSON contains "trace-id" and "span-id".

pragsty (approved):

PR #18468, opened 2026-07-08
feat(rest-api): expose email.branded_senders in gravitee.yml for on-prem configuration

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14192
## Description
Exposes `email.branded_senders` in `gravitee.yml` so on-prem operators can configure branded senders without the console (APIM-14192 / S5 of the branded-senders epic).

Operators write a **native yaml list**; a load-time adapter reconstructs it into the `email.branded_senders` JSON parameter value:

```yaml
email:
  branded_senders:
    - domains: [example.com]
      from: noreply@example.com
      subject: "[Example] %s"
```

The flat JSON-string form (and the `gravitee_email_branded_senders` env var) is also accepted; both forms are normalised through the same `parse` → `write` path, so they get identical `;`-escaping, CR/LF validation and size guard.

### Behaviour (matches existing `email.*` convention)
- A value in `gravitee.yml` seeds `email.branded_senders` at SYSTEM scope and takes precedence over the console (Org/Env) value — same as `email.from` / `email.subject` / `email.host`.
- When a **valid** value is set in `gravitee.yml`, the console marks the branded-senders field read-only (both flat and native-list forms), like `email.host` / `email.port` today. A present-but-invalid value is logged and ignored, and the field stays editable rather than locking on a value that isn't in effect.
- Shipped commented out, so by default the console fully controls branded senders.

### Changes
- **`BrandedSendersEnvironmentReader`** (new) — reconstructs the native yaml list (and normalises the flat/env form) into the JSON parameter value via `BrandedSenders.write`, reusing validation, `;`-escaping and the size guard.
- **`ParameterServiceImpl.getSystemParameter`** — routes `email.branded_senders` (both forms) through the reader so the value is escaped/validated consistently and yaml wins.
- **`ConfigServiceImpl`** — marks the field read-only via `isConfigured()` (valid-value check), covering both storage forms.
- **`BrandedSenders.write`** made public (symmetric with the public `parse`).
- Shipped `gravitee.yml` + 6 docker quick-setup compose files: commented examples.

### Tests
`BrandedSendersEnvironmentReaderTest` (13), plus 5 new cases in `ParameterServiceTest` and 2 in `ConfigServiceTest` — covering both storage forms, `;`-escaping, scalar `domains`, incomplete/invalid entries, the read-only valid-value gate, and the non-overridable-key path. No regressions in the existing suites.

### Additional Context
getSystemParameter / ConfigServiceImpl special-case EMAIL_BRANDED_SENDERS with an explicit branch rather than a Map<Key, …> reader registry is a deliberate choice. With a single key needing custom reconstruction, an inline, commented branch is clearer than a strategy abstraction, and matches how this legacy service module already handles per-key logic.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/BrandedSendersEnvironmentReader.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/BrandedSendersEnvironmentReader.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.service.impl;
+
+import io.gravitee.rest.api.model.parameters.Key;
+import io.gravitee.rest.api.model.settings.BrandedSenderConfig;
+import io.gravitee.rest.api.model.settings.BrandedSenders;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import lombok.CustomLog;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.stereotype.Component;
+
+/**
+ * Reconstructs the {@code email.branded_senders} parameter value from the native yaml list of objects an
+ * on-prem operator writes in {@code gravitee.yml}:
+ *
+ * <pre>
+ * email:
+ *   branded_senders:
+ *     - domains: [example.com]
+ *       from: noreply@example.com
+ *       subject: "[Example] %s"
+ * </pre>
+ *
+ * <p>Spring flattens such a list into indexed properties ({@code email.branded_senders[0].from},
+ * {@code email.branded_senders[0].domains[0]}, …), so {@code environment.getProperty("email.branded_senders")}
+ * returns {@code null} and the value cannot be read as a single string. This reader walks the indexed
+ * properties into a typed {@link BrandedSenderConfig} list and serialises it with {@link BrandedSenders#write(List)},
+ * yielding the same JSON string form the console persists — including the {@code ';'} escaping the parameter
+ * read path relies on and the size guard.</p>
+ *
+ * <p>The equivalent flat JSON string / env-var form ({@code gravitee_email_branded_senders=[{…}]}) is accepted
+ * too and routed through the same {@link BrandedSenders#parse(String)} + {@link BrandedSenders#write(List)}
+ * normalisation, so both forms get the same escaping, CR/LF validation and size guard and behave identically.
+ * It is used by {@code ParameterServiceImpl} to seed the value at SYSTEM scope.</p>
+ *
+ * @author GraviteeSource Team
+ */
+@Component
+@CustomLog
+public class BrandedSendersEnvironmentReader {
+
+    private static final String PREFIX = Key.EMAIL_BRANDED_SENDERS.key();
+
+    private final ConfigurableEnvironment environment;
+
+    @Autowired
+    public BrandedSendersEnvironmentReader(ConfigurableEnvironment environment) {
+        this.environment = environment;
+    }
+
+    /**
+     * Reads the branded-sender configurations, accepting either the native yaml list or the flat JSON string /
+     * env-var form, and normalising both through {@link BrandedSenders#write(List)} so they share the same
+     * {@code ';'}-escaping, CR/LF validation and size guard. Returns {@link Optional#empty()} when none are
+     * declared (so the caller falls through to the env / org / default value), or when the declared value is
+     * invalid or oversized — a broken on-prem config is logged and ignored rather than breaking the whole
+     * settings response, and (crucially) leaves the console field editable rather than locked on a value that
+     * is not actually in effect.
+     */
+    public Optional<String> read() {
+        List<BrandedSenderConfig> configurations = readConfigurations();
+        if (configurations.isEmpty()) {
+            return Optional.empty();
+        }
+        try {
+            return Optional.of(BrandedSenders.write(configurations));
+        } catch (IllegalArgumentException e) {
+            // write() signals invalid/oversized operator config only via IllegalArgumentException; catching that
+            // (not RuntimeException) so an unrelated future bug surfaces instead of being mislabelled as bad config.
+            // Source-neutral: read() serves both the native yaml list and the flat env-var / JSON form.
+            log.error("Ignoring invalid '{}' branded-senders configuration", PREFIX, e);
+            return Optional.empty();
+        }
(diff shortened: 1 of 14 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
carlos-andres-osorio, line 1: **MAJOR:** Write-path/read-path asymmetry for `EMAIL_BRANDED_SENDERS` — `save()` (around line 453, not touched by this diff) still guards its system-override short-circuit with a bare `environment.containsProperty(key.key())`.

`getSystemParameter()` was updated in this PR to detect the native-yaml-list form via `brandedSendersEnvironmentReader.isConfigured()`, but `save()` wasn't updated symmetrically. `containsProperty()` is `false` for the native yaml-list form (Spring flattens it to indexed keys like `email.branded_senders[0].from`, never binding the bare key), so for that form the system-override short-circuit in `save()` never triggers.

Confirmed reachable: `ConfigServiceImpl.saveConfigByReference()` (`ConfigServiceImpl.java:392`) unconditionally calls `parameterService.save()` for every `@ParameterKey` field — the `METADATA_READONLY` flag is advisory for the frontend only, not enforced server-side. If a save request for branded-senders settings is still submitted while yaml locks the field, it gets silently persisted to org/env storage. That's latent while the yaml entry exists, but once the yaml entry is removed, the read path falls through to the stale persisted value instead of the default.

Suggested fix: mirror `getSystemParameter()`'s pattern in `save()` — branch on `Key.EMAIL_BRANDED_SENDERS` and short-circuit via `brandedSendersEnvironmentReader.isConfigured()` before falling through to the existing `containsProperty()` check.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/BrandedSendersEnvironmentReaderTest.java
```diff
@@ -0,0 +1,227 @@
...
+@MockitoSettings(strictness = Strictness.LENIENT)
+class BrandedSendersEnvironmentReaderTest {
+
+    @Mock
+    private ConfigurableEnvironment environment;
+
+    private BrandedSendersEnvironmentReader reader;
+
+    private void stub(String key, String value) {
+        when(environment.getProperty(key)).thenReturn(value);
```
carlos-andres-osorio, line 52: **MAJOR (test coverage gap):** All these tests hand-stub `environment.getProperty(...)` with the assumed indexed-key shape (`email.branded_senders[0].from`, etc.) rather than binding real `gravitee.yml` content through an actual Spring `Environment`/`YamlPropertySourceLoader`.

If Spring's actual yaml-list flattening behavior ever differs from what's assumed here (property naming, indexing, nested-list edge cases), none of these tests would catch it — they'd keep passing against a stubbed shape that no longer matches reality, and the mismatch would only surface at runtime.

Suggested addition: at least one test that loads a real `gravitee.yml` fixture through `YamlPropertySourceLoader` into a `MockEnvironment`/`ConfigurableEnvironment`, asserting `BrandedSendersEnvironmentReader.read()` parses it correctly — supplementing (not replacing) the existing hand-stubbed unit tests.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/BrandedSendersEnvironmentReader.java
```diff
@@ -0,0 +1,163 @@
...
+     * is not actually in effect.
+     */
+    public Optional<String> read() {
+        List<BrandedSenderConfig> configurations = readConfigurations();
+        if (configurations.isEmpty()) {
+            return Optional.empty();
+        }
+        try {
+            return Optional.of(BrandedSenders.write(configurations));
+        } catch (RuntimeException e) {
```
carlos-andres-osorio: **MINOR:** This catch logs the source as hardcoded `"from gravitee.yml"`, but `read()` serves both the native yaml-list form and the flat env-var/JSON form (`gravitee_email_branded_senders`). An invalid value set via the env var gets logged with a misleading source label pointing at `gravitee.yml`. The error is still logged (not silent), just mislabeled.

Suggested fix: make the log label reflect which source actually produced the value, or use a source-neutral phrase like "from branded-senders configuration" instead of hardcoding "gravitee.yml".

carlos-andres-osorio (commented):
This PR is otherwise solid — the write-path/system-guard asymmetry (MAJOR) is the one item I'd want addressed before merge, since it can let a stale org/env value persist under a locked yaml field. The test-coverage gap and log-label mislabeling are lower stakes.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
paulatulis: save() now mirrors getSystemParameter(), resolving branded_senders' system-override through the reader (covers the native-list form where containsProperty is false, and the flat form). A console save can no longer persist to org/env while yaml locks the field. New test asserts save() short-circuits and never calls create/update.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/BrandedSendersEnvironmentReaderTest.java
paulatulis: added a test loading real yaml through YamlPropertiesFactoryBean (same SnakeYAML flattening gravitee-node uses) into a MockEnvironment, asserting the indexed shape and full round-trip — so the hand-stubbed shape can't silently drift.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/BrandedSendersEnvironmentReader.java
paulatulis: now source-neutral ("…branded-senders configuration").
- Also dropped a redundant test of mine (the non-overridable path was already covered).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/BrandedSendersEnvironmentReader.java
```diff
@@ -0,0 +1,164 @@
...
```
carlos-andres-osorio: Nit: `read()` wraps `BrandedSenders.write(...)` in `catch (RuntimeException e)`, which is broader than the `IllegalArgumentException` `write()` actually throws (the size guard and the `JsonProcessingException` wrapper). An unexpected runtime bug (NPE, `IllegalStateException` from a future refactor) would be relabelled "invalid configuration" and silently disable branded senders / unlock the console field — a future defect unrelated to operator config would get misdiagnosed as bad yaml. The exception is still logged with its stack trace, so not fully silent, hence non-blocking.

carlos-andres-osorio (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/BrandedSendersEnvironmentReader.java
paulatulis: Addressed!

PR #18469, opened 2026-07-08
fix(rest-api): fall back to console link when redirect URL isn't whitelisted

dev-marek:
**Issue**

https://gravitee.atlassian.net/browse/PORTAL-33

**Purpose**

Stop the caller-supplied redirect URL used in registration confirmation and password-reset emails from being trusted by default. Previously, when `portal.whitelist` was empty, any URL was accepted, letting an attacker redirect the signed confirmation/reset token to a domain they control.

**Summary of changes**

- `UserServiceImpl.sanitizePortalRedirectUrl` validates the redirect URL against `portal.whitelist`, falling back to the installation's own resolved portal URL(s) (`InstallationAccessQueryService.getPortalUrls`) when the whitelist is empty. A URL that doesn't match either is **dropped**, not rejected: the call proceeds normally and the email link falls back to the installation's console/portal URL, the same fallback `getTokenRegistrationParams` already uses today when no redirect URL is supplied at all.
- The check is a pre-check called only from the two entry points that actually receive a caller-supplied redirect URL: `register()` (`confirmationPageUrl`) and the caller-facing `resetPassword(ctx, id, resetPageUrl)`. It deliberately does **not** live inside `getTokenRegistrationParams` itself — that method is also used by the gamma password-reset target and the registration-approval flow, which build their target URL internally from trusted installation config (gamma console URL, resolved portal URL) rather than from caller input; routing those through the same whitelist check would incorrectly drop them (a gamma URL will never be in the portal whitelist).
- Every dropped URL is logged (`log.warn`) with the URL, environment, and effective whitelist, so operators can see it happening instead of silently mailing the fallback link.
- `fallbackPortalWhitelist` uses `getPortalUrls` (plural, includes non-overriding multi-tenant access points) and filters out the `DEFAULT_PORTAL_URL` sentinel, so an environment where the Portal URL setting was never configured is correctly treated as "no portal URL available" rather than matching on `http://localhost:4100`.
- `UrlSanitizerUtils.checkAllowed` itself is untouched — its existing default behavior is still needed by unrelated callers (OpenAPI/WSDL import, API duplication, page import).
- **Not a breaking change**: no install can have its registration/reset emails start failing because of this — worst case, the email link points at the console/portal instead of the caller-supplied page, which is the existing behavior for a `null` redirect URL.

**Out of scope**

- Any change to `UrlSanitizerUtils` shared semantics.
- Documentation for `portal.whitelist` configuration.
- A first-class Helm chart value for `portal.whitelist` (the existing `api.env` override already covers it).
- **Invitation emails** (`ApplicationInvitationsResource` → `ApplicationInvitationNotificationDomainServiceImpl`): they call `getTokenRegistrationParams` directly, bypassing `register()`/`resetPassword()` entirely, so covering them would mean sanitizing inside `getTokenRegistrationParams` itself — which is exactly what would break the gamma/registration-approval flows above. Same primitive (client-supplied `confirmation_page_url`), authenticated caller so less exposed. Tracked as a follow-up ticket rather than expanding this PR's scope.

**Testing**

- `UserServiceTest`: added tests directly on the private `sanitizePortalRedirectUrl` (via `ReflectionTestUtils.invokeMethod`, matching this file's existing convention for testing internal helpers) covering — URL preserved when it matches the configured whitelist or the fallback portal URL (including a nested path via `checkAllowed`'s `startsWith` matching) or a non-first entry in a multi-URL fallback list; URL dropped when it matches neither, when the whitelist is empty and no portal URL resolves, when the fallback resolves to the `DEFAULT_PORTAL_URL` sentinel, and when the `ExecutionContext` has no environment id. Also: `register()`/`resetPasswordFromSourceId()` no longer reject on a mismatched URL (they now fail downstream for unrelated reasons instead of `UrlForbiddenException`).
- Ran `UserServiceTest` (101 tests), `UrlSanitizerUtilsTest` (13 tests), and the full `gravitee-apim-rest-api-service` module suite — all green except one pre-existing, unrelated failure (`JsonSchemaServiceTest`, confirmed failing identically on this branch without any of this PR's changes).
- Explicitly re-verified `UserServiceRegistrationApprovalTest` and `UserServiceResetPasswordTargetTest` (gamma reset target) still pass — these don't go through `register()`/`resetPassword()` and must stay unaffected by this change.

**Additional context**

Addresses the second review round:
- Fixed the two bugs in the whitelist fallback (`DEFAULT_PORTAL_URL` sentinel not filtered; singular `getPortalUrl` dropping non-overriding multi-tenant access points).
- Added the requested warn logging on every path where the redirect URL is not used.
- Switched from rejecting the request to dropping the untrusted URL and falling back to the installation's own link, per review discussion — this removes the breaking-change risk, so the `gravitee.yml` comment was reworded instead of adding a `BREAKING CHANGE` footer.

### DIFF
--- a/gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-rest-api/src/main/resources/config/gravitee.yml
+++ b/gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-rest-api/src/main/resources/config/gravitee.yml
@@ -631,7 +631,8 @@ email:
 #  themes:
 #    path: ${gravitee.home}/themes
   # Allows domains to be used while generating some emails from the portal. ie. registration, forget password
-  # Empty whitelist means only the installation's own portal URL is allowed (see installation.standalone.portal.url below).
+  # Empty whitelist means only the installation's own portal URL is allowed (see installation.standalone.portal.url below);
+  # any other URL is ignored and the email falls back to the default link.
 #  whitelist:
 #    - https://portal.domain.com
 #    - https://private-portal.domain.com
--- a/gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-rest-api/src/main/resources/config/gravitee.yml
+++ b/gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-rest-api/src/main/resources/config/gravitee.yml
@@ -631,7 +631,7 @@ email:
 #  themes:
 #    path: ${gravitee.home}/themes
   # Allows domains to be used while generating some emails from the portal. ie. registration, forget password
-  # Empty whitelist means all urls are allowed.
+  # Empty whitelist means only the installation's own portal URL is allowed (see installation.standalone.portal.url below).
 #  whitelist:
 #    - https://portal.domain.com
 #    - https://private-portal.domain.com
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
@@ -133,6 +133,7 @@ import io.gravitee.rest.api.service.exceptions.PasswordFormatInvalidException;
 import io.gravitee.rest.api.service.exceptions.ServiceAccountNotManageableException;
 import io.gravitee.rest.api.service.exceptions.StillPrimaryOwnerException;
 import io.gravitee.rest.api.service.exceptions.TechnicalManagementException;
+import io.gravitee.rest.api.service.exceptions.UrlForbiddenException;
 import io.gravitee.rest.api.service.exceptions.UserAlreadyFinalizedException;
 import io.gravitee.rest.api.service.exceptions.UserNotActiveException;
 import io.gravitee.rest.api.service.exceptions.UserNotFoundException;
@@ -309,6 +310,21 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
             portalWhitelist.add(whitelistUrl);
             i++;
         }
+    }
+
+    /**
+     * Registration confirmation and password reset links embed a signed token in the caller-supplied redirect
+     * URL. Without an explicit whitelist, that URL must be rejected outright rather than allowed by default,
+     * otherwise the token can be exfiltrated to an attacker-controlled domain.
+     */
+    private void checkPortalRedirectUrlAllowed(String url) {
+        if (url == null) {
+            return;
+        }
+        if (portalWhitelist.isEmpty()) {
+            throw new UrlForbiddenException();
+        }
+        UrlSanitizerUtils.checkAllowed(url, portalWhitelist, true);
     }
 
     @Override
@@ -835,9 +851,7 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
     ) {
         final ReferenceContext currentContext = executionContext.getReferenceContext();
 
-        if (confirmationPageUrl != null) {
-            UrlSanitizerUtils.checkAllowed(confirmationPageUrl, portalWhitelist, true);
-        }
+        checkPortalRedirectUrlAllowed(confirmationPageUrl);
 
         checkUserRegistrationEnabled(executionContext);
         boolean autoRegistrationEnabled = isAutoRegistrationEnabled(executionContext, currentContext);
@@ -1425,7 +1439,7 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
             throw new UserNotActiveException(sourceId);
         }
 
-        UrlSanitizerUtils.checkAllowed(resetPageUrl, portalWhitelist, true);
+        checkPortalRedirectUrlAllowed(resetPageUrl);
 
         UserEntity foundUser = this.findBySource(executionContext.getOrganizationId(), IDP_SOURCE_GRAVITEE, sourceId, false);
         if ("ACTIVE".equals(foundUser.getStatus())) {
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
@@ -138,6 +138,7 @@ import io.gravitee.rest.api.service.exceptions.PasswordFormatInvalidException;
 import io.gravitee.rest.api.service.exceptions.ServiceAccountNotManageableException;
 import io.gravitee.rest.api.service.exceptions.StillPrimaryOwnerException;
 import io.gravitee.rest.api.service.exceptions.TechnicalManagementException;
+import io.gravitee.rest.api.service.exceptions.UrlForbiddenException;
 import io.gravitee.rest.api.service.exceptions.UserAlreadyFinalizedException;
 import io.gravitee.rest.api.service.exceptions.UserNotActiveException;
 import io.gravitee.rest.api.service.exceptions.UserNotFoundException;
@@ -308,6 +309,21 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
             portalWhitelist.add(whitelistUrl);
             i++;
         }
+    }
+
+    /**
+     * Registration confirmation and password reset links embed a signed token in the caller-supplied redirect
+     * URL. Without an explicit whitelist, that URL must be rejected outright rather than allowed by default,
+     * otherwise the token can be exfiltrated to an attacker-controlled domain.
+     */
+    private void checkPortalRedirectUrlAllowed(String url) {
+        if (url == null) {
+            return;
+        }
+        if (portalWhitelist.isEmpty()) {
+            throw new UrlForbiddenException();
+        }
+        UrlSanitizerUtils.checkAllowed(url, portalWhitelist, true);
     }
 
     @Override
@@ -840,9 +856,7 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
     ) {
         final ReferenceContext currentContext = executionContext.getReferenceContext();
 
-        if (confirmationPageUrl != null) {
-            UrlSanitizerUtils.checkAllowed(confirmationPageUrl, portalWhitelist, true);
-        }
+        checkPortalRedirectUrlAllowed(confirmationPageUrl);
 
         checkUserRegistrationEnabled(executionContext);
         boolean autoRegistrationEnabled = isAutoRegistrationEnabled(executionContext, currentContext);
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
@@ -313,17 +313,26 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
 
     /**
      * Registration confirmation and password reset links embed a signed token in the caller-supplied redirect
-     * URL. Without an explicit whitelist, that URL must be rejected outright rather than allowed by default,
-     * otherwise the token can be exfiltrated to an attacker-controlled domain.
+     * URL. Without an explicit whitelist, that URL must only be accepted if it matches the installation's own
+     * resolved portal URL, otherwise the token can be exfiltrated to an attacker-controlled domain.
      */
-    private void checkPortalRedirectUrlAllowed(String url) {
-        if (url == null) {
+    private void checkPortalRedirectUrlAllowed(ExecutionContext executionContext, String url) {
+        if (url == null || url.isBlank()) {
             return;
         }
-        if (portalWhitelist.isEmpty()) {
+        List<String> effectiveWhitelist = portalWhitelist.isEmpty() ? fallbackPortalWhitelist(executionContext) : portalWhitelist;
+        if (effectiveWhitelist.isEmpty()) {
             throw new UrlForbiddenException();
         }
-        UrlSanitizerUtils.checkAllowed(url, portalWhitelist, true);
+        UrlSanitizerUtils.checkAllowed(url, effectiveWhitelist, true);
+    }
+
+    private List<String> fallbackPortalWhitelist(ExecutionContext executionContext) {
+        if (!executionContext.hasEnvironmentId()) {
+            return List.of();
+        }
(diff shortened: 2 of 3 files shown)


thomasballivet (changes requested):
Please mention explicitly that this is a breaking change in commit : 
```
fix(rest-api)!: reject unconfigured redirect URL in registration/reset password

BREAKING CHANGE: portal.whitelist must now be explicitly configured.
Registration confirmation and password reset links are rejected when it is empty.
```
Update `# Empty whitelist means all urls are allowed.` comments in gravitee.yml.

And make sure helm charts can set it up.

OR

Can we use `InstallationAccessQueryService.getPortalUrl()` as a fallback when the list is empty? That way we would avoid the breaking change

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -310,6 +311,30 @@ public void afterPropertiesSet() {
...
+            throw new UrlForbiddenException();
+        }
+        UrlSanitizerUtils.checkAllowed(url, effectiveWhitelist, true);
+    }
+
+    private List<String> fallbackPortalWhitelist(ExecutionContext executionContext) {
+        if (!executionContext.hasEnvironmentId()) {
+            return List.of();
+        }
+        String portalUrl = installationAccessQueryService.getPortalUrl(executionContext.getEnvironmentId());
```
thomasballivet: Looks like `getPortalUrl()` doesn't hold up as a fallback source here.

In standalone it practically never returns `null`: `Key.PORTAL_URL` is declared without a `defaultValue` (`Key.java:101`), so `ParameterServiceImpl:240` returns `null` when the setting was never set, and `InstallationAccessQueryServiceImpl:344` substitutes `DEFAULT_PORTAL_URL` (the method is `@NonNull`). On an install where 'Portal URL' was never configured, the effective whitelist ends up as `["http://localhost:4100"]`. That list is non-empty, so we don't hit the `throw` above — we go straight to `checkAllowed`, which then requires a match: the URL the portal actually sends (`https://portal.acme.com/...`) matches nothing and gets rejected. The `portalUrl == null` branch below only happens in multi-tenant, where `buildHttpUrl` returns `null` for a missing access point. This file already treats the sentinel as "not configured" at line 1059 (`isEmpty(portalUrl) || DEFAULT_PORTAL_URL.equals(portalUrl)`), like `ConfigServiceImpl:355`.

Second catch in multi-tenant: `getPortalAccessPoint()` uses `includeAll=false`, so when an environment has several portal access points it keeps only the `isOverriding` ones and takes `findFirst()` (`AccessPointQueryServiceImpl:137` + `:262`) — a default access point alongside an overriding custom domain would be rejected.

`getPortalUrls(environmentId)` uses `includeAll=true` and returns the whole list, which fits better here:

```java
List<String> portalUrls = installationAccessQueryService.getPortalUrls(executionContext.getEnvironmentId());
if (portalUrls == null) return List.of();
return portalUrls.stream().filter(u -> !isEmpty(u) && !DEFAULT_PORTAL_URL.equals(u)).toList();
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -310,6 +311,30 @@ public void afterPropertiesSet() {
...
+     * URL. Without an explicit whitelist, that URL must only be accepted if it matches the installation's own
+     * resolved portal URL, otherwise the token can be exfiltrated to an attacker-controlled domain.
+     */
+    private void checkPortalRedirectUrlAllowed(ExecutionContext executionContext, String url) {
+        if (url == null || url.isBlank()) {
+            return;
+        }
+        List<String> effectiveWhitelist = portalWhitelist.isEmpty() ? fallbackPortalWhitelist(executionContext) : portalWhitelist;
+        if (effectiveWhitelist.isEmpty()) {
+            throw new UrlForbiddenException();
```
thomasballivet: This rejection leaves no trace anywhere. On reset, `UsersResource:137` catches `AbstractManagementException` ("do not fail to prevent from brute force attack"), logs a generic warn — "This url is forbidden. Please, contact your administrator", without the URL, the environment or the cause — and returns `204`, so the user is told the mail was sent. On registration there's no catch and nothing is logged at all.

Could we warn on both reject paths (here and on the `checkAllowed` failure below) with the URL, the environment and the effective whitelist, in the style of `portalRegistrationConfirmationUrl` at line 1059? Plus one in `afterPropertiesSet()` when `portal.whitelist` is empty, so it surfaces at startup rather than on the first user.

More fundamentally, worth deciding what should happen when nothing matches. Rejecting is defensible, but we could also drop the untrusted URL and let `getTokenRegistrationParams` build the link from the installation's own URL — what it already does when `targetPageUrl == null`. The exfiltration vector is gone either way, and that variant never breaks a flow: the user still gets a mail pointing at the legitimate portal. If we keep rejecting, the warns above become mandatory.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -310,6 +311,30 @@ public void afterPropertiesSet() {
         }
     }
 
+    /**
+     * Registration confirmation and password reset links embed a signed token in the caller-supplied redirect
+     * URL. Without an explicit whitelist, that URL must only be accepted if it matches the installation's own
+     * resolved portal URL, otherwise the token can be exfiltrated to an attacker-controlled domain.
+     */
+    private void checkPortalRedirectUrlAllowed(ExecutionContext executionContext, String url) {
```
thomasballivet: Same pattern is still open on invitations: `ApplicationInvitationsResource` (`POST /applications/{id}/invitations` and `_resend`) passes a client-supplied `confirmation_page_url` straight into the token link (`ApplicationInvitationNotificationDomainServiceImpl:71` -> `getTokenRegistrationParams`). Authenticated, so less exposed, but same primitive.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -840,9 +865,7 @@ public UserEntity register(
     ) {
         final ReferenceContext currentContext = executionContext.getReferenceContext();
 
-        if (confirmationPageUrl != null) {
-            UrlSanitizerUtils.checkAllowed(confirmationPageUrl, portalWhitelist, true);
-        }
+        checkPortalRedirectUrlAllowed(executionContext, confirmationPageUrl);
```
thomasballivet: Could you refresh the description? It still describes commit 1 only, lists test names that aren't on the branch, and mentions `resetPasswordFromSourceId` instead of `resetPassword(ctx, id, url)`. Small behaviour change worth a line too: `register()` used to reject a blank URL when a whitelist was configured, `isBlank()` now lets it fall through to the console link.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
```diff
@@ -1320,6 +1323,88 @@ public void shouldFailWhileResettingPasswordWhenUserFoundIsNotInternallyManaged(
...
+        assertThrows(UrlForbiddenException.class, () ->
+            userService.resetPasswordFromSourceId(organizationOnlyContext, "my@email.com", "https://attacker.example.com/reset")
+        );
+
+        verifyNoInteractions(emailService);
+        verifyNoInteractions(installationAccessQueryService);
+    }
+
+    @Test
+    public void shouldAllowResetPageUrlWhenPortalWhitelistIsNotConfiguredButUrlMatchesFallbackPortalUrl() throws TechnicalException {
```
thomasballivet: Two cases missing here, and they're the ones that tell us whether real installs break: `getPortalUrl(s)` returning `DEFAULT_PORTAL_URL` (should reject), and the URL the portal actually sends — `registration.component.ts:110` (`origin + pathname + '/confirm'`) and `reset-password.component.ts:63` (`href + '/confirm'`) — which only match through `checkAllowed`'s `startsWith(whitelistUrl + '/')` branch.

gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-rest-api/src/main/resources/config/gravitee.yml
```diff
@@ -631,7 +631,7 @@ email:
 #  themes:
 #    path: ${gravitee.home}/themes
   # Allows domains to be used while generating some emails from the portal. ie. registration, forget password
-  # Empty whitelist means all urls are allowed.
+  # Empty whitelist means only the installation's own portal URL is allowed (see installation.standalone.portal.url below).
```
thomasballivet: If we keep rejecting when nothing matches (see my note on `checkPortalRedirectUrlAllowed`), I'd keep the `fix(rest-api)!:` + `BREAKING CHANGE:` note: an install with neither a whitelist nor a configured portal URL will see its registration/reset emails refused. That note also has to reach users through the release notes, which are generated from the Jira ticket summary (`changelog-helper.mjs:43`) — the commit footer alone will not.

On Helm: no chart value for `portal.whitelist` today, but a default install already renders `installation.standalone.portal.url` from the portal ingress host, so the fallback covers it. Operators who need extra domains can go through `api.env` (`values.yaml:726`, copied into the API container at `api-deployment.yaml:228`):

```yaml
api:
  env:
    - name: gravitee_portal_whitelist_0
      value: https://portal.acme.com
```

So nothing blocks this PR. A first-class chart value would be nicer — a `values.yaml` key, a root `portal:` section in `api-configmap.yaml`, a helm-unittest and a `CHANGELOG.md` entry — but that is a chart release, so that can be a separate PR.

thomasballivet (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
dev-marek: Fixed — switched `fallbackPortalWhitelist` to `getPortalUrls` (plural, `includeAll=true`) and filter out blank/`DEFAULT_PORTAL_URL` entries, mirroring the sentinel check already at `portalRegistrationConfirmationUrl`. Added regression tests for both cases you flagged: the sentinel-as-fallback case, and a multi-tenant list where the matching URL isn't the first entry.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
dev-marek: Went with your second option here: switched from rejecting to silently dropping the untrusted URL and letting `getTokenRegistrationParams` fall through to its existing console/portal link (the same path it already takes for a `null` redirect URL). That removes the exfiltration vector without ever failing the flow, so the mandatory-logging concern is now the main safeguard rather than an exception — added a `log.warn` on every drop path (empty effective whitelist, and `checkAllowed` rejection), naming the URL, environment and effective whitelist. Didn't add a startup warn in `afterPropertiesSet()` for an empty `portal.whitelist` — with the fallback in place that's now a normal, not necessarily misconfigured, state for most single-environment installs.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
dev-marek: Looked into folding this in by moving the check into `getTokenRegistrationParams` itself (invitations call that directly), but reverted it — that method is also used by the gamma password-reset target and the registration-approval flow, both of which build their target URL from trusted installation config rather than caller input. Routing those through the portal-whitelist check breaks them outright (a gamma URL will never be in the portal whitelist), which the full module test suite caught. Covering invitations properly needs a guard in `ApplicationInvitationNotificationDomainServiceImpl`/`ApplicationInvitationsResource`, which is out of scope for this PR — tracking it as a follow-up ticket instead, given the caller there is authenticated.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
dev-marek: Description rewritten to match the current state (two commits, the drop-and-fallback design, corrected method/test names). The `isBlank()` point is moot now — since the check no longer throws, a blank URL and a rejected URL behave identically (both fall through to the default link), so there's no divergent behavior to call out there anymore.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
dev-marek: Added both: a case where the fallback resolves to `DEFAULT_PORTAL_URL` (still treated as unresolved), and a nested-path case matching `checkAllowed`'s `startsWith(whitelistUrl + '/')` branch, which is the shape the Angular portal actually sends. Also added multi-tenant coverage for the `getPortalUrls` fix above (non-first entry in the list). Note the tests now target the private `sanitizePortalRedirectUrl` directly via `ReflectionTestUtils.invokeMethod` rather than through `register()`/`resetPassword()`, since with the drop-instead-of-reject design those entry points no longer produce an observable difference (nothing throws either way) — see the design shift described in the other threads.

gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-rest-api/src/main/resources/config/gravitee.yml
dev-marek: Since this switched from rejecting to dropping-and-falling-back (see the other threads), there's no breaking-change risk left — no install can have its registration/reset emails start failing because of this, so I skipped the `fix(rest-api)!:`/`BREAKING CHANGE:` footer and the Jira release-note follow-up. Reworded the comment instead to describe the actual behavior: an unmatched URL is ignored and the email falls back to the default link. Agreed on Helm — leaving the first-class chart value as a separate, optional follow-up.

thomasballivet (approved):

PR #18474, opened 2026-07-08
fix(APIM-14632): reject multi-address branded sender From at save time via a shared single-address helper

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14632
## Description

What & why

Follow-up M4 from the branded-senders review (APIM-14632). The branded-sender From was validated more leniently than it could actually be sent: an admin could save a config that then failed at send time.

The bug

SenderAddressValidator used a greedy ^.*<...>$ display-name strip. Given a multi-address value like:

alice@example.com, Team <noreply@example.org>

the .* swallowed everything before the last <…>, leaving one valid-looking address, so validation passed. At send time new InternetAddress(from) parses the whole string, rejects the list, and the notification errors out — a failure remote from its cause (save looked fine days earlier). Meanwhile BrandedSenders.extractDomain already rejected multi-address values correctly, so the two code paths disagreed on "what is a single address."

The fix

- New shared helper EmailAddresses.singleAddress(String) — the single source of truth for "extract one sendable address": unwraps a Name <addr> personal name and rejects null/blank/zero-@/multi-@ (counting @ on the whole value before bracket stripping, so a trailing Name <addr> can't mask earlier addresses).
- SenderAddressValidator now routes through it (drops the greedy regex) → save no longer accepts a From delivery can't send.
- BrandedSenders.extractDomain now delegates to the same helper → the two paths can't drift apart again. Behavior unchanged.

Net effect is line-negative: it removes duplicated parsing rather than adding logic.

Tests

- New EmailAddressesTest (9) pins the helper contract.
- New validator regression test for the address-before-display-name case.
- Existing BrandedSenderValidationTest (22) and EmailBrandedSendersTest (27) unchanged and green.
- Full gravitee-apim-rest-api-model suite: 188 passing; prettier clean.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/email/EmailAddressesTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/email/EmailAddressesTest.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.model.email;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/**
+ * Contract of the shared single-address parsing used by both the save-time {@code SenderAddressValidator}
+ * and the send-time branded-sender matching, so the two never disagree on what is a single, sendable address.
+ *
+ * @author GraviteeSource Team
+ */
+class EmailAddressesTest {
+
+    @Test
+    void should_return_a_bare_single_address() {
+        assertThat(EmailAddresses.singleAddress("noreply@example.com")).contains("noreply@example.com");
+    }
+
+    @Test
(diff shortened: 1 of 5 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/email/EmailAddressesTest.java
```diff
@@ -0,0 +1,102 @@
...
+    void should_unwrap_a_bare_bracketed_address() {
+        assertThat(EmailAddresses.singleAddress("<jane@example.com>")).contains("jane@example.com");
+    }
+
+    @Test
+    void should_trim_surrounding_whitespace() {
+        assertThat(EmailAddresses.singleAddress("  Jane <jane@example.com>  ")).contains("jane@example.com");
+    }
+
+    @Test
```
carlos-andres-osorio, line 52: Nit: The empty-result cases (`should_be_empty_for_null`, `should_be_empty_for_blank`, `should_be_empty_when_there_is_no_at_sign`, `should_be_empty_for_a_bare_multi_address_list`, `should_be_empty_for_a_personal_name_multi_address_list`, `should_be_empty_when_an_address_precedes_the_personal_name`, `should_reject_a_quoted_display_name_containing_an_at_sign`, `should_be_empty_when_the_at_is_outside_the_brackets`) share one assertion shape (`assertThat(EmailAddresses.singleAddress(input)).isEmpty()`) and differ only by input. Consider collapsing the non-null cases into a single `@ParameterizedTest` with `@ValueSource(strings = { ... })`, keeping `should_be_empty_for_null` separate since `@ValueSource` can't carry a null literal.

carlos-andres-osorio:
This PR is missing the `docs-needed`/`no-docs` label. Can you confirm which applies and add it?

carlos-andres-osorio (approved):

paulatulis:
> This PR is missing the `docs-needed`/`no-docs` label. Can you confirm which applies and add it?

I keep forgetting that the absence of `docs-needed` doesn't mean `no-docs`. Added!

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/model/email/EmailAddressesTest.java
paulatulis: addressed

PR #18482, opened 2026-07-09
fix(redis): reconnect when Redis connection is closed

mukul-tyagi08:
https://gravitee.atlassian.net/browse/APIM-14658

## Hotfix
This hotfix restores Redis reconnection after dropped rate-limit connections and hardens the reconnect lifecycle against reconnect storms.

https://gravitee.atlassian.net/browse/APIM-14658

## Changes:
- Clear stale connection state when Redis reports a closed/broken connection.
- Reconnect even when `connected=true` was left behind by the previous connection.
- Guard connection handlers with a connection generation so stale `endHandler` / `exceptionHandler` callbacks cannot tear down a newer healthy connection.
- Deduplicate reconnect timers with `reconnectPending` so many in-flight `evalsha` failures do not schedule hundreds of reconnect attempts.
- Treat Redis operation timeouts as recoverable connection failures to cover half-open TCP cases.
- Notify the Redis client from `RedisRateLimitRepository` when Redis operations fail with connection-class errors.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RedisRateLimitRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RedisRateLimitRepository.java
@@ -72,7 +72,10 @@ public class RedisRateLimitRepository implements RateLimitRepository<RateLimit> 
                             convertToList(this.redisClient.scriptSha1(SCRIPT_RATELIMIT_KEY), REDIS_KEY_PREFIX + key, weight, newRate)
                         )
                     )
-                    .onFailure(this::logOperationFailure)
+                    .onFailure(t -> {
+                        logOperationFailure(t);
+                        redisClient.notifyConnectionFailure(t);
+                    })
                     .onComplete(asyncResultHandler)
         )
             .map(response -> {
--- a/gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/vertx/RedisClient.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/vertx/RedisClient.java
@@ -23,6 +23,8 @@ import io.vertx.redis.client.Redis;
 import io.vertx.redis.client.RedisAPI;
 import io.vertx.redis.client.RedisOptions;
 import java.io.InputStream;
+import java.net.SocketException;
+import java.nio.channels.ClosedChannelException;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Map;
@@ -30,6 +32,7 @@ import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 import lombok.CustomLog;
 
@@ -74,7 +77,7 @@ public class RedisClient {
 
     public Future<RedisAPI> redisApi() {
         LoopRedis loop = resolveLoop();
-        synchronized (loop) {
+        synchronized (loop.monitor) {
             if (loop.redisAPIFuture == null) {
                 startConnectLoop(loop, 0);
             }
@@ -83,6 +86,24 @@ public class RedisClient {
             }
             return loop.redisAPIFuture;
         }
+    }
+
+    /**
+     * Notifies the client that a Redis operation failed because the connection is no longer usable.
+     * This covers cases where the TCP session is half-open and Vert.x connection handlers have not
+     * fired yet, which was leaving {@code connected=true} and blocking automatic reconnection.
+     */
+    public void notifyConnectionFailure(final Throwable failure) {
+        if (!isRecoverableConnectionFailure(failure)) {
+            return;
+        }
+        final Context context = Vertx.currentContext();
+        if (context == null) {
+            log.debug("Ignoring Redis connection failure notification without a Vert.x context");
+            return;
+        }
+        final LoopRedis loop = resolveLoop();
+        context.runOnContext(v -> invalidateConnection(loop));
     }
 
     public String scriptSha1(final String key) {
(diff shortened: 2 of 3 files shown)


aqibmohammadkhan:
Fix direction is correct — removing the connected.get() guard restores pre-regression reconnect behavior.

some suggestions: 

- stale-handler flapping. closing flag is reset before the old connection's handlers fire (they're async after close()), so a late handler tears down the new healthy connection, causing a self-sustaining reconnect loop. Use a connection generation counter in the handlers instead.

- timer storm. Every failed evalsha calls notifyConnectionFailure; during the backoff window each one schedules another reconnect timer, and each timer aborts the previous in-flight connect. Dedupe with a reconnectPending flag.
startConnectLoop assigns loop.redis and redisAPIFuture outside the synchronized block — move them inside.

- Recovery test is weak: await_connected calls redisApi(), which itself triggers reconnect, so the test passes even if the automatic timer path is broken. Also runs on a non-context thread (fallback loop) while isConnected() accepts any loop.

- should_ignore_non_connection_operation_failures is racy — the effect is async, assertion runs immediately; add an await.

- Message-substring error classification is fragile; prefer exception types with strings as fallback.

- Half-open connections usually hang → RedisOperationTimeoutException via RxJava timeout, which never reaches onFailure, so notifyConnectionFailure misses the most common half-open symptom.

- Redundant outer closing.get() check in handleConnectionLost.

mukul-tyagi08:
> Fix direction is correct — removing the connected.get() guard restores pre-regression reconnect behavior.
> 
> some suggestions:
> 
> * stale-handler flapping. closing flag is reset before the old connection's handlers fire (they're async after close()), so a late handler tears down the new healthy connection, causing a self-sustaining reconnect loop. Use a connection generation counter in the handlers instead.
> * timer storm. Every failed evalsha calls notifyConnectionFailure; during the backoff window each one schedules another reconnect timer, and each timer aborts the previous in-flight connect. Dedupe with a reconnectPending flag.
>   startConnectLoop assigns loop.redis and redisAPIFuture outside the synchronized block — move them inside.
> * Recovery test is weak: await_connected calls redisApi(), which itself triggers reconnect, so the test passes even if the automatic timer path is broken. Also runs on a non-context thread (fallback loop) while isConnected() accepts any loop.
> * should_ignore_non_connection_operation_failures is racy — the effect is async, assertion runs immediately; add an await.
> * Message-substring error classification is fragile; prefer exception types with strings as fallback.
> * Half-open connections usually hang → RedisOperationTimeoutException via RxJava timeout, which never reaches onFailure, so notifyConnectionFailure misses the most common half-open symptom.
> * Redundant outer closing.get() check in handleConnectionLost.

Done:

What changed (addressing review)
1. Stale handler flapping → connection generation

Removed theclosingflag
Each connect gets agenerationcounter; handlers close over that generation
handleConnectionLost(loop, generation)no-ops if generation doesn't match
invalidateConnection()bumps generation before closing, so asyncendHandlerfrom old connections can't tear down healthy new ones
2. Reconnect timer storm → reconnectPending dedupe

scheduleReconnectIfAbsent()uses CAS — only one reconnect timer at a time
Cleared when the timer fires andstartConnectLoopruns
100 concurrentnotifyConnectionFailurecalls now schedule a single reconnect
3. Synchronization

loop.redisandloop.redisAPIFutureassigned insidesynchronized, with generation check to discard superseded connect attempts
4. Operation timeout path

RedisOperationTimeoutExceptiontriggersnotifyConnectionFailure()(covers Michelin ~30s half-open pattern)
Also matchesClosedChannelException/SocketExceptionby type
5. Tests strengthened

Recovery test runs on Vert.x context,passivewait (noredisApi()polling), verifiesPING
Dedup test: 100 failures → single recovery
Non-connection failure test waits for asyncrunOnContextbefore asserting
All RedisClientTest + RedisClientConcurrencyTest pass.

wbabyte (commented):
Critical findings from review, anchored inline. (Skipping the message-allow-list/locale item per discussion.) Main concern: the timeout handling can turn a Redis latency blip into a cluster-wide fail-open storm.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/vertx/RedisClient.java
```diff
@@ -119,54 +145,149 @@ private static int currentLoopKey() {
...
+                if (loop.connected.get() || loop.connecting.get()) {
                     return;
                 }
                 startConnectLoop(loop, retry + 1);
             }
         });
     }
 
+    private static boolean isRecoverableConnectionFailure(final Throwable failure) {
+        Throwable current = failure;
```
wbabyte, line 258: **Critical:** classifying `RedisOperationTimeoutException` as a connection failure is wrong. A timeout means the operation was *slow* (loaded-but-healthy Redis, slow Lua eval, GC pause), not that the socket is dead. Combined with the repository firing `notifyConnectionFailure` on every timeout (RedisRateLimitRepository.java:97-100), this tears down live connections during ordinary latency. Please remove `RedisOperationTimeoutException` here (or gate it behind a consecutive-timeout threshold).

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/vertx/RedisClient.java
```diff
@@ -85,6 +89,28 @@ public Future<RedisAPI> redisApi() {
...
+        final Context context = Vertx.currentContext();
+        if (context != null) {
+            final LoopRedis loop = resolveLoop();
+            context.runOnContext(v -> invalidateConnection(loop));
+            return;
+        }
+        loops
+            .values()
+            .stream()
+            .filter(loop -> loop.connected.get())
```
wbabyte: **Critical (fail-open storm):** this no-context branch invalidates **every** connected loop, not just the one that failed. The timeout path reaches here: `.timeout(...)` in RedisRateLimitRepository emits on RxJava's `Schedulers.computation()` (no `observeOn`/`subscribeOn` in the chain), so `Vertx.currentContext()` is null and we land in this branch. Net effect: one slow op on one loop → all N loops torn down and reconnected → during each reconnect window `isConnected()` is false → `incrementAndGet` fails open → quota bypass across the fleet. A latency blip gets amplified into cluster-wide fail-open — the exact failure class this PR fixes. Only the loop that actually failed should be invalidated.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/ratelimit/RedisRateLimitRepository.java
```diff
@@ -90,7 +93,12 @@ public Single<RateLimit> incrementAndGet(String key, long weight, Supplier<RateL
 
                 return newRate;
             })
-            .timeout(operationTimeout, TimeUnit.MILLISECONDS, Single.error(new RedisOperationTimeoutException(operationTimeout)));
+            .timeout(operationTimeout, TimeUnit.MILLISECONDS, Single.error(new RedisOperationTimeoutException(operationTimeout)))
+            .doOnError(throwable -> {
```
wbabyte: **Critical:** this `doOnError` runs on the RxJava `.timeout()` fallback thread (computation scheduler), off the Vert.x context, so `notifyConnectionFailure` takes the no-context branch and invalidates *all* loops (see RedisClient.java:110). Together with treating the timeout as a connection failure, this converts transient latency into fleet-wide reconnect + fail-open. Recommend not invalidating on operation timeout at all (or requiring repeated timeouts), and never fanning out to loops that didn't fail.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/vertx/RedisClient.java
```diff
@@ -119,54 +145,149 @@ private static int currentLoopKey() {
     }
 
     private void startConnectLoop(final LoopRedis loop, final int retry) {
-        if (loop.redis != null) {
-            loop.redis.close();
-            loop.redis = null;
+        final long connectionGeneration;
+        synchronized (loop.monitor) {
+            closeClient(loop);
```
wbabyte: **Reconnect race:** `closeClient(loop)` + nulling `redisAPIFuture`/`connected` happens *before* winning the `connecting.compareAndSet` (line 154). A caller that loses the CAS has already torn down live state, then returns — leaving a window where an in-flight/valid connection was closed but `connected` may still read true elsewhere. Do the teardown only after winning the CAS.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/vertx/RedisClient.java
```diff
@@ -119,54 +145,149 @@ private static int currentLoopKey() {
...
             .timeout(options.getNetClientOptions().getConnectTimeout(), TimeUnit.MILLISECONDS)
             .onFailure(t -> {
+                if (connectionGeneration != loop.generation.get()) {
+                    return;
+                }
                 log.error("Unable to connect to Redis on event loop {}", currentLoopKey(), t);
                 loop.connected.set(false);
                 loop.connecting.set(false);
                 loop.redisAPIFuture = null;
                 attemptReconnect(loop, retry);
```
wbabyte: **Reconnect race:** this `onFailure` calls `attemptReconnect(loop, retry)` directly, bypassing the `reconnectPending` dedup (`scheduleReconnectIfAbsent`) that the handler paths use. If the connection drops during `loadScripts`, both the connection's `exceptionHandler` and this failing chain schedule reconnect timers for the same generation → two timers race into `startConnectLoop`. Route all reconnects through a single authority.

gravitee-apim-repository/gravitee-apim-repository-redis/src/test/java/io/gravitee/repository/redis/vertx/RedisClientTest.java
```diff
@@ -116,6 +118,50 @@ void is_connected_false_when_redis_unreachable() throws Exception {
         assertThat(unreachable.isConnected()).isFalse();
     }
 
+    @Test
+    void should_recover_after_connection_failure_notification_without_polling_redis_api() throws Exception {
```
wbabyte: **Tests don't reproduce the regression:** these tests fire a *fabricated* exception at a *healthy* testcontainer Redis — no test drops/pauses/kills a real connection. They would have failed on 4.11.16 only because `notifyConnectionFailure` didn't compile, not because they reproduce the bug. Recovery is also masked: `isConnected()` is an OR across loops (another up loop makes it pass), and `redisApi()` lazily reconnects on a null future so the ping itself re-establishes the connection independent of the fix. Please add an acceptance test that drops a live connection (container pause / `CLIENT KILL` / Toxiproxy) and drives recovery through `RedisRateLimitRepository.incrementAndGet`, asserting recovery on the *same* invalidated loop via the passive path.

wbabyte (approved):

mukul-tyagi08:
> Critical findings from review, anchored inline. (Skipping the message-allow-list/locale item per discussion.) Main concern: the timeout handling can turn a Redis latency blip into a cluster-wide fail-open storm.

Addressed all critical items: removed timeout-based invalidation, dropped no-context fan-out, CAS-before-teardown, unified reconnect dedup, and added CLIENT KILL acceptance test via incrementAndGet.

ThibaudAV (approved):
Sounds good to me, based on the feedback from @wbabyte

PR #18490, opened 2026-07-09
feat(rest-api): expose branded-senders inherited flag and add env-scope reset endpoint

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14612
## Description

Backend half of APIM-14612. Adds the two pieces the S4 Portal branded-senders UI (APIM-14191) couldn't ship without backend support: (1) an inherited flag so the Env settings response can tell an Org-inherited value from an Env override, and (2) a reset-to-inherited endpoint that deletes the Env-scoped email.branded_senders param so the cascade falls back to Org/System.

▎ Reviewer note — the ticket's "mirror notification-templates" framing is inaccurate. That backend pattern doesn't exist: notification-templates "overridden" is computed client-side from enabled/id, is org-scoped only, and has no reset endpoint. This work is built on the ParameterService/ParameterRepository cascade layer instead (SYSTEM → ENV → ORG → default), so existsOnScope / delete / the reset endpoint are net-new surface, not a copy.

Changes

- Email — new derived brandedSendersInherited boolean (not a @ParameterKey; set at read time).
- ParameterService.existsOnScope — scope-exact presence check (no cascade), to distinguish an Env override from inheritance.
- ParameterService.delete — scope-exact, idempotent delete (no-op when absent), reusing the save-path cache-invalidate + cluster command logic.
- ConfigServiceImpl.loadConfigByReference — sets brandedSendersInherited at ENVIRONMENT scope only (!systemConfigured && !envOverride); left false at Org scope.
- ConfigService.resetPortalBrandedSenders — deletes the Env param and returns the refreshed PortalSettings.
- PortalSettingsResource — POST .../settings/email/branded-senders/reset, gated by ENVIRONMENT_SETTINGS[U], returns the refreshed settings — mirroring ThemeResource.resetTheme() (POST+/reset, not DELETE).

Tests

ConfigServiceTest (+5), ParameterServiceTest (+4), and a net-new PortalSettingsResourceTest (+2). All green.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
@@ -469,9 +469,7 @@ public class ParameterServiceImpl extends TransactionalService implements Parame
 
             if (updateMode) {
                 if (value == null) {
-                    parameterRepository.delete(key.key(), refIdToUse, ParameterReferenceType.valueOf(referenceType.name()));
-                    cache.invalidate(computeCacheKey(key.key(), refIdToUse, ParameterReferenceType.valueOf(referenceType.name())));
-                    sendInvalidateParameterCacheCommand(
+                    deleteParameterAndInvalidateCache(
                         key.key(),
                         refIdToUse,
                         ParameterReferenceType.valueOf(referenceType.name()),
@@ -718,6 +716,61 @@ public class ParameterServiceImpl extends TransactionalService implements Parame
 
     private String getDefaultParameterValue(Key key) {
         return key.defaultValue();
+    }
+
+    @Override
+    public boolean existsOnScope(Key key, String referenceId, io.gravitee.rest.api.model.parameters.ParameterReferenceType referenceType) {
+        try {
+            return parameterRepository.findById(key.key(), referenceId, ParameterReferenceType.valueOf(referenceType.name())).isPresent();
+        } catch (TechnicalException e) {
+            throw new TechnicalManagementException(
+                "An error occurs while checking parameter existence with key: " + key.key() + " on scope: " + referenceType,
+                e
+            );
+        }
+    }
+
+    @Override
+    public void delete(
+        ExecutionContext executionContext,
+        Key key,
+        String referenceId,
+        io.gravitee.rest.api.model.parameters.ParameterReferenceType referenceType
+    ) {
+        String refIdToUse = getEffectiveReferenceId(executionContext, referenceId, referenceType);
+        ParameterReferenceType repositoryReferenceType = ParameterReferenceType.valueOf(referenceType.name());
+        try {
+            if (parameterRepository.findById(key.key(), refIdToUse, repositoryReferenceType).isEmpty()) {
+                return;
+            }
+            deleteParameterAndInvalidateCache(key.key(), refIdToUse, repositoryReferenceType, executionContext);
+        } catch (TechnicalException ex) {
+            throw new TechnicalManagementException(
+                "An error occurs while trying to delete parameter with key: " +
+                    key.key() +
+                    " on scope: " +
+                    referenceType +
+                    " for reference: " +
+                    refIdToUse,
+                ex
+            );
+        }
+    }
+
+    /**
+     * Deletes a parameter and invalidates its cache locally and across the cluster. Shared by the value-less
+     * {@code save(...)} path and {@link #delete(ExecutionContext, Key, String, io.gravitee.rest.api.model.parameters.ParameterReferenceType)}
+     * so the delete + invalidation sequence stays in one place.
+     */
+    private void deleteParameterAndInvalidateCache(
+        String key,
+        String referenceId,
+        ParameterReferenceType referenceType,
+        ExecutionContext executionContext
+    ) throws TechnicalException {
+        parameterRepository.delete(key, referenceId, referenceType);
+        cache.invalidate(computeCacheKey(key, referenceId, referenceType));
+        sendInvalidateParameterCacheCommand(key, referenceId, referenceType, executionContext);
     }
 
     @Override
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ConfigServiceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ConfigServiceTest.java
@@ -92,6 +92,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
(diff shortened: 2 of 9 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ConfigServiceTest.java
```diff
@@ -315,6 +315,110 @@ void shouldNotMarkBrandedSendersReadonlyWhenNotConfigured() {
...
+
+        PortalSettingsEntity portalSettings = configService.resetPortalBrandedSenders(GraviteeContext.getExecutionContext());
+
+        verify(mockParameterService).delete(
+            eq(GraviteeContext.getExecutionContext()),
+            eq(Key.EMAIL_BRANDED_SENDERS),
+            eq("DEFAULT"),
+            eq(ParameterReferenceType.ENVIRONMENT)
+        );
+        assertThat(portalSettings).isNotNull();
```
carlos-andres-osorio: `shouldResetPortalBrandedSendersByDeletingEnvParameterAndReturningRefreshedSettings` asserts only `isNotNull()` on the returned settings; it never asserts `getEmail().isBrandedSendersInherited()` is true after the reset. No test proves the real `delete` -> `getPortalSettings` -> `applyBrandedSendersInheritance` path actually produces `brandedSendersInherited=true` — a regression that made reset return non-inherited settings would pass every current test.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
```diff
@@ -720,6 +718,53 @@ private String getDefaultParameterValue(Key key) {
         return key.defaultValue();
     }
 
+    @Override
+    public boolean existsOnScope(Key key, String referenceId, io.gravitee.rest.api.model.parameters.ParameterReferenceType referenceType) {
```
carlos-andres-osorio, line 722: `existsOnScope`'s `TechnicalException` -> `TechnicalManagementException` catch branch is untested. Existing tests only cover the present/absent happy paths — no test forces `parameterRepository.findById` to throw.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
```diff
@@ -720,6 +718,53 @@ private String getDefaultParameterValue(Key key) {
...
+        } catch (TechnicalException e) {
+            throw new TechnicalManagementException(
+                "An error occurs while checking parameter existence with key: " + key.key() + " on scope: " + referenceType,
+                e
+            );
+        }
+    }
+
+    @Override
+    public void delete(
```
carlos-andres-osorio, line 734: `delete`'s `TechnicalException` -> `TechnicalManagementException` catch branch is untested — neither delete test makes the repository throw, so the catch block is never exercised.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
```diff
@@ -720,6 +718,53 @@ private String getDefaultParameterValue(Key key) {
...
+    ) {
+        String refIdToUse = getEffectiveReferenceId(executionContext, referenceId, referenceType);
+        ParameterReferenceType repositoryReferenceType = ParameterReferenceType.valueOf(referenceType.name());
+        try {
+            if (parameterRepository.findById(key.key(), refIdToUse, repositoryReferenceType).isEmpty()) {
+                return;
+            }
+            deleteParameterAndInvalidateCache(key.key(), refIdToUse, repositoryReferenceType, executionContext);
+        } catch (TechnicalException ex) {
+            throw new TechnicalManagementException("An error occurs while trying to delete parameter for key: " + key, ex);
```
carlos-andres-osorio: The `delete` failure message (`"An error occurs while trying to delete parameter for key: " + key`) omits the scope (`referenceType`) and the resolved `referenceId`, unlike the sibling `existsOnScope` message which includes the scope. A parameter with the same key can exist at ENVIRONMENT, ORGANIZATION, and SYSTEM scope; when a delete fails the log entry can't tell which scope was targeted.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ParameterServiceTest.java
```diff
@@ -1093,4 +1095,71 @@ public void shouldSendInvalidateCacheCommandOnDelete() throws TechnicalException
...
+            EMAIL_BRANDED_SENDERS,
+            "DEFAULT",
+            io.gravitee.rest.api.model.parameters.ParameterReferenceType.ENVIRONMENT
+        );
+
+        assertFalse(exists);
+    }
+
+    @Test
+    public void shouldDeleteParameterOnScopeAndInvalidateCacheWhenPresent() throws TechnicalException {
```
carlos-andres-osorio, line 1136: The `delete` tests always pass a concrete `referenceId` (`"DEFAULT"`), so the `getEffectiveReferenceId` null-resolution branch (`refIdToUse == null` -> `executionContext.getEnvironmentId()`) is uncovered for the delete path. Low risk today since `resetPortalBrandedSenders` always passes a concrete `environmentId`, but the contract divergence between `delete` (context-resolving) and `existsOnScope` (non-resolving, documented) is exactly the kind of asymmetry a test should pin.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ConfigServiceTest.java
```diff
@@ -315,6 +315,110 @@ void shouldNotMarkBrandedSendersReadonlyWhenNotConfigured() {
         assertThat(readonly == null ? List.<String>of() : readonly).doesNotContain(Key.EMAIL_BRANDED_SENDERS.key());
     }
 
+    @Test
+    void shouldMarkBrandedSendersInheritedWhenNoEnvOverrideAndNotSystemConfigured() {
```
carlos-andres-osorio: Nit: The three branded-senders-inherited tests (`shouldMarkBrandedSendersInheritedWhenNoEnvOverrideAndNotSystemConfigured`, `shouldNotMarkBrandedSendersInheritedWhenEnvOverrideExists`, `shouldNotMarkBrandedSendersInheritedWhenSystemConfigured`) share the same assertion structure and differ only in inputs and the expected boolean — worth collapsing into a single `@ParameterizedTest`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ParameterServiceTest.java
```diff
@@ -1093,4 +1095,71 @@ public void shouldSendInvalidateCacheCommandOnDelete() throws TechnicalException
         assertEquals(List.of(CommandTags.PARAMETER_CACHE_UPDATE.name()), command.getTags());
         assertTrue(command.getContent().contains(PORTAL_TOP_APIS.key()));
     }
+
+    @Test
+    public void shouldReportParameterExistsOnScopeWhenPresent() throws TechnicalException {
```
carlos-andres-osorio: Nit: `shouldReportParameterExistsOnScopeWhenPresent` and `shouldReportParameterAbsentOnScopeWhenNotPresent` share identical structure and differ only in the `findById` stub (`of(parameter)` vs `empty()`) and the expected boolean — worth collapsing into a single `@ParameterizedTest`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ConfigServiceTest.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ParameterServiceTest.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ConfigServiceTest.java
paulatulis: Addressed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ParameterServiceTest.java
paulatulis: Addressed

carlos-andres-osorio (approved):

PR #18491, opened 2026-07-09
feat(console): add branded-senders inherited badge and reset-to-org action

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14612
## Description

Frontend half of APIM-14612 — completes the Env-scope branded-senders UX that S4 (APIM-14191) couldn't ship without backend support. Adds an "Inherited from Org" indicator and an explicit "Reset to Org settings" action, built against the contract locked with the backend PR.

Changes

- BrandedSendersComponent (shared): new inheritedFromOrg signal input → per-configuration "Inherited from Org" gio-badge. Rendered per-card, so it naturally shows nothing when the list is empty (a true flag doesn't imply the Org has a non-empty config).
- Model/service: brandedSendersInherited?: boolean on PortalSettingsEmail; PortalSettingsService.resetBrandedSenders() → POST {env}/settings/email/branded-senders/reset + env-settings reload.
- Portal (Env) settings page: threads the flag into the embed; adds "Reset to Org settings" — shown only when there's an editable Env override and the user has ENVIRONMENT_SETTINGS[U]. It calls the dedicated reset endpoint (not save) and reloads, so the form falls back to the Org values instead of persisting an empty override.

Tests

Shared component (badge: shown/empty/not-inherited) + Storybook InheritedFromOrg variant; service (resetBrandedSenders); portal component (reset button visibility × 4, reset-routes-to-reset-not-save, empty-save-still-writes-override). Prettier + ESLint clean.
## Additional context

https://github.com/user-attachments/assets/6da0064d-066c-46f8-b460-c417c9a20e48

### DIFF
--- a/gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
@@ -24,6 +24,7 @@ import { MatIconTestingModule } from '@angular/material/icon/testing';
 import { InteractivityChecker } from '@angular/cdk/a11y';
 import { MatButtonHarness } from '@angular/material/button/testing';
 import {
+  GioConfirmDialogHarness,
   GioFormTagsInputHarness,
   GioLicenseService,
   GioLicenseTestingModule,
@@ -32,12 +33,14 @@ import {
 } from '@gravitee/ui-particles-angular';
 import { MatInputHarness } from '@angular/material/input/testing';
 import { MatSlideToggleHarness } from '@angular/material/slide-toggle/testing';
+import { SpanHarness } from '@gravitee/ui-particles-angular/testing';
 import { of } from 'rxjs';
 
 import { PortalSettingsComponent } from './portal-settings.component';
 import { PortalSettingsModule } from './portal-settings.module';
 import { PortalSettingsHarness } from './portal-settings.harness';
 
+import { SnackBarService } from '../../../services-ngx/snack-bar.service';
 import { CONSTANTS_TESTING, GioTestingModule } from '../../../shared/testing';
 import { GioTestingPermission, GioTestingPermissionProvider } from '../../../shared/components/gio-permission/gio-permission.service';
 import { fakePortalSettings } from '../../../entities/portal/portalSettings.fixture';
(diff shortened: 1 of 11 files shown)


gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.ts
```diff
@@ -819,6 +819,33 @@ export class PortalSettingsComponent implements OnInit {
...
+    const isEditable = !!email?.enabled && !this.isReadonly('email.branded_senders');
+    return this.hasEnvironmentSettingsUpdatePermission && isEditable && hasEnvironmentOverride;
+  }
+
+  resetBrandedSenders(): void {
+    this.portalSettingsService
+      .resetBrandedSenders()
+      .pipe(
+        tap(() => this.snackBarService.success('Branded senders reset to the organization configuration.')),
+        catchError(({ error }) => {
```
carlos-andres-osorio, line 874: `resetBrandedSenders`'s `catchError` branch (and the success path) is untested — a regression could swallow errors or skip the settings re-fetch undetected. The only reset test issues the POST but never flushes a response, so neither the success `tap` (snackbar + `ngOnInit` re-fetch) nor the error branch is exercised.

A test that flushes the reset request with an error status and asserts the error snackbar, plus one that flushes success and asserts the success snackbar + re-fetch, would close this gap.

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
```diff
@@ -711,6 +711,86 @@ describe('PortalSettingsComponent', () => {
...
+    it('should hide the reset button when the inheritance flag is absent from the response', async () => {
+      await init();
+      const settings = overriddenSettings();
+      delete settings.email.brandedSendersInherited;
+      expectPortalSettingsGetRequest(settings);
+
+      expect(await resetButton()).toBeNull();
+    });
+
+    it('should reset via the dedicated endpoint, not through save', async () => {
```
carlos-andres-osorio, line 770: Nit: No test verifies the post-reset re-fetch actually re-displays the org-inherited configuration with the badge. `'should reset via the dedicated endpoint, not through save'` stops at asserting the POST is issued and never flushes the reset response, so the refresh (GET `/settings`), the fallback to the inherited configurations, and the badge re-appearance are all unverified end-to-end. The individual pieces are covered in isolation (badge in `branded-senders.component.spec`, refetch in the service spec), but the full reset-then-re-display flow isn't.

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.ts
```diff
@@ -819,6 +819,33 @@ export class PortalSettingsComponent implements OnInit {
...
+          this.snackBarService.error(error?.message ?? 'An error occurred while resetting the branded senders.');
+          return EMPTY;
+        }),
+        takeUntilDestroyed(this.destroyRef),
+      )
+      .subscribe(() => this.ngOnInit());
```
carlos-andres-osorio, line 880: `resetBrandedSenders()` performs the backend mutation and then calls `this.ngOnInit()`, which rebuilds the whole portal form. Any unrelated unsaved edits elsewhere in the form (e.g. host, subject) would be silently discarded on that reload, with no confirmation. This mirrors the existing tag add/edit/delete handlers' pattern, so it may be an accepted convention here too — but since the reset action isn't part of the save-bar flow, the discard could be less expected to a user. Is this intentional, or worth a confirm-dialog gate when the form is dirty?

carlos-andres-osorio:
This PR is missing the `docs-needed`/`no-docs` label. Can you confirm which applies and add it?

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.ts
paulatulis: Added test

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
paulatulis: Adressed

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.ts
paulatulis: Added dialog

carlos-andres-osorio (approved):

PR #18507, opened 2026-07-11
feat(rest-api): enforce Kafka virtual cluster lifecycle guards

callaertanthony:
## Context

A Kafka Virtual Cluster (VC) could be deployed with an empty `backends` list, and its backends
could be emptied or the VC undeployed while Kafka Native APIs were still bound to it and running.
The client then connects successfully but gets an empty topic list, with nothing surfaced to the
operator.

## Approach — guards, not cascades

A started native API must never be left pointing at a backend-less / undeployed virtual cluster.
Rather than auto-stopping APIs, dangerous VC operations are **blocked** with a clear `400`.

| Rule | REST behaviour |
|------|----------------|
| A VC cannot be deployed with no backends | `POST /clusters/{id}/_deploy` → **400** |
| Undeploy, or emptying backends, is blocked while a **started** bound API exists | `POST /clusters/{id}/_undeploy` and `PUT /clusters/{id}` → **400** (message lists the APIs) |
| Emptying backends on a deployed VC with **no** started bound API auto-undeploys it | `PUT /clusters/{id}` → **200**, `lifecycleState = UNDEPLOYED` |
| Starting/deploying a native API bound to a not-deployed VC is blocked | `POST /apis/{id}/_start` and `/deployments` → **400** |

## Changes

- New `@DomainService`s (auto-wired via component scan, no Spring config changes):
  - `VirtualClusterBoundApisQueryService` — resolves started native APIs bound to a VC.
  - `UndeployClusterDomainService` — shared undeploy + event + audit, reused by the explicit
    undeploy and the auto-undeploy-on-empty path.
  - `ValidateApiClusterBindingService` — the start/deploy guard, called from `ApiStateServiceImpl`.
- Guards added to `DeployClusterUseCase`, `UndeployClusterUseCase`, `UpdateClusterUseCase`.
- API↔VC binding = native endpoint type `native-kafka-virtual-cluster`,
  `configuration.virtualClusterCrossId`.

## For the UI (Console / Gamma ESM) — separate repo

Backend-only PR. The lifecycle rules and expected UI behaviour are captured in the table above so a
UI-repo agent can implement matching guards/banners. UI checks are UX sugar; the backend 400s are
the source of truth.

## Testing

`mvn -pl gravitee-apim-rest-api/gravitee-apim-rest-api-service test` — the affected classes run
green (`Tests run: 120, Failures: 0, Errors: 0`), covering every rule above plus the unchanged
start/deploy paths. New unit tests: `VirtualClusterBoundApisQueryServiceTest`,
`ValidateApiClusterBindingServiceTest`, and additions to the three cluster use-case tests and
`ApiStateServiceImplTest`.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
@@ -496,6 +496,15 @@ paths:
             responses:
                 "202":
                     description: API deployment request received
+                "400":
+                    description: |-
+                        The API cannot be deployed. This includes a Kafka Native API bound (via a
+                        `native-kafka-virtual-cluster` endpoint) to a Kafka virtual cluster that is
+                        not currently deployed.
+                    content:
+                        application/json:
+                            schema:
+                                $ref: "#/components/schemas/Error"
                 default:
                     $ref: "#/components/responses/Error"
     /environments/{envId}/apis/{apiId}/deployments/current:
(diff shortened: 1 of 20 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/cluster/VIRTUAL_CLUSTER_LIFECYCLE.md
ThibaudAV, line 1: Are we keeping this file? Or is it just a working temp file?

callaertanthony: intention was to keep it, but I can keep it outside

ThibaudAV (approved):

PR #18509, opened 2026-07-13
Perf/gateway subscription cache scale

ThibaudAV:
## Issue

Gateway-side RCA of a customer performance incident on 4.11.x (CPU spikes, heap growth, slow TLS handshakes under load, degraded at ~3.3M active subscriptions).

## Description

Reduce the memory footprint and allocation pressure of `SubscriptionCacheService` at large subscription counts, and fix a thread-safety issue in its per-API key index:

- **perf** — store the exploded-subscription cache (`cacheBySubscriptionIdAll`) as immutable, usually singleton, sets replaced atomically via `compute()` instead of one `ConcurrentHashMap`-backed keySet per subscription id (~5× smaller per entry at multi-million-subscription scale). Reads no longer make defensive copies (`getAllById`), and the API-key request hot path (`getByApiAndId`) iterates instead of allocating a Stream pipeline per request.
- **fix** — `updateCacheKeyByApiId()` / `evictKeyForApi()` used a get/mutate/put pattern on a plain `HashSet`. Sync appenders register subscriptions of the same API from several threads (`services.sync.appender.parallelism` defaults to 4 since 4.11.9), so concurrent calls could lose updates or corrupt the set. Maintenance is now atomic via `compute()` with a concurrent set.
- **test** — cover the exploded API-key lookup path and add a concurrency regression test (concurrent `register()` for the same API then `unregisterByApiId()` must leave no entry behind — fails on the previous code).

Behavior-preserving: full `gravitee-apim-gateway-handlers-api` test suite green (916 tests).

## Additional context

Targets one contributor identified by the incident analysis (subscription-cache footprint/GC pressure); the other findings of the RCA (event-loop count, V3 policy offload, sync warmup defaults) are tracked separately.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
@@ -178,7 +178,10 @@ public class SubscriptionCacheService implements SubscriptionService {
     @Override
     public void unregisterByApiId(final String apiId) {
         log.debug("Unregistering all subscriptions for API [{}]", apiId);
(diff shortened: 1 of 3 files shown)


ThibaudAV:
Superseded by #18513, which carries this branch's commits plus the follow-up waves in a single reviewable stack.

PR #18510, opened 2026-07-13
feat(gamma-platform): add view plan security type catalog

pragsty:
## Issue

https://gravitee.atlassian.net/browse/FOUND-49

## Description
This adds a read-only reference page where users can see which plan security types the platform supports, including their security type ID, display name, and policy name.

**What's included**
New Plan Security Catalog page under Platform → System & Security
Read-only table with columns:
Security Type ID
Name
Policy Name
Hardcoded catalog entries:
API Key (API_KEY / api-key)
OAuth2 (OAUTH2 / oauth2)
JWT (JWT / jwt)
Keyless (KEY_LESS / key-less)

image


pragsty:
Not needed any more hence closing

PR #18523, opened 2026-07-14
fix: show out of sync status for all apis in api list page

vikrantgravitee:
## Issue

https://gravitee.atlassian.net/browse/APIM-14633

## Description

On the APIs list page in APIM Console, APIs that need redeployment did not show the out-of-sync refresh icon for V4 (and Federated) APIs, even though the API detail page correctly showed the “This API is out of sync” banner.

## Root Cause

The APIs list page and the API detail page used different ways to decide if an API was out of sync.

On the detail page, it looked at a field on the API called deploymentState. If that said the API needed redeploy, it showed the warning.
On the list page, for V4 and Federated APIs, that check was never done. The code always treated them as “in sync,” so the refresh icon never appeared.
For V2 APIs, the list page did check sync status, but by making a separate request per API, not by using the same deploymentState field.
So the backend already knew when V4 APIs were out of sync, and the detail page showed it correctly. The list page simply ignored that information for V4 (and Federated) APIs.

https://github.com/user-attachments/assets/e3b5618d-79b5-40ea-bc6e-4630bf7844f2

## Fix

https://github.com/user-attachments/assets/eb06cedf-b3fc-4c26-a1ae-d71d1817a423

### DIFF
--- a/gravitee-apim-console-webui/src/shared/utils/api-access.util.ts
+++ b/gravitee-apim-console-webui/src/shared/utils/api-access.util.ts
@@ -52,3 +52,5 @@ export function getApiContextPath(api: Api | null | undefined): string | null {
   const access = getApiAccess(api as ApiV4 | ApiV2);
   return access?.[0] ?? null;
 }
+
+export const isApiOutOfSync = (api: Api): boolean => api.deploymentState === 'NEED_REDEPLOY';
--- a/gravitee-apim-console-webui/src/management/api/list/api-list.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/api/list/api-list.component.spec.ts
@@ -233,7 +233,13 @@ describe('ApisListComponent', () => {
 
         await loader.getHarness(GioTableWrapperHarness).then(tableWrapper => tableWrapper.setSearchValue('bad-search'));
         await tick(400);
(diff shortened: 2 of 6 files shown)


gravitee-apim-console-webui/src/shared/utils/api-access.util.ts
```diff
@@ -52,3 +52,11 @@ export function getApiContextPath(api: Api | null | undefined): string | null {
   const access = getApiAccess(api as ApiV4 | ApiV2);
   return access?.[0] ?? null;
 }
+
+export const isApiOutOfSync = (api: Api): boolean | undefined => {
+  if (api.deploymentState == null) {
+    return undefined;
+  }
+
+  return api.deploymentState === 'NEED_REDEPLOY';
```
ankita-gupta21: The undefined vs false distinction doesn’t matter for the list UI — both mean “don’t show the out-of-sync icon.” We can just check deploymentState === 'NEED_REDEPLOY' like the detail page does.
ex:
```
export const isApiOutOfSync = (api: Api): boolean =>
  api.deploymentState === 'NEED_REDEPLOY';
```

ankita-gupta21 (approved):

PR #18535, opened 2026-07-14
feat(repository): add portal category persistence (JDBC + MongoDB)

dev-marek:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-16

## Description

Adds the persistence layer for Portal Next categories: repository model + interface, JDBC implementation with a new Liquibase changelog (`portal_categories` table), MongoDB implementation with a new `environmentId`+`title` index upgrader, and repository tests run against both backends.

## Additional context

Part 1/7 of a stacked chain for PORTAL-16. **Draft — do not review or merge until the full chain lands.**

PR #18536, opened 2026-07-14
feat(rest-api): add portal category core crud/query services

dev-marek:
## Issue

## Description

Adds the core domain model (`PortalCategory`, `CreatePortalCategory`, `UpdatePortalCategory`), `crud_service`/`query_service` interfaces, their infra implementations, and the repository adapter — following the hexagonal layering established by the sibling `portal_page` (navigation items) feature.

## Additional context

Part 2/7 of a stacked chain for PORTAL-16, stacked on #18535 (repository layer).

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_category/model/PortalCategory.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_category/model/PortalCategory.java
@@ -43,7 +43,6 @@ public class PortalCategory {
         this.visible = visible;
     }
 
-    /** New portal category with a random id. */
     public static PortalCategory create(String environmentId, String title, String description, boolean visible) {
         return new PortalCategory(PortalCategoryId.random(), environmentId, title, description, visible);
     }
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_category/model/PortalCategory.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_category/model/PortalCategory.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal_category.model;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * @author GraviteeSource Team
+ */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder(toBuilder = true)
+public class PortalCategory {
+
+    private String id;
+    private String environmentId;
+    private String title;
+    private String description;
+    private boolean visible;
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_category/model/PortalCategory.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_category/model/PortalCategory.java
@@ -15,23 +15,60 @@
  */
 package io.gravitee.apim.core.portal_category.model;
 
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.Data;
-import lombok.NoArgsConstructor;
+import jakarta.annotation.Nonnull;
+import lombok.Getter;
 
 /**
  * @author GraviteeSource Team
  */
-@Data
-@NoArgsConstructor
-@AllArgsConstructor
-@Builder(toBuilder = true)
+@Getter
 public class PortalCategory {
 
-    private String id;
-    private String environmentId;
+    @Nonnull
+    private final PortalCategoryId id;
+
+    @Nonnull
+    private final String environmentId;
+
     private String title;
     private String description;
     private boolean visible;
+
+    /** Raw constructor - only used for database mapping, use {@link #create} or {@link #of} instead. */
+    private PortalCategory(PortalCategoryId id, String environmentId, String title, String description, boolean visible) {
+        this.id = id;
+        this.environmentId = environmentId;
+        this.title = title;
+        this.description = description;
+        this.visible = visible;
+    }
+
+    /** New portal category with a random id. */
+    public static PortalCategory create(String environmentId, String title, String description, boolean visible) {
+        return new PortalCategory(PortalCategoryId.random(), environmentId, title, description, visible);
+    }
+
+    /** Reconstitute a portal category from a known id (database mapping). */
+    public static PortalCategory of(PortalCategoryId id, String environmentId, String title, String description, boolean visible) {
+        return new PortalCategory(id, environmentId, title, description, visible);
+    }
+
+    public void update(UpdatePortalCategory update) {
+        this.title = update.getTitle();
+        this.description = update.getDescription();
+        this.visible = update.isVisible();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        PortalCategory that = (PortalCategory) o;
+        return id.equals(that.id);
+    }
+
+    @Override
+    public int hashCode() {
+        return id.hashCode();
+    }
 }
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalCategoryAdapter.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalCategoryAdapter.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.infra.adapter;
+
+import io.gravitee.apim.core.portal_category.model.PortalCategory;
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+/**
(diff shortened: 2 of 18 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/inmemory/PortalCategoryQueryServiceInMemory.java
```diff
@@ -0,0 +1,53 @@
...
+    final ArrayList<PortalCategory> storage = new ArrayList<>();
+
+    @Override
+    public List<PortalCategory> findByEnvironmentId(String environmentId) {
+        return storage
+            .stream()
+            .filter(pc -> environmentId.equals(pc.getEnvironmentId()))
+            .sorted(Comparator.comparing(PortalCategory::getTitle))
+            .toList();
+    }
```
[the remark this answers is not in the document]
dev-marek: Not applicable — sorting is already performed at the repository query level (JDBC `ORDER BY title` in JdbcPortalCategoryRepository, `Sort.by("title")` in MongoPortalCategoryRepository), so `PortalCategoryQueryServiceImpl` correctly relies on the repository's contract rather than re-sorting in Java. The in-memory fake's `.sorted(...)` call is intentional: it emulates that repository-level guarantee so tests exercise realistic ordering. Keeping it as-is.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/crud_service/portal_category/PortalCategoryCrudServiceImpl.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private final PortalCategoryRepository portalCategoryRepository;
+
+    public PortalCategoryCrudServiceImpl(@Lazy PortalCategoryRepository portalCategoryRepository) {
+        this.portalCategoryRepository = portalCategoryRepository;
+    }
+
+    @Override
+    public PortalCategory create(PortalCategory portalCategory) {
+        try {
```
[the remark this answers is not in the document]
dev-marek: @AikidoSec ignore: this repeats the exact pattern used by every sibling crud_service impl in this codebase (e.g. PortalCrudServiceImpl: 5 methods, same per-method try/catch translating TechnicalException to TechnicalDomainException with a method-specific message, no shared helper). Extracting a helper here would be the odd one out, not the fix — keeping it consistent with the established convention.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/inmemory/PortalCategoryQueryServiceInMemory.java
```diff
@@ -0,0 +1,53 @@
...
+public class PortalCategoryQueryServiceInMemory implements PortalCategoryQueryService, InMemoryAlternative<PortalCategory> {
+
+    final ArrayList<PortalCategory> storage = new ArrayList<>();
+
+    @Override
+    public List<PortalCategory> findByEnvironmentId(String environmentId) {
+        return storage
+            .stream()
+            .filter(pc -> environmentId.equals(pc.getEnvironmentId()))
+            .sorted(Comparator.comparing(PortalCategory::getTitle))
```
thomasballivet, line 36: we don't have this sort in the real implementation, either we should remove this sort or do it there

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_category/model/PortalCategory.java
```diff
@@ -0,0 +1,73 @@
...
+    @Nonnull
+    private final String environmentId;
+
+    private String title;
+    private String description;
+    private boolean visible;
+
+    /** Raw constructor - only used for database mapping, use {@link #create} or {@link #of} instead. */
+    private PortalCategory(PortalCategoryId id, String environmentId, String title, String description, boolean visible) {
+        this.id = id;
```
thomasballivet, line 39: we can prevent from null values here with  `Objects.requireNonNull(id);` on `id` and `environmentId` in addition to `@Nonnull` annotations

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalCategoryAdapter.java
```diff
@@ -0,0 +1,55 @@
...
+        return PortalCategory.of(
+            PortalCategoryId.of(repository.getId()),
+            repository.getEnvironmentId(),
+            repository.getTitle(),
+            repository.getDescription(),
+            repository.isVisible()
+        );
+    }
+
+    default io.gravitee.repository.management.model.PortalCategory toRepository(PortalCategory domain) {
```
thomasballivet: We can use MapStruct here, and let it do the mapping : 

```suggestion
@Mapper
public interface PortalCategoryAdapter {
    PortalCategoryAdapter INSTANCE = Mappers.getMapper(PortalCategoryAdapter.class);

    io.gravitee.repository.management.model.PortalCategory toRepository(PortalCategory domain);

    default String map(PortalCategoryId id) {
        return id.toString();
    }
}
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/crud_service/portal_category/PortalCategoryCrudServiceImplTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+        @Test
+        @SneakyThrows
+        void should_create_a_portal_category() {
+            var category = PortalCategoryFixtures.aPortalCategory();
+            when(repository.create(any())).thenAnswer(invocation -> invocation.getArgument(0));
+
+            var result = service.create(category);
+
+            assertThat(result).isEqualTo(category);
```
thomasballivet: `equals` here will compare only on the `id`, which is good except that here we would verify everything.
```suggestion
            assertThat(result).usingRecursiveComparison().isEqualTo(category);
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/crud_service/portal_category/PortalCategoryCrudServiceImplTest.java
```diff
@@ -0,0 +1,165 @@
...
+        @Test
+        @SneakyThrows
+        void should_return_portal_category_and_adapt_it() {
+            var id = PortalCategoryFixtures.PORTAL_CATEGORY_ID;
+            var category = PortalCategoryFixtures.aPortalCategory();
+            when(repository.findById(id.toString())).thenReturn(Optional.of(PortalCategoryAdapter.INSTANCE.toRepository(category)));
+
+            var result = service.get(id);
+
+            assertThat(result).contains(category);
```
thomasballivet: Same as above 

```suggestion
            assertThat(result).get().usingRecursiveComparison().isEqualTo(category);
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/query_service/portal_category/PortalCategoryQueryServiceImplTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+        @Test
+        @SneakyThrows
+        void should_return_categories_adapted_from_the_repository() {
+            var category = PortalCategoryFixtures.aPortalCategory();
+            when(repository.findAllByEnvironmentId("env")).thenReturn(List.of(PortalCategoryAdapter.INSTANCE.toRepository(category)));
+
+            var result = service.findByEnvironmentId("env");
+
+            assertThat(result).containsExactly(category);
```
thomasballivet: Same as above
```suggestion
            assertThat(result).usingRecursiveFieldByFieldElementComparator().containsExactly(category);
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/inmemory/PortalCategoryQueryServiceInMemory.java
dev-marek: Good catch, this was ambiguous. Sorting does happen in production, just one layer down: `JdbcPortalCategoryRepository` does `ORDER BY title` and `MongoPortalCategoryRepository` does `Sort.by("title")`, per the story's requirement that categories be alphabetically sorted at the repository layer. `PortalCategoryQueryServiceImpl` doesn't re-sort because it doesn't need to. The in-memory fake sorts in Java specifically to reproduce that repository-level contract, since it has no real `ORDER BY` to fall back on. Added a comment above the `.sorted(...)` call explaining this so it doesn't come up again.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_category/model/PortalCategory.java
dev-marek: Decided to leave this as `@Nonnull` only, no `Objects.requireNonNull`. Team convention for core package code: we only need static analysis there (format/null-safety annotations), since we assume inputs entering core are already validated by the layer that constructs them (the `create`/`of` factories here). Adding runtime null checks on top would be defensive coding against a case that shouldn't be reachable given that contract.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalCategoryAdapter.java
dev-marek: Went with a partial version of this — applied MapStruct auto-mapping to `toRepository` (via a `default String map(PortalCategoryId id)` converter for the id, since all other getters already line up by name with the repository model's builder), but kept `toModel` hand-written. MapStruct can't auto-construct `PortalCategory` for that direction: its constructor is deliberately private (to prevent creating it as a plain POJO, keeping construction funneled through `create`/`of`), so there's no public constructor, setter, or builder on the domain type for MapStruct to target without extra `@ObjectFactory` wiring. Left a comment on `toModel` explaining why it stays hand-written so this doesn't come up again either.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/crud_service/portal_category/PortalCategoryCrudServiceImplTest.java
dev-marek: Good catch — applied, using `usingRecursiveComparison().isEqualTo(category)`. Since `PortalCategory.equals()` is identity-only by design, this assertion wasn't actually verifying title/description/visible round-tripped correctly through `create()`. Fixed the same blind spot in the `update` test right above it too, since it had the identical pattern.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/crud_service/portal_category/PortalCategoryCrudServiceImplTest.java
dev-marek: Applied — `assertThat(result).get().usingRecursiveComparison().isEqualTo(category)` since `result` here is an `Optional<PortalCategory>`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/query_service/portal_category/PortalCategoryQueryServiceImplTest.java
dev-marek: Applied — `usingRecursiveFieldByFieldElementComparator().containsExactly(category)`.

thomasballivet (approved):

PR #18537, opened 2026-07-14
feat(rest-api): add portal category domain service and use cases

dev-marek:
## Issue

## Description

Adds `PortalCategoryDomainService` (title validation, uniqueness-on-create, not-found handling) and the four use cases (Create/Update/Delete/List), with unit tests using in-memory fakes.

## Additional context

Part 3/7 of a stacked chain for PORTAL-16, stacked on #18536 (core crud/query services).


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_category/use_case/UpdatePortalCategoryUseCase.java
```diff
@@ -0,0 +1,50 @@
...
+@UseCase
+@RequiredArgsConstructor
+public class UpdatePortalCategoryUseCase {
+
+    private final PortalCategoryDomainService portalCategoryDomainService;
+    private final PortalCategoryCrudService portalCategoryCrudService;
+
+    public Output execute(Input input) {
+        var updatePortalCategory = input.updatePortalCategory();
+        portalCategoryDomainService.validateTitle(updatePortalCategory.getTitle());
```
thomasballivet: We should probably check existence first, so if the category doesn't exist we throw a `PortalCategoryNotFoundException` (404 Not Found) instead of a `ValidationDomainException` (400 Bad Request).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/portal_category/PortalCategoryDomainServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+@Service
+@RequiredArgsConstructor
+public class PortalCategoryDomainServiceImpl implements PortalCategoryDomainService {
+
+    private final PortalCategoryCrudService portalCategoryCrudService;
+
+    @Override
+    public void validateTitle(String title) {
+        if (title == null || title.isBlank()) {
+            throw new ValidationDomainException("Portal category title cannot be blank");
```
thomasballivet: do we want to check if the title exist already, or isn't it necessary?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/portal_category/PortalCategoryDomainServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+import io.gravitee.apim.core.portal_category.model.PortalCategoryId;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+/**
+ * @author GraviteeSource Team
+ */
+@Service
+@RequiredArgsConstructor
+public class PortalCategoryDomainServiceImpl implements PortalCategoryDomainService {
```
thomasballivet: we can use `@DomainService` annotation instead as there will be only one implementation, we could skip the interface: single concrete class in core with `@DomainService`, wdyt?

dev-marek: Note for AI agent: we split domain service into interface/implementation if and only if we need to reach outside of onion architecture (infra) eg. repositories or legacy spaghetti code services

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_category/use_case/UpdatePortalCategoryUseCase.java
dev-marek: Good catch — swapped the order: `execute()` now calls `findByIdAndEnvironmentId` first, then `validateTitle`. Updating a non-existent category with a blank title now correctly returns 404 instead of 400. Added two regression tests covering both orderings (unknown id + blank title -> 404; existing category + blank title -> 400).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/portal_category/PortalCategoryDomainServiceImpl.java
dev-marek: Not necessary — this was a deliberate call made during design: duplicate category titles are allowed, the story doesn't require uniqueness. `validateTitle` stays a blank-check only.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/portal_category/PortalCategoryDomainServiceImpl.java
dev-marek: Done — collapsed into a single concrete `@DomainService` class directly in `core/portal_category/domain_service/`, no interface, no infra impl. It only ever depended on `PortalCategoryCrudService` (a core abstraction), so per the interface/infra split rule (only needed when reaching outside the onion architecture — repositories, legacy services) there was no reason to keep it split. Moved its test to `core/portal_category/domain_service/PortalCategoryDomainServiceTest` to match, mirroring `DashboardDomainServiceTest`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/portal_category/PortalCategoryDomainServiceImpl.java
dev-marek: Correction to my reply above: title uniqueness **is** required after all — legacy Classic categories also enforced unique names on create (case-insensitive, per-environment). Implemented `validateTitleUniqueness` in `PortalCategoryDomainService`, wired into `CreatePortalCategoryUseCase` only (matching legacy, which doesn't re-check on update), throwing a new `PortalCategoryTitleAlreadyExistsException` (409, via the existing generic `ConflictDomainExceptionMapper`).

thomasballivet (approved):

PR #18539, opened 2026-07-14
feat(console): add portal category entities and API service

dev-marek:
## Issue

## Description

Adds the TypeScript entity types/fixtures and a hand-written Angular service (`PortalCategoryService`) for the `/portal-categories` management-api-v2 endpoints, following the `shared-policy-groups.service.ts` convention.

## Additional context

Part 5/7 of a stacked chain for PORTAL-16, stacked on #18538 (v2 REST endpoints).

### DIFF
--- a/gravitee-apim-console-webui/src/entities/management-api-v2/portalCategory/UpdatePortalCategory.ts
+++ b/gravitee-apim-console-webui/src/entities/management-api-v2/portalCategory/UpdatePortalCategory.ts
@@ -1,0 +1,21 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export interface UpdatePortalCategory {
+  title: string;
+  description?: string;
+  visible: boolean;
+}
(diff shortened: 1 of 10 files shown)


gravitee-apim-console-webui/src/services-ngx/portal-category.service.ts
```diff
@@ -0,0 +1,47 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class PortalCategoryService {
+  constructor(
+    private readonly http: HttpClient,
+    @Inject(Constants) private readonly constants: Constants,
+  ) {}
+
+  list(): Observable<PortalCategory[]> {
```
thomasballivet, line 32: we could also add a `get(id)` to fetch only a specific category. Or do we expect only a few categories and this is wanted?

gravitee-apim-console-webui/src/entities/management-api-v2/portalCategory/UpdatePortalCategory.ts
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export interface UpdatePortalCategory {
+  title: string;
+  description?: string;
+  visible: boolean;
```
thomasballivet: let's align with the rest on `visible?: boolean`. (It is not required in other places)

thomasballivet (approved):

gravitee-apim-console-webui/src/services-ngx/portal-category.service.ts
dev-marek: Decided against adding `get(id)` — there's no `GET /portal-categories/{categoryId}` endpoint on the backend either (only list/create/update/delete), and the edit form already loads a single category by fetching the full list and finding the match client-side, which is deliberate: this is a settings-style list expected to stay small (no manual reordering, no pagination anywhere in the design). Adding `get(id)` would need a new backend endpoint too, which felt like scope creep here — worth revisiting as a follow-up if category counts ever grow enough to matter.

gravitee-apim-console-webui/src/entities/management-api-v2/portalCategory/UpdatePortalCategory.ts
dev-marek: Good catch — fixed. The backend OpenAPI schema for `UpdatePortalCategory` only requires `title`; `visible` is optional there too (defaults to true), so the TS interface was over-constraining relative to the actual contract. Now `visible?: boolean`, matching how classic `UpdateCategory.ts` already treats `hidden` as optional.

PR #18556, opened 2026-07-15
fix: preserve colons in relative endpoint paths (forward-port of #17173)

callaertanthony:
## Problem

On 4.12.x, a V4 endpoint whose relative path contains a colon (e.g. `/foo:analyze`) returns **503 NO_ENDPOINT_FOUND**. The invoker parses `/foo:analyze` as endpoint `foo` + path `analyze` and fails to resolve. Changing the colon to a slash works. Reported on 4.12.0–4.12.3 (APIM-14652).

## Root cause

This is a regression of APIM-14220. The fix (PR #17173) landed on `4.11.x` only and was never forward-ported. `HttpEndpointInvoker.ENDPOINT_PATTERN` still used the pre-fix regex `[^:]+` on `master` and `4.12.x`, so colon-containing relative paths were split at the colon.

## Fix

Cherry-pick of the original fix (`2feddfbc3f`, PR #17173): change the endpoint match group from `[^:]+` to `[^/:][^:]*` so a leading `/` (relative path) no longer matches the endpoint group, preserving colon-containing relative paths as literal paths. Includes the original parameterized tests in `HttpEndpointInvokerTest`.

## Backport

Fix is absent from both `master` and `4.12.x`. This PR targets `master`; `apply-on-4-12-x` label added so Mergify backports to `4.12.x` (the customer's line) after merge.

Closes APIM-14652

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/invoker/HttpEndpointInvoker.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/v4/invoker/HttpEndpointInvoker.java
@@ -46,7 +46,7 @@ public class HttpEndpointInvoker implements HttpInvoker, Invoker {
     private static final String MATCH_GROUP_ENDPOINT = "endpoint";
     private static final String MATCH_GROUP_PATH = "path";
     private static final Pattern ENDPOINT_PATTERN = Pattern.compile(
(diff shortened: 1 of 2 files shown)


callaertanthony:

- Regex `[^/:][^:]*` correctly stops treating every colon as an endpoint/target separator; absolute URLs still fall through to the `KNOWN_URL_SCHEMES` handling. No ReDoS risk (`[^:]*` can't backtrack past the next `:`).
- Minor: a path without a leading slash (e.g. `document:translate`) would still be parsed with `document` as the endpoint name. Acceptable since gateway request paths are normalized to start with `/`.

vikrantgravitee (approved):

PR #18557, opened 2026-07-15
fix(console): gate overview dashboard behind analytics permission

callaertanthony:
## Problem

The environment **Overview dashboard** ignores role permissions. A user whose environment role has no permissions still sees every analytics widget (API/Application counts, lifecycle/state, response status, top APIs, request stats, …) — the same data an admin sees — even though the **Observability** and **Analytics** tabs are correctly blocked for that user.

Only the *API Events* row was gated (`environment-platform-r`); every other widget rendered, and the component fetched their data into the browser unconditionally.

Reported on 4.8.x–4.11.x. (APIM-14486)

## Fix

Gate the whole dashboard on `environment-platform-r` — the same permission that already guards the Observability/Analytics tabs (`management-routing.module.ts`) and the existing API Events widget:

- **`home-overview.component.ts`** — `ngOnInit` returns early when the permission is missing, so **no analytics request is issued** (the parent fetches widget data directly and passes it via `[data]`, so template-only hiding would still leak the data to the browser).
- **`home-overview.component.html`** — one `*gioPermission` gate on the dashboard root; removed the now-redundant per-row gate on API Events.

## Tests

- `home-overview.component.spec.ts` — `init()` now takes an optional permissions arg; added a `without environment-platform-r permission` case asserting the dashboard is hidden and (via `httpTestingController.verify()`) that no analytics request fires.

> Not run locally (no `node_modules` in this worktree) — relying on CI.

## Notes

Kept the gate binary (one permission for the whole dashboard), matching the ticket's expectation that a no-permission user sees nothing — same as the tabs. Per-widget granularity (e.g. Summary counts behind `environment-api-r`) was intentionally left out; can be added later if product wants finer roles.

### DIFF
--- a/gravitee-apim-console-webui/src/management/home/home-overview/home-overview.component.html
+++ b/gravitee-apim-console-webui/src/management/home/home-overview/home-overview.component.html
@@ -15,7 +15,7 @@
     limitations under the License.
 
 -->
-<div class="home-overview">
+<div class="home-overview" *gioPermission="{ anyOf: ['environment-platform-r'] }">
   <div class="time-selector">
     <app-dashboard-filters-bar></app-dashboard-filters-bar>
   </div>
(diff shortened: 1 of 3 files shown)


callaertanthony:

Gating on `environment-platform-r` (matching the Observability/Analytics tabs) and the early return in `ngOnInit` are correct. UX nit: a user without the permission now sees a fully blank dashboard area — consider an empty-state message rather than nothing, if that matches conventions elsewhere in the console.

vikrantgravitee (approved):

PR #18558, opened 2026-07-15
fix(portal): prevent subscription metadata leak to users without applications

callaertanthony:
## Problem

`GET /portal/environments/{envId}/applications` (notably with `order=-nbSubscriptions`) leaked subscription metadata to a portal user who owns **no application**. The response `data` was `[]`, but `metadata.subscriptions` contained ACCEPTED subscriptions from across the environment — including other users' subscription `request` comments.

Root cause: an empty (non-null) authorized-application set was passed to the subscription query. Downstream, an empty `applications` filter is treated as "no filter" (JDBC `addStringsWhereClause` skips empty collections), so the query returned every ACCEPTED subscription in the environment.

## Fix

Guard the two source call-sites that pass a user-authorization-scoped application set, so an empty set means "authorized to see nothing" rather than "no filter":

- `ApplicationsResource.fillMetadata` — returns the metadata unchanged when the scoped set is empty.
- `FilteringServiceImpl.getApplicationsOrderByNumberOfSubscriptions` — short-circuits on an empty/null id set (also avoids a full-table subscription scan).

The shared JDBC criteria builder is intentionally left unchanged: empty-means-all is a deliberate convention used by many admin/management queries. Fixing at the authorization boundary is the correct and narrowest change, and the empty set never reaches the repository.

## Tests

- `ApplicationsResourceTest` (18/18) — adds `shouldNotLeakSubscriptionsMetadataWhenUserHasNoAuthorizedApplication`, which asserts the unfiltered subscription search never runs and no subscription metadata is returned. The existing `shouldGetApplicationsWithSubscriptions` was passing because of the bug; it now authorizes the user explicitly.
- `FilteringServiceImplTest` (3/3, new) — covers the empty, null, and populated id-set paths.

Ref: https://gravitee.atlassian.net/browse/APIM-14585

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/test/java/io/gravitee/rest/api/portal/rest/resource/ApplicationsResourceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/test/java/io/gravitee/rest/api/portal/rest/resource/ApplicationsResourceTest.java
@@ -291,6 +291,15 @@ public class ApplicationsResourceTest extends AbstractResourceTest {
         doReturn(true)
             .when(permissionService)
             .hasPermission(any(), eq(RolePermission.APPLICATION_SUBSCRIPTION), any(), eq(RolePermissionAction.READ));
+        doReturn(new HashSet<>(Arrays.asList("A", "B")))
+            .when(applicationService)
+            .findIdsByUserAndPermission(
+                eq(GraviteeContext.getExecutionContext()),
+                any(),
+                any(),
+                eq(RolePermission.APPLICATION_SUBSCRIPTION),
+                eq(RolePermissionAction.READ)
+            );
         SubscriptionEntity sub1 = createSubscriptionEntity("sub-1", "A");
         SubscriptionEntity sub2 = createSubscriptionEntity("sub-2", "A");
         SubscriptionEntity sub3 = createSubscriptionEntity("sub-3", "B");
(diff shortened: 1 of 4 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/filtering/FilteringServiceImpl.java
```diff
@@ -105,6 +105,12 @@ public Collection<String> getApisOrderByNumberOfSubscriptions(Collection<String>
 
     @Override
     public Collection<String> getApplicationsOrderByNumberOfSubscriptions(Collection<String> ids, Order order) {
+        // Security (APIM-14585): an empty application set must not run an unfiltered ranking query.
```
vikrantgravitee: Same here...

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/test/java/io/gravitee/rest/api/portal/rest/resource/ApplicationsResourceTest.java
```diff
@@ -319,6 +328,32 @@ public void shouldGetApplicationsWithSubscriptions() {
         assertEquals(1, ((List) metadataSubscriptions.get("B")).size());
     }
 
+    @Test
+    public void shouldNotLeakSubscriptionsMetadataWhenUserHasNoAuthorizedApplication() {
+        // APIM-14585: a user who owns no application must not receive other users' subscription metadata.
```
vikrantgravitee: Should we keep ticket number in PR?

vikrantgravitee (approved):

PR #18559, opened 2026-07-15
fix(definition): ignore clearTextUpgrade for HTTP/1.1 endpoints

callaertanthony:
## Problem

When a new endpoint is added to a **V2 API** without touching the protocol settings, the exported definition contains the invalid combination:

```json
"clearTextUpgrade": true, "version": "HTTP_1_1"
```

`clearTextUpgrade` (h2c) is an **HTTP/2-only** upgrade mechanism and has no meaning for HTTP/1.1. (APIM-14613, affects 4.8.x–4.11.x)

## Root cause

- The console leaves `clearTextUpgrade` undefined for a new endpoint (and only disables the control for HTTP/1.1), so it does **not** send `true`.
- The value comes from the backend model default `DEFAULT_CLEAR_TEXT_UPGRADE = true`, combined with the default protocol `HTTP_1_1`.
- The custom `HttpClientOptionsSerializer` already omits `clearTextUpgrade`/`version` for HTTP/1.1, but the export path uses **default Jackson bean serialization** (the model's `@JsonProperty` getters), which writes the field regardless of version.

## Fix

`HttpClientOptions.isClearTextUpgrade()` (V2 definition model) now returns `false` when `version == HTTP_1_1`. Since Jackson bean serialization reads the getter, every serialization path is now consistent with the custom serializer.

**No runtime behaviour change:** the only consumer, `HttpEndpointRuleHandler` (health-check), reads `isClearTextUpgrade()` exclusively inside a `version == HTTP_2` guard.

## Tests

New `HttpClientOptionsTest`:
- default (HTTP_1_1) → `isClearTextUpgrade()` is `false`
- HTTP_1_1 + `clearTextUpgrade=true` → default bean serialization writes `false`
- HTTP_2 + `clearTextUpgrade=true` → preserved as `true`

`mvn test -Dtest=HttpClientOptionsTest` → `Tests run: 3, Failures: 0`.

### DIFF
--- a/gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/HttpClientOptions.java
+++ b/gravitee-apim-definition/gravitee-apim-definition-model/src/main/java/io/gravitee/definition/model/HttpClientOptions.java
@@ -162,6 +162,10 @@ public class HttpClientOptions implements Serializable {
     }
 
     public boolean isClearTextUpgrade() {
+        // h2c cleartext upgrade is an HTTP/2-only mechanism; it has no meaning for HTTP/1.1 (APIM-14613)
+        if (version == ProtocolVersion.HTTP_1_1) {
+            return false;
+        }
         return clearTextUpgrade;
     }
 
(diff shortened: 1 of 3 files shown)


callaertanthony:

Normalizing via the `isClearTextUpgrade()` getter is the right approach: it cascades to Jackson serialization (getter-based by default) so the invalid `clearTextUpgrade:true` + `HTTP_1_1` combo stops leaking into exported definitions, with no impact on HTTP_2. The `==` enum check is null-safe, and the tests cover default / serialization / HTTP_2 preservation.

ThibaudAV (approved):

Issue #10, opened 2015-06-27
Generating access logs for each HTTP call


brasseld:
Access Logs should contain:
- Timestamp
- Path
- HTTP method
- Name of the API
- Size of the response
- Time to handle request

brasseld:
Access logs should be done in an asynchronous way.

PR #18576, opened 2026-07-16
feat: propagate IdP claims into DCR requests — REST config, login persistence, injection, docs (APIM-13949)

wbabyte:
## What

Final PR of epic **APIM-13949** — *Tenant/User context propagation in DCR via IdP claims*. Builds on the repository layer merged in #18358 and #18383. This PR delivers the rest-api behavior end-to-end:

- **Config (REST)** — `persistedClaimsWhitelist` on the identity provider (**APIM-13950**) and `claimMappings` on the client registration provider (**APIM-13952**), both with GET/PUT and merge-on-omit so a save that doesn't include the field preserves the stored value.
- **Persist at login** (**APIM-13951**) — on social-IdP login, whitelisted claims are extracted from `id_token` / `access_token` / `userinfo` (first source wins) and persisted on the user; refreshed each login; no-op when no whitelist is configured.
- **Inject into DCR** (**APIM-13953**) — at client registration, the developer's persisted claims are injected into the DCR request body at the mapped (dot-notation) field paths, overriding existing values; missing claims are skipped.
- **Tests** (**APIM-13957**).

## How

- Whitelist surfaced onto `SocialIdentityProviderEntity` at login; claims read via Jackson, stored through the repository (no new API exposure — `idpClaims` is internal only).
- DCR injection done at the JSON-tree layer in `DynamicClientRegistrationProviderClient` just before POST, so both standard and custom/nested fields work without changing the typed `ClientRegistrationRequest`.
- `register()` threads the authenticated user's persisted claims through; skipped for actors with none (admin-on-behalf / service account).

## Tests

rest-api service tests green: `UserServiceTest`, `IdentityProviderServiceImplTest`, `ClientRegistrationService_RegisterTest` (incl. nested injection), `ClientRegistrationService_UpdateTest`, `ApplicationService_CreateTest`.

## Notes

- Stories: **APIM-13950, APIM-13952, APIM-13951, APIM-13953, APIM-13957** (epic **APIM-13949**). Combined into one PR as they form a single cohesive rest-api slice.
- Documentation (APIM-13958) is handled separately, not in this PR.
- No UI (product decision on APIM-13954 — configuration is via the Management REST API).
- Verified end-to-end on a local instance against a mock OIDC provider: login persisted `org_id`, and app registration injected `metadata.organization` into the DCR request.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/configuration/application/registration/UpdateClientRegistrationProviderEntity.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/configuration/application/registration/UpdateClientRegistrationProviderEntity.java
@@ -19,6 +19,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import jakarta.validation.Valid;
 import jakarta.validation.constraints.NotNull;
 import java.util.List;
+import java.util.Map;
 
 /**
  * @author David BRASSELY (david.brassely at graviteesource.com)
@@ -61,6 +62,9 @@ public class UpdateClientRegistrationProviderEntity {
 
     @JsonProperty("software_id")
     private String softwareId;
+
+    @JsonProperty("claim_mappings")
+    private Map<String, String> claimMappings;
 
     @JsonProperty("trust_store")
     @Valid
@@ -164,6 +168,14 @@ public class UpdateClientRegistrationProviderEntity {
 
     public void setSoftwareId(String softwareId) {
         this.softwareId = softwareId;
+    }
+
+    public Map<String, String> getClaimMappings() {
+        return claimMappings;
+    }
+
+    public void setClaimMappings(Map<String, String> claimMappings) {
+        this.claimMappings = claimMappings;
     }
 
     public TrustStoreEntity getTrustStore() {
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
@@ -906,17 +906,13 @@ public class ApplicationServiceImpl extends AbstractService implements Applicati
     ) {
         try {
             // Inject the application owner's IdP claims (not the editor's) so the tenant/user context registered with
-            // the IdP stays tied to who owns the application and does not flip when another user edits it.
-            String ownerId = membershipService.getPrimaryOwnerUserId(
-                executionContext.getOrganizationId(),
-                MembershipReferenceType.APPLICATION,
-                applicationToUpdate.getId()
-            );
+            // the IdP stays tied to who owns the application and does not flip when another user edits it. Resolved
+            // defensively: an unresolvable owner must degrade to "no claims injected", never break the DCR update.
             ClientRegistrationResponse registrationResponse = clientRegistrationService.update(
                 executionContext,
                 registrationPayload,
                 updateApplicationEntity,
-                ownerId != null ? userService.findIdpClaims(executionContext, ownerId) : null
+                resolveOwnerIdpClaims(executionContext, applicationToUpdate.getId())
             );
             metadata.put(METADATA_CLIENT_ID, registrationResponse.getClientId());
             metadata.put(METADATA_REGISTRATION_PAYLOAD, mapper.writeValueAsString(registrationResponse));
@@ -928,6 +924,25 @@ public class ApplicationServiceImpl extends AbstractService implements Applicati
             log.error("Failed to update OAuth client data from client registration. Keeping old OAuth client data.", e);
             metadata.put(METADATA_CLIENT_ID, applicationToUpdate.getMetadata().get(METADATA_CLIENT_ID));
             metadata.put(METADATA_REGISTRATION_PAYLOAD, applicationToUpdate.getMetadata().get(METADATA_REGISTRATION_PAYLOAD));
+        }
+    }
+
+    /**
+     * Resolves the persisted IdP claims of the application's primary owner. {@code getPrimaryOwnerUserId} can throw
+     * (no primary owner, group-owned application, missing role), so this degrades to {@code null} — a missing owner
+     * must mean "no claims injected", never a failure that aborts the DCR update.
+     */
+    private Map<String, String> resolveOwnerIdpClaims(ExecutionContext executionContext, String applicationId) {
+        try {
+            String ownerId = membershipService.getPrimaryOwnerUserId(
+                executionContext.getOrganizationId(),
+                MembershipReferenceType.APPLICATION,
+                applicationId
+            );
+            return ownerId != null ? userService.findIdpClaims(executionContext, ownerId) : null;
+        } catch (Exception e) {
+            log.warn("Could not resolve primary owner IdP claims for application {}; proceeding without injection", applicationId, e);
+            return null;
         }
     }
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
@@ -908,7 +908,8 @@ public class ApplicationServiceImpl extends AbstractService implements Applicati
             ClientRegistrationResponse registrationResponse = clientRegistrationService.update(
                 executionContext,
                 registrationPayload,
-                updateApplicationEntity
+                updateApplicationEntity,
+                userService.findIdpClaims(executionContext, getAuthenticatedUsername())
             );
             metadata.put(METADATA_CLIENT_ID, registrationResponse.getClientId());
             metadata.put(METADATA_REGISTRATION_PAYLOAD, mapper.writeValueAsString(registrationResponse));
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
@@ -905,11 +905,18 @@ public class ApplicationServiceImpl extends AbstractService implements Applicati
         Application applicationToUpdate
     ) {
         try {
+            // Inject the application owner's IdP claims (not the editor's) so the tenant/user context registered with
+            // the IdP stays tied to who owns the application and does not flip when another user edits it.
+            String ownerId = membershipService.getPrimaryOwnerUserId(
+                executionContext.getOrganizationId(),
+                MembershipReferenceType.APPLICATION,
+                applicationToUpdate.getId()
+            );
             ClientRegistrationResponse registrationResponse = clientRegistrationService.update(
                 executionContext,
                 registrationPayload,
                 updateApplicationEntity,
-                userService.findIdpClaims(executionContext, getAuthenticatedUsername())
+                ownerId != null ? userService.findIdpClaims(executionContext, ownerId) : null
             );
             metadata.put(METADATA_CLIENT_ID, registrationResponse.getClientId());
             metadata.put(METADATA_REGISTRATION_PAYLOAD, mapper.writeValueAsString(registrationResponse));
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
@@ -441,7 +441,11 @@ public class ApplicationServiceImpl extends AbstractService implements Applicati
             checkAndSanitizeOAuthClientSettings(newApplicationEntity.getSettings().getOauth());
 
             // Create an OAuth client
-            ClientRegistrationResponse registrationResponse = clientRegistrationService.register(executionContext, newApplicationEntity);
+            ClientRegistrationResponse registrationResponse = clientRegistrationService.register(
+                executionContext,
+                newApplicationEntity,
+                userService.findIdpClaims(executionContext, userId)
+            );
             try {
                 metadata.put(METADATA_CLIENT_ID, registrationResponse.getClientId());
                 metadata.put(METADATA_REGISTRATION_PAYLOAD, mapper.writeValueAsString(registrationResponse));
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
@@ -41,6 +41,8 @@ import com.auth0.jwt.JWT;
 import com.auth0.jwt.JWTVerifier;
 import com.auth0.jwt.algorithms.Algorithm;
 import com.auth0.jwt.interfaces.DecodedJWT;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.jayway.jsonpath.JsonPath;
 import com.jayway.jsonpath.ReadContext;
 import io.gravitee.apim.core.installation.query_service.InstallationAccessQueryService;
@@ -196,6 +198,7 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
     private static final String TEMPLATE_ENGINE_PROFILE_ATTRIBUTE = "profile";
     private static final String TEMPLATE_ENGINE_ACCESSTOKEN_ATTRIBUTE = "accessToken";
     private static final String TEMPLATE_ENGINE_IDTOKEN_ATTRIBUTE = "idToken";
+    private static final ObjectMapper CLAIMS_MAPPER = new ObjectMapper();
 
     // Dirty hack: only used to force class loading
     static {
@@ -1745,7 +1748,76 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
             );
         }
 
+        persistWhitelistedClaims(
+            socialProvider.getPersistedClaimsWhitelist(),
+            user.getId(),
+            userInfo,
+            accessTokenPayloadAsString,
+            idTokenPayloadAsString
+        );
+
         return user;
+    }
+
+    /**
+     * Persists the IdP claims whitelisted on the social identity provider onto the user record, refreshing them on
+     * every login. When no whitelist is configured the feature is inactive and the user record is left untouched.
+     */
+    private void persistWhitelistedClaims(
+        List<String> whitelist,
+        String userId,
+        String userInfo,
+        String accessTokenPayload,
+        String idTokenPayload
+    ) {
+        if (whitelist == null || whitelist.isEmpty()) {
+            return;
+        }
+        Map<String, String> idpClaims = extractWhitelistedClaims(whitelist, userInfo, accessTokenPayload, idTokenPayload);
+        try {
+            Optional<User> optionalUser = userRepository.findById(userId);
+            if (optionalUser.isPresent()) {
+                User user = optionalUser.get();
+                user.setIdpClaims(idpClaims);
+                userRepository.update(user);
+            }
+        } catch (TechnicalException e) {
+            log.warn("Unable to persist IdP claims for user {}", userId, e);
+        }
+    }
+
+    /**
+     * Extracts the whitelisted claims from the IdP sources, looking them up in order id_token, access_token then
+     * userinfo (first source containing the claim wins). Missing claims are skipped; complex values are JSON-serialized.
+     */
+    private Map<String, String> extractWhitelistedClaims(
+        List<String> whitelist,
+        String userInfo,
+        String accessTokenPayload,
+        String idTokenPayload
+    ) {
+        List<JsonNode> sources = new ArrayList<>();
+        for (String source : new String[] { idTokenPayload, accessTokenPayload, userInfo }) {
+            if (source != null && !source.isEmpty()) {
+                try {
+                    sources.add(CLAIMS_MAPPER.readTree(source));
+                } catch (Exception e) {
+                    log.debug("Unable to parse IdP claim source as JSON", e);
+                }
+            }
+        }
+
+        Map<String, String> claims = new HashMap<>();
+        for (String claimName : whitelist) {
+            for (JsonNode source : sources) {
+                JsonNode value = source.get(claimName);
+                if (value != null && !value.isNull()) {
+                    claims.put(claimName, value.isValueNode() ? value.asText() : value.toString());
+                    break;
+                }
+            }
+        }
+        return claims;
     }
 
     private HashMap<String, String> getUserProfileAttrs(Map<String, String> userProfileMapping, String userInfo) {
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
@@ -1759,6 +1759,15 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
         return user;
     }
 
+    @Override
+    public Map<String, String> findIdpClaims(ExecutionContext executionContext, String userId) {
+        try {
+            return userRepository.findById(userId).map(User::getIdpClaims).orElse(null);
+        } catch (TechnicalException ex) {
+            throw new TechnicalManagementException("An error occurs while trying to find IdP claims for user " + userId, ex);
+        }
+    }
+
     /**
      * Persists the IdP claims whitelisted on the social identity provider onto the user record, refreshing them on
      * every login. When no whitelist is configured the feature is inactive and the user record is left untouched.
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
@@ -1819,6 +1819,8 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
         Map<String, String> claims = new HashMap<>();
         for (String claimName : whitelist) {
             for (JsonNode source : sources) {
+                // Deliberate flat lookup (not a dot-notation path like the DCR field side): OIDC namespaced claim
+                // names such as "https://example.com/org_id" contain dots that must be treated as a single literal key.
                 JsonNode value = source.get(claimName);
                 if (value != null && !value.isNull()) {
                     claims.put(claimName, value.isValueNode() ? value.asText() : value.toString());
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/ClientRegistrationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/ClientRegistrationServiceImpl.java
@@ -209,6 +209,11 @@ public class ClientRegistrationServiceImpl extends AbstractService implements Cl
             clientRegistrationProvider.setId(id);
             clientRegistrationProvider.setCreatedAt(clientProviderToUpdate.getCreatedAt());
             clientRegistrationProvider.setUpdatedAt(new Date());
+            // Preserve the claim mappings when the update payload omits them (e.g. saved from a console screen
+            // unaware of the field); clear them explicitly by sending an empty map.
+            if (updateClientRegistrationProvider.getClaimMappings() == null) {
+                clientRegistrationProvider.setClaimMappings(clientProviderToUpdate.getClaimMappings());
+            }
 
             ClientRegistrationProvider updatedClientRegistrationProvider = clientRegistrationProviderRepository.update(
                 clientRegistrationProvider
@@ -507,6 +512,7 @@ public class ClientRegistrationServiceImpl extends AbstractService implements Cl
         entity.setRenewClientSecretMethod(clientRegistrationProvider.getRenewClientSecretMethod());
         entity.setRenewClientSecretEndpoint(clientRegistrationProvider.getRenewClientSecretEndpoint());
         entity.setSoftwareId(clientRegistrationProvider.getSoftwareId());
+        entity.setClaimMappings(clientRegistrationProvider.getClaimMappings());
 
         if (
             clientRegistrationProvider.getInitialAccessTokenType() == null ||
@@ -567,6 +573,7 @@ public class ClientRegistrationServiceImpl extends AbstractService implements Cl
         provider.setRenewClientSecretMethod(newClientRegistrationProvider.getRenewClientSecretMethod());
         provider.setRenewClientSecretEndpoint(newClientRegistrationProvider.getRenewClientSecretEndpoint());
         provider.setSoftwareId(newClientRegistrationProvider.getSoftwareId());
+        provider.setClaimMappings(newClientRegistrationProvider.getClaimMappings());
 
         if (newClientRegistrationProvider.getInitialAccessTokenType() == InitialAccessTokenType.CLIENT_CREDENTIALS) {
             provider.setInitialAccessTokenType(ClientRegistrationProvider.InitialAccessTokenType.CLIENT_CREDENTIALS);
@@ -613,6 +620,7 @@ public class ClientRegistrationServiceImpl extends AbstractService implements Cl
         provider.setRenewClientSecretMethod(updateClientRegistrationProvider.getRenewClientSecretMethod());
         provider.setRenewClientSecretEndpoint(updateClientRegistrationProvider.getRenewClientSecretEndpoint());
         provider.setSoftwareId(updateClientRegistrationProvider.getSoftwareId());
+        provider.setClaimMappings(updateClientRegistrationProvider.getClaimMappings());
 
         if (updateClientRegistrationProvider.getInitialAccessTokenType() == InitialAccessTokenType.CLIENT_CREDENTIALS) {
             provider.setInitialAccessTokenType(ClientRegistrationProvider.InitialAccessTokenType.CLIENT_CREDENTIALS);
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/ClientRegistrationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/ClientRegistrationServiceImpl.java
@@ -73,6 +73,20 @@ import org.springframework.stereotype.Component;
 @Component
 public class ClientRegistrationServiceImpl extends AbstractService implements ClientRegistrationService {
 
+    /**
+     * Security-relevant DCR fields that must never be the target of a claim mapping: their value is IdP/end-user
+     * controlled and claim injection bypasses the OAuth client sanitization applied to operator-supplied settings.
+     * A mapping whose (dot-notation) field path starts with one of these top-level fields is rejected.
+     */
+    private static final Set<String> PROTECTED_DCR_FIELDS = Set.of(
+        "redirect_uris",
+        "grant_types",
+        "response_types",
+        "token_endpoint_auth_method",
+        "jwks_uri",
+        "client_uri"
+    );
+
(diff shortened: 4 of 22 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/client/DynamicClientRegistrationProviderClient.java
```diff
@@ -116,6 +117,32 @@ protected ClientRegistrationResponse register(
         }
     }
 
+    /**
+     * Injects the resolved claim values into the DCR request body at their mapped (dot-notation) field paths.
+     * Intermediate objects are created as needed and existing values at the target path are overridden.
+     */
+    private void injectClaims(ObjectNode body, Map<String, String> claimInjections) {
```
paulatulis, line 124: Claim injection runs **after** OAuth client sanitization, so it can override fields that were just sanitized.

In `ApplicationServiceImpl.create`, the order is `checkAndSanitizeOAuthClientSettings(...)` and *then* `clientRegistrationService.register(...)` — and injection happens inside `register`, at the JSON-tree layer, deliberately overriding existing values (per the APIM-13953 AC).

The field path is operator-configured, but the **value is IdP-controlled**. On an IdP where users can self-edit their own profile attributes, an end user effectively controls the content of whatever field the mapping points at. A mapping onto a security-relevant field (`redirect_uris`, `grant_types`, `client_uri`) therefore reaches the registration endpoint without passing the sanitization that the same request's operator-supplied settings had to pass.

That is the difference between "operator misconfigured it" and "an end user can trigger it", which seems worth an explicit decision rather than falling out of call ordering.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/client/DynamicClientRegistrationProviderClient.java
```diff
@@ -116,6 +117,32 @@ protected ClientRegistrationResponse register(
...
+                    current.set(segments[i], child);
+                }
+                current = (ObjectNode) child;
+            }
+            current.put(segments[segments.length - 1], value);
+            log.debug("Injected IdP claim into DCR field [{}]", fieldPath);
+        }
+    }
+
     public ClientRegistrationResponse update(
```
paulatulis, line 154: The update path never injects claims, so the feature appears to undo itself on the first application update.

`update(...)` takes no `claimInjections` and issues an `HttpPut` against the registration client URI — RFC 7592 replace semantics, so the body sent here is the *complete* new registration. Any field injected at create time (e.g. `metadata.organization`) is absent from that body and is therefore dropped by the provider.

This is quiet in two ways: nothing signals the loss, and `ApplicationServiceImpl.updateClientRegistration` catches `Exception` and logs "Keeping old OAuth client data", so even a provider-side rejection would not surface.

Net effect: claims are correctly injected at creation and silently disappear the first time anyone edits the application. Worth confirming whether update was intentionally out of scope for APIM-13953 — if so, the limitation is probably worth recording, since the observable behaviour is that the feature stops working.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/configuration/application/registration/UpdateClientRegistrationProviderEntity.java
```diff
@@ -62,6 +63,9 @@ public class UpdateClientRegistrationProviderEntity {
     @JsonProperty("software_id")
     private String softwareId;
 
+    @JsonProperty("claim_mappings")
+    private Map<String, String> claimMappings;
```
paulatulis, line 67: This field accepts values the injection code cannot handle, and there is no validation on any layer to stop them.

APIM-13952 lists "Validation ensures DCR field paths are non-empty strings" as an acceptance criterion, but no constraint is declared here, and `ClientRegistrationServiceImpl.convert(...)` copies the map straight to the repository model. Note the contrast within this same class: `trustStore` just below carries `@Valid`, and the sibling fields carry `@NotNull` / `@Size`.

Verified consequences once such a value reaches `DynamicClientRegistrationProviderClient.injectClaims` (JDK 21, measured — `String.split` strips trailing empty strings, which makes these non-obvious):

| configured path | `split("\\.")` | result |
|---|---|---|
| `"."` / `".."` | length **0** | `segments[-1]` → `ArrayIndexOutOfBoundsException` |
| `null` | — | NPE on `fieldPath.split(...)` |
| `""` | length 1, `[""]` | emits a JSON field with an empty name |
| `"a."` | length 1, `["a"]` | trailing segment dropped — writes to `a`, not the intended nested field |

The first two surface as `DynamicClientRegistrationException: Unexpected error while registering client: ...` and fail application creation with a message that names neither the field nor the cause. The last one is the quietest: no error, wrong field.

The same gap applies to `NewClientRegistrationProviderEntity.claimMappings`, and to `persistedClaimsWhitelist` on `New`/`UpdateIdentityProviderEntity` (APIM-13950's "Validation ensures only valid claim names are added").

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -1745,9 +1748,87 @@ public UserEntity createOrUpdateUserFromSocialIdentityProvider(
...
+                } catch (Exception e) {
+                    log.debug("Unable to parse IdP claim source as JSON", e);
+                }
+            }
+        }
+
+        Map<String, String> claims = new HashMap<>();
+        for (String claimName : whitelist) {
+            for (JsonNode source : sources) {
+                JsonNode value = source.get(claimName);
```
paulatulis, line 1824: Dot notation means two different things in this feature, and the mismatch fails silently.

`DynamicClientRegistrationProviderClient.injectClaims` splits the DCR field path on `\.` to build nested objects, so `.` is a path separator there. Here, claim lookup is a flat `source.get(claimName)`, so a whitelisted claim named `address.country` is looked up as a single literal key and simply never resolves.

Because `resolveClaimInjections` skips any claim missing from the persisted map, the two cases are indistinguishable downstream: "the IdP did not send this claim" and "this claim name can never be resolved" both produce the same silent skip, with no log to tell them apart.

Flat lookup may well be the intended contract for claim names — the ACs only specify dot notation on the DCR-field side. If so, the asymmetry is worth making explicit, since the same character behaves oppositely at the two ends of one feature.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/client/DynamicClientRegistrationProviderClient.java
```diff
@@ -116,6 +117,32 @@ protected ClientRegistrationResponse register(
...
+            ObjectNode current = body;
+            for (int i = 0; i < segments.length - 1; i++) {
+                JsonNode child = current.get(segments[i]);
+                if (child == null || !child.isObject()) {
+                    child = mapper.createObjectNode();
+                    current.set(segments[i], child);
+                }
+                current = (ObjectNode) child;
+            }
+            current.put(segments[segments.length - 1], value);
```
paulatulis, line 149: Complex claim values reach the IdP as escaped JSON strings rather than as arrays/objects.

`extractWhitelistedClaims` stores a non-scalar claim via `value.toString()` (UserServiceImpl:1824), which is the right call for a `Map<String, String>` store and matches the APIM-13951 AC ("complex claim values are JSON-serialized"). But this line then re-inserts that stored form with `ObjectNode.put(String, String)`, so it lands in the DCR body as a **string**.

A `groups` claim of `["a","b"]` is therefore transmitted as:

```json
{ "groups": "[\"a\",\"b\"]" }
```

rather than `{ "groups": ["a","b"] }`. The APIM-13953 AC "Ensure the DCR request is valid JSON after claim injection" is technically satisfied — it is valid JSON — but the receiving provider gets a string where it expects an array, and nothing fails loudly.

Scalar claims are unaffected, which is why the existing tests do not catch it: every fixture claim is a scalar.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -1745,9 +1748,87 @@ public UserEntity createOrUpdateUserFromSocialIdentityProvider(
...
+        } catch (TechnicalException ex) {
+            throw new TechnicalManagementException("An error occurs while trying to find IdP claims for user " + userId, ex);
+        }
+    }
+
+    /**
+     * Persists the IdP claims whitelisted on the social identity provider onto the user record, refreshing them on
+     * every login. When no whitelist is configured the feature is inactive and the user record is left untouched.
+     */
+    private void persistWhitelistedClaims(
```
paulatulis, line 1775: Removing a claim from the whitelist does not remove the already-persisted value.

`persistWhitelistedClaims` returns early when the whitelist is null/empty, and otherwise overwrites `idpClaims` with only the currently-whitelisted claims. Both branches are reasonable in isolation, but together they mean the whitelist is only additive in effect:

- Whitelist `[org_id]` → user has `{org_id: ...}` persisted.
- Operator removes `org_id`, leaving the whitelist empty → the early return fires → the stored `{org_id: ...}` **stays on the user record**.
- Any later application registration still injects it into the outbound DCR request.

APIM-13950's "Changes to the whitelist apply to new logins; existing user claims are not retroactively updated" arguably describes this, but the security-relevant reading is different from the config-hygiene one: de-whitelisting a claim looks like revocation and does not act like it. Since the whole point of the whitelist is "sensitive or unnecessary claims are not stored", it may be worth deciding explicitly whether emptying it should purge.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -1745,9 +1748,87 @@ public UserEntity createOrUpdateUserFromSocialIdentityProvider(
...
+        Map<String, String> idpClaims = extractWhitelistedClaims(whitelist, userInfo, accessTokenPayload, idTokenPayload);
+        try {
+            Optional<User> optionalUser = userRepository.findById(userId);
+            if (optionalUser.isPresent()) {
+                User user = optionalUser.get();
+                user.setIdpClaims(idpClaims);
+                userRepository.update(user);
+            }
+        } catch (TechnicalException e) {
+            log.warn("Unable to persist IdP claims for user {}", userId, e);
```
paulatulis, line 1794: This swallow leaves stale claims that the DCR path later injects as if they were fresh.

Not persisting on failure is the right instinct — a claims problem should not break login. But the failure is asymmetric with the read side: `findIdpClaims` (:1763) wraps its `TechnicalException` in a `TechnicalManagementException` and fails the caller hard, while this path logs a warning and continues.

The consequence is that after a failed update the user keeps the *previous* login's claims, and the next application registration injects those into the DCR request with no indication they are stale. A silent write failure becomes a silently wrong value at a third party, which is a worse outcome than the loud failure the read path produces.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/client/DynamicClientRegistrationProviderClient.java
```diff
@@ -61,6 +62,7 @@ public abstract class DynamicClientRegistrationProviderClient {
     protected ClientRegistrationResponse register(
         String initialAccessToken,
         ClientRegistrationRequest request,
+        Map<String, String> claimInjections,
```
paulatulis, line 65: Layering: this parameter brings request-body composition into the HTTP transport client.

`ClientRegistrationServiceImpl.resolveClaimInjections` resolves claim name → field path → value, and then this class re-interprets that field path against the JSON tree — dot-notation splitting, intermediate node creation, override semantics. The mapping rules are consequently split across two layers, and every future caller of `register(...)` inherits the injection behaviour implicitly.

The practical cost shows up in the tests: `should_inject_mapped_claims_into_dcr_request_body` has to assert a *mapping rule* through a WireMock HTTP stub, because there is no seam at which the enriched body exists as a value. Composing the body in the service and handing the finished node to the client would make the path rules testable directly, and keep this class responsible for serialize-and-POST.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
```diff
@@ -1760,6 +1760,104 @@ public void shouldRefreshExistingUser() throws IOException, TechnicalException {
...
+        // claim only present in the id_token
+        assertEquals("foobar", user.getIdpClaims().get("custom_id_token"));
+        // claim only present in the userinfo
+        assertEquals("585252525", user.getIdpClaims().get("service_id"));
+        // claim present in both id_token and userinfo: the id_token value wins
+        assertEquals("john.doe@mycompany.com", user.getIdpClaims().get("email"));
+    }
+
+    @Test
+    public void shouldNotPersistClaimsWhenNoWhitelistConfigured() throws IOException, TechnicalException {
```
paulatulis, line 1790: This assertion cannot fail.

`mockUser()` never sets `idpClaims`, so `user.getIdpClaims()` is already `null` when the test starts. `assertNull(user.getIdpClaims())` at the end therefore passes whether or not `persistWhitelistedClaims` correctly no-ops — deleting the early-return guard in the production code would leave this test green.

To actually pin the no-op, the user needs a non-null starting state that the code under test must leave untouched (e.g. pre-set `idpClaims`, then assert it is unchanged), or `verify(userRepository, never()).update(any())`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ApplicationService_CreateTest.java
```diff
@@ -574,7 +574,7 @@ public void shouldCreateOauthApp() throws TechnicalException {
         // mock response from DCR with a new client ID
         ClientRegistrationResponse clientRegistrationResponse = new ClientRegistrationResponse();
         clientRegistrationResponse.setClientId("client-id-from-clientRegistration");
-        when(clientRegistrationService.register(any(), any())).thenReturn(clientRegistrationResponse);
+        when(clientRegistrationService.register(any(), any(), any())).thenReturn(clientRegistrationResponse);
```
paulatulis, line 581: The new third argument is wildcard-matched, so the wiring this PR adds is unverified.

`register(any(), any(), any())` will match regardless of what `ApplicationServiceImpl` passes as `idpClaims` — including `null`, and including the result of a `findIdpClaims` call that never happened. The `userService.findIdpClaims(...)` → `register(...)` hand-off introduced in this PR is the one new behaviour on this path, and no assertion covers it.

Stubbing `findIdpClaims` to a known map and capturing the third argument (or matching it with `eq(...)`) would make this test fail if the wiring regressed.

paulatulis:
## Review summary — remaining findings

Ten findings are posted inline above. This comment carries the rest so nothing is dropped: test-coverage gaps and minors, which didn't seem worth an inline comment each.

Reviewed at `6547127`, against merge-base `d0a72d6`. Multi-agent review (group reviewers per layer, cross-cutting pass, security/coverage specialists, architect + tester concerns); findings deduped and each one verified against the source before posting.

### Test-coverage gaps

The claim feature is reasonably tested — the precedence test in `shouldPersistWhitelistedClaimsOnLogin` is genuine (the fixtures have different emails in `id_token` vs `userinfo`, so it fails if precedence flips). These are the paths with no coverage:

- [ ] **`access_token` as claim source is never exercised** — the middle of the precedence chain. Every fixture claim resolves from `id_token` or `userinfo`, so removing `accessTokenPayload` from the source list would keep the suite green.
- [ ] **Non-scalar claim value** (`value.toString()` branch, `UserServiceImpl:1824`) — no fixture claim is an array or object, so the JSON-serialization AC (APIM-13951) is unverified. See the inline comment on `DynamicClientRegistrationProviderClient.java:141` — this is where the escaped-string defect hides.
- [ ] **Unparseable / opaque source fall-through** — `extractWhitelistedClaims` catches a parse failure and skips that source; untested.
- [ ] **`findIdpClaims` has no test at all** — neither the `.orElse(null)` path nor the `TechnicalException` → `TechnicalManagementException` wrap.
- [ ] **`persistWhitelistedClaims` `TechnicalException` path** — that login still succeeds when the update fails is the intended behaviour, and nothing pins it.
- [ ] **`resolveClaimInjections` with `idpClaims == null` but mappings configured** — the shape most installations will actually hit (a user who has never logged in via the social IdP).
- [ ] **Injected value overriding a pre-existing DCR field** — the javadoc says existing values are overridden, so a mapping like `org_id` → `client_name` is a *supported* configuration that silently replaces the application name. Only the nested `metadata.organization` path is covered.
- [ ] **Intermediate segment holding a non-object scalar gets replaced** — e.g. mapping to `client_name.foo` when `client_name` is a string.
- [ ] **Flat (single-segment) DCR field path** — only the nested path is tested.
- [ ] **Empty `claimMappings` injects nothing** — never asserted.
- [ ] **Whitelisted claim absent from every source is not stored** (no null entry) — the APIM-13951 AC; untested.
- [ ] **`should_inject_mapped_claims_into_dcr_request_body`** sets up `tenant` as a mapped-but-missing claim and the comment says no `metadata.tenant` is sent — but only `$.metadata.organization` is asserted. The skip (an APIM-13953 AC) is asserted by comment only; the test passes either way.

### Minors

- [ ] `findIdpClaims` returns `null` rather than an empty map, pushing null-handling to callers.
- [ ] `findIdpClaims(ExecutionContext, String)` ignores `executionContext` entirely — `userRepository.findById(userId)` is unscoped, so the parameter implies an organization guarantee the implementation doesn't provide. Worth dropping or honouring.
- [ ] `findIdpClaims` is called unconditionally on every DCR-backed application creation, before any check that a provider declares `claimMappings`. It throws `TechnicalManagementException` on repository error, so a transient user-repo failure now fails app creation on installations that don't use this feature.
- [ ] `UserService` (legacy, widely injected) gains a method returning raw persisted IdP claims, guarded by a javadoc note. Nothing structurally prevents a future caller echoing them into a response DTO.
- [ ] No size bound on serialized claim values written to `idp_claims` and into the outbound DCR body.
- [ ] No timestamp/TTL on persisted claims — a claim revoked at the IdP keeps being asserted until the user's next login.
- [ ] `IdentityProviderServiceImplTest`: blanket `@MockitoSettings(strictness = LENIENT)` that none of its three tests need — it suppresses exactly the strict-stubs signal these merge-semantics tests rely on.
- [ ] Test-naming: `update_should_…` / `should_inject_…` (snake_case) introduced into camelCase classes.
- [ ] New feature logic lands in the legacy `service.impl` layer rather than a `core/` use case. Noted, not blocking — DCR has no `core/` domain today.

### Verified as non-issues

Recording these so they don't get re-litigated:

- `valueToTree` vs `writeValueAsString` — `ClientRegistrationRequest` carries class-level `@JsonInclude(NON_NULL)`, which `valueToTree` honours. No serialization regression.
- `idpClaims` is **not** exposed via the user profile API — `UserEntity` has no such field and nothing outside `UserServiceImpl` reads `User::getIdpClaims`. The APIM-13951 non-exposure AC holds.
- `SocialIdentityProviderEntity.persistedClaimsWhitelist` is `@JsonIgnore` on both field and getter, so it stays off the anonymous login-page IdP listing — consistent with how `clientSecret` is handled in that class.
- The DEBUG log deliberately omits the claim **value**, deviating from APIM-13953's "log the claim name, field path, value". That's the right call — claim values are tenant/user context and don't belong in logs.
- `reset()` usage in `UserServiceTest` matches ~25 existing sites in that file — local convention, not a finding.
- snake_case `claim_mappings` (`@JsonProperty`) vs camelCase `persistedClaimsWhitelist`: each matches its own file's existing convention.

### Coverage gap in this review

One specialist that normally always runs — a swallowed/relabelled error-path hunter — could not run here. That's directly relevant to this diff, which has three such paths (`persistWhitelistedClaims` catch→warn, `catch (Exception)` in the DCR client, `updateClientRegistration` catch→error). Two of the three are raised above on other grounds, but treat error-path coverage as reviewed less thoroughly than the rest.

wbabyte:
Thanks @paulatulis (and the multi-agent pass) — genuinely high-signal review. Pushed `c42542f` addressing it. Grouped below.

### Fixed (code)
- **Update path never injected → feature undid itself on first app edit (#3).** `ClientRegistrationService.update(...)` now takes the user's `idpClaims`, resolves the mappings, and `DynamicClientRegistrationProviderClient.update` re-injects them into the RFC 7592 PUT body. New test `should_inject_mapped_claims_on_application_update` asserts `metadata.organization` is present on the PUT.
- **Security-relevant field targeting (#2).** Decision: **add a blocklist.** Mappings whose (top-level) field path is `redirect_uris`, `grant_types`, `response_types`, `token_endpoint_auth_method`, `jwks_uri`, or `client_uri` are now rejected at provider create/update (`InvalidClaimMappingException`, 400) and defensively skipped at injection. Your end-user-controlled-value framing was the deciding factor — thanks for sharpening it.
- **No field-path validation → crashes/wrong-field (#4).** APIM-13952 non-empty AC now enforced at config save; `injectClaims` also hardened against `.`/`..`/blank paths (skip, no AIOOBE/NPE). Added `persistedClaimsWhitelist` non-blank validation too (APIM-13950). Rejection tests added.
- **Weak test `assertNull` can't fail (#10).** Now pre-sets a sentinel `idpClaims` and asserts it's left untouched — fails if the no-op guard is removed.
- **`register(any(),any(),any())` wildcard didn't verify wiring (#11).** Now stubs `findIdpClaims` to a known map and `verify(...).register(any(), any(), eq(idpClaims))`.
- **Coverage gaps:** added access_token-source, missing-claim-skip, empty-mappings-no-injection, override-existing-standard-field; the skip assertion is now `verify(0, … metadata.tenant)` instead of comment-only.

### Responded — no change (deliberate scope decisions)
- **De-whitelisting doesn't purge persisted claims (#7).** Kept per APIM-13950 ("existing user claims not retroactively updated"); claims refresh on next login. Noted as an explicit limitation.
- **Complex values injected as escaped JSON strings (#6).** This is the storage model chosen for the epic (`Map<String,String>`, APIM-13951 "JSON-serialized"); injecting verbatim keeps store and wire consistent. Documented as a limitation; native array/object passthrough would be a follow-up.
- **Flat claim-name lookup vs dot-split field path (#5).** Intentional and required — a namespaced claim like `https://acme/org` contains dots that must *not* be split on the claim side. The asymmetry is by design.
- **Layering — injection in the HTTP client (#9).** Fair; kept in the client for now since both create and update share it. Noted for a future refactor to compose the body in the service.
- **Aikido "giant file".** `UserServiceImpl` is already large; the claim logic is contained in two private helpers. Extraction deferred (no `core/` domain for this yet).

### Minors
Several folded in (validation, test naming kept consistent per-file). The rest (findIdpClaims empty-vs-null, executionContext scoping, size/TTL bounds) noted; happy to take any as follow-ups if you'd like them in-scope.

Full affected suite green on both backends. Re-request when you've had a look.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/ClientRegistrationServiceImpl.java
```diff
@@ -72,6 +73,20 @@
 @Component
 public class ClientRegistrationServiceImpl extends AbstractService implements ClientRegistrationService {
 
+    /**
+     * Security-relevant DCR fields that must never be the target of a claim mapping: their value is IdP/end-user
+     * controlled and claim injection bypasses the OAuth client sanitization applied to operator-supplied settings.
+     * A mapping whose (dot-notation) field path starts with one of these top-level fields is rejected.
+     */
+    private static final Set<String> PROTECTED_DCR_FIELDS = Set.of(
```
paulatulis: The blocklist covers 6 of the 33 fields on `ClientRegistrationRequest`, and the ones it misses include the sharpest.

First — credit where due: `isProtectedFieldPath` extracts `substring(0, indexOf('.'))`, which is exactly the string `injectClaims` uses as `segments[0]`. So `redirect_uris`, `redirect_uris.foo` and `redirect_uris.` all block identically. That consistency is the part I expected to be wrong, and it isn't.

The problem is enumeration. These are declared on `ClientRegistrationRequest` and unprotected:

| field | consequence of an IdP-controlled value |
|---|---|
| `scope` | privilege escalation — becomes the client's granted scope |
| `software_statement` | RFC 7591 §2.3 signed JWT asserting client metadata; if the OP trusts it, it supersedes the rest |
| `request_uris` | fetched by the OP — SSRF / request-object injection |
| `sector_identifier_uri` | fetched by the OP; controls pairwise subject identifiers |
| `initiate_login_uri` | third-party-initiated login redirect target |
| `logo_uri` / `policy_uri` / `tos_uri` | rendered to end users — phishing surface |

The reasoning that put `jwks_uri` on the list applies at least as strongly to `software_statement` and `scope`.

This is the usual blocklist tradeoff: the list has to stay correct as `ClientRegistrationRequest` grows, and a field added later is unprotected by default. An allowlist of injectable prefixes (in practice `metadata.*` and other vendor-extension namespaces, which is what the epic's motivating use case actually needs) inverts that — a new field is safe by default, and the failure mode for anything unanticipated becomes "mapping rejected" rather than "mapping wins".

If a blocklist is the deliberate call, the six above are worth adding, and it's worth a comment on `PROTECTED_DCR_FIELDS` saying it must be revisited whenever a field is added to the request.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/client/DynamicClientRegistrationProviderClient.java
```diff
@@ -128,17 +164,20 @@ public ClientRegistrationResponse update(
...
             if (request.getScope() != null && !request.getScope().isEmpty()) {
-                ((ObjectNode) reqNode).put("scope", String.join(ClientRegistrationRequest.SCOPE_DELIMITER, request.getScope()));
+                reqNode.put("scope", String.join(ClientRegistrationRequest.SCOPE_DELIMITER, request.getScope()));
             } else {
-                ((ObjectNode) reqNode).remove("scope");
+                reqNode.remove("scope");
             }
 
+            // RFC 7592 update is a full replace, so re-inject the claims to keep them across application updates.
+            injectClaims(reqNode, claimInjections);
```
paulatulis: Injection runs after `client_id` and `scope` are set here, so a mapping onto either overrides them — and neither is on the blocklist.

```java
reqNode.put("client_id", clientId);        // RFC 7592 requires this to match
...
reqNode.put("scope", ...) / remove("scope");
injectClaims(reqNode, claimInjections);    // ← runs last, wins
```

`scope` is a declared DCR field, so a mapping to it is accepted at config time and silently replaces the computed scope on every application update.

`client_id` is the more awkward one: it isn't a `@JsonProperty` on `ClientRegistrationRequest` — it only exists on the tree because this method puts it there — so it wouldn't show up if you enumerate the blocklist from the request class. A mapping onto `client_id` overrides the RFC 7592-mandated identifier in an authenticated PUT to the registration client URI.

Moving `injectClaims(...)` above the `client_id`/`scope` block would make those two structurally un-overridable, which seems better than relying on the blocklist to remember them. (The create path has the same exposure for `scope`, since `scope` is serialized into the body before injection.)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
```diff
@@ -904,7 +908,8 @@ private void updateClientRegistration(
             ClientRegistrationResponse registrationResponse = clientRegistrationService.update(
                 executionContext,
                 registrationPayload,
-                updateApplicationEntity
+                updateApplicationEntity,
+                userService.findIdpClaims(executionContext, getAuthenticatedUsername())
```
paulatulis: This injects the *updater's* claims, not the application owner's.

`create` resolves claims for the actor creating the application; this path resolves them for whoever is saving the edit. (Both resolve to the same principal id — I checked `getAuthenticatedUsername()` against the create path's `userId`, so there's no lookup mismatch.)

The consequence is that if developer B edits developer A's application, the RFC 7592 PUT re-registers it with **B's** tenant context — `metadata.organization` silently flips from A's org to B's. Any user with update permission on the application can rewrite the tenant context the IdP sees for it.

For a feature whose whole purpose is propagating tenant/user context, which identity the registration should carry seems worth deciding explicitly rather than having it follow whoever clicked save. If the app owner's claims are the intended semantic, this wants the owner's id rather than the authenticated actor's.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
paulatulis: Withdrawing this one — you're right and my finding was wrong.

Namespaced claims (`https://example.com/org_id`, the Auth0/OIDC convention) contain dots that must not be split, so flat lookup on the claim side isn't an asymmetry to fix — it's a requirement. Splitting there would break exactly the claims most likely to carry tenant context.

Worth a one-line comment on `extractWhitelistedClaims` saying the flat lookup is deliberate, since the next reader will notice the same mismatch against `injectClaims` and reach for the same wrong conclusion I did.

wbabyte:
Round 2 — pushed `c287898`. All four addressed.

- **Blocklist → allowlist (#1).** Agreed, and thank you for the enumeration — you're right the blocklist missed the sharpest fields and would rot. Inverted it: injection now targets an **allowlist by exclusion**. `STANDARD_DCR_FIELDS` is derived by reflection from `ClientRegistrationRequest`'s `@JsonProperty` declarations (+ `client_id`, which only lives on the tree via the update flow), and a mapping is rejected if its top-level segment is a standard field. Only extension fields (`metadata.*`, `organization`, `tenant_id`, …) are injectable — safe-by-default: a field added to `ClientRegistrationRequest` later is automatically protected. `scope`, `software_statement`, `request_uris`, `sector_identifier_uri`, `initiate_login_uri`, `logo/policy/tos_uri` are all now covered. Documented on the constant.
- **Injection overriding `client_id`/`scope` (#2).** Both are now non-injectable via the allowlist (`scope` is a `@JsonProperty`; `client_id` is added explicitly). Additionally moved `injectClaims(...)` **before** the `client_id`/`scope` block in `update()` so they're structurally un-overridable regardless — belt and suspenders, as you suggested.
- **Updater's vs owner's claims (#3).** Changed to the **application owner's** claims: `updateClientRegistration` now resolves `membershipService.getPrimaryOwnerUserId(...)` and injects that user's claims, so the tenant context stays tied to ownership and doesn't flip when another user edits the app. (Create is unchanged — creator == owner there.) New test asserts the owner's resolved claims are what reaches the DCR update.
- **Flat lookup (#4, withdrawn).** Added the one-line comment on `extractWhitelistedClaims` noting the flat lookup is deliberate (namespaced claim names contain dots that must not be split), so the next reader doesn't reach for the same fix.

Tests green on the affected suite (incl. standard-field-skip, `software_statement` rejection, owner-claims wiring). Re-request when ready.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
```diff
@@ -901,10 +905,18 @@ private void updateClientRegistration(
         Application applicationToUpdate
     ) {
         try {
+            // Inject the application owner's IdP claims (not the editor's) so the tenant/user context registered with
+            // the IdP stays tied to who owns the application and does not flip when another user edits it.
+            String ownerId = membershipService.getPrimaryOwnerUserId(
```
paulatulis: The `ownerId != null` guard below is dead code — `getPrimaryOwnerUserId` does not return null, it throws. And because this call sits inside the existing try, the throw disables the **whole** DCR update, not just claim injection.

Tracing `MembershipServiceImpl`:

- `getPrimaryOwner(...)` returns **`null`** via `.orElse(null)` (:1506) when no primary-owner membership row exists — and `getPrimaryOwnerUserId` (:1519) then calls `primaryOwner.getMemberId()` with no null check → **NPE**.
- **Group-owned application**: the `MembershipMemberType.GROUP` branch resolves through `GroupEntity::getApiPrimaryOwner` and `orElseThrow(NoSuchElementException)` when the group has no API primary owner. (Worth noting separately: for `APPLICATION` that extractor is the group's *API* primary owner, which looks semantically off — but it's pre-existing behaviour, not this PR.)
- `poRole == null` → `RoleNotFoundException`.

All three land in this method's `catch (Exception e)` → *"Failed to update OAuth client data from client registration. Keeping old OAuth client data."* — so the application update silently stops re-registering the OAuth client at all, and the previous metadata is restored. Before this commit `getAuthenticatedUsername()` could not throw, so this is a new failure mode on a path that previously always ran.

The codebase already treats this call as throwing: `SubscriptionServiceImpl.getPrimaryOwnerUserIdOrNull` (:2376) wraps the identical call in `try/catch → return null`.

Resolving the owner defensively — the same `…OrNull` shape, or hoisting the lookup outside the try — makes a missing/unresolvable primary owner degrade to "no claims injected" rather than "OAuth client silently not updated", which matches the intent the null-guard was already reaching for.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/ClientRegistrationServiceImpl.java
```diff
@@ -72,6 +75,32 @@
...
+     * <p>
+     * Claim injection uses an <b>allowlist by exclusion</b>: a mapping may only target a field whose top-level segment
+     * is NOT a standard field, i.e. only vendor/extension fields ({@code metadata.*}, {@code organization}, ...) are
+     * injectable. This is safe-by-default — a standard field added to {@link ClientRegistrationRequest} later is
+     * automatically protected, and the value being IdP/end-user controlled can never override a spec field or bypass
+     * the OAuth client sanitization applied to operator-supplied settings.
+     */
+    private static final Set<String> STANDARD_DCR_FIELDS = computeStandardDcrFields();
+
+    private static Set<String> computeStandardDcrFields() {
```
paulatulis, line 90: Nice inversion — this is safe-by-default in the way the blocklist wasn't, and I verified the reflection is actually complete for the current class: `ClientRegistrationRequest` has no superclass (so `getDeclaredFields()` misses nothing inherited), all 33 instance fields carry a **field-level** `@JsonProperty`, none with an empty value, and none are annotated on getters. So the set genuinely covers every serialized property, plus `client_id`.

One latent gap worth a guard: each of those three properties is load-bearing, and the filter fails **open** if any stops holding. A field added later with no `@JsonProperty` still gets serialized by Jackson under its camelCase bean name, but would be absent from `STANDARD_DCR_FIELDS` — so it would be injectable. Same for a bare `@JsonProperty` (`value()` empty → skipped by the `!isEmpty()` filter) or one moved onto a getter.

A test asserting `STANDARD_DCR_FIELDS` matches the properties Jackson actually serializes would pin it, e.g. comparing against `mapper.valueToTree(new ClientRegistrationRequest())`'s field names, or against `mapper.getSerializationConfig().introspect(...)` — so a future field that slips the convention fails the build rather than quietly becoming injectable.

Not blocking; the current behaviour is correct.

wbabyte:
Round 3 — pushed `239f744`. Both addressed.

- **Owner lookup throws → could silently abort the DCR update (#1).** Correct, and a bug I introduced — `getAuthenticatedUsername()` couldn't throw, `getPrimaryOwnerUserId` can (missing PO → NPE, group-owned → `NoSuchElementException`, missing role → `RoleNotFoundException`), and it sat inside the try so it'd land in the "keeping old data" catch. Extracted `resolveOwnerIdpClaims(...)` mirroring `getPrimaryOwnerUserIdOrNull`: any failure degrades to `null` (→ no claims injected), so an unresolvable owner never aborts the client update. The dead `!= null` guard is gone.
- **Allowlist could fail open for a future field (#2).** Made it structural rather than adding a test: `STANDARD_DCR_FIELDS` is now derived from Jackson's serialization introspection (`SerializationConfig.introspect(...).findProperties()`) instead of hand-rolled `@JsonProperty` field reflection. That's exactly the set Jackson serializes, so a field added later under a getter or bean name is covered automatically — no fail-open. The existing `redirect_uris`/`software_statement` rejection tests exercise it and confirm the introspected names are the snake_case wire names. Thanks for the complete verification of the reflection in the meantime.

Affected suite green.

paulatulis (approved):
We are at the end of the marathon. 

No more blockers! Adding one more nit to take or leave 
One nit, non-blocking. The create path still calls userService.findIdpClaims(executionContext, userId) inline and unguarded, outside any catch — so a transient user-repo error fails application creation outright, while update now degrades to null. That asymmetry is sharper because update got hardened. It's already in my summary comment as a minor; resolveOwnerIdpClaims is right there to reuse. Follow-up material, not a blocker.

PR #18607, opened 2026-07-17
fix(repository): base64-encode DCR claimMappings keys for MongoDB (APIM-13949)

wbabyte:
## What

Follow-up bug fix for epic **APIM-13949**. `ClientRegistrationProvider.claimMappings` is a `Map` **keyed by IdP claim name**. OIDC/Auth0 commonly use namespaced claim names (e.g. `https://acme.example/tenant`), and MongoDB rejects map keys containing dots — so saving such a mapping throws `MappingException` and the client-registration-provider config save fails.

This is the same dotted-key class of bug already fixed for `User.idpClaims` (#18383) and how identity-provider group/role mappings are stored; `claimMappings` was missed because its repo test used a plain `org_id` key. JDBC is unaffected (stored as serialized JSON); only MongoDB's native map storage is impacted.

## How

`MongoClientRegistrationProviderRepository` now Base64-encodes `claimMappings` keys on write and decodes on read, centralised in the `map(...)` conversion methods (plus the manual `update` setter), and `findAll`/`findAllByEnvironment` route through the decoding mapper.

## Tests

`ClientRegistrationProviderRepositoryTest.should_persist_and_read_claim_mappings` now uses a namespaced (dotted) claim-name key across create, update and read — this reproduced the failure before the fix.

> Note: found via local end-to-end verification of the merged feature against a mock OIDC provider. I could not run the Mongo Testcontainers test locally in this session (Docker Desktop was unresponsive in my environment); the change compiles cleanly and mirrors the proven `User.idpClaims` fix — relying on CI to exercise the Mongo path.

Ref: **APIM-13949**.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoClientRegistrationProviderRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoClientRegistrationProviderRepository.java
@@ -21,9 +21,13 @@ import io.gravitee.repository.management.model.ClientRegistrationProvider;
 import io.gravitee.repository.mongodb.management.internal.application.ClientRegistrationProviderMongoRepository;
 import io.gravitee.repository.mongodb.management.internal.model.ClientRegistrationProviderMongo;
 import io.gravitee.repository.mongodb.management.mapper.GraviteeMapper;
+import java.util.Base64;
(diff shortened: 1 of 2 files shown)


wbabyte:
Verified locally now that Docker is back: `ClientRegistrationProviderRepositoryTest` **green on both MongoDB and JDBC (12/12 each)**, including the namespaced/dotted claim-name key (`https://acme.example/tenant`) that reproduced the `MappingException` before the fix.

vikrantgravitee (approved):

PR #18620, opened 2026-07-19
feat(api-product): support AI Workspace product kind and per-plan budget flows

vikrantgravitee:
## What

Introduces an `ApiProductKind` discriminator so an API Product can be a specialized kind (AI Workspace) **without a new table or refType**. Plain (kind-less) API Products behave exactly as before — verified.

## Changes

- `ApiProductKind`, `ApiProductKindFilter`, `ApiProductComposition`; specialized kinds are excluded from the classic API Product list and search.
- Hydrate v4 plan flows into the API Product deployment payload via a single bulk `FlowCrudService.getPlanV4Flows(Set)` lookup, so per-plan budget and rate-limit policies actually reach the gateway (no N+1).
- Persist v4 plan flows on API Product plan create and update.
- Resolve the product-plan policy manager through the ancestor-searching bean lookup (`getBeanNamesForType`) so **reactor plugins** (LLM proxy) receive product security. This was the root cause of a 401 on product-plan-only APIs: the plugin runs in a child Spring context and the direct `applicationContext.getBeanNamesForType(...)` returned empty.
- Bulk api-key lookup by subscription ids.
- Index and persist `allowedInApiProducts` for composable API types (`PROXY`, `LLM_PROXY`).
- Version bumps: `gravitee-reactor-llm-proxy` 3.1.1, `gravitee-gamma-module-aim` 1.1.0-alpha.14.

## Tests

140 affected unit tests pass (product deploy/flow hydration, plan create/update flow persistence, bulk api-key, lucene kind exclusion, mapper composable-type round-trip). Includes a regression test proving the ancestor-searching bean lookup resolves the policy manager from a child context.

## Compatibility

The `kind` discriminator is additive: `ApiProductKindFilter` defaults to `any()`, the lucene `kind` term is only added when present, and classic products carry no `kind` so they are never excluded.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api_product/ApiProductsResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api_product/ApiProductsResource.java
@@ -16,6 +16,7 @@
 package io.gravitee.rest.api.management.v2.rest.resource.api_product;
 
 import io.gravitee.apim.core.api_product.model.ApiProduct;
+import io.gravitee.apim.core.api_product.model.ApiProductKindFilter;
 import io.gravitee.apim.core.api_product.model.CreateApiProduct;
 import io.gravitee.apim.core.api_product.use_case.CreateApiProductUseCase;
 import io.gravitee.apim.core.api_product.use_case.GetApiProductsUseCase;
@@ -158,7 +159,8 @@ public class ApiProductsResource extends AbstractResource {
             paginationParam.toPageable(),
             apiProductSortByParam.toSortable(),
             getAuthenticatedUser(),
-            isAdmin()
+            isAdmin(),
+            ApiProductKindFilter.classicOnly()
         );
         var output = searchApiProductsUseCase.execute(input);
         var page = output.page();
@@ -195,7 +197,8 @@ public class ApiProductsResource extends AbstractResource {
                 executionContext.getEnvironmentId(),
                 executionContext.getOrganizationId(),
                 getAuthenticatedUser(),
-                isAdmin()
+                isAdmin(),
+                ApiProductKindFilter.classicOnly()
             )
         );
         Set<ApiProduct> allApiProducts = output.apiProducts();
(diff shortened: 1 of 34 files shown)


vikrantgravitee:
Addressed the review comments:

- **#1 (log + throw in `getPlanV4Flows(Set)`)** — fixed. The catch now rethrows `TechnicalDomainException` without the preceding `log.error`, so the exception is handled once at a higher level.
- **#3 (inline ternary in `checkSynchronized`)** — fixed. Extracted `effectiveFlows` as a named local before the `checkSynchronized` call in `UpdatePlanDomainService.updatePlanForApiProduct`.
- **#2 (consolidate the bulk flow getters)** — left as-is for now. `getPlanV4Flows(Set)` mirrors the existing `getPlanV2Flows(Set)` / `getNativePlanFlows(Set)` structure; a shared generic helper would also touch those pre-existing methods, so it's better handled as a separate focused cleanup than bundled into this feature PR.

Verified: rest-api-service compiles; `UpdatePlanDomainServiceTest` + `DeployApiProductUseCaseTest` green (59 tests).

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api_product/ApiProductsResource.java
```diff
@@ -158,7 +161,8 @@ public Response searchApiProducts(
             paginationParam.toPageable(),
             apiProductSortByParam.toSortable(),
             getAuthenticatedUser(),
-            isAdmin()
+            isAdmin(),
+            EnumSet.allOf(ApiProductKind.class)
```
araffass: One possible improvement here since EnumSet.allOf(ApiProductKind.class) works today because the only value is AI_WORKSPACE, but if we add a second kind later it'll silently get excluded too. Maybe consider EnumSet.of(AI_WORKSPACE) to make the intent explicit

araffass (approved):

vikrantgravitee:
Good catch on `EnumSet.allOf(ApiProductKind.class)` — addressed by making the search endpoint speak the same intent as the list endpoint.

**Why not `EnumSet.of(AI_WORKSPACE)`?** That would be the opposite bug. Both `/api-products` endpoints are meant to show **classic** products only (the `GET` list already uses `ApiProductKindFilter.classicOnly()`). With `EnumSet.of(AI_WORKSPACE)`, a future second kind would *leak into* the classic search instead of being hidden — and it would drift from the list endpoint.

**Fix:** the search endpoint now passes `ApiProductKindFilter.classicOnly()` too, and `SearchApiProductsUseCase` takes an `ApiProductKindFilter` (matching `GetApiProductsUseCase`). The filter derives the excluded set as *every kind not explicitly included* (`ApiProductKindFilter.excludedKinds()`), so "classic only" always excludes all specialized kinds — present and future — expressed intentionally rather than via `allOf`.

Added `ApiProductKindFilterTest` to lock this in (`classicOnly().excludedKinds()` == all kinds; `any()` == none). Both `/api-products` endpoints now share one `ApiProductKindFilter` abstraction. Verified: rest-api-service + management-v2-rest compile; `SearchApiProductsUseCaseTest` + `ApiProductDocumentSearcherTest` + the new filter test green.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/resource/api_product/ApiProductsResource.java
vikrantgravitee: Done

PR #18634, opened 2026-07-20
fix: http 500 on api product api update

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-14738

## Description

Fixes HTTP 500 when batch-updating an API Product (PUT .../api-products/{id} with a large apiIds list).

**Root cause**: ValidateApiProductService.validateApiIdsForProduct() used Collectors.toMap(Api::getId, api -> api) without a merge function. If apiQueryService.search() emits the same API ID more than once, Java throws IllegalStateException: Duplicate key ... (attempted merging values Api(...) and Api(...)), which is mapped to 500 by ThrowableMapper.
**Fix**: keep the first occurrence:
.collect(Collectors.toMap(Api::getId, api -> api, (existing, duplicate) -> existing));
Safe for this path: validation only needs existence / V4 / allowedInApiProducts. Duplicate search hits do not change the outcome.

Customer evidence : MAPI logs show the exact Duplicate key c661a03d-... / alm_exchange + ThrowableMapper - Internal error. Atlas shows a single Mongo document for that _id (not duplicate DB rows). Single-API PUT succeeds; request-body ID dedupe also succeeds.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/domain_service/ValidateApiProductService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/domain_service/ValidateApiProductService.java
@@ -69,7 +69,9 @@ public class ValidateApiProductService {
         ApiSearchCriteria criteria = ApiSearchCriteria.builder().ids(apiIds).environmentId(environmentId).build();
         ApiFieldFilter fieldFilter = ApiFieldFilter.builder().definitionExcluded(false).pictureExcluded(true).build();
 
-        Map<String, Api> foundApis = apiQueryService.search(criteria, null, fieldFilter).collect(Collectors.toMap(Api::getId, api -> api));
+        Map<String, Api> foundApis = apiQueryService
+            .search(criteria, null, fieldFilter)
+            .collect(Collectors.toMap(Api::getId, api -> api, (existing, duplicate) -> existing));
 
         List<String> nonExistentApiIds = apiIds
             .stream()
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api_product/domain_service/ValidateApiProductServiceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api_product/domain_service/ValidateApiProductServiceTest.java
@@ -149,6 +149,14 @@ class ValidateApiProductServiceTest {
 
             assertThatCode(() -> validateApiProductService.validateApiIdsForProduct(ENV_ID, List.of("api-1"))).doesNotThrowAnyException();
         }
+
+        @Test
+        void should_handle_duplicate_search_results_without_error() {
+            Api api = createV4ProxyApi("api-1", true);
+            apiCrudService.initWith(List.of(api, api));
+
+            assertThatCode(() -> validateApiProductService.validateApiIdsForProduct(ENV_ID, List.of("api-1"))).doesNotThrowAnyException();
+        }
     }
 
     @Nested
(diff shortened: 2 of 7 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/domain_service/ValidateApiProductService.java
```diff
@@ -69,7 +69,9 @@ public void validateApiIdsForProduct(String environmentId, @Nonnull List<String>
         ApiSearchCriteria criteria = ApiSearchCriteria.builder().ids(apiIds).environmentId(environmentId).build();
         ApiFieldFilter fieldFilter = ApiFieldFilter.builder().definitionExcluded(false).pictureExcluded(true).build();
 
-        Map<String, Api> foundApis = apiQueryService.search(criteria, null, fieldFilter).collect(Collectors.toMap(Api::getId, api -> api));
+        Map<String, Api> foundApis = apiQueryService
+            .search(criteria, null, fieldFilter)
+            .collect(Collectors.toMap(Api::getId, api -> api, (existing, duplicate) -> existing));
```
aqibmohammadkhan, line 74: Customer evidence fits unstable Mongo streaming pagination more than duplicate DB rows:

- `apiRepository.search(...)` streams in batches of 100 (ApiRepository.DEFAULT_STREAM_BATCH_SIZE)
- Default sort is name ASC only — no _id tie-breaker (ApiMongoRepositoryImpl)
- Offset pagination + non-unique sort can emit the same API on two pages → IllegalStateException: Duplicate key → 500

That matches: large apiIds fail, single-API PUT works, Atlas shows one document per _id.

After this PR, duplicates no longer 500 — but skipped pages can still produce false “API does not exist” validation errors. The merge function should stay, the real fix belongs in repository search (stable sort, e.g. name + _id, or a non-paginated ID lookup).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api_product/domain_service/ValidateApiProductServiceTest.java
```diff
@@ -149,6 +149,14 @@ void should_not_throw_when_apis_valid_and_allowed() {
 
             assertThatCode(() -> validateApiProductService.validateApiIdsForProduct(ENV_ID, List.of("api-1"))).doesNotThrowAnyException();
         }
+
+        @Test
+        void should_handle_duplicate_search_results_without_error() {
+            Api api = createV4ProxyApi("api-1", true);
+            apiCrudService.initWith(List.of(api, api));
```
aqibmohammadkhan, line 156: apiCrudService.initWith(List.of(api, api)) correctly exercises toMap, but does not model production (no pagination, Mongo can’t store two docs with the same _id). A repo-level test with 101+ APIs and shared names around the page boundary would lock the real bug.

aqibmohammadkhan: do we need to check for the JDBC repository to be checked for the same pattern.

pragsty: Done

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/domain_service/ValidateApiProductService.java
pragsty: Done

aqibmohammadkhan (approved):

PR #18662, opened 2026-07-21
fix(tracing): report distinct-trace total so trace search paginates past page 1

remibaptistegio:
## Problem
The trace explorer showed only one page regardless of how many traces matched. `searchTraces` returned a bare `List<Trace>`, so the gamma adapter derived the total from the fetched slice size (`== perPage`) — `pageCount` was always 1.

## Fix
- Elasticsearch `searchTraces` adds a `cardinality(trace_id)` aggregation; the SPI now returns `Page<Trace>` carrying the distinct-trace total, **capped at `MAX_SEARCH_RESULTS` (200)** so no unreachable pages are advertised.
- The gamma adapter uses that total for the paginated response; `NoOpTracingRepository` returns an empty page.
- No lib/frontend change — the `{ data, pagination: { totalCount } }` wire contract is unchanged.

## Tests
Added/updated Elasticsearch unit tests (distinct total, 200 cap, empty page) and the repository contract test.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
@@ -19,6 +19,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.gravitee.common.data.domain.Page;
 import io.gravitee.elasticsearch.client.Client;
 import io.gravitee.elasticsearch.model.Aggregation;
 import io.gravitee.elasticsearch.model.SearchHit;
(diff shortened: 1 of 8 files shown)


gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
araffass: Sonar's firing here as the new cardinalityAgg adds the 3rd "field" literal in this file (alongside the terms agg and minAgg). Can you maybe extract it into a constant (e.g. private static final String AGG_FIELD_KEY = "field";) and use it at the three call sites, so the quality gate goes green?

araffass (approved):
Apart from the Sonar - All look good to me

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/tracing/ElasticsearchTracingRepository.java
remibaptistegio: Good catch — done in 14cff50. Extracted `private static final String AGG_FIELD_KEY = "field";` and used it at the terms, min and cardinality aggs, so the literal appears once now. Thanks!

PR #18663, opened 2026-07-21
fix(APIM-14729): allow observability dashboard filters without DASHBOARD update permission

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14729
## Description

Users with the stock API_PUBLISHER environment role can open an Observability dashboard but can't filter it — Add filter is hidden, chips are read-only, and Clear is gone. The filter bar was wired to environment-dashboard-u in #16628, which effectively requires DASHBOARD: UPDATE just to read data through a filter.

Frontend only, 3 files, all in dashboard-viewer:

- removed the canEditFilters computed, the GioPermissionService injection, and the early-return guards in openAddFilter() / openEditFilter()
- dropped [editable]="canEditFilters()" — the lib input already defaults to true
- replaced the two specs that asserted the old gating with one regression test: with hasAnyMatching stubbed to false, the Add button renders and clicking it opens the dialog

Dashboard editing is untouched — rename, add widget, edit layout and delete stay gated on environment-dashboard-u in dashboard-detail.component.ts:60-62.

Testing

- dashboard-viewer.component.spec.ts — 6/6 passing
- Confirmed the new test is a real guard: with the production changes stashed (old gating restored) it fails; with them applied it passes
- ESLint and tsc clean on the changed files
- 
## Additional context

Before
Screenshot 2026-07-21 at 5 27 17 PM

After
Screenshot 2026-07-21 at 5 28 40 PM

### DIFF
--- a/gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-viewer.component.html
+++ b/gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-viewer.component.html
@@ -43,7 +43,6 @@
       <div class="observability-dashboard-viewer__filters-row">
         <gd-dynamic-filter-bar
           [conditions]="filtersStore.conditions()"
-          [editable]="canEditFilters()"
           (addRequested)="openAddFilter()"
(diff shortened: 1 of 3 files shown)


araffass (approved):
LGTM

PR #18680, opened 2026-07-22
[master] fix(authz-sync): change-gate to stop authz PDP commit churn

podlesrafal:
## Issue

https://gravitee.atlassian.net/browse/AUTHZ-50 (related)

## Description

Manual master port of #18369 (authz change-gate, shipped in 4.12.4), replacing the stale Mergify backport #18422.

#18422 could not be salvaged: its cherry-pick commit has the conflict markers committed into `AuthzEntityMapperTest.java`, and rebasing it onto current master stacks new conflicts on top of the committed ones. The original conflict was against the pre-refactor master of July 3rd; current master already contains the engine-uid consolidation refactor, so both original 4.12.x commits now apply cleanly:

- `10016785ec` fix(authz-sync): change-gate to stop authz PDP commit churn
- `365ff7b574` fix(authz-sync): correct change-gate evict key, commit-abandon rollback and unknown-timestamp handling

Both picked without conflicts, no manual edits.

## Additional context

- Without this gate, master re-applies every authz event within the ±30s sync timeframe window on every 5s cycle (~12 re-applies + engine commit per event) — the "commit churn" fixed on 4.12.x.
- Verified: `mvn verify` on `gravitee-apim-gateway-services-sync` — 655 tests, 0 failures (includes the 13 change-gate tests).
- Follow-up: the authz sync error-isolation fix (#18672, currently 4.12.x-based) will be ported on top of this once merged.

Replaces #18422.


podlesrafal:
Closing: wrong flow on my side — the isolation PR #18672 (4.12.x) now carries the `apply-on-master` label so Mergify handles the master port, and #18422 stays as the master carrier for the change-gate (its branch just needs a force-push with the clean cherry-picks, which apply without conflicts on current master — see #18422 comments).

PR #18682, opened 2026-07-22
fix(gateway): stop losing distributed sync events during bulk syncs

ThibaudAV:
## Issue

https://gravitee.atlassian.net/browse/APIM-14672

## Description

Follow-up to #18626 (which fixed the distributed sync **read** path). This PR fixes the **write** path on the primary node, root cause of the bulk-sync field report: *5,000 APIs deployed locally, only 832 with data in Redis, logs full of `Redis waiting queue is full`, no surfaced sync failure*.

### `fix(redis): key the per-event-loop connection on the event loop, not the Vertx instance`

`RedisClient.currentLoopKey()` used `System.identityHashCode(context.owner())` — but `Context.owner()` returns the shared **Vertx instance**, so every event loop mapped to the same connection entry. The whole process (rate-limit + distributed sync) serialized on **one TCP connection** with a single `max-waiting-handlers = 1024` in-flight cap, silently defeating the per-event-loop connection feature. The key is now the netty event loop (per-request duplicated contexts still share their loop's connection).

### `fix(gateway): stop losing distributed sync events during bulk syncs`

Two compounding problems on the primary's distribution path:

1. **Unbounded write fan-out** — `distributeIfNeeded` used a `flatMapCompletable` with unlimited concurrency; a single API deployable fans out into one Redis write per subscription and API key (× the parallel deployer pool). Thousands of concurrent HSETs overflow the client-side waiting queue → `Redis waiting queue is full` (Redis itself is healthy; the queue is in the vertx client). Writes are now capped at **32 concurrent commands per deployable**, letting backpressure pace the stream.
2. **Silent data loss** — the repository synchronizers isolate per-item errors (`onErrorResumeNext → empty`), so a failed distribution was logged, dropped, and the sync window advanced anyway: the events were **never re-emitted** and secondary gateways diverged permanently. Distribution failures are now tracked and fail `storeState`, so the sync state is not persisted and **the same time window is replayed on the next 5s cycle** — the sync manager already handles this (window not advanced, throttled error logging), no synchronizer changes needed.

## Additional context

- Root-cause analysis of the field report: with the old code, each queue-full error dropped one API's whole event batch; the `X apis synchronized` counter only counts survivors — consistent with 5,000 → 832 → "178 synchronized".
- New tests: connection-key semantics per event loop / same loop / duplicated context; write-concurrency cap; `storeState` failing after a distribution failure then recovering (window replay).
- Full module test suites green: repository-redis (63), gateway-services-sync (404).

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/distributed/service/DefaultDistributedSyncService.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/distributed/service/DefaultDistributedSyncService.java
@@ -43,12 +43,15 @@ import io.gravitee.node.api.Node;
 import io.gravitee.node.api.cluster.ClusterManager;
 import io.gravitee.repository.distributedsync.api.DistributedEventRepository;
 import io.gravitee.repository.distributedsync.api.DistributedSyncStateRepository;
+import io.gravitee.repository.distributedsync.model.DistributedEvent;
 import io.gravitee.repository.distributedsync.model.DistributedEventType;
 import io.gravitee.repository.distributedsync.model.DistributedSyncAction;
 import io.gravitee.repository.distributedsync.model.DistributedSyncState;
 import io.reactivex.rxjava3.core.Completable;
+import io.reactivex.rxjava3.core.Flowable;
 import io.reactivex.rxjava3.core.Maybe;
 import java.util.Date;
+import java.util.concurrent.atomic.AtomicBoolean;
 import lombok.CustomLog;
 import lombok.RequiredArgsConstructor;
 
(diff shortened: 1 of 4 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/distributed/service/DefaultDistributedSyncService.java
```diff
@@ -258,10 +279,31 @@ public Completable distributeIfNeeded(final NodeMetadataDeployable deployable) {
...
+    /**
+     * A failed distribution must not be silently dropped: the repository synchronizers isolate
+     * (un)deployment errors per item, so without this flag the sync window would advance and the
+     * missing events would never reach the secondary gateways. {@link #storeState(long, long)}
+     * checks the flag and fails the cycle so the same window is replayed.
+     */
+    private Completable trackFailure(final Completable distribution, final String eventType, final String id) {
+        return distribution.doOnError(throwable -> {
+            distributionFailed.set(true);
+            log.error("Unable to distribute {} event for [{}], the sync time window will be replayed", eventType, id, throwable);
```
ThibaudAV: Good catch — this also matches our internal logging rules (log once; wrap with context when the error propagates). Fixed in a1f05fad72: `trackFailure` no longer logs; it sets the failure flag and wraps the error in a contextualized `DistributedSyncException` (event type + id), which the downstream synchronizers/sync manager log once.

schehade-gravitee (approved):

PR #18692, opened 2026-07-22
fix(rest-api): emit audit log on parameter deletion (APIM-14773)

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14773

## Description

Parameter deletion emitted no audit entry. `ParameterServiceImpl` audited its create and update branches but not its delete branch, so removing a parameter left no trace.

Adds a `PARAMETER_DELETED` audit log to the shared delete helper, recording the stored parameter as the old value and `null` as the new value.

## Additional context

Screenshot 2026-07-22 at 10 42 35 PM

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
@@ -17,6 +17,7 @@ package io.gravitee.rest.api.service.impl;
 
 import static io.gravitee.repository.management.model.Audit.AuditProperties.PARAMETER;
 import static io.gravitee.repository.management.model.Parameter.AuditEvent.PARAMETER_CREATED;
+import static io.gravitee.repository.management.model.Parameter.AuditEvent.PARAMETER_DELETED;
 import static io.gravitee.repository.management.model.Parameter.AuditEvent.PARAMETER_UPDATED;
 import static io.gravitee.rest.api.model.parameters.ParameterReferenceType.ENVIRONMENT;
 import static io.gravitee.rest.api.model.parameters.ParameterReferenceType.ORGANIZATION;
@@ -473,6 +474,7 @@ public class ParameterServiceImpl extends TransactionalService implements Parame
                         key.key(),
                         refIdToUse,
                         ParameterReferenceType.valueOf(referenceType.name()),
+                        optionalParameter.get(),
                         executionContext
                     );
                     return null;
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
```diff
@@ -740,10 +742,11 @@ public void delete(
         String refIdToUse = getEffectiveReferenceId(executionContext, referenceId, referenceType);
         ParameterReferenceType repositoryReferenceType = ParameterReferenceType.valueOf(referenceType.name());
         try {
-            if (parameterRepository.findById(key.key(), refIdToUse, repositoryReferenceType).isEmpty()) {
+            Optional<Parameter> optionalParameter = parameterRepository.findById(key.key(), refIdToUse, repositoryReferenceType);
+            if (optionalParameter.isEmpty()) {
                 return;
             }
-            deleteParameterAndInvalidateCache(key.key(), refIdToUse, repositoryReferenceType, executionContext);
+            deleteParameterAndInvalidateCache(key.key(), refIdToUse, repositoryReferenceType, optionalParameter.get(), executionContext);
```
ytvnr: Nothing major, but juggling the `Optional` here without real optional chaining only adds verbosity.

Note that `deleteParameterAndInvalidateCache` throws a checked `TechnicalException`, so `ifPresent(...)` won't compile. Still, we can drop the `.get()`:

```java
Parameter deletedParameter = parameterRepository.findById(key.key(), refIdToUse, repositoryReferenceType).orElse(null);
if (deletedParameter == null) {
    return;
}
deleteParameterAndInvalidateCache(key.key(), refIdToUse, repositoryReferenceType, deletedParameter, executionContext);
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
```diff
@@ -758,17 +761,32 @@ public void delete(
...
+        // Audited as soon as the removal is persisted: the audit records the durable fact, while the cache
+        // invalidation below is best-effort housekeeping that must not sit between the two.
+        auditService.createAuditLog(
+            executionContext,
+            AuditService.AuditLogData.builder()
+                .properties(singletonMap(PARAMETER, key))
+                .event(PARAMETER_DELETED)
+                .createdAt(new Date())
+                .oldValue(deletedParameter)
+                .newValue(null)
```
ytvnr: `AuditServiceImpl` already defaults `createdAt` to `new Date()` when it is null (`AuditServiceImpl.java:369`), so this line can just be dropped. Same for `.newValue(null)`, which is already the builder default.

It also keeps the new test free of a non-deterministic date to match.

ytvnr (approved):
LGTM. Two minor readability suggestions inline, nothing blocking.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
paulatulis: Addressed!

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ParameterServiceImpl.java
paulatulis: Addressed

PR #18695, opened 2026-07-22
fix(console): explain SYSTEM-locked branded senders with a padlock and tooltip

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14775

## Description

When branded senders are set in `gravitee.yml`, the section renders greyed out with no padlock and no tooltip — visually identical to the missing-`ENVIRONMENT_SETTINGS[U]` state. An operator sees a dead form with no stated reason and can't tell "locked by system configuration" from "you lack permission".

Adds a `systemLocked` input to the shared `branded-senders` component, rendering the same padlock + tooltip treatment already used by the sibling default fields. Both hosts pass their existing readonly check `isReadonlySetting('email.branded_senders')` in org settings, `isReadonly('email.branded_senders')` in portal settings.

## Additional context

Screenshot 2026-07-22 at 4 45 55 PM

### DIFF
--- a/gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
@@ -20,7 +20,7 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { HttpTestingController } from '@angular/common/http/testing';
 import { HarnessLoader } from '@angular/cdk/testing';
 import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
-import { MatIconTestingModule } from '@angular/material/icon/testing';
+import { MatIconHarness, MatIconTestingModule } from '@angular/material/icon/testing';
 import { InteractivityChecker } from '@angular/cdk/a11y';
 import { MatButtonHarness } from '@angular/material/button/testing';
 import {
@@ -654,6 +654,20 @@ describe('PortalSettingsComponent', () => {
       await init();
     });
 
+    // Guards the property the padlock specs below silently depend on. They can only catch a mis-keyed
+    // [systemLocked] binding because the fixture's base readonly covers other email.* keys while omitting this one:
+    // a wrong key resolves to a read-only entry and lights the padlock where the negative specs expect none. Adding
+    // 'email.branded_senders' to the fixture would blunt them without failing anything, so fail here instead.
+    it('should keep email.branded_senders out of the fixture base readonly, but other email keys in it', () => {
+      // The component created in beforeEach issues a settings GET; flush it so afterEach's verify() stays clean.
+      expectPortalSettingsGetRequest(fakePortalSettings());
+
+      const readonly = fakePortalSettings().metadata.readonly;
+
+      expect(readonly).not.toContain('email.branded_senders');
+      expect(readonly).toEqual(expect.arrayContaining(['email.from', 'email.subject']));
+    });
+
     it('should render the branded senders control enabled when emailing is on and not read-only', async () => {
       const settings = fakePortalSettings();
       settings.email.enabled = true;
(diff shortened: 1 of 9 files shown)


gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.html
```diff
@@ -45,8 +45,26 @@ <h3 class="branded-senders__title">Default notification email</h3>
...
-  <h3 class="branded-senders__title">Branded notification email</h3>
+  <!--
+    Branded notification email (editable rules).
+
+    The lock tooltip is hosted on the <h3> rather than on the icon, matching the console's container-tooltip pattern
+    and — because a heading is a semantic, AT-navigable node — giving MatTooltip's AriaDescriber somewhere the
+    description is actually announced. The glyph stays decorative (MatIcon defaults it to aria-hidden="true").
+    Note this only holds for the heading: the same placement on a <mat-form-field> lands the description on a
+    role-less div that never reaches the input, which is why the two preview locks above are left alone.
+  -->
```
ytvnr, line 56: The last two lines here are the valuable part: they explain why the two preview locks above were deliberately left alone, which is not deductible from the code and is exactly what a reviewer would wonder. Worth keeping.

The first three are more debatable, since the code already shows that the tooltip sits on the `<h3>` and that the icon is decorative.

Related: the "indistinguishable from missing `ENVIRONMENT_SETTINGS[U]`" rationale is restated four times across the diff (the input's JSDoc, plus one comment in each of the three specs, two of them word for word). The JSDoc is the right home for it, the specs could just point at it.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.html
```diff
@@ -45,8 +45,26 @@ <h3 class="branded-senders__title">Default notification email</h3>
...
+  <h3
+    class="branded-senders__title"
+    data-testid="branded-senders-branded-heading"
+    matTooltip="Configuration provided by the system"
+    [matTooltipDisabled]="!systemLocked()"
+  >
```
ytvnr, line 62: Two minor points on this block:

- `"Configuration provided by the system"` is hardcoded here, while `org-settings-general.component.ts` already exposes it as `providedConfigurationMessage`. That string now lives in 7 HTML files across the console, so a shared constant would be welcome at some point (not necessarily in this PR).
- `MatTooltip` triggers on hover, focus and long-press, and an `<h3>` is not focusable. So a sighted keyboard-only user has no way to read the reason. `aria-describedby` covers screen readers, not the keyboard. Same limitation as the other tooltips on the page, so not a regression here, but it is the one thing I would want solved before this pattern gets replicated.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.scss
```diff
@@ -27,7 +27,21 @@ $warn: map.get(gio.$mat-theme, warn);
 
   &__title {
     @include mat.m2-typography-level($typography, subtitle-1);
+    display: flex;
+    align-items: center;
+    gap: 4px;
+    // Keeps the heading from stretching across the card: the lock tooltip is hosted on this element, so a
+    // full-width h3 would arm it over empty space to the right and anchor the bubble away from the padlock.
+    width: fit-content;
```
ytvnr, line 35: `display: flex` and `width: fit-content` land on `&__title`, so they also apply to the "Default notification email" heading, which never carries a padlock or a tooltip. The `width: fit-content` comment justifies it by tooltip anchoring, and that only holds for the branded heading. Harmless, but a modifier class would keep the rule where it is actually needed.

gravitee-apim-console-webui/src/management/settings/portal-settings/portal-settings.component.spec.ts
```diff
@@ -654,6 +654,20 @@ describe('PortalSettingsComponent', () => {
       await init();
     });
 
+    // Guards the property the padlock specs below silently depend on. They can only catch a mis-keyed
+    // [systemLocked] binding because the fixture's base readonly covers other email.* keys while omitting this one:
+    // a wrong key resolves to a read-only entry and lights the padlock where the negative specs expect none. Adding
+    // 'email.branded_senders' to the fixture would blunt them without failing anything, so fail here instead.
+    it('should keep email.branded_senders out of the fixture base readonly, but other email keys in it', () => {
```
ytvnr, line 661: This asserts on `fakePortalSettings()` rather than on the component, and it has to fire a settings GET purely to keep `afterEach`'s `verify()` clean. The intent is good, but it will fail if someone legitimately adds the key to the fixture even though nothing is actually broken. Setting `settings.metadata.readonly` explicitly in the negative specs would make them self-sufficient and remove the need for this guard.

ytvnr (approved):
The core idea is right: `setDisabledState()` carries no reason, so passing it as a separate input is the proper answer rather than a workaround. The key `email.branded_senders` matches what both hosts already use for their `disabled` logic, so the padlock and the disabled state cannot drift apart. Nice touch covering the intersection case (read-only *and* email disabled) and the negative case where the user is merely unprivileged, since that ambiguity is the whole point of the change.

A few non-blocking remarks inline, nothing that should hold up the merge.

**One thing worth raising with product before this spreads.** This PR does not just fix a field, it introduces a new way of communicating read-only state, and the console is currently inconsistent about it:

| Page | "provided by the system" tooltips | Padlocks |
|---|---|---|
| `org-settings-general` | 23 | 23 |
| `org-settings-entrypoints-and-sharding-tags` | 4 | 4 |
| `api-quality-rules` | 10 | 11 |
| `identity-providers` | 2 | 2 |
| `api-logging` | 12 | 4 |
| `portal-settings` (this page) | 48 | 0 |
| `settings-analytics` | 1 | 0 |

This adds the very first padlock to `portal-settings`. Strictly better for branded senders, but it also leaves one explained field sitting next to 48 that stay silently greyed out. Before rolling this out further we should agree with product on the target treatment, and the question is not only "padlocks everywhere": on a page with 48 locked fields, a section-level notice may read better than a per-field marker on a heavily `gravitee.yml`-driven instance.

(Naming is confusing here, so for the record: both pages touched are console pages. The Portal apps have no `metadata.readonly` handling at all, so there is nothing to mirror on that side.)

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.html
paulatulis: Thank you for flagging! I reached out to @varungoel-sudo and it was determined we should revert to the previous behavior to stay consistent.

paulatulis:
Closing. This introduces a new design pattern too late in the game.
Screenshot 2026-07-23 at 8 15 25 AM

PR #18697, opened 2026-07-23
fix(apim-gamma): shift delete metadata from side to centre modal

arpzilla-dev:
## Issue

https://gravitee.atlassian.net/browse/FOUND-16

## Description

Shift delete metadata from side modal to centre modal

## Additional context

https://github.com/user-attachments/assets/55033a2d-972d-4192-994c-565dfd23615c


ankita-gupta21 (approved):

ankita-gupta21:
Cancel is correctly disabled while isDeleting, but Esc / overlay still call onCancel and can close the dialog mid-delete.

Same pattern exists on other platform delete dialogs today, so leaving this PR as-is for consistency. Can have a small discussion to block dismiss while pending across delete dialogs

PR #18713, opened 2026-07-23
fix(console,portal): trim leading/trailing whitespace from client_id before saving

araffass:
## Issue

https://gravitee.atlassian.net/browse/APIM-14763

## Description

Trim whitespace from client_id in the UI                                                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  A Client ID pasted with leading/trailing whitespace was saved as-is, causing gateway auth/subscription lookup failures with misleading "no subscription found" errors.                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  The value is now trimmed on save in every editable client_id field - console (edit + create), classic portal (edit + create), and new portal (edit) — using the same ?.trim() || undefined pattern the new-portal create flow already had. Whitespace-only input saves as unset. Regression test added on the console General page.

## Additional context

https://github.com/user-attachments/assets/42012493-7aa0-4a9a-a453-c00dfb080ad6


<!-- Add any other context about the PR here -->
<!-- It can be links to other PRs or docs or drawing -->
<!-- Or reproduction steps in case of bug fix -->

### DIFF
--- a/gravitee-apim-console-webui/src/management/application/creation/application-creation.component.ts
+++ b/gravitee-apim-console-webui/src/management/application/creation/application-creation.component.ts
@@ -127,7 +127,7 @@ export class ApplicationCreationComponent implements OnInit {
           ...(applicationPayload.type === 'SIMPLE'
             ? {
                 app: {
-                  client_id: applicationPayload.appClientId,
+                  client_id: applicationPayload.appClientId?.trim() || undefined,
                   type: applicationPayload.appType,
                 },
               }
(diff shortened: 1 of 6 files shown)


gravitee-apim-console-webui/src/management/application/creation/application-creation.component.ts
```diff
@@ -127,7 +127,7 @@ export class ApplicationCreationComponent implements OnInit {
           ...(applicationPayload.type === 'SIMPLE'
             ? {
                 app: {
-                  client_id: applicationPayload.appClientId,
+                  client_id: applicationPayload.appClientId?.trim() || undefined,
```
michel-barret, line 130: what do you think about create a small method `trim(input?: string): string | undefined` for that?

michel-barret (approved):

PR #18724, opened 2026-07-23
fix(console): surface required errors on newly added branded-sender card (APIM-14777)

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14777

## Description

1. Surface card errors on Save, not on Add
- Rewired BrandedSendersComponent from the NG_VALUE_ACCESSOR/NG_VALIDATORS forwardRef providers to the established nested-CVA pattern used by ApiPlanFormComponent (and the design-system's GioFormJsonSchemaComponent): inject the host NgControl, self-assign ngControl.valueAccessor = this, register validation via addValidators.
- In ngOnInit, listen for the host control becoming touched and cascade markAllAsTouched() into the inner card form. Since gio-save-bar already calls form.markAllAsTouched() on an invalid submit, the per-card errors now appear on Save, while a freshly added card stays clean.
- Used Angular 20's control.events + TouchedChangeEvent for the trigger (markAllAsTouched() doesn't emit on statusChanges).

2. Consistent required asterisk
- Passed [required]="true" to the domains gio-form-tags-input so its label shows the asterisk, matching From.

## Additional context
Screenshot 2026-07-23 at 9 37 55 AM

### DIFF
--- a/gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.html
+++ b/gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.html
@@ -76,6 +76,7 @@
         <mat-label>Recipient domains</mat-label>
         <gio-form-tags-input
           formControlName="domains"
+          [required]="true"
           [addOnBlur]="true"
           [placeholder]="group.controls.domains.value.length === 0 ? 'example.com, eu.example.com' : ''"
         ></gio-form-tags-input>
--- a/gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
+++ b/gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
@@ -19,10 +19,12 @@ import {
   Component,
   DestroyRef,
   ErrorHandler,
-  forwardRef,
   inject,
   input,
+  OnInit,
+  Optional,
   output,
+  Self,
   signal,
 } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
@@ -32,14 +34,15 @@ import {
   FormArray,
   FormControl,
   FormGroup,
-  NG_VALIDATORS,
-  NG_VALUE_ACCESSOR,
+  NgControl,
   ReactiveFormsModule,
+  TouchedChangeEvent,
   ValidationErrors,
   Validator,
   ValidatorFn,
   Validators,
 } from '@angular/forms';
+import { filter } from 'rxjs/operators';
 import { MatCardModule } from '@angular/material/card';
 import { MatFormFieldModule } from '@angular/material/form-field';
 import { MatInputModule } from '@angular/material/input';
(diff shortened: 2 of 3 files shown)


gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
```diff
@@ -232,13 +270,13 @@ export class BrandedSendersComponent implements ControlValueAccessor, Validator
   // --- Template actions ---
 
   protected addConfiguration(): void {
+    // Deliberately does not mark the control touched: a freshly added card must start clean. Its required errors
+    // surface only when the user tries to save (the save bar touches the host control -> see ngOnInit cascade).
     this.configurations.push(this.newConfiguration());
```
ytvnr: **Cascade only re-fires on the first `touched` transition**

The touched→cascade wired in `ngOnInit` relies on `TouchedChangeEvent(true)`, but Angular's `markAsTouched` only emits on the `false → true` transition:
```js
const changed = this.touched === false;
if (changed && opts.emitEvent !== false) { this._events.next(new TouchedChangeEvent(true, ...)); }
```
Once the host control is touched it stays touched, so the cascade only runs on the **first** invalid save.

Repro:
1. Save with card A invalid → host `untouched→touched` → errors on A ✅
2. Fix A, add card B (stays clean — intended)
3. Save again (still invalid) → host is already touched → no event → `configurations.markAllAsTouched()` isn't called again → **card B shows no errors**, even though the save is blocked.

That's the same bug this PR fixes, reappearing from the 2nd failed save onwards. Since the form is already in a "post-save" state at that point, we could mark a newly added card touched when the host is already touched:
```ts
protected addConfiguration(): void {
  const card = this.newConfiguration();
  this.configurations.push(card);
  if (this.ngControl?.control?.touched) {
    card.markAllAsTouched();
  }
}
```
Keeps "clean before the first save", stays consistent after. WDYT?

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
```diff
@@ -232,13 +270,13 @@ export class BrandedSendersComponent implements ControlValueAccessor, Validator
...
-    this._onTouched();
   }
 
   protected removeConfiguration(index: number): void {
     this.configurations.removeAt(index);
```
ytvnr, line 287: With both `this._onTouched()` calls removed, `_onTouched` is still registered via `registerOnTouched` but never invoked, so the CVA no longer propagates `touched` to the host on its own (only the save bar does). Consistent with the new design — just flagging that it's effectively dead boilerplate now; maybe worth a one-line comment.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.html
```diff
@@ -76,6 +76,7 @@ <h3 class="branded-senders__title">Branded notification email</h3>
         <mat-label>Recipient domains</mat-label>
         <gio-form-tags-input
           formControlName="domains"
+          [required]="true"
```
ytvnr, line 79: The `domains` FormControl already carries `Validators.required`. Can you confirm `[required]` on `gio-form-tags-input` is purely visual (asterisk + aria) and doesn't register a second `required` validator? Harmless if it's just a duplicate — just ruling out a conflict.

ytvnr (approved):
Nice change. Moving from the `NG_VALUE_ACCESSOR`/`NG_VALIDATORS` providers to the nested-CVA pattern (inject `NgControl`, self-assign `valueAccessor`, `addValidators`) is clean and well-commented, and cleaning up the manually added validator on destroy — with a dedicated test for the license-gated teardown — is exactly right. I checked the underlying mechanism: `gio-save-bar` does call `control.markAllAsTouched()` on invalid submit, and that emits a `TouchedChangeEvent` on the host control, so the approach is sound.

One edge case worth discussing (inline on `addConfiguration`): the cascade only re-fires on the first `touched` transition, so a card added after a first failed save won't surface its errors on later saves. Plus two minor notes inline.

Not blocking — approving.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
paulatulis: TY for flagging! Addressed.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.html
paulatulis: Addressed.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
paulatulis: Addressed.

PR #18726, opened 2026-07-23
fix: align branded-sender From validation with the SMTP send-path parser (backend + console)

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14778

## Description

`SenderAddressValidator` used a hand-rolled regex that disagreed with the `jakarta.mail` parser the SMTP send-path (`EmailServiceImpl → new InternetAddress(from)`) actually uses. It wrongly rejected single-label hosts (`noreply@localhost`, the standard docker/k8s SMTP shape), so an install with such a sender could no longer save settings — and the console's own client-side check had the same false-reject.

**Backend** — validate the whole `From` value with `new InternetAddress(value, true).validate()`, the same parser the send-path uses, so a `From` accepted at save is one delivery can actually send. Closes the gap in both directions:
- False rejects — single-label hosts and apostrophes in the local part now save.
- False accepts — `a..b@x.com` and a broken display name (`Acme, Inc <noreply@acme.com>`, whose unquoted comma parses as two addresses at send time) are now rejected at save instead of failing silently at send.

**Console** — relax the branded-sender `From` check so it stays more permissive than the backend, so single-label hosts are no longer falsely blocked in the browser.

## Changes (6 files)

- `SenderAddressValidator` — parses the whole value against the send-path parser; distinct violation message for the multi-address case (`a@x, b@x` — both valid, but there's more than one).
- `EmailAddresses` — javadoc corrected: the validator no longer routes through this helper (still used by send-time `BrandedSenders` matching).
- `pom.xml` — adds `jakarta.mail:jakarta.mail-api` (version-managed); the `InternetAddress` parser lives in the API artifact, no SMTP transport provider pulled into this module.
- `SenderAddressValidatorTest` (new) — 20 tests: accept/reject table, single-label hosts, broken display names, CR/LF rejection, and end-to-end message wiring through a real `Validator`.
- `branded-senders.component.ts` / `.spec.ts` — lenient client-side `From` check + test.

## Verification

Backend tests 20/20; existing `BrandedSenderValidationTest` (22) + `EmailAddressesTest` (13) still green — no regressions. Verified end-to-end in the console UI (see recording below).

## Additional context

https://github.com/user-attachments/assets/a437bf45-5c30-464e-ab96-3f1dd080c42e

### DIFF
--- a/gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
+++ b/gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
@@ -61,14 +61,20 @@ const SUBJECT_MAX_LENGTH = 255;
 // Mirror it here as an aggregate guard so many individually-valid configurations can't silently overflow it at save time.
 const MAX_SERIALIZED_LENGTH = 4000;
 
-// Client-side format checks so the admin gets a helpful inline error (RFC 1035 host name, and a bare address or a
-// "Name <addr>" sender). These are UX-only and intentionally lenient — the backend only rejects null entries and
-// CR/LF, so this is NOT a mirror of backend validation. The final label allows either an alphabetic TLD or an
-// ACE/punycode IDN TLD (e.g. `xn--p1ai`), which contains digits and hyphens. The `i` flag keeps hostnames
-// case-insensitive (RFC 4343) so e.g. an `XN--P1AI` TLD is accepted the same in a bare domain and in a sender address.
+// Client-side format checks so the admin gets a helpful inline error before saving. The backend
+// (SenderAddressValidator, parsing with the same jakarta.mail parser the SMTP send-path uses) is the
+// authority; these stay deliberately MORE permissive than it for the shapes that matter, so a single-label
+// host like `user@localhost` (the standard docker/k8s SMTP config) or an IDN sender is not locked out
+// client-side even though it saves fine. (A few exotic RFC forms — e.g. a quoted local part containing `@` —
+// are still rejected here though the backend accepts them; negligible for a sender address.)
+//
+// DOMAIN_PATTERN (recipient domains) still requires a real dot-separated host name: the final label allows an
+// alphabetic TLD or an ACE/punycode IDN TLD (e.g. `xn--p1ai`). The `i` flag keeps hostnames case-insensitive
+// (RFC 4343). EMAIL_PATTERN (sender address) only checks that a local and a domain part surround a single `@`
+// with no whitespace — enough to catch an obvious typo (a missing `@` or domain) without mirroring the backend.
 const TLD = '(?:[a-zA-Z]{2,}|xn--[a-zA-Z0-9-]{1,59})';
 const DOMAIN_PATTERN = new RegExp(String.raw`^(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+${TLD}$`, 'i');
-const EMAIL_PATTERN = new RegExp(String.raw`^[A-Za-z0-9._%+-]+@(?:[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?\.)+${TLD}$`, 'i');
+const EMAIL_PATTERN = /^[^\s@]+@[^\s@]+$/;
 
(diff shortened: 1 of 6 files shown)


gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
```diff
@@ -61,14 +61,19 @@ const SUBJECT_MAX_LENGTH = 255;
...
+// config), a quoted local part, or an IDN sender would be locked out client-side even though it saves fine.
+//
+// DOMAIN_PATTERN (recipient domains) still requires a real dot-separated host name: the final label allows an
+// alphabetic TLD or an ACE/punycode IDN TLD (e.g. `xn--p1ai`). The `i` flag keeps hostnames case-insensitive
+// (RFC 4343). EMAIL_PATTERN (sender address) only checks that a local and a domain part surround a single `@`
+// with no whitespace — enough to catch an obvious typo (a missing `@` or domain) without mirroring the backend.
 const TLD = '(?:[a-zA-Z]{2,}|xn--[a-zA-Z0-9-]{1,59})';
 const DOMAIN_PATTERN = new RegExp(String.raw`^(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+${TLD}$`, 'i');
-const EMAIL_PATTERN = new RegExp(String.raw`^[A-Za-z0-9._%+-]+@(?:[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?\.)+${TLD}$`, 'i');
+const EMAIL_PATTERN = /^[^\s@]+@[^\s@]+$/;
```
ytvnr, line 77: The comment above states the client stays strictly more permissive than the backend, but there is one exotic case where it does not: a quoted local part containing an `@` (`"a@b"@x.com`) is accepted by the backend yet rejected here, since `EMAIL_PATTERN` requires a single unescaped `@`. Purely theoretical for a sender address, so I would just soften the comment wording rather than change the pattern. Up to you.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/validator/SenderAddressValidatorTest.java
```diff
@@ -0,0 +1,149 @@
...
+        // overrides it (see below).
+        verify(constraintValidatorContext, never()).buildConstraintViolationWithTemplate(any());
+    }
+
+    @ParameterizedTest
+    @ValueSource(
+        strings = {
+            "alice@example.com, bob@example.com", // a plain address list — both valid, but there are two
+            "Alice <alice@example.com>, Bob <bob@example.com>", // comma-joined personal-name list
+            "noreply@localhost\nBcc: attacker@evil.com", // header-injection attempt reads as multiple addresses
```
ytvnr, line 104: This header-injection case is actually caught by the multi-`@` path, not by CR/LF detection. A single-`@` injection (`noreply@host\nSubject: x`) relies entirely on `validate()` to reject the newline. Behaviour stays consistent with the send path since it is the same parser, so no security regression, but a dedicated single-`@` CR/LF case would make the "CR/LF rejection" claim in the description accurate. Worth adding one assertion.

ytvnr (approved):
Nice fix, the root cause diagnosis is spot on: killing the divergence between the two parsers at the source rather than patching the regex is the right call. Test coverage is convincing (end to end through a real Validator, not just the mock). Two small non-blocking notes inline.

gravitee-apim-console-webui/src/shared/components/branded-senders/branded-senders.component.ts
paulatulis: Softened.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/test/java/io/gravitee/rest/api/validator/SenderAddressValidatorTest.java
paulatulis: Added a dedicated single-@ CR/LF case

PR #18727, opened 2026-07-23
fix: seed default subscription form in new cloud environments.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-81

### Description

Cloud environments created after the original default subscription form upgrader ran never received a form, so the next-gen portal editor showed "Subscription form not found".

New cloud environments are now seeded on first registration, and the existing upgrader is rerun (version v2) to backfill environments that are already missing a form. Created forms include validation constraints derived from the default GMD template.

#### Default form creation use case

Extracted CreateDefaultSubscriptionFormUseCase to idempotently create a disabled default form from the classpath template, with schema-derived validation constraints.

#### Cockpit ENVIRONMENT command

EnvironmentCommandHandler seeds the default subscription form when Cockpit registers an environment for the first time, alongside default portal navigation items.

#### Upgrader backfill

DefaultSubscriptionFormUpgrader now delegates to the shared use case and declares version v2 so it runs again on deploy and heals environments without a form.


jaroslawlakomy-grav (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/cockpit/command/handler/EnvironmentCommandHandler.java
```diff
@@ -96,9 +98,10 @@ public Single<EnvironmentReply> handle(EnvironmentCommand command) {
                 accessPointsToCreate
             );
 
-            // Seed default portal navigation items only when Cockpit registers this environment for the first time
+            // Seed defaults only when Cockpit registers this environment for the first time
             if (existingEnvironment == null) {
                 createDefaultPortalNavigationItemsUseCase.execute(environment.getOrganizationId(), environment.getId());
+                createDefaultSubscriptionFormUseCase.execute(environment.getId());
             }
```
jaroslawlakomy-grav, line 105: What will happen if environment was created but  with old upgrader? Is it ok to skip default subscription form there?

kovaren: Yep, to heal the existing environments which are missing the form, the upgrader's version was increased so that it reruns during the next upgrade https://github.com/gravitee-io/gravitee-api-management/pull/18727/changes#diff-14c7715065ca3c44243326d04750c037c658903ce79bb99c302fa8ce95f0a023R52

PR #18742, opened 2026-07-24
fix: surface encryption cause when AM connection save/read fails

podlesrafal:
## Issue

https://gravitee.atlassian.net/browse/AUTHZ-26

## Description

When `api.properties.encryption.secret` has an invalid AES key length (not 16, 24 or 32 bytes), saving or reading the AM connection failed with an opaque `500 "Failed to encrypt/decrypt am service account access token"` — nothing was logged, and the message blamed the token/AM instead of the misconfigured secret.

`AmConnectionServiceImpl.encryptOrPreserve` (save) and `decryptOrNull` (read) now:
- log the cause at `ERROR`, and
- include the cause **and** the offending property name in the message — which the AM config panel already renders verbatim, so the operator sees *why* the connection can't be saved.

No frontend change.

## Additional context

**Reproduce:** set `gravitee_api_properties_encryption_secret` to a 36-byte value (e.g. a UUID with dashes) and Save the AM connection.

- Before: `500 {"message":"Failed to encrypt am service account access token"}` and no log line.
- After:

  > Failed to encrypt the AM service account token — check the property-encryption secret (api.properties.encryption.secret) on the rest-api, it must be 16, 24 or 32 bytes. Cause: Invalid AES key length: 36 bytes

Covered by `AmConnectionServiceImplTest` — a real `DataEncryptor` with a 36-byte secret reproduces both the save-encrypt failure and the read-decrypt-after-rotation failure.

**Deliberately out of scope** (tracked separately): startup fail-fast length validation, `AES/ECB`→AEAD, a control-plane-only secret.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/AmConnectionServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/AmConnectionServiceImpl.java
@@ -115,7 +115,16 @@ public class AmConnectionServiceImpl implements AmConnectionService {
         try {
             return dataEncryptor.encrypt(plaintextToken);
         } catch (GeneralSecurityException e) {
-            throw new TechnicalManagementException("Failed to encrypt am service account access token", e);
+            log.warn(
+                "Failed to encrypt the AM service account token — check the length of api.properties.encryption.secret on the rest-api: {}",
+                e.getMessage(),
+                e
+            );
+            throw new TechnicalManagementException(
+                "Failed to encrypt the AM service account token — check the property-encryption secret (api.properties.encryption.secret) on the rest-api, it must be 16, 24 or 32 bytes. Cause: " +
+                    e.getMessage(),
+                e
+            );
         }
     }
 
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/AmConnectionServiceImpl.java
```diff
@@ -115,7 +115,16 @@ private String encryptOrPreserve(String plaintextToken, AmConnection existing) {
...
+            log.error(
+                "Failed to encrypt the AM service account token — check the length of api.properties.encryption.secret on the rest-api: {}",
+                e.getMessage(),
+                e
+            );
+            throw new TechnicalManagementException(
+                "Failed to encrypt the AM service account token — check the property-encryption secret (api.properties.encryption.secret) on the rest-api, it must be 16, 24 or 32 bytes. Cause: " +
+                    e.getMessage(),
+                e
+            );
```
leleueri, line 127: question: where is catch the TechnicalManagementException? Is it logged somewhere in Higher level? If so we should avoid logging the error here (at lest not in ERROR level but maybe in DEBUG). If the Tech exception is not logged and is just displayed in the UI, then the log is useful. 

Same for the decrypt one

podlesrafal: `TechnicalManagementException` (a subclass of `AbstractManagementException`) is handled by `ManagementExceptionMapper`, which builds the 500 response via `AbstractExceptionMapper.convert()` without logging — I verified the failing call adds 0 lines to the rest-api log. So it isn't logged higher up, it's only surfaced in the UI, which is the case where you said the log is useful.

I did drop both from `ERROR` to `WARN` though, since an invalid-length encryption secret is a misconfiguration rather than a system error.

leleueri (approved):

PR #18744, opened 2026-07-24
fix(console): base runtime-logs reporting banner on analytics enabled

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-82

## Description

Fixes the **Runtime Logs** _"Reporting is disabled"_ banner on v4 proxy/message APIs, which was wrongly shown whenever both logging modes (entrypoint/endpoint) were off even though reporting was enabled. The banner now reflects only `analytics.enabled`.

## Additional context

Now displayed only when reporter settings is disabled :

Capture d’écran 2026-07-24 à 16 02 08

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs/api-runtime-logs.component.ts
+++ b/gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs/api-runtime-logs.component.ts
@@ -50,9 +50,7 @@ export class ApiRuntimeLogsComponent implements OnInit {
   private apiService = inject(ApiV2Service);
   private api$ = this.apiService.get(this.activatedRoute.snapshot.params.apiId).pipe(shareReplay(1));
 
-  isReportingDisabled$ = this.api$.pipe(
-    map((api: ApiV4) => !api.analytics.enabled || (!api.analytics.logging?.mode?.endpoint && !api.analytics.logging?.mode?.entrypoint)),
-  );
+  isReportingDisabled$ = this.api$.pipe(map((api: ApiV4) => !(api.analytics?.enabled ?? true)));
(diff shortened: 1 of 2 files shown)


gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs/api-runtime-logs.component.ts
```diff
@@ -50,9 +50,7 @@ export class ApiRuntimeLogsComponent implements OnInit {
   private apiService = inject(ApiV2Service);
   private api$ = this.apiService.get(this.activatedRoute.snapshot.params.apiId).pipe(shareReplay(1));
 
-  isReportingDisabled$ = this.api$.pipe(
-    map((api: ApiV4) => !api.analytics.enabled || (!api.analytics.logging?.mode?.endpoint && !api.analytics.logging?.mode?.entrypoint)),
-  );
+  isReportingDisabled$ = this.api$.pipe(map((api: ApiV4) => api.analytics?.enabled === false));
```
kovaren: WDYT about keeping the loose comparison like before? 

Otherwise, if ever `analytics` or `analytics.enabled` are undefined or null, the banner will show up again.
```suggestion
  isReportingDisabled$ = this.api$.pipe(map((api: ApiV4) => !api.analytics?.enabled));

kovaren (changes requested):

gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs/api-runtime-logs.component.ts
thomasballivet: Kept the loose ! and added `?? true` so undefined stays consistent with the backend default (`enabled = true`)

kovaren (approved):

PR #18772, opened 2026-07-28
fix(mongodb): add PageRevisionIndexUpgrader for optimized page revisions query performance

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-14337

## Description

Adds a compound MongoDB index on `page_revisions` (`_id.pageId` ascending, `_id.revision` descending) to fix a response-time regression on `GET /management/organizations/{orgId}/environments/{envId}/applications`.

`PageRevisionMongoRepositoryImpl.findLastByPageId()` filters on `_id.pageId`, a sub-field of the composite `_id = {pageId, revision}`. MongoDB cannot use the default compound `_id` index to satisfy a filter on one sub-field, so
every call falls back to a full collection scan (COLLSCAN) across `page_revisions`. This index lets Mongo serve the filter and the sort directly instead of scanning the collection.

### Root cause

`page_revisions` documents use a composite `_id`, not a plain string:

[`PageRevisionPkMongo.java#L32-L33`](https://github.com/gravitee-io/gravitee-api-management/blob/4.11.x/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/internal/model/PageRevisionPkMongo.java#L32-L33)
```java
private String pageId;
private int revision;
```

[`PageRevisionMongo.java#L36-L37`](https://github.com/gravitee-io/gravitee-api-management/blob/4.11.x/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/internal/model/PageRevisionMongo.java#L36-L37)
```java
@Id
private PageRevisionPkMongo id;
```
`@Id` maps this straight to MongoDB's `_id`, so every document's `_id` is `{pageId, revision}` combined — not a single value.

The query that reads this collection filters on only one piece of that composite key:

[`PageRevisionMongoRepositoryImpl.java#L38-L47`](https://github.com/gravitee-io/gravitee-api-management/blob/4.11.x/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/internal/page/revision/PageRevisionMongoRepositoryImpl.java#L38-L47)
```java
public Optional<PageRevisionMongo> findLastByPageId(String pageId) {
    Query query = new Query();
    query.limit(1);
    query.with(Sort.by(Sort.Direction.DESC, "_id.revision"));
    query.addCriteria(where("_id.pageId").is(pageId));

    PageRevisionMongo revision = mongoTemplate.findOne(query, PageRevisionMongo.class);
    return Optional.ofNullable(revision);
}
```
The filter (`addCriteria`) only references `_id.pageId`; `_id.revision` is used solely for sorting, never for filtering.

MongoDB's automatic index on `_id` is built to match on the *whole* `_id` value as a single unit — it doesn't decompose `_id` into independently searchable sub-fields. A filter on `_id.pageId` alone can't use that index, so MongoDB falls back to a full collection scan (COLLSCAN) on every call. 
### Before Fix: 
```gravitee> db.page_revisions.estimatedDocumentCount()
5
gravitee> const bulk = [];
| for (let i = 0; i < 20000; i++) {
|   const pageId = i < 200 ? "hot-page" : "page-" + (i % 2900);
|   const revision = i < 200 ? i + 1 : Math.floor(i / 2900) + 1;
|   bulk.push({
|     insertOne: {
|       document: {
|         _id: { pageId: pageId, revision: revision },
|         name: "rev",
|         content: "seed data " + i
|       }
|     }
|   });
| }
| db.page_revisions.bulkWrite(bulk, { ordered: false });
| db.page_revisions.estimatedDocumentCount()
20005
gravitee> db.page_revisions.dropIndex("pi1")
| 
| db.page_revisions.find({ "_id.pageId": "hot-page" })
|   .sort({ "_id.revision": -1 })
|   .limit(1)
|   .explain("executionStats")
{
  explainVersion: '1',
  queryPlanner: {
    namespace: 'gravitee.page_revisions',
    indexFilterSet: false,
    parsedQuery: { '_id.pageId': { '$eq': 'hot-page' } },
    queryHash: '3D3C4E15',
    planCacheKey: '3D3C4E15',
    maxIndexedOrSolutionsReached: false,
    maxIndexedAndSolutionsReached: false,
    maxScansToExplodeReached: false,
    winningPlan: {
      stage: 'SORT',
      sortPattern: { '_id.revision': -1 },
      memLimit: 104857600,
      limitAmount: 1,
      type: 'simple',
      inputStage: {
        stage: 'COLLSCAN',
        filter: { '_id.pageId': { '$eq': 'hot-page' } },
        direction: 'forward'
      }
    },
    rejectedPlans: []
  },
  executionStats: {
    executionSuccess: true,
    nReturned: 1,
    executionTimeMillis: 5,
    totalKeysExamined: 0,
    totalDocsExamined: 20005,
    executionStages: {
      stage: 'SORT',
      nReturned: 1,
      executionTimeMillisEstimate: 0,
      works: 20008,
      advanced: 1,
      needTime: 20006,
      needYield: 0,
      saveState: 20,
      restoreState: 20,
      isEOF: 1,
      sortPattern: { '_id.revision': -1 },
      memLimit: 104857600,
      limitAmount: 1,
      type: 'simple',
      totalDataSizeSorted: 0,
      usedDisk: false,
      spills: 0,
      inputStage: {
        stage: 'COLLSCAN',
        filter: { '_id.pageId': { '$eq': 'hot-page' } },
        nReturned: 200,
        executionTimeMillisEstimate: 0,
        works: 20006,
        advanced: 200,
        needTime: 19805,
        needYield: 0,
        saveState: 20,
        restoreState: 20,
        isEOF: 1,
        direction: 'forward',
        docsExamined: 20005
      }
    }
  },
  command: {
    find: 'page_revisions',
    filter: { '_id.pageId': 'hot-page' },
    sort: { '_id.revision': -1 },
    limit: 1,
    '$db': 'gravitee'
  },
  serverInfo: {
    host: 'b23430303480',
    port: 27017,
    version: '6.0.8',
    gitVersion: '3d84c0dd4e5d99be0d69003652313e7eaf4cdd74'
  },
  serverParameters: {
    internalQueryFacetBufferSizeBytes: 104857600,
    internalQueryFacetMaxOutputDocSizeBytes: 104857600,
    internalLookupStageIntermediateDocumentMaxSizeBytes: 104857600,
    internalDocumentSourceGroupMaxMemoryBytes: 104857600,
    internalQueryMaxBlockingSortMemoryUsageBytes: 104857600,
    internalQueryProhibitBlockingMergeOnMongoS: 0,
    internalQueryMaxAddToSetBytes: 104857600,
    internalDocumentSourceSetWindowFieldsMaxMemoryBytes: 104857600
  },
  ok: 1
}
```

This PR adds an index built explicitly on `_id.pageId` (+ `_id.revision` for the sort), so MongoDB can serve the filter directly instead of scanning the collection.

### Verification

Reproduced locally against a seeded `page_revisions` collection (20k docs,
one page with 200 revisions) to confirm the query plan change:

**Before** (index dropped to simulate no covering index): `SORT` → `COLLSCAN`,
20,005 docs examined to return 1.

**After** (`pi1r-1` created): `LIMIT` → `FETCH` → `IXSCAN` on `pi1r-1`,
1 doc examined, no blocking sort stage needed since the index already
returns results in the required order.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/upgrade/upgrader/index/pagerevisions/PageRevisionIndexUpgrader.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/upgrade/upgrader/index/pagerevisions/PageRevisionIndexUpgrader.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.mongodb.management.upgrade.upgrader.index.pagerevisions;
+
+import io.gravitee.repository.mongodb.management.upgrade.upgrader.index.Index;
+import io.gravitee.repository.mongodb.management.upgrade.upgrader.index.IndexUpgrader;
+import org.springframework.stereotype.Component;
+
+/**
+ * Compound index on {@code _id.pageId} (ascending) and {@code _id.revision} (descending) to
+ * cover the {@code findLastByPageId} query: filter on {@code _id.pageId} + sort by
+ * {@code _id.revision DESC} + limit 1.
+ *
+ * <p>Without this index MongoDB falls back to a COLLSCAN across the entire {@code page_revisions}
+ * collection for every call, which becomes a severe bottleneck when the endpoint that triggers
+ * these look-ups is called frequently (e.g. {@code GET /applications}).
+ *
+ * @author GraviteeSource Team
+ */
+@Component("PageRevisionsPageRevisionIndexUpgrader")
+public class PageRevisionIndexUpgrader extends IndexUpgrader {
+
+    @Override
+    protected Index buildIndex() {
+        return Index.builder()
+            .collection("page_revisions")
+            .name("pi1r-1")
+            .key("_id.pageId", ascending())
+            .key("_id.revision", descending())
+            .build();
+    }
+}
--- a/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/upgrade/upgrader/index/pagerevisions/PageRevisionIndexUpgrader.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/upgrade/upgrader/index/pagerevisions/PageRevisionIndexUpgrader.java
@@ -19,17 +19,6 @@ import io.gravitee.repository.mongodb.management.upgrade.upgrader.index.Index;
 import io.gravitee.repository.mongodb.management.upgrade.upgrader.index.IndexUpgrader;
 import org.springframework.stereotype.Component;
 
-/**
- * Compound index on {@code _id.pageId} (ascending) and {@code _id.revision} (descending) to
- * cover the {@code findLastByPageId} query: filter on {@code _id.pageId} + sort by
- * {@code _id.revision DESC} + limit 1.
- *
- * <p>Without this index MongoDB falls back to a COLLSCAN across the entire {@code page_revisions}
- * collection for every call, which becomes a severe bottleneck when the endpoint that triggers
- * these look-ups is called frequently (e.g. {@code GET /applications}).
- *
- * @author GraviteeSource Team
- */
 @Component("PageRevisionsPageRevisionIndexUpgrader")
 public class PageRevisionIndexUpgrader extends IndexUpgrader {
 
--- a/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/test/java/io/gravitee/repository/mongodb/management/upgrade/upgrader/index/pagerevisions/PageRevisionIndexUpgraderTest.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/test/java/io/gravitee/repository/mongodb/management/upgrade/upgrader/index/pagerevisions/PageRevisionIndexUpgraderTest.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.mongodb.management.upgrade.upgrader.index.pagerevisions;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.gravitee.repository.mongodb.management.upgrade.upgrader.index.Index;
+import org.bson.Document;
+import org.junit.jupiter.api.Test;
+
+class PageRevisionIndexUpgraderTest {
+
+    @Test
+    void buildIndex_definesCompoundIndexOnPageRevisionsWithExpectedNameAndKeys() {
+        Index index = new PageRevisionIndexUpgrader().buildIndex();
+
+        assertThat(index.getCollection()).isEqualTo("page_revisions");
+        assertThat(index.options().getName()).isEqualTo("pi1r-1");
+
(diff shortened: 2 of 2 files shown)


aqibmohammadkhan (commented):
@ankita-gupta21  - I would like to challenge the changes here

The index the ticket asks for (_id.pageId: 1) already exists — it's pi1, shipped via `PageIdIndexUpgrader` in 4.0.0, not 4.10.0. If the customer sees a COLLSCAN, pi1 is missing in their environment, and we need to know why before merging: **can we get db.page_revisions.getIndexes() and upgrader logs from the Zendesk ticket?** 
Note that if upgraders aren't applying there, this PR's new upgrader won't reach them either.

Separately, the regression the ticket describes scales with the number of applications because the endpoint calls `findLastByPageId()` once per application (via fillContentRevisionId). An index caps per-query cost but doesn't remove the N+1 - that's the actual 4.10→4.11 regression and should be fixed at the service layer (batch fetch or drop revision resolution from the list path).

On the index itself: the compound version only saves an in-memory sort over one page's revisions for a limit-1 query, which pi1 already makes cheap. If we keep it, the pi1 drop needs to be in this PR, not a follow-up - otherwise every install carries two overlapping indexes on that collection.

callaertanthony (commented):
Reviewed the index definition against the query it is meant to serve, and against what already ships on this branch. The code is correct; my concern is whether it addresses the reported symptom.

## Code correctness — good

`PageRevisionMongoRepositoryImpl.findLastByPageId` issues:

```java
query.limit(1);
query.with(Sort.by(Sort.Direction.DESC, "_id.revision"));
query.addCriteria(where("_id.pageId").is(pageId));
```

`{_id.pageId: 1, _id.revision: -1}` serves the equality on the prefix and the sort on the second key, so the query becomes an IXSCAN with limit 1 and no in-memory SORT stage. The definition matches the query precisely, the package/`@Component` naming follows `PageIdIndexUpgrader`, and the test asserts collection, name and both key directions. No objection to the code itself.

## The premise needs confirming before we call this the fix

`PageIdIndexUpgrader` already creates `pi1` on `{_id.pageId: 1}` and has shipped since 4.10.0 — I compared the blob on `4.11.x`, `4.12.x` and `master`, all identical. That single-field index is already enough to keep this query off a collection scan: the equality predicate is what selects the index, and the sort over the few revisions of one page is cheap.

Yet the reported profiler output on a 4.11.x deployment shows the query still doing a full collection scan. So the situation is not "the codebase is missing an index" — it is "the index the codebase already defines is not present on that database". Adding a second index through the same mechanism has no particular reason to land where the first one did not.

## A mechanism that would explain a missing index

`IndexUpgrader.upgrade()` swallows any error and converts it to a `false` return:

```java
.onErrorResume(e -> {
    log.error("Unexpected error while creating index {} on {}", name, collection, e);
    return Mono.just(false);
})
```

In `UpgraderServiceImpl` (checked against gravitee-node 8.1.0, the version `4.11.x` pins — same logic in 9.x):

```java
.takeWhile(upgrader -> !stopUpgrade.get())
...
case FAILED -> {
    log.error("{} update not applied: it {}. Updater returned false…", name, status);
    stopUpgrade.set(true);
}
...
if (upgradeMode() || stopUpgrade.get()) { … System.exit(1); }
```

A single failing upgrader halts the whole remaining chain, and nothing after it is recorded in `upgrades`. Every index upgrader returns `getOrder() == 0`, so their relative position is just bean-definition order — meaning `PageIdIndexUpgrader` may never have run on a database where some earlier order-0 upgrader failed once.

Four cheap diagnostics from an affected database would settle it:

1. `db.page_revisions.getIndexes()` — is `pi1` present?
2. `db.upgrades.find({_id: /PageIdIndexUpgrader/})` — is it recorded as applied?
3. Management API startup logs, grepped for `update not applied` and `Unexpected error while creating index`.
4. Confirm `management.mongodb.prefix` and the connection URI resolve to the same database the upgrader wrote to — `IndexMongoUpgrader.buildCollectionName()` prefixes the collection name, so a prefix or database mismatch reproduces this symptom exactly.

If `pi1` is absent *and* recorded as applied, the bug is in the upgrader chain rather than in the index set, and this PR will not change the outcome.

## The stated cause of the endpoint regression is not visible in the code

The issue attributes the slowdown to the applications listing endpoint calling `findLastByPageId()` once per application. I could not find that path on `4.11.x`:

- `ApplicationServiceImpl` — no references to pages, page revisions or documentation.
- `ApplicationsResource` — same; its `Page` import is `io.gravitee.common.data.domain.Page`, i.e. pagination.
- The only production callers of `findLastByPageId` are `PageServiceImpl.fillContentRevisionId` (reached from `PageService.findById`, a single-page read) and the page create/update write path (`PageRevisionCrudServiceImpl.create`, `PageRevisionServiceImpl.convert`).

The expensive query is real, but the N+1-from-applications chain is not in the code. That does not block this index — it helps regardless — but it does mean merging this should not be reported as the endpoint regression being fixed.

## Minor

- **Redundant index.** `pi1r-1` has `_id.pageId` as its prefix, so `pi1` becomes fully redundant and only costs write amplification on every page save. Suggestion: drop it on `master` via its own upgrader and keep both on the support branches — dropping an index on a support branch is a worse risk than carrying a duplicate.
- **Startup cost.** `upgrade()` calls `create.block()`, so the Management API blocks on the index build. `page_revisions` documents carry full page content, so this can be slow on a large collection. There is a 10s heartbeat log and `pi1` set the precedent, but it is worth a release note.
- Index name `pi1r-1` is unique across the existing index upgraders and consistent with the local convention.

## Summary

Not requesting changes — the code is right. But I would hold the approval until diagnostics 1–3 come back, because the branch already ships an index that should have prevented this scan, and we do not yet know why it did not.

ankita-gupta21:
Supersedes #18772 (index-based fix) — that addressed a symptom, not the actual bottleneck.

ankita-gupta21:
Supersedes #18772 (index-based fix) — that addressed a symptom, not the actual bottleneck.

PR #18774, opened 2026-07-28
fix(gateway): evaluate v4 flow condition lazily after previous flow

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-83


phiz71:
closed in favour of https://github.com/gravitee-io/gravitee-api-management/pull/18951

PR #18776, opened 2026-07-28
FOUND-2-Environment-Scoped-Dictionaries-cleanup-and-bug-fix

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/FOUND-2

## Description

- **Empty state for dictionaries list** — the dictionaries list at /shell/dictionaries previously fell back to a bare "No dictionaries defined for this environment." table message when an environment had zero dictionaries. Added DictionariesEmptyState, a proper zero-state card (why to configure dictionaries, a "how it works" flow diagram, and key benefits) shown in place of the table when the list is empty, consistent with the empty-state pattern already used elsewhere (e.g. API notifications).
- **Dictionary description placeholder** — the name column showed a literal "—" under the dictionary name whenever no description was set, which read as a broken/missing value. Removed the placeholder; when there's no description, only the name renders, vertically centered in the cell instead of being pinned to the top with dead space below it.
- **Removed unused undeploy dictionary code** — undeployEnvironmentDictionary (service function), useUndeployDictionary (mutation hook), and its associated spec were dead code with no callers anywhere in the dictionaries feature; removed them along with their imports.
- **Dictionary detail page permission parity with classic Console** — `dictionaries/:dictionaryId` was gated on `environment-dictionary-r` only, so a user with only create/update/delete access (no read) was redirected away, whereas classic Console gates that page with `anyOf(-c, -r, -u, -d)`. Widened `PermissionPageGuard` to accept an array of permissions and aligned the route guard with classic; verified all other dictionary action buttons (Deploy, Start/Stop, Add/Edit/Delete Property, list row Edit/Delete) already matched classic's permission gating.
Screenshot 2026-07-28 at 11 47 34 PM

- **Extracted the 403-handling into a shared `useForbiddenResourceRedirect` hook + `isForbiddenApiError` helper** so it isn't duplicated per resource — both `DictionariesPage` and `MetadataPage` use it now, and any future environment-scoped resource can adopt it with one line.
- **"Last Deployed" no longer shown for DYNAMIC dictionaries** — Deploy is a MANUAL-only concept (dynamic dictionaries poll/refresh automatically instead), but the list table and detail page both rendered a "Last Deployed" value/timestamp for DYNAMIC rows too. The list table's "Last Deployed" column now always shows "—" for DYNAMIC dictionaries regardless of the raw `deployed_at` value, and the detail page no longer renders the "Last Deployed" stat at all for DYNAMIC type — matching that Deploy is only ever a MANUAL-type action.
- **Pagination for dictionary properties** — the properties table on the dictionary detail page rendered every key/value entry at once with no pagination, for both MANUAL and DYNAMIC dictionaries. Added client-side pagination (page/page-size controls, same page-size options used elsewhere) to `DictionaryPropertiesTable`, so large property sets no longer render as one long unpaginated list.

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
@@ -25,6 +25,8 @@ import { applicationDetailTabElement } from '../config/applicationDetailPages';
 import { NAV_GROUPS } from '../config/navigation';
 import { PLATFORM_ROUTE_CONFIG } from '../config/routes';
 import { ApplicationDetailIndexRedirect, ApplicationDetailLayout } from '../features/applications/components/detail';
+import { useEnvironmentDictionaries } from '../features/dictionaries/hooks/useEnvironmentDictionaries';
+import { useEnvironmentMetadata } from '../features/metadata/hooks/useEnvironmentMetadata';
 import { SecurityPlanTypesPage } from '../features/security-plan-types/SecurityPlanTypesPage';
 import { ORGANIZATION_USER_ACCESS_PERMISSIONS } from '../features/users/utils/userPermissions';
 import { AccessManagementPage } from '../pages/AccessManagementPage';
@@ -40,6 +42,7 @@ import { retryTransientRequest } from '../shared/api/queryRetry';
 import { ConsoleSettingsProvider } from '../shared/console-settings';
 import { useHasPermission } from '../shared/gamma-modules-sdk';
 import { useEnvironmentPermissions, useEnvironmentPermissionsReady } from '../shared/hooks/useEnvironmentPermissions';
+import { isForbiddenApiError } from '../shared/utils/apiErrors';
 
 const queryClient = new QueryClient({
     defaultOptions: {
@@ -109,8 +112,18 @@ function ModuleLayout() {
     useEnvironmentPermissions();
 
     const permissionsReady = useEnvironmentPermissionsReady();
-    const canReadMetadata = useHasPermission({ anyOf: ['environment-metadata-r'] });
-    const canReadDictionaries = useHasPermission({ anyOf: ['environment-dictionary-r'] });
+
+    // Only probe the resource once the permission map already says "yes" — that's the only case where a
+    // 403 here would disagree with it. Probing unconditionally would hit both list APIs on every platform
+    // page for every user, including those who already correctly lack access.
+    const canReadMetadataPermission = useHasPermission({ anyOf: ['environment-metadata-r'] });
+    const metadataQuery = useEnvironmentMetadata({ enabled: canReadMetadataPermission });
+    const canReadMetadata = canReadMetadataPermission && !isForbiddenApiError(metadataQuery.isError, metadataQuery.error);
+
+    const canReadDictionariesPermission = useHasPermission({ anyOf: ['environment-dictionary-r'] });
+    const dictionariesQuery = useEnvironmentDictionaries({ enabled: canReadDictionariesPermission });
+    const canReadDictionaries = canReadDictionariesPermission && !isForbiddenApiError(dictionariesQuery.isError, dictionariesQuery.error);
+
     const canAccessUsers = useHasPermission({ anyOf: [...ORGANIZATION_USER_ACCESS_PERMISSIONS] });
     const canReadEntrypoints = useHasPermission({ anyOf: ['environment-entrypoint-r', 'organization-entrypoint-r'] });
 
@@ -194,7 +207,15 @@ export function AppRoutes() {
                             <Route
                                 path=":dictionaryId"
                                 element={
-                                    <PermissionPageGuard permission="environment-dictionary-r" unauthorizedTo="../../applications">
+                                    <PermissionPageGuard
+                                        anyOf={[
+                                            'environment-dictionary-c',
+                                            'environment-dictionary-r',
+                                            'environment-dictionary-u',
+                                            'environment-dictionary-d',
+                                        ]}
+                                        unauthorizedTo="../../applications"
+                                    >
                                         <DictionaryDetailPage />
                                     </PermissionPageGuard>
                                 }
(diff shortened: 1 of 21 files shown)


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
```diff
@@ -109,8 +112,15 @@ function ModuleLayout() {
...
-    const canReadDictionaries = useHasPermission({ anyOf: ['environment-dictionary-r'] });
+
+    const canReadMetadataPermission = useHasPermission({ anyOf: ['environment-metadata-r'] });
+    const metadataQuery = useEnvironmentMetadata();
+    const canReadMetadata = canReadMetadataPermission && !isForbiddenApiError(metadataQuery.isError, metadataQuery.error);
+
+    const canReadDictionariesPermission = useHasPermission({ anyOf: ['environment-dictionary-r'] });
+    const dictionariesQuery = useEnvironmentDictionaries();
+    const canReadDictionaries = canReadDictionariesPermission && !isForbiddenApiError(dictionariesQuery.isError, dictionariesQuery.error);
+
```
aqibmohammadkhan, line 126: ModuleLayout always fetches dictionaries & metadata.
Both hooks are enabled: Boolean(env) only, so every platform page (Applications, Users, …) hits list APIs for users who lack those permissions → extra load and noisy 403s.

Prefer gating fetches on the permission flag (e.g. enabled: Boolean(env) && canRead…Permission), or only treating a 403 as authoritative after the permission map already said “yes”.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useForbiddenResourceRedirect.ts
aqibmohammadkhan, line 1: In useForbiddenResourceRedirect:

setQueryData removes environment-dictionary-* / environment-metadata-*
Immediately invalidateQueries refetches permissions

In the stale backend map case the PR describes, the refetch can restore the same grants. Nav stays hidden via the resource-query 403 check, but permissionService / PermissionPageGuard can flip back to “allowed”, so deep-linking /dictionaries becomes allow → 403 → redirect → rinse.

For that scenario, drop the invalidate (or invalidate only when you expect the backend to disagree with the 403). Aikido’s style nits on this file are secondary to this logic issue.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/dictionaries/components/DictionaryPropertiesTable.tsx
```diff
@@ -16,15 +16,23 @@
...
+import { TABLE_PAGE_SIZE_OPTIONS } from '../../applications/utils/paginationConstants';
+
+const DEFAULT_PAGE_SIZE = 10;
 
 export interface DictionaryPropertyRow {
     key: string;
     value: string;
 }
 
+function paginateProperties(rows: DictionaryPropertyRow[], page: number, pageSize: number): DictionaryPropertyRow[] {
```
aqibmohammadkhan, line 31: properties pagination doesn’t clamp page.
Deleting the last item on page 2+ (or shrinking the set) can leave an empty page with a non-zero totalCount. Mirror the metadata clamp (and ideally reset page when the dictionary identity changes).

refer MetadataTable.tsx

aqibmohammadkhan (commented):
Tests:

- No page-level test that empty list renders DictionariesEmptyState (component is mocked away).
- No pagination test for page-size change / last-page after deletes.
- No unit test for isForbiddenApiError / the redirect hook in isolation.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
ankita-gupta21: Fixed. useEnvironmentDictionaries/useEnvironmentMetadata now take an { enabled } option; ModuleLayout passes enabled: canReadXPermission so the probe only fires once the permission map already says "yes" — no more list-API calls for users who are already correctly denied.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useForbiddenResourceRedirect.ts
ankita-gupta21: Fixed. Removed the invalidateQueries call entirely; the hook now only patches the cache locally and redirects, so a stale backend grant can't get silently restored by a refetch.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/dictionaries/components/DictionaryPropertiesTable.tsx
ankita-gupta21: Fixed. Added the same totalPages = Math.max(1, Math.ceil(totalCount / pageSize)) + clamp useEffect pattern from MetadataTable.tsx. Also added key={dictionary.id} on the caller so pagination resets when navigating between dictionaries (the "ideally" part of the comment).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/utils/apiErrors.ts
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ApimApiError } from '../api/apimClient';
+
+export function isForbiddenApiError(isError: boolean, error: unknown): boolean {
+    return isError && error instanceof ApimApiError && error.status === 403;
```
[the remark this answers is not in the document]
ankita-gupta21: ignore: Style-only. A flat && chain is clearer than four statements for a single trivial boolean check — not worth the extra verbosity.

aqibmohammadkhan (approved):

PR #18782, opened 2026-07-28
fix(membership): guard transfer-ownership against a null new-owner member id

ThibaudAV:
## Description

`MembershipServiceImpl.transferOwnership` threw a 500 NPE on the "new owner == current owner"
short-circuit when the new primary owner was identified by `userReference` only (a not-yet-registered
directory user), so `MembershipMember.getMemberId()` was `null` — a documented, valid case
(`newPrimaryOwnerId` "Can be null if userReference is defined").

Fix: use `Objects.equals(...)` for the comparison so a `userReference`-only transfer falls through to
the reference-based user resolution instead of NPE-ing. `previousPrimaryOwner.getMemberId()` is always
non-null for an existing primary owner, so the semantics of the short-circuit are unchanged. Adds a
regression test covering the reference-only path.

## Additional context

- Repro: transfer ownership of a cluster (or API) to a directory user that has no internal id yet
  (identified by `userReference`) → 500 `Cannot invoke "String.equals(Object)" because ... getMemberId() is null`.
- Long-standing latent bug (present at least since ~2026-04 on this history, likely older) — not
  introduced by recent work; more reachable now that the ESM transfer UI offers a user-search flow.
- `MembershipService_TransferOwnershipTest.shouldNotNpeWhenTransferringOwnershipToUserIdentifiedByReferenceOnly`
  asserts the transfer now reaches reference resolution (`UserNotFoundException`) instead of an NPE.


ThibaudAV:
Superseded by #18792 — same changes, cleanly rebased on master. Closing and deleting this branch.

PR #18788, opened 2026-07-28
feat(repository): persist Gamma dashboards in a dedicated store

SamHardi:
## Summary

Gamma has **no custom dashboards today**. The observability library ships read-only *template* dashboards — `DashboardListPage` on `gamma-lib-observability` `main` takes `templates: readonly DashboardTemplate[]` and renders nothing else — so a user cannot create, edit or save a dashboard of their own. What exists on `main` is the seam: a `DashboardPersistence` port and its provider, with no implementation behind them.

This PR adds the store that will back that port, so the feature can ship with server-side persistence from the start: a `GammaDashboardRepository` SPI with Mongo and JDBC implementations, validated by the shared repository TCK.

Building the backend first is what keeps this cheap. There is no existing user data, so there is nothing to migrate and no interim client-side format to be compatible with — the alternative would be to ship the UI against browser storage, then pay for a migration and a coexistence window later. It also means the "dashboard identity filters" work can land with a persistence model instead of without one.

**No API is exposed by this PR.** It is the first of three increments under [OBS-13](https://gravitee.atlassian.net/browse/OBS-13); the endpoints follow in [OBS-15](https://gravitee.atlassian.net/browse/OBS-15) (read) and [OBS-16](https://gravitee.atlassian.net/browse/OBS-16) (write).

Jira: [OBS-14](https://gravitee.atlassian.net/browse/OBS-14)

## Changes

- **SPI** — `GammaDashboardRepository extends CrudRepository<GammaDashboard, String>` plus `findByEnvironmentId`, `findByIdAndEnvironmentId` and `deleteByEnvironmentId`, and the `GammaDashboard` model with its nested `Filter` and `TimeRange` statics.
- **Mongo** — `GammaDashboardMongo` document on a new `gamma_dashboards` collection (honouring the configured prefix), internal Spring Data repository, `MongoGammaDashboardRepository` facade, `GraviteeMapper` MapStruct methods both ways, and an `EnvironmentIdIndexUpgrader` creating index `ei1` on `environmentId` (+ its unit test).
- **JDBC** — `JdbcGammaDashboardRepository` declares `filters` and `time_range` as converted `NCLOB` columns through `JdbcObjectMapper`'s serializer/deserializer `addColumn` overload (the idiom `JdbcPortalPageContentRepository` already uses), and `widgets` as a plain `String` column since it is opaque JSON on the model. That leaves the ORM to build the insert/update SQL and the row mapper, so there is no hand-rolled SQL, no custom `PreparedStatementCreator` and no `findById` override — unlike `JdbcCustomDashboardRepository`, which predates that overload and hand-rolls all of it. Liquibase changeset `v4_13_0/09_create_gamma_dashboards_table.yml` with an index on `environment_id`, `master.yml` include, and `gamma_dashboards` added to the test truncate list.
- **Environment deletion cascade** — `DeleteEnvironmentCommandHandler` now calls `gammaDashboardRepository.deleteByEnvironmentId(...)` alongside the existing `customDashboardRepository` call, with the matching assertion in `DeleteEnvironmentCommandHandlerTest`. Without it, deleting an environment would leave its dashboards orphaned in the collection.
- **TCK** — `GammaDashboardRepositoryTest` and its fixture, wired into `AbstractManagementRepositoryTest` and `RepositoryTestSuite`.

### Design decisions worth knowing

- **A new collection, not `custom_dashboards`.** That collection is already read *unfiltered* by two products — APIM Console v2 (`DashboardsResource`) and the Developer Portal (`PortalAnalyticsDashboardsResource`), both via `ListDashboardsUseCase.execute(new Input(environmentId))`. Adding a discriminator would make leaking the default behaviour. `CoreRulesTest` also forbids the Gamma core from importing `io.gravitee.apim.core.dashboard.model.Dashboard`, so Gamma needs its own model regardless.
- **`widgets` is stored as opaque JSON.** Typing 6 widget variants × ~40 optional fields in Java would go stale on every frontend release and force a backend release per new widget option. It also neutralises the dotted-map-key hazard: `colorByGroup` is keyed by hostnames and API names, and no `MappingMongoConverter#setMapKeyDotReplacement` is configured in this repo. `filters` and `timeRange` stay structured.
- **`filters` and `timeRange` stay structured, but their shape mirrors the frontend's `FilterCondition` exactly** — `{ field, label, operator, value }` with lowercase operators, and `timeRange.type` lowercase. The same type also travels verbatim inside the opaque `widgets` payload, so any divergence would mean storing one type in two shapes and maintaining a mapping table between them. `label` is persisted because nothing derives it: the resolve hook rehydrates `valueLabels` from the API, never `label`. The one deliberate difference is `editable`, which replaces the frontend's optional `locked` tri-state with a required positive boolean.
- **Why structured rather than opaque like `widgets`.** What justifies opacity for widgets is an *open* schema — 6 variants × ~40 optional fields, changing every release, with a dotted-map-key hazard. `FilterCondition` is a closed 5-field shape and `TimeRange` a closed 4-field one, neither with dotted keys. Structuring does not validate values, but it does guarantee shape integrity: a caller has to build `Filter` instances to write, so the column cannot hold arbitrary text.
- **`updated_at`, not `last_modified`** — matches the frontend field and the majority convention here; `custom_dashboards` is the outlier.
- **`findByEnvironmentId` orders by `created_at`, then `id`.** Mongo natural order is not stable — a document can move when an update grows it — so an unordered list would let a user reshuffle their own dashboards just by editing one. The id tie-break matters: the fixtures deliberately share a `createdAt`.
- **Mongo `update()` is a single conditional `findAndReplace` matched on `_id`.** `save()` is an upsert, so a read-then-write pair would silently re-insert a dashboard deleted in between, while JDBC raises on 0 affected rows. Matching on `_id` folds the existence check into the same round trip and keeps the backends aligned; enforcing `version` later is one more criterion on that query.
- **A `version` column is included up front, but nothing manages it yet.** Once dashboards are stored server-side they become a shared artifact that two people can edit at once, so they will need optimistic concurrency control. Adding the column before the table ships costs three lines; adding it afterwards costs a migration changeset, a backfill and coordination with deployed instances. This layer therefore persists the counter **verbatim** — it never initialises, increments or checks it (deliberately no Spring Data `@Version`; the shape follows the existing `Cluster.version`). Incrementing belongs to the update use case in OBS-16, and rejecting a stale value (409) is a separate ticket. `GammaDashboardRepositoryTest#shouldPersistVersionVerbatimOnUpdate` pins that contract so the behaviour cannot drift silently.
- **Model named `GammaDashboard`** because `io.gravitee.repository.management.model.Dashboard` already exists (legacy analytics dashboard).

## Test plan

- [ ] `mvn -f gravitee-apim-repository/pom.xml install -DskipTests -pl gravitee-apim-repository-api,gravitee-apim-repository-test -am` (the TCK ships as a test-jar; without the install, the Mongo/JDBC modules unpack a stale one).
- [ ] Mongo: `mvn -f gravitee-apim-repository/pom.xml test -pl gravitee-apim-repository-mongodb -Dtest=GammaDashboardRepositoryTest -DfailIfNoSpecifiedTests=false` — 22 tests, runs both with and without encryption.
- [ ] Every JDBC dialect: same command on `gravitee-apim-repository-jdbc`, repeated with `-Ddefault-database.jdbcType=` `postgresql` (default), `mariadb`, `mysql`, `sqlserver`.
- [ ] Index upgrader: `-Dtest=EnvironmentIdIndexUpgraderTest` on the mongodb module.
- [ ] Regression on the neighbouring stores: `-Dtest='CustomDashboardRepositoryTest,DashboardRepositoryTest'` on both mongodb and jdbc — the Liquibase changeset and the truncate list are shared.
- [ ] Environment cascade: `mvn -f gravitee-apim-rest-api/pom.xml test -pl gravitee-apim-rest-api-service -Dtest=DeleteEnvironmentCommandHandlerTest`.

Edge cases covered by the TCK: find by environment / other environment / unknown environment, `findById` hit and miss, create, update, update of an unknown id, update of `null`, delete, `deleteByEnvironmentId`; filters round-tripping with `editable` and with the empty `value` "all values" placeholder; `timeRange` in both `relative` and `absolute` form; `version` round-tripping when set, staying `null` when absent, and being stored verbatim across an update (neither auto-initialised nor auto-incremented); a stable list order pinned including the id tie-break; an update of a dashboard deleted beforehand raising rather than resurrecting it; and a widget payload carrying a **dotted map key** (`colorByGroup: {"api.example.com": "#fff"}`) asserted to round-trip byte-for-byte.

The fixtures use payloads the product actually emits — `field` / `label` / lowercase operator / `value`, and real `RelativePeriod` tokens such as `7d` — modelled on the in-repo literal in `gravitee-gamma-module-apim/.../templates/shared.ts`. An earlier revision asserted on an invented shape, which proved serialization rather than compatibility.

`findByIdAndEnvironmentId` is covered both ways: it returns the dashboard for its own environment, and returns empty for an id that exists under a different one (`gd-4` is findable by `findById` and invisible through the scoped lookup).

All of the above was run locally and is green (22/22 on each of the five backends, plus `DeleteEnvironmentCommandHandlerTest`).

## Reviewer notes

- **`liquibase/master.yml`** — the include is appended at the tail, which is order-sensitive and a known conflict point: this branch has already been renumbered once, from `08_` to `09_`, after `08_backfill_cluster_lifecycle_state.yml` landed on master. If another changeset reaches `v4_13_0` before this merges, the ordinal and the tail position both need re-checking.
- **`MongoGammaDashboardRepository.update` reads then saves**, so the existence check and the write are two round trips, whereas the JDBC side is a single statement guarded by the affected-row count. Harmless while `version` is only carried, but there is a comment on the method recording that enforcement must turn this into a single conditional `findAndModify` matching on the expected version — otherwise the gap reopens the very lost-update window the locking is meant to close.
- **The Mongo index upgrader is mandatory, not cosmetic.** The TCK container runs with `notablescan=true` and `MongoTestRepositoryInitializer` executes every `IndexMongoUpgrader` bean, so without it `findByEnvironmentId` fails with a confusing error. Note the explicit bean name `@Component("GammaDashboardsEnvironmentIdIndexUpgrader")`: about thirty sibling classes are also called `EnvironmentIdIndexUpgrader`, and the Spring default bean name would collide.
- **No `NoOpGammaDashboardRepository`.** `NoOpCustomDashboardRepository` exists but is not registered in `NoOpManagementRepositoryConfiguration` — dead code, deliberately not replicated.
- **`findByIdAndEnvironmentId` is on the SPI from the start**, and is what the read endpoints should use rather than `findById` plus a caller-side ownership check. It makes cross-environment isolation a property of the query instead of something each call site has to remember, which is the difference between a 404 that is structural and one that depends on review catching a missing check.
- **The JDBC JSON mapper ignores unknown properties.** `filters` and `time_range` are stored as JSON, so a row written by a newer version must stay readable by an older one; otherwise a rollback or a rolling upgrade turns one unreadable row into a failure of the entire environment listing, since `findByEnvironmentId` maps every row through it.
- **Required-field behaviour differs between the two backends** and is documented on the SPI: JDBC declares `title` / `created_by` / `created_at` / `updated_at` `NOT NULL` and rejects the write, Mongo accepts the document. Callers must validate rather than rely on the store, or the same input succeeds on one deployment and fails on another.
- **No write-time validation of filter names** against the `StaticFilters` catalogue: a dashboard is declarative config, the analytics query path already validates and 400s at read time, and coupling writes to the `FilterRegistry` would make dashboards non-importable across environments.

---

<details>
<summary><b>Verification report</b> — expected / observed / status for all 20 checks run locally (click to expand)</summary>

Everything below was executed locally against real engines — no mocks, no stubs at the persistence boundary. The shared TCK runs on Testcontainers-backed MongoDB and on four JDBC engines; the runtime checks run against the local Docker stack with the Management API image rebuilt from this branch.

### Automated — shared repository TCK

The TCK is an integration suite, not a unit suite: each run starts a real database container, applies the real Liquibase changelog or the real Mongo index upgraders, and exercises the SPI through the actual driver.

| # | Check | Expected | Observed | Status |
|---|-------|----------|----------|--------|
| 1 | `GammaDashboardRepositoryTest` on MongoDB 6.0.26 (replica set, `notablescan=true`) | 22 tests pass, both with and without field-level encryption | 22/22 pass in both surefire executions | PASS |
| 2 | Same suite on PostgreSQL | 22 tests pass | 22/22 pass | PASS |
| 3 | Same suite on MariaDB | 22 tests pass | 22/22 pass | PASS |
| 4 | Same suite on MySQL | 22 tests pass | 22/22 pass | PASS |
| 5 | Same suite on SQL Server | 22 tests pass | 22/22 pass | PASS |
| 6 | `filters` round-trip, including `editable` and the empty-`values` "all values" placeholder | Structured filters survive the round trip on every backend | Asserted field by field; no loss | PASS |
| 7 | `timeRange` round-trip in both forms | `RELATIVE` keeps `period` with null `from`/`to`; `ABSOLUTE` keeps epoch-millis `from`/`to` with null `period` | Both asserted explicitly | PASS |
| 8 | Widget payload with a **dotted map key** (`colorByGroup: {"api.example.com": "#fff"}`) | Stored and returned byte-for-byte; Mongo must not reject the dotted key | Exact string equality asserted; no rejection — the opaque `String` column is what makes this safe | PASS |
| 9 | `version` is persisted verbatim | Round-trips when set, stays `null` when absent, and an update that does not touch it leaves it unchanged | `shouldPersistVersionVerbatimOnUpdate` confirms no auto-initialisation and no auto-increment leak into this layer | PASS |
| 9b | Stable list order | `findByEnvironmentId` returns a deterministic sequence on every backend, including when `created_at` ties | `shouldReturnEnvironmentDashboardsInStableOrder` pins `gd-1, gd-2, gd-3`; the three fixtures share a `createdAt`, so the id tie-break is exercised | PASS |
| 9c | An update must not resurrect a deleted dashboard | Raises on both backends; the row stays deleted | `shouldNotResurrectADashboardDeletedBeforeTheUpdate` green on Mongo and all four dialects. Would have passed on JDBC and failed on Mongo before the `findAndReplace` change | PASS |
| 10 | Cross-environment isolation via `findByIdAndEnvironmentId` | A dashboard belonging to another environment is invisible through the scoped lookup even though `findById` still finds it | `gd-4` found by `findById`, empty via `findByIdAndEnvironmentId("gd-4", "DEFAULT")` | PASS |
| 11 | `EnvironmentIdIndexUpgraderTest` | `buildIndex()` targets `gamma_dashboards`, index name `ei1`, single ascending key on `environmentId` | Matches | PASS |
| 12 | `DeleteEnvironmentCommandHandlerTest` | Deleting an environment cascades to Gamma dashboards | 9/9 pass, including `verify(gammaDashboardRepository).deleteByEnvironmentId(ENV_ID)` | PASS |
| 13 | No regression on the neighbouring stores sharing the changelog and the truncate list | `CustomDashboardRepositoryTest` and `DashboardRepositoryTest` stay green on Mongo and JDBC | 23/23 pass on each | PASS |

### Runtime — local Docker stack, Management API rebuilt from this branch

The distribution was rebuilt from this branch, the `local/apim-management-api:local` image regenerated, and the container force-recreated against the existing MongoDB volume. This is the only check that exercises the **production** upgrader wiring: the TCK deliberately excludes `MongoUpgraderConfiguration` and runs upgraders through test-only wiring, so bean discovery in the real application is otherwise unproven.

| # | Check | Expected | Observed | Status |
|---|-------|----------|----------|--------|
| 14 | Baseline before deploy | `gamma_dashboards` does not exist | Collections matching `/dashboard/i`: `['dashboards', 'custom_dashboards']`; `getIndexes()` returns `ns does not exist` | PASS |
| 15 | Management API boots with the new repository code | Clean startup | `started in 6242 ms`, JVM 21.0.11, version 4.13.0-SNAPSHOT | PASS |
| 16 | Collection created on first boot | `gamma_dashboards` present | `['dashboards', 'gamma_dashboards', 'custom_dashboards']` | PASS |
| 17 | **Mongo index created by the production upgrader path** | `ei1` on `environmentId` | `[{ v: 2, key: { _id: 1 }, name: '_id_' }, { v: 2, key: { environmentId: 1 }, name: 'ei1' }]` | PASS |
| 18 | No bean-name collision from the index upgrader | No `ConflictingBeanDefinition` / `BeanDefinitionOverrideException` | None. Confirms the explicit `@Component("GammaDashboardsEnvironmentIdIndexUpgrader")` is doing its job — production registers upgraders **by bean name** via `context.getBeansOfType(Upgrader.class)`, and ~30 sibling classes share the simple name `EnvironmentIdIndexUpgrader` | PASS |
| 19 | Gamma still loads | Gamma endpoints reachable | `GET /gamma/ui/bootstrap` → `200` | PASS |
| 20 | No Gamma-related error in the startup logs | No exception mentioning the new types | None found | PASS |

### Not covered, and why

| Area | Why it is not covered here | Where it lands |
|------|---------------------------|----------------|
| HTTP-level CRUD (create, read, update, delete over the wire) | Structural: this increment exposes no API. There is nothing to call. | OBS-15 and OBS-16 — their DoD is curl CRUD, 404 across environments, 400 at 51 widgets |
| Liquibase applied to an **already-migrated** JDBC database | The TCK rebuilds the schema from scratch on every run, and the local Docker stack is MongoDB-backed, so Liquibase is not exercised at runtime here. Risk is low by construction: the change adds a new changelog at the tail of `master.yml` and modifies no released changeset, so the usual failure mode — an invalidated checksum on an existing changeset — cannot occur. | CI runs the JDBC suites on all four dialects |
| Concurrent-edit behaviour of `version` | Deliberate: this layer only carries the counter. Enforcement does not exist yet, so there is nothing to test. | OBS-16 for the increment, a dedicated ticket for 409 on a stale version |

One unrelated error appears in the startup logs and is worth naming so it is not mistaken for a regression: `BeanCreationException` on `indexWriter` from `SearchEngineConfiguration`, caused by a stale Lucene `write.lock` in the persisted `data/` volume after the container was force-recreated. It is environmental, touches the search index rather than persistence, and is independent of this change — the API still served requests and the Mongo upgraders still ran.

</details>

[OBS-13]: https://gravitee.atlassian.net/browse/OBS-13?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/GammaDashboard.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/GammaDashboard.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.management.model;
+
+import java.util.Date;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+
+/**
+ * Dashboard owned by the Gamma observability module.
+ *
+ * <p>Distinct from {@link Dashboard} (legacy analytics dashboard) and from {@link CustomDashboard}, whose collection is
+ * read unfiltered by the APIM Console and the Developer Portal.
+ *
+ * <p>{@code widgets} is deliberately stored as opaque JSON: the widget shape belongs to the frontend and would
+ * otherwise force a backend release per new widget option. It also sidesteps Mongo's rejection of dotted map keys,
+ * which widget colour maps use (they are keyed by host and API names).
+ *
+ * @author GraviteeSource Team
+ */
+@Builder(toBuilder = true)
+@NoArgsConstructor
+@AllArgsConstructor
+@Data
+@EqualsAndHashCode(of = "id")
+public class GammaDashboard {
+
+    private String id;
+    private String environmentId;
+    private String title;
+    private String description;
+    private List<Filter> filters;
+    private TimeRange timeRange;
+
+    /** Raw, opaque widget JSON. Never parsed by the backend. */
+    private String widgets;
+
+    private String createdBy;
+    private Date createdAt;
+    private Date updatedAt;
+
+    /**
+     * Optimistic concurrency counter. This layer only persists it verbatim — it never initialises, increments or
+     * checks it. Bumping it belongs to the update use case, and rejecting a stale value is a separate concern.
+     */
+    private Integer version;
+
+    /**
+     * Dashboard-definition filter, applied to every widget of the dashboard.
+     *
+     * <p>Field names and casing deliberately mirror the frontend's {@code FilterCondition}, which is the only producer
+     * and consumer of these values — the same type also travels verbatim inside the opaque {@code widgets} payload, so
+     * any divergence here would mean storing one type in two shapes and maintaining a mapping table between them.
+     */
+    @Builder(toBuilder = true)
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @Data
+    public static class Filter {
+
+        private String field;
+
+        /**
+         * Display name of the field, e.g. {@code API Type}. Persisted because nothing derives it: the frontend's
+         * resolve hook rehydrates {@code valueLabels} from the API, never {@code label}.
+         */
+        private String label;
+
+        /** Lowercase wire operator, matching the frontend vocabulary — {@code eq}, {@code in}, {@code not_in}. */
+        private String operator;
+
+        /** An empty list is the "all values" placeholder and carries no query constraint. */
+        private List<String> value;
+
+        /**
+         * {@code true} when the viewer may change the value; {@code false} locks it. The one deliberate addition to
+         * the frontend shape: it replaces the optional {@code locked} tri-state with a required positive boolean.
+         */
+        private boolean editable;
+    }
+
+    /**
+     * Time window a dashboard opens on. {@code period} is set for {@code relative}, {@code from} / {@code to} (epoch
+     * millis) for {@code absolute}.
+     */
+    @Builder(toBuilder = true)
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @Data
+    public static class TimeRange {
+
+        /** {@code relative} or {@code absolute}, lowercase to match the frontend's discriminated union. */
+        private String type;
+
+        /** Relative period token owned by the frontend, e.g. {@code 7d}, {@code 1h}, {@code currentMonth}. */
+        private String period;
+
+        private Long from;
+        private Long to;
+    }
+}
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcGammaDashboardRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcGammaDashboardRepository.java
@@ -1,0 +1,175 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.jdbc.management;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.gravitee.repository.exceptions.TechnicalException;
+import io.gravitee.repository.jdbc.orm.JdbcObjectMapper;
+import io.gravitee.repository.management.api.GammaDashboardRepository;
+import io.gravitee.repository.management.model.GammaDashboard;
+import java.sql.Types;
+import java.util.Date;
+import java.util.List;
+import java.util.Optional;
+import lombok.CustomLog;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Repository;
+
+/**
+ * @author GraviteeSource Team
+ */
+@CustomLog
+@Repository
+public class JdbcGammaDashboardRepository extends JdbcAbstractCrudRepository<GammaDashboard, String> implements GammaDashboardRepository {
+
+    /**
+     * Unknown properties are ignored on purpose, but this only picks the lesser of two silent failures — it does not
+     * make the store forward-compatible.
+     *
+     * <p>{@link io.gravitee.repository.jdbc.orm.JdbcObjectMapper#setFromResultSet} catches per column and only logs, so
+     * a deserialization failure never fails the query: the row still maps and the field is left {@code null}. With the
+     * feature enabled, a row written by a newer node would therefore come back with {@code filters} entirely
+     * {@code null}, and the next {@code update()} would write that null back. Ignoring the unknown field instead loses
+     * only that field.
+     *
+     * <p>Both paths lose data silently on a rollback or a rolling upgrade. Closing the hole properly means either
+     * storing these columns opaquely, like {@code widgets}, or enforcing {@code version} before the first write
+     * endpoint ships.
+     */
+    private static final ObjectMapper JSON_MAPPER = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+
+    JdbcGammaDashboardRepository(@Value("${management.jdbc.prefix:}") String tablePrefix) {
+        super(tablePrefix, "gamma_dashboards");
+    }
+
+    /**
+     * {@code filters} is declared last on purpose: its element type cannot be expressed as a {@code Class} literal, so
+     * {@code List.class} makes that call an unchecked invocation, which would erase the generics of every builder call
+     * chained after it.
+     */
+    @Override
+    @SuppressWarnings("unchecked")
+    protected JdbcObjectMapper<GammaDashboard> buildOrm() {
+        return JdbcObjectMapper.builder(GammaDashboard.class, this.tableName, "id")
+            .addColumn("id", Types.NVARCHAR, String.class)
+            .addColumn("environment_id", Types.NVARCHAR, String.class)
+            .addColumn("title", Types.NVARCHAR, String.class)
+            .addColumn("description", Types.NVARCHAR, String.class)
+            .addColumn("created_by", Types.NVARCHAR, String.class)
+            .addColumn("created_at", Types.TIMESTAMP, Date.class)
+            .addColumn("updated_at", Types.TIMESTAMP, Date.class)
+            .addColumn("version", Types.INTEGER, Integer.class)
+            .addColumn(
+                "time_range",
+                Types.NCLOB,
+                GammaDashboard.TimeRange.class,
+                JdbcGammaDashboardRepository::serializeTimeRange,
+                JdbcGammaDashboardRepository::deserializeTimeRange
+            )
+            // Already an opaque JSON String on the model, so it needs no conversion.
+            .addColumn("widgets", Types.NCLOB, String.class)
+            .addColumn(
+                "filters",
+                Types.NCLOB,
+                List.class,
+                JdbcGammaDashboardRepository::serializeFilters,
+                JdbcGammaDashboardRepository::deserializeFilters
+            )
+            .build();
+    }
+
+    @Override
+    protected String getId(GammaDashboard item) {
+        return item.getId();
+    }
+
+    @Override
+    public List<GammaDashboard> findByEnvironmentId(String environmentId) throws TechnicalException {
+        log.debug("JdbcGammaDashboardRepository.findByEnvironmentId({})", environmentId);
+        try {
+            return jdbcTemplate.query(
+                getOrm().getSelectAllSql() + " where environment_id = ? order by created_at, id",
+                getRowMapper(),
+                environmentId
+            );
+        } catch (Exception ex) {
+            throw new TechnicalException("Failed to find gamma dashboards by environment id: " + environmentId, ex);
+        }
+    }
+
+    @Override
+    public Optional<GammaDashboard> findByIdAndEnvironmentId(String id, String environmentId) throws TechnicalException {
+        log.debug("JdbcGammaDashboardRepository.findByIdAndEnvironmentId({}, {})", id, environmentId);
+        try {
+            return jdbcTemplate
+                .query(getOrm().getSelectAllSql() + " where id = ? and environment_id = ?", getRowMapper(), id, environmentId)
+                .stream()
+                .findFirst();
+        } catch (Exception ex) {
+            throw new TechnicalException("Failed to find gamma dashboard by id: " + id + " and environment id: " + environmentId, ex);
+        }
+    }
+
+    @Override
+    public void deleteByEnvironmentId(String environmentId) throws TechnicalException {
+        log.debug("JdbcGammaDashboardRepository.deleteByEnvironmentId({})", environmentId);
+        try {
+            jdbcTemplate.update("delete from " + this.tableName + " where environment_id = ?", environmentId);
+        } catch (Exception ex) {
+            throw new TechnicalException("Failed to delete gamma dashboards by environment id: " + environmentId, ex);
+        }
+    }
+
+    private static String serializeFilters(List<?> filters) {
+        return serialize(filters);
+    }
+
+    private static List<GammaDashboard.Filter> deserializeFilters(String json) {
+        return deserialize(json, new TypeReference<>() {});
+    }
+
+    private static String serializeTimeRange(GammaDashboard.TimeRange timeRange) {
+        return serialize(timeRange);
+    }
+
+    private static GammaDashboard.TimeRange deserializeTimeRange(String json) {
+        return deserialize(json, new TypeReference<>() {});
+    }
+
+    private static String serialize(Object value) {
+        if (value == null) {
+            return null;
+        }
+        try {
+            return JSON_MAPPER.writeValueAsString(value);
+        } catch (Exception e) {
+            throw new IllegalStateException("Failed to serialize gamma dashboard JSON column", e);
+        }
+    }
+
+    private static <T> T deserialize(String json, TypeReference<T> typeRef) {
+        if (json == null || json.isEmpty()) {
+            return null;
+        }
+        try {
+            return JSON_MAPPER.readValue(json, typeRef);
+        } catch (Exception e) {
+            throw new IllegalStateException("Failed to deserialize gamma dashboard JSON column", e);
+        }
+    }
+}
--- a/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoGammaDashboardRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoGammaDashboardRepository.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.mongodb.management;
+
+import io.gravitee.repository.exceptions.TechnicalException;
+import io.gravitee.repository.management.api.GammaDashboardRepository;
+import io.gravitee.repository.management.model.GammaDashboard;
+import io.gravitee.repository.mongodb.management.internal.GammaDashboardMongoRepository;
+import io.gravitee.repository.mongodb.management.mapper.GraviteeMapper;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import lombok.CustomLog;
+import lombok.RequiredArgsConstructor;
+import org.springframework.data.mongodb.core.FindAndReplaceOptions;
+import org.springframework.data.mongodb.core.MongoTemplate;
+import org.springframework.data.mongodb.core.query.Criteria;
+import org.springframework.data.mongodb.core.query.Query;
+import org.springframework.stereotype.Component;
+
+/**
+ * @author GraviteeSource Team
+ */
+@CustomLog
+@Component
+@RequiredArgsConstructor
+public class MongoGammaDashboardRepository implements GammaDashboardRepository {
+
+    private final GammaDashboardMongoRepository internalGammaDashboardRepo;
+    private final GraviteeMapper mapper;
+    private final MongoTemplate mongoTemplate;
+
+    @Override
+    public Set<GammaDashboard> findAll() throws TechnicalException {
+        log.debug("Find all gamma dashboards");
+        var dashboards = internalGammaDashboardRepo.findAll();
+        var res = mapper.mapGammaDashboards(dashboards);
+        log.debug("Find all gamma dashboards - Done");
+        return res;
+    }
+
+    @Override
+    public Optional<GammaDashboard> findById(String id) throws TechnicalException {
+        log.debug("Find gamma dashboard by ID [{}]", id);
+        var dashboard = internalGammaDashboardRepo.findById(id).map(mapper::map);
+        log.debug("Find gamma dashboard by ID [{}] - Done", id);
+        return dashboard;
+    }
+
+    @Override
+    public GammaDashboard create(GammaDashboard gammaDashboard) throws TechnicalException {
+        log.debug("Create gamma dashboard [{}]", gammaDashboard.getTitle());
+        var dashboardMongo = mapper.map(gammaDashboard);
+        var createdDashboardMongo = internalGammaDashboardRepo.insert(dashboardMongo);
+        var res = mapper.map(createdDashboardMongo);
+        log.debug("Create gamma dashboard [{}] - Done", gammaDashboard.getTitle());
+        return res;
+    }
+
+    /**
+     * A single conditional replace rather than a read followed by {@code save()}.
+     *
+     * <p>{@code save()} is an upsert, so with a read-then-write pair a dashboard deleted between the two calls would be
+     * silently re-inserted — while the JDBC implementation, guarded by its affected-row count, raises. Matching on
+     * {@code _id} folds the existence check into the same round trip and keeps the two backends behaving alike.
+     *
+     * <p>When {@code version} is eventually enforced, the guard is one more criterion on this same query rather than a
+     * new round trip.
+     */
+    @Override
+    public GammaDashboard update(GammaDashboard gammaDashboard) throws TechnicalException {
+        if (gammaDashboard == null) {
+            throw new IllegalStateException("Unable to update a null gamma dashboard");
+        }
+
+        try {
+            var replaced = mongoTemplate.findAndReplace(
+                Query.query(Criteria.where("_id").is(gammaDashboard.getId())),
+                mapper.map(gammaDashboard),
+                FindAndReplaceOptions.options().returnNew()
+            );
+
+            if (replaced == null) {
+                throw new IllegalStateException(String.format("No gamma dashboard found with id [%s]", gammaDashboard.getId()));
+            }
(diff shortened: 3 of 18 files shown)


gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcGammaDashboardRepository.java
```diff
@@ -0,0 +1,218 @@
...
+            return null;
+        }
+        try {
+            return JSON_MAPPER.writeValueAsString(value);
+        } catch (Exception e) {
+            throw new IllegalStateException("Failed to serialize JSON", e);
+        }
+    }
+
+    private class GammaDashboardPsc implements PreparedStatementCreator {
```
SamHardi: Acknowledged — skipping for now: `Psc` is the established idiom in this package, not a one-off. `JdbcDictionaryRepository`, `JdbcIdentityProviderRepository` and `JdbcPageRepository` each declare an inner `class Psc`, and `JdbcCustomDashboardRepository` — the sibling this class is deliberately modelled on — uses `CustomDashboardPsc`. Renaming only the new class would make the two diverge on a straight diff, and `.agent-rules/general.md` asks to "avoid cosmetic renames in feature PRs". Worth doing across all five in a dedicated refactor if the team wants it.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcGammaDashboardRepository.java
```diff
@@ -0,0 +1,218 @@
...
+            .build();
+    }
+
+    @Override
+    protected String getId(GammaDashboard item) {
+        return item.getId();
+    }
+
+    @Override
+    protected RowMapper<GammaDashboard> getRowMapper() {
```
SamHardi: Acknowledged — skipping for now: the inline `getRowMapper()` lambda is the shape used by every JDBC repository in this package (six of them declare `protected RowMapper<...> getRowMapper()` the same way), and this one mirrors `JdbcCustomDashboardRepository.getRowMapper()` line for line apart from the extra `time_range` column. The body is five straight-line statements with no branching, and it is already covered end to end by the shared TCK (`GammaDashboardRepositoryTest`, green on Mongo plus all four JDBC dialects), so extracting it would buy consistency loss rather than testability.

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoGammaDashboardRepository.java
```diff
@@ -0,0 +1,131 @@
...
+ */
+@CustomLog
+@Component
+@RequiredArgsConstructor
+public class MongoGammaDashboardRepository implements GammaDashboardRepository {
+
+    private final GammaDashboardMongoRepository internalGammaDashboardRepo;
+    private final GraviteeMapper mapper;
+
+    @Override
```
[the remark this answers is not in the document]
SamHardi: Partly applied in dd43de65e5, but not as a shared helper — here is the reasoning.

The four methods return three different shapes and use three different mapping expressions: `findAll` returns a `Set` via `mapper.mapGammaDashboards(Collection)`, `findByEnvironmentId` returns a `List` via `stream().map(mapper::map).toList()`, and the two id-based finders return `Optional`. A generic fetch-and-map helper would have to be parameterised over the return type, the mapping expression, and the log message and its arguments — more indirection than the three lines it would replace, and the parameterised debug messages (`[{}]`, `[{}] and [{}]`) would have to be flattened or passed in as fragments.

What was genuinely duplicated is narrower than the comment suggests: only `findById` and `findByIdAndEnvironmentId` share a shape, and both unwrapped an `Optional` to `null` and re-wrapped it:

```java
var dashboard = internalGammaDashboardRepo.findById(id).orElse(null);
return Optional.ofNullable(mapper.map(dashboard));
```

That is now `internalGammaDashboardRepo.findById(id).map(mapper::map)` in both. Shorter, no null round-trip, and it no longer relies on the mapper's generated null check. Leaving the thread open for a human call on whether the remaining shape is worth abstracting. TCK green on Mongo and all four JDBC dialects.

jourdiw (commented):
Reviewed the SPI, both backends, the changeset, the TCK and the cascade, then compared the persisted model against the `Dashboard` contract in `@gravitee/gamma-lib-observability@1.35.2` and the Gamma templates in this repo. The backend-first sequencing and the opaque-`widgets` argument both hold up; five comments below, the first two being the ones I'd want resolved before OBS-16 builds on this shape.

Things I checked and found fine, so they don't need defending: the index upgrader honours the collection prefix and the explicit bean name is genuinely required; `getClassFromFileName` resolves the fixture correctly and per-test teardown makes the count assertions safe; `Types.NCLOB` + `String.class` is precedented; `JdbcColumn` resolves both Lombok setters and `version` is quoted on every dialect; and `DeleteEnvironmentCommandHandler` really is the only cascade site, with Mongo + JDBC being complete backend coverage.

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/GammaDashboard.java
```diff
@@ -0,0 +1,103 @@
...
+    private Integer version;
+
+    /**
+     * Dashboard-definition filter, applied to every widget of the dashboard.
+     */
+    @Builder(toBuilder = true)
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @Data
+    public static class Filter {
```
jourdiw, line 76: **The persisted model doesn't match the `Dashboard` contract it's persisting.**

The shipped type in `@gravitee/gamma-lib-observability@1.35.2` (`dist/types/domain.d.ts:14`):

```ts
export interface FilterCondition {
    field: string;
    label: string;
    operator: FilterOperator;   // 'eq' | 'in' | 'not_in' | ...  (lowercase)
    value: string[];
    valueLabels?: string[];
}
```

and a real literal from this repo (`gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/config/templates/shared.ts:19`):

```ts
{ field: 'API_TYPE', label: 'API Type', operator: 'in', value: [value], valueLabels: [value] }
```

Against `Filter { name, operator, values, editable }`:

- `label` is **required** on `FilterCondition` and has no column here. A dashboard saved and reloaded cannot be turned back into a valid `FilterCondition` — the write path has to invent a label. (Dropping `valueLabels` is fine; `useResolveFilterLabels` rehydrates that from the API. `label` isn't derived anywhere.)
- `editable` doesn't exist on `FilterCondition`. It's a component prop (`FilterBar.d.ts:5`, `DashboardView.d.ts:8`) meaning "this bar is editable" — persisting it per filter stores a concept the renderer can't consume.
- `field`/`value` vs `name`/`values`, and `operator` documented as uppercase (`EQ`, `NOT_IN`) where the lib's `FILTER_OPERATORS` are lowercase. It's a free `String`, so nothing catches the divergence — it becomes a case-mapping table to maintain, or a filter that silently matches nothing.

`TimeRange` has the same problem: the lib is `{ type: 'relative', period: '7d' }` (and the in-repo template uses exactly that), while the javadoc documents `RELATIVE`/`ABSOLUTE` and the fixtures use `"last7d"` / `"last1h"` — not valid `RelativePeriod` values. So the round-trip tests are green on payloads the product never produces: they prove serialization, not compatibility.

The table is empty, so aligning names, case and the missing `label` is free right now and stops being free the moment OBS-16 writes a row.

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/GammaDashboard.java
```diff
@@ -0,0 +1,103 @@
...
+@AllArgsConstructor
+@Data
+@EqualsAndHashCode(of = "id")
+public class GammaDashboard {
+
+    private String id;
+    private String environmentId;
+    private String title;
+    private String description;
+    private List<Filter> filters;
```
jourdiw, line 49: **The opaque/structured line is drawn in the one place it doesn't hold.**

The case for opaque `widgets` is convincing. But widget-level filters are `FilterCondition[]` too (`BaseWidget.filters`, `domain.d.ts:108`) and `scopeWidgets()` injects one into every widget of the bundled template. So after this PR the same type is persisted twice in two different shapes: verbatim inside the opaque `widgets` blob, and renamed/restructured at the dashboard level.

Whatever argument makes `widgets` opaque applies to `filters` and `timeRange` identically. Structuring them buys no validation — every field is a free `String` or `List<String>` — while guaranteeing the lossy round-trip flagged on `Filter` below. Keeping all three opaque would also collapse the JDBC ORM back to plain columns and delete both serializer pairs.

If OBS-15/16 needs to query by filter, that's a good answer — but it isn't in the description, and it's the decision the rest of the model hangs off.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcGammaDashboardRepository.java
```diff
@@ -0,0 +1,162 @@
...
+@CustomLog
+@Repository
+public class JdbcGammaDashboardRepository extends JdbcAbstractCrudRepository<GammaDashboard, String> implements GammaDashboardRepository {
+
+    /**
+     * Unknown properties are ignored on purpose. {@code filters} and {@code time_range} are stored as JSON, so a row
+     * written by a newer version carrying a field this one does not know must still be readable — otherwise a rollback
+     * or a rolling upgrade turns a single unreadable row into a failure of the whole environment listing.
+     */
+    private static final ObjectMapper JSON_MAPPER = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
jourdiw, line 54: **The failure this documents can't happen; the one it enables is silent.**

`JdbcObjectMapper.setFromResultSet` (`JdbcObjectMapper.java:297-325`) calls the deserializer inside a `try` that ends in `catch (Exception ex) { log.error(...) }`. An exception out of `deserialize` is swallowed per column: the row still maps, the field is left `null`, and `findByEnvironmentId` returns successfully. An unreadable row can't fail "the whole environment listing."

The real hazard is the inverse, and disabling `FAIL_ON_UNKNOWN_PROPERTIES` makes it quiet: an older node reading a row written by a newer one drops the fields it doesn't know, and the next `update()` writes the truncated object back. The dashboard is permanently degraded with nothing in the logs. Mongo has the same hole — Spring Data maps into `GammaDashboardMongo` and `save()` replaces the document.

Keeping `filters`/`timeRange` opaque makes forward-compat structural. Otherwise `version` needs to be enforced before the first write endpoint ships. Either way this javadoc should be corrected — it currently tells the next reader the store is safer than it is.

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoGammaDashboardRepository.java
```diff
@@ -0,0 +1,131 @@
...
+
+        var existingDashboard = internalGammaDashboardRepo.findById(gammaDashboard.getId()).orElse(null);
+
+        if (existingDashboard == null) {
+            throw new IllegalStateException(String.format("No gamma dashboard found with id [%s]", gammaDashboard.getId()));
+        }
+
+        try {
+            var dashboardMongo = mapper.map(gammaDashboard);
+            var updatedDashboardMongo = internalGammaDashboardRepo.save(dashboardMongo);
```
jourdiw: **`save()` is an upsert, so `update()` can resurrect a deleted dashboard.**

The javadoc above is right about the lost-update window, but there's a second consequence of the read-then-write gap: if the dashboard is deleted between `findById` and `save`, `save` re-inserts it. JDBC doesn't — `JdbcAbstractCrudRepository.update` throws when 0 rows are affected. So "update a concurrently deleted dashboard" is an error on Postgres and a silent resurrection on Mongo, which is exactly the class of divergence the SPI javadoc says callers shouldn't have to reason about.

An `updateFirst` / `findAndModify` matched on `_id` fixes this and the version gap together, and folds the existence check into the same round trip.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcGammaDashboardRepository.java
```diff
@@ -0,0 +1,162 @@
...
+    @Override
+    protected String getId(GammaDashboard item) {
+        return item.getId();
+    }
+
+    @Override
+    public List<GammaDashboard> findByEnvironmentId(String environmentId) throws TechnicalException {
+        log.debug("JdbcGammaDashboardRepository.findByEnvironmentId({})", environmentId);
+        try {
+            return jdbcTemplate.query(getOrm().getSelectAllSql() + " where environment_id = ?", getRowMapper(), environmentId);
```
jourdiw: **`findByEnvironmentId` returns an unordered `List`.**

Neither backend sorts — this line, and `MongoGammaDashboardRepository:106`. Mongo natural order in particular can change when a document grows on update, so a user can reorder their own dashboard list just by editing one of them.

A list endpoint needs a stable order and this is the natural place to establish it (`order by title` or `created_at`) — an unordered `List` return type is easy to depend on by accident. `CustomDashboard` has the same gap, but it's cheaper to fix in a store with no rows in it.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcGammaDashboardRepository.java
SamHardi: You are right on both counts, and the javadoc was the worse of the two problems — it told the next reader the store was safer than it is. Corrected in 32006944f3.

I checked `setFromResultSet` (JdbcObjectMapper.java:297-325): the per-column `catch (Exception ex) { log.error(...) }` is exactly as you describe, so a deserialization failure leaves the field null and the query still succeeds. The "whole environment listing fails" claim was simply wrong.

On the flag itself I have kept `false`, but only as the lesser of two silent failures, and the javadoc now says so rather than presenting it as forward-compatibility. With the feature enabled, a row written by a newer node comes back with `filters` entirely null — the exception is swallowed the same way — and the next `update()` writes that null back. Ignoring the unknown field loses only that field. Both are silent data loss.

Your structural point stands and I have not acted on it here: keeping `filters`/`timeRange` opaque is the real fix, and it is the same decision as your comment on the opaque/structured line. I am raising both with the ticket owner rather than settling the epic contract in a review reply.

gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcGammaDashboardRepository.java
SamHardi: Agreed and fixed in 32006944f3. Both backends now order by `created_at` then `id`: `order by created_at, id` on JDBC, and the derived `findByEnvironmentIdOrderByCreatedAtAscIdAsc` on Mongo. The ordering is stated on the SPI method so it reads as contract rather than incidental behaviour.

The id tie-break is not decoration: the three DEFAULT fixtures deliberately share a `createdAt`, so without it the order would still be backend-dependent. `shouldReturnEnvironmentDashboardsInStableOrder` asserts the exact sequence and is green on Mongo and all four JDBC dialects.

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/GammaDashboard.java
SamHardi: Aligned in a839f3ad9e. Every point checked out, including the ones I had taken from the ticket without verifying against the shipped types.

`Filter` is now `{ field, label, operator, value, editable }`, and `TimeRange.type` is lowercase. Concretely: `name`→`field`, `values`→`value`, operators emitted as the lib emits them (`in`, `eq`, `not_in`), and `label` persisted.

On `label` I checked rather than assumed: `use-resolve-filter-labels.ts:9` hydrates `valueLabels` only, never `label`. So the epic's rationale for dropping display labels — they go stale when an entity is renamed — holds for `valueLabels`, which reference entities, and not for `label`, which is the field's own display name. You were right that the write path would otherwise have to invent it.

You were also right about the fixtures, and that was the sharpest part of the review: they used `last7d`/`last1h`, which are not `RelativePeriod` values at all (the legal set is `5m`,`15m`,`1h`,…,`7d`,`30d`,`currentMonth`…). The suite was green on payloads the product cannot produce. Fixtures now use the real shape, modelled on the in-repo literal in `templates/shared.ts:19`.

One deliberate divergence kept, flagged rather than hidden: `editable`. The lib's `FilterCondition` does carry `locked?: boolean`, and OBS-13 replaced that optional tri-state with a required positive boolean. It is documented as the single intentional difference.

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoGammaDashboardRepository.java
SamHardi: Fixed in a839f3ad9e. `update()` is now a single `findAndReplace` matched on `_id` with `returnNew()` and no upsert — a null result means the document is gone and raises `IllegalStateException`, matching what JDBC does on 0 affected rows. The existence check folds into the same round trip, and enforcing `version` later is one more criterion on that query rather than another trip.

`shouldNotResurrectADashboardDeletedBeforeTheUpdate` pins it: read a dashboard, delete it, update the stale copy, expect the raise and expect it to stay deleted. Green on Mongo and all four JDBC dialects — worth noting it would have passed on JDBC before this change and failed on Mongo, which is exactly the divergence you flagged.

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/model/GammaDashboard.java
SamHardi: Kept structured, but aligned to the shipped shape — see the reply on the `Filter` thread. Reasoning, since I did seriously consider going opaque.

Your diagnosis is right: the same type was persisted twice in two divergent shapes. But the remedy that follows is to make the shapes match, not to drop the structure. What justifies opacity for `widgets` is an **open** schema — 6 variants × ~40 optional fields, owned by the frontend, changing every release, with the dotted-map-key hazard. None of that transfers: `FilterCondition` is a closed 5-field shape, `TimeRange` a closed 4-field one, and neither has dotted map keys — the values carrying dots sit in arrays, not in keys.

The cost of opacity is real too. Structuring does not validate values — `field` can be empty, `operator` can be anything — but it does guarantee shape integrity: the column necessarily holds a list of five-field objects, because a caller has to build `Filter` instances to write one. Opaque, a caller could persist `"garbage"` and the store would accept it and hand it back to the frontend. I had originally written that structuring "buys no validation"; that conflated value validation with shape integrity, and it is the latter that is worth keeping here.

With the shapes now identical at both levels, the same type is still stored twice, but not in two different forms — which was the actual complaint.

jourdiw (approved):

PR #18789, opened 2026-07-28
feat: grant AI Workspace access independently of API Product, and audit it

vikrantgravitee:
https://gravitee.atlassian.net/browse/AIAM-195

## Problem

An AI Workspace is stored as an API Product, so it was also *governed* by API Product permissions:

- Granting someone API Product access silently gave them every AI Workspace
- AI Workspace access could not be granted **without** granting API Product access

And there was no audit trail for the workspace actions that write a *different* entity — adding a model mutates the underlying LLM proxy, adding a user mutates a subscription — so neither left a trace on the workspace being administered.

## Decoupling the grant from the storage

`RoleScope.AI_WORKSPACE` maps to the **existing** `API_PRODUCT` membership reference type:

```java
Pair.of(RoleScope.AI_WORKSPACE, MembershipReferenceType.API_PRODUCT)
```

That one line is what keeps this cheap. Memberships stay where they are, so `ApiProductAccessibleIdsDomainService` keeps resolving them and `GetApiProductsUseCase` keeps filtering the listing by membership with the admin bypass. **No migration, no new accessible-ids service, no change to how workspaces are listed** — only who may be granted what.

`PLAN` keeps the platform-wide name even though the workspace UI calls it a tier, so a role matrix reads consistently across scopes.

### Roles

| Role | Source |
|---|---|
| `PRIMARY_OWNER` | system role, per-scope block in `RoleServiceImpl` |
| `OWNER` | `ROLE_AI_WORKSPACE_OWNER` — full CRUD |
| `USER` | `ROLE_AI_WORKSPACE_USER` — read-only, default |

`ADMIN` is deliberately absent: it is an environment/organization-scope role here, not a resource-scope one (`API_PRODUCT` and `CLUSTER` both have exactly the three above). An environment admin already sees every workspace through the `isAdmin` bypass.

## Audit — write and read

**Write.** `AiWorkspaceAuditEvent` covers only the actions that would otherwise leave no trace. Creating, updating and deleting a workspace already emit `ApiProductAuditEvent`s and tier changes emit `PlanAuditEvent`s — deliberately **not** duplicated.

**Read.** The audit chain could only be searched for an API, so workspace entries were written and then unreachable. `AuditRepository` already accepts any reference type through `AuditCriteria`, so this is the same query with a different reference — the two search methods now share one implementation instead of duplicating the criteria and paging. No repository-level work was needed.

> **No management-v2 REST resource on purpose.** An `ApiProductAuditsResource` would be guarded by `API_PRODUCT_DEFINITION` — exactly the permission an AI Workspace role does not carry — so the endpoint belongs in AIM against the `AI_WORKSPACE` scope. That, and the Audit Logs page, land in the AIM PR.

> The two `AI_WORKSPACE_COMPONENT_*` constants have no emitter yet: components do not exist in AIM core on `main`. Happy to drop them until they do.

## Commits

Split for review; they ship together:

- `e549e73e35` — backend: scope, roles, upgrader, audit events
- `e9f18e8672` — console: the roles page enumerates its scopes explicitly, so a new scope stays invisible until listed there too
- `fb366c28ce` — audit read path: filters, query service, domain service, use case

## Verification

| Suite | Result |
|---|---|
| `rest-api-model` (full) | 227/227 |
| `rest-api-rest` (full, incl. `PermissionsFilter`) | 44/44 |
| role / permission / membership in `rest-api-service` | 282/282 |
| audit in `rest-api-service` | 64/64 |
| enum-enumerating tests (`DefaultOrganizationAdminRoleInitializer`, `MembershipService_*`, cluster permissions) | 85/85 |
| console role specs | 27/27, tsc + eslint + prettier clean |

Three existing tests needed updating, all assertions on exact platform inventories: `RoleScopesResourceTest` (scope count 9→10 and the ENVIRONMENT permission list) and `RoleService_CreateOrUpdateSystemRolesTest` (system role count 10→11).

**No Liquibase change.** `role_scope` is a plain varchar and `AI_CATALOG` was added the same way with none.

## Worth a reviewer's eye

- **`AiWorkspaceRolesUpgrader` against an existing organization**, not just a fresh install. That is where role upgraders usually fail and no unit test can prove it.
- Because both scopes hang off `MembershipReferenceType.API_PRODUCT`, an API Product membership on a workspace still surfaces it in the listing. The decoupling is one-way by design — a workspace *is* an API Product — but it should be a stated decision rather than a surprise.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/MembershipReferenceType.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/MembershipReferenceType.java
@@ -29,7 +29,7 @@ import lombok.RequiredArgsConstructor;
 public enum MembershipReferenceType {
     APPLICATION(EnumSet.of(RoleScope.APPLICATION)),
     API(EnumSet.of(RoleScope.API)),
-    API_PRODUCT(EnumSet.of(RoleScope.API_PRODUCT)),
+    API_PRODUCT(EnumSet.of(RoleScope.API_PRODUCT, RoleScope.AI_WORKSPACE)),
     GROUP(
         EnumSet.of(
             RoleScope.GROUP,
@@ -39,7 +39,8 @@ public enum MembershipReferenceType {
             RoleScope.INTEGRATION,
             RoleScope.CLUSTER,
             RoleScope.AI_CATALOG,
-            RoleScope.EXPLORER
+            RoleScope.EXPLORER,
+            RoleScope.AI_WORKSPACE
         )
     ),
     ENVIRONMENT(EnumSet.allOf(RoleScope.class)),
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/model/ApiProductAuditQueryFilters.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/model/ApiProductAuditQueryFilters.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.api_product.model;
+
+import java.util.Optional;
+import java.util.Set;
+
+public record ApiProductAuditQueryFilters(
+    String apiProductId,
+    String organizationId,
+    String environmentId,
+    Optional<Long> from,
+    Optional<Long> to,
+    Set<String> events
+) {
+    public ApiProductAuditQueryFilters {
+        if (apiProductId == null) {
+            throw new IllegalArgumentException("apiProductId must not be null");
+        }
+        if (organizationId == null) {
+            throw new IllegalArgumentException("organizationId must not be null");
+        }
+        if (environmentId == null) {
+            throw new IllegalArgumentException("environmentId must not be null");
+        }
+        if (events == null) {
+            events = Set.of();
+        }
+    }
+}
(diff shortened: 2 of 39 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/MembershipReferenceType.java
```diff
@@ -29,7 +29,7 @@
 public enum MembershipReferenceType {
     APPLICATION(EnumSet.of(RoleScope.APPLICATION)),
     API(EnumSet.of(RoleScope.API)),
-    API_PRODUCT(EnumSet.of(RoleScope.API_PRODUCT)),
+    API_PRODUCT(EnumSet.of(RoleScope.API_PRODUCT, RoleScope.AI_WORKSPACE)),
```
araffass, line 32: Now that API_PRODUCT reference type carries two scopes, the scope can't be inferred from the reference type name anymore. Worth flagging because MembershipServiceImpl.createNewMembership (line 2202) still does exactly that: 
  
new MembershipRole(RoleScope.valueOf(referenceType.name()), roleName)                                                                                                                                                                                                                                                                                                                                                        

For an API_PRODUCT reference that always resolves to RoleScope.API_PRODUCT, never AI_WORKSPACE. So if the AIM member-add PR reuses this convenience method to add a workspace member, it'll quietly write an API_PRODUCT grant and re-couple the exact thing this PR decouples. 

Nothing breaks in this PR since there's no workspace member-add here, but maybe if no solution yet then add a comment (or a guard) on createNewMembership so the follow-up path is forced to pass RoleScope.AI_WORKSPACE explicitly via addRoleToMemberOnReference

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/member/model/MembershipReferenceType.java
```diff
@@ -26,7 +26,7 @@
 public enum MembershipReferenceType {
     APPLICATION(EnumSet.of(RoleScope.APPLICATION)),
     API(EnumSet.of(RoleScope.API)),
-    API_PRODUCT(EnumSet.of(RoleScope.API_PRODUCT)),
+    API_PRODUCT(EnumSet.of(RoleScope.API_PRODUCT, RoleScope.AI_WORKSPACE)),
```
araffass, line 29: One consequence of sharing the reference type that isn't covered by the "one-way listing" note in the description: GetApiProductMembersUseCase (line 39) lists getMembersByReference(API_PRODUCT, id) and only filters member.getType() == USER, never the role scope. 

The classic ApiProductMembersResource does the same. Since a workspace is stored as an API Product, the API Product members view and the workspace members view will now show a merged member list across both scopes, and firstRole can surface a role from the wrong scope. 

If those two admin surfaces are meant to stay isolated, each list needs a .filter on the role scope. 

This ne's also not necessary a blocker for this PR, but it lands as soon as the AIM members page reads this reference.

araffass (approved):
LGTM - With 2 Notes worth checking

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/model/ApiProductAuditQueryFilters.java
```diff
@@ -0,0 +1,43 @@
...
+package io.gravitee.apim.core.api_product.model;
+
+import java.util.Optional;
+import java.util.Set;
+
+public record ApiProductAuditQueryFilters(
+    String apiProductId,
+    String organizationId,
+    String environmentId,
+    Optional<Long> from,
```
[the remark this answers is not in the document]
vikrantgravitee: done

PR #18792, opened 2026-07-29
feat(console): surface the EXPLORER role scope (org roles + group members)

ThibaudAV:
https://gravitee.atlassian.net/browse/ESM-43
## Description

Surfaces the **EXPLORER** role scope (Kafka Explorer, used by the Gamma ESM module) everywhere the console already exposes the `CLUSTER` scope, plus a latent transfer-ownership NPE fix found along the way. Four self-contained commits:

1. **`fix(membership)`** — `MembershipServiceImpl.transferOwnership` threw a 500 NPE on the "new owner == current owner" short-circuit when the new primary owner was identified by `userReference` only (a not-yet-registered directory user), so `getMemberId()` was `null` — a documented, valid case. Now uses `Objects.equals(...)` so a reference-only transfer falls through to reference-based resolution instead of NPE-ing. Adds a regression test.

2. **`feat(console)` — org roles page** — the Organization > Roles screen now lists an **Explorer** scope section (like Cluster), fetching `rolescopes/EXPLORER/roles`. No icon, matching Cluster.

3. **`feat(console)` — group member management** — the group edit page can now set an **Explorer** role per member: a "Default Explorer Role" select in the Add Members and Edit Member dialogs, an "Explorer Role" column in the members table, and the EXPLORER role-list fetch. Adds the scope to the `RoleScope` enum and the group-membership role union.

4. **`fix(console)` — persist the role** — `GroupMembersResource.addOrUpdateGroupMember` only wired API, API_PRODUCT, APPLICATION, GROUP, INTEGRATION and CLUSTER. The EXPLORER role sent by the UI fell into the switch `default` and had no update block, so it was **silently dropped and never persisted** — leaving the Explorer Role column empty on reload. Wire EXPLORER exactly like CLUSTER (previous-role tracking, update block, delete-if-emptied, new-member notification guard).


callaertanthony (approved):
LGTM 👍 — clean, well-scoped, and the four commits split nicely.

### What I checked

- The backend `EXPLORER` wiring in `GroupMembersResource` is a line-for-line mirror of `CLUSTER`: previous-role capture, the update block with the same `e -> true` lock predicate, `deleteIfNewAndPreviousRoleNull`, and the new-member notification guard. No asymmetry between the two scopes.
- `RoleScope.EXPLORER` is already a first-class scope everywhere it needs to be — the three enums, `RoleScopesResource`, `PermissionServiceImpl.ROLE_SCOPE_TO_REFERENCE_TYPE` — and `ExplorerRolesUpgrader` seeds the default OWNER/USER roles, so `rolescopes/EXPLORER/roles` won't come back empty on upgraded installs.
- Console changes follow the `CLUSTER` pattern exactly (column markup, dialog data, form control, submit payload). Leaving `InviteMemberDialogComponent` untouched is right — it only handles API + Application roles.
- CI is green apart from the console SonarCloud gate, whose only failing condition is `new_coverage 0.0 < 40`. That gate reports 0% globally (including `master`), so it's the known-broken measurement rather than a coverage gap — this PR adds a good amount of test surface.

### Non-blocking notes

1. **The NPE class isn't fully closed.** `getPrimaryOwner` ends in `.orElse(null)`, so the right-hand side of the line you fixed can still NPE, as can the `previousPrimaryOwner` dereferences further down (`removeRole`, the `getMemberType()` branches). The method assumes a primary owner exists throughout, so it's pre-existing and fine to leave out of scope — just worth not treating the transfer path as null-safe now.

2. **Sixth copy of the same three lines.** `disableDefaultExplorerRole()` is identical to its five siblings in both dialogs. A loop over the control names inside the existing `disable…` aggregator would drop a few dozen lines across the two components and make scope #7 free.

3. **`org-settings-roles` positional destructuring.** Adding one scope forced reformatting the whole `tap` callback and bumping a 9-arg test helper. A small `SCOPES` table (`[label, scopeId]`) fed into `forkJoin(SCOPES.map(...))` would turn the next scope into a one-line row instead of five coordinated edits.

4. **Comment nits.** The comment above the `Objects.equals` change mostly explains what `Objects.equals` does; the valuable half is *why* `memberId` can be null. And `.isNotInstanceOf(NullPointerException.class)` is redundant right after `.isInstanceOf(UserNotFoundException.class)`.

5. **Question, not a change request:** the Kafka Explorer routes are license-gated behind `APIM_NATIVE_KAFKA_EXPLORER`, but the Explorer role column, both selects, and the org-roles Explorer section render on every install. `CLUSTER` behaves the same way so there's precedent — just want to confirm it's deliberate rather than inherited.

None of these block merging.

PR #18806, opened 2026-07-29
feat(ai-workspace): scope permissions independently and enable the audit trail

vikrantgravitee:
## What

Backend and console support for scoping AI Workspace permissions independently of API Product, and for reading back a workspace's audit trail. This is the APIM-side counterpart to the AIM module work.

## Changes

- **Grant AI Workspace access independently of API Product.** A workspace is stored as an API Product, but holding an API Product permission must not imply access to a workspace. Adds the `AI_WORKSPACE` role scope, its permissions (`DEFINITION`, `PLAN`, `USER`, `MEMBER`), and the default `OWNER`/`USER` roles, seeded by an upgrader.
- **Console — roles page.** Surfaces the `AI_WORKSPACE` scope so its role matrix can be edited alongside the other scopes.
- **Searchable audit trail.** The audit search only accepted an API reference; a workspace records against `API_PRODUCT`, so its entries were written and then unreachable. The two search paths now share one implementation, and result metadata is keyed by the audit's own reference type instead of assuming an API.
- **Audit model.** Adds the `AI_WORKSPACE_COMPONENT` audit property (used to reference a workspace component such as a model, in both the core and repository enums) and trims the `AI_WORKSPACE` audit events to the ones actually recorded.

## Notes

No management-v2 REST resource is added for the workspace audit trail on purpose: such a resource would be guarded by `API_PRODUCT_DEFINITION`, which is exactly the permission an AI Workspace role does not carry. The endpoint is exposed by the AIM module against the `AI_WORKSPACE` scope instead.


vikrantgravitee:
Closing: this branch re-introduces the AI Workspace feature already merged in #18789. Reopening a focused PR against current master with only the audit-property follow-up.

PR #18811, opened 2026-07-29
feat(cluster): server-side lifecycle-state filter + /clusters/_stats counts

ThibaudAV:
## Description

Adds the server-side pieces the Gamma ESM clusters landing needs (drop the client-side `perPage: 100`
cap; resolve search / filter / pagination / stat-strip on the backend). Two additions to the shared v2
clusters API, plus repository coverage.

### 1. Filter the clusters search by lifecycle state
`GET /clusters` gains a repeatable `lifecycleState` query param (bound to the `ClusterLifecycleState`
enum → proper 400 on garbage), threaded through `SearchClusterUseCase` → core `ClusterSearchCriteria` →
repository `ClusterCriteria` (MapStruct maps it by name) → Mongo (`where lifecycleState in`) and JDBC
(`lifecycle_state in (?)`). OpenAPI updated.

### 2. `GET /clusters/_stats` — counts by lifecycle state
New `CountClustersByLifecycleStateUseCase` returns the total plus per-state counts (Deployed / Pending /
Undeployed), optionally scoped to a `type` and restricted to the caller's readable clusters. Reuses the
existing search + lifecycleState filter (one count query per bucket) — no bespoke aggregation to keep in
sync across Mongo/JDBC. Backs the console cluster stat-strip.

## Review follow-ups (all addressed)

- **🔴 Blocking — legacy `NULL lifecycle_state` rows.** Pre-4.12 clusters have a NULL state; the read
  model coerces it to `UNDEPLOYED` but the pushdown filter/counts query the column directly, so a NULL
  cluster would vanish under the Undeployed facet and `_stats` would report `total != sum`. **Fixed by
  backfilling** `NULL → UNDEPLOYED`: a new JDBC Liquibase changeset + a new Mongo upgrader. The pushdown
  is now consistent with the read model.
- **No REST test for `_stats`** → added `ClusterStatsTest` (mapped counts + `ENVIRONMENT_CLUSTER[READ]`
  gate) and registered the use case as a mock bean in `ResourceContextConfiguration`.
- **`_stats` runs 4 queries** — kept the separate `total` (`count(*)`) query as a deliberate defensive
  measure so the strip stays honest even if a row ever falls outside the three buckets; it is now
  redundant-but-correct after the backfill.
- **8 positional args on `SearchClusterUseCase.Input`** → both call sites now use the `@Builder`.
- **enum validation / single source of truth / import cleanup** (earlier round) — done.

## Tests

- Repository: `ClusterRepositoryTest` — lifecycleState filter (match / no-match), run against **Mongo (16)**
  and **JDBC (16)**; the backfill changeset + Mongo upgrader run in setup without breaking fixtures.
- Use case: `CountClustersByLifecycleStateUseCase` (4), `SearchClusterUseCase` (10).
- REST: `ClustersResourceTest` incl. `ClusterStatsTest` (13 total).

## Notes

- No schema migration for the column (it already exists) — only a data backfill.
- Consumer: the Gamma ESM module PR wires `ClustersPage` to these (`useClusters({ lifecycleState })` +
  `useClusterStats` → `/clusters/_stats`). This PR is the backend half.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/api/search/ClusterCriteria.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/management/api/search/ClusterCriteria.java
@@ -28,5 +28,6 @@ public class ClusterCriteria {
     private String environmentId;
     private List<String> ids;
     private String type;
+    private List<String> lifecycleStates;
     private String query;
 }
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcClusterRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcClusterRepository.java
@@ -163,6 +163,11 @@ public class JdbcClusterRepository extends JdbcAbstractCrudRepository<Cluster, S
             andWhereParams.add(criteria.getType());
         }
 
+        if (!CollectionUtils.isEmpty(criteria.getLifecycleStates())) {
+            andWhere.add("lifecycle_state in (" + getOrm().buildInClause(criteria.getLifecycleStates()) + ")");
+            andWhereParams.addAll(criteria.getLifecycleStates());
+        }
+
         if (criteria.getQuery() != null) {
             andWhere.add("((lower(name) like ?) OR (lower(description) like ?))");
             andWhereParams.add("%" + criteria.getQuery().toLowerCase() + "%");
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_13_0/08_backfill_cluster_lifecycle_state.yml
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/changelogs/v4_13_0/08_backfill_cluster_lifecycle_state.yml
@@ -1,0 +1,16 @@
+databaseChangeLog:
+  - changeSet:
+      id: 4.13.0_08_backfill_cluster_lifecycle_state
+      author: GraviteeSource Team
+      changes:
+        # Clusters created before the lifecycle_state column existed (added nullable in 4.12) have a NULL
+        # state. The domain layer reads NULL as UNDEPLOYED, but the server-side lifecycleState filter and
+        # the /clusters/_stats counts query the column directly, so a NULL row would neither match the
+        # Undeployed facet nor be counted. Backfill so the pushdown stays consistent with the read model.
+        - update:
+            tableName: ${gravitee_prefix}clusters
+            columns:
+              - column:
+                  name: lifecycle_state
+                  value: UNDEPLOYED
+            where: lifecycle_state IS NULL
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/master.yml
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/resources/liquibase/master.yml
@@ -401,3 +401,5 @@ databaseChangeLog:
         - file: liquibase/changelogs/v4_13_0/06_add_idp_claims_to_users.yml
     - include:
         - file: liquibase/changelogs/v4_13_0/07_add_api_product_id_to_portal_navigation_items.yml
+    - include:
+        - file: liquibase/changelogs/v4_13_0/08_backfill_cluster_lifecycle_state.yml
--- a/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/internal/clusters/ClusterMongoRepositoryImpl.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/internal/clusters/ClusterMongoRepositoryImpl.java
@@ -58,6 +58,10 @@ public class ClusterMongoRepositoryImpl implements ClusterMongoRepositoryCustom 
             query.addCriteria(where("type").is(criteria.getType()));
         }
 
+        if (!CollectionUtils.isEmpty(criteria.getLifecycleStates())) {
+            query.addCriteria(where("lifecycleState").in(criteria.getLifecycleStates()));
+        }
+
         if (criteria.getQuery() != null) {
             query.addCriteria(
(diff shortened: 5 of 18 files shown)


callaertanthony (commented):
## Review

Nice, focused change — the pushdown is the right move and the in-memory double mirrors the real filter. One blocker though.

### :red_circle: Blocking — legacy `NULL lifecycle_state` rows break both the filter and the counts

The domain layer coerces `NULL` → `UNDEPLOYED`, but the new pushdown filter doesn't:

- `ClusterAdapter#mapLifecycleStateFromString` returns `UNDEPLOYED` when the DB value is null. Every read path (`ClusterQueryServiceImpl#search`) goes through it, so legacy clusters **display as Undeployed**.
- `JdbcClusterRepository:167` — `lifecycle_state in (?)` never matches `NULL`. Same for Mongo, `ClusterMongoRepositoryImpl:62` `where("lifecycleState").in(...)`.
- `liquibase/changelogs/v4_12_0/05_add_lifecycle_to_clusters_table.yml` adds the column **nullable, with no backfill and no default** — so every cluster created before 4.12 has `NULL`.

Two user-visible consequences:

1. A cluster listed as "Undeployed" **vanishes** when you tick the Undeployed filter.
2. `_stats` reports `total != deployed + pending + undeployed` — the strip silently loses clusters.

Reproducible today against the repository test fixture (`clusters.json`: env-1 has 4 clusters, only 1 has a `lifecycleState`) → Total 4 / Deployed 1 / Pending 0 / Undeployed 0.

**Suggested fix:** backfill, rather than teaching two repositories about null. A Liquibase changeset + Mongo upgrader:

```sql
UPDATE ${gravitee_prefix}clusters SET lifecycle_state = 'UNDEPLOYED' WHERE lifecycle_state IS NULL
```

That makes the pushdown honest and eventually lets the adapter's null-default go away.

### :yellow_circle: Non-blocking

**`_stats` runs 4 queries where 3 would do** — `CountClustersByLifecycleStateUseCase#execute`. Once the backfill lands, `total = deployed + pending + undeployed`, so the unfiltered query can go. Right now the 4th isn't redundant, and that is precisely the symptom of the bug above.

**No REST-layer test for `_stats`.** `ClustersResourceTest` is untouched, and `CountClustersByLifecycleStateUseCase` isn't registered as a mock bean in the v2-rest `ResourceContextConfiguration` — so adding a test needs that file too. `@Path("_stats")` vs `@Path("{clusterId}")` precedence and the `ENVIRONMENT_CLUSTER[READ]` permission are both currently unverified. Probably also what the red rest-api Sonar gate is about.

**`SearchClusterUseCase.Input` is now 8 positional args**, with the new one inserted in the middle. `ProxyKafkaConsoleResource#computeKafbatToken` passes four consecutive `null`s. The record already has `@Builder` — worth using it at call sites; adjacent-null swaps are a real hazard at this arity.

### :white_check_mark: Checked and good

`buildInClause` is parameterized (no injection); `lifecycle_state` / `lifecycleState` column and field names match their mappings; the OpenAPI enum `[UNDEPLOYED, DEPLOYED, PENDING]` matches `ClusterLifecycleState`; MapStruct auto-maps the new criteria field (both `List<String>`, same name); the empty-membership → zero-counts guard correctly mirrors `SearchClusterUseCase`.

callaertanthony:
Cross-repo heads-up: `gravitee-io/gravitee-gamma-module-esm#54` consumes both of these (the ESM Clusters landing sends `lifecycleState` and calls `GET /clusters/_stats`), so it blocks on this PR landing and then being present in the host's APIM version.

Worth noting the failure mode if they ship out of order, since neither is loud: an unknown `lifecycleState` param is silently dropped by JAX-RS, so the Lifecycle facet returns the unfiltered list; and `GET /clusters/_stats` falls through to `@Path("{clusterId}")` in `ClustersResource`, resolving as `getCluster("_stats")` → 404, which the ESM stat strip currently renders as `0/0/0/0`.

callaertanthony (approved):
Re-reviewed after the force-push. Blocker is fixed and all checks are green — LGTM. :+1:

### :white_check_mark: Blocker resolved — NULL `lifecycle_state` backfill

`70abff5` adds both halves. I checked each link in the chain:

| Check | Result |
|---|---|
| `liquibase/changelogs/v4_13_0/08_backfill_cluster_lifecycle_state.yml` registered in `master.yml` | :heavy_check_mark: |
| Folder matches project version (`pom.xml` `<revision>4.13.0`) | :heavy_check_mark: |
| `where: lifecycle_state IS NULL`, no rollback block | :heavy_check_mark: consistent — no v4_13_0 changeset defines one |
| Mongo upgrader actually picked up | :heavy_check_mark: `@Component` under `io.gravitee.repository.mongodb.management`, which `ManagementRepositoryConfiguration:52` `@ComponentScan`s |
| `getCollection("clusters")` applies the `management.mongodb.prefix` | :heavy_check_mark: |
| `Filters.eq("lifecycleState", null)` matches null **and** absent fields | :heavy_check_mark: right call, and the comment says so |
| One-shot backfill is enough (no new NULLs) | :heavy_check_mark: `CreateClusterUseCase:84` always sets `UNDEPLOYED`, and it's the only production creation path — every other `clusterCrudService.create` caller is a test |

### :white_check_mark: Other points addressed

- `ClustersResourceTest.ClusterStatsTest` covers both the 200 payload and `ENVIRONMENT_CLUSTER[READ]` → 403; the mock bean is registered in `ResourceContextConfiguration`.
- Both `SearchClusterUseCase.Input` call sites use `Input.builder()` now — `ProxyKafkaConsoleResource` no longer passes four consecutive `null`s.
- The rest-api Sonar gate went from red to green.

### :yellow_circle: Optional follow-ups — none blocking, take or leave

**`_stats` still runs 4 queries.** With NULLs backfilled, `total == deployed + pending + undeployed` by construction, so the unfiltered query is now genuinely redundant (it wasn't before — that was the symptom of the original bug).

**`ClusterLifecycleStateUpgrader` has no test.** `EdgeConfigCollectionRenameUpgraderTest` and the index upgrader tests establish the convention, and the null-or-absent `Filters.eq` semantic is exactly the subtlety worth pinning down.

**Upgrader order is arbitrary.** `THEME_TYPE_UPGRADER_ORDER + 1` collides with `MissingEnvironmentUpgrader`'s order and pulls in an import for a dependency that doesn't exist. Harmless — `PlanReferenceType`/`ThemeType` already collide the same way — but this backfill has no ordering constraint at all, so the coupling buys nothing.

**Repository tests only exercise a single-element `IN`.** The OpenAPI advertises "Repeat the parameter to match several states"; `buildInClause` with more than one placeholder is the untested path.

PR #18820, opened 2026-07-29
fix(console): remove 160-char limit when editing dictionary property value + added hint text

arpzilla-dev:
## Issue

https://gravitee.atlassian.net/browse/APIM-14827

## Description

- Removed the arbitrary md-maxlength: 160 validator from the dictionary property inline edit dialog in Console. Editing a property value now allows the same length as creating one (and as the backend, which stores values as nclob). 
Added unit tests covering the edit path.

- Added hint for edit mode.

## Additional context

- Before :

image

- After :

image

- video : 

https://github.com/user-attachments/assets/2974a10a-d83d-4466-b97e-6c3e01200ccb

### DIFF
--- a/gravitee-apim-console-webui/src/management/settings/dictionaries/dictionary.controller.spec.ts
+++ b/gravitee-apim-console-webui/src/management/settings/dictionaries/dictionary.controller.spec.ts
@@ -1,0 +1,154 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import type {} from 'angular-material';
+
+import DictionaryController from './dictionary.controller';
+
+describe('DictionaryController', () => {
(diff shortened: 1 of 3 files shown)


gravitee-apim-console-webui/src/management/settings/dictionaries/dictionary.controller.ts
```diff
@@ -35,6 +35,7 @@ class DictionaryController {
   private updateMode: boolean;
 
   private selectedProperties: any = {};
+  private propertiesDirty = false;
```
aqibmohammadkhan, line 38: propertiesDirty not cleared when properties are saved via other paths

- saveProperties() clears the flag, but the general Update path (update()) also persists the whole dictionary (including properties) and never sets propertiesDirty = false.
- saveProperties() clears the flag, but the general Update path (update()) also persists the whole dictionary (including properties) and never sets propertiesDirty = false.
- Same risk with Reset (reset()): it restores initialDictionary (including properties) but leaves the hint visible.


Fix: clear (or recompute) propertiesDirty in update() / reset() (and likely deploy() if that reloads dictionary state)

gravitee-apim-console-webui/src/management/settings/dictionaries/dictionary.controller.spec.ts
aqibmohammadkhan, line 1: Tests don’t cover add/delete dirty transitions or the stale-table refresh behavior beyond the edit save callback .

arpzilla-dev: add/delete properties are not part of this ticket, they seems to be missing test from old implementation not this ticket scope.

gravitee-apim-console-webui/src/management/settings/dictionaries/dictionary.controller.ts
arpzilla-dev: fixed

aqibmohammadkhan (approved):

PR #18821, opened 2026-07-29
fix(repository): detect Mongo compatibility via buildInfo storageEngines

callaertanthony:
## Issue

https://gravitee.atlassian.net/browse/APIM-14808

## Description

`RemoveDeletedGroupsFromApisUpgrader` gated itself on `buildInfo` containing a
`storageEngine` key, as a way to tell a real MongoDB apart from DocumentDB.
`buildInfo` never returns that key — `storageEngine` is a `serverStatus` field.
The supported engines are reported by `buildInfo` under **`storageEngines`**
(plural).

The check was therefore always `false`, and the upgrader logged
`Skipping this upgrade because database is not compatible` and did nothing.

**This is broader than the ticket suggests:** it was never Atlas-specific. Every
MongoDB deployment has been silently skipping this upgrader since the check was
introduced, so the deleted-group cleanup has not run for anyone on Mongo.

The fix probes `storageEngines` instead. DocumentDB's `buildInfo` does not
report it, so DocumentDB stays excluded as intended, and the
`majorVersion >= 5` guard is untouched. The now-unused
`DOCUMENTDB_ENGINE_VERSION_PREFIX` constant is removed.

## Additional context

Added `RemoveDeletedGroupsFromApisUpgraderTest` covering the three shapes that
matter — Atlas/MongoDB `buildInfo` (compatible), DocumentDB `buildInfo` with no
`storageEngines` (not compatible), and MongoDB 4.4 (not compatible, version
guard).

Targets `4.11.x` (version reported by the customer); labelled for
`4.12.x` and `master`.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/upgrade/upgrader/groups/RemoveDeletedGroupsFromApisUpgrader.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/upgrade/upgrader/groups/RemoveDeletedGroupsFromApisUpgrader.java
@@ -39,7 +39,6 @@ public class RemoveDeletedGroupsFromApisUpgrader extends MongoUpgrader {
 
     private static final String ATTR_GROUPS = "groups";
     private static final String ATTR_ID = "_id";
-    private static final String DOCUMENTDB_ENGINE_VERSION_PREFIX = "5.0.0";
 
     @Override
     public String version() {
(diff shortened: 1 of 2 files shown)


aditya-goyal01:
Verified 18821's critical finding via bytecode (not just trusting subagent).

**PR 18821** — Mongo `storageEngines` compatibility check fix. **Critical bug, confirmed myself:**

`version()` (`RemoveDeletedGroupsFromApisUpgrader.java:45`) still returns `"v1"` — unchanged by this diff. I decompiled `UpgraderServiceImpl` (gravitee-node-services-upgrader 9.4.1) and confirmed the skip logic directly:

```
UpgraderRepository.findById(upgrader.identifier())
  -> if UpgradeRecord exists: log "already applied. It will be ignored." -> RETURN, upgrade() never called.
```

`identifier()` defaults to `className + "_" + version()`. Also confirmed `upgrade()` (untouched, lines 50-74) `return true` unconditionally — including the "not compatible, skipping" branch and the catch-all safeguard. So SUCCESS gets recorded and an `UpgradeRecord` persisted **every prior boot**, regardless of whether the (broken) compatibility check passed.

Net: every install that already booted this upgrader (introduced ~Sept 2025, backported to 4.6.x-4.8.x) already has `RemoveDeletedGroupsFromApisUpgrader_v1` marked done. Since `identifier()` doesn't change, next boot with this fix will find that record and skip calling `upgrade()` entirely — corrected logic never runs for any existing install. Only brand-new installs benefit. This directly undercuts the PR's stated goal ("deleted-group cleanup never ran for anyone on Mongo" — still won't, for existing deployments).

Rest of 18821 clean: `storageEngines` field genuinely correct per Mongo `buildInfo` docs, `majorVersion >= 5` guard unaffected, tests realistic.

aditya-goyal01 (approved):

PR #18822, opened 2026-07-29
fix: handle requests without a Host header in the gateway dispatcher

callaertanthony:
## Description

APIM-14814 — Debug mode fails with an unhandled `NullPointerException` in the gateway when the request has no `Host` header.

Vert.x 5 returns `null` from `HttpServerRequest.authority()` when the request carries neither a `Host` header nor an HTTP/2 `:authority` pseudo-header. Three call sites dereferenced it unconditionally:

- `DefaultHttpRequestDispatcher.dispatch` (the one in the reported stack trace)
- `io.gravitee.gateway.reactive.http.vertx.VertxHttpServerRequest.hostWithPort`
- `io.gravitee.gateway.http.vertx.VertxHttpServerRequest.hostWithPort`

The dispatcher is what blows up first, but fixing it alone just moves the NPE into `prepareExecutionContext`, so all three are handled.

The fallback is `null`, which is exactly what Vert.x 4 `host()` returned for a request with no `Host` header. `DefaultHttpAcceptor.matchHost` already accepts a null host for context-path-only APIs, so a debug request against `/debug/` now resolves normally; a request with no Host against a vhost API still falls through to not-found, which is correct.

## Tests

- `DefaultHttpRequestDispatcherTest`: dispatching a request with a null authority resolves with a null host instead of throwing.
- `VertxHttpServerRequestTest.HostsTest`: `host()` / `originalHost()` return null instead of throwing.

## Backport

`apply-on-master`.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/http/vertx/VertxHttpServerRequest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/http/vertx/VertxHttpServerRequest.java
@@ -258,8 +258,14 @@ public class VertxHttpServerRequest implements Request {
      * In Vert.x 5, {@code authority().host()} returns only the hostname without the port, which would cause the port
      * to be silently dropped in URLs built from {@code originalHost()} (e.g. in {@code XForwardProcessor} when
      * computing {@code ContextAttributes.ATTR_REQUEST_ORIGINAL_URL}).
+     * <p>
+     * The authority is {@code null} when the request carries neither a {@code Host} header nor an HTTP/2
+     * {@code :authority} pseudo-header. Vert.x 4 returned {@code null} in that case as well.
      */
     private static String hostWithPort(HostAndPort authority) {
+        if (authority == null) {
+            return null;
+        }
         int port = authority.port();
         return port > 0 ? authority.host() + ":" + port : authority.host();
     }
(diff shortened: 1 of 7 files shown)


aditya-goyal01:
## PR #18822 — Gateway crashes on requests with no Host header (in one config)

**What the PR does:** Fixes a crash (NullPointerException) that happens when a request arrives with no `Host` header — this happens routinely in Debug mode. The fix makes the gateway treat a missing host as `null` instead of crashing.

**Confirmed issue:** The fix trades one crash for a narrower one.

If a gateway has the setting `allowOverlappingApiContexts` turned on, **and** at least one deployed API is scoped to a specific virtual host, the code that decides which API a request matches assumes the host is never `null`. It calls a string method directly on it — so a hostless request still crashes, just one step later, in a different file.

- Confirmed by direct code reading, not just PR description.
- No test covers this exact combination — that's a real, verified gap.
- Only reproduces with a non-default setting, so scope is narrow but real.

Where: `OverlappingHttpAcceptor.matchHost()` — [gravitee-apim-gateway-reactor/.../DefaultHttpRequestDispatcher.java#L145-L157](https://github.com/gravitee-io/gravitee-api-management/blob/e61687df239db1d2aca9d67268826d1671ebda3f/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java#L145-L157)

**Verdict: Must resolve before merge.** The whole point of this PR is "requests with no Host header should not crash the gateway." As written, they still can, just under one more condition.

---

aditya-goyal01 (approved):

PR #18823, opened 2026-07-29
fix: build the gRPC endpoint client from a copy of the shared options

callaertanthony:
## Description

APIM-14820 — a single inbound request carrying `Content-Type: application/grpc` permanently breaks an HTTP/1.1 API on a gateway node.

`GrpcHttpClientFactory.buildHttpClient` did:

```java
HttpClientOptions httpOptions = sharedConfiguration.getHttpOptions(); // shared instance
httpOptions.setVersion(ProtocolVersion.HTTP_2);
httpOptions.setClearTextUpgrade(false);
```

`sharedConfiguration` is the endpoint group's shared configuration — the same object `HttpClientFactory` (the plain, non-gRPC one) reads from on every `getOrBuildHttpClient`. Because both factories are held by the same `HttpProxyEndpointConnector`, forcing HTTP/2 there leaks into the plain client: if the gRPC-classified request is the first the API sees on a fresh node, the plain client is then built as HTTP/2 too, and every subsequent request to a HTTP/1.1-only backend fails with a 502 (`First received frame was not SETTINGS`) until the pod restarts.

Classification is header-only, so any anonymous caller can trigger this against a keyless API.

The fix maps the shared options into a fresh `VertxHttpClientOptions` first and forces the version on that copy — the shared configuration is never mutated.

## Tests

Unit:

- `should_not_mutate_the_shared_http_options_when_forcing_http_2` — the regression guard: the shared `HttpClientOptions` still reads `HTTP_1_1` / `clearTextUpgrade=true` after a gRPC client is built.
- `should_build_the_grpc_client_over_http_2_without_clear_text_upgrade` — the gRPC client itself is still HTTP/2 without clear-text upgrade.
- Full `gravitee-apim-plugin-endpoint-http-proxy` suite green (63 tests).

Gateway integration, both against an ordinary v4 HTTP proxy API whose target is a plain `http://` backend, asserting the protocol the backend actually saw:

- `GrpcSharedClientOptionsIsolationV4IntegrationTest` — sends the gRPC-flagged request first, then a plain one, and asserts the plain one is still HTTP/1.1. Reverting `GrpcHttpClientFactory` to the previous code makes it fail with `expected "HTTP/1.1" but was "HTTP/2.0"`.
- `GrpcContentTypeOnHttpTargetV4IntegrationTest` — records the detection rule itself (see below). It passes both before and after this fix, on purpose.

They live in separate classes because the gateway is deployed once per class and `HttpClientFactory` caches its client for the endpoint's lifetime — a plain request made by an earlier test method builds a clean client and masks what a later method is trying to observe.

## Not included: the hardening change

The ticket also proposes gating `isGrpc()` classification on the endpoint being configured for HTTP/2, which would close the remote trigger entirely. That is a behaviour change, so it needs a product decision and is deliberately left out of this fix.

What the new characterization test measures, on an endpoint with a plain `http://` target and no HTTP/2 in its configuration:

| Request | Protocol the backend saw |
|---|---|
| no `Content-Type` | HTTP/1.1 |
| `Content-Type: application/grpc` | HTTP/2.0 |

So the caller, not the configuration, decides the protocol used to reach the backend. Against a backend that accepts h2c the call still succeeds; against one that does not, it fails.

Evidence that the `http://`-target-plus-header combination was never a supported way to configure a gRPC endpoint:

- the `gravitee-apim-plugin-endpoint-http-proxy` README does not mention gRPC at all;
- all 11 gRPC fixtures in this repository — v2 and v4 integration tests, the gateway test SDK, the definition and export fixtures — use a `grpc://` target;
- before this PR, no test exercised the `Content-Type` branch of `isGrpc()`;
- the public gRPC tutorial and the v4 endpoint reference prescribe `grpc://host:port` **and** HTTP/2 on the endpoint, in every version from 4.4 to current;
- the branch arrived inside a broad 2023 protocol-support commit (`85d2314294`) and has since only been touched for a parse-exception guard and a caching change.

Every configuration we document or test satisfies both conditions the hardening would require, so it should be invisible to users following the documentation. If it is taken, it belongs on `master` only — and `GrpcContentTypeOnHttpTargetV4IntegrationTest` is what will flag it.

## Backport

`apply-on-4-11-x`, `apply-on-4-12-x`, `apply-on-master`.

### DIFF
--- a/gravitee-apim-integration-tests/src/test/java/io/gravitee/apim/integration/tests/grpc/v4/AbstractHttpTargetGrpcV4GatewayTest.java
+++ b/gravitee-apim-integration-tests/src/test/java/io/gravitee/apim/integration/tests/grpc/v4/AbstractHttpTargetGrpcV4GatewayTest.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.integration.tests.grpc.v4;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.gravitee.apim.gateway.tests.sdk.AbstractGatewayTest;
+import io.gravitee.apim.gateway.tests.sdk.connector.EndpointBuilder;
+import io.gravitee.apim.gateway.tests.sdk.connector.EntrypointBuilder;
+import io.gravitee.plugin.endpoint.EndpointConnectorPlugin;
+import io.gravitee.plugin.endpoint.http.proxy.HttpProxyEndpointConnectorFactory;
+import io.gravitee.plugin.entrypoint.EntrypointConnectorPlugin;
+import io.gravitee.plugin.entrypoint.http.proxy.HttpProxyEntrypointConnectorFactory;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.rxjava3.core.http.HttpClient;
+import java.util.Map;
+
+/**
+ * Shared setup for the tests that send gRPC-flagged requests to a plain {@code http://} target.
+ *
+ * <p>Each of those tests lives in its own class on purpose. The gateway is deployed once per class and
+ * {@code HttpClientFactory} caches the client it builds for the lifetime of the endpoint, so a plain request made by
+ * an earlier test would build a clean client and hide what a later test is trying to observe.
+ */
+abstract class AbstractHttpTargetGrpcV4GatewayTest extends AbstractGatewayTest {
+
+    protected static final String GRPC_CONTENT_TYPE = "application/grpc";
+
+    @Override
+    public void configureEntrypoints(Map<String, EntrypointConnectorPlugin<?, ?>> entrypoints) {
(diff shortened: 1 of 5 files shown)


ThibaudAV:
### Code review

Reviewed the diff, the parent `HttpClientFactory`, and the mapper. **The production fix is correct and well-targeted** — no correctness issue found in the change itself.

I specifically checked the reordering risk: previously the shared options were mutated *before* the `super.buildHttpClient(...)` call, now a fresh copy is mutated *after* it. `super.buildHttpClient` only sets `httpOptions`/`sslOptions`/`proxyOptions` on the builder — nothing else depends on the version — and the final `.httpOptions(...)` fully overrides it, so the reorder introduces no behavioral difference. The mapper produces a fresh `VertxHttpClientOptions`, so the shared configuration is no longer mutated. 👍

A few non-blocking, test-quality notes:

1. **Both integration tests silently depend on the WireMock backend accepting h2c (prior-knowledge).** The gRPC client is built with `clearTextUpgrade=false` + `HTTP_2`, so for the gRPC-flagged call to return `200` and record `HTTP/2.0` the backend must speak h2c — which is exactly the "permissive backend masks the bug" condition called out in the ticket, and it's load-bearing for the assertions. If a future Jetty/WireMock bump disables h2c, `GrpcSharedClientOptionsIsolationV4IntegrationTest` (the regression guard) would fail for an unrelated reason, or worse turn into a false pass. A short comment making this dependency explicit would help.

2. **`should_build_the_grpc_client_over_http_2_without_clear_text_upgrade`** reads the builder's private `httpOptions` field via `ReflectionTestUtils` — brittle to a Lombok field rename (would surface as `assertNotNull` failing). Consistent with the existing `HttpClientFactoryTest`, so acceptable, but it couples the test to the implementation.

3. **Nit (not a regression):** the mapper now runs twice per gRPC client build — once in `super.buildHttpClient` (result discarded) and once here. This was already the case before the PR; negligible impact since it happens once per endpoint lifetime.

For the record, the DoS vector is only half-closed — by design: a single `Content-Type: application/grpc` request to an HTTP/1.1-only backend still fails on its own request since `isGrpc()` remains header-only. That matches the deliberately-deferred hardening change. Looks good to merge.

ThibaudAV (approved):

PR #18824, opened 2026-07-29
fix: align command delete order to avoid a PostgreSQL deadlock

callaertanthony:
## Description

APIM-14720 — intermittent `deadlock detected` between `CommandServiceImpl.delete()` and `CommandServiceImpl.deleteByExpiredAtBefore()`, on a single Management API pod against PostgreSQL.

`JdbcCommandRepository` has two delete paths that touch the same three tables in **opposite** orders:

| method | order |
|---|---|
| `delete(id)` | `commands` → `command_acknowledgments` → `command_tags` |
| `deleteByExpiredAtBefore(before)` | `command_acknowledgments` → `command_tags` → `commands` |

They are driven by two independently scheduled beans in the same JVM (`ScheduledSearchIndexerService` and `ScheduledCommandsRefresherServiceImpl`). When their transactions overlap, each holds a lock the other needs and PostgreSQL kills one of them. This does not need multiple pods, and it reproduces with an almost-empty `commands` table.

`delete(id)` now deletes the child rows first, matching `deleteByExpiredAtBefore`. Child-then-parent is also the FK-safe order, so nothing else changes.

`deleteByEnvironmentId` / `deleteByOrganizationId` only touch `commands`, so they cannot take part in a cross-table cycle and are left alone.

## Tests

`JdbcCommandRepositoryTest` asserts, with `InOrder` on a mocked `JdbcTemplate`, that both methods issue their deletes as acknowledgments → tags → commands. Checked against the previous ordering: the first test fails on it, so it is a real regression guard rather than a tautology.

## Backport

`apply-on-4-10-x`, `apply-on-4-11-x`, `apply-on-4-12-x`, `apply-on-master`.

Reported on 4.8.28+, but 4.8.x has no mergify rule, so 4.9.x is the oldest branch this lands on.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcCommandRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcCommandRepository.java
@@ -141,9 +141,12 @@ public class JdbcCommandRepository extends JdbcAbstractCrudRepository<Command, S
     public void delete(String id) throws TechnicalException {
         log.debug("JdbcCommandRepository.delete({})", id);
         try {
-            jdbcTemplate.update(getOrm().getDeleteSql(), id);
+            // Delete the child rows before the command itself, in the same table order as
+            // deleteByExpiredAtBefore. Both methods run concurrently from two independently scheduled
+            // services, and taking the row locks in opposite orders is what made PostgreSQL deadlock.
             jdbcTemplate.update("delete from " + COMMAND_ACKNOWLEDGMENTS + " where command_id = ?", id);
             jdbcTemplate.update("delete from " + COMMAND_TAGS + " where command_id = ?", id);
+            jdbcTemplate.update(getOrm().getDeleteSql(), id);
         } catch (final Exception ex) {
             log.error("Failed to delete command:", ex);
             throw new TechnicalException("Failed to delete command", ex);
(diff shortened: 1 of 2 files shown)


aditya-goyal01:
## PR #18824 — Database delete fix reopens an old bug elsewhere

**What the PR does:** Two places in the code delete a "command" record and its related rows (acknowledgments, tags) — but they used to do it in opposite table order. When both run at nearly the same time in production, Postgres can deadlock (each one is waiting on a lock the other is holding). The fix makes both methods delete in the same order, so they stop conflicting.

**Confirmed issue:** There's a third method (`update()`, used when a command is acknowledged) that still deletes those same tables in the *original*, opposite order. Nobody touched it. So instead of two methods conflicting, now a different pair — the newly-fixed `delete()` and the untouched `update()` — can conflict instead. Same deadlock, different pair of methods.

- Confirmed a background service on every server independently decides, every few seconds, whether to delete or acknowledge each command — so two different servers really can pick opposite methods for the same record at nearly the same moment.
- Confirmed there was already a fix for this exact class of problem back in 2023 (commit `892d84915829`, ticket APIM-3113) — this PR quietly undoes that alignment for the `update()`/`delete()` pair, even though it wasn't touching `update()` on purpose.

Where: [JdbcCommandRepository.java#L140-L154](https://github.com/gravitee-io/gravitee-api-management/blob/4b4ec3c52543be57a2b6a4e0622c318d277f3c12/gravitee-apim-repository/gravitee-apim-repository-jdbc/src/main/java/io/gravitee/repository/jdbc/management/JdbcCommandRepository.java#L140-L154)

**Verdict: Must resolve before merge.** Fixing one deadlock pair while opening a different one isn't a net fix — `update()` needs to be reordered to match, or the whole codebase needs one agreed-upon table order for these three tables.

---

aditya-goyal01 (approved):

PR #18826, opened 2026-07-29
fix: report no health check data when no v4 repository is registered

callaertanthony:
## Description

APIM-14383 — opening the v4 endpoint health check dashboard with analytics disabled pops a raw Spring error:

> No qualifying bean of type 'io.gravitee.repository.healthcheck.v4.api.HealthCheckRepository' available […] Dependency annotations: {@Lazy(true)}

`ApiHealthQueryServiceImpl` took the repository as a hard `@Lazy` constructor dependency. `@Lazy` only defers the resolution to the first call, so when no bean is contributed the failure lands on the user as a broken dashboard instead of at startup. Same shape as the API Logs page fix in gravitee-io/issues#11297.

The repository is now resolved through an `ObjectProvider`, and an absent one yields an absent result. All four methods already return `Maybe`, so callers cannot tell the difference between "no repository" and "repository with no data" — which is exactly the behaviour of `NoOpHealthCheckRepository`.

## A caveat worth reviewing

I could not reproduce the missing bean from the code on any maintained branch: `NoOpAnalyticsRepositoryConfiguration` does register a v4 `healthCheckRepositoryV4` bean on 4.9.x → master, so a plain `analytics.type=none` should be served by the no-op repository. The reporter's environment is listed only as "4.10.x", so the exact trigger (older build, a different analytics repository plugin that ships no v4 health check implementation, or a plugin packaging problem) is still open.

What this PR guarantees is that *whatever* the reason no v4 health check repository is present, the dashboard degrades to "no data" instead of surfacing Spring internals. If we want the dashboard to say "analytics are disabled" explicitly rather than show an empty chart, that is a separate product decision and a different change.

## Tests

`ApiHealthQueryServiceImplTest.WithoutHealthCheckRepository` — with no repository available, `averageResponseTime`, `availability` and `averageResponseTimeOvertime` complete empty instead of throwing. Existing tests unchanged (5/5 green).

## Backport

`apply-on-4-11-x`, `apply-on-4-12-x`, `apply-on-master`.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/api_health/ApiHealthQueryServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/api_health/ApiHealthQueryServiceImpl.java
@@ -26,51 +26,69 @@ import io.gravitee.repository.common.query.QueryContext;
 import io.gravitee.repository.healthcheck.v4.api.HealthCheckRepository;
 import io.reactivex.rxjava3.core.Maybe;
 import java.util.ArrayList;
-import org.springframework.context.annotation.Lazy;
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.stereotype.Service;
 
 @Service
 public class ApiHealthQueryServiceImpl implements ApiHealthQueryService {
 
-    private final HealthCheckRepository healthCheckRepository;
+    private final ObjectProvider<HealthCheckRepository> healthCheckRepository;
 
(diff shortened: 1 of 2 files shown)


callaertanthony:
Superseded by #18830, which fixes the actual cause.

The bean *is* declared in `NoOpAnalyticsRepositoryConfiguration`, so I could not explain the failure statically and this PR only made the dashboard degrade gracefully. The reason it is missing at runtime is that `RepositoryPluginHandler` publishes a plugin bean into the host context only when its **name** ends with `Repository`, and the method was `healthCheckRepositoryV4()` — so the bean was built and then left behind in the plugin context.

Closing rather than merging on top: making `ApiHealthQueryServiceImpl` tolerate a missing repository would have turned this into a silently empty dashboard, which is precisely what made the bug hard to track down. If we want graceful degradation for a genuinely absent repository, it deserves its own change.

PR #18827, opened 2026-07-29
fix: keep API key values out of audit records

callaertanthony:
## Description

APIM-14809 — subscription API keys were stored in clear text in Management API audit records, in two places:

1. the `API_KEY` audit **property**, set to the full key string;
2. the audit **patch**, because the serialised `ApiKeyEntity` carries its `key` field.

Anyone with the audit read permission could therefore lift a usable credential out of an `APIKEY_CREATED` / `RENEWED` / `REVOKED` / `EXPIRED` / `REACTIVATED` event and replay it against the gateway.

Three call sites wrote these audits:

- `ApiKeyServiceImpl.createAuditLog` (v3 service, all lifecycle events)
- `GenerateApiKeyDomainService.createAuditLog`
- `RevokeApiKeyDomainService.createAuditLog`

All three now put the **API key id** in the `API_KEY` property and audit a copy of the entity with `key` cleared. Everything that makes the record useful — key id, application, API / API product, event, actor, timestamp — is untouched.

### Why clear the field rather than use `pathsToAnonymize`

`AuditService.AuditLogData` has a `pathsToAnonymize` hook that rewrites a JSON-patch value to `*****`. I did not use it: it only exists on the v3 audit path (the core `ApiAuditLogEntity` has no equivalent), and it matches on an exact patch path, so it would silently stop covering the key if the payload shape ever changed. Clearing the field before serialisation cannot miss, and it works identically on both paths. Happy to switch if you would rather see an explicit `*****` in the diff.

### Scope note

This redacts records written from now on. **Audit records already in the database keep their plaintext keys.** Rewriting them is deliberately not part of this PR: it would mean mutating an audit trail, and the exposure is better addressed by rotating any key that could have been read from an old audit record than by editing history. No migration is planned.

## Tests

- `ApiKeyServiceTest` — the `API_KEY` property is the key id, the plaintext key appears in no property, and the audited new value carries no `key`.
- `GenerateApiKeyDomainServiceTest.should_keep_the_api_key_value_out_of_the_audit` — end to end over the in-memory audit store: neither the properties nor the patch contain the key value.
- The existing audit assertions in the revoke use cases and domain service were asserting the leak (`API_KEY` == key value); they now assert the id.
- Full `*ApiKey*` suite green (225 tests).

## Backport

`apply-on-master`.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_key/domain_service/GenerateApiKeyDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_key/domain_service/GenerateApiKeyDomainService.java
@@ -207,11 +207,15 @@ public class GenerateApiKeyDomainService {
         String referenceId = isApiProduct ? subscription.getReferenceId() : subscription.getApiId();
 
         Map<AuditProperties, String> properties = new HashMap<>();
-        properties.put(AuditProperties.API_KEY, createdApiKeyEntity.getKey());
+        // Audit records are readable by anyone holding the audit permission, so they identify the API key by its
+        // id: the key value itself is a credential that can be replayed against the gateway.
+        properties.put(AuditProperties.API_KEY, createdApiKeyEntity.getId());
         if (referenceId != null) {
             properties.put(isApiProduct ? AuditProperties.API_PRODUCT : AuditProperties.API, referenceId);
         }
         properties.put(AuditProperties.APPLICATION, subscription.getApplicationId());
+
+        var auditedApiKey = createdApiKeyEntity.toBuilder().key(null).build();
 
         if (isApiProduct) {
             auditService.createApiProductAuditLog(
@@ -222,7 +226,7 @@ public class GenerateApiKeyDomainService {
                     .event(event)
                     .actor(auditInfo.actor())
                     .oldValue(null)
-                    .newValue(createdApiKeyEntity)
+                    .newValue(auditedApiKey)
                     .createdAt(createdApiKeyEntity.getCreatedAt())
                     .properties(properties)
                     .build()
@@ -236,7 +240,7 @@ public class GenerateApiKeyDomainService {
                     .event(event)
                     .actor(auditInfo.actor())
                     .oldValue(null)
-                    .newValue(createdApiKeyEntity)
+                    .newValue(auditedApiKey)
                     .createdAt(createdApiKeyEntity.getCreatedAt())
                     .properties(properties)
                     .build()
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_key/domain_service/RevokeApiKeyDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_key/domain_service/RevokeApiKeyDomainService.java
@@ -118,11 +118,16 @@ public class RevokeApiKeyDomainService {
         String referenceId = isApiProduct ? subscription.getReferenceId() : subscription.getApiId();
 
         Map<AuditProperties, String> properties = new HashMap<>();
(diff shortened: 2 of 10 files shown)


ThibaudAV:
### Code review

Solid fix, and I like that the `API_KEY` property now carries the key **id** rather than a blanket `*****` — that keeps the record traceable (you can tell *which* key an event refers to) while removing the secret. I checked completeness: the three audit-writing sites are the only ones. On the v4 side there are exactly two api-key domain services (`GenerateApiKeyDomainService`, `RevokeApiKeyDomainService`), both patched; on v3, `ApiKeyServiceImpl.createAuditLog` is a single method covering every lifecycle event (RENEW / EXPIRE / REACTIVATE included). No leftover leak. The `toBuilder().key(null)` copy also leaves the entity returned to callers untouched (test confirms `result.getKey()` is intact). 👍

Two points:

**1. Backport scope — this looks under-labelled for a security fix.** The PR is `apply-on-master` only, but this is a *Public Security* issue and the plaintext-key code is present on every maintained branch (I confirmed it on 4.11.x, and the ticket reports it on 4.12.x). Its siblings this week (#18828, #18830) went to 4.10.x/4.11.x/4.12.x. As it stands, master-only leaves the supported/LTS releases exposed. I'd add `apply-on-4-11-x` / `apply-on-4-12-x` (and 4.10.x if still supported) unless there's a reason to hold them back.

**2. `key(null)` vs an explicit `*****` in the patch.** For the property, the id is the right call — no change. For the serialised entity, clearing to `null` means the field simply vanishes from the diff (or shows as `"key": null` depending on the audit ObjectMapper's null handling), which a reader could misread as "no key was set" rather than "redacted". Your reasoning for avoiding `pathsToAnonymize` is sound (v3-only, exact-path match, silently drifts), but that argument is specifically against the *path-matching* hook — it doesn't force `null`. You can keep the exact same clear-before-serialise approach and set `key("*****")` (or a `REDACTED` constant) instead of `null`: same robustness (works on both paths, can't miss), but the redaction becomes visible and self-documenting in the diff. That also matches the ticket's own wording ("replaced with `*****` or omitted"). Minor/optional — but since the PR explicitly asked, I'd lean toward the visible marker.

Everything else looks good. Scope note about not rewriting historical records is the right call.

aditya-goyal01:
## Strengths

- All 3 audit call sites (`GenerateApiKeyDomainService`, `RevokeApiKeyDomainService`, `ApiKeyServiceImpl`) correctly switched the `API_KEY` audit property from key value to key id — full coverage, no missed caller.
- key field correctly nulled on the audited entity copy before serialization; serialization path respects it.
- New test `GenerateApiKeyDomainServiceTest.should_keep_the_api_key_value_out_of_the_audit` genuinely asserts raw key absent from both properties and patch.
- Notification hook (`ApiKeyRevokedApiHookContext`) correctly left with the real key — that's email notification, not audit, so untouched is right.
- No schema/migration change; no reliance on new redacted format when reading old audits.

## Issues

**Critical (Must Fix)**
- `ApiKeyServiceImpl.java`:815 — createAuditLog does `key.toBuilder().key(null).build()` but never clears hash (`ApiKeyEntity.hash`, `MD5` of the raw key, doc'd as "Used as username for Native Kafka API Authentication"). No @JsonIgnore on it, so it serializes straight into the audit patch for every `CREATED/RENEWED/REACTIVATED/EXPIRED` event. Reintroduces a real credential leak — same class of bug the fix targets. Fix: `key.toBuilder().key(null).hash(null).build()`.

**Important (Should Fix)**
- `RevokeApiKeyDomainServiceTest.java`:182-195, 220-235 — comparator excludes patch field entirely, so the revoke path's redaction (correct by code inspection) has zero regression coverage.
- `ApiKeyServiceTest.java` — `shouldRenew`, `shouldReactivateRevoked`, `shouldReactivateExpired`, `shouldUpdateExpired` have no key/hash-absence assertions on the captured audit; only `shouldGenerate` does.

**Recommendations**

- Single "sanitize for audit" helper `key(null).hash(null)` instead of repeating `toBuilder().key(null)` per call site — avoids this exact miss if another key-derived field is added later.

aditya-goyal01 (approved):

PR #18828, opened 2026-07-29
fix: make promotion createOrUpdate tolerate a concurrent insert

callaertanthony:
## Description

APIM-14594 — promoting an API fails with `DuplicateKeyException: Violation of PRIMARY KEY constraint 'pk_apim_promotions'`.

`PromotionServiceImpl.createOrUpdate` is a check-then-act:

```java
final Optional<Promotion> existingPromotion = promotionRepository.findById(promotionEntity.getId());
...
if (existingPromotion.isPresent()) { update } else { create }
```

The lookup and the insert are not atomic. The environment that starts the promotion (`create`) and the Cockpit bridge that receives it (`createOrUpdate`) write the same promotion id, and when both sit on the same database the `findById` can run before the other transaction commits. The reporter's SQL trace shows exactly that, two SPIDs on the same primary node, 37 ms apart.

The second writer then takes the `create` branch and blows up. The promotion stays in error and retries with a new id, failing the same way — which is why deleting the offending row does not help.

`create` now falls back to `update` when it conflicts **and** a second lookup confirms the row is there. If the row still is not there, or the failure was something else, it propagates unchanged.

### Why the catch is broad

The conflict does not surface the same way on every repository implementation, so pinning `org.springframework.dao.DuplicateKeyException` would only fix half the problem:

| Repository | What `create()` throws on a PK violation |
|---|---|
| JDBC (the reported case) | `JdbcAbstractCrudRepository.create` catches `Exception` and rethrows a checked `TechnicalException` with the `DuplicateKeyException` as its cause |
| Mongo | the driver's unchecked `DuplicateKeyException` escapes directly |

Catching `Exception` covers both. The `findById` re-check is what keeps it specific: nothing is swallowed unless a promotion with that id genuinely exists.

## Tests

Mongo shape — the unchecked exception escapes `create()`:

- `shouldUpdateWhenThePromotionWasCreatedConcurrently` — first lookup empty, insert conflicts, second lookup finds the row → one `create`, one `update`, no exception.
- `shouldRethrowWhenTheCreateFailureIsNotAConflict` — insert fails and the row is still absent → the original exception propagates.

JDBC shape — the failure arrives wrapped in a `TechnicalException`:

- `shouldUpdateWhenTheConcurrentCreateFailureIsWrappedInATechnicalException` — same fallback through the wrapped conflict.
- `shouldRethrowWhenTheWrappedCreateFailureIsNotAConflict` — a wrapped non-conflict failure surfaces as `TechnicalManagementException` via the existing outer catch.

`PromotionServiceTest` green, 27 tests.

## Backport

`apply-on-4-12-x`, `apply-on-master`.

The ticket carries no reported version, so this targets 4.11.x as the current LTS. The code is identical on 4.9.x and 4.10.x — say the word if the backport should be widened.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/promotion/PromotionServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/promotion/PromotionServiceImpl.java
@@ -238,7 +238,7 @@ public class PromotionServiceImpl extends AbstractService implements PromotionSe
                 createdOrUpdatedPromotion = promotionRepository.update(promotion);
             } else {
                 log.debug("Creating promotion: {}", promotion.getId());
-                createdOrUpdatedPromotion = promotionRepository.create(promotion);
+                createdOrUpdatedPromotion = createOrUpdateOnConflict(promotion);
             }
 
             return convert(createdOrUpdatedPromotion);
@@ -247,6 +247,24 @@ public class PromotionServiceImpl extends AbstractService implements PromotionSe
                 "An error occurs while trying to create or update a promotion using its id {}" + promotionEntity.getId(),
                 e
             );
+        }
+    }
+
+    /**
+     * The existence check above and the insert are not atomic. A promotion is written concurrently by the
+     * environment that starts it and by the Cockpit bridge that receives it, and when both share a database the
+     * check can miss a row the other transaction has not committed yet, which surfaced as a raw primary key
+     * violation. Treat the conflict as what it means - somebody else got there first - and update instead.
+     */
+    private Promotion createOrUpdateOnConflict(Promotion promotion) throws TechnicalException {
+        try {
+            return promotionRepository.create(promotion);
+        } catch (Exception e) {
+            if (promotionRepository.findById(promotion.getId()).isEmpty()) {
+                throw e;
+            }
+            log.debug("Promotion {} was created concurrently, updating it instead", promotion.getId());
+            return promotionRepository.update(promotion);
         }
     }
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/promotion/PromotionServiceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/promotion/PromotionServiceTest.java
@@ -77,6 +77,7 @@ import org.junit.runner.RunWith;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
+import org.springframework.dao.DuplicateKeyException;
 
 /**
  * @author Florent CHAMFROY (florent.chamfroy at graviteesource.com)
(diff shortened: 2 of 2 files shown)


ThibaudAV:
### Code review

The approach is sound and the fix is correct: catch the failed insert, re-check by id, and fall back to `update` only when the row genuinely exists. I checked the transactional model too — `PromotionServiceImpl` is **not** `@Transactional` and each repository call runs on its own `jdbcTemplate` (autocommit), so the failed INSERT does not poison a surrounding transaction and the re-check + `update` execute cleanly. Keeping `catch (Exception)` broad is the right call here (see point 2 below), and the `findById` re-check keeps it from swallowing anything real. 👍

A few points worth a look, one of which touches the root-cause narrative:

**1. Root cause on the JDBC/SQL Server path — the raw `DuplicateKeyException` claim.** The description says the exception *"sails straight past the `catch (TechnicalException)` and reaches the bridge raw"*. On the JDBC path (the customer's SQL Server) that is not what happens: `JdbcAbstractCrudRepository.create` wraps **any** failure into a checked `TechnicalException`:

```java
try {
    jdbcTemplate.update(buildInsertPreparedStatementCreator(item));
    return findById(getId(item)).orElse(null);
} catch (final Exception ex) {
    throw new TechnicalException("Failed to create ... item.", ex); // DuplicateKeyException becomes the cause
}
```

So before this fix, on JDBC the conflict would surface as `TechnicalManagementException` (via the outer `catch (TechnicalException)`), not a raw `DuplicateKeyException`. Two implications: (a) the "reaches the bridge raw" explanation likely holds only for a Mongo-style repo, not the SQL Server case in the ticket — worth reconciling with the customer's actual stack; (b) it confirms why the broad `catch (Exception)` in the fix is necessary — on JDBC the caught type is `TechnicalException`, not `DuplicateKeyException`. The fix is correct either way; it's the narrative I'd double-check.

**2. Test fidelity vs the real repository.** Both new tests throw `org.springframework.dao.DuplicateKeyException` straight out of `create()`, but the JDBC repo (the one in the incident) never throws that — it throws `TechnicalException` wrapping it. So `shouldRethrowWhenTheCreateFailureIsNotAConflict` asserting a *raw* `DuplicateKeyException` escapes only models the Mongo-shaped path; on JDBC a genuine non-conflict failure comes out as `TechnicalManagementException`. Consider adding (or substituting) a `TechnicalException` case so the primary path in the ticket is the one under test.

**3. Semantics: unconditional `update` on conflict = last-writer-wins.** When the insert conflicts, the fix overwrites the concurrently-created row with *our* payload. This matches the pre-existing `update` branch, so it's consistent — but under the race the winner is nondeterministic, so if the two writers carry different promotion state (e.g. one already advanced the status), the fallback `update` can clobber the more-advanced row. Probably fine for promotions, but worth a conscious "last write wins is acceptable here" rather than leaving it implicit.

**4. Very minor.** If the row is found by the re-check but deleted before `update` runs, `update` may return `null` and `convert(null)` would NPE. Not reachable in practice for promotions (nothing deletes them mid-flight), just noting the second non-atomic window the fix opens.

None of these block merge — the fix resolves the reported PK violation correctly. Points 1 and 2 are the ones I'd act on (align the narrative / test with the JDBC path).

aditya-goyal01 (approved):

aditya-goyal01:
**PR 18828** — promotion createOrUpdate race fix. Clean. No findings ≥80 confidence. Broad `catch(Exception)` justified (JDBC wraps in checked TechnicalException, Mongo throws unchecked) and gated by second `findById` check. Tests exercise both repo shapes properly.

PR #18830, opened 2026-07-30
fix: name the no-op v4 health check bean so it reaches the host context

callaertanthony:
## Description

APIM-14383 — with `analytics.type=none`, opening the v4 endpoint health check dashboard fails with:

> No qualifying bean of type 'io.gravitee.repository.healthcheck.v4.api.HealthCheckRepository' available […] Dependency annotations: {@Lazy(true)}

### Root cause

Repository plugins are instantiated in their own Spring context. `RepositoryPluginHandler` (gravitee-plugin-repository) then copies beans into the host context, but **only those whose bean name ends with `Repository`** — or `TransactionManager`:

```java
for (String beanName : pluginContext.getBeanDefinitionNames()) {
    Object bean = pluginContext.getBean(beanName);
    if (beanName.endsWith("Repository")
        || (beanName.endsWith("TransactionManager") && !provider.getClass().equals(bean.getClass()))) {
        if (bean.getClass().getInterfaces().length > 0) {
            beanFactory.registerSingleton(beanName, bean);
        }
    }
}
```

A `@Bean` method names its bean after itself, and ours was `healthCheckRepositoryV4()` — ending in `V4`. The no-op v4 health check repository was built on every startup and then silently left behind in the plugin context. Its neighbours `analyticsV4Repository()` and `logV4Repository()` follow the convention and are published normally, which is why only this one page broke.

Elasticsearch was never affected: it declares `healthCheckElasticsearchRepository(...)`, which ends correctly. That is exactly why the bug only shows up "in case of disabled ElasticSearch DB".

`@Lazy` on the `ApiHealthQueryServiceImpl` constructor parameter is what delays the failure to the first call, so this surfaces as a broken dashboard rather than a startup error.

Present since 4.9.0 (`cf7cce7ba`, the commit that introduced the bean), on every branch through master.

### Fix

Rename the method to `healthCheckV4Repository()`. One line, no behaviour change beyond the bean actually being published.

## Tests

`NoOpRepositoryBeanNamingTest` asserts every `@Bean` in the three `NoOp*RepositoryConfiguration` classes is named so the handler will publish it. Checked against the old name: it fails, naming `healthCheckRepositoryV4` — so it is a real guard, not a tautology. Full no-op suite green (65 tests).

The rule this encodes is invisible in the code today and has now cost one production bug; the test makes a bean that can never be published fail the build instead of a page at runtime.

## Note for reviewers

This supersedes #18826, which I opened earlier tonight before finding the cause. That PR made `ApiHealthQueryServiceImpl` tolerate a missing repository via `ObjectProvider`. I have closed it: it would have turned this bug into a silently empty dashboard, which is what made it hard to find in the first place. If we still want graceful degradation for a genuinely absent repository, it is worth doing deliberately and separately.

## Backport

`apply-on-4-10-x`, `apply-on-4-11-x`, `apply-on-4-12-x`, `apply-on-master`.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-noop/src/main/java/io/gravitee/repository/noop/NoOpAnalyticsRepositoryConfiguration.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-noop/src/main/java/io/gravitee/repository/noop/NoOpAnalyticsRepositoryConfiguration.java
@@ -49,7 +49,7 @@ public class NoOpAnalyticsRepositoryConfiguration {
     }
 
     @Bean
-    public io.gravitee.repository.healthcheck.v4.api.HealthCheckRepository healthCheckRepositoryV4() {
+    public io.gravitee.repository.healthcheck.v4.api.HealthCheckRepository healthCheckV4Repository() {
         return new io.gravitee.repository.noop.healthcheck.v4.NoOpHealthCheckRepository();
     }
 
--- a/gravitee-apim-repository/gravitee-apim-repository-noop/src/test/java/io/gravitee/repository/noop/NoOpRepositoryBeanNamingTest.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-noop/src/test/java/io/gravitee/repository/noop/NoOpRepositoryBeanNamingTest.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.noop;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.List;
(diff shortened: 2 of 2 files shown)


ThibaudAV:
### Code review

Root-cause analysis is spot on and the one-line rename is **correct and safe**. I confirmed:

- `RepositoryPluginHandler` only republishes beans whose name ends in `Repository`/`TransactionManager`, so `healthCheckRepositoryV4` was built and silently dropped — `healthCheckV4Repository` now ends in `Repository` and gets published. 👍
- The rename is safe: the old bean name is referenced nowhere except its own declaration — no `@Qualifier("healthCheckRepositoryV4")`, no `getBean("...")`. The consumer injects by type (`@Lazy HealthCheckRepository`), so nothing resolves this bean by name.

Two non-blocking notes, both on the guard test rather than the fix:

1. **The test encodes only the `Repository` half of the handler rule.** `RepositoryPluginHandler` also publishes names ending in `TransactionManager`, but the test asserts every `@Bean` ends in `Repository`. It passes today only because no NoOp config declares a transaction-manager bean. The day someone legitimately adds a `...TransactionManager` bean to one of these configs, this guard fails as a false negative. Mirroring the full handler predicate (`endsWith("Repository") || endsWith("TransactionManager")`) would keep the test faithful to the rule it protects.

2. **The list of config classes is hardcoded** (`NoOpAnalyticsRepositoryConfiguration`, `NoOpManagementRepositoryConfiguration`, `NoOpRateLimitRepositoryConfiguration`). A future `NoOp*RepositoryConfiguration` added elsewhere won't be covered, and a mis-named bean there would reintroduce exactly this class of bug undetected. Discovering the configs (classpath scan of `NoOp*RepositoryConfiguration`, or driving them off the same list the plugin registers) would make the guard self-maintaining.

For context, not a change request: the underlying fragility is `RepositoryPluginHandler`'s name-suffix heuristic — beans are matched by naming convention instead of by type/interface. The rename works around it correctly and is the right scope here; hardening the handler itself (publish by assignable interface) would be a broader, riskier change best tracked separately.

LGTM.

ThibaudAV (approved):

PR #18831, opened 2026-07-30
chore: build and run APIM on JDK 25

phiz71:
## Issue

N/A — JDK 25 migration.

> **Target**: this PR lands on `alpha-jdk25`, the long-lived integration branch for the migration, not on `master`. `alpha-jdk25-run-e2e` exists so the `-run-e2e` suffix triggers the e2e workflow; it carries the migration itself, not just a CI trigger. Getting `alpha-jdk25` onto `master` is a separate step: that branch holds two commits of its own (dedicated `-alpha-jdk25` versions, CI treating it as a support branch) that are not meant to travel as-is.

## Description

Moves the whole APIM toolchain to JDK 25.

### The migration itself is small

The structural change is a single bump: `gravitee-parent` `24.0.1` → `25.1.0`, which carries `java.version=25` and the `<proc>full</proc>` that JDK 23+ requires for Lombok. With that alone the whole reactor compiles and its ~37 000 unit tests pass — no Java source change was needed. JaCoCo 0.8.13, ArchUnit 1.4.2 and Lombok 1.18.46 all handle class file 69 as-is.

### What actually broke

**`gravitee-archrules-maven-plugin` was reporting a false green.** Its ArchUnit 1.2.1 parses no class file above major 66 (Java 22), so from JDK 23 onwards the importer threw on every class — the project's own and the JDK's `java.base` alike — fell back to a simple import, and the mojos still reported that all rules passed while checking nothing. Fixed upstream in the plugin (2.0.0) and pinned by `gravitee-parent` 25.1.0, so it resolves here purely by inheritance. Verified: `gioArchRules:2.0.0` runs and the build no longer emits a single `Unsupported class file major version`, against 16 865 before.

**Node moved out from under us.** `cimg/openjdk:21.0.9-node` shipped Node 22.21.1, `cimg/openjdk:25.0.3-node` ships 24.15.0, and those images carry no nvm. Nothing in the toolchain asked for Node 22 — `@angular/core` 20.3.16 declares `^20.19.0 || ^22.12.0 || >=24.0.0`, nx declares nothing — so the project's own `^22.12` was the only thing in the way. Declared as a range (`^24.0`) rather than a pinned version, so a future bump of the Node bundled in the images does not break the build again. This is coupled to the JDK bump and cannot ship before it — see the comment below on the attempt to split it out.

**The machine executors were left behind.** Bumping the `cimg/openjdk` tag only moves the Docker jobs. Every current Ubuntu LTS VM image ships OpenJDK 21 — 22.04 and 24.04 on 21.0.10, 26.04 on 21.0.11-ea — so a newer Ubuntu does not help. Added a reusable command that installs Temurin on the machine jobs that invoke Maven, resolving the major from config so the two executor families cannot drift apart again. It caches the archive, retries the transfer and verifies the published sha256 before extracting. A test over the generated config asserts the rule rather than the output: every machine job running Maven installs the JDK, and no other job does.

Also bumped `asm` 9.7.1 → 9.9.1 (9.7.1 stops at major 68) and `maven-lombok` 1.18.36 → 1.18.46 to align the delombok plugin with the version already coming from `gravitee-bom`.

## Additional context

Companion changes, all merged or released:

- `gravitee-archrules-maven-plugin` 2.0.0 — ArchUnit 1.4.2, compiled for Java 25
- `gravitee-parent` 25.1.0 — pins archrules 2.0.0
- `gravitee-docker` #245 — Java 25 Chainguard variants, so `graviteeio.azurecr.io/java:25-chainguard[-fips]` exists for the images this branch builds `FROM`

CI on this branch: 48 of 49 jobs green, including the nine e2e jobs (v3 and v4-emulation across bridge, jdbc and mongo, plus Cypress). The one failure is `aquasec/fs_scan`, which is also red on the last five commits of `master`.

Known gap: the chainguard-FIPS image jobs only run on support/master branches, so this branch never exercised them. They will first run when this lands on `alpha-jdk25`.

Verified locally on Temurin 25.0.3: full reactor build and unit tests green, `prettier` and license checks included. The only failures were environment-dependent or pre-existing and unrelated to the JDK — a locale-sensitive assertion, a DST-sensitive test, two cases of test pollution that pass in isolation, and `ApisResource_CreateApiFromSwaggerTest`, which has been red on any clean build since June 2026 because it loads a test resource its `test-jar` dependency does not package.


phiz71:
Note on scope: the Node 24 commit was briefly split out into a master-targeted PR (#18832) on the assumption it was independent of the JDK bump. It is not, and the attempt is worth recording so nobody retries it.

The Maven build shells out to `yarn` from the **openjdk** executor, and `cimg/openjdk:<jdk>-node` bundles a Node version chosen by the JDK tag — 22.21.1 on `21.0.9`, 24.15.0 on `25.0.3`. Those images carry no nvm. So on master, tightening `engines` to `^24.0` fails every Maven job that runs yarn, `gravitee-gamma` first:

```
error gravitee-apim-workspace@0.0.0: The engine "node" is incompatible with this module.
Expected version "^24.0". Got "22.21.1"
```

The two moves are coupled by the image, and Node 24 rides along with the JDK bump rather than preceding it. The Ubuntu VM images already ship Node 24.15.0, so the machine jobs are consistent with it too.

.circleci/ci/src/commands/cmd-install-jdk.ts
```diff
@@ -0,0 +1,64 @@
...
+
+JDK_HOME=/usr/lib/jvm/temurin-${feature}
+echo "Replacing $(java -version 2>&1 | head -1) with Temurin ${feature}"
+
+# x64 is hardcoded because MachineResourceClass models no Arm variant, so a
+# machine job cannot run on Arm without widening that type first.
+# Resolve the latest GA of the feature release rather than pinning a build number:
+# patch drift inside a major is harmless here, a stale pin is not.
+sudo mkdir -p "\${JDK_HOME}"
+curl -fsSL "https://api.adoptium.net/v3/binary/latest/${feature}/ga/linux/x64/jdk/hotspot/normal/eclipse" \\
```
ytvnr: No retry and no cache on a ~200 MB download.

Every machine job re-downloads the JDK from `api.adoptium.net` on every run: five jobs per PR pipeline, more once the per-database variants fan out. That is recurring CI time, and more importantly a single external dependency with no safety net. A transient Adoptium hiccup fails the job outright.

Cheapest fix, `curl --retry 3 --retry-all-errors --retry-delay 5`. Better, wrap the install in `restore_cache` / `save_cache` on `/usr/lib/jvm/temurin-25` keyed on the feature major, which also removes the download from the critical path entirely.

.circleci/ci/src/commands/cmd-install-jdk.ts
```diff
@@ -0,0 +1,64 @@
...
+  | sudo tar -xz -C "\${JDK_HOME}" --strip-components=1 --no-same-owner
```
ytvnr: No integrity check on the archive.

We extract a tarball fetched from a third-party API into `/usr/lib/jvm` as root, with nothing verifying what came down the wire. HTTPS gives transport security, not artifact provenance.

The Adoptium API exposes the sha256 alongside the binary (`/v3/assets/latest/25/hotspot`). Downloading to a temp file, checking `sha256sum -c`, then extracting is a handful of lines and closes the gap.

.circleci/ci/src/commands/cmd-install-jdk.ts
```diff
@@ -0,0 +1,64 @@
...
+  exit 0
+fi
+
+JDK_HOME=/usr/lib/jvm/temurin-${feature}
+echo "Replacing $(java -version 2>&1 | head -1) with Temurin ${feature}"
+
+# x64 is hardcoded because MachineResourceClass models no Arm variant, so a
+# machine job cannot run on Arm without widening that type first.
+# Resolve the latest GA of the feature release rather than pinning a build number:
+# patch drift inside a major is harmless here, a stale pin is not.
```
ytvnr: Worth being explicit about the consequence of `latest/ga`: two jobs in the same pipeline can end up on two different patch releases if a GA lands between them. The trade-off you describe is defensible, and I would not pin. Just noting that caching the install (see the comment on the curl line) would incidentally freeze the patch for the lifetime of the cache key, which gets you reproducibility within a pipeline for free.

.circleci/ci/src/commands/cmd-install-jdk.ts
```diff
@@ -0,0 +1,64 @@
...
+sudo mkdir -p "\${JDK_HOME}"
+curl -fsSL "https://api.adoptium.net/v3/binary/latest/${feature}/ga/linux/x64/jdk/hotspot/normal/eclipse" \\
+  | sudo tar -xz -C "\${JDK_HOME}" --strip-components=1 --no-same-owner
+
+echo "export JAVA_HOME=\${JDK_HOME}" >> "$BASH_ENV"
+echo 'export PATH=\${JAVA_HOME}/bin:\${PATH}' >> "$BASH_ENV"
```
ytvnr: Single quotes mean this line is written verbatim into `BASH_ENV` and re-evaluated at the start of every subsequent step, so `${JAVA_HOME}/bin` gets prepended to `PATH` once per step. Harmless, but `PATH` ends up with a dozen duplicate entries by the end of a job.

Since `JDK_HOME` is already known at this point, `echo "export PATH=${JDK_HOME}/bin:\$PATH" >> "$BASH_ENV"` writes it once and keeps `$PATH` resolved lazily.

.circleci/ci/src/commands/cmd-install-jdk.ts
```diff
@@ -0,0 +1,64 @@
...
+  public static get(): reusable.ReusableCommand {
+    // Track the major the Docker executors use, so both stay on the same JDK.
+    const feature = config.executor.openjdk.version.split('.')[0];
+
+    return new reusable.ReusableCommand(InstallJdkCommand.commandName, [
+      new commands.Run({
+        name: `Install JDK ${feature}`,
+        command: `if java -version 2>&1 | head -1 | grep -q '"${feature}\\.'; then
+  echo "JDK ${feature} already installed, nothing to do."
+  exit 0
```
ytvnr, line 54: The early-exit path does not set `JAVA_HOME`.

Today it does not matter, since the branch is only taken when the VM already runs the right `java`. But it makes the command idempotent on only one of its two paths: the day an Ubuntu image ships JDK 25 with a `JAVA_HOME` pointing somewhere Maven does not expect, this returns success without having established the contract the rest of the job relies on.

Deriving and exporting `JAVA_HOME` from `readlink -f $(which java)` before `exit 0` would make both paths leave the environment in the same state.

.circleci/ci/src/jobs/backend/abstract-job-test.ts
```diff
@@ -35,11 +35,21 @@ export abstract class AbstractTestJob {
     dynamicConfig.addReusableCommand(saveMavenJobCacheCmd);
     dynamicConfig.addReusableCommand(notifyOnFailureCmd);
 
+    // Docker jobs already run the right JDK through their image; machine jobs get
+    // whatever the Ubuntu VM ships, which is still 21 on every LTS image.
+    const jdkSteps: Command[] = [];
+    if (executor instanceof executors.MachineExecutor) {
```
ytvnr: Three call sites, two different styles: this one guards on `executor instanceof executors.MachineExecutor`, while `job-test-integration.ts` and `abstract-job-test-container.ts` add the command unconditionally because their executor is hardcoded. Both are correct, but the `get()` + `addReusableCommand` + `ReusedCommand` triple is now duplicated three times.

A small helper, say `withJdk(dynamicConfig, executor): Command[]` returning the steps and registering the command, would collapse the three into one and make the next machine job get the JDK by construction rather than by remembering.

ytvnr (commented):
Reviewed the whole diff. Nothing blocking, the migration itself looks clean and the verification work behind it shows.

What I checked and found sound:

- Every referenced tag/artifact actually resolves: `cimg/openjdk:25.0.3`, `:25.0.3-node`, `cimg/node:24.18.0`, `graviteeio/java:25-alpine`, `:25-debian`, and Maven Central serves `gravitee-parent:25.1.0`, `asm:9.9.1`, `lombok:1.18.46`. The Adoptium endpoint answers for 25 GA linux/x64.
- The Node bump is complete: only three `package.json` declare `engines.node` (root, `gravitee-apim-portal-webui`, `.circleci/ci`) and all three are updated. `@angular/core` is 20.3.16 everywhere, which declares `>=24.0.0`.
- `gravitee-apim-e2e/package.json` declares no `engines`, so the `^24.0` bump does not reach the e2e jobs running yarn on the VM's Node.
- All five machine+Maven jobs get `cmd-install-jdk` (rest-api, integration, plugin, reporter, repository), and the e2e jobs are correctly left out.
- No leftover Java 21 reference outside the generated snapshots, apart from two spec fixtures that pass `cimg/openjdk:21` as an arbitrary literal.

Inline comments below. The ones worth acting on before merge are the missing retry/cache on the JDK download, the missing checksum, and the absence of a regression guard on the generator.

One thing I could not anchor to a line: this PR targets `alpha-jdk25` from `alpha-jdk25-run-e2e`, yet the diff carries the full migration. If the intent is to land on `master`, it needs a retarget; if `alpha-jdk25-run-e2e` is only there to trigger an e2e run, worth saying so in the description so reviewers know what they are approving.

Also: the description says the archrules commit "drops the local override", but the diff touches no archrules version and the base pom already had none. Minor, but it sends reviewers looking for something that is not there.

.circleci/ci/src/jobs/backend/job-test-integration.ts
```diff
@@ -29,9 +29,12 @@ export class TestIntegrationJob {
     dynamicConfig.addReusableCommand(restoreMavenJobCacheCmd);
     dynamicConfig.addReusableCommand(saveMavenJobCacheCmd);
     dynamicConfig.addReusableCommand(notifyOnFailureCmd);
+    const installJdkCmd = InstallJdkCommand.get();
```
ytvnr: There is no guard against this regressing.

The updated snapshots pin the current output, but not the invariant behind it: "any machine job that runs Maven installs the JDK first". Someone adding a machine job six months from now gets a green snapshot diff and a job silently running on the VM's JDK 21.

A unit test over the generated config, asserting that every job with a machine executor whose steps contain `mvn` also contains `cmd-install-jdk`, would encode the rule itself. Feels especially worth it in a PR whose headline finding is that archrules was reporting a false green for exactly this kind of reason.

.circleci/ci/src/commands/cmd-install-jdk.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Installs the JDK the project builds with on a machine executor.
+ *
+ * Docker jobs get their JDK from the `cimg/openjdk` tag, but machine jobs run on
+ * the Ubuntu VM image and use whatever it ships. As of the Q2 2026 images that is
+ * OpenJDK 21 on every LTS — 22.04, 24.04 and 26.04 alike — so moving to a newer
+ * Ubuntu does not help: the JDK has to be installed.
+ *
+ * Only needed by the machine jobs that invoke Maven; the e2e and packaging jobs
+ * run on the same executor but never compile Java.
```
ytvnr: I verified the e2e half of this claim: `job-e2e-test` and `job-e2e-cypress` run yarn and Docker only, so excluding them is right.

I did not manage to confirm the packaging half. Could you double check that `job-package-bundle`, `job-publish-rpm-packages`, `job-nexus-staging` and `job-backend-build-and-publish-on-download-website` either do not run on a machine executor or never invoke `mvn`/`java`? The RPM and bundle jobs handle Java distributions, which is where I would expect a surprise. If any of them does both, it has the same silent-JDK-21 problem as the test jobs and did not get the command.

.circleci/ci/src/config.ts
```diff
@@ -72,11 +72,11 @@ const executor = {
   },
   openjdk: {
     image: 'cimg/openjdk',
-    version: '21.0.9',
+    version: '25.0.3',
   },
   node: {
     image: 'cimg/node',
-    version: '22.12.0',
+    version: '24.18.0',
```
ytvnr, line 79: Minor drift worth knowing about: this pins 24.18.0, while `cimg/openjdk:25.0.3-node` ships 24.15.0. Both satisfy the `^24.0` the projects declare, so nothing breaks. It just means a job's Node version depends on which executor family it lands on, which is the kind of thing that costs an hour when a patch-level Node bug eventually shows up on one job and not the other.

pom.xml
```diff
@@ -24,7 +24,7 @@
   <parent>
     <groupId>io.gravitee</groupId>
     <artifactId>gravitee-parent</artifactId>
-    <version>24.0.1</version>
+    <version>25.1.0</version>
```
ytvnr, line 27: The description attributes "drops the local override" for archrules to this bump, but the diff changes no archrules version and the base pom already declares none. The plugin resolves by inheritance either way, so the outcome is right. Just worth fixing the description so reviewers do not go looking for a removal that is not in the diff.

pom.xml
```diff
@@ -24,7 +24,7 @@
   <parent>
     <groupId>io.gravitee</groupId>
     <artifactId>gravitee-parent</artifactId>
```
ytvnr, line 26: Blast radius question rather than a finding: with `java.version=25` coming from the parent, every APIM artifact published from this branch is class file 69. Any plugin or downstream repo still on `gravitee-parent` 24 / JDK 21 that depends on a `gravitee-apim-*` artifact will fail with `UnsupportedClassVersionError` as soon as it picks up a build from here.

The description covers the companion changes on the build side but says nothing about sequencing the plugin repos. Is that tracked somewhere, or is the expectation that plugins move to 25 before this reaches a released version?

phiz71:
Thanks — thorough review, and the three you flagged as worth acting on are now in `58a9e18e89`.

**Retry and cache.** The command now restores `~/.cache/temurin` keyed on the feature major, downloads only on a miss, and passes `--retry 3 --retry-all-errors --retry-delay 5`. Your point about the cache incidentally freezing the patch for the cache lifetime is exactly right, so I documented it as an intended property rather than a side effect.

**Checksum.** Switched from `/v3/binary/latest` to `/v3/assets/latest`, which returns the download link and its sha256 in the same response — so the verification costs one call rather than two, and the URL and digest cannot disagree. `sha256sum -c` now runs before anything is extracted as root.

**Regression guard.** Added `machine-jobs-jdk.spec.ts`, asserting over the generated config that every machine job running Maven installs the JDK, that no machine job without Maven does, and that no Docker job does. I checked it actually bites: removing `withJdk` from `abstract-job-test.ts` fails it with `["job-test-rest-api", "job-test-plugin", "job-test-reporter", "job-test-repository"]`. Agreed on the irony of shipping this one untested in a PR about a rule that passed while checking nothing.

**`withJdk` helper.** Done, all three call sites go through it. It returns nothing for Docker executors, so the executor decides — `job-test-integration` needed its executor hoisted into a variable to take part.

**`JAVA_HOME` on the early-exit path** and **the `PATH` duplication** are both fixed. Good catch on the single quotes; `PATH` was gaining an entry per step.

**Packaging jobs.** Checked all four, no gap:

| job | executor | Java |
|---|---|---|
| `job-package-bundle` | `cimg/base` (Docker) | none |
| `job-publish-rpm-packages` | Ubuntu machine | none — only `docker run` to chown the RPMs |
| `job-nexus-staging` | `cimg/openjdk` (Docker) | Maven, JDK from the image |
| `job-backend-build-and-publish-on-download-website` | `cimg/openjdk` (Docker) | same |

The RPM job was the one worth suspecting, but its Java work happens inside a container it launches, not on the VM. Nothing combines machine and Java.

**Node 24.18.0 vs 24.15.0.** Real and deliberate, and now the third place where a version comes from an image we do not control. Both satisfy `^24.0`; the alternative would be pinning `cimg/node` to whatever `cimg/openjdk:<jdk>-node` happens to bundle, which re-couples them for no benefit. Worth a follow-up if it ever bites.

**Description.** Fixed — you were right that "drops the local override" pointed at nothing. The override existed transiently within the branch (added with the plugin release, removed by the parent bump) so it nets out of the diff; the description now just says it resolves by inheritance. Also added a note at the top about the target branch and why `alpha-jdk25-run-e2e` carries the full migration.

**Blast radius on downstream plugins.** Genuinely open, and I do not have the answer. Every `gravitee-apim-*` artifact published from here is class file 69, so any plugin repo still on `gravitee-parent` 24 that consumes one fails with `UnsupportedClassVersionError`. Nothing in this branch sequences that. It does not bite while this stays on `alpha-jdk25` with `-alpha-jdk25` versions, but it has to be answered before the migration reaches anything released — @phiz71, is the plugin fleet tracked somewhere for this?

pom.xml
phiz71: it's expected, every new plugins that will depends on APIM 4.13, will have to also be compiled with JDK 25

ytvnr (approved):

PR #18832, opened 2026-07-30
chore: move the frontend toolchain to Node 24

phiz71:
## Issue

N/A — toolchain maintenance.

## Description

Moves the frontend toolchain from Node 22 to Node 24.

Node 22 leaves LTS in April 2027 and Node 24 is the current LTS. It is also what the CircleCI images have already moved to: every Ubuntu VM image ships 24.15.0, and `cimg/openjdk:<jdk>-node` now bundles 24 rather than 22. Those images carry no nvm, so the Node they bundle is the Node the build gets — the project either follows or fights them.

**Nothing in the toolchain asks for an older Node.** `@angular/core` 20.3.16 declares `^20.19.0 || ^22.12.0 || >=24.0.0`; `nx` and `@nx/angular` declare no constraint at all; `typescript` asks for `>=14.17`. The project's own `engines` field was the only thing in the way — and it was already inconsistent, the root workspace on `^22.12` while `.circleci/ci` still sat on `^20`. Both are now `^24`.

| | before | after |
|---|---|---|
| `package.json`, `gravitee-apim-portal-webui/package.json` | `^22.12` | `^24.0` |
| `.circleci/ci/package.json` | `^20` | `^24` |
| `.nvmrc` | `22.12.0` | `24.18.0` |
| `cimg/node` executor (`config.ts`) | `22.12.0` | `24.18.0` |
| node orb tag in the setup config | `22.12` | `24.18` |

### A range, not a pin

`engines` declares `^24.0` rather than a fixed version. The Node bundled in `cimg/openjdk:<jdk>-node` is chosen by the CircleCI image maintainers and moves independently of anything we control, so a pinned version guarantees another breakage the next time they bump it. A range absorbs it.

## Additional context

Verified locally on Node 24.16.0:

- `yarn install --immutable` clean, `yarn.lock` untouched
- all 8 nx projects build, console included with the same heap cap the CI applies (`--max_old_space_size=4086`)
- `.circleci/ci`: `tsc --noEmit`, 146/146 generation tests, and license / eslint / prettier all green

Deliberately left alone: `CONTRIBUTING.adoc` still says `java -version (expect 17)` in its environment checks, which is stale — master builds on Java 21. Correcting it belongs in a Java-related change, not here.


phiz71:
Closing: the split this PR was built on does not hold.

The Maven build shells out to `yarn` from the **openjdk** executor, not the `cimg/node` one, and `cimg/openjdk:<jdk>-node` bundles a Node version chosen by the JDK tag:

| image | bundled Node |
|---|---|
| `cimg/openjdk:21.0.9-node` (master today) | 22.21.1 |
| `cimg/openjdk:25.0.3-node` (JDK 25 branch) | 24.15.0 |

So on master, tightening `engines` to `^24.0` fails every Maven job that runs yarn — `gravitee-gamma` first:

```
error gravitee-apim-workspace@0.0.0: The engine "node" is incompatible with this module.
Expected version "^24.0". Got "22.21.1"
```

Getting Node 24 on that path without moving the JDK image would mean installing Node explicitly in those jobs — machinery that exists only to fight the image, and that disappears the moment the JDK 25 images land.

The change itself is sound: gamma builds fine on Node 24 locally (`mvn -f gravitee-gamma/pom.xml install`, green), as do all 8 nx projects. It simply belongs with the image bump.

Folded into #18831, where `cimg/openjdk:25.0.3-node` supplies Node 24 for free.

PR #18838, opened 2026-07-30
fix(plan): enforce native keyless/secure plan exclusivity across DEPRECATED plans

ThibaudAV:
## Issue

https://gravitee.atlassian.net/browse/APIM-14831

## Description

On a Native (Kafka) API, a keyless plan must never coexist with a secure (mTLS or authentication) plan. When both are active the gateway's Kafka reactor throws `KafkaServerUnsupportedSecureAndUnsecurePlansException` and never mounts the SASL listener, so **every** client of the API fails to authenticate (`enabled mechanisms are []`).

The publish-time guard (`PlanServiceImpl#validateNoConflictingAuthenticationForNativePlan`) only looked at `PUBLISHED` plans. A **`DEPRECATED`** keyless plan was therefore ignored, so you could publish an `API_KEY`/`JWT`/`OAUTH2` plan next to a deprecated keyless one (and vice-versa) — an invalid combination that then broke the reactor. The gateway loads `DEPRECATED` plans as active, so "active" for this rule must mean `PUBLISHED` **or** `DEPRECATED`.

What changed:

- Extract the rule into a shared `NativePlanSecurityValidator` — considers a plan conflicting when it is `PUBLISHED` or `DEPRECATED`, excludes the plan being published, keeps the existing keyless / mTLS / authentication mutual-exclusivity semantics.
- `v4 PlanServiceImpl#publish` delegates to it → closes the `DEPRECATED` gap.
- `PublishPlanDomainServiceImpl#publish` now enforces it too (it previously only had the single-keyless guard) → covers the API_PRODUCT / newer publish flow.
- `ValidatePlanDomainService` cross-checks the whole imported plan set → covers the CRD / GKO import path, which had no such check.

Result: the rest-api can no longer produce a plan combination the gateway would reject; the reactor guard becomes a true last resort. Tests added for each path, including the exact `DEPRECATED keyless + publish API_KEY` scenario.

## Additional context

Reproduction: deprecate then keep a keyless plan while publishing an API_KEY plan on a native Kafka API → gateway reactor fails to start, clients get `Unexpected handshake request with client mechanism PLAIN, enabled mechanisms are []`.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/domain_service/ValidatePlanDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/domain_service/ValidatePlanDomainService.java
@@ -20,12 +20,14 @@ import io.gravitee.apim.core.api.model.crd.ApiCRDSpec;
 import io.gravitee.apim.core.api.model.crd.PlanCRD;
 import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.apim.core.documentation.model.Page;
+import io.gravitee.apim.core.plan.model.NativePlanSecurityCategory;
 import io.gravitee.apim.core.plan.model.Plan;
 import io.gravitee.apim.core.plan.model.factory.PlanModelFactory;
 import io.gravitee.apim.core.validation.Validator;
 import io.gravitee.definition.model.v4.ApiType;
 import io.gravitee.definition.model.v4.listener.AbstractListener;
 import io.gravitee.definition.model.v4.nativeapi.NativePlan;
+import io.gravitee.definition.model.v4.plan.PlanStatus;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -98,6 +100,33 @@ public class ValidatePlanDomainService implements Validator<ValidatePlanDomainSe
             }
         });
 
+        if (input.apiCRDSpec.isNative()) {
+            validateNoConflictingNativeSecurity(input.plans(), errors);
+        }
+
         return Result.ofBoth(input.sanitized(sanitizedPlans), errors);
+    }
+
+    /**
+     * Same mutual-exclusivity rule as {@code NativePlanSecurityValidator} (publish paths), applied
+     * here over the whole imported plan set. Keep both in sync via {@link NativePlanSecurityCategory}.
+     */
+    private static void validateNoConflictingNativeSecurity(Map<String, PlanCRD> plans, List<Error> errors) {
+        long distinctSecurityCategories = plans
+            .values()
+            .stream()
+            .filter(plan -> plan.getStatus() == PlanStatus.PUBLISHED || plan.getStatus() == PlanStatus.DEPRECATED)
+            .filter(plan -> plan.getSecurity() != null)
+            .map(plan -> NativePlanSecurityCategory.fromSecurityTypeLabel(plan.getSecurity().getType()))
+            .distinct()
+            .count();
+
+        if (distinctSecurityCategories > 1) {
+            errors.add(
+                Error.severe(
+                    "a Native API cannot combine Keyless, mTLS and authentication plans: keyless, mTLS and authentication plans are mutually exclusive"
+                )
+            );
+        }
     }
 }
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/model/NativePlanSecurityCategory.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/plan/model/NativePlanSecurityCategory.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.plan.model;
+
+import io.gravitee.rest.api.model.v4.plan.PlanSecurityType;
+
+/**
+ * Security categories that are mutually exclusive on a Native (Kafka) API: a plan of a given
+ * category cannot coexist with a plan of a different category. All authentication mechanisms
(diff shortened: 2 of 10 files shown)


aditya-goyal01:
## Review: #18838 

**Verdict:** 
Line 48 `!existingPlan.getId().equals(planToPublish.getId())` NPEs when `planToPublish` (always included in its own `apiPlans` set) has no ID.

Checked `git show master:...PlanService_PublishTest.java` — 4 of the 6 failing tests - (`shouldNotPublishKeylessNativePlanIfAuthPlanPublished`, `shouldNotPublishMtlsNativePlanIfKeylessPlanPublished`, `shouldNotPublishAuthNativePlanIfMtlsPlanPublished`, `shouldNotPublishAuthNativePlanIfKeylessPlanPublished`) already exist unchanged on master — real regression, not just sloppy new tests. Other 2 are new tests this same commit added.

**Recommendation:** 
- `Objects.equals(existingPlan.getId(), planToPublish.getId())` or compare by reference.

callaertanthony (commented):
Reviewed the whole change. The bug is real, the root cause (`PUBLISHED`-only filter vs. the gateway loading `DEPRECATED` plans as active) is correctly identified, and all three write paths are now covered with tests. Nothing blocking from my side — a few things I'd fix before merge.

## Worth fixing before merge

**1. `NativePlanAuthenticationConflictException`'s message is now wrong.**

All three branches say *"is already published for the Native API"* (`NativePlanAuthenticationConflictException.java:34-42`). After this PR the conflicting plan can be **deprecated** — which is exactly the state the user moved it to in order to get rid of it. They deprecate the keyless plan, hit publish, and are told a keyless plan "is already published". One-word fix: *"published or deprecated"*.

**2. The "single source of truth" doesn't actually share the rule.**

`NativePlanSecurityCategory` shares three enum constants; both sites still hand-roll their own mapping (`fromSecurityTypeLabel` over String labels vs. `categoryOf` over the repository enum) *and* their own predicate (`anyMatch(!=)` vs. `distinct().count() > 1`). The two Javadocs saying "keep both in sync" are the tell. Moving the predicate into the enum and letting each site supply only its mapping would leave one rule instead of two:

```java
public static boolean hasConflict(Stream<NativePlanSecurityCategory> activeCategories) {
    return activeCategories.distinct().count() > 1;
}
```

**3. The self-exclusion filter is dead code, and two tests cover unreachable states.**

`NativePlanSecurityValidator`'s `.filter(p -> !Objects.equals(p.getId(), planToPublish.getId()))` can never change the outcome: both callers throw `PlanAlreadyPublished` / `PlanAlreadyDeprecated` / `PlanAlreadyClosed` up front (`PlanServiceImpl.java:611-615`, `PublishPlanDomainServiceImpl.java:70-75`), so the plan being published is always `STAGING` and never passes the `PUBLISHED || DEPRECATED` filter anyway. `should_not_conflict_with_itself` and `should_not_npe_when_plan_to_publish_has_no_id` assert on states the callers can't produce. Dropping the filter also removes the NPE risk that was raised earlier in this thread.

## Non-blocking

**4. Undocumented behavior change for native PUSH plans.** The old code fell into the `else` branch when `getSecurity() == null` and threw (then NPE'd on `planToPublishType.name()` in `getParameters()`). The new `getSecurity() == null` early-return silently makes that combination legal. Probably the right call, but it's unstated and untested.

**5. The CRD error doesn't name the conflicting plans.** Every other error in `ValidatePlanDomainService` is `invalid plan [%s]. Error: %s`; this one is a bare global sentence that also repeats itself ("cannot combine Keyless, mTLS and authentication plans: keyless, mTLS and authentication plans are mutually exclusive"). For someone reading a rejected CRD, the offending plan keys are the whole value. It's also worded differently from what the REST publish path returns, so the same rule surfaces two messages.

**6. `PlanCRD.status` has no default** (`PlanCRD.java:60`, and no defaulting in `PlanModelFactory.fromCRDSpec`), so a plan with a null status is filtered out and skips the new check. Probably unreachable via the GKO schema — nit.

**7. APIs already in the broken state aren't addressed.** One sitting on DEPRECATED keyless + PUBLISHED api-key stays broken, and now also gets a 400 on any further publish. Fine as scope, but worth a release note.

## Things I checked that are fine

- Layering: `core` → `io.gravitee.rest.api.model` is already the norm here (323 core files import it), and `infra` → `rest.api.service.exceptions` was already in `PublishPlanDomainServiceImpl`. No ArchUnit rules in this module, so no new violation.
- Self-conflict, re-publishing a DEPRECATED plan, and the pre-existing `KeylessPlanAlreadyPublishedException` guard all still behave correctly alongside the new validator.

ThibaudAV:
Thanks @callaertanthony, great review. Addressed in the latest push:

**#1 — Wrong exception message.** Fixed: all three branches of `NativePlanAuthenticationConflictException` now say *"is already published or deprecated"*, since after this PR the conflicting plan can be deprecated.

**#3 — Dead self-exclusion filter.** Removed. Confirmed both callers reject `CLOSED`/`PUBLISHED`/`DEPRECATED` up front (`PlanServiceImpl` and `PublishPlanDomainServiceImpl`), so `planToPublish` is always `STAGING` and never matches the `PUBLISHED || DEPRECATED` filter — the id-based exclusion couldn't change the outcome. Dropping it also removes the NPE risk raised earlier in the thread (so no `Objects.equals` band-aid needed), and I removed the two tests that asserted on those unreachable states. Added a short comment explaining why no self-exclusion is required.

**#2 — Single source of truth.** Fair point that only the enum constants are shared while the mapping + predicate are still duplicated. I've left it for now to keep this fix focused; happy to do the `NativePlanSecurityCategory.hasConflict(Stream<...>)` extraction in a follow-up if you'd prefer it here.

**#4–#7 (non-blocking).** Acknowledged — the native PUSH `getSecurity() == null` early-return behavior change, the CRD error not naming the conflicting plans (and wording drift vs. the REST path), the `PlanCRD.status` default, and the release note for already-broken APIs. Leaving these out of this PR's scope unless you want any pulled in.

Let me know if you'd like #2 (or any of #4–#7) in this PR rather than a follow-up.

callaertanthony (approved):
Re-reviewed the force-push (`89bc2d8d`) — three files changed, the other seven are byte-identical.

**#1 (exception message)** — fixed in all three branches. Nothing else in the repo asserts on that string (no test, no console copy), so no fallout.

**#3 (self-exclusion)** — the filter and its import are gone, with a comment in their place. I re-checked the removal on its own terms: even if a caller *did* pass a `PUBLISHED`/`DEPRECATED` plan-to-publish inside its own set, it would compare against itself and find an equal category. So the removal can't produce a false positive under any input, not just under the caller-guard assumption the comment cites. Safe unconditionally.

**#2 and #4–#7** — agreed, follow-up is the right call. None of them change behavior, and keeping this focused matters for something heading to support branches.

CI is 50/50 green, including the backend build that was pending during the first pass.

One leftover, non-blocking: `should_not_conflict_with_itself` is still in `NativePlanSecurityValidatorTest`. The NPE test was removed but this one wasn't, and it's stale in the same way — it names a self-exclusion mechanism that no longer exists, and it builds a `PUBLISHED` plan as the plan-to-publish, precisely the state the new comment says callers can't produce. It passes only because a lone plan trivially can't differ in category from itself. Worth deleting or renaming to what it now asserts, but not worth another round-trip.

LGTM 👍

PR #18849, opened 2026-07-30
fix(console): save group membership roles on the group's own environment

JedrzejJanasiak:
## Issue

https://gravitee.atlassian.net/browse/APIM-14769

## Description

Editing the roles of a user's group membership (Organization > Users > Memberships) always
sent the request to the currently selected environment instead of the environment the group
actually belongs to. In an organization with more than one environment, the backend then
rejected the lookup with `Group [id] cannot be found`, and the change was silently lost.

The memberships sub-component now reports the environment of each group along with the roles
form, and the save action passes it to `addOrUpdateMemberships` — the same way the add and
delete actions already did.

https://github.com/user-attachments/assets/f229105d-f87c-4eda-933e-1c1002eb6aec

### DIFF
--- a/gravitee-apim-console-webui/src/organization/configuration/user/detail/org-settings-user-detail.component.spec.ts
+++ b/gravitee-apim-console-webui/src/organization/configuration/user/detail/org-settings-user-detail.component.spec.ts
@@ -588,6 +588,7 @@ describe('OrgSettingsUserDetailComponent', () => {
 
     expectAddOrUpdateGroupMembershipRequest(
       'groupA',
+      'envAlphaId',
       [fakeGroupMembership({ id: user.id, roles: [{ scope: 'APPLICATION', name: 'ROLE_APP_USER' }] })],
       [
         {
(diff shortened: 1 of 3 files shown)


gravitee-apim-console-webui/src/organization/configuration/user/detail/org-settings-user-detail.component.ts
```diff
@@ -231,18 +239,22 @@ export class OrgSettingsUserDetailComponent implements OnInit, OnDestroy {
...
+                    roles: [
+                      { scope: 'GROUP' as const, name: GROUP ? 'ADMIN' : '' },
+                      { scope: 'API' as const, name: API },
+                      { scope: 'APPLICATION' as const, name: APPLICATION },
+                      { scope: 'INTEGRATION' as const, name: INTEGRATION },
+                      { scope: 'API_PRODUCT' as const, name: API_PRODUCT },
+                    ],
+                  },
+                ],
+                this.groupEnvironmentIds[groupId],
```
paulatulis, line 256: `this.groupEnvironmentIds[groupId]` is passed with no guard, and group.service.ts:60 does:

```
const environmentUrl = environmentId ? `${org.baseURL}/environments/${environmentId}` : this.constants.env.baseURL;
```

A falsy value falls back to the currently selected environment — exactly the bug being fixed, with no error and no diagnostic. Reachable at the type level: Group.environmentId is optional (entities/group/group.ts:38) and gets copied straight into a Record<string, string> unvalidated.

Failure scenario: backend omits environmentId → POST hits the current env → Group [id] cannot be found → change lost, user sees nothing. Same silent failure, new path.

Suggest: look the value up in the mergeMap at line 237 and, if falsy, snackBarService.error(...) + EMPTY for that group rather than calling the service.

gravitee-apim-console-webui/src/organization/configuration/user/detail/org-settings-user-detail.component.spec.ts
```diff
@@ -604,6 +605,50 @@ describe('OrgSettingsUserDetailComponent', () => {
     );
   });
 
+  it('should save group roles on the environment the group belongs to', async () => {
```
paulatulis, line 608: The new test covers the map being refreshed after a tab switch, not what happens to pre-switch dirty state. A future change that keeps stale groupEnvironmentIds entries across a switch would re-introduce the wrong-environment POST undetected. Suggest: edit on tab 1, switch to tab 2, submit, expectNone on both environments' member URLs.

JedrzejJanasiak: Agreed, added. The form and the environment map are rebuilt together in initGroupsRolesForm and emitted as one payload, so the stale state is already dropped today — but that was an unwritten invariant. The new test edits on tab 1, switches to tab 2, submits, and asserts expectNone on both environments plus the old env.baseURL. Verified it actually fails if the parent stops replacing the form on tab switch.

gravitee-apim-console-webui/src/organization/configuration/user/detail/org-settings-user-detail.component.ts
JedrzejJanasiak: Good catch on the silent fallback — fixed, though via a different route. The failure scenario as described isn't reachable here: the tabs always pass environmentId to the v2 endpoint, and it becomes a hard Mongo filter (MongoGroupRepository.search, where("environmentId").is(...)), so every group in the form has a non-null environmentId equal to the tab's. UserGroupResponse.environmentId is also non-optional; the optionality only appears when mapping onto the generic Group entity.

Since it's unreachable, a user-facing error felt wrong, so I closed it at the source instead: group.environmentId ?? this.selectedEnvironmentId. The tab is the group's environment by construction, so the value can no longer be falsy and the fallback branch in addOrUpdateMemberships is unreachable. Happy to switch to the error path if you'd rather have it loud.

paulatulis (approved):
LGTM!

PR #18864, opened 2026-07-30
fix(observability): scope filters per surface and wire transaction/request id in logs search

araffass:
##  Issue 
  https://gravitee.atlassian.net/browse/APIM-14828
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
  On the Observability dashboards, the Transaction ID filter was broken on both surfaces (APIM-14828):                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  1. Overview (analytics) — the filter was offered in the Add-filter dialog, but applying it failed with Filter 'TRANSACTION_ID' is not supported for analytics queries (the analytics engine rejects it by design).                                                                                                                                                                                                             
  2. Logs — the filter was accepted but had no effect: the page's search-param builder only copied a fixed set of filter keys, and TRANSACTION_ID (and REQUEST_ID, same family) was silently dropped before the request was built — even though the whole backend chain (logs use case → CRUD → Elasticsearch adapter, v2 transaction/v4 transaction-id field mapping) already supported it.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                 
##  Fix
                                                                                                                                                                                                                                                                                                                                                                                                                                     
  Per-filter surface signals (backend-derived, no drift possible). The observability filter catalog serves two surfaces; each filter now advertises which ones support it:                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  - New core enum FilterSignal { ANALYTICS, LOGS }; FilterSpec gains a signals list (compat constructor keeps existing call sites untouched).                                                                                                                                                                                                                                                                                    
  - GetAnalyticsFilterDefinitionsUseCase derives signals from the engines' own definitions: AnalyticsQueryValidator.supportsAnalytics(...) (the exact set the validator rejects — now exposed and reused) and the logs engine's FilterName enum (name-matched; HTTP_PATH ↔ URI translation documented). The 4 logs-only filters (TRANSACTION_ID, REQUEST_ID, PAYLOAD, ERROR_KEY) come out as [LOGS].                             
  - Exposed via a new optional signals property in openapi-analytics.yaml (additive; omitted when null on the sibling per-metric endpoints).                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  Frontend gate per surface. New OBSERVABILITY_FILTER_SIGNAL injection token; ObservabilityFiltersApiService.getDefinitions() drops definitions not supporting the provided surface. Overview (dashboard-viewer) provides ANALYTICS, the Logs page provides LOGS. Definitions without signals (older backend during rolling upgrade) stay visible everywhere; an empty signals list means excluded. The shared dialog needs no   
  change — its existing "Logs only" badge now lights up on the Logs page.                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  URL-hydration prune. Conditions restored from a bookmarked/shared URL could still smuggle a logs-only filter onto Overview and reproduce the error — DashboardFiltersStore now prunes hydrated conditions whose filter isn't offered on the surface, then the page re-queries cleanly.                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  Logs wiring. buildSearchParam now passes TRANSACTION_ID and REQUEST_ID (all values — new plural transactionIds/requestIds params serialized as IN, so multi-value filters return results for every value; the scalar requestId deep-link path is untouched).

### DIFF
--- a/gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-filters.store.spec.ts
+++ b/gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-filters.store.spec.ts
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { FilterCondition } from '@gravitee/gravitee-dashboard';
+import { FILTER_DEFINITION_PROVIDER, FilterCondition, FilterDefinition } from '@gravitee/gravitee-dashboard';
 
 import { TestBed } from '@angular/core/testing';
 import { ActivatedRoute, Router } from '@angular/router';
@@ -314,6 +314,82 @@ describe('DashboardFiltersStore', () => {
     expect(hydratedStore.periodControl.value.period).toBe('custom');
     expect(hydratedStore.periodControl.value.from!.valueOf()).toBe(1000000);
     expect(hydratedStore.periodControl.value.to!.valueOf()).toBe(2000000);
+  });
+
+  it('should drop URL-hydrated conditions for filters the surface does not offer, once definitions arrive', () => {
+    const definitions$ = new Subject<FilterDefinition[]>();
+    const definitionProviderMock = { getDefinitions: jest.fn().mockReturnValue(definitions$.asObservable()) };
+    const navigate = jest.fn().mockResolvedValue(true);
+    const snapshot = {
+      queryParams: {
+        q: JSON.stringify({
+          filter: [
+            { field: 'API', operator: 'EQ', value: 'id-1' },
+            { field: 'TRANSACTION_ID', operator: 'EQ', value: 'tx-1' },
+          ],
+        }),
+        v: '1',
+      },
+    };
+
+    TestBed.resetTestingModule();
+    TestBed.configureTestingModule({
+      providers: [
+        DashboardFiltersStore,
+        { provide: FILTER_DEFINITION_PROVIDER, useValue: definitionProviderMock },
+        {
+          provide: ActivatedRoute,
+          useValue: {
+            queryParams: new Subject<Record<string, string>>().asObservable(),
+            snapshot,
+          },
+        },
+        {
+          provide: Router,
+          useValue: { navigate },
+        },
+      ],
+    });
+
+    const hydratedStore = TestBed.inject(DashboardFiltersStore);
+    // Definitions have not loaded yet: restored conditions stay visible.
+    expect(hydratedStore.conditions().map(c => c.field)).toEqual(['API', 'TRANSACTION_ID']);
+
+    definitions$.next([{ name: 'API', label: 'API', type: 'KEYWORD', operators: ['EQ'] }]);
+    definitions$.complete();
+
+    expect(hydratedStore.conditions().map(c => c.field)).toEqual(['API']);
+    // The URL is re-synced so the pruned filter is not shared onward.
+    expect(navigate).toHaveBeenCalled();
+  });
+
+  it('should keep the same conditions reference when nothing is pruned', () => {
+    const definitions$ = new Subject<FilterDefinition[]>();
+    const definitionProviderMock = { getDefinitions: jest.fn().mockReturnValue(definitions$.asObservable()) };
+    const navigate = jest.fn().mockResolvedValue(true);
+    const snapshot = {
+      queryParams: { q: JSON.stringify({ filter: [{ field: 'API', operator: 'EQ', value: 'id-1' }] }), v: '1' },
+    };
+
+    TestBed.resetTestingModule();
+    TestBed.configureTestingModule({
+      providers: [
+        DashboardFiltersStore,
+        { provide: FILTER_DEFINITION_PROVIDER, useValue: definitionProviderMock },
+        { provide: ActivatedRoute, useValue: { queryParams: new Subject<Record<string, string>>().asObservable(), snapshot } },
+        { provide: Router, useValue: { navigate } },
+      ],
+    });
+
+    const hydratedStore = TestBed.inject(DashboardFiltersStore);
+    const before = hydratedStore.conditions();
+
+    definitions$.next([{ name: 'API', label: 'API', type: 'KEYWORD', operators: ['EQ'] }]);
+    definitions$.complete();
+
+    // Same reference — consumers must not re-query when the prune is a no-op.
+    expect(hydratedStore.conditions()).toBe(before);
+    expect(navigate).not.toHaveBeenCalled();
   });
 
   it('should show a loading value label until URL-hydrated ID labels are resolved', () => {
--- a/gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-filters.store.ts
+++ b/gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-filters.store.ts
@@ -19,6 +19,7 @@ import {
   DEFAULT_VIEW_STATE_PERIOD,
   decodeViewState,
   encodeViewState,
+  FILTER_DEFINITION_PROVIDER,
   FilterCondition,
   ID_BASED_FILTER_NAMES,
   RequestFilter,
@@ -33,7 +34,7 @@ import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { FormControl } from '@angular/forms';
 import { ActivatedRoute, Params, Router } from '@angular/router';
 import moment, { Moment } from 'moment';
-import { Subscription } from 'rxjs';
+import { catchError, EMPTY, Subscription } from 'rxjs';
 
 import { FilterLabelResolver } from './filter-label.resolver';
 
@@ -54,6 +55,7 @@ export class DashboardFiltersStore {
   private readonly router = inject(Router);
   private readonly destroyRef = inject(DestroyRef);
   private readonly labelResolver = inject(FilterLabelResolver, { optional: true });
+  private readonly definitionProvider = inject(FILTER_DEFINITION_PROVIDER, { optional: true });
 
   private readonly _conditions = signal<FilterCondition[]>([]);
   readonly conditions = this._conditions.asReadonly();
@@ -72,6 +74,7 @@ export class DashboardFiltersStore {
 
   private _skipNextQueryParamsEmit = false;
   private labelResolutionSubscription?: Subscription;
+  private definitionsPruneSubscription?: Subscription;
   private labelResolutionGeneration = 0;
 
   /**
@@ -224,6 +227,35 @@ export class DashboardFiltersStore {
         .subscribe(enriched => {
           if (generation !== this.labelResolutionGeneration) return;
           this._conditions.update(current => mergeResolvedLabels(current, enriched));
+        });
+    }
+
+    // Conditions restored from the URL may reference filters this surface does not offer (e.g. a
+    // logs-only filter deep-linked onto an analytics dashboard) — drop them instead of querying an
+    // engine that rejects them (APIM-14828). The prune is asynchronous: until definitions arrive,
+    // an initial query may still carry the unsupported filter and briefly error; it self-heals on
+    // the re-query triggered below.
+    this.definitionsPruneSubscription?.unsubscribe();
+    if (conditions.length > 0 && this.definitionProvider) {
+      this.definitionsPruneSubscription = this.definitionProvider
+        .getDefinitions()
+        .pipe(
+          // Fail open: an unavailable catalog must not wipe restored filters.
+          catchError(() => EMPTY),
+          takeUntilDestroyed(this.destroyRef),
+        )
+        .subscribe(definitions => {
+          if (generation !== this.labelResolutionGeneration) return;
+          const offered = new Set(definitions.map(definition => definition.name as string));
+          const current = this._conditions();
+          const kept = current.filter(condition => offered.has(condition.field as string));
+          if (kept.length === current.length) {
+            // Nothing pruned — keep the same reference so no consumer re-queries.
+            return;
+          }
+          this._conditions.set(kept);
+          // The URL must not keep advertising a filter this surface cannot evaluate.
+          this.syncToRouter();
         });
     }
   }
--- a/gravitee-apim-console-webui/src/management/observability/data-access/observability-filters-api.service.ts
+++ b/gravitee-apim-console-webui/src/management/observability/data-access/observability-filters-api.service.ts
@@ -22,7 +22,7 @@ import {
 } from '@gravitee/gravitee-dashboard';
 
 import { HttpClient, HttpParams } from '@angular/common/http';
-import { inject, Injectable } from '@angular/core';
+import { inject, Injectable, InjectionToken } from '@angular/core';
 import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
 
@@ -63,14 +63,26 @@ interface FilterValueItemApi {
   id?: string;
 }
 
+/** Observability surface consuming the shared filter catalog — mirrors the backend FilterSignal enum. */
+export type ObservabilityFilterSignal = 'ANALYTICS' | 'LOGS';
+
+/**
+ * Surface hosting the filter bar. When provided, definitions carrying signals are narrowed to that
+ * surface so a page only offers filters its query engine supports (APIM-14828).
+ */
+export const OBSERVABILITY_FILTER_SIGNAL = new InjectionToken<ObservabilityFilterSignal>('OBSERVABILITY_FILTER_SIGNAL');
+
 @Injectable()
 export class ObservabilityFiltersApiService implements FilterDefinitionProvider, FilterValuesProvider {
   private readonly http = inject(HttpClient);
   private readonly constants = inject(Constants);
+  private readonly surfaceSignal = inject(OBSERVABILITY_FILTER_SIGNAL, { optional: true });
 
   getDefinitions(): Observable<FilterDefinition[]> {
     const url = `${this.constants.env?.v2BaseURL}/observability/filters/definition`;
-    return this.http.get<FilterSpecsResponseApi>(url).pipe(map(res => (res.data ?? []).map(item => this.mapDefinition(item))));
+    return this.http
+      .get<FilterSpecsResponseApi>(url)
+      .pipe(map(res => (res.data ?? []).map(item => this.mapDefinition(item)).filter(def => this.isSupportedOnSurface(def))));
   }
 
   getValues(query: FilterValuesQuery): Observable<FilterValuesResult> {
@@ -86,6 +98,14 @@ export class ObservabilityFiltersApiService implements FilterDefinitionProvider,
       params = params.set('to', String(query.to));
     }
     return this.http.get<FilterValuesResponseApi>(url, { params }).pipe(map(res => this.mapValuesResult(res, query.perPage)));
+  }
+
+  /** Definitions without signals (older backends) stay visible on every surface; an empty list means no surface supports it. */
+  private isSupportedOnSurface(def: FilterDefinition): boolean {
+    if (!this.surfaceSignal || def.signals == null) {
+      return true;
+    }
+    return def.signals.includes(this.surfaceSignal);
   }
 
   private mapDefinition(item: FilterSpecApi): FilterDefinition {
--- a/gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.ts
+++ b/gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.ts
@@ -43,7 +43,11 @@ import { EnvLogsTableComponent } from './components/env-logs-table/env-logs-tabl
 
 import { DashboardFiltersStore } from '../dashboards/ui/dashboard-viewer/dashboard-filters.store';
 import { FilterLabelResolver } from '../dashboards/ui/dashboard-viewer/filter-label.resolver';
-import { ObservabilityFiltersApiService } from '../data-access/observability-filters-api.service';
+import {
+  OBSERVABILITY_FILTER_SIGNAL,
+  ObservabilityFilterSignal,
+  ObservabilityFiltersApiService,
+} from '../data-access/observability-filters-api.service';
 import { EnvironmentLogsService, EnvironmentApiLog, SearchLogsParam, LogApiType } from '../../../services-ngx/environment-logs.service';
 import { SnackBarService } from '../../../services-ngx/snack-bar.service';
 import { GioTableWrapperPagination } from '../../../shared/components/gio-table-wrapper/gio-table-wrapper.component';
@@ -79,6 +83,8 @@ const API_TYPE_LABELS: Record<LogApiType, string> = {
     FilterLabelResolver,
     provideFilterDefinitions(ObservabilityFiltersApiService),
     provideFilterValues(ObservabilityFiltersApiService),
+    // This page queries the logs engine — analytics-only filters must not be offered here (APIM-14828).
+    { provide: OBSERVABILITY_FILTER_SIGNAL, useValue: 'LOGS' satisfies ObservabilityFilterSignal },
   ],
 })
 export class EnvLogsComponent {
@@ -246,6 +252,8 @@ export class EnvLogsComponent {
       apiProductIds: filterMap.get('API_PRODUCT'),
       uri: filterMap.get('HTTP_PATH')?.[0],
       bodyText: filterMap.get('PAYLOAD')?.[0],
+      requestIds: filterMap.get('REQUEST_ID'),
+      transactionIds: filterMap.get('TRANSACTION_ID'),
     };
   }
 
--- a/gravitee-apim-console-webui/src/services-ngx/environment-logs.service.ts
+++ b/gravitee-apim-console-webui/src/services-ngx/environment-logs.service.ts
@@ -70,7 +70,9 @@ export type SearchLogsParam = {
   period?: string;
   from?: string;
   to?: string;
+  /** Single-id lookup used by the log-details view; the filter bar uses `requestIds`. */
   requestId?: string;
+  requestIds?: string[];
   apiIds?: string[];
   applicationIds?: string[];
   planIds?: string[];
@@ -78,6 +80,8 @@ export type SearchLogsParam = {
   statuses?: number[];
   entrypoints?: string[];
   transactionId?: string;
+  /** Multi-value variant used by the filter bar (IN). */
+  transactionIds?: string[];
   uri?: string;
   responseTime?: number;
   errorKeys?: string[];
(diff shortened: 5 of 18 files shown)


gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.ts
```diff
@@ -246,6 +252,8 @@ export class EnvLogsComponent {
       apiProductIds: filterMap.get('API_PRODUCT'),
       uri: filterMap.get('HTTP_PATH')?.[0],
```
paulatulis, line 253: **MAJOR** — Multi-value HTTP Path on Logs silently filters by the first value only.

`supportsLogs` grants the `LOGS` signal to `HTTP_PATH` unconditionally, so the Logs Add-filter dialog now formally advertises it with its catalog operators `[EQ, IN]`. But the logs pipeline only handles it as a single-value `EQ`: this line sends `?.[0]` as the scalar `uri`, and `SearchEnvironmentLogsUseCase` explicitly ignores `ArrayFilter` for `URI`.

The path is reachable: `DashboardFiltersStore.add()` merges two conditions on the same field into one and promotes the operator via `promoteMembershipOperator` once `values.length >= 2`. So adding `HTTP Path = /a` then `HTTP Path = /b` on the Logs page yields a single `HTTP_PATH IN ['/a','/b']` condition, and only `/a` reaches the backend — no error, no warning.

The `?.[0]` truncation predates this PR, but the new signal derivation is what now asserts "the logs engine supports this filter" while only half its advertised operators actually work there. That's the same silent-drop failure class APIM-14828 is fixing. No test covers a multi-value HTTP_PATH condition.

gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-filters.store.ts
```diff
@@ -226,6 +229,21 @@ export class DashboardFiltersStore {
...
+    // engine that rejects them (APIM-14828).
+    this.definitionsPruneSubscription?.unsubscribe();
+    if (conditions.length > 0 && this.definitionProvider) {
+      this.definitionsPruneSubscription = this.definitionProvider
+        .getDefinitions()
+        .pipe(takeUntilDestroyed(this.destroyRef))
+        .subscribe(definitions => {
+          if (generation !== this.labelResolutionGeneration) return;
+          const offered = new Set(definitions.map(definition => definition.name as string));
+          this._conditions.update(current => current.filter(condition => offered.has(condition.field as string)));
```
paulatulis: **MAJOR** — The prune re-queries on every bookmarked-filter load, including when nothing is pruned.

`current.filter(...)` returns a new array reference even when no condition is removed, and Angular signals compare by reference. So on every hydration that has *any* URL filters — including entirely valid ones — the async arrival of definitions re-fires `_conditions`, `requestFilters` recomputes, and every consumer re-queries.

On env-logs the `searchParams -> toObservable -> switchMap` chain cancels the in-flight initial fetch and restarts it, roughly doubling initial load latency for every bookmarked-filter visit. On the dashboard, the `[requestFilters]` input change re-triggers all widget queries. This happens on the path where no pruning was needed at all.

paulatulis: **MAJOR** — When the prune does drop a condition, the URL still carries it.

`syncToRouter()` is never called here, so the `q` query param keeps encoding the pruned filter while the chips and queries no longer include it. Three consequences:

1. The visible state and the shareable URL disagree.
2. The prune is asynchronous, so on every reload the initial query still fires with the unsupported filter *before* definitions arrive — `AnalyticsQueryValidator` rejects it and the analytics widgets flash the error state, then the prune runs and re-queries. The error the ticket is about is still reachable, just transiently.
3. Re-sharing the still-dirty URL keeps propagating the unsupported filter.

So AC4 ("a bookmarked/shared URL must not reintroduce an unsupported filter") looks only half-resolved.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-analytics.yaml
```diff
@@ -1660,12 +1660,26 @@ components:
                         $ref: "#/components/schemas/ApiName"
                     description: List of API types this filter applies to, derived from the metrics that reference it.
                     example: ["HTTP_PROXY", "LLM"]
+                signals:
```
paulatulis, line 1663: **MAJOR** — `signals` is advertised inconsistently across the two endpoints that share this schema, and the contract isn't documented.

`FilterSpecsResponse` is referenced by two operations: `queryObservabilityFilterDefinitions` (line ~259) and `queryFilterSpecsForMetric` (line ~150, via `GetMetricFilterSpecUseCase` and the same `AnalyticsDefinitionMapper.toFilterSpecsResponse`). Signals are derived only in `GetAnalyticsFilterDefinitionsUseCase`, so the identical catalog filter (e.g. `API`) comes back **with** `signals` from one endpoint and `signals: null` from the other.

Meanwhile this description documents the property unconditionally and says nothing about absence or emptiness — yet the console gives those cases opposite meanings: absent = "no restriction, show everywhere" (the rolling-upgrade path), empty array = "no surface supports it, exclude". A consumer or codegen client reading only the spec can't learn that absence is expected on the per-metric endpoint, or that an empty array is meaningful and exclusionary.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetAnalyticsFilterDefinitionsUseCase.java
```diff
@@ -32,6 +43,39 @@ public GetAnalyticsFilterDefinitionsUseCase(AnalyticsDefinitionQueryService defi
...
+        if (AnalyticsQueryValidator.supportsAnalytics(name)) {
+            signals.add(FilterSignal.ANALYTICS);
+        }
+        if (supportsLogs(name)) {
+            signals.add(FilterSignal.LOGS);
+        }
+        return List.copyOf(signals);
+    }
+
+    private static boolean supportsLogs(FilterSpec.Name name) {
```
paulatulis, line 73: **MAJOR** — Surface support is derived by matching two independently-evolving enums by name, with no compile-time or test guard.

`supportsLogs` matches `FilterSpec.Name.name()` against `FilterName.values()`, plus a hard-coded `HTTP_PATH` -> `URI` special case and a comment noting two more (`MCP_METHOD`, `RESPONSE_TIME`) that silently fall through unmapped. Renaming or adding a value in either enum silently changes which filters appear on which surface — nothing fails to compile and no test catches it. Renaming `FilterName.TRANSACTION_ID`, for instance, would quietly remove the filter from the Logs page.

Zooming out, "which surface supports which filter" is now encoded in four independent places that can drift: this name-matching, `AnalyticsQueryValidator.ANALYTICS_UNSUPPORTED_FILTERS` (4 names), `AnalyticsDefinitionMapper`'s `@ValueMapping(source = "PAYLOAD", target = THROW_EXCEPTION)` guard, and the per-metric `filters:` lists in `analytics-definition.yaml`. The last two already disagree with the validator — the mapper hard-fails only on `PAYLOAD` while the validator rejects four names — so adding `TRANSACTION_ID`, `REQUEST_ID`, or `ERROR_KEY` to any metric's yaml filter list would make the per-metric endpoint advertise a filter the analytics engine rejects. That's this exact bug class, one yaml edit away.

(Related: an `analytics_engine` use case now imports `logs_engine.model.FilterName` directly, with no port abstracting the concern.)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/model/FilterSpec.java
```diff
@@ -27,8 +28,29 @@ public record FilterSpec(
     List<String> enumValues,
```
paulatulis, line 33: **MAJOR** — `FilterSpec` is now an 8-component record with a telescoping compatibility constructor.

The record gains an 8th component, a 7-arg compat constructor, and a hand-written `withSignals` wither. `java-conventions.md` ("Records and Builder Pattern") says: *"once a type has more than ~5 fields, positional constructors become fragile in tests and `with*` copy-methods multiply with every new field. Prefer a `@Builder`-annotated class (Lombok) or a companion builder."*

The next component added will force a third constructor overload and another wither, and tests already construct specs with long positional `null` runs.

Acknowledging this is a refactor ask on a targeted bugfix — reasonable to defer, but flagging it as a real deviation rather than let it pass silently.

gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-filters.store.spec.ts
```diff
@@ -316,6 +316,45 @@ describe('DashboardFiltersStore', () => {
     expect(hydratedStore.periodControl.value.to!.valueOf()).toBe(2000000);
   });
 
+  it('should drop URL-hydrated conditions for filters the surface does not offer', () => {
```
paulatulis: **MAJOR** — This test can't exercise the ordering window the prune actually has to survive.

The `FILTER_DEFINITION_PROVIDER` mock returns `of([...])`, which emits synchronously — so by the time the assertion runs, definitions have always already arrived. The real path is the opposite: definitions have *not* loaded when the URL is hydrated, and conditions must stay visible until they arrive and be pruned only on emission.

That window is exactly where the transient error-flash behaviour lives (the initial query fires with the unsupported filter before the prune can run), so it's the case most worth pinning.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/analytics_engine/use_case/GetAnalyticsFilterDefinitionsUseCaseTest.java
```diff
@@ -0,0 +1,67 @@
...
+import io.gravitee.apim.core.observability.model.FilterType;
+import java.util.List;
+import org.junit.jupiter.api.DisplayNameGeneration;
+import org.junit.jupiter.api.DisplayNameGenerator;
+import org.junit.jupiter.api.Test;
+
+@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
+class GetAnalyticsFilterDefinitionsUseCaseTest {
+
+    private final AnalyticsDefinitionQueryService definition = mock(AnalyticsDefinitionQueryService.class);
```
paulatulis, line 43: **MAJOR (low confidence)** — This mocks the `AnalyticsDefinitionQueryService` port; the testing conventions call for in-memory port fakes rather than Mockito mocks on ports.

Flagging with a caveat: no `InMemoryAnalyticsDefinitionQueryService` exists yet, and sibling use-case tests in this module use the same Mockito pattern. If that convention isn't actually enforced here, treat this as noise — it's a maintainer's call, not something I'd block on.

gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.ts
```diff
@@ -246,6 +252,8 @@ export class EnvLogsComponent {
...
       bodyText: filterMap.get('PAYLOAD')?.[0],
```
paulatulis, line 254: **MINOR** — The `PAYLOAD` -> `bodyText` mapping has no component-level wiring test.

Every other mapped field in the `filter wiring` describe block has one (API, PLAN, HTTP_PATH, TRANSACTION_ID, REQUEST_ID, ERROR_KEY, ...). PAYLOAD is the only mapped field left unexercised, so a rename or typo in the map key would ship silently.

gravitee-apim-console-webui/src/services-ngx/environment-logs.service.ts
```diff
@@ -118,6 +122,8 @@ function buildFilters(param?: SearchLogsParam): LogFilter[] {
     { name: 'ENTRYPOINT', values: param.entrypoints },
     { name: 'ERROR_KEY', values: param.errorKeys },
     { name: 'API_PRODUCT', values: param.apiProductIds },
+    { name: 'REQUEST_ID', values: param.requestIds },
```
paulatulis, line 125: **MINOR** — The scalar and plural variants of the same id can both be populated, and precedence is undefined.

`buildFilters()` builds `scalarFilters` from `requestId`/`transactionId` (`EQ`) and `arrayFilters` from `requestIds`/`transactionIds` (`IN`) independently, then concatenates both into the same `filters[]` array with no mutual-exclusion check. A caller populating both would emit two entries with the same `name` but different operators, and `openapi-logs.yaml` says nothing about how the backend resolves duplicate same-name filter entries.

Latent today — no caller sets both, and the scalar deep-link path is untouched — but nothing pins the behaviour either way.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/model/FilterSpec.java
```diff
@@ -27,8 +28,29 @@ public record FilterSpec(
...
     NumberRange range,
     List<FilterOperator> operators,
-    List<ApiSpec.Name> apis
+    List<ApiSpec.Name> apis,
+    List<FilterSignal> signals
```
paulatulis, line 37: **MINOR** — The new `signals` component isn't defensively copied, so a caller-held mutable list can be mutated after construction, silently changing record equality and serialized output.

The sole current producer happens to pass `List.copyOf(...)`, and the pre-existing list components (`enumValues`, `operators`, `apis`) share the same gap — so this is a consistency nit rather than a live defect.

paulatulis, line 37: **MINOR** — The null-vs-empty contract for `signals` is load-bearing but documented nowhere on the record.

Null means "signals were not derived, no gating applies" and an empty list means "supported by no surface, hide everywhere" — the frontend gate depends on that distinction. The compat-constructor Javadoc only says signals are derived elsewhere, and the OpenAPI description omits it too.

Empty-as-"excluded everywhere" is a surprising sentinel that a future catalog author could trip over without ever hitting a test failure.

gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-filters.store.ts
```diff
@@ -226,6 +229,21 @@ export class DashboardFiltersStore {
...
         });
     }
+
+    // Conditions restored from the URL may reference filters this surface does not offer (e.g. a
+    // logs-only filter deep-linked onto an analytics dashboard) — drop them instead of querying an
+    // engine that rejects them (APIM-14828).
+    this.definitionsPruneSubscription?.unsubscribe();
+    if (conditions.length > 0 && this.definitionProvider) {
+      this.definitionsPruneSubscription = this.definitionProvider
+        .getDefinitions()
```
paulatulis, line 241: **MINOR** — Extra definitions fetch per hydration, and a failed fetch is invisible.

This subscribes to a fresh `getDefinitions()` HTTP request on every hydration, and `hydrateFromRoute` runs twice at startup (constructor snapshot plus the first synchronous emission of `route.queryParams`), plus again on any unrelated query-param change — each issuing another GET of `/observability/filters/definition` on top of the filter bar's own call. The service doesn't cache.

There's also no error handler: if the definitions request fails, it surfaces only as an unhandled RxJS error and the prune silently never applies.

gravitee-apim-console-webui/src/management/observability/data-access/observability-filters-api.service.ts
```diff
@@ -63,14 +63,26 @@ interface FilterValueItemApi {
...
 }
 
+/** Observability surface consuming the shared filter catalog — mirrors the backend FilterSignal enum. */
+export type ObservabilityFilterSignal = 'ANALYTICS' | 'LOGS';
+
+/**
+ * Surface hosting the filter bar. When provided, definitions carrying signals are narrowed to that
+ * surface so a page only offers filters its query engine supports (APIM-14828).
+ */
+export const OBSERVABILITY_FILTER_SIGNAL = new InjectionToken<ObservabilityFilterSignal>('OBSERVABILITY_FILTER_SIGNAL');
```
paulatulis, line 73: **MINOR** — Design observation on the ambient token.

`OBSERVABILITY_FILTER_SIGNAL` is injected `{ optional: true }` inside a shared `data-access` service, so `getDefinitions()` silently branches on whatever the enclosing feature component happened to provide. Both current consumers (dashboard-viewer, env-logs) do provide it, so nothing is broken today.

The concern is that a future consumer of `ObservabilityFiltersApiService` that forgets the token gets the ungated legacy behaviour with no compiler warning and no test failure — the failure mode is silent, which is what makes it worth noting.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/analytics_engine/use_case/GetAnalyticsFilterDefinitionsUseCaseTest.java
```diff
@@ -0,0 +1,67 @@
...
+            )
+        );
+
+        var specs = useCase.execute().specs();
+
+        assertThat(specs)
+            .extracting(FilterSpec::name, FilterSpec::signals)
+            .containsExactly(
+                tuple(FilterSpec.Name.API, List.of(FilterSignal.ANALYTICS, FilterSignal.LOGS)),
+                // Logs-only: rejected by the analytics engine (APIM-14828).
```
paulatulis: **MINOR** — Ticket reference in a test-body comment. The conventions permit the explanatory "why" prose but not the embedded ticket tag.

```suggestion
                // Logs-only: rejected by the analytics engine.
```

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/observability/ObservabilityFiltersDefinitionResourceTest.java
```diff
@@ -72,6 +74,28 @@ void should_return_filter_definitions() {
...
+    @Test
+    void should_advertise_filter_signals_per_surface() {
+        var response = rootTarget().request().get();
+
+        assertThat(response)
+            .hasStatus(200)
+            .asEntity(FilterSpecsResponse.class)
+            .extracting(FilterSpecsResponse::getData)
+            .satisfies(filters -> {
+                // Logs-only filters must not be offered on the analytics surface (APIM-14828).
```
paulatulis: **MINOR** — Ticket reference in a test-body comment. The conventions permit the explanatory "why" prose but not the embedded ticket tag.

```suggestion
                // Logs-only filters must not be offered on the analytics surface.
```

gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.ts
araffass: Legit, but the backend uri filter is a wildcard query and single value by design, so real IN support means an or-of-wildcards across the model, crud and the ES adapter. Out of scope here. Keeping the LOGS signal since EQ works today, removing it would drop the filter from the logs page entirely. Opening a follow up ticket for multi value uri.

gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-filters.store.ts
araffass: Good catch, fixed. The prune keeps the same array reference when nothing is dropped, added a test asserting it.

araffass: Fixed the URL part, syncToRouter now runs when something is actually pruned. The transient first query before definitions arrive is still possible, killing it completely would mean holding every initial query on the catalog fetch which felt worse. Documented in the code, it self heals after the prune.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-analytics.yaml
araffass: Fixed. The description now says it is only populated by the observability definitions operation, absent means no restriction and empty means offered nowhere. Same contract documented on the core record.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetAnalyticsFilterDefinitionsUseCase.java
araffass: Added a guard test: every logs FilterName must be a catalog name, the URI translation or a documented exception, otherwise it fails pointing at supportsLogs. Not compile time safe but drift can no longer ship silently.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/model/FilterSpec.java
araffass: Agreed on the direction, keeping it out of this bugfix as you suggested. Will do it when the next field shows up.

gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-filters.store.spec.ts
araffass: Right, rewritten with a Subject. It now asserts conditions are still visible before the emission and pruned after, plus the router sync.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/analytics_engine/use_case/GetAnalyticsFilterDefinitionsUseCaseTest.java
araffass: Kept the mock, sibling use case tests in this module do the same and there is no in-memory fake for this port yet.

gravitee-apim-console-webui/src/management/observability/env-logs/env-logs.component.ts
araffass: Added.

gravitee-apim-console-webui/src/services-ngx/environment-logs.service.ts
araffass: Fixed, the plural variant wins and the scalar is skipped, with a test.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/model/FilterSpec.java
araffass: Leaving as is to stay consistent with the other list components. Happy to do a List.copyOf pass on the whole record separately.

araffass: Documented on the record javadoc and in the openapi description.

gravitee-apim-console-webui/src/management/observability/dashboards/ui/dashboard-viewer/dashboard-filters.store.ts
araffass: Added a catchError so a failed fetch keeps the restored filters instead of failing silently. Left the extra GET, the endpoint is cheap and caching it in the service would change freshness for the add filter dialog.

gravitee-apim-console-webui/src/management/observability/data-access/observability-filters-api.service.ts
araffass: Fair point, it is the tradeoff of the optional token. The two current pages now pin their token value with a test, so at least forgetting it on an existing page fails.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/analytics_engine/use_case/GetAnalyticsFilterDefinitionsUseCaseTest.java
araffass: Removed.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/test/java/io/gravitee/rest/api/management/v2/rest/resource/observability/ObservabilityFiltersDefinitionResourceTest.java
araffass: Removed.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetAnalyticsFilterDefinitionsUseCase.java
paulatulis: AnalyticsDefinitionMapper still carries only @ValueMapping(source = "PAYLOAD", target = THROW_EXCEPTION) while AnalyticsQueryValidator rejects four names (PAYLOAD, ERROR_KEY, REQUEST_ID, TRANSACTION_ID). So adding TRANSACTION_ID to any metric's yaml filter list still makes the per-metric endpoint advertise a filter the analytics engine rejects, with nothing to catch it. That was the "one yaml edit away" hazard, and it stands.

paulatulis (approved):
Left follow up comment to one thing. I will let you decide if you want to address it. Otherwise LGTM!
https://github.com/gravitee-io/gravitee-api-management/pull/18864#discussion_r3691371144

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/use_case/GetAnalyticsFilterDefinitionsUseCase.java
araffass: Fair, the guard I added only covered the logs axis. Now added a test that walks every metric's yaml filter list through the real definition and fails on any analytics rejected name, so the one yaml edit away case is a build failure. Also aligned the mapper's throw list with the validator's four names so runtime stays consistent if anything bypasses the yaml path.

PR #18886, opened 2026-07-31
ci: replace aqua with aikido

phiz71:
## Issue

No JIRA ticket.

## Description

We are moving away from Aqua in favour of Aikido, so this drops the Aqua integration from the CI and wires the new `gravitee-io/aikido` orb in its place. It is deliberately not a one-for-one replacement.

**First commit — remove the Aqua integration.** The `aquasec` orb, the `aquasec/fs_scan` job, Billy artifact registration, the image scans embedded in the docker build jobs, and the five Aqua secrets in Keeper are all gone. This also removes the `aquasec/fs_scan` failures currently breaking master: Trivy resolves the parent POMs of our imported BOMs from Maven Central, which now rate-limits the shared CircleCI IP with a `429` that recent Trivy versions surface as a fatal error.

**Second commit — scan docker images with Aikido.** The Aikido orb only covers container images, so the filesystem scan has no replacement here — Aikido covers repository code through its own SCM integration rather than through CI.

The Aikido local scanner bind-mounts the Docker socket, which `setup_remote_docker` does not provide, so the scan cannot live inside the build jobs the way the Aqua one did. It runs as its own machine-executor job, `job-aikido-scan-docker-images`, which pulls back the images the build jobs have just pushed. A single job scans all of them so that the layers they share — the java and nginx bases, and the variants of a given component — are pulled once instead of once per job.

Three other differences with the previous setup:

- Scans are **report only** for now (`fail_on: ''`). Findings reach the Aikido platform but never break the build, and every image is scanned since the orb stops at the first image failing its gate. Flipping this to `critical` later is a one-line change.
- The **chainguard and chainguard-fips variants are now scanned** as well; Aqua only covered the alpine, debian and UI images. Nothing is scanned on a release dry run, since no image is pushed there.
- **Only the amd64 side of our images is scanned.** The scanner reads images from the local Docker daemon, which only pulls the host architecture, so the `linux/arm64` half of these multi-arch images never reaches it.

## Additional context

- Orb: https://circleci.com/developer/orbs/orb/gravitee-io/aikido — source at https://github.com/gravitee-io/aikido-circleci-orb
- The Aikido CI API key is read from Keeper and exported as `AIKIDO_API_KEY`, following the same pattern as the other secrets in the pipelines.
- Since scans are report only, nothing in the CI will flag a vulnerability any more — this relies on someone watching the Aikido dashboard, or on notifications being configured on the platform side.
- A `4.12.x` branch with the same two commits is ready and will follow. Note that a Mergify backport further down (`4.11.x` and older) will not apply as is: those branches have no chainguard or chainguard-fips images, so the scan job has to be generated without those variants.


callaertanthony:
Went through this one carefully since a broken generated config breaks every pipeline. Looks good to me — a few questions below, none of them blocking.

### What I checked

| Check | Result |
|---|---|
| Aqua removal is complete | ✅ no `aqua`/`aquasec`/`billy` references left on the branch (only base64/cert false positives) |
| `requires:` targets always exist | ✅ traced all 4 workflows. In `pull_requests` the scan sits in `getMasterAndSupportJobs`, only reached when `isSupportBranchOrMaster`, where `filterJobs=false` so every build job is emitted unconditionally — no dangling `requires` on a partially-generated pipeline |
| Dry-run guard | ✅ `isProd && isDryRun` returns `[]`. `publish_docker_images` passes `isProd=false` and *does* push on a dry run (`!isProd \|\| !environment.isDryRun` → `--push`), so scanning there is correct rather than a miss |
| Registry/login split on a release | ✅ `needsAzurecrLogin` correctly picks up the FIPS-only azurecr images, and `cmd-docker-login-azurecr` is the same name the FIPS build job already registers, so the `Config` dedupe is a no-op instead of a silent override |
| `fail_on: ''` accepted by the real orb | ✅ `gravitee-io/aikido@1.0.2` declares `enum: ["", "low", "medium", "high", "critical"]`. The local `CustomEnumParameter` mirrors it exactly, and the `v1.0.2` tag exists |
| Shell in the pull step | ✅ `printf '%s\n' $IMAGES` word-splits on newlines as intended; image names contain no glob characters |
| Generated YAML | ✅ `generate-config` is green, the 19 regenerated resources match |

### Two questions

**1. `force_image_creation` defaults to `true` in the orb** — *"Create a new image in Aikido with the tag instead of updating the tag on existing image."* The scan targets `dockerTagsArgument(...)[0]`, which on master is the stable `master-latest` tag, rescanned on every merge. If that flag does what the description says, we end up with a new Aikido image record per master commit rather than an updated one. Worth a look at the dashboard after the first few runs — `force_image_creation: false` is a one-line fix if the inventory starts filling up.

**2. arm64 is never scanned.** The code comment says it ("only the host architecture of these multi-arch images is pulled — and therefore scanned"), but it isn't in the description's list of differences with the previous setup. Since we publish `linux/arm64`, that feels like it belongs next to the other two.

### Nits, take them or leave them

- `.circleci/ci/src/orbs/aikido.ts` — `docker_image_to_scan` is declared but never used by the job. Dead parameter.
- The scan is deliberately non-blocking (`fail_on: ''`) but the `docker pull` loop above it isn't, so a single registry hiccup fails the job the scan was designed never to fail. Not worth a `|| true` (we'd silently scan nothing), just noting the asymmetry.
- `dockerTagsArgument(...)[0]` couples the scan to tag ordering in the build job. Harmless today since every tag aliases the same image, but it's an implicit contract between the two.

### Unrelated

`ci/circleci: Validate backend` is red and this PR touches no Java — pre-existing or flaky. Worth a rerun before it goes in the queue.

One thing that reassured me on the coverage gap you flagged: the `Aikido Security: check code` GitHub check is already green on this PR, so the SCM-side scanning that replaces `fs_scan` is genuinely wired up. The gap is only that nothing gates the build any more, not that repository code has stopped being scanned.

phiz71:
Thanks for going through the generated config that carefully — the `requires:` tracing across the four workflows and the `cmd-docker-login-azurecr` dedupe were exactly the two things I wanted a second pair of eyes on.

**1. `force_image_creation`** — checked with the orb maintainer, and the flag groups **per tag, not per run**. A production tag like `4.12.3` feeds the `4.12.3` report, and `master-latest` feeds the `master-latest` report, which is reused on every merge. So no inventory growth on master.

Worth recording the other direction too: setting it to `false` would be actively wrong for us. A release scans several tags of the same component (`4.12.3`, `4.12.3-debian`, `4.12.3-chainguard`, `4.12.3-chainguard-fips`), and without the flag they would overwrite each other instead of each getting their own report — that is precisely what the orb's `1.0.2` fix was for. Keeping the default.

**2. arm64** — fair point, it belonged next to the other two. Added to the description of this PR and of the `4.12.x` / `4.11.x` ones.

**Nits** — `docker_image_to_scan` was indeed dead in our orb wrapper, removed (it does not appear in the generated YAML, so the config is unchanged). Leaving the other two as noted: the blocking `docker pull` versus the non-blocking scan is a real asymmetry, but a registry failure meaning "we scanned nothing" is worth a red job, and `dockerTagsArgument(...)[0]` staying coupled to tag ordering is fine while every tag aliases the same image.

**Two things moved since your review:**

- The PRs are now on `gravitee-io/aikido@1.0.3`. `1.0.2` was missing a space before a line continuation, which glued `--force-create-image-for-tag` to `--fail-on` — harmless while `fail_on` is empty, but it would have bitten us the day we turn gating on. Reported and fixed upstream, the `1.0.3` diff is that single space.
- `ci/circleci: Validate backend` is green again, it was flaky. The PR is at 49 passing checks.

And thanks for confirming the `Aikido Security: check code` angle — that was the part I could not verify from the CI config alone.

.circleci/ci/src/jobs/job-aikido-scan-docker-images.ts
```diff
@@ -0,0 +1,146 @@
...
+        label: 'APIM Management API',
+        image: config.components.managementApi.image,
+        variants: ['alpine', 'debian', ...chainguardVariants],
+      },
+      { label: 'APIM Gateway', image: config.components.gateway.image, variants: ['alpine', 'debian', ...chainguardVariants] },
+    ] as { label: string; image: string; variants: (Variant | undefined)[] }[];
+
+    const parsedGraviteeioVersion = parse(environment.graviteeioVersion);
+    const imagesToScan = components.flatMap(({ image, variants }) =>
+      variants.map((variant) => dockerTagsArgument(environment, parsedGraviteeioVersion, isProd, variant, image)[0]),
```
ytvnr, line 142: `[0]` is the mutable tag on non-prod pipelines. `dockerTagsArgument` pushes `<branch>-latest` first and `<branch>-<sha1>` second, so this scans `master-latest` rather than the image this pipeline just built.

CircleCI never auto-cancels redundant workflows on the **default branch** — it is excluded by design ([discuss](https://discuss.circleci.com/t/workaround-auto-cancel-redundant-builds-on-the-default-branch/39468)) — so two `master` pipelines can genuinely overlap, and this job can pull the image the *other* pipeline pushed. Findings then get attributed to the wrong commit. On support branches auto-cancel narrows the window but does not close it: a workflow that gets cancelled may already have pushed the tag.

On a release `[0]` is the version tag, so only the `pull_requests` and `publish_docker_images` workflows are affected. Selecting the sha1-pinned tag on non-prod would make the scan deterministic (careful not to just take the last tag — on a release the last one can be `latest`).

.circleci/ci/src/jobs/job-aikido-scan-docker-images.ts
```diff
@@ -0,0 +1,146 @@
...
+      new commands.Run({
+        name: 'Pull docker images to scan',
+        // The scanner only reads images from the local daemon, and only the host
+        // architecture of these multi-arch images is pulled — and therefore scanned.
+        command: [
+          'IMAGES="',
+          ...imagesToScan,
+          '"',
+          `printf '%s\\n' $IMAGES > ${IMAGE_LIST_FILE}`,
+          `while read -r image; do docker pull "$image"; done < ${IMAGE_LIST_FILE}`,
```
ytvnr: CircleCI runs steps with `bash -eo pipefail`, and `set -e` does apply to the *body* of a `while` loop (only the condition is exempt). So a single failing `docker pull` — image missing, registry throttling, transient network — aborts the job here, before `aikido/scan_docker_image` has run at all. None of the other images get reported.

Since the scan is report-only anyway, failing closed on the pull step is the worst of both worlds. Tolerating per-image failures and scanning whatever was pulled (dropping the failed entries from the list, and logging them) would be more robust.

.circleci/ci/src/jobs/job-aikido-scan-docker-images.ts
```diff
@@ -0,0 +1,146 @@
...
+      steps.push(new reusable.ReusedCommand(azurecrLoginCommand));
+    }
+
+    steps.push(
+      new commands.Run({
+        name: 'Pull docker images to scan',
+        // The scanner only reads images from the local daemon, and only the host
+        // architecture of these multi-arch images is pulled — and therefore scanned.
+        command: [
+          'IMAGES="',
```
ytvnr, line 75: Minor, on the shape of this step: it builds `IMAGES` with embedded newlines, relies on the unquoted `$IMAGES` word splitting to re-split it, then rewrites it one per line. It works — image references contain no whitespace — but it depends on `IFS` and will trip shellcheck. A heredoc writes the file directly:

```sh
cat > /tmp/aikido-images.txt <<'EOF'
graviteeio/apim-gateway:4.2.0
...
EOF
xargs -n1 docker pull < /tmp/aikido-images.txt
```

Same result, one less indirection.

.circleci/ci/src/jobs/job-aikido-scan-docker-images.ts
```diff
@@ -0,0 +1,146 @@
...
+      new reusable.ReusedCommand(dockerLogoutCommand),
+    );
+
+    if (needsAzurecrLogin) {
+      const azurecrLogoutCommand = DockerLogoutCommand.get(environment, false, 'cmd-docker-logout-azurecr');
+      dynamicConfig.addReusableCommand(azurecrLogoutCommand);
+      steps.push(new reusable.ReusedCommand(azurecrLogoutCommand));
+    }
+
+    return new Job(AikidoScanDockerImagesJob.jobName, UbuntuExecutor.create(), steps);
```
ytvnr, line 108: `UbuntuExecutor.create()` defaults to `resource_class: medium` with `docker_layer_caching: false`, and this job pulls 17 images serially — several GB of java and nginx layers, on every master commit and every release.

Grouping them into a single job is the right call — the java/nginx bases and the variants of a component share layers and get pulled once. Just worth measuring the wall time of the first runs: if it lands well above what the inline Aqua scans cost, a larger resource class or parallel pulls would pay for themselves. Not blocking.

.circleci/ci/src/jobs/job-aikido-scan-docker-images.ts
```diff
@@ -0,0 +1,146 @@
...
+          `printf '%s\\n' $IMAGES > ${IMAGE_LIST_FILE}`,
+          `while read -r image; do docker pull "$image"; done < ${IMAGE_LIST_FILE}`,
+        ].join('\n'),
+      }),
+      new reusable.ReusedCommand(orbs.aikido.commands['scan_docker_image'], {
+        built_docker_image_file: IMAGE_LIST_FILE,
+        // Report only: findings are pushed to the Aikido platform but never break the
+        // build. This also keeps every image scanned, since the orb stops at the first
+        // image that fails its gate.
+        fail_on: '',
```
ytvnr, line 97: Combined with the removal of `aquasec/fs_scan`, this means **nothing in the CI fails on a vulnerability any more**, on any branch. That is a deliberate call and it is spelled out in the description, so no objection — but there is no ticket tracking it, and a green CI is exactly what it looked like before.

Could we open follow-ups for:
1. flipping `fail_on` to `critical` once the noise level is known,
2. arm64 coverage — only the amd64 half of these multi-arch images is scanned,
3. and above all, confirming the Aikido SCM integration is actually live on this repo.

(3) is the one that worries me: it is what replaces the filesystem scan, it lives entirely outside this repo, and if it is not enabled we have dropped code scanning with nothing in its place and no signal anywhere.

ytvnr (approved):
Reviewed alongside #18886 / #18894 / #18895 — the three are consistent and the divergences are all justified by the shape of each branch (SDK import, Gamma Console, chainguard variants, second azurecr login). No leftover `aqua` reference under `.circleci`.

The generation looks correct: the `requires` list is derived from the same `components` list as the image list, so the two cannot drift apart; the `isProd && isDryRun` guard matches the workflows that do not push; the scan sits in `getMasterAndSupportJobs`, so it does not appear on branches that build no image; and `cmd-docker-login-azurecr` reuses the naming already introduced by the chainguard-fips job. I checked the generated `pull-requests-master.yml` and `release-4-2-0-no-dry-run.yml` snapshots and every `requires` entry maps to a job that exists.

Approving — the comments below are one correctness point (the mutable tag) plus robustness and follow-up items, none of which need to block the migration.

Two small nits not worth inline comments:
- the logout step has no `when: always`, so credentials survive a failing pull in the machine's docker config (ephemeral executor, so near-zero impact);
- `orb.spec.ts` / `workflow.spec.ts` pin the orb to `1.0.1` while `config.orbs.aikido` is `1.0.3` — harmless literal, but it makes the reader hesitate.

Note: I reviewed the sources and the generated snapshots, I did not run the `.circleci/ci` test suite.

phiz71:
Thanks for reviewing the three together and for checking the snapshots — the `requires`/image-list coupling and the `getMasterAndSupportJobs` placement were the two things I most wanted confirmed from the outside.

**Failing `docker pull`** — you are right, and this is now fixed on the three branches. Each pull is attempted independently, images that fail are dropped from the list with a warning, and the scan runs on whatever was pulled. One guard on top of your suggestion: if *no* image could be pulled the step still exits 1, so the job can never report a green, silently empty scan — that failure mode is infrastructure, not a finding.

```sh
: > /tmp/aikido-images.txt
for image in $IMAGES; do
  if docker pull "$image"; then
    echo "$image" >> /tmp/aikido-images.txt
  else
    echo "WARN: could not pull $image, it will not be scanned"
  fi
done
if [ ! -s /tmp/aikido-images.txt ]; then
  echo "ERROR: none of the images could be pulled"
  exit 1
fi
```

**The mutable tag** — the analysis is correct: on master and support branches `[0]` is `<branch>-latest`, two pipelines can overlap since CircleCI does not auto-cancel on the default branch, and the scan can end up pulling the image the other pipeline pushed.

We are keeping it anyway, and the reason is worth recording here. The orb's `force_image_creation` (default `true`, confirmed with its maintainer on the other thread) groups results **per tag**: `4.12.3` feeds the `4.12.3` report, `master-latest` feeds the `master-latest` report that is reused on every merge. Switching the scan to the sha1-pinned tag would make it deterministic, but every commit would then create its own Aikido report — the inventory would grow one entry per merge, forever. Given the scan is report only and the `master-latest` report claims to describe the current state of master rather than a specific commit, scanning the latest pushed image stays coherent with what the report says it is. We trade per-commit attribution, which Aqua did not give us either, for a bounded inventory.

Happy to revisit if per-commit attribution turns out to matter once we look at real results.

**`orb.spec.ts` / `workflow.spec.ts` pinned to `1.0.1`** — leftover from renaming those SDK fixtures, aligned to `1.0.3`.

**The other nits** — the missing `when: always` on logout and the resource class are noted; the executor is ephemeral for the first, and the second is worth measuring on the first real runs rather than guessing now, since grouping the pulls into one job was precisely meant to keep that cost down. On the heredoc: it reads better, but `<<` is also CircleCI's parameter-interpolation delimiter, so I would rather not put a heredoc inside a generated config without a good reason.

**Follow-ups** — agreed on all three, and (3) is already answered: the `Aikido Security: check code` GitHub check is green on these PRs, so the SCM-side scanning that replaces `fs_scan` is live on this repository. The gap is the gating, not the coverage. I will open tickets for flipping `fail_on` to `critical` and for arm64.

PR #18893, opened 2026-07-31
fix(console): use environment API key header for new plans

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/APIM-14786

## Description

Fixes API Key plan creation so that the `API Key Header` field is no longer pre-filled with the static `X-Gravitee-Api-Key` value.

When no header is explicitly configured, the plan does not persist `apiKeyHeader`, allowing the Gateway to use its environment-level API Key header configuration.

Explicitly configured custom headers and existing plan edit behavior remain unchanged.

## Changes

- Remove the `apiKeyHeader` schema default when creating an API Key plan.
- Preserve explicitly entered API Key headers.
- Add regression tests covering the schema, creation form, and persisted configuration.

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/component/plan/2-secure-step/plan-edit-secure-step.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/api/component/plan/2-secure-step/plan-edit-secure-step.component.spec.ts
@@ -88,14 +88,20 @@ describe('PlanEditSecureStepComponent — API Key header schema', () => {
     });
   });
 
-  it('keeps apiKeyHeader default in schema on create', () => {
+  it('strips apiKeyHeader default from schema on create', () => {
     component.mode = 'create';
     fixture.detectChanges();
 
     schema$.next(apiKeySchemaWithDefault);
     fixture.detectChanges();
 
-    expect(component.securityConfigSchema).toBe(apiKeySchemaWithDefault);
+    expect(component.securityConfigSchema).toEqual({
+      type: 'object',
+      properties: {
+        source: { type: 'string', default: 'HEADER' },
+        apiKeyHeader: { type: 'string' },
+      },
+    });
   });
 
   it('keeps apiKeyHeader default in schema on edit when stored plan already has a header', () => {
--- a/gravitee-apim-console-webui/src/management/api/component/plan/api-plan-form.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/api/component/plan/api-plan-form.component.spec.ts
@@ -92,6 +92,15 @@ const fakeApiKeySchema = {
   additionalProperties: false,
 };
 
+const fakeApiKeySchemaWithHeaderDefault = {
+  ...fakeApiKeySchema,
+  properties: {
+    ...fakeApiKeySchema.properties,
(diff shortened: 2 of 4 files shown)


jaroslawlakomy-grav:
## Manual Verification Plan

### Selected approach

The selected approach removes the static `apiKeyHeader` default from the API Key policy schema before rendering the plan creation form.

The environment-level header is intentionally not loaded into the Console. When no header is explicitly entered, `apiKeyHeader` remains absent from the plan configuration. This allows the Gateway to resolve the header from its environment-level `policy.api-key.header` configuration.

Explicitly entered plan-level headers remain supported and are persisted normally.

### Preconditions

- Use an APIM 4.12 environment containing this fix.
- A V4 API already exists.
- The API Key plan type is available.
- The test account can create, edit, and publish plans.
- For the end-to-end scenario, the Gateway should use a recognizable environment-level API Key header, for example `X-Environment-Api-Key`.

### Scenario 1: Verify the creation form

1. Log in to the APIM Console.
2. Go to **APIs**.
3. Open an existing V4 API.
4. Open **Plans**.
5. Click **Add new plan**.
6. Select **API Key**.
7. Enter a unique plan name in the **General** step.
8. Click **Next**.
9. Locate **API Key Header** in the **API Key authentication configuration** step.
10. Verify that the field is empty.

Expected result:

- `X-Gravitee-Api-Key` is not automatically displayed.
- No plan-level header is introduced without an explicit user action.

### Scenario 2: Verify the saved plan

1. Leave **API Key Header** empty.
2. Complete the remaining plan creation steps.
3. Save the plan.
4. Open the newly created plan for editing.
5. Open its **API Key authentication configuration** step.
6. Verify that **API Key Header** is still empty.
7. Save the plan without modifying the field.
8. Reopen it and verify the field again.

Expected result:

- The field remains empty after creation and subsequent edits.
- The plan does not persist a schema-generated `apiKeyHeader`.

### Scenario 3: Verify an explicit custom header

1. Create another API Key plan.
2. Enter a unique plan name.
3. Open the **API Key authentication configuration** step.
4. Enter `X-Custom-Api-Key` in **API Key Header**.
5. Enable the custom API Key header option if it is displayed by the installed policy version.
6. Complete and save the plan.
7. Reopen the plan.

Expected result:

- `X-Custom-Api-Key` is displayed after reopening the plan.
- Explicit user configuration is not removed by the fix.

### Scenario 4: Verify Gateway fallback

1. Ensure the Gateway is configured with an environment-level API Key header, for example:

   `policy.api-key.header: X-Environment-Api-Key`

2. Publish the plan created with an empty **API Key Header** field.
3. Deploy the API.
4. Create or approve an application subscription to the plan.
5. Retrieve the generated API key.
6. Send a request using:

   `X-Environment-Api-Key: <generated-api-key>`

7. Verify that the request succeeds.
8. Send the same request using only:

   `X-Gravitee-Api-Key: <generated-api-key>`

Expected result:

- The request using `X-Environment-Api-Key` is authenticated.
- The plan does not force the static `X-Gravitee-Api-Key` header.
- The Gateway uses its environment-level API Key header configuration.

dev-marek (approved):
lgtm

PR #18894, opened 2026-07-31
ci: replace aqua with aikido [4.12.x]

phiz71:
## Issue

No JIRA ticket.

## Description

Backport of #18886 to `4.12.x`: we are moving away from Aqua in favour of Aikido, so this drops the Aqua integration from the CI and wires the new `gravitee-io/aikido` orb in its place. It is deliberately not a one-for-one replacement.

**First commit — remove the Aqua integration.** The `aquasec` orb, the `aquasec/fs_scan` job, Billy artifact registration, the image scans embedded in the docker build jobs, and the five Aqua secrets in Keeper are all gone. This also removes the `aquasec/fs_scan` failures: Trivy resolves the parent POMs of our imported BOMs from Maven Central, which now rate-limits the shared CircleCI IP with a `429` that recent Trivy versions surface as a fatal error.

**Second commit — scan docker images with Aikido.** The Aikido orb only covers container images, so the filesystem scan has no replacement here — Aikido covers repository code through its own SCM integration rather than through CI.

The Aikido local scanner bind-mounts the Docker socket, which `setup_remote_docker` does not provide, so the scan cannot live inside the build jobs the way the Aqua one did. It runs as its own machine-executor job, `job-aikido-scan-docker-images`, which pulls back the images the build jobs have just pushed. A single job scans all of them so that the layers they share — the java and nginx bases, and the variants of a given component — are pulled once instead of once per job.

Three other differences with the previous setup:

- Scans are **report only** for now (`fail_on: ''`). Findings reach the Aikido platform but never break the build, and every image is scanned since the orb stops at the first image failing its gate.
- The **chainguard and chainguard-fips variants are now scanned** as well; Aqua only covered the alpine, debian and UI images. Nothing is scanned on a release dry run, since no image is pushed there.
- **Only the amd64 side of our images is scanned.** The scanner reads images from the local Docker daemon, which only pulls the host architecture, so the `linux/arm64` half of these multi-arch images never reaches it.

## Additional context

- Orb: https://circleci.com/developer/orbs/orb/gravitee-io/aikido — source at https://github.com/gravitee-io/aikido-circleci-orb
- The Aikido CI API key is read from Keeper and exported as `AIKIDO_API_KEY`, following the same pattern as the other secrets in the pipelines.
- Differences with the master PR, both mechanical: this branch still uses `@circleci/circleci-config-sdk` instead of the in-repo `src/circleci-config`, so the two new files import from it, and the master-only SDK tests renamed there are not part of this backport.
- The scan is wired into the four workflows Aqua used to scan from (`build-docker-images`, `full-release`, `publish-docker-images`, `pull-requests`). The standalone `build-chainguard-images` and `build-chainguard-fips-images` workflows, which only exist on this branch, are left untouched — Aqua did not scan from them either.
- A Mergify backport further down (`4.11.x` and older) will not apply as is: those branches have no chainguard or chainguard-fips images, so the scan job has to be generated without those variants.


.circleci/ci/src/workflows/workflow-pull-requests.ts
```diff
@@ -845,6 +829,9 @@ export class PullRequestsWorkflow {
             }),
           ]
         : []),
+
+      // Aikido image scans, once every variant of a component has been pushed
+      ...AikidoScanDockerImagesJob.workflowJobs(dynamicConfig, environment, false, '', true),
```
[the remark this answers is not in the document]
phiz71: @AikidoSec ignore: pre-existing file, out of scope for this PR. This change adds one import and one line to a workflow builder that was already this size; splitting it would be an unrelated refactor, which the repository guidelines explicitly rule out for feature PRs.

.circleci/ci/src/jobs/job-aikido-scan-docker-images.ts
```diff
@@ -0,0 +1,147 @@
...
+        label: 'APIM Management API',
+        image: config.components.managementApi.image,
+        variants: ['alpine', 'debian', ...chainguardVariants],
+      },
+      { label: 'APIM Gateway', image: config.components.gateway.image, variants: ['alpine', 'debian', ...chainguardVariants] },
+    ] as { label: string; image: string; variants: (Variant | undefined)[] }[];
+
+    const parsedGraviteeioVersion = parse(environment.graviteeioVersion);
+    const imagesToScan = components.flatMap(({ image, variants }) =>
+      variants.map((variant) => dockerTagsArgument(environment, parsedGraviteeioVersion, isProd, variant, image)[0]),
```
ytvnr, line 143: `[0]` is the mutable tag on non-prod pipelines. `dockerTagsArgument` pushes `<branch>-latest` first and `<branch>-<sha1>` second, so this scans `master-latest` rather than the image this pipeline just built.

CircleCI never auto-cancels redundant workflows on the **default branch** — it is excluded by design ([discuss](https://discuss.circleci.com/t/workaround-auto-cancel-redundant-builds-on-the-default-branch/39468)) — so two `master` pipelines can genuinely overlap, and this job can pull the image the *other* pipeline pushed. Findings then get attributed to the wrong commit. On support branches auto-cancel narrows the window but does not close it: a workflow that gets cancelled may already have pushed the tag.

On a release `[0]` is the version tag, so only the `pull_requests` and `publish_docker_images` workflows are affected. Selecting the sha1-pinned tag on non-prod would make the scan deterministic (careful not to just take the last tag — on a release the last one can be `latest`).

.circleci/ci/src/jobs/job-aikido-scan-docker-images.ts
```diff
@@ -0,0 +1,147 @@
...
+      new commands.Run({
+        name: 'Pull docker images to scan',
+        // The scanner only reads images from the local daemon, and only the host
+        // architecture of these multi-arch images is pulled — and therefore scanned.
+        command: [
+          'IMAGES="',
+          ...imagesToScan,
+          '"',
+          `printf '%s\\n' $IMAGES > ${IMAGE_LIST_FILE}`,
+          `while read -r image; do docker pull "$image"; done < ${IMAGE_LIST_FILE}`,
```
ytvnr: CircleCI runs steps with `bash -eo pipefail`, and `set -e` does apply to the *body* of a `while` loop (only the condition is exempt). So a single failing `docker pull` — image missing, registry throttling, transient network — aborts the job here, before `aikido/scan_docker_image` has run at all. None of the other images get reported.

Since the scan is report-only anyway, failing closed on the pull step is the worst of both worlds. Tolerating per-image failures and scanning whatever was pulled (dropping the failed entries from the list, and logging them) would be more robust.

.circleci/ci/src/jobs/job-aikido-scan-docker-images.ts
```diff
@@ -0,0 +1,147 @@
...
+      steps.push(new reusable.ReusedCommand(azurecrLoginCommand));
+    }
+
+    steps.push(
+      new commands.Run({
+        name: 'Pull docker images to scan',
+        // The scanner only reads images from the local daemon, and only the host
+        // architecture of these multi-arch images is pulled — and therefore scanned.
+        command: [
+          'IMAGES="',
```
ytvnr, line 76: Minor, on the shape of this step: it builds `IMAGES` with embedded newlines, relies on the unquoted `$IMAGES` word splitting to re-split it, then rewrites it one per line. It works — image references contain no whitespace — but it depends on `IFS` and will trip shellcheck. A heredoc writes the file directly:

```sh
cat > /tmp/aikido-images.txt <<'EOF'
graviteeio/apim-gateway:4.2.0
...
EOF
xargs -n1 docker pull < /tmp/aikido-images.txt
```

Same result, one less indirection.

.circleci/ci/src/jobs/job-aikido-scan-docker-images.ts
```diff
@@ -0,0 +1,147 @@
...
+      new reusable.ReusedCommand(dockerLogoutCommand),
+    );
+
+    if (needsAzurecrLogin) {
+      const azurecrLogoutCommand = DockerLogoutCommand.get(environment, false, 'cmd-docker-logout-azurecr');
+      dynamicConfig.addReusableCommand(azurecrLogoutCommand);
+      steps.push(new reusable.ReusedCommand(azurecrLogoutCommand));
+    }
+
+    return new Job(AikidoScanDockerImagesJob.jobName, UbuntuExecutor.create(), steps);
```
ytvnr, line 109: `UbuntuExecutor.create()` defaults to `resource_class: medium` with `docker_layer_caching: false`, and this job pulls 17 images serially — several GB of java and nginx layers, on every push to this branch and every release.

Grouping them into a single job is the right call — the java/nginx bases and the variants of a component share layers and get pulled once. Just worth measuring the wall time of the first runs: if it lands well above what the inline Aqua scans cost, a larger resource class or parallel pulls would pay for themselves. Not blocking.

.circleci/ci/src/jobs/job-aikido-scan-docker-images.ts
```diff
@@ -0,0 +1,147 @@
...
+          `printf '%s\\n' $IMAGES > ${IMAGE_LIST_FILE}`,
+          `while read -r image; do docker pull "$image"; done < ${IMAGE_LIST_FILE}`,
+        ].join('\n'),
+      }),
+      new reusable.ReusedCommand(orbs.aikido.commands['scan_docker_image'], {
+        built_docker_image_file: IMAGE_LIST_FILE,
+        // Report only: findings are pushed to the Aikido platform but never break the
+        // build. This also keeps every image scanned, since the orb stops at the first
+        // image that fails its gate.
+        fail_on: '',
```
ytvnr, line 98: Combined with the removal of `aquasec/fs_scan`, this means **nothing in the CI fails on a vulnerability any more**, on any branch. That is a deliberate call and it is spelled out in the description, so no objection — but there is no ticket tracking it, and a green CI is exactly what it looked like before.

Could we open follow-ups for:
1. flipping `fail_on` to `critical` once the noise level is known,
2. arm64 coverage — only the amd64 half of these multi-arch images is scanned,
3. and above all, confirming the Aikido SCM integration is actually live on this repo.

(3) is the one that worries me: it is what replaces the filesystem scan, it lives entirely outside this repo, and if it is not enabled we have dropped code scanning with nothing in its place and no signal anywhere.

ytvnr (approved):
Reviewed alongside #18886 / #18894 / #18895 — the three are consistent and the divergences are all justified by the shape of each branch. This one correctly imports from `@circleci/circleci-config-sdk` rather than the in-repo `src/circleci-config`, and leaves out the master-only SDK spec changes. No leftover `aqua` reference under `.circleci`.

Same comments as on #18886, repeated here so they are not lost in the backport. Approving.

Note: I reviewed the sources and the generated snapshots, I did not run the `.circleci/ci` test suite.

phiz71:
@ytvnr your five comments apply identically here, so to keep the discussion in one place they are answered on the master PR: https://github.com/gravitee-io/gravitee-api-management/pull/18886#issuecomment-5168146325

Short version: the failing `docker pull` is fixed on this branch too (per-image tolerance, plus an exit 1 when nothing could be pulled at all), and we are deliberately keeping the mutable `<branch>-latest` tag — scanning the sha1-pinned one would create one Aikido report per merge, since the orb groups results per tag.

PR #18913, opened 2026-07-31
fix(observability): apply every filter offered on the environment logs screen

SamHardi:
## Summary

On *Observability → Logs*, most filters shown as active were never applied: 37 of the 46 offered
filters were dropped between the filter bar and the search request, so the chip looked active while
the result set stayed unfiltered ([APIM-14817](https://gravitee.atlassian.net/browse/APIM-14817)).
The mirror problem existed on the dashboards screen, where logs-only filters were offered and then
rejected with a 400 ([APIM-14828](https://gravitee.atlassian.net/browse/APIM-14828)).

The root cause is shared: the filter catalog had no notion of which observability signal a filter
belongs to, so `GET /observability/filters/definition` returned the whole analytics catalog to both
screens. This PR adds that axis and lets each screen narrow by it, removes the console-side
translation that silently lost filters, and wires the filters the logs engine could already express.

## Changes

**Catalog**
- `FilterSpec` gains a `signals` axis, populated from `analytics-definition.yaml`. Entries without it
  default to `ANALYTICS`, so the diff is additive and every `LOGS` exposure is an explicit opt-in.
- `GET /observability/filters/definition` accepts an optional `?signal=`. Without it the response is
  unchanged.
- `AnalyticsQueryValidator` now reads that axis instead of keeping its own copy of the logs-only
  names. A name the catalog does not describe stays accepted, since it omits several the engines
  still know (`URI`, `ENTRYPOINT`, the edge ones).

**Logs engine** — these reach the search for the first time, mostly filling slots `SearchLogsFilters`
already had:
- `HTTP_STATUS_CODE_GROUP` (the replacement for the old "above 400/500" presets)
- `HTTP_STATUS` with `GTE`/`LTE`
- an exact gateway response time
- `API_TYPE`, which narrows the API scope rather than adding a clause, as a log row carries no type
  of its own

**Vocabulary** — the Management API accepts the catalog spellings `HTTP_PATH`,
`HTTP_GATEWAY_RESPONSE_TIME` and `MCP_PROXY_METHOD`, aliasing them to `URI`, `RESPONSE_TIME` and
`MCP_METHOD`. The historical names stay accepted so saved links keep working.

**Refusals** — a path filter with `IN`, and a payload filter with anything but `CONTAINS`, are now
refused instead of being accepted and ignored.

**Console**
- The logs screen forwards every active condition instead of mapping a known subset, and asks for
  `?signal=LOGS`.
- The dashboards screen asks for `?signal=ANALYTICS`, which keeps `TRANSACTION_ID` and the other
  logs-only filters out of its filter bar.
- The `<Signal> only` pill becomes a coloured icon with the wording in a tooltip, always in the same
  reserved slot at the end of the row. On a screen scoped to one signal most rows carried the same
  pill, crowding out the API-type badges that do differ.

## API surface

|---|---|---|---|---|
| `/environments/{envId}/observability/filters/definition` | `GET` | Optional `?signal=` narrows the catalog. Omitted, the response is unchanged. Each `FilterSpec` gains `signals[]`. | `?signal=LOGS` | `{ "data": [ { "name": "HTTP_METHOD", "signals": ["LOGS","ANALYTICS"], ... } ] }` — 15 entries for `LOGS`, 42 for `ANALYTICS` |
| `/environments/{envId}/logs/search` | `POST` | Accepts the catalog spellings (`HTTP_PATH`, `HTTP_GATEWAY_RESPONSE_TIME`, `MCP_PROXY_METHOD`) alongside the historical ones, plus `HTTP_STATUS_CODE_GROUP`, `API_TYPE`, and `HTTP_STATUS` with `GTE`/`LTE`. | `{"filters":[{"name":"HTTP_STATUS_CODE_GROUP","operator":"IN","value":["5XX"]}]}` | `200` — result now narrowed; previously the filter was dropped and the full set returned |
| `/environments/{envId}/logs/search` | `POST` | **Behaviour change**: `HTTP_PATH` with `IN`, and `PAYLOAD` with anything but `CONTAINS`, are refused. | `{"filters":[{"name":"HTTP_PATH","operator":"IN","value":["/a","/b"]}]}` | `400` — previously `200` with the filter silently ignored |

No breaking change: `?signal=` is optional, the added filter names widen a request enum, and `signals[]`
is an added response field. The portal does not consume this catalog.

## Test plan

- [ ] On *Observability → Logs*, open **Add filter**: exactly 15 entries, with a marker icon on
      Error Key, Payload content, Request ID and Transaction ID. `Gateway`, `Endpoint Response Time`,
      `Country` and the other analytics-only filters are gone.
- [ ] Apply **Status Code Group = 5XX**: only 5xx rows come back. Check the request body carries the
      filter.
- [ ] Apply three at once — Status Code Group, API and HTTP Method: all three appear in the request
      body and the result is their intersection, none silently dropped.
- [ ] Apply **Gateway Response Time = <an exact value>**: the result narrows. This one used to return
      the unfiltered total.
- [ ] Apply **API Type = HTTP Proxy**: only HTTP proxy logs. Pick a type the logs signal does not
      serve (Message): zero rows, not every row.
- [ ] On *Observability → Overview*, open **Add filter**: `Transaction ID`, `Request ID`,
      `Payload content` and `Error Key` are absent. Adding any other filter still works.
- [ ] Open a logs URL saved before this change (one carrying `HTTP_PATH`): it still resolves.
- [ ] Regression: the Overview and Dashboards screens still load and filter as before.

### Edge cases to check
- A filter whose value matches nothing must return zero rows, not the unfiltered total — that
  distinction is what the bug was about.
- `HTTP_PATH` now offers `EQ` only; `IN` over a path was never applied and is now refused.

## Reviewer notes

The load-bearing piece is `ObservabilityLogsFilterCoverageTest`: it loads the real definition file
and asserts that every filter advertised for `LOGS`, for every operator advertised, reaches
`SearchLogsFilters`. It builds the request shapes a client can actually send — the schema
discriminates on the operator, so `EQ` always arrives as a string filter — which is what catches a
filter wired on a path no client takes. Adding a `signals: [LOGS]` entry without wiring its
translation now fails there rather than in production.

Two filters stay off the logs screen on purpose: `Gateway` and `Endpoint Response Time` have no slot
in `SearchLogsFilters`, so they cannot be expressed rather than merely being unwired. The ticket
accepts either supporting a filter end to end or not offering it.

Gamma observability, which already solved this problem by the route this PR takes, reaches the same
conclusion: neither appears among its 23 `LOGS` filters, and forcing one returns
`Filter 'GATEWAY' does not apply to signal LOGS` — a 400, same as here. Both are served on its
analytics surface. So this is a missing engine capability, not a catalog entry to flip, and wiring
them is its own ticket — `Gateway` additionally needs an id-to-hostname resolution that does not
exist on the logs path.

#18864 landed on master while this was in review, and this branch is rebased on top of it. It fixed
APIM-14828 alone, by a different route — signals derived at runtime from the two engines rather than
declared in the catalog, and narrowing applied client-side rather than through `?signal=`. Where the
two overlap, this PR's version is kept: it also removes the console-side mapping that dropped the
other 35 filters, which #18864 left in place (it added `REQUEST_ID` and `TRANSACTION_ID` to that same
hardcoded map).

Two pieces of #18864 are **kept** rather than superseded, because nothing here replaced them:
- `DashboardFiltersStore` now prunes URL-hydrated conditions the surface does not offer. Narrowing the
  catalog stops a filter being *offered*; it does nothing about one arriving in a shared link. That
  gap was real and this closes it.
- Its per-metric guard, ported to the declared axis in `GetAnalyticsFilterDefinitionsUseCaseTest` —
  it walks the `metrics[].filters` lists and fails the build if one advertises a filter the analytics
  engine rejects. Nothing else covers that endpoint.

Two files from it are removed: `FilterSignal` (superseded by `Signal`, which adds `TRACES` and a
documented default) and its runtime-derivation tests, which describe logic that no longer exists.

## Verification

Checked against a live stack (Management API built from this branch, 311 log entries across HTTP
proxy, LLM and MCP APIs), comparing each search's `totalCount` to an Elasticsearch count for the same
predicate: **66 cases, all conforming** — 27 single filters, 11 combinations, the 3 legacy spellings,
5 refusals and 20 negative controls. The negative controls are the load-bearing half: each asserts a
value matching nothing returns 0 rows rather than the unfiltered 311, which is the only way to tell
an applied filter from an ignored one. Suites: 8,880 (rest-api-service,
ArchUnit included), 1,821 (management-v2-rest), 279 (gravitee-dashboard), 178 (console
observability).

[APIM-14817]: https://gravitee.atlassian.net/browse/APIM-14817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APIM-14828]: https://gravitee.atlassian.net/browse/APIM-14828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Screenshot 2026-08-03 at 10 15 09
Screenshot 2026-08-03 at 10 16 22
Screenshot 2026-08-03 at 10 16 12
Screenshot 2026-08-03 at 10 15 56
Screenshot 2026-08-03 at 10 15 41
Screenshot 2026-08-03 at 10 15 24

Screenshot 2026-08-03 at 10 27 31


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/domain_service/AnalyticsQueryValidator.java
```diff
@@ -164,6 +148,24 @@ private void validateFilters(List<Filter> filters) {
...
 
+    /**
+     * Whether the analytics engine can evaluate this filter, read from the catalog's signal axis rather than
+     * from a list kept in step with it by hand.
+     *
+     * <p>A name the catalog does not describe at all is accepted, as it was before the axis existed: the
+     * catalog omits several names the engines still know ({@code URI}, {@code ENTRYPOINT}, the edge ones), and
+     * rejecting those here would be a behaviour change rather than a refactor.
+     */
+    private boolean supportsAnalytics(FilterSpec.Name name) {
```
SamHardi, line 159: Fixed. The catalog is now indexed by name once, at construction:

- `AnalyticsDefinitionQueryService` gains `findFilter(FilterSpec.Name)`, mirroring the `findMetric` that was already there.
- The YAML implementation builds an `EnumMap` alongside the `apisByFilter` index it already computes, so the lookup is constant-time and the catalog is still walked exactly once per process.
- `supportsAnalytics` is now a single `definition.findFilter(name)`.

First entry wins on a duplicate name, which is what the previous `findFirst()` over the list returned — the index is a faster spelling of the same rule, not a behaviour change.

Left alone deliberately: `GetFilterValuesUseCase.findFilterSpec` runs the same scan, but once per request rather than once per filter per request. It is not the cost this comment is about, and moving it would mean rewriting eight test stubs for no measurable gain.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/analytics_engine/model/FilterSpec.java
```diff
@@ -34,28 +30,32 @@ public record FilterSpec(
...
-        String label,
-        FilterType type,
-        List<String> enumValues,
-        NumberRange range,
-        List<FilterOperator> operators,
-        List<ApiSpec.Name> apis
-    ) {
-        this(name, label, type, enumValues, range, operators, apis, null);
+    public FilterSpec {
+        signals = signals == null || signals.isEmpty() ? Signal.DEFAULT : Set.copyOf(signals);
```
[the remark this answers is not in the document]
SamHardi: Not taking this one. The suggested `isNullOrEmpty` variable restates the condition it replaces rather than naming anything the reader does not already see, and the line is inside the width the repo's prettier-java config enforces — the formatter is what governs layout here, and it leaves this line as is.

The compact constructor exists precisely to hold this normalisation, and keeping it a single expression is what makes it readable as one rule: absent or empty means the catalog default. Splitting it into two statements would suggest the two halves are separable decisions.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
```diff
@@ -331,16 +385,119 @@ private void updateFilterIds(FilterName name, FilterContext filterContext, Set<S
...
+            var instance = filter.actualInstance();
+            if (filterName(instance) != FilterName.API_TYPE) {
+                continue;
+            }
+            if (requested == null) {
+                requested = new HashSet<>();
+            }
+            for (String value : apiTypeValues(instance)) {
+                var mapped = LOGS_API_TYPES.get(value.trim().toUpperCase(Locale.ROOT));
+                if (mapped != null) {
```
arpzilla-dev, line 510: Inconsistent validation for unknown filter values. An unrecognised `API_TYPE` value here is silently dropped from the requested set (`mapped != null` guard), while an unrecognised `HTTP_STATUS_CODE_GROUP` a few lines above throws `ValidationDomainException` (see `validateStatusCodeGroups`). Both are the same class of bug this PR fixes elsewhere: a client typo produces a request that looks accepted but silently narrows/empties the result set instead of surfacing a 400. Consider rejecting an `API_TYPE` value that maps to nothing, for symmetry with the status-code-group validation and to avoid a new 'silently wrong' path in the very PR that closes the others.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/analytics_engine/definition/AnalyticsDefinitionYAMLQueryService.java
```diff
@@ -39,9 +40,22 @@ public class AnalyticsDefinitionYAMLQueryService implements AnalyticsDefinitionQ
...
         var rawSpec = YAMLDefinitionLoader.load(ANALYTICS_DEFINITION_FILE, AnalyticsDefinition.class).spec();
         spec = enrichFiltersWithApiTypes(rawSpec);
+        filtersByName = indexByName(spec.filters());
+    }
+
+    /** First entry wins on a duplicate name, matching what a scan of the list would have returned. */
+    private static Map<FilterSpec.Name, FilterSpec> indexByName(List<FilterSpec> filters) {
+        Map<FilterSpec.Name, FilterSpec> byName = new EnumMap<>(FilterSpec.Name.class);
+        for (var filter : filters) {
+            byName.putIfAbsent(filter.name(), filter);
```
arpzilla-dev: `putIfAbsent` silently keeps the first entry when the definition YAML declares two filters with the same `name`, matching the old linear-scan behaviour but also silently preserving a config authoring bug (a duplicated/copy-pasted filter block would previously at least appear twice in `getAllFilters()`; now the second one vanishes with no signal). Since this class already validates the file's structure at construction time (`enrichFiltersWithApiTypes`), consider failing fast here too — e.g. `merge(..., (a, b) -> { throw ... })` — so a duplicate name in `analytics-definition.yaml` is caught at startup rather than quietly hidden behind whichever entry happened to load first.

gravitee-apim-console-webui/src/services-ngx/environment-logs.service.ts
```diff
@@ -110,37 +115,38 @@ type SearchLogsRequestBody = {
...
+  if (values.length === 0) {
+    return null;
+  }
+
+  switch (operator) {
+    case 'IN':
+      return { name, operator, value: values };
+    case 'GTE':
+    case 'LTE': {
+      const numeric = Number(values[0]);
```
arpzilla-dev, line 141: If a user manages to submit a non-numeric value for a GTE/LTE condition (e.g. `Number('abc')` → `NaN`), `toLogFilter` returns `null` and the condition is silently dropped from the request rather than surfaced to the user — the chip stays visibly 'active' in the UI while it contributes nothing to the search. That's the exact failure mode (active-looking filter, no effect) this PR sets out to eliminate on the backend; worth at least a comment on why this can't happen from the value-input component, or reporting the drop upstream (e.g. via a returned error/warning) instead of swallowing it here.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
```diff
@@ -1226,6 +1241,9 @@ spec:
 
         - name: HTTP_STATUS
           label: Status Code
+          signals:
+              - LOGS
+              - ANALYTICS
```
arpzilla-dev, line 1246: Possible functional regression to verify: `HTTP_STATUS` only advertises `EQ`/`LTE`/`GTE` (see the `operators` list just below), no `IN`. Before this PR, the env-logs screen sent multi-selected status codes (e.g. 200 + 404) as a hardcoded `{ name: 'HTTP_STATUS', operator: 'IN', value: [...] }` (see the old `arrayFilters` list removed from `environment-logs.service.ts`), and `SearchEnvironmentLogsUseCase` still happily accepts `HTTP_STATUS` via `ArrayFilter`/`updateFilterIds` today. Now that the console strictly follows what the catalog advertises for each filter, users may lose the ability to pick several exact, non-contiguous status codes in one condition — only a single EQ value or a GTE/LTE range is reachable from the UI. If that's intentional (superseded by `HTTP_STATUS_CODE_GROUP`), fine — but worth confirming with product/QA and calling it out explicitly in the PR description/test plan, since it's the kind of silent behaviour change this PR is otherwise very careful to avoid.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
```diff
@@ -1235,12 +1253,17 @@ spec:
               - LTE
               - GTE
 
+        # EQ only: the logs engine matches a single exact path, and IN over full paths has no
+        # known consumer. Gamma's reconciled vocabulary made the same call (URI is EQ-only).
         - name: HTTP_PATH
```
araffass, line 1258: This removes IN for analytics too, not just logs. The analytics engine handles HTTP_PATH IN fine and it was advertised before, so dashboards lose the option to filter on a list of paths. It also breaks old bookmarked logs URLs that have an HTTP Path IN chip. The prune only checks filter names, so the chip stays, and now every search returns 400 because the backend throws on URI with IN. Before this PR those URLs still worked. I would keep IN in the yaml and deal with the logs limit on the logs side.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
SamHardi: Fixed in 791a5011c8. An `API_TYPE` value the catalog does not declare now raises `ValidationDomainException` — the same 400 an unknown status code group gets.

The two cases are kept apart, because they are not the same request:
- a kind the catalog declares but the logs signal cannot serve (`MESSAGE`, `A2A`, `NATIVE`, `EDGE`) still narrows to an empty scope — a valid ask that matches nothing;
- anything else is a typo, and says so.

`UNSERVED_API_TYPES` alongside `LOGS_API_TYPES` spells out the full vocabulary, and a new test pins their union against `analytics-definition.yaml` so a kind added to the catalog cannot silently become a 400.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/analytics_engine/definition/AnalyticsDefinitionYAMLQueryService.java
SamHardi: Fixed in 791a5011c8. `indexByName` now throws `IllegalStateException` on a duplicate name instead of keeping the first entry, so a copy-pasted block fails startup with the offending name rather than disappearing.

You are right that preserving the scan's behaviour was preserving a bug: the point of the index was to make the catalog the single source of truth, and silently dropping half a declaration works against that. New `AnalyticsDefinitionYAMLQueryServiceTest` covers it.

gravitee-apim-console-webui/src/services-ngx/environment-logs.service.ts
SamHardi: Fixed in 791a5011c8. `toLogFilter` now raises on a non-numeric `GTE`/`LTE` bound rather than returning `null`.

Your reading of the reachability is right and it is worth stating: the dialog's input is `type="number"`, so it cannot produce one — but a condition also arrives from a shared or bookmarked URL, where the value is whatever the link carries. That is the path where the chip stayed active while contributing nothing, which is the failure this PR exists to remove. Documented on the function, with a test.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
SamHardi: Not applicable as described: this wire has no `NOT_IN`. The request enum in `openapi-logs.yaml` is `EQ, GTE, LTE, IN, CONTAINS` — five operators — so there is no multi-value exclusion to truncate. The `default` branch is reached by `EQ` and `CONTAINS`, both of which carry a single value by schema.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
SamHardi: Thanks — this is the most useful comment on the PR, and you are right that removing `IN` was wrong. Checking how Gamma resolved the same question changes the shape of the answer, so rather than restore `IN` I would like your view on this.

Gamma's reconciled vocabulary (GMA-422) declares **one** path filter: `URI("HTTP Path", STRING, EQ_ONLY, …, LOGS + ANALYTICS)`. `HTTP_PATH` is dropped outright, with the reason in the code — the field is empty on v4. Its operator rule is `EQ, IN` for KEYWORD/ENUM and `EQ` only for STRING, so `IN` never applied to a path there.

Checked against a live index rather than taking the comment's word for it: across 1 603 065 v4 metrics documents, `exists(path)` is **0** and `exists(uri)` is all of them. Analytics resolves `HTTP_PATH` to `path` (`HTTPFieldResolver`), so that filter has been querying a field no document carries — whatever operator it advertises. Restoring `IN` on it would restore filtering on nothing.

The half of your comment about logs stands entirely: there `HTTP_PATH` is aliased to `URI`, so it hits the populated field, and that is where a bookmarked link now 400s.

So the fix I would propose is Gamma's: one `URI` filter labelled "HTTP Path", `EQ` only, serving both signals, with `HTTP_PATH` kept as a deprecated alias so saved links resolve — and analytics pointed at `uri`, which incidentally makes path filtering on dashboards work for the first time. Does that match how you would want it?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
SamHardi: Apologies — my previous reply on this thread was meant for a bot comment elsewhere and landed here by mistake. Please ignore it. The actual answer to your point:

You are right that the multi-select is lost, and confirmed on both ends: `master`'s console sent `{ name: 'HTTP_STATUS', operator: 'IN', values }` from the hardcoded `arrayFilters` list regardless of the catalog, and `SearchEnvironmentLogsUseCase` still accepts it (`ArrayFilter` → `limitByHttpStatuses`). `HTTP_STATUS` has advertised `EQ/LTE/GTE` only since before this PR, so following the catalog strictly is what makes the capability unreachable.

Adding `IN` to the catalog is not enough on its own, though. `HTTP_STATUS` is typed `NUMBER`, and the dialog renders a NUMBER filter with `gd-number-value-input` — a single `<input type="number">` whose `valuesChange` emits exactly one value. Advertising `IN` would let a user pick the operator and then only ever produce a one-element list: another advertised-but-unusable path, which is what this PR is trying to remove.

So it needs one of:
- a multi-value numeric input in `gravitee-dashboard`, then `IN` in the catalog; or
- retyping `HTTP_STATUS` so it uses the existing multi-select, which costs the `GTE`/`LTE` range and the 100–599 bounds.

Both are real work in the shared UI library rather than a catalog line. My inclination is to call the regression out explicitly in the description and do it in its own ticket, unless you read the loss as severe enough to block — happy to follow your call, and I would rather have product's view than decide it here.

gravitee-apim-console-webui/src/services-ngx/environment-logs.service.ts
```diff
@@ -110,37 +115,38 @@ type SearchLogsRequestBody = {
...
+      return Number.isFinite(numeric) ? { name, operator, value: numeric } : null;
+    }
+    default:
```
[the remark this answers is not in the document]
SamHardi: Not applicable as described: this wire has no `NOT_IN`. The request enum in `openapi-logs.yaml` is `EQ, GTE, LTE, IN, CONTAINS`, so there is no multi-value exclusion to truncate. The `default` branch is reached by `EQ` and `CONTAINS`, both single-valued by schema.

gravitee-apim-console-webui/src/management/observability/data-access/observability-filters-api.service.ts
```diff
@@ -63,26 +74,21 @@ interface FilterValueItemApi {
...
- * surface so a page only offers filters its query engine supports (APIM-14828).
- */
-export const OBSERVABILITY_FILTER_SIGNAL = new InjectionToken<ObservabilityFilterSignal>('OBSERVABILITY_FILTER_SIGNAL');
-
 @Injectable()
 export class ObservabilityFiltersApiService implements FilterDefinitionProvider, FilterValuesProvider {
   private readonly http = inject(HttpClient);
   private readonly constants = inject(Constants);
-  private readonly surfaceSignal = inject(OBSERVABILITY_FILTER_SIGNAL, { optional: true });
+  private readonly signal = inject(OBSERVABILITY_FILTER_SIGNAL, { optional: true });
```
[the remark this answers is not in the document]
SamHardi: Leaving as is. The property is private and reads at its only use site as the signal this surface asks the catalog for; the public token it comes from is already named `OBSERVABILITY_FILTER_SIGNAL`, so the qualifier is carried where it is visible.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
```diff
@@ -318,6 +360,18 @@ private List<Range> buildResponseTimeRanges(FilterContext filterContext) {
...
+    private List<SearchLogsFilters.StatusRange> buildStatusRanges(FilterContext filterContext) {
+        var from = filterContext.statusFrom().orElse(null);
+        var to = filterContext.statusTo().orElse(null);
+        if (from == null && to == null) {
+            return List.of();
+        }
+        if (from != null && to != null && from > to) {
+            throw new ValidationDomainException("Invalid HTTP_STATUS range: 'from' (gte) must not be greater than 'to' (lte).");
+        }
+        return List.of(SearchLogsFilters.StatusRange.builder().gte(from).lte(to).build());
```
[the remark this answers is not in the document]
SamHardi: Leaving as is. The two build different things — `buildStatusRanges` produces `SearchLogsFilters.StatusRange` from status bounds, `buildResponseTimeRanges` produces `Range` from response-time bounds — and they carry different validation messages on purpose, so a user learns which bound is wrong. Merging them would mean a generic message plus a type parameter, for eight lines.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
```diff
@@ -331,14 +387,139 @@ private void updateFilterIds(FilterName name, FilterContext filterContext, Set<S
...
                     filterContext.limitByUri(ids.iterator().next());
                 }
             }
-            case PAYLOAD -> {}
-            case RESPONSE_TIME -> {}
+            // EQ reaches this method as a string — the request schema discriminates on the operator, so a
+            // numeric filter only ever carries GTE/LTE. An exact response time is a single-point range.
+            case RESPONSE_TIME -> {
+                if (!ids.isEmpty()) {
+                    var value = parseResponseTime(ids.iterator().next());
```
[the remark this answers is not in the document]
SamHardi: Good catch — fixed in f8e08251c5. `RESPONSE_TIME` joins `URI` in a `SINGLE_VALUED_FILTERS` set that `applyArrayFilter` refuses outright, so a list now returns 400 rather than keeping an arbitrary element of a `HashSet`.

Worth noting it was not only partial but nondeterministic, as you say: iteration order of a small string `HashSet` is stable for a given set but changes with its contents, so the same three bounds could resolve differently between requests. The catalog advertises `EQ/GTE/LTE` for it so the console cannot produce this, but the API is reachable without the console — which is exactly the gap this PR is about. Test added.

gravitee-apim-webui-libs/gravitee-dashboard/src/lib/components/filter/add-filter-dialog/add-filter-dialog.component.ts
```diff
@@ -68,6 +70,13 @@ export interface AddFilterDialogData {
   timeTo?: number;
 }
 
+/** What the template needs to render the signal marker of a signal-exclusive filter. */
```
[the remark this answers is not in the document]
SamHardi: Leaving as is. The comment states why the type is deliberately wider than the shared `RequestFilter` — that the logs catalog advertises names and an operator that contract does not carry, and that narrowing to it would force a cast at every call site and misdescribe what the endpoint accepts. That is the rationale for the shape, not a restatement of it.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
SamHardi: Decision: no `IN` on `HTTP_STATUS` in this PR, and the regression is called out rather than patched.

The blocker is the UI, not the catalog. `HTTP_STATUS` is typed `NUMBER`, and a NUMBER filter renders as `gd-number-value-input` — one `<input type="number">` emitting a single value. Advertising `IN` would let a user select the operator and still only ever produce a one-element list: an advertised-but-unusable path, which is the class of defect this PR removes. Making it real means either a multi-value numeric input in `gravitee-dashboard`, or retyping the filter and losing `GTE`/`LTE` and the 100–599 bounds.

That is shared-UI work, and this PR is an escalation fix on a filter catalog — so it goes to its own ticket rather than growing this one. Thanks for catching it; it would have shipped unnoticed otherwise.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
SamHardi: Closing this one out with where it lands, and thank you — it changed the answer.

`IN` stays off the path filter, and that is the end state rather than a deferral: Gamma's reconciled vocabulary declares `EQ` only for STRING filters, so `IN` over a path applies in neither product.

The analytics half does not regress the way we both assumed. Analytics resolves `HTTP_PATH` to the ES field `path`, and on a live index `exists(path)` is 0 across 1 603 065 v4 documents while `uri` is on all of them. That filter has been querying a field no document carries, whatever operator it advertised — Gamma dropped `HTTP_PATH` for exactly this reason.

Your logs half was right, and it is fixed: `HTTP_PATH` is aliased to `URI` there, so saved links keep resolving.

What is left is unifying on a single `URI` filter labelled "HTTP Path" for both signals and repointing the analytics resolver at `uri` — which would make path filtering on dashboards work for the first time. That touches the catalog vocabulary and `HTTPFieldResolver` in `gravitee-apim-repository-elasticsearch`, so it gets its own ticket rather than riding along in an escalation fix.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
```diff
@@ -275,11 +289,18 @@ private void clearFilterIds(StringFilter filter, FilterContext filterContext) {
...
     private void applyArrayFilter(ArrayFilter filter, FilterContext filterContext) {
-        // URI only supports EQ operator from StringFilter, ignore all ArrayFilters for
-        // URI
-        if (filter.name() == FilterName.URI) {
-            return;
+        // Say so instead of returning early: silently ignoring it is what made the catalog and the search
+        // disagree (APIM-14817).
+        if (SINGLE_VALUED_FILTERS.contains(filter.name())) {
+            throw new ValidationDomainException("Filter " + filter.name() + " does not support operator " + filter.operator() + ".");
         }
```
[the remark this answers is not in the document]
SamHardi: Intentional, and already documented. The PR description carries an API surface table listing it as a behaviour change: `HTTP_PATH` with `IN` returns 400 where it previously returned 200 with the filter silently ignored — the unfiltered result set, not a filtered one. Preserving the old behaviour is the bug the PR closes (APIM-14817). `HTTP_PATH` remains accepted with `EQ`, so saved links keep working.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/domain_service/FilterContext.java
```diff
@@ -29,6 +29,9 @@ public final class FilterContext {
     private Set<String> planIds;
     private Set<HttpMethod> methods;
     private Set<Integer> statuses;
+    private Set<String> statusCodeGroups;
```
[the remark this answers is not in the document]
SamHardi: Leaving as is. The pattern is the one every other `limitBy...` member on `FilterContext` follows; consolidating these three alone would make them the odd ones out. The class is a candidate for removal in the shared-translator refactor rather than for a local abstraction.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
```diff
@@ -190,10 +195,16 @@ private UserContext loadNames(ExecutionContext executionContext, UserContext use
 
     private SearchLogsFilters buildFilters(UserContext userContext, SearchLogsRequest request) {
         var filterContext = new FilterContext();
+        // API_TYPE narrows which APIs are in scope rather than adding a clause, so it is read before the
```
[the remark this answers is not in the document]
SamHardi: Partly addressed already: `API_TYPE` handling lives in `requestedApiTypes`, and the builder population is the flat run of `builder.x(...)` calls it has always been. Splitting that run into helpers would hide which slots exist, which is precisely what made a filter droppable unnoticed.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/logs_engine/use_case/SearchEnvironmentLogsUseCase.java
```diff
@@ -209,6 +220,7 @@ private SearchLogsFilters buildFilters(UserContext userContext, SearchLogsReques
             .orElseGet(Collections::emptyList)
             .stream()
             .filter(api -> isWantedHttpApi(api.getType()))
+            .filter(api -> requestedApiTypes.map(types -> types.contains(api.getType())).orElse(true))
```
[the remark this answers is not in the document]
SamHardi: Leaving as is. `requestedApiTypes` is already the named extraction; the `.map(...).orElse(true)` in the predicate is what expresses "no API_TYPE filter means every kind", and its javadoc states why empty and absent are different values.

jourdiw (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
araffass: Fair enough, i'm okay with the analytics side. If HTTP_PATH resolves to path and that field is empty across every v4 doc, then keeping IN would just filter on nothing, so leaving it off is fine.                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  The logs half is the one I cared about anyway, and the alias covers it. Only thing I'd note is that i think old links with an HTTP Path IN chip will still 400, since IN is gone from the filter now, not just aliased. Fine by me, just calling it out.                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  +1 on the separate ticket for the URI unification.

PR #18915, opened 2026-08-03
feat(gamma): Group list, expand Create Group form fields and add role/ownership badges

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/FOUND-101

## Description

This adds the **User Groups** page under Identity & Access in the new Gamma Console (Platform Management).

**What you can now do:**

- **See all groups** in a searchable, paginated table, with badges showing at a glance if a group is: the Primary owner of resources, and set to auto-associate with **new APIs, API Products, or Applications.**
- **Create a group** with a name, default roles for APIs/Applications/API Products, whether members can change their own role in each of those, whether the group should be auto-added to new APIs/API Products/Applications, a max member limit, how people can be invited (search existing users vs. email), and whether members get notified when added.
- **Edit** an existing group's settings, same fields as above.
- **Delete** a group, with a confirmation step.
- All actions (create/edit/delete) are gated by permissions, matching classic Console — buttons only show up for users who are actually allowed to perform them.

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupSheet.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupSheet.tsx
@@ -1,0 +1,445 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+    Button,
+    Field,
+    FieldLabel,
+    Input,
+    ScrollArea,
+    Select,
+    SelectContent,
+    SelectItem,
+    SelectTrigger,
+    SelectValue,
+    Sheet,
+    SheetContent,
+    SheetDescription,
+    SheetFooter,
+    SheetHeader,
+    SheetTitle,
+    Switch,
+} from '@gravitee/graphene-core';
+import { useCallback, useEffect, useMemo, useState, type FormEvent, type ReactNode } from 'react';
+
+import { STANDARD_SHEET_WIDTH } from '../../applications/components/sheetLayout';
+import type { Group, GroupRole } from '../types/group';
+
+export type GroupSheetMode = 'create' | 'edit';
+
+export interface GroupFormValues {
+    name: string;
+    apiRole: string;
+    applicationRole: string;
+    apiProductRole: string;
+    lockApiRole: boolean;
+    lockApiProductRole: boolean;
+    lockApplicationRole: boolean;
+    defaultGroupForNewApis: boolean;
+    defaultGroupForNewApiProducts: boolean;
+    defaultGroupForNewApplications: boolean;
+    maxInvitation: string;
+    systemInvitation: boolean;
+    emailInvitation: boolean;
+    notifyOnMemberAdded: boolean;
+}
+
+const NO_ROLE_VALUE = '__none__';
+
+function defaultRoleValue(roles: GroupRole[]): string {
+    return roles.find(role => role.default)?.name ?? '';
+}
+
+function hasEventRule(group: Group, event: string): boolean {
+    return (group.event_rules ?? []).some(rule => rule.event === event);
+}
+
+// Classic Console mirror (group.component.ts:initializeForm): a brand-new group has no
+// server-side state yet, so admins can change every role (canAdminChangeAPIRole defaults true,
+// i.e. lock_api_role saves as false), invitations start off, and membership-change
+// notifications are on (`!disable_membership_notifications` defaults true).
+function buildEmptyForm(apiRoles: GroupRole[], applicationRoles: GroupRole[], apiProductRoles: GroupRole[]): GroupFormValues {
+    return {
+        name: '',
+        apiRole: defaultRoleValue(apiRoles),
+        applicationRole: defaultRoleValue(applicationRoles),
+        apiProductRole: defaultRoleValue(apiProductRoles),
+        lockApiRole: false,
+        lockApiProductRole: false,
+        lockApplicationRole: false,
+        defaultGroupForNewApis: false,
+        defaultGroupForNewApiProducts: false,
+        defaultGroupForNewApplications: false,
+        maxInvitation: '',
+        systemInvitation: false,
+        emailInvitation: false,
+        notifyOnMemberAdded: true,
+    };
+}
+
+function buildFormFromGroup(group: Group): GroupFormValues {
+    return {
+        name: group.name,
+        apiRole: group.roles?.API ?? '',
+        applicationRole: group.roles?.APPLICATION ?? '',
+        apiProductRole: group.roles?.API_PRODUCT ?? '',
+        lockApiRole: Boolean(group.lock_api_role),
+        lockApiProductRole: Boolean(group.lock_api_product_role),
+        lockApplicationRole: Boolean(group.lock_application_role),
+        defaultGroupForNewApis: hasEventRule(group, 'API_CREATE'),
+        defaultGroupForNewApiProducts: hasEventRule(group, 'API_PRODUCT_CREATE'),
+        defaultGroupForNewApplications: hasEventRule(group, 'APPLICATION_CREATE'),
+        maxInvitation: String(group.max_invitation ?? ''),
+        systemInvitation: Boolean(group.system_invitation),
+        emailInvitation: Boolean(group.email_invitation),
+        notifyOnMemberAdded: !group.disable_membership_notifications,
+    };
+}
+
+function ToggleRow({
+    id,
+    title,
+    description,
+    checked,
+    onCheckedChange,
+    disabled,
+}: Readonly<{
+    id: string;
+    title: string;
+    description: string;
+    checked: boolean;
+    onCheckedChange: (checked: boolean) => void;
+    disabled?: boolean;
+}>) {
+    return (
+        <div className="flex items-center justify-between gap-4 rounded-lg border px-3 py-2">
+            <div className="space-y-0.5">
+                <FieldLabel htmlFor={id}>{title}</FieldLabel>
+                <p className="text-xs text-muted-foreground">{description}</p>
+            </div>
+            <Switch id={id} checked={checked} onCheckedChange={onCheckedChange} disabled={disabled} />
+        </div>
+    );
+}
+
+function RoleSelect({
+    id,
+    label,
+    value,
+    roles,
+    disabled,
+    onChange,
+}: Readonly<{
+    id: string;
+    label: ReactNode;
+    value: string;
+    roles: GroupRole[];
+    disabled: boolean;
+    onChange: (value: string) => void;
+}>) {
+    // Classic Console hides system roles (e.g. PRIMARY_OWNER) from the default-role pickers — they're
+    // assigned automatically, not chosen. Keep the currently-selected value visible even if it's a
+    // system role (e.g. an existing group saved with one before this filter existed).
+    const selectableRoles = roles.filter(role => !role.system || role.name === value);
+
+    return (
+        <Field orientation="vertical" className="gap-1.5">
+            <FieldLabel htmlFor={id}>{label}</FieldLabel>
+            <Select value={value || NO_ROLE_VALUE} onValueChange={val => onChange(val === NO_ROLE_VALUE ? '' : val)} disabled={disabled}>
+                <SelectTrigger id={id}>
+                    <SelectValue />
+                </SelectTrigger>
+                <SelectContent>
+                    <SelectItem value={NO_ROLE_VALUE}>None</SelectItem>
+                    {selectableRoles.map(role => (
+                        <SelectItem key={role.name} value={role.name}>
+                            {role.name}
+                        </SelectItem>
+                    ))}
+                </SelectContent>
+            </Select>
+        </Field>
+    );
+}
+
+export function GroupSheet({
+    open,
+    mode,
+    group,
+    apiRoles,
+    applicationRoles,
+    apiProductRoles,
+    rolesLoading,
+    onClose,
+    onSubmit,
+    isSaving,
+}: Readonly<{
+    open: boolean;
+    mode: GroupSheetMode;
+    group?: Group;
+    apiRoles: GroupRole[];
+    applicationRoles: GroupRole[];
+    apiProductRoles: GroupRole[];
+    rolesLoading: boolean;
+    onClose: () => void;
+    onSubmit: (values: GroupFormValues) => void;
+    isSaving: boolean;
+}>) {
+    const [form, setForm] = useState<GroupFormValues>(() => buildEmptyForm([], [], []));
+    const [initialForm, setInitialForm] = useState<GroupFormValues | null>(null);
+    const [maxInvitationError, setMaxInvitationError] = useState<string | null>(null);
+    const [rolesTouched, setRolesTouched] = useState(false);
+
+    useEffect(() => {
+        if (!open) return;
+        if (mode === 'edit' && group) {
+            const initial = buildFormFromGroup(group);
+            setForm(initial);
+            setInitialForm(initial);
+        } else {
+            const initial = buildEmptyForm(apiRoles, applicationRoles, apiProductRoles);
+            setForm(initial);
+            setInitialForm(null);
+        }
+        setMaxInvitationError(null);
+        setRolesTouched(false);
+        // Intentionally re-runs only on open/mode/group, not on role-list identity: resetting the whole
+        // form whenever a role query refetches would wipe out whatever the user already typed.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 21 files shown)


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/queryKeys.ts
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export const groupKeys = {
+    all: ['environment-groups'] as const,
+    list: (envId: string, query: string, page: number, size: number) => [...groupKeys.all, 'list', envId, query, page, size] as const,
+    apiRoles: () => [...groupKeys.all, 'roles', 'API'] as const,
+    applicationRoles: () => [...groupKeys.all, 'roles', 'APPLICATION'] as const,
```
ankita-gupta21: Fixed — extracted a shared roles(scope: 'API' | 'APPLICATION' | 'API_PRODUCT') builder that apiRoles/applicationRoles/apiProductRoles now delegate to.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupRoles.ts
```diff
@@ -0,0 +1,36 @@
...
+        staleTime: 5 * 60_000,
+    });
+}
+
+export function useGroupApplicationRoles() {
+    return useQuery({
+        queryKey: groupKeys.applicationRoles(),
+        queryFn: listGroupApplicationRoles,
+        staleTime: 5 * 60_000,
+    });
```
ankita-gupta21: Fixed — added a useGroupRolesQuery helper and reduced the three hooks to thin wrappers around it.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.ts
```diff
@@ -0,0 +1,64 @@
...
+export async function listGroupApiRoles(): Promise<GroupRole[]> {
+    return apimFetchJsonOrg<GroupRole[]>('/configuration/rolescopes/API/roles');
+}
+
+export async function listGroupApplicationRoles(): Promise<GroupRole[]> {
+    return apimFetchJsonOrg<GroupRole[]>('/configuration/rolescopes/APPLICATION/roles');
+}
+
+export async function listGroupApiProductRoles(): Promise<GroupRole[]> {
+    return apimFetchJsonOrg<GroupRole[]>('/configuration/rolescopes/API_PRODUCT/roles');
```
ankita-gupta21: Fixed — extracted listGroupRolesByScope(scope); the three exported functions now just pass their scope through.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupSheet.tsx
```diff
@@ -0,0 +1,261 @@
...
+    return {
+        name: group.name,
+        lockApiRole: Boolean(group.lock_api_role),
+        defaultGroupForNewApis: (group.event_rules ?? []).some(rule => rule.event === 'API_CREATE'),
+        apiRole: group.roles?.API ?? '',
+        applicationRole: group.roles?.APPLICATION ?? '',
+    };
+}
+
+export function GroupSheet({
```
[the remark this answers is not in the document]
ankita-gupta21: Won't fix as suggested — GroupSheet already extracts ToggleRow and RoleSelect to avoid repeating the 9 toggle rows and 3 role selects inline. At 418 lines it's the largest sheet in this feature area, but that's proportional to field count (11 distinct settings vs. ~5 in EditDictionarySheet, which is 341 lines) rather than poor structure. Splitting further would fragment a single cohesive form without reducing real complexity.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupsPage.tsx
```diff
@@ -0,0 +1,205 @@
...
+    ENVIRONMENT_GROUP_CREATE_PERMISSION,
+    ENVIRONMENT_GROUP_DELETE_PERMISSION,
+    ENVIRONMENT_GROUP_UPDATE_PERMISSION,
+} from '../features/groups/utils/groupPermissions';
+import { DEFAULT_GROUP_LIST_PAGE_SIZE, GROUP_SEARCH_DEBOUNCE_MS } from '../features/groups/utils/paginationConstants';
+import { notify } from '../shared/notify';
+
+type SheetState = { type: 'closed' } | { type: 'create' } | { type: 'edit'; group: Group } | { type: 'delete'; group: Group };
+
+export function GroupsPage() {
```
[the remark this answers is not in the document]
ankita-gupta21: Won't fix as suggested — this follows the same fetch/permission/CRUD-handlers/render shape as every other page in this module (DictionariesPage, MetadataPage, UsersPage)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMutations.ts
```diff
@@ -0,0 +1,50 @@
...
+
+function useGroupMutation<TData, TResult>(mutationFn: (envId: string, data: TData) => Promise<TResult>) {
+    const env = useEnvironment();
+    const queryClient = useQueryClient();
+
+    return useMutation({
+        mutationFn: (data: TData) => mutationFn(env!.id, data),
+        onSuccess: () => {
+            if (env?.id) {
+                void queryClient.invalidateQueries({ queryKey: groupKeys.all });
```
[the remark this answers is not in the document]
ankita-gupta21: Won't fix — invalidating groupKeys.all on every mutation is the established convention across this module (useUserMutations, useCreateApplication, useDictionaryMutations, etc. all do the same).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupSheet.tsx
```diff
@@ -0,0 +1,418 @@
...
+        if (mode === 'edit' && group) {
+            const initial = buildFormFromGroup(group);
+            setForm(initial);
+            setInitialForm(initial);
+        } else {
+            const initial = buildEmptyForm(apiRoles, applicationRoles, apiProductRoles);
+            setForm(initial);
+            setInitialForm(null);
+        }
+        setMaxInvitationError(null);
```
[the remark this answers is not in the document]
ankita-gupta21: Won't fix — inline useEffect callbacks are the standard pattern throughout this codebase

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMutations.ts
```diff
@@ -0,0 +1,50 @@
...
+            }
+        },
```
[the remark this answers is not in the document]
ankita-gupta21: Same as the invalidation comment above — inline onSuccess is the convention used by every other mutation hook in this module

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupsPage.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+    function closeSheet() {
+        setSheet({ type: 'closed' });
+    }
+
+    function openCreateSheet() {
+        setSheet({ type: 'create' });
+    }
+
+    async function handleCreate(values: GroupFormValues) {
```
aqibmohammadkhan, line 82: `NewGroupEntity` doesn't accept roles, create is a POST followed by a PUT. If the PUT fails, the group was created but the user sees "Failed to create group" - and a retry will create a duplicate. 

Suggestion - catching the role-update failure separately ("Group created, but default roles could not be applied") and closing/refreshing the sheet so the user edits the existing group instead of re-creating.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupSheet.tsx
aqibmohammadkhan, line 1: The init effect runs on [open, mode, group] with role arrays deliberately excluded (eslint-disable react-hooks/exhaustive-deps). If a user clicks Create Group before the three role queries resolve (first visit, cold cache), buildEmptyForm runs with whatever has loaded — likely empty arrays — and the default role selects stay on "None" even after roles arrive. The selects are disabled during rolesLoading but the values are never back-filled. A follow-up effect that seeds still-empty role fields when rolesLoading flips false (in create mode, before the user touches them) would fix it.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/groupPayload.ts
```diff
@@ -0,0 +1,60 @@
...
+export function buildEventRules(events: { apiCreate: boolean; applicationCreate: boolean; apiProductCreate: boolean }): GroupEventRule[] {
+    const rules: GroupEventRule[] = [];
+    if (events.apiCreate) rules.push({ event: 'API_CREATE' });
+    if (events.applicationCreate) rules.push({ event: 'APPLICATION_CREATE' });
+    if (events.apiProductCreate) rules.push({ event: 'API_PRODUCT_CREATE' });
+    return rules;
+}
+
+/** Keeps any other role scope untouched — only sets/clears API, APPLICATION, and API_PRODUCT. */
+export function buildRolesMap(
```
aqibmohammadkhan, line 34: buildRolesMap returns undefined when the map ends up empty, so roles is omitted from the PUT body entirely. If the backend treats an absent roles field as "no change" (common for UpdateGroupEntity-style DTOs), a user who sets all three roles to "None" will see no error but the old roles persist. Sending an empty object {} is likely the correct "clear" signal — worth verifying against the v1 API behavior.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.ts
aqibmohammadkhan, line 1: Edit/Delete visibility is driven only by global environment-group-u/d permissions. Classic Console also respects the group-level manageable flag (e.g., groups a user can't administer). If that's intentional for this iteration, fine — otherwise rows should hide/disable actions when manageable === false.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMutations.ts
aqibmohammadkhan, line 1: useGroupsPaged guards with enabled: Boolean(env), but the mutation function dereferences env!.id unconditionally (while onSuccess defensively checks env?.id — inconsistent). If a mutation fires during an environment transition it throws a raw TypeError instead of a clean failure. Throwing an explicit error (or reading the id at call time) would be safer.

ankita-gupta21: Fixed — the mutation function now checks env?.id and rejects with new Error('No active environment') instead of dereferencing env!.id, so a mutation fired during an environment transition surfaces a clean, catchable error instead of a raw TypeError

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.ts
ankita-gupta21: Fixed — GroupsTable's actions column now checks row.original.manageable per row (defaulting to true when the field is absent) and hides Edit/Delete for groups the current user can't administer, on top of the existing global environment-group-u/d permission check. Whole action-menu button is hidden if neither applies to a row.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/groupPayload.ts
ankita-gupta21: Fixed — buildRolesMap now always returns a Record<string, string> (never undefined), so clearing every role to "None" sends roles: {} on the PUT instead of omitting the field entirely.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupSheet.tsx
ankita-gupta21: Fixed — added a second effect that runs when rolesLoading flips to false while the sheet is open in create mode, and backfills apiRole/applicationRole/apiProductRole only if the user hasn't touched them yet (tracked via a new rolesTouched flag set on any manual selection, including explicitly choosing "None"). It's a targeted merge into the existing form state, not a full re-init, so it doesn't clobber anything else the user typed while roles were still loading.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupsPage.tsx
ankita-gupta21: Fixed — handleCreate now closes the sheet and reports "Group "X" was created, but default roles could not be applied. Edit the group to set them." as a warning toast if the follow-up role update fails, instead of the misleading "Failed to create group." The group already exists at that point, so we no longer leave the user in a state where retrying would create a duplicate.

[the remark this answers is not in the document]
ankita-gupta21: Ignore

aqibmohammadkhan (commented):
Issues

Important (should fix)

- Primary Owner badge misses API Product PO-only groups
GroupsTable badges on primary_owner only. Classic uses apiPrimaryOwner || apiProductPrimaryOwner. Backend sets primary_owner for API PO, but API Product PO-only groups get apiProductPrimaryOwner without primary_owner=true.
Fix: Type those string fields; show the badge when primary_owner || apiPrimaryOwner || apiProductPrimaryOwner. Fix the JSDoc (it isn’t “current user is PO”).

- Delete still offered for primary-owner groups
Classic disables delete when apiPrimaryOwner || apiProductPrimaryOwner (backend also rejects via StillPrimaryOwnerException). Gamma only checks canDelete && manageable.
Fix: Hide/disable Delete for PO groups using the same fields as the badge.

 - Create default for “Lock API role” is inverted vs classic
Gamma defaults lockApiRole: true. Classic initializeForm for a new group uses canAdminChangeAPIRole: !group.lock_api_role → true, so lock_api_role saves as false. Labels also diverge (“Members can’t change…” vs classic “Allow Group Admin…”).
Fix: Default lockApiRole: false; align copy; update GroupSheet.spec.tsx.

 - System roles are selectable
Classic disables role.system options (e.g. PRIMARY_OWNER). Gamma allows selecting them as defaults.
Fix: Disable or filter system roles (keep current value visible if already set).

Minor

 - notify mock in GroupsPage.spec.tsx lacks warning; no coverage for the partial-create path.
 - No edit/update success/error tests; table stub doesn’t exercise manageable / PO delete rules.
 - parseMaxInvitation accepts any finite number; UI blocks < 1, but the util could reject non-positive for defense in depth.

ankita-gupta21:
> Issues
> 
> Important (should fix)
> 
> * Primary Owner badge misses API Product PO-only groups
>   GroupsTable badges on primary_owner only. Classic uses apiPrimaryOwner || apiProductPrimaryOwner. Backend sets primary_owner for API PO, but API Product PO-only groups get apiProductPrimaryOwner without primary_owner=true.
>   Fix: Type those string fields; show the badge when primary_owner || apiPrimaryOwner || apiProductPrimaryOwner. Fix the JSDoc (it isn’t “current user is PO”).
> * Delete still offered for primary-owner groups
>   Classic disables delete when apiPrimaryOwner || apiProductPrimaryOwner (backend also rejects via StillPrimaryOwnerException). Gamma only checks canDelete && manageable.
>   Fix: Hide/disable Delete for PO groups using the same fields as the badge.
> * Create default for “Lock API role” is inverted vs classic
>   Gamma defaults lockApiRole: true. Classic initializeForm for a new group uses canAdminChangeAPIRole: !group.lock_api_role → true, so lock_api_role saves as false. Labels also diverge (“Members can’t change…” vs classic “Allow Group Admin…”).
>   Fix: Default lockApiRole: false; align copy; update GroupSheet.spec.tsx.
> * System roles are selectable
>   Classic disables role.system options (e.g. PRIMARY_OWNER). Gamma allows selecting them as defaults.
>   Fix: Disable or filter system roles (keep current value visible if already set).
> 
> Minor
> 
> * notify mock in GroupsPage.spec.tsx lacks warning; no coverage for the partial-create path.
> * No edit/update success/error tests; table stub doesn’t exercise manageable / PO delete rules.
> * parseMaxInvitation accepts any finite number; UI blocks < 1, but the util could reject non-positive for defense in depth.

**Fixed:**
- Primary Owner badge/Delete: now checks primary_owner || apiPrimaryOwner || apiProductPrimaryOwner, so API-Product-only owners are caught, and Delete is blocked for them too.
- Lock API role default: fixed to false for new groups, matching classic. Kept the existing "Lock API role" label text as-is (same wording used by the other two lock toggles) rather than switching just this one to classic's phrasing — let me know if you'd rather I change it.
- System roles: filtered out of the role pickers now; an already-selected one stays visible.
- Tests: added warning toast + partial-create-failure test, edit success/error tests, and a new GroupsTable.spec.tsx covering manageable and primary-owner delete gating.
- parseMaxInvitation: now rejects non-positive numbers.

aqibmohammadkhan (approved):

PR #18922, opened 2026-08-03
feat(analytics): native kafka connections time-series

ThibaudAV:
## What

Adds a **native-Kafka time-series** capability to the analytics engine so the `NATIVE_CONNECTIONS_SUMMARY` metric can be aggregated over time (`date_histogram`) stacked by `NATIVE_CONNECTION_STATUS`. Two commits:

1. **feat**: implement `searchNativeApiTimeSeries` end to end
   - `AnalyticsRepository.searchNativeApiTimeSeries` (ES impl + noop impl)
   - new `NativeTimeSeriesQueryAdapter` (mirrors `HTTPTimeSeriesQueryAdapter`, reuses `NativeFacetsQueryAdapter` + shared `DateHistogramAdapter`/`TimeSeriesResponseAdapter`)
   - `NativeApiAnalyticsQueryService.searchTimeSeries` (was throwing `UnsupportedOperationException`)
   - fix: omit the terms `size` when `limit <= 0` (ES rejects `size:0`; `TimeSeriesQuery` defaults limit to 0)
2. **fix**: accept `"NATIVE"` as an `API_TYPE` filter value in `ApiTypeFilterTransformer` (Gamma/obs-lib spell native "NATIVE", the engine only mapped "KAFKA") — otherwise native analytics queries threw `unknown API type NATIVE` before reaching the query service.

## Why

Backs a "Logs over time" chart for native Kafka logs in Gamma (gravitee-gamma-module-esm). No validator/YAML change needed — `NATIVE_CONNECTIONS_SUMMARY` already declares COUNT + the `NATIVE_CONNECTION_STATUS` facet.

## Tests

- Unit: `NativeTimeSeriesQueryAdapterTest` (query shape), `NativeFacetsQueryAdapterTest` (incl. `size==0`), `NativeApiAnalyticsQueryServiceTest` (forwarding), `ApiTypeFilterTransformerTest` (NATIVE + KAFKA).
- Integration (Testcontainers ES): `AnalyticsElasticsearchRepositoryTest$Engine$NativeTimeSeries` — real date_histogram stacked by status, asserts a CONNECTED bucket with a positive count.

_Ticket: TBD._

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/analytics/engine/api/query/TimeSeriesQuery.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/analytics/engine/api/query/TimeSeriesQuery.java
@@ -30,6 +30,16 @@ public record TimeSeriesQuery(
     Integer limit,
     List<NumberRange> ranges
 ) implements Query {
+    public TimeSeriesQuery {
+        // The convenience constructors below default the limit to 0 to mean "no explicit cap", but a
+        // terms `size` of 0 is rejected by Elasticsearch. Normalise it to null once here (FacetsQuery
+        // already defaults to null) so the terms adapters simply omit `size` — closing the size:0 hazard
+        // for both the HTTP and native time-series paths at the source rather than in each adapter.
+        if (limit != null && limit <= 0) {
+            limit = null;
+        }
+    }
+
     public TimeSeriesQuery(TimeRange timeRange, List<Filter> filters, Long interval, List<MetricMeasuresQuery> metrics) {
         this(timeRange, filters, interval, metrics, List.of(), 0, List.of());
     }
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/log/v4/api/AnalyticsRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/log/v4/api/AnalyticsRepository.java
@@ -74,6 +74,8 @@ public interface AnalyticsRepository {
 
     TimeSeriesResult searchHTTPTimeSeries(QueryContext queryContext, TimeSeriesQuery query);
 
+    TimeSeriesResult searchNativeApiTimeSeries(QueryContext queryContext, TimeSeriesQuery query);
+
     MeasuresResult searchMessageMeasures(QueryContext queryContext, MeasuresQuery query);
 
     FilterValuesResult searchFilterValues(QueryContext queryContext, FilterValuesQuery query);
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/AnalyticsElasticsearchRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/AnalyticsElasticsearchRepository.java
@@ -54,6 +54,7 @@ public class AnalyticsElasticsearchRepository extends AbstractElasticsearchRepos
     private final HTTPFacetsQueryAdapter httpFacetsQueryAdapter = new HTTPFacetsQueryAdapter();
     private final NativeFacetsQueryAdapter nativeFacetsQueryAdapter = new NativeFacetsQueryAdapter();
     private final HTTPTimeSeriesQueryAdapter httpTimeSeriesQueryAdapter = new HTTPTimeSeriesQueryAdapter();
+    private final NativeTimeSeriesQueryAdapter nativeTimeSeriesQueryAdapter = new NativeTimeSeriesQueryAdapter();
     private final MeasuresResponseAdapter measuresResponseAdapter = new MeasuresResponseAdapter();
     private final FacetsResponseAdapter facetsResponseAdapter = new FacetsResponseAdapter();
     private final TimeSeriesResponseAdapter timeSeriesResponseAdapter = new TimeSeriesResponseAdapter();
@@ -316,6 +317,19 @@ public class AnalyticsElasticsearchRepository extends AbstractElasticsearchRepos
         var esQuery = httpTimeSeriesQueryAdapter.adapt(query);
 
         log.debug("HTTP time series query: {}", esQuery);
+
+        return client
+            .search(index, null, esQuery)
+            .map(response -> timeSeriesResponseAdapter.adapt(response, query))
+            .blockingGet();
+    }
+
+    @Override
+    public TimeSeriesResult searchNativeApiTimeSeries(QueryContext queryContext, TimeSeriesQuery query) {
+        var index = this.indexNameGenerator.getWildcardIndexName(queryContext.placeholder(), Type.V4_METRICS, clusters);
+        var esQuery = nativeTimeSeriesQueryAdapter.adapt(query);
+
+        log.debug("Native time series query: {}", esQuery);
 
         return client
             .search(index, null, esQuery)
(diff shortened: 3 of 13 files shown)


callaertanthony (changes requested):
Structure mirrors the HTTP counterpart cleanly, both v4 `AnalyticsRepository` impls (ES + noop) are updated, and the tests are real. One thing I'd fix before merge.

## Should fix

**Native time-series silently drops the 2nd facet.**

`NativeFacetsQueryAdapter.adaptFacets` uses only `queryFacets.getFirst()` and never recurses — unlike `HTTPFacetsQueryAdapter`, which recurses over the remaining facets:

```java
var facet = queryFacets.getFirst();
var aggName = AggregationAdapter.adaptName(metric.metric(), facet);
aggs.put(aggName, toTermsLeaf(facet, limit));
aggs.getJsonObject(aggName).put("aggs", adaptMeasures(metric));
return aggs;
```

Meanwhile `AnalyticsQueryValidator.MAX_TIME_SERIES_QUERY_FACETS_SIZE = 2`, and `NATIVE_CONNECTIONS_SUMMARY` declares facets `API, APPLICATION, PLAN, NATIVE_CONNECTION_STATUS` (`analytics-definition.yaml`). So a query with `facets: [APPLICATION, NATIVE_CONNECTION_STATUS]` passes validation and comes back stacked by `APPLICATION` only — no error, wrong chart.

That's pre-existing behaviour in the facets path, but this PR is what makes it reachable on time-series (it used to throw `UnsupportedOperationException`). Cheapest fix, in `NativeTimeSeriesQueryAdapter.adaptTimeSeries`:

```java
if (query.facets().size() > 1) {
    throw new UnsupportedOperationException("Native time series supports a single facet, got: " + query.facets());
}
```

Loud beats silently-wrong. Recursion only if Gamma actually needs two dimensions.

## Non-blocking

- **The `toTermsLeaf` comment is inaccurate.** Omitting terms `size` is not "no explicit cap" — Elasticsearch defaults `terms` size to **10**. Harmless for `NATIVE_CONNECTION_STATUS` (~3 values), silently truncating for `APPLICATION`/`PLAN`. Worth rewording so the next reader doesn't rely on it.
- **The same `size: 0` hazard is still live on the HTTP path.** `HTTPFacetsQueryAdapter.toTermsLeaf` keeps `if (limit != null)`, and `HTTPTimeSeriesQueryAdapter` passes `query.limit()` straight through — a `TimeSeriesQuery` built with the convenience constructor (limit `0`) breaks there identically. Either apply the same guard, or normalise `limit <= 0 -> null` once in `TimeSeriesQuery`/`toFacetQuery` instead of in two adapters.
- **Test gap:** the integration test covers the faceted path only. The no-facet branch (`adaptMeasures` under `date_histogram`) is asserted at query-shape level but never round-tripped through `TimeSeriesResponseAdapter`. One extra `@Nested` case would close it.
- **Nit:** `NativeApiAnalyticsQueryServiceTest` inlines `io.gravitee.repository.analytics.engine.api.query.Filter.Name.API` fully qualified — an import alias would read better.
- `NativeTimeSeriesQueryAdapter` is missing the `@author GraviteeSource Team` javadoc tag its HTTP twin carries.

## Checked and good

- ES index (`Type.V4_METRICS`) and the `search -> adapt -> blockingGet` shape match `searchNativeApiFacets` / `searchHTTPTimeSeries` exactly.
- Both `implements AnalyticsRepository` v4 classes are updated — no other implementation breaks on the new interface method.
- `ApiTypeFilterTransformer`: 8 pairs, under `Map.of`'s 10-pair ceiling. And `AnalyticsQueryValidator` validates filters globally rather than per-metric, so `API_TYPE` reaches the transformer even though `NATIVE_CONNECTIONS_SUMMARY` doesn't list it as a filter — the `"NATIVE"` alias fix is genuinely reachable.

ThibaudAV:
Thanks @callaertanthony — great catch on the silent facet drop. Addressed in `e032a4a`.

## Should fix

**Native time-series silently drops the 2nd facet** → fixed. `NativeTimeSeriesQueryAdapter.adaptTimeSeries` now rejects `facets().size() > 1` with `UnsupportedOperationException` instead of returning a chart grouped by the first facet only. Added a unit test (`rejects_more_than_one_facet_instead_of_silently_dropping_it`). Went with loud-fail rather than recursion since Gamma only needs the single `NATIVE_CONNECTION_STATUS` dimension today.

## Non-blocking

- **`size: 0` hazard (native + HTTP)** → fixed once at the source. Rather than guarding in each adapter, `TimeSeriesQuery` now normalises `limit <= 0 -> null` in its compact constructor (the convenience constructors default it to `0`; `FacetsQuery` already used `null`). This closes the hazard on **both** the native and HTTP time-series paths, so I reverted `NativeFacetsQueryAdapter.toTermsLeaf` back to the HTTP-style `if (limit != null)` guard for consistency. Verified end-to-end: the faceted native query now emits `terms` with no `size` field.
- **`toTermsLeaf` comment** → reworded to state that a null limit means ES applies its default `terms` size of **10** (fine for `NATIVE_CONNECTION_STATUS`, a cap to be aware of for higher-cardinality facets).
- **Test gap (no-facet branch)** → added `should_return_native_connection_time_series_total_count_without_a_facet`, which round-trips the `date_histogram` + direct-`COUNT` (no terms) shape through `TimeSeriesResponseAdapter` and asserts the buckets carry measures (`ofMeasures`, `buckets() == null`) with a positive total, against the real ES container.
- **`@author` tag** → added to `NativeTimeSeriesQueryAdapter`.

## Two I left as-is (happy to change if you disagree)

- **Fully-qualified `Filter.Name.API` in `NativeApiAnalyticsQueryServiceTest`** — the test imports `io.gravitee.apim.core.analytics_engine.model.*` (wildcard), which brings in a *different* `Filter` (used at `new Filter(FilterSpec.Name.API, ...)`), plus `FilterSpec.Name`/`FacetSpec.Name`/`MetricSpec.Name`. A plain import of the engine `Filter` would clash, and a bare `Name.API` would sit next to `FilterSpec.Name.API` in the same file — I found the FQN actually reads *less* ambiguously here. Since Java has no import aliases, I left it qualified.
- **aikido's `json(...)` → `buildQueryJson(...)` rename** — `json(...)` is the shared private-helper convention across all four adapters (`HTTPTimeSeriesQueryAdapter`, `HTTPFacetsQueryAdapter`, `NativeFacetsQueryAdapter`, this one); renaming just this one would break the symmetry. Left it for a repo-wide follow-up if we want to rename them together.

Adapter unit tests (3+3), the service test (4), and the ES integration group all green.

callaertanthony (approved):
Re-reviewed at `e032a4a9`. All five points from the previous round are addressed — thanks.

- The 2nd facet is now rejected loudly in `NativeTimeSeriesQueryAdapter` instead of silently dropped, with a test covering it.
- The `terms` size comment is accurate now, including what the ES default of 10 means for high-cardinality facets.
- Normalising `limit <= 0` to `null` in the `TimeSeriesQuery` compact constructor is a better fix than the per-adapter guard I suggested — it closes the `size: 0` hazard on the HTTP path too.
- The no-facet branch is now round-tripped through `TimeSeriesResponseAdapter` by an ES integration test.

Two non-blocking notes, take or leave:

1. The 2-facet rejection surfaces as an HTTP 500. `AnalyticsQueryValidator` allows up to 2 time-series facets, and `UnsupportedOperationException` falls through to `ThrowableMapper` → `INTERNAL_SERVER_ERROR`. Same shape as the existing non-COUNT measure rejection, so it's a pre-existing pattern rather than something introduced here — but the validator would be the right place if native metrics ever get a facet-count rule, so callers get a 400.

2. The HTTP half of the `size: 0` fix has no assertion. `HTTPTimeSeriesAdapterTest` never checks `terms/size`, so only the native side locks the normalisation in. A single `assertThat(terms.get("size")).isNull()` there would cover it.

One CI note: `Sonar - gravitee-apim-repository` is red while `Test repository` is green and the SonarCloud gate for that project reports OK (new coverage 94.7%, every condition passing). Looks like a job-level failure rather than a code one — worth a rerun before merging.

PR #18923, opened 2026-08-03
fix: create Overview pages for APIs generated under API Products

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-97

## Description

Fixes the API Product navigation flow so that APIs generated under newly created API Products receive their default **Overview** pages.

After creating API Products, the navigation tree is refreshed to retrieve the generated API navigation items. Their default pages are then created using the existing default-page endpoint, and the tree is refreshed again to display the new pages.

## Changes

- Collect newly created API Product navigation item IDs from the bulk response.
- Identify API navigation items generated directly under those products.
- Seed default pages only for the identified APIs.
- Skip page creation when the products contain no APIs.
- Keep successfully created products when default-page creation fails and display an error notification.
- Add tests for successful creation, empty products, unrelated APIs, partial bulk failures, and default-page creation failures.

https://github.com/user-attachments/assets/b270de57-3210-47a7-b45e-bdb0fe9413db

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/BulkCreatePortalNavigationItemUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/BulkCreatePortalNavigationItemUseCase.java
@@ -16,6 +16,7 @@
 package io.gravitee.apim.core.portal_page.use_case;
 
 import io.gravitee.apim.core.UseCase;
+import io.gravitee.apim.core.portal_page.domain_service.PortalNavigationApiDefaultPageDomainService;
 import io.gravitee.apim.core.portal_page.domain_service.PortalNavigationItemCreationExpansionDomainService;
 import io.gravitee.apim.core.portal_page.domain_service.PortalNavigationItemDomainService;
 import io.gravitee.apim.core.portal_page.domain_service.PortalNavigationItemValidatorService;
@@ -32,6 +33,7 @@ public class BulkCreatePortalNavigationItemUseCase {
     private final PortalNavigationItemDomainService domainService;
     private final PortalNavigationItemValidatorService validatorService;
     private final PortalNavigationItemCreationExpansionDomainService creationExpansionDomainService;
+    private final PortalNavigationApiDefaultPageDomainService defaultPageDomainService;
 
     public Output execute(Input input) {
         final var organizationId = input.organizationId();
@@ -45,6 +47,8 @@ public class BulkCreatePortalNavigationItemUseCase {
         for (CreatePortalNavigationItem itemToCreate : expansion.itemsToCreate()) {
             result.add(domainService.create(organizationId, environmentId, itemToCreate));
         }
+
+        defaultPageDomainService.seedDefaultPages(organizationId, environmentId, expansion.generatedApiNavigationItemIds());
 
         return new BulkCreatePortalNavigationItemUseCase.Output(expansion.selectRequestedItems(result));
     }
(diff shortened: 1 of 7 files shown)


gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
```diff
@@ -396,7 +397,16 @@ export class PortalNavigationItemsComponent implements HasUnsavedChanges {
       .pipe(
         filter((result): result is ApiProductSectionEditorDialogResult => !!result),
         switchMap(result => this.createApiProductsInOrder(parentItem.id, result.apiProducts, result.visibility)),
-        switchMap(result => this.refreshNavigationItems().pipe(map(() => result))),
+        switchMap(result =>
+          this.refreshNavigationItems().pipe(
+            switchMap(navigationItems =>
```
thomasballivet: Suggestion: instead of doing one call to seed everything and another one to get updated data, why don't we expect it to be done from the backend? `BulkCreatePortalNavigationItemUseCase` already creates those API children and drops them from the response, so it has the IDs we're re-deriving here. Calling the seed there (SeedDefaultPagesForApiNavigationItemsUseCase#seedDefaultPage, extracted into a domain service) would simplify things in the frontend and take advantage of data we already have in the backend. Wdyt?

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
```diff
@@ -596,17 +606,49 @@ export class PortalNavigationItemsComponent implements HasUnsavedChanges {
...
         return of({
           createdItemId: null,
+          createdApiProductNavigationItemIds: [],
           errorMessage: this.getApiProductCreateErrorMessage(error),
         });
       }),
     );
   }
 
+  private seedDefaultPagesForGeneratedApis(
```
thomasballivet:
```suggestion
  private seedDefaultPagesForApisUnderCreatedProducts(
```

jaroslawlakomy-grav: No longer applicable—the frontend helper was removed after moving the orchestration to the backend.

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
jaroslawlakomy-grav: Implemented as suggested. Bulk creation now seeds default pages using the generated API navigation IDs through a shared domain service, while the frontend performs a single refresh.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/BulkCreatePortalNavigationItemUseCase.java
```diff
@@ -46,6 +48,8 @@ public Output execute(Input input) {
             result.add(domainService.create(organizationId, environmentId, itemToCreate));
         }
 
+        defaultPageDomainService.seedDefaultPages(organizationId, environmentId, expansion.generatedApiNavigationItemIds());
```
thomasballivet, line 51: Can we do something similar in `CreatePortalNavigationItemUseCase`? It runs the same expansion and drops the generated children too, so creating an API Product through the single-item endpoint still leaves its APIs without an Overview page

thomasballivet (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/BulkCreatePortalNavigationItemUseCase.java
jaroslawlakomy-grav: sounds good, will be delivered in the separate PR

PR #18941, opened 2026-08-04
fix(ci): look up the bumped version property in the distribution pom

phiz71:
## Issue

N/A

## Description

Follow-up fix to the distribution refactor: the **Bump Gamma module version** workflow
(`.github/workflows/bump-from-module.yml`) has been broken since the bundled-plugin version
catalog moved from the root `pom.xml` to `gravitee-apim-distribution/pom.xml`.

The step read the root pom with a hardcoded `open('pom.xml')`, so it now exits with
`ERROR: <gravitee-gamma-module-*.version> not found in pom.xml` for every module bump.

The property is now looked up in **both** poms — the distribution one first, since that is where
bundled plugin versions live, then the root pom which still holds the engine and third-party
versions. The workflow input is a generic Maven artifact ID, so searching both keeps it correct
whichever pom owns the property.

The resolved file is exported as `POM_FILE` and the commit step stages that file instead of a
hardcoded `pom.xml`. Without this second change the workflow would have silently opened an empty
PR — worse than the current visible failure.

## Additional context

Verified locally by extracting the workflow's Python step and running it against the real poms:

| Case | Result |
|---|---|
| `gravitee-gamma-module-aim` (moved to the catalog) | updates `gravitee-apim-distribution/pom.xml`, `POM_FILE` set accordingly |
| `gravitee-connector-http` (still in the root pom) | updates `pom.xml`, `POM_FILE` set accordingly |
| Unknown module | clear error naming both poms, exit code 1 |
| Version already applied | idempotent, no commit (existing `git diff --cached --quiet` guard) |

The workflow YAML still parses and keeps its three steps.

The other GitHub workflows were audited at the same time and need no change: `run-e2e-tests`,
`publish-images-for-test-env` and `trigger-gko-apim-rollout` only POST to the CircleCI API, and
`remove-ready-to-test-labels` only calls the GitHub API. Renovate is unaffected too — its only
`fileMatch` targets the CircleCI manager, while the Maven manager discovers every `pom.xml`.

### DIFF
--- a/.github/workflows/bump-from-module.yml
+++ b/.github/workflows/bump-from-module.yml
@@ -48,7 +48,7 @@ jobs:
                   token: ${{ secrets.BOT_GITHUB_TOKEN || github.token }}
                   ref: master
 
-            - name: Update version property in pom.xml
+            - name: Update version property
               env:
                   MODULE: ${{ inputs.module }}
                   VERSION: ${{ inputs.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-from-module.yml
michel-barret, line 56: Not related to your work but I think that we need to limit the tech stack of the CI node, shell and python looks too much. IMO we need to replace python by node.

michel-barret (approved):

.github/workflows/bump-from-module.yml
phiz71: Yep, and we should focus on CircleCI and renovate instead of Github Actions ... ;)

michel-barret: :+1:

PR #18951, opened 2026-08-04
fix(gateway): evaluate v4 flow conditions once the previous flow completed

phiz71:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-83

## Description

In `DEFAULT` flow mode, flow conditions were evaluated while the flows were being resolved, ahead of their
execution. As soon as a flow held an asynchronous policy, the condition of the next flow was therefore evaluated
while that flow was still in flight, and could not observe the attributes it was about to set. A flow meant to be
skipped was executed, and the other way around.

The condition is now evaluated by the `FlowChain`, right before a flow runs, so it always observes what the
previous flows did. The flows that have been executed are kept in the execution context, and the following phases
replay exactly those, without evaluating their conditions again.

`BEST_MATCH` is deliberately left untouched: a single flow runs per chain, so no previous flow of the same chain
can change the outcome, and the condition has to take part in the selection of that flow. `FlowResolverFactory` is
now the single place knowing which filters it applied while resolving, and tells through `deferredConditionFilter()`
what is left to be evaluated later — so a condition can never be evaluated twice, and the `BEST_MATCH` rule is
written once.

Two commits, meant to be reviewed and backported as such:

1. `test(gateway): cover BEST_MATCH flow selection with conditions` — characterization only, **green on `master`
   without any production change**. It pins down a behaviour that had no coverage at all: in `BEST_MATCH`, a flow
   condition decides eligibility *before* the most specific flow is picked, so a flow ruled out by its condition is
   not a candidate and a less specific flow is selected in its place.
2. `fix(gateway): evaluate v4 flow conditions once the previous flow completed` — the reproduction test (failing
   before the fix) and the fix itself.

### Reproduction

V4 proxy API, `DEFAULT` flow mode, two flows on the same path:

| | flow | request policies | flow condition |
|---|---|---|---|
| 1 | set the marker | Rate Limit, then Assign Attributes (`foo = bar`) | none |
| 2 | runs only without the marker | Transform Headers | `{#context.attributes['foo'] == null}` |

Flow 2 must be skipped, since flow 1 sets `foo` before flow 2 is considered. It was executed anyway. Removing the
Rate Limit policy makes the very same API behave correctly — the defect only shows up when a previous flow crosses
an asynchronous boundary.

The Rate Limit policy is not at fault: it captures the current Vert.x context, runs the increment off the event
loop and comes back with `observeOn(RxHelper.scheduler(context))` before touching the execution context, which is
exactly what a reactive policy is expected to do. Any policy returning a genuinely asynchronous `Completable`
(Cache, HTTP Callout, OAuth2 introspection, Latency) reproduces it identically.

Root cause: `AbstractFlowResolver` evaluates conditions in a `concatMapMaybe`, and both that operator and the
`concatMapCompletable` of `FlowChain` prefetch from upstream. `concatMapMaybe` re-requests an item as soon as it
has emitted one, without waiting for the downstream — so lowering the prefetch would not have been enough, the
evaluation had to leave the resolution pipeline.

### Note on an internal attribute

The internal `flow.<id>` attribute now holds the list of executed flows instead of a cached `Flowable`. Side
effect worth knowing: a phase interrupted halfway no longer resolves, during the response phase, flows that were
never executed during the request phase.

### Impact on the reactors overriding `DefaultApiReactorFactory#flowResolverFactory()`

| reactor | overrides it | after this change |
|---|---|---|
| a2a-proxy, authz, llm-proxy | no | fixed |
| edge, native-kafka | does not use these factories | not concerned |
| message, mcp-proxy | yes, with a single filter | **behaviour strictly unchanged, still exposed to the defect** |

`MessageApiReactorFactory` and `MCPProxyApiReactorFactory` build their `FlowResolverFactory` with a single filter,
conditions included. They keep the two-argument constructor, so `deferredConditionFilter()` returns a no-op filter:
conditions are still evaluated upfront, and never twice. Nothing is degraded for them, but message and MCP APIs
are **not** fixed by this PR.

Checked by compiling and running both reactors against this branch, on their up-to-date default branches:
`gravitee-reactor-message` 329 tests green, `gravitee-reactor-mcp-proxy` 226 tests green.

### Validation

- `gravitee-apim-gateway-handlers-api`: 997 tests green.
- Integration tests: 133 green, covering the new reproduction and characterization suites plus
  `FlowPhaseExecution` (V3/V4/emulation), `BestMatch` (V3/V4/emulation/messages) and the overlapping suite.
- `FlowChainTest` asserts, with a `CompletableSubject`, that the next condition is *not* evaluated until the
  previous asynchronous flow completes, and that the response phase evaluates nothing again.
- `FlowResolverFactoryTest` covers the deferral contract, including the single-filter case that keeps the
  overriding reactors unchanged.

### Follow-ups

- A separate PR will target `4.12.x` with the same change, labelled for 4.11.x / 4.10.x / 4.9.x. The integration
  test module moved to `gravitee-apim-distribution/` on `master`, so a Mergify backport of this PR would conflict
  on the test files; writing them at the right path directly avoids it.
- Message and MCP APIs need a follow-up PR in their own repository, switching to the three-argument constructor of
  `FlowResolverFactory`. Those can only be merged once a release carrying it is available.
- `io.gravitee.gateway.reactive.handlers.api.flow.FlowChain` (v2 and V4 emulation) has the same
  `resolve().cache()` + `concatMapCompletable` shape and is likely to carry the same defect. Not addressed here.

### Related

Same issue, earlier attempts, both still open:
https://github.com/gravitee-io/gravitee-api-management/pull/18774 and
https://github.com/gravitee-io/gravitee-api-management/pull/18780. This PR follows the same overall approach; the
integration fixtures were inspired by the former, the timing unit test by the latter.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/flow/FlowChain.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/flow/FlowChain.java
@@ -21,11 +21,13 @@ import io.gravitee.common.http.HttpStatusCode;
 import io.gravitee.definition.model.v4.flow.Flow;
 import io.gravitee.gateway.reactive.api.ExecutionFailure;
 import io.gravitee.gateway.reactive.api.ExecutionPhase;
+import io.gravitee.gateway.reactive.api.context.base.BaseExecutionContext;
 import io.gravitee.gateway.reactive.api.context.http.HttpExecutionContext;
-import io.gravitee.gateway.reactive.api.context.http.HttpPlainExecutionContext;
 import io.gravitee.gateway.reactive.api.hook.ChainHook;
 import io.gravitee.gateway.reactive.api.hook.Hookable;
+import io.gravitee.gateway.reactive.core.condition.ConditionFilter;
 import io.gravitee.gateway.reactive.core.hook.HookHelper;
+import io.gravitee.gateway.reactive.handlers.api.v4.flow.resolver.FlowResolverFactory;
 import io.gravitee.gateway.reactive.policy.HttpPolicyChain;
 import io.gravitee.gateway.reactive.v4.flow.FlowResolver;
 import io.gravitee.gateway.reactive.v4.policy.PolicyChainFactory;
@@ -49,23 +51,26 @@ import lombok.CustomLog;
 public class FlowChain implements Hookable<ChainHook> {
 
     protected static final String INTERNAL_CONTEXT_ATTRIBUTES_FLOWS_MATCHED = "flowExecution.flowsMatched";
+
     private static final String EXECUTION_FAILURE_KEY_FAILURE = "FLOW_EXECUTION_FLOW_MATCHED_FAILURE";
     private final String id;
     private final FlowResolver flowResolver;
     private final String resolvedFlowAttribute;
     private final PolicyChainFactory<HttpPolicyChain, Flow> policyChainFactory;
+    private final ConditionFilter<BaseExecutionContext, Flow> conditionFilter;
     private final boolean validateFlowMatching;
     private final boolean interruptIfNoMatch;
     private List<ChainHook> hooks;
 
     public FlowChain(final String id, final FlowResolver flowResolver, final PolicyChainFactory<HttpPolicyChain, Flow> policyChainFactory) {
-        this(id, flowResolver, policyChainFactory, false, false);
+        this(id, flowResolver, policyChainFactory, FlowResolverFactory.NO_DEFERRED_CONDITION, false, false);
     }
 
     public FlowChain(
         final String id,
         final FlowResolver flowResolver,
         final PolicyChainFactory<HttpPolicyChain, Flow> policyChainFactory,
+        final ConditionFilter<BaseExecutionContext, Flow> conditionFilter,
         final boolean validateFlowMatching,
         final boolean interruptIfNoMatch
     ) {
@@ -73,6 +78,7 @@ public class FlowChain implements Hookable<ChainHook> {
         this.flowResolver = flowResolver;
         this.resolvedFlowAttribute = "flow." + id;
         this.policyChainFactory = policyChainFactory;
+        this.conditionFilter = conditionFilter;
         this.validateFlowMatching = validateFlowMatching;
         this.interruptIfNoMatch = interruptIfNoMatch;
     }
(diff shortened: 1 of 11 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/flow/FlowChain.java
```diff
@@ -97,69 +106,68 @@ public void addHooks(final List<ChainHook> hooks) {
...
+                return Flowable.fromIterable(alreadyExecutedFlows).concatMapCompletable(flow -> executeFlow(ctx, flow, phase));
+            }
+
+            final List<Flow> executedFlows = new ArrayList<>();
+            ctx.setInternalAttribute(resolvedFlowAttribute, executedFlows);
+
+            final Flowable<Flow> resolvedFlows = flowResolver.resolve(ctx);
+
+            return resolvedFlows
+                .concatMapCompletable(flow -> executeFlowIfConditionMatches(ctx, flow, phase, executedFlows))
```
jhaeyaert: The condition will be re-executed during the response phase. We should avoid this. Conditions are meant to be executed once, after which both the request and response phases apply.

phiz71: The condition is evaluated only once per chain — the response phase replays what was executed, it does not re-evaluate.

Where that is enforced in `FlowChain`:

- `executeFlowIfConditionMatches` (L138) holds the only `conditionFilter.filter(...)` call in the class, and it is reached from L122 only.
- L112-114: as soon as the `flow.<id>` internal attribute is present, the chain calls `executeFlow` directly and never goes through the condition evaluation path, whatever the phase.
- L117: the attribute is published into the context *before* resolution starts, and filled by reference afterwards, so any later phase is guaranteed to find it.

`FlowChainTest#should_not_evaluate_the_condition_again_on_the_response_phase` asserts it: REQUEST then RESPONSE over two flows yields 2 evaluations, both during REQUEST.

One limitation, which predates this PR: the guard is "once per chain", not "during the request phase" — nothing in `execute()` tests the phase. So if the very first invocation of a chain happens to be the response phase (for instance an upstream chain interrupted the request, and the API chain never saw the request phase), resolution and therefore condition evaluation happen there. 

Measured: 2 evaluations on a response-only run. Before this PR, `resolveFlows()` was lazy in the same way and the condition filter lived in the resolver, so that case behaved identically — this is not a regression introduced here.

jhaeyaert (approved):

PR #18956, opened 2026-08-04
feat(observability): add Gamma dashboards read endpoints

SamHardi:
## Summary
Serves dashboards saved by Gamma Observability users via two GET endpoints, so a dashboard opens from any machine instead of being tied to one browser's local storage. Read-only — write endpoints land in a later increment (OBS-16/17).

Related: [OBS-15](https://gravitee.atlassian.net/browse/OBS-15) (depends on [OBS-14](https://gravitee.atlassian.net/browse/OBS-14), repository layer, already Done)

## Changes
- New `dashboard` domain under `gravitee-gamma-rest-api` (core model, port, exceptions, use cases) following the module's Hexagonal layout, mirroring the existing tracing/logs/filters domains
- `GammaDashboardRepositoryAdapter` wraps the `GammaDashboardRepository` SPI shipped by OBS-14, with anticorruption mapping between the persisted `field`/lowercase-`operator`/`value` shape and this endpoint's `name`/UPPERCASE-`operator`/`values` wire shape
- Two REST resources (list + single), declarative `@Permissions` on `ENVIRONMENT_DASHBOARD:READ`
- Environment isolation: a dashboard id from another environment returns 404 (not 403), enforced via `findByIdAndEnvironmentId`
- OpenAPI spec extended with the `Dashboards` tag, both paths, and schemas
- Registered in the three mandatory wiring points (`GammaModuleApplication`, `GammaRootResource`, `StandaloneConfiguration`)

### Endpoint summary

|---|---|---|---|---|
| `/gamma/organizations/{orgId}/environments/{envId}/observability/dashboards` | GET | New | — | `200 { "data": [ { "id", "title", "filters": [...], "timeRange", "widgets", "version", "createdAt", "updatedAt" } ] }` |
| `/gamma/organizations/{orgId}/environments/{envId}/observability/dashboards/{dashboardId}` | GET | New | — | `200 { "id", "title", ... }` / `404` (unknown id or wrong environment) / `403` (missing permission) |

## Test plan
- [x] `mvn test` on `gravitee-gamma-rest-api`: 133 tests, 0 failures, 0 errors
- [x] `mvn prettier:check` clean
- [x] OpenAPI YAML validated (parses, dashboard paths/schemas present)
- [ ] Reviewer: confirm the persisted-to-wire filter mapping (`field`→`name`, lowercase→UPPERCASE `operator`, `value`→`values`, `label` passthrough) reads correctly against a real Mongo/JDBC-backed dashboard once one exists
- [ ] Reviewer: confirm a dashboard id from another environment returns 404, not 403 or 200

## Reviewer notes
- Riskiest part of this diff is the field-mapping in `GammaDashboardRepositoryAdapter.toCore(GammaDashboard.Filter)` — pinned by `GammaDashboardRepositoryAdapterTest`, but worth a close look since the persisted model (OBS-14) and this endpoint's wire format diverge on field names and casing.
- No dashboards can exist yet in any environment (write path is OBS-16), so this PR is effectively unreachable in production until then — safe to merge ahead of it.

[OBS-15]: https://gravitee.atlassian.net/browse/OBS-15?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OBS-14]: https://gravitee.atlassian.net/browse/OBS-14?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-standalone/gravitee-apim-rest-api-standalone-container/src/main/java/io/gravitee/rest/api/standalone/spring/StandaloneConfiguration.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-standalone/gravitee-apim-rest-api-standalone-container/src/main/java/io/gravitee/rest/api/standalone/spring/StandaloneConfiguration.java
@@ -15,10 +15,7 @@
  */
 package io.gravitee.rest.api.standalone.spring;
 
-import io.gravitee.gamma.rest.infra.config.GammaAnalyticsConfiguration;
-import io.gravitee.gamma.rest.infra.config.GammaLogsConfiguration;
-import io.gravitee.gamma.rest.infra.config.GammaObservabilityFilterConfiguration;
-import io.gravitee.gamma.rest.infra.config.GammaTracingConfiguration;
+import io.gravitee.gamma.rest.infra.config.GammaObservabilityConfiguration;
 import io.gravitee.integration.controller.spring.IntegrationControllerConfiguration;
 import io.gravitee.node.api.Node;
 import io.gravitee.node.api.NodeMetadataResolver;
@@ -47,10 +44,7 @@ import org.springframework.context.annotation.Import;
         io.gravitee.rest.api.management.v2.rest.spring.RestManagementConfiguration.class,
         IntegrationControllerConfiguration.class,
         RestPortalConfiguration.class,
-        GammaTracingConfiguration.class,
-        GammaObservabilityFilterConfiguration.class,
-        GammaLogsConfiguration.class,
-        GammaAnalyticsConfiguration.class,
+        GammaObservabilityConfiguration.class,
     }
 )
 public class StandaloneConfiguration {
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/model/Dashboard.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/model/Dashboard.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gamma.rest.core.observability.dashboard.model;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.gravitee.gamma.rest.core.observability.dashboard.exception.InvalidDashboardException;
+import java.time.Instant;
+import java.util.List;
+
+/**
+ * A dashboard saved by a Gamma Observability user, served by the read endpoints in this ticket
+ * (OBS-15) and written by OBS-16/17. Distinct from the legacy v2 analytics dashboard and from
+ * {@code CustomDashboard} — see OBS-13 for the rationale.
+ *
+ * <p>{@code widgets} is deliberately opaque: never parsed, only returned verbatim (see
+ * {@code GammaDashboard.widgets}).
+ *
+ * @author GraviteeSource Team
+ */
+public record Dashboard(
+    String id,
+    String environmentId,
+    String title,
+    String description,
+    List<DashboardFilter> filters,
+    TimeRange timeRange,
+    JsonNode widgets,
+    Integer version,
+    String createdBy,
+    Instant createdAt,
+    Instant updatedAt
+) {
+    public Dashboard {
+        requireNonBlank(id, "id");
+        requireNonBlank(environmentId, "environmentId");
+        requireNonBlank(title, "title");
+        filters = filters == null ? List.of() : List.copyOf(filters);
+    }
+
+    private static void requireNonBlank(String value, String fieldName) {
+        if (value == null || value.isBlank()) {
+            throw new InvalidDashboardException("Dashboard " + fieldName + " is required");
+        }
+    }
+}
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/model/TimeRange.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/model/TimeRange.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gamma.rest.core.observability.dashboard.model;
+
+/**
+ * Time window a dashboard opens on. {@code period} is set for {@link TimeRangeType#RELATIVE},
+ * {@code from} / {@code to} (epoch millis) for {@link TimeRangeType#ABSOLUTE}.
+ *
+ * @author GraviteeSource Team
+ */
+public record TimeRange(TimeRangeType type, String period, Long from, Long to) {}
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/use_case/GetObservabilityDashboardUseCase.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/use_case/GetObservabilityDashboardUseCase.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gamma.rest.core.observability.dashboard.use_case;
+
+import io.gravitee.apim.core.UseCase;
+import io.gravitee.gamma.rest.core.observability.dashboard.exception.DashboardNotFoundException;
+import io.gravitee.gamma.rest.core.observability.dashboard.model.Dashboard;
+import io.gravitee.gamma.rest.core.observability.dashboard.port.repository.DashboardRepository;
+import lombok.AllArgsConstructor;
+
+/**
+ * Fetches a single dashboard, scoped to the caller's environment. A dashboard id from another
+ * environment throws {@link DashboardNotFoundException} — 404, not 403 — via the same
+ * environment-scoped repository lookup used by the list use case, so existence cannot be probed
+ * across environments.
+ *
+ * @author GraviteeSource Team
+ */
+@UseCase
+@AllArgsConstructor
+public class GetObservabilityDashboardUseCase {
+
+    private final DashboardRepository dashboardRepository;
+
+    public record Input(String environmentId, String dashboardId) {}
+
+    public record Output(Dashboard dashboard) {}
+
+    public Output execute(Input input) {
+        Dashboard dashboard = dashboardRepository
+            .findByIdAndEnvironmentId(input.dashboardId(), input.environmentId())
+            .orElseThrow(() -> new DashboardNotFoundException(input.dashboardId()));
+        return new Output(dashboard);
+    }
+}
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/use_case/ListObservabilityDashboardUseCase.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/use_case/ListObservabilityDashboardUseCase.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gamma.rest.core.observability.dashboard.use_case;
+
+import io.gravitee.apim.core.UseCase;
+import io.gravitee.gamma.rest.core.observability.dashboard.model.Dashboard;
+import io.gravitee.gamma.rest.core.observability.dashboard.port.repository.DashboardRepository;
+import java.util.List;
+import lombok.AllArgsConstructor;
+
+/**
+ * Lists every dashboard saved in the caller's environment, paginated.
+ *
+ * <p>{@link DashboardRepository#findByEnvironmentId} has no native pagination (the OBS-14 SPI
+ * returns the full, ordered list) — pagination is applied here by slicing the full result, matching
+ * the "fetch, then slice" precedent used by {@code TracingPortAdapter} for the same reason. Fine for
+ * a per-environment dashboard count, which is expected to stay small.
+ *
+ * @author GraviteeSource Team
+ */
+@UseCase
+@AllArgsConstructor
+public class ListObservabilityDashboardUseCase {
+
+    private static final int DEFAULT_PAGE = 1;
+    private static final int DEFAULT_PER_PAGE = 20;
+    private static final int MAX_PER_PAGE = 100;
+
+    private final DashboardRepository dashboardRepository;
+
+    public record Input(String environmentId, Integer page, Integer perPage) {}
+
+    public record Output(List<Dashboard> dashboards, long totalCount, int page, int perPage) {}
+
+    public Output execute(Input input) {
+        int page = resolvePage(input);
+        int perPage = resolvePerPage(input);
+
+        List<Dashboard> all = dashboardRepository.findByEnvironmentId(input.environmentId());
+        int fromIndex = Math.min((page - 1) * perPage, all.size());
+        int toIndex = Math.min(fromIndex + perPage, all.size());
+
+        return new Output(all.subList(fromIndex, toIndex), all.size(), page, perPage);
+    }
+
+    private static int resolvePage(Input input) {
+        return (input.page() != null && input.page() > 0) ? input.page() : DEFAULT_PAGE;
+    }
+
+    private static int resolvePerPage(Input input) {
+        return (input.perPage() != null && input.perPage() > 0) ? Math.min(input.perPage(), MAX_PER_PAGE) : DEFAULT_PER_PAGE;
+    }
+}
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/GammaDashboardRepositoryAdapter.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/GammaDashboardRepositoryAdapter.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gamma.rest.infra.adapter;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.NullNode;
+import io.gravitee.apim.core.exception.TechnicalDomainException;
+import io.gravitee.gamma.rest.core.observability.dashboard.model.Dashboard;
+import io.gravitee.gamma.rest.core.observability.dashboard.model.DashboardFilter;
+import io.gravitee.gamma.rest.core.observability.dashboard.model.TimeRange;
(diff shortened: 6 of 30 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-standalone/gravitee-apim-rest-api-standalone-container/src/main/java/io/gravitee/rest/api/standalone/spring/StandaloneConfiguration.java
```diff
@@ -51,6 +52,7 @@
         GammaObservabilityFilterConfiguration.class,
         GammaLogsConfiguration.class,
         GammaAnalyticsConfiguration.class,
+        GammaDashboardsConfiguration.class,
```
jourdiw: Maybe a future TODO would be to have a single configuration for all of the gamma observability endpoints?

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/model/Dashboard.java
```diff
@@ -0,0 +1,58 @@
...
+) {
+    public Dashboard {
+        if (id == null || id.isBlank()) {
+            throw new InvalidDashboardException("Dashboard id is required");
+        }
+        if (environmentId == null || environmentId.isBlank()) {
+            throw new InvalidDashboardException("Dashboard environmentId is required");
+        }
+        if (title == null || title.isBlank()) {
+            throw new InvalidDashboardException("Dashboard title is required");
```
jourdiw: WDYT about generalizing the exception message since it is very similar between the different usages?

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/model/TimeRange.java
```diff
@@ -0,0 +1,25 @@
...
+package io.gravitee.gamma.rest.core.observability.dashboard.model;
+
+/**
+ * Time window a dashboard opens on. {@code period} is set for {@code relative}, {@code from} /
+ * {@code to} (epoch millis) for {@code absolute}. Lowercase {@code type} matches the frontend's
+ * discriminated union and mirrors the persisted {@code GammaDashboard.TimeRange} shape.
+ *
+ * @author GraviteeSource Team
+ */
+public record TimeRange(String type, String period, Long from, Long to) {}
```
jourdiw: I would have expected `type` to be an enum 🤔

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/use_case/GetObservabilityDashboardUseCase.java
```diff
@@ -0,0 +1,48 @@
...
+    private final DashboardRepository dashboardRepository;
+
+    public record Input(String environmentId, String dashboardId) {}
+
+    public record Output(Dashboard dashboard) {}
+
+    public Output execute(Input input) {
+        Dashboard dashboard = dashboardRepository
+            .findByIdAndEnvironmentId(input.dashboardId(), input.environmentId())
+            .orElseThrow(() -> new DashboardNotFoundException(input.dashboardId()));
```
jourdiw, line 45: This is more of a query logic than a use case handling. 
It may be worth it to create a query service now so that you may use it in further usecases 🤔

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/use_case/ListObservabilityDashboardUseCase.java
```diff
@@ -0,0 +1,42 @@
...
+public class ListObservabilityDashboardUseCase {
+
+    private final DashboardRepository dashboardRepository;
+
+    public record Input(String environmentId) {}
+
+    public record Output(List<Dashboard> dashboards) {}
+
+    public Output execute(Input input) {
+        return new Output(dashboardRepository.findByEnvironmentId(input.environmentId()));
```
jourdiw: Same thing here, maybe you need to have a query service to control the error thrown?
Likewise, we use pagination as much as possible when listing resources. WDYT about implementing that here?

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/GammaDashboardRepositoryAdapter.java
```diff
@@ -0,0 +1,126 @@
...
+ * Wraps the {@code GammaDashboardRepository} SPI (shipped by OBS-14, {@code gravitee-apim-repository-api})
+ * for the core layer. Anticorruption mapping is non-trivial: the persisted {@code GammaDashboard.Filter}
+ * uses {@code field}/{@code label}/{@code operator}(lowercase)/{@code value}/{@code editable}, not the
+ * {@code name}/UPPERCASE-{@code operator}/{@code values} shape this endpoint's wire format uses via the
+ * shared {@link FilterCondition}/{@link FilterOperator}.
+ *
+ * @author GraviteeSource Team
+ */
+@RequiredArgsConstructor
+public class GammaDashboardRepositoryAdapter implements DashboardRepository {
```
jourdiw, line 45: This is going against the pattern of DDD that we have in apim:
- query service interface in core + implementation in infra
- in the infra file, repository is passed via the bean rather than implementing the interface

Is there a specific need that made you use a different architecture?

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/GammaRootResource.java
```diff
@@ -80,4 +81,13 @@ public LogsResource getLogsResource() {
     public AnalyticsResource getAnalyticsResource() {
         return resourceContext.getResource(AnalyticsResource.class);
     }
+
+    /**
+     * Saved-dashboard read endpoints (list + single). See {@link ObservabilityDashboardsResource} for
+     * the contract.
+     */
+    @Path("/organizations/{orgId}/environments/{envId}/observability/dashboards")
```
jourdiw: It might be helpful to have a GammaObservabilityResource to group all of the resources with `@Path("/organizations/{orgId}/environments/{envId}/observability/...")`

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
```diff
@@ -55,6 +55,15 @@ tags:
 
           All three endpoints return opaque JSON produced by the APIM analytics engine — the
           response schemas below document the stable structure but additional fields may appear.
+    - name: Dashboards
+      description: |
+          Read access to dashboards saved by Gamma Observability users, so a dashboard opens from
+          any machine instead of being tied to one browser's local storage. Write endpoints land in
+          a later increment (OBS-16/17).
```
jourdiw: This is public-facing documentation. Maybe the Jira ticket references could be scrubbed?

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
```diff
@@ -1028,6 +1037,152 @@ paths:
...
+            - $ref: "#/components/parameters/orgIdParam"
+            - $ref: "#/components/parameters/envIdParam"
+        get:
+            tags:
+                - Dashboards
+            summary: List every dashboard saved in the environment
+            description: |
+                Returns every dashboard saved in the context environment, wrapped in
+                `{ "data": [...] }`. Not paginated — the wrapper leaves room to add pagination
+                later without a breaking change.
```
jourdiw: Same, this doesn't feel very pertinent to the user. The user will also see the object body with the schema below, so I don't see the impact of it being referenced here.

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/resources/observability/dashboards/ObservabilityDashboardsResourceTest.java
jourdiw, line 1: Mocking the use case and services limit the expanse that these tests can have.

In general, we encourage no mocking in Resource tests in the DDD part of the application. Can you replace the mocks with the real beans or in memory repos/services?

gravitee-apim-rest-api/gravitee-apim-rest-api-standalone/gravitee-apim-rest-api-standalone-container/src/main/java/io/gravitee/rest/api/standalone/spring/StandaloneConfiguration.java
SamHardi: Fixed in 022944135c. Added `GammaObservabilityConfiguration` as a beanless aggregator (`@Import`ing the five sub-domain configs) — `StandaloneConfiguration` now only imports that single entry point. Documented the pattern in `gravitee-gamma-rest-api/AGENTS.md` §6 so it applies consistently to future observability sub-domains. Verified: 133/133 tests still green (no functional change, just wiring indirection).

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/observability/dashboards/dto/DashboardFilterDto.java
```diff
@@ -0,0 +1,35 @@
...
+package io.gravitee.gamma.rest.resources.observability.dashboards.dto;
+
+import io.gravitee.gamma.rest.core.observability.dashboard.model.DashboardFilter;
+import java.util.List;
+
+/**
+ * Wire shape for one dashboard filter. {@code operator} is UPPERCASE; {@code value} is always an
+ * array, even for single-value operators.
+ */
+public record DashboardFilterDto(String name, String label, String operator, List<String> value, boolean editable) {
```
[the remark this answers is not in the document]
SamHardi: Acknowledged — skipping: this is a false positive. `value` (singular) is the deliberate wire name here, not a naming bug. Three independent sources agree on it: the OBS-15 ticket'\''s own documented response shape (`"value": ["MCP"]`), the OpenAPI schema added in this PR (`DashboardFilter.value: array`), and the existing sibling DTO in this codebase, `FilterConditionDto` (`gravitee-gamma-rest-api/.../observability/logs/dto/FilterConditionDto.java:28`), which also uses singular `value` for its filter-value wire field.

I believe the suggestion is conflating this wire-level DTO field with the *internal* domain model'\''s `FilterCondition.values()` accessor (plural) — a different type, at a different layer, deliberately named differently since the mapping between them is explicit (`DashboardFilterDto.from(DashboardFilter)`). Applying the suggested rename would actually break the documented contract rather than fix a bug.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/model/Dashboard.java
SamHardi: Fixed in 52aa204e13. Extracted a private `requireNonBlank(String, String)` helper in the compact constructor — same messages, no duplication.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/model/TimeRange.java
SamHardi: Fixed in 52aa204e13. Introduced a `TimeRangeType` enum (`RELATIVE`/`ABSOLUTE`), consistent with how `operator` is already handled as `FilterOperator`. The adapter maps the persisted lowercase string to/from the enum (with the same defensive guard as operator — an unsupported value throws `TechnicalDomainException` instead of crashing), and the DTO serializes it back to lowercase on the wire.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/use_case/GetObservabilityDashboardUseCase.java
SamHardi: Acknowledged — skipping for now: this use case is a single 3-line repository call today, so extracting a query service would be premature (no second caller to share it with yet). Genuinely agree it'll be worth it once OBS-16/17 add write-side use cases that need the same "find-or-404" logic — happy to extract then, when the duplication is real rather than anticipated.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/use_case/ListObservabilityDashboardUseCase.java
SamHardi: Fixed pagination in 52aa204e13 — added, using the standard `PaginatedResponseDto` envelope (same shape as `LogsResource`/`ObservabilityFiltersResource`/`TracingResource`), with `page`/`perPage` resolved and clamped the same way `SearchObservabilityLogsUseCase` does. Since `GammaDashboardRepository` (OBS-14) has no native pagination, the use case fetches the full per-environment list and slices it — the same fetch-then-slice approach `TracingPortAdapter` already uses for its own SPI limitation. Updated the OBS-15 ticket and OpenAPI spec to match.

On the query-service point: see my reply on the sibling Get use case thread — same reasoning, skipping until there's a second caller.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/GammaDashboardRepositoryAdapter.java
SamHardi: This mirrors the pattern already merged in this module for every other Observability sub-domain — e.g. `TracingPortAdapter implements TracingPort`, wired via `@Bean` in `GammaTracingConfiguration` (and identically for `ObservabilityLogsDataPortAdapter`/`GammaLogsConfiguration`, and the filters/analytics adapters). It's also what `gravitee-gamma-rest-api/AGENTS.md` §6 documents as this module's convention (adapter implementing the core port, wired via a per-domain `@Configuration`).

If you'd like this module to move to the wider apim-core query-service convention instead, I think that's worth doing consistently across all five Observability sub-domains rather than introducing a one-off different pattern just for dashboards — happy to open a follow-up ticket for that if you want to pursue it. Let me know if I'm missing something about why dashboards specifically should diverge from the existing precedent here.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/GammaRootResource.java
SamHardi: Fixed in 52aa204e13. Added `GammaObservabilityResource` as a routing aggregator for the whole `/observability/*` namespace (traces, filters, logs, analytics, dashboards) — `GammaRootResource` now has a single locator to it instead of one per sub-domain, mirroring the `GammaObservabilityConfiguration` aggregator added for Spring wiring. URL-transparent (JAX-RS concatenates the `@Path` segments across the extra locator hop) — verified all existing resource tests (Tracing/Filters/Logs/Analytics) still pass unchanged.

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
SamHardi: Fixed in 52aa204e13. Removed the `(OBS-16/17)` reference from the Dashboards tag description.

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
SamHardi: Fixed in 52aa204e13 — this became moot once pagination was actually implemented (see the sibling thread on `ListObservabilityDashboardUseCase.java`). The endpoint description no longer explains an internal "why", it just documents the actual paginated behavior.

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/resources/observability/dashboards/ObservabilityDashboardsResourceTest.java
SamHardi: This matches the existing `LogsResourceTest` pattern in this same module (`LogsTestConfiguration` mocks `SearchObservabilityLogsUseCase`/`GetObservabilityLogDetailUseCase` the same way) — I followed that as precedent rather than introducing a different test style for dashboards.

If you'd like resource tests in this module to move to real beans / in-memory repos instead of mocked use cases, I think that's a module-wide testing-convention change worth doing consistently (Tracing/Filters/Logs/Analytics/Dashboards) rather than just for this PR — happy to open a follow-up if you want to pursue it. Note the use-case layer itself (`ListObservabilityDashboardUseCaseTest`, `GetObservabilityDashboardUseCaseTest`) already uses a real in-memory `DashboardRepository`, no mocks — it's only the resource-layer test that mocks the use cases, same as the existing sibling tests.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/use_case/ListObservabilityDashboardUseCase.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public record Output(List<Dashboard> dashboards, long totalCount, int page, int perPage) {}
+
+    public Output execute(Input input) {
+        int page = resolvePage(input);
+        int perPage = resolvePerPage(input);
+
+        List<Dashboard> all = dashboardRepository.findByEnvironmentId(input.environmentId());
+        int fromIndex = Math.min((page - 1) * perPage, all.size());
+        int toIndex = Math.min(fromIndex + perPage, all.size());
```
jourdiw, line 54: Ideally, the pagination would take place at the repository level for better performance of SQL DBs. 

If you find that is overkill, there may also be some utility functions laying around in APIM for resource-level pagination that you could use instead of handling it locally here.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/GammaDashboardRepositoryAdapter.java
```diff
@@ -0,0 +1,148 @@
...
+            return action.get();
+        } catch (TechnicalException e) {
+            throw new TechnicalDomainException(message, e);
+        }
+    }
+
+    @FunctionalInterface
+    private interface TechnicalAction<T> {
+        T get() throws TechnicalException;
+    }
```
jourdiw: These are pretty generic. Perhaps they can be refactored in a shared area so that other classes could use them as well?

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/resources/observability/dashboards/ObservabilityDashboardsResourceTest.java
jourdiw: I understand. For other changes I have recommended, I can understand delaying because it changes the followed architecture.

I would push a bit for the tests to be improved by not mocking use cases now rather than later. It is not an architecture change yet it immediately will improve code quality for each resource to be created.

jourdiw (approved):
Pre-approved

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/resources/observability/dashboards/ObservabilityDashboardsResourceTest.java
SamHardi: Fair enough — you're right that this one is a test-quality change rather than an architecture change, so I've done it now in 3e6c376517.

`DashboardsTestConfiguration` now wires the **real** `ListObservabilityDashboardUseCase`/`GetObservabilityDashboardUseCase` over the in-memory `DashboardRepository`, and the tests seed data through the repository instead of stubbing use-case returns. Only `PermissionService`/`EnvironmentService` stay mocked (platform boundaries the resource can't run without, provided by the shared `ResourceContextConfiguration`).

Concretely it did expand what the tests cover — these cases are now exercised end-to-end instead of being stubbed away, and three of them are new:
- a dashboard in another environment is genuinely filtered out of the list (was previously just a mocked return value)
- `GET /{id}` on a dashboard belonging to another environment really returns 404 via the repository lookup, rather than a mocked `thenThrow`
- pagination actually slices (`?page=2&perPage=2` over 5 seeded dashboards returns `dash-3`/`dash-4` with `totalCount: 5`), instead of asserting that the right `Input` was passed to a mock
- `widgets` and `timeRange` round-trip through the real mapping rather than a fixture handed straight back

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/dashboard/use_case/ListObservabilityDashboardUseCase.java
SamHardi: Agreed on the principle. I went with the second option — with one nuance on where the "utility" lives.

On repository-level pagination: it would mean adding a `Pageable` overload to the `GammaDashboardRepository` SPI plus the Mongo impl, the JDBC impl and the shared TCK — i.e. reopening OBS-14, which is already merged and closed. Given a dashboard list is per-environment and expected to stay in the tens, I don't think that trade is worth it in this PR, but it's a legitimate follow-up if the volume assumption ever breaks — happy to open a ticket for it.

On existing utilities: the closest one is `AbstractResource#computePaginationData` (management-v2-rest), but it isn't reusable here — it's a `protected` method on a v2 REST base class in another module, so using it would mean putting pagination in the JAX-RS layer rather than the use case, and its semantics differ (it throws `PaginationInvalidException` when the offset is past the end, where this endpoint returns an empty page). `PageUtils` (apim `infra.repository`) solves a different problem — iterating pages that a repository already paginates.

What I did instead is follow the precedent that already exists at this exact layer in apim core: `GetFilterValuesUseCase` (`io.gravitee.apim.core.analytics_engine.use_case`) paginates a fully-fetched list with the same `Math.min` + `subList` shape, for the same reason. So this isn't a local invention — it matches how core use cases already do resource-level pagination. Let me know if you'd rather see that extracted into a shared core helper; that felt like it belonged in its own change rather than bundled here, since it'd touch existing apim core code.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/GammaDashboardRepositoryAdapter.java
SamHardi: Done in 3e6c376517 — extracted to `io.gravitee.gamma.rest.infra.adapter.RepositoryCalls`, a small public utility (`RepositoryCalls.wrap(call, message)` + the `RepositoryCall<T>` functional interface), so any future `port/repository/`-backed adapter maps SPI failures the same way instead of repeating the try/catch.

Placed it in `infra.adapter` alongside the other shared adapter helpers already there (`NumericRangeAccumulator`, `FailureOriginClassifier`, `AnalyticsBucketTypeEnricher`).

For context on the scope: this adapter is currently the only one in gamma that touches a *checked* `TechnicalException` — the other ports (tracing, logs, analytics, filters) front reactive or non-throwing SPIs — so there's no second caller to convert today. It should pay off as soon as OBS-16/17 add the write operations and any other domain starts persisting its own data through `port/repository/`.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/RepositoryCalls.java
```diff
@@ -0,0 +1,55 @@
...
+            return call.execute();
+        } catch (TechnicalException e) {
+            throw new TechnicalDomainException(message, e);
+        }
+    }
+
+    /**
+     * A repository SPI call returning a value and declaring the checked {@link TechnicalException}.
+     */
+    @FunctionalInterface
```
SamHardi: Fair point — the comment did just paraphrase the signature. Reworded in 92238f981f to explain why the interface exists (a plain `Supplier` can't declare the SPI's checked exception, which would push the try/catch back into every adapter method) and the contract for implementors (let `TechnicalException` propagate — translating it is `wrap`'s job).

PR #18961, opened 2026-08-04
fix(gateway): harden distributed sync write path (bounding + retry) [APIM-14672]

ThibaudAV:
## Issue

https://gravitee.atlassian.net/browse/APIM-14672

## Description

Follow-up to the cold-start fix (#18781), hardening the distributed-sync **write** path, which was dropping most subscriptions/api-keys during bulk syncs.

- **Bound concurrent distributed writes** — a shared reactive gate caps the aggregate in-flight Redis writes below the client waiting queue (`max-waiting-handlers`), so a large fan-out is back-pressured instead of overflowing and dropping events.
- **Retry transient Redis failures** — recoverable write failures (connection drop/reset/close, reconnect window, command timeout, saturated queue) are retried with an exponential backoff (shared `RxHelper` policy), so a Redis restart / brief outage heals within the sync cycle instead of leaving events unsynced. The key write is an idempotent `HSET`.
- **Clearer sync log** — distinguish initial vs incremental sync and report the total apis currently deployed (the old `N apis synchronized` counted only the current cycle, logging `0` even with thousands deployed).

## Additional context

Root cause: each api fanned out `1 + N_subscriptions + M_apiKeys` Redis writes with only a *per-deployable* concurrency cap; the aggregate overflowed the Redis waiting queue, the overflow was dropped, and the failure was misclassified as permanent (so not re-distributed on the replayed window).

Not covered here (to be tracked separately): redistribution after a **long** Redis outage while the primary stays up with static apis — this self-heals on a primary restart (full initial sync) or on any api change.

Tests: global-concurrency bound + literal `Redis waiting queue is full` reproduction, `WriteGate` permit release/leak guards, and the retry-classification predicate.


phiz71 (commented):
Nice work — the diagnosis is spot on and the `WriteGate` mechanics hold up. I walked through the dispose/hand-off interleavings and the permit accounting is correct: `held` is booked under the lock before any signal is emitted, and `CompletableDoFinally.dispose()` runs the cancellable (which takes the lock) *before* `doFinally`, so there is no window where a counted permit gets orphaned. The javadoc is genuinely good.

One blocker and a few things worth discussing — details inline. Summarised here:

### The fix only exists on `4.11.x`

`WriteGate` isn't on `master` or `4.12.x` (checked both), whereas the earlier APIM-14672 PRs (#18682 / #18750 / #18751) were ported everywhere. As it stands a customer who takes this fix in 4.11 loses it when upgrading to 4.12/4.13. Forward-ports needed — or a note in the description if that's deliberate.

### Blast radius

The gate is process-wide and a permit is held for the entire write *including* its retries, while no command timeout is configured on the client. That combination means a small number of hung writes can freeze all distributed sync on the gateway — a wider failure mode than before this PR, where the impact stayed local to one deployable. See the comment on `createOrUpdateKey`.

### Tests

Good coverage of the gate itself — the two dispose tests target exactly the permit-leak windows, and asserting `peak == 8` rather than `<= 8` also catches over-serialization regressions. Using `doOnTerminate` rather than `doFinally` for the gauge is the right call and well explained.

Two gaps:

- **The retry wiring isn't tested**, only the predicate is. A test with a mocked `RedisClient` failing twice then succeeding would guard that `retryWhen` is actually applied, that `WRITE_RETRY_MAX_ATTEMPTS` is respected, and that the last error still surfaces. That wiring is what silently disappears in a future refactor — the classification is the part least likely to break.
- **No multi-threaded test of the gate**, even though all of its subtlety lives in the dispose/hand-off races. A N-thread hammer asserting `inFlight == 0` at the end would be the real safety net; the three current gate tests are single-threaded.

Also, `should_not_drop_events_when_redis_waiting_queue_would_overflow` largely restates the first test's invariant with the same instrumentation. It documents the ticket well, but it doesn't add much guarantee.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/distributed/service/DefaultDistributedSyncService.java
```diff
@@ -70,8 +73,24 @@ public class DefaultDistributedSyncService implements DistributedSyncService {
...
 
+    /**
+     * Upper bound on the number of Redis writes issued concurrently across <b>all</b> distributions.
+     * {@link #WRITE_MAX_CONCURRENCY} only caps a single deployable's fan-out, but the deployer runs many
+     * deployables in parallel, so the aggregate in-flight writes (≈ parallelism × {@link #WRITE_MAX_CONCURRENCY})
+     * must also be bounded. Kept well below the Redis client waiting queue (max-waiting-handlers, 1024 by
+     * default) — leaving room for reads/state/scan commands — so the primary never triggers "Redis waiting
+     * queue is full", which used to silently drop subscription/api-key events during bulk syncs.
+     */
+    static final int DISTRIBUTION_WRITE_MAX_CONCURRENCY = 512;
```
phiz71: The "1024 by default" is accurate, but only because `RedisConnectionFactory` hardcodes `options.setMaxWaitingHandlers(1024)` — the Vert.x default is actually 2048. So these are two magic numbers, in two different modules, coupled by a comment and nothing else: if someone bumps or lowers the waiting-handler setting, nothing here breaks or even warns.

Worth at least cross-referencing `RedisConnectionFactory` by name in this javadoc so the link is greppable from both ends.

Side note on placement: `WriteGate` lives in `DefaultDistributedSyncService`, which is repository-agnostic, while its entire rationale (waiting queue, `max-waiting-handlers`) is Redis-specific. Not worth churning now, but if a second distributed-sync backend ever shows up, this bound will be sized for the wrong client.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/distributed/service/DefaultDistributedSyncService.java
```diff
@@ -287,7 +306,16 @@ public Completable distributeIfNeeded(final NodeMetadataDeployable deployable) {
...
+        return events.flatMapCompletable(this::distributeSingle, false, WRITE_MAX_CONCURRENCY);
+    }
+
+    /**
+     * Writes a single event, gated by the shared {@link #writeGate}: a permit is acquired before the write
+     * and released once the write terminates or is disposed. When permits are exhausted the acquire step
+     * parks, back-pressuring the (awaited) deployer chain instead of flooding the Redis client.
+     */
+    private Completable distributeSingle(final DistributedEvent event) {
+        return writeGate.runGated(distributedEventRepository.createOrUpdate(event));
```
phiz71: The gate covers everything flowing through `distribute(...)`, but `distributeIfNeeded(ApiReactorDeployable)` calls `distributedEventRepository.updateAll(...)` directly on the UNDEPLOY branch, bypassing it. `updateAll` then does its own `flatMapCompletable(..., UPDATE_MAX_CONCURRENCY = 32)` per api.

So on a mass undeploy, N parallel deployables × 32 ungated writes stack on top of the gate's 512 — which is the exact failure mode this PR fixes, on a neighbouring scenario. The retry does cover those writes (they go through `createOrUpdateKey`), but the bound doesn't.

Either route that path through the gate too, or call out in the description why it's out of scope.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/distributed/service/DefaultDistributedSyncService.java
```diff
@@ -311,4 +339,89 @@ private Completable trackFailure(final Completable distribution, final String ev
...
+        private final int maxConcurrency;
+        private final Deque<Waiter> waiters = new ArrayDeque<>();
+        private int inFlight;
+
+        WriteGate(final int maxConcurrency) {
+            this.maxConcurrency = maxConcurrency;
+        }
+
+        Completable runGated(final Completable write) {
+            AtomicBoolean held = new AtomicBoolean();
```
phiz71: `held` is allocated at **assembly** time, not per subscription — so every subscriber of the returned `Completable` shares one flag.

This is correct today, because every caller reaches `distributeSingle` through a `Completable.defer` and therefore gets a fresh `held` on each re-subscription. But the guarantee lives in the callers, not here: add a `retry()` or `repeat()` above the gate someday and the permit counter corrupts silently — the worst possible failure mode for this class.

Wrapping the body in `Completable.defer(() -> { ... })` makes it locally safe and costs nothing.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/distributed/service/DefaultDistributedSyncService.java
```diff
@@ -311,4 +339,89 @@ private Completable trackFailure(final Completable distribution, final String ev
...
+                    if (next == null) {
+                        inFlight--;
+                        return;
+                    }
+                    if (next.emitter().isDisposed()) {
+                        continue;
+                    }
+                    next.held().set(true);
+                }
+                next.emitter().onComplete();
```
phiz71: Two consequences of handing the permit off by calling `onComplete()` from inside `release()`, both worth a thought:

**1. It changes the subscribing thread, and therefore the Redis connection.** The next write is subscribed on whichever thread completed the previous one — a Vert.x event loop — instead of the deployer thread. `RedisClient.resolveLoop()` keys the connection on `currentLoopKey()`: no Vert.x context → `FALLBACK_LOOP_KEY`, otherwise one connection per event loop. Before this PR all these writes went out from the deployer thread, i.e. a single connection; now the first 512 do and the rest are issued from event loops. That's arguably fine (more total waiting-handler capacity), but it does undercut the "512 well below 1024" reasoning above, which is per connection. Worth verifying and reflecting in the javadoc.

**2. Synchronous recursion.** `release()` → `onComplete()` → next write subscribed → if it terminates synchronously → `doFinally` → `release()` → … Stack depth grows with the number of parked waiters, which can be in the thousands. Unlikely to bite in practice (the Redis write is async and retries go through a timer), but a scheduler hop or a trampoline on the hand-off would rule it out entirely.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/distributedsync/RedisDistributedEventRepository.java
```diff
@@ -57,6 +59,16 @@ public class RedisDistributedEventRepository implements DistributedEventReposito
...
 
+    // A write failing because Redis is momentarily unreachable (restart / brief outage) is retried with an
+    // exponential backoff instead of being surfaced as a failed distribution: the event key is idempotent
+    // (HSET), so retrying is safe, and it lets a short Redis blip heal within the same sync cycle instead of
+    // leaving the event unsynced until the api changes or the node restarts. Bounded so a long outage still
+    // gives up and lets the sync window be replayed: 5 retries after the initial attempt at 200ms doubling
+    // (200/400/800/1600/3200ms) ≈ up to ~6s of blip absorbed.
+    static final int WRITE_RETRY_MAX_ATTEMPTS = 5;
+    static final long WRITE_RETRY_INITIAL_BACKOFF_MS = 200;
+    static final long WRITE_RETRY_MAX_BACKOFF_MS = 5_000;
```
phiz71, line 83: This ceiling is dead code: 5 attempts at factor 2 from 200ms caps the computed delay at 3200ms, so `min(5000, delay)` never applies. Harmless, but either drop it or pick a value that actually engages, otherwise the next person to tune the backoff will assume it's doing something.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/distributedsync/RedisDistributedEventRepository.java
```diff
@@ -139,7 +151,48 @@ public Completable updateAll(
     }
 
     private Completable createOrUpdateKey(final String key, final DistributedEvent distributedEvent) {
-        return send(redisAPI -> redisAPI.hset(buildUpdateArgs(key, distributedEvent))).ignoreElement();
+        return send(redisAPI -> redisAPI.hset(buildUpdateArgs(key, distributedEvent)))
```
phiz71: This is the one I'd want addressed before merge.

`RedisConnectionFactory` configures no command timeout — only `connectTimeout` (`RedisClient.java:177`). So a command that never resolves (half-open connection) is never failed by the client. Combined with the new gate, that write holds its permit **forever**, and since the gate is process-wide, a handful of hung writes is enough to freeze *all* distributed sync on the gateway. Before this PR the damage stayed local to one deployable's fan-out, so this genuinely widens the blast radius.

Secondary, same area: the `retryWhen` sits *inside* the permit, so during an outage each write pins its permit for ~6.2s. With 512 permits the whole distribution goes into an accordion — that's the intended back-pressure, but it's worth sanity-checking that duration against the sync cycle cadence (do cycles end up overlapping?).

A `.timeout(...)` on the write would cap both, and placing it *before* the `retryWhen` makes the timeout itself retryable — the predicate below already matches `"timeout"`.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/distributedsync/RedisDistributedEventRepository.java
```diff
@@ -139,7 +151,48 @@ public Completable updateAll(
...
+            .ignoreElement()
+            // Heal a short Redis blip within the sync cycle: retry recoverable write failures with an
+            // exponential backoff, then surface the last error so the sync window is replayed. HSET is
+            // idempotent, so retrying is safe. Uses the shared RxHelper policy rather than a bespoke one.
+            .retryWhen(
```
phiz71: **The retry is completely silent.** `RxHelper` logs nothing and nothing is added here, so a burst of Redis blips absorbed by the backoff leaves no trace at all. Given this path just cost one investigation, a `doOnError` at debug/warn (key + attempt count) seems cheap insurance against a second blind one.

**Separately — worth a follow-up ticket rather than a change here.** `RedisClient.notifyConnectionFailure(...)` exists precisely to invalidate a half-open connection the Vert.x handlers haven't noticed (its javadoc: `connected=true` is left set and automatic reconnection is blocked). It's only ever called from `RedisRateLimitRepository`, and these are two distinct clients — `redisDistributedSyncClient` vs `redisRateLimitClient`, separate scopes — so nothing will ever heal the sync one.

Net effect: for a clean restart or a brief outage the handlers do fire and this retry works as advertised. But in the half-open case the 5 retries all burn against the same dead connection and then give up, while pinning a permit for 6.2s — so the retry buys nothing exactly where it's most needed. It's pre-existing, not introduced here, hence follow-up rather than blocker.

Note it isn't a one-liner either: `notifyConnectionFailure` does `resolveLoop()` on the *current* thread, whereas the error surfaces on the connection's event loop while `redisApi()` resolved the loop from the subscribing thread (often `FALLBACK_LOOP_KEY`, off-context). A naive `.doOnError(redisClient::notifyConnectionFailure)` could invalidate the wrong `LoopRedis` — or create an empty one via `computeIfAbsent` — leaving the actually-broken connection in place.

gravitee-apim-repository/gravitee-apim-repository-redis/src/main/java/io/gravitee/repository/redis/distributedsync/RedisDistributedEventRepository.java
```diff
@@ -139,7 +151,48 @@ public Completable updateAll(
...
+     * within the sync cycle. Package-private and static so it can be unit-tested as a pure predicate.
+     */
+    static boolean isRetryableWriteFailure(final Throwable error) {
+        if (RedisClient.isRecoverableConnectionFailure(error)) {
+            return true;
+        }
+        for (Throwable current = error; current != null; current = current.getCause()) {
+            final String message = current.getMessage();
+            if (message != null) {
+                final String normalized = message.toLowerCase();
```
phiz71: `toLowerCase()` without a locale (Sonar `java:S1449`). Consistent with `isRecoverableConnectionFailure` right next door, so not a regression — but both are exposed to the Turkish-locale `I` → `ı` case, and here the consequence is that the retry silently stops matching. `Locale.ROOT` costs nothing.

Also, matching on `"closed"` alone is quite broad for a predicate that decides whether to retry — it'll catch unrelated messages that happen to contain the word.

phiz71:
Chokepoint refactor is the right call — cleaner than routing `updateAll` through the gate as I'd suggested — and it closes everything from the last round, forward-port labels included (I checked `apply-on-4-12-x` does fire from a `4.11.x` base, cf. #18682 → #18750/#18751).

One thing worth deciding before merge: the timeout never invalidates the connection, so in the half-open case the retries are lost in advance. `notifyConnectionFailure` gates on `isRecoverableConnectionFailure`, which doesn't know about `TimeoutException` — the same gap you closed in `isRetryableWriteFailure` with the `instanceof` check — and `.timeout(long, TimeUnit)` emits on RxJava's computation scheduler, so the `Vertx.currentContext() == null` guard would bail out anyway. All 6 attempts then hit the same stale connection and a permit is held for ~66s (6 × 10s + backoff), not the "~6s" the class comment describes. Bounded rather than infinite is still a real improvement, but either invalidate on timeout, or align the comments on the real number so nobody reads "~6s" during an incident.

Minor, all fine as follow-ups: the synchronous recursion in `release()` is unchanged; the two `DefaultDistributedSyncServiceTest` cases went away with the refactor, so nothing asserts end-to-end that a bulk sync stays bounded; and no test pins `WRITE_RETRY_MAX_ATTEMPTS`. On Aikido — I'd dismiss all four: `ApiManagerImpl.apis` is a `ConcurrentHashMap` (and its `synchronized` suggestion would be actively worse), the "log and rethrow" is a per-attempt debug log in a retry pipeline, and the two extract-this-lambda ones are cosmetic.

ThibaudAV:
Thanks — and agreed the chokepoint was the cleaner path. On the half-open timeout point: you're right on both counts (`isRecoverableConnectionFailure` doesn't know `TimeoutException`, and `.timeout()` fires off the Vert.x context so `notifyConnectionFailure` bails), so the retries re-hit the same stale socket and the real worst-case hold is ~66s, not ~6s.

For this PR I went with your second option — aligned the comments to the real number rather than changing behaviour on a 4.11.x patch: the backoff comment now says it's only the inter-attempt waiting, `WRITE_COMMAND_TIMEOUT_MS` documents the ~66s half-open worst case and that a timeout doesn't currently invalidate the connection, and the `doOnError` comment notes the invalidation is a no-op for a bare timeout. (Force-pushed, folded into the refactor commit for a clean history.)

Not opening a follow-up ticket for the invalidate-on-timeout change for now — we'll file one later if it proves needed. On your minor points: agree they're fine as follow-ups, and agree on dismissing the four Aikido findings for the reasons you gave.

phiz71 (approved):

PR #18963, opened 2026-08-04
feat(reporter): index authz decisions as event metrics

podlesrafal:
Related to: AUTHZ-2
Indexes authorization decisions as `AuthzEventMetrics` in the `gravitee-event-metrics` data stream, so they are served by the same reporting pipeline as every other event type.

- `ElasticsearchFormatter` gains a formatter for the type, plus `authz-event-metrics.ftl` for es7x/es8x/es9x
- `MetricsType` registers the type — without this entry `canHandle` returns false and the reporter discards the event with no exception and no log
- `event-metrics-mapping.ftl` adds the authz fields, with `matched-policies` as `nested`
- `AuthzEventMetricsFormatter` covers the CSV output of the file reporter

Every string is interpolated through `?json_string`. Decision fields carry request-derived data — the PEP resolves subject, action and resource from EL expressions over the live HTTP request — so an unescaped value could inject arbitrary keys into an audit document. There is a golden-file test with quotes in `resource-id`, `subject-id` and `reasons` covering exactly that.

`MatchedPolicy.annotations` is deliberately not indexed. The PDP fills it with `Map.of()` pending a producer change, so the field would be dead weight, and typing it would require `flattened` / `flat_object` — unavailable on Elasticsearch below 7.3 and OpenSearch below 2.7, both inside the range this connector declares support for. A failed index template aborts preparation for every event-metrics type, not just authz. It goes in together with the producer side.

Requires gravitee-reporter-api 2.7.0.


gravitee-apim-reporter/gravitee-apim-reporter-common/src/main/resources/freemarker/common/es/event/authz-event-metrics-fields.ftl
```diff
@@ -0,0 +1,71 @@
...
+    ,"resource-type": "${metrics.getResourceType()?json_string}"
+</#if>
+<#if metrics.getResourceId()??>
+    ,"resource-id": "${metrics.getResourceId()?json_string}"
+</#if>
+<#if metrics.getDecision()??>
+    ,"decision": "${metrics.getDecision()?json_string}"
+</#if>
+<#if metrics.getMatchedPolicies()??>
+    ,"matched-policies": [<#list metrics.getMatchedPolicies() as p>{
```
sclarkamlab, line 51: The PR description says annotations map to `flattened` / `flat_object` via `annotationsFieldType` from the index preparer, but none of that is in the diff. `MatchedPolicy` in reporter-api 2.7.0 carries an `annotations()` map and this template only emits `id`, `name` and `effect`, so annotations are silently dropped at index time. `annotationsFieldType` doesn't appear anywhere in the repo and the mapping template has no `annotations` property either. Did the wiring get lost in a rebase?

gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/mapping/opensearch/OpenSearchIndexPreparer.java
```diff
@@ -20,6 +20,7 @@
 import io.gravitee.apim.reporter.elasticsearch.mapping.AbstractIndexPreparer;
 import io.gravitee.common.templating.FreeMarkerComponent;
 import io.gravitee.elasticsearch.client.Client;
+import java.util.Map;
```
sclarkamlab: This import is unused. Looks like a leftover from a `getTemplateData()` override that would have supplied `annotationsFieldType`, see my comment on the fields template.

gravitee-apim-reporter/gravitee-apim-reporter-common/src/main/java/io/gravitee/apim/reporter/common/formatter/csv/v4/AuthzEventMetricsFormatter.java
```diff
@@ -0,0 +1,81 @@
...
+        appendString(buffer, data.getSubjectType());
+        appendString(buffer, data.getSubjectId());
+        appendString(buffer, data.getAction());
+        appendString(buffer, data.getResourceType());
+        appendString(buffer, data.getResourceId());
+        appendString(buffer, data.getDecision());
+        appendString(buffer, joinPolicyNames(data));
+        appendString(buffer, join(data.getReasons()));
+        appendString(buffer, data.getSearchType());
+        appendLong(buffer, data.getResultCount());
```
sclarkamlab, line 55: The CSV row skips `batchId`/`batchIndex`/`batchSize` (after `policyGeneration`) and `pageSize`/`hasMore` (after `resultCount`), all of which the ES template indexes. I ran a search-shaped event through both formatters to check: the ES doc has all five, the CSV goes straight from `policyGeneration` to `subjectType` and from `resultCount` to `errorType`. Intentional, or should they be added?

gravitee-apim-reporter/gravitee-apim-reporter-common/src/test/java/io/gravitee/apim/reporter/common/formatter/elasticsearch/ElasticsearchFormatterTest.java
```diff
@@ -62,6 +62,8 @@ protected Type type() {
             "topic event metrics, v4.metric.event.TopicEventMetrics, v4/topic-event-metrics.json, elasticsearch/v4/topic-event-metrics.json",
             "topic event metrics with null api, v4.metric.event.TopicEventMetrics, v4/topic-event-metrics-null-api.json, elasticsearch/v4/topic-event-metrics-null-api.json",
             "operation event metrics, v4.metric.event.OperationEventMetrics, v4/operation-event-metrics.json, elasticsearch/v4/operation-event-metrics.json",
+            "authz event metrics, v4.metric.event.AuthzEventMetrics, v4/authz-event-metrics.json, elasticsearch/v4/authz-event-metrics.json",
+            "authz event metrics with quotes, v4.metric.event.AuthzEventMetrics, v4/authz-event-metrics-with-quotes.json, elasticsearch/v4/authz-event-metrics-with-quotes.json",
```
sclarkamlab, line 66: The other event metrics types also have rows in `should_format_with_options` below, plus `CsvFormatterTest`, `JsonFormatterTest` and `MsgPackFormatterTest`. authz is missing from all four, so the new CSV formatter ships with no test at all.

gravitee-apim-reporter/gravitee-apim-reporter-common/src/main/resources/freemarker/common/es/event/authz-event-metrics-fields.ftl
```diff
@@ -0,0 +1,71 @@
+<#-- @ftlvariable name="metrics" type="io.gravitee.reporter.api.v4.metric.event.AuthzEventMetrics" -->
+,"operation": "${metrics.getOperation()?json_string}"
+,"event-id": "${metrics.getEventId()?json_string}"
+,"status": "${metrics.getStatus()?json_string}"
```
sclarkamlab: These are the only three unguarded fields in the file. If any of them is null the whole format call throws, FreeMarker `InvalidReferenceException` wrapped in an `IllegalArgumentException` (verified by running the formatter with a null `status`), instead of omitting the field like the guarded ones. Worth wrapping them in `<#if>` like the rest unless they're guaranteed non-null.

sclarkamlab (commented):
Ran a review of this locally and verified the findings by actually running the formatters. Main one: the annotations wiring described in the PR body isn't in the diff. Details inline.

podlesrafal:
Thanks — all five fixed, and you were right that the description had drifted.

- **annotations**: the wiring was removed deliberately, not lost in a rebase, and I failed to update the description. The PDP fills `MatchedPolicy.annotations` with `Map.of()` pending a producer change, so indexing it would be dead weight — and typing it needs `flattened` / `flat_object`, which are unavailable on Elasticsearch below 7.3 and OpenSearch below 2.7, both inside the range `BeanFactoryBuilder` declares support for. Since a failed index template aborts preparation for *every* event-metrics type, that risk buys nothing today. Description now says so; the field goes in together with the producer side.
- **unused import**: gone, it was the leftover you suspected.
- **CSV**: added `batchId`/`batchIndex`/`batchSize` and `pageSize`/`hasMore`, so the row now matches what the ES template indexes.
- **tests**: authz added to `should_format_with_options`, `CsvFormatterTest`, `JsonFormatterTest` and `MsgPackFormatterTest`. That was also what tanked the Sonar gate.
- **unguarded fields**: guarded. You're right that the no-args constructor makes them reachable — for an audit trail a document missing one field beats an exception that drops the whole record.

The escaping and the quoted-input golden file were already in from an earlier pass; re-verified on a live gateway that an injected `","injected":"yes` lands as a literal string rather than a new key.

sclarkamlab (approved):

PR #18964, opened 2026-08-04
fix(reporter): stop dropping reportables with no remote address

podlesrafal:
`AbstractFormatter.sanitizeRemoteAddress` dereferences the remote address without a null check, so any reportable without one throws an NPE inside the bulk transformer and is dropped silently — no exception surfaces and no document is written.

One-line null guard.


podlesrafal:
No longer needed. The null remote address only showed up because authz decisions were being reported as request-shaped v4 `Metrics`. They are now a first-class `AuthzEventMetrics` event type (gravitee-reporter-api#139), which never goes through this formatter path, so the guard has no remaining caller.

PR #18966, opened 2026-08-05
feat: add API product navigation and details models

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-101

## Summary

Add the Portal Next models and service layer required to consume API Product navigation and details from the Portal API.

## Changes

- Extend the Portal navigation item union with `API_PRODUCT` and `apiProductId`.
- Add typed API Product details and included API summary models.
- Add reusable API Product navigation and details fixtures.
- Add `ApiProductsService` to retrieve product details by ID.
- Preserve HTTP errors for handling by future UI components.
- Add focused tests for API Product navigation responses and details retrieval.

## Out of scope

- API Product routing and navigation entries.
- API Product details UI.
- Product documentation rendering.
- Backend or HTTP contract changes.

### DIFF
--- a/gravitee-apim-portal-webui-next/src/entities/api-product/api-product.ts
+++ b/gravitee-apim-portal-webui-next/src/entities/api-product/api-product.ts
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2024 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export type ApiProductKind = 'AI_WORKSPACE';
+
+export interface ApiProductApi {
+  id: string;
+  name: string;
+  version: string;
+}
+
+export interface ApiProduct {
+  id: string;
+  name: string;
+  description?: string;
+  version: string;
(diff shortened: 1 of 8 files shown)


gravitee-apim-portal-webui-next/src/services/api-products.service.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ApiProduct } from '../entities/api-product/api-product';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ApiProductsService {
+  private readonly http = inject(HttpClient);
+  private readonly configService = inject(ConfigService);
+
+  details(apiProductId: string): Observable<ApiProduct> {
```
thomasballivet: nit: `get()` or `getById(apiProductId)` would fit better wdyt?

gravitee-apim-portal-webui-next/src/entities/api-product/api-product.ts
```diff
@@ -0,0 +1,34 @@
...
+  name: string;
+  version: string;
+}
+
+export interface ApiProduct {
+  id: string;
+  name: string;
+  description?: string;
+  version: string;
+  kind: ApiProductKind;
```
thomasballivet: Are you sure about kind non-nullability? A classic (non AI Workspace) product has kind = null right? Are we sure the backend always fills it?

gravitee-apim-portal-webui-next/src/services/api-products.service.ts
jaroslawlakomy-grav: Renamed to `getById(apiProductId)` to make the lookup semantics explicit.

gravitee-apim-portal-webui-next/src/entities/api-product/api-product.ts
jaroslawlakomy-grav: Good catch! Confirmed that classic API Products have `kind = null`. The frontend model now accepts `null`, and a regression test has been added. The OpenAPI nullability mismatch will be addressed separately.

thomasballivet (approved):

PR #18967, opened 2026-08-05
feat(gamma-platform): user and group memberships

pragsty:
## Issue

https://gravitee.atlassian.net/browse/FOUND-85
https://gravitee.atlassian.net/browse/FOUND-86

## Description

Add group memberships management to the Gamma Platform user detail page, with classic console parity for memberships, inherited resources, and cross-module navigation.
- **Group memberships card** with per-environment tabs, inline role editing, add-to-group sheet, and remove-with-confirmation
- **Inherited resources** tables for APIs, API Products, and Applications with client-side search and pagination
- **Cross-module links** to APIM and Platform detail pages via `@gravitee/gamma-modules-sdk/routing` `buildModuleNavPath`
- **Permission-aware actions**: add (create), edit roles (update), remove (delete); disabled for archived users
- **API Primary Owner guard**: locked API role select and disabled remove with tooltip
- **Classic console parity**: empty groups show "No group" row while inherited resource tables remain visible

https://github.com/user-attachments/assets/0257a5d1-b327-4354-9d76-194eb7a3aaa9


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/users/services/organizationUsers.ts
aqibmohammadkhan, line 1: `getOrganizationUserGroups` / `listEnvironmentGroups` default to 100, while classic memberships load groups at 9999. Inherited resources correctly use INHERITED_RESOURCES_FETCH_SIZE = 9999. Orgs with >100 groups will silently miss memberships and omit groups from “Add to Group”. 
Align both calls with 9999 (or real pagination).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/users/components/UserGroupMembershipsCard.tsx
```diff
@@ -13,56 +13,548 @@
...
+                    disabled={saving}
+                    onChange={integrationRole => onRoleChange({ integrationRole })}
+                />
+            </TableCell>
+            {deleteCell}
+        </TableRow>
+    );
+}
+
+function EnvironmentGroupMembershipsPanel({
```
aqibmohammadkhan, line 271: In `EnvironmentGroupMembershipsPanel`, `UserInheritedPermissionsSection` only renders after groups finish loading. Inherited hooks already have their own loading UI -> this causes unnecessary blanking on env switch and couples unrelated data.

aqibmohammadkhan (commented):
Minor

- Add success toast says “Roles successfully updated” instead of an add-specific message.
- No user-visible error state if groups/inherited queries fail (falls through to empty).
- Client pagination doesn’t clamp page when search shrinks results (can land on an empty page until “Next”/size change).
- Nav Users Icon change is unrelated noise in an otherwise focused PR.

pragsty:
> Minor
> 
> * Add success toast says “Roles successfully updated” instead of an add-specific message.
> * No user-visible error state if groups/inherited queries fail (falls through to empty).
> * Client pagination doesn’t clamp page when search shrinks results (can land on an empty page until “Next”/size change).
> * Nav Users Icon change is unrelated noise in an otherwise focused PR.

Fixed all
For icon change- Keeping the UsersIcon on the Users nav item, we need it

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/users/services/organizationUsers.ts
pragsty: Fixed. getOrganizationUserGroups and listEnvironmentGroups now default to 9999 via GROUP_MEMBERSHIP_FETCH_SIZE, aligned with classic (getUserGroups(..., 1, 9999)) and inherited resources (INHERITED_RESOURCES_FETCH_SIZE).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/users/components/UserGroupMembershipsCard.tsx
pragsty: Fixed. UserInheritedPermissionsSection now renders independently of group membership loading, so switching environments no longer blanks inherited permissions while groups refetch.

aqibmohammadkhan (approved):

PR #18972, opened 2026-08-05
feat(observability): read authz decisions through the shared logs endpoint

podlesrafal:
## Summary

Authorization decisions have been written to Elasticsearch for a while, but nothing could read them
back as documents. This serves them through the existing `POST /observability/logs/search`, so the
console reads decisions with the same `@gravitee/gamma-lib-observability` every other Gamma module
already uses. No new REST resource and no new signal: a decision is another record kind on the logs
signal, selected by one new filter.

## Why a new read is needed

Decisions are stored as `AuthzEventMetrics` in the `gravitee-event-metrics` data stream. Nothing read
documents from that index: `MetricsRepository` and `LogRepository` only touch `v4-metrics` /
`request` / `v4-log`, and the one `AnalyticsRepository` method that reaches event-metrics returns a
histogram. This adds the document-level query and plugs it in behind the endpoint that already
exists.

## Selecting decisions

A new cross-cutting `RECORD_TYPE` filter (`REQUEST` / `AUTHZ_DECISION`), not `API_TYPE`.

`API_TYPE` narrows **which APIs** are in scope. A decision is attached to the API it guarded — usually
a plain HTTP proxy — so scoping by it would silently drop every decision made in front of one. The
two filters compose: `RECORD_TYPE=AUTHZ_DECISION` + `API_TYPE=HTTP_PROXY` answers "decisions on my
HTTP proxies".

`doc-type: authz` is pinned in the query itself rather than derived from the filter. The data stream
is shared by all five `BaseEventMetrics` subtypes, so without it the result set would mix in api,
application, topic and operation events.

## Changes

**Repository**
- `MetricsRepository` gains `searchAuthzDecisionLogs`. `AuthzDecisionLogQuery` carries the api scope,
  time range and paging, and validates itself before a request is built — including an upper bound on
  page size (`MAX_SIZE = 1000`).
- Elasticsearch side: `SearchAuthzDecisionLogsQueryAdapter` builds the search and pins the doc type,
  `AuthzDecisionLogSourceMapper` and `SearchAuthzDecisionLogsResponseAdapter` read the flat
  `AuthzEventMetrics` source, `AuthzDecisionLogFields` names the fields in one place.
- The sort is `@timestamp desc, event-id asc`. A decision batch stamps every row with the same
  millisecond and tie order is not stable across shards, so paging on the timestamp alone repeats and
  skips rows.
- `NoOpMetricsRepository` gets the empty implementation.

**Core**
- `AuthzDecisionLogsCrudService` port over a core-owned `AuthzDecisionLog` record, returning
  `SearchLogsResponse` exactly as `ConnectionLogsCrudService` does. The infra implementation maps the
  repository record onto it: `io.gravitee.repository` is not on the `CoreRulesTest` allow-list for
  `io.gravitee.apim.core`.

**Gamma REST**
- `RECORD_TYPE` joins the extensible filter catalog on the `LOGS` signal only. Absent means `REQUEST`,
  so every existing caller keeps its behaviour.
- `ApiType` gains `AUTHZ`, and `toGammaApiType` stops mapping `AUTHZ -> null`. That mapping dropped
  the PDP's own API out of the accessible set, which made its decisions unreachable no matter what
  filter was sent.
- Default entrypoint injection is skipped when the record type is a decision. A decision document
  carries no entrypoint, so the injected terms would match nothing.
- The decision itself rides in a typed `authz` object on `LogEntry` / `LogEntryDto`, not in
  `additionalMetrics`. That map is the extension bag for keys plugins attach at runtime, so a known
  contract placed there would be invisible to the schema and readable only by convention — the same
  reason the native Kafka fields were hoisted out of it. All twenty fields are described in
  `openapi-observability.yaml`, which this PR previously did not touch at all.
- The decision search is scoped by api and time only, so any other condition is **refused** rather
  than accepted and dropped. A silently ignored filter is indistinguishable, to the caller, from one
  that matched everything. A multi-valued `RECORD_TYPE` is refused for the same reason instead of
  resolving to its first value.

## API surface

|---|---|---|---|---|
| `/environments/{envId}/observability/logs/search` | `POST` | Accepts the new `RECORD_TYPE` filter (`EQ` / `IN`, values `REQUEST` and `AUTHZ_DECISION`). Omitted, the request behaves exactly as before. | `{"filters":[{"name":"RECORD_TYPE","operator":"EQ","value":"AUTHZ_DECISION"}]}` | `200` — decision rows instead of request logs; previously there was no way to ask for them |
| `/environments/{envId}/observability/logs/search` | `POST` | `LogEntryDto` gains a typed `authz` object (20 fields, new `AuthzDecision` schema). `@JsonInclude(NON_NULL)` throughout, so request rows serialise byte for byte as before. | unchanged | `{ "authz": { "decision": "DENY", "subjectId": "alice", "batchIndex": 2, … } }` on decision rows only |
| `/environments/{envId}/observability/logs/search` | `POST` | **Refusal**: any condition other than `API` / `API_TYPE` sent with `RECORD_TYPE=AUTHZ_DECISION`, and any multi-valued `RECORD_TYPE`. | `{"filters":[{"name":"RECORD_TYPE","operator":"EQ","value":"AUTHZ_DECISION"},{"name":"REQUEST_ID","operator":"EQ","value":"req-1"}]}` | `400` — the decision search cannot apply it, so it is refused rather than ignored |
| `/environments/{envId}/observability/filters/definition` | `GET` | Catalog gains a `RECORD_TYPE` entry, advertised for `LOGS` only, with both enum values. `API_TYPE` gains `AUTHZ` among its values. | `?signal=LOGS` | one additional `FilterSpec`; existing entries unchanged apart from the added `API_TYPE` value |

No breaking change for request logs: `RECORD_TYPE` is a new optional filter that defaults to the
previous behaviour, `authz` is absent on request rows, and `AUTHZ` widens an existing enum rather
than altering it. The refusals only reach requests that could not have worked before — they were
accepted and then dropped.

**Companion change**: the console reads the new shape in
[gravitee-gamma-module-authz@2b61c71](https://github.com/gravitee-io/gravitee-gamma-module-authz/commit/2b61c71)
(branch `authz-decisions-logs`). The unpacking lives in one adapter, so no column changed.

## Test plan

- [ ] On *Observability → Logs*, open **Add filter**: a **Record Type** entry is offered, with
      **Request** and **Authz decision** as values.
- [ ] Search with no `RECORD_TYPE` filter: the result is request logs, unchanged from before this PR.
- [ ] Apply **Record Type = Authz decision**: decision rows come back, carrying decision, caller and
      operation. Request logs are gone from the result, not merely reordered.
- [ ] Check a batch: rows from one batch share a timestamp and each carries its own index. Page
      forward and back through a batch — no row repeats and none is skipped.
- [ ] Apply **Record Type = Authz decision** together with **API Type = HTTP Proxy**: the two compose,
      returning decisions taken in front of HTTP proxies only.
- [ ] The PDP's own API appears in the API scope selector. Before this change it was filtered out and
      its decisions were unreachable.
- [ ] On *Observability → Overview* (analytics signal), **Record Type** is not offered.
- [ ] Send `RECORD_TYPE=AUTHZ_DECISION` together with `REQUEST_ID` over the API: `400`, naming the
      filter that does not apply. Before this change the call returned every decision in scope.
- [ ] Regression: the tracing and dashboards screens still load and filter as before.

### Edge cases to check
- A `RECORD_TYPE` value that matches nothing must return zero rows rather than falling back to the
  unfiltered set — that is the difference between an applied filter and an ignored one.
- A decision row has no entrypoint, method, status or latency. The table must render it without those
  columns rather than showing empty or zeroed cells.

## Reviewer notes

**What changed after review.** The first version of this branch accepted conditions the decision
search cannot apply and dropped them on the floor: 21 of the 24 filters declared for the `LOGS`
signal reached the query object and were never read, so a request combining `RECORD_TYPE` with any
of them returned the unfiltered set with no error. Those are now refused, as is a multi-valued
`RECORD_TYPE` that used to resolve to its first value. An unknown `RECORD_TYPE` value was already a
400 before reaching this code — `ObservabilityFilterValidator.validateEnumValues` rejects any value
the catalog does not advertise, which is verified by the new use-case tests.

**`blockingGet` in the repository is not an oversight.** `MetricsRepository` is synchronous — it
returns `LogResponse<T>`, not `Single<LogResponse<T>>` — so the chain out of `client.search` has to be
collapsed in the implementation, which is what all nine reads in that package do. The management API
runs on Jetty (`JettyEmbeddedContainer` with `JettyQueuedThreadPool`), thread per request, so the cost
is one pooled thread held for one query rather than a stalled event loop. Making it reactive means
changing the SPI signature and both implementations, which belongs in its own change.

**Page size is bounded, deep paging is not.** `AuthzDecisionLogQuery.validate()` now refuses a size
above 1000. That guards the SPI contract rather than a reachable hole, since
`SearchObservabilityLogsUseCase` already clamps `perPage` to 100. The dimension that is genuinely
unbounded is `page`: `resolvePage` accepts any positive value and every adapter in this family builds
`from = (page - 1) * size`, so a large enough page number lands past Elasticsearch's
`index.max_result_window`. That predates this PR and applies equally to connection logs, message logs
and metrics — capping it belongs in a change across the whole family.

## Verification

**Live**, on a local stack: decisions from all three producers (PEP policy, AuthZEN policy, authz
reactor) render in the console's decisions table, including batch rows with their index and search
records.

**Automated**, run on this branch before pushing: the whole `gravitee-gamma-rest-api` module (276,
zero failures, prettier-java check included), of which `SearchObservabilityLogsUseCaseTest` (22,
three new for the refusals), `ObservabilityLogsDataPortAdapterTest` (40) and `LogsResourceTest` (10)
cover this path; plus `SearchAuthzDecisionLogsQueryAdapterTest` (9),
`SearchAuthzDecisionLogsResponseAdapterTest` (6), `AuthzDecisionLogsCrudServiceImplTest` (4) and the
ArchUnit `CoreRulesTest` in both rest-api-service and gamma-rest-api. Console side: 1,179 tests,
four of them new for the `authz` unpacking, which had no coverage before. Full suites run on CI.

The `authz` schema was checked by parsing the OpenAPI document rather than by eye: 20 properties,
matching the record field for field, referenced from `LogEntry` and absent from `LogDetail`, which
does not serve decisions.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/log/v4/model/authz/AuthzDecisionLogQuery.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/log/v4/model/authz/AuthzDecisionLogQuery.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.log.v4.model.authz;
+
+import java.util.Objects;
+import java.util.Set;
+import lombok.Builder;
+import lombok.Data;
+
+/**
+ * @author GraviteeSource Team
+ */
+@Data
+@Builder
+public class AuthzDecisionLogQuery {
+
+    public static final int MAX_SIZE = 1_000;
+
+    private Set<String> apiIds;
+    private Long from;
+    private Long to;
+
+    @Builder.Default
+    private int size = 20;
+
+    @Builder.Default
+    private int page = 1;
+
+    public void validate() {
+        Objects.requireNonNull(apiIds, "apiIds");
+        if (apiIds.isEmpty()) {
+            throw new IllegalArgumentException("apiIds must not be empty");
+        }
+        if (page < 1) {
+            throw new IllegalArgumentException("page must be >= 1, was " + page);
+        }
+        if (size < 1) {
+            throw new IllegalArgumentException("size must be >= 1, was " + size);
+        }
+        if (size > MAX_SIZE) {
+            throw new IllegalArgumentException("size must be <= " + MAX_SIZE + ", was " + size);
+        }
+        if (from != null && to != null && from > to) {
+            throw new IllegalArgumentException("from must be <= to, was from=" + from + " to=" + to);
+        }
+    }
+}
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/MetricsElasticsearchRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/MetricsElasticsearchRepository.java
@@ -24,6 +24,8 @@ import io.gravitee.repository.common.query.QueryContext;
 import io.gravitee.repository.elasticsearch.AbstractElasticsearchRepository;
 import io.gravitee.repository.elasticsearch.configuration.RepositoryConfiguration;
 import io.gravitee.repository.elasticsearch.utils.ClusterUtils;
+import io.gravitee.repository.elasticsearch.v4.log.adapter.authz.SearchAuthzDecisionLogsQueryAdapter;
+import io.gravitee.repository.elasticsearch.v4.log.adapter.authz.SearchAuthzDecisionLogsResponseAdapter;
 import io.gravitee.repository.elasticsearch.v4.log.adapter.connection.SearchConnectionLogErrorKeysQueryAdapter;
 import io.gravitee.repository.elasticsearch.v4.log.adapter.connection.SearchConnectionLogErrorKeysResponseAdapter;
 import io.gravitee.repository.elasticsearch.v4.log.adapter.connection.SearchMetricsQueryAdapter;
@@ -36,6 +38,8 @@ import io.gravitee.repository.elasticsearch.v4.log.adapter.nativeapi.NativeApiMe
 import io.gravitee.repository.elasticsearch.v4.log.adapter.nativeapi.NativeApiMetricsSearchResponseAdapter;
 import io.gravitee.repository.log.v4.api.MetricsRepository;
 import io.gravitee.repository.log.v4.model.LogResponse;
+import io.gravitee.repository.log.v4.model.authz.AuthzDecisionLog;
+import io.gravitee.repository.log.v4.model.authz.AuthzDecisionLogQuery;
 import io.gravitee.repository.log.v4.model.connection.*;
 import io.gravitee.repository.log.v4.model.message.MessageMetrics;
 import io.gravitee.repository.log.v4.model.message.MessageMetricsQuery;
@@ -137,6 +141,22 @@ public class MetricsElasticsearchRepository extends AbstractElasticsearchReposit
         } catch (RuntimeException e) {
             log.error("Failed to search native metrics [apiId={}]", query.getApiId(), e);
             throw new AnalyticsException("Failed to search native metrics for api " + query.getApiId(), e);
+        }
+    }
+
+    @Override
+    public LogResponse<AuthzDecisionLog> searchAuthzDecisionLogs(QueryContext queryContext, AuthzDecisionLogQuery query)
+        throws AnalyticsException {
+        query.validate();
+        var clusters = ClusterUtils.extractClusterIndexPrefixes(configuration);
+        var index = this.indexNameGenerator.getWildcardIndexName(queryContext.placeholder(), Type.EVENT_METRICS, clusters);
+
+        try {
+            return this.client.search(index, null, SearchAuthzDecisionLogsQueryAdapter.adapt(query))
+                .map(SearchAuthzDecisionLogsResponseAdapter::adapt)
+                .blockingGet();
+        } catch (RuntimeException e) {
+            throw new AnalyticsException("Failed to search authz decision logs for apis " + query.getApiIds(), e);
         }
     }
 }
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/adapter/authz/SearchAuthzDecisionLogsQueryAdapter.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/adapter/authz/SearchAuthzDecisionLogsQueryAdapter.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.elasticsearch.v4.log.adapter.authz;
+
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Aggs.SORT;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Keys.BOOL;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Keys.FILTER;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Keys.QUERY;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Keys.SIZE;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Keys.TIMESTAMP;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Keys.TRACK_TOTAL_HITS;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Query.GTE;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Query.LTE;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Query.RANGE;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Query.TERM;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Sort.DESC;
+import static io.gravitee.repository.elasticsearch.utils.ElasticsearchDsl.Tokens.TERMS;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.gravitee.repository.log.v4.model.authz.AuthzDecisionLogQuery;
+
+/**
+ * @author GraviteeSource Team
+ */
+public final class SearchAuthzDecisionLogsQueryAdapter {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+    private static final String FROM = "from";
+    private static final String ORDER = "order";
+    private static final String ASC = "asc";
+    private static final String UNMAPPED_TYPE = "unmapped_type";
+    private static final String KEYWORD = "keyword";
+
+    private SearchAuthzDecisionLogsQueryAdapter() {}
+
+    public static String adapt(AuthzDecisionLogQuery query) {
+        ArrayNode filters = MAPPER.createArrayNode();
+        // The event-metrics data stream is shared by every BaseEventMetrics subtype (api, application,
+        // topic, operation, authz), so this term is what makes the result set decisions and nothing else.
+        filters.add(MAPPER.createObjectNode().set(TERM, MAPPER.createObjectNode().put(AuthzDecisionLogFields.DOC_TYPE, "authz")));
+
+        ArrayNode apiIds = MAPPER.createArrayNode();
+        query.getApiIds().forEach(apiIds::add);
+        filters.add(MAPPER.createObjectNode().set(TERMS, MAPPER.createObjectNode().set(AuthzDecisionLogFields.API_ID, apiIds)));
+
+        if (query.getFrom() != null || query.getTo() != null) {
+            ObjectNode bounds = MAPPER.createObjectNode();
+            if (query.getFrom() != null) {
+                bounds.put(GTE, query.getFrom());
+            }
+            if (query.getTo() != null) {
+                bounds.put(LTE, query.getTo());
+            }
+            filters.add(MAPPER.createObjectNode().set(RANGE, MAPPER.createObjectNode().set(TIMESTAMP, bounds)));
+        }
+
+        ObjectNode root = MAPPER.createObjectNode();
+        root.set(QUERY, MAPPER.createObjectNode().set(BOOL, MAPPER.createObjectNode().set(FILTER, filters)));
+        root.put(FROM, (query.getPage() - 1) * query.getSize());
+        root.put(SIZE, query.getSize());
+        root.put(TRACK_TOTAL_HITS, true);
+        // event-id breaks ties: a batch stamps every decision with the same millisecond, and ordering
+        // within a tie is not stable across shards, so paging without it repeats and skips rows.
+        root.set(
+            SORT,
+            MAPPER.createArrayNode()
+                .add(MAPPER.createObjectNode().set(TIMESTAMP, MAPPER.createObjectNode().put(ORDER, DESC)))
+                .add(
+                    MAPPER.createObjectNode().set(
+                        AuthzDecisionLogFields.EVENT_ID,
+                        MAPPER.createObjectNode().put(ORDER, ASC).put(UNMAPPED_TYPE, KEYWORD)
+                    )
+                )
+        );
+
+        return root.toString();
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/crud_service/log/AuthzDecisionLogsCrudServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/crud_service/log/AuthzDecisionLogsCrudServiceImpl.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.infra.crud_service.log;
+
+import io.gravitee.apim.core.log.crud_service.AuthzDecisionLogsCrudService;
+import io.gravitee.apim.core.log.model.AuthzDecisionLog;
+import io.gravitee.repository.analytics.AnalyticsException;
+import io.gravitee.repository.common.query.QueryContext;
+import io.gravitee.repository.log.v4.api.MetricsRepository;
+import io.gravitee.repository.log.v4.model.authz.AuthzDecisionLogQuery;
+import io.gravitee.rest.api.model.common.Pageable;
+import io.gravitee.rest.api.model.v4.log.SearchLogsResponse;
+import io.gravitee.rest.api.service.common.ExecutionContext;
+import io.gravitee.rest.api.service.exceptions.TechnicalManagementException;
+import java.util.List;
+import java.util.Set;
+import lombok.CustomLog;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Component;
+
+/**
+ * @author GraviteeSource Team
+ */
+@Component
+@CustomLog
+class AuthzDecisionLogsCrudServiceImpl implements AuthzDecisionLogsCrudService {
+
+    private final MetricsRepository metricsRepository;
+
+    public AuthzDecisionLogsCrudServiceImpl(@Lazy MetricsRepository metricsRepository) {
+        this.metricsRepository = metricsRepository;
+    }
+
+    @Override
+    public SearchLogsResponse<AuthzDecisionLog> searchDecisionLogs(
+        ExecutionContext executionContext,
+        Set<String> apiIds,
+        Long from,
+        Long to,
+        Pageable pageable
+    ) {
+        if (apiIds == null || apiIds.isEmpty()) {
+            return new SearchLogsResponse<>(0, List.of());
+        }
+        try {
+            var response = metricsRepository.searchAuthzDecisionLogs(
+                new QueryContext(executionContext.getOrganizationId(), executionContext.getEnvironmentId()),
+                AuthzDecisionLogQuery.builder()
+                    .apiIds(apiIds)
+                    .from(from)
+                    .to(to)
+                    .page(pageable.getPageNumber())
+                    .size(pageable.getPageSize())
+                    .build()
+            );
+            return new SearchLogsResponse<>(
+                response.total(),
+                response.data().stream().map(AuthzDecisionLogsCrudServiceImpl::toDomain).toList()
+            );
+        } catch (AnalyticsException e) {
+            log.error("An error occurs while trying to search authz decision logs [apiIds={}]", apiIds, e);
+            throw new TechnicalManagementException("Unable to search authz decision logs", e);
+        }
+    }
+
+    private static AuthzDecisionLog toDomain(io.gravitee.repository.log.v4.model.authz.AuthzDecisionLog decision) {
+        return AuthzDecisionLog.builder()
+            .eventId(decision.eventId())
+            .timestamp(decision.timestamp())
+            .apiId(decision.apiId())
+            .organizationId(decision.organizationId())
+            .environmentId(decision.environmentId())
+            .gatewayId(decision.gatewayId())
+            .requestId(decision.requestId())
+            .operation(decision.operation())
+            .status(decision.status())
+            .caller(decision.caller())
+            .targetPdpId(decision.targetPdpId())
+            .policyGeneration(decision.policyGeneration())
+            .decision(decision.decision())
+            .matchedPolicyNames(decision.matchedPolicyNames())
+            .reasons(decision.reasons())
+            .subjectType(decision.subjectType())
+            .subjectId(decision.subjectId())
+            .action(decision.action())
+            .resourceType(decision.resourceType())
+            .resourceId(decision.resourceId())
+            .batchId(decision.batchId())
+            .batchIndex(decision.batchIndex())
+            .batchSize(decision.batchSize())
+            .searchType(decision.searchType())
+            .resultCount(decision.resultCount())
+            .durationNanos(decision.durationNanos())
+            .build();
+    }
+}
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/model/RecordType.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/model/RecordType.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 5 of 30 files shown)


gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/log/v4/model/authz/AuthzDecisionLogQuery.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public void validate() {
+        Objects.requireNonNull(apiIds, "apiIds");
+        if (apiIds.isEmpty()) {
+            throw new IllegalArgumentException("apiIds must not be empty");
+        }
+        if (page < 1) {
+            throw new IllegalArgumentException("page must be >= 1, was " + page);
+        }
+        if (size < 1) {
```
sclarkamlab, line 50: do we need a max size limit?

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/MetricsElasticsearchRepository.java
```diff
@@ -139,4 +143,21 @@ public LogResponse<NativeApiMetrics> searchNativeApiMetrics(QueryContext queryCo
...
+    public LogResponse<AuthzDecisionLog> searchAuthzDecisionLogs(QueryContext queryContext, AuthzDecisionLogQuery query)
+        throws AnalyticsException {
+        query.validate();
+        var clusters = ClusterUtils.extractClusterIndexPrefixes(configuration);
+        var index = this.indexNameGenerator.getWildcardIndexName(queryContext.placeholder(), Type.EVENT_METRICS, clusters);
+
+        try {
+            return this.client.search(index, null, SearchAuthzDecisionLogsQueryAdapter.adapt(query))
+                .map(SearchAuthzDecisionLogsResponseAdapter::adapt)
+                .blockingGet();
```
sclarkamlab, line 157: does this NEED to be blockingGet, won't this have an effect on the performance?

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
```diff
@@ -389,6 +398,60 @@ private SearchLogsFilters buildSearchFilters(LogsSearchQuery query) {
         return builder.build();
     }
 
+    private LogsPage searchDecisionLogs(ExecutionContext executionContext, LogsSearchQuery query, PageableImpl pageable) {
+        var result = authzDecisionLogsCrudService.searchDecisionLogs(executionContext, query.apiIds(), query.from(), query.to(), pageable);
```
SamHardi: **Silent filter drop — this is my main concern on the PR.** Only `apiIds`, `from`, `to` and the pagination reach the decision query; `query.conditions()` is never read in this path, even though `SearchObservabilityLogsUseCase` still computes `effectiveConditions` and carries them into the query.

Concretely: a caller combining `RECORD_TYPE=AUTHZ_DECISION` with any other declared logs filter (`REQUEST_ID`, `GATEWAY`, `ENTRYPOINT`, …) gets **unfiltered** results, with no error. That is the worst of the three possible behaviours — the condition is not applied, not rejected, and not documented.

Suggestion: at minimum, reject unsupported conditions with a `ValidationDomainException` when `recordType == AUTHZ_DECISION` (the pipeline already fails fast on other invalid combinations), and ideally support the ones that map naturally onto decision fields (`REQUEST_ID` → `request-id`, `GATEWAY` → `gw-id`).

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/model/RecordType.java
```diff
@@ -0,0 +1,50 @@
...
+    RecordType(String label) {
+        this.label = label;
+    }
+
+    /** Human-readable display label for this record kind. */
+    public String label() {
+        return label;
+    }
+
+    public static RecordType fromNameOrDefault(String value) {
```
SamHardi, line 40: `RECORD_TYPE=TYPO` silently falls back to `REQUEST`: the user believes they are looking at decisions and actually gets request logs. Absent/blank → `REQUEST` is the right default for backward compatibility, but an **unknown** value should be a validation error rather than a silent default — same philosophy as the filter-drop comment on the adapter.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/use_case/SearchObservabilityLogsUseCase.java
```diff
@@ -141,6 +156,20 @@ private static void validateTimeRange(Instant from, Instant to) {
         }
     }
 
+    /**
+     * Which document contract the caller wants. Absent means request logs, so every existing consumer
+     * keeps its behaviour untouched.
+     */
+    private static RecordType extractRecordType(List<FilterCondition> conditions) {
```
SamHardi, line 167: `findFirst()` silently truncates a multi-value condition: `RECORD_TYPE IN (REQUEST, AUTHZ_DECISION)` semantically means "both kinds", but is treated as `REQUEST` only. Until mixed record types are actually supported, a multi-value `RECORD_TYPE` condition should probably be rejected explicitly.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/adapter/authz/SearchAuthzDecisionLogsQueryAdapter.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+            if (query.getTo() != null) {
+                bounds.put(LTE, query.getTo());
+            }
+            filters.add(MAPPER.createObjectNode().set(RANGE, MAPPER.createObjectNode().set(TIMESTAMP, bounds)));
+        }
+
+        ObjectNode root = MAPPER.createObjectNode();
+        root.set(QUERY, MAPPER.createObjectNode().set(BOOL, MAPPER.createObjectNode().set(FILTER, filters)));
+        root.put(FROM, (query.getPage() - 1) * query.getSize());
```
SamHardi, line 74: No cap on `from`: past `index.max_result_window` (10 000 by default) Elasticsearch rejects the request with a raw error that will surface as a 500. If the other log reads share this behaviour it is a pre-existing gap and fine to leave as-is here — but worth either capping the page depth or translating the error into something actionable.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/model/LogEntry.java
```diff
@@ -65,5 +65,9 @@ public record LogEntry(
     FailureOrigin failureOrigin,
     String clientId,
     String brokerId,
-    Long connectionDurationMs
+    Long connectionDurationMs,
+    // Authz decision fields, hoisted from the decision document; the rest rides in additionalMetrics.
+    String decision,
```
SamHardi: Design concern: the real decision contract is mostly **untyped**. Three fields are hoisted here, and the remaining 17 (`subjectId`, `resourceType`, `batchIndex`, `durationNanos`, …) ride in `additionalMetrics` as a `Map<String, Object>` with magic keys the frontend must know by convention — invisible in the OpenAPI contract, and every future PDP field will pile into the same bag.

Consider a typed nested `authz { … }` object on `LogEntry`/`LogEntryDto` instead: the frontend reuse argument is preserved, but the contract becomes explicit and can evolve safely.

SamHardi (commented):
## Review summary

Overall the approach — multiplexing decisions through the shared logs endpoint behind a `RECORD_TYPE` filter — looks like a reasonable trade-off: it reuses the whole `gamma-lib-observability` pipeline (table, pagination, time range, filters) at zero frontend cost, and the `RECORD_TYPE` ⊥ `API_TYPE` reasoning in the description is convincing. The PR size is mostly the layering tax of a new document type (repository interface → ES impl → noop → core service → infra impl → Gamma port) plus ~55% tests, not accidental complexity.

Strong points worth calling out:

- `event-id` tie-break on the sort, so batch decisions sharing a millisecond page stably
- `doc-type: authz` pinned in the query itself rather than derived from the filter
- refusing to query the index at all when no API is in scope
- edge-case test coverage (absent fields, unreadable timestamp, unnamed matched policy)

Main concerns, in order of importance (details inline):

1. **In the decision path, every other filter condition is silently dropped** — neither applied nor rejected.
2. An invalid `RECORD_TYPE` value silently falls back to `REQUEST`.
3. The decision contract is mostly untyped: 17 magic keys in `additionalMetrics`, invisible in the API contract.
4. No guard on deep pagination (`max_result_window`).

## Product question before settling the design

If the product roadmap eventually calls for a real **"Decisions" page** with its own columns and filters (subject, resource, action, decision, matched policies…), it would be cheaper to fork **now** to a dedicated resource (e.g. `/observability/authz-decisions`) with a typed DTO, rather than growing the union `LogEntry` further — every future PDP field will otherwise land in the untyped bag or widen the shared DTO. If instead decisions are meant to stay a "record type" lens inside the shared logs table long-term, the current multiplexing is the right call — but then concerns 1–3 need addressing so the shared contract stays honest. Do we know which way the product is leaning?

gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/log/v4/model/authz/AuthzDecisionLogQuery.java
podlesrafal: Added one: `MAX_SIZE = 1000`, enforced in `validate()`.

Worth being precise about what it does and does not cover, though. `size` was already bounded on the only production path: `SearchObservabilityLogsUseCase.resolvePerPage` clamps `perPage` to `MAX_PER_PAGE = 100`. So the new check keeps the SPI contract honest for the next caller rather than closing a reachable hole. 1000 sits well under elasticsearch's `index.max_result_window` (10 000 by default) and well above anything the REST layer can ask for, so in practice it only ever catches a programming error.

The dimension that is genuinely unbounded is `page`, not `size`. `resolvePage` accepts any positive value, and the adapter computes `from = (page - 1) * size`, so a large enough page number produces a `from` past `max_result_window` and elasticsearch rejects the search outright. That is not specific to this query: `SearchMetricsQueryAdapter`, `SearchConnectionLogDetailQueryAdapter`, `SearchMessageMetricsQueryAdapter` and `SearchMessageLogQueryAdapter` all build `from` the same way, and connection logs reach them through the same use case with the same unbounded `page`.

Capping deep paging therefore looks like it belongs in a separate change across that whole family rather than in this one query. Happy to open a ticket for it if you agree.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/MetricsElasticsearchRepository.java
podlesrafal: It does, given the current contract. `MetricsRepository` is synchronous: it returns `LogResponse<T>` and declares `throws AnalyticsException`, not `Single<LogResponse<T>>`. The reactive chain coming out of `client.search` has to be collapsed somewhere in the implementation, and every other read in this package collapses it the same way, nine `blockingGet()` call sites across `MetricsElasticsearchRepository` and `LogElasticsearchRepository`, including `searchNativeMetrics` fifteen lines above this one.

On the performance question specifically: this is the management API, which runs on Jetty (`JettyEmbeddedContainer` with `JettyQueuedThreadPool`), a thread-per-request container, not a Vert.x event loop like the gateway. Blocking the worker for the duration of the search is the model that container is built around, so the cost is one pooled thread held for one ES query rather than a stalled event loop.

Making it genuinely reactive would mean changing the `MetricsRepository` signature plus both implementations (elasticsearch and noop) and the call chain up to the REST resource. Reasonable to want, but bigger than this PR, so I would rather not do it as a drive-by here.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/crud_service/log/AuthzDecisionLogsCrudServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+                    .to(to)
+                    .page(pageable.getPageNumber())
+                    .size(pageable.getPageSize())
+                    .build()
+            );
+            return new SearchLogsResponse<>(
+                response.total(),
+                response.data().stream().map(AuthzDecisionLogsCrudServiceImpl::toDomain).toList()
+            );
+        } catch (AnalyticsException e) {
```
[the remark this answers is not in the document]
podlesrafal: Keeping this one as is. Log-and-wrap at this layer is the established convention in this package: `ConnectionLogsCrudServiceImpl` does exactly this in all four of its catch blocks, down to the shape of the message. Changing it here alone would make this class the odd one out without making anything better.

The legitimate half of the concern was that the same failure was being logged twice, here and again in the repository. That is fixed on the repository side, so a failed search now produces a single error line.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
```diff
@@ -389,6 +398,60 @@ private SearchLogsFilters buildSearchFilters(LogsSearchQuery query) {
...
+        var entries = result
+            .data()
+            .stream()
+            .map(decision -> mapDecisionToLogEntry(decision, query.apisById()))
+            .toList();
+        return new LogsPage(entries, result.total());
+    }
+
+    private LogEntry mapDecisionToLogEntry(AuthzDecisionLog decision, Map<String, ApiReference> apisById) {
```
podlesrafal, line 412: Not taking this one. The actual overlap between `mapDecisionToLogEntry` and `mapToLogEntry` is five fields (`apiId`, `apiName`, `apiType`, `requestId`, `gateway`) plus a timestamp that is derived differently on each side: `Instant.ofEpochMilli(Long)` here against `parseTimestamp(String)` there. `mapToLogEntry` sets 31 builder properties to this method's 10, and the two inputs are unrelated types from different modules with no common supertype, a record with accessors and a bean with getters.

Extracting a shared helper would mean either introducing an interface over those two types or threading six scalars through a parameter list. That is more machinery than the handful of duplicated lines costs, and it would couple the decision mapping to the connection-log mapping so that changes to one have to be reasoned about against the other.

I did fix the one real inconsistency the diff introduced: the api lookup now uses the same null-guard idiom and the same variable name as `mapToLogEntry`.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
podlesrafal: You were right, and it was worse than the comment suggests. I measured it against the real catalog rather than by eye: 24 filters carry the `LOGS` signal, and after `removeScopeConditions` strips `API`, `API_TYPE` and `RECORD_TYPE`, the remaining **21 reached the query object and were never read** — `REQUEST_ID`, `TRANSACTION_ID`, `ENTRYPOINT`, `ERROR_KEY`, `PAYLOAD`, the HTTP ones, the LLM and MCP ones.

Worth adding for the record: `ObservabilityFilterValidator` checks name, signal, operator and enum values, but not `apiTypes`. So a filter declared only for `HTTP_PROXY` passed validation on a decision search too — the drop was not gated by anything.

Fixed by taking your minimum: any condition surviving scope removal is now refused with a `ValidationDomainException` naming it, so a filter can no longer be accepted and ignored.

I deliberately did not take the second half yet. Mapping `REQUEST_ID` → `request-id` and `GATEWAY` → `gw-id` means widening `AuthzDecisionLogQuery` and the ES adapter, which is new capability rather than closing a hole, and it would land untested in a PR already carrying a contract change. Happy to open it as a follow-up — those two do map cleanly onto decision fields.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/model/RecordType.java
podlesrafal: This one does not reproduce — `RECORD_TYPE=TYPO` is already a 400, and never reaches `fromNameOrDefault`.

`ObservabilityFilterValidator.validateEnumValues` refuses any value an `ENUM` spec does not advertise, and its javadoc gives exactly your reasoning: it exists so that "a typo [does not turn] into an unfiltered result set instead of a 400". `RECORD_TYPE` is `FilterType.ENUM` with both values advertised, and `filterValidator.validate(conditions, Signal.LOGS)` runs at the top of `execute`, before the record type is read.

I checked this by wiring the real catalog into the real validator and asserting the throw, rather than by reading the code, because the same fallback did look suspicious to me.

So the fallback is unreachable for anything arriving over HTTP; it is defence in depth for internal callers constructing conditions directly. I have left it as is — swapping it for a throw would change no observable behaviour and would remove the safety net on the non-HTTP path. If you would rather see it fail loudly there too, say so and I will change it.

Your sibling comments on the adapter and on `findFirst` were both real, and both are fixed.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/use_case/SearchObservabilityLogsUseCase.java
podlesrafal: Agreed and fixed. `RECORD_TYPE` with more than one distinct value is now refused rather than truncated:

> Filter 'RECORD_TYPE' accepts a single value, was [REQUEST, AUTHZ_DECISION]. Search one record kind at a time.

Your framing is the right one — until mixed record kinds are actually supported, silently picking one is the worst of the options. And supporting both is not a wider filter but a second search: the two kinds live in different indices with different shapes and different sort keys, so a union would need merging two result sets with a shared paging cursor.

Covered by a test asserting the refusal and that the data port is never reached.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/adapter/authz/SearchAuthzDecisionLogsQueryAdapter.java
podlesrafal: They do share it, so taking your "fine to leave as-is here". `SearchMetricsQueryAdapter`, `SearchConnectionLogDetailQueryAdapter`, `SearchMessageMetricsQueryAdapter` and `SearchMessageLogQueryAdapter` all compute `from` as `(page - 1) * size` with no ceiling, and connection logs reach them through the same use case, whose `resolvePage` accepts any positive value. So the depth is unbounded across the whole family, not just here.

The size dimension is bounded as of this branch: `AuthzDecisionLogQuery.validate()` refuses a size above 1000, which came out of @sclarkamlab's thread on the query object. That guards the SPI contract rather than a reachable path, since the REST layer already clamps `perPage` to 100.

Capping page depth, or translating the `result window is too large` error into something actionable, is worth doing but belongs in a change that covers all five adapters at once — doing it here alone would leave the other four returning raw 500s. I will open a ticket for it unless you would rather it rode along here.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/model/LogEntry.java
podlesrafal: Adopted. The decision now rides in a typed `authz` object on `LogEntry` / `LogEntryDto`, and `additionalMetrics` is untouched on decision rows:

```json
{ "apiId": "…", "timestampEpochMs": 1754564000000, "requestId": "…",
  "authz": { "eventId": "…", "decision": "DENY", "subjectId": "alice",
             "resourceType": "McpTool", "batchIndex": 2, "durationNanos": 431000, … } }
```

Two things reinforced your point while I was checking it.

First, the contract framing: `additionalMetrics` is declared in the spec as `additionalProperties: true` with the description "Extra metrics attached by plugins or custom policies". It is explicitly the extension bag for what plugins attach at runtime, so putting a fixed, known contract in it was a category error, not just an undocumented choice.

Second, this repo already reached your conclusion once. The native Kafka fields arrived the same way and were hoisted into typed `LogEntry` fields — and documented, `connectionStatus` has its own enum in the spec plus a note saying it is hoisted from `additional-metrics`. So this is finishing a move the model had already made rather than introducing a new pattern.

While doing it I found the branch was not touching `openapi-observability.yaml` at all, so the three top-level fields were undocumented too. There is now an `AuthzDecision` schema with all twenty fields, referenced from `LogEntry` and deliberately not from `LogDetail`, which does not serve decisions.

Console side: the unpacking was already isolated in one adapter (`authz-decisions-source.ts`), exactly as its own comment intended, so this cost one file and no column changes — [gravitee-gamma-module-authz@2b61c71](https://github.com/gravitee-io/gravitee-gamma-module-authz/commit/2b61c71). That mapping had no test; it has four now.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
SamHardi: Verified on the updated branch: `rejectConditionsTheDecisionSearchCannotApply` now refuses every condition the decision search cannot honour with a clear 400, the behaviour is covered by `should_refuse_a_condition_the_decision_search_cannot_apply`, and the OpenAPI description states the contract explicitly ("rejected with 400 rather than accepted and ignored"). Thanks for going further than the comment and auditing the full filter catalog — the `apiTypes` gap you found in the validator is a good catch for a follow-up. Resolving.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/model/RecordType.java
SamHardi: Conceded — you're right, this doesn't reproduce. I checked `ObservabilityFilterValidator.validateEnumValues`: it throws `unknownEnumValue` for any value the ENUM spec doesn't advertise, and `filterValidator.validate(conditions, Signal.LOGS)` runs before the record type is read, so `RECORD_TYPE=TYPO` is a 400 before `fromNameOrDefault` is ever reached. The fallback branch is defensive dead code on the REST path, which is fine. Resolving.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/use_case/SearchObservabilityLogsUseCase.java
SamHardi: Verified: a multi-valued `RECORD_TYPE` is now refused with an actionable message instead of truncated, covered by `should_refuse_a_multi_valued_record_type_rather_than_truncating_it`, and the "second search, not a wider filter" rationale is now documented both in the javadoc and the OpenAPI description. Resolving.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/adapter/authz/SearchAuthzDecisionLogsQueryAdapter.java
SamHardi: Agreed — since the unbounded depth is shared by the whole family of log reads (`SearchMetricsQueryAdapter`, `SearchConnectionLogDetailQueryAdapter`, message adapters…), fixing it only here would be inconsistent. The `MAX_SIZE = 1000` guard on the size dimension is a good addition. Let's track the page-depth ceiling as a separate cross-cutting ticket. Resolving.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/model/LogEntry.java
SamHardi: Verified: the typed `authz` object now flows through `AuthzDecision` → `AuthzDecisionDto` → the `AuthzDecision` schema in the OpenAPI spec, with `additionalMetrics` left to its documented role as the runtime extension bag. The schema descriptions (decision vs status, batch fields) are a nice bonus. Exactly what I was hoping for — resolving.

SamHardi (approved):
## Second pass — feedback verified on the updated branch

All five threads from my first review are settled, each verified against the code rather than the replies alone:

1. **Silent filter drop** — fixed. `rejectConditionsTheDecisionSearchCannotApply` turns every condition the decision search cannot honour into an explicit 400, covered by a dedicated test, and the OpenAPI description now states the contract ("rejected with 400 rather than accepted and ignored"). The `apiTypes` validation gap found along the way is worth its own follow-up ticket.
2. **Unknown `RECORD_TYPE` value** — my comment did not reproduce; conceded. `ObservabilityFilterValidator.validateEnumValues` already 400s any unadvertised enum value before the record type is read.
3. **Multi-valued `RECORD_TYPE`** — fixed: refused with an actionable message instead of silently truncated, with test and spec coverage.
4. **Deep pagination** — agreed to leave as-is here since the unbounded depth is shared by the whole family of log reads; the new `MAX_SIZE = 1000` guard closes the size dimension. Page-depth ceiling to be tracked as a cross-cutting ticket.
5. **Untyped decision contract** — adopted: the typed `authz` object now flows `AuthzDecision` → `AuthzDecisionDto` → the OpenAPI `AuthzDecision` schema, and `additionalMetrics` keeps its documented role as the runtime extension bag.

The two bot findings the author declined (mapper duplication, log-and-wrap) were declined with reasoning I agree with — the duplication is five fields across unrelated types, and log-and-wrap matches the package convention with the double-logging half actually fixed.

**Still open, non-blocking:** the product question from my first review — if a dedicated "Decisions" page with its own columns/filters is on the roadmap, a dedicated resource would be the cheaper long-term shape. Merging this does not foreclose that option: the repository/service stack is reusable as-is and the typed `authz` contract ports cleanly to a dedicated DTO. Worth an answer before the next lot builds on the multiplexed endpoint, though.

Approving — nice turnaround on the feedback, the OpenAPI documentation of the failure modes is above the bar.

PR #18981, opened 2026-08-05
fix(rest-api): load default portal templates via defining classloader

dev-marek:
**Issue**

https://gravitee.atlassian.net/browse/APIM-14880

**Purpose**

Fix default portal navigation items (including the homepage) and the default subscription form failing to seed for Cockpit-created environments due to a classloader bug, and repair environments already stuck in that state.

**Summary of changes**

- Replace `Thread.currentThread().getContextClassLoader()` with `<Class>.class.getClassLoader()` in `CreateDefaultPortalNavigationItemsUseCase`, `SeedDefaultPagesForApiNavigationItemsUseCase`, and `CreateDefaultSubscriptionFormUseCase` (plus a test helper with the same pattern) — the previous lookup could return null when the calling thread's context classloader traced back to the isolated Cockpit WS connector plugin instead of the main app classpath, aborting seeding partway through.
- Make `CreateDefaultPortalNavigationItemsUseCase.execute()` idempotent per item, so it only creates whatever default nav items are still missing.
- Add a guard in `EnvironmentsDefaultPortalNavigationItemsUpgrader`: only repair an environment if it has no Home Page at all — an environment with a homepage was fully seeded at some point, so it's left untouched even if a user later deleted something like the "Guides" folder on purpose.
- Bump `EnvironmentsDefaultPortalNavigationItemsUpgrader` to `v2` and `DefaultSubscriptionFormUpgrader` to `v3` so both re-run once on next boot and repair any environment left stuck by this bug (including subscription forms, since the original handler failure would have skipped that call too).

**Out of scope**

- No change to `EnvironmentCommandHandler`'s new-vs-existing gating logic — the upgrader-based repair is a general safety net for this class of partial-seeding failure regardless of cause.
- No dedicated data-migration script for this specific customer's environment — the general repair upgrader covers it automatically on next boot.

**Testing**

- Unit tests added/updated across `CreateDefaultPortalNavigationItemsUseCaseTest`, `SeedDefaultPagesForApiNavigationItemsUseCaseTest`, `CreateDefaultSubscriptionFormUseCaseTest`, `EnvironmentsDefaultPortalNavigationItemsUpgraderTest` — covering classloader independence, idempotency, partial-state repair, and the new homepage-presence guard. All pass.
- Manual end-to-end verification: built the fixed image locally, connected a real running instance to a protocol-faithful mock Cockpit controller (`gravitee-cockpit-exchange-tester`), and confirmed (a) a brand-new environment seeds fully, (b) an environment simulating the exact reported broken state (only an empty "Guides" folder) gets fully repaired on restart via the version-bumped upgraders, and (c) a fully-seeded environment with a deliberately-deleted "Guides" folder is left untouched.

**Additional context**

Root cause analysis is documented as a comment on the ticket.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/CreateDefaultPortalNavigationItemsUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/CreateDefaultPortalNavigationItemsUseCase.java
@@ -70,14 +70,14 @@ public class CreateDefaultPortalNavigationItemsUseCase {
             PortalArea.TOP_NAVBAR
         );
 
-        var folderGuides = findByTitle(topLevelTopNavbarItems, "Guides");
+        var folderGuides = findByTitleAndType(topLevelTopNavbarItems, "Guides", PortalNavigationItemType.FOLDER);
         if (folderGuides == null) {
             folderGuides = createPortalFolder("Guides", organizationId, environmentId, PortalArea.TOP_NAVBAR, null);
         }
 
         final var guidesChildren = portalNavigationItemsQueryService.findByParentIdAndEnvironmentId(environmentId, folderGuides.getId());
 
-        if (findByTitle(guidesChildren, "Getting started") == null) {
+        if (findByTitleAndType(guidesChildren, "Getting started", PortalNavigationItemType.PAGE) == null) {
             final var contentGettingStarted = createPortalPageContent(organizationId, environmentId, GETTING_STARTED_PATH);
             createPortalPage(
                 "Getting started",
@@ -89,7 +89,7 @@ public class CreateDefaultPortalNavigationItemsUseCase {
             );
         }
 
-        var folderCoreConcepts = findByTitle(guidesChildren, "Core concepts");
+        var folderCoreConcepts = findByTitleAndType(guidesChildren, "Core concepts", PortalNavigationItemType.FOLDER);
         if (folderCoreConcepts == null) {
             folderCoreConcepts = createPortalFolder(
                 "Core concepts",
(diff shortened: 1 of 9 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/CreateDefaultPortalNavigationItemsUseCase.java
```diff
@@ -56,52 +59,96 @@ public class CreateDefaultPortalNavigationItemsUseCase {
...
         );
 
-        final var folderCoreConcepts = createPortalFolder(
-            "Core concepts",
-            organizationId,
-            environmentId,
-            PortalArea.TOP_NAVBAR,
-            folderGuides.getId()
-        );
+        var folderGuides = findByTitle(topLevelTopNavbarItems, "Guides");
```
jaroslawlakomy-grav: Include the expected type when identifying default items

`findByTitle` treats any top-level item named "Guides" as the default folder. If an environment without a homepage already contains a user-created PAGE or LINK with that title, the next child creation uses its id as the parent. `PortalNavigationItemDomainService.resolveParentContainer` then rejects it because it is not a container, causing the upgrader to fail instead of repairing the environment.

Please include the expected `PortalNavigationItemType` in the lookup and cover title/type collisions. The same identity problem applies to "Core concepts", "Getting started", and "Docs".

jaroslawlakomy-grav (approved):
Verification by title is basically weak but I don't think that we have better options here. Please consider checking also PortalNavigationItem type property

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/CreateDefaultPortalNavigationItemsUseCase.java
dev-marek: Good catch — fixed in 343eb3c: `findByTitle` is now `findByTitleAndType`, matching on both title and `PortalNavigationItemType` for all six lookups (Guides/Core concepts as FOLDER, Getting started/Authentication/Making your first API call as PAGE, Docs as LINK). Added a regression test (`should_not_mistake_a_title_colliding_item_of_a_different_type_for_the_default_one`) that plants a user-created PAGE titled "Guides" and asserts the repair still creates a proper FOLDER and parents "Getting started" under it correctly.

PR #18991, opened 2026-08-05
fix(analytics): stop dropping requests from v4 API traffic counts

SamHardi:
## Summary

[APIM-14859](https://gravitee.atlassian.net/browse/APIM-14859)

On a v4 MESSAGE API's *API Traffic* page, the Overview "Total Requests" tile could display a value **lower** than a single entrypoint card for the same API and time range. On the reported customer data the Overview showed **164 774** while the HTTP POST card showed **230 789** — a subtotal larger than its total.

The root cause turned out to sit in a family of counting queries that all silently drop requests. This PR fixes the three of them, in **three commits, one per defect**, so any can be dropped or cherry-picked independently. Please keep them separate on merge.

| # | Commit | Defect |
|---|---|---|
| 1 | `34ea4d8` | The total counted fewer requests than one of its own entrypoints |
| 2 | `c1b175a` | The status breakdown left out requests with no response status |
| 3 | `63b7ee4` | The `type=COUNT` total was hard-capped at 10 000 |

---

## 1. The total counted fewer requests than an entrypoint

Both figures come from one endpoint and one Elasticsearch response (`GET .../apis/{apiId}/analytics/requests-count` → `RequestsCount { total, countsByEntrypoint }`), but they were computed over **different document sets**:

| Value | Aggregation | Documents counted |
|---|---|---|
| `countsByEntrypoint` | `terms` on `entrypoint-id` | **all** matching documents |
| `total` | `range` on `status`, `[100.0, 600.0)` | only documents with `status` in 100..599 |

`io.gravitee.reporter.api.v4.metric.Metrics#getStatus` is a **primitive `int`** (default `0`) and the reporter template serialises it unconditionally. So every request whose response status was never committed — client abort, connection reset, a message connection closed by the client — reaches Elasticsearch as `"status": 0`, lands in its entrypoint bucket, and was dropped from the total. The gap grows with the share of such requests, which is why this only surfaced under sustained load.

The same flaw is **deterministic, at any volume, for metrics carrying no `status` field at all**. Native Kafka connection metrics are that case: `total` came back as `0` while the breakdown held the real count. This half reaches API consumers rather than a screen — the NATIVE analytics page is built entirely from `HISTOGRAM` widgets and has no "Total Requests" tile.

**How it got there.** `925a11b46c` replaced the previous total — the sum of the entrypoint bucket counts — with this status-range count, so that requests carrying **no** `entrypoint-id` would be counted too. A plain revert would reintroduce that bug, so the total now counts *all* matching documents: entrypoint or not, status or not.

**The fix.**
- `SearchRequestsCountQueryAdapter` — add `track_total_hits: true`, drop the `all_apis_status_ranges` range aggregation.
- `SearchRequestsCountResponseAdapter` — derive the total from `hits.total.value`, making it a superset of every bucket by construction. Falls back to the sum of the entrypoint buckets when the response carries no hits, so the invariant survives even a malformed response.
- Explicit `size` on the **four per-entrypoint `terms` aggregations** feeding the page (`SearchRequestsCountQueryAdapter`, `SearchResponseStatusRangesAdapter`, `SearchAverageConnectionDurationQueryAdapter`, `SearchAverageMessagesPerRequestQueryAdapter`). All four relied on the Elasticsearch default of **10 buckets**; the page merges their four maps into one card list, so they could truncate at different points and produce cards with partially missing data. New shared constant `ElasticsearchDsl.Limits.ENTRYPOINT_BUCKETS`.

<details>
<summary>Why <code>track_total_hits</code> rather than an aggregation-based total</summary>

`io.gravitee.elasticsearch.model.Aggregation` exposes `value` and `count` as **`java.lang.Float`**, which silently rounds above 2^24 ≈ 16.7 M — a `value_count`-based total would have introduced a fresh precision bug at exactly the scale that produced this report. `TotalHits.value` is a `long`.

`track_total_hits` is supported across the whole tested matrix (ES 7.17 / 8.17 / 9, OpenSearch 1.3 / 2) and costs nothing here: the `entrypoints` terms aggregation already forces Elasticsearch to visit every matching document, so hit-count early termination was never in play.
</details>

## 2. The status breakdown left out requests with no response status

Raising the total in commit 1 would otherwise have left a second, milder mismatch on the same page. `SearchResponseStatusRangesAdapter` classifies into ranges spanning `[100, 600)` only, so requests with `status: 0` — and metrics with no `status` at all — fell into no bucket. The breakdown would have read ~164 774 next to a total of ~230 789. Before commit 1 the two agreed, because *both* were wrong the same way.

The adapter now derives an explicit `unknown` bucket, per entrypoint and globally, as **the requests the query matched minus the ones the ranges hold**. The console renders it as a neutral **No status** slice; zero-valued slices are already filtered out, so nothing appears for APIs whose traffic is fully classified.

Strictly, the bucket means *outside the ranges*, which also covers a status below 100 or at 600 and above — that takes an upstream returning something that is not a valid HTTP status, so in practice it is a request whose status was never committed. The distinction is spelled out in the javadoc; the user-facing label stays "No status" because it describes what a reader will actually be looking at.

`unknown` is now documented as a reserved key on both the API-level and environment-level `ranges` schemas, since it is part of the response contract.

## 3. The `type=COUNT` total was capped at 10 000

`SearchRequestsCountByEventQueryAdapter` reads its total from `hits.total` on a query that carries **no aggregation**, so Elasticsearch applied its default `track_total_hits` of 10 000: any API busier than that reported a flat 10 000. `TotalHits` drops the `relation` field, so nothing downstream could tell a real 10 000 from a truncated one.

It now asks for the exact count. Unlike the other count queries this one has no aggregation forcing a full collection, so it **does** trade hit-count early termination for correctness — a wrong number is worse than a slower one here. It also no longer dereferences `hits` blindly.

---

## Test plan

- [ ] `mvn -pl gravitee-apim-repository/gravitee-apim-repository-elasticsearch test` — **726 tests, 0 failures** locally.
- [ ] The new regression tests were run **with the production change reverted**, to confirm they are real guards rather than tests that never fail:

| Test | Old code | Fixed |
|---|---|---|
| `should_count_requests_whose_response_status_was_never_committed` | `1` ❌ | `2` ✅ |
| `should_count_native_kafka_metrics_indexed_without_any_status` | `0` ❌ | `4` ✅ |
| `should_never_return_a_total_lower_than_a_single_entrypoint_count` | violated on 2 APIs ❌ | ✅ |

- [ ] Pre-existing expectations are **unchanged** — `total == 13` for the historical fixture API (whose breakdown sums to 12, the thirteenth document having no `entrypoint-id`) and `4` for the v2 API. This is the regression guard for `925a11b46c`.
- [ ] Manual: open a v4 MESSAGE API → API Traffic and check the Overview tile is `>=` every entrypoint tile. To force the symptom without load, index a metrics document with a valid `entrypoint-id` and `"status": 0` into `gravitee-v4-metrics-*`.
- [ ] Manual: on the same API, check the Response Status chart shows a grey **No status** slice and that its slices add up to the Total Requests tile.
- [ ] API-level: call `requests-count` for a Native Kafka API and check `total` is no longer `0`. Not checkable from the console — the NATIVE page has no "Total Requests" tile.
- [ ] Console Jest suite — not runnable locally (`node_modules` absent from the working tree), but `ci/circleci: Lint & test APIM Console` passes on the current head.

## Reviewer notes

⚠️ **Reported figures change visibly — this needs a release note.**
- The Overview "Total Requests" **goes up**: it now includes requests the status filter used to drop (on the reported data, 164 774 → at least 230 789).
- A **"No status" slice appears** in the Response Status chart. On the reported data it holds the ~66 000 requests that had no committed status. It is not new traffic — only traffic that was previously invisible. **This is not limited to the API Traffic page:** `api-analytics-response-status-ranges` is shared with the Home dashboard, whose v4 donut is fed by the environment-level `GET /analytics/response-status-ranges` through the same adapter. A customer who screenshots the home page will see the slice there too.
- APIs busier than 10 000 requests in the window stop reporting a flat 10 000 on `type=COUNT`.

None of these is a regression; all three are corrections. But a customer who screenshots dashboards will see three numbers move, so they should not ship silently.

**Riskiest hunk:** `SearchRequestsCountResponseAdapter#totalCount`, specifically the fallback when `SearchResponse#getSearchHits()` is null. Falling back to the bucket sum rather than `0` is deliberate — it keeps `total >= max(countsByEntrypoint)` true even on a malformed response, and it now logs a warning instead of failing silently.

**Perf note on commit 3:** this is the only query in the PR where `track_total_hits: true` has a real cost, since it has no aggregation forcing a full collection. Worth a second opinion if you expect very large windows on that endpoint.

### Endpoint summary

| Endpoint | Method | Change | Response before | Response after |
|---|---|---|---|---|
| `.../analytics/requests-count` | GET | `total` semantics fixed — no contract change | `{"total": 164774, "countsByEntrypoint": {"http-post": 230789}}` | `{"total": 230789, "countsByEntrypoint": {"http-post": 230789}}` |
| `.../apis/{apiId}/analytics/response-status-ranges` | GET | **new `unknown` key** in `ranges` and in each `rangesByEntrypoint` entry | `{"ranges": {"200.0-300.0": 164774, …}}` | `{"ranges": {"200.0-300.0": 164774, …, "unknown": 66015}}` |
| `.../analytics/response-status-ranges` (environment) | GET | same new `unknown` key — feeds the Home dashboard donut | `{"ranges": {"200.0-300.0": 164774, …}}` | `{"ranges": {"200.0-300.0": 164774, …, "unknown": 66015}}` |
| `.../analytics?type=COUNT` | GET | exact count instead of a 10 000 ceiling | `{"count": 10000}` | `{"count": 4218773}` |

The OpenAPI contract already documents `requests-count`'s `total` as *"The total count of requests"* — this aligns the implementation with it. No API, model or service-layer change for commits 1 and 3; `AnalyticsQueryServiceImpl`, `SearchRequestsCountAnalyticsUseCase` and `ApiAnalyticsResource` are pure pass-through. Commit 2 adds one map key, documented as a reserved key on both the API-level and environment-level `ranges` schemas in the OpenAPI. The console already renders unrecognised keys generically, so the mapping to a **No status** label is the only front-end change.

### Follow-ups found while investigating (deliberately not here)

- **`Float` narrowing on counts — not fixable in this repository.** The lossy step is `io.gravitee.elasticsearch.model.Aggregation`, whose `value` / `count` / `sum` / `avg` / `min` / `max` are `java.lang.Float`, populated by Jackson inside `gravitee-common-elasticsearch` 7.0.0. `AbstractResponseAdapter#lookupForAggregations` hands back those already-deserialised objects, so by the time `StatsQueryAdapter` or the observability engine's `AggregationAdapter` reads them the digits are gone. Fixing it properly means widening those fields upstream and bumping the dependency. Working around it per query — sourcing counts from a bucket `doc_count` read exactly from raw JSON — would additionally require widening `count` from `float` to `long` through `StatsAggregate` → `StatsAnalytics` → the REST model → OpenAPI → the console, a public contract change. Scale of the error: `float` is exact for integers below 2^24 ≈ 16.7 M and drifts by ~1 part in 16 M above that, so the impact on a displayed count is invisible. Not worth a contract change on a maintenance branch.
- **"Top N" widgets capped at ten.** `AggregateValueCountByFieldAdapter` and `SearchTopFailedApisAdapter` have unsized `terms` aggregations on `api-id`. Arguably intentional for a "top" widget, but undocumented.
- **Native Kafka counts connection lifecycle events, not connections** — a `CONNECTED` and its later `DISCONNECTED` are the same session. Pre-existing and untouched by this PR; with no console surface today it only concerns direct API consumers.

Closes APIM-14859

[APIM-14859]: https://gravitee.atlassian.net/browse/APIM-14859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchRequestsCountQueryAdapter.java
```diff
@@ -40,9 +46,10 @@ public static String adapt(RequestsCountQuery query, boolean isEntrypointIdKeywo
     private static JsonObject buildEntrypointIdAggregate(boolean isEntrypointIdKeyword) {
         return JsonObject.of(
             "entrypoints",
-            JsonObject.of("terms", JsonObject.of("field", isEntrypointIdKeyword ? "entrypoint-id" : "entrypoint-id.keyword")),
-            "all_apis_status_ranges",
-            JsonObject.of("range", JsonObject.of("field", "status", "ranges", JsonArray.of(JsonObject.of("from", 100.0, "to", 600.0))))
+            JsonObject.of(
+                "terms",
+                JsonObject.of("field", isEntrypointIdKeyword ? "entrypoint-id" : "entrypoint-id.keyword", "size", ENTRYPOINT_BUCKETS)
+            )
```
jourdiw, line 52: You build a json object similarly in other files with the size + buckets. What about having somewhere shared to centralize this logic?

Additionally, you don't include the ranges any longer in this object, is that normal?

kovaren (approved):
The changes look fine, my only concern is that we are changing the behavior of existing endpoints which might raise questions with the clients who were not aware of the issue.

It's only briefly mentioned in the JIRA comment, I think it's worth making this point more prominent.

Also please add the JIRA ticket to the PR description.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
```diff
@@ -9973,7 +9973,13 @@ components:
                         properties:
                             ranges:
                                 type: object
-                                description: Global status ranges
+                                description: >
+                                    Global status ranges. Alongside the `<from>-<to>` range keys, the
+                                    reserved key `unknown` holds the requests those ranges cannot
+                                    classify — typically a request whose response status was never
+                                    committed, indexed with status 0. Every request counted by
```
jourdiw, line 9980: **The invariant this sentence states is false unless the caller passes `from` and `to`.** The two endpoints default to different time windows:

- `requests-count`: `ApiAnalyticsResource#getApiAnalyticsRequestCount` wraps the params in `Optional`, `SearchRequestsCountAnalyticsUseCase` passes `input.from.orElse(null)` straight through, and `SearchRequestsCountQueryAdapter#buildElasticQuery` then emits **no `@timestamp` range at all** — so an unparameterised call counts *all time*.
- `response-status-ranges`: `SearchResponseStatusRangesUseCase#execute` defaults to `end = now`, `start = end - 1 day` — *the last 24 hours*.

So `GET /apis/{id}/analytics/requests-count` and `GET /apis/{id}/analytics/response-status-ranges` with no query params return numbers over different windows, and the values cannot add up. The console always sends both params, so this doesn't affect the UI — it affects direct API consumers, which is exactly who reads this description.

There's a second, narrower way the same claim fails, independent of time: `searchRequestsCount` unconditionally queries `v4-metrics,request`, while `searchResponseStatusRanges` resolves indices from `definitionVersions`. For an API migrated in place, the `should` clause on `api-id`/`api` matches both eras in the first endpoint and only one in the second.

Suggest either softening this to state the same-window precondition explicitly, or aligning the `requests-count` default with the 24 h one.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusRangesAdapter.java
```diff
@@ -156,14 +177,39 @@ public Optional<ResponseStatusRangesAggregate> adaptResponse(SearchResponse resp
             .stream()
             .collect(Collectors.toMap(jsonNode -> jsonNode.get("key").asText(), jsonNode -> jsonNode.get("doc_count").asLong()));
 
+        final var ranges = totalHits(response)
+            .map(total -> withUnknownRange(allApisStatusRanges, total))
```
jourdiw, line 181: **At environment level this bucket absorbs native Kafka connection events, and the release note doesn't prepare anyone for that.**

`SearchEnvironmentResponseStatusRangesUseCase` searches every V2 + V4 API in the environment with **no `ApiType` filter** — `ApiSearchCriteria` has a `notApiTypes` field and the use case never sets it. Native Kafka metrics are plain `Metrics` objects routed to the same `v4-metrics` index and carry no `status` field at all, so every connection lifecycle event in the window lands in `unknown` and is drawn on the Home "API Response Status" donut. The card's "Nb hits total" inflates with it, since Highcharts computes that from the rendered slices.

The reviewer notes describe the slice as "requests that had no committed status — not new traffic, only traffic that was previously invisible." On a Kafka-heavy environment that framing doesn't hold: it's a different traffic model, it's unbounded relative to HTTP traffic, and it can dominate the donut.

Two ways out, either is fine:
- exclude NATIVE from the environment-level query via `notApiTypes`; or
- split the bucket. A `missing` / `exists` sub-aggregation on `status` separates "status never committed" from "no status field at all" at no extra query cost, and would let the console label both honestly instead of collapsing them into one grey slice.

The javadoc below already concedes the bucket means *outside the ranges*; this is the case where that distinction stops being academic.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
```diff
@@ -9988,9 +9994,13 @@ components:
                                         value: 0
                                     500.0-600.0:
                                         value: 0
+                                    unknown:
+                                        value: 12
                             rangesByEntrypoint:
                                 type: object
-                                description: Status ranges by entrypoint
+                                description: >
+                                    Status ranges by entrypoint. Carries the same reserved `unknown`
```
jourdiw, line 10002: **"the same reserved `unknown` key, scoped to each entrypoint" reads as a clean decomposition, and it isn't one.** The two values are derived from different denominators:

- global `unknown` comes from `hits.total` (`SearchResponseStatusRangesAdapter`, the `totalHits(response).map(...)` line);
- per-entrypoint `unknown` comes from the terms bucket `doc_count`.

Documents with no `entrypoint-id` are counted in the first and form no bucket in the second — and this PR's own fixture proves such documents exist in the wild (`should_return_total_requests_count_including_requests_without_entrypoint` asserts 13 total against a 12-request breakdown, the thirteenth having no `entrypoint-id`).

So on real data `Σ rangesByEntrypoint[*][unknown] < ranges[unknown]`, and a consumer that sums the per-entrypoint buckets expecting to recover the global one will come up short. Worth saying so here, since the whole point of documenting the key is to let consumers rely on it.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusRangesAdapter.java
```diff
@@ -156,14 +177,39 @@ public Optional<ResponseStatusRangesAggregate> adaptResponse(SearchResponse resp
...
+     * committed — an aborted connection, a client that hung up — which is why the console labels the
+     * bucket "No status". A status outside [100, 600) would land here too, but that takes an
+     * upstream returning a value that is not a valid HTTP status.
+     *
+     * @param matchedRequests every request the query matched, whatever its status
+     */
+    private static Map<String, Long> withUnknownRange(Map<String, Long> statusRanges, long matchedRequests) {
+        final long classified = statusRanges.values().stream().mapToLong(Long::longValue).sum();
+        final var result = new LinkedHashMap<>(statusRanges);
+        result.put(UNKNOWN_RANGE, Math.max(0, matchedRequests - classified));
```
jourdiw, line 205: **This clamp is the right behaviour but it should not be silent.** `matchedRequests < classified` is impossible by construction — which is exactly why it's worth knowing if it ever happens. Partial shard failure, or a proxy quietly dropping `track_total_hits`, would produce a `0` here indistinguishable from a genuinely fully-classified window.

`SearchRequestsCountResponseAdapter#totalCount` logs a warning in its analogous impossible branch. A `log.warn` when the subtraction goes negative would make these two consistent, and costs nothing on the hot path since the branch never taken.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/analytics/AnalyticsElasticsearchRepositoryTest.java
```diff
@@ -108,17 +113,54 @@ class AnalyticsElasticsearchRepositoryTest extends AbstractElasticsearchReposito
...
+            var result = cut.searchRequestsCount(new QueryContext("org#1", "env#1"), new RequestsCountQuery(API_ID_UNCOMMITTED_STATUS));
+
+            assertThat(result).hasValueSatisfying(countAggregate -> {
+                assertThat(countAggregate.getTotal()).as("one request with a status, one indexed with status 0").isEqualTo(2L);
+                assertThat(countAggregate.getCountBy()).containsExactlyInAnyOrderEntriesOf(Map.of("http-post", 2L));
+            });
+        }
+
+        @Test
+        void should_count_native_kafka_metrics_indexed_without_any_status() {
```
jourdiw, line 140: **This is a behaviour change on `total`, not the pre-existing issue the follow-up note describes.** The comment below is right that the 4 documents are lifecycle events rather than 4 connections, and I agree the counting-semantics question is a separate product concern worth deferring. But the follow-ups section files it as "pre-existing and untouched by this PR" — that's true of `countsByEntrypoint`, which already reported `native-kafka: 4`, and not of `total`, which this PR moves from `0` to a number that counts a `CONNECTED` and its later `DISCONNECTED` as two.

Not asking for a behaviour change here — the pre-PR pairing of `total: 0` next to `countsByEntrypoint: {native-kafka: 4}` was equally undetectable garbage, and self-consistency is a clear improvement. Just that the follow-up note should say `total` starts reporting a double-counted session figure for NATIVE APIs, rather than implying nothing moved. It's the one number on the endpoint summary table that no console page will surface, so the API docs are the only place a consumer could find out.

jourdiw (approved):
Approving. The three root causes are correctly diagnosed and the fixes are the right ones — deriving the total from `hits.total` instead of a `status`-range aggregation makes `total >= max(countsByEntrypoint)` true by construction rather than by coincidence, which is the invariant that was actually missing. The three-commit split is clean and genuinely independently droppable, and the regression tests were verified against the reverted production code, which is the part that usually gets skipped.

Five comments below. None of them block the counting fixes themselves; #1 and #2 are about claims and blast radius introduced by commit 2, and I'd want them settled before this ships, since both change what a customer reads.

Things I checked and cleared, to save the next reviewer the trip:

- **No consumer parses the range keys numerically.** I went looking for a `parseFloat(key.split('-')[0])` that the new `unknown` key would turn into `NaN`. There isn't one — console, portal, the Java mappers and the e2e SDK all treat `ranges` as an opaque `Map<String, Long>`, and `getLabel` already falls through to the raw key. The new reserved key is safe.
- **`track_total_hits` is fine across the supported matrix** (ES 7.17/8.19/9, OpenSearch 1/2/3). `EventMetricsQueryAdapter` already ships the key on a production path, and `TotalHitsDeserializer` handles both the object and the legacy scalar form.
- **The perf note on commit 3 is correct, not conservative.** I initially thought it was overstated — that with `size: 0` and a filter-only query, Lucene's `Weight#count` short-circuit would make it about as cheap as `_count`. It doesn't apply here: the query is a `bool.must` conjunction, so `BooleanWeight#count` returns -1 and the collector falls back to full doc-id enumeration. The right baseline isn't `_count` (also uncapped) but the pre-PR behaviour, where this aggregation-free query genuinely early-terminated at 10 000. The trade is real and the comment describes it accurately.
- **The `terms` size 10 → 100 bump is a clean fix**, and load-bearing: the real entrypoint plugin set is already around 10, so the truncation was live rather than theoretical. No accuracy change, and no phantom endpoint buckets in `SearchAverageMessagesPerRequestQueryAdapter` — `connector-type: entrypoint` is a query-level filter, not a sub-aggregation one.
- **`SearchResponse#getSearchHits()` cannot be null in practice.** ES and OpenSearch always return the `hits` object from `_search`, including with `size: 0`. So the fallback flagged as the riskiest hunk is unreachable outside a hand-built `new SearchResponse()`. That's worth knowing before spending review attention there — it also means the concern that the bucket-sum fallback silently reintroduces the `925a11b46c` bug (it does drop entrypoint-less documents) is real but can't fire.
- **No existing test, fixture or console spec is broken.** The single exact-equality assertion on the ranges map was caught and updated; `assertRanges` uses `containsAllEntriesOf`; the rest-api-side tests are fake-fed. One follow-up worth a ticket: `HttpStatusCodeGroups` on the analytics-engine surface uses the `"100-199"` key format and gains no `unknown`, so the product now has two status breakdowns that disagree about whether unclassified traffic exists.

PR #18994, opened 2026-08-05
fix(gateway): optimize API product subscription sync

gutek9:
## Summary

This PR optimizes gateway cold start when API Products contain many APIs and subscriptions.

API Product subscriptions are now expanded only for APIs in the current synchronization batch. The initial API Product synchronization also avoids refreshing all subscriptions because they are loaded immediately afterwards during API synchronization.

The fix supports both repository and distributed synchronization.

## Root cause

An API Product subscription applies to every API contained in the product. For large products, this creates a significant Cartesian expansion.

In the affected environment:

- 36,487 logical API Product subscriptions
- 218 API Products
- approximately 5.5 million required runtime subscription entries
- individual products containing up to 792 APIs

During cold start, the gateway previously:

1. Expanded and registered the complete `subscription × API` result during API Product synchronization.
2. Loaded the same subscriptions again during API synchronization.
3. For every API batch, expanded subscriptions to every API in the product.
4. Discarded entries belonging to APIs outside the current batch.
5. Generated warnings containing thousands of subscription IDs for those discarded entries.

This caused excessive temporary allocations, GC pressure, bridge queries, duplicate registration and very large log messages.

## Changes

- Add batch-scoped API Product subscription mapping.
- Expand subscriptions only for APIs present in the current synchronization batch.
- Filter regular API subscriptions to the current batch before allocation.
- Deduplicate plan IDs used in subscription repository queries.
- Replace the warning containing every subscription ID with a bounded count-only warning.
- Skip the full API Product subscription refresh during initial synchronization.
- Keep API Product subscription refresh enabled for incremental updates.
- Apply the same initial-sync behavior to distributed synchronization.
- Preserve distributed subscription and API key events emitted during API synchronization.

## Compatibility

Regular API subscriptions retain their existing 1:1 behavior.

Initial synchronization still loads only active, non-expired subscriptions and API keys using paginated repository queries.

Incremental API Product updates continue to refresh subscriptions and API keys as before.

The final runtime subscription cardinality is unchanged: every valid API Product subscription is still registered for every applicable API. This change removes duplicate and out-of-batch processing rather than changing authorization semantics.

No configuration or database migration is required.

## Expected impact

- Avoid materializing the complete API Product subscription expansion twice.
- Eliminate expansion to APIs outside the current synchronization batch.
- Significantly reduce peak heap usage and temporary allocations.
- Reduce GC and bridge load during gateway startup.
- Eliminate multi-megabyte subscription warning messages.
- Allow gateways with large API Products to complete cold start within a substantially shorter time.

## Testing

The following scenarios are covered:

- API Product subscriptions are expanded only to APIs in the current batch.
- Regular API subscriptions are still attached to the correct API.
- Initial API Product synchronization skips the redundant refresh.
- Incremental API Product synchronization still performs the refresh.
- Repository and distributed synchronization use the correct behavior.
- API Product API keys and subscriptions remain available through distributed events.

Full `gravitee-apim-gateway-services-sync` test suite:

- 675 tests
- 0 failures
- 0 errors

## Remaining limitation

The gateway still needs to retain the required runtime entries. In the affected environment this is approximately 5.5 million entries.

A longer-term improvement could avoid storing the complete `subscription × API` Cartesian representation in the gateway cache, but that requires a broader authorization-cache redesign.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
@@ -42,6 +42,16 @@ public class ApiProductDeployer implements Deployer<ApiProductReactorDeployable>
 
     @Override
     public Completable deploy(ApiProductReactorDeployable deployable) {
+        return deploy(deployable, true);
+    }
+
+    /**
+     * Deploys an API Product and optionally refreshes its subscriptions. During an initial sync,
+     * subscriptions are appended while APIs are deployed immediately after API Products, so doing
+     * the refresh here would build and register the complete product-subscription Cartesian product
+     * twice.
+     */
+    public Completable deploy(ApiProductReactorDeployable deployable, boolean refreshSubscriptions) {
         ReactableApiProduct reactableApiProduct = deployable.reactableApiProduct();
         String apiProductId = deployable.apiProductId();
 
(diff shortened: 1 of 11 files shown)


phiz71:

PR #18995, opened 2026-08-05
fix(gateway): stop duplicating and over-expanding API Product subscriptions

callaertanthony:
> **Overlaps #18993 and #18994**, which were opened independently against `4.12.x` for the same
> incident within minutes of this work. See *Relationship to the other two PRs* at the bottom — this
> PR is the narrowest of the three and should probably be closed in favour of #18993 unless the
> warning treatment here is preferred.

## Problem

An API Product subscription is materialised once per member API. On a deployment where products span
several hundred APIs, gateway initial sync never finishes: the `sync-process` probe stays not-ready,
the pod is killed, and the whole sync restarts against the management API.

Three paths in the sync pipeline each expand the same product subscription set.

### 1. `ApiProductDeployer` registered the whole product subscription set per product

`deploy()` called `ApiProductSubscriptionRefresher.refresh()` for every product, which loads the
product's subscriptions and registers a leg for each member API. The API Product synchronizer
(`Order.API_PRODUCT` = 11) runs immediately before the API synchronizer (`Order.API` = 12), and the
list is walked with `concatMapCompletable`, so this ran strictly serially — one product at a time.

Measured on the reported deployment: **296 products in 280,795 ms, single-threaded on
`gio.sync-deployer-0`.**

### 2. `SubscriptionAppender` expanded across the whole product, then discarded most of it

`SubscriptionAppender` processes APIs in batches of `services.sync.bulk_items` (default 100), but
`SubscriptionMapper` exploded a product subscription across the product's *entire* API list. Legs for
APIs outside the current batch were allocated, grouped into a map, and dropped.

Because a product's APIs are spread across the event stream rather than grouped, nearly every batch
touches a large product.

### 3. Every discarded leg was logged with the full subscription id list

The warning joined all discarded subscription ids into one line:

```java
log.warn("Cannot find api {} for subscriptions [{}]", api,
         subscriptions.stream().map(Subscription::getId).collect(Collectors.joining(",")));
```

The joined string is built as a *method argument*, so it is allocated before the level check — the
allocation happens at any configured log level.

Measured across two log captures from the incident:

| | capture 1 | capture 2 |
|---|---|---|
| warning lines | 1,080 of 1,151 total lines | 205 |
| bytes per line | 64,839 (exactly, every line) | 49,835 avg |
| total | 66.8 MB in 6.06 s | 9.7 MB in 85 ms (~114 MB/s) |
| distinct APIs | 775 | — |

GC in the same window reported `Minor Collection (Allocation Rate)`.

## Changes

| File | Change |
|---|---|
| `SubscriptionMapper` | New overload `to(model, Set<String> apiScope)`. The scope is applied to the product's api ids **before** mapping, so out-of-scope legs are never allocated. `null` keeps the existing expand-everything behaviour. |
| `SubscriptionAppender` | Passes `deployableByApi.keySet()` — the batch it is actually handling — as the scope. |
| `SubscriptionAppender` | The warning now reports a count plus a sample capped at 5 ids, instead of joining every id. |
| `ApiProductDeployer` | Drops the deploy-time subscription registration. |

The unscoped `to(model)` overload is deliberately retained: `SubscriptionSynchronizer` (order 13) and
`LocalApiSynchronizer` handle single subscription events that must still reach every API of the
product.

### Why dropping the deploy-time registration is safe

`SubscriptionAppender` already registers a leg for every member API:

- **at initial sync**, because the API synchronizer runs immediately after the API Product
  synchronizer, over the same event set;
- **on a runtime product change**, because `ApiProductManagerImpl` emits a `DEPLOY` event that
  `RepositoryApiMemberResyncTrigger` consumes and re-runs the API pipeline for the member APIs.

`unregisterRemovedApis` is untouched — removing an API from a product still evicts its legs.

## Testing

`gravitee-apim-gateway-services-sync`: **681 tests, 0 failures, 0 errors**, run on the `4.12.x` base.

New tests:

- `should_not_expand_api_product_subscriptions_to_apis_outside_the_current_batch`
- `should_bound_the_warning_to_a_count_and_a_capped_sample` — 7 orphaned subscriptions, asserts the
  count is reported and that ids 6 and 7 do not appear
- `should_not_register_subscriptions_on_deploy_since_the_api_sync_path_already_registers_them`

## Not measured

There is **no before/after startup timing** and no reproduction at the 296-product / 775-API scale.
The tests prove the mechanism changed. They do not prove the pod starts. Whichever of these three
PRs is taken forward should be validated against a realistic product set before it is relied on.

## Relationship to the other two PRs

Three of us converged on the same batch-scoping fix independently.

| | #18993 | #18994 | this |
|---|---|---|---|
| Batch-scope the product explosion | yes (`Predicate<String>`) | yes (`Set<String>`) | yes (`Set<String>`) |
| Skip the deploy-time refresh | no — paginates it instead | yes, initial sync only | yes, entirely |
| Bound the warning | leaves it | count only | count + 5-id sample |
| Distributed sync path | — | yes | no |
| Paginate the refresher, `ACCEPTED` only | yes | — | no |
| `SyncConfiguration` pools stuck single-threaded | yes | — | no |
| `SubscriptionCacheService` leg index O(P)→O(1) | yes | — | no |

**#18993 covers strictly more than this PR.** It carries two findings absent here that look like they
matter more than the expansion itself: the sync executors never grew past one thread because
`corePoolSize` was 1 behind an unbounded queue, and registering a product leg was O(P), making a
warmup O(P²).

**#18994 covers the same ground as this PR** plus the distributed-sync path, which this PR does not
touch — that path was set aside here after distributed sync was confirmed disabled on the affected
deployment, but it is needed for a general fix.

Two differences here are worth keeping whichever PR wins:

- **The warning keeps a bounded sample rather than dropping the ids entirely.** A count alone tells
  you the batch skipped something; five ids let you go look at one. The whole-list join is what has
  to go, not the ids.
- **`ApiProductDeployer` drops the deploy-time registration outright** rather than gating it on
  initial sync. If the incremental path genuinely still needs it, the gated form in #18994 is the
  safer call and this is wrong — worth settling explicitly rather than by whichever merges first.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
@@ -68,11 +68,12 @@ public class ApiProductDeployer implements Deployer<ApiProductReactorDeployable>
             }
         }
         doBeforeEmit = doBeforeEmit.andThen(Completable.fromRunnable(() -> registerApiProductPlans(deployable)));
-        if (newApiIds != null && !newApiIds.isEmpty()) {
-            doBeforeEmit = doBeforeEmit.andThen(
-                subscriptionRefresher.refresh(deployable.subscribablePlans(), Set.of(reactableApiProduct.getEnvironmentId()))
-            );
-        }
+
+        // Subscriptions are deliberately NOT registered here. SubscriptionAppender already registers
+        // them for every member API: at initial sync because the API synchronizer runs right after
+        // this one, and on a runtime product change because RepositoryApiMemberResyncTrigger re-runs
+        // the API pipeline for the member APIs. Doing it here as well loaded and expanded the whole
+        // product subscription set a second time, serially, per product.
 
         return apiProductManager
             .register(reactableApiProduct, doBeforeEmit)
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/mapper/SubscriptionMapper.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/mapper/SubscriptionMapper.java
@@ -46,9 +46,21 @@ public class SubscriptionMapper {
     private final ApiProductRegistry apiProductRegistry;
 
     public List<Subscription> to(io.gravitee.repository.management.model.Subscription subscriptionModel) {
+        return to(subscriptionModel, null);
+    }
+
+    /**
+     * Maps a repository subscription, expanding API Product subscriptions across the product's APIs.
+     *
+     * @param apiScope when non-null, an API Product subscription is only expanded for the APIs it
+     *                 contains. Callers that process APIs in batches pass the current batch so the
+     *                 expansion does not materialise subscriptions they are about to discard.
+     *                 {@code null} expands across every API of the product.
+     */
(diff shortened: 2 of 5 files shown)


phiz71:

PR #18996, opened 2026-08-05
perf(gateway): make API Product cold start scale with product size

callaertanthony:
Supersedes #18993, #18994 and #18995 — three PRs opened independently against `4.12.x` within
minutes of each other for the same incident. This is the union of all three, plus tests for the two
changes that had none. If it is taken, the other three should be closed.

Credit is preserved: the first four commits are @wbabyte's, unmodified. @gutek9's distributed-sync
work and plan deduplication are carried in the final commit.

## Problem

An API Product subscription is materialised once per member API. On a deployment where products span
several hundred APIs, gateway initial sync never finishes: the `sync-process` probe stays not-ready,
the pod is killed, and the whole sync restarts against the management API.

Notation: **P** = APIs in a product, **S** = its ACCEPTED subscriptions, **B** = `services.sync.bulk_items`.

Six defects compose to produce this.

| # | Defect | Found in |
|---|---|---|
| 1 | Sync executors were permanently single-threaded | #18993 |
| 2 | Registering one subscription leg was O(P), making a product warmup O(P²) | #18993 |
| 3 | Product subscriptions were exploded across all P APIs, then discarded | all three |
| 4 | The refresher issued an unpaginated query including CLOSED | #18993 |
| 5 | The deploy path re-registered the whole expansion, serially, per product | #18994, #18995 |
| 6 | Every discarded leg was logged with the full subscription id list | #18994, #18995 |

### 1. Sync executors were permanently single-threaded

Both pools were built as `new ThreadPoolExecutor(1, poolSize, …, new LinkedBlockingQueue<>(), …)`.
`ThreadPoolExecutor` only creates a thread beyond `corePoolSize` when the queue *rejects* an offer,
and an unbounded `LinkedBlockingQueue` never rejects. The pools ran with exactly one thread
regardless of `poolSize`, so `services.sync.fetcher` / `services.sync.deployer` had no effect and the
`.parallel(getMaximumPoolSize())` fan-out in `AbstractApiSynchronizer` created N rails that all
funnelled through one thread.

This is now covered by a behavioural test — three tasks that each block until all three have
started. It deadlocks and times out on a single-threaded pool; both parameterised cases were
confirmed failing against the exact pre-fix configuration.

### 2. Registering a leg was O(P)

Legs were held in an immutable `Set<Subscription>`, copied on every registration. The singleton fast
path never applies to an exploded product subscription, whose set grows to P legs — so registering a
product's legs cost ~P² element copies per subscription. The index is now
`ConcurrentMap<String, ConcurrentMap<LegKey, Subscription>>` keyed on `(api, environmentId)`, the
exact pair the old code filtered on, making registration an O(1) put.

### 3–6

Product explosion is now restricted to the API set the caller is deploying, applied *before* mapping
so out-of-batch legs are never allocated; the deploy-time refresh is skipped during initial sync
(the API synchronizer runs immediately after over the same event set and registers the legs anyway);
the refresher paginates by `(plan, id)` keyset and requests `ACCEPTED` only on the deploy path; and
the orphan warning reports a count with a sample capped at five ids.

## Where this differs from the three source PRs

Every difference below is a deliberate choice between two source PRs that disagreed, or an addition.

**Explosion restriction walks the smaller set.** #18993 filtered the product's api list (P lookups);
#18994 filtered the batch (B lookups). Taken: whichever is smaller, so a batch of 100 against a
product of 792 costs 100.

**Plain API subscriptions are not filtered.** #18994 filtered them too, which silently drops a
subscription whose api disagrees with the API owning its plan. #18993 deliberately left them
unfiltered so the orphan warning still reports that inconsistency. Taken: #18993's choice, and the
test for it is kept rather than replaced.

**The deploy-time refresh is gated, not removed.** #18995 removed it outright, arguing the API sync
path covers both initial sync and runtime product changes. That is the riskier reading — an
incremental product change has no API event to piggyback on. Taken: #18994's gating, which costs
nothing at cold start and cannot regress the incremental path.

**The warning keeps a bounded sample.** #18994 made it count-only; #18993 left it alone. Taken:
count plus five ids. A count tells you the batch skipped something; ids let you go look at one. The
whole-list join is what had to go, not the ids.

**Two new tests** for changes that shipped without any: the executor concurrency test above, and a
capped-sample test asserting the count is reported in full while ids past the cap are absent.

## Testing

| Module | Tests | Result |
|---|---|---|
| `gravitee-apim-gateway-services-sync` | 688 | 0 failures, 0 errors |
| `gravitee-apim-gateway-handlers-api` | 982 | 0 failures, 0 errors |

`SubscriptionCacheServiceTest` (46 tests, including concurrent-update and multi-leg eviction cases)
passes unmodified against the rewritten leg index — that is the main evidence the O(1) rewrite in
commit 2 preserves behaviour.

**Every behavioural change was mutation-checked** — the fix reverted, the test confirmed failing,
the fix restored:

| Mutation | Failed |
|---|---|
| Pool config restored to `(1, poolSize)` | `should_run_tasks_concurrently_on_a_sync_executor` ×2, timing out at 5 s |
| Explosion restriction disabled | `should_explode_api_product_subscription_only_for_target_apis`, `should_only_materialise_api_product_legs_for_apis_in_the_current_batch` |
| `refreshSubscriptions` gate removed | `should_skip_subscription_refresh_when_requested` |
| Sample cap removed | `should_bound_the_warning_to_a_count_and_a_capped_sample` |
| Sample cap shrunk from 5 to 4 | `should_bound_the_warning_to_a_count_and_a_capped_sample`, on the boundary id |
| Small-product explosion arm filter swapped to `apiIds` | `should_restrict_explosion_to_target_apis_when_product_is_smaller_than_target_set` |
| Removed-api eviction folded under the refresh gate | `should_unregister_removed_apis_even_when_subscription_refresh_is_skipped` |

The last three rows come from a review pass on this PR, which closed three coverage gaps: the
sample cap is now pinned exactly (not just "bounded"), the explosion arm that walks a product
smaller than the batch is exercised, and removed-api eviction is pinned as independent of the
`refreshSubscriptions` flag — a future edit folding it under the gate fails the suite.

## Not measured

There is **no before/after startup timing** and no reproduction at the 296-product / 775-API scale.
The tests prove each mechanism changed and that each test can fail. They do not prove the pod starts.
This should be validated against a realistic product set before it is relied on for the incident.

## Risk

The executor change is the one to watch: it converts API deployment from serial to concurrent across
the whole sync pipeline. That is the fan-out the code was already written for — `.parallel()` rails
that were silently collapsing onto one thread — but any latent ordering assumption in a deployer
would surface here rather than in the pieces above.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
@@ -31,7 +31,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -59,10 +59,13 @@ public class SubscriptionCacheService implements SubscriptionService {
     // ConcurrentSkipListMap), which would double those side effects under contention.
     private final ConcurrentMap<IdentityKey, Subscription> cacheByApiClientId = new ConcurrentHashMap<>();
     private final ConcurrentMap<IdentityKey, Subscription> cacheByClientCertificate = new ConcurrentHashMap<>();
-    // Single by-id index, including exploded API-Product subscriptions. Values are immutable sets
-    // (almost always a singleton), replaced atomically via compute(): keeps the per-entry footprint
-    // minimal at multi-million-subscription scale and lets readers use them without defensive copies.
-    private final ConcurrentMap<String, Set<Subscription>> cacheBySubscriptionIdAll = new ConcurrentHashMap<>();
+    // Single by-id index, including exploded API-Product subscriptions. Each entry maps a leg's
+    // identity (api, environmentId) to its subscription, so registering one leg is an O(1) put.
+    // The previous immutable-set representation had to copy the whole set per registration, which
+    // made warming up an API Product quadratic in its API count: a product spanning P APIs cost
+    // ~P² element copies per subscription, which dominates cold start once P reaches the hundreds.
+    // Inner maps are ConcurrentHashMap so the outer compute() lambdas stay side-effect-safe.
+    private final ConcurrentMap<String, ConcurrentMap<LegKey, Subscription>> cacheBySubscriptionIdAll = new ConcurrentHashMap<>();
     // Holds subscription ids only, for per-API eviction. Identity keys are not tracked here:
     // unregisterByApiId() re-derives them from the cached subscriptions, which keeps this index
     // at one entry per subscription instead of three (id + 2 identity keys) at multi-million scale.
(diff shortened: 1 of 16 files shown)


phiz71:

PR #19000, opened 2026-08-05
fix(gateway): optimize API Product subscription synchronization

gutek9:
## Description

This PR combines and extends the approaches proposed in #18993 and #18994 to address gateway startup failures caused by large API Product subscription expansions.

A logical API Product subscription is expanded into one runtime subscription per member API. In the affected environment, 36,487 database subscriptions expand into approximately 5.5 million gateway entries.

## Root cause

The gateway was doing unnecessary work and allocating excessive amounts of memory during synchronization:

- API Product subscriptions could be processed twice during initial synchronization.
- Subscriptions were expanded for APIs outside the current deployment batch.
- Adding every expanded leg copied an increasingly large immutable set, resulting in O(P²) work for a product containing P APIs.
- The API Product refresher materialized all subscriptions and their expanded runtime entries before registration.
- Some warning messages included very large collections of subscription IDs.
- Concurrent registration and eviction could leave the subscription cache indexes inconsistent.

### Initial synchronization

- Skip API Product subscription refresh during initial synchronization.
- Let the API subscription synchronizer populate the cache once.
- Apply this behavior to both repository and distributed synchronization.

### Expansion scope

- Expand API Product subscriptions only for APIs present in the current deployment batch.
- Deduplicate product plans before loading their subscriptions.
- Replace large subscription-ID warning payloads with a count and bounded sample.

### Subscription cache

- Introduce a hybrid cache representation:
  - regular 1:1 subscriptions remain compact singleton entries;
  - entries are promoted to a concurrent leg map only when a subscription targets multiple API/environment pairs.
- Make registration of expanded legs O(1), avoiding repeated immutable-set copies.
- Serialize mutations per subscription ID using striped locks.
- Keep the by-ID, identity, certificate, trust-store and per-API indexes consistent during registration and eviction.
- Fix the `unregisterByApiId` race without removing an entire live API index bucket before individual entries are evicted.

### Incremental API Product refresh

- Fetch subscriptions using keyset pagination.
- Process each page independently:
  1. fetch repository subscriptions;
  2. expand the current page;
  3. register subscriptions and API keys;
  4. release the page before fetching the next one.
- Fetch only `ACCEPTED` subscriptions on the deployment path.
- Continue fetching all relevant statuses on the eviction path so subscriptions that left `ACCEPTED` can still be removed.
- Preserve correct federated API-key mapping when referenced subscriptions span multiple pages.

## Expected impact

- API Product subscriptions are no longer processed twice during cold start.
- Cache insertion changes from O(P²) copying to approximately O(P).
- Peak refresher memory is bounded by one subscription page and its runtime expansion instead of the entire product.
- Regular subscriptions do not pay the memory cost of a per-subscription concurrent map.
- Repository bridge response sizes are bounded by `services.sync.bulk_items`.
- Both standard and distributed synchronization follow the same initial-sync behavior.
- No global executor or synchronization parallelism change is introduced.

This does not remove the underlying subscription × API runtime expansion, but it substantially reduces duplicate work, temporary allocations, bridge payload size and cache insertion overhead.

## Validation

Executed on JDK 21:

- `SubscriptionCacheServiceTest`
- `ApiProductSubscriptionRefresherTest`
- `SubscriptionAppenderTest`
- repository and distributed `SubscriptionMapperTest`
- `ApiProductDeployerTest`
- `ApiProductSynchronizerTest`
- `DistributedApiProductSynchronizerTest`

Result: 123 tests, 0 failures, 0 errors.

Additional validation:

- Prettier formatting applied.
- `git diff --check` passes.
- Added concurrent cache registration and eviction coverage.
- Added explicit page-by-page processing coverage.
- Added coverage for federated API keys spanning subscription pages.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/ApiKeyCacheService.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/ApiKeyCacheService.java
@@ -28,6 +28,9 @@ import org.springframework.util.DigestUtils;
 @CustomLog
 public class ApiKeyCacheService implements ApiKeyService {
 
+    private static final int MD5_CACHE_STRIPES = 16;
+    private static final int MAX_MD5_VALUES_PER_STRIPE = 8_192;
+
     // ConcurrentMap on purpose (not plain Map): register()/unregister() rely on thread-safe
     // compute()/computeIfPresent() for the per-API index. Unlike SubscriptionCacheService, the
     // lambdas here are idempotent, so any honest ConcurrentMap implementation is acceptable.
@@ -36,6 +39,15 @@ public class ApiKeyCacheService implements ApiKeyService {
     // Holds api-key values only (the API id is already the map key); unregisterByApiId()
     // re-derives the cache keys from them.
     private final ConcurrentMap<String, Set<String>> cacheApiKeysByApi = new ConcurrentHashMap<>();
+
+    // API Product expansion registers the same raw key once per member API. Cache the digest by
+    // raw key so millions of runtime legs do not repeat the byte[] allocation and MD5 operation.
+    // Stripes keep lookups concurrent; each stripe is independently bounded to avoid retaining
+    // every historical key on gateways with millions of unrelated direct subscriptions.
+    @SuppressWarnings("unchecked")
+    private final ConcurrentMap<String, String>[] md5ByRawKey = java.util.stream.IntStream.range(0, MD5_CACHE_STRIPES)
+        .mapToObj(ignored -> new ConcurrentHashMap<String, String>())
+        .toArray(ConcurrentMap[]::new);
 
     /**
      * Cache key referencing the api-key's own field strings instead of a formatted composite
@@ -135,6 +147,25 @@ public class ApiKeyCacheService implements ApiKeyService {
     }
 
     CacheKey buildMd5CacheKey(ApiKey apiKey) {
-        return buildCacheKey(apiKey.getApi(), DigestUtils.md5DigestAsHex(apiKey.getKey().getBytes(StandardCharsets.UTF_8)));
+        return buildCacheKey(apiKey.getApi(), md5(apiKey.getKey()));
+    }
+
+    private String md5(String rawKey) {
+        ConcurrentMap<String, String> stripe = md5ByRawKey[spread(rawKey.hashCode()) & (MD5_CACHE_STRIPES - 1)];
+        String cached = stripe.get(rawKey);
+        if (cached != null) {
+            return cached;
+        }
+
+        // Approximate bounding is sufficient: a concurrent overshoot is small, and clearing one
+        // stripe only causes later cache misses; it cannot affect authentication correctness.
+        if (stripe.size() >= MAX_MD5_VALUES_PER_STRIPE) {
+            stripe.clear();
+        }
+        return stripe.computeIfAbsent(rawKey, key -> DigestUtils.md5DigestAsHex(key.getBytes(StandardCharsets.UTF_8)));
+    }
+
+    private static int spread(int hash) {
+        return hash ^ (hash >>> 16);
     }
 }
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
@@ -31,7 +31,6 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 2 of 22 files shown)


callaertanthony (changes requested):
Reviewed the three commits against `4.12.x`. The direction is right and several of the individual optimizations are clean wins, but there's one likely-blocking gap on the distributed path plus two behaviour changes that aren't called out in the description.

---

## 1. :red_circle: Distributed initial sync has no compensating expansion path

`DistributedApiProductSynchronizer` now calls `deployer.deploy(deployable, !initialSync)`, so `subscriptionRefresher.refresh(...)` is skipped on cold start. On the **repository** path that's fine — `SubscriptionAppender` re-registers the legs while APIs deploy. On the **distributed** path I don't think it is:

- `SubscriptionAppender` is referenced only from `repository/spring/RepositorySyncConfiguration`, `kubernetes/...` and `local/...` — never from `process/distributed/**`. A replica has no appender.
- The only other subscription source on a replica is `DistributedSubscriptionSynchronizer`, and `distributed/mapper/SubscriptionMapper.to(SubscriptionDeployable)` (untouched by this PR) builds each event as `.id(subscription.getId()).type(SUBSCRIPTION)` — one event id per *logical* subscription, not per `(subscription, api)` leg.

So a replica replaying SUBSCRIPTION events cannot reconstruct the P legs of an API Product subscription. Before this PR, `refresh()` rebuilt them from the repository at startup. Expected symptom: after a cold start, a distributed-sync gateway serves at most one API per product subscription until the next incremental API Product event arrives.

One link I could not settle from this repo: whether the distributed-event store upserts on `id` + `type` (that plugin lives elsewhere). If it appends rather than upserts, legs may survive and this is a non-issue — but that needs establishing rather than assuming.

**Suggested resolution:** either verify with an e2e run on a distributed-sync gateway, or scope the initial-sync skip to `ApiProductSynchronizer` only and leave `DistributedApiProductSynchronizer` unchanged. The repository-path win stands on its own.

## 2. :orange_circle: Narrowing `refresh()` to `ACCEPTED` drops a self-healing eviction

`DEPLOY_STATUS = List.of(ACCEPTED)` replaces `INCREMENTAL_STATUS` on the deploy path. Previously the refresher also fetched `CLOSED` / `PAUSED` / `PENDING`, and `register()` routes a non-`ACCEPTED` subscription to `unregisterInternal` — so every product redeploy re-reconciled stale legs out of the cache.

`SubscriptionSynchronizer` covers normal status transitions, so this isn't a leak in the happy path, but the recovery for any *missed* event is gone. `unregisterRemovedApis` deliberately kept `INCREMENTAL_STATUS`, so the asymmetry looks intentional — worth stating explicitly in the PR description so the next reader doesn't have to reconstruct it.

## 3. :orange_circle: Per-page `loadApiKeys` multiplies repository round trips

`loadApiKeys` moved inside the page consumer. The memory bound is real, but the API-key query count goes from `chunks` to `pages × chunks`. Given that bridge payload size and latency are part of the original symptom, it would be good to measure the end-to-end effect on the affected environment before treating this as a net improvement — the memory win could be partly paid back in round trips.

## 4. :yellow_circle: Two overlapping concurrency mechanisms, with stale comments

Every mutator in `SubscriptionCacheService` now runs under `subscriptionLock(id)`, which means the `compute()` lambdas are no longer load-bearing for the reason their comments give:

> `// Use compute() so that the isEmpty-check-then-remove is atomic with respect to concurrent register() calls on the same subscription ID.`

Same for the "must stay backed by `ConcurrentHashMap` … the `ConcurrentMap` contract alone allows implementations that retry the lambda" note above the field declarations. Leaving both mechanisms in place with comments that describe only the older one invites someone to delete the wrong half later. Either drop to plain `get`/`put` under the stripe lock, or rewrite the comments to say the stripe lock is now the primary guard and `compute()` is defence in depth.

## 5. :yellow_circle: `should_not_remove_a_new_registration_while_an_old_leg_is_being_unregistered` is racy toward false negatives

`registerStarted.countDown()` fires *before* `subscriptionService.register(replacement)` is called, so `assertThat(registerFinished).isFalse()` can be evaluated before the register thread would have finished even with the striped lock removed. It will usually catch a regression, but it isn't a guard you can rely on. Signalling from inside a stub that `register()` actually invokes — or asserting after a bounded wait — would make it deterministic.

For contrast, `should_keep_every_leg_of_one_subscription_registered_concurrently` (8 writers × 100 legs on one id) is a genuinely good test: it exercises the promotion to `MultipleSubscriptionLegs` under real contention.

## 6. :yellow_circle: Small perf regressions inside a perf PR

- `getAllById` now returns `List.copyOf(subscriptions.values())` for multi-leg entries — an O(P) copy per call, and `ApiKeySynchronizer` calls it per api-key/subscription pair (`flatMapIterable(subscriptionId -> subscriptionService.getAllById(subscriptionId))`). That's exactly the large-product case this PR targets. The `ConcurrentHashMap` values view is already safe to iterate; the copy only buys immutability of the returned handle.
- `getById` went from `subscriptions.iterator().next()` to `values().stream().findFirst()` — a stream allocation per call on a path this file's own comments describe as hot.
- `SUBSCRIPTION_LOCKS` is `static`, so all `SubscriptionCacheService` instances share the same 1024 monitors. An instance field would express the intent better.

---

## Checked and fine

- Dropping the `newApiIds != null` guard is safe — `newApiIds` is now defaulted to `Set.of()` a few lines above.
- `SubscriptionCursor.byPlanAndId` + `field("plan")` sorting is pre-existing on `4.12.x` and already used by `SubscriptionAppender`; the refresher adopting the same pattern is consistent, and the cursor/sort field pairing is correct.
- Federated API keys spanning pages: correct. `seenKeyIds` is per-call and `subscriptionsByIdMulti` only holds the current page, so key K maps to sub1 on page 1 and sub2 on page 2 — no duplicates and no losses.
- Lock ordering is stripe → trust-store in both the register and unregister paths, and `unregisterByApiId` takes one stripe at a time. I could not find an inversion.
- The hybrid `SingleSubscriptionLeg` / `MultipleSubscriptionLegs` representation is a nice fit for the problem, and keeping regular subscriptions off the per-entry map is the right call.

## Note on test evidence

CircleCI **Build backend** was still `pending` when I looked, so the "123 tests, 0 failures" figure is a local run rather than a CI-confirmed one. Worth waiting for green before merging, given how much of this is concurrency behaviour.

phiz71:

PR #19009, opened 2026-08-06
feat(portal-next): homepage automation via portal documentation with drop-and-replace

konrad-krol-incubly:
## Issue

https://gravitee.atlassian.net/browse/EXT-174

## Description

Enables the Automation API to publish/replace a portal's homepage using the existing Documentation resource. A Documentation with `area=HOMEPAGE` is materialized as a HOMEPAGE-area nav item

According to requirements, applying a Documentation with `area=HOMEPAGE` removes stale homepage (created by webconsole) and replaces with the one managed by automation API.
  
Idempotent when re-applied with the same spec (existing deterministic id → in-place update, no replacement fires).

## Next steps

Replace Portal `navigation` with `structure` property that will respect area specifc navigation with backward compatibility


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_documentation/domain_service/navigation/DocumentationNavigationPageMapper.java
```diff
@@ -23,14 +23,15 @@
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemType;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationPage;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationReference;
 import io.gravitee.apim.core.portal_page.model.PortalPageContentId;
 import io.gravitee.apim.core.portal_page.model.PortalVisibility;
 import io.gravitee.apim.core.portal_page.model.Slug;
 import jakarta.annotation.Nullable;
 
 public final class DocumentationNavigationPageMapper {
```
dev-marek: This class should be removed because:

**Package mismatch**. It lives under `portal_documentation.domain_service.navigation`, but 2 of its 3 callers are in different bounded contexts (`portal_page`, `portal_listing`). `phantomParent()` in particular has nothing to do with documentation — `NavigationItemEntryMaterializer` uses it purely to placeholder a parent container for an API listing row, not a doc page. It's really a generic "navigation container" utility that got parked in the documentation package because that's where it was first needed.

**Static utility vs. the rest of the codebase's style**. Everything around it (`PortalDocumentationSyncDomainService`, `ApiDocumentationSyncDomainService`, `NavigationItemEntryMaterializer`) is an injected `@DomainService`. This is a stateless static class instead — the odd one out.

**build() vs apply() duplicate the same field list two ways**. `build()` constructs via `CreatePortalNavigationItem.builder()`, `apply()` mutates an existing `PortalNavigationPage` via setters (`page.setTitle(...)`, `page.setSegment(...)`, ...). Same 7 fields, two different mechanisms, kept in sync by hand — a missed field in one won't be caught by the compiler. `apply()` also reaches into the domain object's setters from outside, which is more mutation-by-mapper than you'd expect if `PortalNavigationPage` is meant to own its own update behavior.

`PortalNavigationPage` must be updated using dedicated `update` methods instead of setters.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_documentation/domain_service/navigation/DocumentationNavigationPageMapper.java
```diff
@@ -48,16 +49,24 @@ public static PortalNavigationItem build(
...
             .order(meta.order().orElse(DEFAULT_ORDER))
             .portalPageContentId(contentId)
+            .reference(referenceFrom(meta))
             .visibility(PortalVisibility.PUBLIC)
             .published(true)
             .build();
         return PortalNavigationItem.from(create, organizationId, environmentId, parent);
     }
 
+    private static PortalNavigationReference referenceFrom(AutomationMetadata meta) {
```
dev-marek: This should be a method in `AutomationMetadata`

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_documentation/domain_service/navigation/DocumentationNavigationPageMapper.java
konrad-krol-incubly: refactored

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_documentation/domain_service/navigation/DocumentationNavigationPageMapper.java
konrad-krol-incubly: The issue was pre-exisiting, but I made refactoring as requested. I removed `DocumentationNavigationPageMapper` class moved this logic.

Regarding field value mutations - I enclosed all changes within the methods of `PortalNavigationPage` class itself, so right now its better controlled. Please note there are many setter-driven mutations across codebase that are invoked from other APIM features e.g. `PortalNavigationItemDomainService` uses setter directly during `reorderOriginSiblings`, so we fixed just small percent.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationPage.java
```diff
@@ -49,4 +52,46 @@ public final class PortalNavigationPage extends PortalNavigationItem {
     public PortalNavigationItemType getType() {
         return TYPE;
     }
+
+    public static PortalNavigationPage from(
```
dev-marek: This method only maps parameters to `CreatePortalNavigationItem` and executes `PortalNavigationItem.from()`. It would make more sense to construct create object in sync domain service and call `from` directly

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationPage.java
```diff
@@ -49,4 +52,46 @@ public final class PortalNavigationPage extends PortalNavigationItem {
...
+            .reference(meta.reference())
+            .visibility(PortalVisibility.PUBLIC)
+            .published(true)
+            .build();
+        return (PortalNavigationPage) PortalNavigationItem.from(create, organizationId, environmentId, parent);
+    }
+
+    public void update(
+        @Nonnull AutomationMetadata meta,
+        @Nonnull PortalArea area,
```
dev-marek: Area should never be changed

dev-marek (approved):
Please fix those small issues before merging

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationPage.java
```diff
@@ -49,4 +52,15 @@ public final class PortalNavigationPage extends PortalNavigationItem {
     public PortalNavigationItemType getType() {
         return TYPE;
     }
+
+    public void update(@Nonnull AutomationMetadata meta, @Nullable PortalNavigationItemContainer parent, @Nonnull Slug segment) {
```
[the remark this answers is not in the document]
konrad-krol-incubly: this is already established pattern, it doesn't make sense to overcomplicate it

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/mapper/ApiDocumentationMapper.java
```diff
@@ -94,4 +96,10 @@ default PortalPageContentType toDomainType(io.gravitee.apim.rest.api.automation.
     default io.gravitee.apim.rest.api.automation.model.DocumentationType toWireType(PortalPageContentType domain) {
         return domain == null ? null : io.gravitee.apim.rest.api.automation.model.DocumentationType.fromValue(domain.name());
     }
+
+    default io.gravitee.apim.rest.api.automation.model.DocumentationArea toWireArea(
+        io.gravitee.apim.core.portal_page.model.PortalArea domain
+    ) {
+        return domain == null ? null : io.gravitee.apim.rest.api.automation.model.DocumentationArea.fromValue(domain.name());
```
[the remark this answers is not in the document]
konrad-krol-incubly: IMO this way its more compact

PR #19016, opened 2026-08-06
refactor(gateway): cache API Product subscriptions under their scope instead of exploding them

phiz71:
## Issue

No Jira ticket for this one — it started from the production incident behind #19000
and stayed on that thread. The behaviour it touches is specified by
[APIM-12491](https://gravitee.atlassian.net/browse/APIM-12491), whose acceptance criteria
the added tests now cover.

## Description

An API Product subscription was exploded into one runtime subscription per member
API, so the gateway caches grew with (subscriptions × member APIs). On the affected
installation, 36,487 logical subscriptions became roughly 5.5M runtime entries and
the gateway ran out of memory at startup.

This changes what a cache entry represents. A subscription is now stored once, under
the entity it was taken on — its *scope*: the API for a regular subscription, the API
Product for a product one. The `SubscriptionService` contract stays API-scoped, so
lookups still arrive with an API and this resolves the products that API belongs to,
through a reverse index bounded by (APIs × products) — a few thousand entries.

Flattening never bought anything at read time: the request path already knew about
products, since `HttpSecurityChain` builds a product plan's security plan from that
same registry. The expansion only existed to fit index keys designed before API
Products.

Removing it moves ownership as a consequence. The API synchronizer no longer sees
product subscriptions at all — `SubscriptionAppender` collects only each API's own
plans — and `ApiProductDeployer` becomes their single writer: a paginated refresh on
deploy, a bulk eviction by scope on undeploy. That removes the cold-start double
registration by construction, with no `initialSync` flag. Dropping an API from a
product needs no eviction either: the subscription stays cached under the product and
the registry simply stops resolving that API to it.

Four commits fix defects found while working in this area, each independent of the
scope change:

- the member API resync ran during the initial synchronization, where it is pure
  duplication (API Products are synchronized before APIs), and loaded every member
  API's event through an unpaginated search;
- `apiLocks` grew by one `ReentrantLock` per API ever seen, never cleaned up;
- the resync trigger could leave a terminated disposable in its composite forever;
- the product plan index was rebuilt from a **shallow** copy, so adding or removing a
  product mutated, in place, the lists that in-flight requests were walking to build
  their security chain. This is what SonarCloud's `java:S3077` was pointing at. The
  rebuild now works on its own copy and publishes frozen lists.

## Additional context

**Test coverage was the weak point, and it has changed.** The SDK exposes
`SubscriptionService` and `ApiKeyService` as Mockito mocks, so no integration test ever
exercised the real caches: helpers like `denyKeyForApi` stub them, and a scenario
asserting a 401 after detaching an API from a product was really asserting that a mock
told to return empty causes a 401. Three commits address this:

- `ApiProductV4RealCacheIntegrationTest` routes both beans to real cache instances and
  asserts what the gateway itself resolves — one subscription taken on the product is
  reachable from every member API and from none outside it, detaching stops resolution
  for that API alone, reattaching restores it with no new key, undeploying the product
  evicts everywhere at once;
- the same class covers a product plan and an API plan coexisting on one API, and the
  fallback to the API's own plan when nothing matches on the product scope;
- `ApiProductV4ShardingIntegrationTest` covers sharding tags, which had no integration
  coverage at all.

The delegation is wired with `doAnswer`, never `when(...)`: on a partial mock the latter
invokes the delegate while stubbing, with the nulls argument matchers produce — which is
what made a first attempt at flipping the SDK default fail 236 tests. It is private to the
test rather than added to `AbstractGatewayTest`, so **the SDK is untouched** and nothing
changes for the other 80 integration tests or for plugin repositories.

**An asymmetry surfaced while writing the sharding tests**, and is recorded rather than
changed. A product follows the same rule as an API: a tagged gateway refuses it when
untagged. A product *plan* does not — an untagged one is accepted outright, where an
untagged API plan goes through the ordinary rule and can make the gateway skip its API.
The class javadoc of `ApiProductShardingFilter` now states both rules and names the test
pinning each side, so a bug report about a product plan appearing on the wrong gateway
lands there directly. Behaviour is unchanged; whether to align it is a product question.

**Behaviour change.** Two subscriptions sharing an id and an API across environments no
longer coexist — last write wins. Lookups never carried an environment, so the extra legs
were never observable; `getById` already returned an arbitrary one.

**One acceptance criterion remains untested.** APIM-12491 states that API-level plan
checks are skipped when a product subscription matches. That is not observable from a
status code — the chain stopping at the product plan and the API plan answering both
yield a 200 — and proving it means observing the retained plan, which the SDK does not
expose. The ordering itself is covered where it is decided, in `HttpSecurityChain`'s
construction.

**Still open, deliberately out of scope.** `8cee817b82` also widened deployment
eligibility: an API whose tags do not match the gateway becomes eligible as soon as it
belongs to a served product. That is a product decision and is untouched here, but it is
the most likely explanation for the live-set size seen on the affected node, and it
deserves its own discussion. `should_not_let_the_api_own_tags_decide_product_resolution`
now pins the current behaviour down either way.

**Review order.** The commits are meant to be read in sequence and each one builds and
passes on its own. `SubscriptionScope` first (a shared helper, no caller), then the cache
change — necessarily atomic, since the mapper cannot stop exploding before the caches know
how to index by scope — then the ownership move, then the fixes, then the tests.

**Validation.** 2,118 unit tests across `gateway-core`, `gateway-security-core`,
`gateway-handlers-api` and `gateway-services-sync`, and the full integration suite at
1,325 tests — up from 1,309 before this branch, with no regression. Prettier applied.

[APIM-12491]: https://gravitee.atlassian.net/browse/APIM-12491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
@@ -55,6 +55,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import lombok.CustomLog;
 import org.slf4j.MDC;
 
@@ -73,7 +74,15 @@ public class ApiManagerImpl implements ApiManager {
     private final GatewayConfiguration gatewayConfiguration;
     private final LicenseManager licenseManager;
     private final Map<String, ReactableApi<?>> apis = new ConcurrentHashMap<>();
-    private final ConcurrentHashMap<String, Lock> apiLocks = new ConcurrentHashMap<>();
+    /**
+     * Fixed set of locks, one API mapped onto one stripe by hash. A map keyed by API id would grow
+     * with every API ever seen — nothing can safely evict an entry, since removing a lock another
+     * thread is about to acquire hands the two of them different locks for the same API. Striping
+     * bounds the memory instead: two APIs may share a stripe and serialize needlessly, which costs
+     * nothing outside deployment.
+     */
+    private static final int API_LOCK_STRIPES = 64;
+    private final Lock[] apiLocks = Stream.generate(ReentrantLock::new).limit(API_LOCK_STRIPES).toArray(Lock[]::new);
     private final Map<Class<? extends ReactableApi<?>>, ? extends Deployer<?>> deployers;
     private final ApiProductRegistry apiProductRegistry;
 
@@ -464,7 +473,7 @@ public class ApiManagerImpl implements ApiManager {
     }
 
     private <T> T withPerApiLock(String apiId, Callable<T> action) {
-        Lock lock = apiLocks.computeIfAbsent(apiId, k -> new ReentrantLock());
+        Lock lock = apiLocks[Math.floorMod(Objects.hashCode(apiId), API_LOCK_STRIPES)];
         lock.lock();
         try {
             return action.call();
--- a/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
@@ -20,14 +20,16 @@ import static io.gravitee.repository.management.model.Subscription.Status.ACCEPT
 import io.gravitee.gateway.api.service.ApiKeyService;
 import io.gravitee.gateway.api.service.Subscription;
 import io.gravitee.gateway.api.service.SubscriptionService;
+import io.gravitee.gateway.core.subscription.SubscriptionScope;
+import io.gravitee.gateway.handlers.api.ReactableApiProduct;
 import io.gravitee.gateway.handlers.api.manager.ApiManager;
+import io.gravitee.gateway.handlers.api.registry.ApiProductRegistry;
 import io.gravitee.gateway.reactive.api.policy.SecurityToken;
 import io.gravitee.gateway.reactive.handlers.api.v4.Api;
 import io.gravitee.gateway.reactor.ReactableApi;
 import io.gravitee.gateway.security.core.SubscriptionTrustStoreLoaderManager;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
-import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
@@ -44,6 +46,26 @@ import lombok.CustomLog;
 import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
 
+/**
+ * Runtime cache of active subscriptions.
+ *
+ * <h2>Scope</h2>
+ *
+ * A subscription is cached once, under the entity it was taken on — its <em>scope</em>: the API for
+ * a regular subscription, the API Product for an API Product subscription. It is never duplicated
+ * per member API.
+ *
+ * <p>The {@link SubscriptionService} contract is API-scoped, so lookups arrive with an API and this
+ * class resolves the scopes to probe: the API itself, then the products that API belongs to
+ * ({@link ApiProductRegistry#getApiProductIdsForApi(String)}). That reverse index is bounded by
+ * (APIs x products) — a few thousand entries — where caching one subscription per member API grows
+ * with (subscriptions x member APIs) and reaches millions of entries on large products.</p>
+ *
+ * <p>The request path already knows about products: {@code HttpSecurityChain} builds a product
+ * plan's security plan from the same registry. Flattening product subscriptions onto member APIs
+ * therefore bought nothing at read time; it only existed to fit index keys designed before API
+ * Products existed.</p>
+ */
 @CustomLog
 @RequiredArgsConstructor
 public class SubscriptionCacheService implements SubscriptionService {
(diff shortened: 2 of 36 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/services/SubscriptionCacheService.java
```diff
@@ -44,36 +46,59 @@
...
 @RequiredArgsConstructor
 public class SubscriptionCacheService implements SubscriptionService {
 
     private final ApiKeyService apiKeyService;
     private final SubscriptionTrustStoreLoaderManager subscriptionTrustStoreLoaderManager;
     private final ApiManager apiManager;
+    private final ApiProductRegistry apiProductRegistry;
 
-    // Caches only contains active subscriptions.
+    // Caches only contain active subscriptions.
```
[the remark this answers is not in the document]
phiz71: @AikidoSec ignore: this comment states an invariant, not a restatement of the code. The caches hold only ACCEPTED subscriptions — `register()` routes anything else to `unregister()` — which is why no reader filters on status anywhere. That is not inferable from the field names, and dropping the line would invite a future reader to add a redundant status check on the request path.

The line only shows up in this diff because I fixed its grammar ("contains" -> "contain"); the invariant itself predates the PR.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
```diff
@@ -46,29 +45,13 @@ public Completable deploy(ApiProductReactorDeployable deployable) {
...
-                        Set.of(reactableApiProduct.getEnvironmentId())
-                    )
-                );
-            }
-        }
-        doBeforeEmit = doBeforeEmit.andThen(Completable.fromRunnable(() -> registerApiProductPlans(deployable)));
-        if (newApiIds != null && !newApiIds.isEmpty()) {
+        // Removing an API from the product needs no subscription eviction: a product subscription is
+        // cached under the product, and the registry stops resolving the removed API to it, so the
+        // API can no longer reach it. Only the product's own undeployment evicts.
```
jhaeyaert, line 52: To clarify what it means. I'm not clear on this if the subscription isn't evicted.
Removing an API from an API Product does not mean undeploying the API, so do we have an integration test that ensures this case works (e.g., the call is rejected if the API is no longer part of the API Product)?

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
```diff
@@ -73,7 +74,15 @@ public class ApiManagerImpl implements ApiManager {
...
-    private final ConcurrentHashMap<String, Lock> apiLocks = new ConcurrentHashMap<>();
+    /**
+     * Fixed set of locks, one API mapped onto one stripe by hash. A map keyed by API id would grow
+     * with every API ever seen — nothing can safely evict an entry, since removing a lock another
+     * thread is about to acquire hands the two of them different locks for the same API. Striping
+     * bounds the memory instead: two APIs may share a stripe and serialize needlessly, which costs
+     * nothing outside deployment.
+     */
+    private static final int API_LOCK_STRIPES = 64;
+    private final Lock[] apiLocks = Stream.generate(ReentrantLock::new).limit(API_LOCK_STRIPES).toArray(Lock[]::new);
```
jhaeyaert, line 85: Looks more like the job of a Cache with idle eviction, but fine for me

jhaeyaert (changes requested):

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
phiz71: Good catch, and the honest answer is that the integration test you would expect to cover this does not actually cover it.

**What happens.** Removing an API from a product does not undeploy the API, right. What changes is resolution: the product is redeployed with a shorter `apiIds`, `ApiProductRegistry` rebuilds its index, and `getApiProductIdsForApi(removedApi)` stops returning that product. The subscription stays cached under the product — still valid for the remaining member APIs — but the removed API can no longer reach it, so the call is rejected. The reactor of the removed API also refreshes its security chain, because `publishApiProductUpdateEvent` sends the union of previous and current `apiIds`, so the product plan disappears from its chain too.

**Why the existing test proves nothing.** `DeletionScenarios.should_return_401_when_api_is_detached_from_product_but_api_itself_is_still_deployed` reads exactly like the test you are asking for, but `allowKeyForApi`/`denyKeyForApi` stub `ApiKeyService` and `SubscriptionService` with Mockito. The test *tells* the mock to return empty and then asserts a 401 — it never exercises the registry or the cache. Same for the other product scenarios in that class.

**What is actually covered**, in two unit tests, one per link in the chain:
- `ApiProductRegistryImplTest.should_drop_the_api_when_it_leaves_the_product` — after a redeploy without the API, `getApiProductIdsForApi` no longer returns the product.
- `SubscriptionCacheServiceTest$ApiProductScopeTest.should_not_reach_product_subscription_from_an_api_outside_the_product` — an API the registry does not resolve to the product finds nothing.

So the behaviour is verified link by link, but never end to end through a real request. That is the test gap I flagged in the PR description, and your comment lands right on it.

**Getting a real one** is not free: the SDK exposes `SubscriptionService` as a Mockito mock, so its fields are never initialised and `register()` cannot run on it. `ApiProductV4SecurityIntegrationTest` works around this with `thenCallRealMethod()` plus field injection, which is viable for the certificate path because it delegates straight to the trust store, but the api-key path needs the real cache maps. Making this work means having the SDK hand out a real `SubscriptionCacheService`.

Happy to do it in this PR if you want the coverage now, or to track it separately — it is a change to the test SDK rather than to this refactor. Your call.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/manager/impl/ApiManagerImpl.java
phiz71: Agreed that a cache with idle eviction is the shape one reaches for first — I went with striping precisely because eviction is what makes this one unsafe.

Evicting a lock is not like evicting a cached value: if the entry for an API disappears while a thread is about to acquire it, the next caller creates a fresh lock and the two run `deploy` and `undeploy` on the same API concurrently — the exact race the lock exists to prevent. Making that safe means keeping the entry alive for as long as anyone might take it, which is roughly reference counting, and that is a lot of machinery for what is a startup-time concern.

A fixed table bounds the memory with none of that: two APIs may share a stripe and serialize needlessly, which costs nothing outside deployment, and 64 stripes sit well above the deployer parallelism.

Worth noting this only matters as long as there are two concurrent writers on the API lifecycle — the sync pipeline and the product-driven resync. If the eligibility widening from 8cee817b82 is rolled back, the second writer disappears and the whole table can go with it.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductDeployer.java
jhaeyaert: I'm fine, just to keep it in mind for a global work on the integration tests coverage.

phiz71: Ended up doing it here rather than deferring — and it turned up enough about the current coverage to be worth feeding into that global work.

`ApiProductV4RealCacheIntegrationTest` now runs against real `SubscriptionCacheService` and `ApiKeyCacheService` instances and asserts what the gateway resolves on its own: a product key opens API 1 and API 2, the product is redeployed without API 1, and from then on API 1 answers 401 while API 2 keeps answering 200. Sibling scenarios cover reattaching an API, undeploying the product, an API outside any product staying untouched, and a product plan coexisting with the API's own plan.

Three things that seem relevant to a broader review of integration coverage:

1. **The test you were expecting already existed by name.** `DeletionScenarios.should_return_401_when_api_is_detached_from_product_but_api_itself_is_still_deployed` reads exactly like it — but `denyKeyForApi` stubs the mock, so it asserts that a mock told to return empty causes a 401. Same for the other product scenarios in that class. They look like end-to-end coverage and are tautologies. This is not specific to API Products: `GatewayTestContainer` exposes `SubscriptionService` and `ApiKeyService` as bare mocks for every integration test, which is fine for plugin tests and misleading for gateway ones.

2. **Routing those beans to real instances is doable but has a trap.** It has to be `doAnswer`, never `when(...)`: on a partial mock the latter invokes the delegate while stubbing, with the nulls the matchers produce. Flipping the SDK default to a spy failed 236 tests for exactly that reason, so the helper stays private to the test and the SDK is untouched. If the global work wants this available everywhere, that is the constraint to design around.

3. **Sharding had no integration coverage at all**, so `ApiProductV4ShardingIntegrationTest` is new too. Writing it surfaced an asymmetry: a product without tags is refused by a tagged gateway, like an API — but a product *plan* without tags is accepted outright, where an untagged API plan goes through the ordinary rule. Deliberate in the code, never documented. Behaviour left alone; both rules are now in the class javadoc of `ApiProductShardingFilter` with the test pinning each side, so a bug report lands there directly.

One criterion of APIM-12491 stays uncovered and I do not think it can be reached from a status code: *API-level plan checks are skipped when a product subscription matches*. Both paths return 200; proving the order means observing the retained plan, which the SDK does not expose. The ordering itself is covered where it is decided, in `HttpSecurityChain`'s construction.

jhaeyaert (approved):

PR #19034, opened 2026-08-07
chore(deps): bump graphene to 3.8.0 and gamma modules to matching releases

jourdiw:
Bumps Graphene to **3.8.0** and the four gamma module versions, **in a single commit**.

Supersedes the bot PRs #18997 (aim), #18791 (authz), #18373 (edge) and #18079 (esm), which I converted to draft. Those can be closed once this merges.

## Why this has to be one commit

`gamma-console` shares `@gravitee/graphene-core` as a **strict singleton** — `singleton: true, strictVersion: true`, with `requiredVersion` taken from this repo's own (exact) declaration, in `gravitee-gamma/gravitee-gamma-control-plane-webui/module-federation.config.ts`.

All four modules are now built against graphene 3.8.0. If the pom moved first, a remote would register 3.8.0 into the shared scope, the higher version would win, this repo's exact `3.4.0` requirement would no longer be satisfied, and `strictVersion: true` turns that into a throw at load rather than a warning. Bumping graphene first without the pom is harmless but pointless. So: both, together, or neither.

That is the only reason the bot PRs were held — each is individually correct and individually unmergeable.

## Changes

**Graphene → 3.8.0** (exact, as this repo pins):
- `package.json`
- `gravitee-gamma/gravitee-gamma-module-apim/package.json`
- `gravitee-gamma/gravitee-gamma-module-platform/package.json` (core only)

**`gravitee-apim-distribution/pom.xml`:**

| Module | From | To |
|---|---|---|
| aim | 2.19.2 | 2.20.0 |
| authz | 1.10.1 | 1.13.0 |
| edge | 1.1.0 | 1.3.0 |
| esm | 1.0.0 | 1.4.0 |

edge stays on its **1.x** line here. Its 2.0 pre-release line exists and has the same graphene bump (2.0.0-alpha.4), but `master` is not adopting it in this PR.

Every one of those versions is published and verified green through its artifact pipeline — not merely tagged.

## What's in graphene 3.8.0

The response to the August Policy Studio affordance audit, plus items that reach every consumer through the base layer:

- `core`: **pointer cursor restored on all buttons** via the base layer. Tailwind v4 dropped Preflight's `button { cursor: pointer }` and nothing had restored it, so no control in the studio rendered a pointer. This one reaches every Gamma module, not just the studio.
- `core`: spacing scale gains 40/44/56; deterministic Tier 1 class-portability check
- `policy-studio`: focus ring restored on catalog rows; drag handles moved onto the real interactive element; Remove step weight matched to its siblings
- `charts`: `DoughnutChart` `centerStat` prop
- icons: OpenApi, Claude/Cursor brand, ClipboardCheck, Code2, Layers, Link2, ShieldX

No breaking changes in the range — the only `BREAKING CHANGES` entry in the 3.x line is 3.0.0, which this repo is already past.

## Review notes

- All four modules also moved `graphene-core` from exact pins to `^` on their side, so their federation `requiredVersion` now accepts whatever this repo provides rather than demanding one version. That is what makes the shared singleton actually shared.
- `yarn install` reports `doesn't provide react-hook-form, requested by @gravitee/graphene-core`. **Pre-existing** — 3.4.0 and 3.8.0 declare identical peer sets, and the root workspace did not list `react-hook-form` before this change either. Not introduced here, but worth a separate look.
- Graphene 3.8.0 produces 28 Chromatic snapshot changes in the aim module. They are pending review there and are not visible to this repo's CI.
- A companion PR follows for `4.12.x`, with a different module set (aim 1.9.1, edge 1.3.0; authz and esm deliberately unchanged). **Please don't backport this one** — the module versions differ per branch.
- **Changed after first approval:** edge was originally set to `2.0.0-alpha.4` and is now `1.3.0`. Squashed back into the single commit, so the branch history stays clean; the only content difference from the approved version is that one pom line. The graphene half is untouched.


callaertanthony (approved):

jourdiw:
Heads-up @callaertanthony — pushed a change after your approval, so worth a re-look before this queues.

`gravitee-gamma-module-edge` is now **1.3.0** (its 1.x line) instead of `2.0.0-alpha.4`. `master` is not adopting edge's 2.0 pre-release line in this PR after all.

Second commit, `gravitee-apim-distribution/pom.xml` only. The graphene half is untouched, and both edge lines carry the same graphene 3.8.0 bump, so the singleton reasoning in the description is unaffected.

jourdiw:
Squashed the edge correction back into the single commit — force-pushed, so the branch is one commit again (`81cdf3b`).

Content is unchanged from my previous comment: the only difference from the version @callaertanthony approved is `gravitee-gamma-module-edge` `2.0.0-alpha.4` → `1.3.0` in `gravitee-apim-distribution/pom.xml`. Everything else, including the whole graphene half, is identical.

PR #19037, opened 2026-08-07
feat(mapi): gamma reset password

pragsty:
## Issue

https://gravitee.atlassian.net/browse/FOUND-87

## Description
Adds backward-compatible Management API support for the Gamma password reset flow. Classic console behavior is unchanged when resetTarget is omitted.

**Changes**
Password policy endpoint — GET /management/organizations/{orgId}/configuration/password-policy
Returns description from user.password.policy.description in gravitee.yml
Returns pattern and a structured rules[] array parsed from user.password.policy.pattern
Public (unauthenticated) access, same pattern as other registration config endpoints
Admin reset email target — POST /management/organizations/{orgId}/users/{userId}/resetPassword?resetTarget=gamma
Optional resetTarget query parameter (replaces request-body entity)
When resetTarget=gamma, password reset emails link to the Gamma console reset page (/reset-password/:token)
When omitted, existing classic APIM Console reset URL is used (no breaking change)
Unknown resetTarget values return HTTP 400 via ValidationDomainException
Service layer — UserServiceImpl.resetPasswordWithTarget() builds the Gamma reset URL from trusted installation settings
Existing guards unchanged — UserNotInternallyManagedException for external IDP users, ServiceAccountNotManageableException for service accounts, PASSWORD_RESET audit event, email notification

https://github.com/user-attachments/assets/6308368e-52c6-4eb9-947f-d05b8518b7bc

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/PasswordPolicyPatternParser.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/PasswordPolicyPatternParser.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.service.impl;
+
+import io.gravitee.rest.api.model.PasswordPolicyRuleEntity;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Derives UI password requirement rules from {@code user.password.policy.pattern} in gravitee.yml.
+ */
+public class PasswordPolicyPatternParser {
+
+    private static final String POSITIVE_LOOKAHEAD_PREFIX = "(?=.*[";
+    private static final Pattern LENGTH_QUANTIFIER = Pattern.compile("\\.\\{(\\d+)(?:,(\\d+)?)?\\}\\$?");
+
+    public List<PasswordPolicyRuleEntity> parse(String policyPattern) {
+        if (policyPattern == null || policyPattern.isBlank()) {
+            return List.of();
+        }
+
+        List<PasswordPolicyRuleEntity> rules = new ArrayList<>();
+        Set<String> seenRuleIds = new LinkedHashSet<>();
+
+        Matcher lengthMatcher = LENGTH_QUANTIFIER.matcher(policyPattern);
+        if (lengthMatcher.find()) {
+            int minLength = Integer.parseInt(lengthMatcher.group(1));
+            String maxLengthGroup = lengthMatcher.group(2);
+            addRule(
+                rules,
+                seenRuleIds,
+                PasswordPolicyRuleEntity.builder()
+                    .id("minLength")
+                    .label("At least " + minLength + " characters")
+                    .pattern(String.format("^.{%d,}$", minLength))
+                    .build()
+            );
+            if (maxLengthGroup != null && !maxLengthGroup.isBlank()) {
+                int maxLength = Integer.parseInt(maxLengthGroup);
+                addRule(
+                    rules,
+                    seenRuleIds,
+                    PasswordPolicyRuleEntity.builder()
+                        .id("maxLength")
+                        .label("At most " + maxLength + " characters")
+                        .pattern(String.format("^.{0,%d}$", maxLength))
+                        .build()
+                );
+            }
+        }
+
+        for (String charClass : extractPositiveLookaheadCharClasses(policyPattern)) {
+            classifyLookahead(charClass).ifPresent(rule -> addRule(rules, seenRuleIds, rule));
+        }
+
+        if (policyPattern.contains("(?!.*(.)\\1{2,})")) {
+            addRule(
+                rules,
+                seenRuleIds,
+                PasswordPolicyRuleEntity.builder()
+                    .id("noConsecutive")
+                    .label("No more than 2 consecutive equal characters")
+                    .pattern("^(?!.*(.)\\1{2,}).+$")
+                    .build()
+            );
+        }
+
+        return List.copyOf(rules);
+    }
+
+    private static List<String> extractPositiveLookaheadCharClasses(String policyPattern) {
+        List<String> charClasses = new ArrayList<>();
+        int index = 0;
+
+        while ((index = policyPattern.indexOf(POSITIVE_LOOKAHEAD_PREFIX, index)) >= 0) {
+            int start = index + POSITIVE_LOOKAHEAD_PREFIX.length();
+            int end = findCharClassEnd(policyPattern, start);
+            if (end > start) {
+                charClasses.add(policyPattern.substring(start, end));
+                index = end + 1;
+            } else if (end >= 0) {
+                index = end + 1;
+            } else {
+                index += POSITIVE_LOOKAHEAD_PREFIX.length();
+            }
+        }
+
+        return charClasses;
+    }
+
+    private static int findCharClassEnd(String policyPattern, int start) {
+        for (int i = start; i < policyPattern.length(); i++) {
+            if (policyPattern.charAt(i) == '\\' && i + 1 < policyPattern.length()) {
+                i++;
+                continue;
+            }
+            if (policyPattern.charAt(i) == ']') {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    private static void addRule(List<PasswordPolicyRuleEntity> rules, Set<String> seenRuleIds, PasswordPolicyRuleEntity rule) {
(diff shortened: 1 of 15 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/PasswordPolicyPatternParser.java
```diff
@@ -0,0 +1,151 @@
...
+        List<String> charClasses = new ArrayList<>();
+        int index = 0;
+
+        while ((index = policyPattern.indexOf(POSITIVE_LOOKAHEAD_PREFIX, index)) >= 0) {
+            int start = index + POSITIVE_LOOKAHEAD_PREFIX.length();
+            int end = findCharClassEnd(policyPattern, start);
+            if (end > start) {
+                charClasses.add(policyPattern.substring(start, end));
+            }
+            index = end + 1;
```
aqibmohammadkhan: `PasswordPolicyPatternParser.extractPositiveLookaheadCharClasses` — Aikido is right about the hang -> their suggested fix is wrong.

```
int end = findCharClassEnd(policyPattern, start);
if (end > start) {
    charClasses.add(...);
}
index = end + 1;  // end == -1 → index = 0 → loops forever
```
Any pattern with (?=.*[ and no closing ] hangs the unauthenticated GET .../password-policy endpoint (DoS).

Fix: break (or advance past the prefix) when end < 0. Do not set index = start.

Add a unit test for unbalanced [.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -1399,7 +1401,38 @@ private String formatDeletedSourceId(String userId, String sourceId) {
...
+
+        if ("gamma".equalsIgnoreCase(resetTarget.trim())) {
+            doResetPassword(executionContext, id, buildGammaResetPasswordPageUrl(executionContext.getOrganizationId()));
+            return;
+        }
+
+        throw new ValidationDomainException("Unsupported password reset target: " + resetTarget);
+    }
+
+    private String buildGammaResetPasswordPageUrl(final String organizationId) {
```
aqibmohammadkhan, line 1430: Multi-tenant NPE on missing Gamma URL: 
`buildGammaResetPasswordPageUrl` calls `gammaUrl.endsWith("/")` with no null check. In multi-tenant mode getGammaUrl() can return null via buildHttpUrl(null). That turns a misconfigured install into a 500 on admin reset.

Guard null/blank and fail with a clear domain exception (or fall back consistently).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceResetPasswordTargetTest.java
aqibmohammadkhan, line 1: Thin coverage of the actual Gamma email URL -> 
UserServiceResetPasswordTargetTest only reflects into the private builder and checks unknown targets. It never asserts that doResetPassword / getTokenRegistrationParams produce
{gammaUrl}/reset-password/{token}.

Worth one test that stubs installation + exercises resetPasswordWithTarget(..., "gamma") and verifies the notification/email registrationUrl.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/PasswordPolicyServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+
+@Component
+public class PasswordPolicyServiceImpl implements PasswordPolicyService {
+
+    private static final String DEFAULT_DESCRIPTION =
+        "Password must be at least 12 characters long, contain at least one digit, one upper case letter, one lower case letter, one special character, and no more than 2 consecutive equal characters.";
+
+    private final PasswordPolicyPatternParser patternParser = new PasswordPolicyPatternParser();
+
+    @Value("${user.password.policy.description:}")
```
aqibmohammadkhan, line 32: When user.password.policy.description is empty, the hardcoded default still says “at least 12 characters” even if the pattern is .{10,64}. The service test locks that in. Gamma UI can show conflicting copy vs rules[].

Either derive description from rules, or document that description must be set when the pattern chang

aqibmohammadkhan (commented):
Minor

- Commit/PR title feat(gateway) is misleading — this is Management REST API.
- Aikido’s URL-trim duplication note is fair but low priority; extract a tiny helper if you touch that area.
- Parser captures max length in the regex group but never emits a maxLength rule.
- Rule labels are hardcoded English (fine if Gamma owns i18n).
- Org-scoped path reads global gravitee.yml (consistent with how the policy is configured today).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
pragsty: done

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceResetPasswordTargetTest.java
pragsty: done

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/PasswordPolicyServiceImpl.java
pragsty: doen

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/PasswordPolicyPatternParser.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Derives UI password requirement rules from {@code user.password.policy.pattern} in gravitee.yml.
+ */
+public class PasswordPolicyPatternParser {
+
+    private static final String POSITIVE_LOOKAHEAD_PREFIX = "(?=.*[";
+    private static final Pattern LENGTH_QUANTIFIER = Pattern.compile("\\.\\{(\\d+)(?:,(\\d+)?)?\\}\\$?");
```
aqibmohammadkhan, line 32: maxLength still parsed but never emitted.

LENGTH_QUANTIFIER captures group 2 (128 in .{12,128}), but only minLength is added (.{12,}). Checklist can go all-green for a password that the full backend pattern rejects.
Emit a maxLength rule when group 2 is present, or include max in the minLength pattern.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/PasswordPolicyServiceImpl.java
```diff
@@ -0,0 +1,58 @@
...
+@Component
+public class PasswordPolicyServiceImpl implements PasswordPolicyService {
+
+    private final PasswordPolicyPatternParser patternParser = new PasswordPolicyPatternParser();
+
+    @Value("${user.password.policy.description:}")
+    private String passwordPolicyDescription;
+
+    @Value(
+        "${user.password.policy.pattern:^(?=.*[0-9])(?=.*[A-Z])(?=.*[a-z])(?=.*[!~<>.,;:_=?/*+\\-#\\\"'&§`£€%°()|\\[\\]$^@])(?!.*(.)\\1{2,}).{12,128}$}"
```
aqibmohammadkhan, line 36: Unparseable custom patterns → empty rules[] → Gamma hard-blocks.

Parser only understands lookahead/{min,max} style. A custom user.password.policy.pattern that doesn’t match yields rules: []. 

With [#19038](https://github.com/gravitee-io/gravitee-api-management/pull/19038)’s empty-rules handling, reset stays disabled forever even though pattern is present.

Prefer a fallback rule using the full pattern when derivation yields nothing.

aqibmohammadkhan (commented):
Title feat(gateway) is wrong - this is Management REST API.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/PasswordPolicyPatternParser.java
pragsty: Fixed. When the length quantifier includes a max (e.g. .{12,128}), we now emit a dedicated maxLength rule with anchored pattern ^.{0,128}$ alongside minLength ^.{12,}$. UI checklist can no longer go all-green for passwords the full backend pattern would reject on length.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/PasswordPolicyServiceImpl.java
pragsty: Fixed. If rule derivation yields nothing but user.password.policy.pattern is configured, PasswordPolicyServiceImpl now returns a single fallback rule using the full pattern (id: policyPattern). Gamma evaluates that rule directly, so custom/unparseable patterns no longer hard-block reset with rules: [].

aqibmohammadkhan (approved):

PR #19042, opened 2026-08-07
feat(observability): add Gamma dashboards write endpoints

SamHardi:
## Summary
Adds the three write verbs (POST / PUT / DELETE) to the Gamma dashboards API so a dashboard author's work is stored server-side and shared with their team rather than trapped in one browser. Builds on the read endpoints from OBS-15 and the persistence layer from OBS-14.

Jira: [OBS-16](https://gravitee.atlassian.net/browse/OBS-16)

## Changes
- Three new use cases (`Create`/`Update`/`Delete ObservabilityDashboardUseCase`) deriving every server-owned field: `version` starts at 1 and increments on update, `createdBy` comes from the authenticated principal, timestamps are stamped server-side; anything a client sends for them is ignored.
- `DashboardContent` core model carrying the author-editable part of a dashboard, with structural write guardrails: blank title, more than 50 widgets, oversized (~1 MiB) widgets payload, blank/duplicate widget ids, incoherent time range → 400.
- `DashboardRepository` port extended with `create`/`update`/`delete` (OBS-15 shipped it read-only) and `GammaDashboardRepositoryAdapter` gains the reverse anticorruption mapping core `Dashboard` → persisted `GammaDashboard` (lowercase operators, serialized widgets, `Instant` → `Date`).
- Request DTOs (`SaveDashboardRequestDto`, `SaveDashboardFilterDto`): polymorphic filter `value` (scalar or array) normalized by the new shared `FilterValues` helper (promoted from the logs `FilterConditionDto` instead of copied), `editable` defaulting to `false` (locked) when absent.
- OpenAPI: the three paths plus `SaveDashboardRequest`/`SaveDashboardFilter` schemas, `maxItems: 50` documented on `widgets`.

### Endpoints
| --- | --- | --- | --- | --- |
| `/gamma/organizations/{orgId}/environments/{envId}/observability/dashboards` | POST | New — `ENVIRONMENT_DASHBOARD:CREATE` | `{ "id": "d7e2…", "title": "Perf", "widgets": [{"id":"w1"}] }` | `201` + `Location`, body with `version: 1` |
| `…/dashboards/{dashboardId}` | PUT | New — `ENVIRONMENT_DASHBOARD:UPDATE` | `{ "title": "Renamed", "timeRange": {"type":"absolute","from":1000,"to":2000} }` | `200`, body with incremented `version`, preserved `createdAt` |
| `…/dashboards/{dashboardId}` | DELETE | New — `ENVIRONMENT_DASHBOARD:DELETE` | — | `204` |

## Reviewer notes
- **POST and PUT, not a PUT upsert**: `@Permissions` is per method and CREATE / UPDATE are distinct ACLs — collapsing them would silently grant creation to anyone holding update rights.
- **Version is maintained, not enforced**: a stale `version` does not fail — behaviour stays last-write-wins. The 409 guard (and turning the Mongo read-then-save into a single conditional operation) is OBS-17; a partial check here would advertise a guarantee that still has a TOCTOU hole.
- **Filter names are deliberately not validated against the `StaticFilters` catalogue at write time**: a dashboard is declarative configuration, the analytics query path already validates at read time, and catalogue coupling would make dashboards non-importable across environments.
- Cross-environment ids resolve to **404 (not 403)** on every verb, so existence cannot be probed.

## Test plan
- [ ] `mvn test -pl gravitee-gamma-rest-api` from `gravitee-gamma/` — 316 tests green (45 new: use cases, `DashboardContent` validation matrix, adapter write mapping, Jersey resource tests)
- [ ] POST a dashboard with a client-supplied id → `201`, `Location` header, `version: 1`, then `GET /` lists it
- [ ] POST the body with `version: 42` / `createdBy` → both ignored in the response
- [ ] PUT → `200`, `createdAt`/`createdBy` preserved, `updatedAt` bumped, `version` incremented; PUT/DELETE with an id from another environment → `404`
- [ ] 51 widgets, blank title, duplicate widget ids, `{"type":"relative"}` without period → `400`; 50 widgets and `"value": []` filters → accepted
- [ ] Each verb without its ACL → `403`

[OBS-16]: https://gravitee.atlassian.net/browse/OBS-16?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/GammaDashboardRepositoryAdapter.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/GammaDashboardRepositoryAdapter.java
@@ -28,6 +28,7 @@ import io.gravitee.gamma.rest.core.observability.dashboard.port.repository.Dashb
 import io.gravitee.gamma.rest.core.observability.filter.model.FilterCondition;
 import io.gravitee.gamma.rest.core.observability.filter.model.FilterOperator;
 import io.gravitee.repository.management.model.GammaDashboard;
+import java.util.Date;
 import java.util.List;
 import java.util.Optional;
 import lombok.RequiredArgsConstructor;
@@ -53,7 +54,7 @@ public class GammaDashboardRepositoryAdapter implements DashboardRepository {
         return RepositoryCalls.wrap(
             () ->
                 gammaDashboardRepository.findByEnvironmentId(environmentId).stream().map(GammaDashboardRepositoryAdapter::toCore).toList(),
-            "Failed to list dashboards for environment '" + environmentId + "'"
+            "Failed to list dashboards for environment '%s'".formatted(environmentId)
         );
     }
 
@@ -61,7 +62,34 @@ public class GammaDashboardRepositoryAdapter implements DashboardRepository {
     public Optional<Dashboard> findByIdAndEnvironmentId(String id, String environmentId) {
         return RepositoryCalls.wrap(
             () -> gammaDashboardRepository.findByIdAndEnvironmentId(id, environmentId).map(GammaDashboardRepositoryAdapter::toCore),
-            "Failed to fetch dashboard '" + id + "' for environment '" + environmentId + "'"
+            "Failed to fetch dashboard '%s' for environment '%s'".formatted(id, environmentId)
+        );
+    }
+
+    @Override
+    public Dashboard create(Dashboard dashboard) {
+        return RepositoryCalls.wrap(
+            () -> toCore(gammaDashboardRepository.create(toRepository(dashboard))),
+            "Failed to create dashboard '%s'".formatted(dashboard.id())
+        );
+    }
+
+    @Override
+    public Dashboard update(Dashboard dashboard) {
+        return RepositoryCalls.wrap(
+            () -> toCore(gammaDashboardRepository.update(toRepository(dashboard))),
+            "Failed to update dashboard '%s'".formatted(dashboard.id())
+        );
+    }
+
+    @Override
+    public void delete(String id) {
+        RepositoryCalls.wrap(
+            () -> {
+                gammaDashboardRepository.delete(id);
+                return null;
+            },
+            "Failed to delete dashboard '%s'".formatted(id)
         );
     }
 
@@ -95,7 +123,7 @@ public class GammaDashboardRepositoryAdapter implements DashboardRepository {
             return FilterOperator.valueOf(operator.toUpperCase());
         } catch (NullPointerException | IllegalArgumentException e) {
             throw new TechnicalDomainException(
-                "Persisted dashboard filter '" + field + "' has an unsupported operator '" + operator + "'",
+                "Persisted dashboard filter '%s' has an unsupported operator '%s'".formatted(field, operator),
                 e
             );
         }
@@ -112,7 +140,7 @@ public class GammaDashboardRepositoryAdapter implements DashboardRepository {
         try {
             return TimeRangeType.valueOf(type.toUpperCase());
         } catch (NullPointerException | IllegalArgumentException e) {
-            throw new TechnicalDomainException("Persisted dashboard has an unsupported time range type '" + type + "'", e);
+            throw new TechnicalDomainException("Persisted dashboard has an unsupported time range type '%s'".formatted(type), e);
         }
     }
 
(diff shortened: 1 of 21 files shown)


gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/observability/dashboards/ObservabilityDashboardsResource.java
```diff
@@ -71,6 +91,39 @@ public PaginatedResponseDto<DashboardDto> list(@QueryParam("page") Integer page,
...
+        DashboardDto dto = DashboardDto.from(output.dashboard());
+        return Response.created(uriInfo.getAbsolutePathBuilder().path(dto.id()).build()).entity(dto).build();
+    }
+
+    /**
+     * Same resolution as {@code ObservabilityFilterDataPortAdapter#currentAuditInfo()} — the
+     * authenticated principal set by the security filter chain; only {@code createdBy} is needed
+     * here, so the full {@code AuditInfo} is not built.
+     */
+    private static String currentUsername() {
```
jourdiw: This is used in the other resources 🤔 Would be nice to eventually have this centralized somewhere.

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
```diff
@@ -1147,6 +1147,69 @@ paths:
...
                                 message: "Insufficient permissions to access dashboards."
+        post:
+            tags:
+                - Dashboards
+            summary: Create a dashboard
+            description: |
+                Creates a dashboard from a client-supplied `id`. Server-owned fields (`version`,
+                `createdBy`, `createdAt`, `updatedAt`, `environmentId`) sent in the body are
+                ignored: `version` starts at 1, `createdBy` is the authenticated caller and the
+                timestamps are stamped server-side. A dashboard holds at most 50 widgets.
```
jourdiw: Does it make sense to include doc about server-owned fields that are ignored? 
Shouldn't the request schema be clear enough?

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/resources/observability/dashboards/ObservabilityDashboardsResourceTest.java
```diff
@@ -243,5 +513,20 @@ ListObservabilityDashboardUseCase listObservabilityDashboardUseCase(DashboardRep
...
+
+        @Bean
+        UpdateObservabilityDashboardUseCase updateObservabilityDashboardUseCase(DashboardRepository dashboardRepository) {
+            return new UpdateObservabilityDashboardUseCase(dashboardRepository);
+        }
+
+        @Bean
+        DeleteObservabilityDashboardUseCase deleteObservabilityDashboardUseCase(DashboardRepository dashboardRepository) {
+            return new DeleteObservabilityDashboardUseCase(dashboardRepository);
+        }
```
jourdiw: A little odd to me that you would need to have the beans here instead of autowired 🤔

jourdiw (approved):

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
SamHardi: Fixed in d63c405ffb — you were right on both counts.

The schema now carries it: `version`, `createdAt` and `updatedAt` are marked `readOnly: true` on `Dashboard`, and `SaveDashboardRequest` simply doesn't declare them. So the prose no longer needs to repeat what is ignored on write.

Taking the remark further, I rewrote the whole dashboards section from the point of view of someone consuming the API rather than reading our code. Gone: "version starts at 1 server-side", "cross-environment existence cannot be probed", and a reference to a future internal ticket that had no business being in a public spec. In their place, what a client actually needs — that you supply the `id` yourself, that `PUT` is a full replacement and not a patch, that concurrent updates are last-write-wins so you should re-read before saving, that `DELETE` has no undo, and what each 400 actually means.

One real bug surfaced while doing it: the list endpoint was documented as "most recently created first" when the order is ascending. Fixed.

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/resources/observability/dashboards/ObservabilityDashboardsResourceTest.java
SamHardi: Fixed in d63c405ffb. The five hand-written use-case beans are gone — the test config now uses the same `@UseCase`-filtered `@ComponentScan` as `GammaDashboardsConfiguration`, so the test exercises whatever the real context would build instead of a parallel wiring that could drift.

Only the port stays declared by hand, and that one is unavoidable: production derives it from the `GammaDashboardRepository` SPI, which the repository plugin registers at runtime and which doesn't exist in a unit-test context, so the in-memory implementation takes its place. I put that reason in a comment on the config so the next reader doesn't have to ask.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/GammaDashboardRepositoryAdapter.java
```diff
@@ -126,4 +154,46 @@ private static JsonNode parseWidgets(String widgets) {
             throw new TechnicalDomainException("Persisted dashboard widgets payload is not valid JSON", e);
         }
     }
+
+    private static GammaDashboard toRepository(Dashboard source) {
```
SamHardi, line 166: Fixed in d63c405ffb — added a short javadoc covering what the reverse mapping does (lowercasing the operator and time-range type back to the persisted vocabulary, flattening the widgets node to its stored JSON string) and the null handling: nulls are carried through rather than defaulted, and an absent widgets node is stored as SQL/Mongo null rather than the string "null", so an empty dashboard round-trips unchanged.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/observability/dashboards/ObservabilityDashboardsResource.java
SamHardi: Agreed on the duplication — I tried it and hit a wall worth reporting rather than working around.

I extracted an `AuthenticatedUser.currentUserId()` helper, and `OnionArchitectureTest` rejected it. The module enforces `rest_resources_must_not_depend_on_infra` **and** `infra_must_not_depend_on_rest_resources`, and `core` can't hold it either since `CoreRulesTest` forbids Spring Security there. The other three copies live in the observability adapters under `infra`, mine is a resource — so there is no package visible from both sides today. Centralising really means either a new top-level package alongside `core`/`infra`/`resources`, or a core port with an infra implementation.

That's a structural call on the module layout, so it didn't feel right to slip it into this PR. I reverted to the local private method and documented why it's local, so the next person hits the explanation rather than the wall. Happy to open a follow-up ticket for the four call sites if you think it's worth it — say the word and I'll file it.

PR #19043, opened 2026-08-07
chore: generate AI context from the gravitee-ai-foundation hub

FuYu3699:
BX-284

## What this does

Replaces the hand-written root-level AI context with generated context from the shared Gravitee AI foundation hub. This is the first of two PRs migrating this repository onto it; the second declares the modules.

Deleted, with every section covered by the generated root in this same PR:

- root `AGENTS.md` (the router), replaced by the generated `AGENTS.md`
- `.claude/CLAUDE.md`, replaced by the generated root `CLAUDE.md` shim; `.github/AGENTS.md`, a content-free pointer, is dropped — agents working in `.github/` read the root file
- `.agent-rules/general.md`, `java.md`, `angular.md`, now carried by the shared rules or the repo-local rules below (the shared Java rule also corrects the old per-module test advice: `-am test` runs the tests of every upstream module too, so it prescribes building upstream with `-DskipTests install` first)

Added:

- `.ai/manifest.yaml` declares `product: apim`. No modules yet: PR 2 declares all 17 together with their generated files, so no intermediate state exists where a declared module still has a hand-written file.
- `.ai/rules/`: five repo-local rules. Java specifics, the Angular conventions, the Automation API sync triggers, a temporary module index, and authoring guidance for `.ai/`.
- `.ai/guides/`: two on-demand deep dives (Angular async patterns, the Automation API checklist).
- Generated: root `AGENTS.md`, the `CLAUDE.md` shims, `.github/instructions/code-review.instructions.md` for Copilot code review, and `.ai/generated.yaml` (the record that `gbuddy check` verifies).

## Why

Hand-written files drift. The root module index was missing `gravitee-gamma/gravitee-gamma-rest-api`, and the Java rule recommended a Maven command that does not do what it says. Generated context comes from one shared source, updates centrally, and is verified against its record.

## What changes for developers

- At the root, agents load one assembled file carrying the shared Gravitee conventions (Java, TypeScript, security, TDD, planning, review) plus the APIM-specific rules. The conditional "read the matching stack" routing is gone; assembly puts each rule in the file agents already load.
- The 17 module `AGENTS.md` files are untouched and still apply; the root index still points at all of them. An agent launched with its working directory inside a module sees that module's file only, exactly as today; PR 2's generated module files add the pointer back to root.
- New capability: generated review instructions for Copilot code review, and seven skills for local AI tools (installed per developer, not committed).
- Two things you may notice. `CLAUDE.md` is no longer gitignored, because the generated shims need committing; a personal root `CLAUDE.md` will now conflict with the committed shim on pull — delete or move it. And running `gbuddy check` before ever running `gbuddy setup` prints 22 "missing local file" warnings for the per-developer skill files; that is expected, run setup once and they resolve.

## The numbers, honestly

Today a Java developer at the root effectively loads about 190 lines (router + general + java). This PR's root is 307 lines: more than today, and deliberately temporary. PR 2 declares the modules, moves the Java, TypeScript, and Angular layers into the module files agents load where they work, and lands the root near 190. The pitch is coverage (security, TDD, and review rules this repo never had), central maintenance, and drift-proofing, not token savings.

## Review guide

The five files under `.ai/rules/` and two under `.ai/guides/` are the authored content; review those as content. Root `AGENTS.md`, the `CLAUDE.md` shims, `.ai/AGENTS.md`, `.github/instructions/`, and `.ai/generated.yaml` are generated output; spot-check rather than line-read.

---


AGENTS.md
```diff
@@ -1,58 +1,307 @@
...
+Java-specific conventions for Gravitee repositories. Global engineering, security, and TDD rules still apply.
+
+## Follow the repo
+
+- Match the repository's Java version, Maven setup, formatter, package structure, naming style, and test stack. Do not introduce a second convention.
+- Use the parent POM or BOM for dependency versions. Avoid ad hoc version pins and avoid new dependencies when the repo already provides a suitable option.
+
+## Defaults when writing Java
+
+- Use Lombok only where it is already conventional in the repo, such as `@Slf4j` or simple DTO/builders. Avoid introducing Lombok, especially broad annotations like `@Data`, where generated semantics are unclear.
```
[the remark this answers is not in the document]
FuYu3699: Fixed on the current head: this line was corrected in the shared rules and regenerated into this branch in 74f04c2b9d — it now reads "an established logger annotation (`@CustomLog` in some repositories, `@Slf4j` in others)", and the APIM section below states this repository's answer (`@CustomLog`). The comment was anchored to an earlier commit.

AGENTS.md
```diff
@@ -1,58 +1,307 @@
...
+- **No `Co-Authored-By` trailers.**
+- Prefer small, logically-scoped commits over one large commit.
+- Branch name: `<type>/<PROJECT-KEY>-<slug>` (e.g. `feat/ABC-123-add-retry`), using the issue tracker's project key for the target repository.
+- Never bypass hooks (`--no-verify`) or skip signing unless explicitly asked.
+- Create new commits rather than amending; only commit when asked.
+
+# Language and Locale
+
+- Use **US English** for comments, log messages, and user-facing strings (UI copy, API error messages, operator-facing documentation).
+- Match the repository's existing spelling and terminology in the module you are editing — do not introduce a second dialect.
```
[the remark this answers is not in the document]
FuYu3699: Fair observation, two parts to it. The rule's stated scope is code artifacts — comments, log messages, and user-facing strings — not the guidance prose itself, so the document does not break its own rule as written. But you are the second reviewer to flag the mismatch, so the optics clearly cost something: an improvement to align the shared rule prose (or state the scope more visibly) is queued on the shared-rules side, where that text is maintained. Nothing changes in this PR because the wording ships from the shared source, not from this repository.

AGENTS.md
```diff
@@ -1,58 +1,307 @@
...
+- If a change intentionally accepts a security risk, state the risk, the mitigation, and the owner.
+
+# Test-Driven Development
+
+- Write the failing test before the production code: **RED → GREEN → REFACTOR**. Don't skip the refactor step.
+- Tests assert **behaviour**, not implementation details.
+- Prefer **real instances over mocks** for pure logic (no network, DB, or filesystem). Mock only genuine side-effect boundaries.
+- For every behaviour change, name the test that covers it. If a test genuinely can't be written, document the gap and the reason — don't silently skip it.
+
+# Java Conventions
```
FuYu3699, line 110: Temporary placement: this Java section sits at the root only until the follow-up PR declares the Maven modules — it then moves into each Java module's own AGENTS.md, next to the code it governs, and the root shrinks to about 190 lines.

AGENTS.md
```diff
@@ -1,58 +1,307 @@
...
+Treat changes to any of these as compatibility-sensitive — call them out explicitly in the PR description and final response summary:
+
+- Public APIs: REST, gRPC, OpenAPI contracts, generated models.
+- Persistence schemas: Liquibase changelogs, JPA entities, repository interfaces.
+- Configuration keys, environment variables, and feature flags.
+- Serialised DTOs, enum values, and on-wire formats.
+- Authentication, authorisation, and security-relevant paths.
+- Deployment files and CI configuration.
+
+# TypeScript Conventions
```
FuYu3699, line 153: Same as the Java section above: temporary at the root. The follow-up PR moves TypeScript (and Angular) content into the frontend modules' own files.

.gitignore
```diff
@@ -86,9 +86,11 @@ gravitee-apim-distribution/gravitee-apim-distribution-e2e/ui-test/screenshots/
 /.mvn/maven.config
 
 # Claude configurations
-CLAUDE.md
 .claude
 
+# AI agent skills, installed locally per developer (not committed)
+.agents/skills/
```
ytvnr, line 92: The `CLAUDE.md` entry removed just above was a pattern without a leading slash, so it matched at **every** depth, not only the root. Dropping it means any personal `CLAUDE.md` a developer keeps in any subdirectory becomes untracked-visible and committable by accident. It is also what creates the pull conflict described in the PR description.

An alternative that keeps the intent explicit and preserves the protection everywhere else:

```gitignore
CLAUDE.md
!/CLAUDE.md
!/.ai/CLAUDE.md
```

plus the 17 module negations in PR 2. More verbose, but personal files stay ignored outside the generated ones.

.ai/rules/apim-java.md
```diff
@@ -0,0 +1,11 @@
...
+name: APIM Java Conventions
+layer: local
+description: APIM-specific Java conventions - the formatter command, Lombok usage in this codebase, and the JsonNode pitfall
+---
+
+# APIM Java Conventions
+
+- **Formatter:** Google Java Style via the Maven Prettier plugin — run `mvn prettier:write -pl <module>` on every module you changed.
+- **Lombok — the shared Java rule's "already conventional" condition is met here:** this codebase uses `@Data`, `@Builder`, and `@CustomLog`; follow the module's existing conventions and inject loggers with `@CustomLog`, not `@Slf4j`.
+- **JsonNode vs ObjectNode in lists:** `List<ObjectNode>` cannot be assigned to `List<JsonNode>` — Java generics are invariant. Helpers that feed `JsonNode` lists (for example `Aggregation.setBuckets()`) must return `JsonNode`, not `ObjectNode`.
```
ytvnr, line 11: The description says every section of the deleted files is covered by the generated root. The snake_case test naming convention is the one that is not.

The old `.agent-rules/java.md` made it a MUST (`should_return_error_when_api_not_found()`). The shared Java rule now says "Follow the repo's existing test naming style. **Where no clear style exists**, prefer `should_<behavior>`". APIM does have a clear style, so an agent has to infer it instead of reading it. One line in this file would close the gap.

Two smaller drops from the same deleted file, both defensible under the "do not document the inferable" principle in `ai-context-authoring.md`, but worth confirming they are deliberate rather than accidental: `Target Java 21+`, and "use `var` where the type is obvious".

AGENTS.md
```diff
@@ -1,58 +1,307 @@
...
+3. Implement or update JAX-RS resources and mappers using the generated `io.gravitee.rest.api.*.model.*` types.
+
+Do not hand-write DTOs that duplicate generated models or start from JAX-RS before generation has run.
+
+## External plugins
+
+- Package: `target/gravitee-<type>-<name>-<version>.zip`.
+- **Local verification:** after building standalone Gateway and REST API distributions (`gravitee-apim-distribution`), copy the zip into **both** standalone `plugins/` directories (Gateway and Management API) before exercising management + runtime together.
+
+## Operator-visible Gateway behaviour
```
ytvnr, line 216: This section keeps the diagnostics guidance as prose, but the three concrete `ctx.warnWith` / `ctx.interruptWith` snippets from the old `.agent-rules/java.md` are gone. Those are exactly the shape an agent copies rather than reconstructs, and getting the builder chain right (`.key()`, `.message()`, `.cause(e)`) is not obvious from the prose alone.

If they are too heavy for an always-on rule, a `.ai/guides/` file with a one-line breadcrumb from here would keep them reachable, which is the pattern the Angular and Automation rules already use.

AGENTS.md
```diff
@@ -1,58 +1,307 @@
...
+
+When a trigger fires, follow the checklist in `.ai/guides/automation-api-sync.md`; the **Management API: API-first** section of the root `AGENTS.md` names the spec paths and the compile step.
+
+# APIM Java Conventions
+
+- **Formatter:** Google Java Style via the Maven Prettier plugin — run `mvn prettier:write -pl <module>` on every module you changed.
+- **Lombok — the shared Java rule's "already conventional" condition is met here:** this codebase uses `@Data`, `@Builder`, and `@CustomLog`; follow the module's existing conventions and inject loggers with `@CustomLog`, not `@Slf4j`.
+- **JsonNode vs ObjectNode in lists:** `List<ObjectNode>` cannot be assigned to `List<JsonNode>` — Java generics are invariant. Helpers that feed `JsonNode` lists (for example `Aggregation.setBuckets()`) must return `JsonNode`, not `ObjectNode`.
+
+# Module Context Files
```
ytvnr, line 291: Nit: the assembled file ends up with two sections whose names are close enough to send a reader looking twice, `# Module Context Files` here (from the local rule) and `## Modules` at line 303 (generated tail). Renaming one of them, or folding the `.ai/` pointer into this section, would read better.

ytvnr (commented):
Full review of the diff, with every factual claim in the description checked against the repository. Summary here, details inline.

### Verified and correct

- The 17 directories listed in `apim-module-index.md` all carry an `AGENTS.md`, including `gravitee-gamma/gravitee-gamma-rest-api` which the old index was indeed missing.
- The five committed hashes in `.ai/generated.yaml` match the committed file contents (checked with `shasum -a 256`).
- Every path cited in the new rules and guides exists: the Automation mapper package, the CRD package, the four OpenAPI specs, and `CONTRIBUTING.adoc:77` for the `AI Agent Context` section.
- No dangling reference to `.agent-rules/` or `.claude/CLAUDE.md` remains anywhere in the tree.
- The Angular claims hold. `@angular/core` is `20.3.16`, so standalone is the default. Console's `GioTestingModule` uses `provideHttpClient` + `provideHttpClientTesting`, so it is genuinely safe with `rxResource`. Portal's `AppTestingModule` does import `HttpClientTestingModule`.
- The line counts are accurate: 307 for the new root, 188 for the old router + general + java.
- No CI formatting risk on the new root markdown: `nx format:check --libs-and-apps` does not cover it.

I did not find a single incorrect factual claim in the description, which is unusual for a change of this size.

### The one point to settle before merge

`gbuddy check` is not wired into CI. `git grep -l gbuddy` across `.circleci/` and `.github/` returns only the generated instructions file. The value proposition of the PR is that generated context is "verified against its record", and every generated file carries a "Do not edit directly" header, but today nothing stops a hand edit from landing on master. Either add the job in this PR, or say explicitly in the description that it arrives with PR 2.

### Note on the numbers section

The 190 to 307 comparison is the Java case. For an Angular developer the previous effective load was 335 lines (router + general + angular), so this PR is actually a decrease for them. Worth stating: it makes the temporary regression narrower than "more than today" suggests.

Nothing below is a correctness blocker.

ytvnr (approved):
Approving. The findings in my previous review are non-blocking: the only one I would like a decision on before merge is whether `gbuddy check` lands in CI here or with PR 2. Everything else is an improvement you can take or leave.

.gitignore
FuYu3699: Accepted in PR 2 (67efa5e624): your structure verbatim, plus the 17 module negations. Two notes from testing it before folding: plain `git check-ignore` cannot guard the negation list — tracked files never report as ignored — so `git check-ignore --no-index` is the form that catches a forgotten negation; and the `.ai/AGENTS.md` authoring guidance now makes the gitignore negation an explicit step of declaring a module, since forgetting it means a new module's shim silently never enters git.

.ai/rules/apim-java.md
FuYu3699: Accepted — one line in `.ai/rules/apim-java.md` carrying the original example name, now landing in all 12 Java module files (67efa5e624 on PR 2). You are right that a clear repo style beats routed inference; the migration record's disposition is revised accordingly. On the two smaller drops: `Target Java 21+` and `var where obvious` were both deliberate (stated by the root pom and inferable from the codebase, per the authoring bar in `.ai/AGENTS.md`) — confirming that here so the record shows it.

AGENTS.md
FuYu3699: Accepted — the three builder snippets are recovered verbatim into `.ai/guides/gateway-diagnostics.md`, with a one-line breadcrumb from the Java rule in every Java module file (67efa5e624 on PR 2); the prose rules stay in the root's Operator-visible section, the guide carries the copy-ready depth — the same split as the Angular async guide. Your copy-shaped-versus-reconstruct framing also triggered a sweep of every other example the migration dropped: this was the only copy-shaped one.

AGENTS.md
FuYu3699: Resolved by construction in PR 2: the temporary module-index rule is deleted there, so `# Module Context Files` disappears and only the generated `## Modules` list remains.

PR #19044, opened 2026-08-07
feat(observability): filter authz decisions by outcome

podlesrafal:
> Follows #18972, now merged. Rebased onto master, so this diff is the delta alone.
>
> Frontend counterpart: [gravitee-gamma-module-authz#117](https://github.com/gravitee-io/gravitee-gamma-module-authz/pull/117).

## Summary

The decisions screen has no filters. This adds the first one, and the column that
was missing from the table, both from the observability epic
([AUTHZ-2](https://gravitee.atlassian.net/browse/AUTHZ-2)).

`DECISION` goes first because it is the only one of the epic's seven predicates whose
values are a closed set the catalog can carry inline. Principals, actions, resources and
reasons need the filter value endpoint, which this version of `gamma-lib-observability`
does not render, so they need a different approach and a separate change.

## Changes

**The predicate**
- `AuthzDecisionLogQuery` gains `decisions`; the elasticsearch adapter adds a `terms`
  clause on the `decision` field, and omits it entirely when nothing is selected.
- The core port now takes an `AuthzDecisionLogFilters` record rather than loose
  parameters. The remaining predicates of the epic land as fields there instead of
  widening the signature a third time.
- `SearchObservabilityLogsUseCase` stops refusing `DECISION`. Everything else is still
  refused rather than silently dropped.

**Reaching the screen**
- `StaticFilters` gains `DECISION`: `ENUM`, `EQ`/`IN`, three values inline, `LOGS` only.
- `ApiType` gains `AUTHZ_DECISION`, excluded from the `API_TYPE` picker's values.
- `ENTRYPOINT` stops declaring every API kind. Adding the scope token to the enum put every
  `ApiType.ALL` filter on the decisions screen, and that one is not applicable there — a
  decision carries no entrypoint. The registry test now pins what that scope offers.

**The column**
- `Policy version`, from `authz.policyGeneration`. Lands in
  [gravitee-gamma-module-authz](https://github.com/gravitee-io/gravitee-gamma-module-authz)
  since the table config lives there — the value was already on the wire and reached the
  row only inside the untyped `context` bag, so nothing surfaced it.

## API surface

|---|---|---|---|---|
| `/environments/{envId}/observability/logs/search` | `POST` | `DECISION` accepted alongside `RECORD_TYPE=AUTHZ_DECISION`. | `{"filters":[{"name":"RECORD_TYPE","operator":"EQ","value":"AUTHZ_DECISION"},{"name":"DECISION","operator":"IN","value":["FORBID"]}]}` | `200`, narrowed to forbidden decisions; previously `400` |
| `/environments/{envId}/observability/filters/definition` | `GET` | Catalog gains a `DECISION` entry, `LOGS` only, scoped so it is offered on the decisions screen alone. `API_TYPE` values are unchanged. | `?signal=LOGS` | one additional `FilterSpec` |

No breaking change: a filter that was refused is now accepted, and the picker's values
do not move.

## Reviewer notes

**The one thing I want a second opinion on.** The library decides which filters to offer
by matching a filter's `apiTypes` against the value of the field named by its
`scopeFilterField`, which on the decisions screen is `RECORD_TYPE=AUTHZ_DECISION`. A
filter is offered there only if that literal token is among its `apiTypes`. `FilterSpec`
has no separate scope axis, so the token now lives in the `ApiType` enum, where it is not
an API kind, and is filtered out of the `API_TYPE` picker so it never reaches a user.

The clean alternative is a `scopes` axis on `FilterSpec` alongside `apiTypes`, serialised
into the same wire field the library reads. That is the right model, but it touches the
shared filter catalog and all 46 definitions including the analytics surface, so I did
not fold it into a change this size. Happy to do it that way instead if you would rather
pay for it now — the rest of this PR is unaffected either way.

**Why the port signature changed.** Adding a fourth loose parameter to
`searchDecisionLogs` for one predicate, then three more later, is how a five-argument
method becomes an eight-argument one. The record makes each addition a field.

## Test plan

- [ ] On *Observability → Decisions*, open **Add filter**: a **Decision** entry appears
      with Permit / Forbid / Not applicable.
- [ ] Apply **Decision = Forbid**: only forbidden decisions come back. Check the request
      body carries the filter and the total drops.
- [ ] Apply **Decision = Permit, Forbid**: the union comes back, not just the first value.
- [ ] Pick a decision value that matches nothing in the window: zero rows, not the
      unfiltered total. That distinction is the whole point.
- [ ] Combine **Decision** with **API**: both narrow, the result is their intersection.
- [ ] On *Observability → Logs* (request rows), **Decision** is not offered.
- [ ] The **Policy version** column shows a number on evaluated rows and stays blank where
      the decision carries none.

## Verification

Automated, on this branch: `SearchAuthzDecisionLogsQueryAdapterTest` (11, two new for the
clause being present and absent), `AuthzDecisionLogsCrudServiceImplTest` (5, one new for
the filter reaching the repository), `SearchObservabilityLogsUseCaseTest` (23, one new for
the predicate surviving the refusal gate), `ObservabilityLogsDataPortAdapterTest` (42, two
new for the condition translating into the filters record). Whole modules green with the
prettier-java check on: 742 in repository-elasticsearch, 8,441 in rest-api-service, 279 in
gamma-rest-api. Console side 1,185, two of them new for the column.

**Verified on a live stack** — rest-api built from this branch against an Elasticsearch
holding 56 real decision documents, driven through the console:

- `Decision` is offered in **Add filter**, in its own *Authz decision* group, with the
  three values inline. The scope token does what it is supposed to.
- `Decision = FORBID` returns **zero rows**, not the unfiltered set. That negative control
  is the only thing that distinguishes an applied filter from an ignored one.
- `Decision = PERMIT` returns permits only, and drops the search records, which carry no
  decision.

That pass is also what caught the `ENTRYPOINT` scoping bug above: it was offered on the
decisions screen, where the search refuses it. The unit tests were all green at the time.

Two defects turned up that this PR does not introduce and does not fix, both filed
separately: clicking a row opens a drawer that errors with *"Log detail is not supported.
The data source does not implement getById"* (the panel is fully written, the data source
just never implements it), and the histogram above the table never renders because it
reads the analytics signal, which does not aggregate decisions.

[AUTHZ-2]: https://gravitee.atlassian.net/browse/AUTHZ-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/log/v4/model/authz/AuthzDecisionLogQuery.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-api/src/main/java/io/gravitee/repository/log/v4/model/authz/AuthzDecisionLogQuery.java
@@ -32,6 +32,7 @@ public class AuthzDecisionLogQuery {
     private Set<String> apiIds;
     private Long from;
     private Long to;
+    private Set<String> decisions;
 
     @Builder.Default
     private int size = 20;
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/adapter/authz/SearchAuthzDecisionLogsQueryAdapter.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/adapter/authz/SearchAuthzDecisionLogsQueryAdapter.java
@@ -58,6 +58,12 @@ public final class SearchAuthzDecisionLogsQueryAdapter {
         query.getApiIds().forEach(apiIds::add);
         filters.add(MAPPER.createObjectNode().set(TERMS, MAPPER.createObjectNode().set(AuthzDecisionLogFields.API_ID, apiIds)));
 
+        if (query.getDecisions() != null && !query.getDecisions().isEmpty()) {
+            ArrayNode decisions = MAPPER.createArrayNode();
+            query.getDecisions().forEach(decisions::add);
+            filters.add(MAPPER.createObjectNode().set(TERMS, MAPPER.createObjectNode().set(AuthzDecisionLogFields.DECISION, decisions)));
+        }
+
         if (query.getFrom() != null || query.getTo() != null) {
             ObjectNode bounds = MAPPER.createObjectNode();
             if (query.getFrom() != null) {
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/log/adapter/authz/SearchAuthzDecisionLogsQueryAdapterTest.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/log/adapter/authz/SearchAuthzDecisionLogsQueryAdapterTest.java
@@ -109,6 +109,30 @@ class SearchAuthzDecisionLogsQueryAdapterTest {
     }
 
     @Test
+    void narrows_on_the_requested_decisions() {
+        var query = AuthzDecisionLogQuery.builder().apiIds(Set.of("api-1")).decisions(Set.of("FORBID")).build();
+
+        var result = SearchAuthzDecisionLogsQueryAdapter.adapt(query);
+
+        assertThatJson(result)
+            .inPath("$.query.bool.filter[2].terms.decision")
+            .isEqualTo(
+                """
+                [ "FORBID" ]
+                """
+            );
+    }
+
+    @Test
+    void omits_the_decision_clause_when_none_is_requested() {
+        var query = AuthzDecisionLogQuery.builder().apiIds(Set.of("api-1")).build();
+
+        var result = SearchAuthzDecisionLogsQueryAdapter.adapt(query);
+
+        assertThatJson(result).inPath("$.query.bool.filter").isArray().hasSize(2);
+    }
+
+    @Test
     void rejects_a_query_that_would_read_across_every_api() {
         var query = AuthzDecisionLogQuery.builder().apiIds(Set.of()).build();
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/log/crud_service/AuthzDecisionLogsCrudService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/log/crud_service/AuthzDecisionLogsCrudService.java
@@ -16,10 +16,10 @@
 package io.gravitee.apim.core.log.crud_service;
 
 import io.gravitee.apim.core.log.model.AuthzDecisionLog;
+import io.gravitee.apim.core.log.model.AuthzDecisionLogFilters;
 import io.gravitee.rest.api.model.common.Pageable;
 import io.gravitee.rest.api.model.v4.log.SearchLogsResponse;
 import io.gravitee.rest.api.service.common.ExecutionContext;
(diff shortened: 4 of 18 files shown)


sclarkamlab (commented):
Looks good. Followed the DECISION predicate all the way through, catalog to validation to the use case gate to the port adapter to the ES terms clause, and it holds up. Checked the vocabulary matches what the reporter actually writes (PERMIT/FORBID/NOT_APPLICABLE) and that `decision` is mapped as a keyword, so the terms clause will match. Also nice that the request path already throws on an unknown condition name, so a DECISION filter sent with RECORD_TYPE=REQUEST 400s instead of quietly vanishing.

Few small things, none blocking.

`import java.util.Set;` in `AuthzDecisionLogsCrudServiceImpl.java` line 30 is unused now the params moved into the filters record.

The new empty-values guard in `ObservabilityFilterValidator` only covers ENUM specs. Keyword filters still get through with an empty list, `API IN []` just falls back to every accessible API with the chip still sitting there, which is the same failure your comment describes. Not something you introduced, but it's half a fix. Might be worth lifting the guard so it covers any condition with an empty values list.

If you do that, the `condition.values() != null ? condition.values() : List.of()` ternary just below can never take the else branch, so it can go.

`ApiType` now carries a token that isn't an API kind, which means every consumer of `ApiType.ALL` has to remember to exclude it. Two got patched here, `AnalyticsRequestPipeline` line 46 still uses `ALL`. I checked and it's harmless, `computeScope` filters on the API's own type and no real API is AUTHZ_DECISION, so nothing to fix, just flagging it's the third site. `AUTHZ` and `AUTHZ_DECISION` sat next to each other labelled "Authz decision" and "Authz decision record" is also pretty easy to misread.

Last one, there's no direct test in `ObservabilityFilterValidatorTest` for the new empty-enum rejection, it only gets exercised through the use case test.

sclarkamlab (approved):

sclarkamlab (commented):
Went back over this with a sharper eye and found two things I missed first time round, so ignore the "nothing blocking" in my last comment.

The API Type chip on the decisions screen won't populate. `GetObservabilityFilterValuesUseCase` line 109 narrows API_TYPE's enum values to the requested `apiType` set, so `GET filters/API_TYPE/values?apiType=AUTHZ_DECISION` builds `allowedValues = {"AUTHZ_DECISION"}`, and since you deliberately excluded that token from the API_TYPE baseline in `ExtensibleFilters` the list filters down to nothing and you get back `{data: [], totalCount: 0}`. `SpiFilterRegistryTest` asserts API_TYPE is offered on that screen and `narrowApiTypes` honours it in the search, so the filter is advertised and applicable but unpopulatable. This one is on the PR, before it the token didn't exist, and `apiType=AUTHZ` works fine because AUTHZ is still in the baseline.

Same root cause a layer over: for KEYWORD filters, `ObservabilityFilterDataPortAdapter` line 62 maps gamma types to platform types through a partial `EnumMap` with no entry for AUTHZ_DECISION, then drops the nulls. So `apiType=AUTHZ_DECISION` collapses to an empty set, and `GetFilterValuesUseCase` line 131 reads empty as "don't narrow". The API chip's values come back spanning every API type instead of being scoped.

Second one, the OpenAPI `ApiType` enum at line 1259 is still `[HTTP_PROXY, LLM, MESSAGE, MCP, NATIVE, EDGE]`. `FilterSpecDto` line 64 emits raw enum names, so the DECISION spec goes out with `apiTypes: ["AUTHZ_DECISION"]` and anything validating responses against the published schema will reject a legitimate 200. A2A and AUTHZ are already missing so the drift isn't yours, but this is the first value that actually shows up in a response body.

Everything from my previous comment still stands, the empty-keyword-values one is the same class of bug as the API_TYPE hole so might be worth fixing together.

PR #19046, opened 2026-08-08
chore: declare all 21 modules and generate their AI context

FuYu3699:
BX-285

## What this does

Completes the migration started in #19043: declares all 21 code modules in `.ai/manifest.yaml` and replaces every hand-written module `AGENTS.md` with a generated file in the same change. After this PR, every AI context file in the repository is either generated from shared rules or a thin repo-local rule scoped to the directories it serves; nothing is hand-maintained duplication.

Deleted, each replaced by its generated successor in this same diff:

- the 17 hand-written module `AGENTS.md` files
- `.ai/rules/apim-module-index.md`, the temporary hand-maintained module index from PR 1 — the generated root's module list is derived from the manifest, so it cannot drift

Added under `.ai/rules/`, one scoped rule per surface (each lands in the module files it names, next to the shared layers):

- `console-webui.md`, `portal-webui-next.md`, `gravitee-dashboard.md`, `gravitee-kafka-explorer.md`, `gravitee-markdown.md` — the five frontend module files' content, minus what the shared Angular rule now states in the same generated file
- `gamma-clean-architecture.md` — the Clean Architecture body the three gamma files carried three times (the two module files were near-identical twins), authored once. Gamma pom comments, ArchUnit tests, and configuration classes cite "AGENTS.md §n", so the rule keeps the original section numbers; those references still resolve in the generated files.
- `gamma-host.md` — only where `gravitee-gamma-rest-api` deviates from the shared gamma rules
- `apim-jsonnode.md` — the JsonNode list pitfall, scoped to the two modules whose aggregation helpers hit it

Edited: `apim-java.md` and `apim-angular.md` now carry `dirs:` so each lands beside its shared technology layer in the module files and wins by proximity; the Angular rule drops everything the shared Angular rule states in those same files (kept: the settled absolutes, member order, `inject()`-only DI, typed dialogs, naming, lodash-first, `fixture.destroy()`).

## Why

The sequencing rule from PR 1 holds: a module is declared only in the PR that also generates its file, so all 17 land together and no intermediate state exists. Declaring the modules moves the Java, TypeScript, and Angular layers out of the root and into the files agents load where they actually work. That includes Angular content for the three `gravitee-apim-webui-libs` libraries, which the hub detects through their Nx `project.json`.

## What changes for developers

- Root `AGENTS.md` drops from 307 to 199 lines; the technology layers now live in the module files.
- An agent launched inside a module now gets a generated file carrying the shared conventions for that stack plus the module's own rules, with a pointer back to root — before, four of these files were a bare heading with no content, and none pointed at root.
- Every module gets a one-line `CLAUDE.md` shim. Personal `CLAUDE.md` files anywhere else stay gitignored: the review round restored the all-depth ignore with explicit exceptions for the 19 committed shims, so nothing personal becomes committable by accident.
- Nothing under `.cursor/` or `.gemini/` changes.
- From the review round: the gateway diagnostics builder patterns (`ctx.interruptWith` / `ctx.warnWith`) are back as a copy-ready guide (`.ai/guides/gateway-diagnostics.md`), referenced from the Java rule; and the snake_case test-name convention is stated explicitly in every Java module file.

## The numbers, honestly

Root 307 → 203 (four router lines returned with the newly declared modules). Module files: Java modules 59–64 lines, frontend 116–165, the gamma modules 253 and the gamma host 337 — the two biggest are expected, they carry the shared Clean Architecture body that previously existed as 558 hand-written lines across three files. A developer inside a Java module now loads root + 57 lines where today they load root + 1–37. The gain is that those lines are the shared, centrally maintained conventions rather than a blank heading.

## Scope note: every code module is declared

Four directories that never had an `AGENTS.md` are declared alongside the 17 that did, so there is no special case to remember: `gravitee-apim-portal-webui` (gets the shared Angular and TypeScript conventions in their follow-the-repo form — the stricter modern-Angular rules stay scoped to the five modern frontend modules), `gravitee-gamma/gravitee-gamma-control-plane-webui` (TypeScript), and `gravitee-gamma/gravitee-gamma-plugin` + `gravitee-gamma/gravitee-gamma-definition-model` (Java + the APIM Java rule). Deliberately not declared, with reasons: `gravitee-apim-bom` and `gravitee-apim-parent` (dependency-management poms, no source), the `gravitee-gamma` reactor root (workspace configuration; its children are declared), `gravitee-gamma/gamma-ui-shared` (no build file of its own), and `.circleci`/`release` (repository tooling, not product modules).

## Review guide

The ten files under `.ai/rules/` are the authored content — review those as content. The 21 module `AGENTS.md` files, root `AGENTS.md`, the `CLAUDE.md` shims, and `.ai/generated.yaml` are generated output: `gbuddy check` verifies them against the record, so spot-check rather than line-read. One cosmetic quirk: the root's `## Modules` list renders under the last rule's heading; a generator fix is queued separately.


.ai/rules/gamma-clean-architecture.md
```diff
@@ -0,0 +1,196 @@
...
+    ClusterAdapter INSTANCE = Mappers.getMapper(ClusterAdapter.class);
+
+    Cluster toCoreModel(io.gravitee.repository.management.model.Cluster source);
+    io.gravitee.repository.management.model.Cluster toRepository(Cluster source);
+}
+```
+
+## 7. Domain Models
+
+- Domain behavior methods are preferred on model classes. We apply POO principles to domain models. We avoid anemic models.
```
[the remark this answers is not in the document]
FuYu3699: Accepted — a locale artifact carried over from the previous hand-written file ("programmation orientée objet"). Fixed to OOP in 67efa5e624.

.ai/rules/gamma-clean-architecture.md
```diff
@@ -0,0 +1,196 @@
...
+- Domain exceptions extend these bases with descriptive names (e.g., `ClusterNotFoundException`).
+
+## 9. REST Layer Conventions
+
+- JAX-RS: `@Path`, `@GET`/`@POST`/etc., `@Produces(MediaType.APPLICATION_JSON)`.
+- Inject with `@Inject` (Jakarta), not `@Autowired`.
+- **REST resources call use cases** — never Repository directly.
+- Register the parent resource in the module's registration class — `HttpApiManagementModule` (module-apim), `PlatformModule` (module-platform), or `GammaModuleApplication` (the host). Each sub-resource should be referenced in that parent resource.
+- Use records for simple response DTOs.
+- For resource creating stuff, the id should come from the request instead of the usecase generating one.
```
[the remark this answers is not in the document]
FuYu3699: Accepted — rephrased to "For resource-creation endpoints, the id comes from the request, not generated by the use case." in 67efa5e624.

.ai/rules/gamma-clean-architecture.md
```diff
@@ -0,0 +1,196 @@
...
+## 8. Exception Hierarchy
+
+- Use base exceptions from gravitee-apim-common: `NotFoundDomainException`, `ValidationDomainException`, `TechnicalDomainException` in `core.exception`.
+- Domain exceptions extend these bases with descriptive names (e.g., `ClusterNotFoundException`).
+
+## 9. REST Layer Conventions
+
+- JAX-RS: `@Path`, `@GET`/`@POST`/etc., `@Produces(MediaType.APPLICATION_JSON)`.
+- Inject with `@Inject` (Jakarta), not `@Autowired`.
+- **REST resources call use cases** — never Repository directly.
```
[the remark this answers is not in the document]
FuYu3699: Accepted — now "never repositories directly" (the older host file already used the plural). Fixed in 67efa5e624.

.ai/rules/gamma-clean-architecture.md
```diff
@@ -0,0 +1,196 @@
...
+- Extend `AbstractResourceTest` for the Jersey test harness.
+
+### Architecture tests (ArchUnit)
+
+- Create ArchUnit tests enforcing:
+    - Naming conventions (suffixes match table in section 2).
+    - Use cases reside in `use_case` packages.
+    - Use cases do not depend on other use cases.
+    - Core layer only depends on allowed packages (section 5).
+    - Enforce this dependency direction: REST -> core -> infra.
```
ytvnr: This reads as "core depends on infra", which section 5 of this same rule forbids, and which `OnionArchitectureTest` actively fails on.

The four rules enforced in `gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/architecture/OnionArchitectureTest.java` are: core must not depend on infra, core must not depend on rest resources, rest resources must not depend on infra, and infra must not depend on rest resources. So the direction is inward, not a chain.

```suggestion
    - Enforce the inward dependency direction `rest -> core <- infra`: core depends on neither rest nor infra, rest does not reach into infra, and infra does not reach into rest.
```

The wording is inherited from the old `gravitee-gamma-module-apim/AGENTS.md` rather than introduced here, but this file is about to become the single source for three modules, so it is a cheap fix to make on the way through.

.ai/rules/gravitee-markdown.md
```diff
@@ -0,0 +1,32 @@
...
+- Use `slot()` (or the project's slot mixin) for slotted content styling; do not rely on global element selectors.
+- Forward component styles and overrides through the library's SCSS public API so consumers import one entry point (e.g. `@use '.../public-api'`).
+
+## Storybook
+
+- Stories use the component's public API for overrides — the same entry point and mixin/tokens as the main app; do not duplicate override logic in stories.
+
+## Naming
+
+- Use the `gmd-*` prefix only for selectors and component names in this library. Never `gd-*` (reserved for gravitee-dashboard) or other prefixes for GMD-specific UI.
```
ytvnr: Small tension with `portal-webui-next.md`, which tells the main app to use `gmd-*` for GMD wrappers and overrides. "only for selectors and component names in this library" reads as forbidding that.

Dropping the exclusivity and keeping the real constraint would resolve it:

```suggestion
- Never use `gd-*` (reserved for gravitee-dashboard) or other prefixes for GMD-specific UI.
```

ytvnr (approved):
Reviewed the ten authored rules under `.ai/` as content, and spot-checked the generated output (root, one Java module, one frontend module, the React gamma console, the portal). CI is green including the Nx prettier gate.

Approving. The sequencing rule (declare a module only in the PR that generates its file) is the right call given the measured trap, coverage is complete, and the deliberately-not-a-module list is justified in the authoring rule rather than left as tribal knowledge. Side note: I checked `gravitee-gamma-control-plane-webui` and it is React (`src/bootstrap.tsx`), so TypeScript-only generation there is correct, not a missed Angular detection.

Two inline comments, plus notes below. Only the first is worth acting on before merge.

## Worth fixing before merge

See the inline comment on `gamma-clean-architecture.md`: the ArchUnit dependency direction contradicts section 5 of that same rule and the real `OnionArchitectureTest`. It predates this PR (carried verbatim from `gravitee-gamma-module-apim/AGENTS.md` on master, unchanged since 094a7a6d), so it is not a regression. But it is exactly the class of drifted claim the migration set out to correct, and this PR is the moment that line becomes the single source for three modules.

## Description has gone stale against the last two commits

These numbers were correct at 67efa5e6 and went stale at e93daf59, when the four remaining modules were declared:

- "Root `AGENTS.md` drops from 307 to 199 lines" should read 203. The "The numbers, honestly" section already says 203, so the description disagrees with itself.
- "explicit exceptions for the 19 committed shims" should read 23. `.gitignore` carries root plus `.ai/` plus the 21 modules.
- "One cosmetic quirk: the root's `## Modules` list renders under the last rule's heading" can be dropped. 3a5a20cd regenerated it as a top-level `# Modules`, so the hub fix already landed.

## Minor, take or leave

- `.ai/rules/ai-context-authoring.md` says "`gbuddy check` verifies everything matches". The Confluence page makes the sharper point that `check` proves less than a regeneration: it only shows nobody hand-edited since the last one. Worth aligning the rule with the doc, since the rule is what an agent reads.
- Every generated header reads "Generated by gravitee-ai" while the docs and the linked command are `gbuddy`. Two names for one tool, in the artifact developers see most often.
- `gravitee-apim-portal-webui` now receives the full shared Angular rule in its follow-the-repo form. That is the legacy portal being replaced by portal-webui-next. Intentional, or would a scoped "maintenance only" line serve better there?
- Related to the section-numbering promise in the gamma rule, and out of scope here: `OnionArchitectureTest.java` cites "AGENTS.md section 1" for a rule that lives in section 10.
- Hub-side typo, visible in the generated root at line 133: "hubs rules for `product: gamma`".

.ai/rules/gamma-clean-architecture.md
FuYu3699: Accepted. Your wording shipped verbatim in the source rule and all three generated gamma files. The same commit corrects the test's own javadoc, which cited section 1 and repeated the left-to-right chain; it now cites section 10 and states the inward form. Landed as 6d5d552c03 after the rebase.

.ai/rules/gravitee-markdown.md
FuYu3699: Accepted with one adjustment: the exclusivity is gone, but the positive sentence stays. The library's gmd-* convention exists nowhere else in the rule, so deleting the whole line would have dropped the naming rule along with the conflict. Landed as 6d5d552c03 after the rebase.

FuYu3699:
Dispositions for the review-body notes, all landed (CI green on the current head, 46/46):

- **Description numbers**: fixed. The root line count now reads 203, the committed shim count reads 23, and the obsolete sentence about the Modules heading quirk is dropped. One more stale count turned up on the way: the review guide said ten rule files, and there are now 12.
- **`gbuddy check` claim in the authoring rule**: taken. The rule now states exactly what check proves (no hand edits since the last regeneration) and adds the `generated.yaml` merge-conflict guidance: never hand-resolve, rebase and re-run `gbuddy setup`.
- **`OnionArchitectureTest` section citation**: taken here rather than deferred. We maintain this content now, and it is one line in one place. The javadoc now cites section 10 and states the inward direction.
- **"Generated by gravitee-ai" vs gbuddy**: staying, deliberately. Setup recognises previously generated files by byte-matching that exact marker (one shared constant hub-side), so a rename needs a two-step migration. It is on the post-migration improvement list.
- **Legacy portal-webui getting the full Angular rule**: deliberate for now. Declaring it uniformly beat a special case, and the stricter modern-Angular absolutes stay off it. Whether it deserves a scoped maintenance-only line is on the post-migration content batch.
- **"hubs rules" typo at root line 133**: hub-side. Fixed in gravitee-io/gravitee-ai-foundation#66 together with the same left-to-right direction phrasing in the hub's own gamma summary. APIM picks both up at the next regeneration.

Also in this round, after the rebase onto master: the two blocks master added to the hand-written gamma host file during review are carried into the rules. The core-layer logging guardrails went to the shared gamma rule, whose mechanisms are repository-wide, so all three gamma module files carry them. The exception-mapper fall-through trap went to the host rule. Your page comments produced two tickets: [BX-289](https://gravitee.atlassian.net/browse/BX-289) (the CI check that verifies generated context) and [BX-290](https://gravitee.atlassian.net/browse/BX-290) (automatic hub updates).




[BX-289]: https://gravitee.atlassian.net/browse/BX-289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #19048, opened 2026-08-08
feat(observability): read one authz decision by its event id

podlesrafal:
> Stacked on #19044 — base is `authz-decision-filter`, so this diff is only the delta. Merge that one first.
>
> Console counterpart: [gravitee-gamma-module-authz#118](https://github.com/gravitee-io/gravitee-gamma-module-authz/pull/118).

## Summary

Clicking a decision row opens a drawer that errors:

> Log detail is not supported. The data source does not implement getById.

The panel it would render is already written in full — `config/log-detail.tsx` in the
console describes decision and search sections, reasons, matched policies and search
results. It has never had anything to show, because nothing serves a single decision.
This adds that read.

## Why not the existing detail endpoint

`GET /observability/logs/{requestId}` resolves a **connection log**. For a decision that is
the wrong record, in the wrong index, and often absent entirely — an AuthZEN call has no
HTTP exchange behind it.

More fundamentally, **the request id is not a key here**. A batch stamps every decision it
contains with the same one; the decisions screen renders those as `1/2` and `2/2`. The key
is the event id, which is already what the table uses for its row id. Hence its own path
rather than a parameter bolted onto the request-log detail.

## Changes

- `MetricsRepository.findAuthzDecisionLog(queryContext, apiId, eventId)`, with a term query
  on `event-id` reusing the existing source mapper. Implemented for elasticsearch and noop.
- `AuthzDecisionLogsCrudService.findDecisionLog` over the core-owned record.
- `ObservabilityLogsDataPort.getDecision` and `GetAuthzDecisionUseCase`.
- `GET /observability/logs/decisions/{eventId}?apiId=`, returning the same `LogEntryDto` a
  row has — so the console maps the response through the same translation it uses for the
  table rather than keeping a second contract that can drift.

## API surface

|---|---|---|---|---|
| `/environments/{envId}/observability/logs/decisions/{eventId}` | `GET` | New. `apiId` is required. | `?apiId=api-1` | `200` with a `LogEntryDto` carrying the `authz` object; `404` when absent or not permitted; `400` without `apiId` |

Additive: a new path under an existing resource, nothing else moves.

## Reviewer notes

**The api scope is the security control.** The adapter resolves the api through the caller's
accessible set before reading, so an api they cannot see yields empty, and the resource
collapses that into the same `404` as "no such decision". A known event id from another api
cannot be probed through this path — same posture as the request-log detail, which is why
`apiId` is required rather than inferred.

**Null versus throw, on the console side.** A `404` becomes `null` so the drawer can say the
decision is gone; any other status throws, so it never renders a decision assembled from an
error body.

## Test plan

- [ ] On *Observability → Decisions*, click a row: the drawer opens and shows the decision —
      outcome, subject, action, resource, caller, PDP, eval time, request id.
- [ ] Click a row belonging to a batch (`1/2`): the drawer shows that decision, and clicking
      `2/2` shows the *other* one. Before this change the request id could not tell them apart.
- [ ] Click a search record: the drawer opens on its Search section with the result list.
- [ ] Reload the page with a drawer open: it still resolves, since the read is server-side.
- [ ] As a user without `API_LOG[READ]` on that api: `404`, indistinguishable from a missing
      decision.

## Verification

`LogsResourceTest` (14, five new: response shape, forwarding, missing `apiId`, absent
decision, permission denial collapsing to the same 404), plus the modules it touches green:
repository-api, repository-elasticsearch, repository-noop, rest-api-service, gamma-rest-api.
Console side 1,189 with four new for the source, typecheck and eslint clean.

Not yet clicked in a running console — the endpoint and the mapping are covered by the tests
above, but the drawer itself has not been opened against a live stack.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/crud_service/log/AuthzDecisionLogsCrudServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/crud_service/log/AuthzDecisionLogsCrudServiceImpl.java
@@ -27,6 +27,7 @@ import io.gravitee.rest.api.model.v4.log.SearchLogsResponse;
 import io.gravitee.rest.api.service.common.ExecutionContext;
 import io.gravitee.rest.api.service.exceptions.TechnicalManagementException;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import lombok.CustomLog;
 import org.springframework.context.annotation.Lazy;
@@ -74,6 +75,22 @@ class AuthzDecisionLogsCrudServiceImpl implements AuthzDecisionLogsCrudService {
         } catch (AnalyticsException e) {
             log.error("An error occurs while trying to search authz decision logs [apiIds={}]", apiIds, e);
             throw new TechnicalManagementException("Unable to search authz decision logs", e);
+        }
+    }
+
+    @Override
+    public Optional<AuthzDecisionLog> findDecisionLog(ExecutionContext executionContext, String apiId, String eventId) {
+        try {
+            return metricsRepository
+                .findAuthzDecisionLog(
+                    new QueryContext(executionContext.getOrganizationId(), executionContext.getEnvironmentId()),
+                    apiId,
+                    eventId
+                )
+                .map(AuthzDecisionLogsCrudServiceImpl::toDomain);
+        } catch (AnalyticsException e) {
+            log.error("An error occurs while trying to find authz decision [apiId={}, eventId={}]", apiId, eventId, e);
+            throw new TechnicalManagementException("Unable to find authz decision " + eventId, e);
         }
     }
 
--- a/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
+++ b/gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
@@ -123,6 +123,19 @@ public class ObservabilityLogsDataPortAdapter implements ObservabilityLogsDataPo
     }
 
     @Override
+    public Optional<AccessibleApi> loadAccessibleApi(String organizationId, String environmentId, String apiId) {
+        var auditInfo = currentAuditInfo(organizationId, environmentId);
+        var userContext = userContextLoader.loadApi(new UserContext(auditInfo), apiId);
+
+        return userContext
+            .apis()
+            .orElseGet(Collections::emptyList)
+            .stream()
+            .findFirst()
+            .map(api -> new AccessibleApi(api.getId(), api.getName(), toGammaApiType(api.getType())));
+    }
+
+    @Override
     public LogsPage searchLogs(String organizationId, String environmentId, LogsSearchQuery query) {
         var executionContext = new ExecutionContext(organizationId, environmentId);
         var pageable = new PageableImpl(query.page(), query.perPage());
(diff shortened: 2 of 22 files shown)


gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
```diff
@@ -420,6 +420,26 @@ private LogsPage searchDecisionLogs(ExecutionContext executionContext, LogsSearc
         return new LogsPage(entries, result.total());
     }
 
+    @Override
+    public Optional<LogEntry> getDecision(String organizationId, String environmentId, String apiId, String eventId) {
+        // Resolving the api through the accessible set is what scopes this read: an api the caller
+        // cannot see yields empty, which the resource collapses into the same 404 as "no such decision".
+        var accessible = loadAccessibleApis(organizationId, environmentId)
```
sclarkamlab: `loadAccessibleApis` pulls back every API in the environment and then this throws all but one away. For an admin that's an `apiRepository.search` across the whole env, for everyone else it's `findApiIdsByUserId` (full role/membership resolution) plus the fetch, and it runs every time someone opens a decision. Search pays that once per page which is fine, here it's for one row.

Not saying drop the guard, just narrow the query. `ApiCriteria` takes `environmentId` plus `ids` (same builder `UserContextLoaderImpl` uses) so a single-api lookup gets you the same scoping and the same name/type. Also lets the whole thing collapse into one `.map()` and the three `accessible.get()` calls go with it.

sclarkamlab (commented):
Ran a couple of review passes over this and read it through myself. One thing from me, rest looks good.

Test coverage is thorough at every layer and I like that the event id vs request id reasoning is written into the code and the spec instead of only living in the PR description.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
podlesrafal: Good catch, fixed in 660315874d.

`UserContextLoader` gains `loadApi`, the single-api form, and `getDecision` collapses into one `flatMap` as you suggested. The three `accessible.get()` calls are gone with it.

One thing I want to flag, because taking the suggestion literally would have cost us something. Narrowing to `ApiCriteria(environmentId, ids=[apiId])` on its own is not equivalent to what `loadApis` does: for a non-admin, `loadApis` sets `ids` to the result of `findApiIdsByUserId`, so the criteria *is* the membership filter. Replacing that with `ids=[apiId]` would have quietly downgraded the guard from "in the caller's accessible set" to "in the environment". Not a hole, since the resource already checks `API_LOG[READ]`, but this port is deliberately a second independent barrier and I did not want to flatten it.

So `loadApi` keeps the membership check and only narrows the fetch:

```java
if (!isAdmin()) {
    Set<String> userApiIds = apiAuthorizationService.findApiIdsByUserId(executionContext, userId, null, true);
    if (!userApiIds.contains(apiId)) {
        return context.withApis(List.of()).withApiNamesById(Map.of());
    }
}
var criteria = new ApiCriteria.Builder().environmentId(environmentId).ids(Set.of(apiId)).build();
```

Worth being explicit about the payoff: an admin saves the whole environment scan, a non-admin saves the fetch but still pays `findApiIdsByUserId`. Removing that too would mean dropping the membership check, which is exactly what we do not want here.

Tests: `UserContextLoaderImplTest.LoadOneApi` covers all three paths, including a non-admin asking for an api that exists in the environment but that they are not a member of, asserting the repository is never touched. `ObservabilityLogsDataPortAdapterTest` gains a test asserting `loadApi` is used and `loadApis` never is, so the narrowing cannot silently regress.

Pushed as a separate commit rather than amending, so the delta is reviewable; the merge queue squashes anyway.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/crud_service/log/AuthzDecisionLogsCrudServiceImpl.java
```diff
@@ -77,6 +78,22 @@ public SearchLogsResponse<AuthzDecisionLog> searchDecisionLogs(
...
+        try {
+            return metricsRepository
+                .findAuthzDecisionLog(
+                    new QueryContext(executionContext.getOrganizationId(), executionContext.getEnvironmentId()),
+                    apiId,
+                    eventId
+                )
+                .map(AuthzDecisionLogsCrudServiceImpl::toDomain);
+        } catch (AnalyticsException e) {
+            log.error("An error occurs while trying to find authz decision [apiId={}, eventId={}]", apiId, eventId, e);
```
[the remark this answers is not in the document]
podlesrafal: @AikidoSec ignore: this is the existing convention in the file, not a pattern introduced here. `searchDecisionLogs` in the same class already does `log.error(...)` then `throw new TechnicalManagementException(..., e)` on `AnalyticsException`. Changing only the new method would leave two different error-handling styles side by side in one class, which is worse than the duplication being flagged. If the log-and-rethrow pattern should go, it should go for the whole file (or module) in its own change.

sclarkamlab (approved):

PR #19058, opened 2026-08-10
fix: avoid N+1 certificate lookup when listing applications

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-14337

## Description

`GET /applications` regressed from ~500ms to 10-20s in 4.11.x, scaling with app count. Multiple customers affected (APIM-14337).

Initial theory was a missing MongoDB index on `page_revisions` — turned out to be wrong. On the customer's actual DB, the index already existed and `$indexStats`/profiler showed the DB doing 1ms of work for a 3s response. The bottleneck isn't the database.

### Root cause

`convert()` (called once per application from `convertToList`) unconditionally fetches that app's client certificate:

```java
applicationEntity.setSettings(getSettings(executionContext, application, true)); // always true
```

### Fix
Threaded a `fetchCertificate` flag through `convert()/convertApplication()`. List path passes `false` (batched fetch covers it); `create/update/restore/renewClientSecret/findById` still pass `true` — those are single-item lookups where the per-item fetch is correct.

### Test
Added `findByIdsAndStatus_shouldNotFetchCertificatesPerApplication`: asserts the list path calls the batched `findByApplicationIdsAndStatuses` and never the singular `findByApplicationIdAndStatuses`.

### Note
Supersedes #18772 (index-based fix) — that addressed a symptom, not the actual bottleneck.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ApplicationServiceImplTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ApplicationServiceImplTest.java
@@ -19,15 +19,36 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.jupiter.api.Assertions.assertAll;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.ArgumentMatchers.isNull;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import io.gravitee.apim.core.application_certificate.crud_service.ClientCertificateCrudService;
+import io.gravitee.apim.core.application_certificate.model.ClientCertificate;
+import io.gravitee.apim.core.application_certificate.model.ClientCertificateStatus;
+import io.gravitee.common.data.domain.Page;
+import io.gravitee.repository.exceptions.TechnicalException;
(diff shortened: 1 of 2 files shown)


ankita-gupta21:
Supersedes #18772 (index-based fix) — that addressed a symptom, not the actual bottleneck.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ApplicationServiceImplTest.java
aqibmohammadkhan, line 1: - Test doesn’t assert certificates still appear on list items
`findByIdsAndStatus_shouldNotFetchCertificatesPerApplication` stubs the batch to List.of() and only checks call shape. A stubbed cert + assert that settings.tls.clientCertificate is set would lock the contract, not just “no N+1”.

- Varargs Mockito matcher may not catch singular calls reliably
Interface is findByApplicationIdAndStatuses(String, ClientCertificateStatus...), but verify uses anyString(), any(). Prefer matching the real arity, e.g. both status enums, or any(ClientCertificateStatus[].class), so a regression can’t slip past never().

aqibmohammadkhan (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/ApplicationServiceImplTest.java
ankita-gupta21: **Contract not locked** — the test now stubs a real ClientCertificate for app1 in the batched response and asserts it lands on app1Item.getSettings().getTls().getClientCertificate(), while confirming app2 (no certificate) gets null TLS settings.
**Fragile vararg matcher** — both verify calls now use any(ClientCertificateStatus[].class), which matches the vararg's underlying array directly regardless of how many statuses are passed, instead of the arity-sensitive any() that could silently fail to bind.

PR #19062, opened 2026-08-10
feat(portal): filter navigation search by category and expose visible portal categories

dev-marek:
**Issue**

https://gravitee.atlassian.net/browse/PORTAL-22

**Purpose**

Let portal consumers filter navigation search results by category and discover which categories are visible to them.

**Summary of changes**

- `GET /portal-navigation-items/_search` accepts an optional `categoryId` query param, filtering to items whose `categoryIds` array contains the given value. Implemented at the repository layer (Mongo `where("categoryIds").is(...)`, JDBC subquery against the `portal_navigation_item_categories` join table) and threaded through the domain service and use case.
- New `GET /portal-categories` endpoint returning only visible (`visible = true`) portal categories for the current environment.
- Fixed a bug caught during manual testing: the new `/portal-categories` endpoint was unreachable anonymously because the portal-security module uses a URL allow-list (default-deny) and the new path wasn't added to it — added a `permitAll()` matcher alongside the existing one for `/portal-navigation-items/**`.

**Out of scope**

- Exposing the category `visible` flag itself in the API response (only pre-filtered results are returned).
- Filtering `_search` by more than one category at a time.

**Testing**

Unit + integration tests (repository conformance for Mongo/JDBC, domain service, use case, REST resource) — all passing.

Manually verified against a local Docker Compose stack (Mongo-backed):

```bash
# No categoryId param — unaffected baseline
curl -s "http://localhost:8083/portal/environments/DEFAULT/portal-navigation-items/_search?type=api"
# -> 200, all API nav items returned

# Known categoryId — filters to matching items only
curl -s "http://localhost:8083/portal/environments/DEFAULT/portal-navigation-items/_search?type=api&categoryId=<uuid>"
# -> 200, only items tagged with that category

# Unknown categoryId — empty result
curl -s "http://localhost:8083/portal/environments/DEFAULT/portal-navigation-items/_search?type=api&categoryId=00000000-0000-0000-0000-000000000000"
# -> 200, "data": []

# Visible categories only
curl -s "http://localhost:8083/portal/environments/DEFAULT/portal-categories"
# -> 200, returns visible categories; a category manually set to visible:false in Mongo was confirmed excluded
```

**Additional context**

_None._

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoPortalNavigationItemRepository.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoPortalNavigationItemRepository.java
@@ -142,6 +142,9 @@ public class MongoPortalNavigationItemRepository implements PortalNavigationItem
             if (criteria.getUseAutoFetch() != null) {
                 query.addCriteria(where("useAutoFetch").is(criteria.getUseAutoFetch()));
             }
+            if (hasText(criteria.getCategoryId())) {
+                query.addCriteria(where("categoryIds").is(criteria.getCategoryId()));
+            }
         }
         return query;
     }
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/PortalCategoriesResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/PortalCategoriesResource.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.portal.rest.resource;
+
+import static io.gravitee.rest.api.service.common.GraviteeContext.getExecutionContext;
+
+import io.gravitee.apim.core.portal_category.use_case.GetVisiblePortalCategoriesUseCase;
+import io.gravitee.common.http.MediaType;
+import io.gravitee.rest.api.portal.rest.mapper.PortalCategoryMapper;
+import io.gravitee.rest.api.portal.rest.model.PortalCategoriesResponse;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.Response;
+
+public class PortalCategoriesResource extends AbstractResource {
+
+    @Inject
+    private GetVisiblePortalCategoriesUseCase getVisiblePortalCategoriesUseCase;
+
+    private final PortalCategoryMapper portalCategoryMapper = PortalCategoryMapper.INSTANCE;
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getPortalCategories() {
+        var executionContext = getExecutionContext();
+        var output = getVisiblePortalCategoriesUseCase.execute(
+            new GetVisiblePortalCategoriesUseCase.Input(executionContext.getEnvironmentId())
+        );
+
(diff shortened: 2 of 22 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/PortalCategoriesResource.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class PortalCategoriesResource extends AbstractResource {
+
+    @Inject
+    private GetVisiblePortalCategoriesUseCase getVisiblePortalCategoriesUseCase;
+
+    private final PortalCategoryMapper portalCategoryMapper = PortalCategoryMapper.INSTANCE;
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
```
jaroslawlakomy-grav, line 39: The OpenAPI operation is tagged Portal, but the handling resource has no matching @Tag(name = "Portal").

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationApiVisibilityDomainServiceTest.java
```diff
@@ -325,6 +371,22 @@ private PortalNavigationApi publishedApiNavItem(String apiId, PortalVisibility v
...
+            .title("Nav for " + apiId)
+            .segment(PortalNavigationItem.slugify("Nav for " + apiId).value())
+            .area(PortalArea.TOP_NAVBAR)
+            .order(0)
+            .apiId(apiId)
+            .published(true)
+            .visibility(visibility)
+            .categoryIds(categoryIds.stream().map(PortalCategoryId::of).toList())
+            .build();
+    }
```
jaroslawlakomy-grav: Could we extend PortalNavigationItemFixtures with a category-aware API variant and use it here?

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoPortalNavigationItemRepository.java
```diff
@@ -142,6 +142,9 @@ private Query buildQuery(PortalNavigationItemCriteria criteria) {
             if (criteria.getUseAutoFetch() != null) {
                 query.addCriteria(where("useAutoFetch").is(criteria.getUseAutoFetch()));
             }
+            if (hasText(criteria.getCategoryId())) {
+                query.addCriteria(where("categoryIds").is(criteria.getCategoryId()));
```
jaroslawlakomy-grav, line 146: This adds a public equality filter on the categoryIds array, but the Mongo collection has no index containing that field. The production query therefore scans all navigation items in the environment for each category request. Could we add an index upgrader for at least environmentId + categoryIds?

jaroslawlakomy-grav (commented):
Somę minor comments and I see some SonarCloud warning

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/PortalCategoriesResource.java
dev-marek: Added `@Tag(name = "Portal")` to `PortalCategoriesResource`, matching the annotation used on other resources with a non-default Swagger grouping (e.g. `PortalUIBootstrapResource`). Fixed in e9cc9ddca0.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationApiVisibilityDomainServiceTest.java
dev-marek: Added `PortalNavigationItemFixtures.anApi(apiId, visibility, categoryIds)` and switched this test to use it, removing the local `publishedApiNavItemWithCategories` helper. Fixed in e9cc9ddca0.

gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/MongoPortalNavigationItemRepository.java
dev-marek: Added `PortalNavigationItemCategoryIdsIndexUpgrader` (compound `environmentId + categoryIds` index on `portal_navigation_items`), following the same pattern as the other index upgraders in that package (e.g. `PortalNavigationItemEnvironmentIdIndexUpgrader`). Fixed in e9cc9ddca0.

jaroslawlakomy-grav (approved):

PR #19067, opened 2026-08-10
fix: derive service-account sourceId from lastname or email

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-14832

### Root cause
Three compounding issues:
1. `UserServiceImpl.createAndSendEmail()` derives `sourceId` for a service account from `lastname` only. If `lastname` is missing, `sourceId` is set to `null` — it never falls back further to `email`, even though `email` is the one field realistically guaranteed to be present.
2. `NewPreRegisterUserEntityValidator.isValid()` only checked that `firstname` was null for a service account — it never verified that anything usable for `sourceId` derivation was actually present.
3. `MongoUserRepository.findBySource()` returns `Optional.empty()` for a `null` `sourceId`, so the pre-create duplicate check never caught it — the conflict only surfaced later as a unique-index violation, wrapped into a generic `TechnicalManagementException` (500).

Side effect: a legacy null-`sourceId` service account could also trigger an NPE in `UserServiceImpl.isServiceAccount()` (reachable via the password-reset path), since it called `.equalsIgnoreCase()` on `getSourceId()` with no null guard.

### Fix
Rather than forcing callers to always supply `lastname` (which would break existing integrations — see the PostNord case in the ticket comments, where accounts are created with only `email` + `service: true`), the fix widens the `sourceId` derivation instead of narrowing the request contract:

- **`UserServiceImpl`**: `sourceId` is now resolved as `sourceId → lastname → email`, in that priority order. Since `email` is effectively always present, `sourceId` can no longer end up `null` for a service account.
- **`NewPreRegisterUserEntityValidator`**: a service account request is now rejected (`400`) only when `sourceId`, `lastname`, **and** `email` are all blank — the one case where an identifier truly cannot be derived. Every request shape that used to work (including email-only) keeps working.
- **`UserServiceImpl.isServiceAccount()`**: null-guards `getSourceId()` to prevent the NPE on already-corrupted records from before this fix.

### Behavior after the fix
Walking through the ticket's exact reproduction:
1. `{"email": "sa-one@example.com", "service": true}` → no `lastname`/`sourceId` → `sourceId` derived from email → **succeeds**.
2. `{"email": "sa-two@example.com", "service": true}` → different email → different derived `sourceId` → **succeeds, no 500**.
3. Posting the *same* email twice → same derived `sourceId` on both attempts → the pre-create check now catches it → clean **400 "user already exists"** instead of a raw 500.
4. A service account with **no** `email`, `lastname`, or `sourceId` at all → the only case that still gets a **400** (nothing to derive from).

### Before Fix
https://github.com/user-attachments/assets/70c5654f-b563-40c6-b5f7-a03b4491f2b6

### After Fix
https://github.com/user-attachments/assets/7002f884-a827-44fd-91ae-af317cb54d28

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/UsersResourceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/UsersResourceTest.java
@@ -16,6 +16,7 @@
 package io.gravitee.rest.api.management.rest.resource;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.when;
@@ -59,8 +60,54 @@ public class UsersResourceTest extends AbstractResourceTest {
     }
 
     @Test
-    public void shouldCreateServiceAccountWithoutEmail() {
+    public void shouldNotCreateServiceAccountWithoutEmailLastnameOrSourceId() {
+        // sourceId is resolved as sourceId -> lastname -> email; with none of the three set,
+        // sourceId cannot be derived, so the request must be rejected (APIM-14832)
         NewPreRegisterUserEntity newPreRegisterUserEntity = new NewPreRegisterUserEntity();
+        newPreRegisterUserEntity.setService(true);
+
+        when(userService.create(any(), any())).thenReturn(new UserEntity());
+
+        final Response response = orgTarget().request().post(Entity.json(newPreRegisterUserEntity));
+        assertEquals(HttpStatusCode.BAD_REQUEST_400, response.getStatus());
+        assertTrue(
+            response
+                .readEntity(String.class)
+                .contains("a service account requires a sourceId, lastname or email to derive a unique identifier")
+        );
+    }
+
+    @Test
+    public void shouldNotCreateServiceAccountWithFirstname() {
+        // a service account must not have a firstname, even with a usable identifier present (APIM-14832)
+        NewPreRegisterUserEntity newPreRegisterUserEntity = new NewPreRegisterUserEntity();
+        newPreRegisterUserEntity.setFirstname("SA");
+        newPreRegisterUserEntity.setEmail("mail@fake.fake");
+        newPreRegisterUserEntity.setService(true);
+
+        when(userService.create(any(), any())).thenReturn(new UserEntity());
+
+        final Response response = orgTarget().request().post(Entity.json(newPreRegisterUserEntity));
+        assertEquals(HttpStatusCode.BAD_REQUEST_400, response.getStatus());
+        assertTrue(response.readEntity(String.class).contains("a service account must not have a firstname"));
+    }
+
+    @Test
+    public void shouldCreateServiceAccountWithLastnameOnly() {
+        NewPreRegisterUserEntity newPreRegisterUserEntity = new NewPreRegisterUserEntity();
+        newPreRegisterUserEntity.setLastname("SA_LASTNAME");
+        newPreRegisterUserEntity.setService(true);
+
+        when(userService.create(any(), any())).thenReturn(new UserEntity());
+
+        final Response response = orgTarget().request().post(Entity.json(newPreRegisterUserEntity));
+        assertEquals(HttpStatusCode.OK_200, response.getStatus());
+    }
+
+    @Test
+    public void shouldCreateServiceAccountWithSourceIdOnly() {
+        NewPreRegisterUserEntity newPreRegisterUserEntity = new NewPreRegisterUserEntity();
+        newPreRegisterUserEntity.setSourceId("SA_SOURCE_ID");
         newPreRegisterUserEntity.setService(true);
 
         when(userService.create(any(), any())).thenReturn(new UserEntity());
@@ -86,5 +133,6 @@ public class UsersResourceTest extends AbstractResourceTest {
 
         final Response response = orgTarget().request().post(Entity.json(newPreRegisterUserEntity));
         assertEquals(HttpStatusCode.BAD_REQUEST_400, response.getStatus());
+        assertTrue(response.readEntity(String.class).contains("email is required"));
     }
 }
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/NewExternalUserEntity.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/NewExternalUserEntity.java
@@ -16,6 +16,7 @@
 package io.gravitee.rest.api.model;
 
 import io.gravitee.rest.api.sanitizer.HtmlSanitizer;
+import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.constraints.NotNull;
 import java.util.Map;
 
@@ -28,17 +29,20 @@ public class NewExternalUserEntity {
     /**
      * The user first name
      */
+    @Schema(description = "The user first name.")
     private String firstname;
 
     /**
      * The user last name
      */
+    @Schema(description = "The user last name.")
     private String lastname;
 
     /**
      * The user email
      */
     @NotNull
+    @Schema(description = "The user email.")
     private String email;
 
     /**
@@ -54,6 +58,7 @@ public class NewExternalUserEntity {
     /**
      * The user source reference
      */
+    @Schema(description = "The unique identifier of the user for the given source.")
     private String sourceId;
 
     private Boolean newsletter;
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NewPreRegisterUserEntityValidator.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NewPreRegisterUserEntityValidator.java
@@ -18,12 +18,18 @@ package io.gravitee.rest.api.validator;
 import io.gravitee.rest.api.model.NewPreRegisterUserEntity;
 import jakarta.validation.ConstraintValidator;
 import jakarta.validation.ConstraintValidatorContext;
+import org.apache.commons.lang3.StringUtils;
 
 /**
  * @author Yann TAVERNIER (yann.tavernier at graviteesource.com)
  * @author GraviteeSource Team
  */
 public class NewPreRegisterUserEntityValidator implements ConstraintValidator<ValidNewPreRegisterUser, NewPreRegisterUserEntity> {
+
+    static final String SERVICE_ACCOUNT_FIRSTNAME_MESSAGE = "a service account must not have a firstname";
+    static final String SERVICE_ACCOUNT_IDENTIFIER_MESSAGE =
+        "a service account requires a sourceId, lastname or email to derive a unique identifier";
+    static final String EMAIL_REQUIRED_MESSAGE = "email is required";
 
     @Override
     public void initialize(ValidNewPreRegisterUser constraintAnnotation) {
@@ -32,15 +38,37 @@ public class NewPreRegisterUserEntityValidator implements ConstraintValidator<Va
 
     /**
      * Is valid when:
-     * - Is service account and firstName is null.
-     * - Is not service account and have an email
+     * - Is service account, firstName is blank, and at least one of sourceId, lastName or email is non-blank
+     *   (the service layer derives sourceId as sourceId -> lastName -> email, so one of the three must be present).
+     * - Is not service account and have a non-blank email
      */
     @Override
     public boolean isValid(NewPreRegisterUserEntity value, ConstraintValidatorContext context) {
         if (Boolean.TRUE.equals(value.isService())) {
-            return value.getFirstname() == null;
+            if (StringUtils.isNotBlank(value.getFirstname())) {
+                addViolation(context, SERVICE_ACCOUNT_FIRSTNAME_MESSAGE);
+                return false;
+            }
+            if (
+                StringUtils.isBlank(value.getSourceId()) &&
+                StringUtils.isBlank(value.getLastname()) &&
+                StringUtils.isBlank(value.getEmail())
+            ) {
+                addViolation(context, SERVICE_ACCOUNT_IDENTIFIER_MESSAGE);
+                return false;
+            }
+            return true;
         } else {
-            return value.getEmail() != null;
+            if (StringUtils.isBlank(value.getEmail())) {
+                addViolation(context, EMAIL_REQUIRED_MESSAGE);
+                return false;
+            }
+            return true;
         }
+    }
+
+    private void addViolation(ConstraintValidatorContext context, String message) {
+        context.disableDefaultConstraintViolation();
+        context.buildConstraintViolationWithTemplate(message).addConstraintViolation();
     }
 }
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
@@ -38,6 +38,7 @@ import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.isEmpty;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 import com.auth0.jwt.JWT;
 import com.auth0.jwt.JWTVerifier;
@@ -896,13 +897,10 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
     }
 
     private boolean isServiceAccount(User user) {
-        // A service account has the sourceId equals to the lastname or to the email address
-        return (
-            IDP_SOURCE_GRAVITEE.equals(user.getSource()) &&
-            user.getIsServiceAccount() != null &&
-            user.getIsServiceAccount() &&
-            (user.getSourceId().equalsIgnoreCase(user.getEmail()) || user.getSourceId().equalsIgnoreCase(user.getLastname()))
-        );
+        // isServiceAccount is authoritative; a sourceId comparison was a heuristic for records predating the
+        // flag and incorrectly excludes a service account created with an explicit sourceId that differs
+        // from both email and lastname (e.g. sourceId "sa-billing" with email "ops@acme.com").
+        return IDP_SOURCE_GRAVITEE.equals(user.getSource()) && Boolean.TRUE.equals(user.getIsServiceAccount());
     }
 
     /**
@@ -928,13 +926,33 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
         if (isBlank(newExternalUserEntity.getSource())) {
             newExternalUserEntity.setSource(IDP_SOURCE_GRAVITEE);
         } else if (!IDP_SOURCE_GRAVITEE.equals(newExternalUserEntity.getSource())) {
-            // check if IDP exists
             identityProviderService.findById(newExternalUserEntity.getSource());
         }
 
-        if (isBlank(newExternalUserEntity.getSourceId())) {
-            newExternalUserEntity.setSourceId(isServiceUser ? newExternalUserEntity.getLastname() : newExternalUserEntity.getEmail());
+        if (isServiceUser && isBlank(newExternalUserEntity.getFirstname())) {
+            newExternalUserEntity.setFirstname(null);
         }
+
+        final String resolvedSourceId;
+        if (isBlank(newExternalUserEntity.getSourceId())) {
+            if (isServiceUser && isNotBlank(newExternalUserEntity.getLastname())) {
+                resolvedSourceId = newExternalUserEntity.getLastname();
+            } else {
+                resolvedSourceId = newExternalUserEntity.getEmail();
+            }
+        } else {
+            resolvedSourceId = newExternalUserEntity.getSourceId();
+        }
+        final String trimmedSourceId = StringUtils.trim(resolvedSourceId);
+        if (isBlank(trimmedSourceId)) {
+            throw new InvalidUserException(
+                "a service account requires a sourceId, lastname or email to derive a unique identifier",
+                newExternalUserEntity.getSource(),
+                null,
+                organizationId
+            );
+        }
+        newExternalUserEntity.setSourceId(trimmedSourceId);
 
         final Optional<User> optionalUser;
         try {
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
@@ -1347,6 +1347,32 @@ public class UserServiceTest {
         });
     }
 
+    @Test
+    public void shouldNotResetPasswordOnServiceAccountWithNullSourceId() throws TechnicalException {
+        // legacy/corrupted service account persisted with a null sourceId (see APIM-14832): must not NPE,
+        // and the isServiceAccount flag alone must still block the reset - a null sourceId must not be
+        // waved through as "not a service account"
+        when(user.getId()).thenReturn(USER_NAME);
+        when(user.getSource()).thenReturn("gravitee");
+        when(user.getSourceId()).thenReturn(null);
+        when(user.getOrganizationId()).thenReturn(ORGANIZATION);
+        when(user.getIsServiceAccount()).thenReturn(true);
+        when(userRepository.findById(USER_NAME)).thenReturn(of(user));
+
+        assertThrows(ServiceAccountNotManageableException.class, () -> userService.resetPassword(EXECUTION_CONTEXT, USER_NAME));
+    }
+
+    @Test
+    public void shouldNotResetPasswordOnServiceAccountWithMismatchedSourceId() throws TechnicalException {
+        when(user.getSource()).thenReturn("gravitee");
+        when(user.getSourceId()).thenReturn("sa-billing");
+        when(user.getOrganizationId()).thenReturn(ORGANIZATION);
+        when(user.getIsServiceAccount()).thenReturn(true);
+        when(userRepository.findById(USER_NAME)).thenReturn(of(user));
+
+        assertThrows(ServiceAccountNotManageableException.class, () -> userService.resetPassword(EXECUTION_CONTEXT, USER_NAME));
+    }
+
     private String createJWT(long expirationSeconds, String action) {
         Algorithm algorithm = Algorithm.HMAC256(JWT_SECRET);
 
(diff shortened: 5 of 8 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NewPreRegisterUserEntityValidator.java
```diff
@@ -32,13 +33,19 @@ public void initialize(ValidNewPreRegisterUser constraintAnnotation) {
...
      */
     @Override
     public boolean isValid(NewPreRegisterUserEntity value, ConstraintValidatorContext context) {
         if (Boolean.TRUE.equals(value.isService())) {
-            return value.getFirstname() == null;
+            return (
+                value.getFirstname() == null &&
+                (StringUtils.isNotEmpty(value.getSourceId()) ||
+                    StringUtils.isNotEmpty(value.getLastname()) ||
+                    StringUtils.isNotEmpty(value.getEmail()))
```
aqibmohammadkhan: isNotEmpty here doesn't line up with isNotBlank in the service layer, and the gap reopens the original bug.

Payload: service: true, lastname: " ", no sourceId, no email.

Validator: isNotEmpty(" ") → true → passes
createAndSendEmail: isNotBlank(" ") → false → falls through to email, which is null
Result: sourceId persisted as null — same defect, same 500 on the next create
The email-format guard above doesn't catch it either: for a service user with a null email, isNotEmpty(null) is false, so the whole condition short-circuits and nothing is thrown.

Switching to isNotBlank closes the gap and matches the invariant the new comment states ("so it is never left null for a service account"):

```
(StringUtils.isNotBlank(value.getSourceId()) ||
                    StringUtils.isNotBlank(value.getLastname()) ||
                    StringUtils.isNotBlank(value.getEmail()))
```
Worth adding { null, null, " ", null, true, false } to the test matrix to pin it.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -892,6 +893,7 @@ private boolean isServiceAccount(User user) {
             IDP_SOURCE_GRAVITEE.equals(user.getSource()) &&
             user.getIsServiceAccount() != null &&
             user.getIsServiceAccount() &&
+            user.getSourceId() != null &&
```
aqibmohammadkhan: this guard prevents the NPE, but it changes what happens to the account rather than just making it safe.

isServiceAccount() now returns false for a legacy null-sourceId account, so the ServiceAccountNotManageableException check at line 1462 is skipped and the password reset proceeds — a reset email goes out for an account whose isServiceAccount flag is true. The new test locks this in with assertDoesNotThrow(...).

With the fallback in this PR, no new null-sourceId rows can be created, so this only affects already-corrupted records — but those are exactly the accounts sitting in the reporting customer's org today, and they'd go from a 500 to being password-resettable.

The sourceId comparison is a legacy heuristic for records predating the isServiceAccount flag; when the flag says service account, that should stand on its own:

```
private boolean isServiceAccount(User user) {
    return (
        IDP_SOURCE_GRAVITEE.equals(user.getSource()) &&
        user.getIsServiceAccount() != null &&
        user.getIsServiceAccount() &&
        (user.getSourceId() == null ||
            user.getSourceId().equalsIgnoreCase(user.getEmail()) ||
            user.getSourceId().equalsIgnoreCase(user.getLastname()))
    );
}
```

The test would then assert ServiceAccountNotManageableException instead of assertDoesNotThrow — still proves no NPE, and pins the behaviour we actually want.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/UsersResourceTest.java
aqibmohammadkhan, line 1: `shouldCreateServiceAccountWithoutEmail` (line 61) posts service=true with nothing else and asserts 200 — that's now correctly a 400, so the test needs to flip.

`shouldCreateServiceAccountWithEmail` is fine after the switch to the sourceId → lastname → email chain.

aqibmohammadkhan, line 1: the API reference and the error response give no way to work this out - isn't covered.

After this change the remaining 400s for a service account are (a) firstname present and (b) none of sourceId/lastname/email present. Case (a) is exactly what the second customer hit, and it still returns a bare "Pre-registered user is not valid" with nothing pointing at firstname.

Three small additions would close it:

- ConstraintValidatorContext.buildConstraintViolationWithTemplate(...) for each rule  - ConstraintValidationExceptionMapper already surfaces per-violation messages in details[], so this reaches the client with no other plumbing
 - @Schema(description = ...) on firstname / lastname / sourceId / service — swagger-annotations is already a dependency of gravitee-apim-rest-api-model, and the v1 spec is generated from these by swagger-maven-plugin, so it lands in the published console reference
 - @ApiResponse(responseCode = "400", ...) on UsersResource#createUser, which currently documents only 200 and 500

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NewPreRegisterUserEntityValidator.java
ankita-gupta21: Fixed. Switched all three checks to `isNotBlank`, so a whitespace-only `sourceId/lastname/email` no longer passes validation while `createAndSendEmail` treats it as blank and falls through:

```
(StringUtils.isNotBlank(value.getSourceId()) ||
    StringUtils.isNotBlank(value.getLastname()) ||
    StringUtils.isNotBlank(value.getEmail()))
```
Also added your exact repro case to the test matrix` ({ null, null, " ", null, true, false })`, plus whitespace-only sourceId and email variants for the same reason.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
ankita-gupta21: Fixed — the guard was masking the account rather than just fixing the NPE. Switched to your suggested approach: a null `sourceId` now counts as a match instead of an automatic pass, so `ServiceAccountNotManageableException` still fires for legacy corrupted accounts:

```
(user.getSourceId() == null ||
    user.getSourceId().equalsIgnoreCase(user.getEmail()) ||
    user.getSourceId().equalsIgnoreCase(user.getLastname()))
```
Updated the test to assert `ServiceAccountNotManageableException` instead of `assertDoesNotThrow` — still proves no NPE, but now also pins that reset stays blocked for those accounts.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/UsersResourceTest.java
ankita-gupta21: Already Fixed:
`shouldNotCreateServiceAccountWithoutEmailLastnameOrSourceId `and asserts 400:
```
@Test
public void shouldNotCreateServiceAccountWithoutEmailLastnameOrSourceId() {
    NewPreRegisterUserEntity newPreRegisterUserEntity = new NewPreRegisterUserEntity();
    newPreRegisterUserEntity.setService(true);
    ...
    assertEquals(HttpStatusCode.BAD_REQUEST_400, response.getStatus());
}
```
Also added `shouldCreateServiceAccountWithLastnameOnly` and `shouldCreateServiceAccountWithSourceIdOnly` alongside it to cover the two ways sourceId can now be derived.

ankita-gupta21: Addressed all three:

- Per-violation messages via `buildConstraintViolationWithTemplate` — the validator now returns a distinct message for each failure case (firstname set on a service account vs. no usable identifier), so `details[] `no longer just says "Pre-registered user is not valid" with no pointer to which rule failed.
- `@Schema(description = ...) `added on `firstname, lastname, sourceId` (in `NewExternalUserEntity`) and service (in NewPreRegisterUserEntity), documenting the service-account rules directly on the fields the v1 spec generates from.
- `@ApiResponse(responseCode = "400", ...)` added on `UsersResource#createUser`, which previously only documented 200 and 500.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -887,12 +888,16 @@ public UserEntity create(ExecutionContext executionContext, final NewPreRegister
     }
 
     private boolean isServiceAccount(User user) {
-        // A service account has the sourceId equals to the lastname or to the email address
+        // A service account has the sourceId equals to the lastname or to the email address.
```
aqibmohammadkhan: isServiceAccount() was patched for sourceId == null, but it still returns false when sourceId is set and differs from both email and lastname — which is exactly the shape this PR now blesses. shouldCreateServiceAccountWithSourceIdOnly creates it, and the new @Schema on sourceId documents it as supported.

Concretely: {"sourceId": "sa-billing", "email": "ops@acme.com", "service": true} → isServiceAccount() false → the guard at UserServiceImpl.java:1462 is skipped → the reset proceeds and a reset link goes to ops@acme.com, turning a token-only service account into a login-capable one. That path is reachable from the unauthenticated portal endpoint POST /users/_reset_password (portal UsersResource.java:131), keyed by sourceId — the identifier the caller chose and therefore knows. updateServiceAccountStatus (:1405) flips the flag without touching sourceId, so promoted users land in the same state.

The new comment already states the right invariant — "the flag already says it's a service account, so that alone must stand" — but the code still ANDs the heuristic against it. Dropping it implements what the comment claims:

```
private boolean isServiceAccount(User user) {
    // isServiceAccount is authoritative; the sourceId comparison was a heuristic for records
    // predating the flag and lets service accounts with an explicit sourceId slip through.
    return IDP_SOURCE_GRAVITEE.equals(user.getSource()) && Boolean.TRUE.equals(user.getIsServiceAccount());
}
```
That makes the null guard redundant, keeps shouldNotResetPasswordOnServiceAccountWithNullSourceId passing, and wants one more case with sourceId != email/lastname. It widens what ServiceAccountNotManageableException blocks, which is the safe direction. This is pre-existing on master, not introduced here — but the PR both widens the exposed surface and half-fixes it, so finishing it here is cheaper than a follow-up.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NewPreRegisterUserEntityValidator.java
```diff
@@ -18,29 +18,52 @@
...
+                StringUtils.isBlank(value.getSourceId()) &&
+                StringUtils.isBlank(value.getLastname()) &&
+                StringUtils.isBlank(value.getEmail())
+            ) {
+                addViolation(context, SERVICE_ACCOUNT_IDENTIFIER_MESSAGE);
+                return false;
+            }
+            return true;
         } else {
             return value.getEmail() != null;
```
aqibmohammadkhan: Non-service branch has no message. return value.getEmail() != null.
still yields the bare "Pre-registered user is not valid" — inconsistent with the two new per-rule messages, and it's the other 400 a caller can hit. It also uses != null while the service branch uses isBlank, so {"email": " ", "service": false} passes validation and fails later on email format instead.

aqibmohammadkhan (commented):
- firstname != null vs blank. {"firstname": "", "service": true} gets rejected with "must not have a firstname". Pre-existing, but if you're aligning on blank semantics in this file, worth doing here too.

- @Schema scope. The service-account wording sits on NewExternalUserEntity, which is also the body of POST /users/registration (UsersRegistrationResource.java:73), where service doesn't exist — those descriptions will show up there too. Keeping the base descriptions generic and leaving the full rule on NewPreRegisterUserEntity.service + the @ApiResponse avoids it. email is also part of the derivation chain now but has no @Schema.

- No trimming on derivation. {"service": true, "lastname": " sa "} persists sourceId as " sa ", and findBySource is exact-match, so " sa " and "sa" coexist as distinct accounts. A .trim() where sourceId is derived closes it.

- Resource tests assert status only. Given the point of the change is a message that names the failing rule, asserting the message text in the 400 body would pin it. The constants are package-private in another module, so it'd have to be the literal string.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
ankita-gupta21: fixed

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NewPreRegisterUserEntityValidator.java
ankita-gupta21: fixed

aqibmohammadkhan (commented):
resolvedSourceId.trim() can NPE
`UserServiceImpl.java:931`

```
String resolvedSourceId = isBlank(...getSourceId())
    ? (isServiceUser && isNotBlank(...getLastname()) ? ...getLastname() : ...getEmail())
    : ...getSourceId();
newExternalUserEntity.setSourceId(resolvedSourceId.trim());
```
resolvedSourceId is null when sourceId is blank, lastname is blank, and email is null.

I traced both call sites and it is not reachable today: register() (:851) passes isServiceUser=false, so the email-format guard above throws first (EmailValidator.isValid(null) is false), and the service path is reachable only through UsersResource#createUser, where the validator rejects the all-blank case.

It's worth fixing anyway because of how it fails. The old line tolerated null and persisted a null sourceId; the new one throws NPE → TechnicalManagementException → 500 on user creation, which is exactly APIM-14832. So the method now hard-depends on a bean-validation invariant enforced two modules away, and if that coupling ever drifts — a second caller, service-account creation exposed on the v2 API — the bug this PR fixes comes back in its original form. StringUtils.trim() is null-safe and already imported in this file:

`newExternalUserEntity.setSourceId(StringUtils.trim(resolvedSourceId));`

Aikido has flagged the nested ternary here twice ([14:45 Aug 11](https://github.com/gravitee-io/gravitee-api-management/pull/19067#discussion_r0) and again on this head) and it's still open. Its suggested if/else form and this null-safety point are the same edit — the extracted branches make the getEmail() fallback the obvious place null enters.

Smaller

- firstname: "" is now stored, not rejected. The isNotBlank change (which I asked for) means {"firstname": "", "service": true, "email": "..."} goes from 400 to 200, and the account persists with firstname = "" — a non-null firstname on a service account, which is what the rule exists to prevent. TicketServiceImpl.java:182 branches on getFirstname() == null to build a sender name, so such an account would render as " " + lastname instead of its email. Cosmetic and unlikely for a service account, but normalizing blank → null before create would close it properly rather than just relaxing the gate.
- The trim now applies to caller-supplied sourceId too. setSourceId(...) is called unconditionally, so register() also rewrites an externally-supplied IdP subject identifier. Whitespace in an IdP subject is far-fetched and the consistency with exact-match findBySource is arguably right — just noting it's broader than the derived-service-account case and silently rewrites caller data on a path this ticket isn't about.

ankita-gupta21:
> resolvedSourceId.trim() can NPE `UserServiceImpl.java:931`
> 
> ```
> String resolvedSourceId = isBlank(...getSourceId())
>     ? (isServiceUser && isNotBlank(...getLastname()) ? ...getLastname() : ...getEmail())
>     : ...getSourceId();
> newExternalUserEntity.setSourceId(resolvedSourceId.trim());
> ```
> 
> resolvedSourceId is null when sourceId is blank, lastname is blank, and email is null.
> 
> I traced both call sites and it is not reachable today: register() (:851) passes isServiceUser=false, so the email-format guard above throws first (EmailValidator.isValid(null) is false), and the service path is reachable only through UsersResource#createUser, where the validator rejects the all-blank case.
> 
> It's worth fixing anyway because of how it fails. The old line tolerated null and persisted a null sourceId; the new one throws NPE → TechnicalManagementException → 500 on user creation, which is exactly APIM-14832. So the method now hard-depends on a bean-validation invariant enforced two modules away, and if that coupling ever drifts — a second caller, service-account creation exposed on the v2 API — the bug this PR fixes comes back in its original form. StringUtils.trim() is null-safe and already imported in this file:
> 
> `newExternalUserEntity.setSourceId(StringUtils.trim(resolvedSourceId));`
> 
> Aikido has flagged the nested ternary here twice ([14:45 Aug 11](https://github.com/gravitee-io/gravitee-api-management/pull/19067#discussion_r0) and again on this head) and it's still open. Its suggested if/else form and this null-safety point are the same edit — the extracted branches make the getEmail() fallback the obvious place null enters.
> 
> Smaller
> 
> * firstname: "" is now stored, not rejected. The isNotBlank change (which I asked for) means {"firstname": "", "service": true, "email": "..."} goes from 400 to 200, and the account persists with firstname = "" — a non-null firstname on a service account, which is what the rule exists to prevent. TicketServiceImpl.java:182 branches on getFirstname() == null to build a sender name, so such an account would render as " " + lastname instead of its email. Cosmetic and unlikely for a service account, but normalizing blank → null before create would close it properly rather than just relaxing the gate.
> * The trim now applies to caller-supplied sourceId too. setSourceId(...) is called unconditionally, so register() also rewrites an externally-supplied IdP subject identifier. Whitespace in an IdP subject is far-fetched and the consistency with exact-match findBySource is arguably right — just noting it's broader than the derived-service-account case and silently rewrites caller data on a path this ticket isn't about.

Fixed all three:

1. NPE: extracted the nested ternary into if/else and switched to the null-safe `StringUtils.trim(resolvedSourceId)`: [UserServiceImpl.java#L927-L937](https://github.com/gravitee-io/gravitee-api-management/blob/APIM-14832-APIM-Create-service-account-without-lastname-via-mAPI-fails-with-500-instead-of-400/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java#L927-L937)
   Added `shouldNotThrowNPEWhenServiceAccountHasNoDerivableSourceId` as defense-in-depth: [UserServiceTest.java#L3032 ](https://github.com/gravitee-io/gravitee-api-management/blob/APIM-14832-APIM-Create-service-account-without-lastname-via-mAPI-fails-with-500-instead-of-400/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java#L3032)
   

2. `firstname: ""` now normalized to `null` before persisting service accounts: [UserServiceImpl.java#L923-L926](https://github.com/gravitee-io/gravitee-api-management/blob/APIM-14832-APIM-Create-service-account-without-lastname-via-mAPI-fails-with-500-instead-of-400/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java#L923-L926)
   Added `shouldNormalizeBlankFirstnameToNullForServiceAccount`: [UserServiceTest.java#L3073](https://github.com/gravitee-io/gravitee-api-management/blob/APIM-14832-APIM-Create-service-account-without-lastname-via-mAPI-fails-with-500-instead-of-400/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java#L3073)
   
3. Trim on caller-supplied sourceId: left it as broad as your suggested fix — narrowing it to only the derived
   case would let an explicit sourceId with stray whitespace (e.g. " sa " vs "sa") collide again under
   findBySource's exact match, which is the same class of bug this PR closes.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
```diff
@@ -2920,6 +2946,172 @@ public void shouldSetIsServiceAccountToTrueWhenCreatingServiceUser() throws Tech
...
+            membershipService.getRoles(MembershipReferenceType.ORGANIZATION, ORGANIZATION, MembershipMemberType.USER, user.getId())
+        ).thenReturn(new HashSet<>(List.of(roleOrg)));
+
+        userService.create(EXECUTION_CONTEXT, newPreRegisterUserEntity);
+
+        verify(userRepository).create(argThat(userToCreate -> "sa".equals(userToCreate.getSourceId())));
+    }
+
+    @Test
+    public void shouldNotThrowNPEWhenServiceAccountHasNoDerivableSourceId() throws TechnicalException {
```
aqibmohammadkhan: Service layer still persists sourceId = null if the validator is skipped

`UserServiceImpl.createAndSendEmail` trims a possibly-null resolvedSourceId and continues. `shouldNotThrowNPEWhenServiceAccountHasNoDerivableSourceId` asserts that `userRepository.create` is called with sourceId == null.

That is the original APIM-14832 row. `MongoUserRepository.findBySource` still returns empty for null, so a second such create is still a unique-index 500. The NPE is gone, the corrupt write is not.

The API path is safe (`NewPreRegisterUserEntityValidator` rejects all-blank). The test is the problem: it locks in the failure mode this PR exists to remove, for any future caller that bypasses the validator (v2 create, internal UserService.create).

Fix: after trim, if isBlank(resolvedSourceId), throw InvalidUserException (or the same identifier message). Flip the test to expect that exception, not a null persist.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/NewExternalUserEntity.java
aqibmohammadkhan, line 1: @Schema on NewExternalUserEntity still describes service-account derivation

lastname / email / sourceId now say they may be used to derive sourceId. That type is also the body of POST /users/registration (UsersRegistrationResource.java), where service does not exist. Email is also labeled Required, which is true for registration (@Valid + @NotNull) and false for a lastname-only or sourceId-only service account.

Keep parent descriptions generic; leave the full rule on NewPreRegisterUserEntity.service and the 400 @ApiResponse (as in the earlier comment).

aqibmohammadkhan (commented):
PR title is stale

Title: “require lastname or sourceId”. Behavior: email-only succeeds. That will confuse cherry-picks to 4.9–4.12. Something like “derive service-account sourceId from lastname or email” would match the code.

Validator javadoc is stale

It still says “firstName is null”; the code uses isNotBlank.

No service-level test for ticket step 3

Email-only create is covered. Duplicate email → InvalidUserException (400 “user already exists”) is not. Pre-create findBySource should catch it now that sourceId is the email; worth one when(findBySource).thenReturn(of(existing)) assertion.

ankita-gupta21:
> * firstname != null vs blank. {"firstname": "", "service": true} gets rejected with "must not have a firstname". Pre-existing, but if you're aligning on blank semantics in this file, worth doing here too.
> * @Schema scope. The service-account wording sits on NewExternalUserEntity, which is also the body of POST /users/registration (UsersRegistrationResource.java:73), where service doesn't exist — those descriptions will show up there too. Keeping the base descriptions generic and leaving the full rule on NewPreRegisterUserEntity.service + the @apiresponse avoids it. email is also part of the derivation chain now but has no @Schema.
> * No trimming on derivation. {"service": true, "lastname": " sa "} persists sourceId as " sa ", and findBySource is exact-match, so " sa " and "sa" coexist as distinct accounts. A .trim() where sourceId is derived closes it.
> * Resource tests assert status only. Given the point of the change is a message that names the failing rule, asserting the message text in the 400 body would pin it. The constants are package-private in another module, so it'd have to be the literal string.

Fixed all four:

1. `firstname != null` → `isNotBlank`, consistent with the rest of the file: [NewPreRegisterUserEntityValidator.java#L48](https://github.com/gravitee-io/gravitee-api-management/blob/1bb6ccba73/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NewPreRegisterUserEntityValidator.java#L48)
   
2. `@Schema` scoping: base-class descriptions on `NewExternalUserEntity` are now generic (no service-account
   language), `email` now has a description too, and the full service-account rule lives only on
   `NewPreRegisterUserEntity.service` + the 400 `@ApiResponse` on `createUser`:
   [NewExternalUserEntity.java#L32-L61](https://github.com/gravitee-io/gravitee-api-management/blob/1bb6ccba73/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/NewExternalUserEntity.java#L32-L61)
   [NewPreRegisterUserEntity.java#L26-L29](https://github.com/gravitee-io/gravitee-api-management/blob/1bb6ccba73/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/NewPreRegisterUserEntity.java#L26-L29)
   [UsersResource.java#L94-L98](https://github.com/gravitee-io/gravitee-api-management/blob/1bb6ccba73/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/organization/UsersResource.java#L94-L98)
   

3. `.trim()` on the resolved sourceId (now null-safe via `StringUtils.trim`, see the separate NPE thread): [UserServiceImpl.java#L936-L946](https://github.com/gravitee-io/gravitee-api-management/blob/1bb6ccba73/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java#L936-L946)
   
4. Resource tests now assert the actual 400 message text, not just the status code: [UsersResourceTest.java#L72-L77](https://github.com/gravitee-io/gravitee-api-management/blob/1bb6ccba73/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/UsersResourceTest.java#L72-L77) , [UsersResourceTest.java#L92](https://github.com/gravitee-io/gravitee-api-management/blob/1bb6ccba73/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/UsersResourceTest.java#L92), [UsersResourceTest.java#L136](https://github.com/gravitee-io/gravitee-api-management/blob/1bb6ccba73/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/UsersResourceTest.java#L136)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
ankita-gupta21: done

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/validator/NewPreRegisterUserEntityValidator.java
```diff
@@ -18,29 +18,57 @@
...
+                StringUtils.isBlank(value.getLastname()) &&
+                StringUtils.isBlank(value.getEmail())
+            ) {
+                addViolation(context, SERVICE_ACCOUNT_IDENTIFIER_MESSAGE);
+                return false;
+            }
+            return true;
         } else {
-            return value.getEmail() != null;
+            if (StringUtils.isBlank(value.getEmail())) {
```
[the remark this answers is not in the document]
ankita-gupta21: ignoring as it was intentionally done as per the review https://github.com/gravitee-io/gravitee-api-management/pull/19067#discussion_r3752490102

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/NewExternalUserEntity.java
ankita-gupta21: Fixed — stripped all service-account language from the shared NewExternalUserEntity descriptions
(lastname, email, sourceId are back to plain, generic field docs; email no longer says "Required"
either, since that's not true for a lastname-only or sourceId-only service account):

https://github.com/gravitee-io/gravitee-api-management/blob/b43c8b4a0e/gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/NewExternalUserEntity.java#L32-L62

The full rule stays only on NewPreRegisterUserEntity.service and the 400 @ApiResponse on createUser,
both scoped to service-account creation, not the registration endpoint.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -896,13 +897,10 @@ public UserEntity create(ExecutionContext executionContext, final NewPreRegister
...
-        return (
-            IDP_SOURCE_GRAVITEE.equals(user.getSource()) &&
-            user.getIsServiceAccount() != null &&
-            user.getIsServiceAccount() &&
-            (user.getSourceId().equalsIgnoreCase(user.getEmail()) || user.getSourceId().equalsIgnoreCase(user.getLastname()))
-        );
+        // isServiceAccount is authoritative; a sourceId comparison was a heuristic for records predating the
+        // flag and incorrectly excludes a service account created with an explicit sourceId that differs
+        // from both email and lastname (e.g. sourceId "sa-billing" with email "ops@acme.com").
+        return IDP_SOURCE_GRAVITEE.equals(user.getSource()) && Boolean.TRUE.equals(user.getIsServiceAccount());
```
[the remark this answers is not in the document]
ankita-gupta21: ignore: stylistic preference, no behavior change

aqibmohammadkhan (approved):

PR #19070, opened 2026-08-10
feat(portal): filter catalog by category with shareable deep link

dev-marek:
**Issue**

https://gravitee.atlassian.net/browse/PORTAL-23

**Purpose**

Let a portal consumer filter the API catalog by category and share/bookmark the filtered view as a deep link.

**Summary of changes**

- `PortalNavigationApi.categoryIds?: string[]` added to the nav-item entity; threaded through the search response mapping into the catalog's row view model.
- New `PortalCategory` entity + `PortalCategoriesService` (`GET /portal-categories`) to list visible categories.
- `PortalNavigationItemsService.searchNavigationItemsWithApis()` takes an optional `categoryId` and forwards it to `_search`.
- New `CategorySelectComponent`: a compact "Category: All ▾" dropdown (styled to match the Figma spec), placed left of the grid/list toggle in the catalog header.
- Catalog page: hydrates the selected category from the `category` query param on load; selecting/clearing syncs the URL via `Router.navigate` with `queryParamsHandling: 'merge'` and resets the search query + pagination.
- Fixed the search box's navigation to merge query params instead of replacing them, so an active category filter no longer gets wiped out by typing a search term.
- Added a generic inline error state (distinct from the existing "no results" empty state) for both backend search failures and an unknown/hidden `category` id in the URL.
- List view gained a **Category** column (chip badges), matching the Figma mockup.

**Out of scope**

- The `/portal-categories` and `categoryId` search-filter backend support (already shipped in the PORTAL-22 PR this stacks on).
- Category management/admin UI (pre-existing, unrelated).

**Testing**

- Full `portal-webui-next` unit test suite (1171 tests) passes, including new/updated specs for `CatalogComponent`, `CategorySelectComponent`, `PortalCategoriesService`, and `PortalNavigationItemsService`.
- `lint-eslint`, `lint-prettier`, `lint-license`, and a production `build` all pass.
- Not yet manually verified in a running browser — follow-up before merge.

**Additional context**

Stacked on #19062 (PORTAL-22) — base branch is `portal-22-nav-category-filter`, not `master`. Figma reference: the "Categories - Next Gen Portal" file (node 1:4800 / catalog header).

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.spec.ts
+++ b/gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.spec.ts
@@ -17,12 +17,17 @@ import { HarnessLoader } from '@angular/cdk/testing';
 import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
 import { HttpTestingController } from '@angular/common/http/testing';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { ActivatedRoute, Router } from '@angular/router';
+import { of } from 'rxjs';
 
 import { CatalogComponent } from './catalog.component';
 import { ApiCardHarness } from '../../components/api-card/api-card.harness';
+import { CategorySelectHarness } from '../../components/category-select/category-select.component.harness';
 import { PaginationHarness } from '../../components/pagination/pagination.harness';
 import { fakeApi, fakeApisResponse } from '../../entities/api/api.fixtures';
 import { ApisResponse } from '../../entities/api/apis-response';
+import { PortalCategory } from '../../entities/categories/portal-category';
+import { fakePortalCategory } from '../../entities/categories/portal-category.fixture';
 import { PortalNavigationItemsSearchResponse } from '../../entities/portal-navigation/portal-navigation-apis-search';
 import { AppTestingModule, TESTING_BASE_URL } from '../../testing/app-testing.module';
 
@@ -40,6 +45,8 @@ describe('CatalogComponent', () => {
 
     httpTestingController = TestBed.inject(HttpTestingController);
     harnessLoader = TestbedHarnessEnvironment.loader(fixture);
+
+    flushCategories();
 
     fixture.detectChanges();
   };
@@ -60,6 +67,20 @@ describe('CatalogComponent', () => {
     await initBase();
     expectApiList(params.apisResponse, params.page ?? 1, params.size ?? 20, params.hasNextPage ?? false);
     fixture.detectChanges();
+  };
+
+  const initWithQueryParams = async (queryParams: Record<string, string>) => {
+    await TestBed.configureTestingModule({
+      imports: [CatalogComponent, AppTestingModule],
+    })
+      .overrideProvider(ActivatedRoute, {
+        useValue: { snapshot: { queryParams }, queryParams: of(queryParams) },
+      })
+      .compileComponents();
+
+    fixture = TestBed.createComponent(CatalogComponent);
+    httpTestingController = TestBed.inject(HttpTestingController);
+    harnessLoader = TestbedHarnessEnvironment.loader(fixture);
   };
 
   afterEach(() => {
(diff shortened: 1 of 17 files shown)


gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.ts
```diff
@@ -82,23 +88,44 @@ export class CatalogComponent {
...
   private readonly page$ = new BehaviorSubject<number>(1);
   protected readonly query = toSignal(this.route.queryParams.pipe(map(p => p['query'] ?? '')), { initialValue: '' });
-  protected readonly tableColumns = computed(() => (this.isMobile() ? ['name', 'version', 'mcp'] : ['name', 'labels', 'version', 'mcp']));
+  protected readonly categoryId = toSignal(this.route.queryParams.pipe(map(p => p['category'] ?? null)), {
+    initialValue: null as string | null,
+  });
+  private readonly categoriesState = toSignal(
+    this.portalCategoriesService.getCategories().pipe(
+      map(categories => ({ categories, loaded: true })),
+      catchError(_ => of({ categories: [] as PortalCategory[], loaded: true })),
```
thomasballivet: shouldn't `loaded` be set to `false` on error?

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.ts
```diff
@@ -135,14 +173,24 @@ export class CatalogComponent {
...
+        unknownCategory: this.unknownCategory(),
+      })),
       distinctUntilChanged((previous, current) => isEqual(previous, current)),
-      tap(_ => this.loadingPage.set(true)),
-      switchMap(({ page, pageSize, query }) =>
-        this.portalNavigationItemsService
-          .searchNavigationItemsWithApis(page, query, pageSize)
-          .pipe(catchError(_ => of({ data: [], metadata: undefined }))),
-      ),
+      switchMap(({ page, pageSize, query, categoryId, unknownCategory }) => {
```
thomasballivet, line 192: The first search fires before the categories load, so `categoryId` goes to the backend unvalidated. Could we skip emitting while they're in flight? `filter(({ categoryId, status }) => !categoryId || status !== 'loading')`

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.spec.ts
```diff
@@ -170,7 +191,75 @@ describe('CatalogComponent', () => {
 
         const emptyState = fixture.nativeElement.querySelector('.api-list__empty-state');
         expect(emptyState).toBeTruthy();
-        expect(emptyState.textContent).toContain('No APIs available yet');
+        expect(emptyState.textContent).toContain('Something went wrong');
+      });
+    });
+  });
+
+  describe('category filtering', () => {
```
thomasballivet, line 199: Missing case: /portal-categories flushed with a 500 while `?category=` is set → the search should still run.

thomasballivet (changes requested):

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.ts
dev-marek: Good catch — replaced the boolean `loaded` with a `status: 'loading' | 'loaded' | 'error'` union so a failed fetch is no longer indistinguishable from a successful one. Fixed in c3dd079aa2.

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.ts
dev-marek: Added exactly this: `filter(({ categoryId, status }) => !categoryId || status !== 'loading')` right after the initial map, so the search is held back until the category list resolves. Fixed in c3dd079aa2.

gravitee-apim-portal-webui-next/src/app/catalog/catalog.component.spec.ts
dev-marek: Added a test for a /portal-categories 500 with ?category= set — confirms the search still runs with categoryId passed through, since status is now 'error' rather than 'loaded' so it's never treated as an unknown category. Fixed in c3dd079aa2.

thomasballivet (approved):

PR #18973, opened 2026-08-05
feat(gamma): added gateway list ui + code skeleton

arpzilla-dev:
## Issue

https://gravitee.atlassian.net/browse/FOUND-33

## Description

Add the Gamma Platform **Gateways** list (env-scoped) with Graphene `DataTable`, status icons, server-side pagination, and 30s refresh. Includes shared instance types/services and a stub detail route for stacked follow-ups (FOUND-34 / FOUND-35).
- Nav + route under Gateways & Infrastructure
- Permission: `environment-instance-r`
- Name link → detail stub (`:instanceId/environment|monitoring`)

## Stack
PR 1/3 — list + skeleton. Detail (FOUND-34) and monitoring (FOUND-35) will stack on top.

## Additional context

[working video] 

https://github.com/user-attachments/assets/4dca6ed0-af6e-4594-a7d1-d5c536ad3970

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #19079, opened 2026-08-11
feat(gamma): added gateway monitoring details page

arpzilla-dev:
## Issue

https://gravitee.atlassian.net/browse/FOUND-35

## Description

Adds the Gateway Instance **Monitoring** tab in Gamma, showing live health metrics (CPU, memory, heap, and uptime) with near-real-time refresh, and wiring it into the instance detail layout.

## Additional context

https://github.com/user-attachments/assets/40b0bd7e-5c0c-4bd5-bb61-7a07edba84c7


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GatewayInstanceMonitoringPage.tsx
```diff
@@ -0,0 +1,92 @@
...
+                <Skeleton className="h-48 w-full" />
+                <Skeleton className="h-48 w-full" />
+                <Skeleton className="h-48 w-full" />
+                <Skeleton className="h-48 w-full" />
+            </div>
+        </div>
+    );
+}
+
+/** Classic Monitoring tab with 5s polling while the instance is STARTED. */
```
arpzilla-dev: its a classic behaviour, to directly used here

.tmp-apim14718/GatewayInstanceMonitoringView.tsx
aqibmohammadkhan, line 1: why temp file committed here?

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/gateway-instances/components/GatewayInstanceDetailLayout.tsx
aqibmohammadkhan, line 1: Unrelated scope in the same commit

 - GatewayInstancesPage env change → reset page
 - GatewayInstanceDetailLayout back nav → resolveListHrefFromDetailBasePath
Reasonable fixes, but they aren’t FOUND-35. 

Prefer a separate PR/commit so monitoring review stays focused.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/gateway-instances/hooks/useGatewayInstanceMonitoring.ts
```diff
@@ -0,0 +1,41 @@
...
+    gatewayId: string | undefined;
+    enabled: boolean;
+}) {
+    const env = useEnvironment();
+
+    return useQuery({
+        queryKey: gatewayInstanceKeys.monitoring(env?.id ?? '', instanceId ?? '', gatewayId ?? ''),
+        queryFn: () => getGatewayInstanceMonitoring(env!.id, instanceId!, gatewayId!),
+        enabled: Boolean(env?.id && instanceId && gatewayId && enabled),
+        refetchInterval: 5_000,
```
aqibmohammadkhan: refetchInterval always on while mounted (useGatewayInstanceMonitoring.ts)
Matches classic polling intent, but there’s no assertion it stays at 5000, and no refetchInterval: false / conditional when enabled is false (RQ usually ignores when disabled, but making interval conditional is clearer and safer).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/gateway-instances/components/GatewayInstanceMonitoringIndicators.tsx
aqibmohammadkhan, line 1: Hard-coded colors vs design tokens (GatewayInstanceMonitoringView.tsx ProgressBar #f5d0c5, HeroValueCard #1f77b4; gauge #e4e4e7)
Inline styles break Graphene/Tailwind consistency used elsewhere. Prefer tokens/bg-muted/text-primary.

aqibmohammadkhan (commented):
- Low-value JSDoc on GatewayInstanceMonitoringPage (Aikido’s point) — restates visible behavior; either explain why (classic parity / avoid hammering stopped gateways) or drop it.
- Gauge fidelity — classic gio-circular-percentage uses reverseColor; Gamma gauge is flat primary color only.
- Uptime phrasing — classic moment.duration(-ms).humanize(true) (“a few seconds ago”) vs Gamma positive (“a few seconds”); intentional, but worth a product/docs note.
- Stopped-state test doesn’t assert enabled: false on the monitoring hook.
- Duplicate DescriptionRow / Row across view and indicators — fine for now, small DRY opportunity later.

.tmp-apim14718/GatewayInstanceMonitoringView.tsx
arpzilla-dev: Good catch ... that was local scratch and shouldn’t have been committed. Removed in the latest push.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/gateway-instances/components/GatewayInstanceDetailLayout.tsx
arpzilla-dev: it was mentioned here https://github.com/gravitee-io/gravitee-api-management/pull/18973#discussion_r3747224482 in its first PR https://github.com/gravitee-io/gravitee-api-management/pull/18973

so now added as separate commit.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/gateway-instances/hooks/useGatewayInstanceMonitoring.ts
arpzilla-dev: Updated 
to refetchInterval: enabled ? 5_000 : false, and added a test asserting the 5s interval when enabled. React Query already skips polling when disabled; this just makes the intent explicit.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/gateway-instances/components/GatewayInstanceMonitoringIndicators.tsx
arpzilla-dev: We tried Graphene/Tailwind tokens (Progress, text-blue-600, muted track classes) after your note, but in the Gamma MF host those utilities didn’t render reliably (pool bars collapsed / hero blue dropped / gauge track looked wrong). Restored the working approach:

gauge track + rotation via stable inline styles
pool bars via inline height/width + var(--primary)
GC / File Descriptors keep Classic #1f77b4

_it was coming like this, not the intended UX_ :
image

arpzilla-dev:
* Low-value JSDoc on GatewayInstanceMonitoringPage (Aikido’s point) — restates visible behavior; either explain why (classic parity / avoid hammering stopped gateways) or drop it.

> Fair point .. the comment only restated what the code already shows. Removed it.


* Gauge fidelity — classic gio-circular-percentage uses reverseColor; Gamma gauge is flat primary color only.

> Intentional for Gamma: we matched the prototype / product primary-orange gauge rather than Classic’s green→amber→red reverseColor thresholds. Happy to revisit if we want stricter Classic color parity.


* Uptime phrasing — classic moment.duration(-ms).humanize(true) (“a few seconds ago”) vs Gamma positive (“a few seconds”); intentional, but worth a product/docs note.

> Intentional .... Gamma uses a positive uptime phrase instead of moment’s humanize(true) (“ago”). Can align to Classic wording if product prefers

* Stopped-state test doesn’t assert enabled: false on the monitoring hook.

> Adde... the stopped-instance test now asserts useGatewayInstanceMonitoring is called with enabled: false.

* Duplicate DescriptionRow / Row across view and indicators — fine for now, small DRY opportunity later.

> Agreed ...leaving as-is for this PR; can DRY later if we touch these again.

arpzilla-dev:
working UI, after latest changes

image

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GatewayInstanceMonitoringPage.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    if (isMonitoringError && !monitoringData) {
+        return (
+            <p className="text-sm text-muted-foreground" data-testid="gateway-instance-monitoring-error">
+                Failed to load monitoring data. Please refresh and try again.
+            </p>
+        );
+    }
+
+    if (!monitoringData) {
```
aqibmohammadkhan, line 86: Backend `MonitoringServiceImpl.findMonitoring` can return null when the repo has no metrics. After a successful query with no payload, the page hits if (!monitoringData) return <MonitoringSkeleton /> while 5s polling continues — endless loading instead of a no-data message. Specs cover loading/error/stopped, not this path.

Fix: treat settled + falsy data like the stopped empty state (or a dedicated “no monitoring data yet” message), not as loading.

arpzilla-dev: thanks, Added that

aqibmohammadkhan (approved):

PR #19081, opened 2026-08-11
fix(gateway): measure endpoint response time up to the last byte of the response body

phiz71:
## Issue

https://gravitee.atlassian.net/browse/APIM-14903

## Description

`gateway-latency-ms` is meant to be the gateway's own overhead. It is derived by subtracting
`endpoint-response-time-ms` from `gateway-response-time-ms`, but the two stopped at different points:
the endpoint one at the upstream response headers, the total at the last byte sent to the client.
**The whole response body transfer was therefore counted as gateway latency** — on a streamed API, a
gateway that did nothing for 100 seconds reported a latency of about 100 000 ms.

Ten commits, each standing on its own — the first five carry the change, the last five answer review:

1. **`feat(reporter)`** — opens the reporting surface: Elasticsearch templates and mappings for every
   flavour, plus the reflective JSON/MsgPack formatters. Also moves `gravitee-reporter-api` from
   2.5.0 to 2.8.0, purely additive over 2.5.0 and the line `master` already builds against.
2. **`fix(gateway)`** — the fix itself. `endpoint-response-time-ms` now runs until the response body
   is fully received, and what it used to measure is kept as `endpoint-response-ttfb-ms`
   (nginx's `$upstream_header_time`).
3. **`feat(gateway)`** — adds `endpoint-connect-time-ms` (`$upstream_connect_time`), so a wait on a
   saturated connection pool is no longer indistinguishable from a slow backend.
4. **`feat(gateway)`** — measures every duration with the monotonic `System.nanoTime()` and reports
   `*-ns` fields alongside the `*-ms` ones.
5. **`test(gateway)`** — unrelated flaky test fixed along the way, see below.
6. **`fix(gateway)`** — keeps the gateway latency a difference for reactors that only report
   milliseconds. The monotonic branch runs for every request, v2 emulation included, but only the v4
   reactor and the HTTP proxy connector write the nanosecond endpoint field: v2 emulation was having
   its whole response time charged to the gateway latency. Regression from commit 4, caught in review.
7. **`fix(gateway)`** — takes the connect time origin after the HTTP client is built, so building it
   on first use is no longer counted as connection time.
8. **`fix(gateway)`** — gives every request implementation a monotonic origin, so a new one cannot
   silently fall back to the wall clock, with tests pinning both ends of that wiring.
9. **`refactor(reporter)`** — one helper for the "omit a duration that was never measured" rule,
   previously spelled out 22 times.
10. **`test(gateway)`** — display-name alignment on the touched test classes.

The reactor sets the response time to the time to first byte and the HTTP proxy connector overwrites
it once the body is drained, in a `doFinally` covering completion, error and cancellation alike. **A
connector that does not stream a body keeps reporting exactly what it reports today**, so the others
can be migrated one at a time.

## Additional context

**Model side** is https://github.com/gravitee-io/gravitee-reporter-api/pull/141, merged and released
as 2.8.0, which this branch now depends on — nothing left pending there.

**The CSV columns were dropped** from this branch: adding fields to a format whose columns are
already neither stable nor identifiable past the fixed ones is not worth doing on a support branch.
They are tracked separately, behind a redesign of that format.

**Validated end to end against Elasticsearch** on a local stack, not just by unit tests. The same
request — a backend answering immediately, then streaming for 10 seconds — sent a minute apart to a
gateway running 4.12.x and to one running this branch:

| | before | after |
|---|---|---|
| `endpoint-connect-time-ms` | *(no such field)* | 0 |
| `endpoint-response-ttfb-ms` | *(no such field)* | 1 |
| `endpoint-response-time-ms` | 1 | **10 008** |
| `gateway-latency-ms` | **10 010** | 1 |
| `gateway-response-time-ms` | 10 011 | 10 009 |

The total is identical on both sides, as it must be — only its attribution changes: the ten seconds
of streaming move out of the gateway latency and into the endpoint response time, where they were
actually spent. The arithmetic closes: 10 009 − 10 008 = 1.

Scaled to a 100 s stream, the same run gives ttfb 1 ms, response time 100 011 ms, latency 1 ms. A
plain non-streamed request stays within a millisecond of its previous values. With a connection
pool capped at 5 and 20 concurrent requests, the queueing shows up in `endpoint-connect-time`
(2 991 / 5 990 / 8 980 ms across the waves) while the gateway latency stays at 0 — the three endpoint
durations are nested, as nginx's are, so waiting for a connection is upstream time and is subtracted
from the latency rather than charged to it.

**Behaviour change to be aware of**: the value of `endpoint-response-time-ms` rises on APIs that
stream or return large payloads, and `gateway-latency-ms` drops accordingly. Alerts and dashboards
with thresholds on either field will see different numbers — `apiResponseTimeMs` on v2 metrics and
the `upstream-response-time` property sent to Alert Engine derive from the same value.

**Milliseconds are now derived from the nanoseconds, rounded to the nearest** rather than truncated:
a difference of two `currentTimeMillis()` readings counts the millisecond boundaries crossed, so it
averages out to the real duration, whereas a floor would bias every value down by half a millisecond
— enough to report a sub-millisecond gateway latency as a flat zero.

**About the flaky test**: `HttpConnectorTest` started a WireMock server per test and never stopped
any of them, leaving 31 servers holding their ports and thread pools for the life of the JVM. Under
load the suite failed at connection time on no particular test — two runs failed on two different
ones, and one took 23 minutes. Fixed separately in commit 5.


gravitee-apim-reporter/gravitee-apim-reporter-common/src/main/java/io/gravitee/apim/reporter/common/formatter/csv/v4/MetricsFormatter.java
```diff
@@ -67,7 +67,16 @@ protected Buffer format0(Metrics metrics) {
...
+        appendString(buffer, metrics.getApiProductId());
+        // Appended after every fixed column and before the variable ones (additional then custom metrics), so no
+        // existing position moves. -1 marks a duration that was never measured, the endpoint not having been invoked.
+        appendLong(buffer, metrics.getEndpointResponseTtfbMs());
+        appendLong(buffer, metrics.getEndpointConnectTimeMs());
+        appendLong(buffer, metrics.getEndpointResponseTimeNs());
+        appendLong(buffer, metrics.getEndpointResponseTtfbNs());
+        appendLong(buffer, metrics.getEndpointConnectTimeNs());
+        appendLong(buffer, metrics.getGatewayResponseTimeNs());
+        appendLong(buffer, metrics.getGatewayLatencyNs(), customMetrics.isEmpty() && metrics.getAdditionalMetrics().isEmpty());
```
jhaeyaert: This is a BC. It should be removed from the support version and moved to master instead, with a BC notes. WDYT?

jhaeyaert (changes requested):

gravitee-apim-reporter/gravitee-apim-reporter-common/src/main/java/io/gravitee/apim/reporter/common/formatter/csv/v4/MetricsFormatter.java
phiz71: Good catch, you're right — my reasoning was too narrow.

I had checked that no *fixed* column moves (the new ones sat after `api-product-id` and before the
additional/custom metrics) and concluded it was safe. But that isn't the right question: it's not
whether a positional reader of the fixed columns survives, it's whether we change an output contract
on a support branch at all. We shouldn't.

I also brushed off the additional/custom metrics too quickly, on the grounds that their count varies
per request. It varies *across APIs*, but for a given API with a given set of policies it is stable,
so a consumer may well have pinned "column N = my custom metric". Shifting those by seven breaks it.

Removed from this branch — `git diff 4.12.x -- '*csv*'` is now empty. Kept as a patch for master,
where it will land with a BC note.

Worth noting the distinction holds for the rest of the PR: Elasticsearch, JSON and MsgPack are all
accessed by field name, so adding fields there is additive. CSV is the only positional format in the
chain, which is exactly why it is the only one pulled out.

jhaeyaert: To be honest, I don't get why we add the additional/custom metrics, considering they are unbounded and different per request in the CSV. 
This does not work at all, as it makes the CSV columns unpredictable and any CSV client will struggle to extract any information. Additionally, as there is no header line, this is even worse. Definitely a bad design that makes the whole CSV fragile.

This deserves a full redesign before we decide to introduce new fields.

phiz71: Ok when we work on master, we have to think about CSV reporter.
For this PR, we are ok

phiz71: Finally, the best for the CSV part is to wait.
I'll create a ticket to track this issue.
Meanwhile, this PR can be cherrypicked as-is on master. So I've added the "apply-on-master" labek.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/responsetime/ResponseTimeProcessor.java
```diff
@@ -36,12 +36,27 @@ public String getId() {
...
+                // Monotonic clock: measures the elapsed time regardless of any wall-clock adjustment, and at a
+                // resolution the gateway's own overhead — a couple of milliseconds — is actually visible at. Setting
+                // the nanoseconds derives the milliseconds.
+                final long gatewayResponseTimeNs = System.nanoTime() - requestStartNs;
+                final long endpointResponseTimeNs = metrics.getEndpointResponseTimeNs();
+                metrics.setGatewayResponseTimeNs(gatewayResponseTimeNs);
+                // No endpoint involved — a request rejected before reaching one — leaves the whole time to the gateway.
+                metrics.setGatewayLatencyNs(
+                    endpointResponseTimeNs > -1 ? gatewayResponseTimeNs - endpointResponseTimeNs : gatewayResponseTimeNs
+                );
```
ytvnr: The nanosecond branch is entered whenever `requestStartNs > 0`, which holds for every HTTP request: `MetricsProcessor` fills it from `VertxHttpServerRequest.timestampNs`, and `DefaultHttpRequestDispatcher` runs the same platform pre/post chain for every `ApiReactor`, v2 emulation included. Inside the branch the endpoint duration is read only from `getEndpointResponseTimeNs()`, and the only writers of that field are `DefaultApiReactor` and `HttpConnector`, both v4. `SyncApiReactor` still writes milliseconds only.

On a v2 emulation API the ns field therefore keeps its `-1` default, the ternary falls into the no endpoint case, and `gatewayLatencyMs` ends up equal to the full response time instead of the difference. In front of a 500 ms backend the reported latency goes from roughly 2 ms to roughly 502 ms, while `endpointResponseTimeMs` stays correct, so the reported document contradicts itself. `Metrics.toV2()` maps `proxyLatencyMs` from the same value, so v2 analytics follow.

Either fall back to the millisecond endpoint value when the ns one is unset, or route the ms only writers through the ns setter so a single source of truth remains. A test covering requestStartNs set plus only the ms endpoint value would pin it: the current `shouldAddResponseTimeToMetric` never sets `requestStartNs` and only exercises the fallback branch.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/test/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnectorTest.java
```diff
@@ -780,6 +782,72 @@ void shouldHandleServerError() throws InterruptedException {
...
+        assertNoTimeout(obs);
+        obs.assertComplete();
+
+        // Connecting only gets the response head: the body is still to be streamed, so nothing is measured yet.
+        verify(metrics, never()).setEndpointResponseTimeNs(anyLong());
+
+        consumeResponseChunks();
+
+        // The whole response is now in, body included — this is what nginx reports as $upstream_response_time.
+        verify(metrics).setEndpointResponseTimeNs(longThat(responseTimeNs -> responseTimeNs >= BACKEND_ELAPSED_NS));
```
ytvnr: The javadoc on `recordEndpointResponseTime` states that the `doFinally` covers completion, error and cancellation alike, but this test drains the body to completion, and the guard test only covers `getEndpointRequestStartNs() <= 0`. Nothing subscribes and then errors or disposes, so moving the call into `doOnComplete` would leave the whole suite green. The upstream error test above asserts nothing about the endpoint response time either.

Two cases would pin the contract: cancelling mid stream (subscribe, take one chunk, dispose) and an upstream failure, both asserting the setter is still called with a value at least equal to the injected elapsed time.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/metrics/MetricsProcessor.java
```diff
@@ -63,6 +63,9 @@ private void prepareMetrics(final HttpExecutionContextInternal ctx, final boolea
             if (request != null) {
                 metricsBuilder
                     .timestamp(request.timestamp())
+                    // Monotonic counterpart of the timestamp, taken when the request arrived: durations are derived
+                    // from it, the wall clock being reserved for indexing.
+                    .requestStartNs(request.timestampNs())
```
ytvnr, line 68: This is the single seam that turns the whole nanosecond path on: `ResponseTimeProcessor` only takes the monotonic branch when `requestStartNs > 0`, and silently falls back to the wall clock otherwise. `MetricsProcessorTest` is untouched here and asserts nothing about it, and nothing covers `VertxHttpServerRequest` setting `timestampNs`. If either end of the wiring is dropped, every gateway reports through the legacy path with all the new `*-ns` fields left at -1, and no test fails.

`MetricsProcessorTest` already has the structure for it, so a case asserting the built metrics carry the request `timestampNs()`, plus one on the request side, would be enough.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/test/java/io/gravitee/gateway/reactive/reactor/processor/responsetime/ResponseTimeProcessorTest.java
```diff
@@ -35,6 +36,9 @@
 @ExtendWith(MockitoExtension.class)
 class ResponseTimeProcessorTest extends AbstractProcessorTest {
```
ytvnr, line 41: The tests added here are snake_case, but this class has no `@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)`, unlike its siblings in the same package (`MetricsProcessorTest`, `NotFoundProcessorTest`, `ReporterProcessorTest`). Same for `DefaultApiReactorTest`. Reports end up showing raw underscored names next to the legacy camelCase `shouldAddResponseTimeToMetric`.

Adding the annotation on both classes would align them, and the remaining camelCase method could be renamed in the same pass.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
```diff
@@ -318,7 +321,35 @@ private Single<HttpClientResponse> sendEndpointRequestChunks(HttpClientRequest h
                 } catch (Exception e) {
                     ctx.withLogger(log).debug("Can't properly reset endpoint request to backend [{}]", absoluteUri, e);
                 }
-            });
+            })
+            // Last: covers completion, error and cancellation alike, so a truncated response is measured up to the
+            // point it stopped rather than left at the time to first byte.
+            .doFinally(() -> recordEndpointResponseTime(ctx));
```
ytvnr, line 327: This `doFinally` sits on the Flowable handed to `response.chunks(...)`, which is consumed by `nativeResponse.rxSend(chunks()...)` in `VertxHttpServerResponse.end()`. Demand there comes from the client socket, and `endpointResponse.toFlowable()` pauses the upstream read when there is no demand, so the terminal event fires when the last byte was written to the client rather than when it arrived from upstream. A slow client on a large or streamed response therefore inflates `endpoint-response-time-ms` and pushes `gateway-latency-ms` toward zero, which is the mirror image of the bug being fixed.

It also makes the semantics depend on the policy chain: a response phase policy that buffers the body drains this Flowable at upstream speed, and the measure goes back to being upstream read time.

nginx times `$upstream_response_time` on the read from upstream into its own buffers. Worth deciding explicitly whether this field means upstream read time, in which case the hook belongs on `endpointResponse.toFlowable()` before the policy composition, or full proxied transfer time, in which case the javadoc and the nginx analogy should say that downstream write time is included.

gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/HttpRequestInternal.java
```diff
@@ -85,4 +85,17 @@ default HttpRequestInternal debugContextPath(final String contextPath) {
...
+    /**
+     * Returns a monotonic reading taken when this request arrived, as the origin durations covering the whole request
+     * are measured from. It is the counterpart of {@link #timestamp()}, which stays a wall clock because it is what the
+     * request gets indexed on — and an adjustable clock cannot measure a duration.
+     * <p>
+     * The value is only meaningful relative to another {@link System#nanoTime()} reading.
+     *
+     * @return the {@link System#nanoTime()} reading taken when the request arrived, or {@code -1} when unknown.
+     */
+    default long timestampNs() {
```
ytvnr, line 102: `timestampNs()` is added as a `default` returning -1 while its neighbour `connectionTimestamp()` is abstract. The effect is invisible at compile time: `VertxTcpRequest` and `HttpHealthCheckRequest` both extend `AbstractRequest` and assign `this.timestamp` without assigning `timestampNs`, so they stay on the wall clock fallback, and any request implementation added later inherits the same silent downgrade.

Making the accessor abstract would force each implementation to answer. If the default has to stay for binary compatibility, setting the monotonic reading wherever the wall clock is set would keep the two from drifting apart.

gravitee-apim-reporter/gravitee-apim-reporter-common/src/main/java/io/gravitee/apim/reporter/common/formatter/elasticsearch/ElasticsearchFormatter.java
```diff
@@ -252,8 +252,15 @@ private Buffer getSource(io.gravitee.reporter.api.v4.metric.Metrics metrics, Map
...
+        data.put("endpointResponseTtfbMs", metrics.getEndpointResponseTtfbMs() >= 0 ? metrics.getEndpointResponseTtfbMs() : null);
+        data.put("endpointConnectTimeMs", metrics.getEndpointConnectTimeMs() >= 0 ? metrics.getEndpointConnectTimeMs() : null);
         data.put("gatewayResponseTimeMs", metrics.getGatewayResponseTimeMs() >= 0 ? metrics.getGatewayResponseTimeMs() : null);
         data.put("gatewayLatencyMs", metrics.getGatewayLatencyMs() >= 0 ? metrics.getGatewayLatencyMs() : null);
+        data.put("endpointResponseTimeNs", metrics.getEndpointResponseTimeNs() >= 0 ? metrics.getEndpointResponseTimeNs() : null);
+        data.put("endpointResponseTtfbNs", metrics.getEndpointResponseTtfbNs() >= 0 ? metrics.getEndpointResponseTtfbNs() : null);
+        data.put("endpointConnectTimeNs", metrics.getEndpointConnectTimeNs() >= 0 ? metrics.getEndpointConnectTimeNs() : null);
+        data.put("gatewayResponseTimeNs", metrics.getGatewayResponseTimeNs() >= 0 ? metrics.getGatewayResponseTimeNs() : null);
+        data.put("gatewayLatencyNs", metrics.getGatewayLatencyNs() >= 0 ? metrics.getGatewayLatencyNs() : null);
```
ytvnr: Seven more copies of `metrics.getX() >= 0 ? metrics.getX() : null`, with the getter called twice on each line, bringing the block to eleven near identical statements. The key string and the getter have to be kept in sync by hand on every line, a mismatch compiles and ships silently, and the templates consuming these keys live in three other files.

Extracting the put only when non negative rule into one small private method taking the key and the raw value would leave a single line per duration, and one place for the sentinel rule.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/test/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnectorTest.java
```diff
@@ -780,6 +782,72 @@ void shouldHandleServerError() throws InterruptedException {
...
+
+        wiremock.stubFor(get("/team").willReturn(ok(BACKEND_RESPONSE_BODY)));
+
+        final TestObserver<Void> obs = cut.connect(ctx).test();
+
+        assertNoTimeout(obs);
+        obs.assertComplete();
+
+        // Without it, a wait on a saturated pool is indistinguishable from a slow backend.
+        verify(metrics).setEndpointConnectTimeNs(longThat(connectTimeNs -> connectTimeNs >= 0));
```
ytvnr, line 807: `longThat(connectTimeNs -> connectTimeNs >= 0)` holds for a hardcoded 0 as well as for a real measure, so this only proves the setter was invoked. Nothing covers the nesting invariant the PR presents as the contract (connect time <= ttfb <= response time), nor the case where the connection cannot be acquired, since `recordEndpointConnectTime` sits inside the `asyncRequest.succeeded()` branch.

Pairing the connect time with the captured response time in one test and checking the relation, plus a case on an unreachable port asserting the setter is never called, would make it meaningful.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/test/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnectorTest.java
```diff
@@ -101,6 +104,8 @@ class HttpConnectorTest {
     protected static final int REQUEST_BODY_LENGTH = REQUEST_BODY.getBytes().length;
     protected static final String BACKEND_RESPONSE_BODY = "response from backend";
     public static final int TIMEOUT_SECONDS = 60;
+    /** Time already spent on the endpoint when the connector takes over, so a measured duration is provably ≥ it. */
+    private static final long BACKEND_ELAPSED_NS = TimeUnit.MILLISECONDS.toNanos(100);
     private static final String ERROR_ENDPOINT = "/error";
     private static WireMockServer wiremock;
```
ytvnr, line 120: Server creation moved from `@BeforeAll` to `@BeforeEach` and disposal to `@AfterEach`, but this field stayed `static`. A per test instance held in a static field is shared mutable state: it works only because this class runs single threaded, and it breaks as soon as the class runs in parallel or gains `@Nested` groups, with tests hitting the port of an already stopped server.

Making it non static matches the new lifecycle. `vertx` legitimately stays static since it is still bound to `@BeforeAll` and `@AfterAll`.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
```diff
@@ -210,13 +210,16 @@ private Single<HttpClientRequest> acquireUpstreamRequest(
...
+        return Single.create(emitter -> {
+            // Subscription time: this is where the wait on the connection pool starts (see recordEndpointConnectTime).
+            final long connectStartNs = System.nanoTime();
             httpClientFactory
                 .getOrBuildHttpClient(ctx, configuration, sharedConfiguration)
                 .getDelegate()
                 .request(options)
                 .onComplete(asyncRequest -> {
                     if (asyncRequest.succeeded()) {
+                        recordEndpointConnectTime(ctx, connectStartNs);
```
ytvnr, line 222: `connectStartNs` is taken before `httpClientFactory.getOrBuildHttpClient(...)`. On the first request for an endpoint that call enters a synchronized block and builds the Vert.x client, SSL options mapping and keystore loading included, and concurrent requests block on the same monitor. All of that lands in `endpointConnectTimeNs`, which the javadoc presents as nginx `$upstream_connect_time`. The result is a large outlier on the first request per endpoint, and on bursts hitting a cold factory, in a metric whose stated purpose is to isolate connection pool saturation.

Taking the origin after the client has been obtained, right before `.request(options)`, would keep the measure to what the pool and the network contribute. If the client build cost is worth observing, it would deserve its own field.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/test/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnectorTest.java
```diff
@@ -181,7 +181,9 @@ void init() {
 
     @AfterEach
     void cleanUp() {
-        wiremock.resetAll();
+        // Stop it rather than just reset it: each test starts its own server, so anything left running holds on to
+        // its two ports and its thread pool for the rest of the JVM's life.
+        wiremock.stop();
```
ytvnr: Stopping the server here does plug the leak. Worth noting the cost though: the per test `new WireMockServer(...)` plus `start()` in `init()` remains, so the class now pays 31 full server lifecycles, HTTPS listener and self signed keystore included, plus 31 rounds of port binding on the CI host. The previous shape, one class level server with `resetAll()` between tests, was cheaper, and the bug was that `init()` overwrote the static field rather than the class level server itself.

This is an alternative to my comment on the `wiremock` field above, not a complement. Either keep the per test lifecycle and make the field non static, which is the minimal fix, or go back to a single server for the class with `resetAll()` in `@AfterEach` and resolve the target from that server's port, if the CI cost justifies it.

ytvnr (changes requested):
Thorough work, and the end to end validation against Elasticsearch in the description makes the intent easy to follow.

One point needs to be settled before merge. In `ResponseTimeProcessor`, the nanosecond branch is taken as soon as `requestStartNs` is set, which is the case for every HTTP request since `DefaultHttpRequestDispatcher` runs the same platform pre/post chain for every `ApiReactor`. Inside that branch the endpoint duration is read only from `getEndpointResponseTimeNs()`, and the only writers are `DefaultApiReactor` and `HttpConnector`, both v4. `SyncApiReactor`, the v2 emulation reactor, still writes milliseconds. So on a v2 emulation API the field keeps its `-1` default and the whole response time is attributed to the gateway latency. Details inline.

I also want to flag the WireMock field: `wiremock` stayed `static` while its lifecycle became per test, which turns it into shared mutable state. It works today only because the class runs single threaded. That is the second thing I would fix.

Minor: the description says the pom points at a snapshot to be replaced before merge, but it points at `gravitee-reporter-api` 2.8.0, which is released. That point looks settled.

The rest is test coverage and naming remarks, all non blocking.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/responsetime/ResponseTimeProcessor.java
phiz71: Confirmed, and it was a regression I introduced: before the nanosecond work the processor read the millisecond field, which SyncApiReactor does fill, so v2 emulation was correct.

Fixed in 71ab171291 by falling back to the millisecond value when the nanosecond one is unset, which keeps the change out of v2 territory. A request that never reached an endpoint leaves that field at its 0 default and still has the whole time charged to the gateway, as before.

Your point about the existing test was exactly right — it never set requestStartNs, so it only exercised the fallback branch. I put the new test back against the faulty code to check it actually catches it: it fails alone, and passes with the fix.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
phiz71: Agreed — and I had even documented it as intentional, which it should not have been. Building the client, SSL options and keystore included, under a lock other requests queue on, has no place in a field meant to isolate pool saturation.

Moved after the client is in hand, right before the request is issued, in 426ac62a3d.

gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/HttpRequestInternal.java
phiz71: Confirmed: VertxTcpRequest and HttpHealthCheckRequest both set timestamp without setting timestampNs.

Rather than make the accessor abstract — the interface is public, so that would break implementations outside this repo — the reading is now taken in AbstractRequest at construction, so no implementation can forget it. The default stays for external implementors, with a javadoc saying what returning -1 costs. 39763c46c8.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/metrics/MetricsProcessor.java
phiz71: Right — nothing pinned either end of that wiring, so dropping it would have sent every gateway back to the wall clock silently.

Two tests added in 39763c46c8: one on MetricsProcessor asserting the built metrics carry the request timestampNs, and one on VertxHttpServerRequest asserting the request carries a monotonic origin next to its wall-clock timestamp.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/test/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnectorTest.java
phiz71: Fair — the javadoc claimed a contract the tests did not hold. Added a cancellation case in 426ac62a3d: the chunks are subscribed to with a demand of one and then cancelled, and the setter must still have been called with at least the injected elapsed time. Moving the hook to doOnComplete now fails.

I left the upstream-error case out: onErrorResumeNext already turns a stream failure into an empty completion, so the doFinally fires on a path the cancellation test covers, and forcing a mid-stream upstream failure through wiremock reliably is more machinery than the assertion is worth. Happy to add it if you would rather have it explicit.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/test/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnectorTest.java
phiz71: Agreed, it only proved the setter was called. Replaced by a test capturing both the connect time and the response time and asserting the nesting the PR presents as the contract, plus one on an unreachable port asserting the setter is never called since the measure sits in the success branch. 426ac62a3d.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/test/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnectorTest.java
phiz71: Agreed. Took the alternative you describe below rather than this one: one server for the class with resetAll() between tests, which makes the field legitimately static again and saves the 31 server lifecycles. Folded into 426ac62a3d, whose message now says so — it had slipped in there unmentioned.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/test/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnectorTest.java
phiz71: This is the one I went with. You are right that the leak was init() replacing the static field, not the class-level server, so the cheap fix was to stop replacing it.

@BeforeAll now owns the server, @AfterEach resets it, @AfterAll stops it. Resetting clears both the stubs and the request journal, which is all the isolation these tests need — the suite runs green four times in a row at about 4 seconds.

gravitee-apim-reporter/gravitee-apim-reporter-common/src/main/java/io/gravitee/apim/reporter/common/formatter/elasticsearch/ElasticsearchFormatter.java
phiz71: Fixed in 6c5d0419c7 — and it was worse than the seven you flagged: the same rule was spelled out 22 times across the v2 metrics, the v4 metrics and the message metrics. All of them go through one private helper now, so the sentinel rule is stated once.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/test/java/io/gravitee/gateway/reactive/reactor/processor/responsetime/ResponseTimeProcessorTest.java
phiz71: Added on both, and the single remaining camelCase method in ResponseTimeProcessorTest is renamed in the same pass. I left the twenty-six legacy names in DefaultApiReactorTest alone — renaming them would bury this branch in an unrelated diff. 09fb7a1d33.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
phiz71: Your analysis is right, and it is the one point I have not turned into code, because I do not think moving the hook would fix it.

Demand does come from the client socket, so the terminal event fires when the last byte was written downstream. But backpressure propagates through the whole chain: hooking `endpointResponse.toFlowable()` before the policy composition would fire at the same instant, since the source is only drained as fast as the sink asks. In a streaming path with end-to-end backpressure, upstream read time and downstream write time are not separable — which is also why, as you note, a buffering response policy turns the measure back into upstream read time. The semantics end up depending on the policy chain either way.

So what is left is to decide and document, not to relocate. Two things follow, and I would rather settle them with you than pick unilaterally:

- The nginx analogy is misleading as written. nginx buffers by default, so its `$upstream_response_time` really is upstream read time. Ours is proxied transfer time. Either the javadoc says so plainly, or we drop the analogy.
- A slow client inflating `endpoint-response-time` is real, though the time it inflates with is time the gateway is not working either. Envoy splits that out as `%RESPONSE_TX_DURATION%`; if we want that distinction, it is a field of its own rather than a different hook.

For the case that motivated the ticket — a backend trickling for 100 seconds into a client that keeps up — the current measure is the correct one. My inclination is to document it as proxied transfer time, drop the nginx equivalence on that one field, and open a follow-up if the slow-client case turns out to matter. WDYT?

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/test/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnectorTest.java
ytvnr: Your reasoning holds: `onErrorResumeNext` already turns a stream failure into an empty completion, so the doFinally fires on the same path the cancellation test now covers. No need to add the upstream error case, forcing it reliably through wiremock would cost more machinery than the assertion is worth.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
ytvnr: Agreed on both counts. Your point that backpressure makes the two durations inseparable is right, so relocating the hook would buy nothing, and what is left really is a decision to document rather than a fix.

Go with your inclination: document the field as proxied transfer time, and drop the nginx equivalence on that one field, since nginx buffers by default and therefore measures something else. If the slow client case turns out to matter in practice, a dedicated field along the lines of Envoy `%RESPONSE_TX_DURATION%` looks like the right shape for it, in a follow-up rather than here.

ytvnr (approved):
Thanks for turning all of this around so quickly, and for going further than asked on the Elasticsearch formatter: 22 occurrences of the rule rather than the 7 I had spotted.

I checked the latency fix at the new head. The millisecond fallback keeps the v2 emulation path a difference, and the `0` default on the millisecond field preserves the previous behaviour for a request that never reached an endpoint. The two new tests cover both shapes, and putting the test back against the faulty code before trusting it is the right way round.

The one remaining point, the semantics of the endpoint response time, is a documentation change we settled in thread, so nothing blocking. Approving.

gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
phiz71: Documented as you suggested, in 351dfbcde8: the field is now presented as the proxied transfer it is, with the nginx equivalence dropped on it and the reason spelled out — nginx buffers by default, so its figure is a read into its own buffers, while ours includes waiting on a slow client. The ttfb and connect time keep their equivalences, which do hold.

The comparison table in the change note carries the same caveat, with a pointer to Envoy `%RESPONSE_TX_DURATION%` as the shape a dedicated field would take if the slow-client case shows up.

jhaeyaert (approved):

PR #19084, opened 2026-08-11
feat(console): reinstate Add API to Category dialog for Portal Next categories

dev-marek:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-18

## Description

Reinstates the "Add API to Category" dialog on the Portal Next catalog category screen
(Settings > Portal > Catalog), disabled since PORTAL-16 pending the new categoryIds-based
backend model from PORTAL-17.

- Removed the `apiAssociationEnabled = false` hardcode and its disabled-button tooltip.
- Added a dedicated `AddApiToCategoryDialogComponent`, modeled on the shared
  `GioApiSelectDialogComponent`'s autocomplete UX but sourced from already-loaded, published
  API navigation items (not a live API search), restricted to APIs not yet in this category.
- On add/remove, the category's ID is appended to / removed from the API's
  `PortalNavigationItem.categoryIds` via the existing management REST API (PUT, full-replace).
- The category's APIs table is now populated from navigation items whose `categoryIds`
  includes this category, with a new row-level "Remove" action.
- Added `categoryIds?: string[]` to the frontend `PortalNavigationApi` /
  `NewApiPortalNavigationItem` / `UpdateApiPortalNavigationItem` types and fixtures (backend
  support already existed from PORTAL-17; the frontend types were missing the field).
- Fixed two pre-existing spots in `portal-navigation-items.component.ts` (title-edit and
  drag-reorder flows) that rebuilt the nav-item update payload from scratch — since the PUT
  endpoint is a full replace, these would have silently wiped `categoryIds` on any unrelated
  edit of an API nav item once the field started being populated.

## Additional context

- Tested manually against a local full-stack build (Management API + Console UI): add/remove
  flows, empty-state, and the "not offered when already in category" dialog filter all verified
  end-to-end. A demo video/screenshot is pending and will be added before merge.
- Full console-webui test suite passes (4724/4750 green; the 4 pre-existing failures are in
  unrelated files on master and untouched by this branch).


gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
```diff
@@ -182,4 +283,58 @@ export class CategoryCatalogComponent implements OnInit {
...
+    );
+  }
+
+  private loadApiVMs(navItems: PortalNavigationApi[]): Observable<ApiVM[]> {
+    if (navItems.length === 0) {
+      return of([]);
+    }
+    return forkJoin(
+      navItems.map(navItem =>
+        this.apiService.get(navItem.apiId).pipe(
```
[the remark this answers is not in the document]
dev-marek: Not addressed: there's no bulk-get-by-ids endpoint on the management API for this — the closest is `POST /apis/_search` with an `ids` filter, but that's a search endpoint (different semantics/scoping from a direct get), not a batch get, so keeping the per-item `apiService.get()` calls here.

jaroslawlakomy-grav, line 317: Could we consider resolving these API IDs through POST /apis/_search with manageOnly=false instead of individual GET /apis/{apiId} calls? Apart from avoiding the N+1 request pattern, the search endpoint appears better aligned with the read-only permission used to display this card. With the current approach, a user with environment-api-r but no API management permission may receive a 403, causing the entire table to appear empty.

gravitee-apim-console-webui/src/portal/catalog/category/category.component.html
jaroslawlakomy-grav: Could we review the permission combination used for this section? The UI currently checks API permissions, while the underlying /portal-navigation-items endpoints require ENVIRONMENT_DOCUMENTATION/READ and ENVIRONMENT_DOCUMENTATION/UPDATE. Since these permissions can be assigned independently, some custom roles may see actions that end with a 403, while other endpoint-authorized users may not see them. A small permission-matrix test could help document the intended combination.

gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
```diff
@@ -133,6 +158,16 @@ export class CategoryCatalogComponent implements OnInit {
...
     );
+
+    this.apis$ = this.refreshApis.pipe(
+      switchMap(() => {
+        if (this.mode !== 'edit') {
+          return of([]);
+        }
+        return this.publishedApiNavigationItemsInCategory$().pipe(switchMap(navItems => this.loadApiVMs(navItems)));
+      }),
+      catchError(() => of([])),
```
jaroslawlakomy-grav: Could we consider handling the error inside the per-refresh switchMap rather than after the complete refreshApis pipeline? The current catchError(() => of([])) completes the stream after the first failure, so subsequent add/remove operations cannot reload the table. It may also be useful to distinguish a loading error from a genuine empty category and provide a retry path.

gravitee-apim-console-webui/src/portal/catalog/category/category.component.html
```diff
@@ -88,6 +87,21 @@ <h3>APIs</h3>
...
+                <th mat-header-cell *matHeaderCellDef id="api-list-actions"></th>
+                <td mat-cell *matCellDef="let element">
+                  <button
+                    mat-button
+                    (click)="removeApiFromCategory(element)"
+                    matTooltip="Remove API"
+                    *gioPermission="{ anyOf: ['environment-api-u'] }"
+                  >
+                    <mat-icon svgIcon="gio:trash"></mat-icon>
+                  </button>
```
jaroslawlakomy-grav, line 102: Could we add an accessible name to this icon-only action, for example aria-label="Remove {{ element.name }} from category"? The tooltip provides a description but may not give the button a usable name for assistive technologies. This would also let the harness locate the action through its accessible label instead of the mattooltip attribute.

jaroslawlakomy-grav (commented):
There are some permissions consideration, so I made them as questions/sugestions to double check if it's ok or not. Please also take a look on SonarCloud warnings

gravitee-apim-console-webui/src/portal/catalog/category/category.component.html
dev-marek: Good catch — fixed in 6ba33141e9. Swapped the `environment-api-r`/`environment-api-u` checks for `environment-documentation-r`/`environment-documentation-u`, matching the `@Permissions` annotations on `PortalNavigationItemsResource`/`PortalNavigationItemResource` (ENVIRONMENT_DOCUMENTATION READ/UPDATE).

gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
dev-marek: Fixed in 6ba33141e9 — moved `catchError` inside the inner `switchMap` so it only swallows failures for that specific load, instead of terminating the whole `refreshApis` stream and blocking all future add/remove-triggered reloads.

gravitee-apim-console-webui/src/portal/catalog/category/category.component.html
dev-marek: Fixed in 6ba33141e9 — added `aria-label="Remove {{ element.name }} from category"` to the button, and simplified the harness to grab the only button in the actions cell instead of matching on the `mattooltip` attribute.

gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
dev-marek: Agreed — resolved properly in #19092: `GET /portal-navigation-items` now accepts `includes=apis` and returns a `metadata.apis` sidecar map in the same response, scoped to `ENVIRONMENT_DOCUMENTATION:READ` (the permission this screen actually requires), so the per-item `apiService.get()` N+1 — and the permission-mismatch concern — are both gone.

jaroslawlakomy-grav (approved):

PR #19086, opened 2026-08-12
feat(analytics): expose native Kafka metrics in the analytics engine [ESM-106]

ThibaudAV:
https://gravitee.atlassian.net/browse/ESM-106

## Why

The native Kafka reactor already reports throughput, connection and latency counters every 5 seconds, but they were only reachable through the legacy v2 histogram API. The analytics engine knew a single native metric, `NATIVE_CONNECTIONS_SUMMARY`, which is why a Kafka dashboard could say *that* connections fail but never *which service*, *which client*, or *how much traffic* flows.

This exposes the `event-metrics` data stream as a new engine family, and completes the connection family with the dimensions an error breakdown actually needs.

## What

- **New family** (`event-metrics`): 21 metrics — throughput per topic, active connections, authentications, per-operation counters and durations — with a dedicated field resolver, three query adapters, a query service and their catalog entries. All three query kinds are supported, unlike the connection family which cannot answer measures.
- **Connection family**: `fromFacet` now resolves `API`, `APPLICATION` and `PLAN`, which the catalog already declared — a query passed validation and then threw in the adapter. Adds `NATIVE_FAILURE_SIDE` and `NATIVE_CLIENT_ID`, the only two dimensions that survive a failed connection.
- **Gamma**: `NATIVE_TOPIC`, `NATIVE_OPERATION`, `NATIVE_FAILURE_SIDE` and `NATIVE_CLIENT_ID` as observability filters.

## Design notes, all dictated by the data

- **Durations are per-window sums, not samples.** The only sound average is `sum(durations) / sum(samples)`, computed by a `bucket_script`, and each of the three phases has **its own denominator** — sampled where its own counter is incremented. Percentiles are refused: they would describe the distribution of window sums.
- **`doc-type: api` documents carry neither `plan-id` nor `app-id`** — the reactor records those metrics before a connection authenticates. Facets are declared per metric accordingly rather than uniformly.
- **Bucketed queries refuse metrics spanning several `doc-type`.** The bucket sits above the measures, so the per-metric filter envelope would land inside it: the sort path stops resolving, and a derived duration reads back as 0 with no error. The flat measures path keeps the envelope — it has no bucket and resolves by recursion.
- **`ERROR_KEY` is deliberately not exposed for native.** The gateway currently writes `UNKNOWN_SERVER_ERROR` for every failure, including a genuine SASL rejection, so a breakdown by error key would chart one meaningless bar. Worth a reactor fix.

## Validation

Unit tests on the adapters (52), catalog loading, and an end-to-end run against a real stack: 50 messages produced through the gateway read back as 50, `PRODUCE` broker latency at 1.4 ms, `FETCH` at 111 ms (a blocking long-poll, slow by design), and the mixed-doc-type guard rejecting with the offending metrics listed.

## Follow-ups, not in this PR

- No integration test against Elasticsearch for the durations — the infrastructure exists (`AnalyticsElasticsearchRepositoryTest.Engine`) but lacks `operation` fixtures. It is the only level that would have caught the zeroed durations.
- `toTermsLeaf` / `applySorts` are now duplicated across three adapters; the same sort-path issue exists in the HTTP one.
- Value lookup for the new KEYWORD filters throws `UnsupportedOperationException` in `FilterValuesQueryServiceImpl`, as it already does for the EDGE filters.

Related: ESM-106, consumed by gravitee-io/gravitee-gamma-module-esm#ESM-106-kafka-service-dashboards


a-cordier (approved):

a-cordier:
maybe @SamHardi can you 👀 I did check the analytics engine integration, but I can see things around filters that you might want to check

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/EventMetricsTimeSeriesQueryAdapter.java
aqibmohammadkhan, line 1: Can we add a unit test for this file.
Measures and facets have tests, including mixed-doc-type and duration leaves. Time series is the path that nests date_histogram → measures → another single-bucket date_histogram for durations (NativeOperationDurationBuilder, same trick as HttpErrorRateBuilder).
AggregationAdapter unwraps that via "_" + name + inner buckets - it should work, and they ran a live stack (50 messages, PRODUCE 1.4 ms, FETCH 111 ms). There is still no adapter test that the time-series JSON and unwrap stay aligned.

aqibmohammadkhan (approved):

PR #19092, opened 2026-08-12
feat(console): batch-resolve category API summaries via portal-navigation-items includes

dev-marek:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-18

## Description

Resolves review comment #1 on #19084 (the N+1 `apiService.get()` calls in the category APIs
table): `GET /portal-navigation-items` gains an opt-in `includes=apis` query parameter that
returns a `metadata.apis` sidecar map (nav-item-id → `{ id, name, apiVersion, contextPath }`)
in the same response, scoped to the same `ENVIRONMENT_DOCUMENTATION:READ` permission the
endpoint already requires — resolving the permission-mismatch concern the reviewer also raised
(`apiService.get()` needs broader API-management read access than this screen is gated on).

- `ListPortalNavigationItemsUseCase` gains `includeApis` on `Input` and an `apis` map on
  `Output`, resolved via the existing `ApiPortalSearchQueryService` port (version-agnostic,
  already used by the portal-facing `_search` endpoint — no legacy service dependency).
- `PortalNavigationItemsMapper` maps those domain APIs straight to the REST summary DTO,
  computing `contextPath` from the domain `Api`'s own listener/proxy data (mirrors the
  frontend's `getApiAccess`/`getApiContextPath` utilities) — no second network round-trip.
- `category.component.ts` drops the `forkJoin` of individual `apiService.get()` calls entirely
  and consumes `metadata.apis` from the single navigation-items request.

## Additional context

- Full backend suite for the three touched modules passes (only pre-existing, unrelated
  failures elsewhere: a locale-dependent test and two log-pagination tests, neither touched by
  this change).
- Full frontend suite for the touched files passes.
- Verified end-to-end against a local full-stack build via curl: `metadata.apis` populated with
  correct `contextPath` per API type (V2 virtual hosts, V4 HTTP/TCP listeners, V4 native Kafka
  listeners), absent when `includes` is omitted (backward compatible).

### DIFF
--- a/gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
+++ b/gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
@@ -15,7 +15,7 @@
  */
 import { Component, DestroyRef, inject, OnInit } from '@angular/core';
 import { FormControl, FormGroup, FormsModule, ReactiveFormsModule, Validators } from '@angular/forms';
-import { BehaviorSubject, forkJoin, Observable, of } from 'rxjs';
+import { BehaviorSubject, Observable, of } from 'rxjs';
 import { ActivatedRoute, Router, RouterLink } from '@angular/router';
 import { catchError, filter, map, switchMap, tap } from 'rxjs/operators';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
@@ -51,15 +51,14 @@ import { GioPermissionService } from '../../../shared/components/gio-permission/
 import { SnackBarService } from '../../../services-ngx/snack-bar.service';
 import { PortalCategoryService } from '../../../services-ngx/portal-category.service';
 import { PortalNavigationItemService } from '../../../services-ngx/portal-navigation-item.service';
-import { ApiV2Service } from '../../../services-ngx/api-v2.service';
 import {
   CreatePortalCategory,
   PortalCategory,
   PortalNavigationApi,
+  PortalNavigationItemApiSummary,
   UpdateApiPortalNavigationItem,
   UpdatePortalCategory,
 } from '../../../entities/management-api-v2';
-import { getApiContextPath } from '../../../shared/utils';
 
 interface ApiVM {
   id: string;
@@ -118,7 +117,6 @@ export class CategoryCatalogComponent implements OnInit {
   private refreshApis = new BehaviorSubject(1);
   private destroyRef = inject(DestroyRef);
   private readonly portalNavigationItemService = inject(PortalNavigationItemService);
-  private readonly apiService = inject(ApiV2Service);
   private readonly matDialog = inject(MatDialog);
 
   constructor(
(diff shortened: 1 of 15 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/PortalNavigationItemsMapper.java
```diff
@@ -263,4 +271,77 @@ default UUID mapCategoryId(PortalCategoryId id) {
...
+            return tcpHost;
+        }
+
+        String kafkaHost = listeners
+            .stream()
+            .filter(listener -> listener.getType() == ListenerType.KAFKA)
+            .map(KafkaListener.class::cast)
+            .map(KafkaListener::getHost)
+            .findFirst()
+            .orElse(null);
```
jaroslawlakomy-grav: Could we preserve the Kafka listener port in this summary? The previous getApiContextPath implementation formats native Kafka access as host:port, while this replacement returns only host, so the category table displays an incomplete address. Please append the port when present and add a focused native Kafka mapper test.

gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
```diff
@@ -290,41 +293,36 @@ export class CategoryCatalogComponent implements OnInit {
...
-  private selectableApiNavigationItems$(): Observable<PortalNavigationApi[]> {
-    return this.publishedApiNavigationItems$().pipe(
-      map(navItems => navItems.filter(navItem => !(navItem.categoryIds ?? []).includes(this.categoryId))),
+  private publishedApiNavigationItemsWithSummaries$(): Observable<{
+    navItems: PortalNavigationApi[];
+    apisById: Record<string, PortalNavigationItemApiSummary>;
+  }> {
+    return this.portalNavigationItemService.getNavigationItems('TOP_NAVBAR', ['apis']).pipe(
+      map(response => ({
+        navItems: response.items.filter((item): item is PortalNavigationApi => item.type === 'API' && item.published),
```
jaroslawlakomy-grav: Could we keep the published filter only in selectableApiNavigationItems$()? PORTAL-18 defines the category table from API navigation items whose categoryIds contains the current category. With the current filter, an API assigned before being unpublished disappears from the table and its association can no longer be removed here.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/PortalNavigationItemsMapper.java
dev-marek: Good catch — fixed. Kafka context path now formats as `host:port` (matching the frontend's `getApiAccess` formatting), and added `should_include_the_port_in_a_native_kafka_api_context_path` in `PortalNavigationItemsMapperTest`.

gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
dev-marek: Good catch — fixed. The `published` filter is now only applied in `selectableApiNavigationItems$()` (the add-dialog candidates); the main table load (renamed `apiNavigationItemsWithSummaries$`) no longer filters on it, so an API unpublished after being assigned still shows up and can be removed. Added a spec covering that case.

jaroslawlakomy-grav (approved):

PR #19097, opened 2026-08-12
fix(repository-mongodb): backfill null segment on portal navigation items before guard query

dev-marek:
**Issue**

https://gravitee.atlassian.net/browse/PORTAL-116

**Purpose**

Fix a MongoDB-only upgrade failure where `EnvironmentsDefaultPortalNavigationItemsUpgrader` throws a NullPointerException on legacy `portal_navigation_items` documents with a null `segment`, blocking `apim-api` startup on 4.12.x upgrades.

**Summary of changes**

- Added `PortalNavigationItemDefaultSegmentMongoUpgrader`, a raw-driver Mongo pre-upgrader that backfills null/empty `segment` with `__CHANGE_ME__` (same placeholder the JDBC Liquibase changelog already uses), ordered right after the existing `rootId` pre-upgrader and well before the guard query that was failing.
- Added `PortalNavigationItemSegmentIndexUpgrader`, mirroring the existing `rootId` index, so the backfill query isn't an unindexed collection scan (this Mongo test suite runs with `notablescan=true`).
- Added `PortalNavigationItemDefaultSegmentMongoUpgraderTest` covering missing/null/empty/already-set/idempotent cases.

**Out of scope**

- JDBC is unaffected (already backfills `segment` via Liquibase) — no changes there.
- Not replacing `EnvironmentsDefaultPortalNavigationItemsUpgrader`'s guard query with a lighter `.exists()`-style check, as suggested as a longer-term follow-up in the ticket.

**Testing**

- Added `PortalNavigationItemDefaultSegmentMongoUpgraderTest`, run against a real Testcontainers MongoDB instance (`mvn test -Dtest=PortalNavigationItemDefaultSegmentMongoUpgraderTest -pl gravitee-apim-repository/gravitee-apim-repository-mongodb`) — all 5 cases pass.
- Verified end to end: started a 4.11.24 install (Docker Compose, MongoDB), confirmed 7 default navigation items seeded with no `segment` field at all (legacy pre-`segment` state). Built this branch's management-api image from source and swapped it in against the same Mongo data (in-place upgrade). Result: clean startup, no exception, all 7 items preserved, 0 with null/empty `segment`, all correctly backfilled to real slugs by the existing ORM-based upgrader.
- No REST endpoint or UI is touched, so no Postman recording/screenshot applies.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/upgrade/upgrader/index/portalnavigationindex/PortalNavigationItemSegmentIndexUpgrader.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-mongodb/src/main/java/io/gravitee/repository/mongodb/management/upgrade/upgrader/index/portalnavigationindex/PortalNavigationItemSegmentIndexUpgrader.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


jaroslawlakomy-grav (approved):
Looks good to me. The MongoDB upgrader correctly backfills missing, null, and empty segment values before the dependent migration runs, while preserving existing valid values. The implementation is idempotent, the relevant scenarios are covered by tests, and all checks pass

PR #19103, opened 2026-08-12
fix: API Score evaluation fails when many custom rulesets are configured

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-107

## Description

Two defects surfaced by the ticket. Neither is the root cause, which is fixed separately in `spectral-scoring-service`. This PR makes APIM behave correctly when the provider fails.

### 1. The job deadline ignored the number of rulesets

`ScoreApiRequestUseCase#deadLine` computed `Math.min(1, nbRulesets)`, always `1`. The TTL stayed at `10s x nbAssets + 30s` whatever the ruleset count, while the javadoc above it stated "10 seconds per asset **per ruleset**". Past that TTL the job is flipped to `TIMEOUT` on read and the console stops polling.

`min` -> `max`, capped at 15 min so a PENDING job cannot be polled indefinitely. Note this is not what the reporter hit: the provider fails after ~10s, well within even the old 40s deadline.

### 2. Provider failures were stored as successful empty reports

Cockpit reports a failure as `ScoringResult(assetDiagnostics=[], success=false, error=...)`.
`ScoringResponseCommandHandler` never read `success` nor `error`: it was processed as a successful run with zero assets, which **deleted the existing report** and replaced it with an empty one (`score = -1` -> `summary = null` -> "No scorable assets"), job left in `SUCCESS`.

Failures now short-circuit before `deleteByApi`: the job goes to `ERROR` carrying the provider message, and the previous report is left untouched. **This is the path the reporter is on.**

### Root cause

`spectral-scoring-service` spawns one Deno subprocess per (asset x ruleset) with unbounded concurrency. Under its Helm limits (`cpu: 500m`), past ~7 rulesets they starve each other, cross the 10s SIGKILL, and return truncated output — measured: 5 rulesets pass in 7s, 7 fail, 11 crash the service. Bounding the concurrency makes 15 rulesets pass. Fixed in that repo.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/scoring/use_case/ScoreApiRequestUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/scoring/use_case/ScoreApiRequestUseCase.java
@@ -56,6 +56,10 @@ import lombok.RequiredArgsConstructor;
 @UseCase
 public class ScoreApiRequestUseCase {
 
+    private static final long SECONDS_PER_RUN = 10;
+    private static final Duration MARGIN = Duration.ofSeconds(30);
+    private static final Duration MAX_TTL = Duration.ofMinutes(15);
+
     private final ApiCrudService apiCrudService;
     private final ApiDocumentationDomainService apiDocumentationDomainService;
     private final ApiExportDomainService apiExportDomainService;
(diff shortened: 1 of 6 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/scoring/use_case/ScoreApiRequestUseCase.java
```diff
@@ -196,16 +200,21 @@ private record RulesetAndFunctions(List<ScoreRequest.CustomRuleset> rulesets, Li
...
      */
     private Duration deadLine(ScoreRequest request) {
-        var margin = Duration.ofSeconds(30);
         int nbAssets = request.assets().size();
-        int rulesets = Math.min(1, request.customRulesets().size());
-        int numberOfRuns = nbAssets * rulesets;
-        return Duration.ofSeconds(10).multipliedBy(numberOfRuns).plus(margin);
+        int nbRulesets = Math.max(1, request.customRulesets().size());
+        var ttl = Duration.ofSeconds(SECONDS_PER_RUN).multipliedBy((long) nbAssets * nbRulesets).plus(MARGIN);
+        return ttl.compareTo(MAX_TTL) > 0 ? MAX_TTL : ttl;
```
michel-barret:
```suggestion
        return Math.min(ttl, MAX_TTL);
```

michel-barret (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/scoring/use_case/ScoreApiRequestUseCase.java
thomasballivet: Done. `Math.min` doesn't take `Duration`, so applied on seconds:
`Duration.ofSeconds(Math.min(ttlInSeconds, MAX_TTL.toSeconds()))`

michel-barret: my bad I miss that isn’t long

PR #19110, opened 2026-08-12
feat(rest-api): migrate Classic portal categories to Portal Next on upgrade

dev-marek:
**Issue**

https://gravitee.atlassian.net/browse/PORTAL-29

**Purpose**

Automatically migrate existing Classic portal categories to Portal Next categories on upgrade, so portal admins don't need to recreate their category taxonomy.

**Summary of changes**

- `ClassicCategoriesMigrationUpgrader`: one-time, idempotent `Upgrader` (order 722) that, per environment, migrates Classic categories to Portal Next categories and carries over the classic API-to-category assignments onto the corresponding Portal Next API navigation items (`PortalNavigationApi.categoryIds`).
  - Skips an environment entirely if it already has Portal Next categories (defensive idempotency on top of the framework's persisted upgrader marker).
  - Matches classic API category references by both id and key, since `Api.categories` entries are ambiguous depending on the API's history (see `ApiV4CategoriesUpgrader`).
  - A missing navigation item for an API is a normal no-op, not an error — most Classic APIs were never added to a Portal Next listing.
- `ClassicCategoryToPortalCategoryMapper`: pure mapper — Classic `name`→`title`, `description` carried over, `visible` forced `true`, `picture`/documentation page link dropped (no Portal Next equivalent).
- `UpgraderOrder.CLASSIC_CATEGORIES_MIGRATION_UPGRADER = 722`.

**Out of scope**

- Classic category `picture` and documentation page link (no Portal Next equivalent — dropped, not migrated).
- Re-running the migration after manual edits to Portal Next categories (the per-environment skip treats any existing Portal Next category as "already migrated").
- A pre-existing bug found during verification: the portal REST `_search?type=catalog` endpoint silently ignores the `categoryId` filter (only `type=api` honors it). Unrelated to this change — happy to file separately if wanted.

**Testing**

- Unit tests: `ClassicCategoryToPortalCategoryMapperTest` (mapping rules, picture/hidden dropped), `ClassicCategoriesMigrationUpgraderTest` (migration, skip-when-already-migrated, API-category assignment by id and by key, no-op when no nav item exists, idempotency across two `upgrade()` runs for both categories and nav-item assignment).
- Manual end-to-end verification: ran the stack on released 4.12.0 with a fresh MongoDB volume, created Classic APIs/categories/assignments and Portal Next nav items via REST, stopped the stack (data preserved), built and booted the local snapshot against the same volume, and confirmed via REST that Portal Next categories were created correctly, API navigation items got the right `categoryIds`, the portal `_search?type=api&categoryId=...` endpoint returned the correct APIs per category, and a second boot skipped the upgrader with no duplicates.

**Additional context**

No Postman recording/video captured — verification was done via manual curl calls against the real REST endpoints, detailed above. Let me know if you'd like a recorded Postman collection added.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ClassicCategoriesMigrationUpgrader.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ClassicCategoriesMigrationUpgrader.java
@@ -1,0 +1,209 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.service.impl.upgrade.upgrader;
+
+import static io.gravitee.rest.api.service.impl.upgrade.upgrader.UpgraderOrder.CLASSIC_CATEGORIES_MIGRATION_UPGRADER;
+
+import io.gravitee.apim.core.portal_category.model.PortalCategory;
+import io.gravitee.apim.core.portal_category.model.PortalCategoryId;
+import io.gravitee.apim.core.portal_category.query_service.PortalCategoryQueryService;
+import io.gravitee.apim.core.portal_category.use_case.CreatePortalCategoryUseCase;
+import io.gravitee.apim.core.portal_page.crud_service.PortalNavigationItemCrudService;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationApi;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemQueryCriteria;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemType;
+import io.gravitee.apim.core.portal_page.model.UpdatePortalNavigationItem;
+import io.gravitee.apim.core.portal_page.query_service.PortalNavigationItemsQueryService;
+import io.gravitee.node.api.upgrader.Upgrader;
+import io.gravitee.node.api.upgrader.UpgraderException;
+import io.gravitee.repository.exceptions.TechnicalException;
+import io.gravitee.repository.management.api.ApiRepository;
+import io.gravitee.repository.management.api.CategoryRepository;
+import io.gravitee.repository.management.api.EnvironmentRepository;
+import io.gravitee.repository.management.api.search.ApiCriteria;
+import io.gravitee.repository.management.api.search.ApiFieldFilter;
+import io.gravitee.repository.management.model.Category;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import lombok.CustomLog;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Component;
+
+/**
+ * One-time migration of Classic portal categories to Portal Next categories, so portal admins
+ * don't need to recreate their category taxonomy after an upgrade. Also migrates the classic
+ * API-to-category assignments onto the corresponding Portal Next API navigation items.
+ *
+ * Ships in the same release as the Portal Next category feature, and the per-environment skip in
+ * {@link #applyUpgrade()} means it effectively runs once per environment, against that release's
+ * schema. It is deliberately not hardened against the domain/service dependencies it calls (e.g.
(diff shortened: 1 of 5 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ClassicCategoriesMigrationUpgrader.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    private boolean applyUpgrade() throws TechnicalException {
+        for (final var environment : environmentRepository.findAll()) {
+            // Portal Next categories already exist for this environment: either this upgrader
+            // already ran, or an admin already created Portal Next categories by hand. Either way,
+            // migrating classic categories on top would create duplicates.
+            if (!portalCategoryQueryService.findByEnvironmentId(environment.getId()).isEmpty()) {
+                continue;
+            }
```
jaroslawlakomy-grav: Could we make this migration resumable at category and assignment level instead of using “any Portal category exists” as the completion marker? If one category is created and a later create, navigation search, or update fails, a retry skips the whole environment at line 105. The catches here and in the navigation update loop also allow the upgrader to return true, so the migration can be recorded as complete while data is missing. Could we only report success after all required work completes and reconcile partial progress on retries without creating duplicates?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ClassicCategoriesMigrationUpgraderTest.java
```diff
@@ -0,0 +1,326 @@
...
+            assertThat(portalNavigationItemsStorage).isEmpty();
+        }
+    }
+
+    @Nested
+    class Idempotency {
+
+        @Test
+        @SneakyThrows
+        void should_not_create_duplicate_portal_categories_when_upgrade_runs_twice() {
```
jaroslawlakomy-grav, line 360: Could we add the repository-backed integration test requested by PORTAL-29? This unit test manually copies state between independent in-memory fakes before the second invocation, so it does not exercise shared persistence, partial persisted state, or the upgrader completion marker. Running the migration twice against the same repository-backed state would verify idempotency at the actual migration boundary.

jaroslawlakomy-grav (commented):
One general note about the category migration: I’m not sure what options we have within the upgrader framework to verify that every category was migrated successfully. Ideally, retries would reconcile any partial migration, but if that is difficult to implement cleanly, I’m fine with keeping the current approach as is.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ClassicCategoriesMigrationUpgrader.java
dev-marek: Good catch, thanks — fixed in 52e873a. `applyUpgrade()` no longer skips the whole environment just because it has *some* Portal Next categories: it now looks up an existing category by title per classic category (title is already environment-scoped-unique via `PortalCategoryDomainService#validateTitleUniqueness`, so it's a safe, already-enforced reconciliation key) and reuses it instead of recreating, or creates it if missing. And the per-item catches now feed into the return value — `applyUpgrade()` returns `false` if any category or assignment failed, so the upgrader framework does *not* persist its completion marker and the next node startup retries automatically instead of the failure being silently swallowed as "done".

Added `PartialFailureRecovery#should_return_false_and_let_a_retry_reconcile_without_recreating_already_migrated_categories`, which simulates exactly the scenario you described: one category creation fails, `upgrade()` returns `false`, and a subsequent run completes the migration without duplicating the category that had already succeeded.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ClassicCategoriesMigrationUpgraderTest.java
dev-marek: Fair point on the manual-copy step — I looked into what a real repository-backed test would take here before deciding how to handle it.

`gravitee-apim-rest-api-service` currently has zero tests that run against a real database — every `Upgrader` test in this module (this one included, following existing precedent) is Mockito + in-memory-fake based. The only place in this repo that boots a real Mongo for tests (`MongoDBContainer` via Testcontainers) is `gravitee-apim-repository-mongodb`'s own test suite, and it isn't published as a reusable test-jar this module could pull in — adding one here would mean a new cross-module test dependency and Docker becoming a requirement of this module's test phase for the first time, for one test class. That felt like a bigger, precedent-setting infra decision than this PR should make unilaterally, so I've left it out rather than bolt it on quietly.

What I did fix is the substance of what that test gap was pointing at (see the reply on the other thread): the migration is now actually resumable — a retry reconciles by title instead of the completion marker ever getting persisted on a partial failure. Happy to open a follow-up ticket for real DB-backed Upgrader test infra as its own piece of work if that's useful to the team.

jaroslawlakomy-grav (approved):

PR #19115, opened 2026-08-13
feat(gamma): added env alert educational UX

arpzilla-dev:
## Issue

https://gravitee.atlassian.net/browse/APIM-14910

## Description

Adds the environment-level Alerts entry in the Gamma platform module: Monitoring nav item, /alerts route guarded by environment-alert-r, and an educational empty-state page that introduces why/how alerts work. List, create, and activity flows are deferred to follow-up stack PRs

## Additional context

image

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/AlertsPage.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/AlertsPage.tsx
@@ -1,0 +1,57 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/AlertsPage.tsx
ankita-gupta21, line 1: even if you have create permission (environment-alert-c) and the list is empty, there is no “Add alert” button. You only see the educational text. On API `@alerts`, that button still shows when you can edit.

arpzilla-dev: done

ankita-gupta21 (approved):

PR #19120, opened 2026-08-13
feat(gateway): resolve dot segments before routing, behind handlers.request.pathHandling

ytvnr:
**Description**

Reported on 4.11.6 (Zendesk 17952). The gateway resolves the listener context path against the request path exactly as it arrived, enforces the matched API's plan, then appends the remaining path to the endpoint target without resolving it. A conforming upstream applies RFC 3986 §5.2.4 and serves a different resource than the one the gateway authorized.

Reproduced on a clean stack: with `alpha` on `/alpha/api/` carrying no plan and `beta` on `/beta/api/` behind an api-key plan, both pointing at the same upstream host, `GET /alpha/api/../../beta/api/echo` reaches beta's backend with **no credential at all**, while `GET /beta/api/echo` correctly answers 401. The same holds for `%2e%2e`, and `/alpha/api/../../elsewhere/resource` escapes the subtree of the configured endpoint target entirely — no second API needed.

### What this adds

A single gateway-level setting, `handlers.request.pathHandling`, applied **before** the listener is resolved and therefore before any plan is enforced:

| Mode | Behaviour |
|---|---|
| `RAW` | **Default.** The path is used, and forwarded, exactly as received. Today's behaviour. |
| `REJECT` | Answers 400 when the normalized path differs from the one received. Nothing is rewritten, no routing decision changes. |
| `NORMALIZE` | Resolves the dot segments and uses the result for listener resolution, plan enforcement, flow selection and the upstream URI. |

`NORMALIZE` does not block the request, it makes the gateway read it for what it means: `/alpha/api/../../beta/api` is a call to beta, so beta's plan applies and answers 401 for want of a key. Where the resolved path matches no listener, 404.

State of the art, checked rather than assumed: nginx normalizes before location matching, but Envoy documents `normalize_path` as **false** by default ("will default to true in the future"), HAProxy makes it an explicit `http-request normalize-uri` action, and Apache ships `AllowEncodedSlashes Off`. Normalization is not universally on by default — what is universal is that the behaviour is configurable, which is what the gateway was missing.

Normalization decodes percent-encoding **only** in segments that spell a dot segment, so `%2e%2e` is covered while unrelated encoded characters are carried over untouched. Encoded slashes are deliberately left alone: turning `%2F` into a separator has its own breakage profile and belongs to its own option, as Envoy separates it into `path_with_escaped_slashes_action`. Duplicate slashes are preserved for the same reason.

### Scope of this increment

`REJECT` applies to every definition version: it answers before the acceptor is resolved, so it never reaches the code that distinguishes V4 from V3. Only established at code level here — the integration tests cover V4 APIs.

`NORMALIZE` is **V4 only in this increment**. The V3 handler exposes the native request path through a wrapper that cannot be rewritten, so a normalized resolution would leave `pathInfo` derived from the raw path at the wrong offset; when a normalized path resolves to a V3 handler, the dispatcher resolves again on the raw path and V3 keeps its historical behaviour. Extending `NORMALIZE` to the earlier definition versions belongs to the fix, it is simply not in this first step — as does the identical concatenation in `DefaultProxyEndpoint`.

### Coverage

- `PathHandlingV4IntegrationTest` — the three modes against the same pair of APIs, one nested class per mode.
- `PathTraversalV4IntegrationTest` — pins the `RAW` behaviour, including the exact unresolved request line emitted upstream.
- `PathTraversalOrganizationGuardV4IntegrationTest` — the mitigation deployable without upgrading: a Groovy policy on an organization flow. It runs ahead of the security chain, so it closes the traversal for every API at once, but the last test pins its limit — the API is already selected when it runs, so it can only reject, never route to the API the resolved path designates.
- `RequestPathNormalizerTest` — dot segment resolution, encoded variants, above-root climbs, and everything deliberately left untouched.

The 257 existing unit tests of the reactor module still pass.

**Additional context**

**Open question — the default.** `RAW` is kept so that nothing changes on upgrade, which also means the gap stays open unless an operator opts in. The argument for flipping the default to `NORMALIZE` on `master`, while leaving `RAW` on the maintenance branches, is that a security fix shipped off by default protects nobody; the argument against is that switching changes which API a request carrying `../` is routed to, and what is sent upstream. Worth settling before this leaves draft.

Also worth knowing for anyone writing the Groovy guard: `PolicyResult.State` exposes `SUCCESS` and `FAILURE` only. Using a non-existent constant surfaces as `SecurityException: Failed to resolve getter method or field` from the sandbox, which reads like a sandbox denial rather than a typo.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-env/src/main/java/io/gravitee/gateway/env/RequestConfiguration.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-env/src/main/java/io/gravitee/gateway/env/RequestConfiguration.java
@@ -51,4 +51,24 @@ public class RequestConfiguration {
     ) {
         return new RequestClientAuthConfiguration(headerName);
     }
+
+    @Bean
+    public RequestPathConfiguration httpRequestPathConfiguration(@Value("${handlers.request.pathHandling:RAW}") String pathHandling) {
+        RequestPathHandling handling;
+        try {
+            handling = RequestPathHandling.valueOf(pathHandling.trim().toUpperCase());
+        } catch (IllegalArgumentException e) {
+            log.warn(
+                "Unknown value [{}] for handlers.request.pathHandling, falling back to {}. Expected one of {}.",
+                pathHandling,
+                RequestPathHandling.RAW,
+                java.util.Arrays.toString(RequestPathHandling.values())
+            );
+            handling = RequestPathHandling.RAW;
+        }
+        if (handling != RequestPathHandling.RAW) {
+            log.info("Request path handling is set to {}: dot segments are resolved before the listener is resolved.", handling);
+        }
+        return new RequestPathConfiguration(handling);
+    }
 }
--- a/gravitee-apim-gateway/gravitee-apim-gateway-env/src/main/java/io/gravitee/gateway/env/RequestPathConfiguration.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-env/src/main/java/io/gravitee/gateway/env/RequestPathConfiguration.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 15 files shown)


ytvnr:
Superseded: the work continues on the branch renamed after APIM-14917, with a broader scope than the tests this PR carried.

PR #19125, opened 2026-08-13
feat(gamma-platform): added new two column sidebar UI

arpzilla-dev:
## Issue

[no specific ticket, its added as common change for all]
https://gravitee.atlassian.net/browse/APIM-14934

## Description

In gamma platform module, replace the single-column Platform sidebar with a two-column Graphene layout.
- Primary column: Organization, Environment, Team (no General, no “Platform” group label)
- Secondary column: existing Gamma items only, grouped as specified
- Clicking a primary section opens that section’s first visible item
- Permission filtering still hides items and drops empty groups/sections

## Additional context
NEW UI

[image]
image

[video]

https://github.com/user-attachments/assets/85be6ac8-e9e3-40db-8de1-af2231843d1a

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
@@ -14,15 +14,33 @@
  * limitations under the License.
  */
 import { useModuleRouting } from '@gravitee/gamma-modules-sdk/routing';
-import { buildLinearBreadcrumbs, SidebarNavigation, useLayoutConfig } from '@gravitee/graphene-core';
+import {
+    buildLinearBreadcrumbs,
+    ContextSidebar,
+    ContextToggleButton,
+    type NavItem,
+    SidebarGroup,
+    SidebarGroupContent,
+    SidebarMenu,
+    SidebarMenuButton,
+    SidebarMenuItem,
+    useLayoutConfig,
+} from '@gravitee/graphene-core';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { type ReactElement, useMemo } from 'react';
+import { createContext, type ReactElement, useCallback, useContext, useMemo, useState } from 'react';
 import { Navigate, Outlet, Route, Routes, useNavigate } from 'react-router-dom';
 
 import { PlatformToaster } from './PlatformToaster';
 import { APPLICATION_NAV_GROUPS, flattenApplicationDetailNavItems } from '../config/applicationDetailNavigation';
 import { applicationDetailTabElement } from '../config/applicationDetailPages';
-import { NAV_GROUPS } from '../config/navigation';
+import {
+    filterNavSections,
+    findNavSectionKey,
+    firstNavItemKey,
+    NAV_SECTIONS,
+    platformPrimaryNavItems,
+    type PlatformNavSection,
+} from '../config/navigation';
 import { PLATFORM_ROUTE_CONFIG } from '../config/routes';
 import { ApplicationDetailIndexRedirect, ApplicationDetailLayout } from '../features/applications/components/detail';
 import { useEnvironmentDictionaries } from '../features/dictionaries/hooks/useEnvironmentDictionaries';
@@ -124,10 +142,60 @@ function EntrypointsGuard() {
     return <EntrypointsAndShardingTagsPage />;
 }
 
+function PlatformPrimaryNavigation({
+    items,
+    activeItemKey,
+    onItemSelect,
+}: Readonly<{ items: NavItem[]; activeItemKey?: string; onItemSelect: (key: string) => void }>) {
+    return (
+        <SidebarGroup>
+            <SidebarGroupContent>
+                <SidebarMenu>
+                    {items.map(item => {
+                        const Icon = item.icon;
+                        return (
+                            <SidebarMenuItem key={item.key}>
+                                <SidebarMenuButton
+                                    isActive={item.key === activeItemKey}
+                                    tooltip={item.title}
+                                    onClick={() => onItemSelect(item.key)}
+                                >
+                                    {Icon ? <Icon /> : null}
+                                    <span>{item.title}</span>
+                                </SidebarMenuButton>
+                            </SidebarMenuItem>
+                        );
+                    })}
+                </SidebarMenu>
+            </SidebarGroupContent>
+        </SidebarGroup>
+    );
+}
+
+interface PlatformNavContextValue {
+    readonly activeNavKey: string;
+    readonly activeSection: PlatformNavSection | undefined;
+    readonly navigateToKey: (key: string) => void;
+    readonly contextExpanded: boolean;
+    readonly toggleContext: () => void;
+}
+
+const PlatformNavContext = createContext<PlatformNavContextValue | null>(null);
+
+function usePlatformNavContext(): PlatformNavContextValue {
+    const value = useContext(PlatformNavContext);
+    if (!value) {
+        throw new Error('PlatformSectionLayout must render inside ModuleLayout');
+    }
+    return value;
+}
+
 function ModuleLayout() {
     useEnvironmentPermissions();
 
     const permissionsReady = useEnvironmentPermissionsReady();
+    const [contextExpanded, setContextExpanded] = useState(true);
+    const toggleContext = useCallback(() => setContextExpanded(expanded => !expanded), []);
 
     // Only probe the resource once the permission map already says "yes" — that's the only case where a
     // 403 here would disagree with it. Probing unconditionally would hit both list APIs on every platform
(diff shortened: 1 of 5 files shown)


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
```diff
@@ -147,37 +189,67 @@ function ModuleLayout() {
...
+    const activeSectionKey = findNavSectionKey(visibleNavSections, activeNavKey) ?? visibleNavSections[0]?.key;
+    const activeSection = visibleNavSections.find(section => section.key === activeSectionKey);
+
+    const handleNavSelect = useCallback(
+        (key: string) => {
+            const section = visibleNavSections.find(candidate => candidate.key === key);
+            if (section) {
+                if (findNavSectionKey(visibleNavSections, activeNavKey) === section.key) {
+                    return;
+                }
```
arpzilla-dev: ignore: Intentional. Re-clicking the already-active primary section must not reset to the first item.

aqibmohammadkhan (commented):
Blocker — ModuleLayout and ApplicationDetailLayout now fight over the same layout slots
AppRoutes.tsx:235-253 makes the module-level layout own viewMode, contextExpanded, contextSidebar and leading. ApplicationDetailLayout.tsx:152-174 already owns those same four slots, and it renders inside ModuleLayout's <Outlet /> (AppRoutes.tsx:270). useLayoutConfig is a last-writer-wins setSlots, so the ancestor wins.

I reproduced both failure modes with a throwaway spec shaped like the real route tree (deleted after; not left in the tree):

- Deep link or reload on /applications/:id → sidebar owner came back module, expected application. React runs layout effects child-first, so ModuleLayout pushes last and the Environment section nav replaces the application's context sidebar and its ApplicationInfoHeader.
- Clicking a tab inside an application → same result. ModuleLayout's dep array contains breadcrumbs, memoised on react-router's navigate, whose identity is keyed on locationPathname (react-router useNavigateUnstable). Every navigation re-fires the parent's effect, while the detail layout's deps (basePath, application, contextExpanded) stay stable — so the parent steals the slot back mid-session.

Secondary effect of the same overlap: there are now two contextExpanded states (AppRoutes.tsx:171 and ApplicationDetailLayout.tsx:149), and the header toggle drives whichever layout last pushed leading. I also confirmed that a child unmount resets slots the parent still owns (viewMode fell back to global); today that self-heals only because the parent's deps happen to change on navigation.

The apim module avoids this by keeping module-level config to navigation + breadcrumbs only (apim AppRoutes.tsx:142-148). Suggested fix in the same spirit: leave navigation (the primary column) in ModuleLayout, and move viewMode/contextExpanded/contextSidebar/leading into a layout element wrapping only the section pages — never an ancestor of ApplicationDetailLayout.

Worth noting breadcrumbs were already clobbered by this mechanism before the PR, so the ancestor/descendant conflict isn't new — but this change extends it to the sidebar, where it's plainly visible.

Improvement — the headline acceptance criterion is untested
"Clicking a primary section opens that section's first visible item" has no test. firstNavItemKey is covered as a pure function, but the wiring in handleNavSelect (AppRoutes.tsx:212-228) isn't — the spec mocks ContextSidebar/ContextToggleButton to null and never invokes a select handler. It's cheap to cover: pull navigation.props.onItemSelect off the captured layout config, call it with 'team', and assert navigateToKey got 'users'. The already-active-section no-op branch (AppRoutes.tsx:216-218) is likewise unexercised.

Improvement — one callback serving two key namespaces
handleNavSelect is handed both to the primary nav (section keys) and to ContextSidebar (nav item keys), and disambiguates by lookup. If a nav item key ever collides with 'organization' | 'environment' | 'team', a context-sidebar click silently becomes a section jump. Two small handlers would remove the coupling and make each one testable on its own.

Nits

- AppRoutes.tsx:246 — activeSection ? … : null guards a case that can't occur: access-management and security-plan-types fall through isNavItemVisible to return true, so Organization and Environment always survive filterNavSections.
- navigation.ts:98-107 — the Team section's only group is also labelled Team, so the label repeats directly under the section header.
- AppRoutes.tsx:216 — recomputing findNavSectionKey(...) instead of reusing activeSectionKey is deliberate (the ?? visibleNavSections[0] fallback would wrongly no-op), but nothing says so.

arpzilla-dev:
> Blocker — ModuleLayout and ApplicationDetailLayout now fight over the same layout slots AppRoutes.tsx:235-253 makes the module-level layout own viewMode, contextExpanded, contextSidebar and leading. ApplicationDetailLayout.tsx:152-174 already owns those same four slots, and it renders inside ModuleLayout's (AppRoutes.tsx:270). useLayoutConfig is a last-writer-wins setSlots, so the ancestor wins.
> 
> I reproduced both failure modes with a throwaway spec shaped like the real route tree (deleted after; not left in the tree):
> 
> * Deep link or reload on /applications/:id → sidebar owner came back module, expected application. React runs layout effects child-first, so ModuleLayout pushes last and the Environment section nav replaces the application's context sidebar and its ApplicationInfoHeader.
> * Clicking a tab inside an application → same result. ModuleLayout's dep array contains breadcrumbs, memoised on react-router's navigate, whose identity is keyed on locationPathname (react-router useNavigateUnstable). Every navigation re-fires the parent's effect, while the detail layout's deps (basePath, application, contextExpanded) stay stable — so the parent steals the slot back mid-session.
> 
> Secondary effect of the same overlap: there are now two contextExpanded states (AppRoutes.tsx:171 and ApplicationDetailLayout.tsx:149), and the header toggle drives whichever layout last pushed leading. I also confirmed that a child unmount resets slots the parent still owns (viewMode fell back to global); today that self-heals only because the parent's deps happen to change on navigation.
> 
> The apim module avoids this by keeping module-level config to navigation + breadcrumbs only (apim AppRoutes.tsx:142-148). Suggested fix in the same spirit: leave navigation (the primary column) in ModuleLayout, and move viewMode/contextExpanded/contextSidebar/leading into a layout element wrapping only the section pages — never an ancestor of ApplicationDetailLayout.
> 
> Worth noting breadcrumbs were already clobbered by this mechanism before the PR, so the ancestor/descendant conflict isn't new — but this change extends it to the sidebar, where it's plainly visible.
> 
> Improvement — the headline acceptance criterion is untested "Clicking a primary section opens that section's first visible item" has no test. firstNavItemKey is covered as a pure function, but the wiring in handleNavSelect (AppRoutes.tsx:212-228) isn't — the spec mocks ContextSidebar/ContextToggleButton to null and never invokes a select handler. It's cheap to cover: pull navigation.props.onItemSelect off the captured layout config, call it with 'team', and assert navigateToKey got 'users'. The already-active-section no-op branch (AppRoutes.tsx:216-218) is likewise unexercised.
> 
> Improvement — one callback serving two key namespaces handleNavSelect is handed both to the primary nav (section keys) and to ContextSidebar (nav item keys), and disambiguates by lookup. If a nav item key ever collides with 'organization' | 'environment' | 'team', a context-sidebar click silently becomes a section jump. Two small handlers would remove the coupling and make each one testable on its own.
> 
> Nits
> 
> * AppRoutes.tsx:246 — activeSection ? … : null guards a case that can't occur: access-management and security-plan-types fall through isNavItemVisible to return true, so Organization and Environment always survive filterNavSections.
> * navigation.ts:98-107 — the Team section's only group is also labelled Team, so the label repeats directly under the section header.
> * AppRoutes.tsx:216 — recomputing findNavSectionKey(...) instead of reusing activeSectionKey is deliberate (the ?? visibleNavSections[0] fallback would wrongly no-op), but nothing says so.

Thanks — the blocker was real. ModuleLayout was writing contextSidebar/leading/viewMode/contextExpanded as an ancestor of ApplicationDetailLayout, so the parent layout effect won and the Environment nav replaced the application sidebar.

Fix:
- ModuleLayout now only owns the primary column (`navigation`)
- PlatformSectionLayout (list/section pages only) owns viewMode, contextSidebar, leading, contextExpanded, and breadcrumbs
- Application detail is a sibling of that layout, not a child, so it keeps its own context sidebar

Also covered the two improvements:
- Tests for section click → first visible item, already-active no-op, and context-item navigation
- Split into handleSectionSelect / handleItemSelect so the two key namespaces are not mixed

Left the Team/Team group label as specified, and kept the activeSection null guard for the empty-filter case.

aqibmohammadkhan (commented):
- /applications/new has no test. RegisterApplicationPage is mocked at AppRoutes.spec.tsx:88 but never asserted. It works today (I verified), but it's now the one path whose correctness rests on route ranking across two subtrees — a one-line renderPlatform('/applications/new') would pin it.
- Collapse state resets on the detail round-trip. contextExpanded lives in PlatformSectionLayout (AppRoutes.tsx:298), which unmounts on application detail routes — collapse the column, open an application, come back, and it's expanded again. Lifting it to ModuleLayout and passing it through the nav context would make the preference stick.
- handleItemSelect is a pass-through over navigateToKey; navigateToKey could go into the context directly. Likewise breadcrumbs is computed in ModuleLayout but consumed only by PlatformSectionLayout — which could build them itself from the activeNavKey it already receives, shrinking the context to what is genuinely shared.
- layoutConfigs().find(...) in the spec helpers returns the first recorded config for a slot, where the old helper used .at(-1). Fine with the current synchronous mocks (one render pass), but findLast would keep it honest if a layout ever re-renders.
- Carried over and still open: activeSection ? … : null guards a case that can't occur (access-management and security-plan-types fall through isNavItemVisible to true, so Organization and Environment always survive filtering), and the Team section's only group is also labelled Team.

arpzilla-dev:
> * /applications/new has no test. RegisterApplicationPage is mocked at AppRoutes.spec.tsx:88 but never asserted. It works today (I verified), but it's now the one path whose correctness rests on route ranking across two subtrees — a one-line renderPlatform('/applications/new') would pin it.
> * Collapse state resets on the detail round-trip. contextExpanded lives in PlatformSectionLayout (AppRoutes.tsx:298), which unmounts on application detail routes — collapse the column, open an application, come back, and it's expanded again. Lifting it to ModuleLayout and passing it through the nav context would make the preference stick.
> * handleItemSelect is a pass-through over navigateToKey; navigateToKey could go into the context directly. Likewise breadcrumbs is computed in ModuleLayout but consumed only by PlatformSectionLayout — which could build them itself from the activeNavKey it already receives, shrinking the context to what is genuinely shared.
> * layoutConfigs().find(...) in the spec helpers returns the first recorded config for a slot, where the old helper used .at(-1). Fine with the current synchronous mocks (one render pass), but findLast would keep it honest if a layout ever re-renders.
> * Carried over and still open: activeSection ? … : null guards a case that can't occur (access-management and security-plan-types fall through isNavItemVisible to true, so Organization and Environment always survive filtering), and the Team section's only group is also labelled Team.

- Added a `/applications/new` test so ranking across the two subtrees is pinned
- Lifted `contextExpanded` to ModuleLayout so collapsing the column survives an application detail round-trip
- Dropped the `handleItemSelect` wrapper; PlatformSectionLayout now uses `navigateToKey` and builds its own breadcrumbs
- Spec helpers use `findLast`
- Dropped the `activeSection` null guard in favor of `activeSection?.groups ?? []`

arpzilla-dev:
working UI, after latest changes

image

aqibmohammadkhan (approved):
lgtm

PR #19130, opened 2026-08-13
feat(gamma): list, create, and delete shared policy groups

ankita-gupta21:
## Summary

Adds the **Policy Groups** screen to Platform Management: a list page where API Platform Engineers can view, search, create, and delete Shared Policy Groups for the selected environment.

Ports the logic and API calls from console-webui's `shared-policy-groups` feature so behaviour matches classic Console.

### [FOUND-17](https://gravitee.atlassian.net/browse/FOUND-17) — List Shared Policy Groups ✅ done
- [x] List scoped to the selected environment
- [x] Table shows name + description, status, API type, phase, last updated, last deployed
- [x] Search by name/description
- [x] Read-only users see the list with no write actions
- [x] Empty state shows a Create CTA
- [x] Row click opens the detail view
- [x] READ permission required to view the page
- [x] Column sorting (name, phase, last updated, last deployed)
- [ ] Policy count / APIs-using count — not shown; these fields don't exist on the API response (classic Console doesn't show them either) -updated AC

### [FOUND-18](https://gravitee.atlassian.net/browse/FOUND-18) — Create Shared Policy Group ⚠️ partial
- [x] Create form: name (required), description, prerequisite message, API type, phase
- [x] Submit disabled until name is entered
- [x] Success toast
- [x] CREATE permission required
- [ ] Doesn't open Policy Studio after creation — Policy Studio isn't built yet, so it redirects to the new group's detail page instead - future PR

### [FOUND-22](https://gravitee.atlassian.net/browse/FOUND-22) — Delete Shared Policy Group ⚠️ partial
- [x] Confirm dialog before deleting
- [x] Success toast
- [x] DELETE permission required
- [ ] No client-side check for "group is deployed / in use" — relies on the backend rejecting the request; error shows as a toast - Future PR

### Not in this PR
- **FOUND-19** (Configure Policies / Policy Studio)
- **FOUND-20** (Deploy / Undeploy / Version History)
- **FOUND-21** (Import via CRD)

These need Policy Studio and gateway sync work that hasn't landed yet, so they're separate follow-up PRs.

## Screen

**Policy Groups list** (`Platform Management → Policy Groups`)
- Empty state shows a Create CTA  Screenshot 2026-08-13 at 8 00 21 PM
- Table: Name, Status, API Type, Phase, Last updated, Last deployed 
- Row menu (⋯): View/Edit, Delete
- "Add Policy Group" opens a form for Proxy or Message API groups https://github.com/user-attachments/assets/cdf6ebdb-87b7-4be9-b6aa-7e1ac1903897
- Search box and column sorting

https://github.com/user-attachments/assets/b3070c3a-6e69-4c24-9b9c-4962688e7c0e

[FOUND-17]: https://gravitee.atlassian.net/browse/FOUND-17?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FOUND-18]: https://gravitee.atlassian.net/browse/FOUND-18?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupsTable.tsx
```diff
@@ -0,0 +1,272 @@
...
+                            <DropdownMenuContent align="end">
+                                {readOnlyRow ? (
+                                    <DropdownMenuItem onSelect={() => onView(sharedPolicyGroup)}>
+                                        <EyeIcon className="size-4 mr-2" aria-hidden />
+                                        View
+                                    </DropdownMenuItem>
+                                ) : (
+                                    <DropdownMenuItem onSelect={() => onView(sharedPolicyGroup)}>
+                                        <PencilIcon className="size-4 mr-2" aria-hidden />
+                                        Edit
```
arpzilla-dev: **[Score: 71/100]** ⚠️ `Edit` navigates to a read-only details page.

With `canEdit`, the row menu labels the action **Edit**, but `onView` only opens `SharedPolicyGroupDetailPage`, which has no fields to change and no Policy Studio. Classic Console’s Edit actually opens the studio; here the label promises an editor that this PR does not ship.

Until that destination exists, show **View** for every row (keep the Kubernetes / permission gating for Delete). Otherwise operators with UPDATE permission click Edit and land on a metadata card they cannot change.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/SharedPolicyGroupsPage.tsx
```diff
@@ -0,0 +1,210 @@
...
+    return (
+        <div className="space-y-6">
+            <div className="flex items-start justify-between gap-4">
+                <div className="space-y-1">
+                    <h1 className="text-2xl font-semibold tracking-tight">Shared Policy Groups</h1>
+                    <p className="text-sm text-muted-foreground">
+                        Shared Policy Groups let you apply reusable policies to your API event phases.
+                    </p>
+                </div>
+                {canCreate && !isFirstUse ? (
```
arpzilla-dev: **[Score: 58/100]** 💡 Header CTA flashes on first-use empty environments.

`isFirstUse` is `!isLoading && totalCount === 0 && …`. While the first fetch is in flight, `isFirstUse` is false, so this `Add Shared Policy Group` button renders next to a skeleton table, then disappears when the empty collection arrives and `DataTableEmptyState` owns the CTA.

Graphene’s list-page rule is that the header button appears only when data exists (or filters are active). Gate it on loaded data, for example `canCreate && !isLoading && !isFirstUse`, so the header and empty-state CTAs never compete.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupCreateSheet.tsx
```diff
@@ -0,0 +1,217 @@
...
+                                    disabled={isSaving}
+                                />
+                            </Field>
+
+                            <Separator />
+
+                            <h3 className="text-sm font-semibold">Scope</h3>
+
+                            <Field orientation="vertical" className="gap-1.5">
+                                <FieldLabel>
```
arpzilla-dev: **[Score: 56/100]** 💡 API Type and Phase labels are not associated with the controls.

Name / description / prerequisite use `FieldLabel htmlFor="…"`. These two required `ToggleGroup`s have a `FieldLabel` with no `htmlFor` and no `aria-labelledby` on the group, so the accessible name of each radio is only `"Proxy"` / `"Request"` — not that they are the required API Type and Phase fields.

Give each group an id (or label id) and wire it: `FieldLabel htmlFor="spg-api-type"` plus `ToggleGroup id="spg-api-type"` / `aria-labelledby`, same for Phase. Mirror `CreateDictionarySheet`, which binds the Type `SelectTrigger` with `htmlFor` + `id`.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupsTable.tsx
```diff
@@ -0,0 +1,272 @@
...
+            size: 100,
+            enableSorting: false,
+            enableHiding: false,
+            cell: ({ row }: ColCell<SharedPolicyGroup>) => {
+                const sharedPolicyGroup = row.original;
+                const readOnlyRow = !canEdit || isKubernetesOrigin(sharedPolicyGroup);
+                const showDelete = canDelete && !isKubernetesOrigin(sharedPolicyGroup);
+                return (
+                    <div className="flex items-center justify-end gap-1">
+                        {isKubernetesOrigin(sharedPolicyGroup) && <KubernetesIcon className="size-4 text-muted-foreground" aria-hidden />}
```
arpzilla-dev: **[Score: 52/100]** 💡 Kubernetes origin is a mystery icon.

The glyph is `aria-hidden` with no tooltip, so sighted users get an unexplained mark and assistive tech gets nothing. The row is already treated as read-only (`View` only, Delete hidden); that meaning is not announced.

Wrap it in Graphene `Tooltip` (same pattern as the Pending status badge) with copy like **Managed by Kubernetes**, and put an accessible name on the trigger. Do not leave a brand icon as the only signal.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupCreateSheet.tsx
```diff
@@ -0,0 +1,217 @@
...
+        setForm(prev => ({ ...prev, [key]: value }));
+    }
+
+    function handleApiTypeChange(apiType: ApiType) {
+        setForm(prev => ({ ...prev, apiType, phase: PHASE_BY_API_TYPE[apiType][0] }));
+    }
+
+    const isValid = form.name.trim() !== '';
+
+    const [, submitAction, isSaving] = useActionState<void, FormData>(async () => {
```
arpzilla-dev: **[Score: 48/100]** 💡 `useActionState` is a new form pattern for this module, and it forced a global `FormData` test stub.

Sibling create sheets (`GroupSheet`, `TenantFormSheet`, `CreateDictionarySheet`) submit with a normal `onSubmit` / `FormEvent` and drive pending UI from local or mutation state. This sheet is the only `useActionState` consumer, which is why `shared/testing/formAction.ts` has to replace `globalThis.FormData` in jsdom.

Prefer the existing sheet submit pattern: `onSubmit` on the `<form>`, `isSaving` from the parent mutation (`createMutation.isPending`) or a local flag. That matches the module, drops the polyfill, and still prevents double-submit.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupsTable.tsx
```diff
@@ -0,0 +1,272 @@
...
+                    }
+                />
+            }
+            toolbar={
+                <div className="w-64">
+                    <InputGroup>
+                        <InputGroupAddon align="inline-start">
+                            <SearchIcon className="size-3.5 text-muted-foreground" aria-hidden />
+                        </InputGroupAddon>
+                        <InputGroupInput
```
arpzilla-dev, line 265: **[Score: 45/100]** 💡 Search field has no accessible name.

The control is only a placeholder (`Search by name or description…`). Graphene’s accessibility rules want a name via `Label` / `htmlFor` or `aria-label`; `TenantsTable` already uses `aria-label="Search tenants"` on the same `InputGroupInput` pattern.

Add `aria-label="Search Shared Policy Groups"` (or a visually hidden label). Placeholder text is not a reliable accessible name.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupCreateSheet.tsx
```diff
@@ -0,0 +1,217 @@
...
+            return;
+        }
+        await onSubmit({ ...form, name: form.name.trim() });
+    }, undefined);
+
+    return (
+        <Sheet open={open} onOpenChange={isOpen => !isOpen && onClose()}>
+            <SheetContent side="right" className="flex max-h-full flex-col" style={{ maxWidth: STANDARD_SHEET_WIDTH }}>
+                <SheetHeader>
+                    <SheetTitle>Add Policy Group</SheetTitle>
```
arpzilla-dev: **[Score: 28/100]** 🎨 Sheet title does not match the page CTA.

The list header / empty-state button is **Add Shared Policy Group**; this sheet title is **Add Policy Group**. Classic Console uses the full name on the primary action. Pick one string and use it on the button, empty state, and sheet so the create flow reads as the same action.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupsTable.tsx
ankita-gupta21: Agreed, the label promised an editor this PR doesn't ship. The row menu now always shows View, and canEdit was dropped from the table and page entirely (the UPDATE permission lookup is gone until the studio lands). Delete keeps its permission and Kubernetes gating

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/SharedPolicyGroupsPage.tsx
ankita-gupta21: Gated on loaded data: canCreate && !isLoading && !isFirstUse, so the header button never renders next to the skeleton table and never competes with the empty-state CTA. Covered by a new test asserting the button is absent while the first page loads.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupCreateSheet.tsx
ankita-gupta21: Each FieldLabel now has an id (spg-api-type-label, spg-phase-label) and the corresponding ToggleGroup sets aria-labelledby, so both groups expose their field name. Asserted via getByRole('group', { name: /API Type/ }) and the same for Phase.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupsTable.tsx
ankita-gupta21: The glyph is now wrapped in a Graphene Tooltip (same pattern as the Pending status badge) with the copy Managed by Kubernetes, and the trigger carries that accessible name. New test hovers the trigger and asserts the tooltip.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupCreateSheet.tsx
ankita-gupta21: Switched to the module's existing pattern: onSubmit on the <form> with event.preventDefault(), and pending UI driven by an isSaving prop fed from createMutation.isPending. Double-submit is still prevented (guarded in the handler and the button stays disabled). shared/testing/formAction.ts is deleted and both specs no longer install the polyfill

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupsTable.tsx
ankita-gupta21: Added aria-label="Search Shared Policy Groups" on the InputGroupInput, matching TenantsTable, with a test asserting getByRole('textbox', { name: ... }).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/shared-policy-groups/components/SharedPolicyGroupCreateSheet.tsx
ankita-gupta21: Standardised on Add Shared Policy Group across the header button, empty-state CTA, and sheet title.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/SharedPolicyGroupsPage.tsx
```diff
@@ -0,0 +1,210 @@
...
+    ENVIRONMENT_SHARED_POLICY_GROUP_UPDATE_PERMISSION,
+} from '../features/shared-policy-groups/utils/sharedPolicyGroupPermissions';
+import { ConfirmDialog } from '../shared/components/ConfirmDialog';
+import { useForbiddenResourceRedirect } from '../shared/hooks/useForbiddenResourceRedirect';
+import { notify } from '../shared/notify';
+import { isForbiddenApiError } from '../shared/utils/apiErrors';
+
+type SheetState = { type: 'closed' } | { type: 'create' } | { type: 'delete'; sharedPolicyGroup: SharedPolicyGroup };
+
+export function SharedPolicyGroupsPage() {
```
ankita-gupta21, line 51: ignore: SharedPolicyGroupsPage is a route-level orchestration component, which is the established shape for list pages in this module (see DictionariesPage, GroupsPage). Its length is data wiring — query, mutations, permissions, sheet state — not branching logic; splitting it into pass-through hooks would spread one flow across more files without reducing coupling or behaviour.

arpzilla-dev (approved):

PR #19132, opened 2026-08-13
feat(gateway): diagnose connection closures, unexpected 500s and shutdown cut-offs

phiz71:
## Issue
https://gravitee.atlassian.net/browse/APIM-14952

No JIRA ticket. This comes out of a customer investigation on `GATEWAY_CLIENT_CONNECTION_CLOSED`
at scale on streaming APIs, where the reporting could not distinguish three genuinely different
situations, all surfacing under the same key — or under no key at all:

1. the backend really cut the response body short;
2. **the gateway closed the connection itself**, on its own `idleTimeout`, because `readTimeout`
   was configured longer than `idleTimeout` and could therefore never fire;
3. the reactor was shutting down and cut the request short.

Reproduced 11/11 on a bench. The customer's actual root cause is (2), and nothing in the
reporting pointed at it: the failure reads as a backend disconnection.

## Description

Three commits, reviewable in order — each stands on its own.

### 1. `feat` — report a stream cut short under its own key, with timeout diagnostics

*(`gravitee-apim-plugin-endpoint-http-proxy`)*

`HttpConnector.recordBackendResponseStreamFailure` runs when the upstream fails while the
response body is being streamed, i.e. after status and headers were committed downstream.

- **New key** `GATEWAY_CLIENT_STREAM_ENDED_EARLY`, substituted for `CONNECTION_CLOSED` **in this
  phase only**. The caller received a valid response with a truncated body; keying it like a
  connection that never delivered anything is what makes streaming APIs look like they are
  failing. `CONNECTION_CLOSED` is untouched everywhere else.
- **A message meant to be read alone**, by someone holding the analytics entry and not the
  gateway configuration: elapsed time, upstream silence, and — when that silence matches the
  configured `idleTimeout` — that the gateway is the likely closer, naming the misconfiguration.
- **A single WARN log** replacing the generic ERROR at the call site, carrying target, both
  durations, bytes sent, status, the three timeouts, the key and the stack trace. Now emitted
  even when an error was already recorded, so a truncated body is never silent.

**The part deserving the most attention — what is measured.** The detection compares the
**upstream silence** to the effective `idleTimeout`, not the elapsed time of the exchange:

- `idleTimeout` is a channel-level timer that **restarts on every byte received**. On a stream
  that trickles for ten minutes then goes quiet, only the trailing silence is comparable to it.
- The elapsed time additionally includes everything upstream of the backend call (policies,
  connection acquisition, first byte), so it is systematically larger by an unknown amount.

An earlier iteration compared the elapsed time; a simulation showed the detection would
essentially never fire in production. The silence is tracked with an `AtomicLong` initialised at
the response headers and updated in `doOnNext` on the chunk flow.

`effectiveIdleTimeoutMillis()` truncates to the second, because Vert.x applies `idleTimeout` with
whole-second resolution — `12_200 ms` fires at `12_000 ms`.

### 2. `fix` — make the reactor's last-resort 500 diagnosable

*(`gateway-core` + both reactors)*

`handleUnexpectedError` answers a bare `500` for any error reaching the end of the chain
unconverted. It logged `"Unexpected error while handling request"` and nothing else, while
overwriting the status the request had reached.

The observable result: a `500` carrying an error key belonging to a different status (a
`GATEWAY_CLIENT_*` key normally answered with `502`), with an empty body because the error
handling never got to write one — indistinguishable, from analytics alone, from the gateway
simply returning `500`.

New `UnexpectedErrorReporter`, called by both reactors **before** the status is overwritten:

- **In the log**: `requestId` and `transactionId` — the contextual logger injects `apiId`,
  `apiName`, `envId`, `orgId`, `appId`, `planId` and `user`, but neither of those two, and they
  are the only link to the analytics entry the line explains — plus the status being replaced,
  the attributed error, the endpoint, the elapsed time, and the most specific cause in plain text
  (stack traces get truncated by log collectors).
- **In the analytics**, when nothing was attributed: a `GATEWAY_UNEXPECTED_ERROR` key. Such a
  request previously carried **no error key at all** — a `500` reading as a success in any
  dashboard counting failures by key.
- **In the analytics**, when something was already attributed: preserve it (it names what
  actually went wrong) and record the unexpected exception as a **warning** alongside, which
  otherwise exists nowhere but the pod log file.

### 3. `feat` — log the requests an api reactor cuts short when it stops

*(both reactors)*

`stopNow()` closes endpoint connections, policies and resources from under requests still
running. Nothing tied the resulting error spike to the redeploy that caused it, and durations
clustering on `api.pending_requests_timeout` (default 10 000 ms) read as a traffic timeout rather
than the shutdown countdown they are.

One WARN when the reactor tears down with requests in flight, naming how many are cut short and
how long they were waited for. Silent on a clean shutdown — warning there would train operators
to ignore the line.

**Where the count is read matters.** It is captured inside the wait loop, where the decision to
stop anyway is taken, and passed to the log. The first version read `pendingRequests` from the
teardown itself and, on a live gateway, reported *nothing at all* while requests were demonstrably
being cut short: the counter was already back to zero by then, even though the wait loop had just
spent the full grace period precisely because it was not. So the two stop paths capture at
different points — from the wait loop for a graceful stop, and directly in `doStop` for a node
already going down, which cuts everything without waiting. `stopNow()` no longer logs: it serves
both paths and no longer holds the information when it runs.

What decrements the counter in between was not established — the thread hop of
`observeOn(Schedulers.io())` is the plausible suspect. The fix makes the measurement immune to the
race rather than removing it.

### Breaking change to flag in the release notes

`GATEWAY_CLIENT_CONNECTION_CLOSED` → `GATEWAY_CLIENT_STREAM_ENDED_EARLY` for the truncated-body
case. Dashboards filtering the old key will report zero for it. No impact on response templates:
the response is already committed in this phase, so no template applies.

### Design decision worth confirming — no `ExecutionWarn` in commit 1, one in commit 2

The obvious move in commit 1 was `ctx.warnWith(...)`: a truncated body on a `200` is arguably a
warning, not a failure. It was tried and removed, because:

- `ReporterProcessor.translateErrorToDiagnosticFailure` already turns `errorKey`/`errorMessage`
  into the `Diagnostic` carried by the metrics, attributed to `ENDPOINT` via the
  `GATEWAY_CLIENT_` prefix. `warnWith` would emit a **second** diagnostic for the same event.
- That second one would be worse: the `ENDPOINT` component scope is pushed only around
  `invoker.invoke(ctx)`, which completes when the response head arrives — so anything raised
  while the body streams comes out as `Unknown component`.
- `DiagnosticReportHelper` appends the cause to the message, which our message already contains.

In commit 2 it is the right tool for the opposite reason: there is no duplicate, it is the only
trace, and it does not replace the attributed diagnosis.

### And why not a warning *instead* of a failure, for the truncated body?

The semantic argument is real: the caller received a `200` with headers and a body, so the request
did not fail, and a `failure` on a 2xx is odd. It would also match the three existing
`ExecutionWarn` uses (`BAD_REGEX_CONTENT_TYPE`, `EXPRESSION_EVALUATION_ERROR`, `INVALID_XML`) —
anomalies that do not prevent the request from completing.

It was rejected because of what `warnings` costs in visibility. The field is mapped `nested`, so
Kibana does not list it as an ordinary field: it cannot be added as a column through the field
list, filtered in KQL, or aggregated, until runtime fields are declared on the data view. Moving
this case to warnings only would not reclassify it — it would remove it from every dashboard that
watches it today, for a case operators are actively monitoring, and where the truncated body means
the consumer walked away with incomplete data without knowing.

So the failure stays, and the distinction lives in the key instead of the field:
`GATEWAY_CLIENT_STREAM_ENDED_EARLY` lets a dashboard exclude these from its error rate, which was
impossible while they were keyed `CONNECTION_CLOSED`. Moving them to `warnings` becomes the right
call once warnings are usable from analytics out of the box — a product change, not a maintenance
fix.

### Open questions for review

- **`GATEWAY_UNEXPECTED_ERROR` as a key name.** It has no parent key, and no response template
  can target it usefully since this path bypasses templates entirely.
- **Scope: both reactors, v4 and sync.** The sync one is not needed for the customer case, but
  the defect is identical and leaving it out would be arbitrary.
- **`gateway-core` on a maintenance branch** — commit 2 is the most central change of the three.
- **Duplication in commit 3** between `AbstractApiReactor` and `SyncApiReactor`, which share no
  hierarchy. Consistent with how `stopUntil`, `stopNow` and `handleUnexpectedError` are already
  duplicated across the two; extracting a helper for a log line felt like more ceremony than
  value, but happy to change it.

### Deliberately out of scope

- **`GATEWAY_CLIENT_CONNECTION_POOL_EXHAUSTED` returning 502 instead of 503.** It is the only key
  whose status differs from its parent's, and `handleTemplate` applies the template's status
  unconditionally — so a template configured on `GATEWAY_CLIENT_CONNECTION_ERROR`, or a `DEFAULT`
  one, silently downgrades a deliberate 503 into a 502. A narrow fix exists (keep the failure's
  status when the template was reached through the parent fallback rather than an explicit
  match), and it would be a no-op for every other key since they are all status-aligned with
  their parent. It is a behaviour change deserving its own decision.
- **A deployment-time warning when `readTimeout >= idleTimeout`.** That configuration can never
  fire its read timeout; catching it at deploy time would beat diagnosing it in production.
  Belongs to configuration validation.

### Building and testing

Commit 2 changes classes in `gateway-core` that `handlers-api` consumes as an installed artefact.
Run `mvn install -DskipTests -pl gravitee-apim-gateway/gravitee-apim-gateway-core` before testing
`handlers-api` alone, or the tests fail on `NoClassDefFoundError`.

Tests: 104 green on the plugin module, 1 415 on the two gateway modules. No test was weakened or
removed; two existing assertions were updated where the message they asserted changed. Unit tests
cover the two cases the detection must **not** claim (silence far from `idleTimeout`,
`readTimeout` already below `idleTimeout`) and the invariant that broke on the shutdown path.

Commits 1 and 3 were also exercised on a running gateway against a bench backend: an endpoint
sending headers then going silent on an API configured with `readTimeout > idleTimeout` (cut at
12.01 s, reported under the new key, message naming the `idleTimeout` as the likely closer); the
same failure shape with sane timeouts (reported without any claim about the configuration); and a
long-running request while its API is redeployed (`WARN` emitted, naming the single request cut
short, a little over the 10 s grace period).

Commit 2 was exercised too, by deploying an API whose response template contains an expression
that throws (`{#error.thisMethodDoesNotExist()}`). The backend is then cut on its `idleTimeout`,
the failure is classified as a `502`, the template cannot be rendered, and the error reaches the
end of the chain unconverted. Result:

```
status 500, response-content-length 0, error-key GATEWAY_CLIENT_CONNECTION_CLOSED
warnings[0].key     GATEWAY_UNEXPECTED_ERROR
warnings[0].message The gateway could not complete the handling of this request and answered
                    500, replacing status 502 (Expression: #error.thisMethodDoesNotExist()
                    failed with message EL1004E: Method call: Method thisMethodDoesNotExist()
                    cannot be found on type EvaluableExecutionFailure)
```

That is the exact signature this commit set out to make diagnosable — a bare 500, an empty body,
and a key belonging to another status — now carrying the reason. The log line reports
`statusBeingReplaced=502` alongside the request and transaction ids.

Worth knowing when reviewing the analytics side: `warnings` is mapped `nested`, so it is not
listed as an ordinary field in Kibana and needs a runtime field to be filtered or aggregated. The
data is complete; the query path differs.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/diagnostic/UnexpectedErrorReporter.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/diagnostic/UnexpectedErrorReporter.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gateway.reactive.core.context.diagnostic;
+
+import io.gravitee.gateway.reactive.api.ExecutionWarn;
+import io.gravitee.gateway.reactive.api.context.http.HttpPlainExecutionContext;
+import io.gravitee.reporter.api.v4.metric.Diagnostic;
+import io.gravitee.reporter.api.v4.metric.Metrics;
+import org.slf4j.Logger;
+import org.springframework.core.NestedExceptionUtils;
+
+/**
+ * Reports the last-resort failure path of an API reactor: an error that reached the end of the chain without having
+ * been turned into a proper execution failure, and which the reactor answers with a bare 500.
+ *
+ * <p>Left alone, that path is close to undiagnosable. It logs "Unexpected error while handling request" and nothing
+ * else, while overwriting whatever status the request had reached — so the reported outcome can be a 500 carrying an
+ * error key that belongs to a completely different status (a {@code GATEWAY_CLIENT_*} key normally answered with a
+ * 502, for instance), with an empty body because the error handling itself never got to write one. The combination is
+ * indistinguishable, from analytics alone, from the gateway simply returning 500.
+ *
+ * <p>This records what is needed to tell those apart, on both sides:
+ * <ul>
+ *   <li>in the gateway log, a line naming the request and transaction ids, the status being replaced, the error
+ *       already attributed, the endpoint reached and the cause — enough to conclude without correlating anything;</li>
+ *   <li>in the analytics, a key of its own when nothing was attributed, so the request stops being reported as a
+ *       success that happens to carry a 500 — and otherwise a warning carrying the exception, which would otherwise
+ *       exist nowhere but the gateway's own log file.</li>
+ * </ul>
+ */
+public final class UnexpectedErrorReporter {
+
+    /** Key recorded when the failure reached the reactor with nothing attributed to it yet. */
+    public static final String UNEXPECTED_ERROR_KEY = "GATEWAY_UNEXPECTED_ERROR";
+
+    private UnexpectedErrorReporter() {}
+
+    /**
+     * Reports the failure <em>and</em> answers it with the given status, in that order.
+     * <p>
+     * The two are done here together on purpose: every message below states the status the request is answered with,
+     * and the status the request had reached before that. Leaving the caller to apply the status separately would let
+     * the two drift apart, and the reports would then describe a response that was never sent.
+     *
+     * @param statusCode the status to answer with; the caller still decides which one, it is simply written once.
+     */
+    public static void recordAndAnswer(
+        final HttpPlainExecutionContext ctx,
+        final int statusCode,
+        final String reasonPhrase,
+        final Throwable throwable,
+        final Logger log
+    ) {
+        final Metrics metrics = ctx.metrics();
+        final Diagnostic failure = metrics != null ? metrics.getFailure() : null;
+        final String attributedKey = attributedKey(metrics, failure);
+        // Read before the answer below overwrites it.
+        final int replacedStatus = ctx.response().status();
+
+        // The contextual logger already carries apiId, apiName, envId, orgId, appId, planId and user, but neither
+        // request nor transaction id — and those are what tie this line to the analytics entry it explains. The rest
+        // is what one needs to decide whether the answered status is the real outcome or an overwrite: the status it
+        // replaces, the error already attributed, where the request had got to, and the cause spelled out for the
+        // cases where the stack trace is truncated by whatever collects these logs.
+        ctx
+            .withLogger(log)
+            .error(
+                "Unexpected error while handling request, answering {} [requestId={}, transactionId={}, " +
+                    "statusBeingReplaced={}, attributedErrorKey={}, attributedFailure={}, endpoint={}, elapsed={}ms]: {}",
+                statusCode,
+                metrics != null ? metrics.getRequestId() : null,
+                metrics != null ? metrics.getTransactionId() : null,
+                replacedStatus,
+                attributedKey,
+                failure != null,
+                metrics != null ? metrics.getEndpoint() : null,
+                elapsedMillis(metrics),
+                mostSpecificCause(throwable),
+                throwable
+            );
+
+        if (metrics != null) {
+            if (attributedKey == null) {
+                // Nothing was attributed: without a key of its own the request is reported as a success that happens
+                // to carry an error status.
+                metrics.setErrorKey(UNEXPECTED_ERROR_KEY);
+                metrics.setErrorMessage(
+                    "The request failed unexpectedly while being handled by the gateway, and was answered with a " +
+                        statusCode +
+                        " (" +
+                        mostSpecificCause(throwable) +
+                        ")"
+                );
+            } else {
+                // Something was already attributed, and it is the more precise diagnosis: it names what actually went
+                // wrong, whereas reaching this path only says the gateway could not handle the consequence. Keep it,
+                // and record the exception as a warning — otherwise it exists nowhere but the gateway's log file.
+                ctx.warnWith(
+                    new ExecutionWarn(UNEXPECTED_ERROR_KEY)
+                        .message(
+                            "The gateway could not complete the handling of this request and answered " +
+                                statusCode +
+                                ", replacing status " +
+                                replacedStatus
+                        )
+                        .cause(throwable)
+                );
+            }
+        }
+
+        ctx.response().status(statusCode);
+        ctx.response().reason(reasonPhrase);
+    }
+
+    /** How long the request had been running, so a recurring duration (a shutdown countdown) stands out. */
+    private static long elapsedMillis(final Metrics metrics) {
+        if (metrics == null || metrics.timestamp() == null) {
+            return -1L;
+        }
+        return System.currentTimeMillis() - metrics.timestamp().toEpochMilli();
+    }
+
+    /**
+     * Description of the failure to carry into analytics: the message of the most specific cause, which is the one
+     * naming the actual fault, or its type when it carries no message.
+     */
+    private static String mostSpecificCause(final Throwable throwable) {
+        final Throwable cause = NestedExceptionUtils.getMostSpecificCause(throwable);
+        return cause.getMessage() != null && !cause.getMessage().isBlank() ? cause.getMessage() : cause.getClass().getSimpleName();
+    }
+
+    /**
+     * The error already attributed to the request, wherever it was recorded: connectors set {@code errorKey} on the
+     * metrics, while {@code interruptWith} records a {@link Diagnostic} failure without touching it.
+     */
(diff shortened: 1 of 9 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-core/src/main/java/io/gravitee/gateway/reactive/core/context/diagnostic/UnexpectedErrorReporter.java
```diff
@@ -0,0 +1,130 @@
...
+ *   <li>in the analytics, a key of its own when nothing was attributed, so the request stops being reported as a
+ *       success that happens to carry a 500 — and otherwise a warning carrying the exception, which would otherwise
+ *       exist nowhere but the gateway's own log file.</li>
+ * </ul>
+ */
+public final class UnexpectedErrorReporter {
+
+    /** Key recorded when the failure reached the reactor with nothing attributed to it yet. */
+    public static final String UNEXPECTED_ERROR_KEY = "GATEWAY_UNEXPECTED_ERROR";
+    private static final String UNEXPECTED_ERROR_MESSAGE =
```
jhaeyaert: The UnexpectedErrorReporter assumes an error 500, while there is no hard relation with what status the DefaultApiReactor decides to assign (L44 - ctx.response().status(HttpResponseStatus.INTERNAL_SERVER_ERROR.code());)

I believe we should have a stronger relationship between the two

phiz71: You're right, and it is worth fixing: the reporter states `answering 500` while the reactor decides the status two lines below, and nothing keeps the two in sync.

Three ways to get there:

| | Approach | Trade-off |
|---|---|---|
| **A** | The reporter applies the status it is handed — `recordAndAnswer(ctx, throwable, log, INTERNAL_SERVER_ERROR)`. The reactor still decides which status, it is simply written in one place. | Makes the relationship structural, and removes the `status` + `reason` duplication that already exists between `DefaultApiReactor` and `SyncApiReactor`. In exchange, a class living under `core.context.diagnostic` starts mutating the response — the naming would have to follow (`recordAndAnswer`, or renaming the class to `UnexpectedErrorHandler`). Reading `handleUnexpectedError` no longer tells you which status goes out without opening the helper. |
| **B** | The status is passed as a parameter but still applied by the reactor. | Smallest diff, but the link stays conventional — nothing prevents the two from drifting again. Doesn't really answer your point. |
| **C** | Drop the applied status from the messages altogether, keep only `statusBeingReplaced`. | Removes the problem instead of managing it, with no new coupling. The final status is already carried by the `status` field of the metrics, so repeating it in the message is redundant; the status being *replaced* is the part that exists nowhere else. Costs the at-a-glance readability of `answering 500` in the log line. |

I'd go with **A**. The measured impact is nil — the reactor tests assert `verify(response).status(...)` on the mock, so they pass unchanged whoever makes the call, including the `never()` ones — and it removes a duplication that is already there today, independently of this PR.

That said, **C** is defensible if you'd rather keep anything under `diagnostic` strictly passive. Happy to go that way instead — your call, I'll implement whichever you prefer.

jhaeyaert: A) looks ok

phiz71: Went with **A**, pushed as `7e56170a93` (a `fixup!` on top rather than an amend, so the delta since your review stays visible — to be autosquashed before merge).

`record(...)` is now `recordAndAnswer(ctx, statusCode, reasonPhrase, throwable, log)`: it reads the current status, emits the log and the metrics, then applies the status and the reason. Both reactors lose their duplicated `status` + `reason` lines and just pass `INTERNAL_SERVER_ERROR`. The status no longer appears hardcoded anywhere — the three messages that mentioned it now interpolate the parameter, so they cannot describe a response that was never sent.

One deliberate detail: the signature takes `(int, String)` rather than `HttpResponseStatus`. Netty is only transitively available in `gateway-core` (through `vertx-circuit-breaker`) and the module has no other use for it, so importing it there felt like the wrong trade in a PR that is already about responsibilities. Slightly more verbose at the call site, no new dependency.

Two tests pin the behaviour you were after: one asserts the reported status is the applied one (`inOrder` on `status` then `reason`), the other that the *replaced* status stays distinct from the *answered* one — 502 in the message, 500 on the wire.

Also took the Aikido suggestion on the nested ternary in `attributedKey`, and documented why there are two sources at all: connectors set `errorKey` on the metrics, `interruptWith` records a `Diagnostic` failure without touching it.

No existing assertion needed rewriting — the reactor tests verify `verify(response).status(...)` on the mock, so they pass unchanged. 61 green on the reactors, 1003 on `handlers-api`, 415 on `gateway-core`.

jhaeyaert (approved):

PR #19141, opened 2026-08-13
feat(e2e): introduce Playwright UI tests with updated configuration a…

aqibmohammadkhan:
…nd workflows

Integrate Playwright for UI testing with new fixtures, pages, and tests. Update CircleCI workflows to include Playwright job execution, refine NGINX configuration, and add Docker Compose support for Playwright tests.

## Issue

https://gravitee.atlassian.net/browse/APIM-14926

## Description

**Title:** `[Playwright migration] CI wiring + first two converted specs`

---

## What

**CI** — new `job-e2e-playwright.ts` mirroring `job-e2e-cypress.ts`: attaches the workspace so the SDK clients generated by `Generate e2e tests SDK` are present (they're gitignored, so a checkout alone leaves the suite uncompilable), and adds `store_test_results` on the JUnit XML, which the Cypress job doesn't do. Registered in both `workflow-pull-requests.ts` and `workflow-run-e2e-tests.ts`, with the four snapshot fixtures regenerated.

**Docker** — `docker-compose-playwright-tests.yml` (`mcr.microsoft.com/playwright:v1.62.1-jammy`, waits on healthy nginx), a `playwright-test` mode in `run-docker-compose.sh` with `--exit-code-from playwright` plus log capture, and a `yarn test:pw` entrypoint.

**Suite scaffolding** — `pages/base.page.ts` plus login/home/gateways page objects (locators as getters, no assertions), `fixtures/base.fixture.ts` as the single import point for specs, `utils/jira.ts` (`linkJira`, same annotation shape as AM and Cockpit), and a second `api-publisher` persona in `global.setup.ts`.

**Specs** — 6 tests across `tests/login/login.spec.ts` and `tests/gateways/gateway-information.spec.ts`, ported from `ui-login.spec.ts` and `ui-gateway-information.spec.ts` (8 including the two setup projects).

24 files, +821/−18.

## Three bugs this surfaced

**1. `nginx.conf` — pre-existing, and it blocked everything.** `sub_filter` cannot rewrite a compressed body, and browsers always request gzip. The `<base href="/">` → `/console/` rewrite therefore never applied, every relative asset resolved against the origin root, and the containerized Console rendered blank. `curl` hid it by not requesting gzip. Fixed with `proxy_set_header Accept-Encoding "";`.

**This one deserves a second opinion.** The Console image sets `gzip on`, and nginx always gzips `text/html` regardless of `gzip_types` — so the branch-built CI image behaves the same way. Combined with the branch gating below, this would have gone unnoticed on ordinary PRs. I could not confirm whether `Run Cypress UI tests` is currently red on master (CircleCI reports a single aggregate check to GitHub rather than per-job status). If it is green there, my reasoning is incomplete and this change should be scrutinised before merge.

**2. Leading-slash `goto()` paths.** Playwright resolves `page.goto(path)` as `new URL(path, baseURL)`, so `goto('/')` against `http://nginx/console/` lands on the Portal at the origin root. Base URL is now normalised to a trailing slash and page objects pass relative paths.

**3. Yarn 1 vs `packageManager: yarn@4`.** The Playwright image ships Yarn 1, which refuses to run the project — the container exited silently with no output. Fixed with `corepack enable` so it runs the same script as everyone else.

All three are documented in the suite README.

## Deviations from the source Cypress specs

- The non-admin gateway test now asserts the **redirect** (`not.toHaveURL(/gateways/)`) before the absence check. Navigation puts `/gateways` in the URL, so it can only pass once the guard actually fires — a bare visibility check would also pass on a page that simply hadn't rendered.
- Playwright artifacts write to the package-root `.tmp/`, matching where the Cypress job writes and what `.gitignore` already covers.

## Cypress parity

`ui-test/` is untouched. Both source specs remain in place and Cypress stays the gating suite.

## Test evidence

- **8 passed, 3 consecutive runs** against a live stack (`APIM_TAG=4.12` — see the caveat below).
- Perturbed a locator to confirm the specs genuinely fail rather than pass vacuously.
- **Full containerized run** (`yarn test:pw`, the exact path CI takes): exit code 0, 8 passed on 3 workers, JUnit XML written to the path the job archives.
- CI config: `tsc` builds, **148/148 tests**, lint clean. e2e: `tsc --noEmit` and `yarn lint` clean.

**Not verified in real CI.** E2E jobs are gated on `isSupportBranchOrMaster(branch) || isE2EBranch(branch)` (`/.*-run-e2e.*/`), which this branch matches neither — so the whole e2e block, including the new job, is skipped. **The branch needs `-run-e2e` in its name for this to prove itself in CI.**

Separately, `.env` pins `APIM_TAG=3`, so the default local stack runs a 3.x Console against 4.13 source and renders nothing; I verified against 4.12, the newest public tag. CI is unaffected — it builds images from the branch. This is documented in the README.

## Security-sensitive

Touches CI/deployment configuration (`nginx.conf`, docker-compose, CircleCI workflows) and authentication (session-cookie capture into `storageState`, second persona). No credentials added — personas read the existing `.env` values, and generated `.auth/` state is gitignored.

## Known gaps (deliberate, follow-ups)

- **Conventions are unenforced.** The README requires page objects to hold no assertions and specs to import `test`/`expect` from the fixture; AM and Cockpit enforce both via eslint. This package has no eslint for `playwright/` — documentation only for now.
- **No data fixtures yet.** Neither POC spec creates data, so adding SDK-backed setup/teardown fixtures would have been speculative. The next spec batch is where that path gets exercised.

## Checklist

- [x] License headers on new files
- [x] No `page.waitForTimeout()` / `networkidle`
- [x] Page objects contain zero assertions
- [x] Specs import `test`/`expect` from `fixtures/base.fixture.ts`
- [x] `linkJira()` present with the epic key
- [x] Cypress counterpart untouched and still gating


aqibmohammadkhan:
closing this as we no longer need this.

PR #19143, opened 2026-08-13
fix(gateway): retry once when the peer closed or reset the pooled connection

carlos-andres-osorio:
## Issue

https://gravitee.atlassian.net/browse/APIM-14873

## Description

`HttpConnector.connect()` had no retry anywhere in the acquire-and-send chain. When the gateway wrote a request onto a pooled keep-alive connection that the backend (or an intermediary such as an OpenShift Route/HAProxy) had already closed, the failure propagated straight to `HttpProxyEndpointConnector.handleException`, which interrupts the exchange unconditionally — the client saw a 502 for a request the backend never handled.

A pooled connection can only be discovered to be dead by writing to it, so this race isn't avoidable by configuration. `FailoverInvoker` is opt-in per-API and sits a layer above the connector — a workaround, not a fix for this race.

The exchange is now re-subscribed exactly once when the failure classifies as `CONNECTION_CLOSED` or `CONNECTION_RESET` (both are the same pool-reuse race — the peer tore down the idle connection before answering — surfacing as a clean close or a reset depending on OS/timing). Because an error on the upstream `Single` short-circuits `doOnSuccess`, the retry is structurally unreachable once a backend response has started arriving — no new flag needed to guard that. The retry is logged as a WARN via `ctx.withLogger(log)`, naming the endpoint, so it's observable in production rather than looking identical to a first-attempt success.

Live load-testing during development showed a CLOSED-only version of this fix eliminated close failures but left reset failures unchanged; extending the retry to `CONNECTION_RESET` closed that gap. See **Load testing** below for measured before/after failure rates.

## Limitations

**Only requests sent without a body are retried.** This covers GET/HEAD/OPTIONS and the common case of DELETE, but excludes any idempotent method sent with a body (e.g. a PUT that replaces a resource, or a bulk-delete DELETE with a JSON payload) even though replaying an idempotent request is safe by definition. The gate isn't method-based — it's based on whether the request bytes are replayable:

- *Correctness for non-idempotent methods.* Once body chunks are written they may have already reached and been acted on by the backend — replaying a POST/PATCH could apply it twice.
- *Feasibility, independent of method.* `request.chunks()` resolves to the hot `nativeRequest.toFlowable()` stream, which isn't safe to subscribe to twice. Retrying any request with a body — idempotent or not — would require buffering the body first, which is a separate change with its own memory-profile tradeoffs. Not attempted here.

**Not covered in this branch:**
- `InvokerAdapter` (legacy V3/emulation-engine path) has the same no-retry defect — but the root cause isn't in this repo. `InvokerAdapter` is only plumbing; the actual connection acquisition and failure handling live in `io.gravitee.connector:gravitee-connector-http` (`HttpConnection.connect()`/`handleException()`), an external, separately-released artifact this repo pins as a version in `pom.xml`. Fixing it requires a change in that artifact, not this repo — out of scope here. Confirmed via load test, same deterministic backend (see **Load testing** below): a `v4-emulation-engine` API (e.g. any V2-definition API, since V2 creation isn't wired into management-v2 REST yet) failed at 0.39–0.65% across 3 runs on unfixed master and 0.44–0.50% across 3 runs on this branch — essentially unchanged either side of this fix, as expected since this fix never touches that code path.
- `WebSocketConnector` overrides `connect()` so it doesn't inherit the retry.
- `GrpcConnector` inherits the retry but always negotiates HTTP/2, so `sendsRequestBody` is always true and the retry path is inert for it.

## Load testing

Methodology: `ab -k -c 50 -n 20000` against a raw TCP backend built for this repro, which closes every 20th request served on a given connection — deterministic and synchronous (no timer), so the expected number of race-prone events is computable (≈ requests / 20) and checkable against the observed failure count. Each condition below is 3 back-to-back runs with only one gateway+management-API instance active at a time (co-running instances measurably suppress the failure rate through CPU contention, so runs were isolated).

| Condition | Runs (failures / 20,000) | Range | Notes |
|---|---|---|---|
| Before this fix (master, no retry) | 280, 277, 268 | 1.34–1.40% | Baseline — every pool-reuse race is a client-visible 502 |
| After this fix (CLOSED+RESET retry) | 1, 5, 10 | 0.005–0.05% | ~30-45x reduction; residual is the retry-once cap exhausted twice on the same request |

Backend-side stats (`connections=... requests=... closes=...`, printed on shutdown) stayed consistent across every run (~19.6–19.8 requests per connection, matching the every-20th-request close cadence), confirming the race exposure itself didn't drift between runs.

## Additional context

**Security-sensitive:** adds a WARN-level diagnostic log line (`ctx.withLogger(log)`) on the retry path, naming the endpoint — no new data is logged beyond what's already visible via existing connector logging.

**Verification:** full `HttpConnectorTest` suite green (37/37), including the existing regression test that pins bodied requests are never resent after streaming has started.

### DIFF
--- a/gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
+++ b/gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
@@ -70,6 +70,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import lombok.CustomLog;
@@ -107,6 +108,19 @@ public class HttpConnector implements ProxyConnector {
      * it; anything looser would start blaming the configuration for genuine backend failures.
      */
     private static final long IDLE_TIMEOUT_MATCH_TOLERANCE_MS = 250;
+
+    /**
+     * RFC 9110 section 9.2.2 — "PUT, DELETE, and safe request methods are idempotent". An allowlist rather than a
+     * POST/PATCH denylist: an unrecognised verb resolves to OTHER, which must not be re-sent either.
+     */
+    private static final Set<io.gravitee.common.http.HttpMethod> IDEMPOTENT_METHODS = Set.of(
+        io.gravitee.common.http.HttpMethod.GET,
+        io.gravitee.common.http.HttpMethod.HEAD,
+        io.gravitee.common.http.HttpMethod.OPTIONS,
+        io.gravitee.common.http.HttpMethod.TRACE,
+        io.gravitee.common.http.HttpMethod.PUT,
+        io.gravitee.common.http.HttpMethod.DELETE
+    );
 
     private final String relativeTarget;
     protected final String defaultHost;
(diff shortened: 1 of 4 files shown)


gravitee-apim-plugin/gravitee-apim-plugin-endpoint/gravitee-apim-plugin-endpoint-http-proxy/src/main/java/io/gravitee/plugin/endpoint/http/proxy/connector/HttpConnector.java
```diff
@@ -191,12 +195,69 @@ public Completable connect(final HttpExecutionContext ctx) {
...
+                    final boolean retry = attempt == 1 && !cancelled.get() && isConnectionClosedOrResetByPeer(throwable);
+                    if (retry) {
+                        ctx
+                            .withLogger(log)
+                            .warn(
+                                "Pooled connection to [{}] was closed or reset by the peer, retrying once on a fresh connection",
+                                absoluteUri
+                            );
+                    }
+                    return retry;
```
carlos-andres-osorio: Done — split the combined retry predicate inside `retryOnceOnConnectionClosedByPeer` (line ~259) into sequential guard clauses (attempt count, cancellation, then failure classification), matching the suggested diff. Landed in commit 86d813be76.

michel-barret (approved):

PR #19146, opened 2026-08-14
fix(mapi): missing HRID in application

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-14894

## Description

Applications created via Console, Developer Portal, or Management API were persisted without an `hrid`. Only Automation API / GitOps creates had one. That breaks Terraform import, which requires `hrid`.

### Root cause
`hrid` was introduced in 4.9. `ApplicationHRIDUpgrader` (order 958) backfilled `hrid = id` once for existing apps, but:
- `ApplicationServiceImpl.create()` never defaulted `hrid` for manual creates
- that one-time upgrader cannot run again, so apps created after it stayed without `hrid`
- re-running `ApplicationHRIDUpgrader` is unsafe: it overwrites every `hrid`, including GitOps ones

### Fix
Three complementary changes (repository-agnostic via `ApplicationRepository`; same for MongoDB and JDBC):
1. **Create** — if `hrid` is blank, set `hrid = id` in `ApplicationServiceImpl.create()`
2. **Update** — if the stored `hrid` is blank, heal it to `id` on update (preserve existing GitOps hrids)
3. **Upgrader** — new `ApplicationMissingHridUpgrader` (order 963) backfills **only** blank/missing `hrid` to `id`; leaves existing hrids untouched

Before:
image

After:
image

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/ApplicationServiceImpl.java
@@ -471,6 +471,9 @@ public class ApplicationServiceImpl extends AbstractService implements Applicati
         } else {
             application.setId(UuidString.generateRandom());
         }
+        if (isBlank(application.getHrid())) {
+            application.setHrid(application.getId());
+        }
         application.setStatus(ApplicationStatus.ACTIVE);
         metadata.forEach((key, value) -> application.getMetadata().put(key, value));
 
@@ -678,7 +681,7 @@ public class ApplicationServiceImpl extends AbstractService implements Applicati
 
             Application application = applicationConverter.toApplication(updateApplicationEntity);
             application.setId(applicationId);
-            application.setHrid(applicationToUpdate.getHrid());
+            application.setHrid(isBlank(applicationToUpdate.getHrid()) ? applicationToUpdate.getId() : applicationToUpdate.getHrid());
             application.setEnvironmentId(applicationToUpdate.getEnvironmentId());
             application.setStatus(ApplicationStatus.ACTIVE);
             application.setType(applicationToUpdate.getType());
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ApplicationMissingHridUpgrader.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ApplicationMissingHridUpgrader.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


arpzilla-dev:
hi @pragsty 

- The ticket’s expected behaviour is: “All Applications should have an hrid, regardless of how they were created.” This change defaults hrid on create when it is missing. How does that cover applications that already exist without an hrid (created after the 4.9 upgrader ran)?

- After this change, will every newly created application always have an hrid? Including Console / Portal / Management API, and also when hrid is omitted vs sent as empty?

pragsty:
> hi @pragsty
> 
> * The ticket’s expected behaviour is: “All Applications should have an hrid, regardless of how they were created.” This change defaults hrid on create when it is missing. How does that cover applications that already exist without an hrid (created after the 4.9 upgrader ran)?
> * After this change, will every newly created application always have an hrid? Including Console / Portal / Management API, and also when hrid is omitted vs sent as empty?

1- I've added an upgrader, that'll resolve this issue
2- Yes

aqibmohammadkhan (commented):
Important
1. apply-on-master: order 959 is already taken on master

On this 4.12.x PR, APPLICATION_MISSING_HRID_UPGRADER = 959 is fine (last constant was 958).

On current master, 959 is KAFKA_EXPLORER_PERMISSION_UPGRADER, then 960–962 are other role upgraders. Cherry-picking this as-is will conflict in UpgraderOrder.java. When Mergify applies to master, use the next free order (963+), not 959.

The PR description still says “order 963”; the 4.12.x code is 959. Align the description with the branch.

2. Misleading upgrader test name

should_skip_application_with_blank_hrid_only does not test a blank " " hrid. It tests a mixed set and verify(..., times(1)).update(any()) without checking which app was updated. Prefer argThat(app -> "app-without-hrid".equals(app.getId()) && "app-without-hrid".equals(app.getHrid())), and optionally a case for hrid = " ".

3. No create test for empty/whitespace hrid

isBlank handles "" / " ", but tests only stub null. Cheap to add; not blocking.

4. Silent upgrader

A log line with the count of updated apps would help confirm Getlink’s upgrade in Management API logs.

pragsty:
> Important
> 
> 1. apply-on-master: order 959 is already taken on master
> 
> On this 4.12.x PR, APPLICATION_MISSING_HRID_UPGRADER = 959 is fine (last constant was 958).
> 
> On current master, 959 is KAFKA_EXPLORER_PERMISSION_UPGRADER, then 960–962 are other role upgraders. Cherry-picking this as-is will conflict in UpgraderOrder.java. When Mergify applies to master, use the next free order (963+), not 959.
> 
> The PR description still says “order 963”; the 4.12.x code is 959. Align the description with the branch.
> 
> 2. Misleading upgrader test name
> 
> should_skip_application_with_blank_hrid_only does not test a blank " " hrid. It tests a mixed set and verify(..., times(1)).update(any()) without checking which app was updated. Prefer argThat(app -> "app-without-hrid".equals(app.getId()) && "app-without-hrid".equals(app.getHrid())), and optionally a case for hrid = " ".
> 
> 3. No create test for empty/whitespace hrid
> 
> isBlank handles "" / " ", but tests only stub null. Cheap to add; not blocking.
> 
> 4. Silent upgrader
> 
> A log line with the count of updated apps would help confirm Getlink’s upgrade in Management API logs.

Fixed all comments, updated description for order,
branches having conflicts will be taken care of.

benoitgravitee (changes requested):
This is not a bug.
Import of existing resources in Terraform is not supported.
Please close this PR.
It should be a feature request, APIv4 supports it but not Application.
User should (be they can't) export an application using a special header.
Then they would apply it as a new Application with Terraform.

HRID are only valued by Automation API. HRIDs are mashed-up into IDs so they deterministic. You cannot assign an HRID in the DB that would match an existing UUID (as they are random). Using that generated HRID in the Database cannot allow the read the Application.

pragsty:
as per comment above closing this PR

PR #19147, opened 2026-08-14
feat: expose unified API and API Product subscription search

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-120

## Description

Adds unified Portal API subscription search for API and API Product subscriptions.

## Changes

- Add `referenceTypes` filtering for API and API Product subscriptions.
- Add name search across API and API Product targets using the `query` parameter.
- Apply application access filtering before pagination.
- Support combined filtering and pagination in JDBC and MongoDB repositories.
- Preserve legacy API subscriptions without an explicit `referenceType`.
- Preserve existing API-only behavior when `referenceTypes` is omitted.
- Add deterministic ordering for mixed API and API Product results.
- Extend the Portal OpenAPI contract and resource tests.

https://github.com/user-attachments/assets/665c9828-c393-4b27-a442-b191d6d96605

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionsResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionsResource.java
@@ -16,13 +16,13 @@
 package io.gravitee.rest.api.portal.rest.resource;
 
 import static java.util.Collections.emptyList;
-import static java.util.stream.Collectors.toSet;
 
 import io.gravitee.apim.core.portal_page.domain_service.PortalApiProductAccessDomainService;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemViewerContext;
 import io.gravitee.apim.core.subscription.model.SubscriptionConfiguration;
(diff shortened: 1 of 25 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionsResource.java
```diff
@@ -317,19 +322,29 @@ private static List<String> resolveApplicationIds(String applicationId, List<Str
...
     }
 
-    private Page<SubscriptionEntity> fetchSubscriptions(PaginationParam paginationParam, SubscriptionQuery query) {
-        ExecutionContext executionContext = GraviteeContext.getExecutionContext();
-        if (!paginationParam.hasPagination()) {
-            Collection<SubscriptionEntity> resp = subscriptionService.search(executionContext, query);
-            return new Page<>(resp.stream().toList(), 0, resp.size(), resp.size());
+    private static Set<SubscriptionReferenceType> resolveReferenceTypes(
+        List<String> referenceTypes,
+        boolean hasApiFilter,
```
jaroslawlakomy-grav, line 327: hasApiFilter is used below to reject API ID filters combined with reference types other than API.

dev-marek (approved):
looks ok

is it planned to extend permission calculation logic to include apis that are a part of api product that are subscribed to an application which is managed by an user or a grup an user is a member of?

PR #19159, opened 2026-08-14
fix(rest-api): send the registration email only once the request is approved

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-14920

## Description

When automatic validation of registration requests is disabled, the activation email was sent at submission instead of after approval, and `finalizeRegistration` never inspected `UserStatus`: a `PENDING` user could set a password and then fail to log in, with nothing telling them why.

- the registration email is sent when the request is approved, not at submission
- `finalizeRegistration` refuses a `PENDING` user (409, awaiting approval) and a `REJECTED`/`ARCHIVED` one
- a pending account attempting to log in gets a 401 instead of a 500
- both portals tell the user their registration awaits approval
- approving sends a single email, the one carrying the activation link

## Additional context

https://github.com/user-attachments/assets/f9465548-fb7c-46b2-bb15-a18f94f21e3e


dev-marek (approved):
lgtm

PR #19163, opened 2026-08-14
fix(dictionary): prevent key/id collision from evicting another dictionary

arpzilla-dev:
## Issue

https://gravitee.atlassian.net/browse/APIM-14821

## Description

Two dictionaries in the same environment could share one gateway runtime slot when one dictionary’s key equalled another’s id (typical Console + Terraform mix). Deploying the second silently replaced the first. Undeploying or deleting either one emptied the slot, so #dictionaries['…'] started 500ing until a gateway restart.

This PR:

- Rejects create when an explicit key (v2/v4 API / Terraform) matches another dictionary’s id in the same environment.
- Undeploys a gateway slot only if the occupant’s id is the dictionary being removed; otherwise logs a warning and leaves it.
- show in DEBUG logs when deploy overwrites a different dictionary, and logs the dictionary that was actually requested on undeploy.

## Additional context 

- showing the complete flow [creating client's current similar situation]  [video with audio]

https://github.com/user-attachments/assets/cf9095f7-ae35-493b-a067-3edfbacd1d13

- Rejecting the creation 
image (14)

- logs
image

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
@@ -19,6 +19,7 @@ import io.gravitee.gateway.dictionary.model.Dictionary;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import lombok.CustomLog;
 
 /**
@@ -34,13 +35,24 @@ public class MultiEnvironmentDictionaryManager implements DictionaryManager {
     @Override
     public void deploy(Dictionary dictionary) {
         String environmentId = dictionary.getEnvironmentId();
-        //fallback on legacy events
-        String key = dictionary.getKey() == null ? dictionary.getId() : dictionary.getKey();
+        String key = runtimeKey(dictionary);
 
         dictionaries.putIfAbsent(environmentId, new HashMap<>());
         values.putIfAbsent(environmentId, new HashMap<>());
 
         Dictionary existing = dictionaries.get(environmentId).get(key);
+        if (existing != null && !Objects.equals(existing.getId(), dictionary.getId())) {
+            log.error(
+                "Dictionary {} (name={}) collides on runtime key '{}' in environment {} with existing dictionary {} (name={}). Incoming deploy is rejected; occupant is kept",
+                dictionary.getId(),
+                dictionary.getName(),
+                key,
+                environmentId,
+                existing.getId(),
+                existing.getName()
+            );
+            return;
+        }
         if (existing == null || dictionary.getDeployedAt().after(existing.getDeployedAt())) {
             if (dictionary.getProperties() == null) {
                 dictionary.setProperties(Collections.emptyMap());
@@ -55,28 +67,47 @@ public class MultiEnvironmentDictionaryManager implements DictionaryManager {
     @Override
     public void undeploy(Dictionary dictionary) {
         String environmentId = dictionary.getEnvironmentId();
-        //fallback on legacy events
-        String key = dictionary.getKey() == null ? dictionary.getId() : dictionary.getKey();
+        String key = runtimeKey(dictionary);
         Map<String, Dictionary> envDictionaries = dictionaries.get(environmentId);
         if (envDictionaries != null) {
-            Dictionary removed = envDictionaries.remove(key);
+            Dictionary occupant = envDictionaries.get(key);
+            if (occupant == null) {
+                return;
+            }
+            if (!Objects.equals(occupant.getId(), dictionary.getId())) {
+                log.debug(
+                    "Dictionary {} was requested for undeploy but runtime key '{}' is occupied by dictionary {} — ignoring",
+                    dictionary.getId(),
+                    key,
+                    occupant.getId()
+                );
+                return;
+            }
 
+            envDictionaries.remove(key);
             if (envDictionaries.isEmpty()) {
                 dictionaries.remove(environmentId);
             }
 
-            if (removed != null) {
-                Map<String, Map<String, String>> envValues = values.get(environmentId);
-                if (envValues != null) {
-                    envValues.remove(key);
-                    if (envValues.isEmpty()) {
-                        values.remove(environmentId);
-                    }
+            Map<String, Map<String, String>> envValues = values.get(environmentId);
+            if (envValues != null) {
+                envValues.remove(key);
+                if (envValues.isEmpty()) {
+                    values.remove(environmentId);
                 }
-
-                log.info("A dictionary has been undeployed: {}", removed);
             }
+
+            log.info("A dictionary has been undeployed: {}", dictionary);
         }
+    }
+
+    /**
+     * Runtime map key: prefer the dictionary {@code key} when present, otherwise fall back to {@code id}
+     * for legacy events that predate the key field.
+     */
+    private static String runtimeKey(Dictionary dictionary) {
+        String key = dictionary.getKey();
+        return key == null || key.isBlank() ? dictionary.getId() : key;
     }
 
     @Override
--- a/gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/test/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManagerTest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/test/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManagerTest.java
@@ -1,0 +1,158 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gateway.dictionary;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.gravitee.gateway.dictionary.model.Dictionary;
+import java.util.Date;
+import java.util.Map;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayNameGeneration;
+import org.junit.jupiter.api.DisplayNameGenerator;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
+class MultiEnvironmentDictionaryManagerTest {
+
+    private static final String ENV = "DEFAULT";
+    private static final String OTHER_ENV = "OTHER";
+
+    private MultiEnvironmentDictionaryManager cut;
+
+    @BeforeEach
+    void setUp() {
+        cut = new MultiEnvironmentDictionaryManager();
+    }
+
+    @Nested
+    class DeployTest {
+
+        @Test
+        void should_index_console_dictionary_by_id_when_key_is_null() {
+            cut.deploy(dictionary("idp-server-details", null, ENV, "first-value", 1L));
+
+            assertThat(property(ENV, "idp-server-details")).isEqualTo("first-value");
+        }
+
+        @Test
+        void should_keep_both_dictionaries_when_runtime_keys_differ() {
+            cut.deploy(dictionary("idp-server-details", null, ENV, "first-value", 1L));
+            cut.deploy(dictionary("uuid-tf", "tf_idp-server-details", ENV, "tf-value", 2L));
+
+            assertThat(property(ENV, "idp-server-details")).isEqualTo("first-value");
+            assertThat(property(ENV, "tf_idp-server-details")).isEqualTo("tf-value");
+        }
+
+        @Test
+        void should_keep_incumbent_when_another_dictionary_collides_on_runtime_key() {
+            cut.deploy(dictionary("idp-server-details", null, ENV, "first-value", 1L));
+            cut.deploy(dictionary("uuid-tf", "idp-server-details", ENV, "second-value", 2L));
+
+            assertThat(property(ENV, "idp-server-details")).isEqualTo("first-value");
+        }
+
+        @Test
+        void should_replace_properties_when_same_dictionary_is_redeployed() {
+            cut.deploy(dictionary("idp-server-details", null, ENV, "first-value", 1L));
+            cut.deploy(dictionary("idp-server-details", null, ENV, "updated-value", 2L));
+
+            assertThat(property(ENV, "idp-server-details")).isEqualTo("updated-value");
+        }
+
+        @Test
+        void should_not_overwrite_when_same_dictionary_deployed_at_is_older() {
+            cut.deploy(dictionary("idp-server-details", null, ENV, "first-value", 2L));
+            cut.deploy(dictionary("idp-server-details", null, ENV, "second-value", 1L));
+
+            assertThat(property(ENV, "idp-server-details")).isEqualTo("first-value");
+        }
+
+        @Test
+        void should_isolate_dictionaries_per_environment() {
+            cut.deploy(dictionary("idp-server-details", null, ENV, "default-value", 1L));
+            cut.deploy(dictionary("idp-server-details", null, OTHER_ENV, "other-value", 1L));
+
+            assertThat(property(ENV, "idp-server-details")).isEqualTo("default-value");
+            assertThat(property(OTHER_ENV, "idp-server-details")).isEqualTo("other-value");
+        }
+    }
+
+    @Nested
+    class UndeployTest {
+
+        @Test
+        void should_undeploy_when_occupant_id_matches() {
+            Dictionary console = dictionary("idp-server-details", null, ENV, "first-value", 1L);
+            cut.deploy(console);
+
+            cut.undeploy(console);
+
+            assertThat(cut.getDictionaries(ENV)).isNullOrEmpty();
+        }
+
+        @Test
+        void should_keep_incumbent_when_rejected_dictionary_is_undeployed() {
+            cut.deploy(dictionary("idp-server-details", null, ENV, "first-value", 1L));
+            Dictionary terraform = dictionary("uuid-tf", "idp-server-details", ENV, "second-value", 2L);
+            cut.deploy(terraform);
+
+            cut.undeploy(terraform);
+
+            assertThat(property(ENV, "idp-server-details")).isEqualTo("first-value");
+        }
+
+        @Test
+        void should_keep_occupant_when_undeploy_targets_a_different_dictionary() {
+            cut.deploy(dictionary("uuid-tf", "idp-server-details", ENV, "tf-value", 1L));
+
+            cut.undeploy(dictionary("idp-server-details", null, ENV, "first-value", 1L));
+
+            assertThat(property(ENV, "idp-server-details")).isEqualTo("tf-value");
+        }
+
+        @Test
+        void should_keep_the_other_dictionary_when_one_of_two_is_undeployed() {
+            cut.deploy(dictionary("idp-server-details", null, ENV, "first-value", 1L));
+            Dictionary terraform = dictionary("uuid-tf", "tf_idp-server-details", ENV, "tf-value", 2L);
+            cut.deploy(terraform);
+
+            cut.undeploy(terraform);
+
+            assertThat(property(ENV, "idp-server-details")).isEqualTo("first-value");
+            assertThat(cut.getDictionaries(ENV)).doesNotContainKey("tf_idp-server-details");
+        }
+    }
+
+    private String property(String environmentId, String runtimeKey) {
+        Map<String, Map<String, String>> envValues = cut.getDictionaries(environmentId);
+        assertThat(envValues).isNotNull();
+        assertThat(envValues).containsKey(runtimeKey);
+        return envValues.get(runtimeKey).get("MY_PROP");
+    }
+
+    private static Dictionary dictionary(String id, String key, String environmentId, String propertyValue, long deployedAt) {
+        Dictionary dictionary = new Dictionary();
+        dictionary.setId(id);
+        dictionary.setKey(key);
+        dictionary.setEnvironmentId(environmentId);
+        dictionary.setName(id);
+        dictionary.setDeployedAt(new Date(deployedAt));
+        dictionary.setProperties(Map.of("MY_PROP", propertyValue));
+        return dictionary;
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/dictionary/DictionaryServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/dictionary/DictionaryServiceImpl.java
@@ -239,30 +239,13 @@ public class DictionaryServiceImpl extends AbstractService implements Dictionary
             final Dictionary dictionary;
             if (newDictionaryEntity.getKey() == null) {
                 String key = IdGenerator.generate(newDictionaryEntity.getName());
-
-                // Make sure the derived key does not already exist in the same environment
-                // That would mean that a legacy dictionary (prior to multi tenant management) was created with the same name.
                 Optional<Dictionary> idDictionary = dictionaryRepository.findById(key);
-                if (
-                    idDictionary.isPresent() && idDictionary.get().getEnvironmentId().equalsIgnoreCase(executionContext.getEnvironmentId())
-                ) {
-                    throw new DictionaryAlreadyExistsException(newDictionaryEntity.getName());
-                }
-                // Make sure the derived key does not already exist in any environment using the key field which is how it is checked
-                // uniqueness after the multi tenant management is implemented.
-                Optional<Dictionary> optDictionary = dictionaryRepository.findByKeyAndEnvironment(key, executionContext.getEnvironmentId());
-                if (optDictionary.isPresent()) {
-                    throw new DictionaryAlreadyExistsException(newDictionaryEntity.getName());
-                }
+                ensureRuntimeKeyIsAvailable(executionContext, key, newDictionaryEntity.getName(), idDictionary, false);
                 dictionary = convert(newDictionaryEntity, key, idDictionary.isEmpty());
             } else {
                 String key = newDictionaryEntity.getKey();
-                // We shouldn't be here if the caller checked the key uniqueness
-                // But for to respect the contract, we check again
-                Optional<Dictionary> optDictionary = dictionaryRepository.findByKeyAndEnvironment(key, executionContext.getEnvironmentId());
-                if (optDictionary.isPresent()) {
-                    throw new DictionaryAlreadyExistsException(newDictionaryEntity.getName());
-                }
+                Optional<Dictionary> idDictionary = dictionaryRepository.findById(key);
+                ensureRuntimeKeyIsAvailable(executionContext, key, newDictionaryEntity.getName(), idDictionary, true);
                 dictionary = convert(newDictionaryEntity, key, false);
                 // if ID is already set, let's use it
                 if (newDictionaryEntity.getId() != null) {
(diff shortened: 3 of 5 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
```diff
@@ -55,10 +63,23 @@ public void deploy(Dictionary dictionary) {
     @Override
     public void undeploy(Dictionary dictionary) {
         String environmentId = dictionary.getEnvironmentId();
-        //fallback on legacy events
-        String key = dictionary.getKey() == null ? dictionary.getId() : dictionary.getKey();
+        String key = runtimeKey(dictionary);
         Map<String, Dictionary> envDictionaries = dictionaries.get(environmentId);
         if (envDictionaries != null) {
+            Dictionary occupant = envDictionaries.get(key);
```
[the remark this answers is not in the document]
arpzilla-dev: we didnt reinvent this logic....it was existing implementation hashMap implementation.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/test/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManagerTest.java
```diff
@@ -0,0 +1,139 @@
...
+            cut.deploy(dictionary("idp-server-details", null, ENV, "first-value", 1L));
+            cut.deploy(dictionary("uuid-tf", "idp-server-details", ENV, "second-value", 2L));
+
+            cut.undeploy(dictionary("idp-server-details", null, ENV, "first-value", 1L));
+
+            assertThat(property(ENV, "idp-server-details")).isEqualTo("second-value");
+        }
+
+        @Test
+        void should_empty_slot_when_colliding_occupant_is_undeployed() {
```
wbabyte: **Blocking: this test asserts the behaviour the ticket reports as the bug.**

APIM-14821 steps 4-5 undeploy the *colliding* dictionary, and that path is unchanged by this PR:

| step | result |
|---|---|
| deploy A (`id=my-dictionary`, `key=null`) | `200 first-value` |
| deploy B (`id=my-dictionary-b`, `key=my-dictionary`) | `200 second-value` + the new WARN |
| **undeploy B** | **500** — slot emptied, A still recorded as deployed |
| undeploy A instead | `200` ✅ fixed by this PR |

`deploy` overwrites the slot at L58-59 and keeps no record of the displaced A, so when B is undeployed there is nothing to restore from. The incremental synchronizer won't bring A back either — `DictionarySynchronizer` only fetches events inside its `(from, to)` window and A's `PUBLISH_DICTIONARY` predates it — so recovery still needs a gateway restart, which is the workaround the ticket is trying to eliminate.

This is also the *likely* remediation path: on discovering the duplicate, the natural operator move is to delete the Terraform-created dictionary, which silently takes the Console one down with it.

Scope nuance: the create-side guard prevents this arising in new state, so the residual gap only affects already-collided installs — but that is exactly the population the gateway half of this PR exists for.

Two options, both consistent with the ticket ("*Either the two identifier namespaces are kept separate, or the collision is warned about or rejected at deploy time*"):
- **Reject the colliding deploy** — keep the incumbent, log an error, don't take the slot. Simpler, and symmetric with the create-side rejection.
- **Track contenders per slot** so removing the winner promotes the loser.

Whichever you pick, this test needs to stop asserting the loss. Leaving it green as-is means the suite certifies the reported symptom as correct, and a future reader will take it as the spec.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
```diff
@@ -34,14 +35,21 @@ public class MultiEnvironmentDictionaryManager implements DictionaryManager {
...
         String environmentId = dictionary.getEnvironmentId();
-        //fallback on legacy events
-        String key = dictionary.getKey() == null ? dictionary.getId() : dictionary.getKey();
+        String key = runtimeKey(dictionary);
 
         dictionaries.putIfAbsent(environmentId, new HashMap<>());
         values.putIfAbsent(environmentId, new HashMap<>());
 
         Dictionary existing = dictionaries.get(environmentId).get(key);
         if (existing == null || dictionary.getDeployedAt().after(existing.getDeployedAt())) {
```
wbabyte, line 56: **The collision goes completely unlogged in the more dangerous direction.**

The new WARN is *inside* this branch, so it only fires when the incoming dictionary wins the `deployedAt` comparison. When the incoming dictionary collides but is **older**, it falls off the end of the method with no log at any level — the deployment is silently discarded.

That's the worse case: the control plane reports the dictionary as deployed (`deployedAt` set, `PUBLISH_DICTIONARY` written, sync logs "N dictionaries synchronized"), but `#dictionaries['...']` resolves to the other dictionary's values and nothing anywhere explains why. It's also roughly a coin flip which direction a given install hits.

Suggest hoisting detection above the `deployedAt` guard so both outcomes are logged:

```java
Dictionary existing = dictionaries.get(environmentId).get(key);
boolean collision = existing != null && !Objects.equals(existing.getId(), dictionary.getId());
boolean wins = existing == null || dictionary.getDeployedAt().after(existing.getDeployedAt());

if (collision) {
    log.warn(
        "Dictionary {} (key '{}') collides on key '{}' in environment {} with dictionary {}; {}. " +
        "Only one is resolvable via #dictionaries['{}'] - change one dictionary's key.",
        dictionary.getId(), dictionary.getKey(), key, environmentId, existing.getId(),
        wins ? "it replaces the existing one" : "it will NOT be deployed (older deployedAt)", key
    );
}
if (wins) { ... }
```

Also worth adding `environmentId` and `name` to the message — Terraform-created ids are derived UUIDs (`HRIDToUUID`), so an id alone can't be mapped back to anything an operator can see in the Console.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
```diff
@@ -55,10 +63,23 @@ public void deploy(Dictionary dictionary) {
...
-        String key = dictionary.getKey() == null ? dictionary.getId() : dictionary.getKey();
+        String key = runtimeKey(dictionary);
         Map<String, Dictionary> envDictionaries = dictionaries.get(environmentId);
         if (envDictionaries != null) {
+            Dictionary occupant = envDictionaries.get(key);
+            if (occupant == null) {
+                return;
+            }
+            if (!Objects.equals(occupant.getId(), dictionary.getId())) {
+                log.debug(
```
wbabyte, line 78: **`log.debug` here, but the PR description says this "logs a warning".** Worth reconciling one way or the other, since reviewers reading the description will assume a guarantee the code doesn't provide.

Gateways run at INFO by default, so in production this line doesn't exist. The operator experience becomes: delete dictionary A in the Console (204), the gateway keeps serving B's values under A's name, and the only way to find out why is to enable DEBUG on a production gateway — which is the debugging cost this PR is trying to remove.

WARN also seems right on the merits: this branch is unreachable through supported input (the new create guard refuses to produce the state), and it makes a user-visible action have no effect. Suggest including the remedy too:

```java
log.warn(
    "Undeploy of dictionary {} was skipped: key '{}' in environment {} is held by dictionary {}. " +
    "Two dictionaries share this key (one's key equals the other's id), so '{}' still resolves to {}'s values. " +
    "Rename one dictionary's key to clear the collision.",
    dictionary.getId(), key, environmentId, occupant.getId(), key, occupant.getId()
);
```

Separately, the `occupant == null` early return just above logs nothing at all. DEBUG is defensible there since idempotent replay is common, but it shouldn't be zero — and right now the two skip paths are asymmetric, so a reader can't tell whether the omission is deliberate.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
```diff
@@ -55,10 +63,23 @@ public void deploy(Dictionary dictionary) {
...
+                    "Dictionary {} was requested for undeploy but runtime key '{}' is occupied by dictionary {} — ignoring",
+                    dictionary.getId(),
+                    key,
+                    occupant.getId()
+                );
+                return;
+            }
+
             Dictionary removed = envDictionaries.remove(key);
```
wbabyte: **Dead code left by the log-argument change.**

Past the `occupant == null` return above, `envDictionaries.remove(key)` provably returns `occupant`, so `removed` is always non-null and the `if (removed != null)` guard below can never be false. `removed` itself is now unused, since L98 logs `dictionary` instead.

`AGENTS.md`: *"No dead code: after a change, remove the fields, methods, imports, and parameters it left unused"* and *"Inside trusted internal code, don't add handling for scenarios that genuinely can't happen."*

Collapses to `envDictionaries.remove(key);` with the body dedented.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
wbabyte, line 73: **`undeploy` gained a non-atomic check-then-act.** It used to be a single `remove(key)`; it's now `get` -> compare -> `remove`.

This manager is a singleton (`DictionaryConfiguration`) holding plain `HashMap`s, and the sync layer dispatches deployments through `.parallel(syncDeployerExecutor.getMaximumPoolSize()).runOn(...)` in `DictionarySynchronizer`, with `POOL_SIZE = availableProcessors() * 2`. `.parallel()` is round-robin, so two events for the same runtime key in one batch can land on different `gio.sync-deployer-N` threads — meaning the guard can validate one occupant and evict a different one.

Cheap local fix — the value-conditional overload collapses check and act into one step, and it's safe here because `Dictionary` is `@EqualsAndHashCode(onlyExplicitlyIncluded = true)` over `id` + `environmentId` only (so the `deployedAt == null` that `DictionaryServiceImpl.undeploy` puts in the event payload doesn't affect the match):

```java
if (!envDictionaries.remove(key, dictionary)) { /* log skip, return */ }
```

Flagging rather than asking you to fix: the underlying thread-safety gap is pre-existing and wider than this PR. Both maps are unsynchronised `HashMap`s with non-atomic `putIfAbsent`, and `EnvironmentDictionaryTemplateVariableProvider.provide` hands the **live inner map instance** into the EL template context, so request-path threads read it while the sync deployer mutates it. Probably deserves its own ticket.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/dictionary/DictionaryServiceImpl.java
```diff
@@ -257,6 +257,14 @@ public DictionaryEntity create(ExecutionContext executionContext, NewDictionaryE
                 dictionary = convert(newDictionaryEntity, key, idDictionary.isEmpty());
             } else {
                 String key = newDictionaryEntity.getKey();
+                // Gateway indexes deployed dictionaries by key, falling back to id. A new dictionary
+                // whose key equals another dictionary's id in this environment would share that slot.
+                Optional<Dictionary> idDictionary = dictionaryRepository.findById(key);
+                if (
+                    idDictionary.isPresent() && idDictionary.get().getEnvironmentId().equalsIgnoreCase(executionContext.getEnvironmentId())
+                ) {
+                    throw new DictionaryAlreadyExistsException(newDictionaryEntity.getName());
```
wbabyte: **This message describes a conflict that doesn't exist.**

`DictionaryAlreadyExistsException` renders `"A dictionary with name [" + dictionaryName + "] already exists in this environment."` and is passed the **new** dictionary's own name. In the scenario the new test covers (`key=idp-server-details`, `name=tf_idp-server-details`) the caller gets:

> `400 A dictionary with name [tf_idp-server-details] already exists in this environment.`

Nothing of that name exists; the conflict is on `key`, not `name`; and the actual conflicting dictionary (`idp-server-details`) is never named. `getParameters()` carries the same wrong value under `"dictionary"`, so the Console surface inherits it, and `getTechnicalCode()` is `dictionary.exists` — shared with the three other throw sites, so no client can distinguish the three rejection reasons.

This matters more than a wording nit: the 400 is the only signal a Terraform/CRD user gets, and it sends them looking for an object that doesn't exist. The Automation path reaches here with the hrid as the key, so the remedy is "change the hrid" or "migrate the legacy dictionary's id" — neither discoverable from this message.

The information is already in hand via the `idDictionary` Optional. Suggest a distinct exception:

```java
@Override public String getTechnicalCode() { return "dictionary.keyCollidesWithId"; }
@Override public String getMessage() {
    return "Cannot use key [" + requestedKey + "]: it is already the id of dictionary [" + conflictingName +
        "] in this environment. The gateway resolves dictionaries by key, falling back to id, so both would " +
        "share the same runtime slot. Choose a different key.";
}
```

Compare `ValidateDictionaryDomainService`, which states what's wrong *and* what to do: *"Manual dictionary must not have 'dynamic' properties (provider, trigger). Set type to 'DYNAMIC' or remove them."*

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/dictionary/DictionaryServiceImpl.java
```diff
@@ -257,6 +257,14 @@ public DictionaryEntity create(ExecutionContext executionContext, NewDictionaryE
                 dictionary = convert(newDictionaryEntity, key, idDictionary.isEmpty());
             } else {
                 String key = newDictionaryEntity.getKey();
+                // Gateway indexes deployed dictionaries by key, falling back to id. A new dictionary
+                // whose key equals another dictionary's id in this environment would share that slot.
+                Optional<Dictionary> idDictionary = dictionaryRepository.findById(key);
```
wbabyte, line 247: **This is byte-identical to the guard ~15 lines above** in the `key == null` branch. Worth extracting — and the tidiest form removes the asymmetry that makes the two branches easy to get out of sync in future:

```java
boolean explicitKey = newDictionaryEntity.getKey() != null;
String key = explicitKey ? newDictionaryEntity.getKey() : IdGenerator.generate(newDictionaryEntity.getName());

Optional<Dictionary> idDictionary = dictionaryRepository.findById(key);
if (idDictionary.filter(d -> executionContext.getEnvironmentId().equalsIgnoreCase(d.getEnvironmentId())).isPresent()) {
    throw new DictionaryKeyCollidesWithDictionaryIdException(key, idDictionary.get());
}
if (dictionaryRepository.findByKeyAndEnvironment(key, executionContext.getEnvironmentId()).isPresent()) {
    throw new DictionaryAlreadyExistsException(newDictionaryEntity.getName());
}

dictionary = convert(newDictionaryEntity, key, !explicitKey && idDictionary.isEmpty());
if (explicitKey && newDictionaryEntity.getId() != null) {
    dictionary.setId(newDictionaryEntity.getId());
}
```

Using `.filter(...)` also makes the environment comparison null-safe. The current `idDictionary.get().getEnvironmentId().equalsIgnoreCase(...)` NPEs on a row with a null `environmentId`; that pattern is pre-existing elsewhere in this file, but this call site is the only one that does a **global, unscoped** `findById`, so a single bad legacy row anywhere in the installation can now break creates in any environment.

Two smaller notes on this block:
- The comment says a colliding key "*would share that slot*", but that's only true when the other dictionary's `key` is null. If it has its own key it's indexed elsewhere and there's no collision, so the check is broader than its stated reason. It's unreachable in practice (rows with a non-null key always get a UUID id, from `convert` or `HRIDToUUID`), so I'd keep failing closed — but the comment should describe the invariant rather than the mechanism, e.g. *"id and key share one namespace per environment"*.
- The explicit `newDictionaryEntity.getId()` below is still unvalidated. It can't cause a slot collision, but a duplicate id surfaces as a 500 rather than a 400, since Mongo `create` uses `insert` and the `DuplicateKeyException` isn't handled. Probably its own ticket.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/test/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManagerTest.java
```diff
@@ -0,0 +1,139 @@
...
+    class UndeployTest {
+
+        @Test
+        void should_undeploy_when_occupant_id_matches() {
+            Dictionary console = dictionary("idp-server-details", null, ENV, "first-value", 1L);
+            cut.deploy(console);
+
+            cut.undeploy(console);
+
+            assertThat(cut.getDictionaries(ENV)).isNull();
```
wbabyte: **`isNull()` couples these tests to a cleanup detail, and it's why the most important undeploy property is untested.**

`getDictionaries` returns `values.get(environmentId)`, and `undeploy` drops the whole environment entry when it empties — so `null` here conflates "environment cleaned up" with "environment never existed" (`getDictionaries("NEVER")` also returns `null`). `isNullOrEmpty()` would be more honest.

The bigger gap: in every undeploy scenario the environment holds exactly **one** runtime key, so `remove(key)` and `dictionaries.remove(environmentId)` are indistinguishable. The bug being fixed *is* "an environment map got emptied by the wrong undeploy", and no test can currently tell those two apart — a future refactor that dropped the environment entry unconditionally would keep this suite green while re-introducing the incident for any customer with two dictionaries.

```java
@Test
void should_remove_only_the_targeted_dictionary_when_environment_has_others() {
    cut.deploy(dictionary("dic-a", null, ENV, "a-value", 1L));
    cut.deploy(dictionary("dic-b", "b-key", ENV, "b-value", 1L));

    cut.undeploy(dictionary("dic-a", null, ENV, "a-value", 1L));

    assertThat(cut.getDictionaries(ENV)).containsOnlyKeys("b-key");
}
```

Also missing, in rough priority order:
- The cross-environment mirror on the *undeploy* side — `should_isolate_dictionaries_per_environment` only covers deploy.
- The new deploy-time collision WARN, which is the only operator-visible output of the deploy-side change. There's a convention for this already: `MemoryAppender` under `src/test/java/appender/` in the sync and handlers-api modules, with `configureMemoryAppender()` usage in `SubscriptionAppenderTest` — and `logback-classic` is already on the test classpath at `provided` scope, so no pom change is needed. (It would be a third copy of `MemoryAppender`; promoting it to a shared test module is probably a separate item.)
- `should_index_console_dictionary_by_id_when_key_is_null` asserts nothing about "console" — the discriminator is just `key == null`, and the same shape is produced by current code for any name-derived dictionary whose slug is globally free.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/configuration/dictionary/DictionaryServiceImpl_CreateTest.java
```diff
@@ -154,6 +157,7 @@ public void shouldNotCreateWhenIdAlreadyExistsInSameEnvironment() throws Technic
         Dictionary existingById = new Dictionary();
         existingById.setKey("my-key");
         existingById.setEnvironmentId(ENVIRONMENT_ID);
+        when(dictionaryRepository.findById("my-key")).thenReturn(Optional.empty());
```
wbabyte: **This stub makes the test contradict its own name.** With `findById("my-key") -> empty` the test now explicitly asserts there is **no** id match, while the local is named `existingById` and never has an id set — so it reaches the exception through `findByKeyAndEnvironment`, making it input-for-input identical to `shouldNotCreateWhenKeyAlreadyExistsInSameEnvironment` above.

The name was already inaccurate before this PR (master's explicit-key branch never called `findById`), but the added stub turns an ambiguity into a contradiction. Since the new `shouldNotCreateWhenExplicitKeyMatchesExistingIdInSameEnvironment` now covers the real id case, deleting this one looks right; the alternative is repointing it at the id path (`findById -> Optional.of(existing with id="my-key")`, `findByKeyAndEnvironment -> empty`).

Minor, same file: 4 of the 5 added `findById` stubs are no-ops — Mockito already returns `Optional.empty()` by default for `Optional`-returning mocks, and the class is `Strictness.WARN` so nothing flags them. Only the one in `shouldCreateWithExplicitKeyWhenIdAlreadyTaken` earns its place (it stubs a *present* dictionary in `OTHER_ENV`, the sole negative test of the new guard's environment check). The others read as if they're driving behaviour when they aren't.

wbabyte (commented):
Reviewed against APIM-14821. Summary up front, details inline.

**The create-side guard is correct and complete.** Both orderings are now closed: a new explicit key vs. an existing id by the new check here, and a new derived slug vs. an existing key by the pre-existing check above it. The validation asymmetry that allowed the collision (the Console path always checked *both* identifier namespaces, the explicit-key path only checked its own) is genuinely fixed, and it covers the Automation/Terraform path — worth stating in the PR description, since it isn't obvious that a `DictionaryServiceImpl` change fixes the Automation API.

**The gateway side fixes one of the two directions the ticket reports.** The ticket's own reproduction (steps 4-5, undeploying the *colliding* dictionary) still ends in the reported 500, and the new test suite asserts that outcome as expected behaviour. Detail in the comment on `should_empty_slot_when_colliding_occupant_is_undeployed`.

I ran the 8 new gateway tests against both revisions (transcribed, then confirmed by compiling the real classes): **8/8 green on this branch, 7/8 on master**, with `should_keep_occupant_when_undeploy_targets_a_different_dictionary` the only RED, failing with `Expecting actual not to be null`. So the behaviour change is genuinely pinned — by exactly one test. The other 7 are characterization tests of pre-existing behaviour, which are worth keeping on their own merit given this module previously had none.

Inline items, in priority order:
1. The unfixed undeploy direction, ratified by a test *(blocking — needs a design decision)*
2. Collision is unlogged in the more dangerous direction
3. `log.debug` where the PR description says "logs a warning"
4. Rejection message describes a conflict that doesn't exist
5. Dead code left by the log-argument change
6. `undeploy` gained a non-atomic check-then-act under a parallel deployer
7. The new guard duplicates the one 15 lines above it
8. Two test-quality items

For the record, some things I checked that are **not** problems, so they don't get re-raised: existing Terraform state does not break (the Automation API routes through `findById` -> `update` and never reaches the new guard, so no `BREAKING CHANGE` footer is needed); there is no reverse/symmetric collision (a dictionary with a non-null key is never indexed by its id); `update()` needs no equivalent guard (key and id are immutable after create); the em-dash in the new log message matches existing style in this module tree; and switching the undeploy log from `removed` to `dictionary` is correct — the ticket explicitly asked for it, and all four `@ToString.Include` fields are present on the event payload.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/test/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManagerTest.java
arpzilla-dev: Done 

Went with reject colliding deploy (keep the occupant, log an error, do not take the slot). That matches the create-side rejection and the ticket.

Deploying a dictionary whose runtime key is already held by a different id is now a no-op on the gateway. Undeploying the rejected dictionary leaves the occupant in place, so #dictionaries['…'] no longer 500s.

The old test that asserted the emptied slot is gone. Coverage is now: occupant kept on colliding deploy, occupant kept when the rejected dictionary is undeployed, and two non-colliding dictionaries in one env.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
arpzilla-dev: Done. 
Colliding deploy is rejected up front (regardless of deployedAt) and logged at ERROR with both ids, names, runtime key, and environment. The incoming dictionary never takes the slot, so there is no silent overwrite.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
arpzilla-dev: Kept as DEBUG on purpose. Dictionary unpublish is replayed every 5s for ~30s, so WARN would spam at default INFO. 

The PR description is updated to match: skip-undeploy is DEBUG;

arpzilla-dev: Kept as DEBUG on purpose. Dictionary unpublish is replayed every 5s for ~30s, so WARN would spam at default INFO. 

The PR description is updated to match: skip-undeploy is DEBUG;

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
arpzilla-dev: Removed. After the occupant-null / id-mismatch returns, remove(key) is unconditional.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
arpzilla-dev: Agreed — not fixing in this PR. The unsynchronised HashMap and live EL map were already there; this change only adds an id check before remove. 

Let me know, should i create follow up ticket here ? for making the dictionary maps thread-safe.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/dictionary/DictionaryServiceImpl.java
arpzilla-dev: Done. New DictionaryKeyCollidesWithIdException, technical code dictionary.keyCollidesWithId. The 400 names the existing dictionary and the colliding key, not the new Terraform name.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/dictionary/DictionaryServiceImpl.java
arpzilla-dev: Extracted to ensureRuntimeKeyIsAvailable, with a null-safe environmentId filter.

Explicit getId() on create still unvalidated (duplicate id is a 500). Leaving that as its own ticket, as you suggested.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/test/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManagerTest.java
arpzilla-dev: Switched to isNullOrEmpty(). Added should_keep_the_other_dictionary_when_one_of_two_is_undeployed. Also added the occupant-kept cases for colliding deploy/undeploy.

Skipped MemoryAppender in this PR (optional). Cross-environment undeploy is still only covered on the deploy side.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/configuration/dictionary/DictionaryServiceImpl_CreateTest.java
arpzilla-dev: removed shouldNotCreateWhenIdAlreadyExistsInSameEnvironment. The real id-vs-key case is shouldNotCreateWhenExplicitKeyMatchesExistingIdInSameEnvironment.

gravitee-apim-gateway/gravitee-apim-gateway-dictionary/src/main/java/io/gravitee/gateway/dictionary/MultiEnvironmentDictionaryManager.java
```diff
@@ -55,30 +67,49 @@ public void deploy(Dictionary dictionary) {
...
+                log.debug(
+                    "Dictionary {} was requested for undeploy but runtime key '{}' is occupied by dictionary {} — ignoring",
+                    dictionary.getId(),
+                    key,
+                    occupant.getId()
+                );
+                return;
+            }
 
+            envDictionaries.remove(key);
```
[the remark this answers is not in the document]
arpzilla-dev: Same as the earlier comment: we did not introduce this HashMap. Pre-existing, out of scope here. Follow-up for thread-safe dictionary maps.

wbabyte (approved):

PR #19164, opened 2026-08-14
feat(gamma): add members to a group from user search

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/FOUND-106

## Description

Adds **user-search add members** to the Gamma group detail page (classic Console parity for “Add Member to Group”).

Admins can open **Add members**, search platform users, assign default roles across API / API Product / Application / Integration / Cluster / **Explorer**, and add one or more users in bulk.

### What to review

**Group detail — Members**

- New **Add members** button opens `GroupAddMembersSheet`
- Gating: `environment-group-u` **or** `manageable`, plus `system_invitation`; hidden when `max_invitation` is reached
- Info banner when the max-members limit is hit
- Empty-state copy nudges toward **Add members** when the user can add
- Email invitation is out of scope here; it lands in #19165

**Add members sheet**

- User search (≥ 2 chars, 300ms debounce), multi-select, existing members excluded
- Search is disabled while the sheet is closed (`enabled: open && …`)
- Default role selects with lock enforcement (`lock_*` + override via env update permission)
- Six scopes, including **EXPLORER** (classic `roleService.list('EXPLORER')` parity)
- Edit group fetches only API / application / API product catalogs; Integration / Cluster / Explorer load only when Add members is open
- `PRIMARY_OWNER` fail-closed until portal settings load; disabled for a scope that already has one
- Bulk submit via `POST …/configuration/groups/{id}/members`
- Success / error toasts; sheet stays open on failure; form locked while saving

**Shared user search**

- Org-scoped `/search/users`, `SearchableUser`, and `isSameUser` live in `shared/` — not on the group or application service

**Org settings — require a group on applications** (also in this PR)

- `GroupsRequireGroupSetting.tsx` on the groups list: org-level “application must have a group” toggle (classic parity)
- `saveOrgConsoleSettings` (`POST /organizations/{orgId}/console`) — full document, not a partial merge
- `setSettings` / `useSetConsoleSettings` on the shared `ConsoleSettingsProvider` so the rest of the session sees the save without a reload
- Toggle and Save gated on `organization-settings-u` (read-only org-settings users cannot save)

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupAddMembersSheet.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupAddMembersSheet.tsx
@@ -1,0 +1,157 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+    Button,
+    Label,
+    ScrollArea,
+    Sheet,
+    SheetContent,
+    SheetDescription,
+    SheetFooter,
+    SheetHeader,
+    SheetTitle,
+} from '@gravitee/graphene-core';
+
+import { GroupScopedRoleSelects } from './GroupScopedRoleSelects';
+import { GroupUserSearchPicker } from './GroupUserSearchPicker';
+import { useGroupAddMembersForm } from '../hooks/useGroupAddMembersForm';
+import type { GroupMember, GroupMembershipPayload, GroupRole } from '../types/group';
+
+export function GroupAddMembersSheet({
+    open,
+    groupName,
+    groupRoles,
+    members,
+    apiRoles,
+    applicationRoles,
+    apiProductRoles,
+    integrationRoles,
+    clusterRoles,
+    explorerRoles,
+    lockApiRole,
+    lockApiProductRole,
+    lockApplicationRole,
+    canOverrideLocks,
+    maxInvitation,
+    apiPrimaryOwnerMode,
+    apiProductPrimaryOwnerMode,
+    onClose,
+    onSubmit,
+    isSaving,
+}: Readonly<{
+    open: boolean;
+    groupName: string;
+    groupRoles: Record<string, string> | undefined;
+    members: GroupMember[];
+    apiRoles: GroupRole[];
+    applicationRoles: GroupRole[];
+    apiProductRoles: GroupRole[];
+    integrationRoles: GroupRole[];
+    clusterRoles: GroupRole[];
+    explorerRoles: GroupRole[];
+    lockApiRole: boolean;
+    lockApiProductRole: boolean;
+    lockApplicationRole: boolean;
+    canOverrideLocks: boolean;
+    maxInvitation: number | null;
+    apiPrimaryOwnerMode?: string;
+    apiProductPrimaryOwnerMode?: string;
+    onClose: () => void;
+    onSubmit: (memberships: GroupMembershipPayload[]) => void;
+    isSaving: boolean;
+}>) {
+    const form = useGroupAddMembersForm({
+        open,
+        groupRoles,
+        members,
+        lockApiRole,
+        lockApiProductRole,
+        lockApplicationRole,
+        canOverrideLocks,
+        maxInvitation,
+        apiPrimaryOwnerMode,
+        apiProductPrimaryOwnerMode,
+        onSubmit,
+    });
+
+    return (
+        <Sheet open={open} onOpenChange={isOpen => !isOpen && !isSaving && onClose()}>
+            <SheetContent side="right" className="flex max-h-full flex-col" style={{ maxWidth: '480px' }}>
+                <SheetHeader>
+                    <SheetTitle>Add members</SheetTitle>
+                    <SheetDescription>Search platform users and assign roles for membership in {groupName}.</SheetDescription>
+                </SheetHeader>
+
+                <ScrollArea className="min-h-0 flex-1">
+                    <div className="space-y-6 px-4 pb-4">
+                        <div className="space-y-3">
+                            <Label className="text-sm font-medium">Default roles for selected users</Label>
+                            <GroupScopedRoleSelects
+                                idPrefix="add-members-role"
+                                roles={{
+                                    api: apiRoles,
+                                    apiProduct: apiProductRoles,
+                                    application: applicationRoles,
+                                    integration: integrationRoles,
+                                    cluster: clusterRoles,
+                                    explorer: explorerRoles,
+                                }}
+                                values={form.roleValues}
+                                onChange={form.handleRoleChange}
+                                locks={form.roleLocks}
+                                disabled={isSaving}
+                                disabledOptionNames={form.disabledOptionNames}
+                            />
+                        </div>
+
+                        {form.primaryOwnerSelected && (
+                            <p className="text-xs text-muted-foreground">The primary owner role can be granted to a single member only.</p>
+                        )}
+
+                        <GroupUserSearchPicker
+                            search={form.search}
+                            onSearchChange={form.setSearch}
+                            debouncedQuery={form.debouncedQuery}
+                            isFetching={form.isFetching}
+                            candidates={form.candidates}
+                            selected={form.selected}
+                            onToggle={form.handleToggle}
+                            invitationLimitReached={form.invitationLimitReached}
+                            groupMemberCapReached={form.groupMemberCapReached}
+                            disabled={isSaving}
+                        />
+
+                        {form.selected.length > 0 && (
+                            <p className="text-xs text-muted-foreground">
+                                {form.selected.length} user{form.selected.length !== 1 ? 's' : ''} selected
+                            </p>
+                        )}
+                    </div>
+                </ScrollArea>
+
+                <SheetFooter className="shrink-0 flex-row justify-end border-t">
+                    <Button type="button" variant="outline" onClick={onClose} disabled={isSaving}>
+                        Cancel
+                    </Button>
+                    <Button type="button" onClick={form.handleSubmit} disabled={!form.canSubmit || isSaving}>
+                        {isSaving ? 'Adding…' : form.submitLabel}
+                    </Button>
+                </SheetFooter>
+            </SheetContent>
+        </Sheet>
+    );
+}
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupRoleSelect.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupRoleSelect.tsx
@@ -1,0 +1,87 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Label, Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '@gravitee/graphene-core';
+
+import type { GroupRole } from '../types/group';
+import { PRIMARY_OWNER_ROLE } from '../types/group';
+
+const NO_ROLE_VALUE = '__none__';
+
+function isRoleOptionDisabled(
+    role: GroupRole,
+    disabledOptionNames: Set<string> | undefined,
+    disableSystemRoles: 'all' | 'except-primary-owner' | undefined,
+): boolean {
+    if (disabledOptionNames?.has(role.name)) {
+        return true;
+    }
+    if (!role.system || !disableSystemRoles) {
+        return false;
+    }
+    if (disableSystemRoles === 'except-primary-owner' && role.name === PRIMARY_OWNER_ROLE) {
+        return false;
+    }
+    return true;
+}
+
+export function GroupRoleSelect({
+    id,
+    label,
+    roles,
+    value,
+    onChange,
+    disabled,
+    disabledOptionNames,
+    disableSystemRoles,
+    hint,
+}: Readonly<{
+    id: string;
+    label: string;
+    roles: GroupRole[];
+    value: string;
+    onChange: (value: string) => void;
+    disabled?: boolean;
+    disabledOptionNames?: Set<string>;
+    /** Classic add-members: Application/Integration/Cluster disable all system roles; API/API Product keep PRIMARY_OWNER selectable under its own rules. */
+    disableSystemRoles?: 'all' | 'except-primary-owner';
+    hint?: string;
+}>) {
+    return (
+        <div className="space-y-1.5">
+            <Label htmlFor={id} className="text-sm text-muted-foreground">
+                {label}
+            </Label>
+            <Select value={value || NO_ROLE_VALUE} onValueChange={v => onChange(v === NO_ROLE_VALUE ? '' : v)} disabled={disabled}>
+                <SelectTrigger id={id} className="w-full">
+                    <SelectValue />
+                </SelectTrigger>
+                <SelectContent>
+                    {roles.map(role => (
+                        <SelectItem
+                            key={role.name}
+                            value={role.name}
+                            disabled={isRoleOptionDisabled(role, disabledOptionNames, disableSystemRoles)}
+                        >
+                            {role.name}
+                        </SelectItem>
+                    ))}
+                </SelectContent>
+            </Select>
+            {hint && <p className="text-xs text-muted-foreground">{hint}</p>}
+        </div>
+    );
+}
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx
@@ -1,0 +1,129 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Badge, Button, Checkbox, Input, Label, Skeleton } from '@gravitee/graphene-core';
+import { SearchIcon, XIcon } from '@gravitee/graphene-core/icons';
+
+import type { SearchableUser } from '../../../shared/types/userSearch';
+import { isSameUser } from '../../../shared/utils/userSearch';
+import { UserAvatar } from '../../users/components/UserAvatar';
+
+const SEARCH_INPUT_ID = 'group-add-members-search';
+
+export function GroupUserSearchPicker({
+    search,
+    onSearchChange,
+    debouncedQuery,
+    isFetching,
+    candidates,
+    selected,
+    onToggle,
+    invitationLimitReached,
+    groupMemberCapReached = false,
+    disabled = false,
+}: Readonly<{
+    search: string;
+    onSearchChange: (value: string) => void;
+    debouncedQuery: string;
+    isFetching: boolean;
+    candidates: SearchableUser[];
+    selected: SearchableUser[];
+    onToggle: (user: SearchableUser) => void;
+    invitationLimitReached: boolean;
+    /** True when existing members alone already hit max_invitation (not just the current selection). */
+    groupMemberCapReached?: boolean;
+    disabled?: boolean;
+}>) {
+    const searchDisabled = disabled || invitationLimitReached;
+
+    return (
+        <div className="space-y-2">
+            <Label htmlFor={SEARCH_INPUT_ID} className="text-sm font-medium">
+                Search users
+            </Label>
+            <div className="relative">
+                <SearchIcon className="absolute left-3 top-1/2 -translate-y-1/2 size-4 text-muted-foreground pointer-events-none" />
+                <Input
+                    id={SEARCH_INPUT_ID}
+                    className="pl-10"
+                    placeholder="Search by name or email…"
+                    value={search}
+                    onChange={e => onSearchChange(e.target.value)}
+                    disabled={searchDisabled}
+                />
+            </div>
+
+            {selected.length > 0 && (
+                <div className="flex flex-wrap gap-1.5" aria-label="Selected users">
+                    {selected.map(user => (
+                        <Badge key={user.id ?? user.reference} variant="secondary" className="gap-0.5 pr-1 font-normal">
+                            {user.displayName}
+                            <Button
+                                type="button"
+                                variant="ghost"
+                                size="icon-xs"
+                                className="ml-0.5 shrink-0 hover:text-destructive"
+                                onClick={() => onToggle(user)}
+                                aria-label={`Remove ${user.displayName}`}
+                                disabled={disabled}
+                            >
+                                <XIcon className="size-3" aria-hidden />
+                            </Button>
+                        </Badge>
+                    ))}
+                </div>
+            )}
+
+            {invitationLimitReached ? (
+                <p className="px-1 py-2 text-sm text-muted-foreground">
+                    {groupMemberCapReached
+                        ? 'This group has reached its maximum number of members.'
+                        : 'Selection limit reached for this group.'}
+                </p>
+            ) : debouncedQuery.trim().length < 2 ? (
+                <p className="px-1 py-2 text-sm text-muted-foreground">Type at least 2 characters to search for users.</p>
+            ) : isFetching || search !== debouncedQuery ? (
+                <div className="space-y-2 p-1">
+                    <Skeleton className="h-10 rounded" />
+                    <Skeleton className="h-10 rounded" />
+                </div>
+            ) : candidates.length === 0 ? (
+                <p className="px-1 py-2 text-sm text-muted-foreground">No users found.</p>
+            ) : (
+                <div className="rounded-lg border">
+                    {candidates.map(user => {
+                        const isChecked = selected.some(u => isSameUser(u, user));
+                        const inputId = `add-member-${user.id ?? user.reference}`;
+                        return (
+                            <label
+                                key={user.id ?? user.reference}
+                                htmlFor={inputId}
+                                className={`flex items-center gap-3 border-b px-3 py-2.5 last:border-b-0 ${disabled ? 'opacity-50' : 'hover:bg-muted/50 cursor-pointer'}`}
+                            >
+                                <Checkbox id={inputId} checked={isChecked} onCheckedChange={() => onToggle(user)} disabled={disabled} />
+                                <UserAvatar name={user.displayName} />
+                                <div className="min-w-0 flex-1">
+                                    <p className="text-sm font-medium truncate">{user.displayName}</p>
+                                    {user.email ? <p className="text-xs text-muted-foreground truncate">{user.email}</p> : null}
+                                </div>
+                            </label>
+                        );
+                    })}
+                </div>
+            )}
+        </div>
+    );
+}
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts
@@ -1,0 +1,157 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useQuery } from '@tanstack/react-query';
+import { useEffect, useMemo, useState } from 'react';
+
+import { searchUsers } from '../../../shared/services/userSearch';
+import type { SearchableUser } from '../../../shared/types/userSearch';
+import type { GroupMember, GroupMembershipPayload } from '../types/group';
+import { PRIMARY_OWNER_ROLE } from '../types/group';
+import { buildMembershipRoles, getMemberRoleLockFlags, type RoleField } from '../utils/memberRoles';
+import { GROUP_SEARCH_DEBOUNCE_MS } from '../utils/paginationConstants';
+import { groupKeys } from '../utils/queryKeys';
+import { nextSearchableUserSelection } from '../utils/searchableUsers';
+
+const PRIMARY_OWNER_MODE_USER = 'USER';
+
+/** Fail closed: until the mode is known and is not USER, PRIMARY_OWNER must stay unavailable. */
+function isPrimaryOwnerUnavailable(mode: string | undefined): boolean {
+    return mode === undefined || mode.toUpperCase() === PRIMARY_OWNER_MODE_USER;
+}
+
+const DEFAULT_USER_ROLES: Record<RoleField, string> = {
+    apiRole: 'USER',
+    apiProductRole: 'USER',
+    applicationRole: 'USER',
+    integrationRole: 'USER',
+    clusterRole: 'USER',
+    explorerRole: 'USER',
+};
+
+export function useGroupAddMembersForm({
+    open,
+    groupRoles,
+    members,
+    lockApiRole,
+    lockApiProductRole,
+    lockApplicationRole,
+    canOverrideLocks,
+    maxInvitation,
+    apiPrimaryOwnerMode,
+    apiProductPrimaryOwnerMode,
+    onSubmit,
+}: {
+    open: boolean;
+    groupRoles: Record<string, string> | undefined;
+    members: GroupMember[];
+    lockApiRole: boolean;
+    lockApiProductRole: boolean;
+    lockApplicationRole: boolean;
+    canOverrideLocks: boolean;
+    maxInvitation: number | null;
+    apiPrimaryOwnerMode: string | undefined;
+    apiProductPrimaryOwnerMode: string | undefined;
+    onSubmit: (memberships: GroupMembershipPayload[]) => void;
+}) {
+    const [search, setSearch] = useState('');
+    const [debouncedQuery, setDebouncedQuery] = useState('');
+    const [selected, setSelected] = useState<SearchableUser[]>([]);
+    const [roleValues, setRoleValues] = useState(DEFAULT_USER_ROLES);
+
+    useEffect(() => {
+        if (!open) return;
+        setSearch('');
+        setDebouncedQuery('');
+        setSelected([]);
+        setRoleValues({
+            apiRole: groupRoles?.API ?? 'USER',
+            apiProductRole: groupRoles?.API_PRODUCT ?? 'USER',
+            applicationRole: groupRoles?.APPLICATION ?? 'USER',
+            integrationRole: 'USER',
+            clusterRole: 'USER',
+            explorerRole: 'USER',
+        });
+        // eslint-disable-next-line react-hooks/exhaustive-deps -- intentionally keyed on the open transition only
+    }, [open]);
+
+    useEffect(() => {
+        if (!open) return;
+        const timer = setTimeout(() => setDebouncedQuery(search), GROUP_SEARCH_DEBOUNCE_MS);
+        return () => clearTimeout(timer);
+    }, [open, search]);
+
+    const roleLocks = getMemberRoleLockFlags({ lockApiRole, lockApiProductRole, lockApplicationRole }, canOverrideLocks);
+    const groupMemberCapReached = typeof maxInvitation === 'number' && maxInvitation <= members.length;
+    const invitationLimitReached = typeof maxInvitation === 'number' && maxInvitation <= members.length + selected.length;
+
+    const { data: results, isFetching } = useQuery({
+        queryKey: groupKeys.userSearch(debouncedQuery),
+        queryFn: () => searchUsers(debouncedQuery),
+        enabled: open && debouncedQuery.trim().length >= 2,
+        staleTime: 30_000,
+    });
+
+    const existingMemberIds = useMemo(() => new Set(members.map(m => m.id)), [members]);
+    const candidates = useMemo(
+        () => (results ?? []).filter(u => !existingMemberIds.has(u.id ?? u.reference)),
+        [results, existingMemberIds],
+    );
+
+    const apiPrimaryOwnerDisabled =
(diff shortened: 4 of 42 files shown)


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts
```diff
@@ -0,0 +1,118 @@
...
+        search,
+        setSearch,
+        selected,
+        roleValues,
+        roleLocks,
+        invitationLimitReached,
+        deferredQuery,
+        isFetching,
+        candidates,
+        disabledOptionNames: {
```
arpzilla-dev, line 133: **[Score: 80/100]** ⚠️ PRIMARY_OWNER invariants from classic add-members are missing.

Classic `add-members-dialog` does three things this hook does not:

1. Disable `PRIMARY_OWNER` when API / API Product primary-owner mode is `USER`.
2. After one user is selected with `PRIMARY_OWNER`, disable further search so only one PO can be added.
3. Clear the selection when the API / API Product role changes (`onDefaultRoleChange`).

This sheet only disables the option if a current member already holds that role. An admin can pick `PRIMARY_OWNER` and POST several users. `GroupMembersResource.addOrUpdateGroupMember` will write `PRIMARY_OWNER` on every membership and set `apiPrimaryOwner` to the last id — several PO members and a pointer that names only one of them.

Please match classic: read env `primaryOwnerMode`, disable PO in `USER` mode, and allow only one selected user when PO is chosen.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
```diff
@@ -127,6 +144,9 @@ export function GroupDetailPage() {
         );
     }
 
+    const maxInvitationsLimitReached = typeof group.max_invitation === 'number' && group.max_invitation <= members.length;
+    const canAddMembers = (canEdit || canInviteToGroup(group)) && !maxInvitationsLimitReached && Boolean(group.system_invitation);
```
arpzilla-dev: **[Score: 62/100]** 💡 Add members gating is stricter than classic and than the PR description.

Classic shows Add Members when `environment-group-u || (manageable && (system_invitation || email_invitation))`, then disables User Search separately if `!system_invitation`.

This line requires `system_invitation` for everyone, including env-group-u. That matches a user-search-only sheet, but:

- The PR says gating is `environment-group-u` **or** `manageable` + `system_invitation`.
- A group with only `email_invitation` loses the Add members button entirely.

If email invite is a follow-up, keep the button visibility on classic `canAddMembers` and only hide/disable the search action when `!system_invitation`.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx
```diff
@@ -0,0 +1,92 @@
...
+                <Input
+                    className="pl-10"
+                    placeholder="Search by name or email…"
+                    value={search}
+                    onChange={e => onSearchChange(e.target.value)}
+                    disabled={invitationLimitReached}
+                />
+            </div>
+
+            {invitationLimitReached ? (
```
arpzilla-dev, line 90: **[Score: 58/100]** 💡 Hitting the member cap hides selected users, so they cannot be deselected.

When `members + selected === maxInvitation`, this branch replaces the result list with a message and disables search. There are no chips (classic has removable `mat-chip`s). The last selected user disappears from the UI; Cancel is the only undo.

Keep selected users visible (chips or a selected list) and only disable *adding*. Classic: `mat-chip-set` + `deselectUser`, while search is disabled.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupAddMembersSheet.tsx
```diff
@@ -0,0 +1,140 @@
...
+                            </p>
+                        )}
+                    </div>
+                </ScrollArea>
+
+                <SheetFooter className="shrink-0 flex-row justify-end border-t">
+                    <Button type="button" variant="outline" onClick={onClose} disabled={isSaving}>
+                        Cancel
+                    </Button>
+                    <Button type="button" onClick={form.handleSubmit} disabled={!form.canSubmit || isSaving}>
```
arpzilla-dev, line 150: **[Score: 55/100]** 💡 Sheet stays editable while the add-members mutation is in flight.

`isSaving` only disables the footer buttons. Sibling `AddUserGroupSheet` disables fields with `isPending`. The user can still change selection and roles during the POST.

Pass `isSaving` into `GroupUserSearchPicker` and `GroupScopedRoleSelects` (and disable Cancel-via-overlay if you want the mutation to finish without a stale close).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupRoleSelect.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return (
+        <div className="space-y-1.5">
+            <Label className="text-sm text-muted-foreground">{label}</Label>
+            <Select value={value || NO_ROLE_VALUE} onValueChange={v => onChange(v === NO_ROLE_VALUE ? '' : v)} disabled={disabled}>
+                <SelectTrigger className="w-full">
+                    <SelectValue />
+                </SelectTrigger>
+                <SelectContent>
+                    {roles.map(role => (
+                        <SelectItem key={role.name} value={role.name} disabled={disabledOptionNames?.has(role.name)}>
```
arpzilla-dev: **[Score: 52/100]** 💡 System roles stay selectable on Application / Integration / Cluster (and other system roles on API / API Product).

Classic add-members disables `role.system` on Application / Integration / Cluster, and on API / API Product disables system roles other than `PRIMARY_OWNER` (`isSystemRoleDisabled`).

This select only honors `disabledOptionNames`, which today is only `PRIMARY_OWNER` when a member already holds it. A group admin can pick other system roles that classic never allows.

Disable options where `role.system === true`, except `PRIMARY_OWNER` on API / API Product (which has its own rules).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx
```diff
@@ -0,0 +1,92 @@
...
+    deferredQuery: string;
+    isFetching: boolean;
+    candidates: SearchableUser[];
+    selected: SearchableUser[];
+    onToggle: (user: SearchableUser) => void;
+    invitationLimitReached: boolean;
+}>) {
+    return (
+        <div className="space-y-2">
+            <Label className="text-sm font-medium">Search users</Label>
```
arpzilla-dev: **[Score: 48/100]** 💡 Graphene labels are not wired to controls.

`graphene.md` requires `Label` / `FieldLabel` + `htmlFor`. Sibling group/user sheets use `Field` + `FieldLabel`.

This `Label` (“Search users”) has no `htmlFor`, and the `Input` has no `id`. Same issue in `GroupRoleSelect`: the label is not associated with the combobox.

Please pair each label with its control (`htmlFor` + `id`) so the fields are reachable by accessible name.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts
```diff
@@ -0,0 +1,118 @@
...
+            applicationRole: groupRoles?.APPLICATION ?? 'USER',
+            integrationRole: 'USER',
+            clusterRole: 'USER',
+        });
+    }, [open, groupRoles]);
+
+    const roleLocks = getMemberRoleLockFlags({ lockApiRole, lockApiProductRole, lockApplicationRole }, canOverrideLocks);
+    const invitationLimitReached = typeof maxInvitation === 'number' && maxInvitation <= members.length + selected.length;
+
+    const deferredQuery = useDeferredValue(search);
```
arpzilla-dev: **[Score: 42/100]** 💡 User search is not debounced.

Graphene asks for **300ms** on server search. Classic add-members uses `debounceTime(300)`. The groups list already has `GROUP_SEARCH_DEBOUNCE_MS`.

`useDeferredValue` is a render deferral, not a time debounce, so `/search/users` can fire on every keystroke after 2 characters.

Please debounce the query (reuse `GROUP_SEARCH_DEBOUNCE_MS` or 300ms) before hitting `searchUsers`.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts
ankita-gupta21: Done

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
ankita-gupta21: addressed in the stacked invitation [PR #19165](https://github.com/gravitee-io/gravitee-api-management/pull/19165). The Add members button now follows the classic gating logic, while the User search and Email invitation actions are disabled independently based on system_invitation and email_invitation. This allows email-only groups to retain the Add members action without enabling user search.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupAddMembersSheet.tsx
ankita-gupta21: Fixed: Selected users are now always displayed as removable Graphene badges, including when the member limit is reached. Reaching the limit disables only further search/selection; users can still remove a selected member, which immediately re-enables search. Added a regression test covering selection at the limit, removal, and search re-enablement. The fix is also included in stacked PR #19165.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupRoleSelect.tsx
ankita-gupta21: Fixed: While the add-members mutation is in flight, role selects, user search, selection toggles/chips, and overlay dismiss are disabled (same pattern as AddUserGroupSheet). Footer Cancel/Submit remain disabled as before. Also included in #19165.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx
ankita-gupta21: Fixed. Classic parity: Application / Integration / Cluster disable all role.system options; API / API Product disable system roles except PRIMARY_OWNER (still gated by existing primary-owner presence/mode rules). Also included in #19165.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts
ankita-gupta21: Fixed. Search users uses Label htmlFor + matching Input id. Each role select uses Label htmlFor + SelectTrigger id (add-members-role-*). Also included in #19165.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx
ankita-gupta21: addressed in the stacked invitation PR #19165. The Add members button now follows the classic gating logic, while the User search and Email invitation actions are disabled independently based on system_invitation and email_invitation. This allows email-only groups to retain the Add members action without enabling user search.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts
arpzilla-dev: **Follow-up:** addressed on `5acf718`.

USER mode disables `PRIMARY_OWNER`, selection is limited to one user while PO is chosen, and changing the API / API Product role clears the current selection. Covered in `GroupAddMembersSheet.spec.tsx`.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
arpzilla-dev: **Follow-up:** addressed for this PR’s scope on `5acf718`.

`canInviteMembers` vs `canAddMembers` is now split. The Add members button still requires `system_invitation` because user search is the only add path here. Email-only groups still hide the button, which is consistent until email invitation exists.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx
arpzilla-dev: **Follow-up:** addressed on `5acf718`.

Selected users stay visible as removable badges when the cap is hit, and removing one re-enables search.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupAddMembersSheet.tsx
arpzilla-dev: **Follow-up:** addressed on `5acf718`.

Role selects, search, checkboxes, overlay close, and Cancel are locked while `isSaving`.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupRoleSelect.tsx
arpzilla-dev: **Follow-up:** addressed on `5acf718`.

API / API Product keep `PRIMARY_OWNER` selectable under its own rules; Application / Integration / Cluster disable all system roles.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx
arpzilla-dev: **Follow-up:** addressed on `5acf718`.

Search and role selects now pair `Label` with `htmlFor` / `id`.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts
arpzilla-dev: **Follow-up:** still open.

`useDeferredValue(search)` is unchanged, so `/search/users` can still fire on every keystroke after 2 characters. Please debounce with `GROUP_SEARCH_DEBOUNCE_MS` (or 300ms) before calling `searchUsers`.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
```diff
@@ -53,11 +66,19 @@ export function GroupDetailPage() {
...
+    // Role catalogs are only needed when Edit Group or the Add members sheet is open.
+    const rolesNeeded = editOpen || memberSheet !== 'closed';
+    const { data: apiRoles = [], isLoading: apiRolesLoading } = useGroupApiRoles({ enabled: rolesNeeded });
+    const { data: applicationRoles = [], isLoading: applicationRolesLoading } = useGroupApplicationRoles({ enabled: rolesNeeded });
+    const { data: apiProductRoles = [], isLoading: apiProductRolesLoading } = useGroupApiProductRoles({ enabled: rolesNeeded });
+    const { data: integrationRoles = [] } = useGroupIntegrationRoles({ enabled: rolesNeeded });
+    const { data: clusterRoles = [] } = useGroupClusterRoles({ enabled: rolesNeeded });
+
+    // Primary-owner gating only matters while the Add members sheet is open.
+    const { data: environmentSettings } = useEnvironmentSettings({ enabled: memberSheet !== 'closed' });
```
arpzilla-dev: **[Score: 55/100]** 💡 PRIMARY_OWNER mode is fail-open until environment settings load.

`useEnvironmentSettings` is enabled only when the add-members sheet is open. Until that query returns — or if it fails — `apiPrimaryOwnerMode` is `undefined`, and `isPrimaryOwnerModeUser(undefined)` is false, so `PRIMARY_OWNER` stays selectable.

Classic already has a settings snapshot before the dialog opens. If this environment is `USER`, a fast submit on first open can still assign PO.

Please keep PO disabled while settings are loading (or fetch settings with the page, not only when the sheet opens), and fail closed if the settings request errors.

ankita-gupta21: fixed

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts
```diff
@@ -0,0 +1,149 @@
...
+            setSelected(prev => {
+                // A primary owner is a single user, so only one member can be added while it is selected.
+                if (primaryOwnerSelected) {
+                    return prev.some(u => isSameUser(u, user)) ? [] : [user];
+                }
+                return toggleSearchableUser(prev, user);
+            }),
+        handleSubmit: () => {
+            const roles = buildMembershipRoles(roleValues);
+            onSubmit(selected.map(user => ({ id: user.id ?? user.reference, reference: user.reference, roles })));
```
aqibmohammadkhan: SearchableUser.id is nullable precisely because the user may not exist in APIM yet — classic's entity says so outright: "id can be null if user comes from an LDAP". Classic sends id: user.id (possibly undefined) and reference separately.

The backend branches on exactly that (MembershipServiceImpl.findUserFromMembershipMember):

```
if (member.getMemberId() != null) {
    userEntity = userService.findById(executionContext, member.getMemberId());
} else {
    // user is coming from an external system → look up providers, findOrCreateUser
    identityService.findByReference(member.getReference()) ...
}

```
Coercing id to the reference string forces the findById branch → UserNotFoundException, and the provider user is never created. Adding an LDAP/OIDC-sourced user who isn't in APIM yet will fail.

Fix: make GroupMembershipPayload.id optional and pass user.id ?? undefined. The backend's null-id path is what makes reference meaningful at all.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.ts
aqibmohammadkhan, line 1: Wrong environment-settings endpoint → PRIMARY_OWNER permanently disabled for group admins

<p dir="ltr"><span role="button" tabindex="0" class="inline-block rounded-[4px] outline-none hide-focus-ring ring-focus"><code>groups.ts</code></span> uses <code data-epitaxy-inline-code="">/settings</code>; classic's <code data-epitaxy-inline-code="">EnvironmentSettingsService.load()</code> uses <code data-epitaxy-inline-code="">/portal</code>. Those are different resources with different permissions:</p><div dir="ltr" class="overflow-x-auto">
Endpoint | Resource | Permissions
-- | -- | --

…/environments/{env}/portal (classic) | PortalResource.getPortalConfig | none — any authenticated user
…/environments/{env}/settings (this PR) | PortalSettingsResource.getPortalSettings | @Permission(ENVIRONMENT_SETTINGS, READ)

</div><p dir="ltr">Combined with the (correct) fail-closed change made in response to the earlier review:</p><div class="epitaxy-codeblock relative max-w-full w-fit"><div class="relative"><div class="epitaxy-diff rounded-r6 overflow-clip" style=""><pre>function isPrimaryOwnerUnavailable(mode: string | undefined): boolean {
    return mode === undefined || mode.toUpperCase() === PRIMARY_OWNER_MODE_USER;
}</pre></div><div class="pointer-events-none absolute inset-y-0 right-[calc(var(--p4)+var(--p1))]"><div class="pointer-events-auto sticky top-[calc(max(var(--epitaxy-top-scrim-height,0px),var(--fade-top,0px))+var(--p4))] flex gap-[var(--p2)] mt-[var(--p4)]"></div></div></div></div><p dir="ltr">a user who can manage the group but lacks <code data-epitaxy-inline-code="">environment-settings-r</code> gets a 403, <code data-epitaxy-inline-code="">mode</code> stays <code data-epitaxy-inline-code="">undefined</code> forever, and PRIMARY_OWNER is silently unselectable in both API and API Product. Classic doesn't have this problem because <code data-epitaxy-inline-code="">/portal</code> is unguarded. Switching to <code data-epitaxy-inline-code="">/portal</code> fixes it — <code data-epitaxy-inline-code="">PortalConfigEntity</code> carries the same <code data-epitaxy-inline-code="">api</code> / <code data-epitaxy-inline-code="">apiProduct</code> <code data-epitaxy-inline-code="">primaryOwnerMode</code> fields.</p>

aqibmohammadkhan (commented):
<h3 dir="ltr">1. EXPLORER scope is missing from the membership payload</h3><p dir="ltr">Classic's dialog has six selects; this has five. <code data-epitaxy-inline-code="">defaultExplorerRole</code> is rendered unconditionally in <code data-epitaxy-inline-code="">add-members-dialog.component.html</code>, sent as <code data-epitaxy-inline-code="">{ scope: 'EXPLORER' }</code>, and the backend handles <code data-epitaxy-inline-code="">explorerRoleEntity</code> in <code data-epitaxy-inline-code="">GroupMembersResource.addOrUpdateGroupMember</code>. Members added through Gamma will get no EXPLORER role, unlike the same action in classic. Either add the select or state in the PR why EXPLORER is deliberately out of scope — the description lists the five scopes as if that were the complete set.</p><hr><h2 dir="ltr">Reuse — this is the third copy of the user-search stack</h2><p dir="ltr">The <code data-epitaxy-inline-code="">applications/user-permissions</code> feature already has all of this:</p><div dir="ltr" class="overflow-x-auto">
New in this PR | Already exists
-- | --
SearchableUser in types/group.ts | features/applications/types/applicationMembers.types.ts:64
searchUsers() in services/groups.ts | features/applications/services/applicationMembers.ts:130 — byte-identical
isSameUser() in utils/searchableUsers.ts (+ its own spec) | features/applications/components/user-permissions/memberHelpers.ts:35 — same semantics
GroupUserSearchPicker | user-permissions/AddMembersSheet.tsx — same useDeferredValue + ≥2 chars + skeleton + badge-chips structure
UserAvatar usage | MemberAvatar

</div><p dir="ltr">Since <code data-epitaxy-inline-code="">searchUsers</code> hits an org-scoped endpoint that has nothing to do with either applications or groups, it and <code data-epitaxy-inline-code="">SearchableUser</code>/<code data-epitaxy-inline-code="">isSameUser</code> belong in <code data-epitaxy-inline-code="">shared/</code>. Worth doing now rather than after a fourth copy lands.</p><hr><h2 dir="ltr">Undeclared scope: <code data-epitaxy-inline-code="">GroupsRequireGroupSetting</code></h2><p dir="ltr"><span role="button" tabindex="0" class="inline-block rounded-[4px] outline-none hide-focus-ring ring-focus"><code>GroupsRequireGroupSetting.tsx</code></span>, <code data-epitaxy-inline-code="">saveOrgConsoleSettings</code>, and the <code data-epitaxy-inline-code="">setSettings</code> addition to <code data-epitaxy-inline-code="">ConsoleSettingsProvider</code> implement the org-level "require a group on applications" toggle. That's a different feature from FOUND-106 and isn't mentioned anywhere in the PR description. It's competently done and matches classic, but a reviewer scanning the description won't know to look at it. Either split it out or add it to the description.</p><p dir="ltr">One real issue inside it: it's gated on <code data-epitaxy-inline-code="">organization-settings-r</code> while <code data-epitaxy-inline-code="">ConsoleSettingsResource.saveConsoleSettings</code> requires <code data-epitaxy-inline-code="">ORGANIZATION_SETTINGS: CREATE/UPDATE/DELETE</code>. A read-only org-settings user sees an enabled Save and gets a 403 toast. Classic has the same gap, so this is parity, not a regression — but it's cheap to gate the Save on <code data-epitaxy-inline-code="">organization-settings-u</code>.</p><hr><h2 dir="ltr">Comment deletions unrelated to the feature</h2><p dir="ltr">Roughly 30 lines of "why" documentation are removed across <span role="button" tabindex="0" class="inline-block rounded-[4px] outline-none hide-focus-ring ring-focus"><code>types/group.ts</code></span>, <span role="button" tabindex="0" class="inline-block rounded-[4px] outline-none hide-focus-ring ring-focus"><code>groupPayload.ts</code></span>, <span role="button" tabindex="0" class="inline-block rounded-[4px] outline-none hide-focus-ring ring-focus"><code>groupPermissions.ts</code></span>, and both member tables. Some of it is load-bearing — e.g. the <code data-epitaxy-inline-code="">buildRolesMap</code> note explaining that an absent <code data-epitaxy-inline-code="">roles</code> field on the PUT means "no change", so clearing every role must still send <code data-epitaxy-inline-code="">{}</code>. That's exactly the kind of thing someone re-breaks in six months. The <code data-epitaxy-inline-code="">serverSide</code>-is-actually-client-side note on both tables is in the same category. These deletions look incidental; I'd restore them.</p><hr><h2 dir="ltr">Smaller items</h2><ul dir="ltr"><li><strong>Debounce is still open.</strong> <code data-epitaxy-inline-code="">arpzilla-dev</code>'s last comment is accurate — <code data-epitaxy-inline-code="">useDeferredValue</code> is a render deferral, not a time debounce, so <code data-epitaxy-inline-code="">/search/users</code> fires per keystroke past 2 chars. <code data-epitaxy-inline-code="">GROUP_SEARCH_DEBOUNCE_MS</code> already exists in this feature.</li><li><strong><code data-epitaxy-inline-code="">email_invitation</code> is dead in <code data-epitaxy-inline-code="">canInviteToGroup</code>.</strong> <code data-epitaxy-inline-code="">canAddMembers = (canEdit || canInviteToGroup(group)) &amp;&amp; !limitReached &amp;&amp; Boolean(group.system_invitation)</code> — the <code data-epitaxy-inline-code="">system_invitation</code> conjunct makes the <code data-epitaxy-inline-code="">email_invitation</code> branch unreachable. Fine as a forward-looking hook for #19165, but it currently reads as logic that does something it doesn't.</li><li><strong>Misleading limit copy.</strong> "This group has reached its maximum number of members" also shows when it's the <em>current selection</em> that hit the cap (<code data-epitaxy-inline-code="">maxInvitation &lt;= members.length + selected.length</code>), which isn't yet true of the group. Something like "Selection limit reached for this group" would be more accurate in that case.</li><li><strong><code data-epitaxy-inline-code="">useEnvironmentSettings</code> placement and gating.</strong> It lives in <code data-epitaxy-inline-code="">useGroupDetail.ts</code> despite being an environment concern, and runs on every group-detail page load including for users who can never open the sheet. <code data-epitaxy-inline-code="">enabled: canAddMembers</code> would be safe here — the sheet can't open otherwise, so it doesn't reintroduce the fail-open the earlier review flagged.</li><li><strong><code data-epitaxy-inline-code="">groupKeys.userSearch</code> under <code data-epitaxy-inline-code="">groupKeys.all</code></strong> means every group mutation invalidates the user-search cache. Harmless, but the search results aren't group-scoped.</li><li><strong>Form reset deps.</strong> <code data-epitaxy-inline-code="">useEffect(..., [open, groupRoles])</code> resets search/selection/roles. React Query's structural sharing usually preserves <code data-epitaxy-inline-code="">group.roles</code> identity across refetches, so this is unlikely to fire in practice — but keying the reset off the <code data-epitaxy-inline-code="">open</code> transition alone would remove the possibility entirely.</li></ul>

ankita-gupta21:
1. EXPLORER scope

> Deliberately out of scope for FOUND-106. Gamma’s group role catalog doesn’t wire EXPLORER yet (five scopes only). I’ll call that out in the PR description; EXPLORER can land with the Explorer role catalog as a follow-up so we don’t half-add a sixth select without roles/tests.

2. Reuse / shared user-search stack

> Agreed this should live in shared/ — searchUsers is org-scoped and duplicated with applications. Keeping that extract out of this PR to avoid touching the applications permission sheets in the same change; happy to open a follow-up to consolidate SearchableUser / searchUsers / isSameUser (and debounce both call sites).


3. Undeclared GroupsRequireGroupSetting

> Will add it to the PR description. Also gated Save (and the switch) on organization-settings-u so a read-only org-settings user no longer gets an enabled Save → 403.

4. Comment deletions

> Restored the load-bearing notes (buildRolesMap absent-roles semantics, client-side serverSide pagination, version column).

5. Inline: LDAP id

> Fixed — GroupMembershipPayload.id is optional; we only send id when present so the backend takes the reference/provider path.

6. Inline: /settings vs /portal

> Fixed — now uses /portal like classic, so group admins without environment-settings-r still get primaryOwnerMode. Combined with fail-closed PO gating and enabled: canAddMembersPreview.


7. Smaller items

> Debounce, limit copy, settings gating, query keys, and open-only form reset are fixed in 96bcc57. email_invitation dead-branch: intentional for this PR’s user-search-only button; classic split is in #19165 (dropdown + per-item disable).

ankita-gupta21:
### [r3793850282](https://github.com/gravitee-io/gravitee-api-management/pull/19164#discussion_r3793850282) — PRIMARY_OWNER invariants

**Fixed.** Classic rules are in place:

1. Fail-closed when mode is unknown/`USER`: [`useGroupAddMembersForm.ts#L31-L32`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts#L31-L32), applied at [`L116-L118`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts#L116-L118)
2. Single selection while PO is chosen: [`L144-L147`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts#L144-L147)
3. Clear selection on API / API Product role change: [`L138-L141`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts#L138-L141)

---

### [r3793850369](https://github.com/gravitee-io/gravitee-api-management/pull/19164#discussion_r3793850369) — Add members gating stricter than classic

**Addressed for this PR’s scope.** Split into `canInviteMembers` vs `canAddMembers` (search is the only path here):

- [`GroupDetailPage.tsx#L155-L157`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx#L155-L157)

Email-only groups still hide the button until [#19165](https://github.com/gravitee-io/gravitee-api-management/pull/19165) (dropdown + per-item disable).

---

### [r3793850435](https://github.com/gravitee-io/gravitee-api-management/pull/19164#discussion_r3793850435) — Cap hides selected users

**Fixed.** Selected users stay as removable badges; only further search is disabled:

- Badges: [`GroupUserSearchPicker.tsx#L70-L72`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx#L70-L72)
- Cap vs selection copy: [`L93-L94`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx#L93-L94)

---

### [r3793850525](https://github.com/gravitee-io/gravitee-api-management/pull/19164#discussion_r3793850525) — Sheet editable while saving

**Fixed.** Roles, search, and dismiss are locked while `isSaving`:

- Overlay: [`GroupAddMembersSheet.tsx#L90`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupAddMembersSheet.tsx#L90)
- Role selects: [`L113`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupAddMembersSheet.tsx#L113)
- Search picker: [`L132`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupAddMembersSheet.tsx#L132)

---

### [r3793850642](https://github.com/gravitee-io/gravitee-api-management/pull/19164#discussion_r3793850642) — System roles stay selectable

**Fixed.** Classic `disableSystemRoles` rules:

- Logic: [`GroupRoleSelect.tsx#L27-L35`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupRoleSelect.tsx#L27-L35)
- Wired in grid: [`GroupScopedRoleSelects.tsx#L59-L96`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupScopedRoleSelects.tsx#L59-L96)

---

### [r3793850735](https://github.com/gravitee-io/gravitee-api-management/pull/19164#discussion_r3793850735) — Labels not wired

**Fixed.**

- Search: [`GroupUserSearchPicker.tsx#L54-L60`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx#L54-L60)
- Role select: [`GroupRoleSelect.tsx#L65`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupRoleSelect.tsx#L65)

---

### [r3793850839](https://github.com/gravitee-io/gravitee-api-management/pull/19164#discussion_r3793850839) — Debounce

**Fixed.** Real 300ms debounce via `GROUP_SEARCH_DEBOUNCE_MS` (not `useDeferredValue`):

- [`useGroupAddMembersForm.ts#L94`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts#L94)
- Query uses `debouncedQuery`: [`L103-L105`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts#L103-L105)

---

### [r3796160297](https://github.com/gravitee-io/gravitee-api-management/pull/19164#discussion_r3796160297) — PO fail-open until settings load

**Fixed.**

- Fail closed while mode unknown: [`useGroupAddMembersForm.ts#L31-L32`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts#L31-L32)
- Settings gated by `canAddMembersPreview`: [`GroupDetailPage.tsx#L82-L84`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx#L82-L84)
- Loads `/portal` (not `/settings`): [`groups.ts#L106`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.ts#L106)

---

### [r3805641409](https://github.com/gravitee-io/gravitee-api-management/pull/19164#discussion_r3805641409) — LDAP `id` coercion

**Fixed.**

- Optional `id`: [`group.ts#L124-L127`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/types/group.ts#L124-L127)
- Omit `id` when missing: [`useGroupAddMembersForm.ts#L154-L158`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts#L154-L158)

---

### [r3805645163](https://github.com/gravitee-io/gravitee-api-management/pull/19164#discussion_r3805645163) — Wrong `/settings` endpoint

**Fixed.** Uses classic’s unguarded `/portal`:

- [`groups.ts#L98-L106`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.ts#L98-L106)

---

### Review body — smaller / related items ([review](https://github.com/gravitee-io/gravitee-api-management/pull/19164#pullrequestreview-4962966254))
| Item | Status | Link |
|---|---|---|
| Debounce | Fixed | [`useGroupAddMembersForm.ts#L94`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts#L94) |
| Limit copy | Fixed | [`GroupUserSearchPicker.tsx#L93-L94`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupUserSearchPicker.tsx#L93-L94) |
| Settings gating | Fixed | [`GroupDetailPage.tsx#L82-L84`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx#L82-L84) |
| `userSearch` under `groupKeys.all` | Fixed | [`queryKeys.ts#L33-L35`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/queryKeys.ts#L33-L35) |
| Form reset deps | Fixed (`[open]` only) | [`useGroupAddMembersForm.ts#L91`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts#L91) |
| `email_invitation` dead branch | Intentional for this PR | [`GroupDetailPage.tsx#L155-L157`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx#L155-L157) — classic split lands in #19165 |
| Hook placement in `useGroupDetail.ts` | Deferred | Gating fixed; move to env module is follow-up cleanup |
| Comment deletions | Restored | [`groupPayload.ts#L35-L39`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/groupPayload.ts#L35-L39), [`GroupMembersTable.tsx#L135-L138`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupMembersTable.tsx#L135-L138) |
| Save needs `organization-settings-u` | Fixed | [`GroupsRequireGroupSetting.tsx#L29`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupsRequireGroupSetting.tsx#L29), [`L75`](https://github.com/gravitee-io/gravitee-api-management/blob/98aead974c697bad3f017f69a165bb3c5a49772f/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupsRequireGroupSetting.tsx#L75) |
| EXPLORER scope / shared search extract | Out of scope | Documented; follow-ups |

aqibmohammadkhan (approved):

PR #19166, opened 2026-08-14
feat(gateway): add http.pathHandling to normalize or reject non-canonical request paths

ytvnr:
**Issue**

**Description**

The gateway resolves the listener context path against the request path exactly as it arrived, enforces the matched API's plan, then forwards that path upstream without resolving it. Any conforming receiver applies RFC 3986 §5.2.4 and serves a different resource than the one the gateway authorized.

Reproduced on a clean 4.11.6 stack with two v4 APIs sharing an upstream host — `alpha` on `/alpha/api/` with no plan, `beta` on `/beta/api/` behind an api-key plan:

| Request | Credential | Before |
| --- | --- | --- |
| `/beta/api/echo` | none | 401 |
| `/alpha/api/../../beta/api/echo` | **none** | **200, beta's backend served** |
| `/alpha/api/%2e%2e/%2e%2e/beta/api/echo` | **none** | **200, beta's backend served** |

This adds a single gateway-level setting, `http.pathHandling`, applied in `DefaultHttpRequestDispatcher.dispatch()` before the acceptor is resolved — the only point where the routing decision itself can be corrected rather than merely blocked.

| Mode | Behaviour |
| --- | --- |
| `RAW` | Today's behaviour, unchanged. **Default**, so this PR is not a breaking change. |
| `REJECT` | 400 when the path is not already canonical. Nothing is rewritten. |
| `NORMALIZE` | Routes, enforces and forwards on the resolved path. |

**Scope.** Both modes apply to every definition version. `REJECT` by construction, since it answers before the acceptor distinguishes them; `NORMALIZE` because the normalized path is seeded into the request wrapper of the reactive engine and the legacy one alike. Integration tests cover v4, v2 under emulation, and v2 on the legacy engine, in every mode.

**Algorithm.** Derived from Vert.x (`HttpUtils.normalizePath` on 4.x, `RFC3986.normalizePath` on 5.x), carried here with EPL-2.0 / Apache-2.0 attribution — both live in non-public packages and the class moved between the two versions, so there is no public API to delegate to on any branch we maintain. `java.net.URI` was measured and rejected: it leaves `%2e%2e` intact and throws on characters that legitimately appear in a request line.

**One deliberate departure from RFC 3986, and it is the security-relevant one.** A segment may carry parameters after a `;`, and §5.2.4 resolves only the exact strings `.` and `..` — so by the specification, `..;x` is an ordinary segment. The Servlet specification is equally clear that a container strips `;params` from every segment *before* resolving, which is what Tomcat, Jetty and Spring do. The gateway takes the second reading and treats `..;x` as a dot segment, because a gateway cannot know what sits behind it: being stricter than the receiver costs an over-refused request, while being laxer authorises one resource and lets another be served — the escalation this PR closes. **This is the only change that widens what `REJECT` refuses**, so a deployment switching to `REJECT` may now see 400s it did not see on an earlier build of this branch.

**Performance.** A `needsNormalization` pre-check answers in one pass, without allocating, whether normalizing would change anything — so `REJECT` never normalizes and `NORMALIZE` only pays when there is something to resolve. Measured with JMH on a 1000-path corpus at 95/5: **38.8 ns per request** under `NORMALIZE`, 32.0 under `REJECT`, against 0.76 for `RAW`. That is +0.004 % on a 1 ms request. The predicate and `normalize()` are held to each other by a property test over 20 000 generated paths, which also asserts that the normalizer's own output never needs normalizing again — the agreement property alone would be satisfied by a `normalize` that left a `..` behind.

**This PR carries more than the fix.** Four things landed alongside it, each for a reason found while building it, and each in its own commit:

- **The request clock now starts at the entry of `dispatch()`.** It used to be stamped in the wrapper's constructor, so everything the dispatcher did — including normalization — fell outside every latency the gateway reports.
- **Request wrappers are built from an options object** rather than from a growing constructor. Adding a field used to move a `public` constructor on five classes and the `protected` factory that calls them; that is what **silently disabled an override in the debug service**, which kept compiling while no longer overriding anything. All previous constructors are kept, deprecated and delegating.
- **The Helm chart exposes the setting**, in both branches of the gateway configmap, defaulting to `RAW`.
- **A rejected path no longer strands a debug session.** The refusal happens before the acceptor is resolved, so the dispatcher never learns the request was a debug one, the completion processor never runs, and the console spins forever on an event stuck in `DEBUGGING`. The debug port now answers the request through the same chain and then closes the session, showing an empty timeline under a red 400.

**A result worth reading before reviewing.** The product already contains a second normalizer, in `gravitee-policy-resource-filtering`, with different rules and its own on/off switch. An integration test pins what happens when both are in play, on an API that denies `/admin/**` with the policy's own normalization switched off:

- under `RAW`, `/alpha/api/public/../admin/secret` reaches the backend — the reported bug, reproduced **between two of our own components**
- under `NORMALIZE`, the same request is refused with a 403, because the gateway resolved the path before the API was even selected — the policy is not reconfigured and knows nothing about the setting

**Additional context**

ADR, with the full reasoning, the benchmark tables and the review trail: https://gravitee.atlassian.net/wiki/spaces/APIM1/pages/385515526

**What the modes do not protect against.** They assume a receiver that decodes percent sequences once, per RFC 3986 §2.3, and treats only `/` as a separator. Reserved characters are deliberately left encoded, so `/a/..%2f../b` and `/a/%252e%252e/b` are canonical here and are neither resolved nor refused. A receiver configured to decode `%2F` into a separator (nginx with a URI in `proxy_pass`, Apache with `AllowEncodedSlashes On`) or to accept `\` as one resolves paths this normalizer does not. `REJECT` refuses paths that are not canonical; it is not traversal hardening for an arbitrary receiver. This is stated in the class contract so nobody reads the setting as more than it is.

**Branch plan.** This targets `4.11.x` to stay closest to the reporting customer, and deliberately keeps the vendored normalizer. Two follow-ups are agreed: `4.12.x` consumes the class from a released `gravitee-common` and gets the options builder for both request types; `master` does the same and **makes `NORMALIZE` the default**, which is a breaking change to be announced as one. The extraction to `gravitee-common` waits until the review of this branch has settled, so the released class starts from the hardened version rather than needing the same fixes three times.

**Testing.** 369 unit tests in the gateway reactor, 151 in the debug service, 32 in the gateway env module, 42 integration tests on path handling, and four `helm-unittest` cases. Four of the assertions added by this branch were verified by mutation rather than trusted green: breaking either side of the rejection decision fails 6 of 24 cases; reverting the locale fix fails exactly the one test that covers it; re-rendering the Helm key fails the negative assertion; and the property generator now draws from the encodings the threat model names.

**Note for reviewers on JDK 25**: `PathTraversalOrganizationGuardV4IntegrationTest` cannot run locally on a JDK 25 toolchain — the Groovy policy it deploys rejects class file major version 69. All four of its cases fail there, including the one that only serves a regular path. CI runs on JDK 21 and is unaffected.

**Not in this PR**: the documentation, the 4.13 announcement, and surfacing `handlers.rejected.analytics.enabled` in the chart's `values.yaml`. All tracked in the ADR.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/http/vertx/VertxHttpServerRequest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/http/vertx/VertxHttpServerRequest.java
@@ -58,20 +58,28 @@ public class VertxHttpServerRequest implements Request {
 
     private Handler<Long> timeoutHandler;
 
+    /**
+     * @deprecated kept so that anything built against it keeps compiling and keeps working. Prefer
+     *     {@link #VertxHttpServerRequest(HttpServerRequest, IdGenerator, VertxHttpServerRequestOptions)},
+     *     which absorbs new fields without moving this signature.
+     */
+    @Deprecated
     public VertxHttpServerRequest(HttpServerRequest httpServerRequest, IdGenerator idGenerator) {
-        this(httpServerRequest, idGenerator, null);
+        this(httpServerRequest, idGenerator, VertxHttpServerRequestOptions.builder().build());
     }
 
     /**
-     * @param path the path this request reports, so that everything derived from it — starting with
-     *     the {@code pathInfo} a contextualized request computes — matches the path the gateway
-     *     resolved rather than the one received. {@code null} reports the native path, which is the
-     *     historical behaviour. {@link #uri()} reports the untouched native value either way.
+     * @deprecated see {@link #VertxHttpServerRequest(HttpServerRequest, IdGenerator)}.
      */
+    @Deprecated
     public VertxHttpServerRequest(HttpServerRequest httpServerRequest, IdGenerator idGenerator, String path) {
-        this.path = path;
+        this(httpServerRequest, idGenerator, VertxHttpServerRequestOptions.builder().path(path).build());
+    }
+
+    public VertxHttpServerRequest(HttpServerRequest httpServerRequest, IdGenerator idGenerator, VertxHttpServerRequestOptions options) {
+        this.path = options.getPath();
         this.serverRequest = httpServerRequest;
-        this.timestamp = System.currentTimeMillis();
+        this.timestamp = options.getTimestamp() != null ? options.getTimestamp() : System.currentTimeMillis();
         this.id = idGenerator.randomString();
         this.headers = new VertxHttpHeaders(httpServerRequest.headers());
         this.metrics = Metrics.on(timestamp).build();
--- a/gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/http/vertx/VertxHttpServerRequest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/http/vertx/VertxHttpServerRequest.java
@@ -46,6 +46,7 @@ public class VertxHttpServerRequest implements Request {
     private final long timestamp;
 
     protected final HttpServerRequest serverRequest;
+    private String path;
 
     private MultiValueMap<String, String> queryParameters = null;
 
@@ -58,6 +59,17 @@ public class VertxHttpServerRequest implements Request {
     private Handler<Long> timeoutHandler;
 
     public VertxHttpServerRequest(HttpServerRequest httpServerRequest, IdGenerator idGenerator) {
+        this(httpServerRequest, idGenerator, null);
+    }
+
+    /**
+     * @param path the path this request reports, so that everything derived from it — starting with
+     *     the {@code pathInfo} a contextualized request computes — matches the path the gateway
+     *     resolved rather than the one received. {@code null} reports the native path, which is the
+     *     historical behaviour. {@link #uri()} reports the untouched native value either way.
+     */
+    public VertxHttpServerRequest(HttpServerRequest httpServerRequest, IdGenerator idGenerator, String path) {
+        this.path = path;
         this.serverRequest = httpServerRequest;
         this.timestamp = System.currentTimeMillis();
         this.id = idGenerator.randomString();
@@ -89,7 +101,10 @@ public class VertxHttpServerRequest implements Request {
 
     @Override
     public String path() {
-        return serverRequest.path();
+        if (path == null) {
+            path = serverRequest.path();
+        }
+        return path;
     }
 
     @Override
--- a/gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/http/vertx/VertxHttpServerRequest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/http/vertx/VertxHttpServerRequest.java
@@ -68,14 +68,6 @@ public class VertxHttpServerRequest implements Request {
         this(httpServerRequest, idGenerator, VertxHttpServerRequestOptions.builder().build());
     }
 
-    /**
-     * @deprecated see {@link #VertxHttpServerRequest(HttpServerRequest, IdGenerator)}.
-     */
-    @Deprecated
-    public VertxHttpServerRequest(HttpServerRequest httpServerRequest, IdGenerator idGenerator, String path) {
-        this(httpServerRequest, idGenerator, VertxHttpServerRequestOptions.builder().path(path).build());
-    }
-
     public VertxHttpServerRequest(HttpServerRequest httpServerRequest, IdGenerator idGenerator, VertxHttpServerRequestOptions options) {
         this.path = options.getPath();
         this.serverRequest = httpServerRequest;
--- a/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
@@ -20,6 +20,7 @@ import static io.gravitee.gateway.reactive.api.context.InternalContextAttributes
 import static io.gravitee.gateway.reactive.api.context.InternalContextAttributes.ATTR_INTERNAL_TRACING_ERROR;
 import static io.gravitee.gateway.reactive.api.context.InternalContextAttributes.ATTR_INTERNAL_TRACING_ROOT_SPAN;
 
+import io.gravitee.common.http.HttpStatusCode;
 import io.gravitee.common.http.IdGenerator;
 import io.gravitee.common.http.MediaType;
 import io.gravitee.definition.model.ExecutionMode;
@@ -28,6 +29,8 @@ import io.gravitee.gateway.api.handler.Handler;
 import io.gravitee.gateway.core.component.ComponentProvider;
 import io.gravitee.gateway.env.GatewayConfiguration;
 import io.gravitee.gateway.env.RequestClientAuthConfiguration;
+import io.gravitee.gateway.env.RequestPathConfiguration;
+import io.gravitee.gateway.env.RequestPathHandling;
 import io.gravitee.gateway.env.RequestTimeoutConfiguration;
 import io.gravitee.gateway.http.utils.RequestUtils;
 import io.gravitee.gateway.http.vertx.VertxHttp2ServerRequest;
@@ -47,6 +50,7 @@ import io.gravitee.gateway.reactive.core.tracing.TracingHook;
 import io.gravitee.gateway.reactive.http.vertx.ClientCloseClassifier;
 import io.gravitee.gateway.reactive.http.vertx.VertxHttpServerRequest;
 import io.gravitee.gateway.reactive.reactor.handler.HttpAcceptorResolver;
+import io.gravitee.gateway.reactive.reactor.path.RequestPathNormalizer;
 import io.gravitee.gateway.reactive.reactor.processor.DefaultPlatformProcessorChainFactory;
 import io.gravitee.gateway.reactive.reactor.processor.NotFoundProcessorChainFactory;
 import io.gravitee.gateway.reactor.handler.HttpAcceptor;
@@ -97,6 +101,7 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
     private final TracingContext gatewayTracingContext;
     private final RequestTimeoutConfiguration requestTimeoutConfiguration;
     private final RequestClientAuthConfiguration requestClientAuthConfiguration;
+    private final RequestPathConfiguration requestPathConfiguration;
     private final Vertx vertx;
     private final ComponentProvider globalComponentProvider;
     private final TracingHook tracingHook;
@@ -114,6 +119,7 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         TracingContext gatewayTracingContext,
         RequestTimeoutConfiguration requestTimeoutConfiguration,
         RequestClientAuthConfiguration requestClientAuthConfiguration,
+        RequestPathConfiguration requestPathConfiguration,
         Vertx vertx,
         boolean warningsEnabled
     ) {
@@ -128,6 +134,7 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         this.gatewayTracingContext = gatewayTracingContext;
         this.requestTimeoutConfiguration = requestTimeoutConfiguration;
         this.requestClientAuthConfiguration = requestClientAuthConfiguration;
+        this.requestPathConfiguration = requestPathConfiguration;
         this.vertx = vertx;
         this.tracingHook = new TracingHook("Processor chain");
         this.warningsEnabled = warningsEnabled;
@@ -144,9 +151,18 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
      */
     @Override
     public Completable dispatch(HttpServerRequest httpServerRequest, String serverId) {
-        log.debug("Dispatching request on host {} and path {}", httpServerRequest.host(), httpServerRequest.path());
+        final String rawPath = httpServerRequest.path();
+        log.debug("Dispatching request on host {} and path {}", httpServerRequest.host(), rawPath);
 
-        final HttpAcceptor httpAcceptor = httpAcceptorResolver.resolve(httpServerRequest.host(), httpServerRequest.path(), serverId);
+        final String normalizedPath = requestPathConfiguration.isEnabled() ? RequestPathNormalizer.normalize(rawPath) : rawPath;
+
+        // A malformed percent sequence has no normalized form at all, so there is nothing to decide
+        // on and it is refused whatever the mode.
+        if (normalizedPath == null || (normalizedPath != rawPath && requestPathConfiguration.getHandling() == RequestPathHandling.REJECT)) {
+            return handleRejectedPath(httpServerRequest, serverId);
+        }
+
+        final HttpAcceptor httpAcceptor = httpAcceptorResolver.resolve(httpServerRequest.host(), normalizedPath, serverId);
         Context vertxContext = VertxContext.createNewDuplicatedContext(vertx.getOrCreateContext());
         if (httpAcceptor == null || httpAcceptor.reactor() == null) {
             log.debug(
@@ -196,6 +212,12 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         } else if (httpAcceptor.reactor() instanceof ApiReactor<?> apiReactor) {
             log.debug("Request routed to API reactor on path [{}]", httpAcceptor.path());
             MutableExecutionContext mutableCtx = prepareExecutionContext(httpServerRequest, serverId);
+            if (normalizedPath != rawPath) {
+                // Seed the lazily initialized path so that pathInfo, the flow selectors and the
+                // upstream URI are all derived from the value the acceptor actually matched.
+                // uri() and parameters() keep reading the untouched native request.
+                mutableCtx.request().path(normalizedPath);
+            }
             mutableCtx.request().contextPath(httpAcceptor.path());
             markTracingRoute(vertxContext, httpAcceptor.path());
             TracingContext tracingContext = apiReactor.tracingContext();
@@ -256,7 +278,7 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         }
         // V3 execution mode.
         log.debug("Request routed to V3 handler on path [{}]", httpAcceptor.path());
-        return handleV3Request(httpServerRequest, httpAcceptor, vertxContext);
+        return handleV3Request(httpServerRequest, httpAcceptor, vertxContext, normalizedPath == rawPath ? null : normalizedPath);
     }
 
     /**
@@ -326,6 +348,24 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         return context;
     }
 
+    /**
+     * Refuses the request before any API is selected, through a processor chain rather than on the
+     * raw response, so that it is measured and reported like any other request the gateway answers
+     * on its own.
+     */
+    private Completable handleRejectedPath(final HttpServerRequest httpServerRequest, final String serverId) {
+        final MutableExecutionContext ctx = prepareExecutionContext(httpServerRequest, serverId);
+        ctx.request().contextPath("/");
+        final ProcessorChain processorChain = notFoundProcessorChainFactory.rejectedPathProcessorChain();
+        return HookHelper.hook(
+            () -> processorChain.execute(ctx, ExecutionPhase.RESPONSE),
+            processorChain.getId(),
+            List.of(),
+            ctx,
+            ExecutionPhase.RESPONSE
+        );
+    }
+
     private Completable handleNotFound(final MutableExecutionContext ctx, final List<ProcessorHook> notFoundProcessorHook) {
         ctx.request().contextPath("/");
         ProcessorChain processorChain = notFoundProcessorChainFactory.processorChain();
@@ -341,12 +381,15 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
     private Completable handleV3Request(
         final HttpServerRequest httpServerRequest,
         final HttpAcceptor handlerEntrypoint,
-        final Context vertxContext
+        final Context vertxContext,
+        final String normalizedPath
     ) {
         final ReactorHandler reactorHandler = handlerEntrypoint.reactor();
         markTracingRoute(vertxContext, handlerEntrypoint.path());
 
-        io.gravitee.gateway.http.vertx.VertxHttpServerRequest request = createV3Request(httpServerRequest, idGenerator);
+        // The normalized path is handed over at construction: pathInfo, and therefore the upstream
+        // URI, must derive from the path the acceptor matched rather than from the one received.
+        io.gravitee.gateway.http.vertx.VertxHttpServerRequest request = createV3Request(httpServerRequest, idGenerator, normalizedPath);
 
         // Prepare invocation execution context.
         SimpleExecutionContext simpleExecutionContext = createV3ExecutionContext(httpServerRequest, request);
@@ -432,23 +475,31 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         gatewayConfiguration.zone().ifPresent(metrics::setZone);
     }
 
+    /**
+     * The single extension point for building the v3 request: subclasses that decorate it must
+     * override this signature, since it is the one {@link #handleV3Request} calls. A convenience
+     * overload without {@code path} would compile in a subclass and never be invoked.
+     *
+     * @param path the path the acceptor matched, or {@code null} to report the one received
+     */
     protected io.gravitee.gateway.http.vertx.VertxHttpServerRequest createV3Request(
         HttpServerRequest httpServerRequest,
-        IdGenerator idGenerator
+        IdGenerator idGenerator,
+        String path
     ) {
         io.gravitee.gateway.http.vertx.VertxHttpServerRequest request;
 
         if (isV3WebSocket(httpServerRequest)) {
-            request = new VertxWebSocketServerRequest(httpServerRequest.getDelegate(), idGenerator);
+            request = new VertxWebSocketServerRequest(httpServerRequest.getDelegate(), idGenerator, path);
         } else {
             if (httpServerRequest.version() == HttpVersion.HTTP_2) {
                 if (MediaType.APPLICATION_GRPC.equals(httpServerRequest.getHeader(HttpHeaders.CONTENT_TYPE))) {
-                    request = new VertxGrpcServerRequest(httpServerRequest.getDelegate(), idGenerator);
+                    request = new VertxGrpcServerRequest(httpServerRequest.getDelegate(), idGenerator, path);
                 } else {
-                    request = new VertxHttp2ServerRequest(httpServerRequest.getDelegate(), idGenerator);
+                    request = new VertxHttp2ServerRequest(httpServerRequest.getDelegate(), idGenerator, path);
                 }
             } else {
-                request = new io.gravitee.gateway.http.vertx.VertxHttpServerRequest(httpServerRequest.getDelegate(), idGenerator);
+                request = new io.gravitee.gateway.http.vertx.VertxHttpServerRequest(httpServerRequest.getDelegate(), idGenerator, path);
             }
         }
 
--- a/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
@@ -302,13 +302,7 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         }
         // V3 execution mode.
         log.debug("Request routed to V3 handler on path [{}]", httpAcceptor.path());
-        return handleV3Request(
-                httpServerRequest,
-                httpAcceptor,
-                vertxContext,
-                pathWasNormalized ? normalizedPath : null,
-                receivedAt
-            );
+        return handleV3Request(httpServerRequest, httpAcceptor, vertxContext, pathWasNormalized ? normalizedPath : null, receivedAt);
     }
 
     /**
@@ -357,7 +351,12 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         return path.length() > 1 && path.endsWith("/") ? path.substring(0, path.length() - 1) : path;
     }
 
-    private MutableExecutionContext prepareExecutionContext(
+    /**
+     * Visible to subclasses so that a dispatcher serving a single kind of traffic can build the same
+     * context this one builds, rather than a lookalike. The debug dispatcher needs it to answer a
+     * rejected path without losing the debug session — see {@code DebugHttpRequestDispatcher}.
+     */
+    protected MutableExecutionContext prepareExecutionContext(
         final HttpServerRequest httpServerRequest,
         String serverId,
         final String path,
--- a/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
@@ -91,6 +91,7 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
     public static final String ATTR_ENTRYPOINT = ContextAttributes.ATTR_PREFIX + "entrypoint";
     // Key checked by gravitee-node's RouteGetter (OTel span naming) before falling back to the raw request URI.
     private static final String TRACING_ROUTE_CONTEXT_KEY = "VertxRoute";
+
     private final GatewayConfiguration gatewayConfiguration;
     private final HttpAcceptorResolver httpAcceptorResolver;
     private final IdGenerator idGenerator;
@@ -170,6 +171,13 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
             return handleRejectedPath(httpServerRequest, serverId);
         }
 
+        final boolean pathWasNormalized = normalizedPath != rawPath;
+        if (pathWasNormalized) {
+            // Both forms, deliberately: the point of this line is to answer "what did the client
+            // actually send" once the gateway has started deciding on something else.
+            log.debug("Path normalized from [{}] to [{}]", rawPath, normalizedPath);
+        }
+
         final HttpAcceptor httpAcceptor = httpAcceptorResolver.resolve(httpServerRequest.host(), normalizedPath, serverId);
         Context vertxContext = VertxContext.createNewDuplicatedContext(vertx.getOrCreateContext());
         if (httpAcceptor == null || httpAcceptor.reactor() == null) {
@@ -220,11 +228,15 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         } else if (httpAcceptor.reactor() instanceof ApiReactor<?> apiReactor) {
             log.debug("Request routed to API reactor on path [{}]", httpAcceptor.path());
             MutableExecutionContext mutableCtx = prepareExecutionContext(httpServerRequest, serverId);
-            if (normalizedPath != rawPath) {
+            if (pathWasNormalized) {
                 // Seed the lazily initialized path so that pathInfo, the flow selectors and the
                 // upstream URI are all derived from the value the acceptor actually matched.
                 // uri() and parameters() keep reading the untouched native request.
                 mutableCtx.request().path(normalizedPath);
+                // Nothing to record here: MetricsProcessor already reports both forms, since it
+                // builds uri from request.uri(), which keeps reading the native request, and
+                // pathInfo from the value seeded above. Received and routed are both accounted
+                // for without a custom metric.
             }
             mutableCtx.request().contextPath(httpAcceptor.path());
             markTracingRoute(vertxContext, httpAcceptor.path());
--- a/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
@@ -20,7 +20,6 @@ import static io.gravitee.gateway.reactive.api.context.InternalContextAttributes
 import static io.gravitee.gateway.reactive.api.context.InternalContextAttributes.ATTR_INTERNAL_TRACING_ERROR;
 import static io.gravitee.gateway.reactive.api.context.InternalContextAttributes.ATTR_INTERNAL_TRACING_ROOT_SPAN;
 
-import io.gravitee.common.http.HttpStatusCode;
 import io.gravitee.common.http.IdGenerator;
 import io.gravitee.common.http.MediaType;
 import io.gravitee.definition.model.ExecutionMode;
@@ -165,33 +164,45 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         // pays for the resolution when there is something to resolve.
         final boolean needsNormalization = requestPathConfiguration.isEnabled() && RequestPathNormalizer.needsNormalization(rawPath);
 
-        if (needsNormalization && requestPathConfiguration.getHandling() == RequestPathHandling.REJECT) {
-            return handleRejectedPath(httpServerRequest, serverId, receivedAt);
-        }
-
-        final String normalizedPath = needsNormalization ? RequestPathNormalizer.normalize(rawPath) : rawPath;
-
-        // No normalized form at all: the path carries a malformed percent sequence.
-        if (normalizedPath == null) {
-            return handleRejectedPath(httpServerRequest, serverId, receivedAt);
-        }
-
-        final boolean pathWasNormalized = normalizedPath != rawPath;
-        if (pathWasNormalized) {
+        final String normalizedPath;
+        if (needsNormalization) {
+            if (requestPathConfiguration.getHandling() == RequestPathHandling.REJECT) {
+                return handleRejectedPath(httpServerRequest, serverId, receivedAt);
+            }
+            normalizedPath = RequestPathNormalizer.normalize(rawPath);
+            // No normalized form at all: the path carries a malformed percent sequence. Inside this
+            // branch on purpose. Under RAW nothing is inspected, and a null path — which the Vert.x
+            // API declares and an HTTP/2 request without a :path pseudo-header actually produces —
+            // must keep reaching the acceptor exactly as it did before this setting existed.
+            if (normalizedPath == null) {
+                return handleRejectedPath(httpServerRequest, serverId, receivedAt);
+            }
             // Both forms, deliberately: the point of this line is to answer "what did the client
             // actually send" once the gateway has started deciding on something else.
             log.debug("Path normalized from [{}] to [{}]", rawPath, normalizedPath);
+        } else {
+            normalizedPath = rawPath;
         }
+
+        // Exactly needsNormalization: the normalizer answers the same instance if and only if there
+        // was nothing to resolve, and the property test holds the two methods to that. Reusing the
+        // flag rather than comparing references makes that coupling explicit at the call site.
+        final boolean pathWasNormalized = needsNormalization;
 
         final HttpAcceptor httpAcceptor = httpAcceptorResolver.resolve(httpServerRequest.host(), normalizedPath, serverId);
         Context vertxContext = VertxContext.createNewDuplicatedContext(vertx.getOrCreateContext());
         if (httpAcceptor == null || httpAcceptor.reactor() == null) {
-            log.debug(
-                "No acceptor found for host {} and path {}, handling as not found",
-                httpServerRequest.host(),
-                httpServerRequest.path()
+            log.debug("No acceptor found for host {} and path {}, handling as not found", httpServerRequest.host(), normalizedPath);
+            // The resolved path, like the API branch below. The lookup that just failed was made on
+            // it, so an operator investigating a 404 has to see it — reporting the received path
+            // would send them looking for a context path the gateway never tried. Nothing is lost:
+            // uri() still carries the bytes the client sent.
+            MutableExecutionContext mutableCtx = prepareExecutionContext(
+                httpServerRequest,
+                serverId,
+                pathWasNormalized ? normalizedPath : null,
+                receivedAt
             );
-            MutableExecutionContext mutableCtx = prepareExecutionContext(httpServerRequest, serverId, null, receivedAt);
             mutableCtx.tracer(
                 new io.gravitee.gateway.reactive.api.tracing.Tracer(vertxContext, gatewayTracingContext.opentelemetryTracer())
             );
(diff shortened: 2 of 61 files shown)


ytvnr:
## Internal adversarial review, and what it changed

Before opening this for review, the branch was read cold by five independent reviewers, each on a separate angle: the normalizer's security, the dispatcher, the debug service, test quality, and configuration/packaging. None of them shared the authoring context.

They found **two blocking defects and a dozen smaller ones**. All are fixed in `c2654ee`. This comment exists so a reviewer knows what was already attacked, and does not spend their time re-finding it.

---

### 🔴 A traversal that survived the fix

`isDotSegment` only accepted `.` or `..` followed by `/` or end-of-string, so a `;` disqualified the segment:

```
GET /alpha/api/..;/..;/beta/api/secret     (no credential)
```

`needsNormalization` answered `false`. `REJECT` let it through, `NORMALIZE` forwarded it untouched. Tomcat, Jetty and Spring strip `;params` from every segment *before* resolving dot segments, so the backend saw `/beta/api/secret` with alpha's plan enforced — **the reported escalation, still open, against a very common family of backends**.

Worse, the normalizer *produced* that shape: `normalize("/alpha/%2e%2e;/beta")` returned `/alpha/..;/beta`, decoding the dots while leaving the `;`.

Fixed by treating a segment that is entirely dots before the `;` as a dot segment. The reasoning, and the fact that this departs from RFC 3986 on purpose, is in the class contract and in the PR description. **It is the only change here that widens what `REJECT` refuses.**

### 🔴 The drift guard did not exist

The rejection decision is written twice: inlined in `dispatch()` for allocation reasons, and restated in `RequestPathRejection.applies()` for the debug dispatcher. The class javadoc claimed the duplication was "pinned by `RequestPathRejectionTest`, which walks the same table of modes and paths the dispatcher tests use".

That table did not exist. `DefaultHttpRequestDispatcherTest` only ever instantiated the dispatcher with `RAW`, and the test compared `applies()` against the normalizer it is built from — restating its own body, passing by construction.

Replaced by a guard that drives the **dispatcher** over a shared table of 3 modes × 8 paths and asserts it refused exactly what the predicate announces. Verified by mutation: breaking either side fails 6 of 24 cases.

---

### Also fixed

| | |
| --- | --- |
| The debug decorator called the deprecated 2-arg constructor and rebuilt itself from the native request, so a v2 API debugged in V3 execution mode reported the raw path | `VertxHttpServerRequestDebugDecorator` |
| Two different hex parsers backed the two implementations; `Integer.parseInt` accepts a sign, so `%+41` was flagged by one and ignored by the other. `Character.digit` is Unicode-aware, so `%٢e` decoded to a dot no receiver resolves | single ASCII-only helper, shared |
| `toUpperCase()` without a locale: on a Turkish JVM `normalize` uppercases to `NORMALİZE`, `valueOf` throws, and the security control silently turns itself off. `RAW` and `REJECT` carry no `i` and would have kept working | `Locale.ROOT` |
| An unrecognised `http.pathHandling` started the gateway in the permissive mode with only a warning, and had no test at all | covered, and the active mode is now logged in every mode including the default |
| The startup line claimed "dot segments are resolved" even under `REJECT`, which resolves nothing | one sentence per mode |
| The rejected chain's analytics decision was taken from the API's own configuration instead of `handlers.rejected.analytics.enabled`, because the attribute was set before the chain ran | set after |
| `andThen` runs on success only, so a client disconnecting mid-request left the debug event stuck in `DEBUGGING` — the dead end the commit exists to close | `doFinally` |
| A gRPC caller refused under `REJECT` saw `UNKNOWN`; `Content-Length` was counted in characters, so a localised error message truncated the body | `grpc-status` set, length taken from the encoded buffer |
| `preprocessorStep.path` leaked onto the backend response, where the inspector rendered it as a Path a response policy had added | fallback narrowed |

### Test defects, which were the uncomfortable part

Several assertions reported here as coverage could not fail.

- **The fourth `helm-unittest` case was vacuous.** `pattern: |` keeps the trailing newline, so the regex required a newline immediately after the colon — which a rendered key never has. A reviewer put the key back into the render and the suite still reported 4 passed. Rewritten; re-rendering the key now fails it.
- **The debug dispatcher tests stubbed the rejected chain empty and mocked the completion processor**, so double reporting, ordering and the 400 itself were observed by nothing. Added a test across the dispatcher, the real chain and the real processor, asserting the status on the wire and that the event leaves `DEBUGGING`.
- **`REJECT` was never exercised on a V2 definition**, because the shared base fixed the mode to `NORMALIZE`. Lifted into the nested classes; both engines now run it.
- **"Forward on the resolved path" — the headline behaviour — was asserted for no v4 API.** Every traversal under `NORMALIZE` ended 401 or 404 with an empty WireMock, so nothing was ever forwarded. Added a traversal that lands back inside its own context path, where the upstream URL is observable.
- **`RejectedPathReportingIntegrationTest` raced.** It asserted on the report as soon as the 400 arrived, while `ReporterProcessor` runs after the response is ended. Now uses the subject-and-await pattern the rest of the repo uses.
- **The property generator drew from a closed vocabulary** written from what the implementation handles, so it could not produce `%+41`. Widened with `%252e`, `%c0%ae`, `%00`, `%5c`, `%+41`, `%٢e`, `..;`. A second property was added: the normalizer's output must never need normalizing again — the agreement property alone is satisfied by a `normalize` that leaves a `..` behind.

---

### What the reviewers checked and found sound

Worth stating, since it is where a reviewer might otherwise start.

- **No fail-open on the structural rules.** The normalizer was compiled standalone and brute-forced over `{/ . % 2 e a}` exhaustively to 8 characters, plus 3M random strings on a wider alphabet. Zero cases where `needsNormalization` answers `false` while `normalize` would change the path. `null` is handled at every call site.
- **The path each consumer sees** was traced for all three modes across v4, v2-under-emulation and legacy V3: routing, plan enforcement and the upstream URI all take the resolved path; `uri()` keeps the received one.
- **Widening `prepareExecutionContext` to `protected` is safe** — `createExecutionContext` was already protected and already allowed as much.
- **`Content-Length` cannot conflict with chunked encoding**: the response layer removes `Transfer-Encoding` whenever the length is set.
- **The clock move breaks no timeout budget.**
- **The Helm chart renders in both configmap branches**, proven by mutating the template.
- **The `captureInitialPath` window holds on every engine**, including legacy V3, and the non-final field carries no race.

### Known limitations, stated rather than discovered

- The modes assume a receiver that decodes once and treats only `/` as a separator. `..%2f..`, `%252e%252e`, `%c0%ae` and `%00` are canonical here and are neither resolved nor refused. `REJECT` refuses non-canonical paths; it is not traversal hardening for an arbitrary receiver.
- A rejected request is reported to the configured reporter but **is not visible in the Console**: both the legacy platform logs and the new logs explorer scope their query to concrete API ids, and the unknown-service id `"1"` is never among them. Pre-existing, shared with not-found, out of scope here.
- `PathTraversalOrganizationGuardV4IntegrationTest` cannot run on a JDK 25 toolchain — the Groovy policy rejects class file major version 69. All four cases fail locally, including the one that only serves a regular path. CI on JDK 21 is unaffected.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/path/RequestPathNormalizer.java
```diff
@@ -0,0 +1,417 @@
...
+     * mistake is a security defect, which is a poor trade at this size. Revisit if that cap is ever
+     * raised.
+     */
+    private static void decodeUnreservedChars(final StringBuilder path, int start) {
+        while (start < path.length()) {
+            if (path.charAt(start) == '%') {
+                decodeUnreserved(path, start);
+            }
+            start++;
+        }
```
ytvnr: Applied in 814be0f0ce: `decodeUnreservedChars` now walks a local `cursor` and its `start` parameter is `final`. Same treatment as `removeDotSegments` in the same commit.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/path/RequestPathNormalizer.java
```diff
@@ -0,0 +1,403 @@
...
+                    if (out.length() == 0 || out.charAt(out.length() - 1) != SEGMENT_SEPARATOR) {
+                        out.append(SEGMENT_SEPARATOR);
+                    }
+                }
+                final int nextSlash = indexOfSlash(path, i);
+                if (nextSlash != -1) {
+                    out.append(path, i, nextSlash);
+                    i = nextSlash;
+                } else {
+                    out.append(path, i, path.length());
```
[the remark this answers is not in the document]
ytvnr: Applied in 814be0f0ce, though not with the suggested patch: the working copy is named `remaining` rather than `currentPath`, to say that it is RFC 3986 §5.2.4 input buffer rather than a second name for the same thing. The parameter is now `final`.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/path/RequestPathNormalizer.java
```diff
@@ -0,0 +1,403 @@
...
+        }
+
+        final int unescaped = hexValue(path.charAt(start + 1), path.charAt(start + 2));
+        if (unescaped < 0) {
+            throw new IllegalArgumentException("Invalid escape sequence: %" + path.substring(start + 1, start + 3));
+        }
+
+        if (isUnreserved(unescaped)) {
+            path.setCharAt(start, (char) unescaped);
+            path.delete(start + 1, start + 3);
```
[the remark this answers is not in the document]
ytvnr: Correct, and deliberately left as is — documented in 814be0f0ce rather than rewritten.

The input here is a request line, which Vert.x caps at `maxInitialLineLength`, 4096 bytes by default. The worst case an attacker can reach is therefore a full line of `%41`, measured at roughly 60 µs. A single-pass rewrite would be faster on paper, and would also be a rewrite of vendored RFC 3986 resolution logic on the one code path in this PR where a mistake is a security defect rather than a slow response. That is a poor trade at this size.

The javadoc now states the bound and the condition that would justify revisiting it, so the next reader does not have to redo this reasoning.

@AikidoSec ignore: bounded by maxInitialLineLength (4 KB), worst case measured at ~60 µs; rewriting vendored path-resolution logic carries more risk than the quadratic term.

ytvnr:
## Manual test protocol, run on a local 4.11 gateway

The automated suite covers the behaviour; this is what was exercised by hand on a running gateway, and how to reproduce it. Everything below passed.

### Setup

One v4 HTTP proxy API on `/proxyv4`, with a single request flow carrying a `mock` policy that prints what the gateway holds:

```json
"selectors": [
  { "type": "http", "path": "/", "pathOperator": "START_WITH", "methods": [] }
],
"request": [
  {
    "name": "Observe the path",
    "enabled": true,
    "policy": "mock",
    "configuration": {
      "headers": [],
      "status": "200",
      "content": "path={#request.path} pathInfo={#request.pathInfo} uri={#request.uri}"
    }
  }
]
```

Two things will silently waste your afternoon if you skip them:

- **The flow selector must be `START_WITH` on `/`, not `EQUALS`.** Every path under test carries extra segments, so an `EQUALS` selector matches none of them, the policy never runs, and the request is proxied to the endpoint instead. You get a response error rather than the path you asked for.
- **`--path-as-is` is mandatory on every `curl`.** Without it curl resolves the dot segments itself, client side, and sends an already-canonical request line. The gateway then receives nothing to normalize, every mode looks identical, and the test proves nothing. The tell is `uri`: it must still show the `../`, since the gateway never rewrites it.

### 1. The path the gateway decides on, and the one it reports

```
curl -s --path-as-is http://localhost:8082/proxyv4/a/../b
```

| Mode | Result |
| --- | --- |
| `RAW` | `path=/proxyv4/a/../b pathInfo=/a/../b uri=/proxyv4/a/../b` |
| `NORMALIZE` | `path=/proxyv4/b pathInfo=/b uri=/proxyv4/a/../b` |

`uri` keeps the received bytes under both, which is what lets an operator correlate with the logs of whatever sits in front of the gateway. Confirmed at the upstream too: the backend received `/b`.

### 2. Dot segments carrying path parameters

The one change that widens what `REJECT` refuses, so the one worth checking on real traffic shapes.

```
curl -s --path-as-is http://localhost:8082/proxyv4/orders/..;/echo
```

| Mode | Result |
| --- | --- |
| `RAW` | `path=/proxyv4/orders/..;/echo` |
| `NORMALIZE` | `path=/proxyv4/echo` |
| `REJECT` | 400 |

And the counter-test, which matters as much: an ordinary segment carrying parameters must keep working.

```
curl -s --path-as-is http://localhost:8082/proxyv4/orders;v=2/echo
```

200 in all three modes. Only a segment that is entirely dots before the `;` is a dot segment.

### 3. Malformed percent sequences

```
curl -s -o /dev/null -w '%{http_code}\n' --path-as-is 'http://localhost:8082/proxyv4/a%+41b'
```

400 in all three modes — **including `RAW`**, which is worth stating because it is not this PR's doing. Under `RAW` the setting is disabled, the dispatcher takes no rejection branch, and the 400 comes from the HTTP layer refusing a malformed request line before the gateway sees it. Verified by the absence of `RejectedPathProcessor - Rejecting request` in the log under `RAW`, and its presence under the other two.

What the two active modes do change is that `%+41` and `%-41` — equally malformed — are now handled identically. Before, one was refused and the other forwarded unchanged, because the predicate and the normalizer each had their own reading of what a hexadecimal digit is.

### 4. An invalid configuration value

```yaml
http:
  pathHandling: NORMALISE
```

A `WARN` naming the accepted values, then the startup line reporting the mode actually in force, `RAW`. The gateway starts — same house rule as the sibling timeout setting, which also falls back and warns rather than refusing to boot.

### 5. The startup line

Now emitted in all three modes, with wording specific to each. Under `RAW` it is a line that did not exist before; under `REJECT` it no longer claims that dot segments are resolved, which they are not.

### 6. The debug console

Under `NORMALIZE`, the Inspector's **Path** row shows the resolved path rather than the one typed in the form.

Under `REJECT`, a debug session on a path with dot segments terminates on a red **400 - Bad Request** with an empty timeline. Before, the event stayed in `DEBUGGING` and the console span forever with nothing in the logs to explain it.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -144,17 +152,45 @@ public DefaultHttpRequestDispatcher(
...
+        final boolean needsNormalization = requestPathConfiguration.isEnabled() && RequestPathNormalizer.needsNormalization(rawPath);
+
+        if (needsNormalization && requestPathConfiguration.getHandling() == RequestPathHandling.REJECT) {
+            return handleRejectedPath(httpServerRequest, serverId, receivedAt);
+        }
+
+        final String normalizedPath = needsNormalization ? RequestPathNormalizer.normalize(rawPath) : rawPath;
+
+        // No normalized form at all: the path carries a malformed percent sequence.
+        if (normalizedPath == null) {
```
phiz71: **Blocking — RAW is no longer the pass-through it is documented to be.**

This guard sits outside the `requestPathConfiguration.isEnabled()` branch, so it fires in every mode. In RAW, `isEnabled()` is false, `needsNormalization` is false, `normalizedPath` is therefore `rawPath` — and a null `rawPath` reaches this line and returns a 400.

A null path is contractual, not theoretical: `HttpServerRequest.path()` carries `@Nullable` in the Vert.x API, and on the pinned `vertx-core` 4.5.27 `Http2ServerRequest.path()` returns `null` when the stream has no `:path` pseudo-header (`ifnull -> aconst_null` in the bytecode), which is the HTTP/2 CONNECT case. Those requests previously reached `httpAcceptorResolver.resolve(host, null, serverId)`.

That contradicts the table in the description — "RAW | Today's behaviour, unchanged. **Default**, so this PR is not a breaking change" — for the default configuration.

Moving the null check inside the enabled branch is enough.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/path/RequestPathNormalizer.java
```diff
@@ -0,0 +1,418 @@
...
+     * same rules drift the moment one is touched alone, and the drift is silent: a false negative
+     * here is a path that is quietly left unnormalized. That agreement is not maintained by review
+     * but asserted, over a generated corpus, by the property {@code needsNormalization(p) ==
+     * !p.equals(normalize(p))} in {@code RequestPathNormalizerTest}.
+     *
+     * @param path the raw request path
+     * @return {@code true} when normalizing would change the path, or when it cannot be normalized
+     */
+    public static boolean needsNormalization(final String path) {
+        if (path == null || path.isEmpty() || path.charAt(0) != SEGMENT_SEPARATOR) {
```
phiz71: **Blocking — `OPTIONS *` is refused in REJECT and rerouted in NORMALIZE.**

`path.charAt(0) != SEGMENT_SEPARATOR` makes every request target that is not in origin-form non-canonical, and asterisk-form is legal (RFC 9110 §7.1; health probes and proxies use it).

On `vertx-core` 4.5.27 I checked the bytecode of `HttpUtils.parsePath`: for `"*"` it takes neither the `/` branch nor the `://` branch, finds no `?`, and returns `"*"` unchanged. So:

- `REJECT` -> `needsNormalization("*") == true` -> **400 on a legal request**;
- `NORMALIZE` -> `normalize("*")` fails the fast path on `charAt(0) == '/'`, enters `normalizeSlow`, which prepends `/` -> the acceptor lookup runs on `/*`, and an API mounted on a root context path receives an `OPTIONS /*` it never received before.

Authority-form (`CONNECT example.com:443`) behaves the same way.

A `"*".equals(path)` passthrough ahead of the scan costs one comparison. No test covers this target in any mode.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -326,6 +385,26 @@ protected DefaultExecutionContext createExecutionContext(VertxHttpServerRequest
         return context;
     }
 
+    /**
+     * Refuses the request before any API is selected, through a processor chain rather than on the
+     * raw response, so that it is measured and reported like any other request the gateway answers
+     * on its own.
+     */
+    private Completable handleRejectedPath(final HttpServerRequest httpServerRequest, final String serverId, final long receivedAt) {
```
phiz71, line 418: **Blocking — the rejected path skips the platform pre-processor chain, and that undoes the reason the analytics flag defaults to on.**

Both other branches run `platformProcessorChainFactory.preProcessorChain()` before their terminal handling (lines 200 and 250). This one runs the rejected chain only.

That chain is where `ConnectionDrainProcessor`, `MetricsProcessor`, `XForwardProcessor` and `TraceContextProcessor` live (`DefaultPlatformProcessorChainFactory.buildPreProcessorList`, lines 63-79). `XForwardProcessor` is what rewrites the remote address from `X-Forwarded-For`.

So on a gateway behind an ingress or a load balancer, in `REJECT` with `handlers.rejected.analytics.enabled` at its new default of `true`, every 400 is reported with the **load balancer's** address rather than the prober's. The javadoc on `rejectedPathProcessorChain()` justifies the default with "a platform being probed with traversal paths leaves a trace" — the trace is there, but it names the wrong peer. Also lost: `Connection: close` during a drain, and W3C `traceparent` propagation.

`DebugHttpRequestDispatcher.dispatchRejected` has the same gap.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-debug/src/main/java/io/gravitee/gateway/reactive/debug/reactor/DebugHttpRequestDispatcher.java
```diff
@@ -69,9 +91,83 @@ public DebugHttpRequestDispatcher(
...
+            ctx.setInternalAttribute(InternalContextAttributes.ATTR_INTERNAL_REACTABLE_API, debugApi);
+            debugCompletionProcessor
+                .execute((HttpExecutionContextInternal) ctx)
+                .subscribe(() -> {}, throwable -> log.error("Failed to close the debug session for a rejected path", throwable));
+        });
+    }
+
+    private ReactableDebugApi<?> resolveDebugApi(final HttpServerRequest httpServerRequest, final String serverId) {
+        final HttpAcceptor acceptor = httpAcceptorResolver.resolve(httpServerRequest.host(), httpServerRequest.path(), serverId);
+        if (acceptor != null && acceptor.reactor() instanceof ApiReactor<?> apiReactor) {
```
phiz71: **Blocking — the endless spinner is still open for an API in V3 execution mode.**

This only matches the reactive `ApiReactor`. A V2 API whose `executionMode` is `V3` is not deployed behind one: `ApiReactorHandlerFactory.canCreate` accepts any non-V4 definition, and `create()` branches on `isV3ExecutionMode(api)` into `getApiReactorHandler(...)`, which the debug service overrides to `DebugApiReactorHandler extends ApiReactorHandler extends AbstractReactorHandler implements ReactorHandler` — never `io.gravitee.gateway.reactive.reactor.ApiReactor`.

So with `http.pathHandling: REJECT`, debugging such an API on a path carrying `..` returns the 400, `debugApi` is `null`, `DebugCompletionProcessor` never runs, the event stays in `DEBUGGING` and the console spins — the exact dead end this override exists to close. The only clue is a misleading warning: the acceptor *did* match, just not to an `ApiReactor`.

`DebugHttpRequestDispatcherTest` mocks `ApiReactor` only, which is why this passes.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/path/RequestPathNormalizer.java
```diff
@@ -0,0 +1,418 @@
...
+            int segmentEnd = segmentStart;
+            while (segmentEnd < path.length() && path.charAt(segmentEnd) != SEGMENT_SEPARATOR) {
+                segmentEnd++;
+            }
+            int afterDots = segmentStart;
+            while (afterDots < segmentEnd && path.charAt(afterDots) == '.') {
+                afterDots++;
+            }
+            final int dots = afterDots - segmentStart;
+            if ((dots == 1 || dots == 2) && afterDots < segmentEnd && path.charAt(afterDots) == PARAM_SEPARATOR) {
```
phiz71, line 236: **The `;params` departure stops one segment short, and the gap is the security-relevant one.**

The servlet reading is applied only where `dots == 1 || dots == 2`, i.e. to all-dot segments. A `;param` on an ordinary segment therefore leaves the path canonical for the gateway and different for the receiver.

Measured: `needsNormalization("/alpha/api/admin;x/secret")` returns `false` in every mode, so `REJECT` does not refuse it and `NORMALIZE` does not rewrite it. Against this PR's own `api-alpha-resource-filtering.json`, whose blacklist is `/admin/**`: that pattern does not match `pathInfo == "/admin;x/secret"`, the request is allowed and forwarded byte-for-byte, and Tomcat/Jetty/Spring strip `;x` before resolving and serve `/admin/secret`.

The class already states the rule it is applying — "a container strips `;params` from every segment *before* resolving anything" — and then applies it to dot segments only. Meanwhile the contract paragraph that lists what the modes assume of the receiver names `%2F` and `\` but never `;`, while the opening promises "the gateway decides on the same path a conforming receiver will resolve".

Either widen the canonicality test to any segment carrying `;` — `REJECT` can refuse those at no cost to the forwarded bytes, which is the same argument the PR already makes for `..;x` — or say the gap explicitly there and in `gravitee.yml`. What I would not leave is the current state, where the reasoning is written down and stops halfway.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -144,17 +152,45 @@ public DefaultHttpRequestDispatcher(
...
+        }
+
+        final String normalizedPath = needsNormalization ? RequestPathNormalizer.normalize(rawPath) : rawPath;
+
+        // No normalized form at all: the path carries a malformed percent sequence.
+        if (normalizedPath == null) {
+            return handleRejectedPath(httpServerRequest, serverId, receivedAt);
+        }
+
+        final boolean pathWasNormalized = normalizedPath != rawPath;
```
phiz71: **This is the single finding failing the SonarCloud quality gate, and the boolean is redundant anyway.**

The Gateway gate is red on "C Reliability Rating on New Code (required >= A)", and the issues API returns exactly one BUG in new code: *Strings and Boxed types should be compared using `equals()`*, on this line. The required check (`ci/circleci: Validate workflow status`) is green, so nothing is technically blocked — but leaving the gate red on a security PR costs the signal for the next one.

The reference comparison is also avoidable: once the null case is handled, `pathWasNormalized` is exactly `needsNormalization`, computed at line 165. The class contract guarantees `normalize` returns the same instance if and only if `needsNormalization` is false, and the property test already asserts that agreement. Reusing the variable removes the Sonar bug, drops the dependency on reference identity, and makes the coupling between the two methods explicit at the call site.

Worth noting for the reader: the two BLOCKER "Add some tests to this class" on `RequestPathNormalizerTest` and `RequestPathRejectionTest` are java:S2187 false positives on `@Nested`-only classes, and do not affect the reliability rating.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -768,6 +778,12 @@ services:
 #  notfound:
 #    analytics:
 #      enabled: false
+#  rejected:
```
phiz71, line 785: **This block breaks the file when an operator uncomments it.**

The three prose lines below are single-commented, so they are YAML content once the leading `# ` is stripped. The convention in this file is a second `#` for anything that must not become YAML — see the `overrideMode` block a few lines above, written `#      #   - override: ...`.

Uncommenting this the way one uncomments `notfound` yields a mapping key `Reporting of the requests refused by http.pathHandling` with value `REJECT. Enabled by default, so that a`, followed by two lines with no colon: the gateway fails to start on a YAML parse error.

Double-commenting the three prose lines fixes it.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -326,6 +385,26 @@ protected DefaultExecutionContext createExecutionContext(VertxHttpServerRequest
...
 
+    /**
+     * Refuses the request before any API is selected, through a processor chain rather than on the
+     * raw response, so that it is measured and reported like any other request the gateway answers
+     * on its own.
+     */
+    private Completable handleRejectedPath(final HttpServerRequest httpServerRequest, final String serverId, final long receivedAt) {
+        // Nothing is rewritten here, so the path stays the one received — which is what the report
+        // has to carry for an operator to see what was actually sent.
+        final MutableExecutionContext ctx = prepareExecutionContext(httpServerRequest, serverId, null, receivedAt);
```
phiz71, line 421: **A rejected request produces no span at all.**

No `mutableCtx.tracer(...)`, no `markTracingRoute(...)`, no root span — where the not-found branch does all three (tracer at 194, route at 199, root span at 209-228) and the API branch does the same at 245-247.

With OpenTelemetry enabled, the one class of request this feature exists to make visible is the only one absent from APM. That sits awkwardly next to the `RejectedPathProcessor` javadoc, which justifies going through a processor chain precisely so the request is "measured and reported like any other request the gateway answers on its own".

Nothing NPEs — `AbstractBaseExecutionContext` falls back to `NO_OP_TRACER` — which is why it is silent.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -144,17 +152,45 @@ public DefaultHttpRequestDispatcher(
...
+        final HttpAcceptor httpAcceptor = httpAcceptorResolver.resolve(httpServerRequest.host(), normalizedPath, serverId);
         Context vertxContext = VertxContext.createNewDuplicatedContext(vertx.getOrCreateContext());
         if (httpAcceptor == null || httpAcceptor.reactor() == null) {
             log.debug(
                 "No acceptor found for host {} and path {}, handling as not found",
                 httpServerRequest.host(),
                 httpServerRequest.path()
             );
-            MutableExecutionContext mutableCtx = prepareExecutionContext(httpServerRequest, serverId);
+            MutableExecutionContext mutableCtx = prepareExecutionContext(httpServerRequest, serverId, null, receivedAt);
```
phiz71: **The not-found branch resolves on one path and reports another.**

The acceptor lookup at line 184 uses `normalizedPath`, but the wrapper is seeded with `null` here, so `NotFoundProcessor` logs and `MetricsProcessor` records the path as received. The debug line just above (188-192) reads `httpServerRequest.path()` for the same reason. The API branch does the opposite and seeds the normalized path (line 241).

Concretely: an operator turns on `NORMALIZE`, then investigates the remaining 404s and sees `/alpha/api/../../beta/nope`, while the lookup that actually failed was on `/beta/nope`. They go looking for the wrong context path.

If reporting the received path is the deliberate choice here, it would help to say so — the two branches currently make opposite ones.

phiz71 (commented):
Reviewed the whole branch. The core holds: I traced the forwarded upstream path end to end and it does derive from the normalized path in both generations (v4 through `pathInfo()` into `HttpConnector`, v3 into `DefaultProxyEndpoint`, which overwrites the `uri()` that `ProxyRequestBuilder.from` had set), the query string travels in a separate field and cannot be dropped, and I brute-forced the normalizer over the 5,380,840 strings of length <= 7 from a targeted alphabet with zero violations of the three stated invariants. There is no dot-segment bypass left.

What I found sits at the edges. Four items look blocking to me, and two of them hit requests that have nothing to do with an attack, in the default mode the description presents as unchanged. Details inline; each one is a short fix.

gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/http/vertx/VertxHttpServerRequest.java
```diff
@@ -57,9 +58,28 @@ public class VertxHttpServerRequest implements Request {
...
+    @Deprecated
     public VertxHttpServerRequest(HttpServerRequest httpServerRequest, IdGenerator idGenerator) {
+        this(httpServerRequest, idGenerator, VertxHttpServerRequestOptions.builder().build());
+    }
+
+    /**
+     * @deprecated see {@link #VertxHttpServerRequest(HttpServerRequest, IdGenerator)}.
+     */
+    @Deprecated
+    public VertxHttpServerRequest(HttpServerRequest httpServerRequest, IdGenerator idGenerator, String path) {
```
phiz71: **These overloads are introduced by this PR and have no callers — the compatibility argument cannot apply to them.**

`git show origin/4.11.x` on this file shows only the two-argument constructor; the `(HttpServerRequest, IdGenerator, String path)` signature never existed on the branch, so nothing external can have been compiled against it. Same for `VertxHttp2ServerRequest:46`, `grpc/VertxGrpcServerRequest:43`, `ws/VertxWebSocketServerRequest:47` and `DefaultHttpRequestDispatcher.createV3Request(..., String):531`.

That puts five untested public signatures on a support branch, which every forward-port to `4.12.x` and `master` then has to either carry or drop, creating an API asymmetry across branches. Sonar already lists all five as "Do not forget to remove this deprecated code someday".

Keeping the pre-existing two-argument constructors deprecated and delegating is right — those genuinely shipped. I would just delete the String-path ones.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/path/RequestPathNormalizer.java
ytvnr: You are right that the reasoning stops halfway, and that is the part I will fix. I am not going to widen the rule, and here is the argument, so the decision is on the record rather than settled in silence.

**Your demonstration holds.** `/alpha/api/admin;x/secret` is canonical to this normalizer, so `REJECT` does not refuse it and `NORMALIZE` does not rewrite it; the resource-filtering blacklist `/admin/**` does not match `pathInfo == "/admin;x/secret"`; the request is forwarded byte-for-byte and a servlet container serves `/admin/secret`. Same parser differential, one segment over, against this PR's own fixture. Nothing to argue with.

**What differs is the cost of the remedy, and it differs sharply.**

On `..;x`, stripping is free. A dot segment carrying path parameters is not legitimate anywhere — no client emits `..;jsessionid=1` on purpose — so treating it as a dot segment cannot refuse or rewrite traffic anyone actually sends. That is why the departure was worth making.

On `admin;x`, it stops being free. Matrix parameters are legitimate and in use, so:

- widening `REJECT` to any segment carrying a `;` refuses well-formed traffic. That is a much larger blast radius than the `..;` case, and it is not the same argument.
- extending the strip to `NORMALIZE` rewrites the path we forward. Toward a servlet container that is correct. Toward nginx, which does not strip `;params`, we would deliver `/admin/secret` for a request that asked for `/admin;x/secret` — a *different* resource, not merely a stricter outcome. It is not fail-closed, which is the property the `..;` decision rests on.

**And the case you demonstrate is a policy pattern-matching problem more than a path-resolution one.** Fixing it in the normalizer amounts to deciding that matrix parameters do not exist — a product decision of a different order than closing a traversal, and one that would land on every API on the gateway rather than on the ones a policy protects.

So: the asymmetry stays, and it gets stated. The contract paragraph that lists what the modes assume of the receiver names `%2F` and `\` and will now name `;` alongside them, with this reasoning; the same limitation goes into the operator documentation, which is being written for 4.11, 4.12 and 4.13.

If you think the servlet reading should apply to every segment, that is a conversation worth having on its own ticket rather than inside a traversal fix — I would want the matrix-parameter fallout measured before we take it, and it belongs on `master` where the default is changing anyway, not on a support branch.

The three quoted lines in `gravitee.yml` are being double-commented in the same pass, so the block survives being uncommented.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
ytvnr: Fixed in 283eee97, and you are right that the description was wrong, not just the code. The null check now sits inside the `isEnabled()` branch, so under RAW nothing is inspected and a null path reaches the acceptor exactly as it did before this setting existed.

The dispatcher-level table in `DefaultHttpRequestDispatcherTest.The_rejection_decision` now walks `null` and `*` across all three modes, so a regression here fails 3 cases rather than none.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/path/RequestPathNormalizer.java
ytvnr: Fixed in 283eee97. `needsNormalization` and `normalize` both answer a request target that is not in origin-form unchanged, before the scan: the asterisk-form and null, which belong together since neither carries segments to resolve.

I kept the exemption narrow rather than exempting anything that fails to start with a slash, so `relative/path` still gets its leading slash forced — there is a test pinning that, otherwise the fix would have quietly widened into a second behaviour change.

Not covered: authority-form on CONNECT over HTTP/1.1. On HTTP/2 it arrives as the null path and is now handled; over HTTP/1.1 I could not produce a case where it reaches the dispatcher, so I did not write speculative code for it.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
ytvnr: Fixed in 283eee97. The platform pre-processor chain now runs before the rejected chain, in both dispatchers.

The X-Forwarded-For point is the one that convinced me: reporting a rejection with the load balancer address rather than the prober address does not merely lose detail, it makes the default we chose for `handlers.rejected.analytics.enabled` pointless. The javadoc on `rejectedPathProcessorChain()` argues the trace is the whole value of refusing, and the trace was naming the wrong peer.

The connection drain and the W3C trace context come back with it.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-debug/src/main/java/io/gravitee/gateway/reactive/debug/reactor/DebugHttpRequestDispatcher.java
ytvnr: Fixed in 283eee97, and this is the one I am least comfortable about — it is the third time this bug escaped, and twice for the reason you name: the test mocked `ApiReactor`, so it could only ever confirm the branch that already worked.

`resolveDebugApi` now switches on both engines, and `DebugApiReactorHandler` exposes its `ReactableDebugApi` since `reactable` is protected and the interface declares no accessor.

I wrote the test first this time and ran it against the unfixed dispatcher to watch it fail — it mocks a `DebugApiReactorHandler`, so it cannot pass by accident the way the previous one did.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
ytvnr: Fixed in 283eee97, and your second point is the better half of it: `pathWasNormalized` is now simply `needsNormalization`, so the reference comparison is gone rather than replaced by an `equals`.

That is the right shape for the reason you give — it makes the coupling between the scan and the normalizer explicit at the call site instead of relying on identity, and the property test is what guarantees it holds.

Thanks for checking the two BLOCKER java:S2187 on the `@Nested`-only classes. I would have gone looking.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
ytvnr: Fixed in 283eee97 — the three prose lines are double-commented, so the block survives being uncommented the way one uncomments `notfound`.

I checked whether the same hazard applies to the `http.pathHandling` block and concluded it does not, so I left it alone. The rule the file follows, once both cases are side by side: double `#` when the prose sits inside a mapping that has to be uncommented wholesale, single `#` when it is a header above a flat key. `pathHandling` is the second shape, and its neighbour `requestTimeoutGraceDelay` is written exactly that way. Doubling it would have created the local exception rather than removed one.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
ytvnr: Fixed in 283eee97. `handleRejectedPath` now sets the tracer, marks the route and opens a root span, mirroring the not-found branch.

The route is bucketed under `/` rather than the request URI, for the same reason the not-found branch does it: a prober chooses that URI, and it is unbounded.

Your framing is what made it obvious — the class exists to make one kind of request visible, and it was the only kind absent from APM.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
ytvnr: Aligned on the API branch in 283eee97: the not-found branch now seeds the resolved path, and the debug line reads it instead of calling `httpServerRequest.path()` again.

It was not a deliberate choice, it was an oversight — and your example is the one that settles it. An operator investigating 404s under NORMALIZE has to see the path the lookup actually failed on, otherwise they go hunting for a context path the gateway never tried.

Nothing is lost by it: `uri()` still carries the bytes the client sent, so the report now names both instead of one.

gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/http/vertx/VertxHttpServerRequest.java
ytvnr: Deleted in 283eee97 — all five, plus the test that covered one of them, which had no more reason to exist than the signature it exercised.

Your distinction is the one I missed: the compatibility argument holds for the two-argument constructors because they genuinely shipped, and holds for nothing I introduced in this PR. Keeping the latter deprecated protected no caller that can exist, while leaving 4.12.x and master to carry or drop an asymmetry.

The two-argument forms stay deprecated and delegating.

helm/values.yaml
```diff
@@ -1206,6 +1206,16 @@ gateway:
...
     alpn: "true"
+    # How dot segments in the request path are treated, before the listener context path is
+    # resolved and therefore before any plan is enforced. One value for the whole gateway:
+    # it cannot be set per server or per API.
+    #   RAW       (default) the path is used, and forwarded, exactly as received.
+    #   REJECT    answers 400 when the normalized path differs from the one received.
+    #   NORMALIZE resolves the dot segments and routes, enforces and forwards on the result.
+    # Beyond dot segments, REJECT and NORMALIZE also decode percent-encoded unreserved
+    # characters, merge duplicate slashes, and answer 400 on a malformed percent sequence.
+    # Encoded slashes are never decoded.
```
phiz71: **This is the block an operator actually reads before switching a production gateway to REJECT, and it is missing the one caveat that changes what they should conclude.**

The paragraph is good — it names dot segments, percent-decoding of unreserved characters, duplicate slashes, malformed sequences, and it ends on "Encoded slashes are never decoded", which is precisely the kind of limit worth stating. The `;params` gap belongs in the same list and is currently only in the `RequestPathNormalizer` javadoc.

The concrete misreading it invites: someone reads "REJECT answers 400 when the normalized path differs from the one received", concludes an allow/deny rule on `/admin/**` is now protected against path games, and ships. `/a/admin;x/b` is canonical here, forwarded byte for byte, and served as `/a/admin/b` by any Servlet container — so the rule and the backend disagree, silently, on exactly the class of request the setting was turned on for.

Something like: *segment parameters on an ordinary segment are left alone — `/a/admin;x/b` is canonical here, while a Servlet receiver serves `/a/admin/b`* is enough. Same two lines in `gravitee.yml` next to `pathHandling`.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -84,6 +84,16 @@ secrets:
...
+#  How dot segments in the request path are treated, before the listener context path is
+#  resolved and therefore before any plan is enforced.
+#    RAW       (default) the path is used, and forwarded, exactly as received.
+#    REJECT    answers 400 when the normalized path differs from the one received.
+#    NORMALIZE resolves the dot segments and routes, enforces and forwards on the result,
+#              so /alpha/api/../../beta/api is read as the call to beta that it is.
+#  Beyond dot segments, both REJECT and NORMALIZE also decode percent-encoded unreserved
+#  characters, merge duplicate slashes, and answer 400 on a malformed percent sequence.
+#  Encoded slashes are never decoded. RAW is the default for backward compatibility.
+#  pathHandling: RAW
```
phiz71, line 100: Same as `helm/values.yaml`: this is where a standalone operator decides between the three modes, and the `;params` limit is not stated anywhere they will look. Worth the two lines here too — the class javadoc is the wrong audience for a deployment decision.

Everything else in this block reads well, and the `rejected:` comment fix below is correct now.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/path/RequestPathNormalizer.java
```diff
@@ -0,0 +1,461 @@
...
+     *     neither mode has any business inspecting.
+     *     <p>RFC 9110 §7.1 allows a request target in asterisk-form, {@code OPTIONS *}, which proxies
+     *     and health probes do send, and in authority-form for {@code CONNECT}. Neither is an origin
+     *     path: there are no segments to resolve, nothing to compare against a context path, and no
+     *     receiver that could read them differently from us. Left to the general scan they would be
+     *     answered 400 under {@code REJECT} — a legal request refused — and rewritten to {@code /*}
+     *     under {@code NORMALIZE}, which would then be routed to whatever sits on the root context
+     *     path.
+     */
+    private static boolean isNotOriginForm(final String path) {
```
phiz71, line 179: The javadoc above lists authority-form for CONNECT among the targets this exempts, but the method tests `null` and `"*"` only.

On HTTP/2 the two coincide — a CONNECT has no `:path` pseudo-header and arrives as `null`, which is what the inline comment describes. On HTTP/1, `CONNECT host:443` reaches `path()` as `"host:443"`: `HttpUtils.parsePath` finds no leading `/` and no `://`, so it returns the target unchanged, the general scan sees a path not starting with `/`, and REJECT answers 400.

Refusing it is arguably right — a gateway is not a forward proxy and that request had nowhere to go anyway. What I would change is the sentence, so the next reader does not take the exemption as covering a form it does not.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/path/RequestPathNormalizer.java
```diff
@@ -0,0 +1,461 @@
...
+     * !p.equals(normalize(p))} in {@code RequestPathNormalizerTest}.
+     *
+     * @param path the raw request path
+     * @return {@code true} when normalizing would change the path, or when it cannot be normalized
+     */
+    public static boolean needsNormalization(final String path) {
+        if (isNotOriginForm(path)) {
+            return false;
+        }
+        if (path == null || path.isEmpty() || path.charAt(0) != SEGMENT_SEPARATOR) {
```
phiz71: Nit: `path == null` is now unreachable — `isNotOriginForm` four lines up already returned `false` for it. Harmless, but it reads as if null were still a live case here, which is the opposite of what the new contract says.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-debug/src/main/java/io/gravitee/gateway/reactive/debug/reactor/DebugHttpRequestDispatcher.java
```diff
@@ -69,9 +92,95 @@ public DebugHttpRequestDispatcher(
...
+        final MutableExecutionContext ctx = prepareExecutionContext(httpServerRequest, serverId, null, receivedAt);
+        ctx.request().contextPath("/");
+
+        // Resolving here decides nothing about routing — the request is refused either way. It only
+        // names the session to close. The raw path still matches: the acceptor compares prefixes and
+        // the debug context path the daemon prepends sits ahead of whatever the user typed.
+        final ReactableDebugApi<?> debugApi = resolveDebugApi(httpServerRequest, serverId);
+
+        final ProcessorChain rejectedChain = notFoundProcessorChainFactory.rejectedPathProcessorChain();
+        final Completable reject = rejectedChain.execute(ctx, ExecutionPhase.RESPONSE);
```
phiz71: **Not blocking, but this round is the argument for the change I would still make.**

`handleRejectedPath` in the parent just grew the platform pre-processor chain, the tracer, the tracing route and the root span. This copy took none of it and still runs the rejected chain bare.

The consequences here are genuinely small — debug traffic comes from the management API, so the `X-Forwarded-For` rewrite hardly matters, and the tracing block would be inert anyway since this dispatcher hands `TracingContext.noop()` to `super`. That is not the point. The point is that the copy drifted from the original inside a single review cycle, which is the failure mode this PR's own `VertxHttpServerRequestOptions` javadoc is written to prevent, and it drifted despite the previous round naming this exact method.

`prepareExecutionContext` was already made `protected` for this subclass. Doing the same to `handleRejectedPath` and overriding it here to append the session-closing `doFinally` would delete this duplication, and the next thing added to the parent's rejection flow would reach the debug port for free. Fine as a follow-up rather than in this PR.

phiz71 (approved):
Approving. Every blocking point from the previous round is addressed, and two of them more broadly than I asked: the null target is now exempted ahead of the scan rather than only inside the RAW branch, so a CONNECT reaches the acceptor identically in all three modes, and the asterisk-form exemption keeps the `==` contract intact with a guard test proving an ordinary relative path still gets its leading slash. The pre-processor chain, the tracer and the root span on the rejected path, the V3-mode debug session, the reference comparison, the YAML block and the five dead constructors are all done, with tests where tests were the right answer.

On the `;params` gap you took the second branch of the alternative, and the reasoning convinces me: a dot segment carrying parameters is never legitimate, an ordinary segment carrying them is, so refusing them under REJECT would turn away well-formed traffic and stripping them under NORMALIZE would forward a path the client did not send. Framing it as a decision about matrix parameters rather than about traversal is the right framing.

**That said, I want to push on where that reasoning lives, because it is the one thing I would not leave as it is.** It is written in the class javadoc — and nobody choosing a mode reads a class javadoc. An operator turning on REJECT reads `gravitee.yml` and `helm/values.yaml`, and both currently describe what the modes do without ever saying what they do not catch. The `values.yaml` block goes as far as spelling out that encoded slashes are never decoded, which is exactly the right instinct, and then stops one item short of the gap that actually matters for an allow/deny policy. Someone will read "REJECT answers 400 when the normalized path differs from the one received", conclude their `/admin/**` rule is now safe against path games, and be wrong about `/a/admin;x/b`. Two lines in each file close that, and they cost nothing.

Same category, smaller: the `isNotOriginForm` javadoc claims authority-form for CONNECT among the targets it exempts, but the method tests only `null` and `"*"`. On HTTP/2 that is fine, since such a request has no `:path` and arrives null — which is what the comment actually describes. On HTTP/1 a `CONNECT host:443` arrives with `path()` at `"host:443"` and still takes a 400 under REJECT. The behaviour is defensible; the sentence promising otherwise is what I would fix.

The remaining two are inline and neither is blocking. I approved before `Build backend` completed — the tree is byte-identical to the one where `Validate backend` passed, but the gateway and integration test jobs had not run when I looked, so read this as an approval of the code rather than of a green suite.

ytvnr:
Housekeeping: the branch was rebased and the review work folded into a single commit, so the SHAs quoted in my replies above (`283eee97`, `814be0f0ce`) no longer resolve. Everything they described is in `4794fef930`, and the tree is what it was — only the commit list changed.

Two things landed since the approval, both from the remaining comments:

**The `;params` limit is now where an operator reads it** — `helm/values.yaml` and `gravitee.yml`, right after "Encoded slashes are never decoded", with the consequence spelled out rather than just the fact: a path-based allow or deny rule is not protected against `/a/admin;x/b`, and needs its own guard. The same went into the documentation PR ([gravitee-platform-docs#2249](https://github.com/gravitee-io/gravitee-platform-docs/pull/2249)) on all three versions. You were right that a class javadoc is the wrong audience for a deployment decision.

**The `isNotOriginForm` javadoc no longer promises authority-form.** It now states plainly that a `CONNECT` over HTTP/1 arrives as `host:443` and takes a 400 under `REJECT`, and why that is left alone. The unreachable `path == null` went with it.

**And the duplication is gone rather than deferred.** `handleRejectedPath` stayed private; instead the parent exposes `afterRejectedPath(Completable, MutableExecutionContext)`, which hands a subclass the rejection and the context it ran on. The debug dispatcher now overrides only that — its `dispatch()` override and its copy of the flow are deleted. Making `handleRejectedPath` protected would not have been enough on its own: what the copy had to recreate was the *context*, which is created inside it, and that is exactly what made it drift.

One consequence worth flagging, since it changes something you reviewed. With the debug dispatcher no longer predicting a rejection before delegating, `RequestPathRejection` lost its only production caller. Rather than leave a public class nothing calls on a support branch — an invitation to recreate the duplication we just removed — it moved to the test tree, where its role is honest: an oracle, written independently of the dispatcher, that `The_rejection_decision` compares the dispatcher against. That table is what caught `OPTIONS *` and the null target, and it keeps doing so.

helm/values.yaml
phiz71: Done, and checked on `c569f182`: the block now carries the caveat, `helm/values.yaml:1217-1220`. Exactly the shape it needed -- the `;x` spelling named, what a Servlet container does with it, and the conclusion an operator has to draw about their own allow/deny rule. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
phiz71: Same on this side: `gravitee.yml:95-98` now states it too, in the same words. A standalone operator deciding between the three modes reads the caveat where they make the decision, rather than in a class javadoc they will never open. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/path/RequestPathNormalizer.java
phiz71: Fixed -- the javadoc now says the opposite of what it said: *Authority-form is not exempted*, with the HTTP/2 and HTTP/1 cases split and the reason for refusing spelled out.

Worth knowing where it ended up: this class is the one that moved to `gravitee-common` in gravitee-io/gravitee-common#170, and the sentence there goes one step further -- it drops the `REJECT` framing entirely, since a library has no modes, and says the target *is reported as non-canonical*. So the fix outlived this branch. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/path/RequestPathNormalizer.java
phiz71: Gone -- the unreachable `path == null` test is replaced by a comment saying why it is not needed: *Not null by here: isNotOriginForm answered for that case above.* Which is the better outcome, since the thing worth writing down was the reason, not the check. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-debug/src/main/java/io/gravitee/gateway/reactive/debug/reactor/DebugHttpRequestDispatcher.java
phiz71: Fixed, and the shape is better than the one I asked for. I suggested making `handleRejectedPath` protected; what landed is an `afterRejectedPath(Completable, MutableExecutionContext)` extension point, and the debug dispatcher is now down to that single override -- no copy of the flow left in it at all.

That is the stronger answer to the drift I was pointing at: a `protected handleRejectedPath` could still have been overridden wholesale and drifted again. An extension point that receives the rejection and the context cannot. Resolving.

phiz71 (commented):
Went back over the fifteen threads I still had open here and verified each fix against `c569f182` rather than against the replies. All fifteen are addressed, so I am resolving them.

The ten you answered, and what I checked:

| Finding | Verified on `c569f182` |
|---|---|
| RAW no longer a pass-through for a null path | the null check sits inside the `isEnabled()` branch; under RAW nothing is inspected |
| `OPTIONS *` refused / rerouted | `isNotOriginForm` is `path == null \|\| ASTERISK_FORM.equals(path)`, ahead of the scan |
| rejection skipped the platform pre-processor chain | `preProcessorChain` runs at line 431, before the rejected chain |
| endless spinner for a V3-mode API | `resolveDebugApi` switches on `ApiReactor` **and** `DebugApiReactorHandler` |
| `;params` reasoning stopped halfway | behaviour unchanged by decision, and the limitation is now documented in both operator-facing files |
| Sonar bug on the reference comparison | `pathWasNormalized = needsNormalization`; the identity comparison is gone rather than replaced |
| `gravitee.yml` block broke on uncommenting | the three prose lines are double-commented at 786-788 |
| no span for a rejected request | tracer at 425, route bucketed at 428, root span at 458 |
| not-found reported a different path than it resolved | the branch seeds the resolved path, and the debug line reads it |
| five uncallable overloads | none left on `VertxHttpServerRequest` or its subtypes |

Two of them landed better than what I asked for: the reference comparison was removed instead of wrapped in `equals`, and the debug drift was closed with an extension point rather than a `protected` method that could drift again.

One thing worth recording, since it cuts across the four PRs of this change: the authority-form javadoc finding was fixed **here**, and then fixed again in gravitee-io/gravitee-common#170 when the class moved -- where the sentence dropped its `REJECT` framing, because a library has no modes. A thread can be closed by work done in another repository. Checking the repo the code landed in, not only the one the thread was opened on, is the habit this chain needs.

Approval stands. What still blocks the merge is on the other three PRs: gravitee-common#170 has to be released as `5.2.0` before #19387 and #19388 can drop their snapshot pin.

PR #19168, opened 2026-08-14
feat(distribution): init JDBC migrations distribution module with Liquibase changesets

Okhelifi:
**Issue**

https://gravitee.atlassian.net/browse/FOUND-183

**Description**

Adds `gravitee-apim-distribution-jdbc-migrations`, a distribution module that publishes every Liquibase changeset APIM and its bundled Gamma modules apply to a JDBC management database, as one versioned archive on the download website.

Installations running with `management.jdbc.liquibase=false` — controlled upgrades, DBA-driven processes, restricted database credentials — currently have no supported way to apply the pending schema changes themselves. The changesets exist only inside each plugin's jar, inside its zip, and nothing documents how to invoke them. That is already true for APIM alone; Gamma modules turn one undocumented procedure into one per installed module, each with its own tracking tables and prefix parameter.

The archive collects them all:

```
gravitee-apim-jdbc-migrations-<version>.zip
├── README.md
└── liquibase/
    ├── master.yml          APIM
    ├── changelogs/v*/
    └── aim/                one directory per module that owns a schema
```

**How it is built**

The changesets sit inside each plugin's jar, which sits inside the plugin zip, so collecting them takes two stages: `dependency:unpack-dependencies` takes the jar out of each zip, then `antrun` takes `liquibase/**` out of the jars.

The dependencies are the ordinary plugin zips, the same ones the standalone distribution bundles, at the same versions — so **no component had to change**. A module that owns no schema contributes nothing and needs no declaration of its own, which is what keeps the list stable while modules adopt JDBC one at a time.

Two details are load-bearing and commented in place:

- the extraction takes `*.jar`, not `**/*.jar`. The jars under `lib/` include `liquibase-core`, whose own Java package is also called `liquibase`; unpacking those would pour 1426 class files into an archive meant for a DBA. Keeping `lib/` out makes that impossible rather than filtered.
- the `liquibase/` prefix is preserved verbatim. Liquibase identifies a changeset by *(id, author, filename)*, and `filename` is the changelog path as resolved — the same path the application records in `databasechangelog`. Re-rooting the tree would make a manual run record different filenames, and the application would then replay every changeset against an already-migrated database, without reporting an error at the time of the run.

An enforcer rule asserts APIM's own changelog was collected. Without it, a broken include pattern would publish a valid, empty archive and the build would stay green.

**Verified locally**

On Postgres, using only the published archive — unzipped into a directory, and nothing else: no
plugin opened, no jar extracted, no classpath assembled — with a Liquibase installed from Homebrew
(4.33) while the application runs 4.27:

- **Preview.** `update-sql` produced the 4.12 → 4.13 delta: 14 changesets, exactly the 14 files of `v4_13_0`; 4 tables, 9 columns, 6 indexes, each matching the changelogs on master.
- **Apply.** `update` applied them, and the module's own changesets went only to `aim_databasechangelog`, never to APIM's tracking table.
- **Restart.** The application then started normally and replayed nothing — `Run: 0, Previously run: 223` for APIM, `Run: 0, Previously run: 2` for aim, both tracking tables unchanged.

The last point is the one that matters: the paths a manual run records match what the application expects, and checksums written by 4.33 are accepted by 4.27.

Also confirmed the release job picks the archive up with no CI change: the zip matches its `find . -path '*target/gravitee-apim*.zip'`, the module declares `publish-folder-path`, and the name regex resolves to `gravitee-apim-jdbc-migrations`. It lands in `graviteeio-apim/jdbc-migrations/gravitee-apim-jdbc-migrations/`.

**Additional context**

The README shipped in the archive is the module's own `README.md` — one file, so the repository and the archive cannot drift. It leads with the failure a user is most likely to hit: omitting `-Dgravitee_prefix=` leaves the placeholder unresolved, which changes every checksum and produces 141 validation errors that look like a corrupted archive or an incompatible version. Nothing in Liquibase's message points at a missing parameter.

It also tells the reader to discover the components to run (`find . -name master.yml`) rather than listing them, so it cannot go stale as modules adopt JDBC.

Compatibility-sensitive: this publishes Liquibase changelogs as a customer-facing artifact. It adds no changeset and changes none — the files are the ones already inside the plugins.

### DIFF
--- a/gravitee-apim-distribution/gravitee-apim-distribution-es-index-mappings/pom.xml
+++ b/gravitee-apim-distribution/gravitee-apim-distribution-es-index-mappings/pom.xml
@@ -48,7 +48,7 @@
         <maven.deploy.skip>false</maven.deploy.skip>
 
         <!-- Property used by the publication job in CI-->
-        <publish-folder-path>graviteeio-apim/es-index-mappings</publish-folder-path>
+        <publish-folder-path>graviteeio-apim/resources</publish-folder-path>
     </properties>
 
     <dependencies>
--- a/gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/README.md
+++ b/gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/README.md
@@ -1,0 +1,137 @@
+# APIM — JDBC schema migrations
+
+This module packages every Liquibase changeset APIM and its bundled Gamma modules apply to a JDBC
+management database, so that installations running with `management.jdbc.liquibase=false` can apply
+them themselves instead of opening each installed plugin. It ships as this file's own README inside
+the published archive, which is what a customer reads.
+
+Applies to the JDBC backend only.
+
+You need a Liquibase CLI. Nothing else: the changesets in the archive are the ones the application
+runs, and the CLI ships the JDBC driver for every database APIM supports except MySQL, whose licence
+keeps it out. The archive filename carries the APIM version it was built from.
+
+## Read this first
+
+**Always pass the prefix parameters, even when your installation uses no prefix.** Almost every
+changeset names its tables through a parameter — `${gravitee_prefix}api_products` rather than
+`api_products`. Omitting the parameter leaves the placeholder unresolved, which changes the checksum
+of every changeset, and Liquibase then reports well over a hundred failures of this shape:
+
+```
+Validation Failed:
+     141 changesets check sum
+          liquibase/changelogs/v1_15_0/schema.yml::1.15.0::GraviteeSource Team
+              was: 9:89bae981dae8df68c63aa03c4aa12792 but is now: 9:8259b78599409dec370ffa0c10c4395f
+```
+
+Nothing is corrupted and no version is incompatible: a parameter is missing. Pass
+`-Dgravitee_prefix=` with an empty value if you have no prefix.
+
+## What is in the archive, and what to run it against
+
+One entry point per component. List them:
+
+```
+find . -name master.yml
+```
+
+```
+./liquibase/master.yml       APIM itself
+./liquibase/aim/master.yml   the "aim" Gamma module
+```
+
+`liquibase/master.yml` is always APIM. Every other entry point is a Gamma module, and the directory
+name is the module id. Only modules that own a schema appear — a module absent from this list has
+nothing to migrate.
+
+**Each entry point is a separate run.** Modules keep their migration history in their own tracking
+tables, independently of APIM and of each other, so running APIM's changelog does not migrate the
+modules. Derive the arguments from the id:
+
+| | Changelog | Tracking tables | Parameters |
+|---|---|---|---|
+| APIM | `liquibase/master.yml` | `<prefix>databasechangelog` and `<prefix>databasechangeloglock` | `gravitee_prefix`, `gravitee_rate_limit_prefix` |
+| module `<id>` | `liquibase/<id>/master.yml` | `<prefix><id>_databasechangelog` and `<prefix><id>_databasechangeloglock` | `<id>_prefix` |
+
+`<prefix>` is `management.jdbc.prefix` from your `gravitee.yml`, empty unless you configured one.
+`gravitee_rate_limit_prefix` is `ratelimit.jdbc.prefix`.
+
+**The tracking table names carry the prefix too.** The application derives them from
+`management.jdbc.prefix`, so a run that leaves them at Liquibase's default opens a different table
+from the one the application writes to, finds it empty, and replays every changeset against an
+already-migrated database.
+
+Set these once and every command below can be pasted as it is:
+
+```
+URL=<jdbc url>     # e.g. jdbc:postgresql://localhost:5432/gravitee
+USER=<user>
+PASSWORD=<password>
+PREFIX=            # management.jdbc.prefix, leave empty if you configured none
+RL_PREFIX=         # ratelimit.jdbc.prefix
+```
+
+## Reviewing what an upgrade would do
+
+`update-sql` writes the pending statements to a file without applying them. Run it from the
+directory you unzipped the archive into.
+
+APIM:
+
+```
+liquibase --search-path=. --changelog-file=liquibase/master.yml --url="$URL" --username="$USER" --password="$PASSWORD" --database-changelog-table-name="${PREFIX}databasechangelog" --database-changelog-lock-table-name="${PREFIX}databasechangeloglock" --output-file=apim.sql update-sql "-Dgravitee_prefix=$PREFIX" "-Dgravitee_rate_limit_prefix=$RL_PREFIX"
+```
+
+Then one run per module — here `aim`:
+
+```
+liquibase --search-path=. --changelog-file=liquibase/aim/master.yml --url="$URL" --username="$USER" --password="$PASSWORD" --database-changelog-table-name="${PREFIX}aim_databasechangelog" --database-changelog-lock-table-name="${PREFIX}aim_databasechangeloglock" --output-file=aim.sql update-sql "-Daim_prefix=$PREFIX"
+```
+
+Or generate every file in one pass:
+
+```
+liquibase --search-path=. --changelog-file=liquibase/master.yml --url="$URL" --username="$USER" --password="$PASSWORD" --database-changelog-table-name="${PREFIX}databasechangelog" --database-changelog-lock-table-name="${PREFIX}databasechangeloglock" --output-file=apim.sql update-sql "-Dgravitee_prefix=$PREFIX" "-Dgravitee_rate_limit_prefix=$RL_PREFIX"
+for module in $(find liquibase -mindepth 2 -name master.yml | cut -d/ -f2); do
+  liquibase --search-path=. --changelog-file="liquibase/$module/master.yml" --url="$URL" --username="$USER" --password="$PASSWORD" --database-changelog-table-name="${PREFIX}${module}_databasechangelog" --database-changelog-lock-table-name="${PREFIX}${module}_databasechangeloglock" --output-file="$module.sql" update-sql "-D${module}_prefix=$PREFIX"
+done
+```
+
+The generated files carry their own `INSERT INTO ... databasechangelog` statements. Applying one
+therefore leaves the database in the state the application expects, and it will start without
+replaying anything.
+
+`update-sql` does not change your schema, but it does take the Liquibase lock, so it writes to the
+lock table and creates it if absent. It is not a read-only operation.
+
+## Applying
+
+Replace `update-sql` with `update` to apply.
+
+## Three things that will bite you
+
+**Do not move the files inside `liquibase/`.** Liquibase identifies a changeset by *(id, author,
+filename)*, and `filename` is the changelog path as resolved — it is stored in the tracking table.
+The paths in the archive are the ones the application records when it migrates itself. Re-rooting
+them makes your run record different filenames, and the application then considers nothing applied
+and replays everything against an already-migrated database. No error is reported at the time of
+your run.
+
+**The archive describes one combination.** It carries one APIM version and the module versions
+bundled with it. If you upgraded a module on its own, use the changelogs of the version you actually
+run — they are inside that plugin's own archive, under `liquibase/<id>/`.
+
+**MySQL and MariaDB need `--classpath`.** On MySQL, because the CLI carries no driver for it. And on
+both, when the server runs with `sql_require_primary_key=ON`: Liquibase creates its tracking table
+without a primary key, and the server rejects it with error 3750 before any changeset runs. The
+application never hits this, because it registers a generator that adds the key. Take that same jar
+out of the plugin you already have installed:
+
+```
+unzip -j plugins/gravitee-apim-repository-jdbc-*.zip 'gravitee-apim-repository-jdbc-*.jar' -d .
+```
+
+then add `--classpath=<that jar>` to the commands above, prefixed by your MySQL driver and a colon if
+you need one. It only matters for the run that creates the tracking table: once the table exists with
+its key, later runs need nothing.
--- a/gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml
+++ b/gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml
@@ -1,0 +1,346 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright © 2015 The Gravitee team (http://gravitee.io)
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.gravitee.apim.distribution</groupId>
+        <artifactId>gravitee-apim-distribution</artifactId>
+        <version>${revision}${sha1}${changelist}</version>
+    </parent>
+
+    <artifactId>gravitee-apim-distribution-jdbc-migrations</artifactId>
+    <packaging>pom</packaging>
+    <name>Gravitee.io APIM - Distribution - JDBC Migrations</name>
+    <description>Liquibase changesets of APIM and of the bundled Gamma modules, in one archive. JDBC only.</description>
+
+    <properties>
+        <!--
+            The umbrella turns deployment off for the assembly poms; this module is an exception, like
+            es-index-mappings — the archive it produces is the deliverable, and it is meant to reach the download
+            website so that installations running with management.jdbc.liquibase=false have somewhere to get the
+            changesets from.
+        -->
+        <maven.deploy.skip>false</maven.deploy.skip>
+
+        <!-- Property used by the publication job in CI-->
+        <publish-folder-path>graviteeio-apim/jdbc-migrations</publish-folder-path>
+
+    </properties>
+
+    <!--
+        Published so that an installation running with management.jdbc.liquibase=false has one place to get every
+        changelog it needs, instead of opening each installed plugin.
+
+        The dependencies are the ordinary plugin zips, the same ones the standalone distribution bundles, at the
+        same versions. Each carries its changelogs at its root under liquibase/, and unpack-dependencies keeps only
+        that. A component with no schema contributes nothing and needs no declaration of its own — which is what
+        keeps these lists stable while modules adopt JDBC one at a time.
+
+        The trees are disjoint by construction: APIM owns liquibase/master.yml and liquibase/changelogs/, each
+        module owns liquibase/<module id>/. A module taking liquibase/changelogs/ for itself would silently
+        overwrite APIM's files here; that path is reserved.
+
+        Only the open-source repository plugin is declared here. The Gamma modules are enterprise artefacts,
+        resolvable from the private Artifactory only, so they live in the bundle-default profile as they do in the
+        standalone distribution — see the rule at gravitee-apim-distribution-standalone/pom.xml:47-50. Declaring
+        them here would break any build without credentials, job-community-build-backend first.
+    -->
+    <dependencies>
+        <dependency>
+            <groupId>io.gravitee.apim.repository</groupId>
+            <artifactId>gravitee-apim-repository-jdbc</artifactId>
+            <version>${apim.server.version}</version>
+            <type>zip</type>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <!--
+                Both directories accumulate: nothing here overwrites a file that no longer has a source. Bump a
+                module version and rebuild without clean, and the previous jar stays in plugin-jars, its changelogs
+                stay in migrations, and the archive ships changesets that no master.yml references. The build stays
+                green, and the assertions below stay green too, because the files they expect do exist.
+
+                The markers directory has to go with them. unpack-dependencies records what it has already
+                extracted, so leaving the markers behind while deleting plugin-jars would make the unpack a no-op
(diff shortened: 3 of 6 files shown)


gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Reviewing what an upgrade would do
+
+`update-sql` writes the pending statements to a file without applying them. Run it from the
+directory you unzipped the archive into.
+
+APIM:
+
+```
+liquibase --search-path=. --changelog-file=liquibase/master.yml --url=<jdbc url> --username=<user> --password=<password> --output-file=apim.sql update-sql -Dgravitee_prefix= -Dgravitee_rate_limit_prefix=
```
phiz71: **blocker** — the APIM command passes neither `--database-changelog-table-name` nor `--database-changelog-lock-table-name`, but the application derives both from the prefix: `AbstractJdbcRepositoryConfiguration.runLiquibase()` sets `liquibase.databaseChangeLogTableName` to `prefix + "databasechangelog"`. The table at line 54 documents this correctly — the commands do not follow it.

On an installation with `management.jdbc.prefix: gio_`, this command opens Liquibase's default `DATABASECHANGELOG`, finds it absent, creates it, and concludes nothing has been applied — then replays all 223 changesets from `v1_14_0` against a fully migrated database. That is exactly the silent failure described at lines 100-105, reached through the table name instead of the path.

Same on line 80. Note it also affects `update-sql`: per lines 90-91 it takes the lock, so the preview alone creates an unprefixed lock table.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/README.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+Then one run per module — here `aim`:
+
+```
+liquibase --search-path=. --changelog-file=liquibase/aim/master.yml --url=<jdbc url> --username=<user> --password=<password> --database-changelog-table-name=aim_databasechangelog --database-changelog-lock-table-name=aim_databasechangeloglock --output-file=aim.sql update-sql -Daim_prefix=
```
phiz71: **blocker** — the tracking table is hardcoded without the prefix, contradicting line 55 (`<prefix><id>_databasechangelog`).

Verified against the shipped artefact: `AimLiquibaseRunner` in `gravitee-gamma-module-aim-3.7.0.jar` builds the name as `prefix + "aim_databasechangelog"`, and the `liquibase/aim/master.yml` inside the archive says `${aim_prefix}aim_databasechangelog`.

The one-pass loop at line 82 has the same problem. A prefixed installation that copy-pastes either one records module history in `aim_databasechangelog` while the application reads `gio_aim_databasechangelog`, and the module replays its changesets on the next start — which line 87 explicitly promises will not happen.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+            <type>zip</type>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
```
phiz71: **blocker** — the four `com.graviteesource.gamma.module` zips are declared in plain `<dependencies>`, outside any profile.

The standalone distribution puts the same artefacts in the `bundle-default` profile, gated on `-Dbundle` (`gravitee-apim-distribution-standalone/pom.xml:1373-1379`), and states the rule at lines 47-50: *"if the dependency is 'enterprise' … add it in the `bundle-default` profile"*. That pom has zero `com.graviteesource` occurrences outside profiles and 94 inside.

No enterprise repository is declared anywhere in the chain — `gravitee-apim-distribution/pom.xml:304-320` declares only Sonatype snapshots, `gravitee-parent:25.1.0` declares none — so `mvn -f gravitee-apim-distribution/pom.xml install` without credentials now fails at dependency resolution. That is the command `job-community-build-backend.ts:46` runs, with no `--settings` and no `-Dbundle`, under the comment *"whether an outside contributor can build what we ship"*. That job is not referenced by any workflow today, so CI will not catch it.

phiz71 (commented):
The archive itself is well built, and the design notes in the pom are genuinely useful — the `*.jar`-not-`**/*.jar` reasoning and the `filename`-stability warning are both correct. I checked the latter against the source: `AbstractJdbcRepositoryConfiguration` passes `liquibase/master.yml` through a `ClassLoaderResourceAccessor` and every include is classpath-absolute, so a CLI run with `--search-path=. --changelog-file=liquibase/master.yml` does record identical filenames. Same for the module. I also built the module and inspected the resulting archive; unless stated otherwise, the comments below are verified against a real build or a real artefact rather than read off the source.

Comments are graded blocker / improvement / nit.

One cross-cutting point that does not anchor on a line: with #19185 restoring the S3 upload, `graviteeio-apim/resources/` becomes a published directory for the first time. Worth settling the artefact naming (two comments below) before that first release rather than after.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+                <executions>
+                    <execution>
+                        <id>extract-plugin-jars</id>
+                        <phase>prepare-package</phase>
+                        <goals>
+                            <goal>unpack-dependencies</goal>
+                        </goals>
+                        <configuration>
+                            <includes>*.jar</includes>
+                            <outputDirectory>${project.build.directory}/plugin-jars</outputDirectory>
```
phiz71, line 147: **blocker** — neither `plugin-jars` nor `migrations` is emptied before being filled, and `unpack-dependencies` markers make the unpack a no-op on a non-clean rebuild, so a version bump without `clean` ships a mixed archive.

Reproduced: changed `gravitee-gamma-module-aim.version` from 3.7.0 to 3.1.0 and reran `mvn package` without `clean`. `plugin-jars/` held both jars, antrun expanded six instead of five, and the archive shipped 3.7.0's `liquibase/aim/master.yml` alongside an orphan `liquibase/aim/changelogs/2.23.0/schema.yml` from 3.1.0 that no master.yml references. Build green, enforcer green.

The Ant fileset visits jars in filename order, so the winner is not the newer version: bumping aim 1.9.0 to 1.10.1 would make the *older* master.yml win, since `"1.10.1" < "1.9.0"`.

Suggest deleting both directories in `prepare-package` before unpacking.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+                        <id>changelogs-were-collected</id>
+                        <phase>prepare-package</phase>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <rules>
+                                <requireFilesExist>
+                                    <files>
+                                        <file>${project.build.directory}/migrations/liquibase/master.yml</file>
```
phiz71: **improvement** — this asserts a file that comes from `gravitee-apim-repository-jdbc`, the one dependency whose absence is unmissable. Every module-shaped failure stays invisible.

`authz`, `edge` and `esm` already contribute nothing (verified: no `liquibase/` entries in their root jars) and the build is green — so a green build proves nothing about module coverage. A fifth module adopting JDBC without being added above, or an existing one moving its changelogs out of the jar root, both ship a valid but incomplete archive. And `README.md:45-46` explicitly licenses the reader to conclude such a module "has nothing to migrate", so it is not detectable downstream either.

One `requireFilesExist` entry per module known to own a schema (`liquibase/aim/master.yml` today) would turn that red.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/README.md
```diff
@@ -0,0 +1,109 @@
+# APIM — JDBC schema migrations
+
+This module packages every Liquibase changeset APIM and its bundled Gamma modules apply to a JDBC
+management database, so that installations running with `management.jdbc.liquibase=false` can apply
+them themselves instead of opening each installed plugin. It ships as this file's own README inside
+the published archive, which is what a customer reads.
+
+Applies to the JDBC backend only. The `mongodb` backends have no file-based migrations.
```
phiz71: **improvement** — `gravitee-apim-repository-mongodb/src/main/resources/scripts/` holds 54 `.js` migration scripts across about forty version directories. A MongoDB customer reading this concludes their upgrade needs no data migration and skips them.

The pom `<description>` at line 33 also names two backends where this line names one.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+You need a Liquibase CLI and a JDBC driver for your database. Nothing else: the changesets in the
+archive are the ones the application runs. The archive filename carries the APIM version it was
```
phiz71: **improvement** — `finalName` uses `${project.version}`, the distribution's version, not `apim.server.version`.

Built at this HEAD without a profile, it produces `gravitee-apim-jdbc-migrations-4.13.0-SNAPSHOT.zip` whose newest changelog is `liquibase/changelogs/v4_12_0` — the changesets of `gravitee-apim-repository-jdbc-4.12.14`. Only `-Pengine-snapshot` aligns the two, and only the release job and `Taskfile.dist.yml:45` pass it.

A support engineer handing over a locally built archive therefore ships 4.12.14 changesets labelled 4.13.0.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+                <executions>
+                    <execution>
+                        <id>collect-changelogs</id>
+                        <phase>prepare-package</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <unzip dest="${project.build.directory}/migrations">
```
phiz71: **improvement** — nothing enforces the reservation stated at lines 57-59.

All five jars expand into one directory, Ant's `<unzip>` overwrites by default, and the fileset order is alphabetical: `gravitee-apim-repository-jdbc-*.jar` before `gravitee-gamma-module-*.jar`. A module that one day ships `liquibase/master.yml`, or anything under `liquibase/changelogs/`, silently replaces part of APIM's tree — and the rule below still passes, because the file exists, it is just the wrong one.

`useSubDirectoryPerArtifact` plus a duplicate relative-path check would make the reservation real rather than documentary.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+                The changelogs live inside each plugin's jar, which lives inside the plugin zip, so collecting them
+                takes two stages. Maven does the first, Ant the second.
+
+                Only the jar at the root of each zip is taken — deliberately "*.jar" and not "**/*.jar". The jars
+                under lib/ include liquibase-core, whose own Java package is also called liquibase: unpacking it
+                would pour 1426 class files into an archive meant for a DBA, and the changelog assertion below would
+                still pass. Keeping lib/ out is what makes that impossible rather than merely filtered.
+            -->
+            <plugin>
+                <artifactId>maven-dependency-plugin</artifactId>
```
phiz71, line 137: **improvement** — this plugin and `maven-antrun-plugin` (line 153) carry no `<version>`, and nothing in this module's ancestry pins them.

The effective POM resolves 3.7.0 and 3.1.0, both from the Maven 3.9.9 super-POM — so the versions are chosen by the executor image, not by the repo. The `<maven-dependency-plugin.version>3.8.1</maven-dependency-plugin.version>` at `gravitee-apim-distribution/pom.xml:103` is not honoured here, because the `pluginManagement` that consumes it lives in the sibling `gravitee-apim-distribution-standalone/pom.xml:1336-1337`, not in an ancestor.

The same reactor therefore unpacks with 3.8.1 in standalone and 3.7.0 here, and on Maven 3.8.x the super-POM pins antrun 1.3, which has no `<target>` parameter at all. `gravitee-parent` does manage assembly (3.6.0) and enforcer (3.6.3), which is what makes the gap easy to miss.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/README.md
```diff
@@ -0,0 +1,109 @@
+# APIM — JDBC schema migrations
+
+This module packages every Liquibase changeset APIM and its bundled Gamma modules apply to a JDBC
+management database, so that installations running with `management.jdbc.liquibase=false` can apply
+them themselves instead of opening each installed plugin. It ships as this file's own README inside
+the published archive, which is what a customer reads.
+
+Applies to the JDBC backend only. The `mongodb` backends have no file-based migrations.
+
+You need a Liquibase CLI and a JDBC driver for your database. Nothing else: the changesets in the
```
phiz71: **improvement** — two gaps behind "Nothing else".

The archive contains only `.yml` files and this README (verified: zero `.class`, no `META-INF/`). `MySqlChangeLogTableWithPrimaryKeyGenerator`, registered through `META-INF/services/liquibase.sqlgenerator.SqlGenerator`, is therefore not shipped. And it is precisely under `management.jdbc.liquibase=false` that no tracking table was ever created, so the CLI has to create them — Liquibase creates them without a primary key, which `sql_require_primary_key=ON` (HeatWave, Group Replication) rejects with MySQL error 3750 before any changeset runs.

Separately: the README states the driver requirement but no command shows how to give it to Liquibase (`--classpath`, or dropping the jar into `lib/`).

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/README.md
```diff
@@ -0,0 +1,109 @@
...
+The generated files carry their own `INSERT INTO ... databasechangelog` statements. Applying one
+therefore leaves the database in the state the application expects, and it will start without
+replaying anything.
+
+`update-sql` does not change your schema, but it does take the Liquibase lock, so it writes to the
+lock table and creates it if absent. It is not a read-only operation.
+
+## Applying
+
+Replace `update-sql` with `update` to apply, or with `changelog-sync` to record the changesets as
```
phiz71: **improvement** — `update` and `changelog-sync` get one and a half sentences between them.

`changelog-sync` is the only command here that can destroy recoverability: pointed at a fresh or partially migrated database it records every changeset as applied without running any DDL, the application then starts against a schema that does not exist, and no later `update` can repair it because Liquibase considers the work done.

Worth adding a backup step, an explicit "stop the application first" (APIM runs Liquibase at startup by default, so a manual `update` during a rolling restart races the application's own run), and a note that DDL is auto-commit on MySQL/MariaDB, so a changeset failing halfway leaves a partially migrated schema with no rollback — something the `update-sql` review step does not protect against.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+                <executions>
+                    <execution>
+                        <id>make-migrations-assembly</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                        <configuration>
+                            <appendAssemblyId>false</appendAssemblyId>
+                            <finalName>gravitee-apim-jdbc-migrations-${project.version}</finalName>
```
phiz71: **improvement** — `finalName` diverges from the artifactId, so the same archive has two names.

Verified by installing it: with `packaging=pom` and `appendAssemblyId=false`, assembly 3.6.0 attaches with a null classifier, so the coordinates decide the published name. The result is `gravitee-apim-jdbc-migrations-X.zip` on the download site (the publish script uses the on-disk name) and `gravitee-apim-distribution-jdbc-migrations-X.zip` in any Maven repository.

es-index-mappings sets no `finalName`, so once #19185 lands the two sit side by side under `graviteeio-apim/resources/` with two different conventions, one of which matches neither its artifactId nor its GAV. Cheapest to settle now: either give es-index-mappings a matching `finalName`, or drop this one.

gravitee-apim-distribution/gravitee-apim-distribution-es-index-mappings/pom.xml
```diff
@@ -48,7 +48,7 @@
         <maven.deploy.skip>false</maven.deploy.skip>
 
         <!-- Property used by the publication job in CI-->
-        <publish-folder-path>graviteeio-apim/es-index-mappings</publish-folder-path>
+        <publish-folder-path>graviteeio-apim/resources</publish-folder-path>
```
phiz71, line 51: **improvement** — this moves a public download path in a PR whose subject is JDBC migrations, and the description does not mention it. `gravitee-apim-distribution/AGENTS.md` lists "Deployment files and CI configuration" among the surfaces to call out explicitly in the PR description.

Nothing breaks today — both `graviteeio-apim/es-index-mappings/` and `graviteeio-apim/resources/` currently return 404 — which is rather the argument: the change has no observable effect in this PR and would read better as its own commit, with #19185 as its context.

While you are here: the PR description still says the archive lands in `graviteeio-apim/jdbc-migrations/…`, which the second commit made stale.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+        The dependencies below are the ordinary plugin zips, the same ones the standalone distribution bundles, at
+        the same versions. Each carries its changelogs at its root under liquibase/, and unpack-dependencies keeps
+        only that. A component with no schema contributes nothing and needs no declaration of its own — which is
+        what keeps this list stable while modules adopt JDBC one at a time.
+
+        The trees are disjoint by construction: APIM owns liquibase/master.yml and liquibase/changelogs/, each
+        module owns liquibase/<module id>/. A module taking liquibase/changelogs/ for itself would silently
+        overwrite APIM's files here; that path is reserved.
+    -->
+    <dependencies>
```
phiz71, line 66: **nit** — these five blocks are copied from the standalone distribution with the same `<type>zip</type>` and the same wildcard exclusions: lines 62-73 match `standalone/pom.xml:1035-1047`, lines 75-121 match `1765-1815`, modulo `<scope>runtime</scope>`.

The exclusions are genuinely needed — the repository-jdbc pom pulls 21 dependencies including `liquibase-core`, `HikariCP` and `spring-jdbc`, which would land in `plugin-jars/` — but there is nowhere to declare them once: `gravitee-apim-bom` carries no `<type>zip</type>` entry, and the distribution pom has no `<dependencyManagement>` beyond the BOM import.

A `dependencyManagement` block there carrying type plus exclusions once per GA would reduce both consumers to groupId/artifactId, and make "which plugin zips we ship" one list instead of two kept in sync by memory.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml
Okhelifi: Fixed, with one entry per module known to own a schema, plus a negative test to prove it turns red: building with -Dgravitee-gamma-module-aim.version=1.0.0-alpha.123, a version that ships no liquibase/, now fails on module-changelogs-were-collected. Without that check a mis-merged execution would have looked exactly like a passing build.

Two things worth stating rather than leaving implicit.

The third scenario you list, a fifth module adopting JDBC without being added to the dependency list, stays undetected. That is deliberate, not an oversight: a list of expected files cannot detect the absence of a file it does not know to expect. Catching it would mean discovering the modules that ship liquibase/ dynamically, which means resolving every Gamma module rather than the four we bundle. Given how few modules there are and how slowly that list moves, the maintenance cost of the known-list is lower than the machinery that would replace it. Adding a module to the bundle and forgetting to add it here is the same class of omission as adding it to the standalone distribution and forgetting the assembly.

And the scope narrowed slightly compared to what you had in mind: since the Gamma dependencies moved into bundle-default for your other comment, the assertion lives there too and only runs with -Dbundle. Both jobs that produce a shippable archive pass it, so CI is covered, but a local build without the flag no longer checks module coverage.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/README.md
Okhelifi: I hope nobody in the support will never do that and transmit a local archive to a customer 😅

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/README.md
Okhelifi: tested locally with mysql so a note has been added in the readme for the users ending with a `[Failed SQL: (3750) CREATE TABLE gravitee.aim_databasechangelog (ID VARCHAR(255) NOT NULL, ... DEPLOYMENT_ID VARCHAR(10) NULL)]`

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/README.md
Okhelifi: Settled by removal rather than by more prose. changelog-sync is gone from the README: the documented flow never needed it — the files update-sql generates carry their own INSERT INTO databasechangelog, so applying them leaves the tracking table correct — and its only use case is an operator who brought the schema up to date by other means, who does not need us to hand them the footgun. On the race with the application's own run: this archive's audience runs management.jdbc.liquibase=false by definition, so the application never runs Liquibase and there is nothing to race. The MySQL auto-commit point is real but is generic Liquibase-on-MySQL behaviour, not specific to this archive; the DBA running the migration owns it.

gravitee-apim-distribution/gravitee-apim-distribution-es-index-mappings/pom.xml
Okhelifi: Fair on both counts, with one correction on the framing. This does not move a public download path: graviteeio-apim/es-index-mappings/ never existed. The publication job stages the zips and has not uploaded them since d9c7d3dbdb (March 2025) — which is what #19185 restores. So the property was inert, and this changes its value so that both archives land in one resources/ folder from the first release that actually uploads them. It is already its own commit (3fc478df78, the two publish-folder-path lines only), and it stays in this PR because a shared folder only makes sense once there is a second archive to share it with. The PR description will be fixed as soon as GH is not in trouble anymore.

gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml
Okhelifi: Agreed on the diagnosis, and deferred. A dependencyManagement that owns "the plugin zips we ship" only pays off at full scope: the standalone pom declares 201 zip dependencies, 197 with the same wildcard exclusion. Managing five here and leaving 196 there would add a second list, not remove one. That refactor is worth doing — even more so now that the Gamma modules sit in a bundle-default profile mirroring the standalone's — but not inside a PR that adds a module. Follow-up ticket to move the whole catalogue under dependencyManagement in gravitee-apim-distribution/pom.xml.

ytvnr (approved):

PR #19176, opened 2026-08-14
perf(rest-api): give the rest-api test job back the time it spends waiting

phiz71:
## Issue

https://gravitee.atlassian.net/browse/BX-294

## Description

`Test rest-api` spent most of its time on things that were not tests: work `Build backend` had already done, modules the job is not there to exercise, and a containerised module waiting on a broker that was never coming.

**`Test rest-api` goes from 767 s to 497 s, a 35% reduction.** The reference is the median of 843 runs on master, not a single build; the branch figure is stable across five runs (497, 498, 506, 510, 513 s) against a 13% machine variance on this fleet.

But the job is not the deliverable — the chain is. `Sonar - gravitee-apim-rest-api` requires `Test rest-api`, which requires `Build backend`, so the three are one path and only their sum is visible to anyone waiting:

| Chain `Build backend` → `Test rest-api` → `Sonar - rest-api` | |
| --- | --- |
| master (medians over 839–907 runs) | 553 + 767 + 387 = **1 707 s** |
| this branch | 560 + 497 + 327 = **1 384 s** |
| | **−323 s, −19%** |

The pipeline's own wall clock lands at **27.1 min**, and the whole run is 35 jobs green.

Each commit was pushed and measured on its own before the next one was written. What each one is worth:

| Commit | What it changes | Effect |
| --- | --- | --- |
| `f94a8c15` ci(gamma): move gamma out of the rest-api reactor into jobs of its own | Gamma builds and tests in two dedicated jobs | **−9.8%** on the 48 modules the two runs share |
| `c0df4c1f` perf(rest-api): generate the v2 models where their specs live | 22 OpenAPI specs move to a model module of their own | **−49 s** — generation starts in the opening wave instead of behind `service` |
| `89c4f707` perf(rest-api): import the architecture model once per JVM | ArchUnit imports shared across rules | **−9 s** — each fork still pays its own import |
| `8088cfe2` ci(kafka-explorer): give the only containerised module a job of its own | The one module needing Testcontainers leaves the reactor | **−86 s** |
| `792f1e3d` test(kafka-explorer): assert the 502 without waiting for a broker to time out | 46 s of connection timeout become three injected failures | **46 s → 1.3 s** on `Test kafka-explorer` |
| `ec5444af` ci(validate): run the validation job on a large executor | CPU pinned at 100%, memory at 48% | **−53 s** of critical path, **+20 credits** per pipeline |
| `96847ff6` ci(gamma): run the gamma test job on a medium executor | Was using 45% CPU and 20% memory of a large | **+17 s** on that job — invisible, it runs at a quarter of `Test rest-api` — and **−43% credits** |
| `c0041d75` chore(deps): bump expression-language to 4.4.1 and drop the jsonpath load-order hack | Removes the load-order race the next commit would otherwise lose | No direct gain — it is what makes the transport possible |
| `2c78f850` perf(ci): test the rest-api modules without compiling them again | `Build backend` persists the compiled classes; the test job runs goals instead of the `test` phase | **−159 s (−27%)**, no module slower |

The gains do not add up arithmetically: each was measured in the state of the branch at the time, and removing work from one module changes what the next one waits for.

A tenth commit, isolating the security context between test classes, was merged separately on master while this branch was in review and no longer appears here.

**`f94a8c15` also changes what triggers what**, which the table above does not convey. A new `shouldTestGamma` predicate decides whether the two Gamma jobs run:

- a **Gamma-only change now runs tests**. It previously ran none — `shouldTestRestApi` does not match `gravitee-gamma`, and `shouldBuildBackend` compiles it with tests skipped. This closes a gap the ticket uncovered rather than one it set out to fix.
- a **rest-api change now also blocks on `Test gamma` and `Test gamma UI`**, because the predicate ORs in `shouldTestRestApi` — Gamma depends on rest-api, so a rest-api change can break it.

Which triggers run which jobs belongs to BX-292, so this is ground taken early: worth knowing for whoever picks that epic up.

### The side effect the transport had on Sonar, and what it cost

Handing the compiled classes to the test job also handed them to the analysis job, which attaches the same workspace. `sonar.java.binaries=**/target/**` had been matching nothing for years; it suddenly matched, the JavaSensor switched to type-resolved analysis, and `Sonar - rest-api` went **from 387 s to 1 170 s**.

Because it sits at the end of the same chain, that turned a 4-minute saving into a 7-minute loss. On the run before the fix, `Sonar - rest-api` was the only job still running for the **last 17.8 minutes** of the pipeline — it finished at 41.2 min when the next-latest job finished at 23.4. That number is a regression this branch introduced and then removed; it is not what master does.

Three commits deal with it, and only the last one survives in behaviour:

| Commit | Effect |
| --- | --- |
| `42afdb6c` ci(sonar): keep the compiled classes out of the analysis input | Dropped the classes before the scanner ran. **1 170 s → 319 s** |
| `99f3ed18` ci(sonar): drop the classes with globs, the scanner image ships no find | `sonarsource/sonar-scanner-cli:11.2` has no `find`; the first attempt exited 127 on all nine analyses |
| `e2b6f0bd` refactor(ci): carry the compiled classes in a cache only the test job reads | Treats the cause instead: the classes never reach the shared workspace, so that step is gone. Chain 1 367 → 1 384 s, wall clock 27.5 → 27.1 min — **neutral in time**, and some thirty jobs stop downloading hundreds of megabytes they never open |

Worth recording for whoever looks at Sonar next: **no APIM module has ever been analysed with its bytecode.** `sonar.java.binaries` matched nothing anywhere, so the rules that need resolved types have been silent across the whole codebase. Turning that on is a defensible thing to want — this branch puts a number on it: roughly +800 s on rest-api alone.

### Following the review

Six commits answer [@ytvnr's review](https://github.com/gravitee-io/gravitee-api-management/pull/19176), and two more fix what those broke on their way through CI. Only the first has a cost, stated below.

| Commit | Why |
| --- | --- |
| `50fbace0` fix(ci): aggregate coverage over the modules it aggregates | `report-aggregate` on the coverage module alone left an empty input set — a green job publishing 0% rest-api coverage. Verified on a run: 0 bundles before, **91 bundles and 71 193 classes** after. Costs ~60 s, the price of a report that exists |
| `aaabf222` test(rest-api): stop isolating UserServiceTest in its own surefire execution | The split was the other half of the workaround this branch retires, and a CLI `surefire:test` had already stopped honouring it. Verified unnecessary: green alone and inside the module's full 9254-test suite |
| `886e4083` test(kafka-explorer): drive the real service once, at an address that cannot answer | The injected failures left the service's error-to-`TechnicalCode` translation untested. A `.invalid` bootstrap fails instantly, so the real path runs without waiting on a handshake |
| `81250975` ci(gamma): report the gamma UI test results and coverage | The new job ran Jest and reported nothing; it now emits junit through jest-junit and stores the reports and lcov. Verified on a run: **1 913 tests** in the Tests tab |
| `dcf0f8cc` ci(gamma): stop persisting the gamma coverage and align the nx outputs | Nothing consumes those lcov files — there is no Gamma analysis. The Nx `outputs` also disagreed with Jest's `coverageDirectory`, so a cache hit would have emptied the Tests tab silently |
| `62eaf50e` test(kafka-explorer): name what the two accepted codes actually pin | `isIn(TIMEOUT, CONNECTION_FAILED)` cannot fail, since `CONNECTION_FAILED` is the fallback. Tightening it would make the test flaky on a slow agent, so the trade-off is documented in the test rather than hidden |
| `e711496a` fix(ci): keep the new test and the gamma jest configs green in CI | Follow-up on my own two mistakes, both caught on the first run: the junit reporter used `__dirname`, copied from a CommonJS config into ES modules, and a test line sat at exactly the 140-character limit |
| `204e998f` style(kafka-explorer): use the braced lambda form prettier settles on | A multi-line lambda expression inside `satisfies()` is formatted differently depending on the node prettier-java runs under, so the file passed locally and failed in CI |

The transport is the largest single win on the test job, because it takes time off the two modules that hold the chain:

| Module | Before | After |
| --- | --- | --- |
| Management API - Service | 218 s | 145 s |
| Management - v2 - Rest | 146 s | 106 s |
| Management - Rest (v1) | 128 s | 95 s |
| Portal - Rest | 109 s | 86 s |

Two commits cost something, and both are deliberate. `ci(validate)` buys 53 seconds of critical path for about 20 credits a pipeline — roughly three quarters of that is paid back by `ci(gamma)`, which gives up 17 invisible seconds for 43% of its own cost. If the budget matters more than the wall clock, `ec5444af` is the one to drop; nothing else depends on it.

### Following the third review

Four more commits answer [@ytvnr's third pass](https://github.com/gravitee-io/gravitee-api-management/pull/19176), and a fifth takes out the duplication the whole thing left behind. None of them costs wall clock.

| Commit | Why |
| --- | --- |
| `cdee5509` fix(ci): run every backend suite in the nightly build too | The nightly spelled its suites out, so kafka-explorer and gamma stopped running at night the moment they were split into jobs of their own. Both workflows now read one list |
| `cc784df6` fix(rest-api): drop the jsonpath class-loading hack the 4.4.1 bump made dead | Twin of the block removed from `UserServiceImpl`. Settled on evidence: in the 4.4.1 jar, `JsonPathFunction`'s static block only builds a `Configuration`, and `setCache` does not appear — neither block had a race left to win |
| `8b515de3` perf(ci): let the gamma UI job start without waiting for the maven build | It checks out, installs the yarn workspace and runs nx. Nothing it reads comes from `Build backend`, whose workspace it was attaching for nothing |
| `e2b6f0bd` refactor(ci): carry the compiled classes in a cache only the test job reads | Covered above |
| `2d208592` refactor(ci): declare every suite once, analysis included | Having an analysis had become a category — its own type, list and emission loop — where it is an attribute. One `Suite` type with an optional `sonar`, one list per side, one shared emission function: **−70 lines**, and the front-end suite parameters, which were written twice, are written once. Generated configuration verified unchanged |

Two things that surfaced and were deliberately left out. **`Lint & test APIM Libs` and `Lint & test Gamma Console` do not run at night** — the same pattern as kafka-explorer and gamma, on the front-end side; it predates this branch and adding them changes behaviour. And **`analysed-projects.ts` now holds every suite, not only the analysed ones**, so its name is due for a change once the console branch lands.

## Additional context

**Everything here is bound by the critical path through the module graph, never by cores.** `Test rest-api` has a capacity floor of 192 s against a 542 s chain. That is the rule the whole branch follows: a module that is not on the chain costs nothing, however slow it looks in the Reactor Summary.

It also rules things out. **`forkCount` was tried twice on `service` and was worse both times** (+144% per test); an `xlarge` executor would do nothing for the same reason. Restoring the yarn workspace cache in `Build backend` cost 80 seconds — that install was never on the chain. Lazy bean initialisation in the v2 test context was neutral. All four were measured and dropped rather than argued about. The same rule says extending the transport to gateway or repository would not move the wall clock either: those chains already finish at 22.0 and 23.8 minutes, well inside the rest-api one. It would be a credits question, and it would have to be weighed against the larger workspace all 34 jobs attach.

**The dependency bump is not incidental.** Transporting compiled classes changes the order in which classes initialise inside each Surefire fork, and `JsonPathFunction` lost the race for JsonPath's single per-JVM cache registration — taking ten `UserServiceTest` cases with it. Investigating that turned up a thread-safety defect in the shared compiled path (34% crossed results between concurrent evaluations of a path carrying a function), fixed upstream in expression-language 4.4.1. The `// Dirty hack` in `UserServiceImpl` existed only to win that race and goes away with it.

**One review point was left alone on purpose.** `KafkaExplorerResource` declares its dependencies as private `@Inject` fields with no constructor or setter, so the test fills them reflectively; there is no construction path the compiler could check short of standing up an injection container for a unit test.


.circleci/ci/src/jobs/backend/job-test-rest-api.ts
```diff
@@ -13,27 +13,43 @@
...
+          // above it, so the jacoco prefix does not resolve there and Maven would either fail outright or
+          // reach for the latest release instead of the version the build pins.
           name: `Run Rest API tests`,
-          command: `mvn --fail-fast -s ${config.maven.settingsFile} test --no-transfer-progress -Drest-api-modules -Dskip.validation=true -Dgravitee.archrules.skip=true -T 2C`,
+          command: `mvn --fail-fast -s ${config.maven.settingsFile} dependency:properties jacoco:prepare-agent surefire:test --no-transfer-progress -Drest-api-modules -pl '!.,!gravitee-apim-rest-api/gravitee-apim-rest-api-kafka-explorer' -nsu ${mavenParallelism('large')}`,
+        }),
+        new commands.Run({
+          // report-aggregate is bound to the test phase, which no longer runs.
+          name: `Aggregate coverage`,
+          command: `mvn -s ${config.maven.settingsFile} -pl gravitee-apim-rest-api/gravitee-apim-rest-api-coverage org.jacoco:jacoco-maven-plugin:report-aggregate --no-transfer-progress -Drest-api-modules -nsu`,
```
ytvnr: **⚠️ Important** - Running `jacoco:report-aggregate` with `-pl gravitee-apim-rest-api-coverage` produces an empty aggregate report, which silently zeroes the rest-api coverage reported to SonarCloud.

`ReportAggregateMojo` builds its input set from `getProject().getDependencies()` and keeps only the ones it can find in `${reactorProjects}`, i.e. the projects of the *current* session. On master the goal ran inside the full `-Drest-api-modules` reactor (bound to `test` in `gravitee-apim-rest-api-coverage/pom.xml`), so every dependency module was present with its `target/jacoco.exec`. With `-pl` narrowing the session to the coverage module alone, the ~50 dependencies declared in that pom resolve from the local repository, not from the reactor, and contribute nothing. `gravitee-apim-rest-api/sonar-project.properties` points Sonar at `jacoco-aggregate/jacoco.xml`, so this fails green with 0% coverage rather than erroring.

Suggested direction: keep the coverage module's session populated with the modules it aggregates, either by re-running the goal over the same module selection used for the tests, or by going through a phase in the full reactor as before. Worth checking on a pipeline run that `jacoco.xml` is non-empty before merging.

.circleci/ci/src/jobs/backend/job-test-rest-api.ts
```diff
@@ -13,27 +13,43 @@
...
+          // management-v2-rest on kafka-explorer, which has a job of its own — all three come from the
+          // local repository. -nsu keeps a published snapshot from taking the place of the one
+          // `Build backend` just installed. Yarn went with Gamma: nothing here needs it.
+          //
+          // -pl '!.' drops the aggregator from the reactor. It is the one pom with no gravitee-apim-parent
+          // above it, so the jacoco prefix does not resolve there and Maven would either fail outright or
+          // reach for the latest release instead of the version the build pins.
           name: `Run Rest API tests`,
-          command: `mvn --fail-fast -s ${config.maven.settingsFile} test --no-transfer-progress -Drest-api-modules -Dskip.validation=true -Dgravitee.archrules.skip=true -T 2C`,
+          command: `mvn --fail-fast -s ${config.maven.settingsFile} dependency:properties jacoco:prepare-agent surefire:test --no-transfer-progress -Drest-api-modules -pl '!.,!gravitee-apim-rest-api/gravitee-apim-rest-api-kafka-explorer' -nsu ${mavenParallelism('large')}`,
```
ytvnr, line 57: **Worth fixing** - A CLI `surefire:test` runs under the `default-cli` execution, so the `default-test` exclude and the isolated `user-service-test` execution in `gravitee-apim-rest-api-service/pom.xml` (around lines 445-466) no longer apply in CI.

That pom deliberately splits its suite: `default-test` excludes `**/UserServiceTest.java` and a second execution re-runs only that class. Both configurations live inside `<executions>`, which Maven applies only to the matching execution; a goal invoked from the command line sees just the plugin-level `<configuration>` (`forkCount=2`, `reuseForks=true`, the shared `argLine`). `UserServiceTest` therefore now runs in the shared fork pool. That split came from `df8afeb634`, the same commit that added the `JsonPathFunction` static block this branch removes, so the isolation and the load-order hack were two halves of one workaround: one is deleted deliberately, the other disappears as a side effect. Locally `mvn test` still honours it, so CI and developer machines now run different partitions.

Suggested direction: decide explicitly whether the isolation is still needed under expression-language 4.4.1. If not, delete the two executions in the same change so the pom stops describing a behaviour CI no longer performs; if it is, the test command has to trigger the `test` phase for that module.

gravitee-apim-rest-api/gravitee-apim-rest-api-kafka-explorer/src/test/java/io/gravitee/rest/api/kafkaexplorer/resource/KafkaExplorerResourceErrorTest.java
```diff
@@ -0,0 +1,103 @@
...
+        clusters.create(
+            Cluster.builder()
+                .id(CLUSTER_ID)
+                .environmentId(ENVIRONMENT_ID)
+                .configuration(Map.of("bootstrapServers", "unreachable:9092"))
+                .build()
+        );
+
+        var clusterService = mock(KafkaClusterDomainService.class);
+        when(clusterService.describeCluster(any())).thenThrow(failure);
```
ytvnr: **Worth fixing** - The replacement test mocks `KafkaClusterDomainService`, so nothing verifies any more that a SASL-on-plaintext or SSL-on-plaintext misconfiguration is actually turned into a `KafkaExplorerException`.

The two deleted `WrongProtocol` cases drove the real admin client against the plaintext listener and asserted the resulting 502. The new test stubs `clusterService.describeCluster(any())` to throw a ready-made `KafkaExplorerException`, so it only re-asserts the resource's exception to 502 mapping, a path already covered by `should_return_502_with_wrong_credentials` in the integration test. The translation layer between the Kafka client and `TechnicalCode` is now untested for protocol mismatch. The class javadoc says the integration test "covers the handshakes that only a real broker can perform", but those are precisely the handshakes that were removed.

Suggested direction: either keep one real-broker protocol-mismatch case (the timeouts were the cost driver, a rejected handshake fails fast), or drive the real `KafkaClusterDomainService` with an unreachable or plaintext bootstrap and a short client timeout so the mapping itself is still exercised.

.circleci/ci/src/jobs/backend/job-test-gamma-ui.ts
```diff
@@ -0,0 +1,61 @@
...
+      }),
+      new commands.Run({
+        // Jest sizes its worker pool from os.cpus(), which inside a container reports the host's cores
+        // rather than the resource class — the same trap mavenParallelism exists to avoid for -T. Left
+        // alone it opens dozens of jsdom environments on a four-core executor.
+        name: 'Run unit tests',
+        command: `yarn nx run-many --target=test --projects=${projects} --maxWorkers=2`,
+      }),
+      new reusable.ReusedCommand(notifyOnFailureCmd),
+    ];
```
ytvnr, line 61: **Worth fixing** - The new job runs Jest but emits no `store_test_results`, no coverage artefact and no license lint, so its failures are visible only in raw log output.

`job-webui-lint-test.ts`, the pattern this job mirrors, ends with a coverage persist, `StoreArtifacts` and `StoreTestResults`, and starts with a `Check License` step. The generated snapshot confirms the gap: `pull-requests-custom-branch-backend-rest-api-only.yml` ends `job-test-gamma-ui` at `cmd-notify-on-failure`. Since these suites never ran before, this is the first chance to wire them into the Tests tab and into coverage. The Java sibling `job-test-gamma.ts` does collect surefire reports via `AbstractTestJob`.

Suggested direction: align the job with the existing UI test jobs, i.e. run the Nx `test` target with coverage, store the junit report, and persist the lcov so the Gamma modules can later join a Sonar project.

gravitee-apim-rest-api/gravitee-apim-rest-api-kafka-explorer/src/test/java/io/gravitee/rest/api/kafkaexplorer/resource/KafkaExplorerResourceErrorTest.java
```diff
@@ -0,0 +1,103 @@
...
+    @MethodSource("brokerFailures")
+    void should_return_502_carrying_the_technical_code(TechnicalCode technicalCode, String message) throws Exception {
+        var resource = resourceFailingWith(new KafkaExplorerException(message, technicalCode));
+        GraviteeContext.setCurrentEnvironment(ENVIRONMENT_ID);
+
+        var response = resource.describeCluster(new DescribeClusterRequest().clusterId(CLUSTER_ID));
+
+        assertThat(response.getStatus()).isEqualTo(502);
+        assertThat(response.getEntity())
+            .asInstanceOf(org.assertj.core.api.InstanceOfAssertFactories.type(KafkaExplorerError.class))
```
ytvnr: Minor: `org.assertj.core.api.InstanceOfAssertFactories.type(...)` is written inline while every other symbol in the file is imported, including the AssertJ `assertThat` static import. A static import would read more like the rest of the file.

Same spirit, also minor: the subject is built through reflective field writes (`injectField`, lines 98-102), which will fail at runtime rather than at compile time if `KafkaExplorerResource`'s fields are renamed. A construction path the compiler can check would be safer, if one is available.

.circleci/ci/src/workflows/groups/changed-files.ts
```diff
@@ -158,6 +158,13 @@ export function shouldTestPlugin(changedFiles: string[]): boolean {
   );
 }
 
+// Gamma builds against the rest-api modules, so anything that wakes up the rest-api tests wakes up
+// these too. The narrowing works the other way round: a change confined to gravitee-gamma no longer
+// drags in the rest-api suites, and — until this predicate existed — ran no test at all.
+export function shouldTestGamma(changedFiles: string[]): boolean {
+  return shouldTestRestApi(changedFiles) || changedFiles.some((file) => file.includes('gravitee-gamma'));
+}
```
ytvnr, line 166: `shouldTestGamma` is new job-triggering logic, and BX-294 explicitly defers "which triggers run which jobs" to BX-292.

The predicate is a genuine improvement: it closes the gap the ticket only *uncovered* (a Gamma-only change previously ran no test, because `shouldTestRestApi` does not match `gravitee-gamma`). But it also widens the other direction: because it ORs in `shouldTestRestApi`, any rest-api change now blocks `Validate workflow status` on `Test gamma` and `Test gamma UI` as well (`backend-jobs.ts`), which the PR description does not mention; the commit table describes `43c612c1` only as "Gamma builds and tests in two dedicated jobs".

Suggested direction: call the trigger change out explicitly in the PR description, what now runs that did not, and what now runs on rest-api changes that did not, so the reviewers of BX-292 know this ground has been taken.

ytvnr (commented):
Read through the whole branch: the reactor changes, the new Gamma / Kafka Explorer jobs, the compiled-classes transport and the expression-language bump. The job wiring is consistent across every workflow that used to cover those modules, and I found no stale reference to the old v2 spec location.

Two points matter more than the rest, both on `job-test-rest-api.ts`:

- the new `Aggregate coverage` step runs `report-aggregate` with `-pl` on the coverage module alone, so its session contains none of the modules it aggregates and the report is empty. SonarCloud reads that file, so the failure mode is a green job with 0% rest-api coverage. Worth verifying on a pipeline run that `jacoco.xml` is not empty before this merges.
- switching from the `test` phase to a CLI `surefire:test` silently drops the execution-scoped surefire config in `rest-api-service` (the `UserServiceTest` isolation), which was the other half of the workaround this branch retires on purpose.

The remaining comments are smaller: coverage of the Kafka protocol-mismatch path, reporting on the new Gamma UI job, one nit on the new test, and a note on the trigger predicate versus the ticket scope. None of them is meant to block; posting as a comment so the review state stays with whoever approves.

.circleci/ci/src/jobs/backend/job-test-rest-api.ts
phiz71: Confirmed and fixed in c8e1e5f9 — you were right down to the mechanism.

The evidence: the last run before the fix logged **no** \`Analyzed bundle\` line at all, where the pre-transport run logged 34. The goal now runs over the same module selection as the tests, so the coverage module finds its dependencies in the session again.

Worth adding, because it explains why nobody noticed: the rest-api Sonar job is currently timing out (its 10-minute default — the transport puts the compiled classes in the workspace, so `sonar.java.binaries=**/target/**` now matches and the JavaSensor goes from 197 s to over 634 s). The empty report had therefore never been published. The two defects were hiding each other, and fixing the timeout alone would have dropped coverage to zero.

.circleci/ci/src/jobs/backend/job-test-rest-api.ts
phiz71: Both halves of your reading are right, including the archaeology — I had not made the link to `df8afeb634`.

Decided explicitly, as you suggest, and the branch already had the answer: under expression-language 4.4.1 the isolation has nothing left to protect. `UserServiceTest` passes green in the shared fork pool, both on its own (92 tests) and inside the module full 9254-test suite. So the two executions are deleted in 4d5b27a5 rather than restored — the pom stops describing a partition CI had silently abandoned, and local runs match CI again.

gravitee-apim-rest-api/gravitee-apim-rest-api-kafka-explorer/src/test/java/io/gravitee/rest/api/kafkaexplorer/resource/KafkaExplorerResourceErrorTest.java
phiz71: Fair, and the class javadoc I wrote made it worse — it claimed the integration test covered the handshakes only a real broker can perform, which were exactly the ones I had deleted.

Fixed in 5086f2bf, taking your second option. One case now drives the real `KafkaClusterDomainServiceImpl` at a bootstrap address in `.invalid`, which RFC 2606 reserves as never resolvable: the client fails immediately instead of waiting out a handshake, so the error-to-`TechnicalCode` translation runs for real. Four tests, 1.2 s.

It does not restore the SASL-on-plaintext and SSL-on-plaintext cases specifically — those need a listener to refuse the handshake. What it restores is the translation path itself, which is what your comment identified as untested.

.circleci/ci/src/jobs/backend/job-test-gamma-ui.ts
phiz71: Fixed in d4b652b0, and you were right that this is the moment to wire them in — the suites had never run anywhere.

Both modules now emit a junit report through `jest-junit`, with the same reporter options and layout as the console and portal configs, and the job runs with `--coverage`, stores the reports, keeps the lcov as an artefact and persists it so the modules can join a Sonar project later.

One correction to the finding: the license check was already running. `nx.json` makes the `lint` target fan out to `lint-eslint`, `lint-prettier` **and** `lint-license`, so the job has been checking licences since it was written — its step name, "Run Prettier and ESLint", was simply wrong. Renamed rather than duplicated.

gravitee-apim-rest-api/gravitee-apim-rest-api-kafka-explorer/src/test/java/io/gravitee/rest/api/kafkaexplorer/resource/KafkaExplorerResourceErrorTest.java
phiz71: The static import is fixed in 5086f2bf.

The reflective injection stays, and your "if one is available" was the right hedge: `KafkaExplorerResource` declares its dependencies as private `@Inject` fields with no constructor and no setter, so there is no construction path the compiler could check short of standing up an injection container inside a unit test. Noted in the PR description so it reads as a deliberate choice rather than an oversight.

.circleci/ci/src/workflows/groups/changed-files.ts
phiz71: Right on both counts, including the direction I had not mentioned. The PR description now carries it explicitly:

- a Gamma-only change now runs tests, where it previously ran none;
- a rest-api change now also blocks on `Test gamma` and `Test gamma UI`, because the predicate ORs in `shouldTestRestApi` — Gamma depends on rest-api, so a rest-api change can break it.

With the note that this is BX-292 ground taken early, so whoever picks that epic up knows it has moved. Happy to pull the predicate out into its own PR if you would rather keep the scope clean — it stands on its own.

ytvnr: Either way works for me, your call. Keeping it here is fine now that the description carries both directions; splitting it out only if you want BX-292 to own it from the start.

.circleci/ci/src/jobs/backend/job-test-rest-api.ts
ytvnr: Thanks for the extra context on the Sonar job, that explains a lot. One question it leaves open: is the timeout itself addressed on this branch (a narrower `sonar.java.binaries` now that the compiled classes are in the workspace, or a raised limit), or is it left for a follow-up? As it stands, once merged the rest-api analysis would still not publish, so the coverage you just restored would not reach SonarCloud either.

ytvnr: Looked into the mechanism behind the timeout, since it follows from this branch rather than from Sonar itself.

`job-sonarcloud-analysis.ts` does `checkout` then `workspace.Attach({ at: '.' })`, and `backend-jobs.ts` wires the rest-api Sonar job with `requires: ['Test rest-api']`, so it sees everything `Build backend` now persists. The four new globs in `job-build-backend.ts` (around lines 73-76) put every rest-api `target/classes` and `target/test-classes` into that workspace, and `gravitee-apim-rest-api/sonar-project.properties` sets `sonar.java.binaries=**/target/**` with no `sonar.java.test.binaries`, so the JavaSensor now reads main and test classes alike. On master `Build backend` persisted only `console-openapi.*` and the two distributions, so the same pattern matched almost nothing; that is consistent with your 197 s to 634 s figure. The `Run Sonarcloud Analysis` step has no `no_output_timeout`, so it dies at the 600 s default.

Possible directions: keep the transported classes out of the workspace the Sonar job attaches, or narrow the pattern to `**/target/classes` and move test classes to `sonar.java.test.binaries`, and raise the step timeout to whatever the real duration ends up being.

.circleci/ci/src/jobs/backend/job-test-gamma-ui.ts
```diff
@@ -0,0 +1,70 @@
...
+      }),
+      new reusable.ReusedCommand(notifyOnFailureCmd),
+      ...TestGammaUiJob.projects.flatMap((project) => [
+        new commands.StoreTestResults({ path: `gravitee-gamma/${project}/coverage/junit.xml` }),
+        new commands.StoreArtifacts({ path: `gravitee-gamma/${project}/coverage/lcov.info` }),
+      ]),
+      new commands.workspace.Persist({
+        root: '.',
+        paths: TestGammaUiJob.projects.map((project) => `gravitee-gamma/${project}/coverage/lcov.info`),
+      }),
```
ytvnr: **Worth fixing** - This `persist_to_workspace` of the two gamma `lcov.info` files has no downstream consumer yet, so for now it adds a failure mode without adding a signal.

In `job-webui-lint-test.ts` the identical step is labelled "For Sonar analysis" and consumed by a frontend `SonarCloudAnalysisJob`. The gamma block in `backend-jobs.ts` adds no Sonar job and there is no `sonar-project.properties` under `gravitee-gamma/`, so nothing attaches these files, and `store_artifacts` already covers the stated goal. Your own comment in `abstract-job-test.ts` states the cost: a persist whose paths match no file fails the job. The path is also not a declared Nx output: both `project.json` files declare `outputs: ["{workspaceRoot}/coverage/{projectRoot}"]` while jest writes to `<rootDir>/coverage`, so a cache replay of `test` would leave the persist nothing to match.

I read your "so the modules can join a Sonar project later" as deliberate, so this is really about the Nx outputs mismatch: either drop the persist until a gamma Sonar job lands, or align the `test` target's declared outputs with the directory jest writes to so a cache hit cannot break the job.

gravitee-apim-rest-api/gravitee-apim-rest-api-kafka-explorer/src/test/java/io/gravitee/rest/api/kafkaexplorer/resource/KafkaExplorerResourceErrorTest.java
```diff
@@ -0,0 +1,131 @@
...
+        assertThat(response.getEntity())
+            .asInstanceOf(type(KafkaExplorerError.class))
+            .satisfies(error -> {
+                assertThat(error.getTechnicalCode()).isEqualTo(technicalCode.name());
+                assertThat(error.getMessage()).isEqualTo(message);
+            });
+    }
+
+    /**
+     * The cases above inject the failure, so they say nothing about the service turning a real client
```
ytvnr, line 83: Small one: `isIn(TIMEOUT, CONNECTION_FAILED)` means the test cannot fail for the reason it was written, because `CONNECTION_FAILED` is the catch-all branch of `withAdminClient`.

`KafkaClusterDomainServiceImpl.withAdminClient` maps `ExecutionException`, `TimeoutException` and `InterruptedException` to specific codes and everything else to `CONNECTION_FAILED`. Any regression (a new exception type, a broken `buildProperties`, a mapper change) still lands on `CONNECTION_FAILED` and still satisfies the assertion; the only thing genuinely pinned is the 502, which the three parameterized cases already cover. I take it the two-value form absorbs DNS variance (`.invalid` normally fails at `AdminClient` construction, a hijacking resolver runs out the 7 s `defaultGetTimeout`).

If there is a bootstrap address that reliably refuses the connection rather than failing to resolve, a single expected code could be asserted; otherwise fine as is, just naming the trade-off.

ytvnr (commented):
Second pass on the six commits pushed since the first review. On my side: the coverage aggregation, the `UserServiceTest` isolation, the gamma UI reporting, the import and the description update all read as closed. `1486e625` (prettier width plus `__dirname` to `fileURLToPath` in the ESM jest configs) is sound.

Two things remain, both left as non-blocking comments: the Sonar timeout is now explained on the original thread and still needs an answer on this branch, and the gamma UI persist has no consumer yet. One smaller note on the new Kafka test assertion. Still posting as a comment, not a review state.

.circleci/ci/src/jobs/backend/job-test-gamma-ui.ts
phiz71: Both halves confirmed, and I did both in `dcf0f8cc`.

The persist is gone. You read it right that "so the modules can join a Sonar project later" was deliberate — which is exactly what was wrong with it: no gamma Sonar job, no `sonar-project.properties` under `gravitee-gamma`, so I was carrying a failure mode for a use that does not exist. `store_artifacts` already covers what I actually wanted.

The Nx outputs are aligned too, because the mismatch bites wider than the persist. `{workspaceRoot}/coverage/{projectRoot}` against jest's `<rootDir>/coverage` means a cache replay also loses `junit.xml` — and that one does not fail the job, it just empties the Tests tab in silence. That is the worse of the two, and it was the whole point of the commit that added them.

`gravitee-gamma-control-plane-webui` carries the same mismatch but never runs with `--coverage`, so I left it alone rather than widen the diff.

gravitee-apim-rest-api/gravitee-apim-rest-api-kafka-explorer/src/test/java/io/gravitee/rest/api/kafkaexplorer/resource/KafkaExplorerResourceErrorTest.java
phiz71: You are right, and your reading of why is right too: the two-value form absorbs resolver variance.

I went looking for the address you suggest — one that refuses the connection instead of failing to resolve, so a single code could be pinned. A closed port on loopback is the obvious candidate, but it trades a weak assertion for a flaky one: the client retries until `defaultGetTimeout`, and which of `TIMEOUT` or `CONNECTION_FAILED` comes back still depends on timing rather than on the mapping. Nothing gained.

So I took your other option and named it, in `62eaf50e`: the javadoc now says the test pins that the service maps a real client failure at all, not which branch it took, and why narrowing it is not worth a flaky test. Better than leaving the next reader to work out that `CONNECTION_FAILED` is the catch-all.

The three parameterized cases stay the ones that actually pin the 502, as you say.

.circleci/ci/src/jobs/backend/job-test-rest-api.ts
phiz71: Both of your directions were on the table; we took the first one — the classes never reach the scanner.

| `Sonar - gravitee-apim-rest-api` | |
| --- | --- |
| master, median of 839 runs | 387 s |
| with the transported classes in the workspace | 1 018 s, then 1 170 s |
| after `42afdb6c` + `99f3ed18` | **319 s** |

The part that mattered more than the job itself: on the run before the fix, the analysis was the only job still running for the **last 17.8 minutes** of the pipeline — it finished at 41.2 min where the next-latest job finished at 23.4. Pipeline wall clock is now **27.5 min**, and the chain `Build backend` → `Test rest-api` → `Sonar` goes from 1 707 s to 1 367 s against master.

On the timeout: it is addressed, but by no longer being needed. Master added an explicit `no_output_timeout: 30m` on that step (`776bc0f7`) and this branch picked it up in the rebase, so the 600 s default is out of the picture either way.

Your second direction — narrowing to `**/target/classes` and moving test classes to `sonar.java.test.binaries` — we deliberately did not take, and the reason is worth recording. `sonar.java.binaries` has been matching nothing on **every** APIM module, not only rest-api: `Build backend` persisted just `console-openapi.*` and the two distributions, and no other test job persists its classes at all. Type-resolved analysis has therefore never run anywhere in this codebase, and every rule that needs resolved types has been silent since the beginning. Turning it on is a defensible thing to want, but it is a Sonar scope decision rather than a CI one, and this branch now puts a number on it: roughly +800 s on rest-api alone.

The trim sits in the shared `job-sonarcloud-analysis.ts`, before the scanner, so it also covers whoever extends the transport to another module later. It drops `target/classes` and `target/test-classes` only — the `target` directories survive, so the scanner never asks for binaries, and the aggregated `jacoco.xml` you had us restore is untouched. Verified on the run: coverage still reports.

.circleci/ci/src/workflows/groups/changed-files.ts
phiz71: Keeping it here, then. The predicate is what makes the two new Gamma jobs correct rather than an optimisation on top of them, so pulling it out would leave this branch with jobs that nothing triggers properly. The description carries both directions and flags it as BX-292 ground taken early, so whoever picks that epic up will not have to rediscover it.

.circleci/ci/src/workflows/groups/backend-jobs.ts
```diff
@@ -152,7 +156,39 @@ export function backendJobs(
...
+      new workflow.WorkflowJob(testGammaJob, {
+        name: 'Test gamma',
+        context: config.jobContext,
+        requires: ['Build backend'],
+      }),
+      new workflow.WorkflowJob(testGammaUiJob, {
+        name: 'Test gamma UI',
+        context: config.jobContext,
+        requires: ['Build backend'],
+      }),
```
ytvnr: **Worth fixing** - The new gamma UI job is gated behind the whole Maven build although it consumes nothing from it, which works against the wall clock this branch is about.

`TestGammaUiJob` runs `checkout`, `attach_workspace`, yarn install and two `nx run-many` targets. Nothing it reads is produced by `Build backend`, whose workspace carries the rest-api classes, the console OpenAPI file and the distribution. The comparable front-end jobs in the same generated workflow, `job-nx-webui-lint-test` and `job-gamma-webui-build` in `pull-requests-custom-branch-gamma-console-only.yml`, declare no `requires` at all. On a gamma-only pull request this serialises a pure-JS job behind the full backend build, and the `attach_workspace: { at: '.' }` now also pulls the freshly persisted rest-api classes for nothing.

Suggested direction: drop the `requires` edge and the workspace attach on the UI job so it starts alongside the other front-end jobs. `Test gamma`, being Java, keeps both.

.circleci/ci/src/jobs/backend/job-build-backend.ts
```diff
@@ -73,6 +73,13 @@ export class BuildBackendJob {
         root: './',
         paths: [
           './gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/target/classes/console-openapi.*',
+          // The rest-api test job runs surefire against these instead of compiling and generating
+          // again. -DskipTests above compiles the tests without running them, so test-classes are
+          // already here. Two levels of glob because the modules nest one deep in places.
+          './gravitee-apim-rest-api/*/target/classes',
+          './gravitee-apim-rest-api/*/target/test-classes',
+          './gravitee-apim-rest-api/*/*/target/classes',
+          './gravitee-apim-rest-api/*/*/target/test-classes',
```
ytvnr: **Worth fixing** - The compiled rest-api classes go into the single workspace every downstream job attaches, so they are paid for by consumers that never needed them.

`Build backend` has one `persist_to_workspace`, and every job built on `AbstractTestJob`, plus the integration tests, the publish jobs and the nine parameterised Sonar analyses, does `attach_workspace: { at: '.' }`. Widening that payload already produced one regression, and the fix for it lives in `job-sonarcloud-analysis.ts:59-68`, that is, in a job shared by all nine analyses rather than at the source. It works, but every other consumer silently pays the extra download and any future consumer inherits the same trap.

Suggested direction: scope the classes to a workspace path, or a cache key, that only `Test rest-api` attaches, rather than the shared root workspace. That would also make the Sonar `rm` unnecessary.

ytvnr (commented):
Third pass, on the four commits pushed since the last one. `dcf0f8cc`, `62eaf50e`, `42afdb6c` and `99f3ed18` all read as closing what they set out to close: the gamma persist is gone and the Nx outputs now match where jest actually writes, the two accepted Kafka codes are documented for what they pin rather than narrowed, and the classes are stripped from the workspace before the scanner runs, which brings the analysis back to 319 s.

One thing I had missed on both earlier passes, and it is the one that matters: the nightly build no longer runs the kafka-explorer and gamma suites at all. Two remarks are inline; the two below sit outside the diff, so I could not anchor them.

### Notes outside the diff

#### `.circleci/ci/src/workflows/workflow-nightly.ts:66-73`

**⚠️ Important** - The nightly build no longer runs any kafka-explorer or gamma test, because it hard-codes its suite list and was never given the jobs those modules were split into.

`NightlyWorkflow` spells its backend suites out in `backendSuites` instead of calling `backendJobs()`, and this branch only adds `TestKafkaExplorerJob`, `TestGammaJob` and `TestGammaUiJob` inside `backend-jobs.ts`. Meanwhile the nightly `Test rest-api` goes from `mvn test -Drest-api-modules` on master to `-pl '!.,!gravitee-apim-rest-api/gravitee-apim-rest-api-kafka-explorer'`, and `rest-api-modules` no longer lists `gravitee-gamma` now that it has its own `gamma-modules` profile in `pom.xml`. The generated snapshot confirms both halves: `nightly.yml` still declares the same six test jobs as master, with no `Test kafka-explorer`, no `Test gamma` and no `Test gamma UI`. So the two suites that used to be covered every night by that single job now only run on pull requests that touch them.

Worth noting the nightly `Test rest-api` also lost `-Dskip.validation=true -Dgravitee.archrules.skip=true` in the move, which is a change in what it checks rather than a gap, but it is not called out anywhere either.

Suggested direction: add the three suites to the nightly backend list, kafka-explorer with its machine executor, or have the nightly derive its backend suites from the same source as the pull-request group so a future split cannot diverge again.

#### `gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java:81-88`

**Worth fixing** - The twin of the jsonpath load-order hack is still there, so the repo now both asserts and denies that 4.4.1 fixed the cache-provider ordering.

`c0041d75` deletes the static block from `UserServiceImpl` on the grounds that 4.4.1 makes it unnecessary, but the same block, same comment, same `Class.forName(JsonPathFunction.class.getName())`, survives verbatim in `OAuth2AuthenticationResource`. Either the ordering no longer needs forcing and this one is dead code, or it does and the half that went is the one that loaded earliest, since `UserServiceImpl` is instantiated at Spring context startup where the resource is only touched on the OAuth callback.

Suggested direction: decide it once on the evidence from the 4.4.1 change, and apply the same answer to both sites.

gravitee-apim-rest-api/gravitee-apim-rest-api-kafka-explorer/src/test/java/io/gravitee/rest/api/kafkaexplorer/resource/KafkaExplorerResourceErrorTest.java
```diff
@@ -0,0 +1,136 @@
...
+        return Stream.of(
+            Arguments.of(TechnicalCode.TIMEOUT, "Connection to Kafka cluster timed out"),
+            Arguments.of(TechnicalCode.CONNECTION_FAILED, "Failed to connect to Kafka cluster"),
+            Arguments.of(TechnicalCode.AUTHENTICATION_FAILED, "Authentication against Kafka cluster failed")
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("brokerFailures")
+    void should_return_502_carrying_the_technical_code(TechnicalCode technicalCode, String message) throws Exception {
```
ytvnr, line 67: Minor: the two tests are laid out in arrange/act/assert blocks but without the `// Given`, `// When`, `// Then` markers the project uses. Everything else the convention asks for is there (snake_case names, `@DisplayNameGeneration`, a real service instance in the second case), so this is the one gap, and the file is new here. The structure already matches the markers, so it is annotation rather than restructuring.

ytvnr (commented):
Fourth pass, on the rewritten history this time.

I went back over the eleven resolved threads and checked each agreed fix against the final tree. All of them survived the rewrite: the jacoco aggregate runs over the same `-pl` selection as the tests, the dead `user-service-test` execution is gone from the `service` pom, `InstanceOfAssertFactories.type` is statically imported and the `isIn` trade-off is documented, gamma reports through jest-junit with the nx outputs aligned and no orphan `persist_to_workspace`, `Test gamma UI` starts without waiting on `Build backend`, and the compiled classes travel in a workflow-scoped cache that only `Test rest-api` restores, with `Build backend`'s `persist_to_workspace` byte-identical to master.

The seven commits are self-consistent: the two that touch the generator regenerate the fixtures in the same commit. The "generated configuration unchanged" claim of `refactor(ci): declare every suite once` holds. For the ten fixtures it touches, the sorted line multiset is identical before and after and every diff line is paired, so it is pure block reordering. Across the 52 fixtures there is no duplicate job key, every `requires:` resolves to a `name:` in the same workflow, and no job is left orphaned.

Two things on CI rather than on the code:

- `Test gamma UI` was red on `bfcfd2d5`, which kept `Validate workflow status` from running. `c1ca5c25` sorts the platform route imports, so that one is answered: the new job lints gamma UI, which nothing did before this branch.
- `Lint & test APIM Libs` was red on the same commit and I could not attribute it. Nothing in the diff touches `gravitee-apim-webui-libs`. Worth a look before merge, even if it turns out to be a flake.

Nothing blocking on my side beyond those checks going green. One minor comment inline.

ytvnr (approved):

PR #19177, opened 2026-08-16
fix(audit): resolve api product audit names and raise user search limit (AIAM-207)

vikrantgravitee:
https://gravitee.atlassian.net/browse/AIAM-207
## Summary

Companion to gravitee-io/gravitee-gamma-module-aim#651.

- Resolve **API Product** names in audit property metadata (so AI Workspace audit targets show workspace names).

## Test plan

- [x] `AuditMetadataQueryServiceImplTest` — API_PRODUCT name resolution
- [ ] Open AI Workspace → Audit Logs: Target value shows workspace name, not raw id

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/audit/AuditMetadataQueryServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/audit/AuditMetadataQueryServiceImpl.java
@@ -21,6 +21,7 @@ import io.gravitee.apim.core.audit.model.AuditEntity;
 import io.gravitee.apim.core.audit.query_service.AuditMetadataQueryService;
 import io.gravitee.apim.infra.adapter.UserAdapter;
 import io.gravitee.repository.exceptions.TechnicalException;
+import io.gravitee.repository.management.api.ApiProductsRepository;
 import io.gravitee.repository.management.api.ApiRepository;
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/query_service/audit/AuditMetadataQueryServiceImpl.java
```diff
@@ -113,6 +118,8 @@ public String fetchPropertyMetadata(AuditEntity audit, String propertyName, Stri
                     return pageRepository.findById(propertyValue).map(Page::getName).orElse(propertyValue);
                 case PLAN:
                     return planRepository.findById(propertyValue).map(Plan::getName).orElse(propertyValue);
+                case API_PRODUCT:
+                    return apiProductQueryService.findById(propertyValue).map(ApiProduct::getName).orElse(propertyValue);
```
araffass: This one can take the whole audit page down instead of just showing the raw id like the other branches do. The try around the switch catches TechnicalException, the checked one the repositories throw, but findById here turns it into TechnicalDomainException, which extends RuntimeException. So it goes straight through the catch, up through buildMetadata, and the audit search fails.

To be fair it only shows up on a partial failure, since a full DB outage would break auditRepository.search first. Still a one liner: either hit apiProductRepository directly like the branches above, or catch the runtime exception for this case.

araffass (approved):
Approved but one comment worth diagnostic and possibly a fix

PR #19187, opened 2026-08-17
feat: support typo-tolerant API Product catalog search

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-127

## Description

Adds typo-tolerant matching for API Product names in the unified Portal catalog.

The implementation:

- uses the existing `portal.next.catalog.fuzzySearch.enabled` setting;
- preserves case-insensitive partial matching;
- applies the same token length, query length, prefix, and edit-distance limits as API fuzzy search;
- avoids fuzzy matching for unrelated products and when the setting is disabled;
- does not change the Portal API contract.

https://github.com/user-attachments/assets/8e8ae9df-5e45-4deb-9bcd-74acdbdceafc

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/GetVisiblePortalCatalogItemsUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/GetVisiblePortalCatalogItemsUseCase.java
@@ -39,6 +39,7 @@ import io.gravitee.apim.core.portal_page.model.PortalNavigationSearchInclude;
 import io.gravitee.apim.core.portal_page.query_service.PortalNavigationItemsQueryService;
 import io.gravitee.common.data.domain.Page;
 import io.gravitee.rest.api.model.common.Pageable;
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/GetVisiblePortalCatalogItemsUseCase.java
```diff
@@ -54,6 +55,9 @@
 @RequiredArgsConstructor
 public class GetVisiblePortalCatalogItemsUseCase {
 
+    private static final int MIN_FUZZY_TOKEN_LENGTH = 4;
```
thomasballivet, line 58: it would worth to have a separate type for these values to share it between here and `ApiDocumentSearcher`, wdyt?

thomasballivet (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/GetVisiblePortalCatalogItemsUseCase.java
jaroslawlakomy-grav: I would lean toward keeping these constants local for now. Although the values currently match those in `ApiDocumentSearcher`, they protect two different implementations: Lucene query expansion and in-memory API Product name matching. Extracting a shared type would introduce coupling and imply that both implementations must always evolve together. If the fuzzy matching behavior is centralized later, the limits can move together with that behavior.

PR #19193, opened 2026-08-17
feat(console): add fetch actions for sourced portal navigation items

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-90

## Description

Adds manual fetch actions for source-backed portal navigation items in the Console.

- **Fetch now** button in the "Content/Folder from … — read only" banner of the editor, for a sourced page or folder (requires `environment-documentation-u`).
- **Fetch All** entry in the tree context menu, on containers that have at least one sourced page in their subtree.
- Sourced nodes are now flagged with a link icon in the tree.
- Both actions call `POST /portal-navigation-items/{id}/_fetch` and report the outcome in a snackbar: single-page success, the source's fetch error, or a "X of Y pages" summary listing the failed titles (capped at 3).

## Additional context

https://github.com/user-attachments/assets/4b941f79-172f-4cd7-a83b-5f7d647b4987

### DIFF
--- a/gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
+++ b/gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
@@ -29,7 +29,7 @@ import { MatSlideToggleModule } from '@angular/material/slide-toggle';
 import { AbstractControl, FormControl, ReactiveFormsModule, ValidationErrors, ValidatorFn } from '@angular/forms';
 import { rxResource, takeUntilDestroyed, toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { ActivatedRoute, Router } from '@angular/router';
-import { catchError, exhaustMap, filter, map, shareReplay, skip, switchMap, take, tap } from 'rxjs/operators';
+import { catchError, exhaustMap, filter, finalize, map, shareReplay, skip, switchMap, take, tap } from 'rxjs/operators';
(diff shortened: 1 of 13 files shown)


gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
```diff
@@ -263,6 +266,18 @@ export class PortalNavigationItemsComponent implements HasUnsavedChanges {
     return navItem ? this.findSourcedFolderAbove(navItem) : null;
   });
   readonly isContentEditingLocked = computed(() => !!this.selectedItemSource() || !!this.parentSourcedFolder());
+  readonly isFetching = signal(false);
```
jaroslawlakomy-grav, line 269: Could we treat an in-progress fetch as a mutation lock for the navigation editor, rather than disabling only the other fetch actions? The backend loads each page before the external call and later persists the complete navigation item through crudService.update(page). Edit, publish, move, and source changes remain available while that request is running, so the later fetch save can silently overwrite one of those updates. Could we disable these mutations until the fetch and refresh have completed, or make the backend update concurrency-safe and limited to the fetch state?

thomasballivet: good catch, yes we can handle in the backend, will be done in a following PR

jaroslawlakomy-grav (approved):

PR #19199, opened 2026-08-17
chore(kafka): bump reactor to 7.1.0-alpha.7, use the {domain} placeholder in routing host mode examples

ThibaudAV:
Related to [APIM-14788](https://gravitee.atlassian.net/browse/APIM-14788). Companion of gravitee-io/gravitee-reactor-native-kafka#375 (released in reactor `7.0.2`).

## Context

The Kafka routing host mode patterns are evaluated **once per configured domain**, with `{domain}` substituted by the domain being evaluated (`{defaultDomain}` is kept as an alias). The example configuration shipped by APIM predates that and contradicts it:

- **`helm/values.yaml`** is the pressing one: it already advertises a two-entry list
  ```yaml
  domains:
    - mycompany.org
    - mycompany.com
  ```
  while keeping `bootstrapDomainPattern: {apiHost}.mycompany.org` / `brokerDomainPattern: broker-{brokerId}-{apiHost}.mycompany.org`. Copying it verbatim makes both domains resolve to `…mycompany.org`, leaves `mycompany.com` unroutable, and now trips the collapse warning the reactor emits — from the official example.
- **`gravitee.yml`** (gateway distribution) never documented the `domains` list at all, only the deprecated scalar `defaultDomain`.

## Changes

- Both patterns in both files now use the placeholder: `{apiHost}.{domain}` and `broker-{brokerId}-{apiHost}.{domain}`, in the default example and in the "override to fit your DNS configuration" example.
- `{domain}` documented alongside `{apiHost}` / `{brokerId}`: per-domain evaluation, the collapse that hardcoding a domain causes, and `{defaultDomain}` as a deprecated alias.
- `gravitee.yml` gains the `domains:` list and the `defaultDomain` deprecation note that `helm/values.yaml` already carried.
- Corrected an inaccuracy carried by both files: the override note claimed `defaultDomain`, `brokerPrefix` and `domainSeparator` "are not used" with custom patterns. `DefaultKafkaAcceptor#evalDomainPattern` substitutes all of them whenever the pattern references them.
- `helm/CHANGELOG.md`: extended the existing 4.13.0 `domains` entry.

## Reactor bump — this PR is not comment-only

`gravitee-apim-distribution/pom.xml:241` moves from `7.1.0-alpha.6` to `7.1.0-alpha.7`. **This changes the shipped runtime**, and it is required: `alpha.6` was cut before the `alpha` branch was synced with `main`, so it has no `{domain}` substitution. On it, the examples this PR ships would leave `{domain}` literal, `buildPattern` would call `Pattern.compile("myapi\\.{domain}")` and throw `PatternSyntaxException: Illegal repetition` — the acceptor is never built and the API does not deploy.

`alpha.7` brings the `{domain}` alias, the collapse warning, the `defaultDomain`-is-ignored warning, and the three NetClient leak fixes from `7.0.6`.

The quick-setup stacks are updated in the same move: `alpha.7` warns on every startup for the deprecated `defaultDomain`, which `docker/quick-setup/native-kafka` and `native-kafka-mesh` still set. They now use `domains_0`.

**Not identical to #19200**, contrary to what this description said earlier: that PR bumped 4.12.x to `7.0.6`, this one bumps `master` to `7.1.0-alpha.7`, and it carries review corrections that #19200 was merged without.

## The documentation part is comment-only

The `kafka:` block is commented out in its entirety in both config files, so that half changes no rendered configuration and no manifest output. `{defaultDomain}` keeps working, so no existing configuration breaks.

## Follow-up

**4.12.x**: same change in #19200. Filed as a separate PR rather than a backport, because the distribution module was restructured on `master` and `gravitee.yml` has moved (`gravitee-apim-distribution/…-standalone-gateway/` here vs `gravitee-apim-gateway/…-standalone-distribution/` there). Content is identical.

**4.11.x**: not planned. It carries the same misleading `helm/values.yaml` example, but ships reactor `6.3.4`, whose `evalDomainPattern` has no `{domain}` alias — the fix there would have to be written with `{defaultDomain}` (already substituted per domain on that branch). Recorded here so the gap is known.

[APIM-14788]: https://gravitee.atlassian.net/browse/APIM-14788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-distribution/pom.xml
+++ b/gravitee-apim-distribution/pom.xml
@@ -238,7 +238,7 @@
     <gravitee-resource-schema-registry-embedded.version>1.0.0</gravitee-resource-schema-registry-embedded.version>
     <gravitee-resource-storage-azure-blob.version>1.1.0</gravitee-resource-storage-azure-blob.version>
     <gravitee-reactor-message.version>11.0.0</gravitee-reactor-message.version>
-    <gravitee-reactor-native-kafka.version>7.1.0-alpha.6</gravitee-reactor-native-kafka.version>
+    <gravitee-reactor-native-kafka.version>7.1.0-alpha.7</gravitee-reactor-native-kafka.version>
     <gravitee-reactor-mcp-proxy.version>3.2.0</gravitee-reactor-mcp-proxy.version>
     <gravitee-reactor-llm-proxy.version>4.1.0</gravitee-reactor-llm-proxy.version>
     <gravitee-reactor-a2a-proxy.version>2.1.0</gravitee-reactor-a2a-proxy.version>
--- a/helm/CHANGELOG.md
+++ b/helm/CHANGELOG.md
@@ -5,7 +5,7 @@ This file documents all notable changes to [Gravitee.io API Management 3.x](http
 
 ### 4.13.0
 - Render `config/hazelcast.xml` for gateway Hazelcast clustering (`gateway.cluster.type=hazelcast`) with Kubernetes discovery via the `-hz` Service. Cluster name defaults to `<release>-<sharding_tags>` so Redis distributed-event namespace (clusterId) changes when sharding tags change; override with `gateway.cluster.hazelcast.clusterName`. Require `gateway.cluster` when `gateway.distributedSync` is enabled; auto-enable `services.sync.repository` and `services.sync.distributed`.
-- Support multiple custom domains for native Kafka APIs with `gateway.kafka.routingHostMode.domains` (list). `defaultDomain` is deprecated and ignored when `domains` is set.
+- Support multiple custom domains for native Kafka APIs with `gateway.kafka.routingHostMode.domains` (list). `defaultDomain` is deprecated and ignored when `domains` is set. The `bootstrapDomainPattern` / `brokerDomainPattern` examples now use the `{domain}` placeholder, which is substituted with each configured domain in turn; hardcoding a domain there collapses every entry onto the same hostname (`{defaultDomain}` remains a supported alias).
 - Document `gateway.services.metrics.kafka.durations.principalName` to add the `principal_name` tag to Kafka duration metrics (disabled by default).
 - fix gateway requestTimeout ignored when gateway.servers is configured (APIM-14276)
 
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -1454,37 +1454,54 @@ gateway:
 #    # Routing Host Mode
 #    routingHostMode:
 #      brokerPrefix: broker-          # default is broker-
-#      domainSeparator: -             # Used to separate broker's name from api & defaultDomain. Default is '-'
+#      domainSeparator: "-"           # Used to separate broker's name from api & domain. Default is '-'
 #
 #      # The domains where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org` and `myapi.mycompany.com`
 #      # Each domain should be set according to a public wildcard DNS/Certificate.
+#      # Must be a list, even for a single domain: a scalar value is rejected at startup.
 #      domains:
 #        - mycompany.org
 #        - mycompany.com
 #
-#      # The default domain where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org`
-#      # Deprecated: use `domains` instead. Ignored when `domains` is set.
-#      defaultDomain: mycompany.org   # Should set according to the public wildcard DNS/Certificate. Default is empty
+#      # For a single domain, `defaultDomain` is still accepted but deprecated. It is ignored -- with a
+#      # warning -- as soon as `domains` is set, so enable one or the other, never both:
+#      #   defaultDomain: mycompany.org
+#
 #      defaultPort:   9092            # Default public port for Kafka APIs. Default is 9092
 #
-#      # With the upper default configuration, the Gravitee Kafka gateway yields bootstrap and broker domains to be as follows:
-#      bootstrapDomainPattern: {apiHost}.mycompany.org
-#      brokerDomainPattern: broker-{brokerId}-{apiHost}.mycompany.org
+#      # The patterns below are the built-in behaviour, written out explicitly. Each is evaluated once per configured domain:
+#      bootstrapDomainPattern: "{apiHost}.{domain}"
+#      brokerDomainPattern: "{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{domain}"
+#      # {domain} requires a gateway bundling native Kafka reactor 7.0.2 or later (APIM 4.12.14+). On an
+#      # earlier gateway it is left unsubstituted and the API fails to deploy with a PatternSyntaxException;
+#      # use {defaultDomain} there, which resolves to the same value.
 #      # Where:
-#      # {apiHost}  is a placeholder that will be replaced when the API is deployed, by the API Host Prefix.
-#      # {brokerId} is a placeholder that stands for the broker id
+#      # {apiHost}         is a placeholder that will be replaced when the API is deployed, by the API Host Prefix.
+#      # {brokerId}        is a placeholder that stands for the broker id
+#      # {brokerPrefix}    is a placeholder for the 'brokerPrefix' property above
+#      # {domainSeparator} is a placeholder for the 'domainSeparator' property above
+#      # {domain}          is a placeholder for the domain being evaluated: the patterns are applied once per
+#      #                   entry of `domains`, or once to `defaultDomain` when `domains` is not set.
+#      #                   An entry that itself contains {apiHost} becomes the bootstrap pattern, and both
+#      #                   patterns configured here are then ignored.
+#      #                   Writing a domain literally here instead of using {domain} makes every entry resolve
+#      #                   to that same hostname, leaving the other domains unroutable. The gateway logs a
+#      #                   warning when it detects it. `{defaultDomain}` is still accepted as a deprecated
+#      #                   alias of `{domain}`.
+#      # The built-in defaults, used for whichever pattern is left unset, are "{apiHost}.{defaultDomain}" and
+#      # "{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{defaultDomain}".
 #
 #      # It can be overridden to fit your DNS configuration.
-#      # Doing so requires BOTH patterns to be set, as well as 'defaultPort'. Please note that 'defaultDomain', 'brokerPrefix' and 'domainSeparator' are not used in that case, hence optional.
+#      # Each pattern falls back to its own default independently, so overriding only one of them is fine, and 'defaultPort' defaults to 9092. Please note that 'brokerPrefix' and 'domainSeparator' are substituted only if the patterns reference them as {brokerPrefix} / {domainSeparator}, hence optional.
 #      # Example:
 #      #   defaultPort: 9092
-#      #   bootstrapDomainPattern: bootstrap-{apiHost}.mycompany.org
-#      #   brokerDomainPattern: {apiHost}-broker{brokerId}.mycompany.org
+#      #   bootstrapDomainPattern: "bootstrap-{apiHost}.{domain}"
+#      #   brokerDomainPattern: "{apiHost}-broker{brokerId}.{domain}"
 #      #
-#      #   This configuration yields domains that must target the Gravitee Kafka gateway:
-#      #      bootstrap-myapi.mycompany.org
-#      #      myapi-broker0.mycompany.org
-#      #      myapi-broker1.mycompany.org
+#      #   With the two domains above, this configuration yields domains that must target the Gravitee Kafka gateway:
+#      #      bootstrap-myapi.mycompany.org      bootstrap-myapi.mycompany.com
+#      #      myapi-broker0.mycompany.org        myapi-broker0.mycompany.com
+#      #      myapi-broker1.mycompany.org        myapi-broker1.mycompany.com
 #      #      ...
 #
 #    # Kafka probe
--- a/docker/quick-setup/native-kafka-mesh/docker-compose.yml
+++ b/docker/quick-setup/native-kafka-mesh/docker-compose.yml
@@ -117,7 +117,7 @@ services:
             - gravitee_kafka_ssl_keystore_password=gravitee
             - gravitee_kafka_ssl_keystore_path=$${gravitee.home}/ssl/server.keystore.jks
             - gravitee_kafka_ssl_keystore_type=jks
-            - 'gravitee_kafka_routingHostMode_defaultDomain=kafka.local'
+            - 'gravitee_kafka_routingHostMode_domains_0=kafka.local'
         networks:
             storage:
             frontend:
--- a/docker/quick-setup/native-kafka/README.md
+++ b/docker/quick-setup/native-kafka/README.md
@@ -44,7 +44,7 @@ The gateway runs in HOST routing mode (the default) and resolves API hostnames v
 1. Console: `http://localhost:8084/` (`admin` / `admin`).
 2. Organization → Entrypoints & Sharding Tags → Entrypoint Configuration → **Default Kafka Domain** = `kafka.local`.
 
-The gateway is already configured with `gravitee_kafka_routingHostMode_defaultDomain=kafka.local`; the console value above mirrors that to the management plane.
+The gateway is already configured with `gravitee_kafka_routingHostMode_domains_0=kafka.local`; the console value above mirrors that to the management plane.
 
 ## 2. Create a Native Kafka API
 
--- a/docker/quick-setup/native-kafka/docker-compose.yml
+++ b/docker/quick-setup/native-kafka/docker-compose.yml
@@ -116,7 +116,7 @@ services:
             - gravitee_kafka_ssl_keystore_password=gravitee
             - gravitee_kafka_ssl_keystore_path=$${gravitee.home}/ssl/server.keystore.jks
             - gravitee_kafka_ssl_keystore_type=jks
-            - 'gravitee_kafka_routingHostMode_defaultDomain=kafka.local'
+            - 'gravitee_kafka_routingHostMode_domains_0=kafka.local'
         networks:
             storage:
             frontend:
--- a/gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-gateway/src/main/resources/config/gravitee.yml
+++ b/gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-gateway/src/main/resources/config/gravitee.yml
@@ -185,30 +185,54 @@ secrets:
 #  # Routing Host Mode
 #  routingHostMode:
 #    brokerPrefix: broker-          # default is broker-
-#    domainSeparator: -             # Used to separate broker's name from api & defaultDomain. Default is '-'
+#    domainSeparator: "-"           # Used to separate broker's name from api & domain. Default is '-'
 #
-#    # The default domain where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org`
-#    defaultDomain: mycompany.org   # Should set according to the public wildcard DNS/Certificate. Default is empty
+#    # The domains where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org` and `myapi.mycompany.com`
+#    # Each domain should be set according to a public wildcard DNS/Certificate.
+#    # Must be a list, even for a single domain: a scalar value is rejected at startup.
+#    domains:
+#      - mycompany.org
+#      - mycompany.com
+#
+#    # For a single domain, `defaultDomain` is still accepted but deprecated. It is ignored -- with a
+#    # warning -- as soon as `domains` is set, so enable one or the other, never both:
+#    #   defaultDomain: mycompany.org
+#
 #    defaultPort:   9092            # Default public port for Kafka APIs. Default is 9092
 #
-#    # With the upper default configuration, the Gravitee Kafka gateway yields bootstrap and broker domains to be as follows:
-#    bootstrapDomainPattern: {apiHost}.mycompany.org
-#    brokerDomainPattern: broker-{brokerId}-{apiHost}.mycompany.org
+#    # The patterns below are the built-in behaviour, written out explicitly. Each is evaluated once per configured domain:
+#    bootstrapDomainPattern: "{apiHost}.{domain}"
+#    brokerDomainPattern: "{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{domain}"
+#    # {domain} requires a gateway bundling native Kafka reactor 7.0.2 or later (APIM 4.12.14+). On an
+#    # earlier gateway it is left unsubstituted and the API fails to deploy with a PatternSyntaxException;
+#    # use {defaultDomain} there, which resolves to the same value.
 #    # Where:
-#    # {apiHost}  is a placeholder that will be replaced when the API is deployed, by the API Host Prefix.
-#    # {brokerId} is a placeholder that stands for the broker id
+#    # {apiHost}         is a placeholder that will be replaced when the API is deployed, by the API Host Prefix.
+#    # {brokerId}        is a placeholder that stands for the broker id
+#    # {brokerPrefix}    is a placeholder for the 'brokerPrefix' property above
+#    # {domainSeparator} is a placeholder for the 'domainSeparator' property above
+#    # {domain}          is a placeholder for the domain being evaluated: the patterns are applied once per
+#    #                   entry of `domains`, or once to `defaultDomain` when `domains` is not set.
+#    #                   An entry that itself contains {apiHost} becomes the bootstrap pattern, and both
+#    #                   patterns configured here are then ignored.
+#    #                   Writing a domain literally here instead of using {domain} makes every entry resolve
+#    #                   to that same hostname, leaving the other domains unroutable. The gateway logs a
+#    #                   warning when it detects it. `{defaultDomain}` is still accepted as a deprecated
+#    #                   alias of `{domain}`.
+#    # The built-in defaults, used for whichever pattern is left unset, are "{apiHost}.{defaultDomain}" and
+#    # "{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{defaultDomain}".
 #
 #    # It can be overridden to fit your DNS configuration.
-#    # Doing so requires BOTH patterns to be set, as well as 'defaultPort'. Please note that 'defaultDomain', 'brokerPrefix' and 'domainSeparator' are not used in that case, hence optional.
+#    # Each pattern falls back to its own default independently, so overriding only one of them is fine, and 'defaultPort' defaults to 9092. Please note that 'brokerPrefix' and 'domainSeparator' are substituted only if the patterns reference them as {brokerPrefix} / {domainSeparator}, hence optional.
 #    # Example:
 #    #   defaultPort: 9092
-#    #   bootstrapDomainPattern: bootstrap-{apiHost}.mycompany.org
-#    #   brokerDomainPattern: {apiHost}-broker{brokerId}.mycompany.org
+#    #   bootstrapDomainPattern: "bootstrap-{apiHost}.{domain}"
+#    #   brokerDomainPattern: "{apiHost}-broker{brokerId}.{domain}"
 #    #
-#    #   This configuration yields domains that must target the Gravitee Kafka gateway:
-#    #      bootstrap-myapi.mycompany.org
-#    #      myapi-broker0.mycompany.org
-#    #      myapi-broker1.mycompany.org
+#    #   With the two domains above, this configuration yields domains that must target the Gravitee Kafka gateway:
+#    #      bootstrap-myapi.mycompany.org      bootstrap-myapi.mycompany.com
+#    #      myapi-broker0.mycompany.org        myapi-broker0.mycompany.com
+#    #      myapi-broker1.mycompany.org        myapi-broker1.mycompany.com
 #    #      ...
 #
 #  # Kafka probe
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -37,6 +37,8 @@ annotations:
           url: https://github.com/gravitee-io/gravitee-api-management/pull/17691
     - kind: deprecated
       description: "`gateway.kafka.routingHostMode.defaultDomain` is deprecated in favor of `domains` and ignored when `domains` is set"
+    - kind: changed
+      description: "`gateway.kafka.routingHostMode` pattern examples now use the `{domain}` placeholder, substituted with each configured domain in turn; hardcoding a domain there collapses every entry onto the same hostname (`{defaultDomain}` remains a supported alias)"
     - kind: added
       description: "Document `gateway.services.metrics.kafka.durations.principalName` to add the principal_name tag to Kafka duration metrics (disabled by default)"
     - kind: added


ThibaudAV:
Review feedback from #19200 applied here as well (identical content, so the same defects were present):

- `bootstrapDomainPattern` / `brokerDomainPattern` quoted — a scalar starting with `{` parses as a flow mapping, so the block could not be uncommented as-is
- `domainSeparator: -` → `domainSeparator: "-"` — `-` after a colon starts a block sequence; this is actually the first error hit
- `defaultDomain` removed from the active example — with `domains` set alongside it, the reactor warns that it is ignored, which is the very class of problem this PR removes
- `{brokerPrefix}` / `{domainSeparator}` added to the `# Where:` list, since the corrected note refers to them

Verified by extracting the `kafka:` block, stripping one layer of `#` and parsing the result: both files now yield a valid mapping. Thanks @phiz71.

ThibaudAV:
⚠️ **Blocker found while writing the documentation companion — this PR should not be merged as-is.**

`master` pins `<gravitee-reactor-native-kafka.version>7.1.0-alpha.6</...>` (root `pom.xml:241`). That tag does **not** have the `{domain}` placeholder. Its `evalDomainPattern` is:

```java
return pattern
    .replace("{brokerPrefix}", brokerPrefix)
    .replace("{domainSeparator}", domainSeparator)
    .replace("{apiHost}", apiHost)
    .replace("{defaultDomain}", defaultDomain)
    .replaceAll("\\.$", "");
```

No `.replace("{domain}", …)`. So on a build of `master` today, the example this PR ships — `bootstrapDomainPattern: "{apiHost}.{domain}"` — leaves `{domain}` **unsubstituted**, producing the literal hostname `myapi.{domain}`. That is worse than the hardcoded-domain example it replaces.

The alias, the collapse warning, the `defaultDomain`-is-ignored warning and the per-acceptor hostname logs all landed in reactor `7.0.2`. Verified at the tags:

| tag | `{domain}` alias | collapse warning |
| --- | --- | --- |
| `7.0.1` | no | no |
| `7.0.4` | yes | yes |
| `7.0.6` | yes | yes |
| `7.1.0-alpha.6` | **no** | **no** |

`7.1.0-alpha.6` is the newest alpha and is behind `main`, which does have all of it. The alpha line appears to be cut from the `alpha` branch rather than `main`, so it also lacks the three resource-leak fixes released in `7.0.6`.

**Not an issue on #19200**: 4.12.x now pins `7.0.6`, which has the alias.

Two ways out, and the choice is not mine:

1. **Cut a new `7.1.0-alpha` from `main`** and bump `master` to it. The documentation in this PR then becomes correct as written, and `master` also picks up the leak fixes it is currently missing. This is my recommendation.
2. **Rewrite this PR with `{defaultDomain}`**, which works on every version. Correct today, but it means `master` documents the deprecated spelling while 4.12.x documents the canonical one.

Holding this PR in draft until that is settled.

ThibaudAV:
**Update — half of the blocker is cleared, the release is not.**

The `alpha` branch was synced with `main` this morning (`79019be9fb`, *Merge branch 'main' into alpha*, 10:11 UTC), and it now carries everything that was missing. Verified at ref `alpha`:

- `{domain}` alias in `DefaultKafkaAcceptor#evalDomainPattern` — present
- collapse warning in `KafkaApiReactor#hostModeAcceptors` — present
- the `7.0.6` NetClient leak fixes — present (`e57c5c27` is an ancestor of `alpha`)

So my earlier remark that "the alpha line appears to be cut from the `alpha` branch rather than `main`" no longer describes the current state — that gap has been closed at the source.

What is still missing is the **published artifact**. There is no `7.1.0-alpha.7`:

```
git tag  refs/tags/7.1.0-alpha.7  -> 404 Not Found
gh release view 7.1.0-alpha.7     -> release not found
```

The newest prerelease is still `7.1.0-alpha.6` (2026-08-10), which is the version this branch pins and which does **not** have the alias. So the defect described above still applies to a build of `master` today, and I cannot bump `pom.xml:241` to a version that does not resolve.

Looks like the release job did not run, or did not complete, after the merge into `alpha` — worth a look at the pipeline.

As soon as `7.1.0-alpha.7` is published, the fix here is a one-line bump of `pom.xml:241` and this blocker is void; the documentation in this PR is already written for that version.

ThibaudAV:
**Blocker cleared.** `7.1.0-alpha.7` was published at 12:18 UTC and this branch now pins it (`gravitee-apim-distribution/pom.xml:241`).

Verified at the tag itself, not on the `alpha` branch head:

```java
// DefaultKafkaAcceptor#evalDomainPattern @ 7.1.0-alpha.7
.replace("{apiHost}", apiHost)
.replace("{domain}", defaultDomain)
.replace("{defaultDomain}", defaultDomain)
```

- `{domain}` alias — present, so the examples this PR ships resolve correctly
- collapse warning in `KafkaApiReactor#hostModeAcceptors` — present
- `defaultDomain`-is-ignored warning in `RoutingHostModeDomains#read` — present
- the three `7.0.6` NetClient leak fixes — included (`e57c5c27` is an ancestor of the tag)

Nothing left from my earlier objection: the documentation in this PR is now accurate for the reactor `master` actually builds against.

One correction to that earlier comment: I wrote the pin was in the **root** `pom.xml:241`. It is in `gravitee-apim-distribution/pom.xml:241` — the distribution reactor, not the product one. The line number was right, the file was not. It is also the only place the version appears on this branch.

Bump kept as its own commit, mirroring #19200.

phiz71 (commented):
Documentation review — the substance checks out: I verified `{domain}`, the `{defaultDomain}` alias, the `{brokerPrefix}` / `{domainSeparator}` substitutions, both quoted built-in defaults, the collapse warning and the `defaultDomain`-ignored warning against `DefaultKafkaAcceptor` / `RoutingHostModeDomains` / `KafkaApiReactor` at tag `7.1.0-alpha.7`. Quoting the patterns is a real fix too — `bootstrapDomainPattern: {apiHost}.mycompany.org` unquoted is invalid YAML.

Comments below are on the examples that still carry a hardcoded value, on a few statements that describe a stricter contract than the code implements, and on the dependency bump versus the "documentation only" framing. Where a point applies identically to `gravitee.yml`, the twin line is named inline rather than duplicated.

gravitee-apim-distribution/pom.xml
```diff
@@ -238,7 +238,7 @@
     <gravitee-resource-schema-registry-embedded.version>1.0.0</gravitee-resource-schema-registry-embedded.version>
     <gravitee-resource-storage-azure-blob.version>1.1.0</gravitee-resource-storage-azure-blob.version>
     <gravitee-reactor-message.version>11.0.0</gravitee-reactor-message.version>
-    <gravitee-reactor-native-kafka.version>7.1.0-alpha.6</gravitee-reactor-native-kafka.version>
+    <gravitee-reactor-native-kafka.version>7.1.0-alpha.7</gravitee-reactor-native-kafka.version>
```
phiz71, line 241: The PR description says "No behaviour change … documentation only — no rendered configuration and no manifest output changes", but this bump is a runtime change: per fc45066, alpha.7 brings the collapse warning, the `defaultDomain`-ignored warning and the three NetClient leak fixes from 7.0.6.

The bump itself looks right — without it `{domain}` would be left unsubstituted in the very examples this PR ships. It is the description that is stale, and it also still claims "Content is identical" with #19200, which merged with a bump to 7.0.6 rather than 7.1.0-alpha.7. Worth updating so a release manager does not wave this through as comment-only.

helm/values.yaml
```diff
@@ -1454,37 +1454,48 @@ gateway:
...
+#      # For a single domain, `defaultDomain` is still accepted but deprecated. It is ignored -- with a
+#      # warning -- as soon as `domains` is set, so enable one or the other, never both:
+#      #   defaultDomain: mycompany.org
+#
 #      defaultPort:   9092            # Default public port for Kafka APIs. Default is 9092
 #
 #      # With the upper default configuration, the Gravitee Kafka gateway yields bootstrap and broker domains to be as follows:
-#      bootstrapDomainPattern: {apiHost}.mycompany.org
-#      brokerDomainPattern: broker-{brokerId}-{apiHost}.mycompany.org
+#      bootstrapDomainPattern: "{apiHost}.{domain}"
```
phiz71, line 1473: `{domain}` is new in reactor 7.0.2, but nothing here (nor the CHANGELOG entry) states the required gateway version — and the failure mode is hard.

On an earlier reactor `evalDomainPattern` does not substitute it, so `buildPattern` ends up calling `Pattern.compile("myapi\\.{domain}")`, which throws `PatternSyntaxException: Illegal repetition` — the acceptor is never built and the API does not deploy. The chart lets `image.tag` float independently of the chart version, and the PR body already notes 4.11.x ships 6.3.4 without the alias.

A one-liner would make the copy-paste safe, e.g. `# {domain} requires gateway 4.12+ (native Kafka reactor 7.0.2+); use {defaultDomain} on earlier versions.`

helm/values.yaml
```diff
@@ -1454,37 +1454,48 @@ gateway:
...
+#      brokerDomainPattern: "broker-{brokerId}-{apiHost}.{domain}"
```
phiz71: This example still hardcodes `broker-` and `-`, i.e. the values of `brokerPrefix` and `domainSeparator` set 16 lines above — the same hardcoding trap this PR fixes for the domain.

Set `brokerPrefix: kafka-` and the explicit pattern silently ignores it. The contradiction is sharper now that lines 1485-1486 document the built-in default as `{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{defaultDomain}`: the shipped example no longer illustrates the default it sits under.

`brokerDomainPattern: "{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{domain}"` would be both accurate and self-consistent. Same line in gravitee.yml:204.

helm/values.yaml
```diff
@@ -1454,37 +1454,48 @@ gateway:
...
+#      #                   Writing a domain literally here instead of using {domain} makes every entry resolve
+#      #                   to that same hostname, leaving the other domains unroutable. The gateway logs a
+#      #                   warning when it detects it. `{defaultDomain}` is still accepted as a deprecated
+#      #                   alias of `{domain}`.
+#      # The built-in defaults, used when both patterns are left unset, are "{apiHost}.{defaultDomain}" and
+#      # "{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{defaultDomain}".
 #
 #      # It can be overridden to fit your DNS configuration.
-#      # Doing so requires BOTH patterns to be set, as well as 'defaultPort'. Please note that 'defaultDomain', 'brokerPrefix' and 'domainSeparator' are not used in that case, hence optional.
+#      # Doing so requires BOTH patterns to be set, as well as 'defaultPort'. Please note that 'brokerPrefix' and 'domainSeparator' are substituted only if the patterns reference them as {brokerPrefix} / {domainSeparator}, hence optional.
```
phiz71: The first half of this sentence is as inaccurate as the half being fixed: nothing requires both patterns, nor `defaultPort`.

`DefaultKafkaAcceptor` reads each pattern in its own `getProperty(..., DEFAULT_*_PATTERN)` call, and `DefaultBrokerResolver:29` defaults `defaultPort` to 9092. Overriding only `bootstrapDomainPattern` works fine — the broker pattern keeps its default, which stays per-domain correct since `{defaultDomain}` resolves to the domain being evaluated. Same line in gravitee.yml:220.

helm/values.yaml
```diff
@@ -1454,37 +1454,48 @@ gateway:
...
+#      # {brokerId}        is a placeholder that stands for the broker id
+#      # {brokerPrefix}    is a placeholder for the 'brokerPrefix' property above
+#      # {domainSeparator} is a placeholder for the 'domainSeparator' property above
+#      # {domain}          is a placeholder for the domain being evaluated: the patterns are applied once per
+#      #                   entry of `domains`, or once to `defaultDomain` when `domains` is not set.
+#      #                   Writing a domain literally here instead of using {domain} makes every entry resolve
+#      #                   to that same hostname, leaving the other domains unroutable. The gateway logs a
+#      #                   warning when it detects it. `{defaultDomain}` is still accepted as a deprecated
+#      #                   alias of `{domain}`.
+#      # The built-in defaults, used when both patterns are left unset, are "{apiHost}.{defaultDomain}" and
```
phiz71: "used when both patterns are left unset" implies an all-or-nothing fallback that does not exist — each pattern falls back to its own constant independently.

Someone overriding only `bootstrapDomainPattern` reads this and concludes the broker pattern is now undefined, when it silently keeps `{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{defaultDomain}`. "used for whichever pattern is left unset" would state the actual behaviour. Same line in gravitee.yml:216.

helm/values.yaml
```diff
@@ -1454,37 +1454,48 @@ gateway:
...
+#      bootstrapDomainPattern: "{apiHost}.{domain}"
+#      brokerDomainPattern: "broker-{brokerId}-{apiHost}.{domain}"
 #      # Where:
-#      # {apiHost}  is a placeholder that will be replaced when the API is deployed, by the API Host Prefix.
-#      # {brokerId} is a placeholder that stands for the broker id
+#      # {apiHost}         is a placeholder that will be replaced when the API is deployed, by the API Host Prefix.
+#      # {brokerId}        is a placeholder that stands for the broker id
+#      # {brokerPrefix}    is a placeholder for the 'brokerPrefix' property above
+#      # {domainSeparator} is a placeholder for the 'domainSeparator' property above
+#      # {domain}          is a placeholder for the domain being evaluated: the patterns are applied once per
```
phiz71, line 1483: "the patterns are applied once per entry of `domains`" holds only for plain suffixes.

`DefaultKafkaAcceptor.createForDomain` branches on `domain.contains("{apiHost}")`: when an entry carries `{apiHost}`, that entry *becomes* the bootstrap pattern, the broker pattern is forced to `{brokerPrefix}{brokerId}{domainSeparator}` + entry, and `hostDefaultDomain` is set to `""` so any `{domain}` inside it collapses to the empty string — the configured `bootstrapDomainPattern` / `brokerDomainPattern` are ignored entirely.

An edge case, but the sentence is written without qualification.

helm/values.yaml
```diff
@@ -1454,37 +1454,48 @@ gateway:
 #    # Routing Host Mode
 #    routingHostMode:
 #      brokerPrefix: broker-          # default is broker-
-#      domainSeparator: -             # Used to separate broker's name from api & defaultDomain. Default is '-'
+#      domainSeparator: "-"           # Used to separate broker's name from api & domain. Default is '-'
 #
 #      # The domains where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org` and `myapi.mycompany.com`
 #      # Each domain should be set according to a public wildcard DNS/Certificate.
 #      domains:
 #        - mycompany.org
```
phiz71, line 1463: Worth stating that `domains` is list-only: `RoutingHostModeDomains.read` throws `IllegalArgumentException("… it must be a list …")` on a scalar value, and symmetrically rejects `defaultDomain[0]`.

The block being replaced trained operators on the scalar form, so `domains: mycompany.org` is the natural first attempt on migration — and it takes the gateway down at startup rather than falling back.

helm/values.yaml
```diff
@@ -1454,37 +1454,48 @@ gateway:
...
-#      # The default domain where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org`
-#      # Deprecated: use `domains` instead. Ignored when `domains` is set.
-#      defaultDomain: mycompany.org   # Should set according to the public wildcard DNS/Certificate. Default is empty
+#      # For a single domain, `defaultDomain` is still accepted but deprecated. It is ignored -- with a
+#      # warning -- as soon as `domains` is set, so enable one or the other, never both:
+#      #   defaultDomain: mycompany.org
+#
 #      defaultPort:   9092            # Default public port for Kafka APIs. Default is 9092
 #
 #      # With the upper default configuration, the Gravitee Kafka gateway yields bootstrap and broker domains to be as follows:
```
phiz71: This lead-in no longer describes what follows it. It promises the domains the gateway yields; the two lines below are now pure templates, and the only worked example of resolved hostnames sits 18 lines further down in the override block.

Something like "the patterns below are the built-in behaviour" would match what is shown. Same line in gravitee.yml:202.

helm/CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ This file documents all notable changes to [Gravitee.io API Management 3.x](http
 
 ### 4.13.0
 - Render `config/hazelcast.xml` for gateway Hazelcast clustering (`gateway.cluster.type=hazelcast`) with Kubernetes discovery via the `-hz` Service. Cluster name defaults to `<release>-<sharding_tags>` so Redis distributed-event namespace (clusterId) changes when sharding tags change; override with `gateway.cluster.hazelcast.clusterName`. Require `gateway.cluster` when `gateway.distributedSync` is enabled; auto-enable `services.sync.repository` and `services.sync.distributed`.
-- Support multiple custom domains for native Kafka APIs with `gateway.kafka.routingHostMode.domains` (list). `defaultDomain` is deprecated and ignored when `domains` is set.
+- Support multiple custom domains for native Kafka APIs with `gateway.kafka.routingHostMode.domains` (list). `defaultDomain` is deprecated and ignored when `domains` is set. The `bootstrapDomainPattern` / `brokerDomainPattern` examples now use the `{domain}` placeholder, which is substituted with each configured domain in turn; hardcoding a domain there collapses every entry onto the same hostname (`{defaultDomain}` remains a supported alias).
```
phiz71, line 8: `helm/Chart.yaml:34` carries the mirror of this entry in `artifacthub.io/changes` ("Support multiple custom domains for native Kafka APIs with `gateway.kafka.routingHostMode.domains` (list)"), and that file's header reads "# Also update CHANGELOG.md".

It has not been extended, so readers of the release notes on artifacthub.io — the URL that annotation block links to — get no mention of `{domain}` or of the collapse it prevents.

helm/values.yaml
```diff
@@ -1454,37 +1454,48 @@ gateway:
...
 #      # The domains where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org` and `myapi.mycompany.com`
 #      # Each domain should be set according to a public wildcard DNS/Certificate.
 #      domains:
 #        - mycompany.org
 #        - mycompany.com
 #
-#      # The default domain where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org`
-#      # Deprecated: use `domains` instead. Ignored when `domains` is set.
-#      defaultDomain: mycompany.org   # Should set according to the public wildcard DNS/Certificate. Default is empty
+#      # For a single domain, `defaultDomain` is still accepted but deprecated. It is ignored -- with a
```
phiz71, line 1466: While the deprecation is being documented here: the shipped quick-setups still configure the deprecated property — `docker/quick-setup/native-kafka/docker-compose.yml:119`, `docker/quick-setup/native-kafka-mesh/docker-compose.yml:120` and `docker/quick-setup/native-kafka/README.md:47` (plus `MetricsIntegrationTest.java:263`).

The alpha.7 bump in this PR is what makes it visible: `RoutingHostModeDomains` now logs "'kafka.routingHostMode.defaultDomain' is deprecated … configure 'kafka.routingHostMode.domains' (a list) instead" at every startup, so the official demo stacks will emit a deprecation warning while contradicting the documentation this PR writes.

helm/values.yaml
```diff
@@ -1454,37 +1454,48 @@ gateway:
 #    # Routing Host Mode
 #    routingHostMode:
 #      brokerPrefix: broker-          # default is broker-
-#      domainSeparator: -             # Used to separate broker's name from api & defaultDomain. Default is '-'
+#      domainSeparator: "-"           # Used to separate broker's name from api & domain. Default is '-'
```
phiz71, line 1457: Out of diff, but same theme: `helm/tests/gateway/configmap_kafka_test.yaml:69-70` is the repo's only executable example of these patterns, and it combines `brokerPrefix: "fox-"` / `domainSeparator: "|"` with `brokerDomainPattern: "broker-{brokerId}-{apiHost}.mycompany.org"` — both properties ignored, domain hardcoded.

Nothing fails there (the test asserts template passthrough), but a contributor grepping for a usage example lands on it first. `"{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{domain}"` would exercise the same passthrough while modelling the documented form.

gravitee-apim-distribution/pom.xml
ThibaudAV: Applied — the description was written before the bump existed and never updated.

It now says plainly that the bump changes the shipped runtime and why it is not optional: on `alpha.6` the examples this PR ships would leave `{domain}` literal, and `buildPattern` would blow up. I checked that rather than assuming, in actual Java:

```
"myapi.{domain}".replace("{brokerId}","([0-9]+)").replace(".","\\.")
  -> Pattern.compile("myapi\\.{domain}")
  -> PatternSyntaxException: Illegal repetition near index 8
```

So it is a startup failure, not a routing subtlety — worth being explicit about in a PR a release manager reads.

The "content is identical" claim is gone too: #19200 bumped 4.12.x to `7.0.6`, this bumps `master` to `7.1.0-alpha.7`, and this PR now also carries your corrections, which #19200 merged without.

helm/values.yaml
ThibaudAV: Applied, and thank you for tracing the failure mode — I had assumed a degraded hostname, not a hard failure. Reproduced it exactly as you describe (`Illegal repetition near index 8`), which is what convinced me the note is worth the two lines rather than being version trivia:

```
#    # {domain} requires a gateway bundling native Kafka reactor 7.0.2 or later (APIM 4.12.14+). On an
#    # earlier gateway it is left unsubstituted and the API fails to deploy with a PatternSyntaxException;
#    # use {defaultDomain} there, which resolves to the same value.
```

The `image.tag` floating point is the one that settles it — the chart version is no guarantee of the gateway version.

helm/values.yaml
ThibaudAV: Applied: `brokerDomainPattern: "{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{domain}"`.

You are right that it is the same trap, and I walked into it while fixing the domain half. The example now matches, character for character, the built-in default documented eleven lines below — which is what it was supposed to illustrate all along.

helm/values.yaml
ThibaudAV: Applied. Both halves were wrong, and I only fixed the half I had noticed.

Confirmed before rewriting: `DefaultBrokerResolver:29` is `getProperty("kafka.routingHostMode.defaultPort", Integer.class, 9092)`, and each pattern has its own `getProperty(..., DEFAULT_*_PATTERN)` fallback. New wording:

> Each pattern falls back to its own default independently, so overriding only one of them is fine, and 'defaultPort' defaults to 9092.

That sentence had survived from the original file and I copied it forward while correcting the clause next to it.

helm/values.yaml
ThibaudAV: Applied — "used for whichever pattern is left unset".

Same root cause as the comment above: independent fallbacks described as a single unit. The reading you describe is the natural one, and it is wrong in the direction that matters — someone overriding only the bootstrap pattern would go looking for a broker pattern that is quietly still there.

helm/values.yaml
ThibaudAV: Applied as a qualification rather than a rewrite:

```
#    #                   An entry that itself contains {apiHost} becomes the bootstrap pattern, and both
#    #                   patterns configured here are then ignored.
```

Verified in `createForDomain` at `7.1.0-alpha.7`: the `domain.contains("{apiHost}")` branch sets `hostDefaultDomain = ""`, forces the broker pattern to `DEFAULT_BROKER_DOMAIN_PATTERN_ACCESS_POINT + domain`, and never reads the configured patterns. Edge case, but the sentence was unqualified and this is exactly the shape a Cloud access-point config takes.

helm/values.yaml
ThibaudAV: Applied:

```
#    # Must be a list, even for a single domain: a scalar value is rejected at startup.
```

Your reasoning about the migration path is what makes this the most valuable of the batch: the block being replaced showed `defaultDomain: mycompany.org`, a scalar, so `domains: mycompany.org` is the obvious first attempt — and `RoutingHostModeDomains.read` throws `IllegalArgumentException` rather than falling back, which means the gateway does not come up.

helm/values.yaml
ThibaudAV: Applied — the lead-in now reads:

> The patterns below are the built-in behaviour, written out explicitly. Each is evaluated once per configured domain:

You are right that it was stale rather than merely vague: it promised resolved domains, and my change turned the two lines under it into templates without touching the sentence introducing them.

helm/CHANGELOG.md
ThibaudAV: Applied. `helm/Chart.yaml` now carries a `kind: changed` entry mirroring the CHANGELOG line, right after the existing `deprecated` one.

The file header does say "Also update CHANGELOG.md" and I did the reverse — extended the CHANGELOG and left the annotation block alone. Since that block is what feeds the artifacthub.io release notes, the readers most likely to hit the collapse were the ones not being told about it.

Note this is `master`-only: on 4.12.x the block is empty, the CI clears it after each release.

helm/values.yaml
ThibaudAV: Applied to the three shipped stacks — `docker/quick-setup/native-kafka/docker-compose.yml`, `native-kafka-mesh/docker-compose.yml` and the README line that documents the variable. All three now set `gravitee_kafka_routingHostMode_domains_0=kafka.local`, matching the indexed convention already used in the repo (`gravitee_reporters_elasticsearch_endpoints_0`).

Your framing is what made me include it: the deprecation warning is not pre-existing noise here, it is introduced by the bump in this PR, so the demo stacks would start contradicting the documentation being written next to them.

**`MetricsIntegrationTest.java:263` I left alone, deliberately.** `.set("kafka.routingHostMode.defaultDomain", "localhost")` would become `domains[0]`, and I cannot run that suite here to confirm the test configuration API handles the bracket form. Changing a test I cannot execute to silence a log line is a bad trade — happy to do it if you know it works, or to leave it as a follow-up.

helm/values.yaml
ThibaudAV: Not taken, deliberately — the only one of the batch I am leaving.

You are right on the substance: `configmap_kafka_test.yaml:69-70` combines `brokerPrefix: "fox-"` / `domainSeparator: "|"` with a pattern that ignores both and hardcodes the domain, and it is the first hit for a contributor grepping for a usage example. But it is out of diff, the assertions are passthrough so the change is cosmetic there, and this PR has already grown from a comment-only doc change into a runtime bump plus quick-setup migration.

Worth its own small PR — and that one could cover 4.12.x at the same time, which this one deliberately does not.

gravitee-apim-distribution/pom.xml
phiz71: Thanks — the description now says "This changes the shipped runtime" and explains why the bump is required, which is what I was after.

One residual contradiction though: the Follow-up section still ends with "Content is identical" about #19200, eight lines below the new "**Not identical to #19200**, contrary to what this description said earlier". Two smaller leftovers in the same place: the first *Changes* bullet still advertises `broker-{brokerId}-{apiHost}.{domain}`, which the fixup replaced with `{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{domain}`, and the CHANGELOG bullet doesn't mention the new `Chart.yaml` entry.

Leaving this thread open for those.

helm/values.yaml
phiz71: Checked the version claim rather than take it on trust, and it holds: `{domain}` first appears at reactor tag `7.0.2`, `4.12.x` went `7.0.1` → `7.0.3` directly (so 7.0.2 itself never shipped in APIM), and the earliest `4.12.x` tag containing that bump is indeed `4.12.14`.

The fallback advice is right too — `RoutingHostModeDomains` exists as far back as `6.3.4`, so on an earlier gateway `{defaultDomain}` really does resolve per evaluated domain.

helm/values.yaml
phiz71: `{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{domain}` now matches the built-in default documented below it, and the new lead-in says as much. Thanks.

helm/values.yaml
phiz71: "Each pattern falls back to its own default independently, so overriding only one of them is fine, and 'defaultPort' defaults to 9092" — matches the code. Thanks.

helm/values.yaml
phiz71: "used for whichever pattern is left unset" — that's it. Thanks.

helm/values.yaml
phiz71: The `{apiHost}`-in-an-entry caveat is spelled out now. Thanks.

helm/values.yaml
phiz71: "Must be a list, even for a single domain: a scalar value is rejected at startup" — exactly the gap. Thanks.

helm/values.yaml
phiz71: "The patterns below are the built-in behaviour, written out explicitly" now matches what follows it. Thanks.

helm/CHANGELOG.md
phiz71: The `kind: changed` entry is in, and sits inside the changes section as the file's header requires. Thanks.

helm/values.yaml
phiz71: Since this is the one functional change in the fixup, I checked the binding rather than assume it: `RelaxedPropertySource.checkPropertyName` takes the requested `domains[0]`, falls through `encodedArray` to `domains.0`, then to the env key `domains_0` — so `gravitee_kafka_routingHostMode_domains_0` does land on `kafka.routingHostMode.domains[0]`, consistent with the existing `gravitee_plugins_path_0` usage in the repo. No `defaultDomain` residue left under `docker/` either.

helm/values.yaml
phiz71: Still open — the fixture is unchanged, and so is `MetricsIntegrationTest.java:263`, which still sets the deprecated `defaultDomain`. Nothing fails in either case, so leaving it for a follow-up is entirely fair; flagging it so it reads as a decision rather than an oversight.

phiz71 (approved):
Approving — the documentation now matches the code.

I checked the substantive claims against the reactor rather than taking them on trust: `{domain}` and the collapse warning both land in `7.0.2`; `4.12.x` went `7.0.1` → `7.0.3`, so `4.12.14` really is the first release carrying them; and `gravitee_kafka_routingHostMode_domains_0` does bind to `domains[0]` via `RelaxedPropertySource` (`domains[0]` → `domains.0` → `domains_0`), consistent with the existing `gravitee_plugins_path_0`. That last one was the only executable line in the change, and the one place a wrong form would have silently broken quick-setup routing.

Two non-blocking things worth doing before it goes in, both about the PR rather than the diff:

- The `docs(kafka):` title doesn't advertise the runtime bump. On a squash merge the title becomes the commit subject, so nobody later looking for why the native Kafka reactor moved will find it there. #19200 named both halves — either do the same here, or rebase-merge to keep the `docs` and `chore` commits distinct, in which case the `fixup!` needs a `git rebase -i --autosquash` first since GitHub won't autosquash it.
- The description still contradicts itself: "Content is identical" survives in *Follow-up*, eight lines under the new "**Not identical to #19200**". On a squash merge that text becomes the commit body.

Left open and fine as follow-ups: `helm/tests/gateway/configmap_kafka_test.yaml:69-70` still pairs `brokerPrefix: "fox-"` / `domainSeparator: "|"` with a pattern hardcoding `broker-`/`-`, and `MetricsIntegrationTest.java:263` still sets the deprecated `defaultDomain`.

PR #19200, opened 2026-08-17
[4.12.x] chore(kafka): bump reactor to 7.0.6, use the {domain} placeholder in routing host mode examples

ThibaudAV:
Related to [APIM-14788](https://gravitee.atlassian.net/browse/APIM-14788). Companion of gravitee-io/gravitee-reactor-native-kafka#375 (released in reactor `7.0.2`; 4.12.x ships `7.0.5`).

4.12.x counterpart of #19199 — filed separately rather than backported, because the distribution module was restructured on `master` and `gravitee.yml` has moved (`gravitee-apim-distribution/…-standalone-gateway/` vs `gravitee-apim-gateway/…-standalone-distribution/`). Content is identical.

## Context

The Kafka routing host mode patterns are evaluated **once per configured domain**, with `{domain}` substituted by the domain being evaluated (`{defaultDomain}` is kept as an alias). The example configuration shipped by APIM predates that and contradicts it:

- **`helm/values.yaml`** is the pressing one: it already advertises a two-entry list
  ```yaml
  domains:
    - mycompany.org
    - mycompany.com
  ```
  while keeping `bootstrapDomainPattern: {apiHost}.mycompany.org` / `brokerDomainPattern: broker-{brokerId}-{apiHost}.mycompany.org`. Copying it verbatim makes both domains resolve to `…mycompany.org`, leaves `mycompany.com` unroutable, and now trips the collapse warning the reactor emits — from the official example.
- **`gravitee.yml`** (gateway distribution) never documented the `domains` list at all, only the deprecated scalar `defaultDomain`.

## Changes

- Both patterns in both files now use the placeholder: `{apiHost}.{domain}` and `broker-{brokerId}-{apiHost}.{domain}`, in the default example and in the "override to fit your DNS configuration" example.
- `{domain}` documented alongside `{apiHost}` / `{brokerId}`: per-domain evaluation, the collapse that hardcoding a domain causes, and `{defaultDomain}` as a deprecated alias.
- `gravitee.yml` gains the `domains:` list and the `defaultDomain` deprecation note that `helm/values.yaml` already carried.
- Corrected an inaccuracy carried by both files: the override note claimed `defaultDomain`, `brokerPrefix` and `domainSeparator` "are not used" with custom patterns. `DefaultKafkaAcceptor#evalDomainPattern` substitutes all of them whenever the pattern references them.
- `helm/CHANGELOG.md`: extended the existing 4.12.0 `domains` entry.

## Also in this PR: reactor bump 7.0.5 → 7.0.6

`pom.xml` — the single place the version is pinned on this branch. [7.0.6](https://github.com/gravitee-io/gravitee-reactor-native-kafka/releases/tag/7.0.6) carries three resource-leak fixes, none of them related to routing or domains:

- `net`: release runtime NetClients from the Vert.x close future
- `probe`: share a single NetClient across TCP health checks
- `virtualcluster`: close connected clusters when a sibling connect fails

Kept as its own commit. Unlike the doc change below, this one does alter the shipped runtime.

## No behaviour change (documentation part)

The `kafka:` block is commented out in its entirety in both files, so this is documentation only — no rendered configuration and no manifest output changes. `{defaultDomain}` keeps working, so no existing configuration breaks either.

## Note on 4.11.x

Not planned. 4.11.x carries the same misleading `helm/values.yaml` example, but ships reactor `6.3.4`, whose `evalDomainPattern` has no `{domain}` alias — the fix there would have to be written with `{defaultDomain}` (already substituted per domain on that branch). Recorded here so the gap is known.

[APIM-14788]: https://gravitee.atlassian.net/browse/APIM-14788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
@@ -185,30 +185,48 @@ secrets:
 #  # Routing Host Mode
 #  routingHostMode:
 #    brokerPrefix: broker-          # default is broker-
-#    domainSeparator: -             # Used to separate broker's name from api & defaultDomain. Default is '-'
+#    domainSeparator: "-"           # Used to separate broker's name from api & domain. Default is '-'
 #
-#    # The default domain where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org`
-#    defaultDomain: mycompany.org   # Should set according to the public wildcard DNS/Certificate. Default is empty
+#    # The domains where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org` and `myapi.mycompany.com`
+#    # Each domain should be set according to a public wildcard DNS/Certificate.
+#    domains:
+#      - mycompany.org
+#      - mycompany.com
+#
+#    # For a single domain, `defaultDomain` is still accepted but deprecated. It is ignored -- with a
+#    # warning -- as soon as `domains` is set, so enable one or the other, never both:
+#    #   defaultDomain: mycompany.org
+#
 #    defaultPort:   9092            # Default public port for Kafka APIs. Default is 9092
 #
 #    # With the upper default configuration, the Gravitee Kafka gateway yields bootstrap and broker domains to be as follows:
-#    bootstrapDomainPattern: {apiHost}.mycompany.org
-#    brokerDomainPattern: broker-{brokerId}-{apiHost}.mycompany.org
+#    bootstrapDomainPattern: "{apiHost}.{domain}"
+#    brokerDomainPattern: "broker-{brokerId}-{apiHost}.{domain}"
 #    # Where:
-#    # {apiHost}  is a placeholder that will be replaced when the API is deployed, by the API Host Prefix.
-#    # {brokerId} is a placeholder that stands for the broker id
+#    # {apiHost}         is a placeholder that will be replaced when the API is deployed, by the API Host Prefix.
+#    # {brokerId}        is a placeholder that stands for the broker id
+#    # {brokerPrefix}    is a placeholder for the 'brokerPrefix' property above
+#    # {domainSeparator} is a placeholder for the 'domainSeparator' property above
+#    # {domain}          is a placeholder for the domain being evaluated: the patterns are applied once per
+#    #                   entry of `domains`, or once to `defaultDomain` when `domains` is not set.
+#    #                   Writing a domain literally here instead of using {domain} makes every entry resolve
+#    #                   to that same hostname, leaving the other domains unroutable. The gateway logs a
+#    #                   warning when it detects it. `{defaultDomain}` is still accepted as a deprecated
+#    #                   alias of `{domain}`.
+#    # The built-in defaults, used when both patterns are left unset, are "{apiHost}.{defaultDomain}" and
+#    # "{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{defaultDomain}".
 #
 #    # It can be overridden to fit your DNS configuration.
-#    # Doing so requires BOTH patterns to be set, as well as 'defaultPort'. Please note that 'defaultDomain', 'brokerPrefix' and 'domainSeparator' are not used in that case, hence optional.
+#    # Doing so requires BOTH patterns to be set, as well as 'defaultPort'. Please note that 'brokerPrefix' and 'domainSeparator' are substituted only if the patterns reference them as {brokerPrefix} / {domainSeparator}, hence optional.
 #    # Example:
 #    #   defaultPort: 9092
-#    #   bootstrapDomainPattern: bootstrap-{apiHost}.mycompany.org
-#    #   brokerDomainPattern: {apiHost}-broker{brokerId}.mycompany.org
+#    #   bootstrapDomainPattern: "bootstrap-{apiHost}.{domain}"
+#    #   brokerDomainPattern: "{apiHost}-broker{brokerId}.{domain}"
 #    #
-#    #   This configuration yields domains that must target the Gravitee Kafka gateway:
-#    #      bootstrap-myapi.mycompany.org
-#    #      myapi-broker0.mycompany.org
-#    #      myapi-broker1.mycompany.org
+#    #   With the two domains above, this configuration yields domains that must target the Gravitee Kafka gateway:
+#    #      bootstrap-myapi.mycompany.org      bootstrap-myapi.mycompany.com
+#    #      myapi-broker0.mycompany.org        myapi-broker0.mycompany.com
+#    #      myapi-broker1.mycompany.org        myapi-broker1.mycompany.com
 #    #      ...
 #
 #  # Kafka probe
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -1454,7 +1454,7 @@ gateway:
 #    # Routing Host Mode
 #    routingHostMode:
 #      brokerPrefix: broker-          # default is broker-
-#      domainSeparator: -             # Used to separate broker's name from api & defaultDomain. Default is '-'
+#      domainSeparator: "-"           # Used to separate broker's name from api & domain. Default is '-'
 #
 #      # The domains where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org` and `myapi.mycompany.com`
 #      # Each domain should be set according to a public wildcard DNS/Certificate.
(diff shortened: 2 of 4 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -185,30 +185,42 @@ secrets:
...
 #
 #    # The default domain where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org`
+#    # Deprecated: use `domains` instead. Ignored when `domains` is set.
 #    defaultDomain: mycompany.org   # Should set according to the public wildcard DNS/Certificate. Default is empty
 #    defaultPort:   9092            # Default public port for Kafka APIs. Default is 9092
 #
 #    # With the upper default configuration, the Gravitee Kafka gateway yields bootstrap and broker domains to be as follows:
-#    bootstrapDomainPattern: {apiHost}.mycompany.org
-#    brokerDomainPattern: broker-{brokerId}-{apiHost}.mycompany.org
+#    bootstrapDomainPattern: {apiHost}.{domain}
```
phiz71: This is not valid YAML: a scalar starting with `{` is parsed as a flow mapping, so uncommenting the block fails with `did not find expected key while parsing a block mapping` and the gateway does not start.

```suggestion
#    bootstrapDomainPattern: "{apiHost}.{domain}"
```

`brokerDomainPattern` on the next line is fine as-is (it starts with `broker-`), as is `bootstrap-{apiHost}.{domain}` in the override example further down.

The defect is pre-existing (`{apiHost}.mycompany.org` had it too), but this PR rewrites precisely this line to make it copy-pasteable, and that is the one use it does not survive.

helm/values.yaml
```diff
@@ -1468,23 +1468,28 @@ gateway:
 #      defaultPort:   9092            # Default public port for Kafka APIs. Default is 9092
 #
 #      # With the upper default configuration, the Gravitee Kafka gateway yields bootstrap and broker domains to be as follows:
-#      bootstrapDomainPattern: {apiHost}.mycompany.org
-#      brokerDomainPattern: broker-{brokerId}-{apiHost}.mycompany.org
+#      bootstrapDomainPattern: {apiHost}.{domain}
```
phiz71: Same as in `gravitee.yml`: the value starts with `{` and is parsed as a flow mapping, so a user who uncomments this in their values file gets `error converting YAML to JSON` before any template renders.

```suggestion
#      bootstrapDomainPattern: "{apiHost}.{domain}"
```

The configmap template already quotes the pattern (`helm/templates/gateway/gateway-configmap.yaml:227`), so only the values file needs the quotes.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -185,30 +185,42 @@ secrets:
 #  # Routing Host Mode
 #  routingHostMode:
 #    brokerPrefix: broker-          # default is broker-
-#    domainSeparator: -             # Used to separate broker's name from api & defaultDomain. Default is '-'
+#    domainSeparator: -             # Used to separate broker's name from api & domain. Default is '-'
+#
+#    # The domains where the Kafka APIs are exposed. ex: `myapi` will be exposed as `myapi.mycompany.org` and `myapi.mycompany.com`
+#    # Each domain should be set according to a public wildcard DNS/Certificate.
+#    domains:
```
phiz71, line 192: The example now shows `domains:` (two entries) and `defaultDomain: mycompany.org` as active lines side by side. `RoutingHostModeDomains#read` warns whenever both are non-empty:

> `'kafka.routingHostMode.defaultDomain' ('mycompany.org') is ignored because 'kafka.routingHostMode.domains' is configured; remove it to silence this warning.`

So the example, copied verbatim, logs a warning on every acceptor build -- the same class of problem this PR sets out to remove (an official example that trips a reactor warning). `helm/values.yaml` already had this shape; the PR newly introduces it here.

Keeping the deprecation note but leaving `defaultDomain:` out of the active example (or showing it only in the deprecated-alternative comment) would avoid it.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -185,30 +185,42 @@ secrets:
 #  # Routing Host Mode
 #  routingHostMode:
 #    brokerPrefix: broker-          # default is broker-
-#    domainSeparator: -             # Used to separate broker's name from api & defaultDomain. Default is '-'
+#    domainSeparator: -             # Used to separate broker's name from api & domain. Default is '-'
```
phiz71: Also not valid YAML: `-` after the colon starts a block sequence, so this is actually the *first* error hit when the block is uncommented (`block sequence entries are not allowed in this context`), before the pattern lines are even reached.

```suggestion
#    domainSeparator: "-"           # Used to separate broker's name from api & domain. Default is '-'
```

Same at `helm/values.yaml:1457`. Pre-existing too, but the line is touched here and the point of the change is a block a user can lift as-is.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -185,30 +185,42 @@ secrets:
...
 #    # {brokerId} is a placeholder that stands for the broker id
+#    # {domain}   is a placeholder for the domain being evaluated: the patterns are applied once per entry of
+#    #            `domains`, or once to `defaultDomain` when `domains` is not set.
+#    #            Writing a domain literally here instead of using {domain} makes every entry resolve to that
+#    #            same hostname, leaving the other domains unroutable. The gateway logs a warning when it
+#    #            detects it. `{defaultDomain}` is still accepted as a deprecated alias of `{domain}`.
 #
 #    # It can be overridden to fit your DNS configuration.
-#    # Doing so requires BOTH patterns to be set, as well as 'defaultPort'. Please note that 'defaultDomain', 'brokerPrefix' and 'domainSeparator' are not used in that case, hence optional.
+#    # Doing so requires BOTH patterns to be set, as well as 'defaultPort'. Please note that 'brokerPrefix' and 'domainSeparator' are substituted only if the patterns reference them, hence optional.
```
phiz71: The corrected note says `brokerPrefix` and `domainSeparator` are substituted "only if the patterns reference them", but the `# Where:` list above only documents `{apiHost}`, `{brokerId}` and `{domain}` -- the reader is never told the spelling is `{brokerPrefix}` / `{domainSeparator}`.

The old wording ("are not used") did not dangle this way. Worth adding the two placeholders to the `# Where:` list, since they are what the shipped defaults actually use (`{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{defaultDomain}`). Same at `helm/values.yaml:1483`.

phiz71 (commented):
Checked the doc claims against the reactor at tag `7.0.6`: `evalDomainPattern` does substitute `{domain}` with the domain being evaluated and keeps `{defaultDomain}` as an alias, `RoutingHostModeDomains` implements the `domains` > `defaultDomain` precedence, and the collapse warning is in `KafkaApiReactor#hostModeAcceptors`. The correction of the "are not used" note is justified. The 7.0.5 -> 7.0.6 range only carries the three leak fixes announced (plus a CI orb bump), and the root `pom.xml` is indeed the only place the version is pinned on this branch.

Comments below are about the example blocks themselves: as they stand, they still cannot be uncommented as-is, which is the very use this PR optimises for.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
ThibaudAV: Confirmed and applied. Reproduced with Psych before touching anything:

```
bootstrapDomainPattern: {apiHost}.{domain}   -> Psych::SyntaxError: did not find expected key while parsing a block mapping
bootstrapDomainPattern: "{apiHost}.{domain}" -> OK
```

I went one step further than the suggestion and quoted `brokerDomainPattern` too, plus both patterns in the override example. You are right that they parse fine as they stand (`broker-…` and `bootstrap-…` are plain scalars — I checked), but the reader's first edit to these lines is reordering the placeholders, and `{brokerId}-…` at the front puts them straight back into the failure. Uniform quoting makes the rule "patterns are quoted" instead of "patterns are quoted when they happen to start with a brace".

helm/values.yaml
ThibaudAV: Applied, same quoting. Thanks for pinning the failure to the right layer — I checked `gateway-configmap.yaml` and the template quotes every field it renders (`brokerPrefix`, `domainSeparator`, the `domains` entries, `defaultDomain` and both patterns all go through `| quote`), so the rendered configmap was never at risk and only the values file needed it.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
ThibaudAV: Good catch, and it is the sharper version of the bug this PR is about: the previous state was an example that produces a broken *configuration*, this would have been an example that produces a warning about itself.

`defaultDomain` is no longer an active line. It now appears only inside a comment, with the mutual exclusion stated:

```
#    # For a single domain, `defaultDomain` is still accepted but deprecated. It is ignored -- with a
#    # warning -- as soon as `domains` is set, so enable one or the other, never both:
#    #   defaultDomain: mycompany.org
```

I applied the same to `helm/values.yaml`, which carried the shape already. You scoped your comment to `gravitee.yml` since that is where the PR introduces it, but leaving helm as-is would keep the two files out of sync and keep one official example tripping the warning.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
ThibaudAV: Confirmed, applied, and you are right about the ordering — `block sequence entries are not allowed in this context` at `domainSeparator` fires before the parser ever reaches the pattern lines. Fixed in both files.

I verified the end state rather than the individual lines: extracting the `kafka:` block, stripping one layer of `#`, and parsing the result now yields

```
{"brokerPrefix"=>"broker-", "domainSeparator"=>"-", "domains"=>["mycompany.org", "mycompany.com"],
 "defaultPort"=>9092, "bootstrapDomainPattern"=>"{apiHost}.{domain}",
 "brokerDomainPattern"=>"broker-{brokerId}-{apiHost}.{domain}"}
```

for `gravitee.yml` and for `helm/values.yaml`, on both this branch and #19199. `brokerPrefix: broker-` parses fine unquoted so I left it alone.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
ThibaudAV: Fair — the rewrite made the note self-referential without giving the reader the vocabulary. Both placeholders are now in the `# Where:` list, and the note names the spelling explicitly ("substituted only if the patterns reference them as `{brokerPrefix}` / `{domainSeparator}`").

I also added the built-in defaults verbatim, since that is what actually demonstrates the two placeholders in use:

```
#    # The built-in defaults, used when both patterns are left unset, are "{apiHost}.{defaultDomain}" and
#    # "{brokerPrefix}{brokerId}{domainSeparator}{apiHost}.{defaultDomain}".
```

Quoted with `{defaultDomain}` rather than `{domain}` because that is literally what `DEFAULT_BOOTSTRAP_DOMAIN_PATTERN` / `DEFAULT_BROKER_DOMAIN_PATTERN` hold in the reactor; the alias is documented two lines above.

phiz71 (approved):
Thanks -- all five points are addressed, and I checked the result rather than the intent: uncommenting either block by one level now parses cleanly and yields the expected mapping (`domainSeparator => "-"`, `bootstrapDomainPattern => "{apiHost}.{domain}"`, `brokerDomainPattern => "broker-{brokerId}-{apiHost}.{domain}"`, `domains => ["mycompany.org", "mycompany.com"]`). `defaultDomain` being demoted to a nested comment also removes the `defaultDomain ... is ignored` warning a verbatim copy used to trigger. Resolving the threads.

The rest holds up: the `kafka:` block stays fully commented, so no rendered configuration and no manifest output changes; and 7.0.5 -> 7.0.6 is a clean patch (three resource-leak fixes, nothing touching routing or domains, and the `gravitee v5.13.1` entry in the range is the reactor's own CircleCI orb, not a Java dependency).

Two non-blocking nits, take them or leave them:

- "The built-in defaults, used when **both** patterns are left unset" suggests a coupling that does not exist -- `bootstrapDomainPattern` and `brokerDomainPattern` each fall back to their own default independently, so setting one does not stop the other from defaulting. Same looseness as the pre-existing "requires BOTH patterns to be set" just below.
- Those defaults are spelled with `{defaultDomain}`, the deprecated form. It is faithful to the code and the alias is documented three lines above, but "stop using `{defaultDomain}`" sitting next to "the defaults are `{defaultDomain}`" may give a reader pause.

One housekeeping item: the branch carries a `fixup!` commit -- worth autosquashing if the merge does not squash.

PR #19204, opened 2026-08-18
feat(gateway): Adds infrastructure for deploy-time policy warmup

pragsty:
## Issue
https://gravitee.atlassian.net/browse/APIM-14959

## Description
Adds infrastructure for deploy-time policy warmup in the V4 reactive Gateway, enabling policies like XML Validation to pre-fetch and compile schemas before the API is marked ready.
### Key Features
- **WarmablePolicy Integration**: Policies implementing `WarmablePolicy` are automatically warmed up during API deployment
- **Fail-Fast Deployment**: If warmup fails (e.g., registry unreachable), API deployment fails immediately
- **Non-Blocking Warmup**: Uses `subscribeOn(Schedulers.io())` to avoid blocking the event loop
- **Thread-Safe Context**: `ThreadLocal<DeploymentContext>` ensures warmup happens on deployment thread
### Changes
- `HttpPolicyFactory`: Added `beginWarmup()`/`endWarmup()` lifecycle methods and `warmupIfNeeded()` logic
- `FlowChain`: Added `warmupPolicies()` method to pre-create policy chains during deployment
- `DefaultApiReactor`: Calls `warmupPolicies()` after resources are started
- New `PolicyWarmupException` for fail-fast error reporting
### Implementation Details
- Warmup timeout: 30 seconds (configurable in future)
- ThreadLocal context avoids API contract changes
- try-finally ensures ThreadLocal cleanup


pragsty:
Not needed anymore as the approach changed

PR #19207, opened 2026-08-18
fix(gamma): add environment switching handling in `ShellLayout` and `RemoteModuleRoute` with updated navigation logic

Okhelifi:
## Issue

https://gravitee.atlassian.net/browse/FOUND-162

## Problem

Switching environment kept the whole path. From an API detail page, `/environments/env-1/apim/apis/{apiId}` became `/environments/env-2/apim/apis/{apiId}`, with an `apiId` that does not exist in `env-2`.

The Gateway Endpoint repainted immediately (it is keyed on the environment id), then the API detail query refetched and 404'd. Two symptoms, one cause.

## Changes

All host side. No module repository is touched, and graphene is unchanged.

* `shared/config/routes.ts`: on environment change, keep only the first segment after the environment (the module id, or a host area such as `home`/`tasks`). Modules are registered per organization, so every environment exposes the same ones and that root always resolves. This is what makes the fix generic across `apim`, `aim`, `authz`,
  `edge`, `esm`, and any module added later, without those repos participating.
* `features/environment/EnvironmentGuard.tsx`: withhold children until the environment in the URL and the one in the store agree. Children resolve ids from the URL but fetch with the store's environment, so any render where the two disagree issues a request for an entity of one environment against another. The store is synced from a passive effect, so the browser paints that state before it heals, and back/forward plus deep links reach it the same way. While withholding, the guard renders the same `ContentSkeleton` the shell uses as its Suspense fallback for that area, rather than blanking it.
* `features/modules/components/RemoteModuleRoute.tsx`: key the lazy remote on the environment's technical id so the federated subtree remounts, dropping any environment scoped state a module holds. The technical id rather than the URL segment, so canonicalizing an id to its hrid does not remount. The bundle stays cached, so this
  re-renders the remote without re-fetching it.
* `shared/components/ShellLayout.tsx`: the environment selector now only navigates. `EnvironmentGuard` owns adopting the new environment.

## Behaviour change

`/environments/{a}/apim/apis/{apiId}/endpoints` now goes to `/environments/{b}/apim`, not the equivalent deep path. Query string and hash are dropped with the sub-path. In apim nothing is lost by that: query params are only used there as transient deep links into a form on API detail pages (`?editGroup`/`?step`, which the page clears once consumed, and `?tab` on the alert form), all of them under the sub-path being discarded anyway. Back
still returns to the source environment's page.

`EnvironmentGuard`'s id to hrid canonicalization and its unknown-environment redirect are left alone: both rewrite the URL at constant environment, and both still preserve query string and hash.

## Why the module root, and not one level up

Considered: from an API detail page, land on the API list rather than the module home.
Rejected for now, because it cannot be done correctly from the host alone.

The host only sees opaque segments. Keeping "module id plus one segment" works for flat sections but breaks on nested ones: `/aim/catalog/models/{id}` would give `/aim/catalog`, which is neither the intended target (`/aim/catalog/models` is) nor a page at all, that route being a permission guard with no index.

Detecting the id by UUID shape works today but dies with the planned move to hrids everywhere: nothing would match, and every switch would fall back to the module root anyway.

Doing it properly means modules declaring their route paths to the host through the SDK. That is format agnostic, so it survives the hrid migration, but it is a cross-repo contract. The truncation in this PR is its correct fallback for any module that has not adopted it, so this work is not thrown away.

## Out of scope

`permission-sync` clears environment permissions and reloads them asynchronously, so a module can still render briefly without them. The remount makes this less visible but does not close it.

Before:

https://github.com/user-attachments/assets/a68fc0ab-e635-46e5-8d87-9f5c578ebd21

After:

https://github.com/user-attachments/assets/d0ac5555-d22e-4f7c-8231-19f1f91b4f01

### DIFF
--- a/gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/environment/EnvironmentGuard.tsx
+++ b/gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/environment/EnvironmentGuard.tsx
@@ -19,6 +19,7 @@ import { type NavigateFunction, Outlet, useLocation, useNavigate } from 'react-r
 import { useEnvironmentStore } from './environment.store';
 import type { Environment } from './environment.types';
 import { getPrimaryHrid, resolveEnvironmentFromSegment, shouldRewriteIdToHrid, useEnvHrid } from './environment.utils';
+import { ContentSkeleton } from '../../shared/components/ContentSkeleton';
 
 function redirectToFirst(env: Environment, navigate: NavigateFunction, search: string, hash: string) {
     navigate({ pathname: `/environments/${getPrimaryHrid(env)}/home`, search, hash }, { replace: true });
@@ -42,6 +43,7 @@ export function EnvironmentGuard() {
     locationRef.current = location;
 
     const environments = useEnvironmentStore(s => s.environments);
+    const environmentId = useEnvironmentStore(s => s.environmentId);
     const setCurrentEnvironment = useEnvironmentStore(s => s.setCurrentEnvironment);
     const lastSyncedId = useRef<string | null>(null);
 
@@ -72,6 +74,10 @@ export function EnvironmentGuard() {
 
     if (!environments.length) {
         return null;
+    }
+
+    if (resolveEnvironmentFromSegment(environments, envHrid)?.id !== environmentId) {
+        return <ContentSkeleton />;
     }
 
     return <Outlet />;
--- a/gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/modules/components/RemoteModuleRoute.tsx
+++ b/gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/modules/components/RemoteModuleRoute.tsx
@@ -16,6 +16,7 @@
 import { loadRemote } from '@module-federation/runtime';
 import { lazy, type LazyExoticComponent, type ComponentType } from 'react';
 
+import { useEnvironmentStore } from '../../environment/environment.store';
 import type { GammaModule } from '../modules.types';
 
 const lazyComponentCache = new Map<string, LazyExoticComponent<ComponentType>>();
@@ -34,7 +35,21 @@ export function getOrCreateLazyModule(remoteName: string, exposedModule: string)
     return cached;
 }
 
+/**
+ * Mounts a federated module, keyed by environment so that switching environments remounts it.
+ *
+ * Everything a module holds below the SDK -- component state, stores, caches -- is scoped to the
+ * environment it was loaded for. Remounting drops it without asking module authors for teardown
+ * logic, which matters because modules ship from their own repositories. The technical id is the
+ * key rather than the URL segment, so canonicalizing an id to its hrid does not remount.
+ * The lazy component itself is cached, so this re-renders the remote without re-fetching it.
+ *
+ * EnvironmentGuard only renders this once the store environment matches the URL, so the key is
+ * always the environment the URL addresses.
+ */
 export function RemoteModuleRoute({ module }: { readonly module: GammaModule }) {
+    const environmentId = useEnvironmentStore(s => s.environmentId);
     const LazyModule = getOrCreateLazyModule(module.remoteName, module.exposedModule);
-    return <LazyModule />;
+
+    return <LazyModule key={environmentId} />;
 }
--- a/gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/environment/EnvironmentGuard.spec.tsx
+++ b/gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/environment/EnvironmentGuard.spec.tsx
@@ -18,18 +18,23 @@ import { MemoryRouter, Route, Routes } from 'react-router-dom';
 
 import { useEnvironmentStore } from './environment.store';
 import { EnvironmentGuard } from './EnvironmentGuard';
+import { EnvironmentRenderProbe } from '../../testing/EnvironmentRenderProbe';
 import { buildEnvironment, TEST_ENVIRONMENTS } from '../../testing/factories';
 import { resetAllStores, seedBootstrap, seedEnvironments } from '../../testing/helpers';
 import { PathnameProbe } from '../../testing/PathnameProbe';
 
 describe('EnvironmentGuard', () => {
     let lastPath = '';
+    const realSetCurrentEnvironment = useEnvironmentStore.getState().setCurrentEnvironment;
 
     beforeEach(() => {
         resetAllStores();
         seedBootstrap();
         lastPath = '';
     });
+
+    // `reset()` restores state, not actions, so a stubbed action would leak into the next test.
+    afterEach(() => useEnvironmentStore.setState({ setCurrentEnvironment: realSetCurrentEnvironment }));
 
     function renderGuard(initialPath: string) {
         return render(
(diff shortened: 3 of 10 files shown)


gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/environment/EnvironmentGuard.tsx
```diff
@@ -74,5 +76,9 @@ export function EnvironmentGuard() {
         return null;
     }
 
+    if (resolveEnvironmentFromSegment(environments, envHrid)?.id !== environmentId) {
+        return <ContentSkeleton />;
```
ytvnr, line 80: **Worth fixing**: the guard un-withholds on the very store commit that clears the environment permissions, so every module subtree now mounts against an empty permission set until the reload completes.

`setCurrentEnvironment` is called from the guard effect, zustand notifies `onEnvironmentChangeForPermissions`, which runs `permissionService.clear('environment')` synchronously and then fires the async `loadEnvironmentPermissions` (`features/permissions/permission-sync.ts:34-35`). The same state change flips this condition to a match, so `<Outlet />`, and with it a freshly mounted `RemoteModuleRoute`, renders while `permissionService` holds no environment permission. Modules gate their UI on that through the SDK (`useHasPermission`, `PermissionGate`), so each one paints its denied or empty branch for the length of the permissions request, not for a frame.

Before this branch the module was not remounted on a switch, so nothing re-evaluated a mount time permission gate. The PR lists permission-sync as out of scope and says the remount makes this less visible. From here it reads the other way around: the remount moves the empty window to mount time, where it has more effect.

Two ways out, both outside this file or not: extend the withhold condition to cover permission readiness, since the guard is already the single place that decides when the subtree may render, or have permission-sync clear only once the replacement set has arrived.

gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/modules/components/RemoteModuleRoute.tsx
```diff
@@ -34,7 +35,21 @@ export function getOrCreateLazyModule(remoteName: string, exposedModule: string)
...
+ *
+ * EnvironmentGuard only renders this once the store environment matches the URL, so the key is
+ * always the environment the URL addresses.
+ */
 export function RemoteModuleRoute({ module }: { readonly module: GammaModule }) {
+    const environmentId = useEnvironmentStore(s => s.environmentId);
     const LazyModule = getOrCreateLazyModule(module.remoteName, module.exposedModule);
-    return <LazyModule />;
+
+    return <LazyModule key={environmentId} />;
```
ytvnr, line 54: `key={environmentId}` cannot change while the component is mounted, because `EnvironmentGuard` has already unmounted the subtree before the id changes.

The guard returns `<ContentSkeleton />` instead of `<Outlet />` for the whole mismatch window, and a different component type at the same position unmounts the subtree. `RemoteModuleRoute` is only ever rendered under that guard (`app/AppRoutes.tsx:63`), so it always mounts fresh with its final key, and the store commit that changes `environmentId` is the one that restores `<Outlet />`. `RemoteModuleRoute.spec.tsx` exercises the key by rendering the component outside the guard, which is a configuration the app never produces.

Not a bug, and the doc comment already notes the guard ordering. It is worth deciding which of the two mechanisms owns the remount. If the key is deliberate defence in depth, a test that mounts the component under the guard would keep the assertion tracking the real tree.

gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/environment/EnvironmentGuard.tsx
ytvnr, line 80: Rendering `ContentSkeleton` in place of `<Outlet />` also takes the sidebar and the breadcrumbs off screen, not just the content area.

`RouteLayout` and the federated modules register `navigation`, `breadcrumbs` and `contextSidebar` through `useLayoutConfig`, and graphene resets the keys a hook owns on unmount. During the withhold the sidebar navigation and breadcrumbs therefore disappear while the content area shows `ContentSkeleton`, whose card grid shape is the host home dashboard rather than the module page being loaded. Focus also falls back to `<body>`, since the environment menu item is unmounted.

Keeping the previous subtree visible in an inert state, or at least using a neutral skeleton for the module area and moving focus deliberately once the new environment renders, would make the switch read better.

ytvnr (approved):
The fix holds up. I followed the switch, back/forward, deep link, unknown environment and id to hrid canonicalization paths through `EnvironmentGuard`, `AppRoutes` and the store, and found no deadlock, no remount loop and no infinite skeleton. The truncation in `routes.ts` behaves on every path shape I tried, including the host areas and a path with nothing after the environment.

The one point I would look at before merge is the permission window the remount opens at mount time, commented inline. The other two remarks are minor: one on the remount key, one on what the withholding takes off screen with it.

gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/environment/EnvironmentGuard.tsx
Okhelifi: Confirmed on the mechanism: `permissionService.clear('environment')` runs synchronously
inside the zustand subscription, on the same commit that flips this condition to a match.

I do not think the conclusion holds though. I looked for a mount time permission decision
in apim (a redirect, a `useState` initialiser) and there is none: all 61 call sites go
through `useHasPermission`, which subscribes to `permissionService` and re-renders when the
new set lands. So the empty window is not created by the remount. `clear()` bumps the
version, which already flipped every mounted gate to its denied branch before this branch
existed. The remount changes when the gates are first evaluated, not whether they are
evaluated against an empty set.

aim is not affected at all: it loads environment permissions itself through
`useQuery(['env-permissions', envId])` and exposes `useEnvironmentPermissionsReady()`, with
a skeleton while pending and a fail open on error (`RequireCatalogRead`).

Extending the withhold to permission readiness means holding the subtree for a full network
round trip, which also stretches the slot reset you describe in the other comment from one
frame to one request. That is a real improvement to make, but it is a pre-existing window
and not the bug this PR fixes, so I would rather not fold it in here.

Worth a follow up in its own right: aim built a readiness signal because the host does not
expose one, and apim has nothing. A host level `environmentPermissionsReady` in the SDK
would let the guard gate on it and let aim drop its duplicate. I will open a ticket.

gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/modules/components/RemoteModuleRoute.tsx
Okhelifi: Agreed, and thanks for tracing it. The guard swaps `<Outlet />` for `<ContentSkeleton />`,
a different component type at the same position, so the subtree is already unmounted before
`environmentId` changes. The guard owns the remount, not the key.

Decision: the guard owns it, and the key stays as a local contract so the component is
correct on its own. It would become load bearing again if we adopt what you suggest in the
other comment and keep the previous subtree mounted during the transition.

I moved the assertion to the real tree, as you suggested, so it tracks what the app
actually renders: `createMemoryRouter` with `EnvironmentGuard` as the parent route and
`RemoteModuleRoute` as its child, then a navigation from `env-1` to `env-2`. It asserts the
remount as behaviour rather than pinning the mechanism, so it stays valid whichever of the
two owns it. The remaining out of guard test only covers "a re-render must not remount".

gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/environment/EnvironmentGuard.tsx
Okhelifi: Confirmed on the slots: `useLayoutConfig` returns `resetSlots(ownedKeys)` from its
`useLayoutEffect`, so the module's navigation and breadcrumbs are dropped when the subtree
unmounts.

One correction on the duration: the sync runs in a passive effect, so this is a single
painted frame, not a window of the length your comment implies. It would become one network
round trip if we gated on permission readiness as suggested in your other comment, which is
part of why I would rather keep that out of this PR.

On focus, I do not think it happens: `AppContextBar` lives in `ShellLayout`'s
`ContentHeader`, outside the guard's `Outlet`, so the environment selector is never
unmounted. The menu closes on selection, which is normal Radix behaviour, and focus returns
to its trigger.

The skeleton shape point is fair, `ContentSkeleton` is the host dashboard shape and reads
oddly over a module area. Noted for the follow up, together with keeping the previous
subtree visible.

PR #19210, opened 2026-08-18
chore(release): gather changelog tickets from every product board

vikrantgravitee:
## Issue

Ticket: [APIM-14996](https://gravitee.atlassian.net/browse/APIM-14996)

## Description

The changelog was generated from the APIM board alone, so a fix released on one of the modules that now ship with APIM never reached the release note.

### Collecting the tickets

`JIRA_PROJECTS` lists the boards a release can draw from (`APIM,ESM,AIAM,PORTAL,OBS,AUTHZ,FOUND,BX`), overridable so adding a board needs no code change. Each is checked for a version of the released name, and the ones that carry it are searched together with `project IN (...) AND fixVersion = "<name>"`.

- A board the token cannot read is skipped with a warning rather than costing the release its whole changelog. If *no* board could be read the lookup throws instead, because that empty result means "we never found out" rather than "nothing to release" — which the job otherwise takes as a clean no-op, exiting green having published no changelog. One unreadable board among others stays a warning on purpose: this job runs for prereleases too, whose version exists on no board at all.
- An error-shaped search response throws, rather than committing an empty changelog as if nothing public had shipped.
- The search is paged to the end; a release spanning several boards passes one page easily.
- Remote-link lookups for the GitHub issue number run concurrently, and warn when the lookup itself fails — a ticket with no link answers with an empty list, so an error shape means something else went wrong.

### Grouping the sections

Grouping everything by component put every non-APIM ticket under `Other`: the component list is APIM's own and no other board fills that field in. So the board each ticket came from is carried through, and **the product sections replace `Other` for those boards — not the component breakdown**:

- A **component** takes the ticket whichever board raised it, so a ticket carrying a component keeps exactly the placement, and the line count, it has today.
- A **product section** takes only that board's tickets which no component section already covers — the ones that would otherwise have landed in `Other`.
- Two plain lists, each where it is used: `JiraBoards` in `jira-helper.mjs` is the set the release searches, and `ProductSections` in `changelog-helper.mjs` maps the boards that head a section of their own to its name. A board absent from `ProductSections` groups by APIM component instead, and by `Other` when the ticket carries none — `APIM` itself, the team boards that track APIM work rather than a product (`FOUND`, `BX`), and anything reached through `JIRA_PROJECTS`.

So the only thing that changes for any existing ticket is the heading a componentless non-APIM ticket sits under: `Other` becomes the product's name.

## Additional context

**No changelog line is lost or gained.** Checked against the current `master` helper across 189 combinations of board, ticket type and component shape:

```
tickets DROPPED by new            : 0
tickets with a CHANGED line count : 0
total bullet lines old / new      : 144 / 144
```

The only movement is 24 tickets going from `Other` to a product heading — same single line each, non-APIM boards only.

**Verified against 4.13.0.** All 66 public tickets fixed in 4.13.0 are on the APIM board (55 Public Bug, 10 Public Improvement, 1 Public Security), so the rendered changelog is byte-identical to what the current code generates (9,574 bytes both ways). AIAM and FOUND do have 4.13.0 tickets, but all are `Story`, `Epic` or `Private Bug` and are correctly excluded by the public-type filter.

**Tests.** 33 tests in `release/helpers/*.test.mjs`, run with `cd release && yarn test`. They execute in the release-notes job, not on PR CI.

**Out of scope.** `Public Security` tickets are collected by `jira-helper` but `ChangelogSections` defines no section for that type, so they never reach the changelog (`APIM-14809` in 4.13.0). This is pre-existing on `master` — identical before and after this PR — and is being raised separately.

[APIM-14996]: https://gravitee.atlassian.net/browse/APIM-14996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/release/ci-steps/generate-changelog.mjs
+++ b/release/ci-steps/generate-changelog.mjs
@@ -1,7 +1,7 @@
 import { syncProcessCwd } from 'zx';
 import { computeVersion, extractVersion } from '../helpers/version-helper.mjs';
-import { getJiraIssuesOfVersion, getJiraVersion } from '../helpers/jira-helper.mjs';
-import { ChangelogSections, ComponentTypes, getTicketsFor } from '../helpers/changelog-helper.mjs';
+import { getJiraIssuesOfVersions, getJiraVersions } from '../helpers/jira-helper.mjs';
+import { ChangelogSections, ComponentTypes, getTicketsFor, getTicketsForProduct, productKeysOf } from '../helpers/changelog-helper.mjs';
 
 syncProcessCwd(); // restores legacy v7 behavior
 
@@ -29,16 +29,20 @@ echo(chalk.blue(`# Clone ${docRepository} repository`));
 await $`git clone --depth 1  ${docRepositoryURL} --single-branch --branch=main`;
 cd(docRepository);
 
-const version = await getJiraVersion(releasingVersion);
-if (version === undefined) {
-  echo(chalk.blue(`No Jira release found for: ${releasingVersion}, nothing to do.`));
+const jiraProjectKeys = await getJiraVersions(releasingVersion);
+if (jiraProjectKeys.length === 0) {
+  echo(chalk.blue(`No Jira release found for: ${releasingVersion} in any project, nothing to do.`));
   process.exit(0);
 }
-const issues = await getJiraIssuesOfVersion(version.id);
+echo(chalk.blue(`# Jira projects releasing ${releasingVersion}: ${jiraProjectKeys.join(', ')}`));
+
+const issues = await getJiraIssuesOfVersions(jiraProjectKeys, releasingVersion);
 
 let changelogPatchTemplate = `
 ## Gravitee API Management ${releasingVersion} - ${new Date().toLocaleDateString('en-US', dateOptions)}
 `;
+
+const productKeys = productKeysOf(issues);
 
 ChangelogSections.forEach((section) => {
   let changelogSection = '';
@@ -46,6 +50,12 @@ ChangelogSections.forEach((section) => {
     const ticketsForComponent = getTicketsFor(issues, componentType, section.ticketType);
     if (ticketsForComponent) {
       changelogSection += ticketsForComponent;
+    }
+  });
+  productKeys.forEach((projectKey) => {
+    const ticketsForProduct = getTicketsForProduct(issues, projectKey, section.ticketType);
+    if (ticketsForProduct) {
+      changelogSection += ticketsForProduct;
     }
   });
   if (changelogSection) {
--- a/release/ci-steps/generate-changelog.mjs
+++ b/release/ci-steps/generate-changelog.mjs
@@ -1,7 +1,7 @@
 import { syncProcessCwd } from 'zx';
 import { computeVersion, extractVersion } from '../helpers/version-helper.mjs';
 import { getJiraIssuesOfVersions, getJiraVersions } from '../helpers/jira-helper.mjs';
-import { ChangelogSections, ComponentTypes, getTicketsFor, getTicketsForProduct, productKeysOf } from '../helpers/changelog-helper.mjs';
+import { ChangelogSections, getTicketSections } from '../helpers/changelog-helper.mjs';
 
 syncProcessCwd(); // restores legacy v7 behavior
 
@@ -42,22 +42,8 @@ let changelogPatchTemplate = `
 ## Gravitee API Management ${releasingVersion} - ${new Date().toLocaleDateString('en-US', dateOptions)}
 `;
 
-const productKeys = productKeysOf(issues);
-
 ChangelogSections.forEach((section) => {
-  let changelogSection = '';
-  [...ComponentTypes, 'Other'].forEach((componentType) => {
-    const ticketsForComponent = getTicketsFor(issues, componentType, section.ticketType);
-    if (ticketsForComponent) {
-      changelogSection += ticketsForComponent;
-    }
-  });
-  productKeys.forEach((projectKey) => {
-    const ticketsForProduct = getTicketsForProduct(issues, projectKey, section.ticketType);
-    if (ticketsForProduct) {
-      changelogSection += ticketsForProduct;
-    }
-  });
+  const changelogSection = getTicketSections(issues, section.ticketType).join('');
   if (changelogSection) {
     changelogPatchTemplate += `<details>
 
--- a/release/helpers/changelog-helper.mjs
+++ b/release/helpers/changelog-helper.mjs
@@ -1,12 +1,16 @@
 export const ComponentTypes = ['Gateway', 'Management API', 'Console', 'Portal', 'Helm Charts'];
 
-const ProductSections = {
+export const ProductSections = {
   ESM: 'Event Stream Management',
   AIAM: 'AI Agent Management',
-  PORTAL: 'Portal',
   OBS: 'Observability',
   AUTHZ: 'Authorization Management',
 };
+
+// Boards whose tickets belong to a component section that already exists rather than to one of their
+// own: the APIM 'Portal' component and the Portal board are the same product. A board appears in at
+// most one of ProductSections and BoardComponents.
+export const BoardComponents = { PORTAL: 'Portal' };
 
 export const ChangelogSections = [
   {
@@ -22,7 +26,7 @@ export const ChangelogSections = [
 /**
  * Get the Ascii doc formatted changelog for input issues
  *
- * @param issues {Array<{id: string, githubIssue: string, summary: string, components: Array<string>, type: string>}
+ * @param issues {Array<{id: string, githubIssue: string, summary: string, components: Array<{name: string}>, type: string>}
  */
 export function getChangelogFor(issues) {
   return issues
@@ -52,6 +56,10 @@ function hasProductSection(issue) {
   return Boolean(ProductSections[issue.project]);
 }
 
+function hasComponentSection(issue) {
+  return Boolean(BoardComponents[issue.project]);
+}
+
 function headedSection(title, tickets) {
   if (tickets.length === 0) {
     return undefined;
@@ -69,13 +77,15 @@ function carriesNoComponent(issue) {
 }
 
 /**
- * Get the tickets for a given component and a given type of tickets. If type is 'Other', it will return tickets with no known component from the boards that have no product section.
+ * Get the tickets for a given component and a given type of tickets. If type is 'Other', it will return
+ * tickets with no known component from the boards that have neither a product section nor a mapped
+ * component section.
  *
- * <p>A component takes the ticket whichever board raised it. Only 'Other' is narrowed: a board with a
- * product section has that bucket replaced by the section, so catching its tickets here too would list
- * them twice.
+ * <p>A component takes the ticket whichever board raised it, and a board mapped in BoardComponents adds
+ * its componentless tickets to that component's section. Only 'Other' is narrowed: a board with a
+ * section of its own has its buckets replaced, so catching its tickets here too would list them twice.
  *
- * @param issues {Array<{id: string, project: string, githubIssue: string, summary: string, components: Array<string>, type: string>}
+ * @param issues {Array<{id: string, project: string, githubIssue: string, summary: string, components: Array<{name: string}>, type: string>}
  * @param componentType {ComponentTypes | 'Other'}
  * @param ticketType {'Public Bug' | 'Public Improvement'}
  */
@@ -83,23 +93,32 @@ export function getTicketsFor(issues, componentType, ticketType) {
   const ticketsOfType = issues.filter((issue) => issue.type === ticketType);
   const componentsTickets =
     componentType === 'Other'
-      ? ticketsOfType.filter((issue) => !hasProductSection(issue) && carriesNoComponent(issue))
-      : ticketsOfType.filter((issue) => issue.components.some((cmp) => cmp.name === componentType));
+      ? ticketsOfType.filter((issue) => !hasProductSection(issue) && !hasComponentSection(issue) && carriesNoComponent(issue))
+      : ticketsOfType.filter(
+          (issue) =>
+            issue.components.some((cmp) => cmp.name === componentType) ||
+            (BoardComponents[issue.project] === componentType && carriesNoComponent(issue)),
+        );
 
   return headedSection(componentType, componentsTickets);
 }
 
 /**
- * Get the tickets of one product board that no component section already covers, headed by the product's name.
+ * Get the tickets of one product board that no component section already covers, headed by the product's
+ * name. Only boards returned by productKeysOf have a product section; any other key renders nothing.
  *
- * @param issues {Array<{id: string, project: string, githubIssue: string, summary: string, components: Array<string>, type: string>}
+ * @param issues {Array<{id: string, project: string, githubIssue: string, summary: string, components: Array<{name: string}>, type: string>}
  * @param projectKey {string}
  * @param ticketType {'Public Bug' | 'Public Improvement'}
  */
 export function getTicketsForProduct(issues, projectKey, ticketType) {
+  const title = ProductSections[projectKey];
+  if (!title) {
+    return undefined;
+  }
   const productTickets = issues.filter((issue) => issue.project === projectKey && issue.type === ticketType && carriesNoComponent(issue));
 
-  return headedSection(ProductSections[projectKey], productTickets);
+  return headedSection(title, productTickets);
 }
 
 /**
@@ -113,4 +132,21 @@ export function productKeysOf(issues) {
   const present = new Set(issues.filter(hasProductSection).map((issue) => issue.project));
 
   return Object.keys(ProductSections).filter((projectKey) => present.has(projectKey));
+}
+
+/**
+ * The sections of one changelog block as rendered strings, in display order: the long-standing
+ * component sections first, then one section per product board present, then Other last — the catch-all
+ * must not sit between the named sections.
+ *
+ * @param issues {Array<{id: string, project: string, githubIssue: string, summary: string, components: Array<{name: string}>, type: string>}
+ * @param ticketType {'Public Bug' | 'Public Improvement'}
+ * @returns {Array<string>}
+ */
+export function getTicketSections(issues, ticketType) {
+  return [
+    ...ComponentTypes.map((componentType) => getTicketsFor(issues, componentType, ticketType)),
+    ...productKeysOf(issues).map((projectKey) => getTicketsForProduct(issues, projectKey, ticketType)),
+    getTicketsFor(issues, 'Other', ticketType),
+  ].filter(Boolean);
 }
--- a/release/helpers/changelog-helper.mjs
+++ b/release/helpers/changelog-helper.mjs
@@ -1,4 +1,13 @@
 export const ComponentTypes = ['Gateway', 'Management API', 'Console', 'Portal', 'Helm Charts'];
+
+const ProductSections = {
+  ESM: 'Event Stream Management',
+  AIAM: 'AI Agent Management',
+  PORTAL: 'Portal',
+  OBS: 'Observability',
+  AUTHZ: 'Authorization Management',
+};
+
 export const ChangelogSections = [
   {
     ticketType: 'Public Bug',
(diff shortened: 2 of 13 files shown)


release/helpers/jira-helper.mjs
```diff
@@ -1,53 +1,95 @@
...
- * }>}
+ * The Jira projects a release can draw tickets from.
+ *
+ * <p>A release note used to come from APIM alone. The modules that ship alongside it now have boards of
+ * their own, and a ticket fixed in one of them belongs in the same changelog as the APIM ones. Override
+ * with JIRA_PROJECTS (comma separated) to add a board without a code change.
  */
-export function getJiraVersion(versionName) {
-  const token = process.env.JIRA_TOKEN;
+export const JiraProjects = (process.env.JIRA_PROJECTS ?? 'APIM,ESM,AIAM,PORTAL,OBS,AUTHZ')
```
phiz71: why not FOUND or BX?

release/helpers/jira-helper.mjs
```diff
@@ -1,53 +1,95 @@
...
+  const issuesFromJira = [];
+  let nextPageToken;
+  do {
+    const page = await fetch('https://gravitee.atlassian.net/rest/api/3/search/jql', {
+      method: 'POST',
+      headers: { ...jiraHeaders(), 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        jql,
+        maxResults: 100,
+        fields: ['issuetype', 'summary', 'components', 'project', 'customfield_10115'],
```
phiz71: are we sure that 'customfield_10115' is the same field name in other projects than APIM ?

release/helpers/jira-helper.mjs
```diff
@@ -59,22 +101,23 @@ export async function getJiraIssuesOfVersion(versionId) {
     )
     .map((issue) => ({
       key: issue.key,
+      project: issue.fields.project?.key ?? issue.key.split('-')[0],
```
phiz71: This field is never read. `getTicketsFor` and `getChangelogFor` group on `components` only, so `issue.project` is carried through the whole pipeline and dropped — while `jira-helper.test.mjs` asserts it "so the changelog can group by product", which nothing does.

It matters more than dead code usually would, because of what the data looks like: `project IN (ESM, AIAM, PORTAL, OBS, AUTHZ, FOUND, BX) AND component IS NOT EMPTY` returns **0 issues**. Components are an APIM-board habit; the other boards do not use them at all. So `issue.components.every(...)` is vacuously true for every one of them and they all fall into `Other`, mixed in with the APIM tickets that have no component, with no indication of which product they belong to.

So either `changelog-helper.mjs` learns to use `project` (a section per board, or a prefix on the line), or the field and its test come out.

release/helpers/jira-helper.mjs
```diff
@@ -1,53 +1,95 @@
...
+      headers: { ...jiraHeaders(), 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        jql,
+        maxResults: 100,
+        fields: ['issuetype', 'summary', 'components', 'project', 'customfield_10115'],
+        ...(nextPageToken ? { nextPageToken } : {}),
+      }),
+    }).then((response) => response.json());
+
+    issuesFromJira.push(...(page.issues ?? []));
```
phiz71: If Jira answers this call with an error object rather than a result page, `page.issues ?? []` yields nothing and `page.nextPageToken` is `undefined`, so the loop exits and the function returns zero issues — silently. `generate-changelog.mjs` then carries on: it writes a changelog that is nothing but its own heading, commits it, pushes the branch and opens the docs PR.

Before this change, `body.issues` being `undefined` made the following `.filter` throw and the job went red. That was ugly, but it was visible.

It is also asymmetric with `getJiraVersions` just above, which does check the shape of what came back. An `Array.isArray(page.issues)` guard that throws (or at least surfaces `page.errorMessages`) would put the two on the same footing.

release/helpers/jira-helper.mjs
phiz71: The field id itself is fine: custom field ids are global in Jira Cloud, `cf[10115]` resolves against every board, so there is no risk of hitting a different field elsewhere.

What is not fine is that nobody outside APIM fills it in. `project = APIM AND cf[10115] IS NOT EMPTY` → **1890** issues; the same query over `ESM, AIAM, PORTAL, OBS, AUTHZ, FOUND, BX` → **0**.

So every ticket coming from a new board will fall through to the remote-link lookup below — one sequential REST call per ticket, on a loop that was written when only APIM stragglers hit it. And when there is no remote link to `gravitee-io/issues` either, the changelog line ends up with no GitHub link at all. Worth deciding whether that fallback is good enough, or whether those boards should start populating the field.

release/helpers/jira-helper.mjs
phiz71: Why these six and not FOUND or BX? Both projects exist, are readable with the same token, and already accept the `Public Bug` type (there is one on them today). They are APIM working boards — Builder Experience and Foundations — so nothing technical sets them apart from the six listed here.

Either add them, or say in the PR description why they are deliberately out of scope, so the next person does not have to re-derive the answer.

release/helpers/jira-helper.mjs
```diff
@@ -1,53 +1,95 @@
...
+ * @param versions {Array<{projectKey: string, name: string}>}
+ * @returns {Promise<Array<{key: string, project: string, githubIssue: string, summary: string, components: Array<object>, type: string}>>}
+ */
+export async function getJiraIssuesOfVersions(versions) {
+  if (versions.length === 0) {
+    return [];
+  }
+
+  const projects = [...new Set(versions.map((version) => version.projectKey))].join(', ');
+  const jql = `project IN (${projects}) AND fixVersion = "${versions[0].name}"`;
```
phiz71: The signature says "versions", but only `versions[0].name` is ever used — every other `name` in the array is ignored. They happen to be identical, since `getJiraVersions` found them all by that one name, but the signature promises something the body does not honour, and a caller passing a mixed array would silently get results for the first one only.

`getJiraIssuesOfVersions(projectKeys, versionName)` would say what actually happens, and would let `getJiraVersions` return plain project keys.

phiz71: Separate point on the same line: the project keys go into the JQL unquoted, and they can come from `JIRA_PROJECTS`. Quoting them — `project IN ("APIM", "ESM")` — costs nothing and removes the question of what happens with a key that collides with JQL syntax.

release/helpers/jira-helper.test.mjs
```diff
@@ -0,0 +1,180 @@
...
+          { id: '100', name: '4.9.0' },
+          { id: '101', name: '4.8.0' },
+        ];
+      if (url.includes('/project/AIAM/')) return [{ id: '200', name: '4.9.0' }];
+      return []; // the other boards do not release this version
+    });
+
+    const versions = await getJiraVersions('4.9.0');
+
+    assert.deepEqual(versions, [
```
phiz71: This assertion — and the `OBS` one further down — depends on the ambient environment, because `JiraProjects` is resolved once at import time from `process.env`. With `JIRA_PROJECTS=APIM` set, two tests in this file fail.

That is precisely the escape hatch the PR introduces, so it is not a hypothetical: the day the release job exports the variable, the suite starts failing for a reason unrelated to what it is testing. Reading the list at call time (`jiraProjects()`), or taking it as a parameter with a default, would make these tests state their own inputs.

release/helpers/jira-helper.test.mjs
```diff
@@ -0,0 +1,180 @@
+import assert from 'node:assert/strict';
```
phiz71, line 1: Nothing runs these. `release/package.json` has no `test` script and no CircleCI job invokes `node --test` under `release/`. They do pass — `node --test "helpers/*.test.mjs"` is green, 13/13, and prettier is clean — but as it stands they will rot without anyone noticing.

The hook already exists: the PR workflow runs `yarn` with `working_directory: ./release`. Adding `"test": "node --test \"helpers/*.test.mjs\""` and a step next to that install would be enough.

phiz71 (changes requested):
The direction is right and the helper is much better structured than what it replaces — paging in particular was a real bug. Two things need to change before this can land, and there is a third that is not about the code at all.

**The changelog cannot actually tell the boards apart.** `project` is read and carried all the way through, but nothing consumes it: grouping is done on `components` alone. And no ticket on any of the six boards has a component — `project IN (ESM, AIAM, PORTAL, OBS, AUTHZ) AND component IS NOT EMPTY` returns 0. So every non-APIM ticket lands in `Other`, next to the APIM tickets that have no component, with nothing saying which product it came from. Either `changelog-helper.mjs` starts using `project`, or the field and the test that claims it enables grouping should go.

**A failed search now writes an empty changelog instead of failing the job.** Details inline on the paging loop.

**And the process side.** I checked what is actually on those boards today: across ESM, AIAM, PORTAL, OBS, AUTHZ (plus FOUND and BX) there are 13 `Public Bug`/`Public Improvement` tickets and 7 tickets carrying a `fixVersion` — and the intersection is empty. `PORTAL-115` (*Private* Bug) carries `fixVersion 4.12.16`; its public twin `PORTAL-116`, same bug same title, carries none. The habit on these boards is to put the version on the internal ticket, not the public one. So as things stand this code will run correctly and gather nothing, and — because of the point above about silent failures — nobody will notice. Worth agreeing the convention with those teams before merging, otherwise the feature ships dead.

One thing that does hold up nicely: version names line up across boards (AIAM and PORTAL both have `4.12.16`, OBS has `4.12.15`, AIAM has `4.13.0`), so matching by name is sound and the reasoning in the commit message about non-ambiguity is correct.

Minor, not blocking: the commit message has no JIRA footer and the PR body still has the `APIM-XXX` placeholder.

release/helpers/jira-helper.mjs
vikrantgravitee: Removed issue.project and the test claiming it "enables grouping by product". Now, grouping stays component-based aligned with APIM logic.

release/helpers/jira-helper.mjs
vikrantgravitee: getJiraIssuesOfVersions now throws on a non-array search page, surfacing errorMessages, with a test asserting it.

release/helpers/jira-helper.mjs
vikrantgravitee: versions[0].name signature / unquoted JQL- Reshaped to getJiraIssuesOfVersions(projectKeys, versionName); JQL keys are now quoted and embedded quotes escaped.

release/helpers/jira-helper.mjs
vikrantgravitee: Parallelized the remote-link fallback with a concurrency-proving test; field is still unpopulated outside APIM though.

Even on APIM when people miss this field change log enters into OTHERs category and same will be followed. New boards are created so we might need people to remind again that these fields should be filled properly.

release/helpers/jira-helper.mjs
vikrantgravitee: Added both to the default board list. These were missed because list shared with me only had support boards.

release/helpers/jira-helper.mjs
vikrantgravitee: Quoted the project keys in the JQL (project IN ("APIM", "ESM"))

release/helpers/jira-helper.test.mjs
vikrantgravitee: jiraProjects() now reads process.env.JIRA_PROJECTS at call time instead of once at import, and getJiraVersions takes an optional projectKeys param so tests pass their own board list explicitly instead of depending on the ambient default.

release/helpers/jira-helper.test.mjs
vikrantgravitee: Done under this commit: d80ee56ff70eecba1476520d84c0391a7c522beb

phiz71 (commented):
Thanks — all eight points are addressed, and I checked rather than took it on trust:

- `node --test` is green, 17/17, and prettier is clean on `release/`.
- `JIRA_PROJECTS=APIM node --test "helpers/*.test.mjs"` now passes 17/17 too — that was the failing case before, so reading the env at call time genuinely fixed it, not just moved it.
- CI is fully green including `generate-config`, so the six regenerated workflow fixtures match the new `yarn test` step.

The concurrency worry I raised turns out to be smaller than I implied, and it is worth recording so nobody re-opens it: across the whole APIM history there are **77** public tickets with a `fixVersion` and no `customfield_10115`, and **0** for `4.12.0`. So the remote-link fallback fires for a handful of tickets per release at most — the unbounded `Promise.all` is fine at that scale.

Two things left that are not code, and one nuance inline.

**The convention on the new boards is still the blocker for this actually doing anything.** You are right that a missing GitHub-issue field just pushes a ticket into `Other`, and that is survivable. The harder one is the `fixVersion`: on ESM, AIAM, PORTAL, OBS, AUTHZ, FOUND and BX there are 13 public tickets and 7 tickets carrying a `fixVersion`, and no ticket is in both sets. `PORTAL-115` (Private Bug) carries `4.12.16`, its public twin `PORTAL-116` carries none. Until those teams start putting the version on the public ticket, this code runs correctly and gathers nothing. Worth a follow-up ticket so the reminder does not live only in this thread.

**Second**, now that `project` is gone, every non-APIM ticket lands in `Other` — none of those boards uses components at all (`component IS NOT EMPTY` returns 0 across the seven). That was the trade-off I offered and removing the field was a reasonable call, but the changelog reader will see a single undifferentiated `Other` bucket for five products. A line in the PR description saying so would save the next person the archaeology.

Minor: the PR description still has `Ticket:` empty.

Not re-blocking on any of this — the code side is done as far as I am concerned.

release/helpers/jira-helper.mjs
```diff
@@ -1,55 +1,105 @@
...
+ * @param versionName {string}
+ * @returns {Promise<Array<{key: string, githubIssue: string, summary: string, components: Array<object>, type: string}>>}
+ */
+export async function getJiraIssuesOfVersions(projectKeys, versionName) {
+  if (projectKeys.length === 0) {
+    return [];
+  }
+
+  const projects = projectKeys.map((key) => `"${key.replace(/"/g, '\\"')}"`).join(', ');
+  const jql = `project IN (${projects}) AND fixVersion = "${versionName}"`;
```
phiz71, line 80: One nuance now that the search throws rather than swallowing. `getJiraVersions` goes to some trouble to be tolerant board by board: an unreadable project is warned about and skipped so that "one unreadable board must not cost the release its whole changelog". But its output then feeds a single JQL that is all-or-nothing — if the token can list a project's versions yet cannot browse its issues, Jira answers `The value 'X' does not exist for the field 'project'` and the throw above now kills the entire changelog, which is exactly the outcome the per-board guard was written to avoid.

In practice reading versions and browsing issues need much the same permission, so this is unlikely rather than impossible. I would not hold the PR for it, but it is worth knowing that the tolerance upstream is undone one step downstream — and if it ever bites, querying per board and concatenating would restore it.

release/helpers/jira-helper.mjs
```diff
@@ -61,25 +111,33 @@ export async function getJiraIssuesOfVersion(versionId) {
...
+          method: 'GET',
+          headers: jiraHeaders(),
+        }).then((response) => response.json());
 
-    const githubIssue = remoteLinks.find((remoteLink) => remoteLink.object.url.includes('https://github.com/gravitee-io/issues/issues/'));
-    if (githubIssue) {
-      issue.githubIssue = githubIssue.object.url.replace('https://github.com/gravitee-io/issues/issues/', '');
-    }
-  }
+        if (!Array.isArray(remoteLinks)) {
```
phiz71, line 134: This one still swallows silently, which is now inconsistent with the search page above.

For a ticket that simply has no remote links the silence is right — that is an ordinary outcome, not a failure. But the same branch also absorbs a 401 or a 429, and with the batch fired concurrently a rate-limited response is the plausible way to get here. The ticket then quietly loses its GitHub link and the changelog line ships without one.

Given the measured volume (a handful per release) this is theoretical today. If you want it airtight, distinguishing "not an array because there are no links" from "not an array because Jira complained" would be enough — a `console.warn` on the second would do, no need to throw.

phiz71:
Following up on the two open threads, with what I measured since. One thing to change in the code, and two that are being handled outside this PR — worth knowing so you don't try to solve them in the script.

## The `project` field should come back — my call was wrong

On the `project` field I offered you a choice between consuming it and removing it, as if the two were equivalent. They are not, and removing it closed the only workable route. That one is on me, not on you.

The reason is that the alternative — components on the other boards — is not reachable. Across ESM, AIAM, PORTAL, OBS, AUTHZ, FOUND and BX, **0 issues out of 1152** carry a component. That is not a habit that can be nudged into place either: `ComponentTypes` is APIM's own breakdown (Gateway, Management API, Console, Portal, Helm Charts), and asking the Observability team to file their tickets under "Gateway" or "Console" would not mean anything. So component-based grouping can never put a non-APIM ticket anywhere except `Other`.

And this is not only about the new boards. On APIM itself, **234 of the 1350** public tickets that carry a fixVersion have no component — so roughly **one changelog line in six already lands in `Other` today**. Grouping by product would improve the APIM changelog too, not just accommodate the newcomers.

Two ways to do it, whichever you prefer:

- a section per board in `changelog-helper.mjs` — APIM tickets keep their current component breakdown, everything else groups under a readable product name;
- or just a prefix on the line, `* [Portal] <summary> [#11651]`, if you want the smallest possible change.

## Two things outside this PR

**The GitHub issue field.** You were right that the field is unpopulated outside APIM, but the cause is not that people forget to fill it: `customfield_10115` has no context on those projects, so the field does not exist there at all and Exalate has nowhere to write the number back. It also silently drops the value when a ticket is moved between projects — an APIM ticket that travels to another board and back comes home without its number. There is a separate runbook for the Jira-side fix. **Nothing to do here**: your fallback logic is correct, it simply has nothing to read yet.

**fixVersion on those boards.** No public ticket on any of the seven currently carries one — the version tends to go on the internal ticket instead. So the search will run correctly and return nothing until that convention settles. Also outside this PR, and being taken up separately.

Once the grouping change is in, I have nothing further on the code.

phiz71 (commented):
The grouping reads well, and the rule you settled on — a component takes the ticket whichever board raised it, only `Other` gets narrowed — is the right one. 36 tests green, prettier clean, CI green.

One defect and one ordering change, both decided on our side, details inline.

The short version: `Portal` currently comes out **twice**, once as an APIM component and once as a board section, because the two share a name. They also describe the same product, so the answer is to merge them rather than rename either. And the section order should put the long-standing component sections first, the new product sections next, and `Other` last — today `Other` sits in the middle.

release/helpers/changelog-helper.mjs
```diff
@@ -1,4 +1,13 @@
 export const ComponentTypes = ['Gateway', 'Management API', 'Console', 'Portal', 'Helm Charts'];
+
+const ProductSections = {
+  ESM: 'Event Stream Management',
+  AIAM: 'AI Agent Management',
+  PORTAL: 'Portal',
```
phiz71: This collides with the `Portal` entry in `ComponentTypes` on line 1: both produce a section headed `**Portal**`, and both get appended to the same string, so the changelog prints the heading twice with different tickets under each.

Reproduced with two tickets — one APIM ticket carrying the `Portal` component, one PORTAL-board ticket with no component:

```
**Portal**

* apim ticket filed under the Portal component [#1](…)

**Portal**

* ticket raised on the Portal board [#2](…)
```

It is not an edge case: **86** public APIM tickets carrying a fixVersion have the `Portal` component, and every PORTAL-board ticket has no component at all, so any release holding one of each hits it.

**Decision on our side: merge the two, don't rename.** The APIM `Portal` component and the PORTAL board describe the same product, so two sections would be misleading even with distinct titles.

Merging is not just deleting this line, though — drop it and a PORTAL ticket with no component falls into `Other`, which is worse. It needs a board-to-component mapping that `Other` also knows about, something like:

```js
// Boards whose tickets belong to a component section that already exists rather than to one of their
// own: the APIM 'Portal' component and the Portal board are the same product.
const BoardComponents = { PORTAL: 'Portal' };
```

Then in `getTicketsFor`, a named component also takes the tickets of a board mapped to it that carry no component of their own, and `Other` excludes those tickets the same way it already excludes the ones with a product section.

Worth checking while you are there that a PORTAL ticket which *does* carry a component — `Gateway`, say — still lands only under `Gateway` and is not picked up a second time by the mapping. The existing `carriesNoComponent` guard should cover it, but it deserves a test.

release/ci-steps/generate-changelog.mjs
```diff
@@ -29,17 +29,21 @@ echo(chalk.blue(`# Clone ${docRepository} repository`));
...
 
 let changelogPatchTemplate = `
 ## Gravitee API Management ${releasingVersion} - ${new Date().toLocaleDateString('en-US', dateOptions)}
 `;
 
+const productKeys = productKeysOf(issues);
+
 ChangelogSections.forEach((section) => {
   let changelogSection = '';
   [...ComponentTypes, 'Other'].forEach((componentType) => {
```
phiz71: `Other` is emitted inside this loop, so it lands between the component sections and the product sections that follow on line 55 — the catch-all ends up in the middle of the list.

**Decision on our side: keep the long-standing sections at the top, and move `Other` to the end.** So: component sections first, in their current order; then the product sections; then `Other` last.

That means taking `'Other'` out of the spread here and emitting it explicitly after the `productKeys` loop.

One detail worth getting right: `Other` has to be *computed* after the merge above is in place, not just *printed* later — its filter is what decides which tickets nobody else claimed. Printing it last while it still catches PORTAL tickets would move the problem rather than fix it.

release/helpers/changelog-helper.mjs
vikrantgravitee: Fixed in 4832ceafe0 — PORTAL leaves ProductSections for a new BoardComponents map, so its componentless tickets merge into the existing Portal component section and a single `**Portal**` heading renders. Pinned by a direct test plus a structural invariant keeping the two heading namespaces disjoint.

release/ci-steps/generate-changelog.mjs
vikrantgravitee: Fixed in 4832ceafe0 — section assembly moved into changelog-helper as getTicketSections, rendering components, then product boards, then Other last. The exact heading sequence is pinned by an ordering test and an exhaustive board × component × type placement test.

release/helpers/jira-helper.mjs
vikrantgravitee: Addressed — an error shape on the remote-link lookup now warns per ticket (`Could not read the remote links of X`), so a 401/429 surfaces instead of being absorbed. The ordinary no-links case still answers with an empty list and stays silent.

vikrantgravitee:
The `project` field is back and drives the grouping: product boards get their own sections, and the Portal board merges into the Portal component section so no heading prints twice (4832ceafe0). Noted on the GitHub-issue field and fixVersion — agreed, nothing for this script to do there.

phiz71 (approved):
Both points are in, and I checked the rendering rather than just the diff. On a set holding an APIM ticket with the `Portal` component, a Portal-board ticket with none, a Portal-board ticket tagged `Gateway`, plus OBS, ESM, FOUND and a componentless APIM ticket, the headings come out as:

```
**Gateway** | **Portal** | **Event Stream Management** | **Observability** | **Other**
```

One `Portal` section holding both tickets, the long-standing component sections first, the product sections next, `Other` last — and every ticket listed exactly once, including the Portal-board one carrying `Gateway`, which stays under `Gateway` alone. 52 tests green, prettier clean, CI green.

Worth calling out what you did beyond the ask: `keeps the heading namespaces disjoint` pins the *class* of defect, not the case I reported. A board named like a component, or a board landing in both maps, now fails the build instead of quietly printing a heading twice. Pulling the section assembly into `getTicketSections` also turned the display order into one readable list rather than two loops and a special case in the orchestration script. Both are better than what I suggested.

Two things stay open outside this PR, so they don't get lost when it merges:

- **GitHub links on the new boards.** `customfield_10115` has no context on those projects, so Exalate has nowhere to write the issue number back, and the same gap silently erases the value when a ticket moves between projects. Until that is fixed in Jira, non-APIM entries will render without a link — your fallback is correct, it just has nothing to read. A runbook exists for the Jira-side work.
- **fixVersion on those boards.** No public ticket on any of them carries one today, so the search will run correctly and gather nothing until that convention settles with the teams.

Neither is yours to fix, and neither blocks this. Nice work on the iterations.

release/helpers/jira-helper.mjs
phiz71: Closing this one out as it stands: leaving the single all-or-nothing search as it is. Reading a project's versions and browsing its issues need much the same permission, so the case where the per-board tolerance gets undone one step later is unlikely rather than impossible — and if it ever does bite, the job now fails loudly instead of writing an empty changelog, which is the outcome that mattered. Noting it here so the reasoning is on the record rather than lost.

PR #19212, opened 2026-08-18
fix(gateway): use stable route name for OpenTelemetry transaction.name

dev-marek:
**Issue**

https://gravitee.atlassian.net/browse/APIM-14819

**Purpose**

Stop the Gateway's OpenTelemetry root span (transaction.name in Elastic APM / Jaeger) from including the concrete request path and query string, which was producing one distinct transaction name per unique ID/query combination and triggering high-cardinality warnings.

**Summary of changes**

`DefaultHttpRequestDispatcher` now sets the Vert.x context local `"VertxRoute"` to the API's stable acceptor path (`httpAcceptor.path()`) before the root span starts, for both the V4 and V3 request-handling branches. `gravitee-node`'s `RouteGetter` already prefers this context local over the raw request URI when naming the span — it just was never populated. No `gravitee-node` change required.

**Out of scope**

The "not found" branch (no acceptor resolved) still falls back to the raw URI — there's no real route to report there by definition.

**Testing**

- Unit: two new tests in `DefaultHttpRequestDispatcherTest` (V4 and V3 paths) assert the `"VertxRoute"` context local passed to `startRootSpanFrom` equals the API's configured path. Written RED-first, confirmed failing before the fix, passing after. Full module suite green (232/232).
- Manual/docker: built this branch's Gateway into a local Docker image, ran it with Jaeger via `docker/quick-setup/opentelemetry-jaeger`, deployed a v4 HTTP proxy API with a keyless plan and tracing enabled, and called it 10+ times with different resource IDs and query strings. Every trace produced the identical span name (the API's route), with no query string — confirmed via the Jaeger HTTP API against each trace's raw `url.path` attribute, which was genuinely different every time.

**Additional context**

No screenshot/video included — this is an internal tracing behavior fix with no UI surface; verification was done via the Jaeger HTTP API (see Testing). Happy to capture a Jaeger UI screenshot as a follow-up if useful.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
@@ -254,7 +254,15 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
     }
 
     private void markTracingRoute(final Context vertxContext, final String route) {
-        vertxContext.putLocal(TRACING_ROUTE_CONTEXT_KEY, route);
+        // Vert.x 4 exposes String-keyed Context#putLocal directly; on Vert.x 5 (master, 4.12.x) it moved to
+        // io.vertx.core.internal.ContextInternal and requires a cast, so this line needs adjusting on cherry-pick.
+        vertxContext.putLocal(TRACING_ROUTE_CONTEXT_KEY, withoutTrailingSlash(route));
+    }
+
+    // httpAcceptor.path() always has a trailing slash; strip it so http.route matches url.path (both for
+    // correlation in APM tooling and so pre-existing transaction.name-based searches keep matching).
+    private static String withoutTrailingSlash(final String path) {
+        return path.length() > 1 && path.endsWith("/") ? path.substring(0, path.length() - 1) : path;
     }
 
     private MutableExecutionContext prepareExecutionContext(final HttpServerRequest httpServerRequest) {
--- a/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
@@ -156,6 +156,9 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
             mutableCtx.tracer(
                 new io.gravitee.gateway.reactive.api.tracing.Tracer(vertxContext, gatewayTracingContext.opentelemetryTracer())
             );
+            // No route matched: bucket all of it under "/" rather than the unbounded raw request URI
+            // (scanner/probe traffic on internet-facing gateways is otherwise the largest source of cardinality).
+            markTracingRoute(vertxContext, "/");
             ProcessorChain preProcessorChain = platformProcessorChainFactory.preProcessorChain();
             List<ProcessorHook> processHooks = gatewayTracingContext.isVerbose() ? List.of(tracingHook) : List.of();
             Completable handleNotFoundCompletable = HookHelper.hook(
--- a/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
@@ -83,6 +83,8 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
 
     private static final String ATTR_INTERNAL_VERTX_TIMER_ID = ContextAttributes.ATTR_PREFIX + "vertx-timer-id";
     public static final String ATTR_ENTRYPOINT = ContextAttributes.ATTR_PREFIX + "entrypoint";
+    // Key checked by gravitee-node's RouteGetter (OTel span naming) before falling back to the raw request URI.
+    private static final String TRACING_ROUTE_CONTEXT_KEY = "VertxRoute";
     private final GatewayConfiguration gatewayConfiguration;
     private final HttpAcceptorResolver httpAcceptorResolver;
     private final IdGenerator idGenerator;
@@ -190,6 +192,7 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
             log.debug("Request routed to API reactor on path [{}]", httpAcceptor.path());
             MutableExecutionContext mutableCtx = prepareExecutionContext(httpServerRequest);
             mutableCtx.request().contextPath(httpAcceptor.path());
+            markTracingRoute(vertxContext, httpAcceptor.path());
             TracingContext tracingContext = apiReactor.tracingContext();
             mutableCtx.tracer(new io.gravitee.gateway.reactive.api.tracing.Tracer(vertxContext, tracingContext.opentelemetryTracer()));
             mutableCtx.setInternalAttribute(InternalContextAttributes.ATTR_INTERNAL_REACTABLE_API, apiReactor.api());
@@ -250,6 +253,10 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         return handleV3Request(httpServerRequest, httpAcceptor, vertxContext);
     }
 
+    private void markTracingRoute(final Context vertxContext, final String route) {
+        vertxContext.putLocal(TRACING_ROUTE_CONTEXT_KEY, route);
+    }
+
     private MutableExecutionContext prepareExecutionContext(final HttpServerRequest httpServerRequest) {
         VertxHttpServerRequest request = new VertxHttpServerRequest(
             httpServerRequest,
@@ -288,6 +295,7 @@ public class DefaultHttpRequestDispatcher implements HttpRequestDispatcher {
         final Context vertxContext
     ) {
         final ReactorHandler reactorHandler = handlerEntrypoint.reactor();
+        markTracingRoute(vertxContext, handlerEntrypoint.path());
 
         io.gravitee.gateway.http.vertx.VertxHttpServerRequest request = createV3Request(httpServerRequest, idGenerator);
 
--- a/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/test/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcherTest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/test/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcherTest.java
@@ -96,6 +96,8 @@ class DefaultHttpRequestDispatcherTest {
 
     protected static final String HOST = "gravitee.io";
     protected static final String PATH = "/path";
+    // AbstractHttpAcceptor always appends a trailing slash; the real acceptor never returns PATH as-is.
+    protected static final String ACCEPTOR_PATH = "/path/";
     protected static final String MOCK_ERROR_MESSAGE = "Mock error";
     public static final String SERVER_ID = null;
 
@@ -242,7 +244,7 @@ class DefaultHttpRequestDispatcherTest {
         @BeforeEach
         public void prepareV4EmulationMock() {
             when(httpAcceptorResolver.resolve(HOST, PATH, SERVER_ID)).thenReturn(handlerEntrypoint);
-            when(handlerEntrypoint.path()).thenReturn(PATH);
+            when(handlerEntrypoint.path()).thenReturn(ACCEPTOR_PATH);
             when(handlerEntrypoint.reactor()).thenReturn(apiReactor);
             when(apiReactor.tracingContext()).thenReturn(tracingContext);
         }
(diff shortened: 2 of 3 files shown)


jaroslawlakomy-grav (approved):

phiz71 (commented):
Mechanism checked end to end and it holds up: `vertxContext` is a per-request duplicated context (`VertxContext.createNewDuplicatedContext` in `dispatch`, one call per request from `HttpProtocolVerticle.dispatchRequest`), so there is no cross-request leakage of the local; `OpenTelemetrySpan` keeps that very context and `VertxHttpInstrumenterTracer.endSpan` reads it back through `RouteGetter`, whose key and fallback order (`UrlPathTemplate` -> `VertxRoute` -> `uri()`) match gravitee-node 7.27.0. `putLocal(key, null)` is also safe (Vert.x treats it as a remove). Two inline remarks, plus one more general point below.

**The not-found branch keeps the unbounded URI set**

When no acceptor is resolved the route local is never set, so `RouteGetter` falls back to `observableHttpRequest.uri()` — path *and* query string. That is precisely the traffic class whose URI set is genuinely unbounded: scanners hitting `/wp-admin/8f3c...`, `/.env?x=1`, `/api/v1/<random-uuid>`. On an internet-facing gateway this is usually the largest source of distinct URIs, so the Elastic APM high-cardinality warning described in APIM-14819 can keep firing after the fix. The PR declares it out of scope because "there is no real route to report" — but a constant sentinel would bound the set, which is the point of the change: `handleNotFound` already assigns `/` to `ctx.request().contextPath()`, and reusing that (or a fixed `HTTP <method>` name) would cost one line.

Nothing blocking from my side beyond the Vert.x 5 point, which will bite on the labelled forward-ports.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -250,6 +253,10 @@ public Completable dispatch(HttpServerRequest httpServerRequest, String serverId
         return handleV3Request(httpServerRequest, httpAcceptor, vertxContext);
     }
 
+    private void markTracingRoute(final Context vertxContext, final String route) {
+        vertxContext.putLocal(TRACING_ROUTE_CONTEXT_KEY, route);
```
phiz71: `Context.putLocal(String, Object)` only exists in Vert.x 4. This PR carries the `apply-on-master` and `apply-on-4-12-x` labels, and both of those branches pin `<vertx.version>5.0.12</vertx.version>`, where `io.vertx.core.Context` exposes only the `ContextLocal<T>`-keyed overloads — the `String`-keyed `getLocal`/`putLocal` moved to `io.vertx.core.internal.ContextInternal` (checked with `javap` on `vertx-core-5.0.12.jar`).

I replayed the cherry-pick of 6c62119 onto `origin/master` and `origin/4.12.x`: both conflict, and once the conflict is resolved this line does not compile, nor does `vertxContextCaptor.getValue().<String>getLocal("VertxRoute")` in the test. Both Mergify forward-port PRs will land on you with a red build.

Writing it as `((ContextInternal) vertxContext).putLocal(TRACING_ROUTE_CONTEXT_KEY, route)` — which is exactly how gravitee-node's own `RouteGetter` reads it on master — compiles on 4.9.x/4.10.x/4.11.x and survives the forward-port unchanged.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -190,6 +192,7 @@ public Completable dispatch(HttpServerRequest httpServerRequest, String serverId
             log.debug("Request routed to API reactor on path [{}]", httpAcceptor.path());
             MutableExecutionContext mutableCtx = prepareExecutionContext(httpServerRequest);
             mutableCtx.request().contextPath(httpAcceptor.path());
+            markTracingRoute(vertxContext, httpAcceptor.path());
```
phiz71, line 198: `httpAcceptor.path()` always ends with a `/` — `AbstractHttpAcceptor` appends one in its constructor — which is why the integration test had to move from `POST /test` to `POST /test/`.

So for an API declared on context-path `/test`, a call to `POST /test` now reports `http.route="/test/"` while `url.path="/test"`: the route is not a prefix of the path it is supposed to describe. Two consequences: backends that correlate route and path (Elastic APM endpoint grouping, service maps) see an inconsistent pair, and any saved search or alert keyed on `transaction.name: "POST /test"` silently stops matching after the upgrade — a trailing slash is easy to miss when the dashboard just goes empty.

`AbstractHttpAcceptor` already computes `pathWithoutTrailingSlash` for matching purposes, but it is a protected field rather than part of the `HttpAcceptor` interface. Either expose it, or normalise here (strip the trailing `/` unless the path is `/`).

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
dev-marek: Confirmed independently via bytecode inspection (javap on both jars): Vert.x 4.5.27's `io.vertx.core.Context` exposes `getLocal(Object)`/`putLocal(Object, Object)` directly (no cast needed), while Vert.x 5.0.12's public `Context` only has the `ContextLocal<T>`-typed overloads, moving the Object-keyed ones behind `io.vertx.core.internal.ContextInternal`.

One correction though: that exact cast (`(ContextInternal) vertxContext` with `io.vertx.core.internal.ContextInternal`) won't compile on *this* branch either — Vert.x 4.5.27 only has `ContextInternal` at `io.vertx.core.impl.ContextInternal` (different package), and that type isn't needed here since `Context` already exposes the method directly. So there's no single cast that survives both Vert.x major versions unchanged.

Left 4.9.x as-is (it's correct and tested for its own Vert.x version) and added a comment on the `putLocal` call flagging exactly this for whoever resolves the Mergify cherry-pick conflict on master/4.12.x, so they don't have to rediscover the package split. Pushed in a2ead59591.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
dev-marek: Good catch, and agreed on the risk to existing saved searches. Went with normalizing here rather than exposing `pathWithoutTrailingSlash` on the interface — stripped the trailing slash in `markTracingRoute` (unless the path is exactly `/`), so `http.route` now matches `url.path` again. Reran both integration tests with the original (no-trailing-slash) expectations restored — green. Pushed in a2ead59591.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
phiz71: You're right and I was wrong: `io.vertx.core.internal.ContextInternal` doesn't exist on 4.5.27 — the type sits at `io.vertx.core.impl.ContextInternal` there (confirmed with `unzip -l` on both jars: `io/vertx/core/impl/ContextInternal.class` in 4.5.27, `io/vertx/core/internal/ContextInternal.class` in 5.0.12). So there is indeed no single source form that compiles across the two majors, and the cast would have bought nothing here since `Context` already exposes the method.

Agreed with the resolution. The residual risk is identical either way, and the comment puts it exactly where whoever resolves the conflict will read it. Worth noting that a human is in the loop by construction anyway: I replayed the cherry-pick of the first commit onto `origin/master` and `origin/4.12.x` and both conflict on this file, so neither forward-port will land unattended.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
phiz71: Confirmed on `a2ead59`. The `path.length() > 1` guard keeps `/` intact, so an API mounted on the root still reports `http.route = "/"` and `RouteGetter` accepts it (it tests `length() >= 1`). `ctx.request().contextPath()` is left on the trailing-slash form, so the routing semantics are untouched.

Good signal on the shape of the change: the integration test file is now byte-identical to the base branch, and the net diff against `origin/4.9.x` is 2 files / 48 lines with zero deletions.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/test/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcherTest.java
```diff
@@ -283,6 +284,17 @@ void shouldInitiateTracingV4EmulationRequest() {
...
 
+        @Test
+        void shouldUseStableRouteAsTracingSpanNameOnV4EmulationRequest() {
+            when(apiReactor.handle(any(MutableExecutionContext.class))).thenReturn(Completable.complete());
+            ArgumentCaptor<Context> vertxContextCaptor = ArgumentCaptor.forClass(Context.class);
+
+            cut.dispatch(rxRequest, SERVER_ID).test().assertComplete();
+
+            verify(spyNoopTracer).startRootSpanFrom(vertxContextCaptor.capture(), any());
+            assertThat(vertxContextCaptor.getValue().<String>getLocal("VertxRoute")).isEqualTo(PATH);
```
phiz71, line 297: Non-blocking, but worth a one-character change: `PATH` is `"/path"`, and both new tests stub `handlerEntrypoint.path()` with it — a value the production acceptor can never return, since `AbstractHttpAcceptor` always appends a trailing slash in its constructor.

The consequence is that `withoutTrailingSlash` has no unit coverage at all: break it and these two tests stay green, with only the Jaeger integration test standing between the regression and a release. Stubbing `path()` with `"/path/"` while keeping the assertion on `"/path"` would both pin the normalisation here and make the mock reflect what the acceptor actually produces.

phiz71 (commented):
Re-reviewed `a2ead59`. The trailing-slash normalisation is right, and I've conceded the `ContextInternal` point in its thread — my suggested cast doesn't compile on 4.9.x either. One remark left inline on the tests, plus two things below.

**The not-found branch is still open**

No reply on that one yet. Unresolved requests still get their span named from `observableHttpRequest.uri()`, query string included, and that is the traffic class whose URI set is genuinely unbounded. Fine to punt it to a follow-up ticket — I'd just rather it be an explicit decision than an oversight, since it's the same APM warning that opened APIM-14819.

**Branch propagation — one thing to confirm**

A heads-up on sequencing: `master` and `4.12.x` are the two Vert.x 5 targets, and both cherry-picks conflict on `DefaultHttpRequestDispatcher.java` (I replayed them locally), so those two will need the `putLocal` adjustment applied by hand on top of the conflict resolution.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/test/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcherTest.java
dev-marek: Right, that's a real gap — thanks. Stubbed `handlerEntrypoint.path()` with `"/path/"` (a new `ACCEPTOR_PATH` constant, kept separate from `PATH` since that's also used for the incoming request path elsewhere in the class) and left the assertion on the stripped `"/path"`. Mutation-tested it locally by reverting `withoutTrailingSlash` temporarily — both tests fail as expected, confirmed restoring it makes them pass again. Pushed in e9bd585df1.

dev-marek:
**Not-found branch**: fixed in 46cfc9f8ec. This was an explicit call (not an oversight) — went with fixing it in this PR rather than a follow-up ticket, since it's the same mechanism/file/root-cause as the two branches already fixed, and you're right that it's likely the largest cardinality contributor on an internet-facing gateway. Bucketed under `"/"` (reusing the same sentinel `handleNotFound` already assigns to `contextPath`), mutation-tested (temporarily removed the `markTracingRoute` call, confirmed the new test fails, restored, confirmed green).

**Branch propagation**: acknowledged, no further action needed from me right now — the code comment on the `putLocal` call already flags this for whoever resolves the `master`/`4.12.x` cherry-pick conflicts, and it now also covers the not-found branch's `markTracingRoute` call.

phiz71 (approved):
All three points from the review are resolved, and the full pipeline is green on `46cfc9f8` — including `Test gateway` and `Integration tests`.

Re-checked the final state rather than just the deltas:

- **Trailing slash** — `withoutTrailingSlash` keeps `/` intact through the `length() > 1` guard, so a root-mounted API still reports `http.route = "/"` and `RouteGetter` accepts it. `ctx.request().contextPath()` is deliberately left on the trailing-slash form, so routing semantics are untouched.
- **Test coverage** — splitting `ACCEPTOR_PATH` from `PATH` is the right call, since `PATH` also stands in for the incoming request path elsewhere in the class. The normalisation is now genuinely exercised, and no other test in the class asserts on `contextPath`, so moving the stub to the trailing-slash form has no collateral.
- **Not-found branch** — fixing it here rather than in a follow-up was the better call: same file, same mechanism, same root cause, and it's the branch that actually carries the unbounded URI set.

Two things worth recording for later, neither of them blocking:

Unmatched requests and an API mounted on context-path `/` now share the same `transaction.name` (`GET /`), separable only by status code. Rare and largely mutually exclusive in practice — an API at the root leaves almost nothing to fall through — so the trade is worth it. Strictly, the OTel convention wants `http.route` *absent* when nothing matched rather than a placeholder, but `RouteGetter` falls back to `uri()` as soon as the local is empty, so that isn't a choice the dispatcher gets to make.

And the comment on the `putLocal` call does its job for the two Vert.x 5 forward-ports: both cherry-picks conflict on this file anyway, so whoever resolves them will read it at exactly the right moment.

Net diff is 2 files, 67 insertions for 1 deletion, with the integration test back to identical with the base branch. Clean change, well tested. LGTM.

PR #19215, opened 2026-08-18
feat(automation): api-referenced links automation

MateuszJozwiak-Incubly:
## Issue

### What this branch does

This branch introduces the core building blocks for **API-referenced** portal navigation links: pages, folders and links that belong to a specific API rather than to a fixed spot in a portal's navigation tree. Instead of an operator manually placing a link under one portal, it is declared once against the API, and Gravitee takes care of showing it everywhere that API is listed.

### Core use cases

Four new use cases give the Automation API a way to manage an API's own link:

- **Create or update** it
- **Read** it back
- **Validate** it
- **Delete** it

These sit at the domain layer and encapsulate the rules around what a valid API-referenced link looks like (e.g. how it resolves an optional location within the API's own subtree).

### Materialization / sync pipeline

The domain services responsible for turning declared automation config into actual stored navigation items were extended so that an API-referenced link is materialized correctly: as a **root item tagged with a reference back to its owning API**, rather than as an item physically parented somewhere in a portal's tree.

The reconciliation and sync services — subtree reconciliation, documentation sync, listing sync, and the query service that finds automation-managed items — were adjusted to understand and preserve this new kind of reference.

### Rendering: splicing the API's subtree into portal navigation

The most significant behavioural change is in how the portal navigation tree is assembled for display.

Previously, an API-referenced link/page would either be invisible or would incorrectly leak out at the very top level of a portal's navigation, because the query that fetches "root" items had no way to distinguish a genuine portal-level root from an API's own subtree root.

Now, whenever the navigation tree walk reaches a row that represents an API being listed in a portal, the API's own materialized subtree is looked up by its API reference and **spliced in underneath that row** — with a freshly rendered copy of each item whose parent is rewritten to point at that row. The original stored item is never mutated, so the same API subtree renders correctly and identically no matter how many different portals list that API. At the same time, those subtree roots are explicitly excluded from ever appearing as standalone top-level items, closing the leak described above.

Visibility was extended to match: if the API itself is hidden from a given viewer, its subtree is treated as hidden too — even if someone tries to reach it directly by its own item id rather than through the portal row that lists the API. This closes a gap where a client could otherwise bypass the "can this viewer see this API" check by asking for the subtree root directly.

### Outcome

Together, these changes let an API's links, pages and folders be declared once and consistently surfaced wherever the API appears across all portals, with correct visibility and no duplication or leakage — laying the groundwork the follow-up branch (`part2`) builds on to expose this through the Automation REST API and OpenAPI contract.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/ApiDocumentationSyncDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/ApiDocumentationSyncDomainService.java
@@ -75,20 +75,25 @@ public class ApiDocumentationSyncDomainService {
         var meta = pageContent.getAutomationMetadata();
         var apiId = meta.referenceId();
         var contentId = pageContent.getId();
-        for (var navApi : findNavApiRows(auditInfo.environmentId(), apiId)) {
-            var pageId = PortalNavigationItemId.forApiDocumentation(auditInfo, navApi.getId(), contentId);
+        var navApiRows = findNavApiRows(auditInfo.environmentId(), apiId);
+        if (navApiRows.isEmpty()) {
+            return;
+        }
+        for (var navApi : navApiRows) {
+            var pageId = PortalNavigationItemId.forApiDocumentation(auditInfo, apiId, contentId);
             var parent = resolveParent(auditInfo, navApi, meta.location().orElse(null));
             upsertNavPage(auditInfo, pageId, contentId, parent, meta);
         }
     }
 
     public void dematerialize(AuditInfo auditInfo, String apiId, PortalPageContentId contentId) {
-        for (var navApi : findNavApiRows(auditInfo.environmentId(), apiId)) {
-            var pageId = PortalNavigationItemId.forApiDocumentation(auditInfo, navApi.getId(), contentId);
-            var existing = navigationItemsQueryService.findByIdAndEnvironmentId(auditInfo.environmentId(), pageId);
-            if (existing != null) {
-                navigationItemCrudService.delete(pageId);
-            }
+        if (findNavApiRows(auditInfo.environmentId(), apiId).isEmpty()) {
+            return;
+        }
+        var pageId = PortalNavigationItemId.forApiDocumentation(auditInfo, apiId, contentId);
+        var existing = navigationItemsQueryService.findByIdAndEnvironmentId(auditInfo.environmentId(), pageId);
+        if (existing != null) {
+            navigationItemCrudService.delete(pageId);
         }
     }
 
@@ -119,7 +124,7 @@ public class ApiDocumentationSyncDomainService {
         portalPageContentQueryService
             .findByReference(auditInfo.environmentId(), AutomationMetadata.ReferenceType.API, apiId)
             .forEach(pc -> {
-                var pageId = PortalNavigationItemId.forApiDocumentation(auditInfo, navApiId, pc.getId());
+                var pageId = PortalNavigationItemId.forApiDocumentation(auditInfo, apiId, pc.getId());
                 if (navigationItemsQueryService.findByIdAndEnvironmentId(auditInfo.environmentId(), pageId) != null) {
                     navigationItemCrudService.delete(pageId);
                 }
@@ -148,7 +153,7 @@ public class ApiDocumentationSyncDomainService {
         if (location == null || location.isBlank() || "/".equals(location)) {
             return navApi;
         }
-        var folderId = PortalNavigationItemId.forApiFolder(auditInfo, navApi.getId(), location);
+        var folderId = PortalNavigationItemId.forApiFolder(auditInfo, navApi.getApiId(), location);
         var existing = navigationItemsQueryService.findByIdAndEnvironmentId(auditInfo.environmentId(), folderId);
         if (existing instanceof PortalNavigationItemContainer container) {
             return container;
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/CreateOrUpdateApiLinkUseCase.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/CreateOrUpdateApiLinkUseCase.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.portal_page.use_case;
+
+import io.gravitee.apim.core.UseCase;
+import io.gravitee.apim.core.audit.model.AuditInfo;
+import io.gravitee.apim.core.exception.AbstractDomainException;
+import io.gravitee.apim.core.portal.validation.NavigationPathValidator;
+import io.gravitee.apim.core.portal_page.domain_service.PortalLinkSyncDomainService;
+import io.gravitee.apim.core.portal_page.domain_service.PortalNavigationItemValidatorService;
+import io.gravitee.apim.core.portal_page.model.CreatePortalNavigationItem;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemType;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationLink;
+import io.gravitee.apim.core.portal_page.model.UpdatePortalNavigationItem;
+import io.gravitee.apim.core.portal_page.query_service.PortalNavigationItemsQueryService;
+import io.gravitee.apim.core.validation.Validator;
+import java.util.ArrayList;
+import java.util.List;
+import lombok.RequiredArgsConstructor;
+
+@UseCase
+@RequiredArgsConstructor
+public class CreateOrUpdateApiLinkUseCase {
+
+    private final PortalNavigationItemsQueryService navigationItemsQueryService;
+    private final PortalNavigationItemValidatorService navigationItemValidatorService;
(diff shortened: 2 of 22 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalAutomationScopeDomainService.java
```diff
@@ -34,6 +36,15 @@ public boolean isDefaultPortal(AuditInfo auditInfo, PortalId portalId) {
         return portalCrudService.findByIdAndEnvironmentId(portalId, auditInfo.environmentId()).isPresent();
     }
 
+    /**
+     * The environment's single portal, if one has been applied. Automation supports at most one
+     * portal per environment (see {@link #validate}), so API-attached resources resolve their
+     * placement against this one rather than taking a portal from the request path.
+     */
+    public Optional<PortalId> establishedPortalId(AuditInfo auditInfo) {
+        return portalCrudService.findByEnvironmentId(auditInfo.environmentId()).stream().findFirst().map(Portal::getId);
```
konrad-krol-incubly: I think this is wrong. Currently we cannot create multiple portals per env due to explicit "guard". If you add this method, you are restricting whole logic to be single-portal aware and we won't be able to drop "guard" in the future.

Moreover if you add this, we won't be able to attach API into multiple portals through PortalListing, because this method will always return first portal available.

I think we assumed that navigation item can be associated to API instead of portal, so you don't need portal id to realize "always materialize nav items" requirement.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/CreateOrUpdateApiLinkUseCase.java
```diff
@@ -0,0 +1,125 @@
...
+            errors.addAll(NavigationPathValidator.validate(input.location(), "location"));
+        }
+
+        var linkId = PortalNavigationItemId.forApiLink(input.auditInfo(), input.apiId(), input.linkHrid());
+        var existing = navigationItemsQueryService.findByIdAndEnvironmentId(input.auditInfo().environmentId(), linkId);
+
+        // An API-attached link is placed under the API's row in the environment's portal navigation, so
+        // with no portal applied yet there is no tree to place it in — and, unlike a Documentation page,
+        // a link has no content row that could carry the intent in the meantime. Fail loudly instead.
+        var navApiRowId = portalAutomationScopeEnforcer
```
konrad-krol-incubly: The idea here was that nav item can be attached to API or PORTAL. This one should be attached to API, so we should blindly create nav item for this resource.
As soon as renderer starts to render navigation tree for given portal, it will discover navigation item for API (thanks to portal listing) and should jump to rendering navigation items that are associated to that API (through referenceType=API referenceId={API_ID})

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/PortalAutomationScopeDomainService.java
MateuszJozwiak-Incubly: Reconstructed the logic.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/CreateOrUpdateApiLinkUseCase.java
MateuszJozwiak-Incubly: Reconstructed the logic.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/domain_service/ApiDocumentationSyncDomainServiceTest.java
```diff
@@ -210,12 +209,13 @@ void cleanupNavApi_removes_single_nav_api_and_its_pages() {
 
         syncService.cleanupNavApi(AUDIT_INFO, removedNavApiId);
 
-        var keptPageId = PortalNavigationItemId.forApiDocumentation(AUDIT_INFO, keptNavApiId, DOC_ID);
-        var removedPageId = PortalNavigationItemId.forApiDocumentation(AUDIT_INFO, removedNavApiId, DOC_ID);
+        // The API now has a single page, shared by every nav-api row, so cleaning up one row still takes
+        // it. Part 2 moves subtree ownership off the row and this becomes "the page survives".
```
konrad-krol-incubly: what is part 2?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/GetApiLinkUseCase.java
```diff
@@ -0,0 +1,50 @@
...
+        var item = navigationItemsQueryService.findByIdAndEnvironmentId(input.auditInfo().environmentId(), input.linkId());
+        if (!(item instanceof PortalNavigationLink link) || !isApiAttached(link)) {
+            throw new PortalLinkNotFoundException(input.linkId().toString());
+        }
+        return new Output(link);
+    }
+
+    // Scopes the read to API-attached links so an /apis/{apiHrid}/links path cannot reach a
+    // portal-attached one, mirroring GetApiDocumentationUseCase's referenceType filter.
+    private static boolean isApiAttached(PortalNavigationLink link) {
```
konrad-krol-incubly, line 47: Is it possible that /apis/{apiHrid}/links will reach portal-attached link? ID is deterministic and that would mean we have id collission between two entries -> it will be rejected by DB due to ID uniqueness constraint, right?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/domain_service/ApiDocumentationSyncDomainServiceTest.java
MateuszJozwiak-Incubly: I think the comment is a leftover, can be deleted. Part 2 refers to the next PR that will come after this one.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/GetApiLinkUseCase.java
MateuszJozwiak-Incubly: My checks say that it would be rather unlikely. ID generation should look like this:   
- portal link → hash(orgId + "portal" + portalId) then hash again with env + hrid                                                                                             
- api link → hash(orgId + "api" + apiId) then the same

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/ApiDocumentationSyncDomainService.java
```diff
@@ -76,19 +76,20 @@ public void materialize(AuditInfo auditInfo, PortalPageContent<?> pageContent) {
         var apiId = meta.referenceId();
         var contentId = pageContent.getId();
         for (var navApi : findNavApiRows(auditInfo.environmentId(), apiId)) {
-            var pageId = PortalNavigationItemId.forApiDocumentation(auditInfo, navApi.getId(), contentId);
+            var pageId = PortalNavigationItemId.forApiDocumentation(auditInfo, apiId, contentId);
```
konrad-krol-incubly: Please check following case:
We have old API documentation (before changes) attached to portal.
We bump version of APIM and GKO calls reconciliation on the same resource.
What will happen? Are we going to have duplicated nav items? One attached to portal, one to API?

MateuszJozwiak-Incubly: This scenario will produce a duplicate entry :( This self-heals only if that specific portal's listing later gets separately re-synced (triggering the sweep) — but a version bump + GKO reconciling the API resource does not touch the Portal resource, so nothing forces that to happen. In a real upgrade, this duplicate would be visible indefinitely for any portal listing an API whose CRD gets reconciled before its listing portals are separately.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/ApiDocumentationSyncDomainService.java
```diff
@@ -75,20 +75,25 @@ public void materialize(AuditInfo auditInfo, PortalPageContent<?> pageContent) {
...
-            var pageId = PortalNavigationItemId.forApiDocumentation(auditInfo, navApi.getId(), contentId);
+        var navApiRows = findNavApiRows(auditInfo.environmentId(), apiId);
+        if (navApiRows.isEmpty()) {
+            return;
+        }
+        for (var navApi : navApiRows) {
+            var pageId = PortalNavigationItemId.forApiDocumentation(auditInfo, apiId, contentId);
             var parent = resolveParent(auditInfo, navApi, meta.location().orElse(null));
             upsertNavPage(auditInfo, pageId, contentId, parent, meta);
         }
```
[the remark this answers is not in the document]
MateuszJozwiak-Incubly: Leave it as it is for now.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/ListPortalNavigationItemsUseCase.java
```diff
@@ -186,9 +186,135 @@ private List<PortalNavigationItem> searchItems(
...
+     * lists the API. Both are real; this returns their union.
+     */
+    private List<PortalNavigationItem> childrenOf(
+        PortalNavigationItem parent,
+        Input input,
+        PortalNavigationItemVisibilityEvaluator visibilityEvaluator
+    ) {
+        if (parent instanceof PortalNavigationApi navApi) {
+            var physicalChildren = searchItems(input, navApi.getId(), false, visibilityEvaluator);
+            var splicedRoots = queryService.findTopLevelItemsByEnvironmentIdAndPortalAreaAndReference(
```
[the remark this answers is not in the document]
MateuszJozwiak-Incubly: It is a known issue that this part of code is not optimal. Leave it as it is for now.

dev-marek (approved):

PR #19216, opened 2026-08-18
fix(console): group Top Paths widget on the request uri

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-11063

## Summary

The **Top Paths** widget on the v4 API Traffic screen showed a single `(empty)` row instead of the paths being hit.

It grouped on `path-info`, which is the sub-path *after* the API's context path. A request to the API root — the normal case for a simple proxy API — records `path-info: ""`, so Elasticsearch returned one bucket with an empty key, which the table renders via `{{ element || '(empty)' }}`.

The widget now groups on `uri`, the actual called URI.

## Changes

- `api-analytics-proxy.component.ts` — Top Paths `groupByField`: `path-info.keyword` → `uri`
- `analyticsGroupBy.ts` — same swap in the `GroupByField` union
- `api-analytics-proxy.component.spec.ts` — new spec

With Fix: 
image

Without Fix:
image

## Related

Closes the remaining gap found while verifying [APIM-11063](https://gravitee.atlassian.net/browse/APIM-11063). The ticket's original ask — the URI column in the v4 Logs list — already shipped in `b60ec975a9` and is covered by an existing Jest test.

[APIM-11063]: https://gravitee.atlassian.net/browse/APIM-11063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.spec.ts
+++ b/gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.spec.ts
@@ -637,6 +637,58 @@ describe('ApiAnalyticsWidgetService', () => {
           expectGroupByRequest('application-id', mockGroupByResponse);
         });
 
+        it('should merge uri buckets that differ only by query string and rank by merged count', done => {
+          const widgetConfig: ApiAnalyticsDashboardWidgetConfig = {
+            type: 'table',
+            apiId: API_ID,
+            title: 'Top Paths',
+            tooltip: 'Most frequently hit API paths',
+            analyticsType: 'GROUP_BY',
+            groupByField: 'uri',
+            mergeGroupByUriPath: true,
+            shouldSortBuckets: false,
+            orderBy: '-count:_count',
+            tableData: {
+              columns: [
+                { label: 'Path', dataType: 'string' },
+                { label: 'Hits', dataType: 'number' },
+              ],
+            },
+          };
+
+          const mockGroupByResponse: GroupByResponse = fakeGroupByResponse({
+            values: {
+              '/health': 50,
+              '/orders?id=1': 40,
+              '/orders?id=2': 39,
+              '/orders?id=3': 221,
+            },
+            metadata: {
+              '/health': { name: '/health', order: 0 },
+              '/orders?id=1': { name: '/orders?id=1', order: 1 },
+              '/orders?id=2': { name: '/orders?id=2', order: 2 },
+              '/orders?id=3': { name: '/orders?id=3', order: 3 },
+            },
+          });
+
+          service.getApiAnalyticsWidgetConfig$(widgetConfig).subscribe(result => {
+            if (result.state === 'loading') {
+              return;
+            }
+
+            expect(result.state).toBe('success');
+            expect(result.widgetType).toBe('table');
+            if (result.widgetType === 'table') {
+              expect(result.widgetData.data).toHaveLength(2);
+              expect(result.widgetData.data[0]).toEqual(expect.objectContaining({ 'col-0': '/orders', 'col-1': 300, __key: '/orders' }));
+              expect(result.widgetData.data[1]).toEqual(expect.objectContaining({ 'col-0': '/health', 'col-1': 50, __key: '/health' }));
+            }
+            done();
+          });
+
+          expectGroupByRequest('uri', mockGroupByResponse, { order: '-count:_count' });
+        });
+
         it('should use default column names when no columns are provided for table widget', done => {
           const widgetConfig: ApiAnalyticsDashboardWidgetConfig = {
             type: 'table',
--- a/gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.ts
+++ b/gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.ts
@@ -230,16 +230,50 @@ export class ApiAnalyticsWidgetService {
     groupByResponse: GroupByResponse,
     widgetConfig: ApiAnalyticsDashboardWidgetConfig,
   ): ApiAnalyticsWidgetConfig {
+    const normalizedGroupByResponse = this.normalizeGroupByResponse(groupByResponse, widgetConfig);
+
     if (widgetConfig.type === 'pie') {
-      return this.transformGroupByResponseToPieConfig(groupByResponse, widgetConfig);
+      return this.transformGroupByResponseToPieConfig(normalizedGroupByResponse, widgetConfig);
     }
 
     if (widgetConfig.type === 'table') {
-      return this.transformGroupByResponseToTableConfig(groupByResponse, widgetConfig);
+      return this.transformGroupByResponseToTableConfig(normalizedGroupByResponse, widgetConfig);
     }
 
     // Default fallback for unsupported widget types
     return this.createErrorConfig(widgetConfig, 'Unsupported widget type for GROUP_BY analytics');
+  }
+
+  private normalizeGroupByResponse(groupByResponse: GroupByResponse, widgetConfig: ApiAnalyticsDashboardWidgetConfig): GroupByResponse {
+    // Opt-in per widget so a future uri-based widget can still show raw per-URI buckets.
+    if (!widgetConfig.mergeGroupByUriPath) {
+      return groupByResponse;
+    }
+
+    const values: Record<string, number> = {};
+
+    Object.entries(groupByResponse.values).forEach(([uri, count]) => {
+      const path = this.pathWithoutQueryString(uri);
+      values[path] = (values[path] ?? 0) + count;
+    });
+
+    const metadata = Object.entries(values)
+      .sort(([, countA], [, countB]) => countB - countA)
+      .reduce<GroupByResponse['metadata']>((acc, [path], index) => {
+        acc[path] = { name: path, order: index };
+        return acc;
+      }, {});
+
+    return {
+      ...groupByResponse,
+      values,
+      metadata,
+    };
+  }
+
+  private pathWithoutQueryString(uri: string): string {
+    const queryIndex = uri.indexOf('?');
+    return queryIndex === -1 ? uri : uri.slice(0, queryIndex);
   }
 
   private transformGroupByResponseToPieConfig(
--- a/gravitee-apim-console-webui/src/entities/management-api-v2/analytics/analyticsGroupBy.ts
+++ b/gravitee-apim-console-webui/src/entities/management-api-v2/analytics/analyticsGroupBy.ts
@@ -26,5 +26,5 @@ export type GroupByField =
   | 'status'
   | 'application-id'
   | 'plan-id'
-  | 'path-info.keyword'
+  | 'uri'
   | 'additional-metrics.long_mcp-proxy_response-error-code';
--- a/gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-proxy/api-analytics-proxy.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-proxy/api-analytics-proxy.component.spec.ts
@@ -84,6 +84,19 @@ describe('ApiAnalyticsProxyComponent', () => {
     });
   });
 
+  describe('Top Paths widget', () => {
+    it('should group on the request uri', async () => {
+      await initComponent();
+
+      const groupByRequests = httpTestingController.match(req => req.url.includes('type=GROUP_BY'));
+      const groupByFields = groupByRequests.map(req => new URL(req.request.urlWithParams).searchParams.get('field'));
+      groupByRequests.forEach(req => req.flush(fakeGroupByResponse()));
+      handleAllRequests();
+
+      expect(groupByFields).toContain('uri');
+    });
+  });
+
   describe('Query parameters', () => {
     const plan1 = fakePlanV4({ id: '1', name: 'plan 1' });
     const plan2 = fakePlanV4({ id: '2', name: 'plan 2' });
--- a/gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-proxy/api-analytics-proxy.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-proxy/api-analytics-proxy.component.spec.ts
@@ -84,6 +84,20 @@ describe('ApiAnalyticsProxyComponent', () => {
     });
   });
 
+  describe('Top Paths widget', () => {
+    it('should group Top Paths on the request uri', async () => {
+      await initComponent();
+
+      const groupByRequests = httpTestingController.match(req => req.url.includes('type=GROUP_BY'));
+      const topPathsRequest = groupByRequests.find(req => new URL(req.request.urlWithParams).searchParams.get('field') === 'uri');
+      groupByRequests.forEach(req => req.flush(fakeGroupByResponse()));
+      handleAllRequests();
+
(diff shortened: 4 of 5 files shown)


aqibmohammadkhan (commented):
## Review

Nice fix for the `(empty)` bucket bug — confirmed `uri` is a valid, directly-aggregatable `keyword` field in the ES v4-metrics template (unlike `path-info`, which needs the `.keyword` sub-field), so the mechanical part of the change is sound.

One concern on the semantics, though:

### `uri` includes the query string, so "Top Paths" will under-aggregate for APIs with query params
`api-analytics-proxy.component.ts:302`

The gateway populates the `uri` metric from Vert.x's `HttpServerRequest#uri()`:
- `VertxHttpServerRequest.java`: `uri() { return serverRequest.uri(); }`
- `MetricsProcessor.java`: `.uri(request.uri())`

Vert.x's `uri()` includes the query string (e.g. `/pets?id=1`, not just `/pets`). That's also consistent with `V4MetricsPathMappingTest.java`'s comment that `uri` is "uncapped ... and always at least as long" than `path-info`.

So for any API where the same logical path is hit with varying query params (pagination, IDs, filters, cache-busting tokens — i.e. most real APIs), grouping by `uri` will split what should be one bucket into many single-count buckets. The widget's tooltip ("Most frequently hit API paths") and column label ("Path") both promise path-level aggregation, which this won't reliably deliver anymore.

This doesn't bring back the original `(empty)` bug, and may still be a net improvement for path-only traffic, but it swaps one bug for a subtler one. Worth deciding explicitly:
- Is there a query-string-free path field already available server-side (e.g. `mapped-path`) that would aggregate correctly?
- Or should the fix strip the query string client- or server-side before grouping?

Minor, non-blocking: the new spec test just asserts `uri` appears somewhere among all `GROUP_BY` requests rather than specifically on the Top Paths request — fine today since no other widget in this component uses `uri`, but slightly under-specified if that changes.

ankita-gupta21:
> ## Review
> Nice fix for the `(empty)` bucket bug — confirmed `uri` is a valid, directly-aggregatable `keyword` field in the ES v4-metrics template (unlike `path-info`, which needs the `.keyword` sub-field), so the mechanical part of the change is sound.
> 
> One concern on the semantics, though:
> 
> ### `uri` includes the query string, so "Top Paths" will under-aggregate for APIs with query params
> `api-analytics-proxy.component.ts:302`
> 
> The gateway populates the `uri` metric from Vert.x's `HttpServerRequest#uri()`:
> 
> * `VertxHttpServerRequest.java`: `uri() { return serverRequest.uri(); }`
> * `MetricsProcessor.java`: `.uri(request.uri())`
> 
> Vert.x's `uri()` includes the query string (e.g. `/pets?id=1`, not just `/pets`). That's also consistent with `V4MetricsPathMappingTest.java`'s comment that `uri` is "uncapped ... and always at least as long" than `path-info`.
> 
> So for any API where the same logical path is hit with varying query params (pagination, IDs, filters, cache-busting tokens — i.e. most real APIs), grouping by `uri` will split what should be one bucket into many single-count buckets. The widget's tooltip ("Most frequently hit API paths") and column label ("Path") both promise path-level aggregation, which this won't reliably deliver anymore.
> 
> This doesn't bring back the original `(empty)` bug, and may still be a net improvement for path-only traffic, but it swaps one bug for a subtler one. Worth deciding explicitly:
> 
> * Is there a query-string-free path field already available server-side (e.g. `mapped-path`) that would aggregate correctly?
> * Or should the fix strip the query string client- or server-side before grouping?
> 
> Minor, non-blocking: the new spec test just asserts `uri` appears somewhere among all `GROUP_BY` requests rather than specifically on the Top Paths request — fine today since no other widget in this component uses `uri`, but slightly under-specified if that changes.

You can post this as a reply to the review:

---

Valid!!

**On `uri` vs `mapped-path`**

I checked the available v4 metrics fields:

- `path-info` — sub-path after the context path; empty for root hits (the original `(empty)` bug).
- `uri` — full called path, but includes the query string via `HttpServerRequest#uri()`.
- `mapped-path` — only populated when path mappings are configured, and it stores route templates (e.g. `/pets/:id`), not actual request paths.

So `mapped-path` is not a good fit for “Top Paths” on a general proxy API.


I kept grouping on `uri` (fixes the empty-bucket issue and shows the full called path including context path), and added client-side normalization in `ApiAnalyticsWidgetService` to merge buckets that differ only by query string before rendering the table. For example, `/pets?id=1` and `/pets?id=2` are merged into a single `/pets` row with combined hit counts.

This keeps the fix console-scoped for APIM-11063 without a gateway/reporter change. A dedicated query-string-free field (e.g. indexing `request.path()`) would be cleaner long term, but that is a separate server-side change.

**Tests**

- Component spec now asserts the Top Paths widget specifically uses `field=uri` (not just that `uri` appears among all GROUP_BY requests).
- Widget service spec covers query-string merge behavior (`/pets?id=1` + `/pets?id=2` → `/pets` with 5 hits).

---

gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.ts
```diff
@@ -230,18 +230,53 @@ export class ApiAnalyticsWidgetService {
...
+    }
+
+    const values: Record<string, number> = {};
+    const metadata: GroupByResponse['metadata'] = {};
+
+    Object.entries(groupByResponse.values).forEach(([uri, count]) => {
+      const path = this.pathWithoutQueryString(uri);
+      values[path] = (values[path] ?? 0) + count;
+
+      const existingOrder = metadata[path]?.order;
```
aqibmohammadkhan: **Merged buckets keep a stale rank instead of one based on the merged count.** `normalizeGroupByResponse` sums counts for every `uri` sharing a path, but assigns the merged bucket the *minimum of the original per-URI `order` values* rather than a rank derived from the new summed count. `transformGroupByResponseToTableConfig` (line 337) sorts strictly by this `order`, and `shouldSortBuckets: false` on the Top Paths widget means nothing re-sorts by the actual value afterward.

Concretely: ES returns (sorted desc by count via `orderBy: '-count:_count'`) `/health` count=50 order=0, `/orders?id=1` count=40 order=1, `/orders?id=2` count=39 order=2, plus more `/orders?id=N` variants. After merging, `/orders` totals 300 hits but keeps `order=1` (its best constituent's rank), while `/health` stays at `order=0`. The table ends up showing `/health` (50 hits) above `/orders` (300 hits) — the least-popular path first, which is the exact class of bug this PR sets out to fix, just relocated.

Suggestion: after building the merged `values`, re-derive `order` by sorting the merged entries by count descending and assigning order by that final position, instead of carrying over a pre-merge rank.

gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.spec.ts
```diff
@@ -637,6 +637,59 @@ describe('ApiAnalyticsWidgetService', () => {
...
+          service.getApiAnalyticsWidgetConfig$(widgetConfig).subscribe(result => {
+            if (result.state === 'loading') {
+              return;
+            }
+
+            expect(result.state).toBe('success');
+            expect(result.widgetType).toBe('table');
+            if (result.widgetType === 'table') {
+              expect(result.widgetData.data).toHaveLength(2);
+              expect(result.widgetData.data).toEqual(
```
aqibmohammadkhan: **New regression test doesn't assert row order, so it wouldn't catch the ordering bug above.** This spec only checks membership via `toEqual(expect.arrayContaining([...]))`, and the fixture's two merged totals happen to tie at 5, which masks the defect. Worth asserting the resulting row order (or using distinct, non-tied counts in the fixture) once the ordering fix lands, so a future refactor can't silently reintroduce it.

gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.ts
```diff
@@ -230,18 +230,53 @@ export class ApiAnalyticsWidgetService {
...
-      return this.transformGroupByResponseToTableConfig(groupByResponse, widgetConfig);
+      return this.transformGroupByResponseToTableConfig(normalizedGroupByResponse, widgetConfig);
     }
 
     // Default fallback for unsupported widget types
     return this.createErrorConfig(widgetConfig, 'Unsupported widget type for GROUP_BY analytics');
   }
 
+  private normalizeGroupByResponse(groupByResponse: GroupByResponse, widgetConfig: ApiAnalyticsDashboardWidgetConfig): GroupByResponse {
+    if (widgetConfig.groupByField !== 'uri') {
```
aqibmohammadkhan: **Merge behavior is keyed off the field name, not the widget.** `if (widgetConfig.groupByField !== 'uri') return groupByResponse;` means any widget — present or future — that sets `groupByField: 'uri'` silently inherits this query-string-merge/reorder transform, with no way to opt out. Fine today since Top Paths is the only consumer, but a short comment (or scoping this to the widget rather than the field) would save the next person from a surprise if a second `uri`-based widget is added that wants the raw per-URI breakdown.

gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.ts
```diff
@@ -230,18 +230,53 @@ export class ApiAnalyticsWidgetService {
...
+    Object.entries(groupByResponse.values).forEach(([uri, count]) => {
+      const path = this.pathWithoutQueryString(uri);
+      values[path] = (values[path] ?? 0) + count;
+
+      const existingOrder = metadata[path]?.order;
+      const nextOrder = groupByResponse.metadata[uri]?.order;
+      if (nextOrder != null && (existingOrder == null || nextOrder < existingOrder)) {
+        metadata[path] = { name: path, order: nextOrder };
+      } else if (!metadata[path]) {
+        metadata[path] = { name: path, order: nextOrder ?? 0 };
```
aqibmohammadkhan: **Minor: fallback order can collide with a real order-0 bucket.** `order: nextOrder ?? 0` defaults to `0` when a `values` entry has no matching `metadata` entry, which can tie with (or wrongly outrank) a genuine order-0 path. Low risk given how the backend currently builds this response, but a sentinel like `Number.MAX_SAFE_INTEGER` would be safer than `0` if this path is ever hit.

aqibmohammadkhan (commented):
High-effort pass over the `path-info` → `uri` groupBy swap and the new query-string-merge normalization. The field swap itself is correct and well-tested; found one real ordering bug in the new merge logic plus two smaller follow-ups, all inline below.

ankita-gupta21:
> ## Review
> Nice fix for the `(empty)` bucket bug — confirmed `uri` is a valid, directly-aggregatable `keyword` field in the ES v4-metrics template (unlike `path-info`, which needs the `.keyword` sub-field), so the mechanical part of the change is sound.
> 
> One concern on the semantics, though:
> 
> ### `uri` includes the query string, so "Top Paths" will under-aggregate for APIs with query params
> `api-analytics-proxy.component.ts:302`
> 
> The gateway populates the `uri` metric from Vert.x's `HttpServerRequest#uri()`:
> 
> * `VertxHttpServerRequest.java`: `uri() { return serverRequest.uri(); }`
> * `MetricsProcessor.java`: `.uri(request.uri())`
> 
> Vert.x's `uri()` includes the query string (e.g. `/pets?id=1`, not just `/pets`). That's also consistent with `V4MetricsPathMappingTest.java`'s comment that `uri` is "uncapped ... and always at least as long" than `path-info`.
> 
> So for any API where the same logical path is hit with varying query params (pagination, IDs, filters, cache-busting tokens — i.e. most real APIs), grouping by `uri` will split what should be one bucket into many single-count buckets. The widget's tooltip ("Most frequently hit API paths") and column label ("Path") both promise path-level aggregation, which this won't reliably deliver anymore.
> 
> This doesn't bring back the original `(empty)` bug, and may still be a net improvement for path-only traffic, but it swaps one bug for a subtler one. Worth deciding explicitly:
> 
> * Is there a query-string-free path field already available server-side (e.g. `mapped-path`) that would aggregate correctly?
> * Or should the fix strip the query string client- or server-side before grouping?
> 
> Minor, non-blocking: the new spec test just asserts `uri` appears somewhere among all `GROUP_BY` requests rather than specifically on the Top Paths request — fine today since no other widget in this component uses `uri`, but slightly under-specified if that changes.

Agreed — grouping on raw uri alone would split paths like /pets?id=1 and /pets?id=2 into separate buckets.

Fix: We still group on uri in Elasticsearch (fixes the (empty) bucket from path-info), then merge buckets client-side in ApiAnalyticsWidgetService.normalizeGroupByResponse() by stripping the query string before rendering:

api-analytics-widget.service.ts — normalizeGroupByResponse() + pathWithoutQueryString()
/pets?id=1 (3 hits) + /pets?id=2 (2 hits) → one /pets row with 5 hits
Covered by should merge uri buckets that differ only by query string in api-analytics-widget.service.spec.ts.

ankita-gupta21:
> High-effort pass over the `path-info` → `uri` groupBy swap and the new query-string-merge normalization. The field swap itself is correct and well-tested; found one real ordering bug in the new merge logic plus two smaller follow-ups, all inline below.

Thanks for the thorough second pass — all four inline points are addressed in 86c1856f7c:

Merged buckets are re-ranked by summed hit count (descending), not pre-merge ES order.
The regression test now uses non-tied counts and asserts row order explicitly.
Merge behavior is widget-scoped via mergeGroupByUriPath: true on Top Paths only.
The stale order ?? 0 fallback is gone — order is always derived after merge.

gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.ts
ankita-gupta21: **Fixed**: After merging URIs that share the same path, normalizeGroupByResponse() now sorts merged entries by total count descending and assigns fresh order values (0 = highest hits). So /orders at 300 hits ranks above /health at 50, even when ES returned /health first.

gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.spec.ts
ankita-gupta21: Agreed — expect.arrayContaining with tied counts at 5 wouldn't catch an ordering regression.

Fix: The fixture now mirrors your example (/health = 50, /orders?id=* variants summing to 300) and the spec asserts:

row 0 → /orders, 300 hits
row 1 → /health, 50 hits

gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.ts
ankita-gupta21: Fair point — a future uri widget might want raw per-URI buckets.

Fix: Added an opt-in mergeGroupByUriPath?: boolean on the widget config. Only Top Paths sets mergeGroupByUriPath: true. The service checks that flag (with a short comment) instead of groupByField === 'uri'.

gravitee-apim-console-webui/src/management/api/api-traffic-v4/analytics/api-analytics-widget.service.ts
ankita-gupta21: Valid concern on the old order: nextOrder ?? 0 path.

Fix: That logic is removed entirely. After merge, order is always assigned from the post-merge count sort, so there's no fallback that could collide with a genuine order-0 bucket.

aqibmohammadkhan (approved):

PR #19217, opened 2026-08-18
feat: expose API Product subscription details in Portal API

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-125

## Summary

Expose API Product subscription details through the existing Portal API subscription endpoint.

This extends `GET /portal/environments/{environmentId}/subscriptions/{subscriptionId}` in a backward-compatible way while preserving the existing response for direct API subscriptions.

## Changes

- Add an optional API Product details model to the subscription response.
- Return API Product identity, selected plan and included APIs.
- Group accessible entrypoints by API.
- Provide documentation navigation targets when documentation is accessible.
- Represent missing or unpublished products and APIs with explicit availability states.
- Authorize API Product subscription access through the owning application and current environment.
- Preserve the existing `include=keys` and `include=consumerConfiguration` behavior.
- Keep direct API subscription behavior backward compatible.


gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
```diff
@@ -2585,7 +2585,7 @@ paths:
             description: |
                 Get a subscription.
 
-                User must have API_SUBSCRIPTION[CREATE] or APPLICATION_SUBSCRIPTION[CREATE] permission.
+                User must have API_SUBSCRIPTION[READ] or APPLICATION_SUBSCRIPTION[READ] permission.
```
jaroslawlakomy-grav, line 2588: This is a documentation-only correction. The endpoint already checked `RolePermissionAction.READ` for both API and application subscription permissions before this change. The previous `[CREATE]` wording did not reflect the actual authorization logic; no runtime behavior is changed here.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription/use_case/GetPortalApiProductSubscriptionDetailsUseCase.java
```diff
@@ -0,0 +1,277 @@
...
+        return navigationItems
+            .stream()
+            .filter(item -> isAccessibleFromSubscribedProduct(item, input, navigationAncestorsById))
+            .sorted(Comparator.comparing(PortalNavigationItem::getOrder).thenComparing(item -> item.getId().json()))
+            .findFirst()
+            .map(item -> new DocumentationTarget(item.getRootId().json(), item.getId().json()))
+            .orElse(null);
+    }
+
+    private boolean isAccessibleFromSubscribedProduct(
```
thomasballivet, line 241: This only checks `shouldNotShow()` (published + PRIVATE/anonymous), so a product subscriber gets a documentation deep-link for a PRIVATE API. `PortalNavigationApiVisibilityDomainService` would deny it — its `filterAllowedApiIdsBySubscription` keys on `SubscriptionEntity.getApiId()`, null for product subs. Intentional?

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
```diff
@@ -6264,6 +6264,125 @@ components:
...
+                    description: Security used by the plan, when available.
+                    enum:
+                        - API_KEY
+                        - KEY_LESS
+                        - JWT
+                        - OAUTH2
+                        - MTLS
+                mode:
+                    $ref: "#/components/schemas/PlanMode"
+                usageConfiguration:
```
thomasballivet: `usageConfiguration` is declared here but exists neither in `PortalApiProductSubscriptionDetails.PlanSummary` nor in `ApiProductSubscriptionDetailsMapper`, so it will always be absent. We should either map it or drop it from the contract.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
```diff
@@ -81,29 +89,58 @@ public Response getSubscriptionBySubscriptionId(
     ) {
         SubscriptionEntity subscriptionEntity = subscriptionService.findById(subscriptionId);
         final ExecutionContext executionContext = GraviteeContext.getExecutionContext();
+        boolean apiProductSubscription = "API_PRODUCT".equals(subscriptionEntity.getReferenceType());
```
thomasballivet: Hardcoded **"API_PRODUCT"** → `SubscriptionReferenceType.API_PRODUCT.name().equals(...)`, as in ApiProductSubscriptionResource:394.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription/use_case/GetPortalApiProductSubscriptionDetailsUseCase.java
```diff
@@ -0,0 +1,277 @@
...
+            api.getVersion(),
+            api.getType() == null ? null : api.getType().name(),
+            availability,
+            entrypoints,
+            resolveDocumentationTarget(navigationItems, input, navigationAncestorsById)
+        );
+    }
+
+    private ApiAvailability apiAvailability(Api api) {
+        return switch (api.getApiLifecycleState()) {
```
thomasballivet, line 219: are we sure it can't be null?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription/use_case/GetPortalApiProductSubscriptionDetailsUseCase.java
```diff
@@ -0,0 +1,277 @@
...
+                api.getVersion(),
+                api.getType() == null ? null : api.getType().name(),
+                availability,
+                List.of(),
+                null
+            );
+        }
+
+        var entrypoints = Objects.requireNonNullElse(
+            apiExposedEntrypointDomainService.get(input.organizationId(), input.environmentId(), api),
```
thomasballivet, line 194: One call per API, each redoing the same env/org lookups (ApiEntrypointServiceImpl:82-102) — ~35 identical reads for a 10-API product. And it runs on every GET, unlike keys/consumerConfiguration. Hoist the lookups, or gate behind include=apiProduct. WDYT?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription/use_case/GetPortalApiProductSubscriptionDetailsUseCase.java
jaroslawlakomy-grav: Aligned the documentation target with the shared Portal navigation visibility rules. The prepared visibility evaluator is now applied while traversing the API Product ancestor chain, so a target is returned only when the downstream navigation endpoints can expose it. Product-subscription-based access to private APIs remains a separate cross-cutting visibility concern because the current API visibility service only considers direct API subscriptions.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/resources/portal-openapi.yaml
jaroslawlakomy-grav: Removed usageConfiguration from ApiProductSubscriptionPlan. API Product plan usage data is not currently exposed by the core model or mapper, so retaining this property would advertise a value that cannot be returned. Rate and quota exposure can be added once product-plan usage resolution is supported.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
jaroslawlakomy-grav: Replaced the hardcoded value with SubscriptionReferenceType.API_PRODUCT.name().

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription/use_case/GetPortalApiProductSubscriptionDetailsUseCase.java
jaroslawlakomy-grav: Added a null guard and mapped an unknown API lifecycle to UNAVAILABLE, with regression coverage.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/subscription/use_case/GetPortalApiProductSubscriptionDetailsUseCase.java
jaroslawlakomy-grav: Made API Product enrichment opt-in through include=apiProduct, consistently with the existing keys and consumerConfiguration expansions. The OpenAPI contract and resource tests were updated, so entrypoint and documentation resolution no longer runs for every subscription details request.

thomasballivet (approved):

PR #19218, opened 2026-08-18
fix(ci): build and scan the chainguard images on a branch push, not on the nightly

phiz71:
## Issue

https://gravitee.atlassian.net/browse/BX-292

## Description

The development environment runs the chainguard images, and #19087 moved their build to the nightly. A merge stopped refreshing them.

Nothing failed. `Deploy on Azure cluster` is a `kubectl rollout restart` on three namespaces — it never names a tag, so it cannot report a missing image. The pods came back up on whatever `master-latest-chainguard` pointed at before, while the standard variants kept moving.

The claim that the environment did not run them was mine, and it was wrong. It came from reading the `requires` of the deployment job, which listed only the four standard images. That list says what the deployment *waits for*, not what the cluster *pulls*, and the two had drifted apart. The same reasoning is now written down in the code, next to the list, so the next reader does not repeat it.

**What changes**

- **The chainguard images are built on every branch push again**, and the nightly stops building them. The branch pipeline goes from 17 jobs to 22, the nightly from 27 to 22.
- **`Deploy on Azure cluster` waits for the five images the cluster pulls** — the four components plus Gamma Console. Building them again is not enough on its own: without this, the rollout races the push it depends on. Gamma Console was never in that list, which is a pre-existing gap of the same nature.
- **Each workflow scans the variants it builds**: chainguard on a branch push, chainguard-fips on the nightly. The 17 images stay covered between the two.
- **The Aikido scan takes its variant list explicitly.** It used to derive both the images to pull and the jobs to wait for from one boolean; they are now one list, so they cannot drift apart.

**Security-sensitive**: this changes CI/CD configuration and where the container image scan runs. Coverage is unchanged — 17 distinct images, verified across both generated pipelines — but the scan is now split across two workflows instead of one.

## Additional context

**How it was verified.** `tsc` and lint clean, 153 tests green, and every `requires` in all 52 generated fixtures checked to resolve to a job present in its own workflow. That last check is what makes a change like this safe: removing a job from a path silently orphans whatever depended on it, and it happened three times on #19087.

**What was deliberately left alone.**

The standard images stay on the branch push even though the cluster does not pull them. They feed `Trigger SaaS Docker images creation`, which POSTs to the `cloud-distributions` pipeline with `branch-version`, `sha1` and `os-distribution: debian` — so that pipeline resolves images we pushed, by tag. `Publish Helm chart` and `Deploy on NextGen Integration` sit behind it. Removing them would change a contract with another repository on an inference.

The nightly's own `Deploy on Azure cluster` has the same gap this PR closes on the branch path: it waits on `Integration tests` and `Run Cypress UI tests`, so it covers the four product images transitively but neither Gamma nor the chainguard variants. It is left as is because that `requires` carries the open question about fanning in on the E2E matrix, and changing it here would settle that question as a side effect.

**Worth a follow-up.** The nightly still builds the standard images, for the end-to-end suites — they pull `<branch>-<sha7>` from the registry. Since the nightly runs at the same commit as the last merge, those tags already exist and could be reused. The saving is smaller than it looks: the frontend builds have to stay either way, because the chainguard-fips images depend on them. And it would make the nightly depend on the merge pipeline having pushed, without anything in the graph saying so.

### DIFF
--- a/.circleci/ci/src/workflows/workflow-nightly.ts
+++ b/.circleci/ci/src/workflows/workflow-nightly.ts
@@ -19,7 +19,6 @@ import { CircleCIEnvironment } from '../pipelines';
 import { config } from '../config';
 import { devEnvironmentJobs } from './groups/dev-environment-jobs';
 import { e2eJobs } from './groups/e2e-jobs';
-import { chainguardImageJobs } from './groups/chainguard-image-jobs';
 import { chainguardFipsImageJobs } from './groups/chainguard-fips-image-jobs';
 
 /**
@@ -42,10 +41,8 @@ export class NightlyWorkflow {
     return new Workflow('nightly', [
       ...devEnvironmentJobs(dynamicConfig, environment),
 
-      // The chainguard variants are shipped but the environment does not run them, so a branch
-      // push has no reason to build them. They are built here instead, which is also what gives
-      // the Aikido scan below something to look at every night rather than only at release time.
-      ...chainguardImageJobs(dynamicConfig, environment),
+      // Only the FIPS variants. The chainguard images are built — and scanned — by every branch
+      // push, since the environment runs them.
       ...chainguardFipsImageJobs(dynamicConfig, environment),
 
       new workflow.WorkflowJob(testIntegrationJob, {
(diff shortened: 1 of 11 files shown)


.circleci/ci/src/workflows/workflow-nightly.ts
```diff
@@ -70,7 +67,7 @@ export class NightlyWorkflow {
 
       // Every variant, standard and chainguard alike. This is the only place they are scanned
       // between two releases.
-      ...AikidoScanDockerImagesJob.workflowJobs(dynamicConfig, environment, false, '', true),
+      ...AikidoScanDockerImagesJob.workflowJobs(dynamicConfig, environment, false, '', ['chainguard-fips']),
```
phiz71, line 71: Good catch, and the fix is the comment rather than the call.

Scanning chainguard here would be wrong now: those images are built by every branch push, and this PR moves their scan to that same workflow, so each one is scanned where it is built. Adding them back would scan them twice and, worse, make this workflow wait on builds it does not contain.

But you are right that the code and its documentation contradicted each other. Two comments were stale, not one — the line above `chainguardFipsImageJobs` still claimed "the scan below still covers them", and the one above the scan itself still said "every variant, standard and chainguard alike ... the only place they are scanned". Both dated from an earlier iteration of this PR and both are now corrected to say what the code does.

Coverage is unchanged across the two workflows: 17 distinct images, verified against the generated pipelines.

jaroslawlakomy-grav (approved):

PR #19224, opened 2026-08-18
feat(audit-logs): add organization and environment audit log support

aqibmohammadkhan:
Introduce audit log functionality with new endpoints, services, and UI components for organization and environment scopes. Update navigation and routing to include new audit log pages. Add permission handling and tests for access control.

## Issue

https://gravitee.atlassian.net/browse/FOUND-67
https://gravitee.atlassian.net/browse/FOUND-68
https://gravitee.atlassian.net/browse/FOUND-69
https://gravitee.atlassian.net/browse/FOUND-70

## Summary
- Adds Gamma organization and environment Audit pages in `gravitee-gamma-module-platform`, calling the existing v1 `/audit` APIs (no backend changes). Covers [FOUND-48](https://gravitee.atlassian.net/browse/FOUND-48): [FOUND-67](https://gravitee.atlassian.net/browse/FOUND-67) org search, [FOUND-70](https://gravitee.atlassian.net/browse/FOUND-70) env search, [FOUND-68](https://gravitee.atlassian.net/browse/FOUND-68) event types from `GET /audit/events`, [FOUND-69](https://gravitee.atlassian.net/browse/FOUND-69) client-side CSV/JSON export.
- Org search uses `GET /organizations/{org}/audit`; env search uses `GET /organizations/{org}/environments/{env}/audit`. License `apim-audit-trail` shows the existing Enterprise dialog instead of the table. Permissions: `organization-audit-r` / `environment-audit-r`.
- Time filters send `from`/`to` as epoch ms. Relative presets snapshot `Date.now()` when selected so the query key does not churn. The date range picker is always visible (Classic / API Audit parity); a custom range uses start of the first day through end of the last day.

**Out of scope (API cannot support):** summary cards, Result/failure column, actor query param, Labs badge.

## Test plan
- [x] Organization → System & Security → Audit lists org-wide logs; Environment → System & Security → Audit lists the current environment only.
- [x] Event, type, and (org) environment/API/application filters change the `/audit` query as expected.
- [x] Last 24 hours / 7 / 30 / 90 days: one request with `from` and `to`; changing the preset does not loop.
- [x] Calendar range: start and end dates appear as `from`/`to` (end date through 23:59:59.999).
- [x] Export CSV and JSON for the current filters; toast on success; too many rows is rejected.
- [x] Missing `apim-audit-trail` shows the license dialog and skips the table.
- [x] Missing audit read permission hides the nav item and redirects away from the page.

## Additional context

https://github.com/user-attachments/assets/17fa24b4-3aa8-47d1-bc1e-27db62869417

[FOUND-48]: https://gravitee.atlassian.net/browse/FOUND-48?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FOUND-67]: https://gravitee.atlassian.net/browse/FOUND-67?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FOUND-70]: https://gravitee.atlassian.net/browse/FOUND-70?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FOUND-68]: https://gravitee.atlassian.net/browse/FOUND-68?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FOUND-69]: https://gravitee.atlassian.net/browse/FOUND-69?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-gamma/gamma-ui-shared/src/browser/index.ts
+++ b/gravitee-gamma/gamma-ui-shared/src/browser/index.ts
@@ -20,8 +20,12 @@ export function downloadBlob(blob: Blob, fileName: string): void {
     const anchor = document.createElement('a');
     anchor.download = fileName;
     anchor.href = url;
+    document.body.appendChild(anchor);
     anchor.click();
-    URL.revokeObjectURL(url);
+    anchor.remove();
+    // Some browsers (older Safari/Edge) start the download asynchronously; revoking in this
+    // turn can cancel it. Yield so the click is processed first.
+    setTimeout(() => URL.revokeObjectURL(url), 0);
 }
 
 /** Builds a `Blob` from text content and downloads it as `fileName`. */
--- a/gravitee-gamma/gamma-ui-shared/src/browser/index.ts
+++ b/gravitee-gamma/gamma-ui-shared/src/browser/index.ts
@@ -13,15 +13,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { FeatureLicenseDialog } from '../../../shared/components/FeatureLicenseDialog';
-import { SHARDING_TAGS_UPGRADE } from '../license/shardingTagsLicense';
 
-export function ShardingTagsLicenseDialog({
-    open,
-    onOpenChange,
-}: Readonly<{
-    open: boolean;
-    onOpenChange: (open: boolean) => void;
-}>) {
-    return <FeatureLicenseDialog upgrade={SHARDING_TAGS_UPGRADE} open={open} onOpenChange={onOpenChange} />;
+/** Triggers a browser download of `blob` as `fileName`. */
+export function downloadBlob(blob: Blob, fileName: string): void {
+    const url = URL.createObjectURL(blob);
+    const anchor = document.createElement('a');
+    anchor.download = fileName;
+    anchor.href = url;
+    anchor.click();
+    URL.revokeObjectURL(url);
+}
+
+/** Builds a `Blob` from text content and downloads it as `fileName`. */
+export function downloadTextFile(content: string, fileName: string, mimeType: string): void {
+    downloadBlob(new Blob([content], { type: mimeType }), fileName);
 }
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsPageView.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsPageView.tsx
@@ -56,20 +56,42 @@ export function AuditLogsPageView({
     applications,
     apis,
 }: AuditLogsPageViewProps) {
-    if (isError) {
-        return <AuditPageHeading description="Failed to load audit logs. Please refresh and try again." />;
-    }
+    const filters = (
+        <AuditLogsFilters
+            scope={scope}
+            eventTypes={eventTypes}
+            event={state.event}
+            onEventChange={state.onEventChange}
+            referenceType={state.referenceType}
+            onReferenceTypeChange={state.onReferenceTypeChange}
+            environments={environments}
+            environmentId={state.environmentId}
+            onEnvironmentIdChange={state.onEnvironmentIdChange}
+            applications={applications}
+            applicationId={state.applicationId}
+            onApplicationIdChange={state.onApplicationIdChange}
+            apis={apis}
+            apiId={state.apiId}
+            onApiIdChange={state.onApiIdChange}
+            datePreset={state.datePreset}
+            onDatePresetChange={state.onDatePresetChange}
+            customRange={state.customRange}
+            onCustomRangeChange={state.onCustomRangeChange}
+            onReset={state.handleReset}
+            onExport={() => state.setExportOpen(true)}
+        />
+    );
 
     return (
         <div className="space-y-4">
             <AuditPageHeading description={description} />
-
+            {isError ? <p className="text-sm text-destructive">Failed to load audit logs. Please try again.</p> : null}
             <AuditLogsTable
-                rows={rows}
-                loading={loading}
+                rows={isError ? [] : rows}
+                loading={!isError && loading}
                 page={state.page}
                 pageSize={state.pageSize}
-                totalCount={totalCount}
+                totalCount={isError ? 0 : totalCount}
                 onPageChange={state.setPage}
                 onPageSizeChange={size => {
                     state.setPageSize(size);
@@ -78,31 +100,7 @@ export function AuditLogsPageView({
                 selected={state.selected}
                 onSelectRow={state.setSelected}
                 onCloseDetail={() => state.setSelected(null)}
-                toolbar={
-                    <AuditLogsFilters
-                        scope={scope}
-                        eventTypes={eventTypes}
-                        event={state.event}
-                        onEventChange={state.onEventChange}
-                        referenceType={state.referenceType}
-                        onReferenceTypeChange={state.onReferenceTypeChange}
-                        environments={environments}
-                        environmentId={state.environmentId}
-                        onEnvironmentIdChange={state.onEnvironmentIdChange}
-                        applications={applications}
-                        applicationId={state.applicationId}
-                        onApplicationIdChange={state.onApplicationIdChange}
-                        apis={apis}
-                        apiId={state.apiId}
-                        onApiIdChange={state.onApiIdChange}
-                        datePreset={state.datePreset}
-                        onDatePresetChange={state.onDatePresetChange}
-                        customRange={state.customRange}
-                        onCustomRangeChange={state.onCustomRangeChange}
-                        onReset={state.handleReset}
-                        onExport={() => state.setExportOpen(true)}
-                    />
-                }
+                toolbar={filters}
             />
             <AuditExportDialog
                 open={state.exportOpen}
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsPageView.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsPageView.tsx
@@ -1,0 +1,115 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { AuditExportDialog, AuditLogsFilters } from './AuditLogsFilters';
+import { AuditLogsTable } from './AuditLogsTable';
+import type { useAuditLogsPageState } from '../hooks/useAuditLogsPageState';
+import type { AuditGroupedRefs, AuditLogRow, AuditNamedRef, AuditScope } from '../types/auditLog';
+
+export interface AuditLogsPageViewProps {
+    readonly scope: AuditScope;
+    readonly description: string;
+    readonly state: ReturnType<typeof useAuditLogsPageState>;
+    readonly rows: readonly AuditLogRow[];
+    readonly totalCount: number;
+    readonly loading: boolean;
+    readonly isError: boolean;
+    readonly eventTypes: readonly string[];
+    readonly environments?: readonly AuditNamedRef[];
+    readonly applications: readonly AuditNamedRef[] | readonly AuditGroupedRefs[];
+    readonly apis: readonly AuditNamedRef[] | readonly AuditGroupedRefs[];
+}
+
+function AuditPageHeading({ description }: Readonly<{ description: string }>) {
+    return (
+        <div className="space-y-1">
+            <h1 className="text-2xl font-semibold tracking-tight">Audit</h1>
+            <p className="text-sm text-muted-foreground">{description}</p>
+        </div>
+    );
+}
+
+/** Shared chrome for both Audit pages: heading, filter toolbar, results table and export dialog. */
+export function AuditLogsPageView({
+    scope,
+    description,
+    state,
+    rows,
+    totalCount,
+    loading,
+    isError,
+    eventTypes,
+    environments,
+    applications,
+    apis,
+}: AuditLogsPageViewProps) {
+    if (isError) {
+        return <AuditPageHeading description="Failed to load audit logs. Please refresh and try again." />;
+    }
+
+    return (
+        <div className="space-y-4">
+            <AuditPageHeading description={description} />
+
+            <AuditLogsTable
+                rows={rows}
+                loading={loading}
+                page={state.page}
+                pageSize={state.pageSize}
+                totalCount={totalCount}
+                onPageChange={state.setPage}
+                onPageSizeChange={size => {
+                    state.setPageSize(size);
+                    state.setPage(1);
+                }}
+                selected={state.selected}
+                onSelectRow={state.setSelected}
+                onCloseDetail={() => state.setSelected(null)}
+                toolbar={
+                    <AuditLogsFilters
+                        scope={scope}
+                        eventTypes={eventTypes}
+                        event={state.event}
+                        onEventChange={state.onEventChange}
+                        referenceType={state.referenceType}
+                        onReferenceTypeChange={state.onReferenceTypeChange}
+                        environments={environments}
+                        environmentId={state.environmentId}
+                        onEnvironmentIdChange={state.onEnvironmentIdChange}
+                        applications={applications}
+                        applicationId={state.applicationId}
+                        onApplicationIdChange={state.onApplicationIdChange}
+                        apis={apis}
+                        apiId={state.apiId}
+                        onApiIdChange={state.onApiIdChange}
+                        datePreset={state.datePreset}
+                        onDatePresetChange={state.onDatePresetChange}
+                        customRange={state.customRange}
+                        onCustomRangeChange={state.onCustomRangeChange}
+                        onReset={state.handleReset}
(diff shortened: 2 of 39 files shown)


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/OrgAuditLogsPage.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const environmentsQuery = useAuditEnvironments(hasLicense && referenceType === 'ENVIRONMENT');
+    const applicationsQuery = useOrgAuditApplications(hasLicense && referenceType === 'APPLICATION');
+    const apisQuery = useOrgAuditApis(hasLicense && referenceType === 'API');
+
+    const isForbidden = isForbiddenApiError(Boolean(logsQuery.isError), logsQuery.error);
+    useForbiddenResourceRedirect({
+        isForbidden,
+        permissionPrefix: AUDIT_PERMISSION_PREFIXES,
+        redirectTo: '../applications',
+    });
```
ankita-gupta21, line 56: Classic treats org Audit and env Audit as two permissions. A 403 on env Audit should not hide Organization → Audit. Both pages pass `AUDIT_PERMISSION_PREFIXES `(`organization-audit-` and `environment-audit-`), so one 403 wipes both from the cached permission list.

```
// OrgAuditLogsPage
permissionPrefix: 'organization-audit-',

// EnvAuditLogsPage
permissionPrefix: 'environment-audit-',
```

same goes in file [EnvAuditLogsPage.tsx ~L52–L56](https://github.com/gravitee-io/gravitee-api-management/pull/19224/changes#diff-6453e9947c6cfd29b6ca3311245ace676ae7024db9c873ac91eda802e036a730R52-R56)
Let say a person has :
`organization-audit-r `(can use org Audit)
`environment-audit-r `(can use env Audit)

- Someone revokes **only** env Audit. then a person still has org Audit.
- She opens **Environment → Audit.**
- API returns **403**.
- The hook strips everything starting with `organization-audit-` and `environment-audit-.`
- that person is sent to Applications.
- Organization → Audit is gone from the nav, even though she still has organization-audit-r on the server.

Without the bug, only environment-audit-r would be dropped. Org Audit would still show and still work.
The same bug in reverse: org Audit 403 would hide Environment → Audit.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsPageView.tsx
```diff
@@ -0,0 +1,115 @@
...
+    loading,
+    isError,
+    eventTypes,
+    environments,
+    applications,
+    apis,
+}: AuditLogsPageViewProps) {
+    if (isError) {
+        return <AuditPageHeading description="Failed to load audit logs. Please refresh and try again." />;
+    }
```
ankita-gupta21: Cassic shows a snackbar and leaves Event / Type / dates on screen. Gamma replaces the whole page with one heading, so the user cannot change filters or retry without a full refresh.

keep heading + filters + table chrome, and show an error banner (same pattern as Tenants / Alerts):
```
{isError ? (
    <p className="text-sm text-destructive">Failed to load audit logs. Please try again.</p>
) : (
    <AuditLogsTable ... />
)}
```

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/OrgAuditLogsPage.tsx
aqibmohammadkhan: done.
A 403 on Environment Audit was stripping both environment-audit-* and organization-audit-* from the cached permission list, so Organization → Audit disappeared even when organization-audit-r was still valid (and the reverse). Each page now strips only its own prefix: Org uses organization-audit-, Env uses environment-audit-.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsPageView.tsx
aqibmohammadkhan: done.
A failed search no longer replaces the page with a single heading. Heading, Event / Type / dates, table chrome, and Export stay on screen, with an inline error: “Failed to load audit logs. Please try again.”

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/EnvAuditLogsPage.tsx
```diff
@@ -0,0 +1,227 @@
...
+    AuditSearchParams,
+} from '../features/audit-logs/types/auditLog';
+import { auditLogsToCsv, auditLogsToJson, buildAuditExportFileName, downloadAuditExport } from '../features/audit-logs/utils/auditExport';
+import { toAuditLogRow } from '../features/audit-logs/utils/auditListFormat';
+import { AUDIT_PERMISSION_PREFIXES } from '../features/audit-logs/utils/auditPermissions';
+import { useForbiddenResourceRedirect } from '../shared/hooks/useForbiddenResourceRedirect';
+import { notify } from '../shared/notify';
+import { isForbiddenApiError } from '../shared/utils/apiErrors';
+
+export function EnvAuditLogsPage() {
```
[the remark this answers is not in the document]
aqibmohammadkhan: Pages already share useAuditLogsPageState and AuditLogsPageView; splitting further would not change behavior.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/hooks/useAuditLogsPageState.ts
```diff
@@ -0,0 +1,183 @@
...
+import { toAuditLogRow } from '../utils/auditListFormat';
+
+type ExportAudits = (params: Omit<AuditSearchParams, 'page' | 'size'>) => Promise<AuditMetadataPage>;
+
+/**
+ * Filter, pagination, selection and export state shared by the organization and environment
+ * Audit pages. The pages differ only in which services they hit, so they inject `exportAudits`
+ * and read `params` back to drive their own scope-specific queries.
+ */
+export function useAuditLogsPageState(exportAudits: ExportAudits) {
```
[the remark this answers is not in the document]
aqibmohammadkhan: Pages already share useAuditLogsPageState and AuditLogsPageView; splitting further would not change behavior.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+function buildColumns(onSelectRow: (row: AuditLogRow) => void): DataTableProps<AuditLogRow>['columns'] {
+    return [
+        {
+            id: 'Date',
+            accessorFn: (row: AuditLogRow) => row.createdAt,
+            header: 'Date',
+            enableSorting: false,
+            cell: ({ row }: ColCell<AuditLogRow>) => (
+                <button type="button" className="text-left" onClick={() => onSelectRow(row.original)}>
```
pragsty: **[Score: 50/100]** 💡 The Date cell (and the User cell below it) is a `<button>` whose accessible name is only the rendered timestamp/username, with no indication that activating it opens the audit detail sheet. Screen-reader users won’t know the cell is interactive or what it does, and it’s inconsistent with the Type/Reference/Event columns, which are not clickable. Consider adding an `aria-label` (e.g. “View audit details”) to these buttons, or make row activation explicit and consistent across all columns.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
```diff
@@ -0,0 +1,245 @@
...
+    onCloseDetail,
+}: AuditLogsTableProps) {
+    const columns = useMemo(() => buildColumns(onSelectRow), [onSelectRow]);
+
+    return (
+        <>
+            <DataTable
+                aria-label="Audit logs"
+                columns={columns}
+                data={[...rows]}
```
pragsty: **[Score: 20/100]** 🎨 `data={[...rows]}` copies the array on every render. `rows` is already freshly memoized by the caller (via `useMemo` in the page), so the spread is redundant — pass `rows` directly.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
```diff
@@ -0,0 +1,245 @@
...
+            header: 'Event',
+            enableSorting: false,
+            cell: ({ row }: ColCell<AuditLogRow>) => <span className="font-mono text-sm">{row.original.event}</span>,
+        },
+        {
+            id: 'Target',
+            header: 'Target',
+            enableSorting: false,
+            cell: ({ row }: ColCell<AuditLogRow>) => {
+                const text = formatAuditTargetText(row.original.targets);
```
pragsty: **[Score: 20/100]** 🎨 `const text = formatAuditTargetText(...)` is only used as a truthiness check; the cell then re-maps over `targets` directly. Drop the intermediate variable and branch on `row.original.targets.length > 0` instead.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
```diff
@@ -0,0 +1,245 @@
...
+function buildColumns(onSelectRow: (row: AuditLogRow) => void): DataTableProps<AuditLogRow>['columns'] {
+    return [
+        {
+            id: 'Date',
+            accessorFn: (row: AuditLogRow) => row.createdAt,
+            header: 'Date',
+            enableSorting: false,
+            cell: ({ row }: ColCell<AuditLogRow>) => (
+                <button type="button" className="text-left" onClick={() => onSelectRow(row.original)}>
+                    <DateCell value={new Date(row.original.createdAt).toISOString()} format="absolute" />
```
pragsty, line 77: **[Score: 25/100]** 🎨 `format="absolute"` deviates from the Graphene default of relative timestamps with a full-date tooltip (see `graphene.md` → Data table → “Dates (relative format)”). Absolute timestamps are defensible for a forensic audit view, but if that is the intent, add a short comment noting why, or reuse the standard relative format for consistency with sibling list pages.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
```diff
@@ -0,0 +1,245 @@
...
+                        <Button
+                            type="button"
+                            variant="ghost"
+                            size="icon"
+                            className="size-8"
+                            onClick={() => onSelectRow(row.original)}
+                            aria-label="View patch"
+                            title="View patch"
+                        >
+                            <EyeIcon className="size-4" />
```
pragsty: **[Score: 20/100]** 🎨 `<EyeIcon>` carries no `aria-hidden`. Since the button already provides `aria-label="View patch"`, mark the icon decorative with `aria-hidden` (per `graphene.md` icon guidance).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
```diff
@@ -0,0 +1,245 @@
...
+                              pageSizeOptions: [...AUDIT_PAGE_SIZE_OPTIONS],
+                              onPageChange,
+                              onPageSizeChange: (size: number) => {
+                                  onPageSizeChange(size);
+                                  onPageChange(1);
+                              },
+                          }
+                        : undefined
+                }
+                emptyMessage={
```
pragsty, line 200: **[Score: 40/100]** 💡 `emptyMessage` always renders `variant="no-results"` with “Try adjusting or clearing your filters”, even when the user has no active filters and the collection is simply empty. `graphene.md` distinguishes `first-use` (genuinely empty) from `no-results` (filters active). Consider branching on `totalCount === 0 && !hasActiveFilters` so empty collections aren’t told to clear filters that aren’t set.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/hooks/useAuditLogs.ts
```diff
@@ -0,0 +1,93 @@
...
+    return useQuery({
+        queryKey: auditKeys.apis(environmentId),
+        queryFn: () => listAuditApis(environmentId ?? ''),
+        enabled: enabled && Boolean(environmentId),
+    });
+}
+
+// The org-wide pickers fan out per environment, so they depend on the environments query rather than
+// refetching `/environments` themselves — all three consumers then share one cached response.
+export function useOrgAuditApplications(enabled: boolean) {
```
pragsty: **[Score: 40/100]** 💡 `useOrgAuditApplications` and `useOrgAuditApis` each independently call `useAuditEnvironments(enabled)`, and `OrgAuditLogsPage` calls it a third time for the environment picker. React Query dedupes the network by query key, but the `enabled` flags are now duplicated across three call sites and the coupling is implicit. Lifting the environments query to the page (or passing the already-resolved `environments` in) would make the dependency explicit and easier to reason about.

gravitee-gamma/gamma-ui-shared/src/browser/index.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+/** Triggers a browser download of `blob` as `fileName`. */
+export function downloadBlob(blob: Blob, fileName: string): void {
+    const url = URL.createObjectURL(blob);
+    const anchor = document.createElement('a');
+    anchor.download = fileName;
+    anchor.href = url;
+    anchor.click();
+    URL.revokeObjectURL(url);
```
pragsty: **[Score: 45/100]** 💡 `URL.revokeObjectURL(url)` runs synchronously right after `anchor.click()`. In some browsers (notably older Safari/Edge) revoking immediately can race the download. The common safe pattern is to append the anchor to the document, click it, remove it, and defer revocation (e.g. `setTimeout(() => URL.revokeObjectURL(url), 0)`). This shared helper is also used by the API export path, so the fix benefits both consumers.

ankita-gupta21 (approved):

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
aqibmohammadkhan: Date and User buttons now have aria-label="View audit details" so the action is named for screen readers.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
aqibmohammadkhan: Eye icon is aria-hidden="true".

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
aqibmohammadkhan: Target cells branch on targets.length > 0 instead of formatAuditTargetText.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
aqibmohammadkhan: data={rows} is passed through (the spread was only there because the prop was readonly).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
aqibmohammadkhan: Dates stay absolute, with a comment: this is a forensic log, not a recency feed.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/components/AuditLogsTable.tsx
aqibmohammadkhan: Empty collection with no filters uses first-use (“No audit logs”). Filtered empty uses no-results (“Try adjusting or clearing your filters”). Filters stay on screen either way.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/audit-logs/hooks/useAuditLogs.ts
aqibmohammadkhan: useAuditEnvironments lives on the page. App and API org pickers take that result instead of each calling the environments query.

gravitee-gamma/gamma-ui-shared/src/browser/index.ts
aqibmohammadkhan: downloadBlob appends the anchor, clicks, removes it, then revokeObjectURL on a timeout so Safari/Edge don’t cancel the download. Same helper the API export path uses.

pragsty (approved):

PR #19225, opened 2026-08-19
feat(gamma): added activity + updated history env alerts

arpzilla-dev:
## Issue

https://gravitee.atlassian.net/browse/APIM-14913

## Description

Adds the Gamma environment Activity tab: summary cards (Total / Info / Warning / Critical) and a table of alert name, events triggered, and a link to history. My alerts and Activity are sibling tabs; create/edit stay full-page.

Also fixes alert history: events API defaults to page=0, and timestamps use created_at (epoch millis) with classic-style relative time instead of Invalid Date.

## Additional context

https://github.com/user-attachments/assets/61944a85-d403-4b02-a899-55620be96eca

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/gravitee-gamma/gamma-ui-shared/src/time/formatRelativeDateTime.ts
+++ b/gravitee-gamma/gamma-ui-shared/src/time/formatRelativeDateTime.ts
@@ -1,0 +1,77 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const EMPTY = '—';
+const WEEK_MS = 7 * 24 * 60 * 60 * 1000;
+const ABSOLUTE_DATE_TIME_OPTIONS: Intl.DateTimeFormatOptions = {
+    day: 'numeric',
+    month: 'short',
+    year: 'numeric',
+    hour: '2-digit',
+    minute: '2-digit',
+    second: '2-digit',
+    hour12: false,
+};
+
+function parseDate(value: number | string | undefined | null): Date | undefined {
+    if (value === undefined || value === null || value === '') {
+        return undefined;
+    }
+    const date = new Date(value);
+    if (Number.isNaN(date.getTime())) {
+        return undefined;
+    }
+    return date;
+}
+
+/** Classic `date:'medium'` tooltip: always an absolute datetime. */
+export function formatAbsoluteDateTime(value: number | string | undefined | null): string {
+    const date = parseDate(value);
+    if (!date) {
+        return EMPTY;
+    }
+    return date.toLocaleString('en-GB', ABSOLUTE_DATE_TIME_OPTIONS);
+}
+
+/** Classic `humanDatetimeFilter`: relative time if younger than a week, else an absolute datetime. */
+export function formatRelativeDateTime(value: number | string | undefined | null, now = Date.now()): string {
+    const date = parseDate(value);
+    if (!date) {
+        return EMPTY;
+    }
+    const diffMs = now - date.getTime();
+    if (diffMs < WEEK_MS) {
+        return formatRelativeFromNow(diffMs);
+    }
+    return formatAbsoluteDateTime(value);
+}
+
+function formatRelativeFromNow(diffMs: number): string {
+    const rtf = new Intl.RelativeTimeFormat('en', { numeric: 'always' });
+    const seconds = Math.max(0, Math.round(diffMs / 1000));
+    if (seconds < 60) {
+        return rtf.format(-seconds, 'second');
+    }
+    const minutes = Math.round(seconds / 60);
+    if (minutes < 60) {
+        return rtf.format(-minutes, 'minute');
+    }
+    const hours = Math.round(minutes / 60);
+    if (hours < 24) {
+        return rtf.format(-hours, 'hour');
+    }
+    return rtf.format(-Math.round(hours / 24), 'day');
+}
--- a/gravitee-gamma/gamma-ui-shared/src/time/index.ts
+++ b/gravitee-gamma/gamma-ui-shared/src/time/index.ts
@@ -14,11 +14,4 @@
  * limitations under the License.
  */
 
-export const platformAlertKeys = {
-    all: ['platform-alerts'] as const,
-    list: (envId: string) => [...platformAlertKeys.all, 'list', envId] as const,
-    history: (envId: string, alertId: string) => [...platformAlertKeys.all, 'history', envId, alertId] as const,
-    notifiers: (envId: string) => [...platformAlertKeys.all, 'notifiers', envId] as const,
-    notifierSchema: (envId: string, notifierId: string) => [...platformAlertKeys.all, 'notifier-schema', envId, notifierId] as const,
-    analytics: (envId: string, rangeId: string) => [...platformAlertKeys.all, 'analytics', envId, rangeId] as const,
-} as const;
+export { formatAbsoluteDateTime, formatRelativeDateTime } from './formatRelativeDateTime';
--- a/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/pages/detail/alerts/HistoryTab.spec.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/pages/detail/alerts/HistoryTab.spec.tsx
@@ -14,13 +14,26 @@
  * limitations under the License.
  */
 import { render, screen } from '@testing-library/react';
+import type { ReactNode } from 'react';
 
 import { HistoryTab } from './HistoryTab';
+import { formatAbsoluteDateTime } from '../../../../../shared/time';
 
 jest.mock('@gravitee/graphene-core/icons', () => new Proxy({}, { get: () => () => null }));
 
+jest.mock('@gravitee/graphene-core', () => {
+    const actual = jest.requireActual('@gravitee/graphene-core');
+    return {
+        ...actual,
+        Tooltip: ({ children }: { children: ReactNode }) => children,
+        TooltipProvider: ({ children }: { children: ReactNode }) => children,
+        TooltipTrigger: ({ children }: { children: ReactNode }) => children,
+        TooltipContent: ({ children }: { children: ReactNode }) => <span>{children}</span>,
+    };
+});
+
 describe('HistoryTab', () => {
-    it('renders a date from created_at epoch millis, not Invalid Date', () => {
+    it('renders relative time from created_at epoch millis, not Invalid Date', () => {
         const createdAt = Date.now() - 3 * 60_000;
         render(
             <HistoryTab
@@ -32,7 +45,23 @@ describe('HistoryTab', () => {
         );
 
         expect(screen.queryByText('Invalid Date')).toBeNull();
+        expect(screen.getByText(/minutes ago/)).not.toBeNull();
         expect(screen.getByText(/response.response_time: 304/)).not.toBeNull();
+    });
+
+    it('shows an absolute timestamp in a tooltip', () => {
+        const createdAt = Date.now() - 3 * 60_000;
+        render(
+            <HistoryTab
+                historyPage={{
+                    content: [{ message: 'Health check failed', created_at: createdAt }],
+                    totalElements: 1,
+                }}
+            />,
+        );
+
+        expect(screen.getByText(/minutes ago/)).not.toBeNull();
+        expect(screen.getByText(formatAbsoluteDateTime(createdAt))).not.toBeNull();
     });
 
     it('renders two events that share a message', () => {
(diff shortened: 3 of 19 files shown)


aqibmohammadkhan (commented):
1. Duplicate React keys — reproduced

AlertEventEntity has only created_at and message. There is no id field on the wire, so evt.id is always undefined and key={evt.id ?? evt.message} is always the message. Two events with the same message — routine for NODE_HEALTHCHECK-style alerts with static text — collide. I confirmed it with a throwaway test against the real component:

Encountered two children with the same key... Non-unique keys may cause children
to be duplicated and/or omitted — the behavior is unsupported

Affects both copies (HistoryTab.tsx:69 and the apim one). Use the index — key={${evt.created_at}-${i}} — and drop id?: string from AlertHistoryEvent, since it's describing a field the API never sends.

Related: the createdAt camelCase fallback is dead code for the same reason, and the "renders relative time from camelCase createdAt" test only exercises the fallback. Making all three fields optional means TS stops catching a wrong field name — { message: string; created_at: number } would.

2. ~50 lines duplicated verbatim across modules

formatAlertEventTime.ts (apim) and the block appended to alertListFormat.ts (platform) are byte-identical apart from the const EMPTY line. The two HistoryTab.spec.tsx files are literally the same git blob (e8d2e56c066 on both sides of the diff), and the format assertions are duplicated a third time.

There's already a sharing mechanism for exactly this: gravitee-gamma/gamma-ui-shared/src/, wired into both modules via the @gravitee/gamma-ui-shared/* alias in tsconfig.app.json and jest.config.ts. Adding a time/ (or alerts/) entry there and re-exporting collapses this to one implementation and one spec.

3. Refresh wipes the board to a skeleton

refresh() sets a new window, which changes the query key, which means a brand-new cache entry with no data — so isLoading is true and the entire page is replaced by skeletons on every Refresh click and every time-range change. Each click also strands another distinct cache entry.

The same feature already does this correctly: useAlertForm.ts:203 keeps a stable key and uses refetch() + isFetching. Worth matching — keep queryKey: [...keys.analytics(envId, rangeId)], compute the window inside queryFn, and refresh via refetch().

Smaller

- const [window, setWindow] shadows the global window for the whole component (AlertsActivityPage.tsx). timeWindow avoids the trap.
- Two `<h1>s` under one tab bar. AlertsLayout renders tabs, then the outlet renders <h1>Alerts</h1> (AlertsPage.tsx:142) or` <h1>Alerts board</h1>`. The page title ends up below the tabs, and sibling tabs of one section carry different titles. Hoisting "Alerts" into the layout above the tab bar and demoting the per-tab heading would read better.
- total = info + warning + critical is correct today, but a new severity would silently vanish from "Total Alerts". Object.values(data?.bySeverity ?? {}).reduce((a, b) => a + b, 0) is safer and lets you drop the severityCount helper's loose Record<string, number> signature.
- `<Label>Quick time range</Label>` has no htmlFor, so clicking it does nothing; the trigger's aria-label carries the a11y name alone.
- useMemo on summaryCards wraps a 4-element literal whose four primitive deps always change together — it costs more than it saves.
- enabled: !!environmentId: in RQ v5 a disabled query is isPending but not isLoading, so before the environment resolves the page shows "0 / No alert events" rather than a skeleton. Probably unreachable in practice, but it's the failure mode if env loading ever becomes async.

arpzilla-dev:
> 1. Duplicate React keys — reproduced
> 
> AlertEventEntity has only created_at and message. There is no id field on the wire, so evt.id is always undefined and key={evt.id ?? evt.message} is always the message. Two events with the same message — routine for NODE_HEALTHCHECK-style alerts with static text — collide. I confirmed it with a throwaway test against the real component:
> 
> Encountered two children with the same key... Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported
> 
> Affects both copies (HistoryTab.tsx:69 and the apim one). Use the index — key={${evt.created_at}-${i}} — and drop id?: string from AlertHistoryEvent, since it's describing a field the API never sends.
> 
> Related: the createdAt camelCase fallback is dead code for the same reason, and the "renders relative time from camelCase createdAt" test only exercises the fallback. Making all three fields optional means TS stops catching a wrong field name — { message: string; created_at: number } would.
> 
> 2. ~50 lines duplicated verbatim across modules
> 
> formatAlertEventTime.ts (apim) and the block appended to alertListFormat.ts (platform) are byte-identical apart from the const EMPTY line. The two HistoryTab.spec.tsx files are literally the same git blob (e8d2e56c066 on both sides of the diff), and the format assertions are duplicated a third time.
> 
> There's already a sharing mechanism for exactly this: gravitee-gamma/gamma-ui-shared/src/, wired into both modules via the @gravitee/gamma-ui-shared/* alias in tsconfig.app.json and jest.config.ts. Adding a time/ (or alerts/) entry there and re-exporting collapses this to one implementation and one spec.
> 
> 3. Refresh wipes the board to a skeleton
> 
> refresh() sets a new window, which changes the query key, which means a brand-new cache entry with no data — so isLoading is true and the entire page is replaced by skeletons on every Refresh click and every time-range change. Each click also strands another distinct cache entry.
> 
> The same feature already does this correctly: useAlertForm.ts:203 keeps a stable key and uses refetch() + isFetching. Worth matching — keep queryKey: [...keys.analytics(envId, rangeId)], compute the window inside queryFn, and refresh via refetch().
> 
> Smaller
> 
> * const [window, setWindow] shadows the global window for the whole component (AlertsActivityPage.tsx). timeWindow avoids the trap.
> * Two `<h1>s` under one tab bar. AlertsLayout renders tabs, then the outlet renders # Alerts (AlertsPage.tsx:142) or` <h1>Alerts board</h1>`. The page title ends up below the tabs, and sibling tabs of one section carry different titles. Hoisting "Alerts" into the layout above the tab bar and demoting the per-tab heading would read better.
> * total = info + warning + critical is correct today, but a new severity would silently vanish from "Total Alerts". Object.values(data?.bySeverity ?? {}).reduce((a, b) => a + b, 0) is safer and lets you drop the severityCount helper's loose Record<string, number> signature.
> * `<Label>Quick time range</Label>` has no htmlFor, so clicking it does nothing; the trigger's aria-label carries the a11y name alone.
> * useMemo on summaryCards wraps a 4-element literal whose four primitive deps always change together — it costs more than it saves.
> * enabled: !!environmentId: in RQ v5 a disabled query is isPending but not isLoading, so before the environment resolves the page shows "0 / No alert events" rather than a skeleton. Probably unreachable in practice, but it's the failure mode if env loading ever becomes async.

Thanks — addressing each point:

1. Duplicate React keys
Fixed.
Both HistoryTabs use key={`${evt.created_at}-${i}`}. AlertHistoryEvent is { message, created_at } — no id / createdAt. Spec covers two events that share a message.

2. ~50 lines duplicated across modules
Fixed the formatter.
formatRelativeDateTime lives in gamma-ui-shared and both HistoryTabs use it. The two HistoryTab specs are still colocated with each component (they assert render, not just the formatter).

3. Refresh wipes the board to a skeleton
Fixed.
queryKey is stable on rangeId; the time window is computed inside queryFn; Refresh uses refetch() + isFetching.

Smaller

window shadowing — fixed (rangeId / activityTimeWindow).
Label htmlFor — fixed.
useMemo on summary cards — dropped.

Two h1s under the tab bar — left. List vs activity still have their own titles (“Alerts” / “Alerts board”); not hoisting in this PR.
Total = info + warning + critical — left. Those are the three severities the API returns today.
enabled: !!environmentId — left. Environment is available when this page mounts; treating the disabled-query pending state as a skeleton is a follow-up if env loading becomes async.

aqibmohammadkhan (commented):
Should fix
1. Shared formatter is in gamma-ui-shared, but not actually wired as shared

formatRelativeDateTime lives in gamma-ui-shared/src/time/, which is the right place, but both modules still deep-import the file:

APIM: 9 ../ from HistoryTab.tsx
Platform: 7 ../ from HistoryTab.tsx
There is still no @gravitee/gamma-ui-shared/time path (only /api exists in tsconfig.app.json and jest.config.ts), no src/time/index.ts, and the only unit spec sits under the platform module.

That is exactly the failure mode the earlier comment was trying to prevent: two depths, one copy-paste away from a broken import. The APIM path happens to be correct today; it is still not a public module surface.

2. History lost classic’s absolute timestamp

Classic:
```
alert-history.html Ln 38–41
```
```
<span
  >{{event.created_at | humanDatetimeFilter }}
  <md-tooltip>{{ event.created_at | date:'medium' }}</md-tooltip>
```

Gamma only renders formatRelativeDateTime(evt.created_at). For anything younger than a week you get “3 minutes ago” with no way to see the real time. Reuse the Graphene Tooltip already used on AlertsPage counters.

3. “View history” is a <Button> with a duplicate name

navigate(\../${alert.id}?tab=history`) is the right destination (useAlertFormalready honors?tab=history). A Link(orasChild+Link) keeps cmd/ctrl-click and copy-URL. Accessible name should include the alert, e.g. View history for ${alert.name}` — there are N identical “View history” buttons today.

4. Activity rows drop severity

Classic color-codes each alert card. The analytics payload already has severity, and SeverityBadge already exists for the list page. The board table is name / count / history only, so a CRITICAL row and an INFO row look identical. A badge (or a severity column) is a small parity fix.

5. Changing the time range still replaces the board with skeletons

Refresh is fixed. Range change still changes queryKey, so isLoading is true and the page unmounts into skeletons. Elsewhere Gamma uses placeholderData: keepPreviousData (users, groups, audit logs, applications). Same pattern here would finish original comment #3.

Smaller:

- Refresh disables while isFetching but has no busy/spinner, so a click can look like a no-op.
- Error card has no “Try again”; Refresh does refetch, so this is optional.
- History still has no pagination (classic md-table-pagination). Out of scope if this PR is only Invalid Date + page 0, but the first page is now correct.


Out of scope vs classic (flag, don’t block)
- Classic activity also has a severity bar chart and empty states for “no alerting plugin” / “no alerts configured” with a create CTA. This PR’s description is cards + table, so that looks intentional. If APIM-14913’s AC requires visual parity, those are still missing.

arpzilla-dev:
> Should fix
> 
> 1. Shared formatter is in gamma-ui-shared, but not actually wired as shared
> 
> formatRelativeDateTime lives in gamma-ui-shared/src/time/, which is the right place, but both modules still deep-import the file:
> 
> APIM: 9 ../ from HistoryTab.tsx Platform: 7 ../ from HistoryTab.tsx There is still no @gravitee/gamma-ui-shared/time path (only /api exists in tsconfig.app.json and jest.config.ts), no src/time/index.ts, and the only unit spec sits under the platform module.
> 
> That is exactly the failure mode the earlier comment was trying to prevent: two depths, one copy-paste away from a broken import. The APIM path happens to be correct today; it is still not a public module surface.
> 
> 2. History lost classic’s absolute timestamp
> 
> Classic:
> 
> ```
> alert-history.html Ln 38–41
> ```
> 
> ```
> <span
>   >{{event.created_at | humanDatetimeFilter }}
>   <md-tooltip>{{ event.created_at | date:'medium' }}</md-tooltip>
> ```
> 
> Gamma only renders formatRelativeDateTime(evt.created_at). For anything younger than a week you get “3 minutes ago” with no way to see the real time. Reuse the Graphene Tooltip already used on AlertsPage counters.
> 
> 3. “View history” is a  with a duplicate name
> 
> navigate(../${alert.id}?tab=history`) is the right destination (useAlertFormalready honors?tab=history). A Link(orasChild+Link) keeps cmd/ctrl-click and copy-URL. Accessible name should include the alert, e.g. View history for ${alert.name}` — there are N identical “View history” buttons today.
> 
> 4. Activity rows drop severity
> 
> Classic color-codes each alert card. The analytics payload already has severity, and SeverityBadge already exists for the list page. The board table is name / count / history only, so a CRITICAL row and an INFO row look identical. A badge (or a severity column) is a small parity fix.
> 
> 5. Changing the time range still replaces the board with skeletons
> 
> Refresh is fixed. Range change still changes queryKey, so isLoading is true and the page unmounts into skeletons. Elsewhere Gamma uses placeholderData: keepPreviousData (users, groups, audit logs, applications). Same pattern here would finish original comment #3.
> 
> Smaller:
> 
> * Refresh disables while isFetching but has no busy/spinner, so a click can look like a no-op.
> * Error card has no “Try again”; Refresh does refetch, so this is optional.
> * History still has no pagination (classic md-table-pagination). Out of scope if this PR is only Invalid Date + page 0, but the first page is now correct.
> 
> Out of scope vs classic (flag, don’t block)
> 
> * Classic activity also has a severity bar chart and empty states for “no alerting plugin” / “no alerts configured” with a create CTA. This PR’s description is cards + table, so that looks intentional. If APIM-14913’s AC requires visual parity, those are still missing.

Thanks — all five should-fix items are in.

1. Shared formatter
Helper stays in gamma-ui-shared/src/time/ (formatRelativeDateTime + formatAbsoluteDateTime), with src/time/index.ts. Both modules import through the same re-export pattern as shared/browser / shared/api (src/main/ui/shared/time), so we don’t add a second @gravitee/gamma-ui-shared/time tsconfig/Jest path. No more 7-/9-level deep imports.

2. Absolute timestamp
History cells keep the relative label and show the absolute datetime in a Graphene Tooltip (formatAbsoluteDateTime), same idea as Classic’s humanDatetimeFilter + date:'medium'.

3. View history
Button asChild + Link to ../${alert.id}?tab=history. Accessible name is View history for ${alert.name} so cmd/ctrl-click, copy URL, and unique names all work.

4. Severity
Activity table has a Severity column using SeverityBadge.

5. Time range skeletons
Analytics query uses placeholderData: keepPreviousData, same as users/groups/audit. Range change keeps the board; skeletons only on first load.

Smaller
Refresh now shows Refreshing… while isFetching. Error-card “Try again” left optional (Refresh already refetches). History pagination still out of scope.

aqibmohammadkhan (approved):

PR #19231, opened 2026-08-19
test(heartbeat): remove timing race in heartbeat event listener test

phiz71:
## Issue

N/A - no JIRA ticket opened for this one. Spotted on a CI run of `Test gateway`:
https://app.circleci.com/pipelines/github/gravitee-io/gravitee-api-management/92469/workflows/cc04e7b6-ead9-473d-9eca-d8944598f953/jobs/2004953/tests

## Description

`HeartbeatEventListenerTest.should_process_new_event_after_previous_one_completes` is flaky. The listener
releases its `isProcessing` flag in a `finally` block, once `createOrPatch` has returned, while the test used
to wake up on a `CountDownLatch` counted down from inside the `createOrPatch` mock. The main thread could
therefore post the second event before the worker had reached the `finally`: the event was discarded, the
second latch was never counted down, and the test failed after its 5s timeout.

Instead of retrying or sleeping, the test is now given a real synchronization point: the executor becomes
injectable, and since it is single-threaded, an extra task submitted behind an event is only run once that
event has been fully processed, flag release included. The public constructor keeps the exact same behaviour.

While in there, the two `discard` tests were doing their `verify` while the worker was still blocked, so they
would have passed even if the extra events had been accepted and queued. They now assert after the blocked
event has been released and the queue drained.

## Additional context

Failure as seen in CI - the discard of the second event and the 5s timeout are one and the same test:

```
07:56:44.899 [main] WARN HeartbeatEventListener -- Discarding heartbeat event id[2] type[GATEWAY_STARTED] - another heartbeat event is already being processed
[ERROR] should_process_new_event_after_previous_one_completes -- Time elapsed: 5.045 s <<< FAILURE!
Expecting value to be true but was false        (HeartbeatEventListenerTest.java:192)
```

Race made reproducible by delaying the flag release by 100ms (unfavourable scheduling, as on a loaded CI
machine): the previous test then fails exactly as in CI, while the new one stays green. Module suite green
locally (13 tests).

The test file is identical on `4.9.x`, `4.10.x`, `4.11.x`, `4.12.x` and `master`, hence the apply-on labels.
The cherry-pick to `4.11.x` will conflict on a single line, the log annotation (`@Slf4j` here, `@CustomLog`
from 4.11.x on) sitting right above the constructor.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-heartbeat/src/main/java/io/gravitee/gateway/services/heartbeat/impl/HeartbeatEventListener.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-heartbeat/src/main/java/io/gravitee/gateway/services/heartbeat/impl/HeartbeatEventListener.java
@@ -15,6 +15,7 @@
  */
 package io.gravitee.gateway.services.heartbeat.impl;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.gravitee.node.api.cluster.ClusterManager;
 import io.gravitee.node.api.cluster.messaging.Message;
 import io.gravitee.node.api.cluster.messaging.MessageListener;
@@ -44,9 +45,10 @@ public class HeartbeatEventListener implements MessageListener<Event> {
     }
 
     /**
-     * Visible for testing: lets the test provide the executor the events are processed on, so it can wait for a
-     * submitted task to be fully completed instead of relying on timings.
+     * Lets the test provide the executor the events are processed on, so it can wait for a submitted task to be fully
+     * completed instead of relying on timings.
      */
+    @VisibleForTesting
     HeartbeatEventListener(
         final ClusterManager clusterManager,
         final EventRepository eventRepository,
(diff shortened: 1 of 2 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-heartbeat/src/main/java/io/gravitee/gateway/services/heartbeat/impl/HeartbeatEventListener.java
```diff
@@ -23,24 +23,40 @@
...
 
+    public HeartbeatEventListener(final ClusterManager clusterManager, final EventRepository eventRepository) {
+        this(clusterManager, eventRepository, Executors.newSingleThreadExecutor(r -> new Thread(r, "gio-heartbeat-listener")));
+    }
+
+    /**
+     * Visible for testing: lets the test provide the executor the events are processed on, so it can wait for a
+     * submitted task to be fully completed instead of relying on timings.
+     */
+    HeartbeatEventListener(
```
ytvnr, line 52: you can add `@VisibleForTesting` annotation

ytvnr (approved):

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-heartbeat/src/main/java/io/gravitee/gateway/services/heartbeat/impl/HeartbeatEventListener.java
phiz71: Done in 5ff3ca44a7 — Guava is already on the module classpath (33.4.5-jre, provided, transitive), same as the other gateway modules using this annotation. Module suite still green.

PR #19232, opened 2026-08-19
fix(definition): never persist clearTextUpgrade:true for HTTP/1.1 endpoints

dev-marek:
**Issue**

https://gravitee.atlassian.net/browse/APIM-14964

**Purpose**

Stop `clearTextUpgrade` from being persisted as `true` on HTTP/1.1 endpoints, and stop it from silently drifting on unrelated endpoints when the API is saved.

**Summary of changes**

- **Root-cause fix**: `EndpointDeserializer` (`gravitee-apim-definition-jackson`) now normalizes the raw per-endpoint `http.clearTextUpgrade` JSON node to `false` whenever `version` is (or defaults to) `HTTP_1_1`, before it's captured into the endpoint's raw `configuration` string. This is the actual persistence boundary for V2 API endpoint overrides — the typed `HttpClientOptions` model and its getters are never consulted for this per-endpoint data (it's preserved as a raw JSON passthrough "for extendable connector architecture"), so this is where the invariant has to be enforced. Since this deserializer runs on both write and read, invalid data can no longer be newly stored, and pre-existing invalid data self-heals on next read.
- **Defense-in-depth**: added the same `isClearTextUpgrade()` masking getter to the V4 model (`io.gravitee.definition.model.v4.http.HttpClientOptions`), mirroring the existing fix already on the V2 model.
- **Defense-in-depth**: fixed the Console UI's `EndpointHttpConfigComponent.getHttpConfigValue()` to normalize `clearTextUpgrade` to `false` in the submitted payload whenever the endpoint's version is `HTTP_1_1`, so the native HTTP endpoint form can never submit the invalid combo.
- Added regression tests for all of the above.

**Out of scope**

- Not relaxing `Endpoint.configuration`'s raw-JSON-passthrough design more broadly — the fix is scoped to normalizing this one field.
- Not backfilling already-invalid stored data via an explicit migration — it self-heals on next read/write through the fixed deserializer (see testing below).

**Testing**

Unit tests: `gravitee-apim-definition-jackson` (183 tests, incl. 4 new), `gravitee-apim-definition-model`, and console-webui Jest suites (101 tests across 11 endpoint-related suites) all pass. See comment below for manual/dev testing against a live Docker deployment.

**Additional context**

Investigation notes (including why an earlier getter-only approach didn't work) are on the Jira ticket.

### DIFF
--- a/.circleci/ci/src/jobs/job-sonarcloud-analysis.ts
+++ b/.circleci/ci/src/jobs/job-sonarcloud-analysis.ts
@@ -61,6 +61,7 @@ export class SonarCloudAnalysisJob {
         name: 'Run Sonarcloud Analysis',
         command: `sonar-scanner -Dsonar.projectVersion=${apimVersion}`,
         working_directory: '<< parameters.working_directory >>',
+        no_output_timeout: '30m',
       }),
       new reusable.ReusedCommand(notifyOnFailureCmd),
       new commands.cache.Save({
--- a/.circleci/ci/src/pipelines/tests/resources/pull-requests/pull-requests-4-1-x.yml
+++ b/.circleci/ci/src/pipelines/tests/resources/pull-requests/pull-requests-4-1-x.yml
@@ -286,6 +286,7 @@ jobs:
           name: Run Sonarcloud Analysis
           command: sonar-scanner -Dsonar.projectVersion=4.2.0
           working_directory: << parameters.working_directory >>
+          no_output_timeout: 30m
       - cmd-notify-on-failure
       - save_cache:
           paths:
--- a/.circleci/ci/src/pipelines/tests/resources/pull-requests/pull-requests-custom-branch-backend-gateway-only.yml
+++ b/.circleci/ci/src/pipelines/tests/resources/pull-requests/pull-requests-custom-branch-backend-gateway-only.yml
@@ -202,6 +202,7 @@ jobs:
           name: Run Sonarcloud Analysis
           command: sonar-scanner -Dsonar.projectVersion=4.2.0
           working_directory: << parameters.working_directory >>
+          no_output_timeout: 30m
       - cmd-notify-on-failure
       - save_cache:
           paths:
--- a/.circleci/ci/src/pipelines/tests/resources/pull-requests/pull-requests-custom-branch-backend-only.yml
+++ b/.circleci/ci/src/pipelines/tests/resources/pull-requests/pull-requests-custom-branch-backend-only.yml
@@ -202,6 +202,7 @@ jobs:
           name: Run Sonarcloud Analysis
           command: sonar-scanner -Dsonar.projectVersion=4.2.0
           working_directory: << parameters.working_directory >>
+          no_output_timeout: 30m
       - cmd-notify-on-failure
       - save_cache:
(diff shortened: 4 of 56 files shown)


dev-marek:
## Dev testing (live Docker deployment)

Verified end-to-end against a real running stack, not just unit tests.

**Setup**
- Built the Management API distribution from this branch (`mvn ... package -DskipTests`) and rebuilt the `apim-management-api` Docker image from it.
- Redeployed it into a local `docker-compose` stack (`docker/quick-setup/mongodb`), alongside gateway, management UI, portal UI, mongo, elasticsearch.
- Ran a full smoke test: management API login, `/apis` list, gateway, management UI, portal UI all healthy.

**Bug repro, before the fix**
- Via `PUT /management/organizations/DEFAULT/environments/DEFAULT/apis/{id}` (the same call the Console UI's endpoint-edit screen makes), created endpoints on a V2 API with `clearTextUpgrade: true` + `version: HTTP_1_1` explicitly set.
- Confirmed the invalid combo round-tripped unmasked through both the general `GET /apis/{id}` and the dedicated `GET /apis/{id}/export` (the literal repro path in the ticket) — the getter-based masking approach I tried first had zero effect here, which is what led to the root-cause investigation.

**After the fix**
- Rebuilt and redeployed the image with the `EndpointDeserializer` fix.
- Re-fetched the *pre-existing* invalid endpoints (created before the fix, no resubmission) — they now come back as `clearTextUpgrade: false`, confirming the self-heal-on-read behavior.
- Submitted a brand-new endpoint with the invalid combo explicitly set via `PUT` — the response immediately showed `clearTextUpgrade: false`.
- Checked `GET /apis/{id}`, and `GET /apis/{id}/export` — no invalid combo anywhere.
- Re-ran the full smoke test — all green.
- Cleaned up the test endpoints, left the test API in its original state.

dev-marek:
## Verified all 6 QA scenarios from the Jira ticket

Walked through the scenario list from the Jira ticket via curl against the actual Management REST API (same calls the Console UI makes), on a fresh V2 API with this fix deployed:

| # | Scenario | Expected | Result |
|---|---|---|---|
| 1 | Newly created API | `clearTextUpgrade` NOT present | ✅ absent |
| 2 | New endpoint, inheriting group config | `clearTextUpgrade` NOT present | ✅ absent |
| 3 | New endpoint, no inherit, default params | must never be `true` | ✅ present, `false` |
| 4 | New endpoint, no inherit, Keep-Alive/Compression unchecked | must never be `true` | ✅ present, `false` |
| 5 | Second new endpoint group + new endpoint (defaults) | must never be `true` | ✅ present, `false` (group-level default also correctly absent) |
| 6 | New group added near default-endpoint-group + new endpoint | must never be `true` | ✅ present, `false` |

Also checked: re-fetched all endpoints after all 6 scenarios were applied in sequence — no cross-endpoint drift, every previously-set value stayed exactly as it was. And as a defense-in-depth check, simulated a non-Console client explicitly submitting `clearTextUpgrade: true` — still corrected to `false` server-side, since the fix lives in the deserializer rather than relying on the Console UI behaving.

jaroslawlakomy-grav (approved):

PR #19250, opened 2026-08-19
feat(logs): surface Kafka client library and credential on native API logs

ThibaudAV:
## Issue

https://gravitee.atlassian.net/browse/ESM-75

## Description

The native Kafka gateway already writes the client library advertised at the connection handshake and the authenticated credential into Elasticsearch. Neither reached the product: `NativeApiMetricsSourceMapper` never read `security-type` / `security-token`, and `NativeApiLog` projected none of the client-software keywords. Everything the gateway captured was reachable only by querying Elasticsearch directly — not through the console, not through the REST API.

This closes the read path for four fields, from the ES document to the native log detail view.

| Field | Source in the ES document | Written by |
|---|---|---|
| `clientSoftwareName` | `additional-metrics.keyword_native-kafka_client-software-name` | gravitee-reactor-native-kafka ESM-71 |
| `clientSoftwareVersion` | `additional-metrics.keyword_native-kafka_client-software-version` | gravitee-reactor-native-kafka ESM-71 |
| `securityType` | `security-type` (top level) | gravitee-reactor-native-kafka ESM-70 |
| `securityToken` | `security-token` (top level) | gravitee-reactor-native-kafka ESM-70 |

The two client-software values are KIP-511 fields the client advertises at the handshake; the gateway falls them back to `unknown`, so they are always present once a handshake completed. `securityType` / `securityToken` are **top-level** document fields — written alongside the HTTP request documents in the same `v4-metrics` index — so unlike the keywords they have to be read explicitly by the source mapper. `securityToken` is the application's OAuth client id, never a raw API key.

Only the **detail** projection carries them; the list projection is unchanged, matching how `clientId` and `brokerId` are already handled.

## Tests

- `NativeApiMetricsFindResponseAdapterTest` — the source mapper reads both top-level security fields, and leaves them null on a document that omits them (anonymous connection). Verified this fails without the mapper change.
- `NativeApiLogCrudServiceImplTest` — the four fields are flattened onto the detail projection.
- `NativeApiLogsMapperTest` — they survive the core → REST mapping.
- `api-runtime-logs-native-details.component.spec.ts` — the client card renders the library, its version and the credential; the anonymous case still renders the card.

Java: repository + rest-api tests green. Console: 44/44 on the native runtime-logs suites.

## Additional context

- **Depends on nothing in this repo**, but the fields only become non-null once the gateway side ships: ESM-70 is merged on `gravitee-reactor-native-kafka` `alpha`, ESM-71 is in review (gravitee-io/gravitee-reactor-native-kafka#374). Until then the projection is simply null everywhere, which is the same behaviour as today.
- **`failure-side` is deliberately not projected here.** It is defined in `NativeApiMetricKeys` and still unprojected, but it belongs to the analytics facet work, not to client info. Flagged rather than folded in.
- **Design rationale for `securityType` / `securityToken`** (rather than a bespoke `principal_*` pair) is recorded gateway-side in `docs/adr/0001-security-type-token-not-principal.md`. Worth knowing: `securityToken` is application-scoped, so it identifies the Gravitee application, not the individual Kafka client. The per-connection discriminators are `clientId` and the client-software pair.
- **Existing dashboards**: Kafka connection documents populate `security-type` in the same index as HTTP request documents, so a dashboard filtering on `security-type` and counting documents will start including connection-scoped documents.


callaertanthony:
## Code review — reviewed at head `6a1c851`

### :red_circle: Blockers

**1. `securityToken` is exposed for every API type, and for API-key plans it is the raw key**
`gravitee-gamma/.../ObservabilityLogsDataPortAdapter.java:213-214`

`.securityType(...).securityToken(...)` sits inside the unconditional `metricsOpt.ifPresent(...)` block rather than a native-API branch (unlike the surrounding `nativeMetrics.*` calls, which are simply null for non-native documents). So `LogDetail` / `LogDetailDto` now return `securityToken` for **every** API type, HTTP proxy included.

For API-key plans that field holds the credential verbatim: `ApiKeyAuthenticationHandler.java:95-96` does `metrics().setSecurityType(API_KEY); metrics().setSecurityToken(apiKey);`, and `FindApiMetricsDetailResponseAdapter` reads the same `security-token` root field for both v2 and v4 documents (v2 uses the identical field name, `LogBuilder.java:82-83`).

Scenario: an operator opens the log detail of an HTTP proxy API secured by an API-key plan and the response body contains the subscriber's live API key — readable by anyone with log-read permission.

Either scope the field to native connections, or redact/omit it when `securityType == API_KEY`.

**2. `NATIVE_CLIENT_SOFTWARE_NAME` is registered ANALYTICS-only, the inverse of the intent**
`gravitee-apim-rest-api/.../analytics/definition/analytics-definition.yaml:1995`

The new catalog entry declares no `signals:`, and `FilterSpec`'s compact constructor applies `Signal.DEFAULT = Set.of(ANALYTICS)` when the entry is null/empty (`FilterSpec.java:43`). Three comments added in this PR assert the opposite — `AnalyticsMeasuresAdapter.java:115-117`, `AnalyticsDefinitionMapper.java:454-456`, and `StaticFilters.java` all document it as LOGS-only. Two consequences:

- The console requests the catalog with `signal=<screen signal>` (`observability-filters-api.service.ts:89-91`), so **"Kafka Client Library" never appears on the APIM logs screen** — the whole `nativeClientSoftwareNames` chain through `SearchLogsFilters` → `ConnectionLogsCrudServiceImpl` → `SearchMetricsQueryAdapter` is unreachable from APIM's own logs API. Only Gamma, with its own registry, can drive it.
- It **does** appear on the analytics filter bar, where `AnalyticsQueryValidator.supportsAnalytics()` reads the same signal axis and therefore accepts it (`AnalyticsQueryValidator.java:158-163`) — while `AnalyticsMeasuresAdapter.toFilterName` is annotated `THROW_EXCEPTION` for that name. Any caller reaching the mapper gets an `IllegalArgumentException` rather than a clean 400.

Every other `THROW_EXCEPTION` name declares `signals: [LOGS]` (`ERROR_KEY`, `REQUEST_ID`, `TRANSACTION_ID`, `PAYLOAD`). The wire enum `FilterName` in `openapi-analytics.yaml` also omits the constant while `ObservabilityFilterName` has it, which corroborates that logs-only was the intent.

Fix: add `signals: [LOGS]` (and wire the translation in `SearchEnvironmentLogsUseCase`, which `ObservabilityLogsFilterCoverageTest` will then require).

### :yellow_circle: Improvements

**3. The public description of `securityToken` is not true**
`openapi-apis.yaml:7976` states *"the application OAuth client id. Never a raw API key."* — contradicted by finding 1. The same claim is repeated in the javadoc of `NativeApiMetrics`, repository `ApiMetricsDetail`, rest-api-model `ApiMetricsDetail`, and core `NativeApiLog`. Either make the claim true by filtering at the source, or correct the text; a wrong safety claim in a published spec is worse than none.

**4. `NATIVE_CLIENT_ID` is not advertised on the LOGS signal either**
`analytics-definition.yaml:1985` — no `signals:`, so ANALYTICS-only. That is defensible on its own (it is a real dashboard facet, lines 1151/1158), but this PR also wires `nativeClientIds` into the LOGS path and flips gamma's `StaticFilters` to `LOGS_ANALYTICS`. If logs-side filtering by client id is intended, this wants `signals: [LOGS, ANALYTICS]`.

**5. A test that cannot fail for the behaviour it names**
`api-runtime-logs-native-details.component.spec.ts:103` — *"renders a dash on the client card when the connection is anonymous and the client advertises no library"* asserts only `expect(await harness.clientCardText()).toContain('Client library')`, a static `<dt>` label. A mistyped `{{ x || '—' }}` binding, or one bound to the wrong property, keeps it green. Assert on the rendered dash.

### :white_circle: Nit

**6.** `FindApiMetricsDetailResponseAdapterTest.java:91` — `should_leave_the_security_credential_null_for_an_anonymous_connection` uses the `api-proxy-v2-metrics.json` fixture and comments "A key-less plan writes neither field", but the v2 request index uses the *same* `security-type` / `security-token` field names (`LogBuilder.java:82-83`). The test proves only that this fixture omits the keys, not anything about anonymity. The sibling at line 79 also re-asserts what `should_build_api_metrics_detail` already covers at lines 71-72.

### Checked and sound

- `SearchMetricsQueryAdapter` native client filters match the existing `addNativeConnectionStatusesFilter` shape (raw `terms` on `additional-metrics.*`, correct for a v4-only field); null/empty guards are right.
- `SearchLogsFilters` is `@Builder`-constructed everywhere, so inserting record components is safe.
- `NativeApiLogsMapper` and `ApiMetricsDetailAdapter` are name-based MapStruct mappings, so the new fields propagate end to end; `NativeApiLogCrudServiceImpl.map` correctly leaves them out of the list view, matching the "detail view only" OpenAPI wording.
- `FilterValuesQueryServiceImpl` previously fell through to `default -> throw` for `NATIVE_CLIENT_ID`; adding both cases fixes value listing, the path Gamma delegates to via `GetFilterValuesUseCase`.
- The `ObservabilityFiltersDefinitionResourceTest` count bump 50 → 51 is correct for one added catalog entry.

callaertanthony:
## Follow-up review — head `386cfc1`

Re-reviewed after the push. Comparing against [my earlier comment](https://github.com/gravitee-io/gravitee-api-management/pull/19250#issuecomment-5369810215) on `6a1c851`.

### :white_check_mark: Addressed

- `NATIVE_CLIENT_SOFTWARE_NAME` now declares `signals: [LOGS]`, and `NATIVE_CLIENT_ID` `signals: [LOGS, ANALYTICS]` — the catalog no longer inverts the intent, and `AnalyticsQueryValidatorTest` pins the rejected set.
- `openapi-analytics.yaml` correctly adds the name only to `ObservabilityFilterName`, not to `FacetName` or the analytics `FilterName`.
- Gamma's adapter now guards `securityType` / `securityToken` behind `if (nativeMetrics.connectionStatus() != null)`.
- The two tests that could not fail for the behaviour they named.

### :red_circle: Blockers

**1. The signals fix moved the bug one layer out: the catalog advertises two LOGS filters that the search endpoint cannot deserialize**
`.../management-v2-model/src/main/resources/openapi/openapi-logs.yaml:441`

`GET /observability/filters/definition?signal=LOGS` now returns `NATIVE_CLIENT_ID` and `NATIVE_CLIENT_SOFTWARE_NAME` — the PR's own `ObservabilityFiltersDefinitionResourceTest.should_narrow_the_catalog_to_the_requested_signal` asserts it. But the wire enum for `POST /v2/logs/search` is `FilterName` in `openapi-logs.yaml`, and at this head it contains **no `NATIVE_*` value at all** (`grep NATIVE openapi-logs.yaml` → empty). Every other name in the LOGS list is there; these two are not.

The console forwards catalog names verbatim (`environment-logs.service.ts:187`), so the moment an operator picks the "Kafka Client Library" or "Kafka Client ID" chip, the generated enum's `fromValue` throws and the **entire** search request fails — not just that filter. `LogsEngineMapper.mapFilterName` (`LogsEngineMapper.java:75`) is never reached, so the new `FilterName.NATIVE_CLIENT_ID` / `NATIVE_CLIENT_SOFTWARE_NAME` constants and `FilterContext.limitByNativeClientIds` / `…SoftwareNames` are unreachable from REST.

Fix: add both values to the `FilterName` enum in `openapi-logs.yaml`. Gamma is unaffected — it matches filter names as strings.

**2. The credential guard was added to gamma but not to the Console path**
`.../infra/crud_service/log/NativeApiLogCrudServiceImpl.java:110`

`mapFull` projects `securityType` / `securityToken` with no equivalent of the `connectionStatus != null` discriminator, and nothing upstream restricts the query to native APIs: `NativeApiLogsResource` is mounted at `/apis/{apiId}/logs/native` with only an `API_NATIVE_LOG:READ` check and no API-type assertion, `NativeApiLogReadUseCase` passes through, and `NativeApiMetricsFindQueryAdapter.adapt` matches on `api-id` + `request-id` only.

`security-token` is written for every v4 API, not only native ones (`freemarker/es9x/index/v4-metrics.ftl:149`), and for API-key plans it is the credential verbatim (`ApiKeyAuthenticationHandler.java:96`). So a user holding `API_NATIVE_LOG:READ` on an **HTTP v4** API with an API-key plan can call `GET /apis/{httpApiId}/logs/native/{requestId}` and get the subscriber's live API key back in `securityToken` — while `openapi-apis.yaml` documents that field as "Never a raw API key".

Apply the same discriminator here, or gate the resource on `ApiType.NATIVE`.

### :yellow_circle: Improvement

**3. The gamma guard keys on the wrong property, so it fails open on a native API-key plan**
`gravitee-gamma/.../ObservabilityLogsDataPortAdapter.java:224`

The discriminator is "is this a native document", but the property that actually makes the token safe is "the credential type is not one whose token *is* the secret". `PlanValidatorDomainService.validatePlanSecurityAgainstMode` places no restriction on `API_KEY` plans for `ApiType.NATIVE` (it only forbids `PUSH` mode, line 77), so a native Kafka API can carry an API-key plan. And the invariant the comment leans on — `KafkaClientCredential` never writing the key — lives in `gravitee-reactor-native-kafka`, not in this repository, so no test here would catch a regression in it.

An allow-list on `securityType` (emit only for `JWT` / `OAUTH2`, drop otherwise) fails closed and does not depend on another repo's behaviour.

### :white_circle: Nit

**4.** `FilterValuesQueryServiceImpl.java:119` — the two new ES field names are hardcoded literals duplicating `NativeApiMetricKeys.CLIENT_ID` / `CLIENT_SOFTWARE_NAME`, which `SearchMetricsQueryAdapter.java:33` composes from the constants. If a key is renamed there, this switch keeps the stale path and the values dropdown silently returns an empty page rather than failing. The file already hardcodes other fields, so this is consistency-level, not a defect.

### Checked and clean

- `NativeMetrics` arity change (5→7) and positional construction — ordering matches the component list, including the legacy-duration fallback branch.
- `LogDetailDto.from` / `LogEntryDto.from` positional record construction lines up with the record declarations.
- `AnalyticsMeasuresAdapter` and `AnalyticsDefinitionMapper` both map `NATIVE_CLIENT_SOFTWARE_NAME → THROW_EXCEPTION`, matching the four existing LOGS-only filters.
- `NATIVE_CLIENT_ID` exists in the analytics engine (`Filter.Name`, `Facet`, `NativeApiFieldResolver`), so promoting it to `ANALYTICS` is backed by a real dimension.
- `searchFilterValues` aggregates over the `V4_METRICS` wildcard index, so both new keyword filters have values to list.
- `io.gravitee.apim.core.analytics.model.ApiMetricsDetail` was **not** given the new fields, so `/apis/{id}/analytics/requests/{requestId}` does not leak the token.

callaertanthony:
### Re-review at `24bd3d8`

All four findings from the previous pass are resolved. Verified against the PR head, not a local checkout.

| Previously raised | Status at `24bd3d8` |
| --- | --- |
| `FilterName` in `openapi-logs.yaml` missing both `NATIVE_*` values, so picking either chip fails the whole search | Fixed. Parsed both files at this sha: the 18 catalog entries declaring `signals: [LOGS]` are all present in the 22-value enum |
| Gamma adapter returned the raw API key on a native document carrying an API-key plan | Fixed, and better than what I suggested — keyed on the credential type rather than on "is this a native document", so it fails closed |
| `NativeApiLogCrudServiceImpl` projected `securityToken` with no guard | Fixed, with a test that feeds an `API_KEY` document |
| ES field names hardcoded, duplicating `NativeApiMetricKeys` | Fixed, composed from the shared keys |

The security claim underneath `SecurityTokenProjection` holds. `CheckSubscriptionPolicy` (oauth2) and `JwtPlanBasedAuthenticationHandler` both call `setSecurityToken(clientId)`; only `ApiKeyAuthenticationHandler` passes the credential itself. And `v4-metrics.ftl` writes `${metrics.getSecurityType()}`, which renders the enum name, so the `"OAUTH2"` / `"JWT"` literals in the allow-list match what actually lands in Elasticsearch.

`every_logs_filter_the_catalog_publishes_should_exist_in_the_wire_enum` is the right shape for the invariant. `getFilters` filters on `anyMatch(filter::appliesTo)`, so `[LOGS, ANALYTICS]` entries like `NATIVE_CLIENT_ID` are covered, not just the LOGS-only ones — and the `isNotEmpty()` guard stops it passing vacuously if the YAML ever fails to load off the test classpath.

Two things left, neither a blocker.

---

**1. Detached javadoc** — `FilterValuesQueryServiceImpl.java:89` (nit)

`ADDITIONAL_METRICS_PREFIX` landed between the javadoc and the method it documents:

```java
/**
 * Maps domain filter names to Elasticsearch field names.
 * Mirrors the mapping in HTTPFieldResolver (ES module) to keep the repository API simple (String field name).
 */
private static final String ADDITIONAL_METRICS_PREFIX = "additional-metrics.";

private static String resolveEsFieldName(FilterSpec.Name filterName) {
```

That comment now documents a string literal, and `resolveEsFieldName` is undocumented. Moving the constant above the javadoc block restores both.

---

**2. `securityType` is now returned for every document type** — `ObservabilityLogsDataPortAdapter.java:222` (improvement)

`securityType` moved out of the native-document guard and is now set unconditionally:

```java
builder.securityType(metrics.getSecurityType());
if (SecurityTokenProjection.isTokenProjectable(metrics.getSecurityType())) {
    builder.securityToken(metrics.getSecurityToken());
}
```

So the ESM log detail of a plain HTTP v4 request now reports its credential type. That is not a leak — a type name is not a credential — and the field is new in this PR, so nothing regressed. But the widening is only visible as a flipped assertion in `should_not_expose_the_security_credential_on_a_non_native_document` (`null` → `"API_KEY"`), and the PR title scopes this to native API logs. Worth a line in the description, or scoping it back to native documents if it was incidental.

callaertanthony (approved):

PR #19251, opened 2026-08-19
fix(console): default Period to last 5 minutes instead of none

pragsty:
## Issue

https://gravitee.atlassian.net/browse/APIM-14961

## Description
v4 API Runtime Logs defaulted Period to **None**, so Console called `GET .../apis/{apiId}/logs` with no `from`/`to`. Elasticsearch searched the full retention window and large tenants timed out (`NoStackTraceTimeoutException`).
**Runtime Logs (Console)**
- Default Period is **Last 5 Minutes** (`-5m`). First load sends a bounded `from`/`to`.
- **None** stays available. The URL stores `period=0` so refresh and shared links keep None and still omit dates (unbounded API search).
- Custom From/To still force the None sentinel so a reload does not drop the dates and recompute `-5m`.
- Clearing the period chip, or the last custom From/To chip, restores Last 5 Minutes instead of unbounded None.
- Pagination and page-size reuse the last searched `from`/`to` instead of calling `moment()` again. Refresh still recomputes “now” and resets to page 1.
- Error-keys load from the same store update as the table (no eager first call from the store seed). Bookmarked None / custom range / other periods no longer ask error-keys for 5 minutes first.
**Webhook Logs**
- Still default to **None**. `DEFAULT_PERIOD` is now Last 5 Minutes, so webhook code is retargeted to `NONE_PERIOD`. No product change on that page.

## Additional context
**Compatibility:** old Runtime Logs URLs with no `period` used to mean None (unbounded). They now mean Last 5 Minutes. Use `period=0` to keep None.
**Not affected:** Webhook Logs default, native Kafka logs, Management API contract.

https://github.com/user-attachments/assets/aad849db-a01a-4c1b-9236-f581f03ee061

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs/api-runtime-logs.component.harness.ts
+++ b/gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs/api-runtime-logs.component.harness.ts
@@ -115,6 +115,10 @@ export class ApiRuntimeLogsHarness extends ComponentHarness {
       .then(btn => btn.click());
   }
 
+  async getMoreFiltersForm() {
+    return this.moreFiltersHarness().then(moreFilters => moreFilters.getFormOptional());
+  }
+
   async selectPeriodFromMoreFilters() {
     return this.moreFiltersHarness().then(quickFilters => quickFilters.getPeriodSelectInput());
   }
--- a/gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs/api-runtime-logs.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/api/api-traffic-v4/runtime-logs/api-runtime-logs.component.spec.ts
@@ -54,6 +54,8 @@ type ComponentInitData = {
   plans?: PlanV4[] | undefined;
   areLogsEnabled?: boolean;
   apiModifier?: Partial<ApiV4>;
+  from?: number;
+  to?: number;
 };
 
 const ENTRYPOINT_LIST: ConnectorPlugin[] = [
@@ -92,6 +94,12 @@ describe('ApiRuntimeLogsComponent', () => {
   const fromDateTime = new Date(fromDate).getTime();
   const toDate = '2023-10-24 15:21:00';
   const toDateTime = new Date(toDate).getTime();
+  const FAKE_NOW_MS = new Date('2023-10-05T00:00:00.000Z').getTime();
+  const DEFAULT_FROM_MS = FAKE_NOW_MS - 5 * 60 * 1000;
+  const DEFAULT_LOGS_TIME = { from: DEFAULT_FROM_MS, to: FAKE_NOW_MS };
+  const DEFAULT_QUERY_TIME = { period: '-5m', ...DEFAULT_LOGS_TIME };
+  let dateNowSpy: jest.SpyInstance;
+  let momentNowSpy: jest.SpyInstance;
 
   const initComponent = async () => {
     TestBed.configureTestingModule({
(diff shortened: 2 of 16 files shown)


aqibmohammadkhan (commented):
1. Correctness  — quick-filters-store.service.ts:68: periodTimeRange() uses logFilters.from ? ... : computed.from instead of ??. An explicit from/to of 0 (epoch) gets silently dropped and replaced by the period-derived value — three lines away in the same function, the sibling frozenTimeRange branch correctly uses ??, so this looks like an oversight, not a choice.
2. Altitude  — the "None" period option stays fully reachable and still returns an unbounded query; the PR fixes the default but not the underlying code path that caused the original timeout.
3. Reuse  — Webhook Logs' paginationUpdated() wasn't given the new lastSearchTimeRange freeze mechanism, so it still recomputes moment() on every page turn — the same drift bug this PR fixed for Runtime Logs.
4. Test coverage  — a deleted expect.assertions(2) in api-runtime-logs.component.spec.ts:529 weakens a test that verifies the more-filters panel isn't rendered too early.
5. 5–6. Simplification — the from/to derivation logic and the "no period + from/to present ⇒ None" precedence rule are each independently re-derived in two places, requiring lockstep updates.
6. Reuse — preparePeriodFilter() (runtime-logs) and preparePeriodRange() (webhook-logs) remain near-duplicate period-parsing code, and this PR made their guards diverge further rather than converging them.
7. Test coverage — removeFilter()'s new branching and the new frozenTimeRange param have no dedicated unit tests, only indirect coverage via the larger component spec.

pragsty:
> 1. Correctness  — quick-filters-store.service.ts:68: periodTimeRange() uses logFilters.from ? ... : computed.from instead of ??. An explicit from/to of 0 (epoch) gets silently dropped and replaced by the period-derived value — three lines away in the same function, the sibling frozenTimeRange branch correctly uses ??, so this looks like an oversight, not a choice.
> 2. Altitude  — the "None" period option stays fully reachable and still returns an unbounded query; the PR fixes the default but not the underlying code path that caused the original timeout.
> 3. Reuse  — Webhook Logs' paginationUpdated() wasn't given the new lastSearchTimeRange freeze mechanism, so it still recomputes moment() on every page turn — the same drift bug this PR fixed for Runtime Logs.
> 4. Test coverage  — a deleted expect.assertions(2) in api-runtime-logs.component.spec.ts:529 weakens a test that verifies the more-filters panel isn't rendered too early.
> 5. 5–6. Simplification — the from/to derivation logic and the "no period + from/to present ⇒ None" precedence rule are each independently re-derived in two places, requiring lockstep updates.
> 6. Reuse — preparePeriodFilter() (runtime-logs) and preparePeriodRange() (webhook-logs) remain near-duplicate period-parsing code, and this PR made their guards diverge further rather than converging them.
> 7. Test coverage — removeFilter()'s new branching and the new frozenTimeRange param have no dedicated unit tests, only indirect coverage via the larger component spec.

Fixed the rest except:
2. None stays unbounded — true, and we keep it as it is, otherwise we need BE changes, recommended was to copy eixsitng use case behaviour which caps it to 24 hours but then UI will lie & we'll end up with Imact- Automation, GIOs, and Portal-adjacent clients that called this endpoint with no interval will silently lose data.

3. Webhook pagination still calls moment() — true, out of scope

6. preparePeriodFilter vs preparePeriodRange — true, expected
They were already twins. Webhook’s guard still means “is this None?”; the identifier changed because DEFAULT_PERIOD is now -5m. Converging them is a refactor across two pages.(webhook logs not in scope)

aqibmohammadkhan (approved):
as discussed, please create a ticket to track the webhook log and env log es unbounded issue, similar to what we have here.

pragsty:
> as discussed, please create a ticket to track the webhook log and env log es unbounded issue, similar to what we have here.

Ticket is created - https://gravitee.atlassian.net/browse/APIM-14976

PR #19252, opened 2026-08-19
fix(users): purge IdP claims when whitelist is cleared

wbabyte:
## Issue

https://gravitee.atlassian.net/browse/APIM-14840

## Description

The persisted-claims whitelist governed what got **written** to `users.idp_claims`, but nothing governed what **stayed** written. Removing a claim from the whitelist left the already-captured value on every user indefinitely — across logins, restarts and further configuration changes — which defeats the premise of APIM-13950 that sensitive or unnecessary claims are not retained. No REST surface exposes or deletes `idpClaims` either (`UserService.findIdpClaims` is internal, consumed only by `ApplicationServiceImpl`), so direct database access was the only cleanup path.

`UserServiceImpl.persistWhitelistedClaims` returned early on a null/empty whitelist, so the stored map was never rewritten.

**Fix:** treat an absent or empty whitelist as *"retain nothing"* rather than *"do nothing"*.

Two details worth reviewing:

- Claims are cleared to **`null`, not an empty map**. `JdbcUserRepository.serializeIdpClaims` returns null only for null; an empty map serializes verbatim, which would leave the column populated with `{}` rather than cleared. Mongo behaves the same.
- The write is **skipped when nothing changes** (`!Objects.equals(...)`). This path now runs on every login, and most deployments configure no whitelist at all, so without the guard the common case would gain a redundant `update` per login. `UserEntity` carries no `idpClaims` field (only the repository `User` does), so the `findById` cannot be avoided without widening an `Indexable`, REST-serialized model — not worth it here, especially as every social login already performs several read/writes on that row.

A populated whitelist whose claims are simply absent from the token still stores an empty map, unchanged — that is APIM-13951's contract and is covered by an existing test.

### Scope

Purging takes effect on the user's **next login**. Users who never log in again keep their stored claims. Fully remediating those needs an admin-facing read/delete of `idpClaims` or a purge triggered on configuration change — deliberately out of scope here, and worth its own ticket if wanted. APIM-13951's final acceptance criterion ("not exposed unless explicitly requested with a permission check") anticipated such an endpoint and none was implemented.

### Test note

`shouldNotPersistClaimsWhenNoWhitelistConfigured` asserted the previous behaviour deliberately, with a sentinel claim and a comment noting that removing the guard would overwrite it. Rather than quietly rewriting a committed test, calling it out: that test pinned the defect, so it is replaced by `shouldPurgePersistedClaimsWhenWhitelistCleared` / `...WhenWhitelistNotConfigured`, plus `shouldNotUpdateUserWhenThereAreNoClaimsToPurge` covering the write guard.

RED → GREEN verified: both purge tests failed with `expected: <null> but was: <{org_id=org-42}>` before the change. Full `gravitee-apim-rest-api-service` suite green afterwards — 9338 tests, 0 failures.

## Additional context

⚠️ **Security-sensitive**: changes what PII-bearing IdP claim data is retained on the user record (`UserMongo.idpClaims` is `@ToString.Exclude`d as PII for exactly this reason). No change to authentication or authorization logic.

Found during QA of epic APIM-13949. Sibling bug APIM-14841 (OAuth2 client authentication method mismatch) and the reopened UI story APIM-13954 follow in separate PRs.

Reproduction (from the ticket): configure an OIDC provider with `persistedClaimsWhitelist: ["org_id"]`, log in as a federated user, clear the whitelist via `PUT .../configuration/identities/{id}`, log in again — stored claims were previously byte-identical, and are now removed.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
@@ -1891,7 +1891,8 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
 
     /**
      * Persists the IdP claims whitelisted on the social identity provider onto the user record, refreshing them on
-     * every login. When no whitelist is configured the feature is inactive and the user record is left untouched.
+     * every login. An absent or empty whitelist means nothing may be retained, so previously captured claims are
+     * removed instead of being left behind (APIM-14840).
      */
     private void persistWhitelistedClaims(
         List<String> whitelist,
@@ -1900,16 +1901,21 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
         String accessTokenPayload,
         String idTokenPayload
     ) {
-        if (whitelist == null || whitelist.isEmpty()) {
-            return;
-        }
-        Map<String, String> idpClaims = extractWhitelistedClaims(whitelist, userInfo, accessTokenPayload, idTokenPayload);
+        // Null rather than an empty map: both backends serialize an empty map verbatim, which would leave the stored
+        // claims populated with "{}" instead of cleared.
+        Map<String, String> idpClaims = whitelist == null || whitelist.isEmpty()
+            ? null
+            : extractWhitelistedClaims(whitelist, userInfo, accessTokenPayload, idTokenPayload);
         try {
             Optional<User> optionalUser = userRepository.findById(userId);
             if (optionalUser.isPresent()) {
                 User user = optionalUser.get();
-                user.setIdpClaims(idpClaims);
-                userRepository.update(user);
+                // Every login reaches this point and most deployments configure no whitelist at all, so without this
+                // guard the common case would add a redundant write to each login.
+                if (!Objects.equals(user.getIdpClaims(), idpClaims)) {
+                    user.setIdpClaims(idpClaims);
+                    userRepository.update(user);
+                }
             }
         } catch (TechnicalException e) {
             log.warn("Unable to persist IdP claims for user {}", userId, e);
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
@@ -1892,7 +1892,10 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
     /**
      * Persists the IdP claims whitelisted on the social identity provider onto the user record, refreshing them on
      * every login. An absent or empty whitelist means nothing may be retained, so previously captured claims are
-     * removed instead of being left behind (APIM-14840).
+     * removed at the user's next login (APIM-14840).
+     *
+     * <p>Removal is therefore only as timely as that login: a user who never authenticates again keeps whatever was
+     * captured, and no bulk purge or administrative delete exists to reach them.
      */
     private void persistWhitelistedClaims(
         List<String> whitelist,
@@ -1901,24 +1904,31 @@ public class UserServiceImpl extends AbstractService implements UserService, Ini
         String accessTokenPayload,
         String idTokenPayload
     ) {
-        // Null rather than an empty map: both backends serialize an empty map verbatim, which would leave the stored
-        // claims populated with "{}" instead of cleared.
+        // Null, not an empty map: User.setIdpClaims treats null as "no claims stored" and an empty map as a stored
+        // empty value, and only the former clears the column.
         Map<String, String> idpClaims = whitelist == null || whitelist.isEmpty()
             ? null
             : extractWhitelistedClaims(whitelist, userInfo, accessTokenPayload, idTokenPayload);
         try {
             Optional<User> optionalUser = userRepository.findById(userId);
-            if (optionalUser.isPresent()) {
-                User user = optionalUser.get();
-                // Every login reaches this point and most deployments configure no whitelist at all, so without this
-                // guard the common case would add a redundant write to each login.
-                if (!Objects.equals(user.getIdpClaims(), idpClaims)) {
-                    user.setIdpClaims(idpClaims);
-                    userRepository.update(user);
-                }
+            if (optionalUser.isEmpty()) {
+                // The user was created or refreshed a few statements ago, so this means concurrent deletion or replica
+                // lag. On a retention path a silent skip would leave no trace that the purge was even attempted.
+                log.warn("Cannot apply the IdP claims whitelist: user {} was not found", userId);
+                return;
+            }
+
+            User user = optionalUser.get();
+            // Every login reaches this point and most deployments configure no whitelist at all, so without this
+            // guard the common case would add a redundant write to each login.
+            if (!Objects.equals(user.getIdpClaims(), idpClaims)) {
+                user.setIdpClaims(idpClaims);
+                userRepository.update(user);
             }
         } catch (TechnicalException e) {
-            log.warn("Unable to persist IdP claims for user {}", userId, e);
+            // Deliberately not rethrown: a claims failure must not fail the login. Logged as a retention failure rather
+            // than a persistence one, because when the whitelist was cleared the claims are still stored.
+            log.error("Unable to apply the IdP claims whitelist for user {}; previously stored claims may remain", userId, e);
         }
     }
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
@@ -1813,14 +1813,35 @@ public class UserServiceTest {
     }
 
     @Test
-    public void shouldNotPersistClaimsWhenNoWhitelistConfigured() throws IOException, TechnicalException {
+    public void shouldPurgePersistedClaimsWhenWhitelistCleared() throws IOException, TechnicalException {
         reset(identityProvider, userRepository);
         mockDefaultEnvironment();
 
         User user = mockUser();
-        // Pre-set a sentinel so the assertion actually pins the no-op: if the whitelist guard were removed, the code
-        // would overwrite this with an empty map. mockUser() leaves idpClaims null, which would make assertNull vacuous.
-        user.setIdpClaims(Map.of("pre_existing", "value"));
+        user.setIdpClaims(Map.of("org_id", "org-42"));
+        when(userRepository.findBySource(null, user.getSourceId(), ORGANIZATION)).thenReturn(Optional.of(user));
+        when(userRepository.findById(user.getId())).thenReturn(Optional.of(user));
+        when(userRepository.update(any(User.class))).thenAnswer(returnsFirstArg());
+        when(identityProvider.getPersistedClaimsWhitelist()).thenReturn(Collections.emptyList());
+
+        String userInfo = IOUtils.toString(read("/oauth2/json/user_info_response_body.json"), Charset.defaultCharset());
+        String accessToken = IOUtils.toString(read("/oauth2/jwt/access_token.jwt"), Charset.defaultCharset());
+        String idToken = IOUtils.toString(read("/oauth2/jwt/id_token.jwt"), Charset.defaultCharset());
+
+        userService.createOrUpdateUserFromSocialIdentityProvider(EXECUTION_CONTEXT, identityProvider, userInfo, accessToken, idToken);
+
+        // Nothing whitelisted means nothing retained, so an administrator who clears the whitelist to stop capturing a
+        // sensitive claim also gets the already-captured value removed (APIM-14840)
+        assertNull(user.getIdpClaims());
+    }
+
+    @Test
+    public void shouldPurgePersistedClaimsWhenWhitelistNotConfigured() throws IOException, TechnicalException {
+        reset(identityProvider, userRepository);
+        mockDefaultEnvironment();
+
+        User user = mockUser();
+        user.setIdpClaims(Map.of("org_id", "org-42"));
         when(userRepository.findBySource(null, user.getSourceId(), ORGANIZATION)).thenReturn(Optional.of(user));
         when(userRepository.findById(user.getId())).thenReturn(Optional.of(user));
         when(userRepository.update(any(User.class))).thenAnswer(returnsFirstArg());
@@ -1831,8 +1852,28 @@ public class UserServiceTest {
 
         userService.createOrUpdateUserFromSocialIdentityProvider(EXECUTION_CONTEXT, identityProvider, userInfo, accessToken, idToken);
 
-        // No whitelist configured → the persistence path is a no-op and leaves the existing claims untouched
-        assertEquals(Map.of("pre_existing", "value"), user.getIdpClaims());
+        assertNull(user.getIdpClaims());
+    }
+
+    @Test
+    public void shouldNotUpdateUserWhenThereAreNoClaimsToPurge() throws IOException, TechnicalException {
+        reset(identityProvider, userRepository);
+        mockDefaultEnvironment();
+
+        User user = mockUser();
+        when(userRepository.findBySource(null, user.getSourceId(), ORGANIZATION)).thenReturn(Optional.of(user));
+        when(userRepository.findById(user.getId())).thenReturn(Optional.of(user));
+        when(userRepository.update(any(User.class))).thenAnswer(returnsFirstArg());
+
+        String userInfo = IOUtils.toString(read("/oauth2/json/user_info_response_body.json"), Charset.defaultCharset());
+        String accessToken = IOUtils.toString(read("/oauth2/jwt/access_token.jwt"), Charset.defaultCharset());
+        String idToken = IOUtils.toString(read("/oauth2/jwt/id_token.jwt"), Charset.defaultCharset());
+
+        userService.createOrUpdateUserFromSocialIdentityProvider(EXECUTION_CONTEXT, identityProvider, userInfo, accessToken, idToken);
+
+        // The common case is no whitelist and no stored claims: purging must not cost an extra write on every login,
+        // so the only update is the one the profile refresh already performs
+        verify(userRepository, times(1)).update(any(User.class));
     }
 
     @Test
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
@@ -1833,6 +1833,9 @@ public class UserServiceTest {
         // Nothing whitelisted means nothing retained, so an administrator who clears the whitelist to stop capturing a
         // sensitive claim also gets the already-captured value removed (APIM-14840)
         assertNull(user.getIdpClaims());
+        // Production mutates the instance the findById stub returns, so asserting on it only proves the setter ran.
+        // The second write is what proves the cleared value reached the repository: one from the profile refresh, one here.
+        verify(userRepository, times(2)).update(any(User.class));
     }
 
     @Test
@@ -1845,6 +1848,8 @@ public class UserServiceTest {
         when(userRepository.findBySource(null, user.getSourceId(), ORGANIZATION)).thenReturn(Optional.of(user));
         when(userRepository.findById(user.getId())).thenReturn(Optional.of(user));
         when(userRepository.update(any(User.class))).thenAnswer(returnsFirstArg());
+        // Stubbed explicitly: an unstubbed List-returning mock yields an empty list, which is the other branch
+        when(identityProvider.getPersistedClaimsWhitelist()).thenReturn(null);
 
         String userInfo = IOUtils.toString(read("/oauth2/json/user_info_response_body.json"), Charset.defaultCharset());
         String accessToken = IOUtils.toString(read("/oauth2/jwt/access_token.jwt"), Charset.defaultCharset());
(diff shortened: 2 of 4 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
```diff
@@ -1813,26 +1813,67 @@ public void shouldPersistWhitelistedClaimsOnLogin() throws IOException, Technica
...
 
-        // No whitelist configured → the persistence path is a no-op and leaves the existing claims untouched
-        assertEquals(Map.of("pre_existing", "value"), user.getIdpClaims());
+        // Nothing whitelisted means nothing retained, so an administrator who clears the whitelist to stop capturing a
+        // sensitive claim also gets the already-captured value removed (APIM-14840)
+        assertNull(user.getIdpClaims());
+    }
+
+    @Test
+    public void shouldPurgePersistedClaimsWhenWhitelistNotConfigured() throws IOException, TechnicalException {
```
paulatulis, line 1842: This test never exercises the branch it is named for. `identityProvider` is a `@Mock` (line 181) and `getPersistedClaimsWhitelist()` returns `List<String>`, so the unstubbed call returns Mockito's default **empty list** (`ReturnsEmptyValues`), not `null`. The test therefore takes the `whitelist.isEmpty()` path and is an exact duplicate of `shouldPurgePersistedClaimsWhenWhitelistCleared` above it.

The `whitelist == null` half of the new condition at `UserServiceImpl:1906` has no coverage — which matters here because "not configured" is the common deployment state, and it is the case the PR description singles out.

Stubbing it explicitly — `when(identityProvider.getPersistedClaimsWhitelist()).thenReturn(null);` — makes the two tests cover genuinely distinct branches.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
```diff
@@ -1813,26 +1813,67 @@ public void shouldPersistWhitelistedClaimsOnLogin() throws IOException, Technica
...
         String accessToken = IOUtils.toString(read("/oauth2/jwt/access_token.jwt"), Charset.defaultCharset());
         String idToken = IOUtils.toString(read("/oauth2/jwt/id_token.jwt"), Charset.defaultCharset());
 
         userService.createOrUpdateUserFromSocialIdentityProvider(EXECUTION_CONTEXT, identityProvider, userInfo, accessToken, idToken);
 
-        // No whitelist configured → the persistence path is a no-op and leaves the existing claims untouched
-        assertEquals(Map.of("pre_existing", "value"), user.getIdpClaims());
+        // Nothing whitelisted means nothing retained, so an administrator who clears the whitelist to stop capturing a
+        // sensitive claim also gets the already-captured value removed (APIM-14840)
+        assertNull(user.getIdpClaims());
```
paulatulis, line 1835: Both purge tests (here and line 1855) assert only on the in-memory `user` instance — the same object the `findById` stub hands back, which production mutates via `setIdpClaims(null)` *before* calling `userRepository.update(user)`.

That means the assertion passes as soon as the setter runs. Delete `userRepository.update(user)` at `UserServiceImpl:1917` and both tests stay green while nothing is actually purged from the store — i.e. the original bug survives a fully green suite. An `ArgumentCaptor` would not close this either, since it captures that same mutated reference.

Adding `verify(userRepository, times(2)).update(any(User.class));` (one write from the profile refresh, one from the purge) pins that the purge actually reaches the repository — the same technique `shouldNotUpdateUserWhenThereAreNoClaimsToPurge` already relies on.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -1900,16 +1901,21 @@ private void persistWhitelistedClaims(
...
+            : extractWhitelistedClaims(whitelist, userInfo, accessTokenPayload, idTokenPayload);
         try {
             Optional<User> optionalUser = userRepository.findById(userId);
             if (optionalUser.isPresent()) {
                 User user = optionalUser.get();
-                user.setIdpClaims(idpClaims);
-                userRepository.update(user);
+                // Every login reaches this point and most deployments configure no whitelist at all, so without this
+                // guard the common case would add a redundant write to each login.
+                if (!Objects.equals(user.getIdpClaims(), idpClaims)) {
```
paulatulis: The write guard leaves one hole open, in the case this fix exists to close.

A corrupt or unreadable `idp_claims` value degrades to `null` on read in **both** backends — `JdbcUserRepository.deserializeIdpClaims` throws and the ORM's `setFromResultSet` "skips the column, so a corrupt/legacy value degrades to null" (your own comment there), and `MongoUserRepository.toModel` catches and sets null.

So for a user whose stored blob is corrupt, with the whitelist cleared: computed value is `null`, stored value reads back `null`, `Objects.equals` matches, the update is skipped — and the PII-bearing row stays in the database permanently, with the admin believing it was purged.

Not a regression (the old early-return never wrote on that path either), and narrow enough that it may be worth accepting knowingly rather than fixing — but it is the one input where "clear the whitelist" still fails to clear.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -1900,16 +1901,21 @@ private void persistWhitelistedClaims(
         String accessTokenPayload,
         String idTokenPayload
     ) {
-        if (whitelist == null || whitelist.isEmpty()) {
-            return;
-        }
-        Map<String, String> idpClaims = extractWhitelistedClaims(whitelist, userInfo, accessTokenPayload, idTokenPayload);
+        // Null rather than an empty map: both backends serialize an empty map verbatim, which would leave the stored
```
paulatulis: This comment is reasoning about repository internals from inside the service layer — the null-vs-empty-map distinction is meaningful only because of how `JdbcUserRepository.serializeIdpClaims` and the Mongo mapper each encode emptiness.

I verified the claim is accurate today: `serializeIdpClaims` returns NULL for null and `"{}"` for an empty map, and `MongoUserRepository.encodeClaimKeys` returns null for null. The problem is that nothing enforces it. `User.idpClaims` documents what the field means, not that a caller must pass `null` rather than `Map.of()` to get a cleared result — so a third backend (or a refactor of either existing one) that treated the two identically would silently break the purge, with no compile-time or interface-level signal and no failing test.

Worth either documenting the contract on `User.idpClaims` / `UserRepository` and pointing this comment at it, or normalizing at the repository boundary so the service can just express "clear the claims".

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -1891,7 +1891,8 @@ public Map<String, String> findIdpClaims(ExecutionContext executionContext, Stri
 
     /**
      * Persists the IdP claims whitelisted on the social identity provider onto the user record, refreshing them on
```
paulatulis, line 1893: Scope confirmation rather than a defect — the PR description and the ticket both call this out, so flagging it only so it does not get lost.

The javadoc states previously captured claims "are removed", but removal is triggered by the user's next social login. Users who never log in again keep their claims indefinitely, and there is no bulk purge on configuration change and no admin-facing read or delete of `idpClaims`. An administrator who clears the whitelist specifically because a sensitive claim was being captured therefore gets a partial remediation.

APIM-13951's final acceptance criterion ("not exposed unless explicitly requested with a permission check") anticipated such an endpoint and none was implemented. Worth its own ticket so the gap is tracked rather than living only in this PR description — and possibly worth softening the javadoc so it does not read as an unconditional guarantee.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -1900,16 +1901,21 @@ private void persistWhitelistedClaims(
...
-        }
-        Map<String, String> idpClaims = extractWhitelistedClaims(whitelist, userInfo, accessTokenPayload, idTokenPayload);
+        // Null rather than an empty map: both backends serialize an empty map verbatim, which would leave the stored
+        // claims populated with "{}" instead of cleared.
+        Map<String, String> idpClaims = whitelist == null || whitelist.isEmpty()
+            ? null
+            : extractWhitelistedClaims(whitelist, userInfo, accessTokenPayload, idTokenPayload);
         try {
             Optional<User> optionalUser = userRepository.findById(userId);
             if (optionalUser.isPresent()) {
```
paulatulis: The `isEmpty()` case falls through with no `else` and no logging at all. The user was created or refreshed a few statements earlier, so an empty `Optional` here signals something real — concurrent deletion, or replica lag.

On a path whose whole purpose is a data-retention guarantee, that means the purge can silently not happen with zero trace that it was even attempted.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -1900,16 +1901,21 @@ private void persistWhitelistedClaims(
...
-                userRepository.update(user);
+                // Every login reaches this point and most deployments configure no whitelist at all, so without this
+                // guard the common case would add a redundant write to each login.
+                if (!Objects.equals(user.getIdpClaims(), idpClaims)) {
+                    user.setIdpClaims(idpClaims);
+                    userRepository.update(user);
+                }
             }
         } catch (TechnicalException e) {
             log.warn("Unable to persist IdP claims for user {}", userId, e);
```
paulatulis: This message now mislabels the case it most needs to describe. Since this method also purges, "Unable to persist IdP claims" reads as "nothing was saved" when the actual state is the opposite and worse: **sensitive claims are still stored** despite an administrator having cleared the whitelist to remove them.

The catch also spans both `findById` and `update`, so a read failure and a write failure are indistinguishable in the log. And at `warn` on a privacy control, this sits below most alerting thresholds while login still returns success — the operator gets no signal that the purge failed.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
```diff
@@ -1891,7 +1891,8 @@ public Map<String, String> findIdpClaims(ExecutionContext executionContext, Stri
 
     /**
      * Persists the IdP claims whitelisted on the social identity provider onto the user record, refreshing them on
-     * every login. When no whitelist is configured the feature is inactive and the user record is left untouched.
+     * every login. An absent or empty whitelist means nothing may be retained, so previously captured claims are
+     * removed instead of being left behind (APIM-14840).
```
paulatulis: Nit: `(APIM-14840)` in the javadoc, and the same ref in the test comment at `UserServiceTest:1834`. Repo convention keeps ticket ids out of code comments — the linkage lives in the branch, PR, and commit message.

The sentences read fine without them.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
```diff
@@ -1813,26 +1813,67 @@ public void shouldPersistWhitelistedClaimsOnLogin() throws IOException, Technica
...
+
+        String userInfo = IOUtils.toString(read("/oauth2/json/user_info_response_body.json"), Charset.defaultCharset());
+        String accessToken = IOUtils.toString(read("/oauth2/jwt/access_token.jwt"), Charset.defaultCharset());
+        String idToken = IOUtils.toString(read("/oauth2/jwt/id_token.jwt"), Charset.defaultCharset());
+
+        userService.createOrUpdateUserFromSocialIdentityProvider(EXECUTION_CONTEXT, identityProvider, userInfo, accessToken, idToken);
+
+        // The common case is no whitelist and no stored claims: purging must not cost an extra write on every login,
+        // so the only update is the one the profile refresh already performs
+        verify(userRepository, times(1)).update(any(User.class));
```
paulatulis, line 1956: This assertion is sound — I checked that the refresh path performs exactly one `userRepository.update`, so `times(1)` genuinely fails if the guard is removed and a second purge write occurs. It is not vacuous.

But it only covers the `null == null` case. The guard's stated purpose is avoiding a redundant write per login, and the steady state it actually optimizes — whitelist configured, identical claims present on re-login, so the computed map equals the stored one — has no test. `shouldOverwriteClaimsOnReLogin` changes the whitelist between logins, so both of its writes legitimately differ.

A case with a stable whitelist and an unchanged token would pin the equal-non-null-maps path.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
```diff
@@ -1813,26 +1813,67 @@ public void shouldPersistWhitelistedClaimsOnLogin() throws IOException, Technica
...
     }
```
paulatulis, line 1957: No test covers the `TechnicalException` path at `UserServiceImpl:1920`. Two behaviours hang off it and neither is pinned: that a claim-persistence failure must not fail the login, and that the failure is reported.

That is the branch most likely to change meaning silently — the swallow is deliberate today, but nothing would catch a future edit that let the exception escape and started breaking social login.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
```diff
@@ -1813,26 +1813,67 @@ public void shouldPersistWhitelistedClaimsOnLogin() throws IOException, Technica
...
+        when(userRepository.update(any(User.class))).thenAnswer(returnsFirstArg());
+
+        String userInfo = IOUtils.toString(read("/oauth2/json/user_info_response_body.json"), Charset.defaultCharset());
+        String accessToken = IOUtils.toString(read("/oauth2/jwt/access_token.jwt"), Charset.defaultCharset());
+        String idToken = IOUtils.toString(read("/oauth2/jwt/id_token.jwt"), Charset.defaultCharset());
+
+        userService.createOrUpdateUserFromSocialIdentityProvider(EXECUTION_CONTEXT, identityProvider, userInfo, accessToken, idToken);
+
+        assertNull(user.getIdpClaims());
+    }
```
paulatulis, line 1936: Pre-existing gap rather than something this PR introduced, but it is an acceptance criterion on APIM-13951 and this PR is the one revisiting claim persistence.

`extractWhitelistedClaims` has a dedicated branch for non-scalar claim values (`value.isValueNode() ? value.asText() : value.toString()`), but every claim in the fixtures (`user_info_response_body.json`, `id_token.jwt`, `access_token.jwt`) is a scalar — so the JSON-serialization branch is never executed by any test in this file. Real IdPs commonly emit array claims such as `groups` or `roles`.

Adding an array claim to a fixture and whitelisting it would cover it.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
wbabyte: Confirmed and fixed. I checked the mechanism you describe: `getPersistedClaimsWhitelist()` returns `List<String>`, so the unstubbed mock went through `ReturnsEmptyValues` and handed back an empty list — the test took the `isEmpty()` path and was an exact duplicate of the one above it, leaving `whitelist == null` uncovered.

It now stubs `thenReturn(null)` explicitly, with a comment saying why, so the two tests cover distinct branches.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
wbabyte: Agreed, and this was the most valuable comment on the PR. The assertion did pass on the setter alone — production mutates the very instance the `findById` stub returns, so the write could have been deleted with the suite still green.

Both purge tests now add `verify(userRepository, times(2)).update(any(User.class))` — one write from the profile refresh, one from the purge — with a comment recording why asserting on the instance is not sufficient.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
wbabyte: You are right, and I am accepting it knowingly rather than fixing it here, so putting the reasoning on the record.

**Risk:** for a user whose `idp_claims` blob is corrupt, the value reads back as `null` in both backends, the computed value is also `null` once the whitelist is cleared, `Objects.equals` matches, the write is skipped, and the PII-bearing row survives while the administrator believes it was purged.

**Why not fixed here:** the service cannot distinguish "no claims" from "unreadable claims" — that information is lost at the repository boundary. Closing it properly means either normalising there or adding an unconditional purge path, and an unconditional write would undo the guard that keeps the common no-whitelist login from writing on every request.

**Mitigation / owner:** not a regression (the old early-return never wrote on that path either), and it needs the same admin-facing delete that the dormant-user gap needs. Both belong to the follow-up ticket noted on the `:1893` thread, which I have asked the requester to raise since I could not reach Jira in this session.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
wbabyte: Fixed, taking the first of your two options. The rule now lives on `User.idpClaims` as a documented contract:

> `null` and an empty map are not interchangeable, and callers rely on the difference [...] Every implementation of `UserRepository` must preserve that distinction, so a caller wanting to remove what was captured sets `null` rather than an empty map.

The service comment no longer reasons about how each backend encodes emptiness; it points at that contract instead. Thanks for verifying the claim was accurate today — that is exactly why it needed to be written down rather than relied upon.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
wbabyte: Agreed on both counts, and thanks for flagging it rather than assuming it was covered.

The javadoc is softened: it now says claims are removed **at the user's next login**, and adds that removal is only as timely as that login — a user who never authenticates again keeps whatever was captured, and no bulk purge or administrative delete exists to reach them.

On the ticket: I could not raise it, as the Atlassian connection dropped during this session. I have asked the requester to create it, covering this gap and the corrupt-value case from the `:1915` thread together, since both need the same admin-facing read/delete of `idpClaims` that APIM-13951's final criterion anticipated.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
wbabyte: Fixed. The `Optional` is now checked first and logs a warning naming the user before returning, with a comment recording your reasoning — the user was created or refreshed a few statements earlier, so an empty result means concurrent deletion or replica lag rather than an ordinary miss.

You are right that this mattered more than a usual missing-log: on a path whose only purpose is a retention guarantee, the purge could previously fail to happen with nothing recorded.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
wbabyte: All three points fixed. The message is now:

> `Unable to apply the IdP claims whitelist for user {}; previously stored claims may remain`

at `error` rather than `warn`, since you are right that a silently-failed privacy control sitting below alerting thresholds while login returns success is the wrong outcome. The wording now describes the state that actually obtains rather than implying nothing was written.

I have left the catch spanning both `findById` and `update`. Splitting it would let me name read versus write, but the message above is true either way, and the log carries the exception. Happy to split it if you would rather have the distinction.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/UserServiceImpl.java
wbabyte: Pushing back on this one, with evidence — I checked before changing anything.

Ticket ids in comments are established in this repository, not an exception. `git grep -E "APIM-[0-9]{4,5}" origin/master -- '**/*.java'` matches 21 files, and the closest precedent is the same epic, adjacent method:

- `IdentityProviderServiceImpl.java:299` — *"provider-defined free-form strings, so only emptiness is checked — APIM-13950"*
- `Signal.java:27`, `PageUtils.java:41`, `BaseConnectionLog.java:63`, `UsersResourceTest.java:65`, and others.

Since the surrounding code in this feature already carries them, removing mine would introduce the inconsistency rather than remove one, so I have kept them. Glad to strip them across the epic in a separate change if the team wants the convention to go the other way.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
wbabyte: Thanks for checking it was not vacuous — and you are right that it only covered `null == null`.

Added `shouldNotUpdateUserWhenTheWhitelistedClaimsAreUnchanged`, which is the steady state the guard exists for: a stable whitelist and an unchanged token across two logins. It asserts three writes in total (refresh + claims on the first login, refresh only on the second), so removing the guard makes it four and the test fails.

Your point about `shouldOverwriteClaimsOnReLogin` is right too — it changes the whitelist between logins, so both of its writes legitimately differ and it could never have covered this.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
wbabyte: Fixed. `shouldNotFailTheLoginWhenTheClaimsCannotBePersisted` stubs `update` to succeed once and then throw, so the profile refresh writes normally and the claims write fails. It asserts the login still returns a user and that both writes were attempted.

That pins both behaviours you named: the swallow is deliberate, and a future edit letting the exception escape would now break social login in the suite rather than in production.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
wbabyte: Fixed, and thanks for pointing at the fixtures rather than just the branch — every claim in them was a scalar, so `value.toString()` was dead code in tests.

Added a `groups` array to `user_info_response_body.json` and a test whitelisting it, asserting the stored value is `["platform-admin","api-reviewer"]` — the JSON form rather than a Java `toString`. Each module owns its own copy of that fixture, so the change is confined to `rest-api-service`.

paulatulis (approved):
Approving.

Answering the two you left open, so they don't sit:

- **Catch spanning `findById` and `update`** — leave it as is. The message is
  true either way and the exception carries the detail; splitting it buys a
  distinction no operator would act on differently.
- **Asserting the warning log line** — agreed, don't. Pinning log text is the
  wrong trade for its value, and you're consistent about this across the epic.
  I'd rather have that consistency than one asserted line here.

Two gaps we agreed to accept — the corrupt `idp_claims` blob that reads back as
null and skips the write, and dormant users who never log in again. Both need
the same admin-facing read/delete of `idpClaims` that APIM-13951's final
criterion anticipated. Please remember to raise that ticket.

PR #19254, opened 2026-08-19
fix(auth): align OAuth2 client authentication methods

wbabyte:
## Issue

https://gravitee.atlassian.net/browse/APIM-14841

## Description

The two OAuth2 login endpoints demanded **different** client authentication from the same identity provider:

| Endpoint | Credentials sent as | Method required |
|---|---|---|
| `POST .../auth/oauth2/{idp}` | `client_id` + `client_secret` in the token-request body | `client_secret_post` |
| `POST .../auth/oauth2/{idp}/exchange` | hardcoded `Authorization: Basic` on introspection | `client_secret_basic` |

A provider configured for one failed on the other, and the failure was opaque — the code flow returned an **empty 401**, the exchange path a **500**, with only `invalid_client` in the server log and nothing pointing at client authentication.

### The fix

A new `tokenEndpointAuthMethod` key on the identity provider's configuration, applied to **both** calls. Naming follows the OpenID Connect Discovery `token_endpoint_auth_method` value (already present as a String on the DCR DTOs), so a provider's own documentation transfers without translation.

**A provider that declares no method keeps the behaviour each endpoint has always had** — token via body, introspection via Basic. That is deliberate: introspection is only reachable through `/exchange`, only when `tokenIntrospectionEndpoint` is set, and no first-party console or portal code calls `/exchange`, so its only consumers are external integrators we cannot see — and for them Basic is currently the *only* thing that works. Converging both on `client_secret_post` would be cleaner but would silently break those integrators on upgrade, and there is no changelog/upgrade-note mechanism in this repo to warn them. The existing tests passing unchanged is the evidence: the legacy stubs assert the exact token-request body and the `Basic` introspection header, and both still match.

Stored in the `configuration` map rather than as a column, so **no migration** — `JdbcIdentityProviderRepository` persists `configuration` as a single JSON string and absent keys read back null. It is read in the common tail of `SocialIdentityProviderImpl.convert`, alongside `clientId`/`clientSecret`, so Gravitee AM providers get it too, not just OIDC.

### Two further defects fixed in the same paths

- **The 500.** The introspection path logged the response body and then passed the *already-consumed* JAX-RS client `Response.getEntity()` into the outgoing response, which cannot serialize. The body is now read once into a variable, logged, and turned into an error payload.
- **Opaque failures.** Both paths now answer with the provider's own error code, plus a hint naming `tokenEndpointAuthMethod` when the provider says `invalid_client`. Only the error *code* is relayed — never the provider's raw response, which may describe its internals.

An unrecognised configuration value is reported (warning naming the provider and the accepted values) and ignored rather than rejected, so a typo cannot lock every user out of login.

### Behaviour changes to be aware of

- Introspection failures now answer **401 instead of 500**.
- Both failure paths now carry a **JSON body**, where the code flow previously returned an empty 401.

### Scope note — widened beyond the ticket

The ticket names only the management resource. The portal resource (`portal-rest/.../auth/OAuth2AuthenticationResource.java`) carried the **identical** defect on both call sites, so fixing management alone would have left a known twin behind; it is fixed the same way here. Its `_exchange` endpoint had **no test coverage at all** — the test fixture's provider did not even expose an introspection endpoint, so the path was unreachable — and now has coverage.

Not fixed, flagged instead: `ClientCredentialsInitialAccessTokenProvider.java:81-86` hardcodes Basic in the same way, carrying a TODO that says exactly this ("we must being able to handle client authentication according to the `token_endpoint_auth_methods_supported` property from discovery"). That is a different configuration surface — client registration providers, not identity providers — so it is out of scope here.

## Additional context

⚠️ **Security-sensitive**: changes how client credentials are transmitted to an external identity provider, and changes error responses on an authentication path. The new messages relay only the provider's error code and a static hint; the client secret is never logged or returned.

⚠️ **Compatibility-sensitive**: adds a configuration key (`tokenEndpointAuthMethod`) and a new enum (`ClientAuthenticationMethod`). No schema change, no migration, no change to any generated model or OpenAPI contract.

### Verification

RED → GREEN observed on both modules:

- Management: 4 of 16 failed before the change — including `expected: <401> but was: <500>` on introspection and the empty 401 body — then 16/16.
- Portal: tests were written after the fix, so RED was verified by reverting the production file — 3 of the 4 new tests fail, including the same 500 — then 8/8. The fourth correctly passes either way, since it pins the legacy default.

Full suites for all four touched modules green: **236 / 9338 / 92 / 519 / 605, zero failures**.

Pre-existing behaviour, unrelated to epic APIM-13949 — surfaced while testing it. Sibling PR #19252 fixes APIM-14840; the reopened UI story APIM-13954 follows separately.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
@@ -248,10 +248,11 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
             ? identityProvider.getTokenEndpointAuthMethod()
             : endpointDefault;
 
-        // Switched rather than decided by exclusion: token_endpoint_auth_method has values such as none and
-        // private_key_jwt that must not fall through to putting the secret in the body, so adding a constant has to be
-        // a deliberate decision here rather than a silent change on a credential path.
+        // The null label makes this an enhanced switch (JLS 14.11.2), so the compiler requires it to be exhaustive:
+        // adding a token_endpoint_auth_method constant such as none or private_key_jwt fails the build here instead of
+        // silently doing nothing, or worse falling through to putting the secret in the body, on a credential path.
         switch (method) {
+            case null -> throw new IllegalArgumentException("No client authentication method resolved for " + identityProvider.getId());
             case CLIENT_SECRET_BASIC -> request.header(HttpHeaders.AUTHORIZATION, basicAuthorization(identityProvider));
             case CLIENT_SECRET_POST -> {
                 // The authorization-code flow already carries the client_id supplied by the caller. add() registers the
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
@@ -22,6 +22,7 @@ import io.gravitee.rest.api.management.rest.model.ExchangePayloadEntity;
 import io.gravitee.rest.api.management.rest.model.PayloadInput;
 import io.gravitee.rest.api.management.rest.utils.BlindTrustManager;
 import io.gravitee.rest.api.model.UserEntity;
+import io.gravitee.rest.api.model.configuration.identity.ClientAuthenticationMethod;
 import io.gravitee.rest.api.model.configuration.identity.IdentityProviderActivationReferenceType;
 import io.gravitee.rest.api.model.configuration.identity.SocialIdentityProviderEntity;
 import io.gravitee.rest.api.security.utils.AuthoritiesProvider;
@@ -43,6 +44,7 @@ import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.client.Client;
 import jakarta.ws.rs.client.ClientBuilder;
 import jakarta.ws.rs.client.Entity;
+import jakarta.ws.rs.client.Invocation;
 import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.Response;
@@ -50,6 +52,7 @@ import java.io.IOException;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.Base64;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 import javax.inject.Singleton;
@@ -76,6 +79,13 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
     private static final String TEMPLATE_ENGINE_PROFILE_ATTRIBUTE = "profile";
     private static final String ACCESS_TOKEN_PROPERTY = "access_token";
     private static final String ID_TOKEN_PROPERTY = "id_token";
+    private static final String ERROR_PROPERTY = "error";
+    private static final String INVALID_CLIENT_ERROR = "invalid_client";
+    private static final String PROVIDER_ERROR = "identity_provider_error";
+    private static final String CLIENT_AUTHENTICATION_HINT =
+        "The identity provider rejected the client credentials. Check that the tokenEndpointAuthMethod configured on " +
+        "the identity provider matches the method the provider expects (client_secret_basic or client_secret_post); " +
+        "the token and introspection endpoints must both accept it.";
 
     @Autowired
     private SocialIdentityProviderService socialIdentityProviderService;
@@ -126,20 +136,12 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
                 // Step1. Check the token by invoking the introspection endpoint
                 final MultivaluedStringMap introspectData = new MultivaluedStringMap();
                 introspectData.add(TOKEN, token);
-                Response response = client
+                Invocation.Builder introspectRequest = client
                     //TODO: what is the correct introspection URL here ?
                     .target(identityProvider.getTokenIntrospectionEndpoint())
-                    .request(jakarta.ws.rs.core.MediaType.APPLICATION_JSON_TYPE)
-                    .header(
-                        HttpHeaders.AUTHORIZATION,
-                        String.format(
-                            "Basic %s",
-                            Base64.getEncoder().encodeToString(
-                                (identityProvider.getClientId() + ':' + identityProvider.getClientSecret()).getBytes()
-                            )
-                        )
-                    )
-                    .post(Entity.form(introspectData));
+                    .request(jakarta.ws.rs.core.MediaType.APPLICATION_JSON_TYPE);
+                authenticateClient(identityProvider, ClientAuthenticationMethod.CLIENT_SECRET_BASIC, introspectData, introspectRequest);
+                Response response = introspectRequest.post(Entity.form(introspectData));
                 introspectData.clear();
 
                 if (response.getStatus() == Response.Status.OK.getStatusCode()) {
@@ -151,16 +153,11 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
                     } else {
                         return Response.status(Response.Status.UNAUTHORIZED).entity(introspectPayload).build();
                     }
-                } else {
-                    log.error(
-                        "Token exchange failed with status {}: {}\n{}",
-                        response.getStatus(),
-                        response.getStatusInfo(),
-                        getResponseEntityAsString(response)
-                    );
                 }
 
-                return Response.status(response.getStatusInfo()).entity(response.getEntity()).build();
+                String errorBody = getResponseEntityAsString(response);
+                log.error("Token exchange failed with status {}: {}\n{}", response.getStatus(), response.getStatusInfo(), errorBody);
+                return clientAuthenticationFailure(errorBody);
             } else {
                 return Response.status(Response.Status.BAD_REQUEST)
                     .entity("Token exchange is not supported for this identity provider")
@@ -192,15 +189,16 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
             final MultivaluedStringMap accessData = new MultivaluedStringMap();
             accessData.add(CLIENT_ID_KEY, payloadInput.getClient_id());
             accessData.add(REDIRECT_URI_KEY, payloadInput.getRedirect_uri());
-            accessData.add(CLIENT_SECRET, identityProvider.getClientSecret());
             accessData.add(CODE_KEY, payloadInput.getCode());
             accessData.add(CODE_VERIFIER_KEY, payloadInput.getCode_verifier());
             accessData.add(GRANT_TYPE_KEY, AUTH_CODE);
 
-            Response response = client
+            Invocation.Builder tokenRequest = client
                 .target(identityProvider.getTokenEndpoint())
-                .request(jakarta.ws.rs.core.MediaType.APPLICATION_JSON_TYPE)
-                .post(Entity.form(accessData));
+                .request(jakarta.ws.rs.core.MediaType.APPLICATION_JSON_TYPE);
+            authenticateClient(identityProvider, ClientAuthenticationMethod.CLIENT_SECRET_POST, accessData, tokenRequest);
+
+            Response response = tokenRequest.post(Entity.form(accessData));
             accessData.clear();
 
             if (response.getStatus() == Response.Status.OK.getStatusCode()) {
@@ -208,18 +206,78 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
                 final String accessToken = (String) responseEntity.get(ACCESS_TOKEN_PROPERTY);
                 final String idToken = (String) responseEntity.get(ID_TOKEN_PROPERTY);
                 return authenticateUser(identityProvider, servletResponse, accessToken, idToken, payloadInput.getState());
-            } else {
-                log.error(
-                    "Exchange authorization code failed with status {}: {}\n{}",
-                    response.getStatus(),
-                    response.getStatusInfo(),
-                    getResponseEntityAsString(response)
-                );
             }
-            return Response.status(Response.Status.UNAUTHORIZED).build();
+
+            String errorBody = getResponseEntityAsString(response);
+            log.error(
+                "Exchange authorization code failed with status {}: {}\n{}",
+                response.getStatus(),
+                response.getStatusInfo(),
+                errorBody
+            );
+            return clientAuthenticationFailure(errorBody);
         }
 
         return Response.status(Response.Status.NOT_FOUND).build();
+    }
+
+    /**
+     * Applies the client credentials the way the identity provider expects to receive them. A provider that declares no
+     * method keeps the behaviour this endpoint has always had, so existing configurations are unaffected; declaring one
+     * makes the token and introspection calls agree, which they previously did not.
+     */
+    private void authenticateClient(
+        SocialIdentityProviderEntity identityProvider,
+        ClientAuthenticationMethod endpointDefault,
+        MultivaluedStringMap form,
+        Invocation.Builder request
+    ) {
+        ClientAuthenticationMethod method = identityProvider.getTokenEndpointAuthMethod() != null
+            ? identityProvider.getTokenEndpointAuthMethod()
+            : endpointDefault;
+
+        if (method == ClientAuthenticationMethod.CLIENT_SECRET_BASIC) {
+            request.header(
+                HttpHeaders.AUTHORIZATION,
+                String.format(
+                    "Basic %s",
+                    Base64.getEncoder().encodeToString(
+                        (identityProvider.getClientId() + ':' + identityProvider.getClientSecret()).getBytes()
+                    )
+                )
+            );
+        } else {
+            // The authorization-code flow already carries the client_id supplied by the caller, so only add one where
+            // the form has none and the request body keeps the shape providers are already receiving.
+            if (!form.containsKey(CLIENT_ID_KEY)) {
+                form.add(CLIENT_ID_KEY, identityProvider.getClientId());
+            }
+            form.add(CLIENT_SECRET, identityProvider.getClientSecret());
+        }
+    }
+
+    /**
+     * Reports a failed token or introspection call as an authentication failure carrying the provider's own error code,
+     * naming the client authentication method when the provider rejected our credentials. Only the error code is
+     * relayed, never the provider's raw response, which may describe its internals.
+     */
+    private Response clientAuthenticationFailure(String providerResponseBody) {
+        String error = PROVIDER_ERROR;
+        try {
+            Object providerError = getEntity(providerResponseBody).get(ERROR_PROPERTY);
+            if (providerError instanceof String providerErrorCode && !providerErrorCode.isBlank()) {
+                error = providerErrorCode;
+            }
+        } catch (IOException | RuntimeException e) {
+            log.debug("Identity provider error response is not a JSON object", e);
+        }
+
+        Map<String, String> payload = new LinkedHashMap<>();
+        payload.put(ERROR_PROPERTY, error);
+        if (INVALID_CLIENT_ERROR.equals(error)) {
+            payload.put("hint", CLIENT_AUTHENTICATION_HINT);
+        }
+        return Response.status(Response.Status.UNAUTHORIZED).entity(payload).build();
     }
 
     /**
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
@@ -50,6 +50,7 @@ import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.Response;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.Base64;
@@ -83,6 +84,7 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
     private static final String ERROR_PROPERTY = "error";
     private static final String INVALID_CLIENT_ERROR = "invalid_client";
     private static final String PROVIDER_ERROR = "identity_provider_error";
+    private static final String PROVIDER_UNAVAILABLE_ERROR = "identity_provider_unavailable";
     private static final String CLIENT_AUTHENTICATION_HINT =
         "The identity provider rejected the client credentials. Check that the tokenEndpointAuthMethod configured on " +
         "the identity provider matches the method the provider expects (client_secret_basic or client_secret_post); " +
@@ -161,13 +163,21 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
                     if (active) {
                         return authenticateUser(identityProvider, servletResponse, token, null, null);
                     } else {
+                        log.info("Identity provider {} reported the exchanged token as inactive", identityProvider.getId());
                         return Response.status(Response.Status.UNAUTHORIZED).entity(introspectPayload).build();
                     }
                 }
 
                 String errorBody = getResponseEntityAsString(response);
-                log.error("Token exchange failed with status {}: {}\n{}", response.getStatus(), response.getStatusInfo(), errorBody);
-                return clientAuthenticationFailure(errorBody);
+                log.warn(
+                    "Token introspection for identity provider {} at {} failed with status {}: {}\n{}",
+                    identityProvider.getId(),
+                    identityProvider.getTokenIntrospectionEndpoint(),
+                    response.getStatus(),
+                    response.getStatusInfo(),
+                    errorBody
+                );
+                return clientAuthenticationFailure(response.getStatus(), errorBody);
             } else {
                 return Response.status(Response.Status.BAD_REQUEST)
                     .entity("Token exchange is not supported for this identity provider")
@@ -219,13 +229,15 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
             }
 
             String errorBody = getResponseEntityAsString(response);
-            log.error(
-                "Exchange authorization code failed with status {}: {}\n{}",
+            log.warn(
+                "Authorization-code exchange for identity provider {} at {} failed with status {}: {}\n{}",
+                identityProvider.getId(),
+                identityProvider.getTokenEndpoint(),
                 response.getStatus(),
                 response.getStatusInfo(),
                 errorBody
             );
-            return clientAuthenticationFailure(errorBody);
+            return clientAuthenticationFailure(response.getStatus(), errorBody);
         }
 
         return Response.status(Response.Status.NOT_FOUND).build();
@@ -246,24 +258,26 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
             ? identityProvider.getTokenEndpointAuthMethod()
             : endpointDefault;
 
-        if (method == ClientAuthenticationMethod.CLIENT_SECRET_BASIC) {
-            request.header(
-                HttpHeaders.AUTHORIZATION,
-                String.format(
-                    "Basic %s",
-                    Base64.getEncoder().encodeToString(
-                        (identityProvider.getClientId() + ':' + identityProvider.getClientSecret()).getBytes()
-                    )
-                )
-            );
-        } else {
-            // The authorization-code flow already carries the client_id supplied by the caller, so only add one where
-            // the form has none and the request body keeps the shape providers are already receiving.
-            if (!form.containsKey(CLIENT_ID_KEY)) {
-                form.add(CLIENT_ID_KEY, identityProvider.getClientId());
+        // Switched rather than decided by exclusion: token_endpoint_auth_method has values such as none and
+        // private_key_jwt that must not fall through to putting the secret in the body, so adding a constant has to be
+        // a deliberate decision here rather than a silent change on a credential path.
+        switch (method) {
+            case CLIENT_SECRET_BASIC -> request.header(HttpHeaders.AUTHORIZATION, basicAuthorization(identityProvider));
+            case CLIENT_SECRET_POST -> {
+                // The authorization-code flow already carries the client_id supplied by the caller. add() registers the
+                // key even for a null value, so the presence of a value is what decides, not the presence of the key.
+                if (form.getFirst(CLIENT_ID_KEY) == null) {
+                    form.putSingle(CLIENT_ID_KEY, identityProvider.getClientId());
+                }
+                form.add(CLIENT_SECRET, identityProvider.getClientSecret());
             }
-            form.add(CLIENT_SECRET, identityProvider.getClientSecret());
         }
+    }
+
+    private static String basicAuthorization(SocialIdentityProviderEntity identityProvider) {
+        String credentials = identityProvider.getClientId() + ':' + identityProvider.getClientSecret();
+        // Explicit charset: the platform default silently changes the bytes, and therefore the credential, per host
+        return String.format("Basic %s", Base64.getEncoder().encodeToString(credentials.getBytes(StandardCharsets.UTF_8)));
     }
 
     /**
@@ -271,8 +285,10 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
      * naming the client authentication method when the provider rejected our credentials. Only the error code is
      * relayed, never the provider's raw response, which may describe its internals.
      */
-    private Response clientAuthenticationFailure(String providerResponseBody) {
-        String error = PROVIDER_ERROR;
+    private Response clientAuthenticationFailure(int providerStatus, String providerResponseBody) {
+        // A provider that is down and a provider that rejected our credentials are different problems, so they must not
+        // arrive as the same error code. Without this, an outage reads as a configuration mistake.
+        String error = providerStatus >= 500 ? PROVIDER_UNAVAILABLE_ERROR : PROVIDER_ERROR;
         try {
             Object providerError = getEntity(providerResponseBody).get(ERROR_PROPERTY);
             if (providerError instanceof String providerErrorCode && !providerErrorCode.isBlank()) {
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
@@ -49,6 +49,7 @@ import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.Response;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.Base64;
@@ -82,6 +83,7 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
     private static final String ERROR_PROPERTY = "error";
     private static final String INVALID_CLIENT_ERROR = "invalid_client";
     private static final String PROVIDER_ERROR = "identity_provider_error";
+    private static final String PROVIDER_UNAVAILABLE_ERROR = "identity_provider_unavailable";
     private static final String CLIENT_AUTHENTICATION_HINT =
         "The identity provider rejected the client credentials. Check that the tokenEndpointAuthMethod configured on " +
         "the identity provider matches the method the provider expects (client_secret_basic or client_secret_post); " +
@@ -151,13 +153,21 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
                     if (active) {
                         return authenticateUser(identityProvider, servletResponse, token, null, null);
                     } else {
+                        log.info("Identity provider {} reported the exchanged token as inactive", identityProvider.getId());
                         return Response.status(Response.Status.UNAUTHORIZED).entity(introspectPayload).build();
                     }
                 }
 
                 String errorBody = getResponseEntityAsString(response);
-                log.error("Token exchange failed with status {}: {}\n{}", response.getStatus(), response.getStatusInfo(), errorBody);
-                return clientAuthenticationFailure(errorBody);
+                log.warn(
+                    "Token introspection for identity provider {} at {} failed with status {}: {}\n{}",
+                    identityProvider.getId(),
+                    identityProvider.getTokenIntrospectionEndpoint(),
+                    response.getStatus(),
+                    response.getStatusInfo(),
+                    errorBody
+                );
+                return clientAuthenticationFailure(response.getStatus(), errorBody);
             } else {
                 return Response.status(Response.Status.BAD_REQUEST)
                     .entity("Token exchange is not supported for this identity provider")
@@ -209,13 +219,15 @@ public class OAuth2AuthenticationResource extends AbstractAuthenticationResource
             }
 
             String errorBody = getResponseEntityAsString(response);
-            log.error(
-                "Exchange authorization code failed with status {}: {}\n{}",
+            log.warn(
+                "Authorization-code exchange for identity provider {} at {} failed with status {}: {}\n{}",
+                identityProvider.getId(),
+                identityProvider.getTokenEndpoint(),
                 response.getStatus(),
                 response.getStatusInfo(),
                 errorBody
             );
-            return clientAuthenticationFailure(errorBody);
+            return clientAuthenticationFailure(response.getStatus(), errorBody);
         }
 
         return Response.status(Response.Status.NOT_FOUND).build();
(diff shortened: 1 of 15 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/identity/SocialIdentityProviderImpl.java
```diff
@@ -175,6 +177,12 @@ private SocialIdentityProviderEntity convert(IdentityProviderEntity identityProv
             provider.setDescription(identityProvider.getDescription());
             provider.setClientId((String) identityProvider.getConfiguration().get(CLIENT_ID));
             provider.setClientSecret((String) identityProvider.getConfiguration().get(CLIENT_SECRET));
+            provider.setTokenEndpointAuthMethod(
+                clientAuthenticationMethod(
+                    (String) identityProvider.getConfiguration().get(TOKEN_ENDPOINT_AUTH_METHOD),
```
paulatulis: **The console silently deletes this key**, which I think makes the fix not survive normal use.

`tokenEndpointAuthMethod` is settable today — `configuration` is a free-form map on `PUT .../configuration/identities/{id}`, so an admin can set it and login starts working. But the console's OIDC form enumerates a fixed control list (`org-settings-identity-provider-oidc.component.ts:30`: clientId, clientSecret, tokenEndpoint, tokenIntrospectionEndpoint, authorizeEndpoint, userInfoEndpoint, userLogoutEndpoint, scopes, color) with no control for this key — and `org-settings-identity-provider.component.ts:188` submits `{ ...this.initialIdentityProviderValue, ...formSettingsValue }`.

That is a **top-level** spread, so `configuration` is replaced wholesale by the form's version. The next time anyone opens that identity provider in the console and saves — for any unrelated reason — `tokenEndpointAuthMethod` is dropped and login reverts to the pre-fix behaviour, with no warning and nothing in the diff to hint at why. Same for the Gravitee AM form (`org-settings-identity-provider-graviteeio-am.component.ts:30`).

I verified both halves directly: the fixed control list, and the top-level spread in `onSubmit`.

Worth noting your sibling PR #19256 adds a control to this very component for the claims whitelist, so the pattern for exposing a new identity-provider configuration key already exists there.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/identity/SocialIdentityProviderImpl.java
```diff
@@ -185,4 +193,23 @@ private SocialIdentityProviderEntity convert(IdentityProviderEntity identityProv
...
     }
+
+    /**
+     * An unrecognised value is reported rather than rejected: failing here would break every login for the provider,
+     * whereas falling back leaves authentication working while the warning names the typo.
+     */
+    private ClientAuthenticationMethod clientAuthenticationMethod(String configured, String identityProviderId) {
+        ClientAuthenticationMethod method = ClientAuthenticationMethod.fromValue(configured);
+        if (method == null && configured != null && !configured.isBlank()) {
+            log.warn(
```
paulatulis: This warning fires on every `convert` call, not once per configuration change. `convert` runs on the identity-provider config endpoints and on every token exchange, so a single typo in `tokenEndpointAuthMethod` produces a WARN line per request for as long as it goes unnoticed — on an authentication path, which is where log volume hurts most.

The fallback behaviour itself is the right call; it's the reporting cadence that turns one typo into log flooding. Something that reports the bad value once per provider per configuration load, rather than per conversion, would keep the diagnostic without the volume.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/identity/SocialIdentityProviderImpl.java
```diff
@@ -185,4 +193,23 @@ private SocialIdentityProviderEntity convert(IdentityProviderEntity identityProv
 
         return null;
     }
+
+    /**
+     * An unrecognised value is reported rather than rejected: failing here would break every login for the provider,
+     * whereas falling back leaves authentication working while the warning names the typo.
+     */
+    private ClientAuthenticationMethod clientAuthenticationMethod(String configured, String identityProviderId) {
```
paulatulis: No test class exists for `SocialIdentityProviderImpl` anywhere in the repo, so two behaviours this PR relies on are unverified:

- that a configured `tokenEndpointAuthMethod` string actually reaches the entity as the typed enum, and
- that an unrecognised value falls back to `null` (endpoint defaults) instead of failing login.

The second is a central safety claim of this PR — the javadoc immediately above states that failing here would break every login for the provider. That claim currently rests on inspection alone.

The resource-layer tests can't cover this: by the time the resources see it, it is already a typed enum, so the string-to-enum collapse is only exercised here.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/configuration/identity/ClientAuthenticationMethod.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Resolves a configured value, accepting either the OIDC name ({@code client_secret_basic}) or the enum constant
+     * ({@code CLIENT_SECRET_BASIC}).
+     *
+     * @return the matching method, or {@code null} when the value is absent, blank or unrecognised. Callers decide what
+     *         an unrecognised value means rather than having a login fail here on a configuration typo.
+     */
+    public static ClientAuthenticationMethod fromValue(String value) {
```
paulatulis, line 52: `fromValue` has no test anywhere in the repo — no `ClientAuthenticationMethodTest` or equivalent exists. None of its documented resolution cases is exercised directly: null, blank, whitespace-padded, the OIDC name (`client_secret_basic`), the enum constant name (`CLIENT_SECRET_BASIC`), wrong case, or unrecognised. The resource tests only ever pass the enum, never the string round-trip.

This method is the single point where a provider's configured string becomes behaviour. A regression here doesn't fail loudly — it returns `null`, which every caller reads as "use the endpoint default", silently flipping every configured provider back to the pre-fix behaviour. That is the same failure mode the ticket was filed for.

A parameterized test over those inputs would be cheap relative to what it pins.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/configuration/identity/ClientAuthenticationMethod.java
```diff
@@ -0,0 +1,64 @@
...
+        if (value == null || value.isBlank()) {
+            return null;
+        }
+        String normalized = value.trim();
+        for (ClientAuthenticationMethod method : values()) {
+            if (method.value.equalsIgnoreCase(normalized) || method.name().equalsIgnoreCase(normalized)) {
```
paulatulis: Nit: the `method.name().equalsIgnoreCase(normalized)` branch is redundant. The enum constant names are the value strings uppercased with the same underscores (`CLIENT_SECRET_BASIC` vs `client_secret_basic`), so `method.value.equalsIgnoreCase(normalized)` on the preceding line already matches every spelling this branch would catch.

Harmless, but the javadoc promises two distinct accepted forms where there is really only one comparison doing the work.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
```diff
@@ -202,36 +199,97 @@ public Response exchangeAuthorizationCode(
...
+            ? identityProvider.getTokenEndpointAuthMethod()
+            : endpointDefault;
+
+        if (method == ClientAuthenticationMethod.CLIENT_SECRET_BASIC) {
+            request.header(
+                HttpHeaders.AUTHORIZATION,
+                String.format(
+                    "Basic %s",
+                    Base64.getEncoder().encodeToString(
+                        (identityProvider.getClientId() + ':' + identityProvider.getClientSecret()).getBytes()
```
paulatulis: Pre-existing rather than introduced here — I diffed it and this is byte-identical to the code removed above, so there is no regression. Raising it because this PR changes its blast radius.

The credential is built with the platform-default charset (`getBytes()` with no argument) and without the form-urlencoding of client_id/client_secret that RFC 6749 §2.3.1 requires before base64. A secret containing `:` or non-ASCII characters produces an ambiguous or malformed credential.

Previously this quirk only affected introspection. Now that `authenticateClient` also drives the token endpoint, it applies to a second call site. Worst case is an authentication failure rather than a bypass, and passing `StandardCharsets.UTF_8` would be a strictly-safe subset of the fix if the urlencoding change feels out of scope.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
```diff
@@ -202,36 +199,97 @@ public Response exchangeAuthorizationCode(
...
+                    )
+                )
+            );
+        } else {
+            // The authorization-code flow already carries the client_id supplied by the caller, so only add one where
+            // the form has none and the request body keeps the shape providers are already receiving.
+            if (!form.containsKey(CLIENT_ID_KEY)) {
```
paulatulis: This guard doesn't deliver the fallback the comment above it promises, when the caller's payload carries a null `client_id`.

Jersey's `AbstractMultivaluedMap.add(key, value)` fetches-or-creates the value list for the key and only then appends a non-null value — so `form.add(CLIENT_ID_KEY, null)` registers the key with an *empty* list. `containsKey` then returns true, the configured client id is never added, and the token request goes out carrying `client_secret` but no `client_id`.

No functional regression (the old code never added the configured id either, and both first-party UIs always send `client_id`), so this is latent rather than live. Checking for an actual value rather than key presence would close it. Same code in the portal copy at `OAuth2AuthenticationResource.java:238`.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
```diff
@@ -202,36 +199,97 @@ public Response exchangeAuthorizationCode(
...
+            form.add(CLIENT_SECRET, identityProvider.getClientSecret());
+        }
+    }
+
+    /**
+     * Reports a failed token or introspection call as an authentication failure carrying the provider's own error code,
+     * naming the client authentication method when the provider rejected our credentials. Only the error code is
+     * relayed, never the provider's raw response, which may describe its internals.
+     */
+    private Response clientAuthenticationFailure(String providerResponseBody) {
```
paulatulis: Only the `invalid_client` JSON branch of this method is tested. Two branches are not:

- the non-JSON-body fallback (a provider returning HTML or an empty body → `identity_provider_error`, no hint), and
- hint suppression for error codes other than `invalid_client`.

Those are precisely the branches whose regression would relay the provider's raw response — the thing the javadoc here commits to never doing. As written, a change that started passing the body through would not fail any test.

The portal copy at `OAuth2AuthenticationResource.java:250` has the same gap.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResourceTest.java
```diff
@@ -300,6 +301,155 @@ public void should_exchange_token_by_body() throws IOException, NoSuchAlgorithmE
         verifyJwtToken(response);
     }
 
+    @Test
```
paulatulis, line 305: The unrecognised-`tokenEndpointAuthMethod` path has no test in either module — neither that login still succeeds on the endpoint defaults, nor that the warning is emitted.

That is the PR's stated safety property ("a typo cannot lock every user out of login"), and it is the one claim in the description with no test behind it.

Minor, separately: these new methods are declared `public void` in a JUnit 5 class. The file has both styles already, so this only perpetuates the pre-migration modifier rather than diverging from the file.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResourceTest.java
```diff
@@ -300,6 +301,155 @@ public void should_exchange_token_by_body() throws IOException, NoSuchAlgorithmE
...
+        assertEquals(HttpStatusCode.OK_200, response.getStatus());
+        WireMock.verify(
+            postRequestedFor(urlEqualTo("/introspect"))
+                .withHeader(HttpHeaders.AUTHORIZATION, absent())
+                .withRequestBody(containing("client_secret=the_client_secret"))
+                .withRequestBody(containing("client_id=the_client_id"))
+        );
+    }
+
+    @Test
```
paulatulis, line 399: These `invalid_client` tests can't detect the leak they appear to guard against: the stubbed provider response body is equal to the string being asserted, so an implementation that relayed the provider's raw body verbatim would produce the same assertion result as one that correctly extracts only the error code.

Making the stub body carry something that must NOT appear in the response — an extra field describing provider internals, say — would turn these into genuine tests of the "only the error code is relayed" guarantee.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/test/java/io/gravitee/rest/api/portal/rest/resource/auth/OAuth2AuthenticationResourceTest.java
```diff
@@ -365,6 +378,122 @@ void shouldNotConnectUserWhenMissingMailInUserInfo() throws Exception {
         assertFalse(response.getCookies().containsKey(HttpHeaders.AUTHORIZATION));
     }
 
+    @Test
+    void should_authenticate_introspection_with_basic_auth_when_no_auth_method_configured() throws Exception {
```
paulatulis, line 383: **Test parity gap.** Management adds 6 tests here, portal adds 4. Portal is missing:

1. `should_send_client_credentials_in_token_request_body_when_no_auth_method_configured` — the default token-request behaviour. Only incidentally covered by the pre-existing exact-body stub, which would still pass if a spurious `Authorization: Basic` header were also sent.
2. `should_report_invalid_client_on_token_request_instead_of_an_empty_unauthorized` — the portal's pre-existing failure test asserts only the 401 status, not the new body.

This matters more than a usual parity gap because the two production files are **byte-identical** — I diffed the changed lines, 122 each, zero difference. Identical code pinned by non-identical tests means a portal-only regression in either behaviour goes undetected, and the default-behaviour test is the one guarding your central backwards-compatibility claim.

Two smaller notes on the new portal tests: they introduce `// Given` / `// When` / `// Then` markers into a file that previously had none, and use snake_case names in an otherwise all-camelCase class. (The management file already mixes both, so its new tests are consistent with their surroundings.)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/identity/SocialIdentityProviderImpl.java
paulatulis: The PR description's safety property — "a typo cannot lock every user out of login" — holds for a wrong *string*, but not for a wrong *type*.

This is an unchecked `(String)` cast on a value from a free-form, unvalidated configuration map. A non-String value (`"tokenEndpointAuthMethod": 123`, or an unquoted value that parses as a boolean or number) throws a `ClassCastException` here, before `fromValue` ever gets the chance to warn and fall back.

And it does not fail narrowly: `findAll` wraps `.map(this::convert)` in `catch (Exception ex) -> TechnicalManagementException` at line 92, so the exception takes down the whole identity-provider list — every SSO button disappears for **every** provider, not just the misconfigured one. That is the exact failure mode the fallback design was chosen to prevent.

In fairness, `clientId` and `clientSecret` on lines 178-179 use the identical cast, so the pattern is pre-existing and this PR is consistent with it. Raising it because this is the key that carries the explicit safety claim, and the one an administrator hand-edits in a free-form map where a type slip is easy.

Separately and reinforcing the console finding above: `tokenEndpointAuthMethod` is `@JsonIgnore` on both the field (`SocialIdentityProviderEntity:56`) and the getter (`:122`), so no API response exposes the effective value. Between that and the console silently dropping the key, an administrator can neither read back what is in force nor keep it once set.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
```diff
@@ -161,16 +163,11 @@ public Response tokenExchange(
...
-                        "Token exchange failed with status {}: {}\n{}",
-                        response.getStatus(),
-                        response.getStatusInfo(),
-                        getResponseEntityAsString(response)
-                    );
                 }
 
-                return Response.status(response.getStatusInfo()).entity(response.getEntity()).build();
+                String errorBody = getResponseEntityAsString(response);
+                log.error("Token exchange failed with status {}: {}\n{}", response.getStatus(), response.getStatusInfo(), errorBody);
```
paulatulis: Three things about this log line, on a PR whose stated purpose is making an opaque failure diagnosable:

1. **Wrong operation.** This is the introspection branch, but the message says `"Token exchange failed"`. An operator reading it will look at the token endpoint, which is the other call.
2. **No provider id and no endpoint URL.** With several identity providers configured, the status and body alone do not say *which* provider failed, or which of its two endpoints. That is the information the ticket was filed for the absence of.
3. **ERROR level on a path that now answers 401.** This module's convention is that an expected, gracefully-handled failure answered with a 401 logs at warn or info, with error reserved for failures of the operation itself.

One more, related: line 170 routes every non-200 into `clientAuthenticationFailure`, so a provider that is down, unreachable, or returning 503 is reported to the caller identically to one that rejected the credentials. The `invalid_client` hint makes that worse — it will suggest `tokenEndpointAuthMethod` is misconfigured when the provider simply never answered.

Same pattern at line 222 for the authorization-code path (correct operation name there, but the same missing provider/endpoint context).

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
```diff
@@ -161,16 +163,11 @@ public Response tokenExchange(
                     } else {
                         return Response.status(Response.Status.UNAUTHORIZED).entity(introspectPayload).build();
```
paulatulis, line 157: This is the one remaining failure path with no log line at all. The provider answered 200 and explicitly said the token is inactive, the user is refused — and nothing is recorded.

Every other failure branch in this method now logs. An operator debugging "login works for some users but not others" gets a silent 401 here, with no way to distinguish it from the other paths that do leave a trace.

Outside the ticket's scope, but this method is being reworked for diagnosability and this branch was left out of that.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
```diff
@@ -202,36 +199,97 @@ public Response exchangeAuthorizationCode(
...
 
         return Response.status(Response.Status.NOT_FOUND).build();
     }
 
+    /**
+     * Applies the client credentials the way the identity provider expects to receive them. A provider that declares no
+     * method keeps the behaviour this endpoint has always had, so existing configurations are unaffected; declaring one
+     * makes the token and introspection calls agree, which they previously did not.
+     */
+    private void authenticateClient(
```
paulatulis, line 241: `authenticateClient` decides by exclusion: anything that is not `CLIENT_SECRET_BASIC` falls through to putting `client_secret` in the request body.

Today that is safe, because the enum has exactly two constants. But it makes the *unsafe* case the default. `token_endpoint_auth_method` has other standard values — `none`, `private_key_jwt`, `client_secret_jwt`, `tls_client_auth` — and `none` in particular means "send no client credentials at all". If a future change adds any of them to the enum, this method silently posts the secret rather than failing or being caught by the compiler.

A switch over the enum (or an explicit `== CLIENT_SECRET_POST` with a rejected/logged default) would make adding a constant a visible decision instead of a silent behaviour change on a credential path.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
```diff
@@ -202,36 +199,97 @@ public Response exchangeAuthorizationCode(
...
+        SocialIdentityProviderEntity identityProvider,
```
paulatulis, line 242: Not blocking, and I checked that it follows existing precedent rather than introducing anything new — noting it only so the trade-off is on the record.

`authenticateClient` and `clientAuthenticationFailure` are duplicated byte-for-byte in the portal resource. I diffed the two files' changed lines: 122 each, zero difference. Both modules already depend on rest-api-model (for `ClientAuthenticationMethod`) and rest-api-service, and this logic touches nothing module-specific — it works on `SocialIdentityProviderEntity`, `MultivaluedStringMap` and `Invocation.Builder`, all shared — so a common helper would not cross any layer boundary.

Against that: `AbstractAuthenticationResource` is already duplicated per module, and the two `OAuth2AuthenticationResource` files were near-identical before this change, so this follows the codebase's established management/portal split rather than departing from it.

The concrete risk is drift on security-critical code — the Basic-header encoding issue flagged above now has to be fixed in two places, and the test-parity gap in the portal suite is the first instance of exactly that asymmetry. Worth a follow-up ticket rather than expanding this PR.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/identity/SocialIdentityProviderImpl.java
wbabyte: Confirmed both halves myself and fixed it — this was the finding that mattered most, since without it the fix reverted on the next unrelated save.

Added a `tokenEndpointAuthMethod` control to the OIDC form **and** the Gravitee AM form, since both feed the same wholesale `configuration` rebuild. The values come from a shared `CLIENT_AUTHENTICATION_METHODS` list so the two forms cannot drift, with a "Provider default" option for leaving it unset.

There is a regression test for exactly the scenario you describe — load a provider whose configuration carries the key, make an unrelated edit, save, and assert the key is still in the `PUT`. I verified it is not vacuous by removing the control and re-running: it fails with `Cannot find control with name: 'tokenEndpointAuthMethod'`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/identity/SocialIdentityProviderImpl.java
wbabyte: Agreed — one typo producing a WARN per authentication request is worse than the typo.

It is now reported once per provider **and value**: a `ConcurrentHashMap<String, String>` of provider id to last-reported value, so the line appears when the value first appears or changes, and not again while it stays the same. Bounded by provider count.

Kept the fallback behaviour and the diagnostic content unchanged, as you suggested — only the cadence changed.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/identity/SocialIdentityProviderImpl.java
wbabyte: Fixed — added `SocialIdentityProviderImplTest`, 8 tests, covering exactly the two behaviours you named plus the type case from your other comment:

- each configured value resolving to the typed enum
- an unrecognised value falling back to `null` rather than failing
- a non-`String` value falling back rather than throwing

You are right that the resource tests could not reach this: by the time they see it, it is already an enum, so this was the only place the string is interpreted and it had no coverage at all.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/configuration/identity/ClientAuthenticationMethod.java
wbabyte: Fixed — added `ClientAuthenticationMethodTest`, 17 cases, covering every input you listed: `null`, empty, whitespace-only, whitespace-padded, the OIDC name, the enum constant name, mixed case, unsupported-but-real values (`none`, `private_key_jwt`, `client_secret_jwt`), and nonsense.

Your framing of the failure mode is what made this worth more than its size: a regression returns `null`, every caller reads that as "use the endpoint default", and every configured provider silently reverts to the behaviour the ticket was filed for. Nothing would have failed.

gravitee-apim-rest-api/gravitee-apim-rest-api-model/src/main/java/io/gravitee/rest/api/model/configuration/identity/ClientAuthenticationMethod.java
wbabyte: Correct, and removed. The enum constant names are the value strings uppercased, so `value.equalsIgnoreCase(normalized)` already matched every spelling the second branch could catch — it was doing no work.

The javadoc now says the comparison is case-insensitive and that both spellings resolve, rather than implying two distinct comparisons.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
wbabyte: Took the strictly-safe subset you suggested: the credential is now encoded with `StandardCharsets.UTF_8` in both modules, so the bytes no longer depend on the host's default charset.

I have left the RFC 6749 §2.3.1 form-urlencoding out. It is a real gap, but it changes the credential for any client id or secret containing a character that needs escaping, which is a behaviour change for existing working configurations and wants its own change with its own test matrix rather than riding along here. Worth a ticket — happy to raise it if you agree.

Thanks for diffing it and confirming there was no regression; that is what let me treat the charset separately from the encoding.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
wbabyte: Confirmed and fixed. `AbstractMultivaluedMap.add` registers the key and only then appends a non-null value, so `containsKey` was true with an empty list and the configured client id was never used.

It now decides on the presence of a value — `form.getFirst(CLIENT_ID_KEY) == null` — and uses `putSingle` so the key holds exactly one value. Applied to the portal copy at the same time, and the comment now says the value decides rather than the key.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
wbabyte: Fixed — both branches now have tests, in the management module:

- `should_report_a_non_json_provider_response_without_relaying_it`: an HTML body from a 502, asserting the response names the provider as unavailable and that the upstream address in the body does not come back
- `should_not_suggest_the_client_authentication_method_for_an_unrelated_error`: an `invalid_grant`, asserting the hint is absent

You were right that these were the branches whose regression would relay the raw response — the thing the javadoc commits to never doing.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResourceTest.java
wbabyte: Covered, though at a different layer than you suggested, so flagging the reasoning rather than claiming a test where you looked for one.

An unrecognised **string** cannot reach the resource — `SocialIdentityProviderImpl` collapses it to `null` before the resource sees anything. So the safety property is pinned in the new `SocialIdentityProviderImplTest` (unusable value resolves to `null`), and "login still succeeds on the endpoint defaults" is pinned by the existing default-behaviour tests in both modules, which run with the method unset. A resource test taking an unrecognised string would have had to bypass the type system to construct the state.

The warning itself I have deliberately not asserted on — pinning a log line felt like the wrong trade for its value, though I will add it if you would rather have it.

On the `public void` nit: left as-is, since the file already mixes both and matching the surroundings seemed better than a partial migration.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResourceTest.java
wbabyte: Good catch — the stub body and the asserted string were the same, so a verbatim relay was indistinguishable from correct extraction.

Both tests now stub a body carrying detail that must never come back:

```json
{"error":"invalid_client","error_description":"client not found in realm internal-realm-7","trace":"provider-node-3"}
```

and a shared `assertRelaysOnlyTheErrorCode` asserts `invalid_client` is present while `internal-realm-7`, `provider-node-3` and `error_description` are all absent. Applied in the portal module too.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/test/java/io/gravitee/rest/api/portal/rest/resource/auth/OAuth2AuthenticationResourceTest.java
wbabyte: Fixed — both missing tests added, and I agree the byte-identical production code is what made this more than a usual parity gap.

Portal now has `should_send_client_credentials_in_token_request_body_when_no_auth_method_configured`, asserting no `Authorization` header alongside the credentials in the body, so a spurious Basic header would fail it — which, as you note, the pre-existing exact-body stub would not have caught. It also has the token-endpoint `invalid_client` test, asserting the new body rather than only the status.

On the two smaller notes: I removed the `// Given` / `// When` / `// Then` markers from the portal tests to match that file. I kept snake_case, since the module's conventions specify it and the management file already mixes both — but say the word and I will align them to the portal file's camelCase instead.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/identity/SocialIdentityProviderImpl.java
wbabyte: Both points taken.

The cast is gone: the helper now takes `Object`, resolves only when it is a `String`, and falls back otherwise. Your escalation path is what made this worth fixing rather than noting — `findAll` wrapping `convert` meant one provider with `"tokenEndpointAuthMethod": 123` removed **every** SSO button, which is precisely the failure the fallback exists to prevent. There is a test for the non-`String` case.

I left `clientId`/`clientSecret` on the neighbouring lines alone: same latent issue, but changing them is unrelated to this ticket and belongs in its own change.

On `@JsonIgnore`: I have kept it. The effective value is now readable — the management configuration endpoint returns the raw `configuration` map, and the console control added for your other comment displays it. What `@JsonIgnore` suppresses is the portal-facing social-identities payload, which is public and has no reason to describe how APIM authenticates itself. So the "can neither read back nor keep it" half is resolved by the control; the deliberate part is only the public payload.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
wbabyte: All three fixed, plus the point in your last paragraph.

The introspection branch now says `Token introspection for identity provider {} at {} failed...`, naming the provider id and the endpoint, at `warn` rather than `error` since it answers a 401. The authorization-code branch gained the same provider and endpoint context.

On conflating an outage with a rejection: a 5xx from the provider now reports `identity_provider_unavailable` rather than the generic code, and the `tokenEndpointAuthMethod` hint stays reserved for `invalid_client`, so an unreachable provider is no longer blamed on the configuration. There is a test for it using a 502 with an HTML body.

I have deliberately **not** changed the HTTP status — it stays 401 for both. Mapping 5xx to 502 would be the more correct answer, but it is a response-contract change beyond this ticket, and an automated check on this PR has independently flagged the status change risk. Worth deciding separately.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
wbabyte: Fixed — that branch now logs at `info`, naming the provider: the provider answered 200 and explicitly said the token was inactive, so it is an expected refusal rather than a failure, but it should not be the only silent one.

Agreed it was outside the ticket; it would have been odd to rework the method for diagnosability and leave the one branch that recorded nothing.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
wbabyte: Agreed, and changed — this was the right kind of objection: the current two-value behaviour is fine, but the shape made the unsafe case the default.

`authenticateClient` now switches over the enum with a case per constant and no fall-through, so adding `none`, `private_key_jwt`, `client_secret_jwt` or `tls_client_auth` becomes a compile-visible decision instead of silently posting the secret. The comment names those values so the reason survives.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
wbabyte: Noted, and I agree with your conclusion — following the established management/portal split rather than departing from it, with the duplication on the record.

Two things that came out of it: the Basic-header charset fix from your other comment was applied to both copies in the same change, and the portal parity gap you identified is now closed, so the asymmetry that would have let them drift is gone for the moment.

Worth the follow-up ticket you suggest. If it is taken, the natural home is a small shared helper in `rest-api-model` alongside `ClientAuthenticationMethod`, since as you note both modules already depend on it and nothing here is module-specific.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/auth/OAuth2AuthenticationResource.java
```diff
@@ -136,18 +140,21 @@ public Response tokenExchange(
...
+                String errorBody = getResponseEntityAsString(response);
+                log.warn(
+                    "Token introspection for identity provider {} at {} failed with status {}: {}\n{}",
+                    identityProvider.getId(),
+                    identityProvider.getTokenIntrospectionEndpoint(),
+                    response.getStatus(),
+                    response.getStatusInfo(),
+                    errorBody
+                );
+                return clientAuthenticationFailure(response.getStatus(), errorBody);
```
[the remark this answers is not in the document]
wbabyte: The premise needs correcting: there was no reliable prior contract on this path.

The previous code was `Response.status(response.getStatusInfo()).entity(response.getEntity()).build()` after the body had already been consumed by the log statement above it. The entity was spent, so this did not return the provider's status and body — it produced a **500**. That is one of the two defects APIM-14841 was filed for; the ticket records "the exchange path returns 500". Preserving that behaviour is neither possible nor desirable.

The concern about clients distinguishing provider 4xx from 5xx is fair as forward-looking design, and the body now carries it: a 5xx from the provider reports `identity_provider_unavailable`, a credential rejection relays the provider's own error code, and the client-authentication hint appears only for `invalid_client`.

I have deliberately left the status at 401 rather than mapping 5xx to 502, since that is a contract change beyond this ticket's scope. Leaving the finding open rather than dismissing it, so a human can make that call.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
```diff
@@ -159,18 +163,21 @@ public Response tokenExchange(
...
+                String errorBody = getResponseEntityAsString(response);
+                log.warn(
+                    "Token introspection for identity provider {} at {} failed with status {}: {}\n{}",
+                    identityProvider.getId(),
+                    identityProvider.getTokenIntrospectionEndpoint(),
+                    response.getStatus(),
+                    response.getStatusInfo(),
+                    errorBody
+                );
+                return clientAuthenticationFailure(response.getStatus(), errorBody);
```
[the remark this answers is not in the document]
wbabyte: Same as the portal occurrence — the premise needs correcting.

Before this change, the introspection failure path read the response body into the log statement and then passed the already-consumed `Response.getEntity()` into the outgoing response, which produced a **500** rather than the provider's status and body. That is one of the two defects this PR exists to fix; APIM-14841 records "the exchange path returns 500". There was no stable status or payload for a client to depend on.

Forward-looking, the distinction you raise is now expressed in the body: `identity_provider_unavailable` for a 5xx, the provider's own error code for a rejection, and the client-authentication hint only for `invalid_client`.

The status stays 401, deliberately — changing it is a contract decision beyond this ticket. Leaving this open rather than dismissing it so a human decides.

gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/main/java/io/gravitee/rest/api/management/rest/resource/auth/OAuth2AuthenticationResource.java
```diff
@@ -202,36 +209,103 @@ public Response exchangeAuthorizationCode(
...
+        Invocation.Builder request
+    ) {
+        ClientAuthenticationMethod method = identityProvider.getTokenEndpointAuthMethod() != null
+            ? identityProvider.getTokenEndpointAuthMethod()
+            : endpointDefault;
+
+        // Switched rather than decided by exclusion: token_endpoint_auth_method has values such as none and
+        // private_key_jwt that must not fall through to putting the secret in the body, so adding a constant has to be
+        // a deliberate decision here rather than a silent change on a credential path.
+        switch (method) {
```
paulatulis, line 254: This doesn't deliver the guarantee it's written for, and the comment above it states that it does — which is the part that worries me, because the next person adding a constant will read it and trust it.

`switch (method) { case CLIENT_SECRET_BASIC -> …; case CLIENT_SECRET_POST -> … }` is a switch *statement* over an enum with constant labels. Under JLS §14.11.2 that is not an *enhanced* switch, so exhaustiveness is not required — the arrow form doesn't change that, it only changes fall-through. Adding a constant compiles clean and silently does nothing at runtime.

I compiled a mirror of it rather than going from memory:

```java
enum Method { CLIENT_SECRET_BASIC, CLIENT_SECRET_POST, NONE }

static void authenticateClient(Method method) {
    switch (method) {
        case CLIENT_SECRET_BASIC -> System.out.println("basic header");
        case CLIENT_SECRET_POST  -> System.out.println("secret in body");
    }
}
```

`javac --release 21 -Xlint:all` → compiles cleanly, no error, no warning. Calling it with `NONE` prints nothing.

So the outcome for a future `none` or `private_key_jwt` is a silent no-op rather than a silent secret-in-body. That is genuinely safer than what we had, and I'd still take this over the `if/else`. But it isn't compile-visible, which was the whole reason I asked for the switch. A switch expression, or a `default -> throw`, would make it real — the former at compile time, the latter at least loudly.

Second, smaller point on the same block: `switch` on a null selector throws NPE (confirmed in the same probe). Not reachable today — both call sites pass a non-null `endpointDefault`, `CLIENT_SECRET_BASIC` at :155 and `CLIENT_SECRET_POST` at :219 — so this is latent, not a bug. Worth noting only because the old `if/else` degraded to the post branch on null and this turns the same input into a 500 on the credential path. If the `default` above ends up throwing anyway, that covers both.

Both points apply equally to the byte-identical copy in the portal module — raising it once here rather than twice, but it needs the same change in both.

wbabyte: Right on both counts, and the comment was the worse half of the problem — it asserted a guarantee the code did not carry, which is exactly the failure mode you describe. Reproduced your probe with `javac --release 21 -Xlint:all`: adding a third constant to the switch as written compiles clean, no warning, and the new constant is a silent no-op.

Fixed in 9de8280, both modules, with `case null -> throw` rather than a switch expression or a `default`. The null label makes it an *enhanced* switch under JLS 14.11.2, so exhaustiveness becomes a compile requirement on a plain statement — no `yield`-shaped refactor needed to get the compile-time half. Same third-constant probe against the fixed shape:

```
FixedPlusNone.java:6: error: the switch statement does not cover all possible input values
    switch (method) {
    ^
```

Deliberately no `default`: a `default` re-covers every input, so the compiler stops caring and we are back to a runtime-only signal. And as you noted, the same line covers your second point — a null selector is now an `IllegalArgumentException` naming the provider instead of an NPE, still unreachable today since both call sites pass a constant (`:155`, `:219`).

The comment is rewritten to state what the compiler actually enforces, and why the `default` is absent, so the next person adding a constant meets the build rather than the prose.

One thing I checked before choosing this over the switch expression: `case null` is already used in ~20 prettier-checked files under `gravitee-apim-rest-api` (`ApiCRDMapper`, `PlanAdapter`, `ImportExportApiMapper`, …), so prettier-java 2.7.5 from `gravitee-parent:25.1.0` parses it — and the `prettier:check` on both modules passed locally, so this is not a formatter risk.

Same change applied to the portal copy, as you asked. Verified: management `OAuth2AuthenticationResourceTest` 18/18, portal 10/10, both modules BUILD SUCCESS.

paulatulis (approved):
Other than sonar failure, LGTM!

Two non-blocking nits: 
1. SocialIdentityProviderImpl warn-dedup never clears on success. Verified with a probe: typo → warns; admin fixes it → resolves fine but the map still holds "bogus"; admin re-introduces the same typo → silently no warning. One line (reportedClientAuthenticationMethods.remove(id) on the success path) 
2. clientAuthenticationFailure: the >= 500 → identity_provider_unavailable default is unconditionally overridden by any JSON error in the body, so the outage-vs-rejection separation the comment promises only holds for non-JSON 5xx bodies — which is exactly what the test stubs. Narrow, arguably fine behaviour.

PR #19256, opened 2026-08-19
feat(console): configure persisted IdP claims and DCR mappings

wbabyte:
## Issue

https://gravitee.atlassian.net/browse/APIM-13954

(Reopened — it had been closed Won't Do, leaving the epic's configuration API-only.)

## Description

Both halves of the claims-to-DCR epic shipped without any console UI. That was more than a convenience gap: the API **preserves a field it does not receive** (`IdentityProviderServiceImpl.java:164`, `ClientRegistrationServiceImpl.java:249`), so an administrator working through the console could never *clear* a claims whitelist — the one action you need after realising a sensitive claim was being captured.

Two new sections, both saving through the existing endpoints:

- **Settings ▸ Identity Providers ▸ [provider] → Persisted Claims** — `gio-form-tags-input` for `persistedClaimsWhitelist`.
- **Settings ▸ Client Registration ▸ [provider] → Claim Mappings** — `gio-form-headers` for `claimMappings`.

### Things worth a reviewer's attention

**The whitelist is a top-level entity field, not provider configuration.** `IdentityProvider.java:86` puts it beside `configuration`, not inside it, so the control belongs to the *parent* form. Putting it in the OIDC sub-form — the obvious place — would have nested it under `configuration` and silently never persisted.

**Offered for every provider type.** I first gated the card to OIDC and then reversed that: `persistWhitelistedClaims` runs for every social provider and `SocialIdentityProviderImpl.convert` sets the whitelist for all types, so gating would have hidden a capability the API already supports — the very gap this ticket exists to close. The help text names all three claim sources (ID token, access token, user info endpoint) because GitHub has no ID token.

**Both `update()` field lists had to change.** `services-ngx/identity-provider.service.ts` and `client-registration-providers.service.ts` send an explicit whitelist of fields rather than the whole object. Without adding the new fields there they would round-trip on create and vanish on save — exactly what already happens to `requiredUrlParams`/`optionalUrlParams`. Emitting an empty collection rather than omitting the field is what makes clearing possible at all, and it means the console is now authoritative for these two fields.

**Reused, not reinvented.** Claim mappings use Particles `gio-form-headers` with the existing `toGioFormHeader` / `toDictionary` / `uniqueKeysValidator` utils, following the `additional_client_metadata` precedent (`application-general.component.ts:147-153`) instead of a hand-rolled FormArray.

**A trap for anyone else using `gio-form-headers`:** `headerFieldMapper` is a **data property mapper**, not display labels — `writeValue` reads `header[keyName]`. Passing `{ keyName: 'Claim' }` to relabel the column silently broke hydration and rendered an empty row. The default `{ key, value }` is what `toGioFormHeader` produces, so it must be left alone. Caught by the hydration test.

### Validation

Mirrors the server: claim names are checked for blankness only (`IdentityProviderServiceImpl.java:301`); DCR field paths reject **every standard registration field**, extension fields only (`ClientRegistrationServiceImpl.java:408`). The help text says so explicitly, including that `client_name` is rejected despite APIM-13952's AC2 listing it as supported.

Added `nonBlankEntriesValidator` to `util/gio-form-header.util.ts` (a new export — the ~20 other `gio-form-headers` usages are untouched). `toDictionary` keeps blank entries and the server validates only the field *path*, so a row filled on one side alone would have been stored under an empty claim name. Written test-first; the RED test confirmed it was previously accepted.

## Additional context

⚠️ **Security-sensitive**: exposes configuration that controls which PII-bearing IdP claims are retained on user records.

⚠️ **Compatibility-sensitive**: changes the submitted payload of two existing forms — both `PUT`s now carry an additional field, and the console becomes authoritative for it.

**Merge-order note:** the Persisted Claims help text says emptying the list removes claims already stored, on the user's next login. That is only true once #19252 (APIM-14840) lands — same epic, but these PRs can merge in either order.

### Verification

- New specs: whitelist hydrates, an added claim saves, emptying it sends `[]`, it is offered for a non-OIDC provider; claim mappings hydrate, an added mapping saves as a dictionary, a blank-named mapping blocks save.
- Collateral fix: `services-ngx/identity-provider.service.spec.ts` asserts the exact `PUT` body.
- IdP spec 21/21, client registration spec 14/14, util spec 15/15, both service specs 9/9.
- **Full console suite green: 518 suites, 4803 tests, 0 failures.** Formatter (`nx format:check`) and eslint clean.
- Not caused by this change: `api-import-file-picker` and `no-mcp-entrypoint` specs failed in one earlier parallel run and pass in isolation and in the final run; they touch nothing here.

Completes the epic's console surface alongside #19252 (APIM-14840) and #19254 (APIM-14841). APIM-13958 (documentation) remains open.


gravitee-apim-console-webui/src/entities/client-registration-provider/clientRegistrationProvider.ts
```diff
@@ -31,6 +31,7 @@ export class ClientRegistrationProvider {
   public renew_client_secret_method: string;
   public software_id: string;
   public updated_at: number;
+  public claimMappings?: Record<string, string>;
```
paulatulis: **This field never binds server-side, so the whole claim-mappings half of the PR is a silent no-op.**

The console declares `claimMappings`, but every server DTO maps it as `claim_mappings` — `@JsonProperty("claim_mappings")` on `ClientRegistrationProviderEntity:71`, `NewClientRegistrationProviderEntity:69` and `UpdateClientRegistrationProviderEntity:66`, all unchanged on this branch. It is also the one camelCase outlier in this interface: `discovery_endpoint`, `client_id`, `renew_client_secret_method`, `software_id`, `trust_store` and `key_store` are all snake_case.

All three directions break:

- **GET** — the server emits `claim_mappings`, the component reads `clientRegistrationProvider?.claimMappings` → `undefined` → configured mappings are never displayed, the form always renders empty.
- **PUT** — the server binds `claimMappings` to nothing, so `getClaimMappings()` is null, so the preserve-on-null branch at `ClientRegistrationServiceImpl:249` keeps the old value. Every save, *including an attempted clear*, is a no-op while the UI reports success. (`FAIL_ON_UNKNOWN_PROPERTIES` is disabled, so there is no 400 to give it away.)
- **POST** — mappings entered at creation are silently dropped.

The consequence is the part that worries me most: an administrator who removes a mapping to stop injecting personal data into DCR requests gets a success toast, and the gateway keeps injecting it.

The specs pass because they assert the same wrong key (`req.request.body.claimMappings` at `client-registration-provider.component.spec.ts:123,163,190`) — the mocked HTTP layer never round-trips through Jackson, so nothing in the suite can see this.

The identity-provider half is fine, for contrast: `UpdateIdentityProviderEntity` has no `@JsonProperty` and there is no snake_case strategy on those endpoints, so `persistedClaimsWhitelist` binds correctly and its clear path works end to end.

Renaming console-side to `claim_mappings` (here, in the service payload, the component read/write, and the spec assertions) matches the sibling fields; renaming the server DTOs instead would break an already-published REST contract.

gravitee-apim-console-webui/src/services-ngx/client-registration-providers.service.ts
```diff
@@ -62,6 +62,7 @@ export class ClientRegistrationProvidersService {
         renew_client_secret_endpoint: clientRegistrationProvider.renew_client_secret_endpoint,
         renew_client_secret_method: clientRegistrationProvider.renew_client_secret_method,
         software_id: clientRegistrationProvider.software_id,
+        claimMappings: clientRegistrationProvider.claimMappings,
```
paulatulis: This is the second place the `claimMappings` / `claim_mappings` wire-name mismatch needs fixing — see the detail on `clientRegistrationProvider.ts:34`. The key sent here binds to nothing on the server.

Separately, and part of why this was able to slip through: `client-registration-providers.service.spec.ts` ("should update provider", line 107) asserts only the method, the URL and the flushed response — there is no `req.request.body` assertion anywhere in that spec. So this explicit field allowlist is entirely unpinned, and dropping or misnaming any field in it (claimMappings, trust_store, key_store, …) passes silently.

`identity-provider.service.spec.ts` does assert the full update body with `toStrictEqual`, which is why the equivalent change on that side is protected. Mirroring that assertion here would have caught this one.

gravitee-apim-console-webui/src/organization/configuration/identity-provider/org-settings-identity-provider.component.html
```diff
@@ -147,6 +147,30 @@ <h2 gioTableOfContents>General</h2>
...
 
+    @if (identityProviderType) {
+      <mat-card class="identity-provider__card">
+        <mat-card-content>
+          <h2 gioTableOfContents>Persisted Claims</h2>
+
+          <p class="identity-provider__card__description">
+            Claims listed here are read at each login, from the ID token, the access token or the user info endpoint, and stored on the user
+            profile, where a client registration provider can inject them into dynamic client registration requests. Only list the claims
+            you need: everything stored here is personal data. Emptying the list stops the capture and removes the claims already stored, on
```
paulatulis: This copy promises something the platform does not currently do: "removes the claims already stored, on the user's next login." On the code in this repo, `UserServiceImpl.persistWhitelistedClaims` returns early when the whitelist is null **or empty**, leaving `user.idpClaims` untouched — emptying the list stops the refresh but never purges what was already captured.

You already call this out as a merge-order note, so the gap isn't that it was missed. My concern is the conclusion that the PRs "can merge in either order." There is no enforced ordering, so if this lands before #19252 there is a window where the console tells an administrator their personal data was removed when it was not — and the administrator's whole reason for emptying the list is usually that they realised a sensitive claim was being captured. That is the worst case in which to be wrong, and nothing signals it.

Either make #19252 a merge prerequisite, or promise only what is true today here and restore the purge wording in #19252 where the behaviour actually ships.

gravitee-apim-console-webui/src/management/settings/client-registration-providers/client-registration-provider/client-registration-provider.component.spec.ts
```diff
@@ -178,6 +181,70 @@ describe('ClientRegistrationProvider', () => {
...
+        .expectOne({
+          method: 'GET',
+          url: `${CONSTANTS_TESTING.env.baseURL}/configuration/applications/registration/providers/${provider.id}`,
+        })
+        .flush(provider);
+      fixture.detectChanges();
+      await fixture.whenStable();
+      fixture.detectChanges();
+    }
+
```
paulatulis, line 201: The happy-path claim-mapping saves call `fixture.componentInstance.onSubmit()` directly rather than driving `GioSaveBarHarness`, and the only save-bar assertion among the new tests is the *invalid* case (`isSubmitButtonInvalid() === true` for a blank key).

That combination leaves a real gap: if the new `nonBlankEntriesValidator` / `uniqueKeysValidator` wiring wrongly marked a valid form invalid, the Save button in the actual UI would be permanently disabled — and every test here would still pass, because the invalid-case test passes trivially and the save tests bypass the bar entirely.

The org-settings spec gets this right — it asserts `isSubmitButtonInvalid() === false` and then `clickSubmit()`. (The pre-existing tests in this file that call `onSubmit()` directly are established style and not worth changing.)

gravitee-apim-console-webui/src/management/settings/client-registration-providers/client-registration-provider/client-registration-provider.component.spec.ts
```diff
@@ -178,6 +181,70 @@ describe('ClientRegistrationProvider', () => {
     });
   });
 
+  describe('claim mappings', () => {
```
paulatulis, line 184: No test removes an existing claim mapping through the form. The "empty dictionary" test starts from a provider that already has `claimMappings: {}` and submits, so the remove interaction — deleting a configured row and seeing the key drop out of the payload — is never exercised.

That is the one path most likely to misbehave: removal is where a trailing empty row in `gio-form-headers` could leak an `"": ""` entry through `toDictionary`.

It is also asymmetric — the identity-provider side does test removal (`removeTag('org_id')`), so the two halves of the same feature are pinned to different standards.

gravitee-apim-console-webui/src/util/gio-form-header.util.ts
```diff
@@ -34,6 +35,23 @@ export function toDictionary(headers: Header[] | undefined): Record<string, stri
   return headers.reduce((acc, { key, value }) => ({ ...acc, [key]: value }), {});
 }
 
+/**
+ * Rejects entries where either side is blank. `toDictionary` keeps them, so a half-filled row would otherwise be sent
+ * as an empty key or an empty value.
+ */
+export function nonBlankEntriesValidator(): ValidatorFn {
```
paulatulis, line 42: The `!headers` guard is never executed by the spec. The suite covers `[]`, blank key, blank value, whitespace-only key and fully-populated entries — but `![]` is false, so the empty-array case falls straight past this branch and the null/undefined control value is never exercised.

A one-line `nonBlankEntriesValidator()(new FormControl(null))` case would cover it. (`uniqueKeysValidator` has the same untested guard, pre-existing.)

gravitee-apim-console-webui/src/organization/configuration/identity-provider/org-settings-identity-provider.component.spec.ts
```diff
@@ -495,6 +495,83 @@ describe('OrgSettingsIdentityProviderComponent', () => {
       fixture.detectChanges();
     });
 
+    describe('persisted claims', () => {
```
paulatulis, line 498: Nothing asserts that either new section renders its explanatory message and examples, so the copy could be deleted or detached from its section without a single test failing.

Normally that would be a minor point about hint text. Here it is worth more: the Persisted Claims copy is the only thing telling an administrator what emptying the list does to already-stored personal data, and it currently overstates that (see the comment on `org-settings-identity-provider.component.html:158`). Copy carrying a privacy claim is closer to behaviour than decoration, and is worth pinning.

gravitee-apim-console-webui/src/management/settings/client-registration-providers/client-registration-provider/client-registration-provider.component.html
```diff
@@ -314,6 +314,32 @@ <h3>Renew client_secret (outside DCR specification)</h3>
...
+      </gio-form-headers>
+      @if (providerForm.get('claimMappings')?.hasError('nonUniqueKeys')) {
+        <mat-error>Claim names must be unique</mat-error>
+      }
+      @if (providerForm.get('claimMappings')?.hasError('blankEntries')) {
+        <mat-error>Every mapping needs both a claim name and a registration request field</mat-error>
+      }
+      <p class="client-registration-provider__card__content__description">
+        Only extension fields can be written, using dot notation for nested ones such as <code>metadata.organization</code>. Standard
+        registration fields defined by the specification, <code>client_name</code> among them, are rejected.
```
paulatulis, line 338: Recording the console/server validation drift rather than objecting to it — I think the call you made is the right one.

The console validates uniqueness and non-blankness; the server additionally rejects any field path whose top-level segment is a standard DCR field (`ClientRegistrationServiceImpl.validateClaimMappings:408`). So `client_name` passes console validation and fails at save. Since the server derives that set at runtime by introspecting `ClientRegistrationRequest`, the console cannot mirror it without a new endpoint, and hardcoding a blocklist here would drift from it. Keeping the console permissive with the help text as the guard is the better trade.

One thing worth carrying back to the epic: **APIM-13952's AC2 lists `client_name` as a supported mapping target, but the shipped server rejects it.** Your help text sides with the code, which is correct — the AC is what should be amended, and right now the ticket and the product disagree.

gravitee-apim-console-webui/src/organization/configuration/identity-provider/org-settings-identity-provider.component.html
```diff
@@ -147,6 +147,30 @@ <h2 gioTableOfContents>General</h2>
       ></org-settings-identity-provider-graviteeio-am>
     </ng-container>
 
+    @if (identityProviderType) {
```
paulatulis, line 150: Coordination note, not a defect in this PR.

Sibling PR #19254 needs a `tokenEndpointAuthMethod` control on this same identity-provider form — without one, the key it adds is silently erased by this form's `onSubmit` spread, since `configuration` is rebuilt wholesale from the enumerated controls. This PR restructures exactly this area of the template.

Whichever of the two lands second will need to rebase onto the other's template changes. Landing this one first and having #19254 add its control on the rebased template seems cleanest — a placeholder control here would just be dead UI.

gravitee-apim-console-webui/src/entities/client-registration-provider/clientRegistrationProvider.ts
wbabyte: You are right, and this was mine — thank you for catching it. I verified all three DTOs: `ClientRegistrationProviderEntity:71`, `NewClientRegistrationProviderEntity:69` and `UpdateClientRegistrationProviderEntity:66` all carry `@JsonProperty("claim_mappings")`, unchanged on this branch. The console half was a silent no-op in all three directions exactly as you describe.

Renamed console-side to `claim_mappings`, as you recommend — the server DTOs are a published contract and it matches the snake_case siblings on the same interface. That covers the entity, the service payload, the form control name, the template, and the spec assertions.

The consequence you singled out is the one that made this urgent: an administrator removing a mapping to stop injecting personal data got a success toast while the gateway kept injecting it.

And you are right about why the suite could not see it. The fix for that is on your `client-registration-providers.service.ts` comment.

gravitee-apim-console-webui/src/services-ngx/client-registration-providers.service.ts
wbabyte: Fixed here too, and I have closed the hole that let it through: `client-registration-providers.service.spec.ts` now asserts `req.request.body` with `toStrictEqual` across the whole 15-field allowlist, mirroring `identity-provider.service.spec.ts`.

I confirmed your diagnosis — that spec asserted only method, URL and the flushed response, so the allowlist was entirely unpinned and any dropped or misnamed field passed silently. With the assertion in place, dropping `claim_mappings`, `trust_store` or any other field now fails there.

Your observation that the identity-provider side was protected by exactly this assertion is what made the fix obvious.

gravitee-apim-console-webui/src/organization/configuration/identity-provider/org-settings-identity-provider.component.html
wbabyte: Agreed, and I have taken your first option rather than relying on merge order.

The copy now promises only what is true on this branch: *"Emptying the list stops new claims being captured."* No claim about removing what was already stored.

You are right that "can merge in either order" was the wrong conclusion. There is no enforced ordering, and the window you describe — telling an administrator their personal data was removed when it was not, at the exact moment they are acting on a sensitive claim — is not a risk worth taking for a sentence.

I have not moved the purge wording into #19252, because that branch predates this card and the paragraph does not exist there to edit. So the copy stays truthful-today permanently, which is never wrong, and adding the stronger sentence once both have landed is an optional follow-up rather than a requirement.

gravitee-apim-console-webui/src/management/settings/client-registration-providers/client-registration-provider/client-registration-provider.component.spec.ts
wbabyte: Fixed. The happy-path claim-mapping save now gets the save bar, asserts `isSubmitButtonInvalid() === false`, and clicks submit — so the scenario you describe, a validator wrongly marking a valid form invalid and disabling Save in the real UI, now fails the test instead of passing trivially.

Thanks for noting that the pre-existing `onSubmit()` tests in this file are established style and not worth changing; I left those alone and only changed the ones this PR added.

gravitee-apim-console-webui/src/management/settings/client-registration-providers/client-registration-provider/client-registration-provider.component.spec.ts
wbabyte: Added, and it immediately earned its keep — but not in the way either of us expected.

`should_drop_a_removed_mapping_from_the_payload` removes a configured row and asserts the key is gone from the payload, with `{}` sent rather than an `"": ""` entry leaking through `toDictionary`. That part passes.

What it surfaced is that **the remove button submits the form on its own**. `gio-form-headers`' remove button carries no `type="button"`, so inside a form with an `(ngSubmit)` handler the click triggers implicit submission. I verified it directly: one `PUT` fires from the remove click alone, before any Save. It is latent in the other ~20 usages only because those forms have no `(ngSubmit)` — `application-general`'s does not — and live here because this one does.

So removing a mapping currently persists the whole form, and if the form is invalid `onSubmit()` throws. The fix is one attribute in `@gravitee/ui-particles-angular`, so I have not worked around it here; the test asserts the resulting payload rather than clicking Save, which keeps it valid whichever way the submit is triggered. Worth a bug against Particles — flagged to the requester.

gravitee-apim-console-webui/src/util/gio-form-header.util.ts
wbabyte: Fixed — added `nonBlankEntriesValidator()(new FormControl(null))`, since as you say `![]` is false and the empty-array case walked straight past the guard.

I left `uniqueKeysValidator`'s identical guard untested rather than expanding the diff into pre-existing code, but it is a one-liner if you would rather have both covered here.

gravitee-apim-console-webui/src/organization/configuration/identity-provider/org-settings-identity-provider.component.spec.ts
wbabyte: Agreed with the reasoning, and both sections now have their copy pinned via `data-testid`.

The persisted-claims assertion checks the text says *personal data* and states what emptying the list does — for the reason you give: that sentence is the only thing telling an administrator what happens to data already captured, which makes it closer to behaviour than decoration.

The claim-mappings assertion checks the guidance names *"Only extension fields can be written"* and `client_name`, which is the console's only guard against a validation rule it cannot mirror.

And you were right that the copy overstated things — that is fixed on the `:158` thread, so what is now pinned is a statement that is actually true.

gravitee-apim-console-webui/src/management/settings/client-registration-providers/client-registration-provider/client-registration-provider.component.html
wbabyte: Thanks for recording it rather than objecting — and agreed on the trade-off. The server derives that set at runtime by introspecting `ClientRegistrationRequest`, so any console-side blocklist would drift from it the moment a field is added, and a new endpoint to expose it is a lot of machinery for a validation hint. Keeping the console permissive with the help text as the guard is the call I would make again.

The copy is now pinned by a test (see the `org-settings-identity-provider.component.spec.ts:498` thread), so the guidance cannot be silently detached from the section it explains.

On carrying it back to the epic: fully agree, and I have raised it with the requester — **APIM-13952's AC2 lists `client_name` as a supported mapping target while the shipped server rejects it**, so the AC is what should be amended, not the code. I could not edit it myself; the Atlassian connection dropped partway through this session, so it needs a human to apply.

gravitee-apim-console-webui/src/organization/configuration/identity-provider/org-settings-identity-provider.component.html
wbabyte: Resolved, and I took your recommendation on placement.

The `tokenEndpointAuthMethod` control lives in **#19254**, not here — a control in this PR would have been dead UI, as you say, since nothing reads the key until that branch lands. It went onto the OIDC child form and the Gravitee AM child form, since both feed the same wholesale `configuration` rebuild you describe. There is a regression test there that I verified fails without the control (`Cannot find control with name: 'tokenEndpointAuthMethod'`).

That also keeps the two PRs mergeable in either order: #19254 carries both the key and its control, so it is self-contained, and this PR never references the key.

On the rebase: confirmed, and it will be needed. #19254 touches `org-settings-identity-provider.component.spec.ts` too — adding the control changes the asserted `configuration` payload in the OIDC and AM tests — and this PR modifies the same spec. The conflict is confined to that file plus the child components, which this PR does not touch. Whichever lands second rebases; I have noted it on #19254 as well.

paulatulis (approved):
Approving, but please note two things:
- Please raise the Particles bug and link it here, so this thread isn't the only record. Worth saying in it that the other ~20 usages aren't safe by design — they're latent only because those forms have no `(ngSubmit)`, and any one of them becomes live the moment its form gains one.
- Your call to assert the resulting payload rather than clicking Save was the right one. It keeps `should_drop_a_removed_mapping_from_the_payload` valid whichever way the submit ends up being triggered, so the test survives the Particles fix instead of needing rewriting after it.

PR #19257, opened 2026-08-19
ci(sonar): analyse the reference branch in the nightly build

phiz71:
## Issue

https://gravitee.atlassian.net/browse/BX-300

## Description

Nothing has analysed `master` since **14 August**. The workflow refactoring reduced a push on a branch to refreshing the development environment, so the per-module suites — and the analyses that depend on them — no longer run there. Two consequences, both invisible until you look for them:

- a pull request targeting `master` has **no new-code baseline**, so its quality gate compares against nothing;
- the server-side cache finds nothing to reuse — `0 out of 5,514 files` in the scanner log — so every pull request analysis re-parses the whole project.

This adds a second chain to the nightly build: the suites that produce the coverage reports, each followed by the analysis that reads them. They are replayed for their reports, not to retest what pull requests already cover.

```mermaid
flowchart LR
  subgraph env["Environment chain — unchanged"]
    direction LR
    setup["Setup"] --> build["Build backend"]
    build --> imgback["Backend docker images"]
    front["Frontend builds"] --> imgfront["Frontend docker images"]
    build --> integ["Integration tests"]
    build --> sdk["Generate e2e tests SDK"] --> e2ebuild["Lint & Build APIM e2e"]
    e2ebuild --> e2e["E2E (matrix)"]
    e2ebuild --> cypress["Run Cypress UI tests"]
    imgback --> e2e
    imgback --> cypress
    imgfront --> cypress
    integ --> deploy["Deploy on Azure cluster"]
    cypress --> deploy
    imgback --> aikido["Scan docker images with Aikido"]
    imgfront --> aikido
  end

  subgraph ana["Analysis chain — added"]
    direction LR
    testback["6 backend suites"] --> sonarback["6 SonarCloud analyses"]
    testfront["3 frontend suites"] --> sonarfront["3 SonarCloud analyses"]
  end

  build --> testback
```

## Additional context

**One junction, and nothing waits on an analysis.** The six backend suites hang off the `Build backend` the workflow already runs — the same way `Integration tests` does — so nothing is recompiled. The three frontend suites are roots: `Lint & test APIM Console` does not depend on `Build APIM Console`. And the nine analyses are leaves: no job lists them in `requires`, so a slow or failing analysis can neither delay the deployment nor block the image scans in the dependency graph. Runner contention is a separate matter: nine extra jobs queued alongside the e2e matrix can still push the deployment back under org concurrency limits.

**The jobs are spelled out rather than borrowed from the pull-request groups.** `backendJobs` and `frontendJobs` carry a changed-files filter and a build-and-publish scope this workflow has no use for, and reusing them would have meant either declaring `Setup` and `Build backend` twice, or reordering the jobs the pull-request fixtures pin. Only `workflow-nightly.ts` and its own fixture change here — **the other 152 test cases pass untouched, so the pull-request YAML is identical byte for byte.**

**Cost**, from the durations observed today:

| | today | with the analyses |
| --- | --- | --- |
| declared jobs | 22 | 40 |
| wall clock | 31 min | + 0 to 16 min |
| cumulative machine time | 161 min | ≈ 240 min |

Those are **per-branch** figures: the nightly is one scheduled pipeline per branch, so a support branch that gets its own schedule pays them again.

The longest added branch is `Build backend → Test rest-api (629 s) → Sonar rest-api (319 s)`, about sixteen minutes after the build, against a workflow that already takes thirty-one. It has a fair chance of running in the shadow of the end-to-end chain — the first run will tell.

**Support branches are unaffected by this PR.** `workflow-nightly.ts` exists only on `master` — the four live support branches do not carry it, and they still analyse on every push because they do not carry the workflow refactoring either. Nothing here runs on a branch by itself: `nightly` is a `ci_action` value, so a branch is analysed only once a schedule is created for it. `4.13.x`, born from `master`, will need that schedule like any other.

### DIFF
--- a/.circleci/ci/src/workflows/workflow-nightly.ts
+++ b/.circleci/ci/src/workflows/workflow-nightly.ts
@@ -13,8 +13,20 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { Config, workflow, Workflow } from '../circleci-config';
-import { AikidoScanDockerImagesJob, DeployOnAzureJob, TestIntegrationJob } from '../jobs';
+import { Config, Job, workflow, Workflow } from '../circleci-config';
+import {
+  AikidoScanDockerImagesJob,
+  DeployOnAzureJob,
+  SonarCloudAnalysisJob,
+  TestDefinitionJob,
+  TestGatewayJob,
+  TestIntegrationJob,
+  TestPluginJob,
+  TestReporterJob,
+  TestRepositoryJob,
+  TestRestApiJob,
+  WebuiLintTestJob,
+} from '../jobs';
 import { CircleCIEnvironment } from '../pipelines';
 import { config } from '../config';
 import { devEnvironmentJobs } from './groups/dev-environment-jobs';
@@ -24,11 +36,19 @@ import { chainguardFipsImageJobs } from './groups/chainguard-fips-image-jobs';
 /**
  * The scheduled build, one per branch: the default branch and every live support branch.
  *
- * It is the dev-environment refresh plus the two suites that run nowhere else. Pull requests
- * cover their own scope — the changed-files predicates are dependency-aware, so anything a
- * change could affect is retested before merge — and a push to a branch only refreshes the
- * environment. What no one runs, then, is the real-plugin integration tests and the end-to-end
- * suites. It deliberately does not replay the per-module or frontend test suites.
+ * Two chains hang off a single `Build backend`, for two unrelated purposes.
+ *
+ * The first refreshes the development environment: the web UIs and the docker images the
+ * cluster runs, then the end-to-end suites and the deployment.
+ *
+ * The second gives SonarCloud an analysis of the reference branch. A push to a branch only
+ * refreshes the environment, so nothing has analysed the default branch since that became
+ * true — and without it a pull request has no new-code baseline and the server-side cache
+ * finds nothing to reuse. The suites are replayed here for the coverage reports they leave
+ * behind, not to retest what pull requests already covered.
+ *
+ * The jobs are spelled out rather than borrowed from the pull-request groups: those carry a
+ * changed-files filter and a build-and-publish scope this workflow has no use for.
  */
 export class NightlyWorkflow {
   static create(dynamicConfig: Config, environment: CircleCIEnvironment): Workflow {
@@ -37,6 +57,30 @@ export class NightlyWorkflow {
 
     const deployOnAzureJob = DeployOnAzureJob.create(dynamicConfig, environment);
     dynamicConfig.addJob(deployOnAzureJob);
+
+    // The analysis chain. One parameterized analysis job, reused with a different working
+    // directory; one test job per module, each waiting on the shared `Build backend`.
+    const sonarAnalysisJob = SonarCloudAnalysisJob.create(dynamicConfig, environment);
+    dynamicConfig.addJob(sonarAnalysisJob);
+
+    const backendSuites: { job: Job; name: string; project: string }[] = [
+      { job: TestDefinitionJob.create(dynamicConfig, environment), name: 'Test definition', project: 'gravitee-apim-definition' },
+      { job: TestGatewayJob.create(dynamicConfig, environment), name: 'Test gateway', project: 'gravitee-apim-gateway' },
+      { job: TestRestApiJob.create(dynamicConfig, environment), name: 'Test rest-api', project: 'gravitee-apim-rest-api' },
+      { job: TestPluginJob.create(dynamicConfig, environment), name: 'Test plugins', project: 'gravitee-apim-plugin' },
+      { job: TestReporterJob.create(dynamicConfig, environment), name: 'Test reporters', project: 'gravitee-apim-reporter' },
+      { job: TestRepositoryJob.create(dynamicConfig, environment), name: 'Test repository', project: 'gravitee-apim-repository' },
+    ];
+    backendSuites.forEach(({ job }) => dynamicConfig.addJob(job));
+
+    const consoleLintTestJob = WebuiLintTestJob.createNx(dynamicConfig, environment);
+    dynamicConfig.addJob(consoleLintTestJob);
+
+    const portalNextLintTestJob = WebuiLintTestJob.createNx(dynamicConfig, environment);
+    dynamicConfig.addJob(portalNextLintTestJob);
+
+    const portalLintTestJob = WebuiLintTestJob.create(dynamicConfig, environment);
+    dynamicConfig.addJob(portalLintTestJob);
 
     return new Workflow('nightly', [
       ...devEnvironmentJobs(dynamicConfig, environment),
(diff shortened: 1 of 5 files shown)


.circleci/ci/src/workflows/workflow-nightly.ts
```diff
@@ -24,11 +36,19 @@ import { chainguardFipsImageJobs } from './groups/chainguard-fips-image-jobs';
...
- * cover their own scope — the changed-files predicates are dependency-aware, so anything a
- * change could affect is retested before merge — and a push to a branch only refreshes the
- * environment. What no one runs, then, is the real-plugin integration tests and the end-to-end
- * suites. It deliberately does not replay the per-module or frontend test suites.
+ * Two chains hang off a single `Build backend`, for two unrelated purposes.
+ *
+ * The first refreshes the development environment: the web UIs and the docker images the
+ * cluster runs, then the end-to-end suites and the deployment.
+ *
+ * The second gives SonarCloud an analysis of the reference branch. A push to a branch only
```
ytvnr, line 44: **Worth fixing** - The description says support branches are out of scope and still analyse on every push, but neither seems to hold.

The class docstring just above states the scheduled build runs "one per branch: the default branch and every live support branch", and nothing in the new code gates on the branch, so `4.13.x` gains the 18 jobs the moment this merges. On the push side, `workflow-pull-requests.ts:49` routes `isSupportBranchOrMaster` to `devEnvironmentJobs` plus chainguard plus publish only, with no Sonar job anywhere, so support branches lost their analysis exactly like `master` did.

That also makes the declared cost (22 to 40 jobs, around 240 min) a per-branch figure rather than a total. Worth deciding whether the chain is wanted on support branches: if yes, the description and the cost need correcting; if no, the block needs gating on the branch.

.circleci/ci/src/workflows/workflow-nightly.ts
```diff
@@ -50,6 +94,69 @@ export class NightlyWorkflow {
...
+      ...backendSuites.flatMap(({ job, name, project }) => [
+        new workflow.WorkflowJob(job, {
+          name,
+          context: config.jobContext,
+          requires: ['Build backend'],
+        }),
+        new workflow.WorkflowJob(sonarAnalysisJob, {
+          name: `Sonar - ${project}`,
+          context: config.jobContext,
+          requires: [name],
```
ytvnr, line 109: **Worth fixing** - Because each analysis requires its suite, a flaky test leaves that module with no `master` baseline until the next night.

CircleCI skips a job whose dependency failed. The suites run `mvn --fail-fast ... test`, so a single failure aborts the module and the jacoco aggregate is never produced or persisted, and the `Sonar - <project>` job never runs. The resulting state, no reference analysis and a cold server-side cache, is indistinguishable from the one this PR is fixing, and the only signal is a red nightly among other red jobs.

Worth deciding explicitly whether these replayed suites should tolerate failure so they still emit their reports, or adding a distinguishable alert when an analysis leaf is skipped, so a missing baseline is visible rather than inferred.

.circleci/ci/src/workflows/workflow-nightly.ts
```diff
@@ -38,6 +58,30 @@ export class NightlyWorkflow {
...
+    dynamicConfig.addJob(sonarAnalysisJob);
+
+    const backendSuites: { job: Job; name: string; project: string }[] = [
+      { job: TestDefinitionJob.create(dynamicConfig, environment), name: 'Test definition', project: 'gravitee-apim-definition' },
+      { job: TestGatewayJob.create(dynamicConfig, environment), name: 'Test gateway', project: 'gravitee-apim-gateway' },
+      { job: TestRestApiJob.create(dynamicConfig, environment), name: 'Test rest-api', project: 'gravitee-apim-rest-api' },
+      { job: TestPluginJob.create(dynamicConfig, environment), name: 'Test plugins', project: 'gravitee-apim-plugin' },
+      { job: TestReporterJob.create(dynamicConfig, environment), name: 'Test reporters', project: 'gravitee-apim-reporter' },
+      { job: TestRepositoryJob.create(dynamicConfig, environment), name: 'Test repository', project: 'gravitee-apim-repository' },
+    ];
```
ytvnr, line 73: **Worth fixing** - This mapping duplicates what `backend-jobs.ts` already hardcodes, and nothing asserts the two stay in sync.

The six entries mirror the suite factories and Sonar project names in `backend-jobs.ts`, and the three frontend blocks mirror `frontend-jobs.ts` parameter for parameter (`resource_class: xlarge`, `max-workers: '7'`, and so on). They match today, but a seventh backend module or a renamed Sonar project added to the pull-request groups would silently never be analysed on `master`. The only nightly test is a whole-file snapshot, which accepts any omission as long as the fixture is regenerated.

One option is to extract the (suite factory, workflow job name, sonar project, cache type) tuples into a shared list consumed by both. A cheaper one is a test asserting the set of `Sonar - *` job names in the nightly equals the set the unfiltered pull-request workflow produces.

.circleci/ci/src/workflows/workflow-nightly.ts
```diff
@@ -50,6 +94,69 @@ export class NightlyWorkflow {
...
+          requires: ['Build backend'],
+        }),
+        new workflow.WorkflowJob(sonarAnalysisJob, {
+          name: `Sonar - ${project}`,
+          context: config.jobContext,
+          requires: [name],
+          working_directory: project,
+          cache_type: 'backend',
+        }),
+      ]),
```
ytvnr, line 113: **Worth fixing** - Every job in this chain carries `cmd-notify-on-failure`, which on `master` posts to the general team channel.

`computeSlackChannelToUse` returns `apiManagementTeamNotifications` for `nightly`, and the command's `branch_pattern` is `master,[0-9]+\.[0-9]+\.x`, so each replayed suite can post a `basic_fail_1` message indistinguishable from a real regression, for work the description says is not meant to retest anything. The three other periodic actions (`repositories_tests`, `helm_tests`, `bridge_compatibility_tests`) already have dedicated channels, so routing this chain the same way would keep the noise out.

Separately, on the cost note in the description: "cannot delay deployment" is true of the dependency graph, but six machine and large executors plus three frontend jobs queued alongside the e2e matrix can still push `Integration tests`, Cypress and `Deploy on Azure` back under org concurrency limits. Worth phrasing as runner contention rather than only the absence of `requires` edges.

ytvnr (approved):
Adds a second chain to the nightly build so `master` gets a SonarCloud analysis again: nine suites replayed for their coverage reports, each followed by the analysis that reads them, all as leaves of the graph.

I checked the plumbing and it holds. The nine analyses really are leaves: `Deploy on Azure cluster` requires only `Integration tests` and `Run Cypress UI tests`, and the Aikido scan only the image jobs. Each suite persists its jacoco aggregate exactly where the matching `sonar-project.properties` reads it, the three frontend jobs likewise for `coverage/lcov.info`, all nine repo projects are covered with no omission and no extra, and no new secret or context is introduced.

Two points I would settle before merging, neither of them a blocker. The nightly is generated per branch, so the chain lands on every live support branch too, which contradicts the scope and the cost figure in the description. And each analysis sits behind `requires: [<suite>]`, so a single flaky test quietly reproduces the missing-baseline state this PR exists to fix.

The two others are smaller: the module to project mapping is duplicated from the pull-request groups with only a whole-file snapshot guarding it, and the 18 new jobs each notify the general team channel.

.circleci/ci/src/workflows/workflow-nightly.ts
phiz71: Checked, and the branch side turns out to be the reverse of the worry: nothing here runs on a branch by itself.

`nightly` is a value of the `ci_action` pipeline parameter, so a branch is only analysed once a scheduled pipeline is created for it. And `workflow-nightly.ts` exists **only on `master`** — none of the four live support branches carry it, so the chain cannot land on them at all. They also still analyse on every push, because they do not carry the workflow refactoring either: SonarCloud has analyses for `4.9.x` through `4.12.x` dated 17 and 18 August, against 14 August for `master`. The routing you quote is real, but it is `master` code that the support branches have not received.

`4.13.x` will need its own schedule like any other branch, which is when the decision gets made rather than inherited.

You are right about the figure though: 22 to 40 jobs and ~240 min are per-branch, not a total. Description corrected on that point, and reworded to say plainly that the chain runs where a schedule exists.

.circleci/ci/src/workflows/workflow-nightly.ts
phiz71: The mechanism is exactly as you describe, and it is not hypothetical — `Test gateway` and `Test rest-api` both failed on this branch today, on flaky tests unrelated to it.

We discussed it and decided to leave the suites strict for now. Where we disagree is on the size of the gap: a skipped analysis costs one night, not the open-ended state this PR fixes. The consequences are a server-side cache one day older, so slower pull-request analyses, and a new-code baseline shifted by a day, so a pull request briefly carries a few lines merged in between. Both are repaired by the next run. `master` had been without an analysis for five days when we started — that is a different order of magnitude.

What does need fixing is the visibility, which was the real point of your remark: a missed night has to be seen, not inferred. That is now the dedicated Slack channel below — a failing suite in the scheduled build posts where someone is watching for it, and relaunching is a click.

If a suite turns out to miss several nights in a row, the parameterised `-Dmaven.test.failure.ignore=true` you are pointing at is the way out: the jacoco aggregate is still produced, so the analysis happens anyway. Worth doing once we know it is needed rather than in advance.

.circleci/ci/src/workflows/workflow-nightly.ts
phiz71: Took the cheaper of your two options — the test is in `pipeline-nightly.spec.ts`, in the commit above.

It asserts that the set of `Sonar - *` job names in the nightly equals the set the unfiltered pull-request workflow produces, and that there are nine of them. I checked it actually guards: removing the `reporter` entry from `backendSuites` makes it fail, restoring it makes it pass. So a seventh module added to the pull-request groups now breaks the build instead of silently never being analysed.

I left the shared-tuple extraction aside on purpose. The frontend blocks mix lint-and-test, build, docker image and analysis in one `if`, so extracting the pairs would reorder the emitted jobs and rewrite eleven pull-request fixtures for a diff that changes no dependency edge. The test buys the same guarantee without touching what the fixtures pin.

.circleci/ci/src/workflows/workflow-nightly.ts
phiz71: Fixed — `#apim-ci-notifications-nightly` now exists and `computeSlackChannelToUse` routes `nightly` to it, alongside the three other periodic actions.

Worth noting it routes the **whole workflow**, not just the analysis chain: the integration tests, the e2e suites and the Azure deployment stop posting to the general channel too. That is consistent with how the other periodic actions behave, but it does move existing alerts, so it should not come as a surprise. The generated diff is a single line — `channel: C02JENTV2AX` becomes `C0BR8DK75N2`.

This also happens to answer your point above about a flaky suite: the missed analysis it causes now has somewhere to be seen.

On runner contention: you are right, and the description was overstating it. "Cannot delay the deployment" holds for the dependency graph only; nine extra jobs queued next to the e2e matrix can still push it back under org concurrency limits. Reworded to say exactly that.

.circleci/ci/src/workflows/workflow-nightly.ts
phiz71: Will open a new PR to implement the first option (instead of the cheapest) It allows me to merge this one before the nightly of today.

PR #19261, opened 2026-08-19
feat(gamma): invite group members by email and manage invitations

ankita-gupta21:
hem instead of duplicating the applications-feature copies.

## Issue

https://gravitee.atlassian.net/browse/FOUND-109

## Summary

Adds **email invitation** and **pending-invitation management** to Gamma group detail. User search already existed; this PR is the classic Console invite path that was still missing.

This is **FOUND-109 only**. Edit roles, remove member, ownership transfer, Explorer column, and delete-dialog rename stay on the stacked FOUND-107 branch.

## What this covers

On the group detail page:

- **Add members** is a dropdown, matching classic:
  - **User search** — disabled when `system_invitation` is false
  - **Email invitation** — disabled when `email_invitation` is false
- The button is shown when the user can update the group **or** the group is `manageable`, and at least one of search/email is enabled, and `max_invitation` is not already reached.
- **Members / Invitations** tabs. Invitations load **only after** the Invitations tab is selected (classic `onTabChange` / lazy tab).
- **Invite sheet:** email + default API role + default application role.
- **Invitations table:** email, API role, application role, invitation date, delete (when the user can manage).
- **Ambiguous email (HTTP 202):** “Many users found” dialog → Continue opens user search with that email prefilled (`initialSearch`).
- Delete invitation uses the shared `ConfirmDialog`.

## Classic Console alignment

Source of truth: `gravitee-apim-console-webui/.../groups/group/` (`group.component.ts` / `.html`, `invite-member-dialog`).

| Classic | Gamma |
| --- | --- |
| Add Members menu: User Search / Email Invitation | Same dropdown + disable flags |
| `canAddMembers = (canUpdate \|\| canInvite) && !maxReached` | Same, with `manageable` as the invite-admin path |
| Invite form: API role, application role, required email | `GroupInviteMemberSheet` — same three fields |
| Defaults: `group.roles.API/APPLICATION ?? USER` | Same |
| Locks: API/application locked unless `environment-group-u` | `getMemberRoleLockFlags` + `canOverrideLocks={canEdit}` |
| API: disable PRIMARY_OWNER when USER mode **or** an API PO already exists; other system roles disabled | `isPrimaryOwnerUnavailable` + `disableSystemRoles="except-primary-owner"` |
| Application: all system roles disabled | `disableSystemRoles="all"` |
| Invite has **no** API Product / Integration / Cluster / Explorer roles | Invite sheet stays API + Application only |
| POST `/invitations` → 200 success, 202 ambiguous users | `inviteGroupMember` treats an array body as `{ ambiguous: true }` |
| 202 → TooManyUsers → add-members dialog | `GroupTooManyUsersDialog` → search sheet + `searchSeed` |
| Invitations tab: email, API role, application role, date, delete | `GroupInvitationsTable` |
| Delete copy: “You are trying to delete an invitation sent to {email}…” | Same via `ConfirmDialog` |
| Unpaged invitations list, client search/pagination | Same; fetch gated on the tab to avoid the extra call on every detail load |

Intentionally **not** in this PR (classic has them; FOUND-107 will): edit member, remove member, ownership transfer, Explorer column on the members table.

## Code map for reviewers

**Page / orchestration**

- `pages/GroupDetailPage.tsx` — dropdown, tabs, sheets, dialogs. Does not own the mutations.
- `hooks/useGroupMemberActions.ts` — sheet/tab state, invite/add/delete invitation handlers, invitations query enabled only when `memberTab === 'invitations'`.
- `hooks/useGroupDetail.ts` — `useGroupInvitations`.
- `hooks/useGroupMutations.ts` — `useInviteGroupMember`, `useDeleteGroupInvitation`.
- `services/groups.ts` — POST/GET/DELETE `.../configuration/groups/{id}/invitations`. Ambiguous invite is detected as `Array.isArray(result)` because the 202 body is a user list (classic checks `response.status === 202`; `apimFetchJsonV1Env` only returns JSON).

**Invite UI**

- `components/GroupInviteMemberSheet.tsx` — Graphene `Sheet` / `Input` / `Label` / `GroupRoleSelect`. `aria-invalid` + “Enter a valid email”. Submit disabled until `isValidEmail`.
- `shared/utils/email.ts` — shared validator (also used from users).
- `components/GroupTooManyUsersDialog.tsx` — Graphene `Dialog` + `Alert` (classic Material dialog).
- `components/GroupInvitationsTable.tsx` — Graphene `DataTable`, client pagination, search, empty states. Single-action delete is an icon button (Graphene single-action exception).

**Reuse / DRY**

- `shared/layout/sheetLayout.ts`, `shared/utils/{clientPagination,dataTableTypes,paginationConstants,email}.ts`
- Applications feature files re-export those so existing imports keep working.
- `groups/utils/clientPagination.ts` re-exports the shared helpers.
- `isPrimaryOwnerUnavailable` lives in `memberRoles.ts` (invite + add-members). Edit/remove ownership helpers are **not** here.

**Add-members continuation**

- `useGroupAddMembersForm` / `GroupAddMembersSheet` accept `initialSearch` so Continue from 202 seeds the search field.

## Security

Invitation create/list/delete and membership add. Gating is the same as classic (env group update **or** `manageable`, plus `email_invitation` / `system_invitation` and max members). Backend still enforces `GROUP_INVITATION` / `ENVIRONMENT_GROUP` on the v1 resources.

## Out of scope

- Edit member roles, remove member, ownership transfer
- Explorer column on members
- Renaming `GroupDeleteSheet` → `GroupDeleteDialog`
- Paged invitations API (classic is unpaged too)

## Test plan

- `GroupInviteMemberSheet.spec.tsx` — validation, locks, PO / system-role disable, no “None” option
- `GroupInvitationsTable.spec.tsx` — columns, empty, delete
- `GroupDetailPage.spec.tsx` — dropdown wiring, invite success → Invitations tab, 202 → search seed, invitations query only after tab select
- `useGroupAddMembersForm.spec.tsx` — `initialSearch`
- `shared/utils/email.spec.ts`

Manual: invite a unique email (200 + row in Invitations), invite an email with multiple org users (202 → Continue → search), delete invitation, max members banner disables both menu items.

---

Classic has **no** invitation status (pending / accepted / rejected), **no** resend, **no** invitation history, and **no** invitation audit UI. The invitation entity is the same shape: `id`, email, roles, `created_at`.

FOUND-109 still lists those extras as acceptance criteria. They go beyond Console and need backend/API work, so they are **follow-ups** unless product explicitly reduces FOUND-109 to Console parity or expands this story beyond it.

| FOUND-109 AC | Classic Console | This PR |
| --- | --- | --- |
| Send invitations | Yes | Yes |
| Cancel pending invitations | Yes (delete) | Yes |
| Track status (pending / accepted / rejected) | No | Follow-up |
| Resend invitations | No | Follow-up |
| Display invitation history | No | Follow-up |
| Audit event for invitations | No | Follow-up |

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/sheetLayout.ts
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/sheetLayout.ts
@@ -13,9 +13,4 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-/** Default right sheet width (30rem / 480px). */
-export const STANDARD_SHEET_WIDTH = '480px';
-
-/** Wider sheet for multi-column forms or category sections (48rem / 768px). */
-export const WIDE_SHEET_WIDTH = '48rem';
+export { STANDARD_SHEET_WIDTH, WIDE_SHEET_WIDTH } from '../../../shared/layout/sheetLayout';
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/utils/dataTableTypes.ts
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/utils/dataTableTypes.ts
@@ -13,10 +13,4 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import type { DataTableColumnHeaderProps } from '@gravitee/graphene-core';
-
-/** Column header render props for Graphene `DataTable` column definitions. */
-export type ColHeader<T> = { column: DataTableColumnHeaderProps<T, unknown>['column'] };
-
-/** Column cell render props for Graphene `DataTable` column definitions. */
-export type ColCell<T> = { row: { original: T } };
+export type { ColCell, ColHeader } from '../../../shared/utils/dataTableTypes';
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/utils/paginationConstants.ts
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/utils/paginationConstants.ts
@@ -14,8 +14,7 @@
  * limitations under the License.
  */
 
-/** Shared page-size options for application tables (UI-COMP-12). */
-export const TABLE_PAGE_SIZE_OPTIONS = [10, 25, 50, 100] as const;
+export { TABLE_PAGE_SIZE_OPTIONS } from '../../../shared/utils/paginationConstants';
 
 export const DEFAULT_APPLICATION_LIST_PAGE_SIZE = 25;
 
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupAddMembersSheet.spec.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupAddMembersSheet.spec.tsx
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 import { useQuery } from '@tanstack/react-query';
-import { act, fireEvent, render, screen } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
 
 import { GroupAddMembersSheet } from './GroupAddMembersSheet';
+import { installFormActionTestEnvironment } from '../../../shared/testing/formAction';
 import type { SearchableUser } from '../../../shared/types/userSearch';
 import type { GroupMember } from '../types/group';
 import { GROUP_SEARCH_DEBOUNCE_MS } from '../utils/paginationConstants';
@@ -26,8 +27,15 @@ jest.mock('@tanstack/react-query', () => ({
     useQuery: jest.fn(),
 }));
 
+let restoreTestEnvironment: () => void;
+
 beforeAll(() => {
     Element.prototype.scrollIntoView = jest.fn();
+    restoreTestEnvironment = installFormActionTestEnvironment();
+});
+
+afterAll(() => {
+    restoreTestEnvironment();
 });
 
 const mockUseQuery = jest.mocked(useQuery);
@@ -42,6 +50,13 @@ function typeSearch(value: string) {
     fireEvent.change(screen.getByPlaceholderText('Search by name or email…'), { target: { value } });
     act(() => {
         jest.advanceTimersByTime(GROUP_SEARCH_DEBOUNCE_MS);
+    });
+}
+
+async function submitMembers() {
+    await act(async () => {
+        fireEvent.click(screen.getByRole('button', { name: 'Add member' }));
+        await Promise.resolve();
     });
 }
 
@@ -70,7 +85,6 @@ function renderSheet(overrides: Partial<React.ComponentProps<typeof GroupAddMemb
             maxInvitation={null}
             onClose={onClose}
             onSubmit={onSubmit}
-            isSaving={false}
             {...overrides}
         />,
     );
@@ -109,6 +123,11 @@ describe('GroupAddMembersSheet', () => {
         expect(screen.getByText('Explorer')).not.toBeNull();
     });
 
+    it('seeds the search field from initialSearch', () => {
+        renderSheet({ initialSearch: 'anna@lufthansa.com' });
+        expect(screen.getByPlaceholderText('Search by name or email…')).toHaveProperty('value', 'anna@lufthansa.com');
+    });
+
     it('has no "make selected users group admins" option — that only applies when editing an existing member', () => {
         renderSheet();
         expect(screen.queryByText(/group admin/i)).toBeNull();
@@ -134,13 +153,13 @@ describe('GroupAddMembersSheet', () => {
         expect(screen.getByRole('button', { name: 'Add member' })).toHaveProperty('disabled', true);
     });
 
-    it('submits selected users with all classic group-member role scopes', () => {
+    it('submits selected users with all classic group-member role scopes', async () => {
         const { onSubmit } = renderSheet();
 
         typeSearch('an');
         fireEvent.click(screen.getByText('Anna Schmidt'));
 
-        fireEvent.click(screen.getByRole('button', { name: 'Add member' }));
+        await submitMembers();
 
         expect(onSubmit).toHaveBeenCalledWith([
             {
@@ -158,12 +177,12 @@ describe('GroupAddMembersSheet', () => {
         ]);
     });
 
-    it('omits id for LDAP users so the backend resolves via reference', () => {
+    it('omits id for LDAP users so the backend resolves via reference', async () => {
         const { onSubmit } = renderSheet();
 
         typeSearch('ld');
         fireEvent.click(screen.getByText('LDAP Anna'));
-        fireEvent.click(screen.getByRole('button', { name: 'Add member' }));
+        await submitMembers();
 
         expect(onSubmit).toHaveBeenCalledWith([
             {
@@ -181,12 +200,12 @@ describe('GroupAddMembersSheet', () => {
     });
 
     describe('default role pre-fill', () => {
-        it('pre-fills API/API product/Application from the group’s configured default roles', () => {
+        it('pre-fills API/API product/Application from the group’s configured default roles', async () => {
             const { onSubmit } = renderSheet({ groupRoles: { API: 'OWNER', API_PRODUCT: 'OWNER', APPLICATION: 'USER' } });
 
             typeSearch('an');
             fireEvent.click(screen.getByText('Anna Schmidt'));
-            fireEvent.click(screen.getByRole('button', { name: 'Add member' }));
+            await submitMembers();
 
             expect(onSubmit).toHaveBeenCalledWith([
                 {
@@ -204,12 +223,12 @@ describe('GroupAddMembersSheet', () => {
             ]);
         });
 
-        it('falls back to USER for any scope missing from the group’s configured roles', () => {
+        it('falls back to USER for any scope missing from the group’s configured roles', async () => {
             const { onSubmit } = renderSheet({ groupRoles: { API: 'OWNER' } });
 
             typeSearch('an');
             fireEvent.click(screen.getByText('Anna Schmidt'));
-            fireEvent.click(screen.getByRole('button', { name: 'Add member' }));
+            await submitMembers();
 
             expect(onSubmit).toHaveBeenCalledWith([
                 {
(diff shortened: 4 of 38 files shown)


aqibmohammadkhan (commented):
### Issues

#### Important

**1. HTTP 200 for an existing unique user adds a member, then the UI jumps to Invitations**

Backend `InvitationServiceImpl.create`: 0 matches → persist invitation + email; 1 match → `addMember(...)` and `return null`; >1 → 202 list.

Gamma always does this on non-ambiguous success:

- toast “Member invited successfully”
- `setMemberTab('invitations')`

If the email already belongs to one org user, they land on Invitations (often empty) while the new person is on Members. Classic stays put and refreshes both lists.

Fix: branch on the POST body:

| Body | Meaning | UI |
| --- | --- | --- |
| array | 202 / ambiguous | too-many-users dialog |
| object with `id` | pending invitation | Invitations tab + invite toast |
| `null` / empty | member added | stay on Members + add-member toast |

Add a service test for a `null`/empty 200 body.

**2. Page tests miss the invite gating matrix that classic covers**

Implementation looks right, but `GroupDetailPage.spec.tsx` never asserts:

- `email_invitation` alone still shows Add members
- User search disabled when `system_invitation` is false
- Email invitation disabled when `email_invitation` is false

Classic has those harness tests. Easy to regress the dropdown.

#### Minor

**3. Duplicate email validator** — `shared/utils/email.ts` copies `features/users/utils/userDisplay.ts`. The PR says users already use the shared helper; they don’t. Re-export from one place.

**4. Success copy** — “Member invited successfully” is wrong for a pending invite and for an immediate add. Classic: “Successfully invited user to the group.”

**5. Truncated PR body** — first line is leftover: “hem instead of duplicating the applications-feature copies.”

**6. Aikido comments — dismiss both**

| Aikido | Reality |
| --- | --- |
| Add-members form no longer resets on open | `GroupAddMembersSheet` remounts with `key={open-…-initialSearch}` |
| Invitations query runs on every detail load | `useGroupInvitations(memberTab === 'invitations' ? groupId : undefined)` so `enabled` is false until the tab is selected |

An `enabled` argument on `useGroupInvitations` would be clearer, but this is not a bug.

---

### Scope vs Jira

FOUND-109 AC also asks for invitation status, resend, history, and audit. Classic group invitations have none of that (pending rows + delete only). Treat those as follow-ups unless the ticket is explicitly reduced to Console parity.

ankita-gupta21:
**1. Invitation response wasn't branched (invite vs. direct add vs. ambiguous)**
`inviteGroupMember` now returns a discriminated `InviteGroupMemberResult` instead of a boolean, mirroring what the backend actually sends: an array means several users match the email, an empty body means an existing user was added straight to the group, and an object with an `id` means a pending invitation was persisted. An unrecognised shape now fails loudly instead of being silently treated as an invitation.
- https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.ts#L109-L128
- https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/types/group.ts#L138

**2. Wrong tab / wrong success message after inviting**
The caller now switches to Members with "Member added successfully" for a direct add, to Invitations with "Successfully invited user to the group." for a pending invitation, and opens the too-many-users dialog for an ambiguous email — matching the classic Console.
- https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMemberActions.ts#L72-L85

**3. Over-broad cache invalidation**
Invalidation is now derived from the outcome: members only, invitations only, or nothing at all when the response was ambiguous and no data changed.
- https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMutations.ts#L78-L92
- Invitations are also only fetched while that tab is active: https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMemberActions.ts#L41

**4. Duplicated email validation**
The local copy is gone; `userDisplay` re-exports the shared validator so there is a single regex in the module.
- https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/users/utils/userDisplay.ts#L17

**5. Duplicated table search input**
The search field was promoted to a shared component and both the users and invitations tables consume it, so the label, icon, and clear behaviour are identical.
- https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/components/ClientSideTableSearchField.tsx
- https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupInvitationsTable.tsx#L180-L195

**6. Missing coverage for the gating and the response branches**
New tests cover each invitation-settings combination that enables or disables the menu items, plus one test per response branch end to end.
- Gating matrix: https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.spec.tsx#L555-L581
- Tab and message per outcome: https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.spec.tsx#L625-L672
- Service response parsing: https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.spec.ts#L40-L66
- Invalidation per outcome: https://github.com/gravitee-io/gravitee-api-management/blob/e65937532752b181299bd2351e14f60b1592815c/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMutations.spec.tsx#L71-L132
---
Added in PR description about what AC and how is it not aligned with classic console

aqibmohammadkhan (commented):
1. Sheet content remounts on close, not just open (GroupAddMembersSheet.tsx:61, same pattern in GroupInviteMemberSheet.tsx:46) — the sheet's inner content is keyed on open ? 'open' : 'closed', so it unmounts/remounts on both transitions. This diverges from the codebase's own convention (GroupSheet.tsx resets via a useEffect keyed only on opening) and risks a visible flash of reset form state during the close animation, plus needless remount work.
2. Fail-closed security comment dropped in a code move (memberRoles.ts:42) — isPrimaryOwnerUnavailable's "Fail closed..." comment, explaining a non-obvious auth-default invariant, wasn't carried over when the function moved out of useGroupAddMembersForm.ts into the now-shared utility.
3. Unrelated why-comments deleted outside the feature's scope (groups.ts:51, types/group.ts:39 and :17) — explanatory comments on listGroupMembers, EnvironmentPrimaryOwnerSettings, and the primary_owner field's semantics were stripped even though those areas aren't part of this feature.
4. setState-in-render to clamp pagination (GroupInvitationsTable.tsx:127) — works correctly but forces an extra render; clamping the value used for rendering (Math.min(page, totalPages)) would be simpler.

ankita-gupta21:
Addressed!!

**1. Sheet content remounted on close, not just open**
Extracted `useOpenRemountKey`, which bumps the key only on the `false → true` transition (and when identity/prefill changes). Closing keeps the same key, so the form no longer remounts mid close-animation. Both sheets use it, so the two copies of the keying logic are gone.
- https://github.com/gravitee-io/gravitee-api-management/blob/882bd73cad/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useOpenRemountKey.ts#L23-L33
- https://github.com/gravitee-io/gravitee-api-management/blob/882bd73cad/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupInviteMemberSheet.tsx#L46-L49
- https://github.com/gravitee-io/gravitee-api-management/blob/882bd73cad/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupAddMembersSheet.tsx#L59-L62

**2. Fail-closed comment dropped in the code move**
Restored verbatim on the shared helper, so the deny-until-known default stays documented where the function now lives.
- https://github.com/gravitee-io/gravitee-api-management/blob/882bd73cad/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/memberRoles.ts#L42-L45

**3. Unrelated why-comments deleted outside this feature**
Restored all three from `master` with no behaviour change: the unpaged-members rationale, the `/portal` vs `/settings` note, and the `primary_owner` / `GroupEventName` semantics.
- https://github.com/gravitee-io/gravitee-api-management/blob/882bd73cad/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.ts#L51-L53
- https://github.com/gravitee-io/gravitee-api-management/blob/882bd73cad/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.ts#L96-L97
- https://github.com/gravitee-io/gravitee-api-management/blob/882bd73cad/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/types/group.ts#L40-L47
- https://github.com/gravitee-io/gravitee-api-management/blob/882bd73cad/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/types/group.ts#L17

**4. setState-in-render to clamp pagination**
Gone. The rendered page is now `Math.min(page, totalPages)`, used for both slicing and the pagination control, so an out-of-range page paints once instead of render → setState → render.
- https://github.com/gravitee-io/gravitee-api-management/blob/882bd73cad/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupInvitationsTable.tsx#L122-L125
- https://github.com/gravitee-io/gravitee-api-management/blob/882bd73cad/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupInvitationsTable.tsx#L161

aqibmohammadkhan (approved):

PR #19262, opened 2026-08-19
refactor(ci): derive the analysed projects from one shared list

phiz71:
## Issue

https://gravitee.atlassian.net/browse/BX-300

## Description

The nine suite-to-project pairs were written twice: once in the pull-request groups, once in the nightly. They now live in `groups/analysed-projects.ts`, consumed by the pull-request groups with their changed-files predicate and by the nightly without it.

## Additional context

Pulling the analyses out of the frontend blocks — where lint-and-test, build, docker image and analysis share one `if` — moves them earlier in the emitted list, hence the ten regenerated fixtures. Every job keeps the same set of dependencies (`pull-requests-run-e2e.yml` reorders one `requires` array, which CircleCI reads as a set), and the nightly's fixture is untouched.


.circleci/ci/src/workflows/groups/frontend-jobs.ts
```diff
@@ -259,5 +224,16 @@ export function frontendJobs(
     }
   }
 
+  const sonarAnalysisJob = SonarCloudAnalysisJob.create(dynamicConfig, environment);
+  dynamicConfig.addJob(sonarAnalysisJob);
```
ytvnr: **Worth fixing** - Pipelines that analyse nothing now emit an orphan `job-sonarcloud-analysis` definition.

On `master` the analysis job was instantiated inside each guarded block, so a pipeline with no analysed module never defined it. Here (and at `backend-jobs.ts:66`) it is created and added before the predicate filter on line 231, so it is always emitted.

Comparing the fixtures on both sides confirms it: `pull-requests-custom-branch-backend-distribution-only.yml`, `...-backend-integration-tests-only.yml`, `...-gamma-console-only.yml` and `...-helm-only.yml` all gain a `job-sonarcloud-analysis` with zero `Sonar - *` workflow entry pointing at it. `helm-only.yml` additionally gains `cmd-restore-maven-job-cache` and `cmd-save-maven-job-cache`, which `SonarCloudAnalysisJob.create` registers as a side effect even though the sonar job uses neither.

Harmless at runtime, but it is the one place where the diff is not the pure reordering the description claims. Creating the job lazily, on the first project that survives the predicate, would keep those four pipelines byte-identical.

.circleci/ci/src/workflows/groups/analysed-projects.ts
```diff
@@ -0,0 +1,151 @@
...
+      resource_class: 'large',
+    },
+  },
+];
+
+/** The analysis job that reads the report a suite left behind. Its only dependency is that suite. */
+export function analysisJobFor(
+  project: AnalysedProject,
+  sonarAnalysisJob: Job,
+  cacheType: 'backend' | 'frontend',
```
ytvnr: `cache_type` stays outside the shared list, so the single source of truth is incomplete.

`cacheType` comes from the caller rather than from the `AnalysedProject`, so `backend-jobs.ts:77`, `frontend-jobs.ts:234` and the two nightly call sites each restate `'backend'` / `'frontend'`, even though the value is a fixed property of the list a project belongs to. A project moved between the two arrays, or a fifth consumer added later, can drift on the cache key exactly the way the suite/project pairs used to.

Same idea for `suiteParameters?: Record<string, string>` on line 56: a misspelt parameter name still compiles.

.circleci/ci/src/workflows/groups/analysed-projects.ts
```diff
@@ -0,0 +1,151 @@
...
+): { job: Job; parameters: Record<string, string | string[]> } {
+  return {
+    job: sonarAnalysisJob,
```
ytvnr: `analysisJobFor` echoes its own argument back.

`job` is the `sonarAnalysisJob` the caller just handed in, so every consumer writes `const analysis = analysisJobFor(project, sonarAnalysisJob, cache)` then `new workflow.WorkflowJob(analysis.job, analysis.parameters)`, unwrapping a pair whose first half it already owns. Only `analysis.parameters` carries information. Returning just the parameter record, or building the `WorkflowJob` here, would drop the unwrap at all three call sites.

.circleci/ci/src/pipelines/tests/pipeline-nightly.spec.ts
```diff
@@ -36,9 +36,9 @@ describe('Nightly', () => {
     expect(result.stringify()).toStrictEqual(expected);
   });
 
-  // The nightly spells out its own analyses, so nothing stops the two lists from drifting: a
-  // seventh module added to the pull-request groups would never be analysed on the reference
-  // branch, and the snapshot above would accept it.
+  // Both sides now derive from `analysed-projects.ts`, so this guards the derivation rather
+  // than the lists: the nightly takes them unfiltered, and a filter creeping in — or a project
+  // added to one consumer only — shows up here rather than in a missing baseline weeks later.
```
ytvnr: This comment describes a drift the shared list makes unconstructible.

"a project added to one consumer only" was a real risk when the nightly and the pull-request groups held two independently written lists. Both now read the same arrays, so that state cannot exist and `expect(sonarJobs(nightly)).toEqual(sonarJobs(pullRequest))` on line 65 is tautological.

The assertion that still earns its keep is `toHaveLength(9)` on line 66: it is what catches a project silently dropped from the shared list, which is the failure mode this whole epic exists to prevent. Worth pointing the comment at that instead.

ytvnr (approved):
Nice cleanup: the nine suite-to-project pairs now live in one place, and both consumers read the same list instead of two hand-kept copies. I checked the regenerated fixtures structurally (job definitions, commands, workflow entries) and ran the CI test suite in a worktree: 154/154 pass, so the fixtures are genuinely regenerated, all nine pairs keep their Sonar project keys and working directories, each predicate matches the block condition it replaces one for one, and the nightly still takes the list unfiltered.

One point worth a look before merge: the description says the only change is ordering, and that holds for ten of the fourteen fixtures, but not for four of them. Moving `SonarCloudAnalysisJob.create` out of the guarded blocks makes it run unconditionally, so pipelines that analyse nothing now emit an analysis job definition no workflow references (details inline).

Minor note on the description itself: `pull-requests-run-e2e.yml` does move `Integration tests` inside the `Validate workflow status` requires array. Same set, no behavioural effect since CircleCI treats `requires` as a set, but "No `requires` changes" is slightly stronger than what the diff shows.

.circleci/ci/src/workflows/groups/frontend-jobs.ts
phiz71: Right, and that was the one place the diff was not what the description claimed. Fixed in the commit above: the analysis job is now created on the first project that survives the predicate, in both groups.

The four fixtures you named — `backend-distribution-only`, `backend-integration-tests-only`, `gamma-console-only` and `helm-only` — are byte-identical to `master` again, `helm-only` included, so the two maven cache commands the factory registers as a side effect are gone with it. Ten fixtures still differ, and all ten are now pure displacement: same line count added and removed, checked file by file.

.circleci/ci/src/workflows/groups/analysed-projects.ts
phiz71: Moved: `cacheType` is now a property of `AnalysedProject`, set on the nine entries, and the four call sites no longer restate it. Moving a project between the two arrays carries its cache key with it.

On `suiteParameters?: Record<string, string>`, I left it as is. Typing it against the job parameters would mean naming them per suite factory, which buys back one class of typo at the cost of a shape that has to be kept in step with the jobs — the coupling this PR is trying to remove. Worth revisiting if a third frontend project shows up.

.circleci/ci/src/workflows/groups/analysed-projects.ts
phiz71: Fair — it returned the caller its own argument back. `analysisJobFor` now builds and returns the `workflow.WorkflowJob` directly, so the three call sites push it without unwrapping anything.

.circleci/ci/src/pipelines/tests/pipeline-nightly.spec.ts
phiz71: Correct, and the comment was describing the old design. Rewritten to point at the count, which is the assertion that still catches something:

> Both consumers derive from `analysed-projects.ts`, so the equality below cannot drift on its own. What earns its keep is the count: a project dropped from the shared list leaves the nightly and every pull request at once, which is the failure this epic exists to prevent.

PR #19278, opened 2026-08-20
perf(ci): cut the console lint & test job wall clock

phiz71:
## Issue

https://gravitee.atlassian.net/browse/BX-294

## Description

`Lint & test APIM Console` is the single most expensive job of the `pull_requests`
workflow: **100 747 credits over the last 30 days, 9.4 % of the whole workflow**, nearly
twice the runner-up. It runs on `xlarge` (40 credits/min) for a median of 11:54, and 71 %
of that wall clock is the Jest suite.

Two changes, both aimed at work that is done twice:

**1. The console suite runs transpile-only.** `ts-jest` was type checking every test file
as it transformed it — 521 suites, each paying the full program check. Types are still
checked, once, by a dedicated `console:type-check` step (`tsc --noEmit -p tsconfig.spec.json`,
**10 s**, 3 525 files) that runs right before the suite, so a typing error still fails the
job — and now fails it in 10 s instead of 8 minutes.

**2. The nx lint & test jobs no longer run prettier.** `Check prettier formatting for nx
projects` already checks every nx project and runs in parallel of them, so console,
portal-next and gamma-console were paying for the same check twice, on the critical path
of their own job.

## Additional context

**Measured on this branch's own CI runs.** The CircleCI fleet turned out to be noisy
enough to matter, so the numbers below come from four runs, two of them on the very same
SHA — which is what sets the error bar:

| run | commits | Jest | ESLint | wall clock |
|---|---|---|---|---|
| master (2008884 / 2008925) | — | — | 110 / 118 s | **10:06 / 11:00** (11:54 median) |
| 2009563 | 2 | 321.7 s | 68.0 s | 7:58 |
| 2010238 | 7 | 392.6 s | 86.6 s | 9:36 |
| 2010601 | 7 — *same SHA as above* | 349.7 s | 70.8 s | 8:33 |

The last two runs are the same code: **±12 % on Jest, ±22 % on ESLint** from the machine
alone. So the honest figure is **11:54 → around 8:30–9:00, roughly −25 %**, not the −33 %
the first run suggested. The type check itself is stable at ~20 s, visible as the gap
between the test step and Jest's own reported time.

At 40 credits/min and ~215 runs a month, that is roughly **33 000 credits a month**, ~3 %
of the whole `pull_requests` workflow.

**And on a console-only PR, this job _is_ the critical path.** The filtered workflow then
holds five jobs — Danger JS (0.7 min), prettier format check (3.0), Build APIM Console
(6.5), this one, and Sonar console (4.3), the only one that `requires` it. Everything else
runs in its shadow, so the workflow lasts `Lint & test` + `Sonar` = **16.2 min today, of
which this job is 73 %** — and **around 13 min with this PR, roughly −20 % end to end**. That is the
workflow a frontend dev waits on at every iteration.

**Measured locally**, full suite, `maxWorkers=7`, coverage on, cold Jest cache (which is
what CI always has — the cache is never persisted between runs):

| | idle machine | back-to-back A/B, loaded machine |
|---|---|---|
| master | 342 s | 547 s |
| this branch | 215 s | 322 s |
| | **−37 %** | **−41 %** |

`4792 passed, 0 failed` on this branch, same as master (master actually has one flaky
failure in both runs). Prettier: I checked that `nx format:check --libs-and-apps` does
catch a badly formatted console file, and that the format-check job is scheduled whenever
any of these jobs is.

**Coverage moves by −0.7 pt** (75.14 → 74.48 % statements). Transpile-only emits slightly
different code, so istanbul counts slightly differently. Sonar will see it once, and there
is no coverage threshold for it to trip.

**Following review** (thanks @ytvnr — all four points held up, and the second one was worth
more than it looked):

- `isolatedModules` was set on the transform but not on the compiler, so the type check was
  not enforcing what the runner assumes. Declaring it surfaced **78 errors** — 77 TS1272
  raised only because `emitDecoratorMetadata` is on. Nothing here reads decorator metadata
  (no `reflect-metadata`, no `Reflect.getMetadata`, no consuming dependency, and Ivy does
  not need it), so the spec scope now turns it off — which also silences the ts-jest warning
  each worker was printing. The two remaining errors were real and are fixed.
- The type check runs on its own `tsconfig.typecheck.json`, keeping the app's
  `moduleResolution: bundler`. Against the `node10` first pushed here, `tsc --listFiles`
  differs by **408 lines** — `node10` takes CJS builds and ignores `exports` maps — so the
  check no longer answers a different question from the build.
- `yarn console:test` type checks too: the nx `test:jest` target now depends on a
  `type-check` target.
- The nightly workflow schedules the prettier check, which it never did — it had three
  `job-nx-webui-lint-test` and zero `job-nx-format-check`, so dropping prettier from the
  lint jobs would have left nightly with no format check at all.

**Second review** — two of the four fixes above did not go far enough:

- The nightly format check was a **no-op**: `yarn prettier` is affected based, and on a
  nightly build of master the affected set is empty. Measured with a planted badly
  formatted file: reported against `origin/master`, silent with `base = head`. The nightly
  now runs `prettier:all`, built on the projects' own `lint-prettier` target —
  `format:check --all` is not an option, it reaches 594 files outside the nx projects that
  were never formatted.
- Hanging the type check on `test:jest` moved a whole-app check onto the **developer inner
  loop** (17 s, re-paid on every edit since the target has no `inputs`). It moves to the
  `console:test` script instead: the full local command still matches CI, a single spec run
  costs what it did before.
- `Lint & test APIM Libs` was still on the aggregate `lint` target, so the duplicate this PR
  removes survived there. It now uses `lint-eslint` + `lint-license` like the others.
- The base-branch fallback moves to a `diffRef` helper, covered by the suite — previously
  only the chunking half of that fix had a test.


.circleci/ci/src/jobs/frontend/job-webui-lint-test.ts
```diff
@@ -112,8 +112,10 @@ export class WebuiLintTestJob {
         command: 'yarn nx run << parameters.nx-project >>:lint-license',
       }),
       new commands.Run({
-        name: 'Run Prettier and ESLint',
-        command: 'yarn nx lint << parameters.nx-project >>',
+        // Prettier is deliberately left out: the 'Check prettier formatting for nx projects' job
+        // already checks every nx project, and runs in parallel of this one.
```
ytvnr, line 116: **Worth fixing** - The comment holds for pull requests but not for nightly, where the format check job is never scheduled, so nightly now runs no prettier check at all on the nx projects.

`workflow-nightly.ts:76-82` builds `WebuiLintTestJob.createNx` for console and portal-next, but its import list (lines 17-29) has no `NxFormatCheckJob` and the workflow never adds one. On the regenerated `nightly.yml`, `grep -c job-nx-format-check` returns 0 while `job-nx-webui-lint-test` appears three times, and all six pull-request snapshots return 2. Before this change `nx lint` was the only prettier check nightly had.

Either add the format check job to the nightly workflow group, or make the comment say explicitly that nightly relies on the PR gate on purpose.

gravitee-apim-console-webui/jest.config.js
```diff
@@ -4,6 +4,18 @@ module.exports = {
...
   preset: 'jest-preset-angular',
+  // Transpile-only: skipping the type check makes the suite ~35% faster on a cold cache.
+  // Types are still checked, by `yarn console:type-check` right before this suite runs.
+  transform: {
+    '^.+\\.(ts|js|mjs|html|svg)$': [
+      'jest-preset-angular',
+      {
+        tsconfig: '<rootDir>/tsconfig.spec.json',
+        stringifyContentPathRegex: '\\.(html|svg)$',
+        isolatedModules: true,
```
ytvnr, line 15: **Worth fixing** - `isolatedModules` is set on the ts-jest transform only, not in `tsconfig.spec.json`, so the compensating `tsc --noEmit` cannot report the errors that transpile-only introduces.

The constructs single-file transpilation breaks (TS1205, a type re-exported without `export type`, cross-file or ambient `const enum`) are exactly the ones the type-check step is there to catch, and it currently runs without the flag. Today the exposure is empty, `grep -rn "const enum" gravitee-apim-console-webui/src` returns nothing, which is why the suite is green. The gap is for later: a PR introducing one of those passes `console:type-check` and fails only inside whichever test happens to import the file.

Mirroring the flag in `tsconfig.spec.json` would make the compiler enforce the constraint the runner now assumes.

gravitee-apim-console-webui/tsconfig.spec.json
```diff
@@ -3,6 +3,8 @@
   "compilerOptions": {
     "outDir": "./out-tsc/spec",
     "module": "CommonJs",
+    // The root config sets "bundler", which tsc rejects together with CommonJs.
+    "moduleResolution": "node10",
```
ytvnr: **Worth fixing** - Forcing `node10` makes the type check resolve dependencies differently from both the ts-jest check it replaces and the app build, and `noImplicitAny: false` turns a resolution miss into a silent `any` rather than an error.

`tsconfig.json:28` sets `"moduleResolution": "bundler"`, which is what ts-jest used before this PR and what `tsconfig.app.json` still inherits. `node10` ignores package `exports` maps, so a dependency whose types it cannot locate is typed `any`, and since `tsconfig.json:22` sets `"noImplicitAny": false` that produces no diagnostic. The replacement check can therefore be quietly narrower than the one removed without the job ever going red.

Keeping `bundler` and picking a `module` value tsc accepts alongside it would avoid the question, otherwise a diff of the error count between the two configurations would settle it.

package.json
```diff
@@ -244,7 +244,8 @@
         "console:build": "NODE_OPTIONS=--max_old_space_size=7168 nx build console",
         "console:build:prod": "NODE_OPTIONS=--max_old_space_size=7168 nx build console --configuration=production",
         "console:test": "nx test console",
-        "console:test:ci": "node --max-old-space-size=10240 --expose-gc ./node_modules/.bin/jest --config gravitee-apim-console-webui/jest.config.js --ci --maxWorkers=7 --collectCoverage=true --logHeapUsage && node scripts/rewrite-lcov-for-sonar.js gravitee-apim-console-webui",
+        "console:type-check": "tsc --noEmit --incremental false -p gravitee-apim-console-webui/tsconfig.spec.json",
+        "console:test:ci": "yarn console:type-check && node --max-old-space-size=10240 --expose-gc ./node_modules/.bin/jest --config gravitee-apim-console-webui/jest.config.js --ci --maxWorkers=7 --collectCoverage=true --logHeapUsage && node scripts/rewrite-lcov-for-sonar.js gravitee-apim-console-webui",
```
ytvnr: **Worth fixing** - The type check was chained into `console:test:ci` only, so `yarn console:test`, the command developers run locally, now transpiles with no type checking at all.

Line 246 is unchanged and goes through the same `jest.config.js` with `isolatedModules: true`, so a developer iterating locally gets a green suite on code that fails `tsc` and only finds out on CI. That is a signal the local command used to give.

Putting the type check on the local entry point too, or on the nx `test` target's `dependsOn`, would keep the two commands answering the same question.

ytvnr (approved):
Nice work, the measurements make the case on their own and the two changes are cleanly separated into their own commits.

I checked the part that carries the risk, whether anything stops being checked, and it holds for pull requests: `job-nx-format-check` is present in all six PR snapshots, both as a job definition and as a workflow entry. `tsconfig.spec.json` declares no `include`/`files`, so `tsc -p` covers the whole console directory rather than a narrower set, the type check runs before the suite and aborts it, there is no `const enum` or `Reflect.getMetadata` in the console sources, and no coverage threshold anywhere that the -0.6 pt could trip.

One point I would look at before merging: the nightly workflow schedules the nx lint & test job but never schedules the format check, so nightly ends up with no prettier check at all on the nx projects. The other three are smaller, mostly about the new type-check step being a slightly different check from the one it replaces.

Approving since none of this blocks the perf win.

.circleci/ci/src/jobs/frontend/job-webui-lint-test.ts
phiz71: Confirmed and fixed. `grep -c job-nx-format-check` on the regenerated `nightly.yml` returned 0 against 3 `job-nx-webui-lint-test`, exactly as you measured. The nightly workflow now builds and schedules `NxFormatCheckJob` like the PR pipelines do — 18 lines added to the fixture, nothing else moved, 154 config tests green.

gravitee-apim-console-webui/jest.config.js
phiz71: You were right, and the exposure was not empty — it was just invisible. Declaring `isolatedModules` in `tsconfig.spec.json` surfaced **78 errors**.

77 of them are TS1272, raised only because `emitDecoratorMetadata` is on. Nothing here reads decorator metadata: no `reflect-metadata` polyfill, no `Reflect.getMetadata` anywhere in the console or lib sources, no dependency that consumes it (class-transformer, tsyringe, inversify…), and Ivy does not need it — which is exactly what the ts-jest warning says. So the spec scope now sets `emitDecoratorMetadata: false`, and 77 errors go away with it, along with the warning that every worker was printing.

What remains is two real defects the flag caught: a TS1205 in `src/util/index.ts` (`export { Json }` where `Json` is a type) and one in `gravitee-dashboard/src/public-api.ts`, which the console pulls in through its path mapping. Both fixed. Full suite: 4792 passed, 0 failed.

gravitee-apim-console-webui/tsconfig.spec.json
phiz71: Settled by the diff you suggested. `tsc --listFiles` differs by **408 lines** between the two: `node10` resolves `@sinclair/typebox/build/cjs/**` and `@lit/reactive-element/*.d.ts`, `bundler` follows the `exports` maps to `development/**`. Both report 0 errors today, so nothing was actually broken — but the check was answering a different question from the build, which is the part worth avoiding.

The type check now runs on its own `tsconfig.typecheck.json`, extending `tsconfig.spec.json` and overriding only `module` to `es2020` so `moduleResolution: bundler` is kept. `tsconfig.spec.json` goes back to no explicit resolution, so ts-jest sees exactly what it saw before this PR.

package.json
phiz71: Fixed. The console's nx `test:jest` target now has `dependsOn: ["type-check"]`, and `type-check` is a target running the same `tsc --noEmit`, so `yarn console:test` and `console:test:ci` answer the same question. The command definition lives in one place — `package.json`'s `console:type-check` now just calls `nx run console:type-check`.

.circleci/ci/src/workflows/workflow-nightly.ts
```diff
@@ -112,6 +116,12 @@ export class NightlyWorkflow {
         }),
       ]),
 
+      // The nx lint & test jobs below do not check formatting — this one does, for every nx project.
+      new workflow.WorkflowJob(formatCheckJob, {
+        name: 'Check prettier formatting for nx projects',
+        context: config.jobContext,
+      }),
```
ytvnr, line 85: **Worth fixing** - `NxFormatCheckJob` is a no-op in the nightly workflow, so dropping prettier from the nx lint jobs removes nightly formatting coverage rather than relocating it.

The job runs `yarn prettier`, which is `nx format:check --libs-and-apps && nx format:check --files=nx.json,package.json,jest.config.js` (`package.json:232`). In nx 21.6.10 `format:check` is affected based: `format.js` calls `splitArgsIntoNxArgsAndOverrides(args, 'affected')`, and `--libs-and-apps` routes to `getPatternsFromApps` then `filterAffected(...)` with `base = affected.defaultBase = "master"` and `head = HEAD`. On a nightly build of `master` that diff is empty, so no project is affected, `patterns` is `[]`, `check([])` returns `[]` and the job exits 0. Only the three listed root files are really checked.

Pull requests are fine, an affected project contributes its whole root as a pattern, so the gap is nightly specific. But that is exactly the coverage `yarn nx lint <project>` gave unconditionally there before this PR. Having the nightly job check all files, rather than reusing the PR oriented `yarn prettier` script, would close it.

gravitee-apim-console-webui/project.json
```diff
@@ -184,6 +184,14 @@
...
       "dependsOn": ["test:jest"]
     },
+    "type-check": {
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "yarn console:type-check",
+        "cwd": "{workspaceRoot}"
+      },
+      "cache": true
+    },
```
ytvnr, line 194: **Worth fixing** - wiring `type-check` as a `dependsOn` of `test:jest` moves a whole application type check onto the developer inner loop, where it was previously scoped to the files a run touched.

The chain is `test` then `test:jest` then `type-check` then `tsc --noEmit -p tsconfig.typecheck.json`. None of `tsconfig.typecheck.json`, `tsconfig.spec.json` or `tsconfig.json` declares `include` or `files`, so the program covers every `.ts` under the console plus the libs reached through `paths`. Running one spec now type checks the whole app, where ts-jest previously checked only what that run imported. And `type-check` (line 209) sets `cache: true` with no `inputs`, so the project wide default applies and any source edit busts the cache, which means every loop iteration pays it.

The CI side of this is clearly worth it. The local side is the part I would measure before merging: a single spec run before and after. Narrowing the target's nx `inputs`, or keeping the dependency on the CI target only, would keep the CI gain without the local cost.

.circleci/ci/src/jobs/frontend/job-webui-lint-test.ts
ytvnr, line 116: The comment says the format check job covers every nx project, but `Lint & test APIM Libs` still runs prettier itself, so the duplicate this PR removes here survives there.

`createNxLibs` (line 164) was left on the aggregate `lint` target, `yarn nx affected -t lint --exclude=console,portal-next`, and `nx.json` declares `lint` as `dependsOn: [lint-eslint, lint-license, lint-prettier]`. `frontendJobs` schedules the format check job whenever the libs job runs, so that prettier run is redundant in exactly the same way the console one was.

Either move the libs job to the two granular targets like this one, or narrow the comment.

.circleci/ci/src/index.ts
```diff
@@ -41,7 +41,9 @@ if (isBlank(CIRCLE_SHA1)) {
  *     - if the branch is supported ( CIRCLE_BRANCH is master or a support branch )
  *     - if we are working on a branch with changes committed on the base branch
  */
-const changed = isSupportBranchOrMaster(CIRCLE_BRANCH) ? Promise.resolve([]) : changedFiles(GIT_COMMON_COMMIT_HASH ?? GIT_BASE_BRANCH);
+const changed = isSupportBranchOrMaster(CIRCLE_BRANCH)
+  ? Promise.resolve([])
+  : changedFiles(isBlank(GIT_COMMON_COMMIT_HASH) ? GIT_BASE_BRANCH : GIT_COMMON_COMMIT_HASH);
```
ytvnr: The base branch fallback half of `3543eff9a0` has no test, while the chunking half does.

`git.spec.ts` exercises `changedFiles` and `toChangedPaths` only. The fallback lives here, and its trigger is reachable: `.circleci/config.yml:117` sets `GIT_COMMON_COMMIT_HASH` from `jq -r '.base.sha // ""'`, so a GitHub API response without `base.sha` yields the empty string that `??` did not catch and `isBlank` now does. That path selects `GIT_BASE_BRANCH` and, combined with the new three dot diff, changes which jobs get scheduled, with nothing asserting that `''` is treated as absent.

Extracting the ref selection into a helper under `utils` would make it coverable by `npm run test`.

ytvnr (approved):
Second pass, focused on whether the four points from the first review actually hold in the code rather than in the replies. They do: the spec scope now declares `isolatedModules` and the two defects it surfaced are genuinely neutral (`Json` is a type alias, the three dashboard exports are interfaces), `emitDecoratorMetadata: false` is safe since no `reflect-metadata` polyfill exists anywhere in the repo so `design:paramtypes` was never readable at test runtime, `tsconfig.typecheck.json` covers a superset of what ts-jest checked, and there is no double type check on CI because `console:test:ci` bypasses nx. The fixture diffs are explainable line by line: every remaining `Run Prettier and ESLint` belongs to the legacy `job-webui-lint-test` or to `job-e2e-lint-build`, not to the nx job.

Two things worth a look before merge, neither of them blocking.

The nightly format check job that was added does not check anything on `master`. `yarn prettier` runs `nx format:check --libs-and-apps`, which is affected based, and on a nightly build of `master` the affected set is empty. The gap the fix commit targeted is still open, it just moved one level down.

The other is the local inner loop. `test:jest` now depends on `type-check`, and that target has no `include` and no nx `inputs`, so a single spec run type checks the whole application and any edit busts the cache. The CI win is real, part of it is now paid on the developer side. Per the epic's own rule, that is the kind of transfer worth measuring before it ships.

The last two comments are minor and can be left as they are.

.circleci/ci/src/workflows/workflow-nightly.ts
phiz71: Confirmed, and the fix commit did exactly what you say — it moved the gap down a level instead of closing it. Measured with a badly formatted console file: `nx format:check --libs-and-apps --base=origin/master` reports it, the same command with `--base=HEAD` (an empty affected set, which is the nightly's situation on master) reports nothing at all.

The nightly now runs a new `prettier:all` script. `format:check --all` was the first thing I tried and it is not usable here: it reaches **594 files** outside the nx projects — Java templates under `gravitee-apim-distribution`, gateway test resources — none of which were ever formatted. So `prettier:all` goes through the projects' own `lint-prettier` target instead (`nx run-many -t lint-prettier --all --exclude-task-dependencies`), which keeps the same file patterns as the PR path and picks up new projects on its own. Verified it fails on the same planted file.

gravitee-apim-console-webui/project.json
phiz71: Fair, and the transfer is real: the target has no `inputs`, so any edit busts the cache and the type check is paid again — 17 s cold, on every iteration. Rather than narrow the inputs (a whole-app type check is legitimately invalidated by any source edit, so the cache would rarely hit anyway), I moved it off `test:jest` and onto the `console:test` script, next to `console:test:ci`.

So `yarn console:test` still answers the same question as CI, and the inner loop — a single spec through nx or jest directly — costs exactly what it did before this PR. Your first review asked for the local command to type check; this keeps that without putting it on every iteration.

.circleci/ci/src/jobs/frontend/job-webui-lint-test.ts
phiz71: Right, and it is the same duplicate. `createNxLibs` now runs `nx affected -t lint-eslint lint-license --exclude=console,portal-next` instead of the aggregate `lint`, so the libs job matches the nx jobs above it and the comment holds for both. Checked the two targets run green on `markdown` and `dashboard`.

.circleci/ci/src/index.ts
phiz71: Agreed — the half that could silently change which jobs run was the half without a test. The selection now lives in a `diffRef(commonCommitHash, baseBranch)` helper in `utils/git.ts`, covered by two cases: the common commit wins when present, and an empty string falls back to the base branch, with the `jq -r '.base.sha // ""'` origin of that empty string spelled out in the test.

One note while writing it: `isBlank` is length-based, not trim-based, so only a genuinely empty string is treated as absent — which is what `jq` produces. I left `isBlank` alone rather than widen a helper used elsewhere.

package.json
phiz71: Correction to my reply above, since the second review changed the answer: the `dependsOn` I described is **gone**. You pointed out it put a whole-app type check on the inner loop, re-paid on every edit, so the type check now hangs off the `console:test` script instead — right next to `console:test:ci`.

Net effect for this thread is the same and is what you asked for: `yarn console:test` type checks, so the local command and CI answer the same question. What changed is that running a single spec through nx no longer pays for it.

PR #19283, opened 2026-08-20
feat(console): import portal page content from a file

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-91

## Description

Import portal page content from a local file, on both entry points: the "Import from file" option of the Add page dialog (until now a disabled placeholder) and a new Import file button on an existing page. Accepts `.md`, `.yaml`, `.yml`, `.json`.

The content type is derived from the document itself rather than its extension, so an AsyncAPI spec no longer lands as an OpenAPI page. On the backend, the page content update endpoint accepts an optional `type` to convert a page in place, and rejects content over 10 MB.

## Additional context

https://github.com/user-attachments/assets/05994638-a8c6-4ac2-9a97-d57b5dd4ae4f

### DIFF
--- a/gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
+++ b/gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
@@ -22,6 +22,8 @@ import {
   GioCardEmptyStateModule,
   GioConfirmAndValidateDialogComponent,
   GioConfirmAndValidateDialogData,
+  GioConfirmDialogComponent,
+  GioConfirmDialogData,
 } from '@gravitee/ui-particles-angular';
 import { Component, computed, DestroyRef, effect, HostListener, inject, NgZone, Signal, signal, viewChild } from '@angular/core';
 import { MatButtonModule } from '@angular/material/button';
@@ -63,6 +65,7 @@ import {
   PublishNavigationItemDialogData,
   PublishNavigationItemDialogResult,
 } from './publish-navigation-item-dialog/publish-navigation-item-dialog.component';
+import { ImportFileError, IMPORTABLE_FILE_EXTENSIONS, readImportedFile, validateImportFile } from './portal-page-content-import.util';
 
 import { PortalHeaderComponent } from '../components/header/portal-header.component';
 import { EmptyStateComponent } from '../../shared/components/empty-state/empty-state.component';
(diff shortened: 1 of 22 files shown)


gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
```diff
@@ -953,6 +978,74 @@ export class PortalNavigationItemsComponent implements HasUnsavedChanges {
...
+          if (error instanceof ImportFileError) {
+            this.snackBarService.error(error.message);
+          } else if (!(error instanceof HttpErrorResponse)) {
+            this.snackBarService.error(`Failed to import content from "${file.name}"`);
+          }
+          return EMPTY;
+        }),
+        takeUntilDestroyed(this.destroyRef),
+      )
+      .subscribe(updatedContent => {
```
jaroslawlakomy-grav, line 1039: Could we make the response update conditional on the imported page still being selected, or otherwise cancel or lock the operation when navigation changes? If the user selects another page while the PUT is pending, page A’s response can replace page B’s editor state and create a path to saving the wrong content into page B. A delayed-response navigation test would cover this race.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalPageContentValidatorService.java
```diff
@@ -16,21 +16,34 @@
...
 @RequiredArgsConstructor
 public class PortalPageContentValidatorService {
 
+    private static final int MAX_CONTENT_SIZE_IN_MEGA_BYTES = 10;
+    private static final int MAX_CONTENT_SIZE_IN_BYTES = MAX_CONTENT_SIZE_IN_MEGA_BYTES * 1024 * 1024;
+
     private final List<PortalPageContentValidator> validators;
 
     public void validateForUpdate(PortalPageContent<?> existingContent, UpdatePortalPageContent updateContent) {
+        validateSize(updateContent.getContent());
```
jaroslawlakomy-grav: Could we decide whether the 10 MB limit is a portal-content invariant or only a constraint of the Management API update endpoint, and enforce it consistently? The shared validator affects Automation API same-type updates, while its create and type-change branches bypass the limit.

jaroslawlakomy-grav (commented):
Please also check SonarCloud warnings

gravitee-apim-console-webui/src/portal/navigation-items/portal-navigation-items.component.ts
thomasballivet: Done, the response handler now bails out if the imported page is no longer selected (a snackbar still confirms the import succeeded). Covered by a delayed-response navigation test.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalPageContentValidatorService.java
thomasballivet: Done, scoped the cap to the Management API update endpoint only (UpdatePortalPageContentUseCase): validateForUpdate no longer enforces it, so Automation API create/update/type-change flows are untouched and can't regress. Covered by use-case tests on both sides.

jaroslawlakomy-grav (approved):

PR #19284, opened 2026-08-20
fix(console): edit portal source auto-fetch frequency with the cron builder

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-89

## Description

The source editor rendered the auto-fetch frequency as a raw cron text input, while fetcher plugin schemas declare `fetchCron` with format: "`gio-cron`" — losing the frequency builder UI users had in the v4 documentation screens.

The raw input is replaced with the `GioFormCron` widget (same as health-check). As a bonus, the widget validates the expression, so saving is now blocked on an invalid cron instead of sending any string to the backend.

## Additional context

Capture d’écran 2026-08-20 à 10 52 22

### DIFF
--- a/gravitee-apim-console-webui/src/portal/navigation-items/navigation-item-source-editor/navigation-item-source-editor.component.html
+++ b/gravitee-apim-console-webui/src/portal/navigation-items/navigation-item-source-editor/navigation-item-source-editor.component.html
@@ -51,11 +51,10 @@
     <mat-slide-toggle data-testid="auto-fetch-toggle" [formControl]="useAutoFetchControl">Auto-fetch</mat-slide-toggle>
 
     @if (useAutoFetch()) {
-      <mat-form-field appearance="outline" class="source-editor__auto-fetch__cron">
-        <mat-label>Update frequency (cron expression)</mat-label>
-        <input matInput data-testid="fetch-cron-input" [formControl]="fetchCronControl" placeholder="0 */10 * * * *" />
-        <mat-hint>Example: "0 */10 * * * *" fetches every 10 minutes</mat-hint>
-      </mat-form-field>
+      <gio-form-cron data-testid="fetch-cron-input" class="source-editor__auto-fetch__cron" [formControl]="fetchCronControl">
+        <gio-form-cron-label>Update frequency</gio-form-cron-label>
+        <gio-form-cron-hint>Required when auto-fetch is enabled</gio-form-cron-hint>
+      </gio-form-cron>
     }
   </div>
 
--- a/gravitee-apim-console-webui/src/portal/navigation-items/navigation-item-source-editor/navigation-item-source-editor.component.ts
+++ b/gravitee-apim-console-webui/src/portal/navigation-items/navigation-item-source-editor/navigation-item-source-editor.component.ts
@@ -21,7 +21,7 @@ import { MatFormFieldModule } from '@angular/material/form-field';
 import { MatInputModule } from '@angular/material/input';
 import { MatSelectModule } from '@angular/material/select';
 import { MatSlideToggleModule } from '@angular/material/slide-toggle';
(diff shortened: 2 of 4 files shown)


gravitee-apim-console-webui/src/portal/navigation-items/navigation-item-source-editor/navigation-item-source-editor.component.html
```diff
@@ -51,11 +51,10 @@
     <mat-slide-toggle data-testid="auto-fetch-toggle" [formControl]="useAutoFetchControl">Auto-fetch</mat-slide-toggle>
 
     @if (useAutoFetch()) {
-      <mat-form-field appearance="outline" class="source-editor__auto-fetch__cron">
-        <mat-label>Update frequency (cron expression)</mat-label>
-        <input matInput data-testid="fetch-cron-input" [formControl]="fetchCronControl" placeholder="0 */10 * * * *" />
-        <mat-hint>Example: "0 */10 * * * *" fetches every 10 minutes</mat-hint>
-      </mat-form-field>
+      <gio-form-cron data-testid="fetch-cron-input" class="source-editor__auto-fetch__cron" [formControl]="fetchCronControl">
```
jaroslawlakomy-grav, line 54: Could we align the cron builder’s generated values with the backend before using it here? In ui-particles 17.8.0 the weekly selector displays Monday through Sunday but writes values 0 through 6, while Spring interprets 0 as Sunday. Selecting Monday therefore schedules the fetch on Sunday. The same control also offers zero for interval fields and emits */0, which it considers valid but Spring rejects. Could we fix or upgrade the component and add representative builder-mode tests?

gravitee-apim-console-webui/src/portal/navigation-items/navigation-item-source-editor/navigation-item-source-editor.component.ts
```diff
@@ -101,6 +102,7 @@ export class NavigationItemSourceEditorComponent {
   readonly selectedType = toSignal(this.typeControl.valueChanges, { initialValue: this.typeControl.value });
   readonly useAutoFetch = toSignal(this.useAutoFetchControl.valueChanges, { initialValue: this.useAutoFetchControl.value });
   readonly fetchCron = toSignal(this.fetchCronControl.valueChanges, { initialValue: this.fetchCronControl.value });
+  private readonly fetchCronStatus = toSignal(this.fetchCronControl.statusChanges, { initialValue: this.fetchCronControl.status });
```
jaroslawlakomy-grav, line 105: Could we cover and correctly observe the initial validity of a persisted cron? fetchCronStatus is initialized before GioFormCron registers its validator, and Angular recalculates that validity with emitEvent: false, so the signal can remain VALID. A test initialized with useAutoFetch: true and a rejected cron would expose this case.

thomasballivet: Done, reproduced with a persisted invalid cron test, then fixed: validation re-runs once the cron field renders, so the initial status is observed.

gravitee-apim-console-webui/src/portal/navigation-items/navigation-item-source-editor/navigation-item-source-editor.component.html
thomasballivet: Confirmed in **ui-particles** 17.8.0, but pre-existing in every consumer: the doc screens render the same widget through the **gio-cron** schema format, and health-check uses it directly, this PR is at feature parity with them. Tracked in [PORTAL-133](https://gravitee.atlassian.net/browse/PORTAL-133) to fix it once in **ui-particles** and bump here.

jaroslawlakomy-grav (approved):

PR #19285, opened 2026-08-20
feat(automation): api-referenced links automation - openApi and resources

MateuszJozwiak-Incubly:
## Issue

### What this branch does

This branch exposes the API-referenced link feature (introduced in `part1`) through the **Automation REST API**, and completes a structural change part1 started: an API's navigation content — its folders, its documentation pages, and now its links — is materialized exactly once, owned by the API itself, instead of being duplicated per portal listing. Existing deployments are migrated to that model automatically on upgrade.

### REST layer & OpenAPI contract

New resources let automation config declare and read back an API's own link(s) through the Automation API:

- New endpoints for reading and managing a single API-referenced link, and listing an API's links
- The `open-api.yaml` contract extended with the corresponding schemas and operations
- A mapper translating between the REST/automation model and the domain model
- Resource wiring updated so the new endpoints are reachable from the existing API resource tree

### API-owned navigation: materialize once, not per listing

Previously, an API's folder subtree and documentation pages were each reconciled and materialized **per portal listing** — an API listed in two portals produced two independent copies of the same subtree, keyed to that listing rather than to the API. This branch removes that fan-out:

- The folder subtree reconciler now reconciles an API's `portalNavigation` **once per API**, regardless of how many portals list it, or whether any currently do
- Documentation pages drop the per-listing materialization loop entirely — one page per API and content, materialized with no listing and no portal required
- Ownership queries (used to protect automation-managed rows from an unrelated cascade delete) move onto the item's own API reference instead of enumerating by listing row
- Removing an API from a listing no longer deletes its automation-managed documentation, folders or links — only deleting the API itself does. Re-adding the listing simply re-renders what was never actually removed
- The walks that resolve an item's owning API (for documentation-template scoping) and that order or validate siblings at the root level now recognize an API-owned root directly, rather than assuming every top-level item belongs to a portal

### Fixing a root-level segment conflict

An API declaring an ordinary navigation path like `/guides` could never actually be reconciled: the check that rejects a segment already taken by a foreign item was scoped across the *entire* environment rather than by the item's own reference, so it collided with the portal's own seeded "Guides" folder — a false conflict on the most common input there is. Fixed by scoping that check the same way sibling ordering was already scoped: by reference, not environment-wide, so an API's own navigation and a portal's own navigation stop competing for the same segment names.

### Migrating existing navigation data

Rows materialized under the old, per-listing model don't disappear just because the code changed. A new boot-time upgrader re-keys every pre-existing folder, documentation page and link into the API-owned model in place: legacy folders and pages are identified by matching their stored id against the old id formula — the same check the runtime itself already uses to detect a foreign row — then migrated to their new identity, with any duplicate left behind by the old per-listing fan-out converging onto a single row. Links, whose identity never depended on the listing, are simply re-parented and re-tagged in place. The migration runs once, requires no downtime and no manual step, and replaces an earlier runtime-sweep approach that turned out not to work for folders at all.

### Outcome

With this branch, operators can declare an API's links via the Automation API/GitOps config exactly like they already do for documentation and listing, and have them reliably created, updated, retracted and exposed for reading. An API's navigation content is now owned by the API rather than by any one listing, existing deployments migrate to that model automatically on upgrade, and a class of root-level path collisions between an API's own declared navigation and a portal's seeded folders is closed.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/ApiLinkResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/ApiLinkResource.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.rest.api.automation.resource;
+
+import io.gravitee.apim.core.portal_page.exception.PortalLinkNotFoundException;
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
+import io.gravitee.apim.core.portal_page.use_case.DeleteApiLinkUseCase;
+import io.gravitee.apim.core.portal_page.use_case.GetApiLinkUseCase;
+import io.gravitee.apim.rest.api.automation.exception.HRIDNotFoundException;
+import io.gravitee.apim.rest.api.automation.mapper.PortalLinkMapper;
+import io.gravitee.common.http.MediaType;
+import io.gravitee.rest.api.model.permissions.RolePermission;
+import io.gravitee.rest.api.model.permissions.RolePermissionAction;
+import io.gravitee.rest.api.rest.annotation.Permission;
+import io.gravitee.rest.api.rest.annotation.Permissions;
+import io.gravitee.rest.api.service.common.HRIDToUUID;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.DELETE;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.Response;
+
+/**
+ * @author GraviteeSource Team
+ */
+public class ApiLinkResource extends AbstractResource {
+
+    @Inject
+    private GetApiLinkUseCase getApiLinkUseCase;
+
+    @Inject
+    private DeleteApiLinkUseCase deleteApiLinkUseCase;
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Permissions({ @Permission(value = RolePermission.API_DOCUMENTATION, acls = RolePermissionAction.READ) })
+    public Response getApiLinkByHRID(@PathParam("apiHrid") String apiHrid, @PathParam("linkHrid") String linkHrid) {
+        var auditInfo = getAuditInfo();
+        var apiId = HRIDToUUID.api().context(auditInfo).hrid(apiHrid).id();
+        var linkId = PortalNavigationItemId.forApiLink(auditInfo, apiId, linkHrid);
+        try {
+            var output = getApiLinkUseCase.execute(new GetApiLinkUseCase.Input(auditInfo, linkId));
+            return Response.ok(PortalLinkMapper.INSTANCE.toApiLinkState(output.link(), linkHrid, apiHrid)).build();
+        } catch (PortalLinkNotFoundException e) {
+            throw new HRIDNotFoundException(linkHrid);
+        }
+    }
+
+    @DELETE
+    @Permissions({ @Permission(value = RolePermission.API_DOCUMENTATION, acls = RolePermissionAction.DELETE) })
+    public Response deleteApiLinkByHrid(@PathParam("apiHrid") String apiHrid, @PathParam("linkHrid") String linkHrid) {
+        var auditInfo = getAuditInfo();
+        var apiId = HRIDToUUID.api().context(auditInfo).hrid(apiHrid).id();
+        var linkId = PortalNavigationItemId.forApiLink(auditInfo, apiId, linkHrid);
+        try {
+            deleteApiLinkUseCase.execute(new DeleteApiLinkUseCase.Input(auditInfo, linkId));
+        } catch (PortalLinkNotFoundException e) {
+            throw new HRIDNotFoundException(linkHrid);
+        }
+        return Response.noContent().build();
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/ApiLinksResource.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/ApiLinksResource.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.rest.api.automation.resource;
+
+import static io.gravitee.rest.api.model.permissions.RolePermissionAction.CREATE;
+import static io.gravitee.rest.api.model.permissions.RolePermissionAction.UPDATE;
+
+import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
+import io.gravitee.apim.core.portal_page.use_case.CreateOrUpdateApiLinkUseCase;
+import io.gravitee.apim.core.portal_page.use_case.ValidateApiLinkUseCase;
+import io.gravitee.apim.core.validation.Validator;
+import io.gravitee.apim.rest.api.automation.mapper.PortalLinkMapper;
+import io.gravitee.apim.rest.api.automation.model.PortalLinkSpec;
+import io.gravitee.common.http.MediaType;
+import io.gravitee.rest.api.model.permissions.RolePermission;
+import io.gravitee.rest.api.rest.annotation.Permission;
+import io.gravitee.rest.api.rest.annotation.Permissions;
+import io.gravitee.rest.api.service.common.HRIDToUUID;
+import jakarta.inject.Inject;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotNull;
+import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.PUT;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.QueryParam;
+import jakarta.ws.rs.container.ResourceContext;
+import jakarta.ws.rs.core.Context;
+import jakarta.ws.rs.core.Response;
+
+/**
+ * @author GraviteeSource Team
+ */
+public class ApiLinksResource extends AbstractResource {
+
+    @Context
+    private ResourceContext resourceContext;
+
+    @Inject
+    private CreateOrUpdateApiLinkUseCase createOrUpdateApiLinkUseCase;
+
+    @Inject
+    private ValidateApiLinkUseCase validateApiLinkUseCase;
+
+    @Path("/{linkHrid}")
+    public ApiLinkResource getApiLinkResource() {
+        return resourceContext.getResource(ApiLinkResource.class);
+    }
+
+    @PUT
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Permissions({ @Permission(value = RolePermission.API_DOCUMENTATION, acls = { CREATE, UPDATE }) })
+    public Response createOrUpdate(
+        @PathParam("apiHrid") String apiHrid,
+        @Valid @NotNull PortalLinkSpec spec,
+        @QueryParam("dryRun") boolean dryRun
+    ) {
+        var auditInfo = getAuditInfo();
+        var apiId = HRIDToUUID.api().context(auditInfo).hrid(apiHrid).id();
+        var linkId = PortalNavigationItemId.forApiLink(auditInfo, apiId, spec.getHrid());
+
+        var input = new CreateOrUpdateApiLinkUseCase.Input(
+            auditInfo,
+            apiId,
+            spec.getHrid(),
+            spec.getName(),
+            spec.getHref(),
+            spec.getLocation(),
+            spec.getOrder()
+        );
+        var output = dryRun ? validateApiLinkUseCase.execute(input) : createOrUpdateApiLinkUseCase.execute(input);
+
+        var state = PortalLinkMapper.INSTANCE.toApiLinkState(spec, linkId.toString(), output.errors(), auditInfo, apiHrid);
+
+        // A dry run is a preview: severe findings are its payload, so it always answers 200.
+        // A real apply that produced severe errors persisted nothing — it must not report success.
+        var applyFailed = !dryRun && output.errors().stream().anyMatch(Validator.Error::isSevere);
+
+        return Response.status(applyFailed ? Response.Status.BAD_REQUEST : Response.Status.OK).entity(state).build();
+    }
+}
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/query_service/AutomationManagedNavigationItemsQueryService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/query_service/AutomationManagedNavigationItemsQueryService.java
@@ -24,7 +24,6 @@ import io.gravitee.apim.core.portal_page.model.PortalNavigationItem;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemType;
 import io.gravitee.apim.core.portal_page.query_service.PortalNavigationItemsQueryService;
-import io.gravitee.apim.core.portal_page.query_service.PortalPageContentQueryService;
 import java.util.Set;
 import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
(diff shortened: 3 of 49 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/query_service/AutomationManagedNavigationItemsQueryService.java
```diff
@@ -76,4 +75,17 @@ public Set<PortalNavigationItemId> automationManagedPortalLinks(AuditInfo auditI
             .map(PortalNavigationItem::getId)
             .collect(Collectors.toSet());
     }
+
+    /**
+     * Unlike {@link #automationManagedApiDocPages}, this needs no {@code PortalNavigationApi} row: a
```
konrad-krol-incubly: Is this comment relevant? Both (links and doc) are not queryable through navigation items.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/NavigationItemEntryMaterializer.java
```diff
@@ -76,6 +79,37 @@ void dematerialize(AuditInfo auditInfo, PortalId portalId, String apiId) {
...
+        var environmentId = auditInfo.environmentId();
+        var navApiId = rowId(auditInfo, portalId, apiId);
+        var toDelete = new ArrayList<PortalNavigationItemId>();
+        var queue = new LinkedList<PortalNavigationItemId>();
+        queue.add(navApiId);
+        while (!queue.isEmpty()) {
+            var children = navigationItemsQueryService.findByParentIdAndEnvironmentId(environmentId, queue.poll());
+            for (var child : children) {
+                queue.add(child.getId());
+                if (isAutomationOwnedByApi(child)) {
```
konrad-krol-incubly: Method complexity is too high,nested loops and if statements.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/NavigationItemEntryMaterializer.java
```diff
@@ -76,6 +79,37 @@ void dematerialize(AuditInfo auditInfo, PortalId portalId, String apiId) {
         apiDocumentationSyncDomainService.cleanupNavApi(auditInfo, rowId(auditInfo, portalId, apiId));
     }
 
+    /**
+     * Rows created before the api-owned navigation re-key are still physically parented under the
```
konrad-krol-incubly: I am considering if maybe Upgrader isn't better choice in this case. If you introduce this code, it will stay forever here, just to address legacy case.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemApiOwnedRekeyUpgrader.java
```diff
@@ -0,0 +1,326 @@
...
+            .collect(Collectors.groupingBy(PortalNavigationItem::getParentId));
+
+        // Folder new-id -> new-rootId, accumulated across every nav-api row in this environment, so a
+        // doc page or link parented under a re-keyed folder inherits that folder's root regardless of
+        // which nav-api row originally owned the folder tree it came from.
+        Map<String, String> folderNewIdToRootId = new HashMap<>();
+
+        for (var apiRow : items) {
+            if (apiRow.getType() != PortalNavigationItem.Type.API) {
+                continue;
```
konrad-krol-incubly: instead of for loop with continue statement -> can we collect list of apiRows into variable (of list/set type) and then iterate over that?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemApiOwnedRekeyUpgrader.java
```diff
@@ -0,0 +1,326 @@
...
+                newParentId = folderId(organizationId, environmentId, apiId, parentPath);
+                newRootId = folderNewIdToRootId.getOrDefault(newParentId, newParentId);
+            }
+            folderNewIdToRootId.put(newId, newRootId);
+
+            rekeyItem(pathed.folder(), newId, newParentId, newRootId, PortalNavigationReferenceType.API, apiId);
+        }
+    }
+
+    private void rekeyDocPages(
```
konrad-krol-incubly, line 147: this method is unreadable, for loop with 3 if statements, each with "continue" jump.

please collect list of items using filtering and write it in local var, then iterate

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemApiOwnedRekeyUpgrader.java
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    private void fixLinks(
+        String organizationId,
+        String environmentId,
+        List<PortalNavigationItem> items,
+        Map<String, String> folderNewIdToRootId
+    ) throws TechnicalException {
+        for (var link : items) {
+            if (link.getType() != PortalNavigationItem.Type.LINK) {
```
konrad-krol-incubly: same as above - not readable, for loop with if statement, each with continue jumps

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemApiOwnedRekeyUpgrader.java
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    private void migrateEnvironment(String organizationId, String environmentId, List<PortalNavigationItem> items)
+        throws TechnicalException {
+        Map<String, List<PortalNavigationItem>> childrenByParentId = items
+            .stream()
+            .filter(item -> item.getParentId() != null)
+            .collect(Collectors.groupingBy(PortalNavigationItem::getParentId));
+
+        // Folder new-id -> new-rootId, accumulated across every nav-api row in this environment, so a
```
konrad-krol-incubly: I personally prefer if code is self-descriptive, here we have a lot of comments that explains what code does. Maybe you can try to refactor and extract some parts into private method - method name can explain what this code does.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/NavigationItemEntryMaterializer.java
MateuszJozwiak-Incubly: Done the upgrader so moving on to next coming regarding it.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ApiNavigationSubtreePaths.java
```diff
@@ -0,0 +1,92 @@
...
+                );
+                continue;
+            }
+            var path = parentPath + "/" + segment;
+            var legacyId = HRIDToUUID.navigation().context(organizationId, environmentId).api(navApiRowId).folder(path).id();
+            if (!legacyId.equals(child.getId())) {
+                continue;
+            }
+            result.add(new PathedFolder(child, path));
+            walk(organizationId, environmentId, navApiRowId, child.getId(), path, childrenByParentId, result);
```
MateuszJozwiak-Incubly: We have accepted this issue for now.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/PortalNavigationItemApiOwnedRekeyUpgrader.java
```diff
@@ -0,0 +1,348 @@
...
+        String newId,
+        String newParentId,
+        String newRootId,
+        PortalNavigationReferenceType referenceType,
+        String referenceId
+    ) throws TechnicalException {
+        if (newId.equals(original.getId())) {
+            return;
+        }
+        if (portalNavigationItemRepository.findById(newId).isEmpty()) {
```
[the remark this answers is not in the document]
MateuszJozwiak-Incubly: let's leave it for now

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/ApiLinkResource.java
```diff
@@ -0,0 +1,76 @@
...
+            var output = getApiLinkUseCase.execute(new GetApiLinkUseCase.Input(auditInfo, linkId));
+            return Response.ok(PortalLinkMapper.INSTANCE.toApiLinkState(output.link(), linkHrid, apiHrid)).build();
+        } catch (PortalLinkNotFoundException e) {
+            throw new HRIDNotFoundException(linkHrid);
+        }
+    }
+
+    @DELETE
+    @Permissions({ @Permission(value = RolePermission.API_DOCUMENTATION, acls = RolePermissionAction.DELETE) })
+    public Response deleteApiLinkByHrid(@PathParam("apiHrid") String apiHrid, @PathParam("linkHrid") String linkHrid) {
```
dev-marek, line 65: The DELETE endpoint (deleteApiLinkByHrid) has zero test coverage — no test ever issues a DELETE request. Its permission check, the PortalLinkNotFoundException→404 mapping, and the use-case wiring are all unverified. Compare to the sibling PortalLinkResourceTest (same module, same PR context), which tests both the 204 happy path and the 404 case for the equivalent portal-link DELETE. This is a destructive, publicly-reachable operation shipped with no test that would fail if it were broken.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/ApiLinksResource.java
```diff
@@ -0,0 +1,96 @@
...
+            spec.getLocation(),
+            spec.getOrder()
+        );
+        var output = dryRun ? validateApiLinkUseCase.execute(input) : createOrUpdateApiLinkUseCase.execute(input);
+
+        var state = PortalLinkMapper.INSTANCE.toApiLinkState(spec, linkId.toString(), output.errors(), auditInfo, apiHrid);
+
+        // A dry run is a preview: severe findings are its payload, so it always answers 200.
+        // A real apply that produced severe errors persisted nothing — it must not report success.
+        var applyFailed = !dryRun && output.errors().stream().anyMatch(Validator.Error::isSevere);
```
dev-marek, line 92: The new applyFailed = !dryRun && output.errors().stream().anyMatch(Validator.Error::isSevere) branch (200 vs 400 status) is untested for API links — ApiLinkResourceTest.Run only covers the all-clear 200 case. The sibling PortalLinksResourceTest.Run has an explicit should_return_400_with_severe_errors_when_apply_fails_validation test for the identical logic; nothing here would catch a flipped ternary or a dropped !dryRun.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/main/java/io/gravitee/apim/rest/api/automation/resource/ApiLinkResource.java
```diff
@@ -0,0 +1,76 @@
...
+    @Inject
+    private GetApiLinkUseCase getApiLinkUseCase;
+
+    @Inject
+    private DeleteApiLinkUseCase deleteApiLinkUseCase;
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Permissions({ @Permission(value = RolePermission.API_DOCUMENTATION, acls = RolePermissionAction.READ) })
+    public Response getApiLinkByHRID(@PathParam("apiHrid") String apiHrid, @PathParam("linkHrid") String linkHrid) {
```
dev-marek, line 51: No happy-path GET test — only the 404 case is covered. PortalLinkResourceTest.Get.should_return_the_link asserts the returned field values (name, href, order, portalHrid); no equivalent exists here to assert apiHrid/other fields on a successful GET. Dry-run is also untested for ApiLinksResource (PortalLinksResourceTest.DryRun has three tests: pass, fail, and 400-on-missing-hrid; none exist for the API variant).

dev-marek (changes requested):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/upgrade/upgrader/ApiNavigationSubtreePaths.java
```diff
@@ -0,0 +1,105 @@
...
+            .getOrDefault(parentId, List.of())
+            .stream()
+            .filter(child -> child.getType() == PortalNavigationItem.Type.FOLDER)
+            .map(child -> matchLegacyFolder(organizationId, environmentId, navApiRowId, parentPath, child))
+            .flatMap(Optional::stream)
+            .toList();
+
+        for (var match : matches) {
+            result.add(match);
+            walk(organizationId, environmentId, navApiRowId, match.folder().getId(), match.path(), childrenByParentId, result);
```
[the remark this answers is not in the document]
MateuszJozwiak-Incubly: @AikidoSec ignore: accepted, deterministic id check prevents real cycles

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/ApiDocumentationSyncDomainService.java
```diff
@@ -97,43 +83,40 @@ public void dematerialize(AuditInfo auditInfo, String apiId, PortalPageContentId
...
 
     /**
-     * Cleans up navigation rows materialized by the API itself (api-folder subtree + phantom-parented doc pages).
-     * Nav-api rows are owned by their {@code PortalListing} and {@code PortalPageContent} rows by their
-     * {@code Documentation}, so neither is touched here.
+     * Cleans up the API's own subtree — its folder subtree and its documentation pages, wherever they are
+     * rooted — enumerated by reference rather than by iterating nav-api rows: the subtree belongs to the
+     * API, not to any listing. Nav-api rows are owned by their {@code PortalListing}, so none are touched here.
      */
     public void cleanupForApi(AuditInfo auditInfo, String apiId) {
```
dev-marek, line 90: **Blocker (widened-by-change):** `cleanupForApi` deletes FOLDER roots (+ descendants) and PAGE-type automation rows, but never LINK-type rows.

A root-level API link (created with no `location`) is a parentless root — see `PortalLinkSyncDomainService#upsert`: `if (parent == null) existingLink.markAsRoot();`. It's unreachable by the FOLDER-root cascade (wrong type, no parent folder to cascade from) and excluded by the `item.getType() == PortalNavigationItemType.PAGE` filter below.

**Repro:** `PUT .../apis/{apiHrid}/links/{hrid}` with no `location` → root API link. `DELETE .../apis/{apiHrid}` → `cleanupForApi` leaves the link row behind, permanently orphaned (no `apiHrid` left to resolve for a manual `DELETE .../links/{hrid}` afterward).

This contradicts the PR description's claim that deleting the API removes its links. No test exercises `cleanupForApi` with a LINK item (root or nested), so this looks like a genuine coverage gap rather than an intentional deferral.

dev-marek (changes requested):
**Verdict: Request changes** — one confirmed data-integrity bug in the API-deletion cleanup path; everything else in this large refactor is well-designed and well-tested.

### Summary

This PR exposes API-referenced links through the Automation REST API (`ApiLinkResource`/`ApiLinksResource`) and completes a structural move of API navigation ownership (folders, doc pages, links) from per-listing rows onto the API itself, including a carefully-reasoned, well-tested boot-time upgrader (`PortalNavigationItemApiOwnedRekeyUpgrader` + `ApiNavigationSubtreePaths`) that re-keys legacy per-listing rows and converges duplicates. I traced the reference-scoped segment-conflict fix (`NavigationItemReference.sharesRootNamespaceWith`), the folder-subtree reconciler, and the migration's duplicate-convergence/idempotency logic in detail — all correct and covered by targeted tests (e.g. `two_portals_listing_the_same_api_converge_on_one_subtree`, `a_second_run_changes_nothing`).

The one gap: `ApiDocumentationSyncDomainService#cleanupForApi`, rewritten in this PR to enumerate by reference instead of by nav-api row, only cleans up FOLDER and PAGE rows — never LINK rows. A root-level API link (no `location`, i.e. a parentless root, which `PortalLinkSyncDomainService` explicitly supports) is invisible to both branches of the cleanup and survives API deletion as an orphaned row with no remaining way to delete it through the REST API. This directly contradicts the PR description's own claim ("removing an API from a listing no longer deletes its... documentation, folders or links — only deleting the API itself does"). No test exercises `cleanupForApi` with a link, root or otherwise, which is consistent with this being a genuine coverage gap rather than an intentional deferral. See inline comment for details.

Note: this method's underlying gap traces back to `part1`'s `resolveApiParent` (root links were already parentless there), but this is the PR that ships the REST surface making API links reachable/creatable by real users and GitOps config for the first time, and it's the PR that rewrote `cleanupForApi` itself — so it's the right place to fix it.

### Findings

| Severity | Origin | Location | Finding |
|---|---|---|---|
| Blocker | widened-by-change | `ApiDocumentationSyncDomainService.java:90-111` (`cleanupForApi`) | Deleting an API leaves root-level API-attached LINK rows permanently orphaned — the method only deletes FOLDER and PAGE types, never LINK. |

### Residual risks

- **Compatibility-sensitive surfaces**: this PR changes the Automation API OpenAPI contract (new `API Links` paths/schemas) and `portal_navigation_items` semantics via a boot-time upgrader. Worth calling out explicitly as compatibility-sensitive alongside the migration notes already in the description.
- **Migration at scale**: `PortalNavigationItemApiOwnedRekeyUpgrader` runs a single full-environment scan held in memory (documented as intentional, mirroring the existing metadata upgrader) — worth confirming this is acceptable for the largest expected navigation-table sizes in production.
- **No GET-all/list endpoint for API links** (`ApiLinksResource` only has PUT, plus GET/DELETE on the singular resource) — this exactly mirrors the existing `PortalLinksResource` pattern, so not filed as a finding, but the PR description's "listing an API's links" phrasing slightly overstates what ships here if a dedicated list endpoint was intended.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/ApiDocumentationSyncDomainService.java
MateuszJozwiak-Incubly: Valid point, fixed :)

MateuszJozwiak-Incubly:
Explanation of the PR's size:  
The bulk of the diff is a structural rework of API-folder-subtree ownership that those new endpoints needed as a foundation:                                                             
                                                                                                                                                                           
  - Old model: API-owned folders/pages were parented under a materialized PortalNavigationApi "row" — a listing entry with its own ID.                                     
  - New model: ownership moved onto a NavigationItemReference field (PortalReference/ApiReference) carried directly on PortalNavigationItem. API-folder roots now have parentId = null and are found via reference.equals(...) + isRoot() instead of looking up a nav-api row.

PR #19288, opened 2026-08-20
fix(ci): fall back to git when the GitHub API cannot answer for the base branch

phiz71:
## Issue

No ticket: CI reliability fix.

## Description

When the GitHub API stopped returning `base.sha`, every pipeline with an open pull request
went red. The `generate-config` setup job resolves the base branch of the current branch,
and everything else depends on the answer: it is what the changed files are diffed against,
and therefore what decides which jobs get scheduled at all. That resolution had no way to
carry on without the API.

Three changes, and deliberately nothing else.

**The API call is allowed to fail.** It had no `--fail` and no status check, so an error
body — a rate limited 403 included — was parsed like a successful answer, and
`jq -r '.base.ref // "master"'` turned it into `origin/master` for every branch, silently. A
branch cut from `4.12.x` was then diffed against `master`. The call now fails properly, and
a response without a usable `base.ref` and `base.sha` — this morning's case exactly — falls
through to the path below instead of taking the pipeline down.

**The call is bounded in time.** A connection that never answers is the one failure `--fail`
cannot turn into a fall through: it does not return, so the fallback is never reached and
the job sits there until the pipeline times out. `--max-time 10` returns after 10.02s with
exit 28, and the fallback runs.

**The history walk it falls back to can no longer poison `$BASH_ENV`.** That path already
existed, for branches with no open pull request. It greps `git branch -r --contains`, which
returns *one line per containing branch*, and `origin/HEAD -> origin/master` alone makes
`origin/master` match twice. The result went straight into `$BASH_ENV`, so the file gained
lines that are not assignments; CircleCI sources it under `bash -eo pipefail` before every
step, so each later step died before it started. It was rarely hit while it only served
branches without a pull request. It is the fallback now, so it takes a `head -1`.

The two values written into `$BASH_ENV` are quoted with `printf '%q'` in the same spirit: on
the API path `BASE_REF` is `base.ref` verbatim, and `git check-ref-format` allows `;`, `&`
and backticks in a branch name.

## Additional context

Exercised locally against this repository, by extracting the step and running it under
`bash -eo pipefail` with a real `$BASH_ENV`:

| case | result |
|---|---|
| open pull request, API answering | base from the API, unchanged behaviour |
| API failing (`401`) | `curl: (56)`, then the history walk, pipeline carries on |
| API answering without `base.sha` — this morning | history walk, pipeline carries on |
| API never answering | `--max-time` returns at 10.02s, exit 28, history walk |
| no pull request | history walk |
| all of them | `$BASH_ENV` sources cleanly, one line per variable |

Measured on 8 local branches, the multi-line value the `head -1` fixes was produced by 6 of
them, and sourcing such a file under `bash -eo pipefail` exits 1 before the step body runs.

On the quoting, a branch named `feat;<command>` as the base of a pull request writes
`export GIT_BASE_BRANCH=origin/feat;<command>` into `$BASH_ENV`, and *every* later step of
the job runs `<command>` while sourcing it — verified locally. `%q` keeps the value intact
and inert. Creating such a branch takes write access to the repository, so this is
hardening rather than an open door.

Two known defects are left alone on purpose, since neither is what broke: the walk gives up
after 100 commits, and giving up calls `circleci step halt`, which leaves a green pipeline
that has run no test at all.

Backport labels are set for `4.12.x`, `4.11.x`, `4.10.x` and `4.9.x`: the block is
byte-identical on all of them.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,43 +78,47 @@ jobs:
                       - run:
                             name: Find base branch
                             command: |
-                                if [ "x${GITHUB_TOKEN}" == "x" ]; then
-                                  echo "[GITHUB_TOKEN] is not set, can't find the base branch. Use 'master'"
-                                  echo "export GIT_BASE_BRANCH=master" >> $BASH_ENV
-                                fi;
+                                # `--fail`, so that an error body -- a rate limited 403 included -- stops being
+                                # parsed like an answer: the jq defaults below used to turn it into `origin/master`.
+                                # Nothing here is fatal any more; a missing answer falls through to the history walk.
+                                if [ "x${CIRCLE_PULL_REQUEST}" != "x" ] && [ "x${GITHUB_TOKEN}" != "x" ]; then
+                                  PULL_REQUEST_URL=$(echo https://api.github.com/repos/${CIRCLE_PULL_REQUEST:19} | sed "s/\/pull\//\/pulls\//")
 
-                                if [ "x${CIRCLE_PULL_REQUEST}" == "x" ]; then
-                                  echo "No opened pull request for this branch -> try to find the base branch with commits."
-                                  # Loop over the last 100 commits maximum and find the first commit that is on a branch matching `origin/master` or `origin/[0-9].[0-9].x`
+                                  if PULL_REQUEST_RESPONSE_CONTENT=$(curl -sS --fail -H "Authorization: token ${GITHUB_TOKEN}" ${PULL_REQUEST_URL}); then
+                                    PULL_REQUEST_BASE=$(echo $PULL_REQUEST_RESPONSE_CONTENT | jq -r '.base.ref // ""')
+                                    COMMON_COMMIT=$(echo $PULL_REQUEST_RESPONSE_CONTENT | jq -r '.base.sha // ""')
+
+                                    if [ "x${PULL_REQUEST_BASE}" != "x" ] && [ "x${COMMON_COMMIT}" != "x" ]; then
+                                      BASE_REF=origin/${PULL_REQUEST_BASE}
+                                    else
+                                      COMMON_COMMIT=""
+                                    fi
+                                  else
+                                    echo "The GitHub API did not answer, looking for the base branch in the history."
+                                  fi
+                                fi
+
+                                # No pull request, or no usable answer: find the first commit that is also on `origin/master`
+                                # or on a support branch.
+                                if [ "x${COMMON_COMMIT}" == "x" ]; then
                                   MAX_COMMITS=100
-
-                                  # Define the pattern for matching branch names: origin/master or origin/[0-9].[0-9].x
                                   PATTERN="origin/master|origin/[0-9]+\.[0-9]+\.x"
 
-                                  # Loop over the last $MAX_COMMITS commits
                                   for (( i = 0; i < MAX_COMMITS; i++ )); do
-                                    # Get the commit hash
                                     hash=$(git rev-parse HEAD~$i)
-                                  
-                                    # Get the branch name that contains the commit hash and matches the pattern
-                                    branch=$(git --no-pager branch -r --contains $hash | grep -oE $PATTERN || true)
-                                  
-                                    # If the branch name is not empty, then the matching branch has been found
+
+                                    # `head -1`: a commit usually sits on several branches, and `origin/HEAD -> origin/master`
+                                    # alone makes `origin/master` match twice. Several lines here means several lines in
+                                    # $BASH_ENV, which every later step sources under `bash -eo pipefail` before running.
+                                    branch=$(git --no-pager branch -r --contains $hash | grep -oE $PATTERN | head -1 || true)
+
                                     if [ -n "$branch" ]; then
                                       BASE_REF=$branch
                                       COMMON_COMMIT=$hash
                                       break
                                     fi
                                   done
-
-                                # else use the pull request info
-                                else
-                                  PULL_REQUEST_URL=$(echo https://api.github.com/repos/${CIRCLE_PULL_REQUEST:19} | sed "s/\/pull\//\/pulls\//")
-                                  PULL_REQUEST_RESPONSE_CONTENT=$(curl -s -H "Authorization: token ${GITHUB_TOKEN}" ${PULL_REQUEST_URL})
-
-                                  BASE_REF=origin/$(echo $PULL_REQUEST_RESPONSE_CONTENT | jq -r '.base.ref // "master"')
-                                  COMMON_COMMIT=$(echo $PULL_REQUEST_RESPONSE_CONTENT | jq -r '.base.sha // ""')
-                                fi;
+                                fi
 
                                 # if BASE_REF is empty just stop
                                 if [ "x${BASE_REF}" == "x" ]; then
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,9 @@ jobs:
                                 if [ "x${CIRCLE_PULL_REQUEST}" != "x" ] && [ "x${GITHUB_TOKEN}" != "x" ]; then
                                   PULL_REQUEST_URL=$(echo https://api.github.com/repos/${CIRCLE_PULL_REQUEST:19} | sed "s/\/pull\//\/pulls\//")
 
-                                  if PULL_REQUEST_RESPONSE_CONTENT=$(curl -sS --fail -H "Authorization: token ${GITHUB_TOKEN}" ${PULL_REQUEST_URL}); then
+                                  # `--max-time`, because a connection that never answers is the one failure
+                                  # `--fail` cannot turn into a fall through: it would sit here until the job times out.
+                                  if PULL_REQUEST_RESPONSE_CONTENT=$(curl -sS --fail --max-time 10 -H "Authorization: token ${GITHUB_TOKEN}" ${PULL_REQUEST_URL}); then
                                     PULL_REQUEST_BASE=$(echo $PULL_REQUEST_RESPONSE_CONTENT | jq -r '.base.ref // ""')
                                     COMMON_COMMIT=$(echo $PULL_REQUEST_RESPONSE_CONTENT | jq -r '.base.sha // ""')
 
@@ -126,10 +128,13 @@ jobs:
                                   circleci step halt
                                 fi;
 
-                                echo "export GIT_BASE_BRANCH=${BASE_REF}" >> $BASH_ENV
+                                # `%q`, because on the pull request path these values come from the API rather than from
+                                # the pattern above, and `git check-ref-format` allows `;`, `&` and backticks in a branch
+                                # name — which $BASH_ENV would hand to every later step.
+                                printf 'export GIT_BASE_BRANCH=%q\n' "${BASE_REF}" >> $BASH_ENV
                                 echo "The base branch is ${BASE_REF}"
 
-                                echo "export GIT_COMMON_COMMIT_HASH=${COMMON_COMMIT}" >> $BASH_ENV
+                                printf 'export GIT_COMMON_COMMIT_HASH=%q\n' "${COMMON_COMMIT}" >> $BASH_ENV
                                 echo "The common commit hash is ${COMMON_COMMIT}"
 
             - node/install-packages:


.circleci/scripts/find-base-branch.sh
```diff
@@ -0,0 +1,124 @@
...
+  echo "${ref}"
+}
+
+# Among master and the support branches, the one whose merge base with HEAD is the closest.
+closest_base_ref() {
+  local candidates candidate merge_base distance
+  local best_ref="" best_distance=""
+
+  candidates=$(git for-each-ref --format='%(refname:short)' refs/remotes/origin \
+               | grep -E '^origin/(master|[0-9]+\.[0-9]+\.x)$' || true)
```
ytvnr: **Worth fixing**

`closest_base_ref` builds its candidate set from whatever `refs/remotes/origin` already carries and never fetches a missing one, so on a clone without `origin/4.12.x` a branch cut from `4.12.x` is answered with `origin/master`.

`remote_ref_for` (l.61-71) is built on the premise that "this clone does not carry it" is a real case, and fetches on demand. Here a missing candidate is indistinguishable from one that lost on distance, and nothing is logged. It is also the path taken whenever the API call above fails, which is the scenario this PR is written for, and the only one where a wrong answer stays green.

Making the candidate set explicit rather than discovered, or fetching missing ones as `remote_ref_for` already does, would close it. Logging the candidates actually compared would make the next incident readable.

.circleci/scripts/find-base-branch.sh
```diff
@@ -0,0 +1,124 @@
...
+    exit 1
+  fi
+fi
+
+COMMON_COMMIT=$(git merge-base HEAD "${BASE_REF}")
+
+log "The base branch is ${BASE_REF}"
+log "The common commit hash is ${COMMON_COMMIT}"
+
+printf 'export GIT_BASE_BRANCH=%s\nexport GIT_COMMON_COMMIT_HASH=%s\n' "${BASE_REF}" "${COMMON_COMMIT}"
```
ytvnr: **Worth fixing**

The two values are interpolated unquoted into a file that is sourced under `bash -eo pipefail`, and on the pull request path `BASE_REF` is a ref name that comes straight from `base.ref`.

The fallback constrains its answer with `^origin/(master|[0-9]+\.[0-9]+\.x)$`, but this path does not. `git check-ref-format` allows `;`, `&`, `|`, `$` and backticks in a branch name, and any of them reaches `$BASH_ENV` verbatim and takes down every later step before it runs, which is the failure mode the header at l.15-18 says is closed.

`printf '%q'` on the two values would make the stated invariant hold for everything the script can produce, not only for what the fallback pattern happens to sanitise.

.circleci/scripts/find-base-branch.sh
```diff
@@ -0,0 +1,124 @@
...
+  local best_ref="" best_distance=""
+
+  candidates=$(git for-each-ref --format='%(refname:short)' refs/remotes/origin \
+               | grep -E '^origin/(master|[0-9]+\.[0-9]+\.x)$' || true)
+
+  for candidate in ${candidates}; do
+    merge_base=$(git merge-base HEAD "${candidate}") || continue
+    distance=$(git rev-list --count "${merge_base}..HEAD")
+
+    if [ -z "${best_distance}" ] || [ "${distance}" -lt "${best_distance}" ]; then
```
ytvnr: Ties in merge base distance go to the lexicographically first ref rather than to master: `-lt` keeps the incumbent, and `for-each-ref` sorts by refname.

On this repository the order is `origin/3.10.x` ... `origin/4.12.x` ... `origin/master`, so a branch cut from master before the recent support branches were cut ties across all of them and gets the oldest one. Impact is bounded: the merge base is the same commit either way, so `GIT_COMMON_COMMIT_HASH` stays right and only `GIT_BASE_BRANCH` is off, whose consumer is the cache key in `cmd-restore-maven-cache.ts`.

An explicit tie break preferring `origin/master` would remove the surprise.

.circleci/scripts/find-base-branch.sh
```diff
@@ -0,0 +1,124 @@
...
+closest_base_ref() {
+  local candidates candidate merge_base distance
+  local best_ref="" best_distance=""
+
+  candidates=$(git for-each-ref --format='%(refname:short)' refs/remotes/origin \
+               | grep -E '^origin/(master|[0-9]+\.[0-9]+\.x)$' || true)
+
+  for candidate in ${candidates}; do
+    merge_base=$(git merge-base HEAD "${candidate}") || continue
+    distance=$(git rev-list --count "${merge_base}..HEAD")
```
ytvnr: The pattern matches every support branch back to `origin/3.5.x`, 22 of them here, and each one costs a `merge-base` plus a full `rev-list --count`. Run unmodified on a warm local clone that is 5.8s wall and 4.8s CPU, paid in the job that gates the whole pipeline.

The counts for the oldest branches run into tens of thousands of commits and can never win. An ancestry test against the current best, or a pattern narrowed to the branches still supported, would keep it cheap.

.circleci/config.yml
```diff
@@ -77,56 +77,7 @@ jobs:
...
-                                  echo "No base branch found, stopping"
-                                  circleci step halt
-                                fi;
-
-                                echo "export GIT_BASE_BRANCH=${BASE_REF}" >> $BASH_ENV
-                                echo "The base branch is ${BASE_REF}"
-
-                                echo "export GIT_COMMON_COMMIT_HASH=${COMMON_COMMIT}" >> $BASH_ENV
-                                echo "The common commit hash is ${COMMON_COMMIT}"
+                            command: .circleci/scripts/find-base-branch.sh >> $BASH_ENV
```
ytvnr: The step just below lints and unit tests the TypeScript generator, but the script that decides what everything else is diffed against has no coverage at all.

Extracting it into a real file made it runnable in isolation, which was most of the work. The two failure modes fixed here are still guarded only by manual testing. A shellcheck step, or a small harness over throwaway repositories next to the existing lint and test step, would keep them fixed.

ytvnr (approved):
Solid fix. Extracting the resolution into a script that can be run on its own, and turning the GitHub call into a hint that is allowed to fail, are both the right calls. The `$BASH_ENV` poisoning is genuinely closed on every path I could reach: `curl --silent`, `git fetch --quiet` and the `log()` redirection keep stdout clean, so the file gets exactly two lines.

Two things I would still look at, neither of them blocking:

- the git only fallback compares whatever `refs/remotes/origin` happens to carry and never fetches a missing candidate, the way `remote_ref_for` does. That is the path taken exactly when GitHub is unreachable, and a missing `origin/4.12.x` is silently answered with `origin/master`.
- the final `printf` interpolates its values unquoted, and on the pull request path the value is a ref name straight from `base.ref`, which the fallback pattern never gets to sanitise.

The rest is minor. Approving.

phiz71:
@ytvnr heads up: I have rewritten the branch, so your review points at a version that no
longer exists. Sorry for the detour, and thanks — half of what you found is what convinced
me to cut it back.

The scope had drifted. What broke was one thing: the base branch resolution had no way to
carry on when the API could not answer. The branch had grown a script, a merge-base
distance comparison over 22 candidate branches, an on-demand fetch, and a `halt` turned into
a failure. Only the first of those is the fix; the rest was work I had given myself, and
your review was mostly about defects in that extra work rather than in the repair.

It is now two changes on the existing block, 29 insertions and 25 deletions in
`config.yml`, no new file: `--fail` plus a fall-through to the history walk when the answer
is unusable, and a `head -1` on that walk, which becomes reachable far more often now that
it is the fallback.

Where that leaves your five points:

- **fetching missing candidates** and **the cost of 22 `rev-list --count`** are gone with the
  code that introduced them. There is no candidate set and no distance comparison any more.
- **the unquoted interpolation into `$BASH_ENV`** still stands. It predates this branch — the
  `echo "export GIT_BASE_BRANCH=${BASE_REF}"` lines are untouched from `master` — but your
  reasoning holds: on the API path the value is `base.ref` verbatim, and
  `git check-ref-format` allows `;`, `&` and backticks. It is the same failure mode this pull
  request closes from the other side, so it is a fair thing to fix here; I left it out rather
  than widen the scope again right after narrowing it. Happy to add the `printf '%q'` if you
  think it belongs in this one.
- **the tie break** survives in a smaller form: `head -1` takes the first line of
  `git branch -r`, which is sorted, so a commit on several branches is labelled
  `origin/3.10.x` rather than `origin/master`. As you noted, `GIT_COMMON_COMMIT_HASH` is
  right either way and only the cache key sees the difference.
- **no test coverage** is still true, but there is no longer a standalone script to run, so
  it is back to being a property of the whole block rather than something this change makes
  worse.

Two defects I deliberately left alone, since neither is what broke: the walk gives up after
100 commits, and giving up calls `circleci step halt`, which leaves a green pipeline that has
run no test at all. Worth their own change if we want them.

The backport labels are unchanged; the block is still byte-identical on `4.9.x` through
`4.12.x`.

ytvnr:
Re-read the new version. The narrowing is the right call, and the two changes do the job: `--fail` plus the fall through means an unusable answer no longer becomes `origin/master`, and dropping the token-less early `echo` removes a second `GIT_BASE_BRANCH` line that used to be appended alongside whatever the walk found.

On your question: yes, I would put the quoting in this one. The lines are untouched from master, so it is a widening, but it is two lines in the block you are already editing, it is the same failure mode this pull request closes from the other side, and the value it protects is the one the API hands you. The backport labels are already set and the block is still byte identical on 4.9.x through 4.12.x, so it ships everywhere for free here, where a separate change means doing the backport dance again for two lines.

One thing worth carrying over from the version you deleted: the `curl` there had `--max-time 10` and this one has none. Not a regression against master, which had none either, but a hang is the one API failure that `--fail` does not catch: it never returns, so it never falls through to the walk, and the setup job sits there until the CircleCI timeout. One flag, same goal as the rest of the change.

The two you left alone read right to me as separate work.

.circleci/config.yml
```diff
@@ -78,43 +78,47 @@ jobs:
...
+                                # parsed like an answer: the jq defaults below used to turn it into `origin/master`.
+                                # Nothing here is fatal any more; a missing answer falls through to the history walk.
+                                if [ "x${CIRCLE_PULL_REQUEST}" != "x" ] && [ "x${GITHUB_TOKEN}" != "x" ]; then
+                                  PULL_REQUEST_URL=$(echo https://api.github.com/repos/${CIRCLE_PULL_REQUEST:19} | sed "s/\/pull\//\/pulls\//")
 
-                                if [ "x${CIRCLE_PULL_REQUEST}" == "x" ]; then
-                                  echo "No opened pull request for this branch -> try to find the base branch with commits."
-                                  # Loop over the last 100 commits maximum and find the first commit that is on a branch matching `origin/master` or `origin/[0-9].[0-9].x`
-                                  MAX_COMMITS=100
+                                  if PULL_REQUEST_RESPONSE_CONTENT=$(curl -sS --fail -H "Authorization: token ${GITHUB_TOKEN}" ${PULL_REQUEST_URL}); then
```
ytvnr: The `curl` in the version you deleted had `--max-time 10` and this one has none, so a hang is the one API failure that `--fail` does not catch.

`--fail` turns a bad status into a non-zero exit, which is what makes the fall through work. A connection that never answers produces neither: the call does not return, the walk below is never reached, and the setup job sits there until the CircleCI timeout takes the pipeline down. Same outcome as before this change, on the one path the change was written to close.

Not a regression against master, which had no timeout either, but it is one flag and it belongs with the rest of the hardening.

ytvnr (approved):
Re-approving on `36beb433`, since the previous approval was attached to a version that no longer exists.

The rewritten block holds up: an unusable answer from the API now falls through to the history walk instead of becoming `origin/master`, `head -1` keeps `$BASH_ENV` to one line per variable, and the token-less early `echo` that used to append a second `GIT_BASE_BRANCH` alongside the walk result is gone.

One point inline, and one that lands outside the diff: the quoting on the two `echo "export ..."` lines at l.129 and l.132, which is the question you raised in the comment above. My answer is there.

.circleci/config.yml
phiz71: Applied in `d2ebba2c`, thanks — you named the hole in the change: the call was made non blocking everywhere except where it actually blocks.

Measured against an unroutable address: `--max-time 10` returns at 10.02s with exit 28, so it takes the same route as the 401 and lands in the walk. The comment on the line says why the flag is not redundant with `--fail`, since that is the part that is easy to drop again later.

phiz71:
Both in `d2ebba2c`, and thanks for the re-approval on the rewritten branch.

On the quoting: your backport argument is what settled it. Two lines in a block I am already
editing, shipping to `4.9.x` through `4.12.x` off labels that are already set, against doing
the whole backport dance again later for the same two lines.

Worth noting it is a bit sharper than "takes down every later step". Verified locally with a
base branch named `feat;<command>`:

```
$ cat $BASH_ENV                       # before
export GIT_BASE_BRANCH=origin/feat;touch /tmp/pwned-$$
$ bash -eo pipefail -c 'source $BASH_ENV; echo "step ran, value=$GIT_BASE_BRANCH"'
step ran, value=origin/feat
/tmp/pwned-82990                      # the command ran

$ cat $BASH_ENV                       # after
export GIT_BASE_BRANCH=origin/feat\;touch\ /tmp/pwned-\$\$
$ bash -eo pipefail -c 'source $BASH_ENV; echo "step ran, value=$GIT_BASE_BRANCH"'
step ran, value=origin/feat;touch /tmp/pwned-$$
```

So the step does not die, it runs the command, once per step of the job. Creating such a
branch takes write access to the repository, so it is hardening rather than an open door,
but that is the reason it is worth two lines. The description carries both, along with the
`--max-time` measurement.

The 100-commit horizon and the `halt` stay out, as we agreed.

PR #19289, opened 2026-08-20
feat(gamma): edit and remove group members with ownership transfer

ankita-gupta21:
## FOUND-107 — Edit and remove group members 

Stacked on top of `FOUND-109-gamma-invite-members`. **Please set the base branch to
`FOUND-109-gamma-invite-members`**, otherwise the diff also shows the invites work.

### What this covers

Group administration in Gamma could add and invite members, but there was no way to
change or revoke access afterwards — any role correction meant switching back to the
classic Console. This PR closes that gap on the group detail page:

- **Edit member roles** — a sheet on the member row action that edits every scoped role
  (API, API product, Application, Integration, Cluster, Explorer) plus the *group admin*
  flag, in one save.
- **Primary-ownership transfer** — demoting a primary owner requires picking a successor;
  promoting a member automatically demotes the incumbent to `OWNER`. The sheet explains
  in prose exactly what will change before the operator saves.
- **Remove member** — a confirmation dialog that asks for a successor first when the
  member being removed holds primary ownership.
- **Explorer role column** on the members table, so the role the group grants is visible
  without opening the edit sheet.
- **Rename** `GroupDeleteSheet` → `GroupDeleteDialog`, so the name matches what the
  component actually renders (it has always been a `Dialog`, not a `Sheet`).

### Alignment with the classic Console

Behaviour was checked against
`gravitee-apim-console-webui/src/management/settings/groups/group/`.

At parity:

| Classic Console | Gamma |
| --- | --- |
| `edit-member-dialog` — all scoped roles + group admin | `GroupEditMemberSheet` |
| `delete-member-dialog` — successor picker for primary owners | `GroupRemoveMemberDialog` |
| Ownership transfer message ("… will become the API primary owner of this group.") | `buildEditOwnershipTransferMessage` |
| `disableDeleteMember()` — sole primary owner cannot be removed | `isRemoveDisabled` |
| Group admins manage members without `environment-group-u` | `useCurrentUserIsGroupAdmin` |
| `PRIMARY_OWNER` disabled when the API primary-owner mode is `USER` | `isPrimaryOwnerUnavailable` |

Deliberate differences, all in Gamma's favour:

- **Transfer happens before removal, not after.** `group.component.ts` deletes the
  membership first and only then calls `addOrUpdateMemberships` with the successor, so a
  failure in between leaves the group with no primary owner. Gamma transfers first, then
  deletes, and if the delete fails it restores both memberships to their previous roles
  and tells the operator that ownership was restored. If the restore itself fails, the
  dialog stays open with an explicit reconcile message rather than a silent success.
- **All five primary-owner scopes are guarded**, not just API and API product. Classic's
  `disableDeleteMember()` only inspects `API`/`API_PRODUCT`, so removing the sole
  Application primary owner slips through; Gamma uses `primaryOwnerScopesOf`, which also
  covers `APPLICATION`, `INTEGRATION`, and `CLUSTER`. `EXPLORER` is intentionally *not* a
  transferable scope.
- **Disabled "Remove member" explains itself** via a tooltip instead of being inert.

### Technical background

**Management API.** Both flows use the existing endpoints — no backend change:

- `POST /configuration/groups/{group}/members` (`addOrUpdateGroupMember`) for role edits
  and for the ownership transfer; it takes a list, so a transfer and the edited member's
  new roles go out as one request.
- `DELETE /configuration/groups/{group}/members/{member}` (`removeGroupMember`) for
  removal.

**Where the logic lives.** All ownership reasoning is pure and unit-tested in
`utils/primaryOwnership.ts` (`analyzeEditOwnershipTransfer`,
`buildEditOwnershipTransferMessage`, `buildEditMembershipPayloads`), so the components
stay declarative. `useGroupEditMemberForm` owns the sheet's form state; orchestration and
toasts live in `useGroupMemberActions`; `MemberSuccessorCombobox` is shared by the edit
sheet and the remove dialog.


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMemberActions.ts
```diff
@@ -94,6 +98,56 @@ export function useGroupMemberActions(groupId: string | undefined) {
...
+            return;
+        }
+
+        if (transferMembership && !(await transferPrimaryOwnership(groupId, transferMembership))) {
+            return;
+        }
+
+        try {
+            await removeMemberMutation.mutateAsync({ groupId, memberId: removingMember.id });
+        } catch (error) {
```
aqibmohammadkhan, line 181: **1. No rollback when removal fails after ownership transfer**

When ownership transfer succeeds but the subsequent member removal fails, no rollback of the transfer is attempted — this contradicts the PR description's stated restore-on-failure behavior.

Admin removes a primary owner, picking a successor: `transferPrimaryOwnership` succeeds (successor is now `PRIMARY_OWNER`) but `removeMemberMutation` then fails (network blip, 5xx). The catch block only shows an error toast and returns — it never demotes the successor or restores the original owner. The PR's own test is titled *"reports partial success without attempting an unsafe rollback"* and asserts no compensating call happens, directly contradicting the PR description's claim that Gamma "restores both memberships to their previous roles" on delete failure. If the admin closes the dialog instead of retrying, the group is left with an unintended ownership change while the member meant to be removed is still present.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
```diff
@@ -111,32 +115,40 @@ export function GroupDetailPage() {
...
     const canAddMembers =
         group !== undefined &&
-        canManageMembers &&
-        Boolean(group.system_invitation || group.email_invitation) &&
-        !maxInvitationsLimitReached;
-    const { data: environmentSettings } = useEnvironmentSettings({ enabled: memberSheet !== 'closed' });
+        (canEdit || canInviteToGroup(group)) &&
+        !maxInvitationsLimitReached &&
+        Boolean(group.system_invitation || group.email_invitation);
+    const canManageMemberActions = canEdit || isCurrentUserGroupAdmin;
```
aqibmohammadkhan: **2. Member-action permission gate diverges from invite gate**

`canManageMemberActions` was changed from a server-computed `manageable` flag to a client-derived `isCurrentUserGroupAdmin` check, while the sibling `canInviteToGroup` gate still honors the server flag — creating an inconsistent access-control surface.

Server-side, `group.manageable` is set true for any user holding CREATE, UPDATE, or DELETE on `ENVIRONMENT_GROUP`, or who holds the GROUP-scope `ADMIN` role for that group. Client-side, `canEdit` only checks the UPDATE permission and `isCurrentUserGroupAdmin` only matches `GROUP===ADMIN` membership. A user granted only CREATE or DELETE on `ENVIRONMENT_GROUP` (not UPDATE), who isn't a GROUP ADMIN member, still gets `manageable=true` from the backend and sees "Add members," but `canManageMemberActions` is now false, hiding Edit-roles/Remove-member menu items and invitation-delete buttons they previously had access to.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupEditMemberForm.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    const disabledOptionNames = {
+        api: isPrimaryOwnerUnavailable(apiPrimaryOwnerMode) ? PRIMARY_OWNER_DISABLED_OPTIONS : undefined,
+        apiProduct: isPrimaryOwnerUnavailable(apiProductPrimaryOwnerMode) ? PRIMARY_OWNER_DISABLED_OPTIONS : undefined,
+    };
+
+    function handleRoleChange(field: RoleField, value: string) {
+        setRoleValues(previous => ({ ...previous, [field]: value }));
+        if (field !== 'explorerRole') {
+            setSelectedSuccessorId(null);
```
aqibmohammadkhan, line 93: **3. Successor pick cleared by unrelated role edits**

`handleRoleChange` clears the already-picked successor on any role field change except Explorer, even when the change is unrelated to the scope requiring a successor.

Admin demotes a member from API `PRIMARY_OWNER` (triggers `needsSuccessor` for the API scope), picks a successor, then also changes the member's unrelated Cluster role in the same sheet. `handleRoleChange` resets `selectedSuccessorId` to `null` for every field except `explorerRole`, so the already-chosen successor is silently discarded and Save re-disables — appearing as a stuck Save button with no explanation.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/components/ConfirmDialog.tsx
```diff
@@ -64,19 +73,17 @@ export function ConfirmDialog({
...
     destructive = false,
     isPending = false,
+    confirmDisabled = false,
     confirmKeyword,
     icon,
+    children,
     onConfirm,
 }: Readonly<ConfirmDialogProps>) {
     const inputId = useId();
     const [typed, setTyped] = useState('');
```
aqibmohammadkhan, line 83: **4. Confirm-dialog typed text no longer clears on close**

The effect that cleared type-to-confirm input text immediately on dialog close was removed; the typed value now only resets on the next open via a remount key.

A destructive-action dialog requires typing a resource name to confirm. The dialog's `DialogContent` stays mounted during its close animation (standard Radix/shadcn Presence behavior), so with `useEffect(() => { if (!open) setTyped('') }, [open])` removed, the previously typed text remains visibly rendered in the input for the duration of the close transition instead of clearing immediately. No test in `ConfirmDialog.spec.tsx` covers close-then-reopen or the close-animation window.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/primaryOwnership.ts
```diff
@@ -0,0 +1,221 @@
...
+    transfer: EditOwnershipTransfer,
+    selectedSuccessor: GroupMember | null,
+    selections?: Pick<MemberRoleSelections, RoleField>,
+): string | null {
+    const parts: string[] = [];
+    const isApiDowngrade = transfer.downgradeScopes.includes('API');
+    const isApiProductDowngrade = transfer.downgradeScopes.includes('API_PRODUCT');
+
+    if (selectedSuccessor) {
+        if (isApiDowngrade && isApiProductDowngrade) {
```
aqibmohammadkhan, line 198: **5. Ownership-transfer message can render a blank role clause**

`buildEditOwnershipTransferMessage` can produce a sentence with a dangling, descriptor-less clause when a member's role is cleared to "No role" while a compound API/API_PRODUCT downgrade is in effect.

`GroupRoleSelect` exposes an explicit "No role" option that maps back to `''`. A member previously `PRIMARY_OWNER` for both API and API_PRODUCT can have `apiRole` set to `''` while `apiProductRole` is set to a different role, making both `isApiDowngrade` and `isApiProductDowngrade` true with `selections.apiRole` falsy. `roleUpdatePhrase('')` returns `''`, so the compound branch produces a message like *"...will be updated (API) and as owner (API Product)"* — a malformed sentence shown directly to the operator confirming a sensitive ownership change.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupMembersTable.tsx
```diff
@@ -17,35 +17,55 @@
...
 
-function roleCell(member: GroupMember, scope: 'API' | 'APPLICATION' | 'API_PRODUCT' | 'INTEGRATION' | 'CLUSTER') {
+function roleCell(member: GroupMember, scope: 'API' | 'APPLICATION' | 'API_PRODUCT' | 'INTEGRATION' | 'CLUSTER' | 'EXPLORER') {
     const role = member.roles?.[scope];
     return <span className="text-sm text-muted-foreground">{role ?? '—'}</span>;
 }
 
-function buildColumns(): DataTableProps<GroupMember>['columns'] {
-    return [
+function isRemoveDisabled(member: GroupMember, totalMemberCount: number): boolean {
```
aqibmohammadkhan, line 48: **6. Needs-successor rule duplicated across three files**

"Does this member need a successor to be removed" is independently re-derived in `GroupMembersTable` (`totalMemberCount===1 && primaryOwnerScopesOf(member).length>0`), `useGroupMemberActions` (`primaryOwnerScopesOf(removingMember).length>0`, no member-count factor), and `GroupRemoveMemberDialog` (same primitive, combined differently into `canConfirm`). None share a single named helper. They currently agree in practice only because the successor pool is unrestricted, but any future tightening of successor eligibility in one file won't propagate to the other two, silently reintroducing the "member removed with no valid successor" bug this PR was written to prevent.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupEditMemberForm.ts
```diff
@@ -0,0 +1,111 @@
...
+import { PRIMARY_OWNER_ROLE } from '../types/group';
+import { getMemberRoleLockFlags, sortedSuccessorCandidates, type MemberRoleSelections, type RoleField } from '../utils/memberRoles';
+import {
+    analyzeEditOwnershipTransfer,
+    buildEditMembershipPayloads,
+    buildEditOwnershipTransferMessage,
+    isPrimaryOwnerUnavailable,
+} from '../utils/primaryOwnership';
+
+const PRIMARY_OWNER_DISABLED_OPTIONS = new Set([PRIMARY_OWNER_ROLE]);
```
aqibmohammadkhan: **7. `PRIMARY_OWNER_DISABLED_OPTIONS` constant duplicated 3x**

The literal `new Set([PRIMARY_OWNER_ROLE])` constant and its guarding `isPrimaryOwnerUnavailable` check are copy-pasted verbatim across `useGroupEditMemberForm.ts:29`, `useGroupAddMembersForm.ts:38`, and `GroupInviteMemberSheet.tsx:31`, instead of living once beside the shared `primaryOwnership.ts` helper. A future change to which roles get disabled must be applied in three separate files; the edit, add, and invite flows can silently drift out of sync.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupRemoveMemberDialog.spec.tsx
```diff
@@ -0,0 +1,151 @@
...
+    });
+
+    it('names the member and the group in the confirmation message', () => {
+        renderSheet();
+        expect(screen.getByText('Anna Schmidt')).not.toBeNull();
+        expect(screen.getByText('API Team')).not.toBeNull();
+    });
+
+    it('calls onConfirm with no transfer membership for a non-primary-owner member', () => {
+        const { onConfirm } = renderSheet();
```
aqibmohammadkhan, line 56: **8. Successor picker + transfer message UI duplicated 2x**

`GroupRemoveMemberDialog` and `GroupEditMemberSheet` each hand-assemble the same successor-combobox-plus-transfer-message JSX/string pairing independently instead of sharing a composed component or message builder. `GroupRemoveMemberDialog` builds its own inline transfer-message string rather than reusing `scopedTransferSentence` from `primaryOwnership.ts`. A future wording or UX change to the transfer flow must be applied in multiple places.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMemberActions.ts
```diff
@@ -94,6 +98,56 @@ export function useGroupMemberActions(groupId: string | undefined) {
...
+    async function handleRemoveMember(transferMembership?: GroupMembershipPayload) {
+        if (!groupId || !removingMember) return;
+
+        const requiresOwnershipTransfer = primaryOwnerScopesOf(removingMember).length > 0;
+        if (requiresOwnershipTransfer && !transferMembership) {
+            notify.warning('Primary ownership must be transferred before removing this member');
+            return;
+        }
+
+        if (transferMembership && !(await transferPrimaryOwnership(groupId, transferMembership))) {
```
aqibmohammadkhan: **9. Transfer + removal double-invalidates members query**

`handleRemoveMember`'s transfer-then-remove sequence triggers two independent `invalidateQueries` calls against the same `groupKeys.members` key. Both the add-members mutation (transfer step) and the remove-member mutation independently invalidate the same query on success, so the first invalidation's refetch is thrown away seconds later by the second — one unnecessary members-list refetch on every ownership-transfer removal.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupDeleteDialog.tsx
```diff
@@ -33,9 +33,6 @@ export function GroupDeleteSheet({
     onConfirm: () => void;
     isDeleting: boolean;
 }>) {
```
aqibmohammadkhan, line 35: **10. Deleted explanatory comment during rename-only change**

A "why" comment explaining that Delete is always offered because the backend rejects primary-owner deletion with `StillPrimaryOwnerException` was deleted during the Sheet→Dialog rename, even though the guarded logic is unchanged. Root `AGENTS.md` states comments explaining a non-obvious constraint should be kept, and changes should touch only what the task requires. Losing this context risks a future contributor reintroducing a client-side pre-check that duplicates or conflicts with the backend's own guard.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMemberActions.ts
ankita-gupta21: Fixed. Transfer and removal are now one mutation that restores both memberships if the delete fails: [`useGroupMutations.ts#L115-L162`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMutations.ts#L115-L162). Distinct toasts per failure phase, including the case where rollback itself fails: [`useGroupMemberActions.ts#L131-L152`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMemberActions.ts#L131-L152). The old "no unsafe rollback" test is replaced by one asserting restoration.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
ankita-gupta21: Fixed. Reverted to the server-computed flag so both gates read the same source: [`GroupDetailPage.tsx#L145`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx#L145). The client-derived `useCurrentUserGroupAdmin` hook and its spec are deleted. This is an authorization change, so it deserves a second look on review.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupEditMemberForm.ts
ankita-gupta21: Fixed. The pick is only cleared when the set of downgraded scopes actually changes, so editing an unrelated Cluster/Explorer role keeps it: [`useGroupEditMemberForm.ts#L77-L84`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupEditMemberForm.ts#L77-L84).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/components/ConfirmDialog.tsx
ankita-gupta21: Fixed. The effect is restored so text clears immediately on close rather than waiting for the remount: [`ConfirmDialog.tsx#L85-L91`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/components/ConfirmDialog.tsx#L85-L91). Covered by a close-then-reopen test.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/primaryOwnership.ts
ankita-gupta21: Fixed. A cleared role now renders "with no role" instead of an empty clause: [`primaryOwnership.ts#L48-L50`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/primaryOwnership.ts#L48-L50), with a test for the compound API/API Product case.
**6

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupMembersTable.tsx
ankita-gupta21: Fixed. One named helper, [`requiresPrimaryOwnerSuccessor`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/primaryOwnership.ts#L80), is now the only definition, used by [`GroupMembersTable.tsx#L49`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupMembersTable.tsx#L49), [`GroupRemoveMemberDialog.tsx#L50`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupRemoveMemberDialog.tsx#L50), and [`useGroupMemberActions.ts#L123`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMemberActions.ts#L123).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupEditMemberForm.ts
ankita-gupta21: Fixed. Defined once beside the ownership helpers at [`primaryOwnership.ts#L30`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/primaryOwnership.ts#L30) and imported by the [edit](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupEditMemberForm.ts#L73), [add](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupAddMembersForm.ts#L118), and [invite](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupInviteMemberSheet.tsx#L120) flows.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupRemoveMemberDialog.spec.tsx
ankita-gupta21: Fixed. Extracted [`MemberOwnershipTransferField`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/MemberOwnershipTransferField.tsx#L31), consumed by [`GroupRemoveMemberDialog.tsx#L93`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupRemoveMemberDialog.tsx#L93) and [`GroupEditMemberSheet.tsx#L160`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupEditMemberSheet.tsx#L160). The dialog's inline string is gone; it reuses the shared sentence builder at [`primaryOwnership.ts#L214-L215`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/primaryOwnership.ts#L214-L215).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupDeleteDialog.tsx
ankita-gupta21: Restored, in the renamed dialog at [`GroupDeleteDialog.tsx#L36-L38`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupDeleteDialog.tsx#L36-L38) and the matching table rationale at [`GroupsTable.tsx#L111`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/GroupsTable.tsx#L111).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMemberActions.ts
ankita-gupta21: Fixed. A single `onSettled` invalidation for the whole sequence: [`useGroupMutations.ts#L156-L160`](https://github.com/gravitee-io/gravitee-api-management/blob/f09307f976/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMutations.ts#L156-L160), asserted as exactly one call.

aqibmohammadkhan (approved):

PR #19291, opened 2026-08-20
fix(ci): restore the S3 upload for es-mapping and jdbc-migration resources

Okhelifi:
## Issue

https://gravitee.atlassian.net/browse/FOUND-183

## What

Restores the `cmd-sync-folder-to-s3` step in `job-backend-build-and-publish-on-download-website`, and scopes the artefact search to the distribution reactor so that only `gravitee-apim-distribution-es-index-mappings` and `gravitee-apim-distribution-jdbc-migrations` are published.

## Why

The job has two halves: shell steps that *stage* zips and checksums into a local `folder_to_sync/`, and one reusable command that *uploads* that folder to S3.

`d9c7d3dbdb` ("chore: remove backend standalone zip and sync on S3", March 2025) legitimately removed the two staging steps for the mAPI and Gateway standalone zips, replaced by the Docker images. But the four entries were contiguous in the same `steps.push(...)`, and the deletion took the fourth with them:

```diff
-        new reusable.ReusedCommand(syncFolderToS3Cmd, {
-          'folder-to-sync': 'folder_to_sync',
-        }),
```

That one was not specific to the standalone zips: it was the shared upload, serving also the generic `publish-folder-path` loop, which survived.

Since then the job stages everything correctly and the container dies with the folder. Nothing fails and the logs look clean. Two things kept it hidden. Lines 80-81 still register the reusable command, so it keeps appearing under `commands:` in the generated YAML and a grep looks reassuring. And the download website does not look empty, because `job-package-bundle` builds its own `folder_to_sync` and calls the sync itself; it is currently the only caller left in the repo.

Confirmed against the bucket:

| Prefix | Newest object | |
|---|---|---|
| `graviteeio-apim/distributions/` | `graviteeio-full-4.12.15.zip`, 2026-08-12 | alive, published by `job-package-bundle` |
| `graviteeio-apim/plugins/repositories/` | `...elasticsearch-4.4.23.zip`, 2025-03-27 | dead since `d9c7d3dbdb` |
| `graviteeio-apim/resources/` | 0 objects | `es-index-mappings` never landed |

No plugin zip exists for 4.8, 4.9, 4.10, 4.11 or 4.12, checked prefix by prefix.

## Scope of the search

The `find` is narrowed to `./gravitee-apim-distribution`. Restoring the upload without narrowing it would silently resume publishing thirteen plugin artefacts (repositories, reporters, endpoints, entrypoints) that have been absent from the download website for a year and a half.

Whether to resume them, or instead to drop the now-inert `publish-folder-path` properties, is a product decision, and one to backport across 4.8.x to 4.12.x. This PR deliberately keeps master's current behaviour and only adds the two distribution archives.

## What will be published

Replayed the job's shell script against a local build:

```
gravitee-apim-distribution-es-index-mappings-4.13.0.zip -> graviteeio-apim/resources/gravitee-apim-distribution-es-index-mappings/
gravitee-apim-jdbc-migrations-4.13.0.zip                -> graviteeio-apim/resources/gravitee-apim-jdbc-migrations/
```

These two and nothing else. Three filters stack up:

- the narrowed `find` returns 2 zips instead of 40;
- `publish-folder-path` is declared by only those two poms in the reactor, and the script reads each module's own `pom.xml`, so nothing inherits it by accident;
- `gravitee-apim-distribution-integration-tests` is profile-gated and produces no zip.

The case worth noting: `target/distribution/plugins/` under the two standalone modules holds 372 plugin zips. They do not match, because the pattern requires the literal `target/gravitee-apim` and here `target/` is followed by `distribution/`. Even if they did, `${pathToArtefactFile%/target*}` would attribute them to the standalone module, whose pom carries no `publish-folder-path`. The guard holds at two levels.

No double upload: nothing persists `folder_to_sync` to the workspace, so `job-package-bundle`'s `attach_workspace` cannot pick up the backend job's folder; the two prefixes are disjoint anyway; and `aws s3 sync` runs without `--delete`, so re-running a release overwrites the same keys rather than accumulating.

The restored step needs no companion. The reusable command already carries the two `keeper/env-export` for the AWS credentials and the `aws-cli/setup`, and the job already runs with `context: cicd-orchestrator`, the same context `job-package-bundle` uses and the one this job already relies on for GPG.

## Tested

- `npx jest`: 153 tests, 29 suites, green
- `npm run validate:test`: 52 configs valid, 0 invalid
- the 7 regenerated fixtures were diffed: they contain the two changes replicated 7 times, with no absorbed drift

Dry-run is safe to try: the sync targets `s3://gravitee-dry-releases-downloads`, and a qualified version goes under `/pre-releases`.

## Note

`npm run build` fails on `circleci-config/config.ts:103` (`lineWidth` missing from `Options`). Verified pre-existing by stashing, unrelated to this change, likely a `yaml` version drift.

Backport policy: none (master only). The published modules only exist from 4.10.x (es-index-mappings) and master (jdbc-migrations). On 4.8.x/4.9.x the narrowed find would match nothing and the sync step would fail on a missing directory, so this must not be cherry-picked there.

### DIFF
--- a/.circleci/ci/src/jobs/backend/job-backend-build-and-publish-on-download-website.ts
+++ b/.circleci/ci/src/jobs/backend/job-backend-build-and-publish-on-download-website.ts
@@ -82,18 +82,20 @@ sed -i "s#<changelist>.*</changelist>#<changelist></changelist>#" gravitee-apim-
 
       steps.push(
         /**
-         * In order to upload repositories, endpoints and entrypoints embedded in APIM mono-repository, we browse for all ZIP files in the project and check if they have a "publish folder path" property in pom.xml.
-         * Because we don't want to publish EVERY plugins (we don't want, apim-services or rest-api-idp-memory for instance), we only rely on this publish-folder-path maven property to determine if a ZIP has to be published or not.
-         * Each plugins is uploaded into a folder based on its name.
+         * We browse the distribution reactor for ZIP files and check if they have a "publish folder path" property in
+         * pom.xml. Because we don't want to publish every artefact, we only rely on that maven property to determine
+         * whether a ZIP has to be published. Each artefact is uploaded into a folder based on its name.
          * Example:
-         *   gravitee-apim-repository-mongodb-x.x.x.zip is published into graviteeio-apim/plugins/repositories/gravitee-apim-repository-mongodb
+         *   gravitee-apim-jdbc-migrations-x.x.x.zip is published into graviteeio-apim/resources/gravitee-apim-jdbc-migrations
          *
-         *
+         * The search is scoped to gravitee-apim-distribution on purpose. Repositories, reporters, endpoints and
+         * entrypoints also carry publish-folder-path, and have not been published since 4.8 — widening the search
+         * here would silently resume publishing thirteen more artefacts, which is a product decision of its own.
          */
         new commands.Run({
           name: 'Prepare plugin zip to upload',
           command: `workingDir=$(pwd)
-for pathToArtefactFile in $(find . -path '*target/gravitee-apim*.zip'); do
+for pathToArtefactFile in $(find ./gravitee-apim-distribution -path '*target/gravitee-apim*.zip'); do
   # Extract folder of the artefact to publish
   # e.g. ./gravitee-apim-repository/gravitee-apim-repository-mongodb/target/gravitee-apim-repository-mongodb-4.4.21.zip => ./gravitee-apim-repository/gravitee-apim-repository-mongodb
   artefactFolder=\${pathToArtefactFile%/target*}
@@ -126,6 +128,9 @@ for pathToArtefactFile in $(find . -path '*target/gravitee-apim*.zip'); do
     cd $workingDir
   fi
 done`,
+        }),
+        new reusable.ReusedCommand(syncFolderToS3Cmd, {
+          'folder-to-sync': 'folder_to_sync',
         }),
       );
     }
--- a/.circleci/ci/src/pipelines/tests/resources/full-release/release-4-2-0-alpha.yml
+++ b/.circleci/ci/src/pipelines/tests/resources/full-release/release-4-2-0-alpha.yml
@@ -510,7 +510,7 @@ jobs:
           name: Prepare plugin zip to upload
           command: |-
             workingDir=$(pwd)
-            for pathToArtefactFile in $(find . -path '*target/gravitee-apim*.zip'); do
+            for pathToArtefactFile in $(find ./gravitee-apim-distribution -path '*target/gravitee-apim*.zip'); do
               # Extract folder of the artefact to publish
               # e.g. ./gravitee-apim-repository/gravitee-apim-repository-mongodb/target/gravitee-apim-repository-mongodb-4.4.21.zip => ./gravitee-apim-repository/gravitee-apim-repository-mongodb
               artefactFolder=${pathToArtefactFile%/target*}
@@ -543,6 +543,8 @@ jobs:
                 cd $workingDir
               fi
             done
+      - cmd-sync-folder-to-s3:
+          folder-to-sync: folder_to_sync
       - persist_to_workspace:
           root: .
           paths:
--- a/.circleci/ci/src/pipelines/tests/resources/full-release/release-4-2-0-dry-run.yml
+++ b/.circleci/ci/src/pipelines/tests/resources/full-release/release-4-2-0-dry-run.yml
@@ -510,7 +510,7 @@ jobs:
           name: Prepare plugin zip to upload
           command: |-
             workingDir=$(pwd)
-            for pathToArtefactFile in $(find . -path '*target/gravitee-apim*.zip'); do
+            for pathToArtefactFile in $(find ./gravitee-apim-distribution -path '*target/gravitee-apim*.zip'); do
               # Extract folder of the artefact to publish
               # e.g. ./gravitee-apim-repository/gravitee-apim-repository-mongodb/target/gravitee-apim-repository-mongodb-4.4.21.zip => ./gravitee-apim-repository/gravitee-apim-repository-mongodb
               artefactFolder=${pathToArtefactFile%/target*}
@@ -543,6 +543,8 @@ jobs:
                 cd $workingDir
               fi
             done
+      - cmd-sync-folder-to-s3:
+          folder-to-sync: folder_to_sync
       - persist_to_workspace:
           root: .
           paths:
--- a/.circleci/ci/src/pipelines/tests/resources/full-release/release-4-2-0-latest.yml
+++ b/.circleci/ci/src/pipelines/tests/resources/full-release/release-4-2-0-latest.yml
@@ -510,7 +510,7 @@ jobs:
           name: Prepare plugin zip to upload
           command: |-
             workingDir=$(pwd)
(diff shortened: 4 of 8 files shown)


.circleci/ci/src/jobs/backend/job-backend-build-and-publish-on-download-website.ts
```diff
@@ -82,18 +82,20 @@ sed -i "s#<changelist>.*</changelist>#<changelist></changelist>#" gravitee-apim-
...
          *
+         * The search is scoped to gravitee-apim-distribution on purpose. Repositories, reporters, endpoints and
+         * entrypoints also carry publish-folder-path, and have not been published since 4.8 — widening the search
+         * here would silently resume publishing thirteen more artefacts, which is a product decision of its own.
          */
         new commands.Run({
           name: 'Prepare plugin zip to upload',
           command: `workingDir=$(pwd)
-for pathToArtefactFile in $(find . -path '*target/gravitee-apim*.zip'); do
+for pathToArtefactFile in $(find ./gravitee-apim-distribution -path '*target/gravitee-apim*.zip'); do
```
ytvnr, line 98: **Worth fixing** - backported to 4.8.x or 4.9.x, this narrowed `find` matches nothing, so `folder_to_sync` is never created and the restored sync step runs `aws s3 sync` on a missing directory.

On those branches `git ls-tree origin/4.8.x -- gravitee-apim-distribution` lists a single `pom.xml`: neither `gravitee-apim-distribution-es-index-mappings` nor `gravitee-apim-distribution-jdbc-migrations` exists before 4.10.x. The un-narrowed `find .` used to match the repository and reporter zips there, so the folder always existed; with this pathspec it does not. On master the es-index-mappings assembly is in the unconditional `<build>`, so the folder is always created and the problem is invisible.

Either make the sync conditional on `folder_to_sync` existing, or state in the PR which branches this is meant to be backported to so the ones that cannot carry it are ruled out explicitly.

ytvnr, line 98: **Worth fixing** - on every currently supported release line the jdbc-migrations zip still will not be published, because the module only exists on master.

`git ls-tree -r --name-only origin/4.1{0,1,2}.x | grep jdbc-migrations` returns nothing: 4.10.x, 4.11.x and 4.12.x carry `gravitee-apim-distribution-es-index-mappings` alone. The fix is right on master, but a customer running with `management.jdbc.liquibase=false` on a supported release gets nothing out of it until the migrations module is itself backported.

Worth saying explicitly in the PR or in FOUND-183, so this CI fix is not read as the artefact becoming available on 4.12.x.

.circleci/ci/src/jobs/backend/job-backend-build-and-publish-on-download-website.ts
```diff
@@ -82,18 +82,20 @@ sed -i "s#<changelist>.*</changelist>#<changelist></changelist>#" gravitee-apim-
...
-         * In order to upload repositories, endpoints and entrypoints embedded in APIM mono-repository, we browse for all ZIP files in the project and check if they have a "publish folder path" property in pom.xml.
-         * Because we don't want to publish EVERY plugins (we don't want, apim-services or rest-api-idp-memory for instance), we only rely on this publish-folder-path maven property to determine if a ZIP has to be published or not.
-         * Each plugins is uploaded into a folder based on its name.
+         * We browse the distribution reactor for ZIP files and check if they have a "publish folder path" property in
+         * pom.xml. Because we don't want to publish every artefact, we only rely on that maven property to determine
+         * whether a ZIP has to be published. Each artefact is uploaded into a folder based on its name.
          * Example:
-         *   gravitee-apim-repository-mongodb-x.x.x.zip is published into graviteeio-apim/plugins/repositories/gravitee-apim-repository-mongodb
-         *
+         *   gravitee-apim-jdbc-migrations-x.x.x.zip is published into graviteeio-apim/resources/gravitee-apim-jdbc-migrations
```
ytvnr, line 89: The example documents a download path that will not exist: the artifactId is `gravitee-apim-distribution-jdbc-migrations`, so the published name carries `-distribution-`.

`gravitee-apim-distribution/gravitee-apim-distribution-jdbc-migrations/pom.xml:30` sets that artifactId and `pom.xml:333` sets `<finalName>${project.artifactId}-${apim.server.version}</finalName>`, so `BASH_REMATCH[1]` yields `gravitee-apim-distribution-jdbc-migrations`. Since this comment is the only place in the repo describing the resulting URL, a support answer built from it would 404.

Side note while you are there: that `finalName` keys the published version off `apim.server.version` rather than the module version, which is only correct because this job passes `-P engine-snapshot`. Might deserve a word next to the example.

ytvnr (approved):
Nice write-up, the description made this reviewable. I checked the claims that were checkable from the repo and they hold up:

- the `*target/gravitee-apim*.zip` pathspec really does exclude the 372 plugin zips under `target/distribution/plugins/` (the standalone assembly uses `<finalName>distribution</finalName>`, so there is no `target/gravitee-apim` literal) and `gravitee-apim-distribution-integration-tests` is property gated;
- both target zips are produced by this job's own build: es-index-mappings from the unconditional `<build>`, jdbc-migrations from the `bundle-default` profile activated by the `-Dbundle` already passed;
- `${pathToArtefactFile%/target*}` still resolves to the module dir with the new search root, and the alpha qualifier survives because it lives in `<sha1>`;
- the restored step sits after both maven builds and before `persist_to_workspace`, which only persists the two `target/distribution` dirs, so no double upload with `job-package-bundle` and the two prefixes stay disjoint;
- credentials are carried by the reusable command itself (two `keeper/env-export` plus `aws-cli/setup`), so nothing is missing at that point;
- the 7 fixtures carry exactly the same two hunks, no absorbed drift, and only the two workflows passing `publishOnDownloadWebsite: true` are affected.

Approving. Three remarks, none blocking, and the two that matter are about the backports rather than about master.

### Notes outside the diff

`.circleci/ci/src/pipelines/tests/pipeline-release.spec.ts:39-41` (untouched, so no inline anchor): the release specs only assert `result.stringify()` against a checked-in YAML file. Deleting `cmd-sync-folder-to-s3` again passes as soon as the fixtures are regenerated, which is exactly the path `d9c7d3dbdb` took in March 2025. One behavioural assertion, that the job contains the sync command when `publishOnDownloadWebsite` is true, would make a second silent removal require deleting a test that names the intent.

.circleci/ci/src/jobs/backend/job-backend-build-and-publish-on-download-website.ts
Okhelifi: Good catch it was a feedback from previous PR I changed the comment.

.circleci/ci/src/jobs/backend/job-backend-build-and-publish-on-download-website.ts
Okhelifi: This fix is master-only and is not meant to be backported: the modules it publishes do not exist before 4.10.x, so there is nothing to publish on older lines. I'm stating it in the PR description. If we ever need the es-mappings zip on a supported line, the backport is safe from 4.10.x onward only.

Okhelifi: Correct, this only restores the publication mechanism. The jdbc-migrations artefact becomes available starting with the release that carries the module (master as of today). Noted in the PR description / FOUND-183.

phiz71 (approved):

PR #19304, opened 2026-08-21
feat(gamma): list all organization groups and bulk-associate group with existing APIs/API Products/Applications

ankita-gupta21:
# FOUND-108 + FOUND-111 — Group associations & organization-wide group listing

Two Gamma platform-module changes on the Groups feature:

- **[FOUND-108](https://gravitee.atlassian.net/browse/FOUND-108)** — bulk-associate a group with existing APIs / API Products / Applications
- **[FOUND-111](https://gravitee.atlassian.net/browse/FOUND-111)** — list all groups in the organization (cross-environment)

**Frontend only — no backend changes.** 24 files, all under `gravitee-gamma/gravitee-gamma-module-platform`.

---

## 1. Alignment with the classic Console

Classic Console has **no organization-wide groups screen**. Its only groups list
(`Settings → Groups`) is environment-scoped, and `GET {org}/groups` is used there
only to populate dropdowns in sharding-tag and identity-provider configuration.
So this page is new surface area — alignment means it reuses Classic's existing API
and repeats its table conventions, adding nothing beyond them.

| Concern | Classic Console | This PR |
| --- | --- | --- |
| Org-wide groups endpoint | `GroupService.listByOrganization()` → `GET {org}/groups` | `listOrganizationGroups()` → same endpoint, **unchanged contract** |
| Permission | `OrganizationGroupsResource` guarded by `ORGANIZATION_TAG READ` | Route guarded by `organization-tag-r`; nav item hidden without it |
| Bulk association | `addToExistingComponents(groupId, type)` → `POST /configuration/groups/{id}/memberships?type=` | Same endpoint and `type` semantics |
| Search semantics | Group name only | Group name only |
| Paginated table + page-size selector | `gio-table-wrapper` | Graphene `DataTable` pagination with the module's shared page-size options |
| Permission-gated navigation | `GioPermissionService` / `*gioPermission` | `useHasPermission` + `PermissionPageGuard`, same hide-then-redirect behaviour |

### Two unavoidable differences

1. **Pagination is client-side.** Classic's env-scoped list pages server-side via
   `/configuration/groups/_paged?page&size&query`. The organization endpoint returns
   the full list and has no paged variant, so search, environment filter and paging are
   applied over the fetched array. Table behaviour is identical to the user; matching
   Classic here would require a new backend endpoint.
2. **Environment column and filter have no Classic counterpart** — they exist only
   because this list spans environments, which Classic's list cannot do. They are
   inherent to the feature, not additions on top of it.

---

### Organization groups (FOUND-111)

- **`OrganizationGroupsPage`** — page shell with loading and error states.
- **`OrganizationGroupsTable`** — Graphene `DataTable` with `Name` and `Environment`
  (as a `Badge`); search field, environment `Select` derived from the returned data,
  `first-use` and `no-results` empty states, and a "Clear filters" action.
- **Data layer** — `OrganizationGroup` type, `listOrganizationGroups()` on
  `apimFetchJsonOrg`, `useOrganizationGroups()` query hook, and a dedicated
  `groupKeys.organizationList()` cache key so the org-wide list invalidates
  independently of env-scoped group queries.
- **Permissions** — `ORGANIZATION_TAG_READ_PERMISSION` constant.
- **Routing / navigation** — `organization-groups` route and label in `config/routes.ts`,
  nav entry under the existing **Team** section, guarded route in `AppRoutes.tsx`.

### Group associations (FOUND-108)

- `associateGroupToExisting()` service and `useAssociateGroupToExisting()` mutation,
  invalidating only the affected membership key.
- `GroupAssociationSection` / `GroupDetailPage` wiring for APIs, API Products
  and Applications.

### Shared-code refactor (no behaviour change)

`useClientSideTableState` and `clientSideTableUtils` (`clampPage`,
`filterClientSideTableItems`, `paginateClientSideTableItems`,
`CLIENT_SIDE_TABLE_DEFAULT_PAGE_SIZE`) were promoted from `features/users/` to
`shared/` so the new table reuses them instead of duplicating the logic. The original
`features/users` paths re-export from `shared/`, so no user-feature call sites changed.

### Tests

304 tests / 24 suites pass; type-check and ESLint/Prettier clean.

- `OrganizationGroupsTable.spec.tsx` — renders name and environment, search by name
  (and *not* by environment name), environment filtering, client-side pagination.
- `OrganizationGroupsPage.spec.tsx` — title, group rendering, error state.
- `AppRoutes.spec.tsx` — route renders with `organization-tag-r`; nav item hidden and
  direct navigation redirected without it.
- `navigation.spec.ts` — Team section contents.
- `useGroupMutations.spec.tsx`, `groups.spec.ts`, `GroupDetailPage.spec.tsx` — association flow.
- JSDOM shims for `hasPointerCapture` / `setPointerCapture` / `releasePointerCapture` /
  `scrollIntoView`, which Graphene's `Select` needs and JSDOM does not implement.

---

## 3. Out of scope — in classic Console, not here

Present in Classic's `Settings → Groups`, intentionally **not** in this PR. FOUND-111
scopes the page to listing, and `GET {org}/groups` is read-only.

| Classic capability | Note |
| --- | --- |
| Create group | `environment-group-c`; env-scoped `POST /configuration/groups` |
| Edit / rename group | `PUT /configuration/groups/{id}` |
| Delete group, with primary-owner groups protected | `environment-group-d`; Classic disables delete for API / API-Product primary-owner groups |
| Member management from the list (add, invite, edit role, remove, "too many users" dialog) | Lives in Classic's group detail screen |
| "Add to new APIs" / "Add to new Applications" event-rule toggles | Per-group `event_rules` (`API_CREATE`, `APPLICATION_CREATE`) |
| `userGroupRequired` console setting on the groups page | Needs `organization-settings-r` and a save bar |
| Member count per group | No Classic equivalent; would require extending `GroupSimpleEntity` server-side |
| Navigating from a row into group detail | Rows are not links; no cross-environment group detail route exists |
| Server-side search and pagination | Needs a paginated `GET {org}/groups` |
| Group roles column (default API / Application / Integration role) | Not requested by FOUND-111 |

---

## Security-sensitive

Touches authorization:

- New route and nav item gated on `organization-tag-r`, matching the
  `ORGANIZATION_TAG READ` check already enforced on `OrganizationGroupsResource`. The UI
  guard is defence in depth, not the only check.
- No API contract, response payload, or permission definition was changed.

[FOUND-108]: https://gravitee.atlassian.net/browse/FOUND-108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FOUND-111]: https://gravitee.atlassian.net/browse/FOUND-111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.ts
```diff
@@ -27,9 +27,14 @@ import type {
     GroupsPagedResponse,
     InviteGroupMemberResult,
     NewGroupPayload,
+    OrganizationGroup,
     UpdateGroupPayload,
 } from '../types/group';
 
+export async function listOrganizationGroups(): Promise<OrganizationGroup[]> {
```
arpzilla-dev: **[Score: 55/100]** 💡 `listOrganizationGroups()` hits the same `GET /groups` org endpoint as the existing `listOrgGroups()` in `features/entrypoints/services/groups.ts`, but under a different React Query key (`groupKeys.organizationList()` vs `org-groups`).

That means two independent fetches and caches for identical server data. Consider consolidating into one shared service/hook (or at least sharing the query key) so sharding-tag screens and this page stay in sync after a refetch.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/queryKeys.ts
```diff
@@ -18,6 +18,7 @@ import type { GroupMembershipType } from '../types/group';
 
 export const groupKeys = {
     all: ['environment-groups'] as const,
+    organizationList: () => [...groupKeys.all, 'organization-list'] as const,
```
arpzilla-dev: **[Score: 25/100]** 🎨 `organizationList()` is nested under `groupKeys.all = ['environment-groups']`, so the cache-key root still reads as environment-scoped.

A dedicated org prefix (e.g. `['organization-groups']`) would make invalidation intent obvious and avoid accidentally treating this list as env-scoped later.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/OrganizationGroupsTable.tsx
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    return (
+        <DataTable
+            aria-label="Organization groups"
+            columns={columns}
+            data={paginatedItems}
+            loading={loading}
+            skeletonCount={pageSize}
+            serverSide
```
arpzilla-dev, line 123: **[Score: 20/100]** 🎨 Sibling tables in this folder (`GroupMembershipTable`, `GroupMembersTable`) comment why `serverSide` is set even though pagination is client-side. This table does the same pattern without that note.

A one-liner here would match the local convention and stop the next reader from treating this as a real server-paged fetch.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/groupPermissions.ts
```diff
@@ -23,6 +23,7 @@ export const ENVIRONMENT_GROUP_DELETE_PERMISSION = 'environment-group-d' as cons
 
 export const ORGANIZATION_SETTINGS_READ_PERMISSION = 'organization-settings-r' as const;
 export const ORGANIZATION_SETTINGS_UPDATE_PERMISSION = 'organization-settings-u' as const;
+export const ORGANIZATION_TAG_READ_PERMISSION = 'organization-tag-r' as const;
```
arpzilla-dev, line 27: **[Score: 22/100]** 🎨 Gating Organization Groups on `organization-tag-r` is correct (it matches `OrganizationGroupsResource` / `ORGANIZATION_TAG READ`), but the name is non-obvious in a groups file.

A one-line comment pointing at that backend permission would save the next reader from assuming this is a copy-paste of the sharding-tags check.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroups.ts
```diff
@@ -17,9 +17,17 @@
 import { useEnvironment } from '@gravitee/gamma-modules-sdk';
 import { keepPreviousData, useQuery } from '@tanstack/react-query';
 
-import { listGroupsPaged } from '../services/groups';
+import { listGroupsPaged, listOrganizationGroups } from '../services/groups';
 import { groupKeys } from '../utils/queryKeys';
 
+export function useOrganizationGroups() {
```
arpzilla-dev: **[Score: 30/100]** 🎨 FYI: this loads the entire organization group list into the browser. There is no paged variant of `GET {org}/groups`, and the PR description already calls this out as an accepted trade-off.

Fine for typical org sizes. Worth revisiting if some tenants grow into thousands of groups — search, environment filter, and paging all run client-side over that full array.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
```diff
@@ -210,6 +220,24 @@ export function GroupDetailPage() {
         startMemberTabTransition(() => setMemberTab(value as 'members' | 'invitations'));
     }
 
+    async function handleAssociate() {
```
arpzilla-dev, line 284: **[Score: 28/100]** 🎨 FYI: this associates the group with *every* existing resource of that type in the environment, with only a generic confirm and no count/preview of affected APIs, API Products, or applications.

That matches Classic Console, so it is not a regression. Still a high-blast-radius mutation — a count in the dialog (or type-to-confirm) would make the impact clearer without changing the API contract.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
```diff
@@ -526,6 +560,19 @@ export function GroupDetailPage() {
                 isPending={deleteInvitationMutation.isPending}
                 onConfirm={handleDeleteInvitation}
             />
+
+            {associatingType ? (
+                <ConfirmDialog
```
aqibmohammadkhan, line 605: Confirm dialog doesn't disable Continue on permission loss

The "Add group to existing X" ConfirmDialog never passes confirmDisabled, so its Continue button stays clickable even after canAssociateGroup becomes false while the dialog is open. If permission is revoked or group.manageable flips to false via a background refetch while the dialog is open, clicking Continue silently no-ops (handleAssociate returns early) — no toast, no dialog close, no error shown.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
```diff
@@ -159,6 +167,8 @@ export function GroupDetailPage() {
 
     const updateMutation = useUpdateGroup();
     const deleteMutation = useDeleteGroup();
+    const associateMutation = useAssociateGroupToExisting();
+    const canAssociateGroup = Boolean(group && canEdit && group.manageable !== false);
```
aqibmohammadkhan: canAssociateGroup diverges from canInviteToGroup on undefined manageable

canAssociateGroup = Boolean(group && canEdit && group.manageable !== false) treats a missing/undefined manageable as permitted. The existing canInviteToGroup helper (groupPermissions.ts) treats the same undefined value as not permitted (Boolean(group.manageable)). Same field, same file, opposite default — an inconsistent authorization rule computed inline instead of added to the shared permission predicates.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useClientSideTableState.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export function useClientSideTableState<T extends object>(
+    items: readonly T[],
+    searchIgnoreKeys: readonly string[],
+    options: UseClientSideTableStateOptions = {},
+) {
+    const [search, setSearch] = useState('');
+    const [page, setPage] = useState(1);
+    const [pageSize, setPageSize] = useState(CLIENT_SIDE_TABLE_DEFAULT_PAGE_SIZE);
+    const deferredSearch = useDeferredValue(search);
```
aqibmohammadkhan, line 41: useDeferredValue desyncs search flag from filtered rows

search/hasActiveSearch update synchronously on each keystroke, but filteredItems/totalCount/paginatedItems are computed from useDeferredValue(search) — one render behind. Affects OrganizationGroupsTable, UserPersonalAccessTokensCard, UserEnvironmentRolesCard, UserInheritedPermissionsSection: a keystroke can show "search active" UI paired with stale rows/count for a frame. The carried-over spec drives handleSearchChange inside a synchronous, non-awaited act(), which isn't guaranteed to flush a deferred-value render — risk of intermittent test failures.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/OrganizationGroupsTable.tsx
```diff
@@ -0,0 +1,162 @@
...
+];
+
+interface OrganizationGroupsTableProps {
+    readonly groups: OrganizationGroup[];
+    readonly loading: boolean;
+}
+
+export function OrganizationGroupsTable({ groups, loading }: OrganizationGroupsTableProps) {
+    const searchInputId = useId();
+    const [environmentId, setEnvironmentId] = useState(ALL_ENVIRONMENTS);
```
aqibmohammadkhan, line 66: Stale environment filter can silently resurrect

Raw environmentId state is never cleared when the selected environment temporarily disappears from groups — only the derived selectedEnvironmentId falls back to "all" for rendering. If a later refetch brings a group in that environment back, the filter re-activates on its own, even though the Select showed "All environments" in between and the user never re-selected anything.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroups.ts
```diff
@@ -17,9 +17,17 @@
 import { useEnvironment } from '@gravitee/gamma-modules-sdk';
 import { keepPreviousData, useQuery } from '@tanstack/react-query';
 
-import { listGroupsPaged } from '../services/groups';
+import { listGroupsPaged, listOrganizationGroups } from '../services/groups';
```
aqibmohammadkhan: Full org group list rescanned on every keystroke

useOrganizationGroups fetches the entire org-wide group list unpaginated (no backend paged variant), and every keystroke re-scans all of it via filterClientSideTableItems's per-item Object.entries pass to render one page of rows. For an org with thousands of groups, this is an O(n × fields) scan on every character typed.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/OrganizationGroupsTable.tsx
```diff
@@ -0,0 +1,162 @@
...
+    const environmentOptions = useMemo(
```
aqibmohammadkhan, line 67: Environment dropdown rebuilt on every background refetch

environmentOptions rebuilds a Map and re-sorts it from the full, unfiltered groups array on every groups reference change, including 30s-staleTime background refetches that almost never change the actual environment set.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
```diff
@@ -210,6 +220,24 @@ export function GroupDetailPage() {
...
+        try {
+            await associateMutation.mutateAsync({ groupId: group.id, type: associationType });
+            notify.success(`Successfully added the group to existing ${typeLabel}.`);
+            setAssociatingType(null);
+        } catch (error) {
+            notify.error(error, `Error occurred while adding the group to existing ${typeLabel}.`);
+        }
+    }
+
+    function handleAssociateDialogOpenChange(isOpen: boolean) {
```
aqibmohammadkhan, line 299: Third copy-pasted confirm-mutate-toast dialog in one file

The new associate flow (associatingType state, handleAssociate, handleAssociateDialogOpenChange, inline ConfirmDialog) re-implements the same confirm→mutate→toast→close pattern already used twice in this file (delete group, delete invitation) instead of extracting a shared hook. A future fix has to be hand-applied three times.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
```diff
@@ -389,6 +417,8 @@ export function GroupDetailPage() {
                     ariaLabel="APIs"
                     searchPlaceholder="Search APIs…"
                     emptyTitle="No dependent APIs to display"
+                    actionLabel={`Add group to existing ${ASSOCIATION_TYPE_LABELS.api}`}
+                    onAction={canAssociateGroup ? () => setAssociatingType('api') : undefined}
                 />
 
                 <GroupAssociationSection
```
aqibmohammadkhan: Three near-identical GroupAssociationSection blocks

The three call sites (APIs, API Products, Applications) repeat the same actionLabel/onAction shape almost verbatim, varying only by membership type — ~40 lines that could be a single array-driven render.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
```diff
@@ -210,6 +220,24 @@ export function GroupDetailPage() {
         startMemberTabTransition(() => setMemberTab(value as 'members' | 'invitations'));
     }
 
+    async function handleAssociate() {
+        if (!group || !associatingType || !canAssociateGroup) return;
+        const associationType = associatingType;
+        const typeLabel = ASSOCIATION_TYPE_LABELS[associationType];
+        try {
+            await associateMutation.mutateAsync({ groupId: group.id, type: associationType });
+            notify.success(`Successfully added the group to existing ${typeLabel}.`);
```
aqibmohammadkhan: Association dialog copy re-interpolated in 7 places

ASSOCIATION_TYPE_LABELS[type] is inlined into "Add group to existing X" / "all the existing X" phrasing at 7 separate call sites with no single source of copy. Already slightly inconsistent — the success message has a trailing period the action labels and dialog title/description lack.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/types/organizationGroup.ts
aqibmohammadkhan, line 1: Old hook/util files left as unmigrated re-export shims

Reduced to pass-through re-exports of the new shared/ implementation, but the 5 existing consumers (UserEnvironmentRolesCard, UserPersonalAccessTokensCard, UserInheritedPermissionsSection, UserGroupMembershipsCard, TenantsTable) still import the old path instead of shared/ directly — two import paths for identical logic.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/services/groups.ts
ankita-gupta21: Fixed. `listOrgGroups()` / `org-groups` and `listOrganizationGroups()` / `groupKeys.organizationList()` are now one shared service, hook, and query key.

Sharding-tag screens and the organization groups page both use `useOrganizationGroups()` (`queryKey: organizationGroupKeys.list()`). Create / update / delete group mutations also invalidate `organizationGroupKeys.all`, so both surfaces stay in sync.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useOrganizationGroups.ts#L21-L26

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/entrypoints/hooks/useShardingTags.ts#L39

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroupMutations.ts#L69-L81

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/OrganizationGroupsTable.tsx
ankita-gupta21: Fixed. Added the same kind of comment as the sibling tables: `serverSide` only tells DataTable not to re-paginate; this table still slices the unpaginated organization response client-side.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/OrganizationGroupsTable.tsx#L122-L123

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/queryKeys.ts
ankita-gupta21: Fixed. The org-wide list is no longer nested under `groupKeys.all` (`environment-groups`). It now uses a dedicated `organization-groups` prefix.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/utils/queryKeys.ts#L27-L30

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/groupPermissions.ts
ankita-gupta21: Fixed. Added a one-line comment that this matches `OrganizationGroupsResource` / `ORGANIZATION_TAG READ`, not a copy-paste of the sharding-tags check.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/groupPermissions.ts#L26-L27

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroups.ts
ankita-gupta21: Acknowledged — no change to the API contract. `GET {org}/groups` has no paged variant, so this page still loads the full list and applies search / environment filter / paging client-side.

That matches Classic Console’s existing endpoint. Search now uses a name-only predicate plus deferred filtering so we do not rescan every field on each keystroke. A paged org endpoint would be a follow-up if tenants grow into thousands of groups.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useOrganizationGroups.ts#L21-L26

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/OrganizationGroupsTable.tsx#L40-L41

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
ankita-gupta21: Acknowledged — left as-is for Classic Console parity. 

**Classic Console:**
same API (POST .../groups/{id}/memberships?type=api)
same meaning: every existing resource of that type
same generic confirm, no count, no type-to-confirm
If Gamma showed “This will update 47 APIs” or asked you to type the group name, that would be new UX Classic does not have.

**Also, a count is not free:**
The associated-APIs table is already associated APIs, not “how many APIs exist in the env.”
To show “47 APIs will be affected” you would need another API (list all APIs in the env). Classic does not do that on this dialog.
So we left the dialog as Classic: generic wording, no extra number, no type-to-confirm.



https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx#L605-L616

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
ankita-gupta21: Fixed. The associate ConfirmDialog now passes `confirmDisabled={!canAssociateGroup}`, so Continue is disabled if update permission or `manageable` is lost while the dialog is open. The handler still guards the mutation as well.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx#L614

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
ankita-gupta21: Fixed. Association now uses a shared `canAssociateGroupToExisting()` predicate that fails closed on missing / undefined `manageable` (`Boolean(group.manageable)`), same rule as `canInviteToGroup`. The page no longer treats undefined as permitted.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/utils/groupPermissions.ts#L37-L39

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx#L233

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useClientSideTableState.ts
ankita-gupta21: Fixed. `hasActiveSearch`, `filteredItems`, and `paginatedItems` now all derive from `deferredSearch`, so the “search active” UI no longer leads the rows/count by a frame. Specs wait for the deferred render instead of asserting inside a synchronous `act()`.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useClientSideTableState.ts#L49-L63

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/OrganizationGroupsTable.tsx
ankita-gupta21: Fixed. When the selected environment disappears from the list, we now write the fallback (`all`) back into `environmentId` instead of only using it for rendering. A later refetch that brings that environment back no longer re-applies the old filter.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/OrganizationGroupsTable.tsx#L86-L90

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/hooks/useGroups.ts
ankita-gupta21: Fixed as far as we can without a paged backend. Search now uses a name-only `matchesSearch` predicate instead of `Object.entries` over every field, and filtering still runs on `useDeferredValue` so typing does not block the UI. The unpaginated `GET {org}/groups` contract is unchanged.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/OrganizationGroupsTable.tsx#L40-L41

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useClientSideTableState.ts#L49-L56

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/OrganizationGroupsTable.tsx
ankita-gupta21: Left as-is. `environmentOptions` is already memoized on `groups`. React Query structural sharing keeps that reference stable when a 30s refetch returns the same data, so we do not rebuild the Map/sort unless the list actually changed. Adding a second fingerprint cache would duplicate that without a behaviour change.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/groups/components/OrganizationGroupsTable.tsx#L67-L73

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
ankita-gupta21: Not extracted in this PR. Delete group, delete invitation, and associate look similar at the toast layer, but they differ in navigation (leave the page vs close dialog), pending-state ownership, and authorization. A shared hook would either leak those details or hide them behind options that still need per-call-site wiring.

The associate flow stays local; the copy/guards for that flow are now centralized.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx#L285-L298

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
ankita-gupta21: Fixed. APIs / API Products / Applications now come from one `associationSections` config and a single `.map()` render.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx#L475-L487

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/types/organizationGroup.ts
ankita-gupta21: Fixed. Removed the `features/users` re-export shims. User cards and TenantsTable now import `useClientSideTableState` / `clientSideTableUtils` from `shared/` directly.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/users/components/UserEnvironmentRolesCard.tsx#L34

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/tenants/components/TenantsTable.tsx#L35

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
ankita-gupta21: Fixed. APIs / API Products / Applications now come from one `associationSections` config and a single `.map()` render.

https://github.com/gravitee-io/gravitee-api-management/blob/457b6b8c4d39a93bd7d0d6b2ef35c8b1b6bf5ff9/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx#L475-L487

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/GroupDetailPage.tsx
ankita-gupta21: I did not ignore the duplication entirely. I centralized the associate flow’s own copy and guards (ASSOCIATION_COPY, canAssociateGroupToExisting, confirmDisabled). I just did not invent a third abstraction that all three dialogs would share.

**1. What happens after success is different**
- Delete group: leave the page (navigate('..')).
- Delete invitation: stay, just close the dialog.
- Associate: stay, close dialog, refresh that section’s list.

**_A shared hook would need extra flags like afterSuccess: 'navigate' | 'close'. That is still per-dialog wiring._**

**2. Who owns “busy” / pending is different**
- Delete group uses deleteMutation.isPending.
- Delete invitation uses deleteInvitationMutation.isPending from the member-actions hook.
- Associate uses associateMutation.isPending.

**_A shared hook cannot own those without taking the mutation as an argument — then the page still has to pass it in._**

**3. Authorization is different**
- Delete group: canDelete.
- Delete invitation: member-management permission.
- Associate: canEdit and group.manageable.

**_If you hide that in a generic hook, the next person still has to pass confirmDisabled / guards in. The “shared” part becomes a thin wrapper._**

aqibmohammadkhan (approved):

arpzilla-dev:
1: @ankita-gupta21  Good for my review point, confirm with Aqib's review comment

PR #19306, opened 2026-08-21
feat(analytics): expose authz decisions through the analytics engine

podlesrafal:
## Issue

https://gravitee.atlassian.net/browse/AUTHZ-96

## Description

Adds an authz family to the analytics engine so Gamma modules can build dashboards on authorization decisions. Pairs with https://github.com/gravitee-io/gravitee-gamma-module-authz/pull/130, which is the dashboard itself.

Decisions were already reported to the `gravitee-event-metrics` data stream with `doc-type: "authz"` and every field was already indexed — what was missing was vocabulary. This adds a field resolver plus measures/facets/time-series adapters modelled on the HTTP family, eight metrics in the analytics catalog, and the matching constants in the v2 OpenAPI spec. **No new endpoint:** the dashboard runs on the existing `/observability/analytics/{measures,facets,time-series}`, the same ones aim uses. No reporter change.

## Additional context

**One call for the reviewer.** `ObservabilityAnalyticsDataPortAdapter.toGammaApiType` mapped v4 `AUTHZ -> null`, so AUTHZ-typed APIs fell out of the accessible-API scope and their decisions were silently never counted — 35 of 43 in the dev dataset. This PR maps it to the existing `ApiType.AUTHZ`. Permission checks are untouched; the fix stops discarding APIs that already passed them. If AUTHZ-typed APIs were meant to be invisible in observability, this is the line to object to.

**Compatibility-sensitive:** additive constants on five enums in `openapi-analytics.yaml`. MapStruct joins the core and generated enums by name, so a core constant without a spec counterpart fails the build.

**Verified live**, not by curl: dashboard values match raw Elasticsearch aggregations over the same window — Forbid 416, Searches 722, Failed 133, Not applicable 299 — and the Decisions screen agrees at 416.


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
```diff
@@ -1534,6 +1534,227 @@ spec:
               - PLAN
               - NATIVE_OPERATION
 
+        # Authorization decision events, from the authz event-metrics doc-type (see AuthzFieldResolver).
+        - name: AUTHZ_DECISIONS
+          label: Decisions
+          apis:
+              - AUTHZ
```
sclarkamlab, line 1543: All eight new metrics declare `apis: [AUTHZ]` but the `spec.apis` list at the top of the file still stops at `EDGE`. `GetApiSpecUseCase` hands back `spec.apis()` as is, so `GET /analytics/definition/apis` never mentions AUTHZ. Needs an `- name: AUTHZ` entry up there with a label.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
```diff
@@ -2006,6 +2227,62 @@ spec:
           operators:
               - EQ
               - IN
+
+        - name: AUTHZ_DECISION
+          label: Decision
+          type: KEYWORD
```
sclarkamlab: All eight authz filters are KEYWORD, and the openapi says KEYWORD filters return distinct values from the data store. But `resolveEsFieldName` in `FilterValuesQueryServiceImpl` has no case for any `AUTHZ_` name, so it falls into `default -> throw new UnsupportedOperationException` and the picker gets a 500 the moment someone opens it. `searchFilterValues` also reads `V4_METRICS` while these docs are in `EVENT_METRICS`, so adding the field mapping on its own won't be enough. Same gap already exists for `NATIVE_TOPIC` and friends so it's not something you introduced, but this adds eight more of them. Either wire up the mapping and the index, or make them STRING like the gamma catalog does for subject/action/resource.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/use_case/SearchObservabilityLogsUseCase.java
```diff
@@ -184,13 +184,13 @@ private static RecordType extractRecordType(List<FilterCondition> conditions) {
...
     // decisions screen but missing here is advertised and then refused with a 400.
     static final Set<String> DECISION_SUPPORTED_FILTERS = Set.of(
-        StaticFilters.DECISION.filterName(),
-        StaticFilters.SUBJECT.filterName(),
-        StaticFilters.ACTION.filterName(),
-        StaticFilters.RESOURCE.filterName(),
-        StaticFilters.CALLER_KIND.filterName(),
-        StaticFilters.STATUS.filterName(),
-        StaticFilters.OPERATION.filterName(),
+        StaticFilters.AUTHZ_DECISION.filterName(),
```
sclarkamlab, line 187: The rename didn't reach the error text. The `ValidationDomainException` at line 217 still tells people to send `DECISION`, `SUBJECT`, `ACTION`, `RESOURCE`, `CALLER_KIND`, `STATUS`, `OPERATION`, and `openapi-observability.yaml` lines 598 to 599 document the same old vocabulary. So someone reads the 400, sends `DECISION`, and gets the same 400 back.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
```diff
@@ -1534,6 +1534,227 @@ spec:
...
+              - GATEWAY
+              - AUTHZ_ACTION
+              - AUTHZ_SUBJECT_ID
+              - AUTHZ_RESOURCE_ID
+              - AUTHZ_REASON
+              - AUTHZ_CALLER
+          filters:
+              - API
+              - GATEWAY
+              - AUTHZ_DECISION
```
sclarkamlab: Your new test in `AnalyticsDefinitionYAMLQueryServiceTest` says a search document carries no decision, which is why the decision facet came off searches. Same logic kills the filter, it'll always match nothing. I'd drop it from the filters list too.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/analytics_engine/use_case/GetAnalyticsFilterDefinitionsUseCaseTest.java
```diff
@@ -52,6 +52,20 @@ void should_narrow_to_the_requested_signal() {
         assertThat(specs).hasSizeLessThan(catalog.getAllFilters().size());
     }
 
+    // TODO(authz): apis stays empty until an authz metric lists AUTHZ_REASON under filters:.
```
sclarkamlab: This TODO says apis stays empty until an authz metric lists `AUTHZ_REASON` under `filters:`, but the same PR does list it, so the test passes and the comment is already out of date. Delete it.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
podlesrafal: Fixed. Added `- name: AUTHZ` / `label: Authz` to `spec.apis`.

Verified live rather than by inspection: `GET /analytics/definition/apis` now returns 200 with `{"name":"AUTHZ","label":"Authz"}` as the last entry.

Also added a regression test in `AnalyticsDefinitionYAMLQueryServiceTest` that walks every `ApiSpec.Name` and fails if a kind declared by any metric is missing from `getApis()`. It is generic, so the next api kind cannot drift the same way. Confirmed it fails without the fix.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
podlesrafal: Good catch, and it was worse than "the picker 500s": all eight were 500, not 400, because `resolveEsFieldName` throws `UnsupportedOperationException` while the gamma adapter only catches `ValidationDomainException`.

Rather than wiring the ES mapping, I fixed the typing, because `KEYWORD` was the wrong call in the first place:

- `AUTHZ_DECISION`, `AUTHZ_OPERATION`, `AUTHZ_STATUS`, `AUTHZ_CALLER` are closed sets, so they are now `ENUM` with `enumValues` taken from the wire values the reporter writes. The picker no longer needs Elasticsearch at all, and values show even when a bucket is empty.
- `AUTHZ_SUBJECT_ID`, `AUTHZ_ACTION`, `AUTHZ_RESOURCE_ID`, `AUTHZ_REASON` are open sets, so `STRING`. `reasons` are full sentences (`["Permitted by policy x"]`), so a distinct-value picker over them is a second view of the data, not a picker.

This also repairs the error contract for free: `STRING` makes `GetFilterValuesUseCase` throw `ValidationDomainException`, which the gamma adapter already converts to a clean 400.

Verified live on a running stack:

| filter | before | after |
|---|---|---|
| the four ENUM ones | 500 | **200** (`AUTHZ_DECISION` returns PERMIT/FORBID/NOT_APPLICABLE) |
| the four STRING ones | 500 | **400** |
| `NATIVE_TOPIC` | 500 | 500 (untouched) |

`enumValues` is only read for listing (`GetFilterValuesUseCase:146`), never to validate incoming conditions, so no new rejection path.

On the shared gap you mentioned: `NATIVE_CLIENT_ID`, `NATIVE_TOPIC` and `NATIVE_OPERATION` stay broken and I did not hide that. `FilterValuesQueryServiceImplTest` now walks every `KEYWORD` filter in the catalog through the real `searchFilterValues` and fails on any without a mapping, with those three named in an explicit `knownGaps` set that is only allowed to shrink. The proper fix is routing filter-values through `FieldResolver` and deleting `resolveEsFieldName` (which duplicates `HTTPFieldResolver` in the wrong layer, and is also why the index is pinned to `V4_METRICS`), but that is a shared-engine change and belongs in its own ticket.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/use_case/SearchObservabilityLogsUseCase.java
podlesrafal: Fixed, and derived rather than restated so it cannot drift again:

```java
"Filters " + unsupported + " do not apply to RECORD_TYPE=AUTHZ_DECISION. Supported: " +
    Stream.concat(Stream.of("API", "API_TYPE"), DECISION_SUPPORTED_FILTERS.stream().sorted())
        .collect(Collectors.joining(", ")) + " and the time range."
```

`API` and `API_TYPE` are prepended because `stripScopeConditions` removes them before this check runs.

`openapi-observability.yaml` updated too.

Chasing this exposed something bigger, which I fixed in the same pass: the PR renamed seven filters in place but `REASON` could not follow, because `FilterSpec` carries one operator list for all signals and `REASON` needs `CONTAINS` on logs while analytics has none. So a second `AUTHZ_REASON` was added under the same label, and the Decisions picker listed "Reason" twice, distinguishable only by a signal icon. Picking the analytics one built a filter the frontend then dropped before sending, leaving an active-looking chip that filtered nothing.

I confirmed the constraint instead of trusting the comment: temporarily allowing `CONTAINS` on the analytics path returns **500 `Unexpected enum constant: CONTAINS`**, matching `@ValueMapping(source = "CONTAINS", target = THROW_EXCEPTION)` in `AnalyticsMeasuresAdapter`. `CONTAINS` is used by exactly one other filter in the whole catalog, `PAYLOAD`, which is also logs-only.

So the duplicate is gone and the vocabulary is uniform: every decision filter now carries the `AUTHZ_` prefix, including the logs-only ones (`AUTHZ_PDP`, `AUTHZ_MATCHED_POLICY`, `AUTHZ_POLICY_VERSION`, `AUTHZ_REASON`), and `AUTHZ_REASON` is a single logs-only `CONTAINS` filter shaped exactly like `PAYLOAD`. The dashboard is unaffected because it reads reasons as a facet, which needs no entry in the filter catalog. `SpiFilterRegistryTest` now fails if two filters share a label within one api kind.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
podlesrafal: Right, dropped it from the `filters` list of `AUTHZ_SEARCHES`. Facets and filters now agree about a field the search document does not carry.

Guarded by a sibling of the existing facet test in `AnalyticsDefinitionYAMLQueryServiceTest`:

```java
@Test
void should_not_offer_decision_as_a_filter_of_searches() {
    var filterNames = service.getFilters(MetricSpec.Name.AUTHZ_SEARCHES).stream().map(FilterSpec::name).toList();
    assertThat(filterNames).doesNotContain(FilterSpec.Name.AUTHZ_DECISION);
}
```

Confirmed it fails when the entry is put back.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/analytics_engine/use_case/GetAnalyticsFilterDefinitionsUseCaseTest.java
podlesrafal: Deleted.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/analytics/use_case/AnalyticsRequestPipeline.java
```diff
@@ -123,4 +127,15 @@ private static List<FilterCondition> applyDefaultEntrypointScoping(List<FilterCo
         result.add(new FilterCondition("ENTRYPOINT", FilterOperator.IN, List.copyOf(DEFAULT_ENTRYPOINT_IDS)));
         return List.copyOf(result);
     }
+
+    private static boolean isAuthzDecisionRequest(List<FilterCondition> conditions) {
+        return conditions.stream().anyMatch(c -> "RECORD_TYPE".equals(c.name()) && c.values().contains("AUTHZ_DECISION"));
```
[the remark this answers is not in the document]
podlesrafal: Not taking this one, because the suggested patch would break the feature.

It pins the check to `FilterOperator.IN`, but every real caller sends `RECORD_TYPE` with `EQ`: `AnalyticsRequestPipelineTest`, `SearchObservabilityLogsUseCaseTest`, `ObservabilityFilterValidatorRecordTypeAnalyticsScopeTest` and the openapi description all use `EQ`, and the docs state the filter takes `EQ`/`IN`. Applying it would stop routing authz analytics requests entirely.

The underlying point about ignoring the operator is not empty, since `RECORD_TYPE` lives in `ExtensibleFilters`, which declares no operator list. But the scenario is unreachable: `NOT_IN` is deliberately absent from the whole catalog (`StaticFilters`: "NOT_IN is intentionally absent until a translator supports it"), so no filter advertises it. If we ever want to close it, the correct condition is `EQ || IN`, not `IN` alone.

podlesrafal:
Review addressed, force-pushed as a single commit. Summary of what changed and how it was verified.

**The five review points**

1. `AUTHZ` added to `spec.apis`, plus a generic regression test so no future api kind can drift the same way.
2. The eight `AUTHZ_*` filters retyped: four `ENUM` (closed sets, picker served from the catalog), four `STRING` (open sets, honest 400). This turns eight 500s into working pickers or clean 400s, and fixes the error contract rather than papering over it.
3. The 400 message is now derived from `DECISION_SUPPORTED_FILTERS` instead of restating it, and `openapi-observability.yaml` was updated.
4. `AUTHZ_DECISION` dropped from the `filters` of `AUTHZ_SEARCHES`, guarded by a test.
5. Stale TODO deleted.

**One thing beyond the review**

The rename had left `Reason` duplicated: seven filters were renamed in place, but `REASON` could not follow because `FilterSpec` carries one operator list per filter and analytics has no `CONTAINS`, so a second `AUTHZ_REASON` was added under the same label. The Decisions picker listed "Reason" twice and picking the analytics one produced a chip that filtered nothing.

I checked the constraint instead of trusting the comment: allowing `CONTAINS` on the analytics path returns 500 `Unexpected enum constant: CONTAINS`. So the duplicate is gone and the vocabulary is now uniform, with every decision filter carrying the `AUTHZ_` prefix, including the logs-only ones. `AUTHZ_REASON` is a single logs-only `CONTAINS` filter, shaped like `PAYLOAD`, the only other `CONTAINS` filter in the catalog. The dashboard is unaffected because it reads reasons as a facet.

**Verification**

Not just green tests. I ran the stack, seeded six authz decisions with a known distribution, and compared what the API returns against what is actually in Elasticsearch.

Decisions, all thirteen filters exact: baseline 6, `AUTHZ_DECISION=PERMIT` 3, `FORBID` 1, `AUTHZ_SUBJECT_ID=alice` 3, `AUTHZ_ACTION=read` 4, `AUTHZ_RESOURCE_ID=doc-1` 3, `AUTHZ_CALLER=pep` 4, `AUTHZ_STATUS=error` 1, `AUTHZ_OPERATION=search` 1, `AUTHZ_PDP=tagged` 1, `AUTHZ_MATCHED_POLICY=alpha` 3, `AUTHZ_POLICY_VERSION=2` 2, `AUTHZ_REASON CONTAINS Forbidden` 1, `REQUEST_ID` 1.

Dashboard metrics exact too: `AUTHZ_DECISIONS` 5, `AUTHZ_PERMITS` 3, `AUTHZ_FORBIDS` 1, `AUTHZ_NOT_APPLICABLE` 1, `AUTHZ_SEARCHES` 1, `AUTHZ_FAILURES` 1, plus the seven shared filters applied to `AUTHZ_DECISIONS`. The rendered dashboard matched, including Top Reasons, which confirms the facet path survived the filter change. Seeded documents were removed afterwards.

Suites: rest-api-service 9367, gamma 385, no failures.

**Left deliberately undone**

`NATIVE_CLIENT_ID`, `NATIVE_TOPIC` and `NATIVE_OPERATION` still 500 on value listing. They are now named in an explicit `knownGaps` set in `FilterValuesQueryServiceImplTest` that is only allowed to shrink, instead of being invisible. The real fix is routing filter-values through `FieldResolver` and deleting `resolveEsFieldName`, which is a shared-engine change and deserves its own ticket.

Separately, `LogsExplorerPage` in `gamma-lib-observability` passes `visibleSignals={ACTIVE_SIGNALS}` unconditionally and renders the desync banner even when `widget: false`, so a logs-only screen still shows a banner about a histogram that is not there. Worth a ticket on the library; nothing in this PR can reach it.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
podlesrafal: Correction to my earlier reply on this thread: I have narrowed the guard and dropped the `NATIVE_*` part.

The test I described walked the whole catalog and carried a `knownGaps` set naming `NATIVE_CLIENT_ID`, `NATIVE_TOPIC` and `NATIVE_OPERATION`. That was overreach. Those filters are not ours, and the test would have gone red on somebody else`s PR with a list this PR happens to own.

It was also pointless for us in the end: after the retyping there is no `KEYWORD` authz filter left, so the whole-catalog version protected everyone except the code it shipped with.

Replaced with an authz-scoped property in `AnalyticsDefinitionYAMLQueryServiceTest`:

```java
@Test
void should_not_type_any_authz_filter_as_keyword() {
    var keywordAuthzFilters = service.getAllFilters().stream()
        .filter(spec -> spec.name().name().startsWith("AUTHZ_"))
        .filter(spec -> spec.type() == FilterType.KEYWORD)
        .map(FilterSpec::name)
        .toList();

    assertThat(keywordAuthzFilters).isEmpty();
}
```

Confirmed it fails when `AUTHZ_SUBJECT_ID` is put back to `KEYWORD`: `Expecting empty but was: [AUTHZ_SUBJECT_ID]`.

The `NATIVE_*` value listing is still broken on master, exactly as it was before this PR. It just is not this PR`s business to track it.

sclarkamlab (approved):

PR #19318, opened 2026-08-21
feat: manage API Products in Portal categories

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-131

## Description

Adds API Product assignment management to the existing Console Portal category editor.

This PR:

- adds a separate API Products section,
- lists API Products assigned to the category, including unpublished assignments,
- allows searching for and adding published, unassigned API Products,
- preserves existing category assignments when adding a product,
- allows removing the current category assignment with confirmation,
- applies the existing documentation permissions,
- provides success and error notifications.

No HTTP contract changes are introduced.

## Dependency

This is a stacked PR based on `PORTAL-130-backend-support-api-product-navigation-categories`.

https://github.com/user-attachments/assets/4ec42117-92b5-40bf-98dd-c9af952e3f71

### DIFF
--- a/gravitee-apim-console-webui/src/portal/catalog/category/add-api-product-to-category-dialog/add-api-product-to-category-dialog.component.ts
+++ b/gravitee-apim-console-webui/src/portal/catalog/category/add-api-product-to-category-dialog/add-api-product-to-category-dialog.component.ts
@@ -1,0 +1,96 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Component, computed, inject } from '@angular/core';
+import { FormControl, FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { toSignal } from '@angular/core/rxjs-interop';
+import { MAT_DIALOG_DATA, MatDialogModule, MatDialogRef } from '@angular/material/dialog';
+import { MatAutocompleteModule } from '@angular/material/autocomplete';
+import { MatButtonModule } from '@angular/material/button';
+import { MatFormFieldModule } from '@angular/material/form-field';
+import { MatIconModule } from '@angular/material/icon';
+import { MatInputModule } from '@angular/material/input';
+
+import { PortalNavigationApiProduct } from '../../../../entities/management-api-v2';
+
+export interface ApiProductCategoryCandidate {
+  navigationItem: PortalNavigationApiProduct;
+  name: string;
+  version: string;
+}
+
+export interface AddApiProductToCategoryDialogData {
+  title: string;
+  candidates: ApiProductCategoryCandidate[];
+}
+
+export type AddApiProductToCategoryDialogResult = ApiProductCategoryCandidate;
+
+@Component({
+  selector: 'add-api-product-to-category-dialog',
+  templateUrl: './add-api-product-to-category-dialog.component.html',
+  styleUrl: './add-api-product-to-category-dialog.component.scss',
+  imports: [
+    FormsModule,
+    ReactiveFormsModule,
+    MatAutocompleteModule,
+    MatButtonModule,
+    MatDialogModule,
+    MatFormFieldModule,
+    MatIconModule,
+    MatInputModule,
+  ],
+})
+export class AddApiProductToCategoryDialogComponent {
+  private readonly dialogRef = inject(MatDialogRef<AddApiProductToCategoryDialogComponent, AddApiProductToCategoryDialogResult>);
+  protected readonly data = inject<AddApiProductToCategoryDialogData>(MAT_DIALOG_DATA);
+
+  protected readonly searchApiProductControl = new FormControl<string | ApiProductCategoryCandidate>('', { nonNullable: true });
+
+  private readonly searchValue = toSignal(this.searchApiProductControl.valueChanges, {
+    initialValue: this.searchApiProductControl.value,
+  });
+
+  protected readonly isApiProductSelected = computed(() => this.isApiProductCandidate(this.searchValue()));
+
+  protected readonly filteredOptions = computed(() => {
+    const value = this.searchValue();
+    const term = (this.isApiProductCandidate(value) ? value.name : value).trim().toLowerCase();
+    return this.data.candidates.filter(candidate => candidate.name.toLowerCase().includes(term));
+  });
+
+  protected displayFn(option: ApiProductCategoryCandidate): string {
+    return option?.name ?? '';
+  }
+
+  protected resetSearchTerm(): void {
+    this.searchApiProductControl.setValue('');
+  }
+
+  protected onCancelClick(): void {
+    this.dialogRef.close();
+  }
+
+  protected submit(): void {
+    const value = this.searchApiProductControl.getRawValue();
+    if (this.isApiProductCandidate(value)) {
+      this.dialogRef.close(value);
+    }
+  }
+
+  private isApiProductCandidate(value: string | ApiProductCategoryCandidate): value is ApiProductCategoryCandidate {
+    return typeof value !== 'string';
+  }
+}
--- a/gravitee-apim-console-webui/src/portal/catalog/category/category.component.html
+++ b/gravitee-apim-console-webui/src/portal/catalog/category/category.component.html
@@ -45,14 +45,14 @@
     @if (mode === 'edit') {
       <mat-card *gioPermission="{ anyOf: ['environment-documentation-r'] }">
         <mat-card-content class="category__content">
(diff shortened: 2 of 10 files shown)


gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
```diff
@@ -157,21 +177,40 @@ export class CategoryCatalogComponent implements OnInit {
...
+          .filter(navItem => (navItem.categoryIds ?? []).includes(this.categoryId)),
+      ),
+      switchMap(navItems =>
+        this.toApiProductVMs$(navItems).pipe(
+          catchError(({ error }) => {
+            this.snackBarService.error(error?.message ?? 'Unable to load API Products');
+            return of([]);
+          }),
+        ),
+      ),
```
thomasballivet, line 218: Sharing the nav-items call via `shareReplay` is a nice win, but the `catchError` moving downstream breaks the failure path: on `apis$` (L197) it now emits `[]` and **completes**, so a later `refreshNavigationItems.next(1)` can't revive the list (stale until reload) — it used to sit inside the `switchMap` and the outer stream survived. And `apiProducts$` doesn't catch that error at all, so it hits the `async` pipe and the card renders with no table and no message.

Moving it back inside the `switchMap` of the shared source fixes both and makes the one on L197 redundant:

```ts
switchMap(() =>
  this.mode === 'edit'
    ? this.portalNavigationItemService.getNavigationItems('TOP_NAVBAR', ['apis']).pipe(
        catchError(({ error }) => {
          this.snackBarService.error(error?.message ?? 'Unable to load navigation items');
          return of({ items: [] } satisfies PortalNavigationItemsResponse);
        }),
      )
    : of({ items: [] } satisfies PortalNavigationItemsResponse),
),
```

gravitee-apim-console-webui/src/portal/catalog/category/category.harness.ts
```diff
@@ -36,15 +36,19 @@ export class CategoryHarness extends ComponentHarness {
...
+    return await harnessLoader.getHarnessOrNull(MatButtonHarness.with({ selector: '[data-testid="add-api-product-button"]' }));
   }
 
   async getSaveBar(rootLoader: HarnessLoader): Promise<GioSaveBarHarness> {
     return await rootLoader.getHarness(GioSaveBarHarness);
   }
 
   async getTableRows(harnessLoader: HarnessLoader): Promise<MatRowHarness[]> {
-    return await harnessLoader.getHarness(MatTableHarness).then(table => table.getRows());
+    return await this.getApiTable(harnessLoader).then(table => table.getRows());
```
thomasballivet, line 51: `getTextByColumnNameAndRowIndex` just below still does `getHarness(MatTableHarness)` with no selector, so it grabs the first table in the DOM — fine today only because the APIs card renders first, and it would silently read the API Products table if that order or the permissions change. Route it through `getApiTable(...)` too, like this one? wdyt?

gravitee-apim-console-webui/src/portal/catalog/category/add-api-product-to-category-dialog/add-api-product-to-category-dialog.component.ts
```diff
@@ -0,0 +1,96 @@
...
+    ReactiveFormsModule,
+    MatAutocompleteModule,
+    MatButtonModule,
+    MatDialogModule,
+    MatFormFieldModule,
+    MatIconModule,
+    MatInputModule,
+  ],
+})
+export class AddApiProductToCategoryDialogComponent {
```
thomasballivet, line 56: This is nearly line-for-line `add-api-to-category-dialog` (same `FormControl`, `toSignal`, `filteredOptions`, `submit`, plus template/scss/harness). A single dialog parameterised by a label accessor would drop ~200 duplicated lines and avoid maintaining both. Fine as a follow-up if you'd rather not grow this PR — wdyt?

gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
```diff
@@ -300,19 +408,52 @@ export class CategoryCatalogComponent implements OnInit {
...
+  }
+
+  private findApiProductsByIds$(apiProductIds: string[]): Observable<Map<string, ApiProduct>> {
+    const idBatches = chunk([...new Set(apiProductIds)], 100);
+    if (idBatches.length === 0) {
+      return of(new Map());
+    }
+
+    return forkJoin(idBatches.map(ids => this.apiProductService.search({ ids }, 'name', 1, 100))).pipe(
+      map(responses => new Map(responses.flatMap(response => response.data ?? []).map(apiProduct => [apiProduct.id, apiProduct] as const))),
```
thomasballivet, line 471: `findApiProductsByIds$` (id chunking + `Map` aggregation) reads like reusable service logic rather than screen logic — would `ApiProductV2Service` be a better home? Also `sortBy: 'name'` on L455 is a no-op: results go into a `Map` and the display order comes from the nav items. wdyt?

gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
```diff
@@ -254,12 +293,81 @@ export class CategoryCatalogComponent implements OnInit {
...
           this.snackBarService.success(`'${apiVM.name}' removed successfully`);
-          this.refreshApis.next(1);
+          this.refreshNavigationItems.next(1);
         },
         error: ({ error }) => this.snackBarService.error(error?.message ?? 'Error during API removal'),
       });
   }
 
+  addApiProductToCategory(): void {
+    this.selectableApiProductCandidates$()
```
thomasballivet, line 313: `matDialog.open()` runs only after the candidates fully resolve, so clicking does nothing visible during the nav-items GET + `_search` calls, and the button stays clickable (two clicks = two dialogs). Opening straight away with a loading state — the way `ApiProductSectionEditorDialogComponent` already does with its `isLoading`/`hasLoadError` signals — and disabling the button while loading would cover it. Bonus: it also moves the failure handling into the dialog, so the current "Error during API Product addition" snackbar stops firing on what is really a load failure.

Related: `selectableApiProductCandidates$()` (L411) re-GETs the nav items even though `navigationItemsResponse$` — the `shareReplay` this PR introduces — is already subscribed by the two tables and holds the response. Reading it back with `take(1)` would serve the cached value with zero HTTP; the `take(1)` matters because that stream is driven by a `BehaviorSubject` and never completes, so without it the `switchMap` would re-fire on every refresh — including the one this very flow triggers after a successful add, reopening the dialog. Trade-off is a possibly slightly stale list (a product published elsewhere since the page loaded), so happy either way — wdyt?

gravitee-apim-console-webui/src/portal/catalog/category/category.component.html
```diff
@@ -45,14 +45,15 @@ <h3>General</h3>
     @if (mode === 'edit') {
       <mat-card *gioPermission="{ anyOf: ['environment-documentation-r'] }">
         <mat-card-content class="category__content">
-          <div class="category__apis__title">
+          <div class="category__assignments__title">
             <div class="category__content__header">
               <h3>APIs</h3>
             </div>
             <button
               class="add-button"
```
thomasballivet: nit: `class="add-button"` is dead now — nothing styles it (no rule in the component scss or the global styles) and the harness moved to `data-testid`. The API Products button doesn't carry it, so dropping it keeps the two consistent.

gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
jaroslawlakomy-grav: The error handling has been moved into the shared source switchMap, so the outer stream remains active after a failed request. The redundant downstream catch has been removed, and a regression test now verifies recovery on a subsequent refresh.

gravitee-apim-console-webui/src/portal/catalog/category/category.harness.ts
jaroslawlakomy-grav: Updated to use getApiTable(), so the helper no longer depends on the order of tables in the DOM.

gravitee-apim-console-webui/src/portal/catalog/category/add-api-product-to-category-dialog/add-api-product-to-category-dialog.component.ts
jaroslawlakomy-grav: The duplication is acknowledged. Consolidating both dialogs is intentionally deferred to a follow-up to avoid changing the existing API assignment flow in this PR.

gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
jaroslawlakomy-grav: The no-op sort has been removed. The batching helper remains private for now because it has a single consumer and returns a Map tailored to this screen. It can be extracted once a reusable service-level contract is needed.

gravitee-apim-console-webui/src/portal/catalog/category/category.component.ts
jaroslawlakomy-grav: A loading guard now disables the button and prevents multiple dialogs while candidates are being resolved. Candidate-loading failures are handled separately from update failures. The fresh navigation-items request has been retained to avoid stale candidates and remain consistent with the existing API assignment flow.

gravitee-apim-console-webui/src/portal/catalog/category/category.component.html
jaroslawlakomy-grav: The unused class has been removed.

thomasballivet (approved):

PR #19323, opened 2026-08-24
feat(portal-next): expose theme automation REST endpoints

konrad-krol-incubly:
## Issue

https://gravitee.atlassian.net/browse/EXT-175

## Description

Wires the REST layer for the automation **portal theme** use cases created in the base PR, and extends the Portal resource with a theme reference.

- Theme automation endpoints - themes become GitOps-managed via the Automation API: create, update, delete, and validate (dry-run).
- `activeThemeHrid` on PortalSpec - a portal now declares which theme it uses. On apply, the referenced theme is marked active for the environment and the previously-active PORTAL_NEXT theme is disabled.
- Symmetric clear semantics - leaving `activeThemeHrid` unset clears the reference and disables the previously-active theme. On the next portal-next visit the UI lazily falls back to a fresh default theme; the previous customized theme is preserved (disabled, not deleted) and can be re-attached later.
- Portal deletion mirrors the same rule - deleting an automation-managed portal disables whichever theme it was pointing at.
- GET reverse-resolves the HRID - retrieving a portal returns the current `activeThemeHrid` so the CRD round-trips cleanly.

## To be confirmed

- [ ] New API resource is named: portal-theme. Other type of themes (e.g. gamma) will have dedicated resources.


MateuszJozwiak-Incubly:
lgtm;

dev-marek (approved):

PR #19324, opened 2026-08-24
feat: integrate API Products into category-filtered catalog

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-136

## Summary

Integrates API Products into the category-filtered Portal Next catalog.

API Products assigned directly to the selected category are now returned and displayed alongside matching standalone APIs.

## Changes

- expose API Product `categoryIds` through the Portal API
- filter visible API Products by their direct category assignments
- preserve visibility, search, sorting, and pagination behavior
- propagate API Product categories to the Portal Next catalog view
- display assigned categories in the existing catalog list
- prevent API Products from inheriting categories from their nested APIs
- cover authenticated and anonymous catalog access

https://github.com/user-attachments/assets/64696e28-2415-48ea-9d72-2f3914cc6d8a

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/test/java/io/gravitee/rest/api/portal/rest/resource/PortalNavigationItemsResourceNotAuthenticatedTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/test/java/io/gravitee/rest/api/portal/rest/resource/PortalNavigationItemsResourceNotAuthenticatedTest.java
@@ -21,6 +21,7 @@ import inmemory.ApiProductQueryServiceInMemory;
 import inmemory.PortalNavigationItemsQueryServiceInMemory;
 import io.gravitee.apim.core.api_product.model.ApiProduct;
 import io.gravitee.apim.core.portal.model.PortalArea;
+import io.gravitee.apim.core.portal_category.model.PortalCategoryId;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationApiProduct;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItem;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
@@ -43,6 +44,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 public class PortalNavigationItemsResourceNotAuthenticatedTest extends AbstractResourceTest {
 
     private static final String ENV_ID = "DEFAULT";
+    private static final String CATEGORY_ID = "11111111-1111-1111-1111-111111111111";
 
     @Autowired
     private PortalNavigationItemsQueryServiceInMemory portalNavigationItemsQueryService;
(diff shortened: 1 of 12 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/GetVisiblePortalCatalogItemsUseCase.java
```diff
@@ -178,6 +178,20 @@ private List<PortalNavigationApiProduct> findVisibleApiProducts(
...
 
+    private List<PortalNavigationApiProduct> filterApiProductsByCategory(
+        List<PortalNavigationApiProduct> apiProducts,
+        Optional<PortalCategoryId> categoryId
+    ) {
+        return categoryId
+            .map(selectedCategoryId ->
+                apiProducts
+                    .stream()
+                    .filter(item -> item.getCategoryIds().contains(selectedCategoryId))
```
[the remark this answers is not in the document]
jaroslawlakomy-grav: `categoryIds` is a non-null invariant of `PortalNavigationApiProduct`. It defaults to `List.of()`, is marked as `@Nonnull`, and both construction and update paths normalize a missing value to an empty list. The repository model follows the same default, and the omitted-category paths are covered by model tests. Therefore, the direct `contains` call is safe and avoids duplicating domain normalization in the use case.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/GetVisiblePortalCatalogItemsUseCase.java
```diff
@@ -112,10 +112,10 @@ public Output execute(Input input) {
...
             navigationItemsById
         );
-        // Category assignment and catalog filtering are delivered separately; exclude products until catalog filtering is implemented.
-        List<PortalNavigationApiProduct> visibleApiProducts = input.categoryId().isPresent()
-            ? List.of()
-            : findVisibleApiProducts(navigationItems, input, navigationItemsById, accessibleApiNavigationItemIds, accessibleApiProductIds);
+        List<PortalNavigationApiProduct> visibleApiProducts = filterApiProductsByCategory(
+            findVisibleApiProducts(navigationItems, input, navigationItemsById, accessibleApiNavigationItemIds, accessibleApiProductIds),
+            input.categoryId()
+        );
```
thomasballivet, line 113: With a category filter, `visibleApis` only holds API nav items carrying that category (L90-94), and `resolveIncludedApiProducts` intersects each product's `apiIds` with that set (L372-381). Child API items never carry the product's category, so a product surfaced by this new filter comes back with an empty `apis` summary — the catalog renders "Included APIs: 0" for a product that lists them fine unfiltered.

Resolving the summary's APIs from a category-free visible set would keep it intact. The new tests all use `apiIds = Set.of()`, which is why it slips through.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/test/java/io/gravitee/rest/api/portal/rest/resource/PortalNavigationItemsResourceNotAuthenticatedTest.java
```diff
@@ -108,7 +110,7 @@ void should_return_navigation_items_when_user_is_not_authenticated() {
     }
 
     @Test
-    void should_not_return_private_api_product_in_catalog_search() {
+    void should_not_return_private_api_product_in_category_filtered_catalog_search() {
```
thomasballivet, line 150: Renaming this one drops the anonymous "private product stays hidden in an *unfiltered* catalog search" case — nothing covers it now. Keep the original alongside the category-filtered one? wdyt?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/use_case/GetVisiblePortalCatalogItemsUseCaseTest.java
```diff
@@ -411,17 +411,54 @@ void should_exclude_api_products_from_catalog_when_category_id_is_present() {
...
+            null,
+            PortalVisibility.PUBLIC,
+            List.of(PortalCategoryId.of(CATEGORY_ID_1))
+        );
+        var productOutsideCategory = apiProductItem(
+            "product-item-outside-category-id",
+            "product-outside-category-id",
+            null,
+            PortalVisibility.PUBLIC,
+            List.of(PortalCategoryId.of(UNKNOWN_CATEGORY_ID))
```
thomasballivet: nit: `UNKNOWN_CATEGORY_ID` means "a category that doesn't exist" elsewhere here (L475). A `CATEGORY_ID_2` would state what this case actually asserts: a product in *another* category must not show up.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/GetVisiblePortalCatalogItemsUseCase.java
jaroslawlakomy-grav: Fixed by resolving a category-independent visible API set for API Product summaries while keeping category filtering for standalone catalog entries. A regression test now verifies that a public nested API is included and a private one remains excluded.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/test/java/io/gravitee/rest/api/portal/rest/resource/PortalNavigationItemsResourceNotAuthenticatedTest.java
jaroslawlakomy-grav: The original unfiltered catalog test has been restored alongside the category-filtered case.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/use_case/GetVisiblePortalCatalogItemsUseCaseTest.java
jaroslawlakomy-grav: Added `CATEGORY_ID_2` for the product assigned to another category. `UNKNOWN_CATEGORY_ID` is now used only for the nonexistent-category case.

thomasballivet (approved):

PR #19341, opened 2026-08-24
feat(portal-next): portal-scoped current theme resolution

konrad-krol-incubly:
## Issue

https://gravitee.atlassian.net/browse/EXT-175

## Description

Follow-up to the theme automation work: flips theme resolution from env-first to portal-first, matching where ownership already lives (`Portal.activeThemeId`). 

Unblocks multi-portal-per-env without further code - the current "one enabled per env" invariant stops being coherent the moment `allowMultiplePortalPerEnv` flips, and removes a source of dual-truth between `Portal.activeThemeId` and `Theme.enabled`.

Portal-webui-next `GET /theme` gains an optional `portalId` query param; when supplied, resolution follows the portal's reference, otherwise it falls back to today's env-wide `enabled` lookup with lazy default-create. Console v2 untouched, nullable param, no OpenAPI/DB changes. 

Stacked on #19301.


dev-marek (approved):
Looks ok from technical/logical point of view, but please avoid abbreviations in names, eg. `envId`, `ctx` etc. - we're not getting charged more by github for using full names 😄

PR #19348, opened 2026-08-24
fix(api-product): search by kind and report deployment state

vikrantgravitee:
## What

API Product listings that ask for a single kind now narrow in the store, carry a real `deploymentState`, and resolve primary owners in one batch instead of three queries per row.

Raised from review feedback on [gravitee-gamma-module-aim#679](https://github.com/gravitee-io/gravitee-gamma-module-aim/pull/679), whose AI Workspace list consumes these paths. **This must merge first** — the AIM PR is correct only once these fixes are in the snapshot.

## Why

**Kind filtering was ineffective.** A listing wanting only one kind asked the store to *exclude* every other kind. A classic API Product carries no kind term, so it passed every exclusion and padded the page — a page of 25 could return fewer than 25 of the kind asked for, or none. Such listings now *require* the kind. Listings that also want classic products still narrow by exclusion, since no kind term can match a classic product.

**`deploymentState` was never resolved on the search path.** The field the API Product schema declares came back `null`, so anything counting deployed products counted none. Now resolved for the page.

**Primary owners cost 3 queries per row** (role → membership → user/group). A 25-row page cost ~75 round trips; the listing use case ran it across an entire environment.

## Cost

| Path | Before | After |
|---|---|---|
| Primary owner | 3 × N queries | 3 (user owners) to 6 (mixed), fixed |
| Deployment state | 2 × N queries | 2 queries, fixed |

Both are pinned by tests asserting the count does not grow with product count, so a regression fails CI rather than fading into a slow listing.

### Known cost, not yet addressed

The deployment-state lookup uses the existing `findAllByTypeAndEnvironments`, which scopes by **environment, not by id**. The product ids are filtered in memory afterwards, and because page/size are null the Mongo pipeline skips its limit and looks every matching row back up with its payload. So while the *query* count is fixed at 2, a 10-row page still reads every `DEPLOY_API_PRODUCT` payload in the environment — and AIM's summary walk pays that once per 500-page.

Scoping it store-side is possible without new plumbing: `EventCriteria.properties` already accepts a collection and both stores push it down (`JdbcEventLatestRepository:128`, `EventMongoRepositoryImpl:146`). It needs an id-scoped finder on `EventLatestQueryService`, which this PR deliberately avoided to keep the change API-Product-scoped. Flagged here rather than left implicit — happy to add it if reviewers prefer.

No shared query-service interface was changed. Every bulk finder used already existed: `findByReferencesAndRoleId`, `findBaseUsersByIds`, `findByIds`, `findAllByTypeAndEnvironments`, `findAllByReferenceIdsAndEnvironments`.

## Behaviour change worth noting

The classic Console API Products list will now show a real deployment state where it previously showed none. The v2 `ApiProduct` schema has always declared `deploymentState`; returning `null` was the defect.

## Tests

- `ApiProductSearchQueryServiceImplTest` (new) — the store is asked to require the kind
- `ApiProductDeploymentStateDomainServiceTest` (new) — state per product, fixed query count
- `ApiProductPrimaryOwnerDomainServiceTest` — user owners, group owners (including a group with no owner of its own), both in one pass, and the pinned query count
- `SearchApiProductsUseCaseTest` — deployment state on searched products
- 18 pre-existing `GetApiProductsUseCaseTest$DeploymentState` tests pass unchanged against the extracted domain service

`*ApiProduct*` + `*PrimaryOwner*` sweep: **643 tests, 0 failures**. Full REST API tree compiles.

## Scope

Every changed file is API-Product-scoped. Nothing touches APIs, proxies, or shared query-service interfaces.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/domain_service/ApiProductDeploymentStateDomainService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/domain_service/ApiProductDeploymentStateDomainService.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.apim.core.api_product.domain_service;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.gravitee.apim.core.DomainService;
+import io.gravitee.apim.core.api_product.model.ApiProduct;
+import io.gravitee.apim.core.event.model.Event;
+import io.gravitee.apim.core.event.query_service.EventLatestQueryService;
+import io.gravitee.apim.core.plan.model.Plan;
+import io.gravitee.apim.core.plan.query_service.PlanQueryService;
+import io.gravitee.definition.model.v4.plan.PlanStatus;
+import io.gravitee.rest.api.model.EventType;
+import io.gravitee.rest.api.model.v4.plan.GenericPlanEntity;
+import java.time.Instant;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import lombok.CustomLog;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Tells whether an API Product on screen still matches what was last deployed to the gateway.
+ *
+ * <p>A product is {@code DEPLOYED} when the last deploy event carries the same APIs and sharding tags it
+ * carries now, and no non-staging plan has been changed since. Anything else — never deployed, an
+ * unreadable payload, a since-changed plan — is reported as {@code NEED_REDEPLOY}, so the operator is
+ * told to redeploy rather than told nothing.</p>
+ *
+ * <p>The state is resolved for a whole collection at a time and costs two queries however many products
+ * are in it, so listing a page does not cost a query per row.</p>
+ */
+@DomainService
+@RequiredArgsConstructor
+@CustomLog
+public class ApiProductDeploymentStateDomainService {
+
+    private final EventLatestQueryService eventLatestQueryService;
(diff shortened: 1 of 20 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/api_product/domain_service/ApiProductDeploymentStateDomainService.java
```diff
@@ -0,0 +1,164 @@
...
+            )
+        );
+    }
+
+    private Map<String, Event> findLatestDeployEvents(Set<String> productIds, Set<String> environmentIds) {
+        if (environmentIds.isEmpty()) {
+            return Map.of();
+        }
+        return eventLatestQueryService
+            .findAllByTypeAndEnvironments(Set.of(EventType.DEPLOY_API_PRODUCT), environmentIds, Event.EventProperties.API_PRODUCT_ID)
```
araffass: This is two queries, but each one reads the whole environment.

There is no id scoping on the call, so the store returns every DEPLOY_API_PRODUCT latest event in the env and the productIds filter runs in memory on the next line. Page and size are null, so the Mongo pipeline skips its limit and looks every matching row back up with its payload. A 10 row page ends up deserializing every API Product deploy payload in the environment.

`EventCriteria.properties` already takes a collection and both stores push it down (`properties.API_PRODUCT_ID IN (...)`, Mongo `EventMongoRepositoryImpl:146`, JDBC `JdbcEventLatestRepository:128`), so the ids could go to the store instead. It needs widening `findAllByTypeAndEnvironments`, which I know you set out not to do, so your call. But the cost table reads as if the row count didn't change and it did.

It compounds on the AIM side: the summary walk pays this once per 500-page.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/membership/domain_service/ApiProductPrimaryOwnerDomainService.java
```diff
@@ -84,6 +91,111 @@ public PrimaryOwnerEntity getApiProductPrimaryOwner(final String organizationId,
...
+            .findBaseUsersByIds(List.copyOf(userIds))
+            .stream()
+            .collect(Collectors.toMap(BaseUserEntity::getId, Function.identity(), (first, second) -> first));
+    }
+
+    /**
+     * A group's own primary-owner member supplies the contact email, so groups cost one extra membership
+     * query for the whole batch rather than one per group.
+     */
+    private Map<String, PrimaryOwnerEntity> findGroupOwnersById(Set<String> groupIds, String primaryOwnerRoleId) {
```
araffass, line 156: The group path is the trickiest thing here and nothing covers it.

Three joins to get one owner: group membership by role, the groups themselves, then the member emails. I read it against the singular `findGroupPrimaryOwner` and it looks faithful, but `GetApiProductPrimaryOwners` only tests `Membership.Type.USER`, so a regression here lands silently.

Also worth pinning the query count the way `ApiProductDeploymentStateDomainServiceTest` does. Right now the "~2-4 total instead of 3xN" half of the claim isn't held by anything.

araffass (approved):
Approved with 2 comments worth flagging

PR #19349, opened 2026-08-24
fix(gateway): stop advertising mTLS subscription certificates in the TLS handshake

ThibaudAV:
## Issue

https://gravitee.atlassian.net/browse/APIM-14863

## Description

With `http.ssl.clientAuth` set to `request` or `required`, `SubscriptionTrustStoreLoader` loads each mTLS application **leaf** into the gateway trust store. JSSE uses `X509TrustManager#getAcceptedIssuers()` as the TLS `CertificateRequest.certificate_authorities`, so that list was sent to **every** client of the listener — including callers of the other, non-mTLS APIs of the same listener, who present no certificate at all. Those entries are leaves, not CAs: advertising them disclosed which client identities the gateway accepts.

The fix itself is [gravitee-node#603](https://github.com/gravitee-io/gravitee-node/pull/603), released in **9.8.0**. This PR brings it into APIM.

**⚠️ security-sensitive**: this changes TLS handshake behaviour and certificate handling.

### Commits

| Commit | What |
|---|---|
| `fix(gateway)` | bump `gravitee-node` 9.7.0 → 9.8.0, document `sendClientCertificateAuthorities` on `http`/`tcp`/`edge`/`kafka`, and add the integration tests pinning what the listener advertises |
| `feat(helm)` | move the Hazelcast node plugin URLs to 9.8.0 (Danger enforces they match `gravitee-node.version`), expose the option under `gateway.ssl`, `gateway.servers[].ssl` and `gateway.kafka.ssl`, and carry the warning into the chart CHANGELOG |

### Behaviour change

**Nothing is advertised by default now**, where the configured trust store used to be. Validation is untouched — the full trust store is still used to validate client certificates, so mTLS plan matching is unaffected. An empty list means "no constraint": clients keep presenting their certificate.

The one case that changes is a client that *relied* on the list to choose which certificate to present — typically a JDK client holding several client certificates, which without the list can pick the wrong one. Those deployments can restore the previous behaviour with:

```yaml
http:
  ssl:
    sendClientCertificateAuthorities: true
```

which sends **only the configured trust store**. Certificates registered at runtime by subscriptions are never advertised, whatever the setting — so enabling this on a listener serving mTLS plans cuts those subscriptions off. That consequence is now asserted by a test, not only documented.

### Tests

Three new integration tests in `plan/mtls`. They capture what actually reaches the wire rather than what the trust manager reports: JSSE hands the decoded `certificate_authorities` to the *client* key manager, so an instrumented client records exactly what the gateway sent. Parameterized on TLS 1.2 and TLS 1.3.

| Test | Config | Asserts |
|---|---|---|
| `…NotAdvertisedIntegrationTest` | default | nothing advertised, and the subscribed certificate still gets a 200 |
| `…AdvertisedIntegrationTest` | `sendClientCertificateAuthorities: true` | only the configured trust store, never the subscription leaf, and the subscription is cut off |

Both fail against gravitee-node 9.7.0 with the subscription leaf DN (`CN=APIM_CN`) in the failure message, which is the reported behaviour.

- `plan/mtls` + `tls` suite: **174 tests, 0 failures**
- Helm: `helm lint` clean, **942 unit tests**, 0 failures
- Mutation-checked: switching `clientAuth` to `none` turns the headline test red; a shifted indentation and a wrong `hasKey` guard each turn exactly one Helm case red

### Verified separately

Reproduced end to end on a running gateway with `openssl s_client` (no client certificate), before and after, on TLS 1.2 and 1.3 — leak visible before, `No client certificate CA names sent` after, subscribed leaf still 200 / no cert still 401.

### Heads-up

Cockpit's exchange controller WS server runs `clientAuth: REQUIRED` through the same code path and is affected by the new default.

### Backports

**master only** — decided, no `apply-on-*` label. `4.12.x` is on node 9.7.0 and would be a straight bump; `4.11.x` (8.1.0) and `4.10.x`/`4.9.x` (7.27.0) would each need `gravitee-node#603` backported to `origin/8.x` and `origin/7.x` respectively, and released, first.

### Known deviation

The branch is `APIM-14863-…` without the `<type>/` prefix `AGENTS.md` asks for. Renaming it now would retarget the PR mid-review; flagging rather than churning.


graviteeio:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Fails</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Gravitee node plugins defined in Helm values.yaml needs to be consistent with the version defined in the pom.xml.</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 9d8531d26849d466d95ef42873391289d835df77
</p>

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/plan/mtls/AbstractPlanMutualTLSCertificateAuthoritiesIntegrationTest.java
```diff
@@ -0,0 +1,298 @@
...
+
+    /**
+     * Records the {@code certificate_authorities} the gateway sent, then declines to present a certificate.
+     */
+    private static class CapturingKeyManager extends X509ExtendedKeyManager {
+
+        private volatile Principal[] sentAuthorities;
+
+        List<String> sentAuthorities() {
+            return sentAuthorities == null ? List.of() : Arrays.stream(sentAuthorities).map(Principal::getName).toList();
```
phiz71, line 251: `chooseClientAlias` only runs when the server actually sends a `CertificateRequest`. Mapping "never invoked" onto the same empty list as "sent an empty `certificate_authorities`" means `should_not_advertise_any_certificate_authority` cannot tell the two apart.

Concretely: drop `.set("http.ssl.clientAuth", "request")` from `configureGateway` above, or let a regression stop the listener requesting a client certificate at all, and `sentAuthorities` stays null, this returns `List.of()`, and the assertion is green. That is the headline non-regression test of the PR passing without exercising the behaviour it pins.

A `volatile boolean captured` set alongside `sentAuthorities` in `chooseClientAlias`, exposed and asserted before the `isEmpty()` in the test method, is enough to close it.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/plan/mtls/AbstractPlanMutualTLSCertificateAuthoritiesIntegrationTest.java
```diff
@@ -0,0 +1,298 @@
...
+        context.init(new KeyManager[] { keyManager }, new TrustManager[] { new TrustEverything() }, null);
+
+        try (SSLSocket socket = (SSLSocket) context.getSocketFactory().createSocket("localhost", gatewayPort)) {
+            socket.setEnabledProtocols(new String[] { protocol });
+            socket.startHandshake();
+            // TLS 1.3 sends the CertificateRequest after the client considers the handshake done, so exchange
+            // application data to make sure it has been processed before asserting
+            socket.getOutputStream().write("GET / HTTP/1.1\r\nHost: localhost\r\n\r\n".getBytes());
+            socket.getOutputStream().flush();
+            socket.getInputStream().read();
```
phiz71, line 228: No timeout anywhere on this socket: no connect timeout, no `setSoTimeout`, and neither test method carries `@Timeout` — unlike `PlanMutualTLSClientAuthRequestIntegrationTest`, which uses `@Timeout(value = 30, timeUnit = SECONDS)`.

`GET /` matches no deployed API, so this depends on the gateway answering on a path nothing else in the suite exercises. If it ever accepts the connection and holds it without responding, this `read()` blocks forever and the CircleCI job burns to `no_output_timeout` with no failure message rather than a red test.

`socket.setSoTimeout(30_000)` before `startHandshake()` turns that into a readable `SocketTimeoutException`.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/plan/mtls/AbstractPlanMutualTLSCertificateAuthoritiesIntegrationTest.java
```diff
@@ -0,0 +1,298 @@
...
+
+    @Override
+    public void configurePolicies(final Map<String, PolicyPlugin> policies) {
+        policies.put("mtls", PolicyBuilder.build("mtls", MtlsPolicy.class, MtlsPolicyConfiguration.class));
+    }
+
+    @Override
+    public void configureEntrypoints(Map<String, EntrypointConnectorPlugin<?, ?>> entrypoints) {
+        entrypoints.putIfAbsent("http-proxy", EntrypointBuilder.build("http-proxy", HttpProxyEntrypointConnectorFactory.class));
+        entrypoints.putIfAbsent("http-get", EntrypointBuilder.build("http-get", HttpGetEntrypointConnectorFactory.class));
```
phiz71: `http-get` here, `mock` on line 122 and `MessageApiReactorFactory` on line 134 are unreachable: `@DeployApi` on this class lists only `/apis/plan/v4-proxy-api.json`, which is `type: proxy` with a single `http` listener on the `http-proxy` entrypoint and one `http-proxy` endpoint group.

They are carried over from `PlanMutualTLSClientAuthRequestIntegrationTest`, which also deploys `v4-message-api.json`. Here they load the message reactor and two connector plugins into both new gateway instances for nothing, and they suggest to the next reader that a message API is under test. `AGENTS.md`, *Clean code*: "No dead code: after a change, remove the fields, methods, imports, and parameters it left unused."

Same line of thought for `isV4Api(definitionClass)` on line 127, always true with a single v4 API.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/plan/mtls/AbstractPlanMutualTLSCertificateAuthoritiesIntegrationTest.java
```diff
@@ -0,0 +1,298 @@
...
+
+    protected SubscriptionTrustStoreLoaderManager subscriptionTrustStoreLoaderManager;
+
+    /**
+     * @return the value of {@code http.ssl.sendClientCertificateAuthorities}, or {@code null} to leave it unset.
+     */
+    protected abstract Boolean sendClientCertificateAuthorities();
+
+    @Override
+    public void configurePolicies(final Map<String, PolicyPlugin> policies) {
```
phiz71, line 112: From here down to `registerSubscription`, this is a near-verbatim copy of the block already present in each of the four existing `PlanMutualTLS*IntegrationTest` classes: `configurePolicies` / `configureEntrypoints` / `configureEndpoints` / `configureApi` / `configureReactors` / `configureHttpClient`, the `@BeforeEach` reflection cheat on `SubscriptionCacheService`, and the subscription factory.

This PR is the first to extract an abstract base for it, which is the right move — but the four existing classes are not moved onto it, so the package now carries five copies instead of fewer. The next change to the mTLS fixture has to land in five places, and the two variants have already drifted: the existing `aSubscription` reads the certificate with `Files.readString(...).getBytes()`, `registerSubscription` here with `Files.readAllBytes(...)`.

Not a blocker, but rebasing the four onto this base while it is fresh is much cheaper than later.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/plan/mtls/PlanMutualTLSCertificateAuthoritiesAdvertisedIntegrationTest.java
```diff
@@ -0,0 +1,58 @@
...
+class PlanMutualTLSCertificateAuthoritiesAdvertisedIntegrationTest extends AbstractPlanMutualTLSCertificateAuthoritiesIntegrationTest {
+
+    @Override
+    protected Boolean sendClientCertificateAuthorities() {
+        return true;
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = { "TLSv1.2", "TLSv1.3" })
+    void should_advertise_only_the_configured_trust_store(String protocol, GatewayDynamicConfig.Config gatewayConfig) throws Exception {
```
phiz71, line 61: The failure mode this PR documents three times — in `gravitee.yml`, in `values.yaml` and in the node javadoc — is that enabling the flag cuts off every mTLS subscription whose certificate is not issued by an authority of the configured trust store. It is asserted in prose and tested nowhere.

The fixtures for it are already here: with the flag on, `client2.cer` as the configured trust store and `client.cer` registered by the subscription, a JDK client should withhold its certificate entirely, so the call that returns 200 in the default configuration should not. Adding the `@WithCert` counterpart of `should_still_accept_the_certificate_registered_by_the_subscription` would pin the documented consequence — and would go red the day the client stops behaving that way, at which point the warning in both configuration files becomes wrong.

I checked node 9.8.0 too: `TrustManagerHandshakeTest` only covers *everything advertised* and *nothing advertised*, so this is not covered upstream either.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/plan/mtls/PlanMutualTLSCertificateAuthoritiesAdvertisedIntegrationTest.java
```diff
@@ -0,0 +1,58 @@
...
+        Subscription subscription = registerSubscription();
```
phiz71, line 62: Nothing in this test proves the subscription leaf ever reached the gateway trust store, so both assertions below pass just as well when the registration silently did nothing.

`SubscriptionTrustStoreLoaderManager.registerSubscription` swallows `MalformedCertificateException` (`log.error(...); return null;`) and `registerSubscription()` returns a `Subscription` either way. Move or re-encode `plans/mtls/client.cer`, or change the base64 wrapping, and the leaf never enters the trust store while `containsExactly(configured)` and `doesNotContain(leaf)` both stay green — the test would then assert nothing about the disclosure it exists to pin.

`PlanMutualTLSCertificateAuthoritiesNotAdvertisedIntegrationTest` is protected by `should_still_accept_the_certificate_registered_by_the_subscription`, which fails loudly if the leaf is missing. This class has no equivalent guard.

helm/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
...
 
 ### 4.13.0
+- Bump the Gravitee node Hazelcast cache and cluster plugins to 9.8.0, to stay consistent with `gravitee-node.version` in the APIM `pom.xml`.
+- The gateway no longer sends its truststore as the list of acceptable certificate authorities during the TLS handshake (APIM-14863). Certificates registered at runtime by mTLS plan subscriptions were application leaves, not authorities, and advertising them told every caller of the listener which client identities are accepted. Validation is unchanged, so mTLS plan matching is unaffected, and an empty list means "no constraint": clients keep presenting their certificate. Only a client that *relied* on the list to choose which certificate to present is affected — typically a JDK client holding several client certificates. Those deployments can send the configured truststore again with the new `gateway.ssl.sendClientCertificateAuthorities` (also available per server under `gateway.servers[].ssl` and for the Kafka listener under `gateway.kafka.ssl`); certificates registered by subscriptions are never sent, whatever the setting.
```
phiz71: This offers `sendClientCertificateAuthorities` as the way to "send the configured truststore again", without the warning that `values.yaml` and the distribution `gravitee.yml` both carry: enabling it cuts off every mTLS plan subscription on that listener.

That matters because the CHANGELOG is the one document the affected operator reads. The path is: they upgrade, their JDK client stops picking a certificate, they find this line, they set the flag — and every mTLS subscription breaks, since subscription certificates are self-signed leaves and are never issued by an authority of the configured trust store, so a non-empty advertised list makes the client withhold its certificate.

"certificates registered by subscriptions are never sent, whatever the setting" currently reads as reassurance rather than as the trap it is. One sentence lifted from `values.yaml` closes it.

helm/values.yaml
```diff
@@ -1837,6 +1837,14 @@ gateway:
...
   #    path: ${gravitee.home}/security/truststore.jks
   #    password: secret
+  #  # Send the configured truststore to clients as the list of acceptable certificate authorities, so a client
+  #  # holding several certificates can pick the right one. Disabled by default: the list reaches every client of
+  #  # the listener, and an empty list means "no constraint" so clients keep presenting their certificate.
+  #  # Only enable it when every expected client certificate is issued by an authority of the truststore above:
+  #  # a client whose issuer is absent from a non-empty list withholds its certificate entirely. Certificates
+  #  # registered at runtime by mTLS plan subscriptions are never sent, so enabling this cuts off every
+  #  # subscription whose certificate is not issued by an authority of the configured truststore.
+  #  sendClientCertificateAuthorities: false
```
phiz71, line 1853: The chart renders this key under `gateway.ssl`, `gateway.servers[].ssl` and `gateway.kafka.ssl`, and the CHANGELOG names all three, but it is documented only here.

An operator running the multi-server layout or the Kafka listener finds no trace of it in the file they configure from. `README.md` (the `gateway.ssl.*` table around line 348) and `README.adoc` do not carry it either — though those tables are already stale for `gateway.edge` and `gateway.kafka`, so that half is a pre-existing habit rather than something this PR broke.

helm/tests/gateway/configmap_ssl_test.yaml
```diff
@@ -166,3 +166,50 @@ tests:
             \s   crl:
             \s     path: \$\{gravitee\.home\}/security/crl\.pem
             \s     watch: true
+
+  - it: Client certificate authorities not sent by default
```
phiz71, line 170: Both new cases exercise the `gateway.ssl` rendering only; the `gateway.servers[].ssl` and `gateway.kafka.ssl` renderings added in the same commit have none.

I mutation-tested the `matchRegex` case below by shifting the key two spaces left in the template and it correctly went red, so the assertion is meaningful for the path it covers. The other two paths have nothing behind them: a wrong indentation or a wrong `hasKey` guard renders a misplaced key and the suite stays green.

Worth noting for later: the "not sent by default" case uses `notMatchRegex` over the whole document, so once the other blocks are covered it will fail for the wrong block if a default is ever added there.

gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-gateway/src/main/resources/config/gravitee.yml
```diff
@@ -108,6 +108,16 @@ secrets:
...
+#    Whether the configured truststore is sent to clients as the list of acceptable certificate authorities during
+#    the TLS handshake, so a client holding several certificates can pick the right one. Disabled by default: the
+#    list is sent to every client of the listener, including consumers of non-mTLS APIs, and an empty list simply
+#    means "no constraint" so clients keep presenting their certificate.
+#    Only enable it when every expected client certificate is issued by an authority present in the truststore
+#    configured below. A non-empty list is a constraint, not a hint: a client whose issuer is absent from it
+#    withholds its certificate entirely. In particular, client certificates registered at runtime by mTLS plan
+#    subscriptions are never sent, so enabling this cuts off every subscription whose certificate is not issued
+#    by an authority of the configured truststore.
+#    sendClientCertificateAuthorities: false
```
phiz71, line 120: Documented under `http.ssl` only, while the same file documents `tcp.ssl` (line 165), `edge.ssl` (line 210) and `kafka.ssl` (line 337) — all governed by the same code path, all losing their advertised list with this bump.

The option is read generically as `<prefix>.ssl.sendClientCertificateAuthorities` in `VertxServerOptions`, and the Kafka listener does reach it: `KafkaServerFactory.initTcpOptions()` builds `VertxTcpServerOptions.builder().prefix("kafka").environment(environment)`. So `kafka.ssl.sendClientCertificateAuthorities` works, the chart already exposes it, and an operator reading this file next to `kafka.ssl.clientAuth` has no way to learn it exists.

The Edge block is the sharpest case: its own comment recommends `clientAuth: required` in production and states its trust store holds "the certificate authority of the Edge fleet, not the enrolled leaf certificates" — the one listener where advertising was legitimate stops silently, with no pointer to the escape hatch.

phiz71 (commented):
Reviewed the three commits. The fix itself holds up, and I checked the parts that a bump like this usually gets wrong: gravitee-node 9.8.0 is published (POM and both Hazelcast plugin zips return 200), the 9.7.0 -> 9.8.0 delta is the fix commit plus the release bump and nothing else, the two-argument `TrustStoreLoaderManager` constructor survives so `SubscriptionTrustStoreLoaderManagerTest` still compiles, no 9.7.0 reference is left anywhere in the tree, and `kafka.ssl.sendClientCertificateAuthorities` is genuinely reachable rather than inert. `helm unittest` is green on the new file, and I mutation-tested the new `matchRegex` case — shifting the key two spaces left in the template turns it red — so that assertion is meaningful.

Ten findings inline. The two I would not merge without are both about the new tests being able to stay green while asserting nothing: see the comments on `sentAuthorities()` and on `registerSubscription()` in the Advertised test. The missing warning in the chart CHANGELOG is worth folding into the same pass, since that file ships to customers.

One point with no line to anchor to: the description replaces the sections of `.github/pull_request_template.md` (`## Issue` / `## Description` / `## Additional context`) with its own headings, the commit footers carry the bare key `APIM-14863` rather than the full browse link, and the branch is `APIM-14863-...` without the `<type>/` prefix `AGENTS.md` asks for.

ThibaudAV:
Thanks — the two you flagged as blocking were both right, and both would have shipped a test that asserts nothing.

**Taken, in `test(gateway)`:**

- **`sentAuthorities()` conflated "never invoked" with "sent an empty list".** Fixed with a `volatile boolean askedForACertificate` set in `chooseClientAlias` and asserted in `advertisedCertificateAuthorities` before returning. I ran your mutation: switching `configureGateway` to `clientAuth: none` now fails all three cases with *"the gateway did not send a CertificateRequest, so nothing was advertised to capture"*. Before the fix it passed.
- **`registerSubscription()` proved nothing.** Confirmed `registerSubscription` returns `void` and swallows `MalformedCertificateException` in both the `computeIfPresent` and `computeIfAbsent` branches. Rather than duplicating the 200-call guard into the Advertised class, the check now lives at the registration point in the base and protects both: `getByCertificate(api, plan, CertificateUtils.generateThumbprint(leaf, "SHA-256"))` must be present.
- **No timeout.** `setSoTimeout(30s)` before `startHandshake()`, plus `@Timeout(30, SECONDS)` on every test method, matching `PlanMutualTLSClientAuthRequestIntegrationTest`.
- **Dead fixture.** Dropped `http-get`, `mock`, `MessageApiReactorFactory` and the `isV4Api` guard — only `v4-proxy-api.json` is deployed here.
- **The documented consequence, untested.** Added `should_cut_off_a_subscription_whose_certificate_is_not_issued_by_an_advertised_authority`. It passes, which is the interesting part: it confirms the warning the three configuration files carry is actually true, and it goes red the day a JDK client stops withholding.

**Taken, in `docs(gateway)`:**

- **`tcp.ssl` / `edge.ssl` / `kafka.ssl`.** All three documented now. I verified the `tcp` prefix directly (`VertxReactorConfiguration:91`, `.prefix(TCP_PREFIX)`); `KafkaServerFactory` is not in this repo so I took your reading of it, and the chart already exposes `kafka.ssl.sendClientCertificateAuthorities`. Edge got its own wording instead of a cross-reference, for the reason you gave: its trust store holds the fleet CA and no subscription certificate is ever registered on that private server, so it is the one listener where enabling this is both safe and useful.
- **`values.yaml` for `servers[]` and `kafka`.** Documented. I left `README.md` / `README.adoc` alone — as you noted, those tables are already stale for `gateway.edge` and `gateway.kafka`, and fixing them properly is its own change.
- **The CHANGELOG read as reassurance rather than as the trap.** Rewritten: enabling it on a listener serving mTLS plans breaks every subscription on it.
- **Helm coverage of the other two renderings.** Two cases added. Mutation-checked both: shifting the key two spaces in the `$server` block reds exactly the servers case, and typoing the `hasKey` guard reds exactly the Kafka case.

**Not taken:**

- **Rebasing the four existing `PlanMutualTLS*IntegrationTest` onto the new base.** You're right that the package now carries five copies and that the two variants have already drifted on how the certificate is read. But it means rewriting four test classes that currently guard the mTLS behaviour this PR changes, in the same PR that changes it — if one of them breaks I lose the ability to tell a fixture bug from a real regression. Happy to do it as a follow-up where a diff of "no behaviour change, four classes moved onto a base" is reviewable on its own.
- **Commit footers as full browse links.** Done for the two new commits. I did not rewrite the three already-reviewed ones: `AGENTS.md` says to create new commits rather than amend.
- **Branch prefix.** Correct, it should be `fix/APIM-14863-…`. Renaming retargets the PR mid-review, so I flagged it in the description instead.

The description now follows `.github/pull_request_template.md` and carries the security-sensitive flag.

Suites after the pass: `plan/mtls` + `tls` **174 tests, 0 failures**; Helm **942 tests, 0 failures**, `helm lint` clean.

phiz71 (commented):
Re-reviewed after the two commits and the fixup. I checked the claims rather than taking them, and they hold: I re-ran both Helm mutations myself (breaking the Kafka `hasKey` guard reds exactly the Kafka case, shifting the `$server` indentation reds exactly the servers case, nothing else moves), and I verified the new registration guard uses the right key — `SubscriptionCertificate` derives its fingerprint with exactly `CertificateUtils.generateThumbprint(cert, "SHA-256")`, and `CacheKey` scopes on `SubscriptionScope.of(subscription)`, which returns `getApi()` when no API Product is set, which is the case here.

Nine of the ten threads are addressed; I am resolving those. Placing the registration guard in the base rather than duplicating the 200-call into the Advertised class is better than what I asked for — it protects both classes at the point where the failure is swallowed.

Three new notes inline, one of which I would act on before merge (the `isNotEqualTo(OK_200)` assertion); the other two are recorded rather than requested.

Left open deliberately: the thread about rebasing the four existing `PlanMutualTLS*IntegrationTest` onto the new base. Your reason for deferring is a good one and I am not asking you to do it here — the thread is only a pointer, so resolve it once the follow-up exists.

One point with no line to anchor to, worth catching before merge: `a91d9594ba` is a literal `fixup!` commit, and this repository merges by rebase — each commit lands on `master` as it is, not squashed. Unless it is autosquashed into `6ccc20d7f4` first, `master` gains a commit titled `fixup! test(gateway): …`. That is not hypothetical; two have already leaked into the last 3000 commits of `master`.

For what it is worth on the `Validate backend` failure: I had reproduced it locally before it went red — running the repository's own prettier-java 2.7.5 with `.prettierrc` against the file, the single rewrite was that import pair — and 0 of 600 sampled Java files in the repository have an unsorted import block, so the ordering is enforced rather than conventional. `Validate backend` is green again on the fixup, and this is the first push where the backend chain gets past the format gate, so the new cut-off test is about to run somewhere other than a laptop for the first time.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/plan/mtls/PlanMutualTLSCertificateAuthoritiesAdvertisedIntegrationTest.java
```diff
@@ -0,0 +1,102 @@
...
+        wiremock.stubFor(get("/endpoint").willReturn(ok("endpoint response")));
+        Subscription subscription = registerSubscription();
+        try {
+            client
+                .rxRequest(GET, PlanHelper.getApiPath("v4-proxy-api"))
+                .flatMap(HttpClientRequest::rxSend)
+                .test()
+                .awaitDone(30, TimeUnit.SECONDS)
+                .assertValue(response -> {
+                    assertThat(response.statusCode()).isNotEqualTo(OK_200);
```
phiz71: This is the one assertion in the package that only says "not 200". The four existing mTLS tests next door assert `isEqualTo(UNAUTHORIZED_401)` and that the body carries `MtlsPolicy.FAILURE_MESSAGE`; this accepts 401, 404, 500 and 502 alike.

That gap is reachable here, not theoretical: a Vert.x `rxRequest(GET, path)` sends no `Accept` header, and the gateway's failure-template path dereferences `request.headers().get(ACCEPT)` — an empty 500 satisfies `isNotEqualTo(OK_200)` exactly as well as the 401 the test exists to demonstrate. Since the whole point is to prove the warning now carried by `gravitee.yml`, `values.yaml` and the chart CHANGELOG is true, the assertion should pin the mechanism rather than the absence of success: 401, plus the mTLS failure message in the body.

Worth adding `.assertComplete()` alongside `.assertValue(...)` while you are here, as the sibling tests do.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/plan/mtls/AbstractPlanMutualTLSCertificateAuthoritiesIntegrationTest.java
```diff
@@ -0,0 +1,316 @@
...
+     */
+    protected List<String> advertisedCertificateAuthorities(int gatewayPort, String protocol) throws Exception {
+        CapturingKeyManager keyManager = new CapturingKeyManager();
+        SSLContext context = SSLContext.getInstance("TLS");
+        context.init(new KeyManager[] { keyManager }, new TrustManager[] { new TrustEverything() }, null);
+
+        try (SSLSocket socket = (SSLSocket) context.getSocketFactory().createSocket("localhost", gatewayPort)) {
+            // without this a gateway that accepts the connection and never answers hangs the whole CI job instead
+            // of failing the test
+            socket.setSoTimeout((int) HANDSHAKE_TIMEOUT.toMillis());
```
phiz71, line 221: Small follow-on to the timeout fix, which is otherwise right: this socket timeout and the `@Timeout` on the test methods are both 30 s, so they race. Whichever fires first decides the message you get, and JUnit's generic "execution timed out after 30000 ms" is the less useful of the two — the `SocketTimeoutException` names the line that actually stalled.

Dropping this one to ~10 s while leaving the method timeout at 30 s makes the informative error win deterministically.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/plan/mtls/AbstractPlanMutualTLSCertificateAuthoritiesIntegrationTest.java
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    @Override
+    public void configureEndpoints(Map<String, EndpointConnectorPlugin<?, ?>> endpoints) {
+        endpoints.putIfAbsent("http-proxy", EndpointBuilder.build("http-proxy", HttpProxyEndpointConnectorFactory.class));
+    }
+
+    @Override
+    public void configureApi(ReactableApi<?> api, Class<?> definitionClass) {
+        configurePlans((Api) api.getDefinition(), Set.of("mtls"));
```
phiz71, line 128: Consequence of removing the `isV4Api` guard, worth a note rather than a change: this cast is now unguarded. It is correct today — `@DeployApi` on this class lists only `v4-proxy-api.json`, which is a v4 definition — and adding a guard back for a case that cannot happen would be exactly the speculative defence `AGENTS.md` argues against.

Recording it so the next person is not surprised: the day a second entry joins `@DeployApi`, this needs the guard again or it throws `ClassCastException` at deploy time.

ThibaudAV:
All three taken, and the `fixup!` point was the important one.

**`isNotEqualTo(OK_200)` was too loose.** You're right that it was the one assertion in the package that only says "not 200", and right about why it matters here: the test exists to prove the warning three configuration files now carry, so it has to pin the mechanism. Now `isEqualTo(UNAUTHORIZED_401)` plus `MtlsPolicy.FAILURE_MESSAGE` in the body, with `.assertComplete()`, matching the four siblings. It passes as written — so the client really does withhold and the gateway really does answer 401 through the mTLS policy, which is a stronger statement than the test made before.

**The two 30 s timeouts raced.** Socket timeout down to 10 s, method timeout left at 30 s, with a comment saying why they differ so the next person does not "tidy" them back into agreement.

**The unguarded cast.** Recorded, no change — agreed that guarding a case `@DeployApi` makes impossible is the speculative defence `AGENTS.md` argues against. Your comment is the note.

**The `fixup!` commit.** Verified before acting rather than taking it: `git log origin/master | grep -c '^[0-9a-f]* fixup!'` over the last 3000 commits returns **2**, so this is a demonstrated leak and not a hypothetical. Autosquashed into `6ccc20d7f4` via `GIT_SEQUENCE_EDITOR=true git rebase -i --autosquash`, and checked the squash was lossless — tree `e64954fa` before and after, identical. The branch is five clean commits with no `fixup!`, force-pushed with `--force-with-lease`.

I should have caught that one myself: my own working rules say to autosquash a fixup rather than leave it dangling, and I left it.

On the format gate — thanks for reproducing it. Root cause on my side was worse than a stale prettier: I was passing `-Dskip.validation=true` on every local run, and the plugin is configured `<skip>${skip.validation}</skip>`, so my local "prettier clean" checks were verifying nothing at all. Same shape as the two findings you opened on the tests. Re-checked without the flag: 0 unformatted Java files in the module.

Local after this pass: the six advertised-authorities cases green, including the tightened cut-off assertion.

callaertanthony (approved):
Approving. Reviewed at `0fe2f8d`. The node bump, the three config surfaces and the tests line up; CI is green (49/49). Three findings, none blocking.

## Findings

### 1. No backport labels — improvement

The PR carries no label and targets `master` only, but the disclosure exists on every maintained branch.

- `4.12.x` is already on `gravitee-node` 9.7.0, so `apply-on-4-12-x` is a straight bump.
- `4.11.x` is on 9.8.0's predecessor line (8.1.0), so it needs a node 8.x backport of gravitee-io/gravitee-node#603 first. Same for 4.10.x and 4.9.x.

Worth deciding explicitly rather than defaulting to master-only — a fix for what the listener discloses is exactly the kind that customers on a support branch are waiting for.

### 2. The two other Helm render paths have no chart test — nit

`helm/tests/gateway/configmap_ssl_test.yaml` covers `gateway.ssl` in both directions (rendered / not rendered), which is the right shape. But the option renders in three places, and the two others are untested:

- `helm/templates/gateway/gateway-configmap.yaml:108` — `gateway.servers[].ssl`
- `helm/templates/gateway/gateway-configmap.yaml:284` — `gateway.kafka.ssl`

A wrong indent or a `hasKey` on the wrong map there ships silently. Two more cases mirroring the existing ones would close it.

### 3. Ticket reference in test javadoc — nit

`AbstractPlanMutualTLSCertificateAuthoritiesIntegrationTest.java:76` and the class javadoc of both subclasses carry `APIM-14863`. There is precedent both ways in this module, so this is a preference, not a rule — the commit messages and this PR already hold the trail. Take it or leave it.

## What I checked

Stating these so the approval means something more than "diff looks fine":

- **Property name and default match the dependency.** `gravitee-node` 9.8.0 (released today) binds `<prefix>.ssl.sendClientCertificateAuthorities` with `DEFAULT_SEND_CLIENT_CERTIFICATE_AUTHORITIES = false` (`VertxServerOptions.java:62,192`) — exactly the key written by the distribution `gravitee.yml`, `values.yaml` and all three configmap blocks.
- **No compile break from the bump.** node#603 added a 3-arg `TrustStoreLoaderManager` constructor but kept the 2-arg overload, which is the one `SubscriptionTrustStoreLoaderManagerTest:120` uses.
- **The bump is minimal.** 9.7.0…9.8.0 is one feature commit plus the release commit; no other surface change riding along.
- **The plugin URLs resolve.** Both Hazelcast 9.8.0 zips return 200 on download.gravitee.io, and all three references to 9.7.0 in the repo were updated — none left.
- **`gateway.edge.ssl` needs no template change**: it renders through `toYaml`, so the new key passes through as-is. Good call not touching it.
- **The guards cannot hit a nil map**: both new `hasKey` calls sit inside an existing `if .Values.gateway.ssl.enabled` / `if .Values.gateway.kafka.ssl`.
- **The chart-test regex really matches**: the 4-space pattern lines up with the actual indent inside the `gravitee.yml` block scalar (8 in the template, minus the 4 of the block).
- **The registration guard is not vacuous**: it looks the certificate up by the same `CertificateUtils.generateThumbprint(cert, "SHA-256")` that `SubscriptionCertificate` keys on, so a leaf that failed to load really does fail the test rather than passing quietly.

The `should_cut_off_a_subscription_whose_certificate_is_not_issued_by_an_advertised_authority` test is the best thing in here: it turns a warning that lived only in three config files into something that goes red the day it stops being true.

ThibaudAV:
Thanks for the approval, and for listing what you actually checked — the `gateway.edge.ssl` `toYaml` point and the `hasKey`-cannot-hit-a-nil-map one are both things I had reasoned about but never wrote down anywhere.

**On finding 2 — the two other render paths do have chart tests.** They were added in `docs(gateway)` after the previous review round, and they are present on the commit you reviewed. At `0fe2f8d`, `helm/tests/gateway/configmap_ssl_test.yaml`:

- line 217, `Client certificate authorities sent on a server of the servers list` → `gateway.servers[].ssl`
- line 241, `Client certificate authorities sent on the Kafka listener` → `gateway.kafka.ssl`

Both were mutation-checked when they landed: shifting the key two spaces in the `$server` block reds exactly the servers case, and typoing the `hasKey` guard reds exactly the Kafka case — nothing else moves. So the gap you describe is closed; no action needed.

**On finding 1 — backports.** Decision taken: **master only**, no `apply-on-*` label on this PR. The support branches stay candidates if the need comes up, so recording the actual state here rather than leaving it implicit:

| APIM branch | `gravitee-node` | What a backport would take |
|---|---|---|
| `master` | 9.8.0 | done, this PR |
| `4.12.x` | 9.7.0 | straight bump — same 9.x line, no node work |
| `4.11.x` | 8.1.0 | backport node#603 to `origin/8.x`, release, then bump |
| `4.10.x` | 7.27.0 | backport node#603 to `origin/7.x`, release, then bump |
| `4.9.x` | 7.27.0 | same as 4.10.x |

One correction to your version reading: 4.10.x and 4.9.x are on **7.27.0**, not on 4.11.x's 8.1.0 line — so covering everything is two separate node backports (`8.x` and `7.x`), not one. Both branches exist upstream, so it is mechanically possible whenever it is wanted.

**On finding 3** — leaving the ticket keys in the test javadoc, as you say there is precedent both ways and the trail is already in the commits.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/plan/mtls/AbstractPlanMutualTLSCertificateAuthoritiesIntegrationTest.java
phiz71: Closing this one. The deferral is the right call and I am not asking for it here — but a review thread is a poor place to track it, and with `required_conversation_resolution` on `master` this was the only thing left blocking a green, approved PR.

Worth a story rather than a comment: the package now carries five copies of the same setup, and the two variants have already drifted on how the certificate is read (`Files.readString(...).getBytes()` in the older classes, `Files.readAllBytes(...)` in the new base). "No behaviour change, four classes moved onto the existing base" is exactly the diff that reviews well on its own.

PR #19350, opened 2026-08-24
test(reporter): extend the data-stream lifecycle coverage to opensearch and the unset case

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14875

## Description

Test-only, and smaller than it started.

The production fix this PR opened with is already on master. #19421 (`fix(authz): index authorization decisions in their own data stream`) put `indexLifecyclePolicyEventMetrics` into `getTemplateData()`, dropped the `rollover_alias` line from the event-metrics templates and escaped the opensearch policy — as collateral to splitting authz decisions out of the shared event-metrics stream, without reference to this ticket. Rebasing this branch onto master left none of the original diff behind.

That commit also brought the first data-stream test coverage, `es_trees_and_data_stream_types`. This PR extends it rather than sitting beside it: the cases here that duplicated it are gone, and what remains fills the gaps it left. The rendered-assertion file ends up shorter than it is on master while covering strictly more.

## What the coverage gains

`IndexTemplateTest` — the data-stream block now runs against **all four trees** instead of es7x/es8x/es9x, keyed per tree (`index.lifecycle.name`, or `index.plugins.index_state_management.policy_id` for opensearch), and against **both** data-stream types, so `AUTHZ_DECISIONS` inherits every new case.

- The opensearch tree is driven through `OpenSearchIndexPreparer` for the first time. It was previously asserted only by reading its templates off the classpath, which proves an interpolation is spelled with `?json_string` but not that the escape applies.
- Escaping is now proved by rendering a policy containing a quote, replacing that classpath check.
- The configured ISM property name is honoured on a data stream. The existing case hard-coded `index.lifecycle.name`, so the override was never exercised on these templates.
- With no policy configured — the state most installations are in — neither key renders at all. `configurationWithPolicies()` always set one, so the unset path was untested.
- Every tree × type × policy-state renders a body that parses as JSON. Every other assertion here is a substring match and cannot see a body broken into invalid JSON by a stray separator, which is the failure a cluster reports only as an opaque parse error.

`IndexPreparerIntegrationTest`

- Reads the stored template back off a real OpenSearch instead of trusting the PUT's 200. A cluster accepts a template whose lifecycle block is missing, so acceptance alone cannot tell a working configuration from a silently dropped one.
- Deletes stored templates between trees. All three push the same template names into one shared container, so without this a tree whose PUT never happened would pass on the previous tree's leftovers.
- Keeps an alias-managed type as a control, through the legacy template API on es7x.

## Tests

Module suite green under `TZ=UTC` against a real Docker daemon, nothing skipped.

Mutation-verified against this head: dropping the `data.put` line fails the rendered assertions; re-adding a `rollover_alias` to the opensearch event-metrics template fails, and so does removing its `?json_string` — neither of which any test on master catches through a rendered body.

## Backports

`apply-on-4-11-x` / `apply-on-4-12-x` are still set but cannot be honoured as they stand. #19421 was not backported, so both branches still omit the policy from the template data and still emit the rollover alias — these assertions would fail there. Those backports need the production change alongside them, either by backporting #19421 or by carrying the fix explicitly. Left open deliberately; not resolved in this PR.

## Related

Helm: #19361 exposes `es.lifecycle.policies.eventMetrics` so chart users can reach the fix. The two are independent — the earlier merge-order note no longer applies, since the gateway already honours the key on master.

### DIFF
--- a/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/mapping/AbstractIndexPreparer.java
+++ b/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/mapping/AbstractIndexPreparer.java
@@ -111,6 +111,7 @@ public abstract class AbstractIndexPreparer implements IndexPreparer {
         data.put("indexLifecyclePolicyMonitor", this.configuration.getIndexLifecyclePolicyMonitor());
         data.put("indexLifecyclePolicyRequest", this.configuration.getIndexLifecyclePolicyRequest());
         data.put("indexLifecyclePolicyLog", this.configuration.getIndexLifecyclePolicyLog());
+        data.put("indexLifecyclePolicyEventMetrics", this.configuration.getIndexLifecyclePolicyEventMetrics());
         data.put("extendedRequestMappingTemplate", this.configuration.getExtendedRequestMappingTemplate());
         data.put("extendedSettingsTemplate", this.configuration.getExtendedSettingsTemplate());
         return data;
--- a/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/resources/freemarker/es7x/mapping/index-template-event-metrics.ftl
+++ b/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/resources/freemarker/es7x/mapping/index-template-event-metrics.ftl
@@ -5,7 +5,6 @@
     "template": {
         "settings": {
             <#if indexLifecyclePolicyEventMetrics?has_content>"${indexLifecyclePolicyPropertyName?json_string}": "${indexLifecyclePolicyEventMetrics?json_string}",</#if>
-            <#if indexLifecyclePolicyEventMetrics?has_content>"${indexLifecycleRolloverAliasPropertyName?json_string}": "${indexName}",</#if>
             "index.number_of_shards":${numberOfShards},
             "index.number_of_replicas":${numberOfReplicas},
             "index.refresh_interval": "${refreshInterval}"
--- a/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/resources/freemarker/es8x/mapping/index-template-event-metrics.ftl
+++ b/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/resources/freemarker/es8x/mapping/index-template-event-metrics.ftl
@@ -5,7 +5,6 @@
     "template": {
         "settings": {
             <#if indexLifecyclePolicyEventMetrics?has_content>"${indexLifecyclePolicyPropertyName?json_string}": "${indexLifecyclePolicyEventMetrics?json_string}",</#if>
-            <#if indexLifecyclePolicyEventMetrics?has_content>"${indexLifecycleRolloverAliasPropertyName?json_string}": "${indexName}",</#if>
             "index.number_of_shards":${numberOfShards},
             "index.number_of_replicas":${numberOfReplicas},
             "index.refresh_interval": "${refreshInterval}"
--- a/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/resources/freemarker/es9x/mapping/index-template-event-metrics.ftl
+++ b/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/resources/freemarker/es9x/mapping/index-template-event-metrics.ftl
@@ -5,7 +5,6 @@
     "template": {
         "settings": {
             <#if indexLifecyclePolicyEventMetrics?has_content>"${indexLifecyclePolicyPropertyName?json_string}": "${indexLifecyclePolicyEventMetrics?json_string}",</#if>
-            <#if indexLifecyclePolicyEventMetrics?has_content>"${indexLifecycleRolloverAliasPropertyName?json_string}": "${indexName}",</#if>
             "index.number_of_shards":${numberOfShards},
             "index.number_of_replicas":${numberOfReplicas},
             "index.refresh_interval": "${refreshInterval}"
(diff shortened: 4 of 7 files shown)


callaertanthony (approved):
Approving — test-only, CI green, and the added coverage is real (the unset-policy case and the opensearch preparer both close genuine holes). Findings below are all non-blocking; #1 and #2 are worth a follow-up because they make part of the new coverage inert.

### 1. The authz-decisions policy set in the `@Bean` method is silently discarded

`IndexPreparerIntegrationTest.java:185`

`ReporterConfiguration`'s fields are `@Value`-annotated and the production bean is built with a bare `new ReporterConfiguration()`, so every value arrives via field injection. Spring runs `createBeanInstance` (the `@Bean` body) before `populateBean`, so anything the factory method sets on a `@Value` field is overwritten afterwards.

`indexLifecyclePolicyAuthzDecisions` is `@Value("${reporters.elasticsearch.lifecycle.policies.authz_decisions:#{null}}")` and that property is not in `@TestPropertySource`, so `setIndexLifecyclePolicyAuthzDecisions("policy-authz-decisions")` resolves back to `null`. The authz-decisions template is PUT with its `<#if ...?has_content>` lifecycle block absent — exactly what the comment on line 185 says must not happen. It is also why line 80 had to be added for `event_metrics`.

Suggested: add `"reporters.elasticsearch.lifecycle.policies.authz_decisions=policy-authz-decisions"` to `@TestPropertySource` and drop lines 186-187 — they're dead, and `setIndexLifecyclePolicyEventMetrics` now reads as if it were the source of the value when the property already is.

### 2. The read-back covers only `event-metrics`

`IndexPreparerIntegrationTest.java:121`

Follows from #1. `AUTHZ_DECISIONS` is the type the description says inherits every new case, but the integration half never reads back `gravitee-ism-override-authz-decisions`, and `deleteStoredTemplates()` doesn't clean it either. A regression that dropped the lifecycle block from only the authz-decisions template would pass all three tree invocations.

### 3. The "valid JSON" guard is weaker than the comment claims

`IndexTemplateTest.java:48`

`new ObjectMapper()` leaves `FAIL_ON_TRAILING_TOKENS` and `STRICT_DUPLICATE_DETECTION` off, and `readTree(String)` only checks trailing tokens when the former is enabled. Two regressions the 80-case matrix would pass:

- a stray `}` closing the root after `settings` — mappings and `_meta` become trailing content and `readTree` returns the truncated object silently, which is precisely the "cluster reports it only as a parse error" case the comment at line 150 names;
- a settings block emitting the lifecycle key twice (a merge leaving both the hard-coded `index.lifecycle.name` and the interpolated `${indexLifecyclePolicyPropertyName}`) — valid JSON, one key silently lost on the cluster.

Enabling both features on the mapper closes it.

### 4. `all_types()` duplicates `Type.TYPES` and will rot silently

`IndexTemplateTest.java:159`

The hard-coded list matches `Type.TYPES` exactly today. But `AUTHZ_DECISIONS` was itself added to that enum recently; the next type added upstream gets a template in all four trees, gets PUT by `indexMapping()` (which iterates `Type.TYPES`), and is excluded from every new assertion here with nothing to signal it. The javadoc's reason — avoiding a tree-specific source — doesn't apply to `Type.TYPES`, which is the exact set the preparer iterates. `Stream.of(Type.TYPES)` would do.

### 5. Two comments this diff made false (nit)

`IndexTemplateTest.java:102` and `IndexTemplateTest.java:127`

Both say opensearch "has no preparer this harness can drive", which is why `should_bound_additional_keyword_metrics_in_every_template_family` reads templates off the classpath. Line 311 of this diff adds `case "opensearch" -> new OpenSearchIndexPreparer(...)`, so the premise is gone. Either update both, or — consistent with what this PR did to the escape test — render the opensearch tree there too and delete the classpath variant.

### 6. The matrix omits the arrangement this class was written for (nit)

`IndexTemplateTest.java:174`

`all_trees_and_all_types_and_policy_state` varies the policy between set and unset but always leaves `indexLifecyclePolicyPropertyName` at its default. The AWS-OpenSearch-compat arrangement from `IndexPreparerIntegrationTest`'s javadoc — a user-supplied property *name* interpolated as a JSON key — is never parsed as JSON, only substring-asserted on a single tree/type. A third policy state setting the property-name override would cover the riskiest key in the body.

---

Checked and sound:

- `legacyApi = "es7x".equals(esDir) ? "_template" : "_index_template"` is right — `ES7IndexPreparer.useOldClient(dataStream)` returns `!dataStream`, so only es7x's dated types use the legacy API and data streams use `_index_template` everywhere.
- Excluding opensearch from `should_render_the_data_stream_policy_under_the_configured_property_name` is correct, not a masked inconsistency: that tree hard-codes `index.plugins.index_state_management.policy_id` in all its templates.
- The escape assertion isn't vacuous — FreeMarker's `JSON` output format does no auto-escaping, so `?json_string` is load-bearing.
- `should_render_no_lifecycle_settings_...` can fail: dropping `?has_content` renders `"index.lifecycle.name": ""` and trips the `doesNotContain`.
- No coverage lost by replacing the classpath escape check; the rendered version subsumes existence, escaping and the rollover-alias ban.
- `isIlmManagedIndex()` is false in the integration test (`index_mode` defaults to `daily`), so `ensureAlias` never runs and the added `deleteStoredTemplates()` has no alias leftovers to contend with.

paulatulis:
Thanks — all six applied, and the branch is rebased onto master (now `bd3b3f6`).

1. Fixed. `authz_decisions` now arrives through `@TestPropertySource` and the two setter calls are gone. Your diagnosis checks out against the cluster: dropping that one property line makes the new read-back fail on all three trees, and the stored `gravitee-ism-override-authz-decisions` template comes back with a `settings` block carrying shards/replicas/refresh and no lifecycle key at all.
2. Both data-stream templates are read back now, and `deleteStoredTemplates()` cleans `authz-decisions` too.
3. The mapper enables `FAIL_ON_TRAILING_TOKENS` and `STRICT_DUPLICATE_DETECTION`, with the two regressions it closes noted on the field.
4. `all_types()` is `Stream.of(Type.TYPES)`.
5. Took your second option rather than fixing the comments: the keyword-metrics assertion now renders the opensearch tree through its preparer, and the classpath variant and its stale premise are gone.
6. Added a third policy state to the JSON matrix that sets both property-name overrides, so the user-supplied key is parsed as JSON on every tree and type.

PR #19351, opened 2026-08-24
feat(ci): release a hotfix from a branch cut off a released tag

phiz71:
## Issue

https://gravitee.atlassian.net/browse/BX-347

## Description

Nothing needs this today. It is a capability we do not have, for a situation that will eventually come up: an urgent fix has to ship, and the support branch already carries work that is not ready to go out with it. The only answer available right now is to move commits around on the branch that serves customers — back it up, rewrite the history, put it back — while someone waits. That is the worst possible moment to be inventing a procedure.

The way out is to leave the support branch alone: cut a branch from the released tag, add the one commit that matters, release from there, and cherry-pick the fix back afterwards.

**Most of the machinery already handled it.** The release job commits and tags on the branch the pipeline runs on; the version parser reads `4.12.17-hotfix.1` correctly; a qualified version leaves the `4.12` and `latest` Docker tags untouched, so a hotfix cannot overwrite the minor; and the bump that follows a release moves the branch to `-hotfix.2`, the same way it does for alpha and milestone. Maven orders the versions the way the scheme needs — verified against `ComparableVersion`: `4.12.17 < 4.12.17-hotfix.1 < 4.12.17-hotfix.2 < 4.12.18`, since an unknown qualifier sorts after the release.

Three things were missing.

**The pipeline refused the branch.** `generateReleaseConfig` threw for anything that did not match `^\d+\.\d+\.x$`.

**And the branch had to be prepared, which is the part that is easy to get wrong.** The published version comes from the poms — `<revision><sha1><changelist>` — not from the `--version` passed to the release, which only feeds the commit message, the tag and the UI `build.json` files. A branch cut from a tag therefore sits at the released version with an empty `<changelist>`: releasing it untouched would **republish `4.12.17` under a tag claiming `4.12.17-hotfix.1`**, without failing. `yarn prepare_hotfix --version=4.12.17` cuts the branch and sets the qualifier in `<sha1>`, the slot pre-releases already use.

**And a hotfix shipped no RPM.** The publish job replaces its steps with an `echo` for any qualified version — a guard added in March 2025, after an alpha published into the production repository outranked the release everyone was running. A hotfix is the one qualified version that belongs there: its RPM version is the released one and only the release number moves, so it sits above the release it fixes and below the next patch. The job now makes that exception. The encoding it relies on landed in [gravitee-io/packages#29](https://github.com/gravitee-io/packages/pull/29).

### How to review this

Five commits, each standing on its own.

The first two are unrelated to the feature and can be read first: a dead branch removed from the version resolver, and a rewrite of the release README, which had drifted from the scripts — it documented a `yarn docker_rpms` that does not exist, a Node version three majors behind `.nvmrc`, and no `maven_release` at all.

Then the CI side, then the tooling. That order is deliberate: the command prints the release line to run next, and that line relies on the branch being derived from the version, which the CI commit introduces. Reversed, the command would advertise something that does not work yet. The README section documenting the whole flow rides with the command, so no commit documents what it does not contain.

The RPM comes last and stands alone. It touches one job and nothing else reads it.

## Additional context

**Verified by running it**, not by reading: the script was exercised on a throwaway clone against the real `4.12.17` tag. It leaves the branch at `4.12.17-hotfix.1-SNAPSHOT`, which the release turns into `4.12.17-hotfix.1` — no collision with the published `4.12.17`.

That run found two defects worth mentioning, both fixed here. The script runs from `release/`, so relative paths resolved against the wrong directory and every edit was silently skipped — a run that edits nothing now fails rather than committing a branch that would republish. And the distribution pom only carries a version triplet of its own since the reactor cut, which is master-only: on a support branch it is still a module and inherits the root's, so it is left alone there. Since hotfixes are cut from support-branch tags, that is the common case rather than the exception.

**Two things stay manual, deliberately.** Cherry-picking the fix into the support branch is a judgement call about what belongs in the next patch. And deleting the hotfix branch once the tag exists — the tag is what is kept.

**Noticed, not addressed:** only one of the nine release pipelines checks the branch it runs on, and it is not the one that matters most. `generateFullReleaseConfig` — the `full_release` action, the path the README recommends — has no branch check at all, so the guard added here covers `release` only. Nothing prevents triggering `nexus_staging` or `build_docker_images` from anywhere either. Worth its own ticket.

**Security-sensitive**: this changes CI/CD configuration and adds a script that writes version metadata and commits. It publishes nothing on its own and reads no secret beyond the `CIRCLECI_TOKEN` the other release scripts already use.

### DIFF
--- a/.circleci/ci/src/pipelines/pipeline-release.ts
+++ b/.circleci/ci/src/pipelines/pipeline-release.ts
@@ -16,13 +16,35 @@
 import { Config } from '../circleci-config';
 import { ReleaseWorkflow } from '../workflows';
 import { CircleCIEnvironment } from './circleci-environment';
-import { isSupportBranch } from '../utils';
+import { isHotfixBranch, isSupportBranch } from '../utils';
 import { initDynamicConfig } from './config-factory';
 
-export function generateReleaseConfig(environment: CircleCIEnvironment): Config {
-  if (!isSupportBranch(environment.branch)) {
-    throw new Error('Release is only supported on support branches');
+const HOTFIX_QUALIFIER = /-hotfix\.\d+$/;
+
+/**
+ * The version and the branch reach this pipeline as two independent inputs, and nothing downstream
+ * brings them together: the published version comes from the poms of whichever branch runs, while
+ * the tag comes from the version. A mismatch therefore tags one version and publishes another,
+ * without failing. A hotfix is where the two diverge most easily, since `hotfix/4.12.17` and
+ * `4.12.17-hotfix.1` name the same release twice.
+ */
+function assertVersionMatchesBranch(version: string, branch: string): void {
+  if (isHotfixBranch(branch)) {
+    const releasedVersion = branch.substring('hotfix/'.length);
+    if (!version.startsWith(`${releasedVersion}-`)) {
+      throw new Error(`${branch} releases ${releasedVersion} with a qualifier, not ${version}`);
+    }
+  } else if (HOTFIX_QUALIFIER.test(version)) {
+    throw new Error(`${version} is only released from hotfix/${version.replace(HOTFIX_QUALIFIER, '')}, not from ${branch}`);
   }
+}
+
+export function generateReleaseConfig(environment: CircleCIEnvironment): Config {
+  if (!isSupportBranch(environment.branch) && !isHotfixBranch(environment.branch)) {
+    throw new Error('Release is only supported on a support branch (X.Y.x) or a hotfix branch (hotfix/X.Y.Z)');
+  }
+
+  assertVersionMatchesBranch(environment.graviteeioVersion, environment.branch);
 
   const dynamicConfig = initDynamicConfig();
   const workflow = ReleaseWorkflow.create(dynamicConfig, environment);
--- a/.circleci/ci/src/pipelines/tests/pipeline-release.spec.ts
+++ b/.circleci/ci/src/pipelines/tests/pipeline-release.spec.ts
@@ -43,7 +43,7 @@ describe('Release tests', () => {
     },
   );
 
-  it('should throw error when branch is not a support branch', () => {
+  it('should throw error when branch is neither a support nor a hotfix branch', () => {
     expect.assertions(1);
 
     try {
@@ -60,7 +60,47 @@ describe('Release tests', () => {
         isDryRun: false,
       });
     } catch (e) {
-      expect(e).toStrictEqual(new Error('Release is only supported on support branches'));
+      expect(e).toStrictEqual(new Error('Release is only supported on a support branch (X.Y.x) or a hotfix branch (hotfix/X.Y.Z)'));
     }
+  });
+
+  it('should build release config on a hotfix branch cut from a released tag', () => {
+    const result = generateReleaseConfig({
+      action: 'release',
+      sha1: '784ff35ca',
+      changedFiles: [],
+      buildNum: '1234',
+      buildId: '1234',
+      graviteeioVersion: '4.2.0-hotfix.1',
+      baseBranch: '4.2.x',
+      branch: 'hotfix/4.2.0',
+      apimVersionPath: './src/pipelines/tests/resources/common/pom-snapshot.xml',
+      isDryRun: false,
+    });
+
+    const expected = fs.readFileSync('./src/pipelines/tests/resources/release/release-4-2-0-hotfix.yml', 'utf-8');
+    expect(result.stringify()).toStrictEqual(expected);
+  });
+
+  it.each`
+    graviteeioVersion   | branch            | expectedError
+    ${'4.2.0'}          | ${'hotfix/4.2.0'} | ${'hotfix/4.2.0 releases 4.2.0 with a qualifier, not 4.2.0'}
+    ${'4.2.1-hotfix.1'} | ${'hotfix/4.2.0'} | ${'hotfix/4.2.0 releases 4.2.0 with a qualifier, not 4.2.1-hotfix.1'}
+    ${'4.2.0-hotfix.1'} | ${'4.2.x'}        | ${'4.2.0-hotfix.1 is only released from hotfix/4.2.0, not from 4.2.x'}
+  `('should throw for version $graviteeioVersion on branch $branch', ({ graviteeioVersion, branch, expectedError }) => {
+    expect(() =>
+      generateReleaseConfig({
+        action: 'release',
+        sha1: '784ff35ca',
+        changedFiles: [],
+        buildNum: '1234',
+        buildId: '1234',
+        graviteeioVersion,
+        baseBranch: '4.2.x',
+        branch,
+        apimVersionPath: './src/pipelines/tests/resources/common/pom-snapshot.xml',
+        isDryRun: false,
+      }),
+    ).toThrow(expectedError);
   });
 });
--- a/release/README.adoc
+++ b/release/README.adoc
@@ -1,129 +1,97 @@
-= Release process
+= Releasing APIM
 
-== 📝 Description
+The scripts in this folder do not build anything. Each one is a *trigger*: it posts to the CircleCI API and hands over. Understanding that explains most of what follows — in particular why the branch you are standing on has no effect on what gets released.
 
-This repo contains some https://github.com/google/zx[zx] scripts to execute the different release steps for APIM.
+== What decides what
 
-== 🛠 Requirement
+Three things, and they are not the same thing.
 
-To run the scripts, you need to have:
+[cols="1,3"]
+|===
+| Where the release runs
+| Derived from `--version`: `4.12.18` releases from `4.12.x`. Never from your current branch — no script reads your local git state. `--branch` overrides it, and you are asked to confirm when it differs from the derived one.
 
-* Node.js >= 16.0.0 (configured in link:.nvmrc[.nvmrc] with v18.4.0)
-* https://github.com/google/zx[zx], to install it you just have to run:
+| Which version is published
+| The poms, through `<revision><sha1><changelist>`. The release clears `<changelist>` to drop `-SNAPSHOT`, and publishes what is left. `--version` does *not* set it — it feeds the commit message, the git tag and the UI `build.json` files.
+
+| What happens next
+| After releasing, the branch is bumped to the next version. `4.12.18` leaves it at `4.12.19-SNAPSHOT`; a qualified version like `4.13.0-alpha.1` leaves it at `4.13.0-alpha.2-SNAPSHOT`.
+|===
+
+The second row is the one that surprises people. If a branch's poms and the `--version` you pass disagree, the tag will say one thing and the artefacts another, without failing.
+
+== Setup
+
+* Node — the version in link:../.nvmrc[.nvmrc]. `nvm use` in this folder picks it up.
+* https://github.com/google/zx[zx], globally: `npm i -g zx`.
+* Dependencies: `yarn install`, here in `release/`.
+* A CircleCI personal token, in a `.env` file next to this README:
++
+[source,shell]
+----
+echo "CIRCLECI_TOKEN=<your token>" > .env
+----
++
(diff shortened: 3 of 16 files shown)


release/steps/prepare_hotfix.mjs
```diff
@@ -0,0 +1,119 @@
...
+    [/^appVersion:.*$/m, `appVersion: ${nextRelease}`],
+  ],
+  { required: true },
+);
+
+if (edited === 0) {
+  console.log(chalk.red('Nothing was edited. Refusing to commit a branch that would republish the released version.'));
+  process.exit(1);
+}
+
```
phiz71: Correct, and removed.

The counter came from a real defect found by running the script: it executes from `release/`, so the relative paths resolved against the wrong directory and every edit was silently skipped. But the fix for that was the `cd` to the repository root, plus two guards inside `edit()` — absence of a file the version depends on, and a file that is present but does not match. Both exit. The counter sat on top of those and could never fire.

Verified by running it rather than reading it: a pom without `<sha1>` now aborts with `Nothing to replace in pom.xml`, before anything is committed.

release/README.adoc
```diff
@@ -1,129 +1,148 @@
...
+Then add the fix as one commit, and release from there:
 
-Thanks to link:package.json[package.json], you can run those scripts as simple `yarn` commands (do not forget to use `yarn install` before starting).
+[source,shell]
+----
+yarn full_release --version=4.12.17-hotfix.1 --branch=hotfix/4.12.17
+----
 
-Each of the commands needs:
+The support branch is untouched throughout, and the Docker `4.12` and `latest` tags are left where
```
ytvnr: **⚠️ Important** — `4.12.17-hotfix.1` carries a qualifier, and `PublishRpmPackagesJob` replaces its whole publish step with an echo for any qualified version, so the RPM listed in the acceptance criteria never ships.

`.circleci/ci/src/jobs/job-publish-rpm-packages.ts:30` branches on `parsedGraviteeioVersion.qualifier.full` and, when it is non-empty, emits a single `Run` printing "Publishing RPMs is not supported for prerelease version". The job is wired into the full-release workflow unconditionally (`workflows/workflow-full-release.ts:86,304`), so the hotfix pipeline goes green while producing no RPM. Docker, bundle, Nexus and Helm have no such gate, only RPM does.

Worth deciding whether a `hotfix` qualifier should count as a release rather than a prerelease in that job, and saying in the README which artefacts a hotfix actually produces.

.circleci/ci/src/pipelines/pipeline-release.ts
```diff
@@ -16,12 +16,12 @@
...
 import { ReleaseWorkflow } from '../workflows';
 import { CircleCIEnvironment } from './circleci-environment';
-import { isSupportBranch } from '../utils';
+import { isHotfixBranch, isSupportBranch } from '../utils';
 import { initDynamicConfig } from './config-factory';
 
 export function generateReleaseConfig(environment: CircleCIEnvironment): Config {
-  if (!isSupportBranch(environment.branch)) {
-    throw new Error('Release is only supported on support branches');
+  if (!isSupportBranch(environment.branch) && !isHotfixBranch(environment.branch)) {
```
ytvnr, line 43: **Worth fixing** — Nothing cross-checks the version against the branch, and `branch('4.12.17-hotfix.1')` derives `4.12.x`, so `yarn full_release --version=4.12.17-hotfix.1` without `--branch` releases from the support branch, the exact outcome the feature exists to avoid.

`getTargetBranch` (`release/helpers/option-helper.mjs:11`) falls back to `versions.branch`, which is `4.12.x` for any `-hotfix.N` version. The confirmation prompt in `full_release.mjs:47` only fires when `--branch` *is* passed and differs, so the dangerous case is the silent one. The support branch poms sit at `4.12.18-SNAPSHOT`, so the run would tag `4.12.17-hotfix.1` over `4.12.18` artefacts, the mismatch the README's own second table row warns about.

`generateReleaseConfig` has both `environment.branch` and `graviteeioVersion` in hand and checks neither, so the guard could reject a `hotfix` qualifier that does not match the branch it is running on.

release/steps/prepare_hotfix.mjs
```diff
@@ -0,0 +1,112 @@
...
+edit(
+  'helm/Chart.yaml',
+  [
+    [/^version:.*$/m, `version: ${nextRelease}`],
+    [/^appVersion:.*$/m, `appVersion: ${nextRelease}`],
+  ],
+  { required: true },
+);
+
+await $`git add --update`;
```
ytvnr, line 159: **Worth fixing** — The script never checks the working tree is clean, and `git add --update` stages every tracked modification, so unrelated local work is carried onto the hotfix branch and committed into the prepare commit.

`git switch -c` at line 39 carries uncommitted changes across, and this line stages all tracked modifications rather than the files `edit()` touched.

The same gap shows on the abort path: `edit()` exits at line 61 after the branch was already created and earlier files rewritten, so "Aborting rather than half-preparing the branch" describes a state the script does leave behind, uncommitted, with the user standing on the new branch.

Refusing to start on a dirty tree, and staging explicitly what the script edited, would close both.

release/steps/prepare_hotfix.mjs
```diff
@@ -0,0 +1,112 @@
...
+  edit(buildJson, [[/"version": ".*"/, `"version": "${nextRelease}-SNAPSHOT"`]]);
+}
+
+edit(
+  'helm/Chart.yaml',
+  [
+    [/^version:.*$/m, `version: ${nextRelease}`],
+    [/^appVersion:.*$/m, `appVersion: ${nextRelease}`],
+  ],
+  { required: true },
```
ytvnr, line 156: **Worth fixing** — `4.12.17-hotfix.1` is a semver prerelease and a Maven qualifier, so both Helm and Maven order it below `4.12.17`, and the fixed chart is never selected over the broken one by any resolution that is not pinned.

The qualifier deliberately reuses the `<sha1>` slot that alpha and milestone prereleases use (comment at line 8-10), and `helm/Chart.yaml` gets `version: 4.12.17-hotfix.1`. Helm hides prereleases from `helm search` and from upgrade unless `--devel` is passed, and Maven ranges rank the qualifier below the plain release.

Not a code defect, but the consequence belongs in the README's Hotfixes section: consumers have to pin the hotfix version explicitly.

release/README.adoc
```diff
@@ -1,129 +1,148 @@
...
 
-* A `CIRCLECI_TOKEN` in `.env` file
-* the `--version` parameter
+Two things are on you afterwards. The fix has to be cherry-picked into the support branch, so the
+next patch carries it too; deciding what belongs in that patch is a judgement call, not a step that
+can be automated. And the hotfix branch can be deleted once the tag exists — the tag is what is
+kept.
 
-Optional flag:
+`--hotfix=2` prepares a second one from the same tag, if the first was not enough.
```
ytvnr: **Worth fixing** — This offers `--hotfix=2` for a second hotfix from the same tag, but the branch name does not carry the number, so `git switch -c hotfix/4.12.17` fails on the branch the first run created.

`branch` is `hotfix/${releasedVersion}` (`prepare_hotfix.mjs:22`), independent of `hotfixNumber` (line 21), so a second run aborts on an unhandled zx failure. And if the branch was deleted first, cutting again from the tag drops the first hotfix's commit while numbering the result `.2`.

The release job already bumps the branch to `-hotfix.2-SNAPSHOT` after publishing `.1`, so the supported second hotfix looks like another commit on the existing branch, which is not what this line says.

.circleci/ci/src/pipelines/tests/pipeline-release.spec.ts
```diff
@@ -60,7 +60,24 @@ describe('Release tests', () => {
...
+      buildNum: '1234',
+      buildId: '1234',
+      graviteeioVersion: '4.2.0-hotfix.1',
+      baseBranch: '4.2.x',
+      branch: 'hotfix/4.2.0',
+      apimVersionPath: './src/pipelines/tests/resources/common/pom-snapshot.xml',
+      isDryRun: false,
+    });
+
+    expect(result).toBeDefined();
```
ytvnr: The hotfix test asserts `toBeDefined()` while every neighbouring case compares the generated config to a checked-in YAML, so nothing verifies what the hotfix pipeline actually generates.

The parameterised block at line 20-42 asserts `result.stringify()` against `resources/release/*.yml`; this case does not, so a config that pushes to the wrong branch or drops a job would still pass.

`isHotfixBranch` also has no entry in `utils/tests/branch.spec.ts`, which has a table-driven `describe` for each of the other predicates, so lookalikes such as `hotfix/4.12` or `feature/hotfix/4.12.17` are untested.

ytvnr (commented):
This adds a hotfix release path: a branch cut from a released tag, prepared by `yarn prepare_hotfix`, and accepted by the release pipeline through a new `isHotfixBranch` guard. The flow holds together and the README rewrite is a real improvement, so most of what follows is about the edges rather than the design.

Two points I would want settled before this is used under pressure. A qualified version turns `PublishRpmPackagesJob` into an echo, so a hotfix pipeline goes green without producing the RPM the ticket lists among the artefacts. And nothing cross-checks the version against the branch: `yarn full_release --version=4.12.17-hotfix.1` with no `--branch` resolves back to the support branch and releases from there, which is the outcome this feature exists to prevent.

The rest is smaller: a dirty-tree guard on the script, and two README statements that do not match what the code does.

Dropping the unreachable `master` case in the version resolver checks out, the `split[2] === 0` comparison was genuinely dead and `getTargetBranch` is the only consumer.

phiz71:
Closing the review threads — here is what each one led to, since the fixes were pushed without a reply on the line.

| Finding | Outcome |
| --- | --- |
| Nothing cross-checks the version against the branch | `assertVersionMatchesBranch` in the release pipeline, and `getTargetBranch` now derives the hotfix branch from the version instead of falling back to the support branch |
| The script never checks the working tree | It refuses to start on a dirty tree — `git switch` would carry the changes onto the branch and commit them with the version |
| `4.12.17-hotfix.1` is a semver prerelease | Documented rather than worked around: there is no semver version between a release and the next patch, and build metadata is ignored in comparisons. The README now covers what that means for `helm search`, for range constraints, and for ArgoCD |
| `--hotfix=2` for a second hotfix | The flag is gone. The release bumps the qualifier on the branch it ran on, so a second hotfix needs no preparation at all — re-running `prepare_hotfix` checks the branch out as it stands |
| The hotfix test asserts `toBeDefined()` | It compares against a fixture like its neighbours, and `isHotfixBranch` has its own table of cases in `branch.spec.ts` |
| A hotfix ships no RPM | Fixed in the last commit. The publish job makes an exception for the `hotfix` qualifier, and the version encoding it relies on landed in [gravitee-io/packages#29](https://github.com/gravitee-io/packages/pull/29) |

The ordering claim in the first review was the one thing that did not hold: Maven sorts an unknown qualifier *after* the release, so `4.12.17 < 4.12.17-hotfix.1`. Helm sorts the same string the other way. Both behaviours are now in the README.

ytvnr (approved):
Approving. I went back through the five commits and the resolutions hold.

What I checked rather than took on trust:

- The version/branch mismatch is closed twice over, in `assertVersionMatchesBranch` and in `getTargetBranch`, and the three rejection cases are covered by the parameterised test. Dropping `--branch` from the documented command follows from it rather than being a separate claim.
- `prepare_hotfix` refuses a dirty tree before it creates anything, and a second run on an existing branch checks it out instead of resetting the qualifier to `.1`. The `--hotfix` flag going away is the better answer to what I raised.
- The RPM exception is scoped to the `hotfix` qualifier alone, so pre-releases keep their guard, and both the build-rpm and full-release pipelines now carry a hotfix fixture rather than a `toBeDefined()`.
- `isHotfixBranch` has its own table, lookalikes included.

You were right about the ordering and I was not. Maven sorts an unknown qualifier after the release, which the ticket already stated and I contradicted without checking it. The Helm half stands, and having both behaviours in the README is a better outcome than the note I asked for.

One thread still open, and it is not mine: the Aikido comment about an early return on `job-publish-rpm-packages.ts:36`. Style only, your call.

PR #19352, opened 2026-08-24
feat(portal): add API Product subscription lifecycle actions

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-135

## Summary

Add Pause, Resume, and Close lifecycle actions for API Product subscriptions in Portal Next.

## Changes

- Add Pause and Resume actions using the existing consumer status endpoint.
- Add Close action with an explicit confirmation dialog.
- Display actions according to the subscription status and application permissions.
- Disable lifecycle actions while a request is pending.
- Refresh subscription details and display inline success or error feedback.
- Disable API access and hide credentials while the consumer status is stopped.
- Validate the API Product subscription environment and allowed states in Portal API.
- Preserve the existing audit and API key revocation behavior.
- Keep the existing behavior of direct API subscriptions unchanged.
- Document the invalid API Product subscription state response in the existing OpenAPI contract.

Uploading Nagranie z ekranu 2026-08-25 o 09.21.40.mov…

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/subscription-details/api-product-subscription-details/api-product-subscription-details.component.html
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/subscription-details/api-product-subscription-details/api-product-subscription-details.component.html
@@ -15,7 +15,73 @@
     limitations under the License.
 
 -->
-<h1 class="next-gen-h3" i18n="@@apiProductSubscriptionDetailsTitle">Subscription Details</h1>
+<div class="subscription-header">
+  <h1 class="next-gen-h3" i18n="@@apiProductSubscriptionDetailsTitle">Subscription Details</h1>
+  <div class="subscription-actions" [attr.aria-busy]="isActionPending()">
+    @if (canPause()) {
+      <button
+        mat-stroked-button
+        type="button"
+        data-testid="pause-subscription"
+        [disabled]="isActionPending()"
+        (click)="pauseSubscription()"
+        i18n="@@apiProductSubscriptionPause"
+      >
+        Pause
+      </button>
+    }
+    @if (canResume()) {
+      <button
+        mat-stroked-button
+        type="button"
+        data-testid="resume-subscription"
+        [disabled]="isActionPending()"
+        (click)="resumeSubscription()"
+        i18n="@@apiProductSubscriptionResume"
+      >
+        Resume
+      </button>
+    }
+    @if (canRetry()) {
+      <button
+        mat-stroked-button
+        type="button"
+        data-testid="retry-subscription"
+        [disabled]="isActionPending()"
+        (click)="retrySubscription()"
+        i18n="@@apiProductSubscriptionRetry"
+      >
+        Retry subscription
+      </button>
+    }
+    @if (canClose()) {
+      <button
+        mat-stroked-button
+        color="warn"
+        type="button"
+        data-testid="close-subscription"
+        [disabled]="isActionPending()"
+        (click)="closeSubscription()"
+        i18n="@@apiProductSubscriptionClose"
+      >
+        Close
+      </button>
+    }
+  </div>
+</div>
+
+@if (actionFeedback(); as feedback) {
+  <p
+    class="action-feedback"
+    data-testid="subscription-action-feedback"
+    [class.action-feedback--error]="feedback.type === 'error'"
+    [class.action-feedback--success]="feedback.type === 'success'"
+    [attr.role]="feedback.type === 'error' ? 'alert' : 'status'"
+    [attr.aria-live]="feedback.type === 'success' ? 'polite' : null"
+  >
+    {{ feedback.message }}
+  </p>
+}
 
 @if (product(); as productDetails) {
   <mat-card appearance="outlined">
@@ -24,7 +90,10 @@
     </mat-card-header>
     <mat-card-content class="details-grid" data-testid="product-subscription-summary">
       <span class="details-grid__label" i18n="@@apiProductSubscriptionStatus">Status</span>
-      <span>{{ subscription().status | capitalizeFirst }}</span>
+      <span>{{ displayedSubscription().status | capitalizeFirst }}</span>
+
+      <span class="details-grid__label" i18n="@@apiProductSubscriptionAccessStatus">Access status</span>
+      <span>{{ accessStatusLabel() }}</span>
 
       <span class="details-grid__label" i18n="@@apiProductSubscriptionProduct">API Product</span>
       <span
@@ -40,29 +109,29 @@
       } @else if (applicationResource.isLoading()) {
         <span i18n="@@apiProductSubscriptionApplicationLoading">Loading application...</span>
       } @else {
-        <span>{{ subscription().application }}</span>
+        <span>{{ displayedSubscription().application }}</span>
       }
 
       <span class="details-grid__label" i18n="@@apiProductSubscriptionId">Subscription ID</span>
-      <span>{{ subscription().id }}</span>
+      <span>{{ displayedSubscription().id }}</span>
 
       <span class="details-grid__label" i18n="@@apiProductSubscriptionPlan">Plan</span>
-      <span>{{ productDetails.plan?.name || productDetails.plan?.id || subscription().plan }}</span>
+      <span>{{ productDetails.plan?.name || productDetails.plan?.id || displayedSubscription().plan }}</span>
 
       @if (planSecurityLabel()) {
         <span class="details-grid__label" i18n="@@apiProductSubscriptionAuthentication">Authentication</span>
         <span>{{ planSecurityLabel() }}</span>
       }
 
-      @if (subscription().created_at; as createdAt) {
+      @if (displayedSubscription().created_at; as createdAt) {
         <span class="details-grid__label" i18n="@@apiProductSubscriptionCreatedAt">Created at</span>
         <span>{{ createdAt | date: 'medium' }}</span>
       }
-      @if (subscription().start_at; as startAt) {
+      @if (displayedSubscription().start_at; as startAt) {
         <span class="details-grid__label" i18n="@@apiProductSubscriptionStartsAt">Starts at</span>
         <span>{{ startAt | date: 'medium' }}</span>
       }
-      @if (subscription().end_at; as endAt) {
+      @if (displayedSubscription().end_at; as endAt) {
         <span class="details-grid__label" i18n="@@apiProductSubscriptionEndsAt">Ends at</span>
         <span>{{ endAt | date: 'medium' }}</span>
       }
@@ -97,8 +166,8 @@
       <mat-card-title class="next-gen-h4" i18n="@@apiProductSubscriptionCredentialsTitle">API access</mat-card-title>
     </mat-card-header>
     <mat-card-content class="credentials">
-      @if (subscription().status !== 'ACCEPTED') {
-        @switch (subscription().status) {
+      @if (displayedSubscription().status !== 'ACCEPTED') {
+        @switch (displayedSubscription().status) {
           @case ('PENDING') {
             <p i18n="@@apiProductSubscriptionPending">Your subscription request is being reviewed.</p>
           }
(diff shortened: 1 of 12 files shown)


gravitee-apim-portal-webui-next/src/app/dashboard/subscription-details/api-product-subscription-details/api-product-subscription-details.component.html
```diff
@@ -97,8 +154,10 @@ <h1 class="next-gen-h3" i18n="@@apiProductSubscriptionDetailsTitle">Subscription
       <mat-card-title class="next-gen-h4" i18n="@@apiProductSubscriptionCredentialsTitle">API access</mat-card-title>
     </mat-card-header>
     <mat-card-content class="credentials">
-      @if (subscription().status !== 'ACCEPTED') {
-        @switch (subscription().status) {
+      @if (displayedSubscription().consumerStatus === 'STOPPED') {
```
FuYu3699: Pause then close leaves this card saying "This subscription is paused." while the header says Closed. `SubscriptionEntity.close()` sets status to CLOSED and never touches consumerStatus, and this template checks `consumerStatus === 'STOPPED'` before status; `accessStatusLabel()` uses the same order. The ticket requires the updated state right after each action, and close is irreversible, so I grade this a blocker even though the fix is small.

Suggested shape (untested), here and in `accessStatusLabel()`:

```html
@if (displayedSubscription().status !== 'ACCEPTED') {
  <!-- closed / rejected / pending copy -->
} @else if (displayedSubscription().consumerStatus === 'STOPPED') {
  <!-- paused copy -->
}
```

The close spec passes today because its fixture keeps `consumerStatus: STARTED`; a pause-then-close case with the fixture on STOPPED would cover the real sequence.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
```diff
@@ -251,4 +265,18 @@ public Response resumeFailedSubscription(@PathParam("subscriptionId") String sub
     public SubscriptionKeysResource getSubscriptionKeysResource() {
         return resourceContext.getResource(SubscriptionKeysResource.class);
     }
+
+    private static boolean isApiProductSubscription(SubscriptionEntity subscription) {
+        return SubscriptionReferenceType.API_PRODUCT.name().equals(subscription.getReferenceType());
+    }
+
+    private static void validateApiProductEnvironment(
```
FuYu3699, line 276: This helper runs on three of the resource's five operations. `updateSubscription` and `resumeFailedSubscription` check only the application permission, which is not environment-scoped, so an API Product subscription from another environment still passes them. Not an open door, since the permission is still required. But the ticket asks for environment ownership enforced in the backend, and a half-applied rule reads as optional to the next editor.

Suggested (untested), in both methods:

```java
validateApiProductEnvironment(executionContext, subscriptionEntity, isApiProductSubscription(subscriptionEntity));
```

with a test per method shaped like `should_not_close_api_product_subscription_from_another_environment`. A follow-up ticket noted here would work too.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
```diff
@@ -151,11 +154,16 @@ private boolean hasReadPermission(ExecutionContext executionContext, Subscriptio
...
         final ExecutionContext executionContext = GraviteeContext.getExecutionContext();
-        if (hasPermission(executionContext, APPLICATION_SUBSCRIPTION, subscriptionEntity.getApplication(), RolePermissionAction.DELETE)) {
-            closeSubscriptionUsecase.execute(new CloseSubscriptionUseCase.Input(subscriptionId, getAuditInfo()));
-            return Response.noContent().build();
+        boolean apiProductSubscription = isApiProductSubscription(subscriptionEntity);
+        validateApiProductEnvironment(executionContext, subscriptionEntity, apiProductSubscription);
+        if (!hasPermission(executionContext, APPLICATION_SUBSCRIPTION, subscriptionEntity.getApplication(), RolePermissionAction.DELETE)) {
+            throw new ForbiddenAccessException();
+        }
+        if (apiProductSubscription && !API_PRODUCT_CLOSABLE_STATUSES.contains(subscriptionEntity.getStatus())) {
```
FuYu3699, line 163: Question: is removing pending-request withdrawal intended? `_close` on a PENDING subscription previously reached `CloseSubscriptionDomainService`, which treats it as withdrawing the request and has an API Product branch for that case. Now an API Product consumer gets a 400 with no way to cancel a pending request, while a direct API consumer still can. If intended, a line in the PR description makes it a visible product decision; if not, letting PENDING through to the domain service restores the old behaviour. Also, this 400 carries no body while the endpoint's other error paths return an ErrorResponse, so the portal can only show a generic failure message.

FuYu3699 (commented):
**This is a test code review to evaluate the gravitee-ai-foundation review skills.**

Reviewed against PORTAL-135 (https://gravitee.atlassian.net/browse/PORTAL-135) and the repository's code-review instructions. The lifecycle flows, permission gating, cross-environment 404s, audit and key-revocation path, and unchanged direct-API behaviour all hold up; test coverage is thorough.

Three inline comments are worth a look before merge: the pause-then-close display state, environment-check coverage on two operations, and a question about pending-request withdrawal.

Smaller items, take or leave:

- No test fails if the three `[disabled]="isActionPending()"` bindings are removed. A test whose request never completes would cover the disable-while-running criterion. REJECTED also lacks an actions-absent assertion; CLOSED has one.
- `aria-live="polite"` on the same element as `role="alert"` downgrades error announcements. The api-keys-list component binds the two conditionally; same pattern would work here.
- `getSubscriptionBySubscriptionId` still inlines the comparison the new `isApiProductSubscription` helper wraps.
- `@@apiProductSubscriptionConsumerPaused` and `@@apiProductSubscriptionPaused` render identical copy. Once state precedence is fixed, both are reachable and the user cannot tell a self-pause from a publisher pause.

Notes only, no action asked. The parent component's subscription resource is not refreshed after an action, so a later parent reload would show the older state. The new service tests fix in place that API Product pauseConsumer skips validateConsumerStatus. Every claim traced against source at cf1c89cae9; suites not run locally, CI green.

This review was prepared with an AI reviewer; every comment was checked and approved by me before posting.

gravitee-apim-portal-webui-next/src/app/dashboard/subscription-details/api-product-subscription-details/api-product-subscription-details.component.html
jaroslawlakomy-grav: Fixed. The publisher-managed subscription status now takes precedence over consumerStatus in both the credentials section and the access status label.

A regression test now covers the complete Pause → Close sequence with status=CLOSED and consumerStatus=STOPPED, ensuring that the view displays Closed and Unavailable instead of Paused.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
jaroslawlakomy-grav: Fixed. API Product environment ownership is now also validated by updateSubscription and resumeFailedSubscription.

Focused tests verify that both operations return 404 for an API Product subscription belonging to another environment. Direct API subscription behavior remains unchanged. The existing isApiProductSubscription helper is now also reused by the read operation.

gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-rest/src/main/java/io/gravitee/rest/api/portal/rest/resource/SubscriptionResource.java
jaroslawlakomy-grav: Fixed. PENDING API Product subscriptions are now allowed through the existing _close flow, preserving the previous withdrawal behavior.

The existing domain service converts a pending request to REJECTED, while ACCEPTED and PAUSED subscriptions are closed normally. A focused resource test covers the pending case, and CLOSED and REJECTED remain rejected by the endpoint.

gravitee-apim-portal-webui-next/src/app/dashboard/subscription-details/api-product-subscription-details/api-product-subscription-details.component.ts
```diff
@@ -47,23 +61,163 @@ import { ApplicationService } from '../../../../services/application.service';
...
   );
   protected readonly clientSecret = computed(() => this.applicationResource.value()?.settings.oauth?.client_secret);
+  protected readonly isActionPending = computed(() => this.pendingAction() !== null);
+  protected readonly canPause = computed(() => {
+    const subscription = this.displayedSubscription();
+    return (
+      subscription.status === 'ACCEPTED' &&
+      subscription.consumerStatus === SubscriptionConsumerStatusEnum.STARTED &&
+      (this.permissionsResource.value()?.SUBSCRIPTION?.includes('U') ?? false)
+    );
```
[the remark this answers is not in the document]
jaroslawlakomy-grav: Acknowledged. This is a readability-only suggestion and does not identify a behavioral or security issue. The current expression is short and remains consistent with the related canResume and canClose predicates, so it is intentionally kept unchanged in this PR.

jaroslawlakomy-grav:
@FuYu3699 
Thanks for the detailed review. All three inline comments and the additional observations have been reviewed.

Changes made:

- Publisher-managed subscription status now takes precedence over consumerStatus, including the Pause → Close sequence.
- API Product environment ownership is consistently validated for read, update, resume-failed, pause, resume, and close operations.
- The previous pending-request withdrawal behavior has been restored. _close accepts PENDING and delegates to the existing domain service, which transitions the request to REJECTED.
- In-progress tests now verify the disabled state of the Pause, Resume, and Close actions.
- CLOSED and REJECTED subscriptions are both covered by assertions verifying that lifecycle actions are unavailable.
- Feedback accessibility semantics now use role="alert" without aria-live for errors and role="status" with aria-live="polite" for success messages.
- getSubscriptionBySubscriptionId now reuses the isApiProductSubscription helper.
- Publisher-paused and consumer-paused subscriptions now display distinct messages.
- A regression test covers the complete Pause → Close sequence with status=CLOSED and consumerStatus=STOPPED.

The two informational notes were also reviewed:

- Action refresh remains local to displayedSubscription in the details component; changing the parent resource lifecycle is not required for this story.
- The existing API Product pauseConsumer domain behavior remains unchanged. Status eligibility continues to be enforced at the Portal API boundary.

Focused backend validation:
- SubscriptionResourceTest: 45 tests passed.

The separate Aikido comment was also reviewed. It is a readability-only suggestion concerning guard clauses and does not identify a behavioral or security issue.

gravitee-apim-portal-webui-next/src/app/dashboard/subscription-details/api-product-subscription-details/api-product-subscription-details.component.ts
```diff
@@ -47,23 +61,163 @@ import { ApplicationService } from '../../../../services/application.service';
...
   protected readonly planSecurityLabel = computed(() => getPlanSecurityTypeLabel(this.plan()?.security));
+  protected readonly accessStatusLabel = computed(() => {
+    const subscription = this.displayedSubscription();
+    if (subscription.status !== 'ACCEPTED') {
+      return $localize`:@@apiProductSubscriptionAccessStatusUnavailable:Unavailable`;
+    }
+    if (subscription.consumerStatus === SubscriptionConsumerStatusEnum.STOPPED) {
+      return $localize`:@@apiProductSubscriptionAccessStatusPaused:Paused`;
+    }
+    return subscription.consumerStatus === SubscriptionConsumerStatusEnum.STARTED
```
thomasballivet, line 87: `consumerStatus === 'FAILURE'` isn't handled, so the page contradicts itself: status shows **Accepted**, access status falls through to **Unavailable**, and the API cards say *"API access is not available for the current subscription status"* — but the credentials block only tests `STOPPED`, so it still displays the API key / client ID. On top of that neither Pause nor Resume is shown, so the consumer is stuck.

At minimum, align the three signals (label, credentials block, API cards) so we never show credentials next to "access unavailable". Ideally handle `FAILURE` like the direct API subscription does (failure cause + Resume through the existing `_resumeFailure` / `SubscriptionService.resumeConsumerStatus()`), or open a follow-up if that's out of scope here.

thomasballivet (approved):

gravitee-apim-portal-webui-next/src/app/dashboard/subscription-details/api-product-subscription-details/api-product-subscription-details.component.ts
jaroslawlakomy-grav: Fixed. API Product subscriptions with consumerStatus=FAILURE no longer display credentials and now show the failure cause together with a Retry subscription action.

Retry uses the existing _resumeFailure endpoint, requires the application update permission, refreshes the subscription details after completion, and disables concurrent lifecycle actions while running. Focused tests cover credential hiding, permission gating, successful retry, and retry failure.

PR #19353, opened 2026-08-24
fix(test): make the elasticsearch sample data independent of the clock

phiz71:
## Issue

https://gravitee.atlassian.net/browse/FOUND-235

## Description

The Elasticsearch repository sample data depends on the clock in two independent ways. Each makes tests fail at some hours and pass at others, and both look like analytics regressions without being one. Test configuration and test data only — no production code.

### 1. The sample data and production disagree on which day it is

The templates under `src/test/resources` spell their timestamps out in UTC — `"@timestamp": "${dateToday}T06:54:30.047Z"` — so `TimeProvider` resolves days in UTC to match. Production resolves the day it reads from the **default zone**: `AbstractIndexNameGenerator.getTodayIndexName()` calls `LocalDate.now()`, and `DateUtils#rangedIndices` converts through `ZoneId.systemDefault()`.

The two agree only when the JVM runs on UTC, as CI does. Anywhere else they designate different days for the hours where the local date is not the UTC date — on CEST, every day between 00:00 and 02:00 local. Repositories carrying no time range of their own then read a day the fixtures were never written to.

The surefire fork now runs with `-Duser.timezone=UTC`, which is what the dataset already assumes.

### 2. Half the traffic samples followed the clock, the other half did not

A sample may follow the clock, as the health-check and monitoring ones do, as long as every assertion about it follows too — those tests build their window from the same reference, so the two never drift apart.

The traffic samples were different: some documents carried an hour written into the template while others interpolated the reference instant. The distance between the two families therefore changed with the hour of the run, so documents falling in distinct time buckets at one hour merged into a single bucket at another, and any assertion counting buckets became a function of when the build ran.

They now sit at a fixed time of day, `TimeProvider.TRAFFIC_SAMPLE_TIME`, exposed to tests as `trafficSampleInstant()`. Early in the day, because `ElasticsearchAnalyticsRepositoryTest` queries `lastDays(30)`, whose upper bound is the system clock: a sample dated after the moment the build starts is invisible to it, and anything later would break the nightly.

`should_return_response_status_for_api_v2_and_v4` loses the tolerance that hid this:

```java
// before — either outcome accepted, "to avoid making this test flaky"
assertThat(array).satisfiesAnyOf(c -> assertThat(c).containsOnly(36.25, 20.0), c -> assertThat(c).containsOnly(33.0));

// after
assertThat(array).containsOnly(36.25, 20.0, 245.0);
```

The third value is the current day's traffic, in the bucket the window closes on — the interval starting at `to`, a point of the series like any other. The assertion refused it; that was the mistake, not the data.

## Additional context

Measured on a machine in CEST, same commit, only the JVM zone and the local hour changing:

| Zone | Local clock | Result |
| --- | --- | --- |
| `Europe/Paris` | 13:05 | 827 tests, 0 failures |
| `Etc/GMT-8` | 19:06 | 827 tests, 0 failures |
| `Etc/GMT+9` | 02:07 | 827 tests, 0 failures |

The third row is the case that used to fail: the local date is a day behind the UTC date. Before the zone pin, the same commit reported 2 errors there — `ElasticsearchLogRepositoryTest.testFindById` and `ElasticsearchMonitoringRepositoryTest.testQuery`, exactly the two repositories that resolve an index off the system clock.

After rebasing on the current master: **864 tests, 0 failures**, licence and prettier checks included.

Two implementation notes for the backports:

- `gravitee-parent` offers a `surefireArgLine` property, but it does not work here: `gravitee-apim-parent` overrides surefire's `argLine` in `<build><plugins>`, so the module has to declare `argLine` itself — which replaces the parent's rather than adding to it, hence its `add-opens` flags being repeated with a comment.
- There are two fixture sets. `AnalyticsElasticsearchRepositoryTest` overrides the template path to `src/test/resources/freemarker-v4-analytics`, so `request.ftl` exists twice and both copies needed the change.


gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/TimeProvider.java
```diff
@@ -62,6 +63,22 @@ public class TimeProvider {
...
+     * drift apart. The traffic samples are different: some of their documents carry an hour written into the
+     * template while others interpolate the reference, so a moving reference changes the distance between the
+     * two. Documents that fall in distinct time buckets at one hour of the day then merge into a single bucket
+     * at another, and any assertion that counts buckets becomes a function of when the build runs.
+     *
+     * <p>Early in the day, because {@code ElasticsearchAnalyticsRepositoryTest} queries {@code lastDays(30)},
+     * whose upper bound is the system clock: a sample dated after the moment the build starts is invisible to
+     * it. Anything later would break the nightly, which starts a few minutes past midnight.
+     */
+    private static final LocalTime TRAFFIC_SAMPLE_TIME = LocalTime.of(0, 5);
```
ytvnr, line 87: **Worth fixing**: `TRAFFIC_SAMPLE_TIME = 00:05` narrows the clock dependent window instead of closing it.

`SinceDateRange` sets `end = System.currentTimeMillis()`, so the upper bound of `lastDays(30)` is the wall clock at query time. `ElasticsearchAnalyticsRepositoryTest.testCountWithTwoApis` asserts `count == 7`, and three of those documents are the `e2c0ecd5-...` requests now stamped `${trafficDateTimeToday}`, so today at 00:05 UTC. A run that reaches that test before 00:05 UTC sees 4.

The javadoc just above justifies the value with "the nightly, which starts a few minutes past midnight", which puts the build start and the sample time inside the same five minute band. The start of the day satisfies the same constraint and still sits inside the `truncatedTo(DAYS)` windows the module queries. If 00:05 is deliberate, the nightly's real trigger time is worth recording next to it, since the ticket asks for the chosen hour to be backed by a run.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/TimeProvider.java
```diff
@@ -93,13 +112,25 @@ public TimeProvider() {
...
+        trafficDateTimeYesterday = DATE_TIME_FORMATTER_WITH_DASH.format(trafficToday.minus(1, ChronoUnit.DAYS));
+
         todayWithDot = DATE_FORMATTER_WITH_DOT.format(now);
         yesterdayWithDot = DATE_FORMATTER_WITH_DOT.format(yesterday);
     }
 
     /**
      * @return the reference instant every Elasticsearch test must use instead of {@link Instant#now()}.
      */
+    /**
```
ytvnr: **Worth fixing**: both new javadoc blocks are inserted below the existing ones, so `now()` and `REFERENCE` end up undocumented.

Here the block describing "the reference instant every Elasticsearch test must use" now sits above `trafficSampleInstant()`, and `now()` a few lines further down is bare. Same shape at the `TRAFFIC_SAMPLE_TIME` declaration: the block explaining `REFERENCE` is followed immediately by a second javadoc, so only the new one attaches and `REFERENCE` loses its own.

While you are in there, the paragraph starting "Early in the day, because `ElasticsearchAnalyticsRepositoryTest` queries `lastDays(30)`" reads as a subordinate clause with no main clause. This file is where the ticket asks the rationale to live, so it is worth having it land cleanly.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/analytics/AnalyticsElasticsearchRepositoryTest.java
```diff
@@ -395,20 +395,18 @@ void should_return_response_status_for_api_v2_and_v4() {
...
                 .mapToDouble(l -> l)
                 .filter(d -> d > 0)
                 .toArray();
-            // Depending on the exact timestamps of test data relative to the 10-minute bucket boundaries,
-            // the same underlying response times can be aggregated either into two separate buckets
-            // (values [36.25, 20.0]) or into a single bucket (value [33.0]). Both outcomes are valid
-            // representations of the same dataset with slightly different time bucketing, so we accept
-            // either result to avoid making this test flaky while still asserting on the actual values.
-            assertThat(array).satisfiesAnyOf(c -> assertThat(c).containsOnly(36.25, 20.0), c -> assertThat(c).containsOnly(33.0));
+            assertThat(array).containsOnly(36.25, 20.0, 245.0);
```
ytvnr, line 411: **Worth fixing**: `245.0` holds only while `TRAFFIC_SAMPLE_TIME` stays under the 10 minute interval used here, and that coupling is written nowhere.

The window is `[yesterday 00:00, today 00:00]`, and `ResponseTimeRangeQueryAdapter` widens the filter to `HistogramGrid.endOfLastBucket(...)`, so today 00:10. That is the only reason today's samples at 00:05 (response times 225, 5, 505, average 245.0) are visible at all. Move the constant past 00:10 and the value disappears, with the failure surfacing three files away from its cause.

The comment you added above explains why a third bucket exists, which already helps. What is missing is the link back to the constant that makes it true, either stated where `TRAFFIC_SAMPLE_TIME` is defined or derived from `trafficSampleInstant()` right here.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/pom.xml
```diff
@@ -270,6 +271,33 @@
...
+                        that carry no time range of their own read a different day than the fixtures were
+                        written to, and fail for anyone whose local date differs from the UTC date.
+
+                        The add-opens flags below repeat gravitee-apim-parent's surefire argLine: that
+                        configuration sits in <build><plugins>, so declaring argLine here replaces it rather
+                        than adding to it. Keep the two in sync.
+                    -->
+                    <argLine>
+                        @{argLine}
+                        -Duser.timezone=UTC
```
ytvnr, line 292: The UTC pin covers surefire forks only, so an IDE run on `Europe/Paris` still hits the day mismatch this PR fixes.

Not a problem with the change itself, the flags all match the parent. Just that a developer running these tests from their IDE between 00:00 and 02:00 local gets the old failure with no hint as to why. A cross reference from `TimeProvider.ZONE`'s javadoc, or a fail fast in the test setup when `ZoneId.systemDefault()` is not UTC, would say it out loud.

ytvnr (approved):
Two clock dependencies removed from the elasticsearch fixtures, and the tolerance in `should_return_response_status_for_api_v2_and_v4` dropped rather than backported. The `argLine` override checks out: the five `add-opens` and `@{argLine}` match `gravitee-apim-parent` exactly, so Jacoco propagation stays intact, and the two `request.ftl` copies are byte identical.

The one point I would look at again is `TRAFFIC_SAMPLE_TIME = 00:05`. `lastDays(30)` closes on `System.currentTimeMillis()`, so a build that reaches `testCountWithTwoApis` before 00:05 UTC does not see the three samples dated at 00:05. The javadoc justifies the value by the nightly starting a few minutes past midnight, which puts the build start and the sample time in the same band.

Nothing blocking, and all of it open to discussion.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/TimeProvider.java
phiz71: Measured before answering, and 00:00 does not hold: it breaks `should_return_a_page_of_connection_logs_from_today_to_whenever_from_multi_indexes` and `..._from_yesterday_from_multi_indexes`. Those windows open at `withHour(0).withMinute(1)`, so a sample dated midnight falls one minute short of them.

So the value is squeezed between two bounds rather than freely chosen: after 00:01 for those two, before the moment the build starts for `lastDays(30)`. The nightly of 2026-08-24 started at 00:16 UTC — I took that from its own log, `Finished at 00:23:33Z` minus a 7m03s wall clock — which leaves room but not much, and you are right that the javadoc made it sound like a comfortable margin.

You are also right that the window is narrowed, not closed: a build reaching `testCountWithTwoApis` between 00:01 and 00:05 counts four instead of seven. No constant closes it, since the two bounds are four minutes apart. Closing it for good means moving those `withMinute(1)` bounds, which belongs to those tests rather than to this change — I would rather say so in the javadoc than pretend the value is safe.

Rewritten in 432d7077b2, with both bounds, the measurement behind each, and the residual gap stated.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/TimeProvider.java
phiz71: Confirmed and fixed in 432d7077b2. Both new blocks sat above an existing one, so `REFERENCE` and `now()` were left bare — the javadoc I added attached to the declaration that followed, not to the one it described. Reordered so each block sits on its own member, and checked that no block is preceded by the end of another.

The dangling sentence is gone too: the rationale is now a list of the two bounds rather than a clause hanging off the previous paragraph.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/analytics/AnalyticsElasticsearchRepositoryTest.java
phiz71: Agreed — the comment explained why a third bucket exists without saying what makes its value reachable. Stated in both directions in 432d7077b2: here, that 245.0 is only in range because `TimeProvider.TRAFFIC_SAMPLE_TIME` sits inside the interval the window closes on, and next to the constant, that it has to stay under the ten minute interval this test uses.

I kept the literal rather than deriving it from `trafficSampleInstant()`. Deriving would hide which samples the test is actually reading, and the value is one of three the assertion pins; the coupling is worth naming, not worth computing.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/pom.xml
phiz71: Took the fail-fast rather than the cross-reference: `TestConfiguration` now refuses to build the clock when `LocalDate.now()` and `LocalDate.now(UTC)` disagree, naming both dates and the flag to add.

It fires on the mismatch rather than on the zone, so an IDE run on `Europe/Paris` still works all day and only stops during the hours it would have failed anyway — with a message instead of two repositories coming back empty.

One limit worth stating: I could not exercise it through Maven, precisely because the module's `argLine` pins the fork to UTC, so the condition can never be true there. I checked the predicate itself under two zones — `Europe/Paris` false, `Etc/GMT-11` true, its local date being a day ahead of UTC at the time. The wiring into `TestConfiguration` is not covered by a test.

PR #19354, opened 2026-08-24
fix(analytics): stop returning histogram buckets outside the requested window

phiz71:
## Issue

https://gravitee.atlassian.net/browse/FOUND-240

## Description

Three time-series adapters widened their range filter by one interval on each side while `extended_bounds` stayed on `[from, to]`. `extended_bounds` extends the returned range — it does not clamp it; clamping is what `hard_bounds` does. A document admitted in the margin **before** `from` therefore came back as a bucket before the start of the window, and since these responses carry no timestamp per point — only `timeRange` plus ordered lists — the consumer can only place values by position. That leading bucket shifts **every point of the series by one interval**.

Asking for response statuses between 10:00 and 11:00 with a 10-minute interval, with one request at 09:52, three at 10:15 and two at 10:45:

```
before   8 buckets   [1, 0, 3, 0, 0, 2, 0, 0]   09:52 shown at 10:00, 10:15 shown at 10:20, …
after    7 buckets   [0, 3, 0, 0, 2, 0, 0]      each point on its true slot
```

### The fix

The root of it is that the bucket grid is aligned on epoch multiples of the interval, so the bucket holding `from` generally starts before it. That left no good option: filter strictly on the window and the first bucket is short; filter wider and the extra traffic comes back as a bucket outside the window.

The grid is now shifted onto the window itself, through the `offset` parameter of `date_histogram`:

```java
Instant from = query.from();
Instant to = query.to().plus(query.interval());
…
.put("offset", HistogramGrid.offsetMillis(query.from(), query.interval()) + "ms")
```

With a bucket boundary sitting exactly on `from`, the filter starts there — nothing that happened before the window is counted — and reaches the end of the last bucket, so that bucket stays complete. The upper bound is exclusive so a document landing exactly on it cannot open a bucket past `extended_bounds.max`.

`offset` is a long-standing `date_histogram` parameter on both Elasticsearch and OpenSearch, unlike `hard_bounds`, so no version gate is needed.

The series keeps its `(to − from) / interval + 1` points and the last one keeps its value. `TimeRangeAdapter` and `DateHistogramAdapter` never widened and are untouched.

## Additional context

**The existing repository tests already asserted the corrected behaviour** — the exact bucket count and the positional correspondence `index i ↔ from + i × interval`. They are unchanged and green.

New coverage, since the previous round of this PR showed that pinning the emitted query is not enough to catch a regression on either bound:

- `AnalyticsElasticsearchRepositoryTest.ResponseStatusOverTime.WindowBoundaries` asserts the returned series rather than the query: a document sitting exactly on `from` lands on the first point with the expected bucket count, a document one millisecond before `from` is not reported, and a document past the end of the last bucket is not reported either. The windows are derived from the instant the sample documents sit on, so the cases do not depend on the hour the build runs at.
- `HistogramGridTest` covers the grid offset itself, including windows already on a boundary, windows before the epoch, and intervals that do not amount to a millisecond.

Full module: **814 tests, 0 failures**, licence and prettier checks included.

### History of this branch

The first implementation rounded the filter bounds down to the epoch-aligned grid and dropped the trailing bound. Review showed both halves were wrong in different ways: rounding down still counted traffic from before `from` into the first bucket, and dropping the trailing bound left the last bucket emitted but empty — a regression on the final point of every series, since that bucket is emitted by `extended_bounds.max` whether or not the filter reaches it.

A claim made in the first description, that this branch unblocked the second half of FOUND-235, rested on that regression and does not hold. The constraint analysis on that ticket is being redone.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapter.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapter.java
@@ -23,6 +23,7 @@ import io.gravitee.definition.model.DefinitionVersion;
 import io.gravitee.elasticsearch.model.Aggregation;
 import io.gravitee.elasticsearch.model.SearchResponse;
 import io.gravitee.elasticsearch.version.ElasticsearchInfo;
+import io.gravitee.repository.elasticsearch.utils.HistogramGrid;
 import io.gravitee.repository.log.v4.model.analytics.ResponseStatusOverTimeAggregate;
 import io.gravitee.repository.log.v4.model.analytics.ResponseStatusOverTimeQuery;
 import java.time.Instant;
@@ -103,11 +104,15 @@ public class SearchResponseStatusOverTimeAdapter {
             apiFilter.add(filterV2);
         }
 
-        // we just ensure to fetch full bucket interval (a bit too)
-        Instant from = query.from().minus(query.interval());
-        Instant to = query.to().plus(query.interval());
+        // Align the bucket grid on the window itself. With the default epoch-aligned grid the bucket holding
+        // `from` starts before it, so the filter would have to reach back past `from` to fill it — and anything
+        // admitted there comes back as a bucket outside the window, because extended_bounds extends the
+        // returned range without clamping it. The upper bound stays open to the end of the last bucket so that
+        // bucket is complete, and stops at the end of that bucket so nothing can open one beyond the window.
+        Instant from = query.from();
+        Instant to = HistogramGrid.endOfLastBucket(query.from(), query.to(), query.interval());
 
-        ObjectNode timestamp = json().put("gte", from.toEpochMilli()).put("lte", to.toEpochMilli());
+        ObjectNode timestamp = json().put("gte", from.toEpochMilli()).put("lt", to.toEpochMilli());
         JsonNode rangeFilter = json().set("range", json().set(TIME_FIELD, timestamp));
 
         var v2orv4 = json().put("minimum_should_match", 1).set("should", apiFilter);
@@ -125,6 +130,7 @@ public class SearchResponseStatusOverTimeAdapter {
         ObjectNode histogram = json()
             .put("field", TIME_FIELD)
             .put(intervalFieldName, query.interval().toMillis() + "ms")
+            .put("offset", HistogramGrid.offsetMillis(query.from(), query.interval()) + "ms")
             .put("min_doc_count", 0)
             .set(
                 "extended_bounds",
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapterTest.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapterTest.java
@@ -72,6 +72,7 @@ class SearchResponseStatusOverTimeAdapterTest {
                          "extended_bounds": { "max": 1697969730000, "min": 1697883330000 },
                          "field": "@timestamp",
                          "fixed_interval": "600000ms",
+                         "offset": "330000ms",
                          "min_doc_count": 0
                        }
                      }
@@ -83,8 +84,8 @@ class SearchResponseStatusOverTimeAdapterTest {
                          {
                            "range": {
                              "@timestamp": {
-                               "gte": 1697882730000,
-                               "lte": 1697970330000
+                               "gte": 1697883330000,
+                               "lt": 1697970330000
                              }
                            }
                          }
@@ -122,6 +123,7 @@ class SearchResponseStatusOverTimeAdapterTest {
                          "extended_bounds": { "max": 1697969730000, "min": 1697883330000 },
                          "field": "@timestamp",
                          "fixed_interval": "600000ms",
+                         "offset": "330000ms",
                          "min_doc_count": 0
                        }
                      }
@@ -133,8 +135,8 @@ class SearchResponseStatusOverTimeAdapterTest {
                          {
                            "range": {
                              "@timestamp": {
-                               "gte": 1697882730000,
-                               "lte": 1697970330000
+                               "gte": 1697883330000,
+                               "lt": 1697970330000
                              }
                            }
                          }
@@ -172,6 +174,7 @@ class SearchResponseStatusOverTimeAdapterTest {
                          "extended_bounds": { "max": 1697969730000, "min": 1697883330000 },
                          "field": "@timestamp",
                          "interval": "600000ms",
+                         "offset": "330000ms",
                          "min_doc_count": 0
                        }
                      }
@@ -183,8 +186,8 @@ class SearchResponseStatusOverTimeAdapterTest {
                          {
                            "range": {
                              "@timestamp": {
-                               "gte": 1697882730000,
-                               "lte": 1697970330000
+                               "gte": 1697883330000,
+                               "lt": 1697970330000
                              }
                            }
                          }
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/utils/HistogramGrid.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/utils/HistogramGrid.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.elasticsearch.utils;
+
+import java.time.Duration;
+import java.time.Instant;
+
+/**
+ * Placement of the bucket grid of a date histogram run over a caller-supplied window.
+ *
+ * <p>By default Elasticsearch aligns buckets on epoch multiples of the interval, so the bucket holding the
+ * start of the window generally begins before it. That leaves no good option: filtering strictly on the window
+ * yields a short first bucket, and filtering wider admits documents that come back as a bucket outside the
+ * window — {@code extended_bounds} extends the returned range, it does not clamp it. Since these responses
+ * carry no timestamp per point, such a bucket shifts the whole series by one interval.
+ *
+ * <p>Shifting the grid onto the window itself removes the dilemma: the first bucket starts exactly at the
+ * start of the window, so the filter can begin there and the first bucket is still complete.
+ */
+public final class HistogramGrid {
+
+    private HistogramGrid() {}
+
+    /**
+     * Offset, in milliseconds, that places a bucket boundary exactly on {@code windowStart}.
+     *
+     * @return zero when the window already starts on a boundary, or when the interval is unusable (see
+     *         {@link #usableMillis(Duration)}).
+     */
+    public static long offsetMillis(final Instant windowStart, final Duration interval) {
+        final long intervalMillis = usableMillis(interval);
+        return intervalMillis == 0 ? 0L : Math.floorMod(windowStart.toEpochMilli(), intervalMillis);
+    }
+
+    /**
+     * End of the last bucket the histogram emits for the window, which is the instant a range filter must stop
+     * short of.
+     *
+     * <p>{@code extended_bounds.max} rounds down onto the grid, so the last emitted bucket is the one holding
+     * {@code windowEnd}, not the one starting on it. Filtering up to {@code windowEnd + interval} instead is
+     * only equivalent when the window spans a whole number of intervals; otherwise it admits documents that
+     * open a bucket past {@code extended_bounds.max} — and since {@code extended_bounds} does not clamp, that
+     * bucket is returned, adding a point that carries traffic from after the window.
+     *
+     * @return {@code windowEnd} itself when the interval is unusable (see {@link #usableMillis(Duration)}).
+     */
+    public static Instant endOfLastBucket(final Instant windowStart, final Instant windowEnd, final Duration interval) {
+        final long intervalMillis = usableMillis(interval);
+        if (intervalMillis == 0) {
+            return windowEnd;
+        }
+        final long startMillis = windowStart.toEpochMilli();
+        final long span = windowEnd.toEpochMilli() - startMillis;
+        final long buckets = Math.floorDiv(span, intervalMillis) + 1;
+        return Instant.ofEpochMilli(startMillis + buckets * intervalMillis);
+    }
+
+    /**
+     * @return the interval in milliseconds, or zero when it does not amount to at least one — which covers a
+     *         null, zero or negative duration, and also a positive one shorter than the millisecond the
+     *         histogram is expressed in.
+     */
+    private static long usableMillis(final Duration interval) {
+        if (interval == null) {
+            return 0L;
+        }
+        final long intervalMillis = interval.toMillis();
+        return intervalMillis <= 0 ? 0L : intervalMillis;
+    }
+}
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/ResponseTimeRangeQueryAdapter.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/ResponseTimeRangeQueryAdapter.java
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.gravitee.elasticsearch.model.Aggregation;
 import io.gravitee.elasticsearch.model.SearchResponse;
 import io.gravitee.elasticsearch.version.ElasticsearchInfo;
+import io.gravitee.repository.elasticsearch.utils.HistogramGrid;
 import io.gravitee.repository.log.v4.model.analytics.AverageAggregate;
 import io.gravitee.repository.log.v4.model.analytics.ResponseTimeRangeQuery;
 import io.reactivex.rxjava3.core.Maybe;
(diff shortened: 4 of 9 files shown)


gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapter.java
```diff
@@ -103,9 +104,11 @@ private ObjectNode query(ResponseStatusOverTimeQuery query) {
...
         }
 
-        // we just ensure to fetch full bucket interval (a bit too)
-        Instant from = query.from().minus(query.interval());
-        Instant to = query.to().plus(query.interval());
+        // Align the lower bound down to the bucket boundary so the first bucket is complete, and never look
+        // past `to`: extended_bounds extends the returned range without clamping it, so any document admitted
+        // beyond the requested window comes back as an extra bucket and shifts the whole series by an interval.
+        Instant from = HistogramBounds.alignDown(query.from(), query.interval());
+        Instant to = query.to();
```
ytvnr: **⚠️ Important** - Dropping the upper widening without touching `extended_bounds.max` leaves the last bucket cut mid span, and effectively empty when `to` sits on a bucket boundary.

`extended_bounds.max` is still `query.to()` (line 137 here, and `ResponseTimeRangeQueryAdapter.java:117`, `AverageHealthCheckResponseTimeOvertimeAdapter.java:92`). Elasticsearch rounds that bound down to the bucket boundary and always emits that bucket, but the range filter now stops at `to`, inside it. With the fixture used by this PR's own adapter test (`FROM`/`TO` at 10:15:30, interval 10 min) the final bucket `[10:10, 10:20)` only receives documents up to 10:15:30. When `to` is aligned on the interval, the case `AnalyticsElasticsearchRepositoryTest` and `HealthCheckElasticsearchRepositoryTest` exercise, it only admits the exact millisecond `to`, so the last point is a phantom zero. On master `lte` was `to + interval` and that bucket was complete, so this is a regression on the final data point, and acceptance criterion 3 ("the last bucket is not partially filled") has no implementation.

Either both ends drop the trailing incomplete bucket (filter and `extended_bounds.max` together), or the histogram aligns on the window rather than on the epoch.

ytvnr (commented):
Thanks for the detailed write-up. The diagnosis of the head shift, and the `extended_bounds` vs `hard_bounds` distinction, both hold up against the code.

My main concern is that the fix is asymmetric. The head widening is gone, but `extended_bounds.max` still carries `query.to()` while the range filter now stops at `to`, so the last bucket is still emitted and is cut mid span. When `to` is aligned on the interval, which is the case the repository tests exercise, that bucket only admits the exact millisecond `to` and the last data point becomes a phantom zero. On master `lte` was `to + interval` and that bucket was complete, so the final point of every series regresses.

Underneath both ends, the bucket grid is still aligned on epoch multiples: nothing in the diff changes it. That is also why acceptance criterion 2 of FOUND-240 ("a document one minute before `from` does not appear") is not met, since `alignDown` pulls `gte` back before `from`. The `offset` parameter of `date_histogram` would align the grid on `from` itself and needs no version gate, unlike `hard_bounds`.

`alignDown` itself is sound: `Math.floorDiv` handles pre epoch instants, the multiplication cannot overflow past the input magnitude, and the `<= 0` guard covers null, zero, negative and sub millisecond intervals.

### Notes outside the diff

- `AnalyticsElasticsearchRepositoryTest.java:444-446`: the third scope item of the ticket, turning the size and positional assertions into deterministic ones, does not appear in this branch. `contains(1L, atIndex(28))` still derives from `TimeProvider.now()`, and lines 398-409 still carry the `satisfiesAnyOf(containsOnly(36.25, 20.0), containsOnly(33.0))` workaround with its comment about avoiding flakiness.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/utils/HistogramBounds.java
```diff
@@ -0,0 +1,53 @@
...
+        if (interval == null) {
+            return instant;
+        }
+        final long intervalMillis = interval.toMillis();
+        if (intervalMillis <= 0) {
+            return instant;
+        }
+        final long epochMilli = instant.toEpochMilli();
+        return Instant.ofEpochMilli(Math.floorDiv(epochMilli, intervalMillis) * intervalMillis);
+    }
```
ytvnr: **Worth fixing** - Rounding the lower bound down to the epoch aligned boundary counts traffic that happened before `from` into the first bucket, which acceptance criterion 2 forbids.

With `from` at 10:15:30 and a 10 minute interval, `gte` becomes 10:10:00, so a request at 10:14:30 lands in the first point. The branch buys "first bucket complete" at the cost of "nothing before `from`", and the trailing end fails the mirror criterion. Both follow from the bucket grid staying on epoch multiples.

The `offset` parameter of `date_histogram` shifts that grid onto `from` itself, and is available on every supported Elasticsearch and OpenSearch version, unlike `hard_bounds`. The filter could then be strictly `[from, to]`, with complete first and last buckets and no extra points.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapterTest.java
```diff
@@ -83,8 +83,8 @@ void adapt_query_for_ElasticSearch() {
                          {
                            "range": {
                              "@timestamp": {
-                               "gte": 1697882730000,
-                               "lte": 1697970330000
+                               "gte": 1697883000000,
+                               "lte": 1697969730000
                              }
                            }
                          }
```
ytvnr, line 91: **Worth fixing** - The new coverage pins the emitted JSON and the arithmetic of `alignDown`, but nothing asserts the behaviour the fix claims to deliver: the bucket count, and the exclusion of out of window documents.

Acceptance criteria 1 and 2 are statements about the returned series, not about the query. The repository level tests that could carry them (`AnalyticsElasticsearchRepositoryTest`, `HealthCheckElasticsearchRepositoryTest`) were not extended with a document seeded just outside `[from, to]`, so a regression on either bound would still pass CI.

A repository level case seeding one document just before `from` and one just after `to`, asserting the exact series length and the absence of both, would pin it.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/utils/HistogramBoundsTest.java
```diff
@@ -0,0 +1,72 @@
...
+package io.gravitee.repository.elasticsearch.utils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.time.Duration;
+import java.time.Instant;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class HistogramBoundsTest {
```
ytvnr: **Worth fixing** - This test class does not carry `@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)`, unlike its siblings in the same package, and there is no module wide default to supply it.

`JsonNodeUtilsTest` and `OtelDataStreamIndexUtilsTest` both declare it, and the module has no `junit-platform.properties`, so this class reports raw method names. The tests also drop the `// Given / When / Then` structure used elsewhere, and `should_return_the_instant_unchanged_when_the_interval_is_not_usable` bundles four independent cases in one method, so the first failure hides the rest. A parameterised test would keep it short.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapter.java
phiz71: You are right, and checking it changed my understanding of the defect.

I went back to the evidence in the ticket. The extra bucket I measured sat at `2026-08-22T23:50` for a window starting `2026-08-23T00:00`, and the example in the description is a request at 09:52 for a window starting at 10:00. Both are the **head**. The head widening is what creates a bucket before `from` and shifts every point.

The tail widening does not do that. `lte = to + interval` feeds the bucket `[to, to+interval)`, which is the bucket `extended_bounds.max` already emits, so it adds no bucket — it only fills the last one. Dropping it, as this branch does, leaves that bucket emitted and empty. That is a regression on the final point, and it is mine, not master's.

So the fix I want is the second option you name — align the grid on the window — with the trailing filter kept:

- `offset` on `date_histogram` so buckets start at `from` rather than on epoch multiples;
- range filter `gte from` strictly, and `lt to + interval` rather than `lte`, so the last bucket is complete and a document exactly at `to + interval` cannot open one beyond `extended_bounds.max`;
- `extended_bounds` unchanged at `[from, to]`.

That gives `(to - from) / interval + 1` buckets, all complete, none outside the window, and nothing before `from`. The N+1 contract and the last data point both stay as they are on master.

One correction I owe the description: I claimed this branch unblocks the second half of FOUND-235, on the strength of a probe where the last bucket stopped being populated. That was this regression, not a fix. With the design above the value comes back and that claim does not hold — I will restate it there.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/utils/HistogramBounds.java
phiz71: Conceded — this contradicts an acceptance criterion I wrote myself in the ticket, "a document one minute before `from` does not appear in the result", and I did not reconcile the two.

I reasoned about the trade-off in the description, choosing "first bucket complete" over "nothing before `from`", and then never went back to check it against the criteria. Rounding down buys the first property at the cost of the second; both are only reachable by moving the grid, which is what `offset` does.

So `alignDown` should not be feeding the filter at all. With `offset` set to the window's own phase, the first bucket starts exactly at `from`, the filter can be `gte from` strictly, and no traffic before `from` enters. The method may still be useful for computing that phase — `from` minus its aligned-down value is the offset — but the query no longer widens either bound to reach it.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapterTest.java
phiz71: Agreed, and this is the gap that lets the other two through. The tests I added pin the emitted JSON; acceptance criteria 1 and 2 are statements about the returned series, and nothing asserts them. A regression on either bound passes CI today — which is precisely what happened at the tail.

I will add the repository-level case you describe: one document seeded just before `from`, one just after `to`, asserting the exact series length and the absence of both. It has to come with the bounds design settled, since the assertion encodes it — writing it against the current branch would pin the regression.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/utils/HistogramBoundsTest.java
phiz71: Fixed in d2873571e5: `@DisplayNameGeneration` declared, Given/When/Then restored, and the four unusable-interval cases split into a `@ParameterizedTest` so one failure no longer hides the others.

Small precision on the premise: `ClusterUtilsTest` in the same package does not declare it either, so it is two of four rather than all siblings. The convention is the more recent one and worth following — I just did not want to leave the count wrong in the thread.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapter.java
phiz71: Implemented in 35f1d7402a, along the second option you named.

The grid is shifted onto the window with the `offset` parameter of `date_histogram`, so a bucket boundary sits exactly on `from`. The filter can then be `gte from` strictly — nothing before the window is counted, which is acceptance criterion 2 — and `lt to + interval`, so the last bucket is complete and a document landing exactly on that bound cannot open one past `extended_bounds.max`. Both edge buckets are complete, the N+1 contract and the final data point are unchanged from master, and no version gate is needed.

`HistogramBounds.alignDown` is gone from the query path; the helper is now `HistogramGrid.offsetMillis`, which computes the phase of the window relative to the interval.

On your third point, the behavioural gap: `AnalyticsElasticsearchRepositoryTest.ResponseStatusOverTime.WindowBoundaries` now asserts the returned series rather than the query — a document on `from` lands on the first point with the expected bucket count, a document one millisecond before `from` is not reported, and one past the end of the last bucket is not reported either. The windows derive from the instant the sample documents sit on, so the cases hold whatever hour the build runs at. That class fails on the previous revision of this branch, which is what makes it worth having.

Full module: 814 tests, 0 failures. The PR description has been rewritten — the defect is described as what it is, a head-side shift, and the FOUND-235 claim is withdrawn.

ytvnr: The `offset` does what it claims: `floorMod(from, interval)` puts a boundary exactly on `from`, so `gte from` no longer loses the first bucket. Criterion 2 is met, and the three adapters are consistent.

One case is left at the other end. `to = query.to().plus(query.interval())` is the end of the bucket that *starts* at `to`, which is the end of the last emitted bucket only when `to` sits on the grid. When the window duration is not a whole multiple of the interval, the filter reaches past it.

`DurationUtils.buildIntervalFromTimePeriod` picks the interval from a fixed ladder rather than dividing the window, and the console's custom picker takes seconds, so this is reachable. Window 09:00:07 to 10:00:00: duration 59m53s, interval 15s, so `(to - from) / i + 1 = 240` buckets. The grid puts the last emitted bucket at 09:59:52, since `extended_bounds.max` rounds down. The filter is `lt 10:00:15`. A document at 10:00:10 opens the bucket at 10:00:07, past `extended_bounds.max`, so the series comes back with 241 points and the last one carries traffic that happened after `to`. Same defect as the head one, quieter because it does not shift the series.

The bound that holds in both cases is the end of the last emitted bucket, `from + (floorDiv(to - from, interval) + 1) * interval`, which is exactly `to + interval` when `to` is on the grid, so it is a no-op on the common path.

Worth noting for the coverage: the three `WindowBoundaries` cases all use a one hour window with a ten minute interval, and the adapter fixture is 86400000 ms over 600000 ms, so every window currently under test is aligned.

ytvnr (approved):
The `offset` rework answers the two bounds questions and the coverage one. I left a note on the trailing bound for windows whose duration is not a whole multiple of the interval. Nothing blocking from my side.

Still open from the ticket, and my fault for putting it in the review body rather than a thread the first time round: the third scope item, turning the size and positional assertions into deterministic ones. `AnalyticsElasticsearchRepositoryTest` still carries `satisfiesAnyOf(containsOnly(36.25, 20.0), containsOnly(33.0))` with its "avoid making this test flaky" comment at lines 410-411, `hasSize((int) nbBuckets)` derived from `TimeProvider.now()` at lines 445-447, and two more of the same shape at lines 1418 and 1446. The new `WindowBoundaries` class shows the pattern that would replace them, since it derives its windows from the instant the documents sit on.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapter.java
phiz71: Reproduced your case before touching anything — 09:00:07 to 10:00:00 by 15s gives 240 buckets, last emitted at 09:59:52, and a document at 10:00:10 opens a bucket at 10:00:07, past `extended_bounds.max`. Your bound lands exactly on 10:00:07 and excludes it.

Fixed in bf3aba42be: the upper bound is now `HistogramGrid.endOfLastBucket(from, to, interval)`, `from + (floorDiv(to - from, interval) + 1) * interval`, which collapses to `to + interval` on the aligned path — the three adapter fixtures are a day over ten minutes, so their expected JSON is unchanged, which is the no-op you predicted.

On the coverage gap you flagged in the same note: `WindowBoundaries` gained a case with a 65 minute window bucketed by 10, so it spans six whole intervals plus five minutes. The last emitted bucket ends one minute before the documents, while `to + interval` would have reached four minutes past them. I checked it fails on the previous revision of this branch and passes on this one, rather than assuming it — that was the lesson of the first round.

`HistogramGridTest.EndOfLastBucket` covers the helper directly, with your 15s window as one of the cases.

Full module: 821 tests, 0 failures.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapterTest.java
phiz71: On the third scope item, from your approval note — I went through the four sites and I do not think they say the same thing.

`hasSize((int) nbBuckets)` at 445-447 and the two event-analytics sequences at 1418 and 1446 are already deterministic. In each case the window and the documents come from the same reference: `nbBuckets` is derived from the same `from` and `to` the query uses, `atIndex(28)` targets the fixture hardcoded at 14:08 against a window starting at midnight, and the event-metrics sequences are built from `nowMinus`/`nowPlus`, which `TimeProvider.setTimestamps` derives from that same instant. They have passed every run today, spread across the morning, afternoon and evening.

The one that genuinely is not deterministic is the tolerance at 410-411, and re-deriving its window does not fix it. Its two contributors are a moving family — the V2 documents at `dateTimeYesterday` — and a fixed one, the metrics document at 14:08. What decides whether they share a ten minute bucket is the distance between the two, and that distance changes with the hour whatever window you wrap around them. It stops moving only when the sample data does.

Which is to say I put that item on the wrong ticket. I wrote it into FOUND-240 believing the adapters caused it; they do not. It is FOUND-235 cause 2, and it is recorded there — along with a correction, since the constraint analysis I posted on that ticket was built on the same misreading as the first revision of this branch.

Happy to be wrong on this one: if you see a way to pin 410-411 without pinning the fixtures, I will take it.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/adapter/SearchResponseStatusOverTimeAdapter.java
ytvnr: Confirmed on bf3aba42.

`endOfLastBucket` is the formula that holds at both ends. `from + (floorDiv(span, interval) + 1) * interval` gives `to + interval` when the window spans whole intervals, so the aligned path is untouched and the fixture literal in `SearchResponseStatusOverTimeAdapterTest` stayed valid without a change. On the 09:00:07 to 10:00:00 window with a 15s interval it lands on 10:00:07, the end of the 09:59:52 bucket, so the document at 10:00:10 falls outside and the series is 240 points rather than 241.

The repository case is the one that earns its keep: a 65 minute window on 10 minute buckets now stops at `to + 5min`, where the previous formula reached `to + 10min` and reported an eighth point. It fails on the previous revision, which is what makes it worth having.

Extracting `usableMillis` also means an unusable interval no longer has to be handled twice. Nothing further from me on this thread.

PR #19355, opened 2026-08-24
feat(gamma): serve Message APIs and their messages in observability logs

ThibaudAV:
## Issue

https://gravitee.atlassian.net/browse/ESM-143

## Description

Brings v4 **MESSAGE** APIs into the Gamma observability LOGS signal, and adds the message-level
trace behind a log detail. Two commits, each standing on its own.

### 1. Serve Message APIs in the LOGS signal

A Message API returned an empty page with no error: `narrowApiTypes()` intersects the requested
`API_TYPE` with `LOGS_SUPPORTED_API_TYPES`, which did not list `MESSAGE`, so the scope came out
empty. Widening that set is not enough on its own — two further gaps, both measured against real
connection documents on the SDK stack:

- **The default entrypoint scoping is an allow-list**, and `entrypoint-id` carries the plugin id. It
  held `http-get` and `http-post` but not `sse`, `websocket` or `webhook`, so a Message API exposed
  over any of those three stayed invisible — silently, since the filter narrows rather than fails.
- **The filters catalog answered six filters** for MESSAGE + LOGS (API, APPLICATION, PLAN,
  ENTRYPOINT, API_TYPE, RECORD_TYPE), leaving the screen with nothing useful to narrow on. The
  connection leg of a Message API *is* an HTTP request — which is why the analytics catalog already
  declares MESSAGE on every `HTTP_*` metric — so method, status, path, response time, error key,
  request id and transaction id all read fields its rows carry.

Those are grouped under a new `HTTP_CONNECTION_KINDS` set whose name states that intent, rather than
widening `HTTP_LLM_MCP_A2A` into a name that would no longer describe its members.

`PAYLOAD` stays out on purpose: `AnalyticsValidationServiceImpl` rejects `logging.content.payload`
on a MESSAGE api type, so there is no connection-level body for it to search.

No adapter change was needed — `buildSearchFilters()` translates by filter name, and every widened
filter already had its arm.

### 2. Serve the messages of a connection

A connection log answers "who connected and did it work". For an event-driven API the question worth
asking is what actually flowed, and Gamma had no way to answer it.

`GET /observability/logs/{requestId}/messages` returns one page of the messages that crossed the API
during that connection. Each entry joins the two documents the gateway records per message — the
entrypoint leg (client side) and the endpoint leg (broker side) — so a payload can be compared on
both sides of the gateway. Delegates to the existing `AggregatedMessageLogCrudService`.

Kept as a sub-resource rather than folded into the log detail: the detail serves every api type and a
message query would be wasted on the HTTP and native ones, and a single connection can carry
thousands of messages while a debugging session reads the first few. The console lazy-loads them in a
separate tab for the same reasons. `perPage` is capped at 100.

Guarded exactly like the detail, collapse to 404 included: a caller who may not read this API's logs
must not be able to tell a forbidden connection from an absent one.

## Tests

387 green in `gravitee-gamma-rest-api`. Three existing tests pinned the exclusion of MESSAGE and were
flipped — one used `MESSAGE` literally as its example of an unsupported type (now `EDGE`).

Verified end to end on the SDK stack, against an API built for it
(`MSG http-post+http-get+sse -> kafka`, ~150 metric documents):

| | Before | After |
|---|---|---|
| Logs of a Message API | 0 | **72** |
| Filters offered for MESSAGE + LOGS | 6 | **14** |
| `sse` connections visible | 0 (silently dropped) | **4** |
| `HTTP_STATUS=202` on a message API | unavailable | **65 rows** |
| `/{requestId}/messages` | — | PUBLISH, http-post → kafka, full payload |

Guards checked on the stack too: missing `apiId` → 400, unknown `requestId` → 200 with an empty page,
native API → empty page rather than an error, `perPage=5000` → capped.

## Additional context

- **Compatibility-sensitive**: adds a REST endpoint and widens the advertised filter catalog. Both
  are additive — no existing response shape changes.
- **The empty page is the normal answer for most APIs.** Message documents are only indexed when
  `analytics.logging` enables the message content flags, which are off by default and rejected
  outright on non-message api types.
- The UI side needs `@gravitee/gamma-lib-observability` with `loadMessages` on `messageLogPreset`
  (gravitee-io/gamma-lib-observability#82, released in v3.1.0).

## Scope beyond the gamma module

Rebased onto master after `ObservabilityEntrypoints` landed. That registry now declares, per
`entrypoint-id`, the scope observability gives it, and Gamma logs reads it through
`EntrypointScopeProvider` instead of its own `DEFAULT_ENTRYPOINT_IDS` — which this branch used to
widen. Two consequences worth calling out:

- **`SSE`, `WEBHOOK` and `WEBSOCKET` move from `Scope.EXCLUDED` to `Scope.LOGS_ONLY`.** The registry
  classified them as "Message and TCP APIs, outside what the observability signals cover today",
  which is the opposite of what this PR does. `LOGS_ONLY` is the scope `native-kafka` already uses:
  served by the logs signal, still absent from analytics, a divergence tracked by OBS-18. Resolving
  the conflict the other way would have left the old constant as dead code and silently returned SSE
  logs to invisible.
- `ObservabilityEntrypointsTest` asserted `containsExactly` on the logs scope, and its name said
  `native_kafka_only`. Both updated.

This touches a declaration shared with the analytics engine and owned by OBS-18/OBS-69, so it is no
longer gamma-only. Flagging it rather than leaving it to be discovered in the diff.

The NOTE about the logs/analytics divergence that this branch previously carried in
`SearchObservabilityLogsUseCase` is gone with the constant it documented; what it described now lives
in the registry, next to the scope it belongs to.

The last commit, `fix(analytics): scope message queries by api on the message documents`, changes the
**shared ES analytics engine**, not only Gamma. It lands here rather than in a second PR, so it is
worth stating plainly what it touches:

- `FilterAdapter.MESSAGE_FILTER_NAMES` gains `Filter.Name.API`. The API scope reached the message
  index only through the request ids the first phase resolved; message documents carry their own
  `api-id` (`MessageMetrics#apiId`), so the filter now applies there directly and the second phase
  stops depending on that set being complete.
- `MessageMeasuresQueryAdapter` stops inlining its own request-id clause and calls the shared
  `MessageRequestIdFilter.restrictTo`, which its javadoc already claimed. The two had come to
  disagree about an empty set — the shared one emits `terms` on an empty array, the inlined one
  dropped the clause and so widened the query.

**This reaches the v2 console message dashboards, not just Gamma.** Both changes narrow or leave
results unchanged; neither widens them.

Follow-ups from the review of #19356, both deliberately out of scope here:

- The ceiling on the first phase is untouched. `searchMessageConnectionRequestIDs` still pages 10k
  ids at a time up to 1000 iterations into a single `terms` clause, against Elasticsearch's default
  `index.max_terms_count` of 65,536. Skipping the join when only api-level filters are present is the
  change that would remove it.
- `AUTHZ` is missing from `REQUEST_BEARING_KINDS`, so authz APIs are offered neither `REQUEST_ID` nor
  `TRANSACTION_ID`. Pre-existing, now recorded in the javadoc rather than contradicted by it.


gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/observability/logs/LogsResource.java
```diff
@@ -147,6 +152,42 @@ public Response getLogDetail(@PathParam("requestId") String requestId, @QueryPar
      * id, so that is not a key here, and a decision is not an HTTP exchange so the merged log detail
      * shape does not fit. Same {@code apiId} defense and 404 collapse.
      */
+    /**
```
callaertanthony, line 172: **blocker (docs):** the new method landed between `getDecision`'s javadoc and `getDecision` itself.

At this point in the file there are now two javadoc blocks in a row: the one ending `"...a decision is not an HTTP exchange so the merged log detail shape does not fit. Same {@code apiId} defense and 404 collapse."` describes the **decisions** endpoint, but it is now attached to `getLogMessages`. `getDecision` further down is left with no javadoc at all.

Move `getLogMessages` and its own javadoc below `getDecision`.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
```diff
@@ -864,4 +875,63 @@ private static io.gravitee.gamma.rest.core.observability.filter.model.ApiType to
...
+            apiId,
+            requestId,
+            new PageableImpl(page, perPage)
+        );
+
+        if (result == null || result.logs() == null || result.logs().isEmpty()) {
+            return MessageLogsPage.EMPTY;
+        }
+
+        return new MessageLogsPage(result.logs().stream().map(ObservabilityLogsDataPortAdapter::mapToMessageLog).toList(), result.total());
```
callaertanthony: **blocker:** an out-of-range page loses `totalCount`.

The early return fires whenever `logs()` is empty, which includes "this page is past the end" — and it throws away `result.total()`. Ask for `page=3` of a 30-message connection at `perPage=20` and the response is `data: []` with `totalCount: 0` and `pageCount: 0`, so the client cannot tell "past the end" from "this connection recorded no messages", and cannot render a paginator to get back.

`searchLogs` in this same class already does it the right way: it null-guards `logs()` but keeps `result.total()`. Suggest matching it:

```java
var logs = result.logs() == null ? List.<AggregatedMessageLog>of() : result.logs();
return new MessageLogsPage(logs.stream().map(ObservabilityLogsDataPortAdapter::mapToMessageLog).toList(), result.total());
```

That also removes the need for the `result == null` guard (nothing else in this class defends against it) and leaves `MessageLogsPage.EMPTY` unused in main code — worth deleting if so.

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapterTest.java
```diff
@@ -727,6 +735,16 @@ private SearchLogsFilters captureSearchFilters() {
...
 
+        /**
+         * The detail resolves the api type from the caller's accessible set, so every detail test goes
+         * through the loader. These cases are about the credential, not about the api type: hand the
+         * context straight back so the lookup finds no api and the type stays null.
+         */
+        @BeforeEach
+        void stubUserContext() {
+            when(userContextLoader.loadApi(any(), eq("api-1"))).thenAnswer(invocation -> invocation.getArgument(0));
+        }
```
callaertanthony, line 787: **improvement:** the two projection commits ship with no test that would fail without them.

Grepping the branch diff, `LogDetail.apiType` and `LogEntry.entrypointId` appear only in main code and in the OpenAPI — never in an assertion. Delete `.apiType(apiType)` in `getLogDetail`, or `.entrypointId(log.getEntrypointId())` in `mapToLogEntry`, and the suite stays green.

This stub makes it slightly worse: it hands the context straight back so the lookup finds nothing and the type stays **null**, which pins the absent path rather than the feature. `AGENTS.md` asks for a test that would fail without the change, for every behaviour change.

Two small additions would close it: one detail test where `loadApi` returns an api and `apiType` comes back as e.g. `"MESSAGE"`, and one assertion on `entrypointId` in the existing search-logs mapping test.

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/core/observability/filter/use_case/GetObservabilityFilterDefinitionsUseCaseTest.java
```diff
@@ -72,7 +72,36 @@ void should_intersect_both_axes_independently() {
 
         List<String> names = output.filters().stream().map(FilterSpec::name).toList();
         assertThat(names).contains("MESSAGE_OPERATION_TYPE", "MESSAGE_SIZE", "API_TYPE");
-        // URI is HTTP_PROXY/LLM/MCP only → excluded for MESSAGE; REQUEST_ID is LOGS-only → excluded for ANALYTICS.
-        assertThat(names).doesNotContain("URI", "REQUEST_ID");
+        // REQUEST_ID is LOGS-only → excluded for ANALYTICS.
+        assertThat(names).doesNotContain("REQUEST_ID");
```
callaertanthony: **nit:** `should_intersect_both_axes_independently` now only exercises one axis.

Dropping `URI` from the assertion was necessary, but it was the only thing covering the **api-type** axis; what is left checks the signal axis alone, which no longer matches the test's name.

`HOST` is still a good stand-in — it is `Defs.ANALYTICS` + `Defs.HTTP_LLM_MCP_A2A`, so it stays excluded for MESSAGE:

```java
// HOST is HTTP/LLM/MCP/A2A only → excluded for MESSAGE; REQUEST_ID is LOGS-only → excluded for ANALYTICS.
assertThat(names).doesNotContain("HOST", "REQUEST_ID");
```

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/model/StaticFilters.java
```diff
@@ -284,11 +296,27 @@ private static final class Defs {
...
+         * already declares MESSAGE on every {@code HTTP_*} metric. Filters reading a field of that
+         * document (method, status, path, response time) therefore apply to it unchanged.
+         */
+        private static final Set<ApiType> HTTP_CONNECTION_KINDS = Set.of(
             ApiType.HTTP_PROXY,
             ApiType.LLM,
             ApiType.MCP,
             ApiType.A2A,
+            ApiType.MESSAGE
+        );
```
callaertanthony, line 316: **nit:** this widening reaches the ANALYTICS signal too, and only half-way.

All five filters moved onto `HTTP_CONNECTION_KINDS` are declared `Defs.LOGS_ANALYTICS`, so MESSAGE now also gets `HTTP_METHOD`, `HTTP_STATUS`, `HTTP_STATUS_CODE_GROUP`, `URI` and `HTTP_GATEWAY_RESPONSE_TIME` on analytics screens. Meanwhile `HTTP_GATEWAY_LATENCY` and `HOST` stay excluded purely because they happen to be `Defs.ANALYTICS`-only — even though `FilterAdapter.HTTP_FILTER_NAMES` carries all seven, so the connection phase of a message query supports them identically.

So the MESSAGE boundary now follows the signal, which was never the reason MESSAGE was excluded. Either widen those two as well, or add a line to this javadoc saying analytics-only filters are deliberately left for the analytics PR.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/use_case/SearchObservabilityLogsUseCase.java
```diff
@@ -54,23 +54,27 @@ public class SearchObservabilityLogsUseCase {
...
 
     /**
      * Canonical entrypoints applied when no explicit entrypoint filter is set. The HTTP subset
      * mirrors {@code FilterAdapter.httpFilter()} from the analytics ES adapter; {@code native-kafka}
-     * is additionally included so native connection documents are served by the LOGS signal.
+     * is additionally included so native connection documents are served by the LOGS signal, and
+     * {@code sse} / {@code websocket} / {@code webhook} so are the async entrypoints of Message
+     * APIs — {@code entrypoint-id} carries the plugin id, so a Message API exposed over any of the
+     * three would otherwise yield an empty page with no error (their http-get / http-post siblings
+     * were already covered).
```
callaertanthony, line 77: **nit:** the NOTE just below is now stale.

It warns about the logs/analytics divergence for `native-kafka` only, but `sse` / `websocket` / `webhook` create exactly the same one: `FilterAdapter.httpFilter()` lists only `http-get`, `http-post`, `http-proxy`, `llm-proxy`, `mcp-proxy`, `a2a-proxy`. A Message API exposed over SSE will therefore show N connections in the logs list and 0 on the dashboard — the largest case of the divergence, and the one a user is most likely to hit now that Message APIs are served here.

Worth extending the NOTE to name the three new ids.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/model/StaticFilters.java
```diff
@@ -304,14 +332,15 @@ private static final class Defs {
 
         /**
          * Every kind whose rows carry a request id, decisions included. Named rather than
-         * {@link ApiType#ALL}: that would also advertise the filter for MESSAGE and EDGE, which the
-         * logs search does not serve.
+         * {@link ApiType#ALL}: that would also advertise the filter for EDGE, which the logs search
+         * does not serve.
```
callaertanthony, line 340: **nit:** the reworded sentence now claims EDGE is the only exclusion, but the set also omits `AUTHZ` — which *is* in `SearchObservabilityLogsUseCase.LOGS_SUPPORTED_API_TYPES`, so `REQUEST_ID` and `TRANSACTION_ID` are not offered for authz APIs.

Pre-existing gap, not introduced here; just worth not asserting the opposite while the line is being touched.

callaertanthony (commented):
Reviewed the whole branch. The feature holds up: the security guard on `/{requestId}/messages` is right (apiId required, `API_LOG[READ]` checked, 404 collapse, env-scoped), all five widened filter names already have an arm in `buildSearchFilters()`, `entrypointId` really is populated (MapStruct maps `Metrics.entrypointId` -> `BaseConnectionLog.entrypointId`), and the message index sorts `@timestamp desc` so the spec's "newest first" is accurate.

Two things I'd fix before merge (inline), plus a few smaller ones.

**Scope: the PR body describes two commits, the branch has five.**

`fix(analytics): scope message queries by api on the message documents` changes the shared ES analytics engine — `FilterAdapter.MESSAGE_FILTER_NAMES` gains `Filter.Name.API`, and `MessageMeasuresQueryAdapter` now emits an empty `terms` on `request-id` instead of dropping the clause. That is a behaviour change to v2 console message dashboards, not only to Gamma, and the body says the opposite ("Message analytics is a separate concern and is fixed in #TBD on this repo"). Please describe it here or split it out.

The change itself looks correct, for the record: `MessageFieldResolver.fromFilter` falls through to `HTTPFieldResolver` so `API` resolves to `api-id`, which that resolver's own javadoc confirms is stamped on every message document; and `AnalyticsElasticsearchRepository.searchMessageMeasures` already returns early on an empty id set, so the empty-`terms` clause is unreachable in practice.

The other two undescribed commits (`project the connection entrypoint onto a log row`, `project the api type onto a log detail`) are the untested ones — see the inline comment.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/resources/observability/logs/LogsResource.java
ThibaudAV: Fixed. `getLogMessages` and its own javadoc now sit below `getDecision`, and the decisions javadoc is
back on the method it describes. Insertion mistake on my side — it landed inside the javadoc/method
pair rather than after it.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
ThibaudAV: Fixed, following `searchLogs` in this class as you suggested:

```java
var logs = result.logs() == null ? List.<AggregatedMessageLog>of() : result.logs();
return new MessageLogsPage(logs.stream().map(ObservabilityLogsDataPortAdapter::mapToMessageLog).toList(), result.total());
```

The `result == null` guard is gone with it, since nothing else in the class defends against that.

I kept `MessageLogsPage.EMPTY` rather than deleting it: it is still the value the port contract names
in its javadoc, and four tests assert against it. It stays reachable by equality too — an empty page
with a zero total is `new MessageLogsPage(List.of(), 0)`, which is `EMPTY`. What changed is that a
page past the end no longer collapses onto it.

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapterTest.java
ThibaudAV: You are right, and the stub made it worse — it pinned the null path, which is the opposite of what the
commits do.

Added a `Projections` nested class with one test per field: `should_carry_the_connection_entrypoint_onto_the_log_row`
maps a `BaseConnectionLog` carrying `entrypointId` and asserts it on the returned `LogEntry`, and
`should_resolve_the_api_type_onto_the_log_detail` stubs `loadApi` to return a `MESSAGE` api and asserts
`apiType` comes back as `"MESSAGE"`.

Checked they guard rather than assuming it: with `.entrypointId(...)` and `.apiType(...)` neutralised in
the adapter, both fail — one per projection.

The two projection commits are now one. They answer the same question and share this test class, so
splitting them was artificial.

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/core/observability/filter/use_case/GetObservabilityFilterDefinitionsUseCaseTest.java
ThibaudAV: Fixed, but `HOST` no longer works as the witness — it is one of the two filters widened in the sibling
comment, so it is now offered for MESSAGE.

Used `GEO_IP_COUNTRY` instead: same shape you were reaching for (`Defs.ANALYTICS` + `Defs.HTTP_LLM_MCP_A2A`),
so it stays excluded by api type after the widening.

```java
// One exclusion per axis, so the test keeps covering both: GEO_IP_COUNTRY is HTTP/LLM/MCP/A2A
// only → excluded by api type; REQUEST_ID is LOGS-only → excluded by signal.
assertThat(names).doesNotContain("GEO_IP_COUNTRY", "REQUEST_ID");
```

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/model/StaticFilters.java
ThibaudAV: Widened both, rather than documenting the gap — your reasoning holds and the "left for the analytics
PR" argument no longer applies now that the analytics change rides in this PR.

`HOST` and `HTTP_GATEWAY_LATENCY` move to `HTTP_CONNECTION_KINDS`. The javadoc now states the rule
explicitly, since that was the part that was implicit:

> The boundary is the document, not the signal: `FilterAdapter.HTTP_FILTER_NAMES` carries host and
> latency alongside the rest, so the analytics-only filters belong here too. Splitting them off would
> make MESSAGE follow the signal, which was never why it was excluded.

Knock-on effect on the test in the sibling comment — see there.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/logs/use_case/SearchObservabilityLogsUseCase.java
ThibaudAV: Extended. The NOTE now names the divergence rather than attributing it to `native-kafka` alone, and
says which case matters:

> `FilterAdapter.httpFilter()` lists only http-get, http-post, http-proxy, llm-proxy, mcp-proxy and
> a2a-proxy, so every id added here — `native-kafka`, and now `sse` / `websocket` / `webhook` — shows
> N connections in the logs list and 0 on the dashboard. The async trio is the largest case and the
> one a user is most likely to hit, since a Message API exposed over SSE is served here in full while
> its dashboard stays empty.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/model/StaticFilters.java
ThibaudAV: Corrected — the line now records the AUTHZ gap instead of implying it away:

> AUTHZ is absent too, though the logs search does serve it — a pre-existing gap, not a decision taken
> here: an authz API is offered neither `REQUEST_ID` nor `TRANSACTION_ID`.

Left as a gap rather than fixed in this PR, since offering those two for authz rows is a behaviour
change with no test coverage here and nothing in this branch depends on it.

ThibaudAV:
Rebased onto master — all seven points are addressed, and the rebase turned one of them into something
worth flagging before you re-read.

**`ObservabilityEntrypoints` changed what this branch had to touch.**

The registry landed while this PR was open, and Gamma logs now reads its scope through
`EntrypointScopeProvider` instead of the `DEFAULT_ENTRYPOINT_IDS` constant this branch used to widen.
The registry classified `SSE`, `WEBHOOK` and `WEBSOCKET` as `Scope.EXCLUDED` — *"Message and TCP APIs,
outside what the observability signals cover today"* — which is the opposite of what this PR does.

Taking the constant's side of the conflict would have compiled and stayed green: nothing reads it any
more, so the widening would have become dead code and SSE logs would have silently gone back to
invisible, undoing the 0 → 4 connections this PR measured. So the three moved to `Scope.LOGS_ONLY`
instead — the scope `native-kafka` already uses: served by logs, still absent from analytics, the
divergence OBS-18 tracks. `EXCLUDED`'s javadoc lost its "Message and" accordingly.

**This is a shared declaration owned by OBS-18/OBS-69, so it is no longer gamma-only.** Same category
as the scope point you raised on the analytics commit, which is why I am not leaving it to the diff.
It is in the PR body too.

**One of your seven changed shape rather than being dropped.** The NOTE you asked me to extend
(`SearchObservabilityLogsUseCase`, the logs/analytics divergence) lived in the block master deleted, so
it is gone from the diff. What it described is now encoded in the registry — a `LOGS_ONLY` scope plus a
comment on the three ids — which is the better place for it. Flagging it so it does not read as a
regression.

`ObservabilityEntrypointsTest` asserted `containsExactly` on the logs scope and was named
`..._with_native_kafka_only_for_logs`; both updated. It did its job — it caught the scope change on CI.

The other six are unchanged and re-verified against the rebased tree: javadoc order in `LogsResource`,
`result.total()` kept on an out-of-range page, the two `Projections` tests (checked they still fail with
the projections neutralised), `GEO_IP_COUNTRY` as the api-type witness, `HOST` and
`HTTP_GATEWAY_LATENCY` widened, and the AUTHZ gap recorded rather than contradicted.

35 checks green.

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
```diff
@@ -1870,6 +2031,14 @@ components:
                 endpoint:
                     type: string
                     description: Backend endpoint the request was proxied to.
+                entrypointId:
```
callaertanthony: **blocker (contract):** the two new projections are documented on each other's schema.

This `entrypointId` block sits in the `LogDetail` schema (opens at line 1989; `LogEntry` does not start until 2192). But `entrypointId` was added to `LogEntryDto`, not `LogDetailDto` — `LogDetailDto` has no such component:

```
LogDetailDto.java:40   String apiType,       <- new, absent from the LogDetail schema
LogDetailDto.java:48   String endpoint,
LogDetailDto.java:49   String host,          <- no entrypointId between them
LogEntryDto.java:59    String entrypointId,  <- new, absent from the LogEntry schema
```

So the published contract has three problems at once: it declares `LogDetail.entrypointId`, which `GET /observability/logs/{requestId}` never returns; and it declares neither `LogDetail.apiType` nor `LogEntry.entrypointId`, both of which the API now does return. A client generated from this spec — the `@gravitee/gamma-lib-observability` side named in the PR body — drops both new fields and gains one that is always null.

The description text gives the slip away: "one API commonly exposes several at once, which makes it the column that tells its rows apart" is about list rows, i.e. `LogEntry`. The `endpoint:` -> `host:` adjacency is identical in both records, which is how it landed one schema up.

Move this block to the `LogEntry` schema, and add `apiType` to `LogDetail` next to `apiId`.

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapterTest.java
```diff
@@ -1051,4 +1110,52 @@ private LogsSearchQuery decisionQueryWith(FilterCondition... conditions) {
...
+            var mapped = page.data().getFirst();
+            assertThat(mapped.operation()).isEqualTo("PUBLISH");
+            assertThat(mapped.timestamp()).isEqualTo("2026-08-24T13:14:59.395Z");
+            assertThat(mapped.entrypoint().connectorId()).isEqualTo("http-post");
+            assertThat(mapped.entrypoint().headers()).containsKey("content-type");
+            assertThat(mapped.endpoint().connectorId()).isEqualTo("kafka");
+        }
+
+        @Test
+        void should_return_an_empty_page_when_the_api_records_no_message_logs() {
```
callaertanthony, line 1175: **blocker (test gap):** the `result.total()` fix from the last round has no test that can fail.

`SearchMessages` has exactly two cases: one message with `total=1`, and this one with `total=0` plus an empty list. Revert `searchMessages` to the early return it had before —

```java
if (result.logs() == null || result.logs().isEmpty()) return MessageLogsPage.EMPTY;
```

— and **both still pass**, because this case's total is already 0. The suite is green either way, so it is not evidence the fix is in.

The reported failure was the out-of-range page: page 3 of a 30-message connection at `perPage=20` answered `totalCount: 0`, so the client could not tell "past the end" from "nothing recorded". That is precisely the case with no test.

One more case pins it:

```java
@Test
void should_keep_the_total_on_a_page_past_the_end() {
    when(aggregatedMessageLogCrudService.searchApiAggregatedMessageLog(any(), eq("api-1"), eq("req-1"), any())).thenReturn(
        new SearchLogsResponse<>(30L, List.of())
    );

    var page = adapter.searchMessages("org-1", "env-1", "api-1", "req-1", 3, 20);

    assertThat(page.data()).isEmpty();
    assertThat(page.totalCount()).isEqualTo(30);
}
```

Worth mutation-checking it rather than trusting it: restore the early return, confirm it goes red, put the fix back.

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
```diff
@@ -393,6 +393,112 @@ paths:
...
+                                              endpoint:
+                                                  id: "e279b068-7d94-4d2b-8853-9f5ed4bb4c1d"
+                                                  connectorId: "kafka"
+                                                  payload: '{"order":200}'
+                                                  isError: false
+                                        pagination:
+                                            totalCount: 1
+                                            page: 1
+                                            pageCount: 1
+                "403":
```
callaertanthony: **improvement:** this operation documents a 403 it cannot return, and omits the 400 it does.

`getLogMessages` calls `requireParam("apiId", apiId)` -> `MissingLogScopeException` -> **400**, then `checkApiLogReadPermissionOrCollapse(apiId, requestId)` -> **404**. It never calls `checkReadObservabilityPermission()`, and neither `LogsResource` nor `GammaObservabilityResource` carries a class-level environment guard, so nothing on this path can produce a 403.

The two sibling detail operations in this same file get it right — `getObservabilityLogDetail` and `getAuthzDecision` both document 200/400/404 and no 403, which is the guard shape this endpoint deliberately copies.

Concretely: a caller who omits `apiId` gets an undocumented 400, and a caller who handles the documented 403 writes dead code. Swap the `"403"` entry for a `"400"` describing the missing `apiId`.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
```diff
@@ -176,7 +181,12 @@ public Optional<LogDetail> getLogDetail(String organizationId, String environmen
             return Optional.empty();
         }
 
-        var builder = LogDetail.builder().requestId(requestId).apiId(apiId);
+        // The connection indices do not store the api type, so it comes from the caller's accessible
+        // set — the same source the list rows use. Resolved here rather than left to the client: a
+        // direct link to a detail has no list row to carry it from.
+        var apiType = loadAccessibleApi(organizationId, environmentId, apiId).map(AccessibleApi::type).map(Enum::name).orElse(null);
```
callaertanthony: **improvement:** this puts a full authorization enumeration on the log-detail path to obtain one enum name.

`loadAccessibleApi` delegates to `UserContextLoaderImpl.loadApi`, which for any non-admin does:

```java
Set<String> userApiIds = apiAuthorizationService.findApiIdsByUserId(executionContext, userId, null, true);
if (!userApiIds.contains(apiId)) { ... }
```

Every API id the user can reach is materialised, purely to `contains(apiId)`, before the by-id repository search that actually yields the type. On an endpoint whose own javadoc calls itself heavy and lazy, an operator with access to thousands of APIs pays that scan on every click into a log row.

The difference from `getDecision`, which uses the same call legitimately: there `loadAccessibleApi` *is* the scoping guard. Here it is only a projection — the resource already ran `checkApiLogReadPermissionOrCollapse(apiId, requestId)`, so this `apiId` is known and already authorized, and a direct by-id lookup would return the same value.

Not blocking, and the comment above the line explains the *why* of resolving it server-side well. It is the *how* that is more expensive than the value it fetches.

callaertanthony (commented):
Re-read against 85d6050. The seven earlier points all check out on the rebased tree: javadoc order restored in `LogsResource`, `result.total()` kept, the two `Projections` tests genuinely pin their fields, `GEO_IP_COUNTRY` works as the api-type witness, `HOST` and `HTTP_GATEWAY_LATENCY` widened, AUTHZ gap recorded. I also verified the `ObservabilityEntrypoints` call: `DEFAULT_ENTRYPOINT_IDS` really is gone from master, so taking the constant's side would have been dead code; and `LOGS_SCOPE_IDS` has exactly one consumer (`EntrypointScopeProviderAdapter`) while `HTTP_SCOPE_IDS` is untouched, so `LOGS_ONLY` on the async trio cannot move analytics totals. Agreed with the resolution.

Four new points below, two blocking. Both blockers are on the same theme: something the code now does that nothing outside the code knows about — the OpenAPI contract in one case, the test suite in the other.

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
ThibaudAV: Fixed, and confirmed all three problems before touching it: `LogDetailDto` has `apiType` and no
`entrypointId`, `LogEntryDto` has `entrypointId`, and `apiType` appeared nowhere as a schema property.

- `entrypointId` moved out of `LogDetail` into `LogEntry`, between `endpoint` and `host` to match the
  record.
- `apiType` added to `LogDetail`, next to `apiId`.
- `LogEntry.apiType` was already documented, so it needed nothing.

Your read of the cause is right — the `endpoint:` → `host:` adjacency is identical in both records.

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapterTest.java
ThibaudAV: You are right, and it is the same miss as the projections last round: I applied the "must fail without
the change" rule where you pointed it, not to the fix I was writing at the same time.

Added `should_keep_the_total_on_a_page_past_the_end` with `total=30` and an empty list at page 3.
Mutation-checked as you asked: with the early return restored, 66 tests run and exactly one fails —
the new one. Restored the fix, 410 green.

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
ThibaudAV: Fixed — `"403"` swapped for a `"400"` describing the missing `apiId`, which lines this operation up
with `getObservabilityLogDetail` and `getAuthzDecision` in the same file.

Verified the path rather than taking it on trust: `getLogMessages` runs `requireParam` →
`MissingLogScopeException` → 400, then `checkApiLogReadPermissionOrCollapse` → 404 on refusal, and
never calls `checkReadObservabilityPermission()`.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/infra/adapter/ObservabilityLogsDataPortAdapter.java
ThibaudAV: Done — `apiCrudService.findById(apiId)` replaces the `loadAccessibleApi` call on the detail path.
`ApiCrudService` is injected into the adapter and wired in `GammaLogsConfiguration`.

Checked your premise first, since this is authorization-adjacent: `getLogDetail` in `LogsResource`
runs `requireParam` then `checkApiLogReadPermissionOrCollapse(apiId, requestId)` — which is
`permissionService.hasPermission(ctx, API_LOG, apiId, READ)` with a 404 collapse — before the use case
is reached. So the id is authorized by the time the adapter sees it, and `getDecision` keeps
`loadAccessibleApi` where it really is the scoping guard.

Two things the swap would have broken silently, so they are explicit in the code:

- **Environment scope.** `loadAccessibleApi` scopes through `ApiCriteria.environmentId`; `findById`
  has no scope of its own. Kept as `.filter(api -> api.belongsToEnvironment(environmentId))`, with
  `should_not_resolve_the_api_type_of_another_environment` pinning it.
- **The wire name.** `toGammaApiType` is not the identity — `PROXY` → `HTTP_PROXY`, `LLM_PROXY` → `LLM`,
  `MCP_PROXY` → `MCP`, `A2A_PROXY` → `A2A`. Calling `Enum::name` on the definition type directly would
  have published `"PROXY"` and broken the client's discriminant, with every existing test still green
  since they all use MESSAGE or AUTHZ. The conversion is kept.

410 tests green in the module, 905 in repository-elasticsearch, 71 in repository-api.

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
```diff
@@ -1840,6 +2001,14 @@ components:
                     type: string
                     description: Identifier of the API that handled the request.
                     example: "67e246cb-ab02-4da4-8f47-f9fa3e061d6b"
+                apiType:
+                    type: string
+                    description: |
+                        API kind as stored in the definition (e.g. `HTTP_PROXY`, `MESSAGE`, `NATIVE`).
+                        Resolved from the caller's accessible set rather than read off the connection
```
callaertanthony: **nit:** this sentence describes the mechanism the `findById` swap replaced.

`apiType` is no longer "resolved from the caller's accessible set" — the adapter now does `apiCrudService.findById(apiId).filter(api -> api.belongsToEnvironment(environmentId))`. The distinction is not academic for whoever reads this description: the accessible set was a per-caller thing, so the old wording implies the field could differ between two callers on the same connection, which is exactly what it no longer does.

Same sentence in `LogDetail.java:40`:

```java
* Api type of the log's API, resolved from the caller's accessible set rather than from the
* document — the connection indices do not store it.
```

The adapter's own inline comment was updated correctly in the same commit; these two lagged. "Read by id from the API definition, scoped to the environment" covers both, and the rest of each paragraph — why the client needs it at all — still holds unchanged.

callaertanthony (commented):
All four verified fixed on 7fbac974, threads resolved.

I re-checked the premise of the `findById` swap myself rather than taking it on trust, since it moves an authorization-adjacent call: `getLogDetail` runs `requireParam` then `checkApiLogReadPermissionOrCollapse` (`permissionService.hasPermission(ctx, API_LOG, apiId, READ)`, 404 collapse) before the use case, so the id really is authorized by the time the adapter reads it; `ApiCrudService.findById` returns `Optional<Api>` rather than throwing, so an unknown id yields a null type instead of a 500; and `belongsToEnvironment` restores the one dimension the accessible-set read was providing, pinned by `should_not_resolve_the_api_type_of_another_environment`. Good catch keeping `toGammaApiType` — `PROXY` on the wire would have been a silent client break.

One nit below, on the wording the swap left behind. 35 checks green.

callaertanthony (approved):
Approving on 7fbac974. Eleven points across three rounds, all verified against the head rather than the reply.

What I checked directly, since these are the parts that reach beyond gamma:

- **`ObservabilityEntrypoints`.** `DEFAULT_ENTRYPOINT_IDS` really is gone from master, so keeping the constant's side of the rebase would have been dead code and would have returned SSE logs to invisible. `LOGS_SCOPE_IDS` has exactly one consumer (`EntrypointScopeProviderAdapter`) and `HTTP_SCOPE_IDS` is untouched, so `LOGS_ONLY` on the async trio cannot move analytics totals. The resolution is the right one.
- **`FilterAdapter.MESSAGE_FILTER_NAMES += API`.** Resolves to `api-id`, which message documents carry (`MessageMetrics#apiId`). Narrows only.
- **`MessageRequestIdFilter.restrictTo`.** The empty-set difference between the shared filter and the inlined one is unobservable: all three callers early-return before reaching the adapter.
- **The `findById` swap.** Authorization-adjacent, so I traced it rather than trusting the reply: `checkApiLogReadPermissionOrCollapse` runs before the use case, `findById` returns an `Optional` rather than throwing, and `belongsToEnvironment` restores the one dimension the accessible-set read was providing. Keeping `toGammaApiType` avoided a silent client break on `PROXY`.

The one open thread is a doc-wording nit — the `apiType` description still names the mechanism the swap replaced. Not a merge blocker; worth folding in if you touch the file again.

Good discipline on the mutation checks. The `totalCount` case was the one that mattered: the fix was correct from the first round, but nothing would have told us if it regressed.

gravitee-gamma/gravitee-gamma-rest-api/src/main/resources/openapi/observability/openapi-observability.yaml
ThibaudAV: Folded in on `fe6d574e`, since the approval survives a push here (`dismiss_stale_reviews=false`) and a
published description naming a mechanism that no longer exists ages badly.

Both places you named now read "Read by id from the API definition, scoped to the environment", and the
rest of each paragraph is unchanged.

Grepping the phrase turned up a third one you did not name — the `Projections` javadoc in
`ObservabilityLogsDataPortAdapterTest`, which still said the type came from the accessible set. Fixed
too.

The two remaining mentions are deliberate: `GetAuthzDecisionUseCase:30` and the `getAuthzDecision`
description in the spec both describe the decision path, which still goes through `loadAccessibleApi`
(adapter line 502) — there the accessible-set read *is* the scoping guard, so the wording is accurate
and stays.

410 tests green in the module.

PR #19356, opened 2026-08-24
fix(analytics): unbreak message analytics and add facets and time series

ThibaudAV:
## Issue

https://gravitee.atlassian.net/browse/ESM-143

## Description

Message analytics were broken in a way nothing surfaced: **a Message API exposed over `http-get` or
`http-post` had none at all** — not a missing chart, zero everywhere, measures included. On top of
that, only measures were implemented; facets and time series threw.

### 1. Stop hiding Message APIs served over the shared HTTP entrypoints

The message query selected its connection documents with `must_not(httpFilter())`. That read as "not
a plain HTTP proxy" when it was written (GKO-1609, `db8442a2`, Nov 2025), because `httpFilter()` then
held a single id, `http-proxy`. It has been widened three times since, **each time for a need on the
HTTP side**, and each widening silently narrowed this one:

| Commit | Added to `httpFilter()` | Why |
|---|---|---|
| `db8442a2` | `http-proxy` | creation |
| `a67d1c50` — **GMA-513** | `http-get`, `http-post` | widgets returned 0 for LLM/MCP |
| `a610d378` | `a2a-proxy` | A2A support |

GMA-513 finished it: adding those two ids to fix LLM/MCP dashboards removed the same entrypoints from
the message side. Nothing caught it — no test covers this predicate from the message side.

The root cause is one predicate answering two opposite questions: "is this HTTP" and "is this
message", defined as its complement. Neither can move without moving the other.

**Removing it is safe rather than merely convenient.** Which API a connection belongs to is already
carried by the `API` filter `ApiTypeFilterTransformer` appends to *every* query, and the second phase
keeps only request ids that have message documents — an index no other api type writes to. The
predicate only ever narrowed the request-id set; it was never what made the result correct.

An entrypoint allow-list was the other option and is rejected on purpose: it restates the same guess,
and the next async connector would reproduce the bug. Entrypoint ids name plugins, not api types.

Adds the regression guard that was missing, stated from the side that broke.

### 2. Facets and time series

`MessageDataPlaneQueryService` answered measures and threw `UnsupportedOperationException` for the
other two, so a Message API could show totals and nothing else: no split by operation or connector,
no trend. Both new adapters run on the second phase of the message join and take the request ids the
first phase resolved. Measure building is shared with the measures adapter rather than copied.

### 3. A trap the wiring exposed

The catalog declared `APPLICATION`, `PLAN`, `TENANT` and `ZONE` as facets of every message metric,
and `MessageFieldResolver` delegated unknown facets to the HTTP resolver — which answers `plan-id`
and `application-id`, **fields the message index does not have**. Elasticsearch does not fail a terms
aggregation on a missing field; it returns no buckets. "Messages by application" would have rendered
as an empty chart with nothing to diagnose.

Verified against real documents:

| Facet | ES field | Buckets |
|---|---|---|
| `MESSAGE_OPERATION_TYPE` | `operation` | ✅ 2 |
| `MESSAGE_CONNECTOR_TYPE` | `connector-type` | ✅ 2 |
| `MESSAGE_CONNECTOR_ID` | `connector-id` | ✅ 3 |
| `API` / `GATEWAY` | `api-id` / `gateway` | ✅ |
| `APPLICATION` · `PLAN` · `TENANT` · `ZONE` | absent | ❌ **0** |

Fixed at both ends — the catalog declares only what the index carries, and the resolver throws
instead of delegating, so a future drift fails loudly. **Breaking messages down by application or
plan needs the dimension carried across the request-id join first**; until then those axes belong to
the `HTTP_*` metrics, which hold them on the connection document.

## Tests

815 green in `gravitee-apim-repository-elasticsearch`, Testcontainers suites included. New:
`MessageFacetsQueryAdapterTest`, `MessageTimeSeriesQueryAdapterTest`, `MessageFieldResolverTest`, and
the GMA-513 guard in `HTTPFacetsQueryAdapterTest`.

Verified on the SDK stack against an API built for it (`MSG http-post+http-get+sse -> kafka`):

| | Before | After |
|---|---|---|
| `MESSAGES` COUNT | **30** (the SSE messages only) | **260** |
| `MESSAGE_PAYLOAD_SIZE` AVG | 32 B | **45.9 B** |

The 45 `http-post` and 6 `http-get` message documents that were structurally invisible are now
counted.

## Additional context

- **Compatibility-sensitive**: `AnalyticsRepository` gains two methods (`NoOpAnalyticsRepository`
  updated); the analytics catalog drops four facets that never worked. `FacetsResponseAdapter.empty`
  and `TimeSeriesResponseAdapter.empty` become public, matching their `MeasuresResponseAdapter`
  sibling — they keep the requested metric names in an empty result, which rebuilding by hand would
  drop.
- **Worth a second opinion from whoever owns GMA-513** — the fix is in their blast radius, and the
  regression guard is written so the two concerns cannot collide again.
- Noticed while testing, out of scope here: a malformed filter (`"values"` instead of `"value"`) is
  silently ignored and answers 200 with zeros rather than being rejected.
- The logs half of ESM-143 is #19355.

### DIFF
--- a/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/MessageFacetsQueryAdapter.java
+++ b/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/MessageFacetsQueryAdapter.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.repository.elasticsearch.v4.analytics.engine.adapter;
+
+import io.gravitee.repository.analytics.engine.api.query.Facet;
+import io.gravitee.repository.analytics.engine.api.query.FacetsQuery;
+import io.gravitee.repository.analytics.engine.api.query.MetricMeasuresQuery;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Breaks message measures down by a dimension — operation, connector, API, application.
+ *
+ * <p>Runs on the second phase of the message join: the caller resolves the connection documents
+ * first and passes their request ids in, because the message index carries none of the connection
+ * dimensions (no plan-id, no app-id, no entrypoint-id) and could not be filtered by them directly.
+ */
+public class MessageFacetsQueryAdapter {
+
+    private final MessageMeasuresQueryAdapter measuresAdapter = new MessageMeasuresQueryAdapter();
+    private final FilterAdapter filterAdapter = new FilterAdapter(new MessageFieldResolver());
+    private final BoolQueryAdapter boolAdapter = new BoolQueryAdapter(filterAdapter);
+
+    public String adapt(FacetsQuery query, Set<String> requestIDs) {
+        return json(query, requestIDs).toString();
+    }
+
+    private JsonObject json(FacetsQuery query, Set<String> requestIDs) {
+        var boolQuery = boolAdapter.messageFilter(query);
+        MessageRequestIdFilter.restrictTo(boolQuery, requestIDs);
+
+        return new JsonObject()
+            .put("size", 0)
+            .put("query", JsonObject.of("bool", boolQuery))
+            .put("aggs", adaptFacets(query.metrics(), query.facets(), query.limit()));
+    }
+
+    JsonObject adaptFacets(List<MetricMeasuresQuery> metrics, List<Facet> facets, Integer limit) {
+        var aggs = new JsonObject();
+        for (var metric : metrics) {
+            aggs.mergeIn(adaptFacets(metric, facets, limit));
+        }
+        return aggs;
+    }
+
+    /**
+     * Nests the requested dimensions, outermost first, and hangs the metric's measures off the
+     * innermost bucket — the shape {@code AggregationAdapter} unwraps on the way back.
+     *
+     * <p>Built top-down, one level per recursion, so every level is keyed by its own facet. Building
+     * it bottom-up is what let a child be keyed by its parent's name, which the response side —
+     * resolving the child strictly by name — cannot survive.
+     */
+    JsonObject adaptFacets(MetricMeasuresQuery metric, List<Facet> facets, Integer limit) {
+        var aggs = new JsonObject();
+        if (facets == null || facets.isEmpty()) {
+            return aggs;
+        }
+
+        var facet = facets.getFirst();
+        var isLast = facets.size() == 1;
+        var aggName = AggregationAdapter.adaptName(metric.metric(), facet);
+
+        aggs.put(aggName, isLast ? toTermsLeaf(metric, facet, limit) : toTerms(facet));
+
+        if (isLast) {
+            aggs.getJsonObject(aggName).put("aggs", measuresAdapter.adaptMeasures(metric));
+        } else {
+            aggs.getJsonObject(aggName).put("aggs", adaptFacets(metric, facets.subList(1, facets.size()), limit));
+        }
+
+        return aggs;
+    }
+
+    /**
+     * Only the innermost level carries {@code size} and {@code order}. Both rank buckets by a
+     * measure, and the measures hang off the leaf: Elasticsearch rejects the whole search when an
+     * order path names an aggregation absent from the level that declares it. {@code limit} belongs
+     * here for the same reason — applied at every level it would mean "top N at each depth" rather
+     * than "top N of the ranked dimension".
+     */
+    private JsonObject toTermsLeaf(MetricMeasuresQuery metric, Facet facet, Integer limit) {
+        var terms = new JsonObject().put("field", measuresAdapter.fieldResolver().fromFacet(facet));
+        // A null limit leaves Elasticsearch its default terms size of 10. Fine for the closed
+        // dimensions (operation, connector type); a caller ranking a high-cardinality one — connector
+        // id, application — must ask for the size it wants.
+        if (limit != null) {
+            terms.put("size", limit);
+        }
+        if (metric.sorts() != null && !metric.sorts().isEmpty()) {
+            var order = new JsonObject();
+            for (var sort : metric.sorts()) {
+                order.put(AggregationAdapter.adaptName(metric.metric(), sort.measure()), sort.order().name().toLowerCase());
+            }
+            terms.put("order", order);
+        }
+        return new JsonObject().put("terms", terms);
+    }
+
+    private JsonObject toTerms(Facet facet) {
+        return new JsonObject().put("terms", new JsonObject().put("field", measuresAdapter.fieldResolver().fromFacet(facet)));
+    }
+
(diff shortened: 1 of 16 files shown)


gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/MessageFacetsQueryAdapter.java
```diff
@@ -0,0 +1,123 @@
...
+
+        var leaf = toTerms(metric, innermost, limit);
+        leaf.put("aggs", measuresAdapter.adaptMeasures(metric));
+
+        var current = leaf;
+        for (var facet : nested.reversed()) {
+            var parent = toTerms(metric, facet, limit);
+            parent.put("aggs", new JsonObject().put(AggregationAdapter.adaptName(metric.metric(), facet), current));
+            current = parent;
+        }
```
callaertanthony: **Blocker — nested aggregations are named off by one, so any query with 2+ facets breaks.**

The loop keys each child agg with the *parent's* facet name. For facets `[A, B]` this emits:

```
MESSAGES#A -> terms(A)
  aggs: { MESSAGES#A -> terms(B) { aggs: measures } }
```

`MESSAGES#B` is never emitted, and `MESSAGES#A` appears at two depths.

The response side resolves the child strictly by name and has no null guard on this path — `AggregationAdapter#toFacetBucketResult`:

```java
var nextAggName = adaptName(metric, nextFacet);   // MESSAGES#B
var next = bucket.get(nextAggName);               // null
var buckets = next.get(ES_BUCKETS_PROP);          // NPE
```

This is reachable: `AnalyticsQueryValidator` accepts up to 3 facets on a facets request and 2 on a time series, and `MessageTimeSeriesQueryAdapter#adaptTimeSeries` delegates into the same method.

`HTTPFacetsQueryAdapter#adaptFacets` gets this right by recursing top-down and keying each level by its own facet — building the same way here would remove the divergence rather than mirror it.

The reason the new tests stay green is that `should_nest_dimensions_outermost_first` walks the tree with `.elements().next()`, i.e. positionally, so it never observes an agg name. Asserting on the name would have caught it, and would be worth keeping.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/MessageFacetsQueryAdapter.java
```diff
@@ -0,0 +1,123 @@
...
+        if (limit != null) {
+            terms.put("size", limit);
+        }
+        if (metric.sorts() != null && !metric.sorts().isEmpty()) {
+            var order = new JsonObject();
+            for (var sort : metric.sorts()) {
+                order.put(AggregationAdapter.adaptName(metric.metric(), sort.measure()), sort.order().name().toLowerCase());
+            }
+            terms.put("order", order);
+        }
```
callaertanthony, line 112: **Blocker — `order` (and `size`) applied to non-leaf terms aggregations; Elasticsearch rejects the search.**

`toTerms` is called for every nesting level, so with 2+ facets and a non-empty `sorts` the outer bucket carries `order: {"MESSAGES#COUNT": "desc"}` while its only sub-aggregation is another terms agg. Elasticsearch answers `Invalid aggregation order path [MESSAGES#COUNT]. Unknown aggregation [MESSAGES#COUNT]` and the whole search fails.

The HTTP adapter splits this deliberately: `toTermsLeaf` sets `size` + `order`, and non-leaf levels go through `toTerms(facet)`, which puts only `field`. Same split is needed here.

`limit` has the milder version of the same problem — applying it at every level changes the semantics from "top N of the ranked dimension" to "top N at each level", which is also not what the HTTP path does.

No test covers `sorts`, which is why this passes.

callaertanthony (commented):
Two blockers in `MessageFacetsQueryAdapter`, both only firing with more than one facet. The core fix (dropping `must_not(httpFilter())`) reads right to me, and the reasoning in the description holds up against `ApiTypeFilterTransformer` — comments are inline.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/MessageFacetsQueryAdapter.java
ThibaudAV: Confirmed. Traced it with `facets = [A, B]`: the bottom-up loop keyed the child with the *parent's* name, so `MESSAGES#B` was never emitted and `MESSAGES#A` appeared at two depths, exactly as you described.

Rebuilt top-down, one level per recursion, so every level is keyed by its own facet — same shape as `HTTPFacetsQueryAdapter#adaptFacets` rather than a second way of doing it.

Your point about the test is the most useful part of this review. `.elements().next()` walked the tree positionally, so it passed regardless of the names. Rewritten to address by name, plus the inverse assertion that the parent's name does not reappear underneath it. I checked it actually guards: against the previous adapter the rewritten test fails.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/analytics/engine/adapter/MessageFacetsQueryAdapter.java
ThibaudAV: Confirmed. Took the HTTP path's split rather than mirroring the divergence: `toTermsLeaf` carries `field` + `size` + `order`, and the levels above it go through `toTerms(facet)` with only `field`.

Your note on `limit` is right and has the same cause — applied at every level it means "top N at each depth" instead of "top N of the ranked dimension" — so it moved to the leaf with the rest. Added `should_order_and_size_only_the_innermost_bucket`, which fails against the previous version.

You were also right about the blast radius: `MessageTimeSeriesQueryAdapter:62` delegates into `adaptFacets`, so both paths were fixed by the one change.

On the Aikido comment on `restrictTo` in this file (it seems to have dropped off after the force-push): it is not reachable today — `searchMessageFacets` and `searchMessageTimeSeries` both return `empty(query)` before reaching the adapter when the id set is empty. I hardened it anyway, since an early return that *widens* the query is the wrong default: an empty set now emits `terms` on an empty array, which matches nothing.

callaertanthony (approved):
Both blockers are fixed, and fixed in the right place.

`adaptFacets` is now built top-down, one level per recursion, so every level is keyed by its own facet — structurally the same shape as `HTTPFacetsQueryAdapter#adaptFacets` rather than a second way of expressing it. `toTermsLeaf` / `toTerms` reproduces the same split for `size` and `order`. `MessageTimeSeriesQueryAdapter` inherits both through `facetsAdapter.adaptFacets`.

The test changes matter as much as the code: asserting `/aggs/MESSAGES#MESSAGE_CONNECTOR_TYPE` by name — plus the explicit check that the parent's name does not reappear underneath it — is what the positional walk could never have caught. The two sorted-query cases pin the leaf-only ordering.

Good catch on the side: `restrictTo` emitting `terms: {"request-id": []}` on an empty set rather than skipping the filter is the correct direction, since skipping would widen the query.

I also checked the catalog split after the rebase — `filters:` keeping APPLICATION / PLAN / TENANT / ZONE while `facets:` drops them is coherent, since those resolve against the connection documents in the first phase. Worth stating in the release note that a facets request naming one of them now fails validation with a 400 rather than rendering an empty chart; that is the intended outcome, just a visible one for anything already configured that way.

Two non-blocking things for later, no need to hold this PR:

1. Without the entrypoint predicate, an environment-wide message query resolves request ids for every connection in the window. `searchMessageConnectionRequestIDs` will page up to 10k x 1000 ids into a single `terms` filter, against Elasticsearch's default `index.max_terms_count` of 65,536. `MESSAGE_FILTER_NAMES` still excludes `API` even though message documents carry `api-id` — pushing the API filter down into the second phase would decouple correctness from how large the id set gets.
2. `MessageMeasuresQueryAdapter#adaptMetrics` is still a verbatim copy of the new `adaptMeasures`, and `json()` still inlines its own request-id filter while `MessageRequestIdFilter` is documented as the shared one. Unreachable today — all three callers return early on an empty set — but the two now disagree about what an empty set means.

PR #19359, opened 2026-08-24
feat(authz): route schema documents to the PDP engines

podlesrafal:
> Stacked on #19357 (`authz-33-schema-wire`). Base is that branch, so this diff shows only its own changes. Rebases onto `master` once #19357 merges.

## What

Second step of the gateway half of AUTHZ-33: the engine-port contract and the deployment path for schema documents. Nothing produces an `AuthzSchemaReactorDeployable` yet, so this is inert until the synchronizer lands in the next PR, and it merges independently — same shape as the wire PR.

Section 8 came to ~1400 lines including tests, so it is split in two along the line where the two halves fail differently. This half asks "does the command shape and the routing match the contract the PDP already accepts?". The next asks "does inserting a synchronizer step break the other synchronizers?".

## Changes

- `AuthzEnginePort` gains `addOrUpdateSchema` / `removeSchema`, implemented in `EventBusAuthzEnginePort` and `NoopAuthzEnginePort`.
- `AuthzSchemaReactorDeployable`, `AuthzSchemaMapper`, `AuthzSchemaDeployer`, mirroring their policy counterparts.

## Three things worth a careful look

**The bus command carries `docId` and `schemaText` and nothing else.** The PDP half of AUTHZ-33 is already merged and its consumer reads exactly those two fields. A `name` on the command would be dropped silently, so it stays on the deployable, where the gateway uses it for logging. Section 7 of the spec lists `name` in the port signature; that is right for the event record and wrong for the command.

**D7 gets its own expansion rather than a flag inside the shared one.** `expandWildcard` is shared with policies and entities and unconditionally adds the bootstrap scope. Schema needs the opposite, so `expandWildcardWithoutBootstrap` sits beside it and `routeGated` / `routeForget` now take already-expanded scopes. The alternative, a mode flag threaded through the shared method, puts the policy and entity paths one boolean away from silently losing their bootstrap fan-out.

`*` on a schema therefore reaches every named engine the node hosts and never the bootstrap engine, which is shared across environments. Targeting `default` by name still works and is still cross-environment; D7 removes the implicit fan-out only.

**Distributed replay for schema is deferred and marked.** `DistributedSyncService` has a per-type overload, and reaching parity with authz policy needs a `DistributedEventType`, a distributed mapper and a `DistributedAuthzSchemaSynchronizer`. None of that is in section 8. `AuthzSchemaDeployer.doAfterDeployment` therefore returns `complete()` with a `TODO(AUTHZ-33)` naming what is missing.

Consequence while it stands, stated precisely: a secondary node with distributed sync enabled receives **no schema at all**, not a delayed one. `DefaultSyncManager` runs either the distributed synchronizers or the repository ones and never both, so there is no fallback cycle to pick it up. Policies and entities are unaffected because they have distributed synchronizers. Under D5 an engine with no schema behaves as it does today, so the same request would resolve a bare action name on the primary and not on the secondary, silently.

Distributed sync is opt-in and the default wiring is `NoopDistributedSyncService`, so the default deployment is untouched. That lowers the urgency, not the obligation: this belongs to the definition of done for AUTHZ-33 and lands as its own PR after the synchronizer, where the reviewer's question is "does a secondary end up with the same effective schema as the primary?" rather than "does inserting a step break the other synchronizers?".

## Tests

`mvn verify` on `gravitee-apim-gateway-services-sync`: 708 tests green, prettier and the logging ArchUnit rules included.

New: `EventBusAuthzEnginePortSchemaTest` (12), `AuthzSchemaMapperTest` (9), `AuthzSchemaDeployerTest` (5).

Both load-bearing assertions were checked by mutation rather than assumed:

- pointing the schema ops back at the shared `expandWildcard` (undoing D7) turns `wildcard_reaches_every_hosted_engine_but_never_the_bootstrap_engine` and `wildcard_removal_also_spares_the_bootstrap_engine` red;
- adding `name` to the command turns `the_add_command_carries_only_docId_and_schemaText` red.

The port test also carries two regression cases asserting that a policy wildcard and an entity wildcard still reach the bootstrap engine, since that is what the shared expansion could have lost.

The four hand-written `AuthzEnginePort` implementations in tests get recording queues rather than empty stubs, and `AuthzPolicyDeployerTest` / `AuthzEntityDeployerTest` now assert that neither path emits a schema command.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzSchemaDeployer.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzSchemaDeployer.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.gateway.services.sync.process.common.deployer;
+
+import io.gravitee.gateway.services.sync.process.repository.synchronizer.authz.AuthzEnginePort;
+import io.gravitee.gateway.services.sync.process.repository.synchronizer.authz.AuthzSchemaReactorDeployable;
+import io.reactivex.rxjava3.core.Completable;
+import java.util.Set;
+import lombok.CustomLog;
+import lombok.RequiredArgsConstructor;
+
+@CustomLog
+@RequiredArgsConstructor
+public class AuthzSchemaDeployer implements Deployer<AuthzSchemaReactorDeployable> {
+
+    private final AuthzEnginePort enginePort;
+
+    @Override
+    public Completable deploy(AuthzSchemaReactorDeployable deployable) {
+        Set<String> removed = deployable.removedTargetPdpIds();
+        Completable eviction = removed == null || removed.isEmpty()
+            ? Completable.complete()
+            : enginePort.removeSchema(deployable.environmentId(), deployable.docId(), removed);
+        return eviction
+            .andThen(
+                enginePort.addOrUpdateSchema(
+                    deployable.environmentId(),
+                    deployable.docId(),
+                    deployable.name(),
+                    deployable.schemaText(),
+                    deployable.targetPdpIds(),
+                    deployable.updatedAt()
+                )
+            )
+            .doOnComplete(() -> log.debug("Authz schema '{}' staged for next commit", deployable.docId()))
+            .doOnError(e -> log.warn("Failed to stage authz schema '{}': {}", deployable.docId(), e.getMessage()));
+    }
+
+    // TODO(AUTHZ-32): distributed replay for schema documents, tracked there with the rest of distributed
+    // sync. Until it lands, a secondary node with distributed sync enabled receives no schema at all:
+    // DefaultSyncManager runs either the distributed synchronizers or the repository ones and never both,
+    // so there is no fallback cycle. Opt-in only, the default wiring is NoopDistributedSyncService.
+    // No DistributedSyncService is injected until there is something to distribute.
+    @Override
+    public Completable doAfterDeployment(AuthzSchemaReactorDeployable deployable) {
+        return Completable.complete();
+    }
+
+    @Override
+    public Completable undeploy(AuthzSchemaReactorDeployable deployable) {
+        return enginePort
+            .removeSchema(deployable.environmentId(), deployable.docId(), deployable.targetPdpIds())
+            .doOnComplete(() -> log.debug("Authz schema '{}' staged for removal on next commit", deployable.docId()))
+            .doOnError(e -> log.warn("Failed to stage authz schema '{}' removal: {}", deployable.docId(), e.getMessage()));
+    }
+
+    @Override
+    public Completable doAfterUndeployment(AuthzSchemaReactorDeployable deployable) {
+        return Completable.complete();
+    }
+}
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
@@ -47,6 +47,8 @@ public class EventBusAuthzEnginePort implements AuthzEnginePort {
     static final String OP_REMOVE_ENTITY = "removeEntity";
     static final String OP_ADD_OR_UPDATE_POLICY = "addOrUpdatePolicy";
     static final String OP_REMOVE_POLICY = "removePolicy";
+    static final String OP_ADD_OR_UPDATE_SCHEMA = "addOrUpdateSchema";
+    static final String OP_REMOVE_SCHEMA = "removeSchema";
     static final String OP_COMMIT = "commit";
 
     // Vertx default reply timeout is 30s — long enough to wedge a sync cycle when the PDP service
@@ -106,12 +108,17 @@ public class EventBusAuthzEnginePort implements AuthzEnginePort {
         if (parents != null && !parents.isEmpty()) {
             command.put("parents", new JsonArray(parents));
         }
-        return routeGated(environmentId, command, targetPdpIds, uid, updatedAt);
+        return routeGated(environmentId, command, expandWildcard(environmentId, targetPdpIds), uid, updatedAt);
     }
 
     @Override
     public Completable removeEntity(String environmentId, String uid, Set<String> targetPdpIds) {
-        return routeForget(environmentId, new JsonObject().put("op", OP_REMOVE_ENTITY).put("uid", uid), targetPdpIds, uid);
+        return routeForget(
+            environmentId,
+            new JsonObject().put("op", OP_REMOVE_ENTITY).put("uid", uid),
+            expandWildcard(environmentId, targetPdpIds),
+            uid
+        );
     }
 
     @Override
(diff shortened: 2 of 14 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
```diff
@@ -288,6 +328,22 @@ private Set<String> expandWildcard(String environmentId, Set<String> targetPdpId
         return expanded;
     }
 
+    // D7: "*" on a schema means every NAMED engine of the environment. Unlike expandWildcard it does not
+    // add the bootstrap engine, which is shared across environments — a wildcard schema landing there would
+    // merge two environments' contracts into one. Targeting "default" explicitly still works.
+    private Set<String> expandWildcardWithoutBootstrap(String environmentId, Set<String> targetPdpIds) {
```
sclarkamlab, line 350: D7 means `*` on a schema expands to `hostedFor(env)` only, no `default`. But `AbstractAuthzReactorSynchronizer.deploy` clears `removedTargetPdpIds` whenever the target contains `*`, and its comment leans on a wildcard never being able to narrow a scope away. True for policy and entity since their `expandWildcard` adds `default`, not true here. So retargeting a schema from `{"default"}` to `{"*"}` stages the new targets and never emits `removeSchema` for `default`, leaving it orphaned on the shared bootstrap engine. Nothing produces the deployable yet so it can't bite in this PR, just flagging it so the synchronizer PR does schema aware dropped scopes instead of reusing that branch.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzSchemaDeployer.java
```diff
@@ -0,0 +1,76 @@
...
+                enginePort.addOrUpdateSchema(
+                    deployable.environmentId(),
+                    deployable.docId(),
+                    deployable.name(),
+                    deployable.schemaText(),
+                    deployable.targetPdpIds(),
+                    deployable.updatedAt()
+                )
+            )
+            .doOnComplete(() -> log.debug("Authz schema '{}' staged for next commit", deployable.docId()))
```
sclarkamlab, line 48: On a node hosting no named PDP engine `hostedFor(env)` is empty, so a `*` schema expands to nothing, `routeGated` returns complete and no command goes out. This still logs staged for next commit. A wildcard policy on that node always reaches `default`, so the silent no-op is new to schema. I'd warn when a wildcard expansion comes back empty, otherwise a plain single node install gets policies with no schema and nobody notices.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/test/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzSchemaDeployerTest.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @Test
+    void deploy_evicts_dropped_scopes_before_staging_the_new_targets() {
+        deployer.deploy(schema("doc-2", "n", "entity User;", Set.of("orders"), Set.of("stock"))).blockingAwait();
+
+        assertThat(port.ops).hasSize(2);
+        assertThat(port.ops)
+            .extracting(RecordingPort.SchemaOp::op)
+            .as("eviction must precede staging, otherwise a retarget leaves the schema live on the old engine")
```
sclarkamlab, line 82: This one can't fail. `RecordingPort` adds to `ops` in the method body, so it records when `deploy()` assembles the chain rather than when it subscribes, and Java evaluates the `addOrUpdateSchema` argument before `andThen` runs. I swapped `andThen` for `mergeWith` and it stays green. Record inside a `Completable.fromAction` and the order will actually mean something.

sclarkamlab (commented):
Ran a couple of independent reviews over this. Routing and command shape look right and the tests are good, the mutation notes in the description especially. Three things inline worth a look.

Smaller stuff: `distributedSyncService` on `AuthzSchemaDeployer` is never read, either say why in a comment or leave it out until the distributed path lands. The `schemaOps` queues added to `AuthzEventToEngineIntegrationTest` and `AuthzHydrationPlacementTest` aren't asserted on by anything. And the `AuthzHostedScopes` class javadoc still says `*` expands to `hostedFor` plus `default`, which isn't true for schema now.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzSchemaDeployer.java
podlesrafal: Fixed. `EventBusAuthzEnginePort` now routes both schema operations through a `schemaScopes(...)` helper that warns when the expansion comes back empty while the wildcard was requested:

    Authz schema '{}' targets every named engine of environment [{}], which hosts none, so it reaches nothing

That is the case you describe: a single-node install ends up running policies against no schema, and previously nothing in the log said so, because the "staged for next commit" line is emitted by the deployer regardless of whether any command actually went out.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/test/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzSchemaDeployerTest.java
podlesrafal: Good catch, and it was worse than vacuous: `ops.add(...)` in the method body records at assembly time, so both operations were recorded before anything subscribed.

`RecordingPort` now records inside `Completable.fromAction(...)`, so the order reflects subscription. The ordering assertion is also no longer the only guard: `deploy_does_not_start_staging_until_the_eviction_has_completed` holds the eviction open with a `CompletableSubject` and asserts that staging has not begun yet, because as you note a synchronous `mergeWith` would still produce remove-then-add.

Verified with your mutation: swapping `andThen` for `mergeWith` now fails that test, where before it stayed green.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/EventBusAuthzEnginePort.java
podlesrafal: This one was already handled in #19360, which is what you asked for: `AuthzSchemaSynchronizer` overrides `scopesNotCoveredByWildcard()` so the schema route computes its own dropped scopes instead of reusing the branch that clears them. The override now compares on the scope base, so it also covers a tagged bootstrap engine.

Leaving the thread here rather than changing anything in this PR, since as you say nothing produces the deployable yet at this point in the stack.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzSchemaDeployer.java
```diff
@@ -0,0 +1,76 @@
...
+            .doOnError(e -> log.warn("Failed to stage authz schema '{}': {}", deployable.docId(), e.getMessage()));
```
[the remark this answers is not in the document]
podlesrafal: @AikidoSec ignore: this is the established pattern for every deployer in this package. AuthzPolicyDeployer and AuthzEntityDeployer log the same way in doOnError and propagate, so changing only the schema deployer would make it the odd one out. If the pattern is wrong it should be changed across all three in its own PR rather than diverging here.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzSchemaDeployer.java
```diff
@@ -0,0 +1,76 @@
...
+    public Completable doAfterDeployment(AuthzSchemaReactorDeployable deployable) {
+        return Completable.complete();
+    }
+
+    @Override
+    public Completable undeploy(AuthzSchemaReactorDeployable deployable) {
+        return enginePort
+            .removeSchema(deployable.environmentId(), deployable.docId(), deployable.targetPdpIds())
+            .doOnComplete(() -> log.debug("Authz schema '{}' staged for removal on next commit", deployable.docId()))
+            .doOnError(e -> log.warn("Failed to stage authz schema '{}' removal: {}", deployable.docId(), e.getMessage()));
```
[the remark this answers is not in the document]
podlesrafal: @AikidoSec ignore: this is the established pattern for every deployer in this package. AuthzPolicyDeployer and AuthzEntityDeployer log the same way in doOnError and propagate, so changing only the schema deployer would make it the odd one out. If the pattern is wrong it should be changed across all three in its own PR rather than diverging here.

sclarkamlab (approved):

PR #19360, opened 2026-08-24
feat(authz): synchronize schema documents and hydrate them first

podlesrafal:
> Stacked on #19359 (`authz-33-schema-port`), which is itself stacked on #19357. Base is that branch, so this diff shows only its own changes. Rebases down as the stack merges.

## What

Third and last step of the gateway half of AUTHZ-33: the schema route joins the sync cycle. With this a schema authored in the control plane reaches the PDP engines it targets, without a restart.

Section 8 came to ~1400 lines including tests and was split in two. #19359 answered "does the command shape and the routing match the contract the PDP already accepts?". This one answers "does inserting a synchronizer step break the other synchronizers?".

## Changes

- `AuthzSchemaSynchronizer extends AbstractAuthzReactorSynchronizer`, with mapper, deployable and deployer already in #19359.
- `Order.AUTHZ_SCHEMA` at 9, between `AUTHZ_PDP` and `AUTHZ_ENTITY`. Everything below renumbers. The enum's static duplicate-index guard turns a mistake into a class-initialisation failure rather than a silent reordering, and `OrderTest` exercises it.
- Hydration groups three aggregates instead of two and stages the schema **first**, so a freshly provisioned scope carries it on its very first commit.
- Spring wiring, including a schema-specific `AuthzScopePlacement` rather than sharing the policy one.

## The D7 eviction defect, found in review of #19359

`AbstractAuthzReactorSynchronizer.deploy` dropped the whole eviction set on a wildcard:

```java
if (target.contains(WILDCARD)) {
    // A wildcard targets every hosted scope, so it can never narrow a previously-applied scope away
    dropped.clear();
}
```

That premise is true for policies and entities and **false for schemas**, by construction of D7: a schema wildcard deliberately excludes the bootstrap engine. So a schema retargeted from `["default"]` to `["*"]` was never evicted from the bootstrap engine and `addOrUpdateSchema` deliberately skipped it, leaving it on a cross-environment engine forever. `placement.replace` then recorded `{"*"}`, so no later narrowing could evict it either: there was no state from which this recovers.

The fix replaces the unconditional `clear()` with `retainAll(scopesNotCoveredByWildcard())`. For policies and entities that hook returns the empty set, so `retainAll` is **exactly** the `clear()` it replaces, by the definition of the operation rather than by assurance. An explicitly named scope is then removed from the drop set unconditionally, so targeting both `"*"` and `"default"` still keeps the document on the bootstrap engine, which is what D7 intends: it removes the implicit fan-out, not explicit targeting.

Two alternatives were considered and rejected. Evicting `default` unconditionally in the deployer runs on every deploy rather than on the transition, so it would send a removal to the shared engine every cycle for documents that were never there. Keeping `default` in the drop set whenever the target contains `"*"` evicts a document that named `default` explicitly.

## Two behaviour changes the tests now assert rather than absorb

**A third staging step means a third `commitScope` per hydrated scope.** `stageScope` seals the scope per synchronizer, so the existing `times(2)` expectations become `times(3)`. Worth a look: three round-trips per scope where one would do is a cost on a path already known to matter for API deployment latency. Not changed here because it means moving the commit out of the shared `stageScope`.

**A dead engine now fails a hydration cycle earlier.** With the schema staged first, `failed_hydration_is_retried_on_the_next_cycle` sees the failure at the schema step, before any policy work: 1 `addOrUpdatePolicy` instead of 2, and 4 `commitScope` instead of 3. That is the ordering doing its job, and the comment in that test now says so.

## Tests

`mvn verify` on `gravitee-apim-gateway-services-sync`: 714 green, prettier and the logging ArchUnit rules included. 539 insertions, 34 deletions.

New `AuthzWildcardEvictionTest`, three cases:
- a schema retargeted from `default` to `*` **is** evicted from the bootstrap engine;
- a schema targeted at both `*` and `default` **stays** on it;
- a policy retargeted from `default` to `*` is still not evicted, pinning the `retainAll(Set.of()) == clear()` equivalence for the shared path.

Reverting the hook to the old `clear()` turns the first of those red; checked by mutation, not assumed.

`AuthzPdpHydrationTest.hydration_stages_the_schema_before_the_policies_of_the_same_scope` pins the section 8 ordering requirement with an `InOrder` on the engine port.

`AuthzEventToEngineIntegrationTest` and `AuthzHydrationPlacementTest` get an assertion proving their schema recorder is live, so their emptiness assertions are not vacuous. Note the substantive hydration-ordering assertion lives in `AuthzPdpHydrationTest` rather than in those two, because it needs a mockable engine port to express ordering.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzSchemaDeployer.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzSchemaDeployer.java
@@ -22,6 +22,18 @@ import java.util.Set;
 import lombok.CustomLog;
 import lombok.RequiredArgsConstructor;
 
+/**
+ * TODO(AUTHZ-32): schema documents are not replayed over distributed sync, unlike policies and entities,
+ * which have DistributedAuthzPolicySynchronizer and DistributedAuthzEntitySynchronizer. The consequence is
+ * specific and worth stating precisely: the bootstrap engine IS present on every node, and
+ * AuthzHostedScopes#serves admits the bare default scope unconditionally, so a secondary node running
+ * distributed sync does receive policies and entities there while never receiving any schema. Evaluation
+ * outcomes do not diverge, because the PDP builds its Authorizer without a schema by design and therefore
+ * never validates against one; the schema's only runtime consumer is SearchExecutor. So what diverges is
+ * action search: the same query answers from the schema on the primary and from nothing on a secondary.
+ * Distributed sync is opt-in and the default wiring is NoopDistributedSyncService, so this is inert unless
+ * it is switched on.
+ */
 @CustomLog
 @RequiredArgsConstructor
 public class AuthzSchemaDeployer implements Deployer<AuthzSchemaReactorDeployable> {
@@ -49,26 +61,11 @@ public class AuthzSchemaDeployer implements Deployer<AuthzSchemaReactorDeployabl
             .doOnError(e -> log.warn("Failed to stage authz schema '{}': {}", deployable.docId(), e.getMessage()));
     }
 
-    // TODO(AUTHZ-32): distributed replay for schema documents, tracked there with the rest of distributed
-    // sync. Until it lands, a secondary node with distributed sync enabled receives no schema at all:
-    // DefaultSyncManager runs either the distributed synchronizers or the repository ones and never both,
-    // so there is no fallback cycle. Opt-in only, the default wiring is NoopDistributedSyncService.
-    // No DistributedSyncService is injected until there is something to distribute.
-    @Override
-    public Completable doAfterDeployment(AuthzSchemaReactorDeployable deployable) {
-        return Completable.complete();
-    }
-
     @Override
     public Completable undeploy(AuthzSchemaReactorDeployable deployable) {
         return enginePort
             .removeSchema(deployable.environmentId(), deployable.docId(), deployable.targetPdpIds())
             .doOnComplete(() -> log.debug("Authz schema '{}' staged for removal on next commit", deployable.docId()))
             .doOnError(e -> log.warn("Failed to stage authz schema '{}' removal: {}", deployable.docId(), e.getMessage()));
-    }
-
-    @Override
-    public Completable doAfterUndeployment(AuthzSchemaReactorDeployable deployable) {
-        return Completable.complete();
     }
 }
--- a/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
@@ -61,6 +61,7 @@ public class AuthzPdpSynchronizer implements RepositorySynchronizer {
 
     private final LatestEventFetcher eventsFetcher;
     private final AuthzPdpMapper mapper;
+    private final AuthzSchemaSynchronizer schemaSynchronizer;
     private final AuthzPolicySynchronizer policySynchronizer;
     private final AuthzEntitySynchronizer entitySynchronizer;
     private final Node node;
@@ -114,6 +115,7 @@ public class AuthzPdpSynchronizer implements RepositorySynchronizer {
     public AuthzPdpSynchronizer(
         LatestEventFetcher eventsFetcher,
         AuthzPdpMapper mapper,
+        AuthzSchemaSynchronizer schemaSynchronizer,
         AuthzPolicySynchronizer policySynchronizer,
         AuthzEntitySynchronizer entitySynchronizer,
         Node node,
@@ -126,6 +128,7 @@ public class AuthzPdpSynchronizer implements RepositorySynchronizer {
     ) {
         this.eventsFetcher = eventsFetcher;
         this.mapper = mapper;
+        this.schemaSynchronizer = Objects.requireNonNull(schemaSynchronizer, "schemaSynchronizer must not be null");
         this.policySynchronizer = Objects.requireNonNull(policySynchronizer, "policySynchronizer must not be null");
         this.entitySynchronizer = Objects.requireNonNull(entitySynchronizer, "entitySynchronizer must not be null");
         this.node = Objects.requireNonNull(node, "node must not be null");
@@ -447,6 +450,7 @@ public class AuthzPdpSynchronizer implements RepositorySynchronizer {
         Set<String> scopes = byScope.keySet();
         // One scan+map of policies and one of entities for the whole environment, grouped by scope.
         return Single.zip(
+            schemaSynchronizer.groupByScope(environmentId, scopes),
             policySynchronizer.groupByScope(environmentId, scopes),
             entitySynchronizer.groupByScope(environmentId, scopes),
             BackfillGroups::new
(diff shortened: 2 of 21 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzSchemaSynchronizer.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * D7: a schema wildcard expands to the named engines only, so unlike a policy or an entity it does
+     * NOT cover the bootstrap engine. Retargeting a schema from "default" to "*" must therefore still
+     * evict it from the bootstrap engine, which is shared across environments.
+     */
+    @Override
+    protected Set<String> scopesNotCoveredByWildcard() {
+        return Set.of(EventBusAuthzEnginePort.DEFAULT_SCOPE);
```
sclarkamlab, line 59: Exact string match, but a routing scope can be tagged. `addressFor` strips the tag so `default@us` resolves to the same bootstrap address, and if a PDP ever gets provisioned as `default` with a tag then hydration records `default@us` in the placement, the `retainAll` in `AbstractAuthzReactorSynchronizer` drops it and the schema never leaves the bootstrap engine. Comparing on the base would cover it:

```java
dropped.removeIf(s -> !scopesNotCoveredByWildcard().contains(EventBusAuthzEnginePort.baseScope(s)));
```

`expandWildcardWithoutBootstrap` has the same gap on the way in, though that one is already on the base branch. Can a PDP actually be called `default`? If not this is theoretical.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
```diff
@@ -466,10 +470,13 @@ private Completable hydrateEnv(List<AuthzPdpProvisionDeployable> provisions) {
...
-        // Per scope: policies then entities (policy failure short-circuits the scope), so a single scope's
-        // engine failure re-pends only that scope, not the whole environment.
-        return policySynchronizer
-            .stageScope(environmentId, scope, groups.policies().getOrDefault(scope, List.of()))
+        // Per scope: schema, then policies, then entities (a failure short-circuits the scope), so a single
+        // scope's engine failure re-pends only that scope, not the whole environment. Schema goes first so
+        // the freshly provisioned scope's very first commit already carries it; staged later it would take
+        // effect a full cycle after the policies that depend on it.
+        return schemaSynchronizer
+            .stageScope(environmentId, scope, groups.schemas().getOrDefault(scope, List.of()))
```
sclarkamlab: `stageScope` commits even when the doc list is empty, so hydrating a scope is now three `commitScope` round trips to the same engine address instead of two. A gateway with no schemas at all pays the third one every time. Worth skipping the commit when `docs` is empty?

sclarkamlab (commented):
Looks good, the D7 eviction fix and staging the schema first both read right to me. Two small things inline.

One on tests: there's no `AuthzSchemaSynchronizerTest`. `AuthzPolicySynchronizerTest` and `AuthzEntitySynchronizerTest` both cover `order()`, publish and unpublish in one batch, unparseable payload dropped, one commit per cycle and placement eviction on retarget, and the schema one only gets hit indirectly through `AuthzWildcardEvictionTest` and the hydration ordering test. Not blocking.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzSchemaSynchronizer.java
podlesrafal: Fixed, and the case is real rather than theoretical: `CreateAuthzPdpUseCase` normalises a blank `targetPdpId` to `default`, so a PDP created without an explicit id and with a tag ends up on the routing scope `default@us`.

The comparison now runs on the scope base, in `AbstractAuthzReactorSynchronizer`, so the carve-out covers both the bare and the tagged bootstrap engine:

    dropped.removeIf(scope -> !scopesNotCoveredByWildcard().contains(EventBusAuthzEnginePort.baseScope(scope)));

Worth adding that the control plane had exactly the same defect and it was already fixed there during review: `SchemaDeploymentDomainService` now compares `AuthzTargetResolver.baseScope(scope)` instead of the full string. The two sides disagreeing on this was the dangerous part, since it would have been silent at runtime.

On `expandWildcardWithoutBootstrap` having the same gap on the way in: it does not, and by construction. A tag-scoped default needs no provisioning, so it never enters `hostedByBase`, and `hostedFor` therefore cannot return `default@us`. The gap only existed on the eviction side.

There was no test covering any of this. I checked by reverting the expression to the exact-string comparison and the whole suite stayed green, so `AuthzWildcardEvictionTest` now has `a_schema_retargeted_from_a_tagged_bootstrap_engine_to_wildcard_is_still_evicted`, which fails against the old expression.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzPdpSynchronizer.java
podlesrafal: Changed, though not in the way the suggestion proposed, because skipping the commit when `docs` is empty is not safe.

`commitScope` on an empty scope is deliberate: a freshly provisioned engine has to leave commit generation 0, and `== 0L` is what `AuthzPdpEvalBusConsumer`, `AuthzPdpSearchBusConsumer` and `AuthzPdpEngine.search` use as their readiness gate. Suppressing that first commit would make every evaluation on a new engine return NOT_READY.

`stageScope` now takes a `commit` flag instead, and hydration stages schema, then policies, then entities into the same scope and seals it once at the end. Same saving, and a scope with no documents at all still gets exactly one commit.

One consequence worth stating rather than hiding: the first commit used to double as an early failure detector, so a dead engine was discovered at the first aggregate boundary. With a single commit per scope the staging calls all run first. `AuthzPdpSynchronizerGapsTest.failed_hydration_is_retried_on_the_next_cycle` now pins that behaviour and its comment says so.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/AuthzSchemaDeployer.java
```diff
@@ -51,11 +51,11 @@ public Completable deploy(AuthzSchemaReactorDeployable deployable) {
             .doOnError(e -> log.warn("Failed to stage authz schema '{}': {}", deployable.docId(), e.getMessage()));
     }
 
-    // TODO(AUTHZ-33): distributed replay for schema documents. Needs a DistributedEventType, a
-    // distributed mapper and a DistributedAuthzSchemaSynchronizer. Until then a secondary node with
-    // distributed sync enabled never receives schema at all: DefaultSyncManager runs either the
-    // distributed synchronizers or the repository ones, never both, so there is no fallback cycle.
-    // Opt-in only, the default wiring is NoopDistributedSyncService.
+    // TODO(AUTHZ-32): distributed replay for schema documents, tracked with the rest of the authz
```
podlesrafal: Addressed. The marker now points at AUTHZ-32, which is the ticket that owns distributed sync for authz, so the deferred work is tracked rather than open-ended. The comment also states the observable consequence while it is outstanding: a secondary node with distributed sync enabled receives no schema at all, because DefaultSyncManager runs either the distributed synchronizers or the repository ones and never both.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/repository/synchronizer/authz/AuthzSchemaMapper.java
```diff
@@ -41,6 +41,12 @@ public Maybe<AuthzSchemaReactorDeployable> toDeploy(Event event) {
                     log.warn("Skipping authz schema DEPLOY event [{}] — missing or blank schemaText", event.getId());
                     return null;
                 }
+                // Without it AuthzHostedScopes#serves looks up "null:<base>", matches nothing, and the
+                // document is dropped with no trace. AuthzPdpMapper guards the same field.
+                if (wire.getEnvironmentId() == null || wire.getEnvironmentId().isBlank()) {
```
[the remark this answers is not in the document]
podlesrafal: @AikidoSec ignore: none of the sibling mappers extract their field guards — AuthzPolicyMapper inlines five of them, AuthzEntityMapper and AuthzPdpMapper two each — so pulling a helper out here alone would make this the only mapper in the package with a different shape. The two occurrences also differ in their warning text (DEPLOY vs UNDEPLOY), which is deliberate: the log line names the path that dropped the event. If this validation is to be centralised it should be done across all four mappers in its own change.

sclarkamlab (approved):

PR #19361, opened 2026-08-24
fix(helm): expose es.lifecycle.policies.eventMetrics for the gateway

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-14875

## Description

The gateway reads `reporters.elasticsearch.lifecycle.policies.event_metrics`, but the chart rendered only `monitor` / `request` / `health` / `log` under `es.lifecycle`, so the key could only be set by hand through `gateway.env`. Adds `es.lifecycle.policies.eventMetrics` to `values.yaml` and the gateway configmap, with CHANGELOG and artifacthub entries.

Defaults to `""`, unlike the four siblings which default to the `my_policy` placeholder. The siblings have an installed base; this key has none. Defaulting it empty means no existing `es.lifecycle.enabled` deployment gets a placeholder policy attached to its event-metrics stream on upgrade — it renders as `event_metrics:` (null), which the gateway ignores.

## Compatibility

The gateway honours this key on master (#19421), so unlike when this PR was opened, the chart key is live rather than inert: a policy set here takes effect on deploy, including its delete phase. Nothing renders at all unless `es.lifecycle.enabled` is set and the key is given a value.

## Tests

`helm/tests/gateway/configmap_es_reporter_test.yaml` sets the new key and asserts it renders under `es.lifecycle`. Full chart suite green.

## Related

Reporter: #19350 pins the gateway-side behaviour with test coverage. The two are independent — the earlier merge-order note no longer applies.

### DIFF
--- a/helm/CHANGELOG.md
+++ b/helm/CHANGELOG.md
@@ -9,6 +9,7 @@ This file documents all notable changes to [Gravitee.io API Management 3.x](http
 - Add `gateway.kafka.routingHostMode.virtualClusterBrokerDomainPattern`, a broker domain pattern applied to Kafka APIs backed by a Virtual Cluster only. Such APIs advertise rewritten broker ids, and setting their pattern separately keeps the domains of the other Kafka APIs on the gateway unchanged. Optional: when unset, Virtual Clusters follow `brokerDomainPattern`. Two placeholders come with it, usable in either pattern: `{realBrokerId}` (the id configured on the backend broker) and `{clusterIndex}` (the backend's position in the Virtual Cluster).
 - Document `gateway.services.metrics.kafka.durations.principalName` to add the `principal_name` tag to Kafka duration metrics (disabled by default).
 - fix gateway requestTimeout ignored when gateway.servers is configured (APIM-14276)
+- Add `es.lifecycle.policies.eventMetrics`, the ILM policy for the `gravitee-event-metrics-*` data streams. The gateway already read this key, but the chart did not render it under `es.lifecycle`, so it could only be set by hand through `gateway.env` (APIM-14875).
 
 ### 4.12.0
 - Add support for Kubernetes Gateway API HTTPRoute for all components (API, Gateway, User Interface, Portal).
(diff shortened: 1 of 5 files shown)


callaertanthony (approved):
Approving — the change is correct and minimal. Reviewed at `2d6c1e5`.

I checked the claims in the description rather than taking them:

- The gateway does read the key: `ReporterConfiguration.java:236` binds `reporters.elasticsearch.lifecycle.policies.event_metrics`, defaulting to null.
- The empty default really is inert. All four flavors guard the property with `?has_content` (`freemarker/*/mapping/index-template-event-metrics.ftl:7`), which is false for null *and* for the empty string — so the `event_metrics:` line an upgraded deployment now gets cannot attach an empty policy name to the data stream template.
- `es.lifecycle` is rendered only by the gateway configmap, so no other component needed the same key.
- The CHANGELOG entry and the Chart.yaml annotation both land under 4.13.0, matching `version:`.

Two things, neither blocking:

**1. The default path has no test.** The only lifecycle case, "Check override es lifecycle settings", sets all five policies explicitly. Nothing covers `es.lifecycle.enabled: true` with `eventMetrics` left at `""` — which is what every existing deployment gets on upgrade, and the only behaviour change here that reaches the installed base.

The risk isn't today, it's the next person: defaulting `eventMetrics: my_policy` "for consistency with its four siblings" is an easy and plausible follow-up, the suite stays green, and every upgraded deployment silently points its event-metrics data streams at an ILM policy that doesn't exist. A case asserting the line renders with no value when the key is untouched pins the decision the description argues for.

**2. `helm/values.yaml:375` — the comment argues rather than documents.** The "unset by default, unlike its siblings: no deployment can be relying on it yet…" half is rationale for the choice, and it already lives in the commit message and the CHANGELOG. The four siblings are a single clause; suggest matching them:

```yaml
eventMetrics: "" ## ILM policy for the gravitee-event-metrics-* data streams (unset by default)
```

Unrelated, for whenever someone is next in this file: `ReporterConfiguration.java:239` also binds `lifecycle.policies.authz_decisions`, and there is no `authz` key anywhere under `helm/` — the identical gap, one line further down. Out of scope here.

paulatulis:
Thanks — both applied, and the branch is rebased onto master (now `436535b`; `Chart.yaml` conflicted with the HTTP/2 flow-control annotation that landed in the same `- kind: added` slot, both entries are kept).

1. Added a case for `es.lifecycle.enabled: true` with `eventMetrics` left untouched, asserting the key renders with no value. Negative control for the scenario you described: defaulting it to `my_policy` in `values.yaml` makes exactly that case fail.
2. `values.yaml` now matches its four siblings — the rationale stays in the commit message and the CHANGELOG.

The `lifecycle.policies.authz_decisions` gap under `helm/` is real; leaving it out of this PR as you suggested.

PR #19363, opened 2026-08-25
feat(gamma): improved create env alert filters, conditions, checks

arpzilla-dev:
## Issue

https://gravitee.atlassian.net/browse/APIM-14912

## Description

Align Gamma environment-alert create with Classic: rule catalog (metrics, types, filters, aggregations, projections), Value vs Pattern loaders, hidden STRING payloads on info-only node rules, Create gating, and hide Node rules when cloudHosted.enabled is true. List templates now show a [Template] prefix. CRUD, Notifications, and History were already in place; this PR is Classic create-form parity on the platform module only.

## Additional context

https://github.com/user-attachments/assets/5016e99f-ff64-4adb-b61a-0e9ffcb561b9


aqibmohammadkhan (commented):
Issues:

1. Value field dead-end for Application/Plan metrics — the Select renders with zero options and no text fallback, so filters on these metrics can never be configured.
2. event_rules silently restored — unchecking "auto-create on every new API" (or "template") on an existing alert doesn't actually remove the association; the save silently re-injects the old value.
3. Empty-state screen leaks hidden Node rules to cloud-hosted customers — it lists the raw, unfiltered catalog instead of applying the same cloudHosted filter used in the real create form (directly undermines this PR's stated goal).
4. Stale pattern reused as a Value option when switching Operator between MATCHES and EQUALS/NOT_EQUALS — can silently save a condition that never matches anything.
5. Description field silently overwritten when the user reselects Rule after already typing custom text — a new regression, not present pre-PR.
6. COUNT aggregation doesn't clear property — the hidden field's stale value still gets sent to the backend.

Plausible:
7. Tenant Value option uses tenant.id where the rest of the app treats tenant.key as the real matching identifier — likely wrong but not confirmable from the frontend repo alone.

Design/latent-risk (altitude):
8. AlertsTab's rules/ruleCategories default to the unfiltered catalog — today's only caller does it right, but nothing stops a future one from reintroducing Node rules for cloud-hosted envs, and tests don't guard it.
9. getMetricsForRuleId vs. getFilterMetricsForRuleId disagree on Node-rule metrics — currently masked, a latent trap for future changes.
10. "Info-only rule" IDs are hand-enumerated in three separate files instead of one catalog flag — easy to drift out of sync when a new rule is added.

arpzilla-dev:
> Issues:
> 
> 1. Value field dead-end for Application/Plan metrics — the Select renders with zero options and no text fallback, so filters on these metrics can never be configured.
> 2. event_rules silently restored — unchecking "auto-create on every new API" (or "template") on an existing alert doesn't actually remove the association; the save silently re-injects the old value.
> 3. Empty-state screen leaks hidden Node rules to cloud-hosted customers — it lists the raw, unfiltered catalog instead of applying the same cloudHosted filter used in the real create form (directly undermines this PR's stated goal).
> 4. Stale pattern reused as a Value option when switching Operator between MATCHES and EQUALS/NOT_EQUALS — can silently save a condition that never matches anything.
> 5. Description field silently overwritten when the user reselects Rule after already typing custom text — a new regression, not present pre-PR.
> 6. COUNT aggregation doesn't clear property — the hidden field's stale value still gets sent to the backend.
> 
> Plausible: 7. Tenant Value option uses tenant.id where the rest of the app treats tenant.key as the real matching identifier — likely wrong but not confirmable from the frontend repo alone.
> 
> Design/latent-risk (altitude): 8. AlertsTab's rules/ruleCategories default to the unfiltered catalog — today's only caller does it right, but nothing stops a future one from reintroducing Node rules for cloud-hosted envs, and tests don't guard it. 9. getMetricsForRuleId vs. getFilterMetricsForRuleId disagree on Node-rule metrics — currently masked, a latent trap for future changes. 10. "Info-only rule" IDs are hand-enumerated in three separate files instead of one catalog flag — easy to drift out of sync when a new rule is added.

Thanks for the review. Addressing in order:

1. Application / Plan Value — Same as Classic on environment alerts: those loaders return an empty list and EQUALS still shows Value. MATCHES / CONTAINS / Pattern still work. Not a Gamma regression; a Pattern fallback when the list is empty would be a follow-up.

2. event_rules restore — Agreed. Save was sending undefined when auto-create was off, and the update path re-injected the old API_CREATE rule. Fixed: we now send [] when it is unchecked.

3. Empty-state Node list — Educational copy only. Cloud-hosted users still cannot pick Node rules on create. Cheap to filter later; it does not leak create.

4. Stale pattern as a Value option — Real UX when switching MATCHES → EQUALS. The inject-current-pattern path is there so existing unknown values still display. Clearing the pattern when it is not in the catalog is a follow-up.

5. Description overwrite — Intentional: selecting a rule seeds the Classic description. Protecting a dirty custom description is nicer; not treating it as a blocker.

6. COUNT still sending property — Metric is hidden; Classic also kept a property on the model. Harmless unless the API rejects it. Omitting it is a one-liner follow-up.

7. Tenant id vs key — Classic’s loader is new Tuple(tenant.id, tenant.name). Gamma matches that. Tenant CRUD uses key; alert matching uses id. Leaving as-is unless the engine proves otherwise.

8–10. Fair design notes, not current bugs. AlertsTab’s only caller already passes the filtered catalog; Node metric/filter split is masked by info-only rules; info-only IDs could be a catalog flag later.

Also fixed the two copy mismatches (range High < Low, RELAXED_COUNT “higher” vs >=).

aqibmohammadkhan (commented):
1. Value field is a dead end for Application/Plan metrics — the dropdown renders with zero options and no text input, so a filter/condition on these two metrics can never be completed. This looks like a functional blocker for a documented use case, not an edge case.
2. Stale pattern reused as Value when switching Operator between MATCHES ↔ EQUALS/NOT_EQUALS — can silently save a nonsense condition.
3. Description silently overwritten if the user reselects Rule after typing custom text — new regression, no confirmation shown.
4. COUNT aggregation leaks stale property to the backend even though the UI hides the field.
5. Empty-state screen advertises Node rules on cloud-hosted environments that the create form itself hides — directly undercuts this PR's stated purpose.

Plus one plausible (tenant id vs key mismatch, needs backend confirmation) and two design-risk items (default-prop filtering gap, triplicated info-only rule list) that are lower urgency but worth a comment.

arpzilla-dev:
Thanks for the thorough review @aqibmohammadkhan! Went through each point:

**Fixed:**
- **#2 Stale pattern reused as Value (MATCHES ↔ EQUALS/NOT_EQUALS):** switching the operator now clears `pattern` when it isn't a valid option in the Value catalog, so a leftover regex can no longer be silently saved as if it were an enumerated value (`sanitizePatternForOperator` in `alertMetricValues.ts`, wired into both `SimpleConditionForm` and `FilterRow`).
- **#5 Empty-state lists Node rules on cloud-hosted:** `AlertsEducationalEmptyState` now filters through the same `getAlertRulesForEnvironment(cloudHostedEnabled)` helper the create form already uses, so it no longer advertises rules the form itself hides.

**Not changing in this PR, with reasoning:**

- `#1` Application/Plan Value is empty:  this matches Classic's own behavior exactly — both metrics use an async/empty loader there too, so an empty Value list for EQUALS isn't a Gamma regression. Worth a follow-up to wire a real lookup, but not something this PR broke or owes.
- **#3 Description overwritten on Rule reselect:** also mirrors Classic's seed-on-select behavior. Debatable UX, but not a regression introduced here.
- **#4 COUNT aggregation leaking stale `property`:** cosmetic payload noise carried over from Classic; no evidence the backend rejects it. Happy to remove it if that changes.
- **Tenant `id` vs `key`:** Gamma uses `tenant.id` here because Classic's loader does the same. Other console screens use `tenant.key`, so this *might* be the wrong identifier — but changing it without knowing which one the alert engine actually matches on could break tenant filters. Leaving it as Classic until that's confirmed.
- **Design-risk items (default unfiltered catalog, metric/filter split, triplicated info-only list):** good future-hardening notes, but they're not bugs in the current behavior, so leaving them as follow-up rather than scope creep here.

aqibmohammadkhan (approved):

PR #19380, opened 2026-08-25
feat(console): import a portal navigation tree from a remote repository

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-92

## Description

Import a portal navigation subtree from a remote repository (GitHub, GitLab) in one action.

**Backend** — `POST /v2/portal-navigation-items/_import` (`ENVIRONMENT_DOCUMENTATION[UPDATE]`) walks a files fetcher and creates the navigation tree:
- a `.gravitee.json` manifest drives the hierarchy when present, otherwise the remote tree is mirrored; unsupported file types are ignored and `.adoc` is rejected, as the NG portal has no AsciiDoc content type.
- the source lands on the created root folder only, marked `subtreeImport` — server-owned, read-only in the REST schema and kept coherent by `FileListingSourceOnFolderRule`. Children carry no source and are read-only; re-fetching the folder re-runs the import, updating contents in place and pruning orphans.
- failures are visible and never destructive: a listing or manifest failure is stamped on the source and leaves the existing subtree untouched; per-file failures are reported without aborting the run.

**Console** — an Import button in *Portal > Navigation* opens an "Import documentation from a repository" dialog (folder title + source editor limited to the directory-capable fetchers). On import, a success or per-file failure snackbar, then navigation to the created folder. The imported folder shows the read-only source banner with *Fetch now*; its children show "Managed by parent folder". Fetchable containers are read from the server-owned `subtreeImport` marker, so the set matches exactly what `_fetch` accepts.

## Additional context

This change is security-sensitive: it adds an authenticated endpoint that fetches content from an external repository. Three points worth a reviewer's attention — the endpoint is gated by `ENVIRONMENT_DOCUMENTATION[UPDATE]`, the fetcher configuration may hold credentials and is never included in error messages, and a failed fetch leaves the already-imported pages untouched instead of pruning them.

https://github.com/user-attachments/assets/86ed12a5-63c4-4417-8080-ae7dc3c6221a

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/PortalNavigationItemsMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/PortalNavigationItemsMapper.java
@@ -17,8 +17,10 @@ package io.gravitee.rest.api.management.v2.rest.mapper;
 
 import io.gravitee.apim.core.exception.TechnicalDomainException;
 import io.gravitee.apim.core.portal_category.model.PortalCategoryId;
+import io.gravitee.apim.core.portal_page.domain_service.PortalNavigationBulkImportDomainService;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
 import io.gravitee.apim.core.portal_page.use_case.FetchPortalNavigationItemUseCase;
+import io.gravitee.apim.core.portal_page.use_case.ImportPortalNavigationUseCase;
 import io.gravitee.apim.core.portal_page.use_case.SeedDefaultPagesForPortalNavigationItemsUseCase;
 import io.gravitee.definition.model.VirtualHost;
 import io.gravitee.definition.model.v4.listener.ListenerType;
@@ -33,6 +35,8 @@ import io.gravitee.rest.api.management.v2.rest.model.CreatePortalNavigationFolde
 import io.gravitee.rest.api.management.v2.rest.model.CreatePortalNavigationLink;
 import io.gravitee.rest.api.management.v2.rest.model.CreatePortalNavigationPage;
 import io.gravitee.rest.api.management.v2.rest.model.FetchPortalNavigationItemResponse;
+import io.gravitee.rest.api.management.v2.rest.model.ImportPortalNavigationRequest;
+import io.gravitee.rest.api.management.v2.rest.model.ImportPortalNavigationResponse;
 import io.gravitee.rest.api.management.v2.rest.model.PortalNavigationItem;
 import io.gravitee.rest.api.management.v2.rest.model.PortalNavigationItemApiSummary;
 import io.gravitee.rest.api.management.v2.rest.model.PortalNavigationItemFetchResult;
@@ -174,6 +178,34 @@ public interface PortalNavigationItemsMapper {
 
     FetchPortalNavigationItemResponse map(FetchPortalNavigationItemUseCase.Output output);
 
+    PortalNavigationItemFetchResult map(PortalNavigationBulkImportDomainService.BulkImportResult.FileImportResult result);
+
+    default ImportPortalNavigationResponse map(ImportPortalNavigationUseCase.Output output) {
+        var results = output.result().files().stream().map(this::map).toList();
+        var succeeded = (int) output
+            .result()
+            .files()
+            .stream()
+            .filter(file -> file.success())
+            .count();
+        return new ImportPortalNavigationResponse()
+            .rootFolder(map((io.gravitee.apim.core.portal_page.model.PortalNavigationItem) output.rootFolder()))
+            .summary(new PortalNavigationItemsFetchSummary().succeeded(succeeded).failed(results.size() - succeeded).results(results));
+    }
+
+    default ImportPortalNavigationUseCase.Input map(String organizationId, String environmentId, ImportPortalNavigationRequest request) {
+        return ImportPortalNavigationUseCase.Input.builder()
+            .organizationId(organizationId)
+            .environmentId(environmentId)
+            .title(request.getTitle())
+            .parentId(map(request.getParentId()))
+            .visibility(map(request.getVisibility()))
+            .source(map(request.getSource()))
+            .build();
+    }
+
+    io.gravitee.apim.core.portal_page.model.PortalVisibility map(io.gravitee.rest.api.management.v2.rest.model.PortalVisibility visibility);
+
     // Hand-built because the fetch state is readOnly in the OpenAPI spec: the generated model only
     // exposes it through its @JsonCreator constructor, which MapStruct cannot target.
     default io.gravitee.rest.api.management.v2.rest.model.PortalNavigationItemSource map(
@@ -185,7 +217,8 @@ public interface PortalNavigationItemsMapper {
         return new io.gravitee.rest.api.management.v2.rest.model.PortalNavigationItemSource(
             DateMapper.INSTANCE.map(source.getLastFetchedAt()),
             DateMapper.INSTANCE.map(source.getLastFetchAttemptAt()),
-            source.getLastFetchError()
+            source.getLastFetchError(),
+            source.isSubtreeImport()
         )
             .type(source.getSourceType())
             .configuration(ConfigurationSerializationMapper.INSTANCE.deserializeConfiguration(source.getSourceConfiguration()))
@@ -199,6 +232,7 @@ public interface PortalNavigationItemsMapper {
     @Mapping(target = "lastFetchedAt", ignore = true)
     @Mapping(target = "lastFetchAttemptAt", ignore = true)
     @Mapping(target = "lastFetchError", ignore = true)
+    @Mapping(target = "subtreeImport", ignore = true)
     io.gravitee.apim.core.portal_page.model.PortalNavigationItemSource map(
         io.gravitee.rest.api.management.v2.rest.model.PortalNavigationItemSource source
     );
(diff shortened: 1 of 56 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationBulkImportDomainService.java
```diff
@@ -0,0 +1,464 @@
...
+                rootFolder,
+                resolved.fromManifest()
+                    ? "The %s manifest of the source lists no importable pages.".formatted(manifestParser.manifestFileName())
+                    : "The source of type %s listed no importable files.".formatted(source.getSourceType())
+            );
+        }
+
+        var touchedItemIds = new HashSet<PortalNavigationItemId>();
+        var results = importEntries(rootFolder, entries, touchedItemIds);
+        deleteOrphans(rootFolder, touchedItemIds);
```
jaroslawlakomy-grav: Could we avoid pruning orphans when any entry failed, or reconcile them by a stable source identity? If a manifest renames `a.md` from `Old` to `New` and fetching it transiently fails, this run touches no page under the new title and the unconditional orphan pass deletes the last-good `Old` page. Could we keep stale items until a fully successful reconciliation and cover this case with a regression test?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/ImportedFileContentType.java
```diff
@@ -0,0 +1,49 @@
...
+    private static final Pattern ASYNCAPI_MARKER = Pattern.compile("(?m)^\\s*\"?asyncapi\"?\\s*:");
+
+    private ImportedFileContentType() {}
+
+    public static Optional<PortalPageContentType> from(String fileName, String content) {
+        var extension = extensionOf(fileName);
+        return switch (extension) {
+            // No AsciiDoc content type in the NG portal: .adoc is rejected rather than rendered broken
+            case "md" -> Optional.of(PortalPageContentType.GRAVITEE_MARKDOWN);
+            case "yaml", "yml", "json" -> Optional.of(
```
jaroslawlakomy-grav: Could we detect structured document types from the parsed root object instead of defaulting every JSON/YAML file to OpenAPI? In mirror mode this imports files such as `package.json` as broken OpenAPI pages; compact `{"asyncapi":"3.0.0"}` is classified as OpenAPI, while a nested `asyncapi` key can classify an OpenAPI document as AsyncAPI. The Console import already checks root `asyncapi`, `openapi`, and `swagger` properties; could we align this path and add regression tests?

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/PortalNavigationItemsMapper.java
```diff
@@ -174,6 +178,34 @@ default SeedDefaultPagesForPortalNavigationItemsUseCase.Input mapSeedDefaultPage
 
     FetchPortalNavigationItemResponse map(FetchPortalNavigationItemUseCase.Output output);
 
+    PortalNavigationItemFetchResult map(PortalNavigationBulkImportDomainService.BulkImportResult.FileImportResult result);
```
jaroslawlakomy-grav, line 181: Could we align failed import results with the REST contract? Every failure carries a null `navigationItemId`, while `PortalNavigationItemFetchResult` requires a non-null UUID, so expected listing and per-file failures produce responses generated clients consider invalid. Could we make the field optional for import results, or introduce an import-specific result, and add a resource-level failure test?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationBulkImportDomainService.java
```diff
@@ -0,0 +1,464 @@
...
+    private PortalNavigationItemId resolveTargetFolder(
+        PortalNavigationFolder rootFolder,
+        String destinationPath,
+        Map<String, PortalNavigationItemId> folderIdsByPath,
+        Set<PortalNavigationItemId> touchedItemIds
+    ) {
+        touchedItemIds.add(rootFolder.getId());
+        var currentParentId = rootFolder.getId();
+        var currentPath = "";
+        for (var pathElement : destinationPath.split("/")) {
```
jaroslawlakomy-grav, line 364: Could we reject imports whose resulting depth reaches the sourced-ancestor limit, or make depth overflow fail closed? A destination with 50 nested folders puts the sourced root beyond the 50-hop check for the imported page, so regular update/delete operations no longer recognize that page as source-managed and read-only. A boundary test would help preserve this invariant.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationBulkImportDomainService.java
```diff
@@ -0,0 +1,464 @@
...
+        }
+        return currentParentId;
+    }
+
+    private PortalNavigationItemId findOrCreateChildFolder(
+        PortalNavigationFolder rootFolder,
+        PortalNavigationItemId parentId,
+        String title
+    ) {
+        var existing = queryService
```
jaroslawlakomy-grav: Could we serialize re-imports per root, including manual and scheduled runs, or make reconciliation idempotent? This check-then-create uses random IDs and there is no sibling uniqueness constraint, so overlapping `_fetch` runs can both create the same folder/page; their orphan passes may also interfere with each other.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationBulkImportDomainService.java
```diff
@@ -0,0 +1,464 @@
...
+        Set<PortalNavigationItemId> touchedItemIds
+    ) {
+        var folderIdsByPath = new HashMap<String, PortalNavigationItemId>();
+        var results = new ArrayList<BulkImportResult.FileImportResult>();
+        var seenTitles = new HashSet<String>();
+        for (var entry : entries) {
+            // Two entries landing on the same (destination, title) — api.md next to api.yaml, or two
+            // manifest pages sharing name and dest — would silently overwrite each other's content
+            // while both report success; failing the second one keeps every failure visible
+            if (!seenTitles.add(entry.destinationPath() + "/" + entry.title())) {
```
jaroslawlakomy-grav: Could we canonicalize the destination before using it as the duplicate key? `resolveTargetFolder` ignores blank path segments, so `/guides`, `guides/`, and `guides//` resolve to the same folder while this set treats them as different destinations. Two entries with the same title can therefore update one page twice and both report success.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/ImportPortalNavigationUseCase.java
```diff
@@ -0,0 +1,99 @@
...
+
+        validatorService.validateAll(List.of(itemToCreate), input.environmentId());
+        if (!sourceDomainService.supportsFileListing(input.source())) {
+            throw InvalidPortalNavigationItemSourceException.sourceCannotListFiles(input.source().getSourceType());
+        }
+        // The marker is what later routes a fetch of this folder to a re-import
+        input.source().setSubtreeImport(true);
+
+        var rootFolder = (PortalNavigationFolder) domainService.create(input.organizationId(), input.environmentId(), itemToCreate);
+        var result = bulkImportDomainService.importSubtree(rootFolder);
```
jaroslawlakomy-grav: Could we put server-side bounds around this synchronous import, such as file count, resulting depth, per-file size, and total bytes? The root is persisted before `importSubtree`, then every remote file is fetched within the use-case transaction and ultimately read with `readAllBytes`; a large or slow repository can hold a JDBC transaction/connection and consume unbounded memory. The direct MAPI page import already applies a 10 MB content limit, so defining corresponding limits here seems worth considering.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationBulkImportDomainService.java
thomasballivet: Done. Orphans are only pruned when every entry imported successfully, so a transient failure on a renamed entry no longer deletes the last-good page. Regression test covers the `Old` -> `New` rename case.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/ImportedFileContentType.java
thomasballivet: Done. The type now comes from the parsed root object (`asyncapi` / `openapi` / `swagger`), aligned with the Console; anything else is not a document and is left out of a mirror import. Tests cover compact JSON and the nested `asyncapi` key.

gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/java/io/gravitee/rest/api/management/v2/rest/mapper/PortalNavigationItemsMapper.java
thomasballivet: Done. `navigationItemId` is no longer required on `PortalNavigationItemFetchResult` -- `_fetch` returns the same results for an import-managed folder, so both endpoints need it optional. Resource-level tests added for a per-file and a listing failure.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationBulkImportDomainService.java
thomasballivet: Done. An entry whose destination reaches `MAX_TREE_DEPTH` now fails instead of creating a page the read-only walk cannot protect. The boundary test asserts `findSourcedAncestor` still resolves at the deepest allowed level.

jaroslawlakomy-grav (approved):
The PR looks noticeably better after the latest changes. Thanks for addressing the previous feedback and adding the regression coverage. 

Before merging please take care of SonarCloud warnings and errors

PR #19381, opened 2026-08-25
refactor(portal-next): consolidate portal folder and listing validation upfront through shared rule set

konrad-krol-incubly:
## Issue

## Description

Preperation work for EXT-179 (visibility on automation-managed navigation): consolidates portal folder and listing validation into a single upfront pass through the shared rule collector, so any invalid plan is rejected before any DB write.

Previously, segment-conflict checks were duplicated across three ad-hoc call sites (folder executor, listing materializer, link sync) and fired mid-mutation - a failing item mid-plan could leave earlier items already committed.
  
Docs and links stay untouched - they already validate immediately before their single mutation, so they're atomic by construction.

##  Design note
  
New `PortalNavigationValidator` interface in `core.portal.domain_service.navigation`, implemented by the existing `PortalNavigationItemValidatorService`. This is standard dependency inversion - avoids a back-edge from `core.portal.domain_service` into `core.portal_page.domain_service`.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanExecutor.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanExecutor.java
@@ -19,7 +19,6 @@ import io.gravitee.apim.core.DomainService;
 import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.apim.core.portal.domain_service.navigation.NavigationFolderMapper;
 import io.gravitee.apim.core.portal.domain_service.navigation.actions.FolderActions;
-import io.gravitee.apim.core.portal.exception.PathConflictException;
 import io.gravitee.apim.core.portal.model.PortalArea;
 import io.gravitee.apim.core.portal_page.crud_service.PortalNavigationItemCrudService;
 import io.gravitee.apim.core.portal_page.crud_service.PortalPageContentCrudService;
@@ -114,8 +113,6 @@ public final class NavigationSyncPlanExecutor {
         Function<String, PortalNavigationItemId> idFactory
     ) {
         final var folderId = idFactory.apply(df.path());
-        final var parentId = parent == null ? null : parent.getId();
-        rejectIfSegmentTakenByForeignItem(auditInfo, parentId, df.segment().value(), folderId, df.path());
         final var create = CreatePortalNavigationItem.builder()
             .id(folderId)
             .title(df.title())
@@ -129,21 +126,6 @@ public final class NavigationSyncPlanExecutor {
         return (PortalNavigationFolder) crudService.create(
             PortalNavigationItem.from(create, auditInfo.organizationId(), auditInfo.environmentId(), parent)
         );
-    }
-
-    private void rejectIfSegmentTakenByForeignItem(
-        AuditInfo auditInfo,
-        @Nullable PortalNavigationItemId parentId,
-        String segment,
-        PortalNavigationItemId expectedId,
-        String path
-    ) {
-        queryService
-            .findByParentIdAndSegment(auditInfo.environmentId(), parentId, segment)
-            .filter(sibling -> !sibling.getId().equals(expectedId))
-            .ifPresent(squatter -> {
-                throw PathConflictException.folderPath(path);
-            });
     }
 
     private PortalNavigationFolder applyUpdate(
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanExecutor.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanExecutor.java
@@ -19,6 +19,7 @@ import io.gravitee.apim.core.DomainService;
 import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.apim.core.portal.domain_service.navigation.NavigationFolderMapper;
 import io.gravitee.apim.core.portal.domain_service.navigation.actions.FolderActions;
+import io.gravitee.apim.core.portal.exception.PathConflictException;
 import io.gravitee.apim.core.portal.model.PortalArea;
 import io.gravitee.apim.core.portal_page.crud_service.PortalNavigationItemCrudService;
 import io.gravitee.apim.core.portal_page.crud_service.PortalPageContentCrudService;
@@ -113,6 +114,8 @@ public final class NavigationSyncPlanExecutor {
         Function<String, PortalNavigationItemId> idFactory
     ) {
         final var folderId = idFactory.apply(df.path());
+        final var parentId = parent == null ? null : parent.getId();
+        rejectIfSegmentTakenByForeignItem(auditInfo, parentId, df.segment().value(), folderId, df.path());
         final var create = CreatePortalNavigationItem.builder()
             .id(folderId)
             .title(df.title())
@@ -126,6 +129,21 @@ public final class NavigationSyncPlanExecutor {
         return (PortalNavigationFolder) crudService.create(
             PortalNavigationItem.from(create, auditInfo.organizationId(), auditInfo.environmentId(), parent)
         );
+    }
+
+    private void rejectIfSegmentTakenByForeignItem(
+        AuditInfo auditInfo,
+        @Nullable PortalNavigationItemId parentId,
+        String segment,
+        PortalNavigationItemId expectedId,
+        String path
+    ) {
+        queryService
+            .findByParentIdAndSegment(auditInfo.environmentId(), parentId, segment)
+            .filter(sibling -> !sibling.getId().equals(expectedId))
+            .ifPresent(squatter -> {
+                throw PathConflictException.folderPath(path);
+            });
     }
 
     private PortalNavigationFolder applyUpdate(
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/ApiFolderSubtreeReconciler.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/ApiFolderSubtreeReconciler.java
@@ -71,19 +71,22 @@ class ApiFolderSubtreeReconciler {
         planExecutor.execute(plan, auditInfo, navApi.getArea(), navApi, idMapper, deleteStrategy);
     }
 
-    void validateConflicts(AuditInfo auditInfo, String apiId, List<NavigationPath> desired, List<PortalNavigationItem> currentFolders) {
-        var safeDesired = desired == null ? List.<NavigationPath>of() : desired;
-        var ownership = ownership(auditInfo, apiId, safeDesired);
-        NavigationSyncPlanner.plan(safeDesired, currentFolders, safeDesired, ownership);
-    }
-
     ValidationItems itemsForValidation(
         AuditInfo auditInfo,
         String apiId,
         PortalNavigationApi navApi,
         List<PortalNavigationItem> currentFolders
     ) {
-        var desired = desiredPaths(apiId);
+        return itemsForValidation(auditInfo, apiId, navApi, desiredPaths(apiId), currentFolders);
+    }
+
+    ValidationItems itemsForValidation(
+        AuditInfo auditInfo,
+        String apiId,
+        PortalNavigationApi navApi,
+        List<NavigationPath> desired,
+        List<PortalNavigationItem> currentFolders
+    ) {
         var ownership = ownership(auditInfo, apiId, desired);
         var plan = NavigationSyncPlanner.plan(desired, currentFolders, List.of(), ownership);
         Function<String, PortalNavigationItemId> idFactory = path -> apiFolderId(auditInfo, apiId, path);
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/ApiFolderSubtreeReconciler.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/ApiFolderSubtreeReconciler.java
@@ -19,21 +19,29 @@ import io.gravitee.apim.core.DomainService;
 import io.gravitee.apim.core.api.crud_service.ApiCrudService;
 import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.apim.core.portal.domain_service.navigation.PortalNavigationTreeWalker;
+import io.gravitee.apim.core.portal.domain_service.navigation.PortalNavigationValidator.PendingUpdate;
 import io.gravitee.apim.core.portal.domain_service.navigation.PortalNavigationVisitor;
+import io.gravitee.apim.core.portal.domain_service.navigation.actions.FolderActions;
 import io.gravitee.apim.core.portal.domain_service.navigation.plan.NavigationOwnership;
 import io.gravitee.apim.core.portal.domain_service.navigation.plan.NavigationSyncPlanExecutor;
 import io.gravitee.apim.core.portal.domain_service.navigation.plan.NavigationSyncPlanner;
 import io.gravitee.apim.core.portal.model.NavigationPath;
+import io.gravitee.apim.core.portal.model.PortalArea;
 import io.gravitee.apim.core.portal.query_service.AutomationManagedNavigationItemsQueryService;
+import io.gravitee.apim.core.portal_page.model.AutomationMetadata;
+import io.gravitee.apim.core.portal_page.model.CreatePortalNavigationItem;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationApi;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationFolder;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItem;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemQueryCriteria;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemType;
+import io.gravitee.apim.core.portal_page.model.PortalVisibility;
+import io.gravitee.apim.core.portal_page.model.UpdatePortalNavigationItem;
 import io.gravitee.apim.core.portal_page.query_service.PortalNavigationItemsQueryService;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import lombok.RequiredArgsConstructor;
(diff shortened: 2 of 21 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/plan/NavigationSyncPlanExecutor.java
```diff
@@ -131,21 +128,6 @@ private PortalNavigationFolder createFolder(
         );
     }
 
-    private void rejectIfSegmentTakenByForeignItem(
```
konrad-krol-incubly, line 134: FIY: This was moved to specific validation rule.

MateuszJozwiak-Incubly:
a small remark to consider: 
When the PortalNavigationApi row already exists, this branch returns validation items only for its folder subtree (PortalListingSyncDomainService.java:126-131, via
  ApiFolderSubtreeReconciler.itemsForValidation() at ApiFolderSubtreeReconciler.java:80-105). It does not include the listing row itself in the updates passed to
  validatorService.validateAllUpdates() at PortalListingSyncDomainService.java:107.

  However, the later upsert changes the existing row’s segment and parent in NavigationItemEntryMaterializer.applyUpdate() (NavigationItemEntryMaterializer.java:121-129) and
  persists it at lines 53-56. If another item already occupies the resulting (parent, segment), this creates duplicate navigation routes.

  Please consider including a PendingUpdate for the existing navApi row in this branch and ensure SegmentConflictRule is applied to updates—currently it appears only in createRules at
  PortalNavigationItemValidatorService.java:87-102, not updateRules at lines 103-114.

MateuszJozwiak-Incubly:
And one more suggestion to analyze:
The API CRD path does not use the new shared item validator. ImportApiCRDUseCase calls validateApiFolderConflictsForApi() before persistence (ImportApiCRDUseCase.java:198-202
  and 273-277), but that method only runs NavigationSyncPlanner.plan() over existing folders (PortalListingSyncDomainService.java:136-143, ApiFolderSubtreeReconciler.java:74-
  78). Because currentFolders is folder-only, this cannot detect a desired API folder whose (parent, segment) is already occupied by a link or another non-folder item.

  The CRD is then applied through syncApiFolders() (ImportApiCRDUseCase.java:306 → PortalListingSyncDomainService.java:78-85), which reaches createFolder() here without another
  conflict check (NavigationSyncPlanExecutor.java:108-128).

  Please consider retaining the previous findByParentIdAndSegment() check in createFolder(), or route this CRD path through the shared segment-conflict validator before mutation.

konrad-krol-incubly:
> a small remark to consider: When the PortalNavigationApi row already exists, this branch returns validation items only for its folder subtree (PortalListingSyncDomainService.java:126-131, via ApiFolderSubtreeReconciler.itemsForValidation() at ApiFolderSubtreeReconciler.java:80-105). It does not include the listing row itself in the updates passed to validatorService.validateAllUpdates() at PortalListingSyncDomainService.java:107.
> 
> However, the later upsert changes the existing row’s segment and parent in NavigationItemEntryMaterializer.applyUpdate() (NavigationItemEntryMaterializer.java:121-129) and persists it at lines 53-56. If another item already occupies the resulting (parent, segment), this creates duplicate navigation routes.
> 
> Please consider including a PendingUpdate for the existing navApi row in this branch and ensure SegmentConflictRule is applied to updates—currently it appears only in createRules at PortalNavigationItemValidatorService.java:87-102, not updateRules at lines 103-114.

Fixed, SegmentConflictRule now also implements UpdatePortalNavigationItemValidationRule. The existing nav-api row goes into the update batch.

konrad-krol-incubly:
> And one more suggestion to analyze: The API CRD path does not use the new shared item validator. ImportApiCRDUseCase calls validateApiFolderConflictsForApi() before persistence (ImportApiCRDUseCase.java:198-202 and 273-277), but that method only runs NavigationSyncPlanner.plan() over existing folders (PortalListingSyncDomainService.java:136-143, ApiFolderSubtreeReconciler.java:74- 78). Because currentFolders is folder-only, this cannot detect a desired API folder whose (parent, segment) is already occupied by a link or another non-folder item.
> 
> The CRD is then applied through syncApiFolders() (ImportApiCRDUseCase.java:306 → PortalListingSyncDomainService.java:78-85), which reaches createFolder() here without another conflict check (NavigationSyncPlanExecutor.java:108-128).
> 
> Please consider retaining the previous findByParentIdAndSegment() check in createFolder(), or route this CRD path through the shared segment-conflict validator before mutation.

fixed, validateApiFolderConflictsForApi now goes through the shared validator (creates + updates)

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/validation/SegmentConflictRule.java
```diff
@@ -0,0 +1,130 @@
...
+            .pendingItemsById()
+            .values()
+            .stream()
+            .anyMatch(
+                other ->
+                    !Objects.equals(other.getId(), item.getId()) &&
+                    Objects.equals(other.getParentId(), item.getParentId()) &&
+                    Objects.equals(other.getSegment(), item.getSegment())
+            );
+    }
```
dev-marek, line 100: **Blocker — create/update collisions in the same batch aren't cross-checked.**

`collidesWithPendingBatch` only scans `CreateValidationContext.pendingItemsById()` — i.e. other pending *creates*. `UpdateValidationContext` carries no pending-batch data at all, and `validateAllUpdates` (the default method on `PortalNavigationValidator`) validates each update against a fresh, isolated context, one at a time. So a create and an update that land on the same `(parentId, segment)` within one `validateForConflicts()` call are never compared against each other — only against the DB (which doesn't yet reflect either pending write) and, for creates, against other pending creates.

Concretely: `PortalNavigationSyncDomainService.validateForConflicts` builds `creates` and `updates` across all areas of one sync, then calls `validatorService.validateAll(creates, envId)` and `validatorService.validateAllUpdates(updates)` separately. If the desired state simultaneously creates a new folder at `(P, "docs")` and updates a different, already-persisted folder to move it to that same `(P, "docs")`, both calls succeed and both mutations get applied — leaving two navigation items sharing the same parent+segment, which is exactly the invariant this rule exists to enforce.

I verified this with a runtime reproduction against this branch (compiled and run with JDK 25): a `PortalNavigationItemValidatorService.validateAll` call for the create and a separate `validateAllUpdates` call for the colliding update both pass with zero exceptions.

This also removes a previously-existing safety net without a full replacement: pre-PR, `NavigationSyncPlanExecutor.createFolder` called `rejectIfSegmentTakenByForeignItem` against the DB immediately before each create, so if the colliding update happened to execute first in the same run, the DB write would already be visible and the later create would have been rejected. That check is gone now.

Suggested fix: give `SegmentConflictRule` (or the two validation entry points) a single shared "pending items" view spanning both creates and updates — or run one unified validation pass instead of two independent ones — and add a test for the create-vs-update collision case alongside `SegmentConflictRuleTest`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/PortalListingSyncDomainService.java
```diff
@@ -44,6 +50,7 @@ public class PortalListingSyncDomainService {
     private final ApiDocumentationSyncDomainService apiDocumentationSyncDomainService;
```
dev-marek, line 52: **Nit — inconsistent with the PR's own design rationale.**

This injects the concrete `PortalNavigationItemValidatorService`, while `PortalNavigationSyncDomainService` injects the new `PortalNavigationValidator` interface for the identical `validateAll`/`validateAllUpdates` calls. Not a functional issue — `portal_listing.domain_service` already imports other `portal_page.domain_service` classes (e.g. `ApiDocumentationSyncDomainService`), so there's no cyclic-package risk here — but it undercuts the stated "standard dependency inversion" rationale from the PR description by applying it to only one of the two call sites doing the same job. Worth using the interface here too for consistency, or noting why this one doesn't need it.

dev-marek (changes requested):
The consolidation is well-executed and the goal fit is otherwise good — `SegmentConflictRule` cleanly centralizes what used to be three ad-hoc call sites, and the routing tests (creates vs. updates) are solid. But the core promise of the PR ("any invalid plan is rejected before any DB write") has a confirmed gap: a create and an update in the same sync batch that resolve to the same `(parentId, segment)` both pass validation and both get persisted, since `validateAll` and `validateAllUpdates` run as two independent passes with no shared view of each other's pending items. I reproduced this on the PR branch (JDK 25): a folder renamed via update to `(root, "shared")` plus a new folder created at `(root, "shared")` in the same call — neither `validateAll` nor `validateAllUpdates` throws. This is a regression, not just an incomplete new feature: the pre-PR `NavigationSyncPlanExecutor.createFolder` ran a DB-backed segment check immediately before every create, which was deleted here with no equivalent replacement for the cross-batch case. See inline comment for details and a suggested fix direction.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/validation/SegmentConflictRule.java
konrad-krol-incubly: Fixed. Went with your unified pass suggestion - added a shared `pendingSegmentClaims` view on both contexts and collapsed the two calls into one `validate(creates, updates, envId)`. Test mirroring your scenario added in SegmentConflictRuleTest

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/PortalListingSyncDomainService.java
konrad-krol-incubly: Swapped it to the interface in the same commit for consistency.

FYI the interface exists because `PortalNavigationSyncDomainService` is placed in `core.portal.domain_service` and `core.portal_page.domain_service` already depends back that way via `ValidatePortalDocumentationDomainService → PortalAutomationScopeDomainService` - so injecting the concrete there closes a cycle. Listing doesn't have that pressure (as you noted), but keeping it on the interface for consistency.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/validation/SegmentConflictRule.java
```diff
@@ -0,0 +1,130 @@
...
+        return collidesWithPersistedSibling(item.getId(), item.getParentId(), item.getSegment(), environmentId);
+    }
+
+    private boolean collidesWithPersistedSibling(
+        io.gravitee.apim.core.portal_page.model.PortalNavigationItemId itemId,
+        io.gravitee.apim.core.portal_page.model.PortalNavigationItemId parentId,
+        String segment,
+        String environmentId
+    ) {
+        return navigationItemsQueryService
```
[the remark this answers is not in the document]
konrad-krol-incubly: This mirrors how per-item rules work across the codebase - UniqueItemIdRule, HomepageUniquenessRule, ParentRule all query per item. Batch sizes here are small (folder plans, listings), so the overhead is negligible in practice, but it will overcomplicate whole solution.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/NavigationItemEntryMaterializer.java
```diff
@@ -84,26 +85,53 @@ PortalNavigationItem findExistingRow(AuditInfo auditInfo, PortalNavigationItemId
...
-    }
-
-    private void rejectIfSegmentTakenByForeignItem(
+    PortalNavigationValidator.PendingUpdate updateForValidation(
         AuditInfo auditInfo,
-        PortalNavigationItemContainer parent,
-        String segment,
-        PortalNavigationItemId expectedId,
-        String location
+        PortalId portalId,
```
konrad-krol-incubly, line 92: already removed

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/PortalListingSyncDomainService.java
```diff
@@ -89,32 +96,72 @@ public void dematerialize(AuditInfo auditInfo, PortalId portalId, PortalListing
...
+
+    private PortalNavigationItem findExistingRow(AuditInfo auditInfo, PortalId portalId, String apiId) {
+        return navigationItemEntryMaterializer.findExistingRow(
+            auditInfo,
+            navigationItemEntryMaterializer.rowId(auditInfo, portalId, apiId)
+        );
+    }
+
+    private ValidationItems validationItemsForNewRow(AuditInfo auditInfo, PortalId portalId, String apiId, PortalListingApiEntry entry) {
+        var create = navigationItemEntryMaterializer.itemForValidation(auditInfo, portalId, apiId, entry);
```
dev-marek: **Blocker — new listing rows skip folder-subtree validation entirely.**

`PortalListingSyncDomainService.validationItemsForNewRow` (this file, ~L131-134) only builds the listing row's own create item — it never calls `apiFolderSubtreeReconciler.itemsForValidation(...)` for the API's declared `portalNavigation` folder subtree. But `sync()` (L71) unconditionally materializes that subtree via `ApiFolderSubtreeReconciler.sync()` → `NavigationSyncPlanExecutor.createFolder()`, and that method's own guard (`rejectIfSegmentTakenByForeignItem`) was deleted by this PR with nothing replacing it for this path.

The pre-existing `NavigationSyncPlanner.rejectConflicts` (unchanged by this PR) doesn't backstop it either — it only sees FOLDER-type siblings (`loadAllFoldersInEnv` filters `type=FOLDER`), so it can't catch a new API folder colliding with a non-folder item (e.g. a `Link`) at the same `(parentId, segment)`.

Net effect: adding an API with a declared folder path to a listing for the first time can silently create a folder at the same navigation slot as a pre-existing `Link`/`API`/`API_PRODUCT` item — two items sharing one slot, with no exception from either the new upfront validator or the mutation path.

I verified this with a runtime reproduction against `7266452826` (current head): seeded a `PortalNavigationLink` at `(parentId = deterministic navApiId, segment = "reports")`, gave the not-yet-listed API `portalNavigation = [/reports]`, then called `validateForConflicts` (no throw) followed by `sync` (no throw; storage ends up with 2 items at that slot instead of 1). Swapping the foreign item for a `PortalNavigationFolder` instead *does* throw — via the untouched `NavigationSyncPlanner.rejectConflicts` — confirming the gap is specifically the cross-type case, exactly what the deleted `rejectIfSegmentTakenByForeignItem` (unfiltered by type) used to catch.

None of the PR's new tests would catch this — they never populate the API's `portalNavigation`, so `hasSize(1)` on the new-row test is trivially true regardless of whether folder items are included.

This is the same root pattern as the two issues already fixed on this PR (existing-row updates missing the row itself; the CRD path using a folder-only planner) — just the remaining "new row" case.

Suggested fix: have `validationItemsForNewRow` also build the folder-subtree creates for the new row — the deterministic row id (`navigationItemEntryMaterializer.rowId(...)`) is computable before the row is persisted, so `ApiFolderSubtreeReconciler`'s existing `toCreateItem`/`toUpdateItem` logic can be reused with that id as the parent — or keep an equivalent guard inside `NavigationSyncPlanExecutor.createFolder()` as a last line of defense regardless of caller.

dev-marek (changes requested):
The consolidation is close to complete now that the create/update cross-batch collision and the CRD-path folder-only check are fixed. One more instance of the same root pattern remains: brand-new listing rows skip folder-subtree validation entirely, so the upfront pass doesn't yet fully deliver on "any invalid plan is rejected before any DB write." I verified this with a runtime reproduction against the current head (`7266452826`) — see inline comment for details and a suggested fix direction.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/PortalListingSyncDomainService.java
konrad-krol-incubly: Fixed - `validationItemsForNewRow` now also emits the folder-subtree creates, so `SegmentConflictRule` sees them upfront. Added a test coverage.

Also I decided to restore both `rejectIfSegmentTakenByForeignItem` guards. Upfront batch validation stays as the primary check, but good to keep original guards in place, as you flagged too many gaps within PR to be able to safely remove it.

Also, ParentRule now checks the pending-parent map on the update side too (mirror of the create-side lookup).

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/ApiFolderSubtreeReconciler.java
```diff
@@ -63,10 +71,82 @@ void sync(
...
+
+    ValidationItems itemsForValidation(
+        AuditInfo auditInfo,
+        String apiId,
+        PortalNavigationItemId navigationApiId,
+        PortalArea area,
+        List<NavigationPath> desired,
+        List<PortalNavigationItem> currentFolders
+    ) {
+        var ownership = ownership(auditInfo, apiId, desired);
```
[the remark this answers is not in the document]
konrad-krol-incubly: I added safe resolution at the domain entry level

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/validation/SegmentConflictRule.java
```diff
@@ -0,0 +1,134 @@
...
+            .stream()
+            .anyMatch(
+                claim ->
+                    !Objects.equals(claim.id(), itemId) &&
+                    Objects.equals(claim.parentId(), parentId) &&
+                    Objects.equals(claim.segment(), segment)
+            );
+    }
+
+    private boolean collidesWithPersistedSibling(
```
dev-marek, line 102: **Improvement — false conflict when a batch vacates and fills the same slot in one pass.**

`collidesWithPersistedSibling` (here and the create-side overload) excludes a DB row only when it matches the *item being validated*. It has no way to see that a different item's pending update, elsewhere in the same `pendingSegmentClaims` batch, is about to move away from that exact `(parentId, segment)`.

Concretely, in one `PortalListingSyncDomainService.validateForConflicts()` call: entry A (segment "foo", currently at parent P) is relocated to parent Q — its `PendingUpdate` claim only records the *new* target `(Q, "foo")`, never that `(P, "foo")` is being freed. If entry B (a different, unrelated entry whose hrid also slugifies to "foo") is newly placed at parent P in the same payload, B's create checks `(P, "foo")`: `collidesWithPendingBatch` sees no claim there (A's claim is at `(Q, "foo")`), but `collidesWithPersistedSibling` still finds A's row sitting at `(P, "foo")` in the DB (`sibling.getId() != B.getId()`) and throws — even though the plan is internally consistent and A is vacating that slot in this very batch.

I verified this directly against `SegmentConflictRule` on this branch: a `CreateValidationContext` with a `PendingSegmentClaim` recording another item's *new* target (not its old slot), plus a DB row still sitting at the *old* slot, causes `rule.validate()` to throw for a create targeting that old slot — confirmed with a temporary unit test in an isolated worktree (removed afterward).

This is narrower than the two blockers already fixed here (over-rejection of a valid plan vs. silently persisting a duplicate slot), and likely requires a segment/slug collision between two distinct entries to trigger via the listing path specifically (folder and API-CRD paths key IDs by path, so a genuine reparent-while-keeping-id — and thus this race — isn't reachable there). Still, it's the same underlying gap: pending claims record only destinations, never departures.

Non-blocking — suggest tracking this as a follow-up rather than holding up this PR. Suggested fix direction: when building `pendingSegmentClaims` (or when checking `collidesWithPersistedSibling`), also record each update's *origin* `(existingItem's current parentId, segment)` so a persisted sibling occupying a slot that's being vacated in the same batch is recognized as free rather than compared only by id.

dev-marek (approved):
Re-reviewed after the fixes for the two prior blockers (create/update cross-batch collision, new-row folder-subtree validation) — both check out against current head. I also resolved the one thread that had a "Fixed" reply but was left unresolved (new-row folder-subtree validation).

Approving. One more instance of the same root pattern (batch-aware slot accounting) survives — see inline comment for a verified repro — but it's narrower than the two already-fixed blockers (over-rejection of a valid plan, not silent duplication of a slot) and doesn't need to hold up this PR. Suggest tracking it and fixing in a follow-up.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/validation/SegmentConflictRule.java
dev-marek: Since this is still open and we're actively following up on the same validator area in #19502 — want to pick this up as a small follow-up, either here or bundled into the next EXT-179 PR? It's non-blocking, but the underlying gap (pending claims only record destinations, not departures) is the same class of thing we've been tightening up elsewhere.

konrad-krol-incubly: Noted for follow-up PR

PR #19387, opened 2026-08-25
fix(gateway): resolve dot segments before routing [4.12.x]

ytvnr:
## Issue

## Description

Port of the path-normalization fix from 4.11.x (#19166) to 4.12.x, consuming `RequestPathNormalizer` from `gravitee-common` instead of vendoring it into the gateway reactor.

The gateway reads `..;jsessionid=1` as an ordinary segment per RFC 3986 §5.2.4, while Tomcat/Jetty/Spring strip `;params` first and see a dot segment. That parser differential lets a request bypass a policy written against the normalized path. The new `http.pathHandling` setting takes `RAW` (default, unchanged behaviour), `REJECT` or `NORMALIZE`, and resolves dot segments before the listener context path is resolved — therefore before any plan is enforced. Rejected requests are reported under `handlers.rejected.analytics.enabled` (default `true`).

`;params` on **ordinary** segments are deliberately left alone; the limitation is documented in `gravitee.yml`, `helm/values.yaml` and the docs PR ([gravitee-platform-docs#2249](https://github.com/gravitee-io/gravitee-platform-docs/pull/2249)).

The 12 commits from the 4.11.x branch are cherry-picked unchanged. Two commits sit on top:

- `refactor(gateway): consume RequestPathNormalizer from gravitee-common` — drops the vendored class and its two test classes, rewrites the imports, bumps `gravitee-common`.
- `fix(gateway): adapt the ported path handling to the Vert.x 5 APIs` — 4.12.x no longer carries `io.grpc` on the reactor classpath, so `RejectedPathProcessor` now uses `io.vertx.grpc.common.GrpcStatus`, the way its neighbour `NotFoundProcessor` already does on this branch. The value put on the wire is unchanged (`"3"` for `INVALID_ARGUMENT`). The debug dispatcher test moves to `authority()`, `io.vertx.core.MultiMap` and `channelHandlerContext().channel()`.

## Additional context

**Do not merge on the snapshot.** This branch pins `gravitee-common` to `5.2.0-feat-request-path-normalizer-SNAPSHOT` ([gravitee-common#170](https://github.com/gravitee-io/gravitee-common/pull/170)). It must move to `5.2.0` once that release is out. Merge order with #19166 (4.11.x), gravitee-common #170 and the master port is still to be settled.

Verified locally on JDK 21: `clean install` green on the gateway chain (1003 tests in `gateway-reactor`) and on `gateway-services-debug` (152 tests), the 7 `pathtraversal` integration test classes green (46 tests), and the ported Helm unit test green (4 tests).

Manual testing needs `curl --path-as-is`, and the test API needs a `START_WITH` flow selector — without either you get silent false negatives.


pom.xml
```diff
@@ -55,7 +55,7 @@
     <gravitee-cockpit-api.version>3.12.0</gravitee-cockpit-api.version>
     <gravitee-cloud-initializer.version>2.3.1</gravitee-cloud-initializer.version>
     <logback-ecs-encoder.version>1.7.0</logback-ecs-encoder.version>
-    <gravitee-common.version>5.1.0</gravitee-common.version>
+    <gravitee-common.version>5.2.0-feat-request-path-normalizer-SNAPSHOT</gravitee-common.version>
```
phiz71: **MUST** -- the snapshot pin, which you flag yourself in the description. Recording the order here so it does not get lost between four PRs:

1. merge gravitee-io/gravitee-common#170 -- it is approved, and the only thing still open there is a missing JIRA footer on its commit;
2. wait for the release, which will be **`5.2.0`** and not `5.1.0` -- `5.1.0` was already tagged on 2026-08-12, and a `feat` commit takes the minor;
3. pin `5.2.0` here.

Note that `ci/circleci: Build backend` was still running when I reviewed, so I have not seen the snapshot actually resolve in CI.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -84,6 +84,20 @@ secrets:
...
+#  resolved and therefore before any plan is enforced.
+#    RAW       (default) the path is used, and forwarded, exactly as received.
+#    REJECT    answers 400 when the normalized path differs from the one received.
+#    NORMALIZE resolves the dot segments and routes, enforces and forwards on the result,
+#              so /alpha/api/../../beta/api is read as the call to beta that it is.
+#  Beyond dot segments, both REJECT and NORMALIZE also decode percent-encoded unreserved
+#  characters, merge duplicate slashes, and answer 400 on a malformed percent sequence.
+#  Encoded slashes are never decoded, and segment parameters on an ordinary segment are left
+#  alone: /a/admin;x/b is canonical here and forwarded as sent, while a Servlet receiver strips
+#  the ;x and serves /a/admin/b. An allow or deny rule written against /admin/** therefore still
```
phiz71, line 97: **SHOULD** -- this paragraph documents one limitation of `NORMALIZE` carefully, and leaves out its twin.

In `NORMALIZE`, `uri()` keeps reporting the bytes as received. `EvaluableRequest.getUri()` reads `request.uri()`, so a flow condition or a policy written on `{#request.uri}` still sees `/alpha/api/../../beta/api/echo` while routing, plan enforcement and the upstream call all work on `/beta/api/echo`. It is deliberate -- request signing depends on it, and `should_still_expose_the_uri_as_received` pins it -- but it is evaded exactly the way the `;x` case above is.

An operator reading this block learns to keep a guard against `/a/admin;x/b` and learns nothing about `{#request.uri}`. Two lines, right here, and the same in `helm/values.yaml`.

gravitee-apim-gateway/gravitee-apim-gateway-env/src/main/java/io/gravitee/gateway/env/RequestConfiguration.java
```diff
@@ -51,4 +53,38 @@ public RequestClientAuthConfiguration httpRequestClientAuthConfiguration(
...
+            // of bug that survives every test suite that does not set a locale.
+            handling = RequestPathHandling.valueOf(pathHandling.trim().toUpperCase(Locale.ROOT));
+        } catch (IllegalArgumentException e) {
+            log.warn(
+                "Unknown value [{}] for http.pathHandling, falling back to {}. Expected one of {}.",
+                pathHandling,
+                RequestPathHandling.RAW,
+                Arrays.toString(RequestPathHandling.values())
+            );
+            handling = RequestPathHandling.RAW;
```
phiz71, line 74: **SHOULD** -- a misspelt value falls back to `RAW` silently, and `RAW` is the mode with no protection at all.

This branch is only ever reached when somebody explicitly asked for something other than the default. An operator who writes `NORMALIZ` or `Reject ` believes the control is on, gets a `warn` line in a startup log nobody is watching, and runs unprotected. The `log.info` below helps whoever goes looking, which is not the same as telling whoever does not.

Failing to start would be the safer read: refusing to boot on a typo is loud, recoverable in seconds, and cannot be mistaken for success. The cost of the current behaviour is silent and indefinite.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -155,13 +167,53 @@ public Completable dispatch(HttpServerRequest httpServerRequest, String serverId
...
+
+        // One scan answers what both active modes ask. REJECT needs nothing more: a path that is
+        // not in the form it claims to be is refused without ever being rewritten. NORMALIZE only
+        // pays for the resolution when there is something to resolve.
+        final boolean needsNormalization = requestPathConfiguration.isEnabled() && RequestPathNormalizer.needsNormalization(rawPath);
+
+        final String normalizedPath;
+        if (needsNormalization) {
+            if (requestPathConfiguration.getHandling() == RequestPathHandling.REJECT) {
+                return handleRejectedPath(httpServerRequest, serverId, receivedAt);
```
phiz71, line 187: **COULD** -- rejection happens before the acceptor is resolved, so under `REJECT` an unknown host or context path answers 400 where it used to answer 404. I suspect that is deliberate and right -- resolving first would turn the gateway into an enumeration oracle, telling a prober which context paths exist before refusing them.

It is neither asserted in the integration tests nor stated in `gravitee.yml`, though, so today it reads as a side effect rather than as the choice it probably is. One test named for it would pin it.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
```diff
@@ -0,0 +1,93 @@
...
+            if (MediaType.MEDIA_APPLICATION_GRPC.equals(mediaType)) {
+                ctx.response().headers().set("grpc-status", GrpcStatus.INVALID_ARGUMENT.toString());
+                ctx.response().headers().set("grpc-message", message);
+            }
+
+            final Buffer body = Buffer.buffer(message);
+            // Counted on the encoded body, not on the string: a localised http.errors[400].message
+            // holding any non-ASCII character has more bytes than it has characters, and a short
+            // Content-Length truncates the response.
+            ctx.response().headers().set(HttpHeaderNames.CONTENT_LENGTH, Integer.toString(body.length()));
```
phiz71, line 84: **COULD** -- this is right, and it makes its neighbour visibly wrong: `NotFoundProcessor.java:66` still does `Integer.toString(message.length())` on the string, so a localised `http.errors[404].message` carrying any non-ASCII character still truncates the 404 body.

Out of scope for this PR and it should stay out. Worth a ticket rather than a fix here.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
```diff
@@ -0,0 +1,93 @@
...
+
+            ctx.response().status(HttpStatusCode.BAD_REQUEST_400);
+
+            final String message = environment.getProperty("http.errors[400].message", DEFAULT_MESSAGE);
+
+            // A gRPC caller reads the status from the trailers, not from the HTTP code: without
+            // these it sees UNKNOWN and has no idea why it was refused. Same treatment as an
+            // unmatched context path, which is the closest thing this rejection resembles.
+            final MediaType mediaType = MediaType.parseMediaType(ctx.request().headers().get(HttpHeaderNames.CONTENT_TYPE));
+            if (MediaType.MEDIA_APPLICATION_GRPC.equals(mediaType)) {
```
phiz71, line 75: **COULD** -- `MediaType.MEDIA_APPLICATION_GRPC.equals(mediaType)` matches `application/grpc` exactly, so a caller sending `application/grpc+proto` gets the 400 without `grpc-status`, and reads `UNKNOWN`.

Copied from `NotFoundProcessor`, so the two agree, which is the more important property. Naming it only so the limitation is chosen rather than inherited.

phiz71 (commented):
Reviewed the chain rather than the diff alone, and the substance holds up.

**The differential is closed end to end**, which is the point everything else rests on. Decision and forward both run on the resolved path: `path()` -> `pathInfo()` (`AbstractRequest.updatePathInfoFromContextPath` on V4, `VertxHttpServerRequest.pathInfo()` on V3) -> `HttpConnector.java:691` (`uri + request.pathInfo()`) and `DefaultProxyEndpoint.java:36` (`target() + request.pathInfo()`). No path re-derives the upstream URI from the raw `uri()`.

**No entrypoint left behind**: one production `HttpRequestDispatcher` plus the debug one, and the debug one is handled here through an extension point rather than a second copy of the flow.

**The claims in the description check out.** `GrpcStatus.INVALID_ARGUMENT.toString()` is indeed `"3"` -- verified in the `vertx-grpc-common` 5.0.12 sources, where the constructor does `this.string = Integer.toString(code)` -- so moving off `io.grpc` changes nothing on the wire. And `MediaType.parseMediaType(null)` returns `null` rather than throwing, so a request with no Content-Type, which is the common case, does not blow up in the rejection path.

Worth calling out on the way past: `Locale.ROOT` on the `toUpperCase`, with the Turkish trap spelled out; and every integration mode has a `RAW` control class beside it, which is what makes those assertions conclusive rather than decorative.

Findings inline. One that has no line to sit on: **none of the 14 commits carries the JIRA footer**, though the PR references APIM-14917. Separately, `84efdf53 fix(gateway): harden path handling after review` describes the process rather than the change -- it is a cherry-pick from 4.11.x, so rewording it here would desync the two, which may well be the better trade anyway.

Verdict: sound on the merits, blocked only by the snapshot pin. Findings 2 and 3 are two paragraphs of documentation and one fail-fast decision; neither asks for the mechanism to be touched.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
ytvnr: Done in c9121b1a70 — four lines in `gravitee.yml`, five in `helm/values.yaml`, right after the `;x` paragraph. You are right that the block was documenting one half of the same hazard: it warned about the spelling a Servlet receiver strips, and said nothing about the one an EL expression reads. Both are now named, and the new text points back at the `;x` case explicitly so the two read as the pair they are.

gravitee-apim-gateway/gravitee-apim-gateway-env/src/main/java/io/gravitee/gateway/env/RequestConfiguration.java
ytvnr: Leaving as is, deliberately. The fallback mirrors `requestTimeout` a few lines up in the same file, and it is the behaviour we settled on in 4.11.x — changing it here would desync the branches for a decision that was already taken, which is the same trade you named for the commit message. Your reading of the fail-open is right on the merits; the counterweight is that refusing to boot turns a config typo into an outage, and that cost lands on the same operator. If we revisit it, it should be on all three branches at once rather than starting here.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
ytvnr: Agreed, and pinned in cc7b35dcbd — both ways you suggested, because they do different jobs. There is now a comment at the `return handleRejectedPath(...)` saying the order is the point and naming the cost (a traversal at an unknown context path, a 404 under RAW, is a 400 here), and a test `should_answer_the_same_whether_the_context_path_exists_or_not` in `With_reject_handling`. The test asserts 400 on `/nowhere/api/../../beta/api/echo`: `/nowhere` is deployed by nothing in the suite, so were resolution to move ahead of rejection there would be no acceptor and the answer would be 404. It also names its counterpart under NORMALIZE, `should_answer_not_found_when_the_resolved_path_matches_no_api`, so the two modes now answer each other in the file.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
ytvnr: Confirmed the defect is real: `NotFoundProcessor.java:66` still counts `message.length()` on the string while its neighbour now counts `body.length()` on the encoded buffer, so a localised `http.errors[404].message` carrying non-ASCII truncates. Agreed it stays out of this PR — it predates the change and fixing it here would put an unrelated 404 fix inside a security backport. I will raise a ticket rather than touch it.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
ytvnr: Leaving it. `application/grpc+proto` indeed falls through and the caller reads UNKNOWN, but the exact match is inherited from `NotFoundProcessor` and the two agreeing is worth more than one of them being subtly better. Noting it here so the limitation is on the record as chosen rather than overlooked, which is what you asked for.

pom.xml
ytvnr: Still the blocker, and the order you recorded is the one we will follow. Two updates since your review: `ci/circleci: Build backend` has since gone green, so the snapshot does resolve in CI; and `ci/circleci: Test gateway` went red on a flake — reproducing the CI command locally failed in a different module (`standalone-container`, forked VM terminating on `LoggableClientResponseWithRequestTimeoutTest`), and the same job is green on every other recent PR against 4.12.x. Not touching the pin until 5.2.0 is out.

ytvnr:
Thanks for reviewing the chain rather than the diff — the `pathInfo()` trace through `HttpConnector:691` and `DefaultProxyEndpoint:36` is the check that mattered, and it is the one a reader of the diff alone cannot make.

Where each thread landed, and the two points from this comment that have no line to sit on:

- **`pom.xml:58`** — unchanged and still the blocker. Your merge order is the one we follow. Two things moved since: `Build backend` has gone green, so the snapshot does resolve in CI; and `Test gateway` went red on a flake, not on this change (reproducing the CI command locally failed in `standalone-container`, a forked VM terminating on `LoggableClientResponseWithRequestTimeoutTest`, and that job is green on every other recent PR against 4.12.x).
- **`gravitee.yml:97`** — taken, in c9121b1a70, in both `gravitee.yml` and `helm/values.yaml`.
- **`RequestConfiguration:74`** — left as is. It mirrors `requestTimeout` in the same file and was settled in 4.11.x; if we revisit the fail-open it should be on all three branches at once.
- **`dispatch:187`** — taken both ways, in cc7b35dcbd: a comment stating the order is the point and naming the cost, and a test that fails if resolution ever moves ahead of rejection.
- **`RejectedPathProcessor:84` and `:75`** — both left, both now on the record as chosen.

**The JIRA footer.** Added to all sixteen commits. Worth saying that it is not yet settled here: of the last thirty commits on `4.12.x`, twenty-three carry no APIM reference at all. So this is us adopting a good practice rather than us complying with one, which is a better reason anyway.

While in there, the twelve cherry-picks also lost their `(cherry picked from commit ...)` lines. Backports on this repo do carry them, but theirs point at commits that exist once merged; ours pointed at #19166, still an open branch, so the references would have dangled the moment it is squashed. Removing them beat leaving a trail to nothing.

**`84efdf53 fix(gateway): harden path handling after review`** — left untouched, for the reason you gave yourself. It is a cherry-pick, and rewording it here would desync the two branches for a message that is only imprecise, not wrong. Same trade as `RequestConfiguration`, decided the same way.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -146,6 +154,10 @@ public DefaultHttpRequestDispatcher(
      */
     @Override
     public Completable dispatch(HttpServerRequest httpServerRequest, String serverId) {
+        // Before anything else, so that whatever the dispatcher does with this request — the scan,
```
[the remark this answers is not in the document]
ytvnr: Not doing it here. `dispatch` was already long before this PR, and the change adds about twenty lines to it. The same method is being ported to three branches (4.11.x, 4.12.x, master), so extracting helpers in this one would enlarge the diff of a security fix and leave the three copies structurally different from each other. That costs more than the readability it buys. Splitting the method is worth doing, on its own, on master.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
ytvnr: Ticket raised: https://gravitee.atlassian.net/browse/APIM-15007 — `404 response body is truncated when http.errors[404].message contains non-ASCII characters`. It records the reproduction (a 27-character, 28-byte message loses its trailing `.`), notes that the default message is ASCII so nothing shows out of the box, and points at the byte count this PR uses in `RejectedPathProcessor` as the fix pattern. Staying out of this branch as you said.

gravitee-apim-gateway/gravitee-apim-gateway-standalone/gravitee-apim-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
phiz71: Verified on `c9121b1a`: `gravitee.yml:99-102` and `helm/values.yaml:1452-1456`, both right after the `;x` paragraph, both pointing back at it so the two hazards read as the pair they are. That was the gap -- the block warned about the spelling a Servlet receiver strips and said nothing about the one an EL expression reads. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-env/src/main/java/io/gravitee/gateway/env/RequestConfiguration.java
phiz71: Taking the refusal, and the argument is the stronger one. The fallback mirroring `requestTimeout` a few lines up in the same file is the part that settles it for me: a lone exception to a local convention is its own kind of trap, and the decision was already made on 4.11.x. Reopening it here would desync three branches for a call that should be made once, across all of them.

My reading of the fail-open stands on the merits, and so does your counterweight -- refusing to boot turns a config typo into an outage, on the same operator. Noting for whoever does revisit it: the question is already settled the other way on master, where the fallback is `NORMALIZE`, so only 4.11.x and 4.12.x carry the fail-open. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
phiz71: Both checked. The comment at the `return` names the order, the reason and the cost, and `should_answer_the_same_whether_the_context_path_exists_or_not` pins the 400 on `/nowhere/api/../../beta/api/echo` while naming its NORMALIZE counterpart, so the two modes now answer each other in the file.

The part worth the credit is that you verified the test's premise instead of assuming it -- a probe under `RAW` to confirm the same path answers 404, then disabling the early reject to watch the test go red. A test asserting 400 on a path that would have answered 400 anyway pins nothing, and that is only visible if someone checks. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
phiz71: Checked the ticket rather than taking the reply for it: [APIM-15007](https://gravitee.atlassian.net/browse/APIM-15007) exists, raised yesterday, *To Do*, with the title given. Worth knowing it is typed **Public Bug**, so Exalate will mirror the title and description onto `gravitee-io/issues` -- nothing in either is sensitive, so that is fine, just not silent. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
phiz71: Which is exactly what I asked for -- the limitation chosen rather than inherited, and now on the record where the next reader meets it. Agreed that two rejection processors on the same path agreeing beats one of them being individually better. Resolving.

phiz71 (approved):
Approving.

Everything I raised is addressed, checked on `c9121b1a` rather than on the replies:

- The `uri()` limitation is documented where an operator decides, in both files (`gravitee.yml:99-102`, `helm/values.yaml:1452-1456`), next to the `;x` paragraph and pointing back at it.
- The rejection order is now stated and pinned -- a comment naming the cost, and `should_answer_the_same_whether_the_context_path_exists_or_not` asserting the 400 on `/nowhere/api/../../beta/api/echo`.
- [APIM-15007](https://gravitee.atlassian.net/browse/APIM-15007) exists and says what it claims to; I looked rather than assuming.
- The `application/grpc+proto` limitation is on the record as chosen.
- All 16 commits now carry `APIM-14917`.

On the `RAW` fallback I am taking your refusal, and the argument is the stronger one: mirroring `requestTimeout` in the same file beats a lone local exception, and the decision was already made on 4.11.x. It should be revisited across all three branches or not at all -- noting that master already answers it the other way.

The thing worth the credit is smaller and better than any of the above: you verified the new test's premise instead of assuming it -- probing under `RAW` to confirm the path really answers 404, then disabling the early reject to watch the test go red. A test asserting 400 where 400 would have come anyway pins nothing, and only checking reveals which one you have.

**The snapshot thread stays open on purpose.** It is the only one left, and it is doing a job: with conversation resolution required on this branch, an open thread holds the merge button while the approval sits ready behind it. So the pin cannot slip through, and you do not need anyone to come back once `gravitee-common` 5.2.0 is out -- change the line, resolve that thread, merge.

I checked the branch protection so the arrangement actually works as intended: `dismiss_stale_reviews` and `require_last_push_approval` are both off here, so this approval survives the push that moves the pin.

ytvnr:
**Manual test run locally**

Ran the manual scenarios against this branch on a local gateway, on a v4 HTTP proxy API whose flow carries a `mock` policy printing `{#request.path}`, with `curl --path-as-is` on every request. Without `--path-as-is` curl resolves the dot segments client side and the gateway receives an already canonical request line, so every mode looks identical.

| Request path sent | `RAW` (default here) | `NORMALIZE` | `REJECT` |
| --- | --- | --- | --- |
| `/proxyv4/orders/a/../echo` | `/proxyv4/orders/a/../echo` | `/proxyv4/orders/echo` | 400 |
| `/proxyv4/orders/..;/echo` | as received | `/proxyv4/echo` | 400 |
| `/proxyv4/orders/..;jsessionid=1/echo` | as received | `/proxyv4/echo` | 400 |
| `/proxyv4/orders/%2e%2e/echo` | as received | `/proxyv4/echo` | 400 |
| `/proxyv4/orders/%41%42` | `/proxyv4/orders/%41%42` | `/proxyv4/orders/AB` | 400 |

The `..;` and `..;jsessionid=1` rows are the parser differential the fix targets: strictly per RFC 3986 those are ordinary segments, while Tomcat, Jetty and Spring strip `;params` first and see a dot segment.

The bypass itself, with no credential, against a second API carrying an API key plan:

`/proxyv4/orders/../../proxyv4/guarded/x` answers **200** under `RAW` (authorized against the keyless API), **401** under `NORMALIZE`, **400** under `REJECT`.

Also checked:
- `{#request.uri}` still reports the path as received in every mode, including `NORMALIZE`. This is the documented behaviour and what lets an operator correlate with upstream logs.
- The startup log line reports the mode the gateway actually came up in.

No deviation from the expected behaviour.

pom.xml
ytvnr: Unblocked. `gravitee-common` 5.2.0 was released this morning (gravitee-io/gravitee-common#170) and `pom.xml:58` now pins `5.2.0`. The bump went into the commit that introduced it, `refactor(gateway): consume RequestPathNormalizer from gravitee-common`, rather than a follow-up, so no trace of the snapshot is left anywhere in the branch history.

Worth recording: the released `RequestPathNormalizer` is byte-for-byte identical to the snapshot this PR was reviewed against, so nothing about the behaviour you assessed has moved. Built the gateway chain against the real 5.2.0 and it is green, with one exception unrelated to this change: `DebugHttpProtocolVerticleTest.http_server_should_listen` failed once with `Vertx Pool closed` and passed on a rerun of identical code. That test is untouched by this PR and identical to `origin/4.12.x`, and it is in the module that produced the `Test gateway` red you saw earlier, so the two are almost certainly the same flake.

PR #19388, opened 2026-08-25
feat(gateway)!: normalize request paths by default

ytvnr:
## Issue

## Description

Ports the path-normalization fix from 4.11.x ([#19166](https://github.com/gravitee-io/gravitee-api-management/pull/19166)) to master, and additionally makes `NORMALIZE` the default.

### The bypass

A parser differential. The gateway reads `..;jsessionid=1` as an ordinary segment per RFC 3986 §5.2.4, while Tomcat, Jetty and Spring strip `;params` first and see a dot segment. A request to `/alpha/api/../../beta/api/echo` is therefore matched against `alpha`, whose plan is enforced, then forwarded unresolved to a backend that serves `beta` — reaching an API behind an api-key plan without a key.

Resolving dot segments **before** the listener context path is resolved closes it: the request is routed, authorized and forwarded on the same path.

### What this adds

`http.pathHandling` takes `RAW`, `REJECT` or `NORMALIZE`. Rejected requests are reported under `handlers.rejected.analytics.enabled` (default `true`), independently of the `notfound` flag. Debug sessions report the executed path and close cleanly on a rejection. Exposed in the Helm chart.

The normalizer itself is **not** vendored here: it is `io.gravitee.common.http.RequestPathNormalizer`, from gravitee-common ([#170](https://github.com/gravitee-io/gravitee-common/pull/170)).

## Breaking change

`NORMALIZE` becomes the default in 4.13. It was `RAW` up to 4.12.

A gateway that carries no `http.pathHandling` setting changes how it routes: a path containing dot segments now reaches the API the **resolved** path designates, not the one the raw path did. Depending on the deployment, such a request may now answer `401` (the resolved API's plan applies) or `404` (the resolved path matches no API) where it previously answered `200`.

Set `pathHandling: RAW` explicitly to keep the previous behaviour.

An unreadable value falls back to `NORMALIZE`, not to `RAW` — a typo in a configuration file must not be a way to reopen the bypass. The warning is still emitted, and the active mode is logged at startup in every mode.

## Known limitation

Segment parameters on an **ordinary** segment are left alone, on purpose. `/a/admin;x/b` is canonical here and forwarded as sent, while a Servlet receiver strips the `;x` and serves `/a/admin/b`. An allow or deny rule written against `/admin/**` still needs its own guard against that spelling. Documented in `gravitee.yml` and `helm/values.yaml`.

## Additional context

- `gravitee-common` [5.2.0](https://github.com/gravitee-io/gravitee-common/releases/tag/5.2.0), carrying `RequestPathNormalizer` ([#170](https://github.com/gravitee-io/gravitee-common/pull/170))
- Docs: [gravitee-platform-docs#2249](https://github.com/gravitee-io/gravitee-platform-docs/pull/2249)
- 4.12.x port: same change, default stays `RAW`
- `PathTraversalV4IntegrationTest` and `PathTraversalOrganizationGuardV4IntegrationTest` were written when `RAW` was the default and now set it explicitly, so they go on describing the mode they document. The closed behaviour is covered by `PathHandlingV4IntegrationTest`.
- Manual testing needs `curl --path-as-is` and a `START_WITH` flow selector on the test API — without either you get silent false negatives.


pom.xml
```diff
@@ -55,7 +55,7 @@
     <gravitee-cockpit-api.version>3.12.0</gravitee-cockpit-api.version>
     <gravitee-cloud-initializer.version>2.3.1</gravitee-cloud-initializer.version>
     <logback-ecs-encoder.version>1.7.0</logback-ecs-encoder.version>
-    <gravitee-common.version>5.1.0</gravitee-common.version>
+    <gravitee-common.version>5.2.0-feat-request-path-normalizer-SNAPSHOT</gravitee-common.version>
```
phiz71: **MUST** -- the snapshot pin, as you flag in the description. Same ordering as on #19387:

1. merge gravitee-io/gravitee-common#170 -- approved, and the only thing still open there is a missing JIRA footer on its commit;
2. wait for the release, which will be **`5.2.0`**, not `5.1.0`: that tag already exists from 2026-08-12 and a `feat` commit takes the minor;
3. pin `5.2.0` here.

`ci/circleci: Build backend` was still running when I reviewed, so I have not seen the snapshot resolve in CI on either PR.

gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-gateway/src/main/resources/config/gravitee.yml
```diff
@@ -84,6 +84,22 @@ secrets:
...
+#    RAW       the path is used, and forwarded, exactly as received.
+#    REJECT    answers 400 when the normalized path differs from the one received.
+#    NORMALIZE (default) resolves the dot segments and routes, enforces and forwards on the
+#              result, so /alpha/api/../../beta/api is read as the call to beta that it is.
+#  Beyond dot segments, both REJECT and NORMALIZE also decode percent-encoded unreserved
+#  characters, merge duplicate slashes, and answer 400 on a malformed percent sequence.
+#  Encoded slashes are never decoded, and segment parameters on an ordinary segment are left
+#  alone: /a/admin;x/b is canonical here and forwarded as sent, while a Servlet receiver strips
+#  the ;x and serves /a/admin/b. An allow or deny rule written against /admin/** therefore still
+#  needs its own guard against that spelling.
```
phiz71, line 98: **SHOULD** -- this paragraph documents one limitation of `NORMALIZE` carefully and leaves out its twin. On 4.12.x that gap only reaches operators who opt in; here `NORMALIZE` is the default, so it reaches everybody.

`uri()` keeps reporting the bytes as received in every mode. `EvaluableRequest.getUri()` reads `request.uri()`, so a flow condition or a policy written on `{#request.uri}` still sees `/alpha/api/../../beta/api/echo` while routing, plan enforcement and the upstream call all work on `/beta/api/echo`. It is deliberate -- request signing depends on it, and `should_still_expose_the_uri_as_received` pins it -- but it is evaded exactly the way the `;x` case above is.

An operator reading this block learns to keep a guard against `/a/admin;x/b` and learns nothing about `{#request.uri}`. Two lines here, and the same in `helm/values.yaml` around line 1453.

gravitee-apim-gateway/gravitee-apim-gateway-env/src/main/java/io/gravitee/gateway/env/RequestConfiguration.java
```diff
@@ -51,4 +53,41 @@ public RequestClientAuthConfiguration httpRequestClientAuthConfiguration(
     ) {
         return new RequestClientAuthConfiguration(headerName);
     }
+
+    @Bean
+    public RequestPathConfiguration httpRequestPathConfiguration(@Value("${http.pathHandling:NORMALIZE}") String pathHandling) {
```
phiz71, line 58: **SHOULD** -- this line is the breaking change, and the commit that carries it says nothing.

`18a1c043 feat(gateway)!: normalize request paths by default` is 50 characters and an empty body: no `BREAKING CHANGE:` footer, nothing about what breaks. This repository has no semantic-release configuration -- releases are driven by hand -- so the `!` feeds no tool. It is a signal to a human, and it is on its own.

Everything that explains the change lives in the PR description, which does not survive the merge. Someone running `git log` on master a year from now sees the subject and has to find this PR to learn that a request carrying dot segments may now answer 401 or 404 where it answered 200, and that `pathHandling: RAW` is the way back. That belongs in the commit body, as a `BREAKING CHANGE:` footer.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -155,13 +168,54 @@ public Completable dispatch(HttpServerRequest httpServerRequest, String serverId
...
+
+        // One scan answers what both active modes ask. REJECT needs nothing more: a path that is
+        // not in the form it claims to be is refused without ever being rewritten. NORMALIZE only
+        // pays for the resolution when there is something to resolve.
+        final boolean needsNormalization = requestPathConfiguration.isEnabled() && RequestPathNormalizer.needsNormalization(rawPath);
+
+        final String normalizedPath;
+        if (needsNormalization) {
+            if (requestPathConfiguration.getHandling() == RequestPathHandling.REJECT) {
+                return handleRejectedPath(httpServerRequest, serverId, receivedAt);
```
phiz71, line 188: **COULD** -- same as on #19387: rejection happens before the acceptor is resolved, so under `REJECT` an unknown host or context path answers 400 where it used to answer 404. Probably deliberate and right -- resolving first would turn the gateway into an enumeration oracle -- but neither asserted nor stated anywhere, so it currently reads as a side effect rather than the choice it likely is.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
```diff
@@ -0,0 +1,93 @@
...
+            if (MediaType.MEDIA_APPLICATION_GRPC.equals(mediaType)) {
+                ctx.response().headers().set("grpc-status", GrpcStatus.INVALID_ARGUMENT.toString());
+                ctx.response().headers().set("grpc-message", message);
+            }
+
+            final Buffer body = Buffer.buffer(message);
+            // Counted on the encoded body, not on the string: a localised http.errors[400].message
+            // holding any non-ASCII character has more bytes than it has characters, and a short
+            // Content-Length truncates the response.
+            ctx.response().headers().set(HttpHeaderNames.CONTENT_LENGTH, Integer.toString(body.length()));
```
phiz71, line 84: **COULD** -- correct here, and it makes the neighbour visibly wrong: `NotFoundProcessor.java:66` still measures the string, so a localised `http.errors[404].message` with any non-ASCII character truncates the 404 body. Out of scope for this PR and it should stay out -- worth a ticket.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
```diff
@@ -0,0 +1,93 @@
...
+
+            ctx.response().status(HttpStatusCode.BAD_REQUEST_400);
+
+            final String message = environment.getProperty("http.errors[400].message", DEFAULT_MESSAGE);
+
+            // A gRPC caller reads the status from the trailers, not from the HTTP code: without
+            // these it sees UNKNOWN and has no idea why it was refused. Same treatment as an
+            // unmatched context path, which is the closest thing this rejection resembles.
+            final MediaType mediaType = MediaType.parseMediaType(ctx.request().headers().get(HttpHeaderNames.CONTENT_TYPE));
+            if (MediaType.MEDIA_APPLICATION_GRPC.equals(mediaType)) {
```
phiz71, line 75: **COULD** -- matches `application/grpc` exactly, so a caller sending `application/grpc+proto` gets the 400 without `grpc-status` and reads `UNKNOWN`. Copied from `NotFoundProcessor`, so the two agree, which matters more. Naming it so the limitation is chosen rather than inherited.

phiz71 (commented):
Reviewed alongside the 4.12.x port (#19387), so this focuses on what the default change brings rather than restating the mechanism -- which I checked there and which holds end to end: decision and forward both run on the resolved path, and no entrypoint is left out.

**Three things the default change does better than its 4.12.x sibling**, worth saying because they are the parts that could easily have been got wrong.

An unreadable `http.pathHandling` falls back to `NORMALIZE`, not to `RAW`, and the test says why in one line: a typo in a configuration file must not be a way to reopen the bypass. On 4.12.x I raised the mirror image of this as a finding; here it is already the right way round, and it makes the fail-fast argument moot on this branch since the fallback *is* the safe mode.

The default is covered end to end, not only in a unit test. `With_no_handling_configured` deliberately does not override `configureGateway` and asserts the 401; `With_an_unreadable_handling_value` sets `NORMALISE` -- the British spelling, which is the typo somebody actually makes -- and asserts it too. Plus the unit test pinning the `${http.pathHandling:NORMALIZE}` literal, whose own comment is honest about being the only assertion covering the bean's side. That was the gap I went looking for, and it is closed.

The operator-facing documentation carries the breaking change in both files a person reads, with the version on it (`default since 4.13`, `was RAW up to 4.12`), the consequence for routing, and the way back. `helm/values.yaml` even adds something 4.12.x does not say: the setting is gateway-wide, not per server or per API.

Findings inline. One that has no line to sit on: **none of the 15 commits carries the JIRA footer**, though the PR references APIM-14917.

Verdict: sound on the merits, blocked only by the snapshot pin. The default change is well conducted; findings 2 and 3 are two paragraphs of documentation and one commit body.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-debug/src/main/java/io/gravitee/gateway/debug/vertx/VertxHttpServerRequestDebugDecorator.java
```diff
@@ -18,13 +18,26 @@
...
+     * <p>That is not hypothetical: the path is resolved before the listener is chosen, and until the
+     * options below were passed on, a debug session running in V3 execution mode reported the path
+     * as received while the request it described had been routed on the resolved one.
+     */
     public VertxHttpServerRequestDebugDecorator(VertxHttpServerRequest delegate, IdGenerator idGenerator) {
-        super(delegate.getNativeServerRequest(), idGenerator);
+        super(
+            delegate.getNativeServerRequest(),
+            idGenerator,
+            VertxHttpServerRequestOptions.builder().path(delegate.path()).timestamp(delegate.timestamp()).build()
```
ytvnr, line 39: Not applying this one: there are two distinct `VertxHttpServerRequestOptions` in the tree, and the suggestion targets the wrong one.

- `io.gravitee.gateway.http.vertx.VertxHttpServerRequestOptions` — the Lombok `@Builder` class, carrying `path` and `timestamp` only. This is the one imported here.
- `io.gravitee.gateway.reactive.http.vertx.VertxHttpServerRequest.VertxHttpServerRequestOptions` — a nested record `(clientAuthHeaderName, path, timestamp)`, used by the reactive stack.

This decorator extends `io.gravitee.gateway.http.vertx.VertxHttpServerRequest`, which contains no occurrence of `clientAuthHeaderName` at all. Nothing is being dropped, and the suggested `.clientAuthHeaderName(delegate.clientAuthHeaderName())` would not compile: neither the builder nor the delegate has that method.

The "carry dispatcher decisions across" intent is satisfied: `path` and `timestamp` are the only two decisions the dispatcher makes about the wrapper on this stack.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-debug/src/main/java/io/gravitee/gateway/reactive/debug/reactor/DebugHttpRequestDispatcher.java
```diff
@@ -17,32 +17,52 @@
...
+import io.gravitee.gateway.debug.definition.ReactableDebugApi;
+import io.gravitee.gateway.debug.handlers.api.DebugApiReactorHandler;
 import io.gravitee.gateway.debug.vertx.VertxHttpServerRequestDebugDecorator;
 import io.gravitee.gateway.env.GatewayConfiguration;
 import io.gravitee.gateway.env.RequestClientAuthConfiguration;
+import io.gravitee.gateway.env.RequestPathConfiguration;
 import io.gravitee.gateway.env.RequestTimeoutConfiguration;
+import io.gravitee.gateway.http.vertx.VertxHttpServerRequestOptions;
 import io.gravitee.gateway.opentelemetry.TracingContext;
+import io.gravitee.gateway.reactive.api.ExecutionPhase;
```
ytvnr: Correct, removed. The import came in with the ported commit and was never referenced; `javac` does not fail on it and prettier does not strip it, so it survived. Fixed in `refactor(gateway): drop an unused import in the debug dispatcher`.

gravitee-apim-gateway/gravitee-apim-gateway-env/src/main/java/io/gravitee/gateway/env/RequestPathHandling.java
```diff
@@ -0,0 +1,50 @@
...
+ * Left untouched, a request can be authorized as one API and, once a conforming upstream applies
+ * RFC 3986 §5.2.4, delivered to another one's backend.
+ *
+ * @author GraviteeSource Team
+ */
+public enum RequestPathHandling {
+    /**
+     * Byte-preserving pass-through: the path is used, and forwarded, exactly as it arrived. This is
+     * the historical behaviour and remains the default, so that request signing and encoded slashes
+     * keep working without any change.
```
ytvnr: Correct, and the most useful finding on this PR — thank you. This was a fifth place carrying the old default that I had missed: the enum javadoc contradicted the very change it documents.

Fixed in `docs(gateway): document the uri limitation and the flipped default where operators read them`, on both constants rather than only `RAW`:

- `RAW` — "This **was the default up to 4.12** and is what a deployment sets explicitly to keep that behaviour"
- `NORMALIZE` — "**The default since 4.13.** Resolves the dot segments and uses the result for listener resolution, plan enforcement, flow selection and the upstream URI."

Note it is 4.13, not 25.1.0: `25.1.0` at `pom.xml:27` is the `gravitee-parent` version, while APIM master is `4.13.0-SNAPSHOT` via `${revision}`.

pom.xml
ytvnr: Agreed, and it stays blocking. Your ordering is the one recorded in the brief, including `5.2.0` rather than `5.1.0`. Nothing here is merged until gravitee-common 5.2.0 is out and this line points at it — the PR description carries a "Do not merge as is" section for exactly that.

gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-gateway/src/main/resources/config/gravitee.yml
ytvnr: Good catch, and the asymmetry was real: the block warned carefully about `;x` on an ordinary segment and said nothing about the twin. Your point about the blast radius is what settles it — on 4.12.x the gap only reaches those who opt in, here it reaches deployments that configured nothing at all.

Added to both files. In `gravitee.yml`:

```
#  Under NORMALIZE the request exposes two forms of the path to policies and flow conditions:
#  {#request.path} and {#request.pathInfo} carry the resolved path, {#request.uri} still carries
#  the bytes as received, because request signing depends on them. Write security rules on
#  {#request.path}: a condition on {#request.uri} is evaded by the very traversal being resolved.
```

and the same in `helm/values.yaml`. The concrete failure it now warns about: a flow condition `{#request.uri.startsWith('/admin')}` does not fire on `/public/api/../../admin/api/secret`, while the gateway routes, authorizes and forwards on `/admin/api/secret`. Written on `{#request.path}` it fires.

gravitee-apim-gateway/gravitee-apim-gateway-env/src/main/java/io/gravitee/gateway/env/RequestConfiguration.java
ytvnr: You are right, and the `!` alone was not enough. Added a `BREAKING CHANGE:` footer to the flip commit:

```
feat(gateway)!: normalize request paths by default

BREAKING CHANGE: http.pathHandling now defaults to NORMALIZE; it was RAW up to 4.12.
A request whose path carries dot segments is routed, authorized and forwarded on the
resolved path, so it may answer 401 or 404 where it answered 200. Set pathHandling: RAW
to keep the previous behaviour.

APIM-14917
```

On the JIRA footer raised in your summary: strictly it is not a convention here — 1 of the last 40 commits on master carries an APIM reference in its body — but it is good practice and it costs nothing, so all 19 commits now carry `APIM-14917`. The `(cherry picked from commit ...)` provenance lines are gone at the same time: the 4.11.x sha means nothing to someone reading master, and the PR already says where this comes from.

This required rewriting the branch, so the force-push has cost you the incremental view since your review. Sorry about that. The trees were diffed before and after the rewrite and are identical.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
ytvnr: Deliberate, and you read the intent correctly: the request is validated first, and only then does the gateway decide what to do with it. Reversing the two would let a caller sending a dot segment tell a deployed context path from an unknown one by the status code alone.

You were right that nothing said so. Now two things do.

A comment at the `return`, and a test — `With_reject_handling.should_reject_before_looking_for_an_acceptor`:

```java
assertStatus(httpClient, "/nowhere/api/../../beta/api/echo", 400);
assertThat(wiremock.findAll(getRequestedFor(anyUrl()))).isEmpty();
```

No API in the suite is deployed on `/nowhere`, so resolution-first would find no acceptor and answer 404. I checked that premise rather than assuming it: a temporary probe under `RAW` confirmed the same path answers 404 when used as received. And I checked the test actually bites, by disabling the early reject — it goes red. Worth noting it goes red at 401 rather than 404, because neutralising the reject also moves normalization ahead; the ordering is still what it pins.

Its NORMALIZE counterpart, `should_answer_not_found_when_the_resolved_path_matches_no_api`, answers 404. The two modes now answer each other.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
ytvnr: Agreed on both counts: real, and out of scope here. `NotFoundProcessor` measures the string rather than the encoded bytes, so a localised `http.errors[404].message` with any non-ASCII character gets a `Content-Length` shorter than the body and is truncated. Not touching it in this PR — it is a pre-existing defect on a neighbouring path and it deserves its own change with its own test. Raising a ticket.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
ytvnr: Fair, and leaving it as is. `application/grpc+proto` gets the 400 without `grpc-status` and reads `UNKNOWN` — but this matches `NotFoundProcessor` exactly, and two rejection processors on the same path agreeing matters more than one of them being individually more correct. Fixing it in one place would create a divergence a reader has to explain. If it is worth fixing it is worth fixing for both, in a change that says so.

ytvnr:
## Review addressed

Thanks both — replies are inline on each thread. Summary here because the branch was rewritten to add the commit footers, so the incremental diff since the review is gone. The trees were compared before and after the rewrite and are identical; only messages changed.

### Applied

| Finding | Commit |
|---|---|
| `RequestPathHandling` javadoc still advertised `RAW` as the default | `docs(gateway): document the uri limitation and the flipped default where operators read them` |
| `{#request.uri}` limitation undocumented under the new default | same commit, `gravitee.yml` + `helm/values.yaml` |
| Reject-before-resolve ordering unstated | `refactor(gateway): name the reject-before-resolve ordering as a choice` + `test(gateway): pin that a rejected path is refused before any acceptor is resolved` |
| Unused `ExecutionPhase` import | `refactor(gateway): drop an unused import in the debug dispatcher` |
| No `BREAKING CHANGE:` footer | flip commit now carries one |
| No ticket reference on the commits | all 19 carry `APIM-14917`; the `(cherry picked from commit ...)` lines are gone |

The enum javadoc was the best catch of the review: a fifth place carrying the old default, contradicting the very change this PR makes. It would have shipped.

### Not applied

**`clientAuthHeaderName` dropped in `VertxHttpServerRequestDebugDecorator`** — the finding conflates two same-named types. The decorator uses `io.gravitee.gateway.http.vertx.VertxHttpServerRequestOptions` (`path` + `timestamp`), not the reactive nested record that carries `clientAuthHeaderName`. The V3 `VertxHttpServerRequest` it extends has no such field, and the suggested patch would not compile. Detail on the thread.

**`application/grpc` exact match** — real, but it matches `NotFoundProcessor`. Two rejection processors agreeing is worth more than one being individually more correct; if it is fixed it should be fixed for both.

### Out of scope, ticket to follow

`NotFoundProcessor:66` measures the string rather than the encoded bytes for `Content-Length`, so a localised 404 message with any non-ASCII character is truncated. Pre-existing, neighbouring path, deserves its own change and its own test.

### On the new ordering test

`With_reject_handling.should_reject_before_looking_for_an_acceptor` asserts 400 on `/nowhere/api/../../beta/api/echo`. Two checks beyond it being green:

- **the premise** — a temporary probe under `RAW` confirmed the same path answers 404 when used as received, which is what makes the 400 meaningful;
- **the detection power** — disabling the early reject makes it go red. Worth recording that it goes red at 401 rather than 404, because neutralising the reject also moves normalization ahead. The ordering is still what it pins.

### Verification

- gateway unit tests: 647 green across the four impacted modules
- `gravitee-apim-distribution-integration-tests`, path traversal: 50 green
- helm chart: 4 green

One flake unrelated to this change: `DebugHttpProtocolVerticleTest.http_server_should_listen` intermittently fails with `Vertx Pool closed`. It is not in this diff and passes 3/3 in isolation.

### Still blocking

`gravitee-common` is pinned to `5.2.0-feat-request-path-normalizer-SNAPSHOT`. Not merging until 5.2.0 is released and this line points at it.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
ytvnr: Ticket raised: **APIM-15007**.

It carries the comparison that makes the fix obvious — `RejectedPathProcessor` builds `Buffer.buffer(message)` and counts `body.length()` (bytes), while `NotFoundProcessor:66` counts `message.length()` (characters) — plus the note that no test covers the 404 `Content-Length` today, which is exactly why it went unnoticed. Pre-existing on 4.11.x and 4.12.x as well.

gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-gateway/src/main/resources/config/gravitee.yml
phiz71: Verified on `dd0556c4`: `gravitee.yml:99-102` and `helm/values.yaml:1454-1457`. The wording here is better than what I asked for -- naming `{#request.path}` and `{#request.pathInfo}` as the pair that carries the resolved path, and telling the reader which one to write security rules on, rather than only warning them off `{#request.uri}`. On a branch where `NORMALIZE` is what everyone gets, that difference matters. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-env/src/main/java/io/gravitee/gateway/env/RequestConfiguration.java
phiz71: The footer is in, and it carries what the `!` could not: what changes, that it may answer 401 or 404 where it answered 200, and `pathHandling: RAW` as the way back. That now survives the merge, which was the whole point.

I checked the rest of the branch too: all 19 commits carry `APIM-14917`, and the `(cherry picked from ...)` lines are gone -- a 4.11.x sha means nothing to someone reading master. Thanks for measuring the local convention before following mine rather than just deferring to it. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
phiz71: Both checked. The comment at the `return` states the order and why reversing it would let a caller enumerate deployed context paths by status code, and `should_reject_before_looking_for_an_acceptor` pins it, naming its NORMALIZE counterpart. The name is the better of the two branches' -- it says what is being pinned rather than what is observed.

And you verified the premise rather than assuming it, which is what makes the test worth having: the `RAW` probe confirming the 404, and disabling the early reject to watch it go red. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
phiz71: Checked the ticket rather than taking the reply for it: [APIM-15007](https://gravitee.atlassian.net/browse/APIM-15007) exists, raised yesterday, *To Do*, with the title given. Worth knowing it is typed **Public Bug**, so Exalate will mirror the title and description onto `gravitee-io/issues` -- nothing in either is sensitive, so that is fine, just not silent. Resolving.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/processor/rejected/RejectedPathProcessor.java
phiz71: Which is exactly what I asked for -- the limitation chosen rather than inherited, and now on the record where the next reader meets it. Agreed that two rejection processors on the same path agreeing beats one of them being individually better. Resolving.

phiz71 (approved):
Approving.

Everything I raised is addressed, checked on `dd0556c4` rather than on the replies:

- The `uri()` limitation is documented in both files, and better than I asked for: it names `{#request.path}` and `{#request.pathInfo}` as the pair carrying the resolved path and tells the reader which one to write security rules on, instead of only warning them off `{#request.uri}`. On the branch where `NORMALIZE` is what everybody gets, that difference is the one that matters.
- The `BREAKING CHANGE:` footer carries what the `!` could not -- what changes, the 401/404 where there was a 200, and `pathHandling: RAW` as the way back -- and it survives the merge, which was the point.
- The rejection order is stated and pinned by `should_reject_before_looking_for_an_acceptor`, whose name says what is being pinned rather than what is observed.
- [APIM-15007](https://gravitee.atlassian.net/browse/APIM-15007) exists and says what it claims to; I looked rather than assuming.
- All 19 commits carry `APIM-14917`, and the `(cherry picked from ...)` lines are gone -- a 4.11.x sha means nothing to someone reading master.

Two things I would single out. You measured the local convention before adopting mine -- 1 of the last 40 commits on master carries an APIM reference -- rather than just deferring; and you verified the new test's premise instead of assuming it, probing under `RAW` for the 404 and then disabling the early reject to watch the test go red.

**The snapshot thread stays open on purpose.** It is the only one left, and it is doing a job: with conversation resolution required on this branch, an open thread holds the merge button while the approval sits ready behind it. So the pin cannot slip through, and you do not need anyone to come back once `gravitee-common` 5.2.0 is out -- change the line, resolve that thread, merge.

I checked the branch protection so the arrangement actually works as intended: `dismiss_stale_reviews` and `require_last_push_approval` are both off here, so this approval survives the push that moves the pin.

ytvnr:
**Manual test run locally**

Ran the manual scenarios against this branch on a local gateway, on a v4 HTTP proxy API whose flow carries a `mock` policy printing `{#request.path}`, with `curl --path-as-is` on every request. Without `--path-as-is` curl resolves the dot segments client side and the gateway receives an already canonical request line, so every mode looks identical.

| Request path sent | `NORMALIZE` (default here) | `RAW` | `REJECT` |
| --- | --- | --- | --- |
| `/proxyv4/orders/a/../echo` | `/proxyv4/orders/echo` | `/proxyv4/orders/a/../echo` | 400 |
| `/proxyv4/orders/..;/echo` | `/proxyv4/echo` | as received | 400 |
| `/proxyv4/orders/..;jsessionid=1/echo` | `/proxyv4/echo` | as received | 400 |
| `/proxyv4/orders/%2e%2e/echo` | `/proxyv4/echo` | as received | 400 |
| `/proxyv4/orders/%41%42` | `/proxyv4/orders/AB` | `/proxyv4/orders/%41%42` | 400 |

The `..;` and `..;jsessionid=1` rows are the parser differential the fix targets: strictly per RFC 3986 those are ordinary segments, while Tomcat, Jetty and Spring strip `;params` first and see a dot segment.

The bypass itself, with no credential, against a second API carrying an API key plan:

`/proxyv4/orders/../../proxyv4/guarded/x` answers **401** under the new default, **200** under `RAW` (authorized against the keyless API, which is the bug), **400** under `REJECT`.

Specific to this branch:
- With no `http.pathHandling` set in the configuration, the gateway comes up in `NORMALIZE` and the startup log line confirms it. The whole first column above is what an operator gets out of the box after this change.
- A mistyped value falls back to `NORMALIZE`, not to `RAW`, with the warning logged and the startup line reporting the effective mode.

Also checked:
- `{#request.uri}` still reports the path as received in every mode, including `NORMALIZE`. This is the documented behaviour and what lets an operator correlate with upstream logs.
- Debug mode on a v2 API: the policy steps still appear, under both `RAW` and `NORMALIZE`.

No deviation from the expected behaviour.

pom.xml
ytvnr: Unblocked: `gravitee-common` **5.2.0** was released this morning and the pin now points at it, following exactly the ordering you set out.

The bump is folded into `refactor(gateway): consume RequestPathNormalizer from gravitee-common` rather than sitting in a commit of its own, so no commit on this branch ever points at the snapshot. Verified across the series: `pom.xml` reads `5.1.0` up to that commit and `5.2.0` from there on, and the snapshot string appears nowhere on the branch.

Green against the release: 647 unit tests across the four impacted gateway modules, 50 integration tests on the path-traversal suite. The "Do not merge as is" section is gone from the description.

One note in case you hit it too: the first resolution attempt failed with a 409 from the *snapshots* repo, which aborts Maven before it reaches the releases repo. It was a cached negative marker from an attempt made before the artifact had synced — clearing `~/.m2/repository/io/gravitee/common/gravitee-common/5.2.0/*.lastUpdated` and retrying resolved it.

PR #19394, opened 2026-08-25
fix(ci): accept only a hotfix version on a hotfix branch

phiz71:
## Issue

https://gravitee.atlassian.net/browse/BX-347

## Description

The guard added in [#19351](https://github.com/gravitee-io/gravitee-api-management/pull/19351) checks that the version and the branch agree before a release runs. On a hotfix branch it only tested the `<tag>-` prefix, so **any** qualifier passed. Probed against the generator on `master`:

```
4.2.0-alpha.1       on hotfix/4.2.0   →  accepted
4.2.0-milestone.3   on hotfix/4.2.0   →  accepted
4.2.0-hotfix        on hotfix/4.2.0   →  accepted
```

Releasing `4.2.0-alpha.1` from `hotfix/4.2.0` would tag `4.2.0-alpha.1` while the branch's poms carry `4.2.0-hotfix.1` — the tag says one thing, the artefacts another. That is the exact mismatch the guard exists to prevent, and it survived on the one branch the guard was written for.

Reachable through `yarn release --version=4.2.0-alpha.1 --branch=hotfix/4.2.0`: the client-side check does not object either, since the version carries no hotfix qualifier.

The condition now requires the version to carry a hotfix qualifier *and* its base to equal the branch's released version, rather than testing the prefix alone. The error message names the shape it expects.

## Additional context

Raised by a review comment on the 4.10.x backport ([#19389](https://github.com/gravitee-io/gravitee-api-management/pull/19389)). It was not fixed there on purpose: a backport has to stay a faithful copy of what merged into `master`, or the support branches drift apart. This is the `master` fix, and it is backportable — unlike [#19393](https://github.com/gravitee-io/gravitee-api-management/pull/19393), which regenerated a fixture that only `master` needed.

**How to verify**: `npm ci && npx jest` in `.circleci/ci` — 182 tests. Five cases were added to the throw table; restoring the old condition turns three of them red, which are the three the change is about. The other two only assert the new message.

**Security-sensitive**: this changes CI/CD configuration. It publishes nothing on its own; it makes an existing release guard stricter.

### DIFF
--- a/.circleci/ci/src/pipelines/pipeline-release.ts
+++ b/.circleci/ci/src/pipelines/pipeline-release.ts
@@ -26,13 +26,15 @@ const HOTFIX_QUALIFIER = /-hotfix\.\d+$/;
  * brings them together: the published version comes from the poms of whichever branch runs, while
  * the tag comes from the version. A mismatch therefore tags one version and publishes another,
  * without failing. A hotfix is where the two diverge most easily, since `hotfix/4.12.17` and
- * `4.12.17-hotfix.1` name the same release twice.
+ * `4.12.17-hotfix.1` name the same release twice. A hotfix branch releases hotfix versions of the
+ * tag it was cut from, and nothing else: any other qualifier on it would tag a version its poms do
+ * not carry.
  */
 function assertVersionMatchesBranch(version: string, branch: string): void {
   if (isHotfixBranch(branch)) {
     const releasedVersion = branch.substring('hotfix/'.length);
-    if (!version.startsWith(`${releasedVersion}-`)) {
-      throw new Error(`${branch} releases ${releasedVersion} with a qualifier, not ${version}`);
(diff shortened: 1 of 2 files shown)


phiz71:
Closing without merging — the subject is covered differently in [#19395](https://github.com/gravitee-io/gravitee-api-management/pull/19395).

The finding was real and stands: on a hotfix branch the check tested only the `<tag>-` prefix, so `4.12.17-alpha.1` passed. But that check lives in `generateReleaseConfig`, the pipeline behind `yarn release`, and #19395 removes that command and its pipeline along with the six others that duplicated `full_release`.

What replaces it is broader and fires earlier. Before triggering anything, `full_release` and `maven_release` now read the root `pom.xml` of the branch they are about to release from and refuse when it does not carry the version passed on the command line. That catches every mismatch rather than the hotfix shape alone, and it fails in the terminal instead of after a pipeline has started.

PR #19395, opened 2026-08-25
refactor(release): remove the per-step release path

phiz71:
## Issue

https://gravitee.atlassian.net/browse/BX-348

## Description

Eight ways existed to start a release: `full_release`, which triggers the whole pipeline, and seven commands that each triggered one step of it. Each of the seven had a CircleCI pipeline of its own behind it.

Two commits.

**1. All of it is removed** — the commands and the pipelines both. 61 files, 8448 lines deleted, 11 added.

**2. A check replaces part of what went with it.** Before triggering anything, `full_release` and `maven_release` read the root `pom.xml` of the branch they are about to release from and refuse when it does not carry the version passed on the command line. The published version comes from the poms; `--version` only names the tag. Without that comparison the two can disagree and nothing downstream notices.

```
kept       full_release · prepare_hotfix · maven_release
removed    release · package_zip · docker · rpms · helm_chart · release_notes · nexus_sync
           and the 7 pipelines, 7 specs, 23 fixtures and 6 workflows behind them
```

`full_release` never used any of it: it composes the jobs itself and posts its own `gio_action`. Every job the removed workflows referenced is still used elsewhere, so nothing that builds or publishes has gone away — only the seven ways of triggering a subset of it by hand.

Replaying after a failure is now CircleCI's own *Rerun from failed*, which picks the workflow up where it stopped instead of starting a separate pipeline on a fresh workspace. The README says so.

## Additional context

**What referenced the removed pipelines.** The seven commands were referenced only by their own entries in `package.json`, and no CI job invoked them. One thing outside that did depend on a removed pipeline, and a review caught it: `release/staging/stage-release-helm.mjs` posted `gio_action: 'release_helm'` directly. It is deleted here — written in June 2023, untouched since except for a dependency bump and a Node 24 fix, and recoverable from history if it turns out to be missed. `fast-xml-parser` was its only user and goes with it.

`workflow-nexus-staging.ts` is removed too: `pipeline-nexus-staging.ts` was its only consumer, and it built its workflow differently from the other six, which is how it escaped the first pass.

**Verified**:

* `node --test "helpers/*.test.mjs"` in `release/` — 58 tests, green.
* `npx jest` in `.circleci/ci` — 132 tests, green (182 before; the 50 that went are the specs of the removed pipelines).
* `.circleci/ci/README.md` used `build_rpm` as its worked example in three places, for an action that no longer resolves. It now uses `publish_docker_images`, which the same file already used further up.
* **No surviving fixture changed.** The generated configuration of every remaining pipeline is byte-identical, which is the evidence that nothing shared was disturbed.
* `tsc --noEmit` clean, prettier clean on both sides.

**What the first commit removes along the way**: `isHotfixBranch` and the version/branch cross-check lived in `generateReleaseConfig`, the pipeline behind `yarn release`. They had no other consumer, so they go with it.

The second commit covers the part that mattered, and covers it better: comparing the version against the poms catches every mismatch, not only the hotfix shape, and it fails in your terminal before a pipeline is triggered rather than after. What it deliberately does **not** restore is the branch whitelist — `full_release` has accepted any branch since `bc3fcb7cb2` removed its `isSupportBranch` guard in June, on purpose, and that stays.

**The check reads GitHub through the contents API**, with `Accept: application/vnd.github.raw`, rather than `raw.githubusercontent.com`. Measured: the latter is served with `cache-control: max-age=300`, long enough for a release started right after a version bump to read the previous pom and refuse a correct release. The API is cached for one minute, returns the file itself rather than a base64 envelope — same 35 084 bytes, same parsing — and the refusal message names the cache. It also runs before any interactive prompt, so a deterministic refusal no longer waits behind three questions.

**Exercised against the real branches**, not only unit-tested:

```
master   HTTP 200  →  would publish 4.13.0
4.12.x   HTTP 200  →  would publish 4.12.18
4.11.x   HTTP 200  →  would publish 4.11.26
4.9.x    HTTP 200  →  would publish 4.9.33
a branch that does not exist  →  HTTP 404, reported and refused
```

Six unit tests cover the pom reading itself, including the self-closing `<sha1 />` of a final release and the changelist that the release clears.

The frontend build jobs tested `['build_rpm', 'build_docker_images', 'release', 'full_release']` to decide where the version comes from; three of those actions no longer exist, so the test is now an equality on `full_release`. Same outcome for every surviving action, and the fixtures confirm it.

**Scheduled triggers are safe.** They are configured in the CircleCI UI and invisible from here, so they were checked with their owner: they use `repositories_tests`, `bridge_compatibility_tests`, `nightly` and `helm_tests`. None is removed, and all four still resolve — present in `config.yml` and in the dispatch.

**Noticed, not addressed**: `config-factory.ts` redeclares the `gio_action` enum for the dynamic configuration, and it was already missing `nightly`, `helm_tests`, `integration_tests` and `maven_release` before this change. Those actions work anyway, so the redeclaration is apparently not what validates them. Left as it was.

**Security-sensitive**: this changes CI/CD configuration. It removes ways to trigger a publication; it adds none.


release/helpers/version-helper.mjs
```diff
@@ -46,3 +46,46 @@ function trimmed(releasingVersion) {
...
+/**
+ * Stops the release when --version and the branch's poms disagree.
+ *
+ * They are two independent inputs and nothing downstream reconciles them: the artefacts carry what
+ * the poms say, the git tag carries what --version says. A mismatch therefore ships one version
+ * under the name of another, without failing anywhere. This is the only place they are compared.
+ * @param {string} version the version passed to the command
+ * @param {string} branch the branch the release will run on
+ */
+export async function assertVersionMatchesPom(version, branch) {
```
Okhelifi, line 76: ⚠️ [58/100] raw.githubusercontent.com is CDN-cached, so the check can read a stale pom —

Branch content is served with a roughly five-minute TTL, which means a release run shortly after a pom bump is pushed (code freeze, hotfix prep) reads the previous version and refuses a correct release. It fails closed so this is friction, not a hazard, but a cache-busting query param or api.github.com/repos/.../contents/pom.xml?ref=<branch> removes it.

.circleci/config.yml
```diff
@@ -14,18 +14,11 @@ parameters:
...
-                nexus_staging,
                 pull_requests,
-                build_rpm,
-                build_docker_images,
                 build_chainguard_images,
                 build_chainguard_fips_images,
-                release_notes_apim,
                 bridge_compatibility_tests,
                 publish_docker_images,
-                release_helm,
```
Okhelifi, line 28: ⚠️ [85/100] `release_helm` is removed from the enum but still triggered by a maintained script
→ release/staging/stage-release-helm.mjs:18 (file not in this diff)

That script POSTs `gio_action: 'release_helm'`, the value deleted here and in the
`buildCIPipeline` switch, so CircleCI now rejects the request and the script prints
`Expected status 201, got 400`. The description asserts nothing outside `package.json`
referenced what was removed; that holds for the seven commands but not for this pipeline,
last maintained in 080ed0fc63. Delete it with the rest, or keep `release_helm` alive.

.circleci/config.yml
```diff
@@ -14,18 +14,11 @@ parameters:
         enum:
             [
                 full_release,
-                release,
-                package_bundle,
-                nexus_staging,
                 pull_requests,
-                build_rpm,
```
Okhelifi, line 21: 💡 [48/100] The CI README still documents `build_rpm` as its worked example
→ .circleci/ci/README.md:63,68,81 (file not in this diff)

Both the local `npm run generate` recipe and the temporary-commit trick use an action
that no longer exists, and for which `buildCIPipeline` now returns `null`. Swap it for a
surviving action, `build_chainguard_images` or `publish_docker_images`.

release/steps/full_release.mjs
```diff
@@ -44,6 +44,7 @@ if (argv.latest) {
 }
 
 const targetBranch = getTargetBranch(versions);
+await assertVersionMatchesPom(releasingVersion, targetBranch);
```
Okhelifi: 💡 [40/100] The check sits behind three interactive prompts

A deterministic, sub-second refusal runs only after the JIRA confirmation, the alpha confirmation and the --latest question. getTargetBranch(versions) is pure and computable right after line 10; is there a reason it has to wait until line 46?

release/helpers/version-helper.mjs
phiz71: Measured and taken. `curl -I` on the branch pom confirms `cache-control: max-age=300`, so the window you describe is real: a release started right after a version bump would read the previous pom and refuse a correct release.

Switched to the contents API, with `Accept: application/vnd.github.raw`. That returns the file itself rather than a base64 envelope — same 35 084 bytes as the raw host, same regex, no parsing change — and it is served with `max-age=60`. Five times less staleness for the same code.

The refusal message now names the cache for the residual minute, so the reader is not left guessing. Rate limit is 60 requests per hour unauthenticated; a release makes one.

For the record I also measured `github.com/.../blob/...`, which has no cache at all: 580 KB of HTML, with the file content either unicode-escaped inside a JSON payload or split across syntax-highlighting spans. Fresh, but a dependency on an undocumented page shape.

.circleci/config.yml
phiz71: You are right, and this is the finding that mattered.

My verification looked for two things: that nothing referenced the removed *filenames*, and that no CI job ran `yarn <command>`. I never searched for the action strings themselves — which is exactly how an independent script triggers a pipeline. The description asserted that nothing outside `package.json` referenced what was removed; that was wrong, and it has been corrected.

The same gap left something else behind, which you did not have to catch: `workflow-nexus-staging.ts` had no consumer once its pipeline went. It built its workflow differently from the other six, so it escaped a delete list I had written by hand. Removed now.

On the choice you offer — the script is deleted rather than the action kept. It dates from June 2023 and neither commit since changed its behaviour, only a dependency bump and a Node 24 fix. `fast-xml-parser` was its only user and goes with it. If it turns out to be missed, it comes back from history in one commit.

.circleci/config.yml
phiz71: Fixed. Three occurrences, at lines 63, 68 and 81, and `buildCIPipeline` does return `null` for that action now.

Swapped for `publish_docker_images` — which the same file already used at line 45, so the README was quietly inconsistent with itself before this change. It now uses one action throughout.

release/steps/full_release.mjs
phiz71: No reason at all — it was where the branch happened to be computed, not where it belonged.

`getTargetBranch(versions)` is pure and needs nothing but the version, so both it and the check now run immediately after `computeVersion`, in `full_release` and `maven_release` alike. A mismatch refuses before the JIRA confirmation rather than after three questions.

release/package.json
```diff
@@ -6,7 +6,6 @@
   "license": "MIT",
   "devDependencies": {
     "dotenv": "17.4.2",
-    "fast-xml-parser": "5.10.1",
```
Okhelifi, line 9: Can you regenerate the yarn.lock and add it to the PR please ?

release/README.adoc
```diff
@@ -45,29 +45,20 @@ yarn full_release --version=4.12.18
...
-----
-yarn release       --version=4.12.18
-yarn package_zip   --version=4.12.18
-yarn docker        --version=4.12.18   # --latest to also tag as latest
-yarn rpms          --version=4.12.18
-yarn helm_chart    --version=4.12.18
-yarn release_notes --version=4.12.18
-yarn nexus_sync    --version=4.12.18
-----
+Nothing resumes it on its own. Rerun the failed job from the CircleCI workflow — *Rerun from failed* picks the release up where it stopped, on the same workspace.
```
Okhelifi: It replays the failed job and everything downstream, which for a mid-graph failure
means a dozen jobs that each publish something. CircleCI can rerun one chosen job
instead, via its v2 API or the UI itself. The README should say so, and list which jobs of full_release are safe to replay on their own, since the PR already notes the steps are not idempotent. wdyt ?

release/package.json
phiz71: Done — `yarn.lock` regenerated with yarn 4.1.1, in `update-lockfile` mode. 71 lines removed, all of them `fast-xml-parser` and its transitive dependencies, nothing added.

release/README.adoc
phiz71: Agreed, and the graph makes it worse than it sounds in one place and better in another — so the README now says which is which rather than listing all 33 jobs, a list that would rot.

Most of what publishes is a leaf or close to one: the bundle, the RPMs, the Helm chart, the Nexus sync, the API docs ingestion. A failure there replays little more than itself, which is why *Rerun from failed* is still the right first answer.

The case worth the warning is a build job. `Backend build and publish on download website` and the three UI builds feed the Docker images, their chainguard and FIPS variants, and through them the job that commits the version and pushes the tag — so rerunning from failed replays all of that, tagging included.

The section now leads with "restarts the failed job **and everything that depends on it**", names that chain, and gives the single-job call:

```
curl -X POST -H "Circle-Token: $CIRCLECI_TOKEN" \
  https://circleci.com/api/v2/workflow/<workflow-id>/rerun \
  -d '{"jobs": ["<job-id>"]}'
```

Okhelifi (approved):
👍🏼

PR #19399, opened 2026-08-26
feat: enable API key renewal for API product subscriptions

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-139

## Summary

This PR enables API key renewal for API Product subscriptions in Portal Next.

- Adds the **Renew API Key** action for accepted API Product subscriptions using an API Key plan.
- Displays the action only when the user has the required application subscription update permission.
- Reuses the existing confirmation dialog and API key renewal endpoint.
- Refreshes the subscription details after successful renewal.
- Provides accessible success and error feedback.
- Ensures API Product renewals use application-level permissions without falling back to API permissions.
- Prevents renewal of API Product subscriptions belonging to another environment.

No HTTP contract changes are introduced.

### DIFF
--- a/gravitee-apim-portal-webui-next/src/app/dashboard/subscription-details/api-product-subscription-details/api-product-subscription-details.component.ts
+++ b/gravitee-apim-portal-webui-next/src/app/dashboard/subscription-details/api-product-subscription-details/api-product-subscription-details.component.ts
@@ -14,16 +14,17 @@
  * limitations under the License.
  */
 import { DatePipe } from '@angular/common';
-import { Component, computed, DestroyRef, inject, input, linkedSignal, signal } from '@angular/core';
+import { Component, computed, DestroyRef, inject, input, linkedSignal, output, signal } from '@angular/core';
 import { rxResource, takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { MatButtonModule } from '@angular/material/button';
 import { MatCardModule } from '@angular/material/card';
 import { MatDialog } from '@angular/material/dialog';
+import { MatTooltipModule } from '@angular/material/tooltip';
 import { RouterLink } from '@angular/router';
-import { catchError, map, Observable, of, switchMap } from 'rxjs';
+import { catchError, filter, finalize, map, Observable, of, switchMap, tap } from 'rxjs';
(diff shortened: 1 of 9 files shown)


gravitee-apim-portal-webui-next/src/app/dashboard/subscription-details/api-product-subscription-details/api-product-subscription-details.component.ts
```diff
@@ -64,15 +67,20 @@ interface ActionFeedback {
...
   private readonly destroyRef = inject(DestroyRef);
 
   readonly subscription = input.required<Subscription>();
+  readonly apiKeyRenewed = output<void>();
 
   protected readonly displayedSubscription = linkedSignal(() => this.subscription());
   protected readonly pendingAction = signal<LifecycleAction | null>(null);
   protected readonly actionFeedback = signal<ActionFeedback | null>(null);
+  protected readonly isRenewingApiKey = signal(false);
+  protected readonly apiKeyFeedback = signal<ApiKeyFeedback | undefined>(undefined);
```
thomasballivet: nit: `apiKeyFeedback` is never cleared after the parent's `subscriptionResource.reload()`, so the success message stays on the card indefinitely. Could be reset on the next subscription input (e.g. a `linkedSignal` on `subscription()` instead of a plain `signal`).

jaroslawlakomy-grav: Implemented. `apiKeyFeedback` is now reset when the parent provides the refreshed subscription after `subscriptionResource.reload()`. The renewal test also verifies that the previous success feedback is cleared.

jaroslawlakomy-grav:
The SonarCloud finding covers 25 duplicated lines of the API key renewal confirmation flow shared with `ApiAccessComponent`.

Extracting this code would require introducing a shared abstraction and modifying both the existing generic API access component and the new API Product subscription component. This would add indirection and broaden the scope of the change for a small, stable flow with only two call sites.

thomasballivet (approved):

PR #19400, opened 2026-08-26
fix(gateway): do not fail requests on an unparseable logging condition

SamHardi:
## Summary

A logging condition that cannot be parsed answered **every** request to the API with a 500. The gateway now
skips logging for that request, says why — a gateway `WARN` and an `EXPRESSION_EVALUATION_ERROR` diagnostic on
the analytics record — and lets the request through. Policy and flow conditions are unchanged and keep failing
closed.

Closes [APIM-14947](https://gravitee.atlassian.net/browse/APIM-14947).

## Root cause

`SpelTemplateEngine.eval(String, Class)` (gravitee-expression-language 4.4.1) reports the two EL failure modes
with two different exception types:

| Failure | Exception surfaced | Handled by `ExpressionLanguageConditionFilter` |
| --- | --- | --- |
| Condition parses, evaluation fails | `ExpressionEvaluationException` | warn, element filtered out, execution continues |
| Condition **does not parse** | `IllegalArgumentException` from `ExpressionEvaluationException.buildCause` | propagated → `DefaultApiReactor` answers 500 |

`LogInitProcessor` used the filter as is, so the second row turned a typo in a logging condition into an outage.

## Why the fix is in `LogInitProcessor`, not in the condition filter

Making the filter fail open for every error breaks a deliberate, long-standing contract:
`InvalidConditionalPolicyTest` and `InvalidConditionalStreamPolicyTest` (standalone container, both engine
executions) assert that a policy whose condition cannot be parsed answers 500 and never reaches the backend. A
filter-wide fail-open would silently disable a conditional policy — a security policy included — on every
request, because of a typo. Logging enforces nothing, so logging is the only consumer that should degrade
gracefully.

## Changes

- `LogInitProcessor` — an error from evaluating the logging condition skips logging and is reported once: a
  gateway `WARN` carrying the failure message (the stack trace goes to `DEBUG`, since a parse failure repeats on
  every request) and an `EXPRESSION_EVALUATION_ERROR` execution warning carrying the cause. The handling sits
  **before** the log entity is built, so a failure there still propagates. Evaluation failures are still
  reported by the filter itself, so nothing is reported twice.
- `LogInitProcessorTest` — three tests:
  - two drive the **real** template engine with the ticket's condition `{#request.headers["x"][0] ==}`: the
    processor completes without setting a `Log`, and the warning carries the key, the condition and the
    `EL1042E` cause;
  - one pins the placement: a failure while building the log entity reaches the subscriber and is not reported
    as a condition failure (verified to fail when the error handling is moved below `initLogEntity`).
- `LoggingV4IntegrationTest` — nested `UnparseableLoggingCondition`: real HTTP call through a real gateway →
  200, backend reached, no `Log`, `EXPRESSION_EVALUATION_ERROR` on the analytics record.
- `ExpressionLanguageConditionFilter` and its test — **unchanged** compared to `master`.

No HTTP endpoint, OpenAPI contract or configuration key changes.

## Surfaces affected

| Surface | Path | Behaviour after this PR |
| --- | --- | --- |
| V4 APIs with analytics logging | `v4.processor.ApiProcessorChainFactory` → `LogInitProcessor` | 200, logging skipped, warning |
| V2 APIs on the V4 emulation engine | `processor.ApiProcessorChainFactory` → `LogInitProcessor` | 200, logging skipped, warning |
| V2 APIs on the legacy V3 engine | `LoggableRequestProcessor` (already catches and warns) | unchanged |
| Policy, flow and shared-policy-group conditions | `ExpressionLanguageConditionFilter` | unchanged |

## Security-sensitive

Touches error handling, logging and operator-visible diagnostics on the gateway request path. The fail-open is
confined to logging conditions, which enforce nothing; conditional policies keep failing closed on a parse
failure.

## Test plan

- [ ] `LogInitProcessorTest` parse-failure tests: RED on `master` (`IllegalArgumentException … EL1042E: Problem parsing right operand`), GREEN with the change
- [ ] `LogInitProcessorTest.should_propagate_a_failure_raised_while_building_the_log_entity`: RED if `.onErrorComplete()` is moved below `doOnSuccess(initLogEntity)`
- [ ] `gravitee-apim-gateway-standalone-container`, both `v3-engine-test` and `emulate-v4-engine-test` executions: `InvalidConditional*PolicyTest` still expect 500
- [ ] Integration: `LoggingV4IntegrationTest$UnparseableLoggingCondition` → 200, no `Log`, `EXPRESSION_EVALUATION_ERROR` warning
- [ ] Manual, V4 proxy API: logging condition `{#request.headers["x"][0] ==}` → 200, one `WARN` line without stack trace per request, and the warning on the API's analytics entry
- [ ] Manual, same condition on a **policy** → still 500 (regression guard for the fail-closed contract)

### Local verification

| Module | Tests | Result |
| --- | --- | --- |
| `gravitee-apim-gateway-core` | 415 | pass |
| `gravitee-apim-gateway-handlers-api` | 1030 | pass |
| `gravitee-apim-gateway-standalone-container` — `v3-engine-test` / `emulate-v4-engine-test` | 141 / 140 (3 skipped each) | pass |
| Integration tests — `Logging{V3,V4,V4Emulation}`, `FlowPhaseExecution*`, `BestMatchWithCondition`, `SharedPolicyGroup`, `FlowConditionAfterAsyncPolicy` | 70 | pass |

The review fixes only change `LogInitProcessor`'s log calls and add a unit test; the standalone-container and
integration suites were run before them and are covered again by CI. The full integration-tests module was not
run locally.

## Reviewer notes

Follow-ups surfaced while working this, **not** addressed here:

- A **policy** condition that cannot be parsed still answers 500 with a bare `internal_error`:
  `DefaultApiReactor.processThrowable` only logs `Unexpected error while handling request` and attributes no
  diagnostic. That is the ticket's "if a request does fail, the API log should say why".
- Policy conditions are asymmetric: an evaluation failure skips the policy with a warning (a6a49be), a parse
  failure answers 500. Whether both should fail closed is a gateway product decision.
- V4 flow conditions (`ConditionSelectorConditionFilter`) resume on **any** error with no warning at all.
- Rejecting an unparseable condition when the API is saved — the ticket's first expected behaviour — is not
  covered.

## Backport

Reported on 4.11.17. As measured at review on the previous head: `4.12.x` cherry-picks cleanly; `4.11.x`
conflicts on `LogInitProcessorTest.java` and on `LoggingV4IntegrationTest.java`, which lives under
`gravitee-apim-integration-tests/` on that branch. Not re-measured after the review fixes.

[APIM-14947]: https://gravitee.atlassian.net/browse/APIM-14947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


phiz71 (commented):
Reviewed at `bec05663d1`. Two findings that anchor to no line of the diff.

**SHOULD** — `761ef311a2` is undone by `bec05663d1`. It opens the filter for every error without touching `InvalidConditionalPolicyTest` / `InvalidConditionalStreamPolicyTest`, which assert `500` and `verify(0, anyRequestedFor(anyUrl()))` — so it lands red. Merged with *Rebase and merge* (used on this repo alongside squash), master inherits that commit and Mergify replays both onto the support branches.

Measured, cherry-picking onto `origin/4.11.x`: the two commits conflict on the first one (`LoggingV4IntegrationTest.java`); squashed into one they conflict on `LogInitProcessorTest.java` and `LoggingV4IntegrationTest.java`. `4.12.x` is clean either way. Squashing does not remove the `4.11.x` conflict — the integration test lives under `gravitee-apim-integration-tests/` there — but it removes the red commit and reduces the resolution to a single pass.

**SHOULD** — APIM-14947 also asks that "if a request does fail, the API log should say why". A **policy** condition carrying the same typo still answers 500: `DefaultApiReactor.processThrowable`, `else` branch (line 550), only logs `Unexpected error while handling request` and runs the error chain — no diagnostic is attributed, so the analytics record stays `internal_error` / "Internal Server Error". The three follow-ups listed under *Reviewer notes* are referenced by no ticket; opening them and linking them to APIM-14947 before it is closed would keep that half of the ticket from being lost.

Purely personal preference, ignore freely: the subject of `bec05663d1` is 101 characters and its body narrates the abandoned approach — both moot if you squash and write a fresh subject.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/processor/logging/LogInitProcessor.java
```diff
@@ -64,12 +67,26 @@ public Completable execute(final HttpExecutionContextInternal ctx) {
                 return Completable.complete();
             }
 
-            return CONDITION_FILTER.filter(ctx, analyticsContext.getLoggingContext())
+            final LoggingContext loggingContext = analyticsContext.getLoggingContext();
+            return CONDITION_FILTER.filter(ctx, loggingContext)
+                // The filter propagates a condition that cannot be parsed, so that a broken policy condition fails closed.
+                // Logging enforces nothing: a condition it cannot evaluate only skips logging, never the request.
+                .doOnError(throwable -> reportUnusableCondition(ctx, loggingContext.getCondition(), throwable))
+                .onErrorComplete()
```
phiz71, line 75: **SHOULD** — This placement is load-bearing and nothing pins it: moving `.onErrorComplete()` below `.doOnSuccess(initLogEntity)` keeps all 14 `LogInitProcessorTest` tests green, so a later refactor would silently swallow any failure while the log entity is being built, and no test would fail.

Add one: make `mockMetrics.setLog(any())` throw, then assert the error still reaches the subscriber.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/processor/logging/LogInitProcessor.java
```diff
@@ -64,12 +67,26 @@ public Completable execute(final HttpExecutionContextInternal ctx) {
...
                 .doOnSuccess(activeLoggingContext -> initLogEntity(ctx, activeLoggingContext))
                 .ignoreElement();
         });
     }
 
+    private void reportUnusableCondition(final HttpExecutionContextInternal ctx, final String condition, final Throwable throwable) {
+        ctx.withLogger(log).warn("Logging condition {} cannot be evaluated, logging is skipped for this request", condition, throwable);
```
phiz71: **COULD** — A parse failure is deterministic: it repeats on every request until the operator fixes the condition, so this writes a full stack trace per request indefinitely — on the API that triggered the ticket, that is all of its traffic. The message at WARN and the stack at DEBUG would be enough.

The filter already does the same for evaluation failures, so keeping it is a defensible consistency call.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/processor/logging/LogInitProcessor.java
SamHardi: Fixed in b0c3284075. Added `should_propagate_a_failure_raised_while_building_the_log_entity`: `setLog` throws, the error reaches the subscriber and no `EXPRESSION_EVALUATION_ERROR` warning is recorded. Checked that it fails (`No errors (latch = 0, values = 0, errors = 0, completions = 1)`) when `.onErrorComplete()` is moved below `doOnSuccess(initLogEntity)`.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/processor/logging/LogInitProcessor.java
SamHardi: Fixed in b0c3284075. The `WARN` now carries only the failure message (`Logging condition {} cannot be evaluated, logging is skipped: {}`); the stack trace moved to `DEBUG`. The filter's own `WARN` for evaluation failures is left unchanged here, since the filter is out of this PR's scope.

SamHardi:
Thanks @phiz71 — addressed at `b0c3284075`.

**Red intermediate commit** — squashed into a single commit with a fresh subject (70 characters) whose body describes only the final change, so *Rebase and merge* no longer carries the filter-wide commit that failed `InvalidConditional*PolicyTest`. As you measured, the `4.11.x` backport still needs one manual resolution; the PR description now says so and notes it was not re-measured after the review fixes.

**"If a request does fail, the API log should say why"** — agreed, and confirmed: a policy condition with the same typo still reaches `DefaultApiReactor.processThrowable`'s `else` branch, which only logs `Unexpected error while handling request` and attributes no diagnostic. This PR stays scoped to the reported logging case so it backports cleanly. That gap and the other follow-ups listed under *Reviewer notes* have been created for follow-up and linked to APIM-14947:

- [APIM-15112](https://gravitee.atlassian.net/browse/APIM-15112) — umbrella story: [Gateway] EL conditions that cannot be evaluated
  - [APIM-15113](https://gravitee.atlassian.net/browse/APIM-15113) — [Gateway] Attribute a diagnostic when a policy condition cannot be parsed
  - [APIM-15114](https://gravitee.atlassian.net/browse/APIM-15114) — [Gateway] Align policy condition failure modes (evaluation vs parse)
  - [APIM-15115](https://gravitee.atlassian.net/browse/APIM-15115) — [Gateway] Warn when a V4 flow condition fails to evaluate
  - [APIM-15116](https://gravitee.atlassian.net/browse/APIM-15116) — [Console] Reject unparseable EL conditions when saving an API

The two line comments are fixed and resolved: the placement of `.onErrorComplete()` is now pinned by a test (verified red when moved), and the stack trace moved to `DEBUG`.

[APIM-15112]: https://gravitee.atlassian.net/browse/APIM-15112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APIM-15113]: https://gravitee.atlassian.net/browse/APIM-15113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/processor/logging/LogInitProcessor.java
phiz71: Verified on `b0c3284075`: 15/15 green, and moving `.onErrorComplete()` back below `doOnSuccess(initLogEntity)` turns the new test red with exactly the message you quote. The added assertion that no `EXPRESSION_EVALUATION_ERROR` is recorded also catches the log-entity failure being reported as a condition failure — stricter than what I asked for.

gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/processor/logging/LogInitProcessor.java
phiz71: Verified on `b0c3284075`: the `WARN` carries the message alone, the stack trace sits on its own `DEBUG` line. Leaving the filter's own `WARN` alone is the right call here.

phiz71:
Re-measured on the squashed head, since the description notes it was not: cherry-picking `b0c3284075` onto `origin/4.12.x` is clean, onto `origin/4.11.x` it conflicts on **both** `LogInitProcessorTest.java` and `LoggingV4IntegrationTest.java`. The manual resolution stands, on two files.

The squash checks out: one commit, 70-character subject, body describing only the final change, and `ExpressionLanguageConditionFilter` still byte-identical to `master`.

phiz71 (approved):
Approving `b0c3284075`.

Both line comments are fixed and re-verified rather than taken on trust: `LogInitProcessorTest` is 15/15 green, and moving `.onErrorComplete()` back below `doOnSuccess(initLogEntity)` turns `should_propagate_a_failure_raised_while_building_the_log_entity` red, so the placement is genuinely pinned now. The `WARN` carries the message alone and the stack trace sits on its own `DEBUG` line.

The squash removes the intermediate commit that failed `InvalidConditional*PolicyTest`, and `ExpressionLanguageConditionFilter` is byte-identical to `master`, so policy conditions keep failing closed. Backport re-measured on this head: `4.12.x` clean, `4.11.x` conflicts on `LogInitProcessorTest.java` and `LoggingV4IntegrationTest.java`.

The one point left is outside the code: opening the follow-up tickets — the missing diagnostic on a 500 from a policy condition, the policy-condition asymmetry, the silent V4 flow conditions, save-time validation — and linking them to APIM-14947 before it is closed.

PR #19401, opened 2026-08-26
feat(reporter-elasticsearch): let operators stop indexing captured log bodies

SamHardi:
## Summary

The Elasticsearch reporter indexed every request, response and message body the logging policy captures as a full-text field. That is the heaviest work the reporter gives the cluster — `reporters.logging.max_size` is a limit **per captured body**, so a v4 log document at 256KB carries up to four of them — and on a gateway running thousands of requests per minute it is what puts log ingestion hours behind. Bodies are now indexed for prefix search only, and operators who need more can stop indexing them altogether while keeping them visible in the Logs UI.

Raised from [APIM-13313](https://gravitee.atlassian.net/browse/APIM-13313) / [TT-10329](https://gravitee.atlassian.net/browse/TT-10329), where the Platform team's conclusion was *"open a ticket on APIM team as the mapping is not performant and we have this issue on multiple customer"*, and Antoine Cordier's suggestion was to test disabling the body analyzer on that field.

## Changes

- **Bodies are indexed for prefix search only** — `index_options: docs` and `norms: false` on every body/payload field. Positions serve phrase queries and norms serve relevance scoring; the only query that ever reads these fields is the Logs UI prefix `query_string` over `*.body` ([`SearchConnectionLogDetailQueryAdapter#addBodyTextFilter`](https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/adapter/connection/SearchConnectionLogDetailQueryAdapter.java#L87), [`ElasticLogRepository#getQuery`](https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/log/ElasticLogRepository.java#L158)), which needs neither.
- **New setting `reporters.elasticsearch.settings.index_body`** (default `true`). Set to `false` and the bodies' inverted index disappears: they stay in `_source`, so the Logs UI still displays them in the log detail view, but its "search in payloads" filter no longer matches them.
- The body mapping was repeated 36 times across the four template trees; it moves to a shared freemarker macro, `freemarker/common/mapping/log-body-field.ftl`.
- The macro reads `indexBody`, so every renderer of these templates has to supply it. The elasticsearch repository tests render them through their own `DatabaseHydrator` rather than through the reporter's `IndexPreparer`; it now passes the production default.
- Documented in the gateway `gravitee.yml` and in the reporter README.

Both are **mapping** settings: they apply to indexes created after the change (next daily index or ILM rollover), not to existing ones.

## Test report

All numbers below come from real calls against Elasticsearch 9.2.1 (Lucene 10.3.1) in Docker — single node, 2 GiB heap, 1 shard, 0 replicas, `refresh_interval: 5s`. The corpus is seeded v4 log documents whose four captured bodies are ~64 KiB each (256 KiB of payload per request), compact JSON with human-text values, as a real API response would be.

### A. What the templates render — `LogBodyMappingTest`, 48 cases, no container

| # | Test | Expected | Observed | Result |
|---|---|---|---|---|
| A1 | Every body field, 4 template trees × 3 log types | `type: text`, `index_options: docs`, `norms: false` | as expected, 12/12 combinations | **PASS** |
| A2 | Same, with `index_body: false` | `type: text`, `index: false`, and no `analyzer` / `index_options` / `norms` (Elasticsearch rejects those on an unindexed field) | as expected, 12/12 | **PASS** |
| A3 | Existing analyzer per tree is untouched | `gravitee_body_analyzer` everywhere except es7x v4, which keeps the Elasticsearch default | as expected, 12/12 | **PASS** |
| A4 | Every captured payload still mapped | 4 bodies for `log` and `v4-log`, 1 payload for `v4-message-log` | as expected, 12/12 | **PASS** |

### B. Behaviour on a real cluster — `LogBodyIndexingIntegrationTest`, Testcontainers ES 9.2.1

Each case runs the reporter's own `IndexPreparer.prepare()` (a real `PUT _index_template` for every type), then a real refreshed `_bulk`, then the query the Logs UI issues.

| # | Test | Expected | Observed | Result |
|---|---|---|---|---|
| B1 | Cluster accepts the rendered templates, both modes | `prepare()` completes with no error | completes in both modes | **PASS** |
| B2 | Payload search with body indexing on | 1 hit | 1 hit | **PASS** |
| B3 | Payload search with `index_body: false` | 0 hits | 0 hits | **PASS** |
| B4 | Log detail view with `index_body: false` | body still returned in `_source` | `entrypoint-response.body` returned intact | **PASS** |

### C. Payload search returns exactly the same thing — direct HTTP, same corpus indexed under both mappings

The query is the one the adapter builds: the Lucene-escaped term plus `*`, over `\*.body`.

| # | Search input | Expected | Observed (current → new) | Result |
|---|---|---|---|---|
| C1 | `order-4711` (whole token) | identical | 1 → 1 | **PASS** |
| C2 | `order` (token prefix) | identical | 1 → 1 | **PASS** |
| C3 | `ACME` | identical | 0 → 0 | **PASS** |
| C4 | `acme` (case folding) | identical | 0 → 0 | **PASS** |
| C5 | `SKU-1234` | identical | 0 → 0 | **PASS** |
| C6 | `{"status":"accepted"}` (raw JSON typed as a term) | identical | HTTP 400 → HTTP 400 | **PASS** |
| C7 | `ACME\ Corp` (multi-word, escaped as the adapter escapes it) | identical | 0 → 0 | **PASS** |
| C8 | `nothing-here` | identical | 0 → 0 | **PASS** |
| C9 | `42.5` | identical | 0 → 0 | **PASS** |
| C10 | `match_phrase` on a body — **not a query APIM issues** | breaks without positions | 0 hits → HTTP 400 *"field was indexed without position data"* | **known consequence** |

C10 is the one thing `index_options: docs` takes away. Nothing in APIM issues a phrase query on a body field: the only `match_phrase` in the Elasticsearch repository is on `additional-metrics.string_*` in the message-metrics index, which this change does not touch. Note that on the current mapping the phrase query returns 0 hits anyway, because `gravitee_body_analyzer` splits on whitespace only and never produces the tokens a phrase query looks for.

C3/C5 returning 0 on both mappings is pre-existing behaviour, not a regression: with a whitespace tokenizer a compact JSON body is one single token, so a prefix search only reaches a term that starts a whitespace-delimited word.

### D. The transition window — old and new mappings behind one index pattern

A mapping change only applies to new indexes, so for a while the Logs UI searches a wildcard spanning both. If Elasticsearch refused a `query_string` over a pattern where one index has the field unindexed, the Logs UI would break for every operator flipping the flag.

| # | Index pattern searched | Expected | Observed | Result |
|---|---|---|---|---|
| D1 | current mapping only | 1 hit | 1 hit | **PASS** |
| D2 | new mapping only | 1 hit | 1 hit | **PASS** |
| D3 | `index_body: false` only | 0 hits, no error | 0 hits | **PASS** |
| D4 | current + new | 2 hits | 2 hits | **PASS** |
| D5 | current + unindexed | 1 hit, no error | 1 hit | **PASS** |
| D6 | wildcard over all three | 2 hits, no error | 2 hits | **PASS** |

### E. What it costs the cluster

**E1 — body index structures**, measured field by field with `_disk_usage?run_expensive_tasks=true`, 400 documents:

| Mapping | Bodies' inverted index | `_source` | Result |
|---|---|---|---|
| current master | 26.8 MiB | 39.4 MiB | baseline |
| this PR, default | **20.2 MiB (−25%)** | 39.4 MiB | **PASS** |
| this PR, `index_body: false` | **0.0 MiB (−100%)** | 39.4 MiB | **PASS** |

**E2 — indexing and store**, 1200 documents (316 MiB of JSON), one throwaway container per mapping, two runs each:

| Mapping | Shard indexing time | Merge time | Store after force-merge |
|---|---|---|---|
| current master | 8045 / 7931 ms | 4595 / 5277 ms | 198.4 MiB |
| this PR, default | 7652 / 7709 ms (−4%) | 4213 / 4453 ms (−12%) | 165.7 / 178.7 MiB |
| this PR, `index_body: false` | **3170 / 2876 ms (−62%)** | **562 / 466 ms (−89%)** | **118.4 MiB (−40%)** |

Read honestly: **the default change is a disk and merge win, not an indexing-time win.** Analysis dominates shard indexing time and the analyzer is unchanged, so the −4% sits inside run-to-run noise and should not be quoted as a speedup. The lever for a cluster that cannot keep up is `index_body: false`.

An earlier version of this benchmark ran all three mappings against one container and produced incoherent numbers (the unindexed mapping came out *larger* than the baseline) because background merges and pre-merge segment files leaked across mappings. The figures above come from one container per mapping, with the store read only after the shard settles to a single segment of stable size.

### F. Regression

| # | Test | Expected | Observed | Result |
|---|---|---|---|---|
| F1 | `mvn -pl gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch test` | all green | **144 tests, 0 failures, 0 errors** | **PASS** |

Includes the pre-existing `IndexTemplateTest` (50), `IndexPreparerIntegrationTest` against a real OpenSearch container (3), and `ElasticsearchReporterTest` against a real Elasticsearch container (6).

## Answers to the questions on the ticket

**"They currently have `gravitee_reporters_logging_max_size` set to 256KB. Should we suggest a smaller value?"**
It helps, but it is the wrong lever to reach for first, and it is bigger than it looks: the limit is applied **per captured body**, so 256KB means up to 1 MB of payload per request document. Lowering it trades away log content for indexing cost. `index_body: false` removes almost all of that indexing cost while keeping every byte of the payload visible in the Logs UI.

**"Are there any other parameters we can work with?"**
Before this PR, no — nothing exposed the body mapping. Now, in order of impact:
1. `reporters.elasticsearch.settings.index_body: false` — the measured −62% / −40% above. Costs payload search.
2. `reporters.logging.excluded_response_types` — already exists, already defaults to excluding video, audio, images, octet-stream and PDF. Worth confirming it is not overridden.
3. `reporters.logging.max_size` — remember it is per body.
4. `reporters.elasticsearch.bulk.actions` / `flush_interval` and `settings.refresh_interval` — these shape how the reporter talks to the cluster, not how much work the cluster does per document.

**Antoine's suggestion — "disable the body_analyser indexer on that field, worth testing on our side, to check how the logs UI reacts"** — tested, sections B, C and D. The Logs UI keeps rendering payloads in the log detail view (they come from `_source`); only the "search in payloads" filter goes quiet, and it goes quiet without an error, including while old and new indexes coexist.

**Scope note:** this addresses the Elasticsearch side that APIM owns. The Logstash saturation described in TT-10329 is a separate, Platform-owned bottleneck; less data reaching it will help, but this PR does not remove it.

## Reviewer notes

- The interesting file is `freemarker/common/mapping/log-body-field.ftl` — everything else in the template trees is a call into it.
- The analyzer is a macro parameter rather than a constant because **es7x maps v4 bodies without `gravitee_body_analyzer`** while every other tree uses it. That inconsistency predates this PR and is deliberately left alone: aligning it would change what payload search matches on existing ES 7 deployments. Worth a separate ticket.
- **Security-sensitive:** touches a configuration key and deployment configuration (`gravitee.yml`). No change to authentication, authorization, input parsing or logging redaction. `index_body: false` reduces what is derivable from the index but does not remove payloads from `_source`; it is not a data-protection control.
- **Compatibility-sensitive:** new configuration key `reporters.elasticsearch.settings.index_body`, and a change to the shipped index templates. Existing indexes are untouched; the new mapping applies from the next daily index or ILM rollover.

## Test plan

- [ ] `mvn -pl gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch test` is green (needs Docker for the Testcontainers cases).
- [ ] On a gateway with an Elasticsearch reporter and a v4 API with payload logging on, let a new daily index roll, then `GET gravitee-v4-log-*/_mapping` and confirm the body fields carry `index_options: docs` and `norms: false`.
- [ ] In the Console Logs view, search a term inside a payload and confirm the results are what they were before this change.
- [ ] Set `reporters.elasticsearch.settings.index_body: false`, restart, let a new index roll. Confirm: the log detail view still shows request and response payloads; the "search in payloads" filter returns nothing rather than an error; a search spanning the day before and the day after returns the older matches without erroring.
- [ ] Confirm existing indexes created before the upgrade still answer payload searches normally.

[APIM-13313]: https://gravitee.atlassian.net/browse/APIM-13313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TT-10329]: https://gravitee.atlassian.net/browse/TT-10329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/config/ReporterConfiguration.java
+++ b/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/config/ReporterConfiguration.java
@@ -52,6 +52,7 @@ public class ReporterConfiguration {
     public static final int DEFAULT_NUMBER_OF_SHARDS = 1;
     public static final int DEFAULT_NUMBER_OF_REPLICAS = 1;
     public static final String DEFAULT_REFRESH_INTERVAL = "5s";
+    public static final boolean DEFAULT_INDEX_BODY = true;
     public static final boolean DEFAULT_ENABLED = true;
     public static final String DEFAULT_INDEX_LIFECYCLE_POLICY_PROPERTY_NAME = "index.lifecycle.name";
     public static final String DEFAULT_INDEX_LIFECYCLE_ROLLOVER_ALIAS_PROPERTY_NAME = "index.lifecycle.rollover_alias";
@@ -179,6 +180,17 @@ public class ReporterConfiguration {
      */
     @Value("${reporters.elasticsearch.settings.refresh_interval:" + DEFAULT_REFRESH_INTERVAL + "}")
     private String refreshInterval = DEFAULT_REFRESH_INTERVAL;
+
+    /**
+     * Settings: index the request, response and message bodies captured by the logging policy.
+     *
+     * <p>Indexing them is what makes the "search in payloads" filter of the Logs UI work, and it is also the
+     * heaviest thing the reporter asks Elasticsearch to do: a gateway capturing 256KB payloads hands the
+     * cluster up to four analysed bodies per request. Turning it off keeps the bodies in {@code _source}, so
+     * the Logs UI still displays them, but no longer matches them.
+     */
+    @Value("${reporters.elasticsearch.settings.index_body:" + DEFAULT_INDEX_BODY + "}")
+    private boolean indexBody = DEFAULT_INDEX_BODY;
 
     @Value("${reporters.elasticsearch.enabled:" + DEFAULT_ENABLED + "}")
     private boolean enabled = DEFAULT_ENABLED;
@@ -390,6 +402,14 @@ public class ReporterConfiguration {
 
     public void setRefreshInterval(String refreshInterval) {
         this.refreshInterval = refreshInterval;
+    }
+
+    public boolean isIndexBody() {
+        return indexBody;
+    }
+
+    public void setIndexBody(boolean indexBody) {
+        this.indexBody = indexBody;
     }
 
     public boolean isEnabled() {
--- a/gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-gateway/src/main/resources/config/gravitee.yml
+++ b/gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-gateway/src/main/resources/config/gravitee.yml
@@ -697,6 +697,10 @@ reporters:
 #      number_of_shards: 1
 #      number_of_replicas: 1
 #      refresh_interval: 5s
+#      index_body: true         # Index request, response and message bodies captured by the logging policy.
+                                # Set to false to stop indexing them: the Logs UI still displays them, but its
+                                # "search in payloads" filter no longer matches. Applies to indexes created
+                                # after the change, not to existing ones.
 #    pipeline:
 #      plugins:
 #        ingest: geoip, user_agent      # geoip and user_agent are enabled by default
--- a/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/README.adoc
+++ b/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/README.adoc
@@ -54,6 +54,28 @@ Then you can activate the ILM feature by adding these settings in the gravitee.y
     index_mode: ilm         # "daily" indexes, suffixed with date. Or "ilm" managed indexes, without date
 ```
 
+=== Body indexing
+
+The logging policy can capture request, response and message bodies, and the reporter indexes them so the
+Logs UI can search inside payloads. That indexing is by far the heaviest work the reporter gives the cluster:
+with `logging.max_size` at 256KB a single v4 log document carries up to four analysed bodies. On a gateway
+running thousands of requests per minute it is the first thing to look at when Elasticsearch falls behind.
+
+Bodies are indexed for prefix search only — no positions, no norms — because that is all the Logs UI query
+needs. A cluster that still cannot keep up can stop indexing them altogether:
+
+```yaml
+  elasticsearch:
+    settings:
+      index_body: false # default: true
+```
+
+Bodies then stay in `_source`, so the Logs UI keeps displaying them in the log detail view, but its
+"search in payloads" filter no longer matches them.
+
+Both are mapping settings, so they only apply to indexes created after the change. Existing indexes keep the
+mapping they were created with until the next daily index or ILM rollover.
+
 == Testing
 
 By default, unit tests are run with a TestContainer version of ElasticSearch 8.5.2, but sometimes it can be useful to run them against other version of ElasticSearch.
--- a/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/mapping/AbstractIndexPreparer.java
+++ b/gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/mapping/AbstractIndexPreparer.java
@@ -105,6 +105,7 @@ public abstract class AbstractIndexPreparer implements IndexPreparer {
         data.put("numberOfShards", this.configuration.getNumberOfShards());
         data.put("numberOfReplicas", this.configuration.getNumberOfReplicas());
         data.put("refreshInterval", this.configuration.getRefreshInterval());
+        data.put("indexBody", this.configuration.isIndexBody());
         data.put("indexLifecyclePolicyPropertyName", this.configuration.getIndexLifecyclePolicyPropertyName());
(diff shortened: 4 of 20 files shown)


gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/config/ReporterConfiguration.java
```diff
@@ -389,6 +401,14 @@ public void setRefreshInterval(String refreshInterval) {
         this.refreshInterval = refreshInterval;
     }
 
+    public boolean isIndexBody() {
```
jourdiw, line 407: Splitting hairs, but it would maybe be better as `hasIndexBody` or `includesIndexBody` 🤔

jourdiw (approved):

PR #19413, opened 2026-08-26
refactor: allow callers to set api ids

a-cordier:
this is required by the automation usecase, where the id is infered from an HRID.

see https://gravitee.atlassian.net/browse/AIAM-233

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/model/factory/ApiModelFactoryTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/model/factory/ApiModelFactoryTest.java
@@ -17,6 +17,7 @@ package io.gravitee.apim.core.api.model.factory;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import fixtures.core.model.NewApiFixtures;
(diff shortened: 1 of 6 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/model/factory/ApiModelFactoryTest.java
```diff
@@ -41,6 +42,26 @@ class ApiModelFactoryTest {
 
     private static final String ENVIRONMENT_ID = "environment-id";
 
+    @Test
+    void fromNewHttpApi_should_generate_an_id_when_none_is_supplied() {
+        var newHttpApi = NewApiFixtures.aProxyApiV4();
+
```
benoitgravitee, line 49: might sound silly but you don't have a hard proof in your test if you don't assert that it is null before the call

benoitgravitee (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/api/model/factory/ApiModelFactoryTest.java
a-cordier: done

PR #19414, opened 2026-08-26
feat(gamma): implement organization level policy studio

Okhelifi:
## Issue

https://gravitee.atlassian.net/browse/FOUND-230

## Description

Adds the organization-level Policy Studio to the gamma platform module, under **Organization → Design → Policy Studio**.

The page hosts `@gravitee/graphene-policy-studio` at `scope="ORGANIZATION"`; everything specific to this host lives in `features/platform-policies`:

- **Flow adapter** between the organization entity (v2 wire format: `path-operator`, `pre`/`post`, `consumers`) and the studio's v4 `Flow` (`selectors`, `request`/`response`, `tags`). Sharding tags are stored by id, as the backend's flow schema does.
- **Save** re-reads the organization and PUTs it whole, since the endpoint replaces the entity: fields the studio does not own survive, and an untouched half (flows when only the flow mode changed) is sent back as stored rather than emptied.
- **Deploy confirmation** before the write, matching Classic Console: saving deploys to every gateway in the organization. Cancelling rejects the studio's save so the changes stay dirty.
- **A banner** stating what platform flows reach: the request and response phases of every API in the organization, native Kafka excepted.

### Permissions

The route and the nav entry are gated on `organization-policies-c`, `-d` or `-u`, and `readOnly` follows `-u`.

Not on `-r`, which Classic Console uses: `OrganizationResource.get()` is annotated `ORGANIZATION_POLICIES` with `CREATE`, `DELETE` and `UPDATE` only, so a reader would open a studio whose very first request answers 403. Serving read-only users takes a backend change to accept `READ` on that GET; it ships in a dedicated PR right after this one, front and back together.

### Shared code

Three behaviour-preserving moves came out of review:

- the policy catalog service (`listPolicies`, `getPolicySchema`, `getPolicyDocumentation`) hoisted to `shared/services/policyPlugins.ts` with the protocol as a parameter, replacing the copy in `shared-policy-groups` and typing `documentation-ext` as the document it returns rather than a string;
- the two v2 flow mappings that were identical on both sides shared with the apim module through `gamma-ui-shared/src/v2Flows`;
- the organization model and service moved to `shared/types/organization.ts` and `shared/services/organization.ts`, since the endpoint serves the whole `OrganizationEntity` rather than a platform-policies projection. The model now lists every field the entity carries, so a full-entity PUT cannot drop one.

## Additional context

- Depends on [gravitee-ui-graphene#448](https://github.com/gravitee-io/gravitee-ui-graphene/pull/448), released: `@gravitee/graphene-core` and `@gravitee/graphene-policy-studio` are pinned to `3.14.0`.
- 2024 tests green on the platform module (32 new), 654 unchanged on the apim module. `tsc --noEmit` clean on both modules and on each commit taken on its own, eslint and prettier clean.
- Not covered by tests: the widened organization policy catalog, which lives in the design system. Worth an eyeball on a running gateway, `request-content-limit` and `circuit-breaker` should now appear.


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
```diff
@@ -592,6 +602,17 @@ export function AppRoutes() {
                                 }
                             />
                             <Route path="entrypoints-and-sharding-tags" element={<EntrypointsGuard />} />
+                            <Route
+                                path="policy-studio"
+                                element={
+                                    <PermissionPageGuard
+                                        permission={ORGANIZATION_POLICIES_READ_PERMISSION}
```
ytvnr: **Worth fixing** — A user holding only `organization-policies-r` passes this guard and the nav check, then gets a 403 on `GET /organizations/{orgId}` and lands on the load-error alert instead of a read-only studio.

`OrganizationResource.get()` is annotated `@Permission(value = ORGANIZATION_POLICIES, acls = { CREATE, DELETE, UPDATE })`, with no `READ`, and `PermissionsFilter` only accepts an acl that is listed. This guard, the nav entry above, and `usePlatformPolicies` are all keyed on `-r`, so the page renders, the GET is refused, `studioData.isError` is true and the page shows "Failed to load the platform policies." `readOnly={!canUpdate}` is then only reachable for the odd role that holds `-c`/`-d` without `-u`. The page spec asserts `readOnly` with `getOrganization` mocked as succeeding, so nothing catches it.

Classic Console guards on `-r` too, so this is a faithful port of an existing quirk rather than something this PR broke, but FOUND-230 now states it as an acceptance criterion ("read-only users see the studio but cannot save"). Worth deciding which contract wins: widen the backend GET to accept `READ` on `ORGANIZATION_POLICIES`, or gate the page on the permission the API actually enforces. Either way, telling a 403 apart from a transport failure in the alert would spare the user a pointless refresh.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/services/organization.ts
```diff
@@ -0,0 +1,32 @@
...
+    return apimFetchJsonOrg<Organization>('');
+}
+
+/**
+ * Full-entity update: every field read from the organization has to be sent back. The resource binds the
+ * body to `UpdateOrganizationEntity`, which is `OrganizationEntity` without `id`; the extra `id` we send
+ * is ignored, the mapper has `FAIL_ON_UNKNOWN_PROPERTIES` disabled.
+ */
+export function updateOrganization(organization: Organization): Promise<Organization> {
+    return apimFetchJsonOrg<Organization>('', { method: 'PUT', body: JSON.stringify(organization) });
```
ytvnr: The PUT resource answers `204`, so this resolves `undefined` at runtime despite being typed `Promise<Organization>`.

`OrganizationResource.update()` returns `Response.ok(updatedOrganization).status(204)`, and `doFetch` short-circuits on `res.status === 204` with `return undefined as T`. Nothing reads the value today — `useSavePlatformPolicies.onSuccess` ignores it — so there is no live bug, but `useSavePlatformPolicies.spec.tsx` mocks `mockUpdateOrganization.mockResolvedValue(STORED_ORGANIZATION)`, which cements a contract the API does not honour for the next caller who trusts the return.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/OrganizationPolicyStudioPage.spec.tsx
```diff
@@ -0,0 +1,214 @@
...
+        expect(screen.queryByText('Deploy the policies?')).toBeNull();
+        expect(mockUpdateOrganization).not.toHaveBeenCalled();
+    });
+
+    it('reports a failure to load the platform policies', async () => {
+        mockGetOrganization.mockRejectedValue(new Error('boom'));
+        renderPage();
+
+        expect(await screen.findByText('Failed to load the platform policies. Please refresh and try again.')).toBeTruthy();
+    });
```
ytvnr, line 235: The save-failure path has no test, while the cancelled save and the failing load both do.

The spec covers cancel (`mockSaveRejection` is `'Deployment cancelled'`) and a rejecting GET, but never a rejecting `updateOrganization`. That leaves `confirmDeployment`'s catch branch untested, including the `error instanceof Error` fallback and the `finally` that clears `pendingDeployment`. A failed deploy is exactly the case where the user must keep their unsaved flows, so it is the branch most worth pinning: a case alongside the cancel test that rejects `updateOrganization` and asserts the studio's save promise rejects, the dialog closes and the error notification fires.

ytvnr (commented):
Organization-level Policy Studio hosted in the platform module, with a v2/v4 flow adapter, a whole-entity save and a deploy confirmation. I traced the round trip of both adapters against `io.gravitee.definition.model.flow.Flow` and the `Organization` interface against `UpdateOrganizationEntity`, and did not find a field that gets lost, renamed or reordered on the way through. Cache invalidation is prefix-correct, the shared-code moves leave the apim module's behaviour unchanged, and the graphene bump is coherent across the root and both module manifests.

The one point worth settling before merge is the permission the page is guarded on: `organization-policies-r` opens the route, but the GET it needs does not accept `READ`, so the read-only case in the ticket cannot hold as written. Details inline.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
Okhelifi: Good catch I'll update the backend and handle readonly mode in the next PR. I'll push it this afternoon. For now I assume the ui is aligned with the backend permissions.

ytvnr: Thanks, the direction you took is the safer one: gating the route on the acls the API actually accepts means nobody reaches a studio that 403s on its first request, and the comment on the constant records why `-r` is excluded. The rewritten nav test pins it from the read-only side, which is the case that was invisible before.

One thing worth keeping visible: with the page now behind `-c`/`-d`/`-u`, the FOUND-230 criterion "read-only users see the studio but cannot save" is not met by this PR, and `readOnly={!canUpdate}` only bites for the rare role holding `-c` or `-d` without `-u`. Since you are handling the backend and the read-only mode in a follow-up, a ticket linked to FOUND-230 would keep the criterion from being closed by accident.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/services/organization.ts
ytvnr: Fixed: `Promise<void>` now matches what a 204 yields, the note above says callers have to re-read the organization for the stored state, and both specs mock `undefined` so they stop describing a response the API never sends.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/OrganizationPolicyStudioPage.spec.tsx
ytvnr: Covered, and wider than what I asked for: the rejection handed back to the studio, the non-`Error` fallback message, and on the hook side that `notify.error` fires while `notify.success` does not. The failed-deploy branch is the one where the user must keep their unsaved flows, so it is good to have it pinned.

ytvnr (approved):
Re-reviewed at `ab27cc9`. The three points from the first pass are addressed and the scope did not drift: same 34 files, the extra lines are the fixes and their tests.

The permission fix went the safer way round. Rather than widening the backend GET, the route and the nav entry now mirror the acls `GET /organizations/{orgId}` actually accepts, so nobody lands on a studio that 403s on load, and `ORGANIZATION_POLICIES_ACCESS_PERMISSIONS` carries a comment saying why `-r` is not among them. `PermissionPageGuard` already supported `anyOf`, so this follows the pattern already used for organization users.

`updateOrganization` is typed `Promise<void>` and both specs mock `undefined`, so nothing cements a response the 204 never sends. The failed-deploy branch is now pinned from three angles: the rejection handed back to the studio, the non-`Error` fallback message, and `notify.error` firing while `notify.success` does not.

Worth tracking rather than blocking: the FOUND-230 criterion about read-only users seeing the studio without being able to save is not met by this PR, since `-r` alone no longer opens the page. That is a deliberate call, tied to the backend change you are doing next, and a linked ticket would keep the criterion from being closed by accident.

Approving.

PR #19421, opened 2026-08-26
fix(authz): index authorization decisions in their own data stream

podlesrafal:
## Issue

https://gravitee.atlassian.net/browse/AUTHZ-126

## Description

Depends on gravitee-io/gravitee-common-elasticsearch#250, **merged and released as 7.1.0**, which this PR now pins.

Authorization decisions were appended to the shared `event-metrics` data stream. That stream already exists on every installation reporting Kafka or native metrics, and Elasticsearch never applies a changed index template to backing indices created earlier. So on any upgraded environment the authz fields are mapped dynamically as `text` the moment the first decision lands, and `text` has no doc_values.

The whole decision read path is down as a result, not degraded:

- the Decisions list fails inside the `sort` clause on `event-id`, before any filter is evaluated, so the bare page with nothing selected already returns 500
- every dashboard widget counts through `value_count` or `terms` on an authz field, so all of them fail too

Live on `apim-master`, 19 Elasticsearch 400s in one hour, all against `.ds-dev_apim_master-event-metrics-2025.12.30-000004`:

```
Fielddata is disabled on [event-id]    7x   list sort
Fielddata is disabled on [status]      3x   aggregation
Fielddata is disabled on [action]      3x   aggregation
Fielddata is disabled on [reasons]     2x   aggregation
Fielddata is disabled on [decision]    2x   aggregation
Fielddata is disabled on [subject-id]  1x   aggregation
Fielddata is disabled on [caller]      1x   aggregation
```

Only string fields appear there. `api-id`, `doc-type`, `org-id`, `env-id` and `gw-id` predate authz and were already keyword; `duration-nanos` and the other numerics are fine because dynamic mapping guesses `long`, which does have doc_values.

The field type of an existing index cannot be changed, so this is not repairable in place, and a reindex would push manual work onto every installation. Decisions therefore get their own data stream, always created fresh from a template that already declares the mapping. New and upgraded installations are both correct with no migration step, and existing environments recover on deploy because the mis-mapped index simply stops being queried.

What changed:

- the authz block moves out of `event-metrics-mapping.ftl` into `authz-decisions-mapping.ftl`, completed with the `BaseEventMetrics` fields the documents also carry
- four `index-template-authz-decisions.ftl`, one per flavour (es7x, es8x, es9x, opensearch), mirroring their siblings
- `AbstractPerTypeIndexNameGenerator` branches `AuthzEventMetrics` **before** `BaseEventMetrics`, since authz extends it and would otherwise fall into the shared branch
- reads move to the new type: `searchAuthzDecisionLogs`, `findAuthzDecisionLog`, and `searchAuthzMeasures` / `searchAuthzFacets` / `searchAuthzTimeSeries`. The three `searchEventMetrics*` methods deliberately stay on `EVENT_METRICS`
- the new mapping carries a `dynamic_templates` strings-to-keyword guard. `event-metrics` was the only template of its family without one, which is what let Elasticsearch guess `text` here in the first place
- the stream gets its own lifecycle setting, `reporters.elasticsearch.lifecycle.policies.authz_decisions`, wired through `ReporterConfiguration` and `AbstractIndexPreparer.getTemplateData()` and documented in `gravitee.yml`. An audit trail wants its own retention, not the traffic-metrics one
- the stream gets its own lifecycle setting, `reporters.elasticsearch.lifecycle.policies.authz_decisions`, wired through `ReporterConfiguration` and `AbstractIndexPreparer.getTemplateData()` and documented in `gravitee.yml`. An audit trail wants its own retention, not the traffic-metrics one

## Additional context

**Tests.** All four existing authz tests in `gravitee-apim-repository-elasticsearch` mock the Elasticsearch client or assert JSON strings, while every other repository in that module has a testcontainers-backed test. The query shape was verified, its execution never was. This PR adds `MetricsElasticsearchRepositoryTest_SearchAuthzDecisionLogs`, running against a real Elasticsearch and covering the sort tiebreaker on a deliberate three-way timestamp tie, paging across that tie, exact-match filters on `decision` / `subject-id` / `action` / `caller` / `status`, the nested `matched-policies` filter and the `reasons` fragment filter.

Worth being explicit that an ordinary integration test would **not** have caught the original bug: it creates its index fresh from the current template, gets keyword and passes, exactly like a fresh environment. The bug class is upgrade behaviour over pre-existing data, and the dedicated data stream removes that class rather than testing around it.

**Verified locally** against `gravitee-common-elasticsearch` built from #250:

```
reporter (common + elasticsearch)   95/95    BUILD SUCCESS
repository-elasticsearch           864/864   BUILD SUCCESS
```

The repository suite matters here because `DatabaseHydrator`, which every test in that module uses, had to learn the new data stream.

**Pre-existing bug found next door, deliberately not fixed here.** `indexLifecyclePolicyEventMetrics` has a config field, a getter, a setter and a documented `event_metrics: my_policy` key in `gravitee.yml`, but `AbstractIndexPreparer.getTemplateData()` never puts it into the template model. Setting that policy is therefore silently a no-op, and has been since the key was added. One line would fix it, but it would start applying policies that users configured and that never took effect, which is a behaviour change that deserves its own ticket rather than a ride along a bug fix.

**Pre-existing bug found next door, deliberately not fixed here.** `indexLifecyclePolicyEventMetrics` has a config field, a getter, a setter and a documented `event_metrics: my_policy` key in `gravitee.yml`, but `AbstractIndexPreparer.getTemplateData()` never puts it into the template model. Setting that policy is silently a no-op and has been since the key was added. One line would fix it, but that would start applying policies users configured and that never took effect, which is a behaviour change deserving its own ticket rather than a ride along a bug fix.

**Known cosmetic point for review.** `base-event-metrics.ftl` hardcodes `"type": "event-metrics"` on its non-bulk branch for all five document types. Authz documents keep emitting it although they now live elsewhere. The field is in no mapping and nothing queries it, so I left the shared macro alone rather than touch four other document types and their fixtures. Say the word if you would rather it were split.

**Data already written.** Decisions indexed before the switch stay in `event-metrics` and stay unreadable. They are unreadable today, so nothing is lost, but the history does not come back on its own. Their `_source` is intact, so a reindex can recover it if anyone wants it.


gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/MetricsElasticsearchRepository.java
```diff
@@ -165,7 +165,7 @@ public LogResponse<AuthzDecisionLog> searchAuthzDecisionLogs(QueryContext queryC
     public Optional<AuthzDecisionLog> findAuthzDecisionLog(QueryContext queryContext, String apiId, String eventId)
         throws AnalyticsException {
         var clusters = ClusterUtils.extractClusterIndexPrefixes(configuration);
-        var index = this.indexNameGenerator.getWildcardIndexName(queryContext.placeholder(), Type.EVENT_METRICS, clusters);
+        var index = this.indexNameGenerator.getWildcardIndexName(queryContext.placeholder(), Type.AUTHZ_DECISIONS, clusters);
```
sclarkamlab, line 168: `MetricsElasticsearchRepositoryTest_FindAuthzDecisionLog:87` still does `verify(indexNameGenerator).getWildcardIndexName(any(), eq(Type.EVENT_METRICS), any())`, which contradicts this line. Line 61 stubs the generator to return `"gravitee-event-metrics-*"` for any type and line 98 asserts that string, so those two won't guard anything once 87 is fixed. Worth updating all three.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/log/MetricsElasticsearchRepositorySearchAuthzDecisionLogsTest.java
```diff
@@ -0,0 +1,112 @@
...
+import io.gravitee.repository.elasticsearch.TimeProvider;
+import io.gravitee.repository.log.v4.model.authz.AuthzDecisionLog;
+import io.gravitee.repository.log.v4.model.authz.AuthzDecisionLogQuery;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * Runs the decision search against a real Elasticsearch, so the query is checked against the mapping
+ * the index template actually produces. Sorting, exact-match filters and the nested policy filter all
```
sclarkamlab, line 32: This test never runs. `gravitee-parent` pins the surefire includes to `**/Test*.java`, `**/*Test.java`, `**/*TestCase.java` and `**/*Suite.java`, and this filename matches none of them. Proof is that CI is green on this head even though `MetricsElasticsearchRepositoryTest_FindAuthzDecisionLog:87` asserts the old enum. Rename it to `MetricsElasticsearchRepositorySearchAuthzDecisionLogsTest` and it'll actually execute.

gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/resources/freemarker/es8x/mapping/index-template-authz-decisions.ftl
```diff
@@ -0,0 +1,21 @@
+<#ftl output_format="JSON">
+{
+    "index_patterns": ["${indexName}*"],
+    "data_stream": {},
+    "template": {
+        "settings": {
+            <#if indexLifecyclePolicyAuthzDecisions?has_content>"${indexLifecyclePolicyPropertyName?json_string}": "${indexLifecyclePolicyAuthzDecisions?json_string}",</#if>
```
sclarkamlab, line 7: This is the first data stream template that will really render these two settings, since `getTemplateData` never put `indexLifecyclePolicyEventMetrics` into the model so the event-metrics block is dead. Elastic says don't set `index.lifecycle.rollover_alias` on a data stream, rollover comes from the stream itself. Worth trying against a real cluster with `reporters.elasticsearch.lifecycle.policies.authz_decisions` set, `IndexPreparerIntegrationTest` doesn't set it so nothing covers this.

sclarkamlab (commented):
The split looks right and the mapping move is clean. Main things are the stale test assertion and the fact that neither authz test class actually runs.

Few smaller ones with no line in the diff to hang off. `SearchAuthzDecisionLogsQueryAdapter:72` and `:114` still describe the shared event-metrics stream in their comments, and the `doc-type: authz` filter at 116 is redundant now the index only holds decisions. `IndexTemplateTest.es_trees_and_lifecycle_types` excludes `EVENT_METRICS` because its policy isn't in the model, that reason doesn't apply to `AUTHZ_DECISIONS` so it should go in the list. And `searchAuthzMeasures`, `searchAuthzFacets` and `searchAuthzTimeSeries` all changed index with no test at all.

gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/mapping/AbstractIndexPreparer.java
```diff
@@ -111,6 +111,7 @@ protected Map<String, Object> getTemplateData() {
         data.put("indexLifecyclePolicyMonitor", this.configuration.getIndexLifecyclePolicyMonitor());
         data.put("indexLifecyclePolicyRequest", this.configuration.getIndexLifecyclePolicyRequest());
         data.put("indexLifecyclePolicyLog", this.configuration.getIndexLifecyclePolicyLog());
+        data.put("indexLifecyclePolicyAuthzDecisions", this.configuration.getIndexLifecyclePolicyAuthzDecisions());
```
sclarkamlab, line 115: While you're here, `event_metrics` is still missing from this map, so `reporters.elasticsearch.lifecycle.policies.event_metrics` is dead config. One more line and `IndexTemplateTest` can drop its exclusion comment.

podlesrafal:
Thanks, the two you led with were both real and the second one is worse than it looked. All points addressed, pushed.

**Neither authz test class ran.** Confirmed exactly as you described: `gravitee-parent` pins the includes to `Test*`/`*Test`/`*TestCase`/`*Suite` and neither filename matches. Renamed both:

- `MetricsElasticsearchRepositoryTest_SearchAuthzDecisionLogs` -> `MetricsElasticsearchRepositorySearchAuthzDecisionLogsTest`
- `MetricsElasticsearchRepositoryTest_FindAuthzDecisionLog` -> `MetricsElasticsearchRepositoryFindAuthzDecisionLogTest`

The module went from 864 to 877 tests, and the rename is what turned `:87` red. Worth stating plainly that I had reported this suite as green as evidence the change was safe, and it never touched either file.

**Stale assertion.** Fixed all three sites you pointed at, not just the one: the stub at `:61`, the assertion at `:87`, the literal at `:98`, plus the method name and the comment.

**rollover_alias.** You were right that mine is the first template to really render it. Removed from all four new templates, and from the four event-metrics ones for the same reason, since wiring `event_metrics` below would otherwise have started setting it there too. `IndexTemplateTest` now asserts no data-stream template carries it, and `IndexPreparerIntegrationTest` sets both policies so a real cluster has to accept the rendered body. That is what proves it, and it passes.

**Escaping.** `?json_string` added to the OpenSearch variants of both data-stream templates. Covered by a classpath test across all four families, so OpenSearch is held to it despite having no preparer to drive.

**`event_metrics` dead config.** Added, and `IndexTemplateTest` dropped its exclusion comment as you suggested. Both data streams now have a rendering test.

**`doc-type: authz`.** Removed from the two decision-log adapters and from `AuthzMeasuresQueryAdapter`, which is the single funnel for measures, facets and time series. The two tests that existed only to assert that filter are gone, and the index-dependent assertions shifted. Comments at `:72` and `:114` rewritten, and the `AuthzDecisionLogsCrudService` javadoc too.

**No coverage on the three analytics methods.** Added `AnalyticsElasticsearchRepositoryTest.Engine.AuthzDecisions`, running against a real Elasticsearch over the seeded decisions: a count through `searchAuthzMeasures`, a terms facet through `searchAuthzFacets` and a bucketed series through `searchAuthzTimeSeries`. The facet one is the interesting one, since a terms aggregation is exactly what needed doc_values and what was failing in production.

Also in: `ignore_above` on `subject-id` and `resource-id`, since those come straight off the PEP payload. I left `reasons` unbounded on purpose: its filter is a `*fragment*` wildcard that needs the term indexed, so `ignore_above` would make the filter quietly return nothing instead of matching. If a reason ever approaches the term limit that is a PDP bug worth seeing, not one to hide. `DatabaseHydrator` now derives the data streams from `Type.isDataStream()` instead of hardcoding the pair. The restating javadoc Aikido flagged is gone, and the four-behaviour test is split into four.

One thing outside this PR: `MetricsElasticsearchRepositoryTest_FindNativeApiMetrics` and `..._SearchNativeApiMetrics` have the same naming and have never run either. Filed as AUTHZ-127 rather than renamed here, because turning them on may well surface assertions nobody has checked since they were written, and that deserves its own change.

Green locally with validation on: reporter 124/124, repository-elasticsearch 877/877.

podlesrafal:
Follow-up: I pulled the two native-metrics classes into this PR after all.

`MetricsElasticsearchRepositoryTest_FindNativeApiMetrics` and `..._SearchNativeApiMetrics` were skipped by the same naming rule. I renamed them and ran them before deciding what to do, and all 8 pass, so there is nothing hidden behind them and no reason to keep the rename in a separate change. The module is now at 885 tests and no `*Test_*` file is left in it.

Deleting them was suggested as the cheaper option and I did not take it: they genuinely cover `findNativeApiMetrics` and `searchNativeApiMetrics` against a real Elasticsearch, so removing them would have swapped silent missing coverage for outright missing coverage. AUTHZ-127 is closed as fixed here.

podlesrafal:
Correction to my earlier reply, before you spend time re-checking it.

I said the unbounded-keyword point was handled, but I had only bounded `subject-id` and `resource-id` and quietly dropped the rest of the list. Now closed by a rule rather than a selection: every field whose value the PEP supplies is bounded, which is `subject-type`, `subject-id`, `action`, `resource-type` and `resource-id`.

The keywords still unbounded are deliberate. `caller`, `status`, `decision`, `operation`, `search-type` and `error-type` are fixed vocabularies the PDP writes, and `event-id`, `request-id`, `batch-id` and `target-pdp-id` are generated ids. Nothing a caller controls reaches any of them.

`reasons` also stays unbounded, for the reason above: its filter is a `*fragment*` wildcard that needs the term indexed.

One other place I did not follow the suggestion literally: `AUTHZ_DECISIONS` did not go into `es_trees_and_lifecycle_types`, because that source asserts a rollover alias is present and a data stream must not have one. It has its own source, `es_trees_and_data_stream_types`, with the inverted assertion.

Still green: reporter 124/124, repository-elasticsearch 885/885.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/MetricsElasticsearchRepository.java
podlesrafal: Fixed in dd26f59. All three spots now name the new stream: the generator stub returns `gravitee-authz-decisions-*` (line 61), the verification asserts `eq(Type.AUTHZ_DECISIONS)` (line 87), and the query assertion matches the same string (line 98). You were right that the three guarded nothing together.

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/log/MetricsElasticsearchRepositorySearchAuthzDecisionLogsTest.java
podlesrafal: Good catch, and it went further than these two: four classes in this module were dark under the `Test*`/`*Test`/`*TestCase`/`*Suite` includes. Renamed in dd26f59 to `MetricsElasticsearchRepositorySearchAuthzDecisionLogsTest`, `MetricsElasticsearchRepositoryFindAuthzDecisionLogTest`, plus the two native-metrics ones that were skipped for the same reason. Running the renamed find test is exactly what surfaced the stale `EVENT_METRICS` assertion you flagged above.

gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/java/io/gravitee/apim/reporter/elasticsearch/mapping/AbstractIndexPreparer.java
podlesrafal: Added in dd26f59, together with `indexLifecyclePolicyAuthzDecisions`, so `reporters.elasticsearch.lifecycle.policies.event_metrics` stops being dead config. `IndexTemplateTest` no longer needs the exclusion: `es_trees_and_data_stream_types` now parameterises both `EVENT_METRICS` and `AUTHZ_DECISIONS`.

gravitee-apim-reporter/gravitee-apim-reporter-elasticsearch/src/main/resources/freemarker/es8x/mapping/index-template-authz-decisions.ftl
podlesrafal: Both parts are covered now. No data-stream template sets a rollover alias any more, in any of the four families, and two tests hold that: `should_not_set_a_rollover_alias_on_a_data_stream` over the rendered es7x/8x/9x output, and the classpath read that also covers opensearch, which has no preparer to drive here.

On the real cluster: `IndexPreparerIntegrationTest` now sets both policies, so the lifecycle block does render in the PUT that OpenSearch accepts. Worth recording why the setters are enough there, since it looked wrong to me too at first: the `@Value` defaults on these fields are `#{null}`, and `AutowiredFieldElement.inject` only assigns when the resolved value is non-null, so field injection on the `@Bean` instance leaves the setter values in place. Checked with a control field in the same context, `index` came back as the injected `gravitee` while both policies kept what the setters put there.

podlesrafal:
Picking up the four items from your summary that had no line to hang off, all in dd26f59:

- `SearchAuthzDecisionLogsQueryAdapter:72` no longer describes the shared stream, it now says the search spans every backing index of the stream, which is still the reason `ignore_unmapped` has to be there.
- The redundant `doc-type: authz` filter is gone, along with the comment at 114 that explained it. `AuthzDecisionLogFields.DOC_TYPE` and the two orphaned `DOC_TYPE_*` constants in `AuthzFieldResolver` went with it.
- `IndexTemplateTest` dropped the `EVENT_METRICS` exclusion, since its reason disappeared with the missing template-model entry. Both data streams are parameterised now.
- `searchAuthzMeasures`, `searchAuthzFacets` and `searchAuthzTimeSeries` have integration tests against a real Elasticsearch: the count asserts 6 decisions, the facet asserts the action buckets read=4, write=1, delete=1 with their measures, and the time series asserts the counts sum to 6 across the hourly buckets. The time series deliberately asserts the sum rather than a bucket shape, because `TimeProvider` backdates from `Instant.now()` and the fixture spans three minutes, so which side of an hourly boundary each document lands on moves with the clock. Verified they fail when the index is put back to `EVENT_METRICS`, which the earlier `isNotNull()` versions did not.

sclarkamlab (approved):

PR #19423, opened 2026-08-26
feat(gamma): add Roles management to the platform module

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/FOUND-75

## Summary

Adds Roles management to `gravitee-gamma-module-platform` (Team → Roles), porting the Classic console's org-settings-roles feature: list roles by scope, create/edit with the CRUD permission matrix, lock system roles read-only, delete, and manage members for Organization-scope roles.

Role changes made here invalidate the same query keys the User detail page's role dropdowns and the Groups sheets already read, so a role added/removed/edited shows up there without a manual refresh.

## What changed

- New `features/roles/` — types, services (`/configuration/rolescopes/...`), permission
  matrix utilities, license gating, hooks for all 9 role scopes, CRUD, and memberships.
- New pages: `RolesPage`, `RoleFormPage`, `RoleMembersPage`.
- Route/nav wiring: `roles`, `roles/:roleScope`, `roles/:roleScope/:roleName`,
  `roles/:roleScope/:roleName/members` under Team, guarded by `organization-role-r`/`-u`.
- `RolesTableOfContents` — sticky scope jump-nav with scroll-spy highlighting, matching
  Classic's `<gio-table-of-contents>`.
- `ConsoleSettings` extended with `management.systemRoleEdition` (same `/console`
  endpoint, no backend change) to preserve the ADMIN-role read-only exception.

Tested and verified : 
[Roles-Platform-vs-Classic-QA-Report.pdf](https://github.com/user-attachments/files/31486882/Roles-Platform-vs-Classic-QA-Report.pdf)
[report.html](https://github.com/user-attachments/files/31688866/report.html)


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
```diff
@@ -495,6 +510,52 @@ export function AppRoutes() {
...
+                                            <RoleFormPage />
+                                        </PermissionPageGuard>
+                                    }
+                                />
+                                <Route
+                                    path=":roleScope/:roleName/members"
+                                    element={
+                                        <PermissionPageGuard
+                                            anyOf={[ORGANIZATION_ROLE_UPDATE_PERMISSION]}
+                                            unauthorizedTo="../../applications"
```
arpzilla-dev: Left as `../../applications`. These role sub-routes (`:roleScope`, `:roleScope/:roleName`, and `:roleScope/:roleName/members`) are siblings under `path="roles"`, not nested Route trees — React Router resolves `..` against that hierarchy, not URL segments. The same depth is used for group/dictionary detail. Added a regression test that an unauthorized members visit lands on Applications, not a nested roles path.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolePermissionsTable.tsx
```diff
@@ -0,0 +1,120 @@
...
+        return <p className="text-sm text-muted-foreground">No permission</p>;
+    }
+
+    return (
+        <Table id="rolePermissionsTable" aria-label="Role CRUD permissions table">
+            <TableHeader>
+                <TableRow>
+                    <TableHead scope="col">Permission</TableHead>
+                    {ROLE_RIGHTS_CONFIG.map(right => {
+                        const state = computeSelectAllState(value, right.key, permissionNames);
```
arpzilla-dev: Fixed. Select-all state now uses the same toggleable permission set as `toggleAll` (moved-to-organization rows are excluded), so the header checkbox can show fully selected.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/AddRoleMembersSheet.tsx
```diff
@@ -0,0 +1,186 @@
...
+import { useDeferredValue, useEffect, useState } from 'react';
+
+import { searchUsers } from '../../../shared/services/userSearch';
+import type { SearchableUser } from '../../../shared/types/userSearch';
+import { isSameUser } from '../../../shared/utils/userSearch';
+import { STANDARD_SHEET_WIDTH } from '../../applications/components/sheetLayout';
+import type { RoleMembershipListItem } from '../types/role';
+import { roleKeys } from '../utils/queryKeys';
+
+export function AddRoleMembersSheet({
```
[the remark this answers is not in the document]
arpzilla-dev: Leaving as one sheet component. Length is in line with other Graphene search-and-select sheets in this module (add members, add users); splitting it would be a drive-by refactor without a behavior change.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleMembersPage.tsx
```diff
@@ -0,0 +1,146 @@
...
+    function goBack() {
+        navigate('..');
+    }
+
+    async function handleAdd(users: SearchableUser[]) {
+        try {
+            await addMutation.mutateAsync({
+                scope,
+                roleName: roleName!,
+                users: users.map(user => ({ id: user.id ?? user.reference, reference: user.reference })),
```
arpzilla-dev: **[Score: 90/100]** 🚨 Adding a member from an external identity provider will fail.

`SearchableUser.id` is intentionally nullable (Classic documents it: *"id can be null if user comes from an LDAP"*). Falling back to `user.reference` forces the wrong backend branch in `MembershipServiceImpl.findUserFromMembershipMember`: `userService.findById(<idp-reference>)` instead of resolving through `identityService` and auto-provisioning.

Classic sends `{ id: user.id, reference }` with `id` absent. The sibling feature in this module already does it correctly (`useGroupAddMembersForm.ts`: `...(user.id ? { id: user.id } : {})`).

**Fix:** omit `id` when it is null/undefined, and add a spec with `id: null` — no current test covers it.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleFormPage.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const createMutation = useCreateRole();
+    const updateMutation = useUpdateRole();
+
+    // Mirrors organization-settings-routing.module.ts: only the create route (role/:roleScope) requires the
+    // license and redirects back to the list; the edit and members routes do not.
+    if (!isEditMode && !hasCustomRolesLicense) {
+        return <Navigate to=".." replace />;
+    }
+
+    if (isRoleLoading || arePermissionsLoading) {
```
arpzilla-dev, line 60: **[Score: 82/100]** ⚠️ Failed role fetch in edit mode renders a create form.

This gates on `isRoleLoading` only, never `isError`. With `role === undefined`, `RoleForm` computes `isEditMode = Boolean(role) = false`, so `/roles/API/CUSTOM` shows an editable name field and a **Create role** button. The user only learns something is wrong on submit.

The guard at lines 71–75 prevents the bad write, but the fix belongs here: surface `useRole`'s error state and render an error panel instead of a form. Your own spec (`RoleFormPage.spec.tsx`) currently clicks a button named `'Create role'` while sitting on the edit route, which locks this gap in.

arpzilla-dev (commented):
Inline review of FOUND-75 Roles port (scores ≥ 30). One blocker; High band is mostly unmodelled error / post-mutation state.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/hooks/useRoleMutations.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * Beyond this feature's own role list, the user detail page's role-assignment dropdowns
+ * (organizationUserKeys.organizationRoles()/.environmentRoles()) and the group sheets' role selects
+ * (groupKeys.roles(scope)) already read from the same `/configuration/rolescopes/{scope}/roles` endpoint.
+ * Invalidating their existing cache keys here — rather than touching those features — is what makes a role
+ * change made in this feature show up there without a refresh.
+ */
+async function invalidateRoleQueries(queryClient: QueryClient, scope: RoleScope): Promise<void> {
```
arpzilla-dev: **[Score: 78/100]** ⚠️ `roleKeys.detail` is never invalidated, so a saved edit looks unsaved.

This invalidates `listByScope` and the cross-feature catalogs, but not `roleKeys.detail(scope, name)`. Classic re-runs `ngOnInit()` after update, refetching and resetting the save bar.

Here `RoleForm`'s `initialForm` is captured once in a `useState` initializer, so after a successful save `isDirty` stays `true` and Save never returns to disabled. `useRoleMutations.spec.tsx` asserts `invalidateQueries` was called exactly once, which pins the gap.

**Fix:** also invalidate `roleKeys.detail(scope, name)`. The refetch remounts the form through the loading gate and resets dirty state.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/hooks/useRoles.ts
```diff
@@ -0,0 +1,68 @@
...
+            queryFn: () => listRolesByScope(scope),
+        })),
+    });
+
+    const groups = useMemo<RolesByScopeGroup[]>(
+        () =>
+            ROLE_SCOPES.map((scope, index) => ({
+                scope,
+                label: ROLE_SCOPE_LABELS[scope],
+                roles: results[index]?.data ?? [],
```
arpzilla-dev, line 48: **[Score: 75/100]** ⚠️ Every load failure renders as "empty" instead of as an error.

`results[index]?.data ?? []` drops `isError` entirely: a 403 or 500 on a scope renders **No role**, indistinguishable from a scope that genuinely has none. Same shape in `useRole` and `useRoleMemberships` (empty members table).

On a permissions screen, "you have no roles" and "we couldn't load your roles" are very different statements. Surface `isError` per group (and on detail/memberships) and render an error state.

Classic is no defense here: its unguarded `combineLatest` hangs on the loading state, which is a different bug, not a precedent.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
```diff
@@ -540,6 +552,49 @@ export function AppRoutes() {
...
+                                            <RolesPage />
+                                        </PermissionPageGuard>
+                                    }
+                                />
+                                <Route
+                                    path=":roleScope"
+                                    element={
+                                        <PermissionPageGuard
+                                            anyOf={[ORGANIZATION_ROLE_UPDATE_PERMISSION]}
+                                            unauthorizedTo="../../applications"
```
arpzilla-dev: **[Score: 72/100]** ⚠️ Read-only users are ejected from Roles by an ordinary row click.

Classic's roles routes omit `unauthorizedFallbackTo`, so the Angular guard cancels navigation and leaves the user on the list. `PermissionPageGuard` always navigates on denial, and `RolesByScopeSection` renders every role row as an unconditional `<button>`.

A user with `organization-role-r` but not `-u` (the auditor persona — `user/password` and `api1/api1` in the HTML report) clicking any role name is thrown onto Applications. The same `unauthorizedTo` is repeated on the edit and members routes below.

Direct-URL bounce is a reasonable access denial; reached by an in-app click it reads as the app losing your place.

**Fix:** either disable the row when `-u` is missing, or point `unauthorizedTo` at the roles list (`..`) instead of `../../applications`.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolePermissionsTable.tsx
```diff
@@ -0,0 +1,152 @@
...
+                                            disabled={disabled}
+                                            onCheckedChange={checked => toggleAll(right.key, checked === true)}
+                                            aria-label={
+                                                isIndeterminate
+                                                    ? `Some ${right.label} permissions selected — select all ${right.label}`
+                                                    : `${state === true ? 'Deselect' : 'Select'} all ${right.label}`
+                                            }
+                                            className={
+                                                isIndeterminate
+                                                    ? 'border-primary bg-background data-[state=indeterminate]:bg-background [&_svg]:invisible'
```
arpzilla-dev, line 110: **[Score: 65/100]** 💡 Graphene `Checkbox` internals overridden to fake an indeterminate state.

`[&_svg]:invisible` plus `data-[state=indeterminate]:bg-background` plus an absolutely-positioned dash. `graphene.md` lists overriding Graphene internals with nested selectors as a mistake to avoid, and says not to recreate what a variant should do.

There is no `[&_svg]` precedent anywhere else in this module. The only other indeterminate handling (`NotificationHookCategorySection`) sidesteps it.

The diagnosis is right and the UX fix is real — which is why it belongs upstream in `graphene-core` as an `indeterminate` visual state, not patched per-feature. File it against Graphene and keep this as a referenced temporary workaround.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleFormPage.tsx
```diff
@@ -0,0 +1,112 @@
...
+        return <Skeleton className="h-64 w-full rounded-md" />;
+    }
+
+    // EXPLORER/AI_WORKSPACE have no permissions-by-scope entry (Angular's getPermissionsByScope throws for
+    // them); treating that as "no permissions to grant" keeps the rest of the form usable instead of crashing.
+    const permissionNames = isPermissionEligibleScope(scope) ? (permissionsByScopes?.[scope] ?? []) : [];
```
arpzilla-dev, line 75: **[Score: 60/100]** 💡 EXPLORER / AI_WORKSPACE role creation is new capability, not parity.

Classic's `RoleService.getPermissionsByScope` *throws* for these scopes, so the create form never renders. Here an empty matrix renders and a permission-less role can be created.

The QA report flags this as pending product confirmation. That answer should land before merge, because this is capability the reference deliberately blocks rather than a port.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx
```diff
@@ -0,0 +1,201 @@
...
+            <CardContent>
+                {group.isLoading ? (
+                    <div className="space-y-2">
+                        <Skeleton className="h-10 w-full rounded-md" />
+                        <Skeleton className="h-10 w-full rounded-md" />
+                    </div>
+                ) : group.roles.length === 0 ? (
+                    <p className="text-sm text-muted-foreground">No role</p>
+                ) : (
+                    <ul className="divide-y">
```
arpzilla-dev, line 147: **[Score: 55/100]** 💡 Role list hand-rolls a `<ul>` where the design system mandates `DataTable`, and loses search.

`graphene.md`: use `DataTable` for any list that could exceed 5 rows. Role lists routinely do (system + custom per scope). The port is faithful to Classic's `mat-list`, and Graphene's "minimum 3 data columns" cuts the other way, so this is a genuine tension — but the practical cost is that a 9-section page has no search at all, while the members page right next to it does.

Worth stating in the PR as a deliberate Classic-faithful choice if you keep the list.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/hooks/useRoleMemberships.ts
```diff
@@ -0,0 +1,59 @@
...
+    roleName: string;
+    users: RoleMembershipPayload[];
+}
+
+/** Mirrors OrgSettingsRoleMembersComponent.onAddMemberClicked: one membership call per selected user, in parallel. */
+export function useAddRoleMembers() {
+    const queryClient = useQueryClient();
+    return useMutation({
+        mutationFn: async ({ scope, roleName, users }: AddRoleMembersInput) => {
+            await Promise.all(users.map(user => createRoleMembership(scope, roleName, user)));
```
arpzilla-dev: **[Score: 55/100]** 💡 Partial failure on multi-member add is reported as total failure.

`Promise.all` means 3 of 5 succeeding surfaces only a generic error toast, with three memberships silently created and no list refresh (`onSuccess` never runs).

Classic's `combineLatest` shares the flaw, so this is faithful — `Promise.allSettled` with a succeeded/failed count in the message is a cheap upgrade, and you can still invalidate after any success.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx
```diff
@@ -0,0 +1,201 @@
...
+                        {group.roles.map(role => (
+                            <li key={role.name} className="flex items-center justify-between gap-4 py-3">
+                                <button
```
arpzilla-dev: **[Score: 50/100]** 💡 Full-row-click and a raw `<button>` instead of name-as-link.

`graphene.md`: the primary identifier is a clickable link with `font-medium hover:underline`. **No full-row-click.** Here the whole row — icon, name, badges, description — is one native `<button>` with hand-written classes, which also collides with "use `Button` for actions" and "don't recreate variants with custom classes".

This also compounds the read-only eject issue: shrinking the hit target to the name (and gating it on `organization-role-u`) shrinks the trap.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleFormPage.tsx
```diff
@@ -0,0 +1,112 @@
...
+import type { RoleScope } from '../features/roles/types/role';
+import { isPermissionEligibleScope, isRoleReadOnly } from '../features/roles/utils/rolePermissions';
+import { ROLE_SCOPE_LABELS } from '../features/roles/utils/roleScopeLabels';
+import { useConsoleSettings } from '../shared/console-settings';
+import { notify } from '../shared/notify';
+
+export function RoleFormPage() {
+    const { roleScope, roleName } = useParams<{ roleScope: string; roleName?: string }>();
+    const navigate = useNavigate();
+    const scope = roleScope as RoleScope;
```
arpzilla-dev: **[Score: 48/100]** 💡 Route params are cast, never validated.

`roleScope as RoleScope` accepts anything. `/roles/FOO` renders "Create role in the **undefined** scope" and POSTs to `/rolescopes/FOO/roles`. Same pattern in `RoleMembersPage`.

`/roles/API/USER/members` also renders a full page with a working-looking **Add a member** button, though the backend returns `[]` for non-ORGANIZATION scopes and `400 Can't determine context` on add.

Classic shares both holes. A scope-validity check (against `ROLE_SCOPES`) redirecting to the list would close them, and avoids the `as` cast Graphene's TypeScript rules discourage.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/services/roles.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { apimFetchJsonOrg } from '../../../shared/api/apimClient';
+import type { NewRolePayload, PermissionsByScopes, Role, RoleMembershipListItem, RoleMembershipPayload, RoleScope } from '../types/role';
+
+export async function listRolesByScope(scope: RoleScope): Promise<Role[]> {
```
arpzilla-dev, line 32: **[Score: 45/100]** 💡 Classic's defensive scope normalization was dropped.

Classic `RoleService.list`/`get` both map `scope: role.scope.toUpperCase()`. This service does not, and `isRoleReadOnly` compares the *server's* `role.scope` against `'ORGANIZATION'` to decide the ADMIN lock.

Harmless today (`RoleEntity.scope` is a Java enum, and `RoleServiceImpl.isReserved` enforces the rule server-side), but `RoleFormPage` already holds the authoritative scope from the URL — passing that into `isRoleReadOnly` removes the coupling for free. Restoring `.toUpperCase()` on list/get would also match Classic.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RoleMembersTable.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+function buildColumns({
+    canManage,
+    onDeleteMember,
+}: {
+    canManage: boolean;
+    onDeleteMember: (member: RoleMembershipListItem) => void;
+}): DataTableProps<RoleMembershipListItem>['columns'] {
+    const columns: DataTableProps<RoleMembershipListItem>['columns'] = [
+        {
```
arpzilla-dev: **[Score: 42/100]** 💡 `DataTable` used for a single-column table.

`graphene.md`: "Minimum 3 data columns (excluding actions). If an entity has fewer than 3 meaningful attributes, use a simpler component (Item list, card grid) instead." Members has exactly one (`displayName`) plus actions, so sorting and column-visibility affordances are dead weight.

Note the inversion with the roles list: a plain `<ul>` where the DS wants `DataTable`, and `DataTable` where the DS wants a plain list. Both are Classic-faithful — worth stating as the deliberate tie-breaker if you keep it.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RoleMembersTable.tsx
```diff
@@ -0,0 +1,135 @@
...
+                        title="No member matches your search."
+                        description="Try adjusting your search terms."
+                        action={
+                            <Button size="sm" variant="outline" onClick={() => handleSearchChange('')}>
+                                Clear search
+                            </Button>
+                        }
+                    />
+                ) : (
+                    <DataTableEmptyState variant="first-use" title="No member" description="No one has been added to this role yet." />
```
arpzilla-dev: **[Score: 35/100]** 🎨 `variant="first-use"` is passed through `emptyMessage`, but Graphene says first-use renders *instead of* `DataTable`, wrapped in `rounded-lg border`.

No `icon` prop on either empty-state variant, and `RoleMembersPage` always shows **Add a member** where Graphene gives the empty state the CTA during first-use (`graphene.md` header-button coordination).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolePermissionsTable.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+function computeSelectAllState(
+    value: RolePermissionsForm,
+    right: RoleRight,
+    permissionNames: readonly string[],
+    scope: RoleScope,
+): boolean | 'indeterminate' {
+    // Mirror toggleAll's exclusion below: a permission moved to ORGANIZATION scope can never be checked here,
+    // so it must not count against "all selected" either — otherwise select-all could never reach `true`.
+    const manageablePermissionNames = permissionNames.filter(permission => !isPermissionMovedToOrganizationScope(scope, permission));
```
arpzilla-dev, line 36: **[Score: 30/100]** 🎨 Select-all counts only manageable permissions; Classic counts all controls (`org-settings-role.component.ts`), which is why its ENVIRONMENT column can never reach "checked" while TAG/TENANT/ENTRYPOINT are disabled.

Your version is a genuine fix (and the comment explains it well), but the PR body presents this as parity. Call it out as an intentional improvement so reviewers don't treat the mismatch as a regression.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesTableOfContents.tsx
```diff
@@ -0,0 +1,97 @@
...
+                    if (entry.isIntersecting) {
+                        intersecting.add(scope);
+                    } else {
+                        intersecting.delete(scope);
+                    }
+                });
+                const topMost = sections.find(section => intersecting.has(section.scope));
+                if (topMost) setActiveScope(topMost.scope);
+            },
+            { rootMargin: '-96px 0px -70% 0px', threshold: 0 },
```
arpzilla-dev: **[Score: 28/100]** 🎨 `rootMargin: '-96px 0px -70% 0px'` hardcodes the shell header height.

`graphene.md` exposes an `h-topnav` layout token for this. If the shell header resizes, scroll-spy highlighting will drift without anyone touching this file. Prefer the token (or a CSS variable derived from it) over a magic pixel value.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx
```diff
@@ -0,0 +1,201 @@
...
+}
+
+/** Card-list row actions: a lone icon button for a single action, a dropdown once there are two (mirrors DirectMembersTable.tsx's row-actions pattern). */
+function RoleRowActions({ role, scope, canSeeMembers, canDeleteRole, onViewMembers, onDeleteRole }: Readonly<RoleRowActionsProps>) {
+    if (canSeeMembers && canDeleteRole) {
+        return (
+            <DropdownMenu>
+                <DropdownMenuTrigger asChild>
+                    <Button variant="ghost" size="icon" aria-label={`Actions for ${role.name}`}>
+                        <MoreHorizontalIcon className="size-4" aria-hidden />
```
arpzilla-dev: **[Score: 25/100]** 🎨 `MoreHorizontalIcon` for row actions; Graphene says `MoreVerticalIcon` and reserves horizontal for toolbars and card headers.

In fairness this module is already split (~8 vertical / ~9 horizontal), and `DirectMembersTable.tsx` (cited as this file's model) is on the horizontal side — so this copies an existing inconsistency rather than creating one. The 1-action / 2-action dropdown split and destructive-last-after-separator are correct.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleMembersPage.tsx
```diff
@@ -0,0 +1,146 @@
...
+            <div>
+                <Button variant="ghost" size="sm" className="-ml-2 mb-3 text-muted-foreground" onClick={goBack}>
+                    <ArrowLeftIcon className="size-4" aria-hidden />
+                    Back to roles
+                </Button>
+                <h1 className="text-2xl font-semibold tracking-tight">
+                    {ROLE_SCOPE_LABELS[scope]} - {roleName}
+                </h1>
+                <p className="mt-1 text-sm text-muted-foreground">
+                    List all the users having the role {roleName} in the {scope} scope.
```
arpzilla-dev: **[Score: 22/100]** 🎨 Members heading formatting is inconsistent with itself.

The title uses the pretty label (`ROLE_SCOPE_LABELS[scope]` → "API Product") while the subtitle right below uses the raw scope (`{scope}` → "API_PRODUCT"). Classic used `titlecase` for both. Use `ROLE_SCOPE_LABELS[scope]` here as well.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/AddRoleMembersSheet.tsx
```diff
@@ -0,0 +1,179 @@
...
+    isAdding: boolean;
+}>) {
+    const [search, setSearch] = useState('');
+    const [selectedUsers, setSelectedUsers] = useState<SearchableUser[]>([]);
+
+    const deferredQuery = useDeferredValue(search);
+    const { data: results, isFetching } = useQuery({
+        queryKey: roleKeys.userSearch(deferredQuery),
+        queryFn: () => searchUsers(deferredQuery),
+        enabled: deferredQuery.trim().length >= 2,
```
arpzilla-dev: **[Score: 20/100]** 🎨 User search now needs 2 characters; Classic's `gio-users-selector` fires at `term.length > 0`.

A one-character last name or email local-part will not search. If the 2-char floor is intentional (to spare `/search/users`), say so; otherwise match Classic and search from the first character (with debounce).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleMembersPage.tsx
ankita-gupta21: Omitting id when it's null/undefined now matches Classic's { id: user.id, reference } pattern (and the sibling useGroupAddMembersForm.ts convention already in this module).

Fix: [RoleMembersPage.tsx:76](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleMembersPage.tsx#L76)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleFormPage.tsx
ankita-gupta21: Agreed — the fetch's isError is now surfaced and edit mode renders an error panel instead of silently falling through to a create form.

Fix: [RoleFormPage.tsx:66-68](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleFormPage.tsx#L66-L68) (query wired at [RoleFormPage.tsx:39](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleFormPage.tsx#L39))

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/hooks/useRoleMutations.ts
ankita-gupta21: Right — roleKeys.detail(scope, name) is now invalidated alongside listByScope, so the refetch remounts the form through the loading gate and resets dirty state.

Fix: [useRoleMutations.ts:43](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/hooks/useRoleMutations.ts#L43)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/hooks/useRoles.ts
ankita-gupta21: Fixed — isError is now surfaced per scope group instead of being dropped by the ?? [] fallback, and the roles list renders an explicit error state instead of "no roles."

Fix: [useRoles.ts:50](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/hooks/useRoles.ts#L50) (shape at [useRoles.ts:29](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/hooks/useRoles.ts#L29))

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
ankita-gupta21: Agreed this reads worse than Classic's "cancel navigation" behavior. Changed unauthorizedTo on all three roles routes to point back at the roles list (..) instead of ejecting to Applications.

Fix: [AppRoutes.tsx:594-618](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx#L594-L618)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolePermissionsTable.tsx
ankita-gupta21: Agreed this belongs upstream in graphene-core. Kept the workaround (there's no DS alternative today) but marked it explicitly as temporary and referenced, so it isn't copied elsewhere before the real fix lands. Will file the Graphene ticket separately.

Fix: [RolePermissionsTable.tsx:90-97](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolePermissionsTable.tsx#L90-L97)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleFormPage.tsx
ankita-gupta21: Correct, this isn't a port — Classic blocks it entirely. Called it out explicitly in code as a product decision pending sign-off, not something this comment's presence should be read as approving.

Fix: [RoleFormPage.tsx:70-74](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleFormPage.tsx#L70-L74)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx
ankita-gupta21: Kept the plain list — it's the more faithful port of Classic's mat-list, and Graphene's own "≥3 data columns" rule argues against DataTable here. Documented this as a deliberate tradeoff (search-affordance loss accepted) rather than an oversight.

Fix: [RolesByScopeSection.tsx:159-160](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx#L159-L160)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/hooks/useRoleMemberships.ts
ankita-gupta21: Took the suggested upgrade — switched to Promise.allSettled, now reports succeeded/failed counts and invalidates the list after any success instead of only on total success.

Fix: [useRoleMemberships.ts:51-53](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/hooks/useRoleMemberships.ts#L51-L53)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx
ankita-gupta21: Agreed, switched to the Graphene name-as-link pattern — only the role name is now the interactive Button variant="link", the row itself is a non-interactive container.

Fix: [RolesByScopeSection.tsx:174-178](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx#L174-L178)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleFormPage.tsx
ankita-gupta21: Fixed in both pages — added an isRoleScope guard against ROLE_SCOPES and redirect to the roles list on an invalid scope, replacing the unchecked as RoleScope cast.

Fix: [RoleFormPage.tsx:37,49-50](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleFormPage.tsx#L37) · same pattern in [RoleMembersPage.tsx:39](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleMembersPage.tsx#L39)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/services/roles.ts
ankita-gupta21: Restored — added normalizeScope, applied to both outgoing request paths and incoming Role/Role[] responses, matching Classic's .toUpperCase() mapping.

Fix: [roles.ts:21-22](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/services/roles.ts#L21-L22)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RoleMembersTable.tsx
ankita-gupta21: : Agreed — reverted to a plain <ul> per graphene.md's "<3 data columns → simpler component" rule.

Fix: [RoleMembersTable.tsx:105](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RoleMembersTable.tsx#L105)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RoleMembersTable.tsx
ankita-gupta21: Fixed — DataTableEmptyState now renders in place of the list (wrapped in rounded-md border) rather than through a DataTable prop.

Fix: [RoleMembersTable.tsx:67-69](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RoleMembersTable.tsx#L67-L69) and [92-93](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RoleMembersTable.tsx#L92-L93)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolePermissionsTable.tsx
ankita-gupta21: Fair point — added a comment explaining the intentional deviation from Classic so reviewers don't mistake it for a regression.

Fix: [RolePermissionsTable.tsx:32-33](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolePermissionsTable.tsx#L32-L33)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesTableOfContents.tsx
ankita-gupta21: No design token exists yet for this, so extracted the magic number into a named constant with a comment flagging it as a temporary stand-in until one does.

Fix: [RolesTableOfContents.tsx:30](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesTableOfContents.tsx#L30)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx
ankita-gupta21: Switched to MoreVerticalIcon per graphene.md, rather than propagating the existing module inconsistency further.

Fix: [RolesByScopeSection.tsx:55](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx#L55)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleMembersPage.tsx
ankita-gupta21: Fixed — both the heading and subtitle now use ROLE_SCOPE_LABELS[scope] consistently.

Fix: [RoleMembersPage.tsx:116,119](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/RoleMembersPage.tsx#L116-L119)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/AddRoleMembersSheet.tsx
ankita-gupta21: Matched Classic — search now fires from a single character, with a 300ms debounce added to protect /search/users from the lower threshold.

Fix: [AddRoleMembersSheet.tsx:59](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/AddRoleMembersSheet.tsx#L59) (threshold at [line 111](https://github.com/gravitee-io/gravitee-api-management/blob/ef2305f8ad067a4ac60eaccf730f05a342e5d7d6/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/AddRoleMembersSheet.tsx#L111))

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RoleForm.tsx
```diff
@@ -0,0 +1,185 @@
...
+    // systemRoleEdition-aware isReadOnly getter, which only governs the permission matrix and the banner below.
+    const areBasicFieldsDisabled = isSaving || Boolean(role?.system);
+    const nameValid = form.name.trim().length > 0;
+    // A pristine, untouched field shouldn't show "required" — only once the user has interacted with it
+    // (matches Angular Material's default ErrorStateMatcher, which gates on touched/dirty).
+    const showNameError = nameTouched && !nameValid;
+    // Mirrors gio-save-bar's dirty-gated Save button (creationMode skips this — a filled-in create form is
+    // always submittable, there is nothing to be "unchanged" from).
+    const isDirty = JSON.stringify(form) !== JSON.stringify(initialForm);
+    const canSubmit = nameValid && !isSaving && (!isEditMode || isDirty);
```
arpzilla-dev: **[Score: 55/100]** 💡 `!isReadOnly` was dropped from `canSubmit`, justified by an unreachable test fixture.

This is the only read-only guard on the submit path — `handleSubmit` just early-returns on `!canSubmit`, and `RoleFormPage.handleSubmit` calls `updateMutation.mutateAsync` unconditionally. Removing it means a read-only role can, in principle, be PUT from a form that is simultaneously displaying "System role are not editable".

The new test that drives this (`does not let isReadOnly block submitting an edit to a field it does not disable`) uses `{ name: 'ADMIN', scope: 'ORGANIZATION', system: false }` — read-only but *not* a system role. That state isn't reachable in the product:

- `isRoleReadOnly` returns true only when `role.system` is true, **or** when `systemRoleEdition` is enabled and the role is `ORGANIZATION`/`ADMIN`.
- The real `ORGANIZATION`/`ADMIN` role is created as a system role by `createOrUpdateSystemRoles`.
- A non-system role with that name can't be created either — `RoleServiceImpl.create` → `generateId` → `isReserved(ORGANIZATION, 'ADMIN')` throws `RoleReservedNameException` whether or not `systemRoleEdition` is on.

So `isReadOnly && !system` can't occur, which means the premise in the test comment ("a non-system role can still have its description edited") never applies. In every reachable read-only case `areBasicFieldsDisabled` (line 72) already disables description and default, and `name` is disabled in edit mode, so the form can't go dirty and Save stays off regardless.

Net effect: the change is inert today and the backend would reject the request anyway, which is why this is scored Medium rather than High. But it trades a cheap invariant ("a read-only form cannot submit") for a test asserting behaviour that can't be reached, and it becomes a live bug the moment any field is made editable while `isReadOnly` is true.

Suggestion: restore `!isReadOnly` in `canSubmit` and drop the test. If the intent was really to allow editing metadata on a locked role, that's a product change and the reachable path for it needs to exist first — `isRoleReadOnly` would have to stop implying `role.system`.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx
```diff
@@ -0,0 +1,213 @@
...
+    DropdownMenu,
+    DropdownMenuContent,
+    DropdownMenuItem,
+    DropdownMenuSeparator,
+    DropdownMenuTrigger,
+    Skeleton,
+} from '@gravitee/graphene-core';
+import { LockIcon, MoreVerticalIcon, PlusIcon, Trash2Icon, UsersIcon } from '@gravitee/graphene-core/icons';
+
+import { SectionError } from '../../groups/components/SectionError';
```
arpzilla-dev: **[Score: 40/100]** 💡 Cross-feature import: `roles` reaches into `features/groups/` internals for a generic component.

`SectionError` is a generic error panel that happens to live in `features/groups/components/`. With this PR it gains three consumers in the roles feature — this file, `RoleFormPage.tsx`, and `RoleMembersPage.tsx` — so three of its four call sites are now outside the feature that owns it, and roles has a compile-time dependency on groups' internal structure for no domain reason.

Worth calling out because the same class of issue was fixed correctly elsewhere in this pass: `STANDARD_SHEET_WIDTH` was moved out of `features/applications/components/sheetLayout` into `shared/layout/sheetLayout`, with a re-export shim left behind so existing importers kept working. That's exactly the right treatment.

Suggestion: apply the same move here — relocate `SectionError` to `shared/components/SectionError.tsx` and re-export from the old path (or update `GroupDetailPage.tsx`, the single remaining groups consumer, directly). It's a file move plus four import updates, and it keeps the error-panel component discoverable for the next feature that needs one instead of it accumulating more cross-feature importers.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RoleForm.tsx
ankita-gupta21: You're right, thanks for tracing that through — the state I built the test around isn't reachable. isRoleReadOnly only returns true for role.system, or for ORGANIZATION/ADMIN, and the backend always creates that role as system: true and refuses to let it be recreated as non-system (reserved name). Restored !isReadOnly in canSubmit and dropped the test that asserted the unreachable case.

Fix: [RoleForm.tsx:80](https://github.com/gravitee-io/gravitee-api-management/blob/21c7128048f4bc8ff3326b03f9f36e60d2f03fbb/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RoleForm.tsx#L80)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx
ankita-gupta21: Agreed, and good catch on the precedent — applied the same treatment as STANDARD_SHEET_WIDTH. Moved SectionError out of features/groups/components/ into shared/components/ (alongside ConfirmDialog, FeatureLicenseDialog) and updated all five import sites: the three in roles plus groups' own two remaining consumers (GroupAssociationSection.tsx, GroupDetailPage.tsx). No shim left behind — it's an internal import, not a public API, so there were no other callers to preserve compatibility for.

Fix: [RolesByScopeSection.tsx:32](https://github.com/gravitee-io/gravitee-api-management/blob/21c7128048f4bc8ff3326b03f9f36e60d2f03fbb/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesByScopeSection.tsx#L32) · new location: [shared/components/SectionError.tsx](https://github.com/gravitee-io/gravitee-api-management/blob/21c7128048f4bc8ff3326b03f9f36e60d2f03fbb/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/components/SectionError.tsx)

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/roles/components/RolesTableOfContents.tsx
```diff
@@ -0,0 +1,102 @@
...
+// No shell header-height design token exists to reference yet; centralizing the pixel value here (rather
+// than inline in the IntersectionObserver options below) documents what it stands for and gives future
+// callers one place to update it if the shell header's height changes.
+const SHELL_HEADER_HEIGHT_PX = 96;
+
+/**
+ * Sticky jump-nav for the scope cards on RolesPage — mirrors the Angular reference's
+ * <gio-table-of-contents>: one link per scope, highlighting whichever card is currently in view.
+ */
+export function RolesTableOfContents({ items }: Readonly<{ items: readonly RolesTableOfContentsItem[] }>) {
```
arpzilla-dev, line 36: **[Score: 65/100]** 💡 This component is now dead code, and removing it from the page is a parity regression.

The latest push dropped `<RolesTableOfContents />` from `pages/RolesPage.tsx`, but the component and its spec are both still in the tree. Nothing renders it any more — the only remaining import is its own `RolesTableOfContents.spec.tsx`. Because that spec exercises the component in isolation, it keeps passing, so nothing flags the rot.

The bigger issue is that this wasn't just unused scaffolding — Classic's roles page really does have a table of contents, which is what this component's own docstring on line 33 says it mirrors:

```html
<!-- gravitee-apim-console-webui/src/organization/configuration/roles/org-settings-roles.component.html:87 -->
<gio-table-of-contents
  class="org-settings-roles__toc"
  [sectionNames]="{ '': 'Roles' }"
  scrollingContainer="#gio-toc-scrolling-container"
></gio-table-of-contents>
```

With nine scope cards stacked vertically on a single page, losing the jump-nav is a real usability step back from Classic, and it compounds the earlier decision to keep the plain `<ul>` instead of `DataTable` — the page now has neither a search affordance nor a way to jump between scopes.

There's also leftover surface area pointing at the removed feature: `roleSectionId` still stamps an anchor id onto every scope `Card` and each one still carries `scroll-mt-20`, but with no TOC there is nothing left to scroll to those anchors. And the `SHELL_HEADER_HEIGHT_PX` extraction on line 30 — added in response to the earlier comment about the hardcoded `rootMargin` — now lives entirely inside dead code.

Suggestion: pick one direction and apply it fully.

- **Restore the TOC** (my preference, since it's the Classic-faithful behaviour and the component is already written and tested), or
- **Delete it deliberately** — remove `RolesTableOfContents.tsx`, `RolesTableOfContents.spec.tsx`, `utils/roleSectionId.ts`, the `id={roleSectionId(...)}` and `scroll-mt-20` on the scope `Card`, and note in the PR description that the jump-nav is intentionally out of scope for this parity port.

Either way it shouldn't merge in the current half-removed state.

arpzilla-dev:
@ankita-gupta21 heads up — CI is red on the current head (`84b0ce8`). `ci/circleci: Test gamma UI` is [failing](https://circleci.com/gh/gravitee-io/gravitee-api-management/2038129) with 7 test failures. The other three Gamma jobs (Build Gamma Console, Lint & test Gamma Console, Test gamma) are green, so it's isolated to the platform module's UI tests.

Both causes are specs that weren't updated alongside the last push — the components are fine, the tests just still describe the old markup.

**1. `features/roles/components/RolesByScopeSection.spec.tsx` — 5 failures.** `RoleRowActions` now always renders a single dropdown trigger (`Actions for <role>`) with items `View members` / `Delete role`, but the spec still queries the old per-row icon buttons and the old label:

```
Unable to find an accessible element with the role "button" and name "Button to delete a role"
Unable to find an accessible element with the role "button" and name "Button to see members with this role"
Unable to find an accessible element with the role "menuitem" and name /See members/   (present: "View members", "Delete role")
Found multiple elements with the role "button" and name /ADMIN/
RolesByScopeSection.spec.tsx:205  expect(queryByRole('button', {name: /Actions for/})).not.toBeInTheDocument()
                                  -> found <button aria-label="Actions for USER">
```

Failing cases: *truncates a long role name…*, *selecting a role name calls onSelectRole…*, *only offers delete for roles that are neither system nor default*, *collapses two available row actions into a dropdown menu…*, *only offers "see members" for ORGANIZATION-scope roles*. The last one is worth a second look rather than a mechanical fix — it asserts that no actions trigger exists, which is no longer reachable now that the dropdown always renders when any action is permitted. Note the `/ADMIN/` ambiguity is because the role name became a `Button variant="link"`, so scoping that query to the link specifically will be clearer than loosening it.

**2. `pages/RolesPage.spec.tsx` — 2 failures.** These are expectations for the `RolesTableOfContents` that the same push removed:

```
RolesPage.spec.tsx:92  expect(getAllByText('Organization')).toHaveLength(2) -> Received length: 1
                       Unable to find an accessible element with the role "navigation" and name "Roles sections"
```

Failing cases: *renders the page title and every scope section*, *renders a table-of-contents link for every scope*. How you fix these depends on the TOC decision — I've left a separate comment on `RolesTableOfContents.tsx` about that, since the component and its spec are still in the tree with nothing rendering them, and Classic's roles page does have a table of contents. If the TOC comes back these two tests mostly stand as-is; if it's staying out, they should be deleted along with the component rather than trimmed.

Everything else from the previous round looks good, thanks for the quick turnaround — `canSubmit` has `!isReadOnly` back with the unreachable test removed, and `SectionError` moved to `shared/components/` with all five importers updated and no stale paths left behind. The `RoleRowActions` change is right on its own merits too: always-dropdown matching `DirectMembersTable` is more consistent than the previous lone-icon branching. One small thing while you're in there — that delete item uses `className="text-destructive focus:text-destructive"` where `DirectMembersTable` uses the component's own `variant="destructive"` prop.

arpzilla-dev (approved):

PR #19425, opened 2026-08-26
feat(analytics): make the Kafka client library an analytics dimension

ThibaudAV:
Closes the analytics half of the client-info work on [ESM-2](https://gravitee.atlassian.net/browse/ESM-2).

## The gap

`NATIVE_CONNECTIONS_SUMMARY` could already be broken down by `NATIVE_CLIENT_ID`, but not by the client library — which is the wrong way round for a fleet question.

| | Cardinality | Good facet? |
|---|---|---|
| `client.id` | caller-chosen, near-unique per pod | thousands of bars |
| client library | a handful across a whole estate | yes |

So these had no answer:

- how many connections per client library this week
- what share of traffic still advertises no library at all (`unknown`)
- is the migration off an old client actually progressing, week over week

Filtering logs by library shipped in #19250; **aggregating** by it did not.

## Why this is small

The field is already indexed and already aggregated — it is what backs the value list of the filter added in #19250. This is wiring, not new data, and it therefore applies to **history** as much as to new traffic. No gateway change, no migration.

It follows `NATIVE_CLIENT_ID` step for step: the two repository enums, the two switch arms in `NativeApiFieldResolver`, the `FilterAdapter` allow-list, the catalog entry (`signals` + the metric's `facets`/`filters`) and the two wire enums.

## About the two removed `THROW_EXCEPTION` mappings

They came from #19250 and are not a mistake being undone. At that point the catalog had to carry the filter so the filter bar could list its values, while the analytics engine had no dimension to group by — the choice was between lying in the catalog and refusing cleanly. Refusing cleanly was right then; the dimension exists now, so the refusal does not belong.

## A test that earned its keep

`FilterAdapterFieldResolverParityTest` caught a real omission mid-change: wiring `NativeApiFieldResolver` alone makes the filter *resolvable* while `FilterAdapter.NATIVE_FILTER_NAMES` still drops it from the ES query — accepted by the API, silently ignored in the result. Exactly the failure mode it exists to prevent.

## Tests

`repository-elasticsearch` **864** · `rest-api-service` **9525** · `management-v2-rest` **1855** · `gamma-rest-api` **392**, plus a full backend build. Two new cases pin the facet and the filter to `additional-metrics.keyword_native-kafka_client-software-name`.

One snapshot moved on purpose: `AnalyticsQueryValidatorTest` no longer lists this name among the filters withheld from analytics.

[ESM-2]: https://gravitee.atlassian.net/browse/ESM-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/resources/analytics/definition/analytics-definition.yaml
```diff
@@ -1151,13 +1151,17 @@ spec:
               # can actually group by.
               - NATIVE_FAILURE_SIDE
               - NATIVE_CLIENT_ID
+              # The fleet question, and the one client.id cannot answer: an id is near-unique per pod,
+              # so grouping by it yields thousands of bars, while the library is a handful of values.
+              - NATIVE_CLIENT_SOFTWARE_NAME
```
callaertanthony, line 1156: **Blocker: this facet never reaches any UI.** It is listed on the metric, but the top-level `facets:` catalog (starts line 2323) has no entry for it — `NATIVE_CLIENT_ID` has one at line 2460.

The facet-spec endpoint intersects the two:

```java
// AnalyticsDefinitionYAMLQueryService.java:136
public List<FacetSpec> getFacets(MetricSpec.Name metricSpecName) {
    var metric = getMetricByName(metricSpecName);
    return spec.facets().stream()
        .filter(facet -> metric.facets().contains(facet.name()))
        .toList();
}
```

`spec.facets()` never contains `NATIVE_CLIENT_SOFTWARE_NAME`, so `GetMetricFacetSpecUseCase` drops it and the breakdown picker never offers "Kafka Client Library" — the fleet question this PR sets out to answer.

It fails silently: `AnalyticsQueryValidator.validateFacets` (line 102) checks `metricSpec.facets()`, not the catalog, so a hand-crafted request still works and nothing throws. Nothing validates the two lists against each other at load time either.

Fix — alongside `NATIVE_CLIENT_ID` at line 2460:

```yaml
        - name: NATIVE_CLIENT_SOFTWARE_NAME
          label: Kafka Client Library
          type: KEYWORD
```

callaertanthony (commented):
Two things beyond the inline blocker.

### Gamma still refuses the filter for analytics — improvement

`gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/model/StaticFilters.java:207` declares `Defs.LOGS`, while its sibling `NATIVE_CLIENT_ID` (line 191) is `Defs.LOGS_ANALYTICS`. Gamma's `AnalyticsRequestPipeline:75` calls `filterValidator.validate(conditions, Signal.ANALYTICS)`, and `ObservabilityAnalyticsDataPortAdapter` delegates to the same core engine — so after this PR the filter is accepted through management-v2 and rejected in Gamma.

Its javadoc is also now false:

> LOGS-only, like {@link #FAILURE_ORIGIN}: the analytics definition does not declare this dimension yet

Either flip it to `Defs.LOGS_ANALYTICS` in this PR, or say explicitly that Gamma is a follow-up — right now the file reads as if nothing changed.

### No parity test between metric facet lists and the facets catalog — nit

`FilterAdapterFieldResolverParityTest` exists for exactly this failure mode on the filter side, and the description credits it with catching one mid-change. The facet side has no equivalent, which is what let the inline blocker through: adding a name to a metric's `facets:` list while the catalog has no `FacetSpec` for it compiles, validates, and passes every test — it just never surfaces.

A test asserting that every name in every `metric.facets()` resolves to an entry in the top-level `facets:` catalog would pin it, and would fail today.

---

The rest checks out: `NativeApiMetricKeys.CLIENT_SOFTWARE_NAME` resolves to `keyword_native-kafka_client-software-name` as described, both wire enums are updated (`FacetName` and `FilterName`; `ObservabilityFilterName` already carried it from #19250), and dropping the two `THROW_EXCEPTION` mappings is right — MapStruct auto-maps now that source and target enums both carry the constant.

ThibaudAV:
All three taken. The blocker was the important one — the feature this PR exists for would have shipped invisible.

## Blocker — the facet never reached any UI

Verified before fixing: `getFacets` intersects `spec.facets()` with `metric.facets()`, and the top-level catalog had no entry for the name. So the breakdown picker would never have offered "Kafka Client Library" — the exact question the PR sets out to answer — and nothing would have said so. Entry added next to `NATIVE_CLIENT_ID`, as suggested.

Worth naming what made this dangerous: it fails silently *in both directions*. `AnalyticsQueryValidator` checks the metric's list rather than the catalog, so a hand-crafted request works and the dimension is genuinely queryable — which is precisely how it survives manual testing.

## Gamma refused the filter — fixed here, not deferred

Correct, and the javadoc had become false. `Defs.LOGS` → `Defs.LOGS_ANALYTICS`, with the prose rewritten to say why: Gamma runs the same core engine through `ObservabilityAnalyticsDataPortAdapter`, so leaving it LOGS-only would have made the same filter work through management-v2 and fail in Gamma. Splitting that across two PRs would have shipped an inconsistency on purpose.

## The parity test — added, and it found more than mine

This is the one I should have written with the filter-side equivalent in #19250; the asymmetry is what let the blocker through.

Rather than reaching for internal state, it compares what a metric declares against what the public accessor actually returns — measuring the silent loss directly. Verified by re-removing the catalog entry: it reports `NATIVE_CONNECTIONS_SUMMARY -> NATIVE_CLIENT_SOFTWARE_NAME`.

It also surfaced **nine pre-existing gaps**, none of them from this PR:

```
facets   EDGE_HEARTBEAT_COUNT -> EDGE_VERSION
         EDGE_TOKENS_IN/OUT   -> EDGE_MODEL, EDGE_TOOL
filters  the same five, plus
         MESSAGES / MESSAGE_ERRORS / MESSAGE_GATEWAY_LATENCY / MESSAGE_PAYLOAD_SIZE -> MESSAGE_CONNECTOR_ID
```

I have not catalogued them. Whether each is an omission or a deliberate withholding belongs to the teams owning EDGE and MESSAGE, and silently adding them could surface pickers someone removed on purpose. They are pinned as an exact set that can only shrink — fixing one fails the test until it is struck off, and nothing may be added. Happy to raise it separately if you want it tracked.

## On the version dimension

Fair point from the #70 review, and it applies here: the facet is name-only, so every librdkafka version collapses into one bar and a version-specific fault stays invisible. `clientSoftwareVersion` is on the documents already. Deliberately not in this PR — a second facet is a product decision, not a side effect of this one — but it is the obvious next step if the fleet view proves useful.

Full build green. `rest-api-service` **9527**, `gamma-rest-api` **392**.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/domain_service/analytics_engine/definition/AnalyticsDefinitionYAMLQueryServiceTest.java
```diff
@@ -103,6 +103,87 @@ void should_advertise_every_api_kind_a_metric_declares() {
...
+        @Test
+        void should_surface_every_filter_a_metric_lists() {
+            assertThat(
+                droppedDimensions(MetricSpec::filters, (service, name) -> service.getFilters(name).stream().map(FilterSpec::name).toList())
+            ).containsExactlyInAnyOrderElementsOf(KNOWN_UNCATALOGUED_FILTERS);
+        }
+
+        private static <T> List<String> droppedDimensions(
+            java.util.function.Function<MetricSpec, List<T>> declared,
+            java.util.function.BiFunction<AnalyticsDefinitionYAMLQueryService, MetricSpec.Name, List<T>> surfaced
```
callaertanthony: Nit: `java.util.function.Function` / `BiFunction` are fully qualified here — `AGENTS.md` asks for explicit imports, and the file already imports everything else it uses. Adding the two imports makes the signature read a fair bit shorter too.

callaertanthony (approved):
All three points addressed. Approving — the two below are nits, neither worth holding the PR for.

I checked the new parity test's allowlists against the catalog rather than taking them on trust: `EDGE_VERSION` / `EDGE_MODEL` / `EDGE_TOOL` appear only in metric lists and never as a `- name:` in either catalog, and `MESSAGE_CONNECTOR_ID` has a facets entry (line 2404) but no filters entry — so both lists are accurate, and the filter-side entries are exactly the ones the facet side does not need. All 52 metrics declare `apis:`, so iterating `ApiSpec.Name.values()` misses no metric. Pinning it as an exact set rather than a floor is the right call.

### The Gamma flip has no test — nit

`StaticFilters.NATIVE_CLIENT_SOFTWARE_NAME` moving from `Defs.LOGS` to `Defs.LOGS_ANALYTICS` is a behaviour change with nothing asserting it. `SpiFilterRegistryTest:224` pins the `LOGS` + `NATIVE` list, which contained the filter before and after, and no test snapshots the `ANALYTICS` + `NATIVE` list — so reverting the signal today breaks no test. Given the javadoc now explains precisely why the two signals must agree with the core catalog, a one-line assertion on `getFilters(Set.of(Signal.ANALYTICS), Set.of(ApiType.NATIVE))` would keep the next person from undoing it.

ThibaudAV:
Both nits taken.

**The Gamma flip now has a test.** You were right that nothing pinned it — `SpiFilterRegistryTest` only snapshotted the LOGS+NATIVE list, which contained the filter before and after. Added an assertion on `getFilters(Set.of(Signal.ANALYTICS), Set.of(ApiType.NATIVE))` covering both Kafka client dimensions, plus `doesNotContain("FAILURE_ORIGIN")` so it says what the axis is *for* rather than just listing names — that one is derived from the error key rather than stored, so the engine genuinely cannot group by it.

Verified by reverting the signal on that declaration: the build fails. It took me two tries — my first mutation hit a different filter's `Defs.LOGS_ANALYTICS` block and the suite stayed green, which would have let me claim a guard that wasn't there.

**Imports made explicit**, per AGENTS.md. The signature reads much shorter, as you said.

Thanks for checking the allowlists against the catalog rather than taking them on trust — the `MESSAGE_CONNECTOR_ID` asymmetry (facets entry at 2404, no filters entry) is exactly the kind of thing I would have assumed symmetric.

`gamma-rest-api` **393**, `rest-api-service` **9527**. Squashed into the single commit.

PR #19426, opened 2026-08-26
feat(portal): create gamma module for developer portals.

kovaren:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-153

### Description

Scaffolds a UI-only Developer Portals Gamma module and wires it into the control plane as a core product, same pattern as APIM and Platform. Operators get a home card and app-switcher entry; there is no REST API yet.

#### Module

New gravitee-gamma-module-portals plugin (id portals): Java stub with no REST resource, Nx/Rspack federated UI on port 3003, registered in the Gamma reactor and standalone distribution.

#### Host console

Home catalog, labels, GioDeveloperPortalIcon, and ShellLayout switcher. HomePage tests cover the heading and empty-state CTA.

#### CI and docs

Gamma UI CI lints and tests the new Nx project. CONTRIBUTING.adoc and AI-context sources declare the module.


gravitee-gamma/gravitee-gamma-module-portals/README.md
```diff
@@ -0,0 +1,22 @@
+# Gravitee Gamma — Developer Portals module
+
+Developer Portals module for [Gravitee Gamma](../gravitee-gamma-plugin/gravitee-gamma-plugin-module/README.md): backend (Jersey) and React micro-frontend (Module Federation remote).
+
+## Local development
+
+Commands run from the **repository root** (Nx workspace).
+
+1. Run the Management API on port 8083 with **`gamma.enabled: true`** in `gravitee.yml` (or `GAMMA_ENABLED=true`). See the **Gamma Console** section in [`CONTRIBUTING.adoc`](../../CONTRIBUTING.adoc#dev-guide-gamma-console).
+2. Start this module’s UI dev server (port **3003**, hot reload):
```
jourdiw: I think the 3003 may be used by another module (ESM or Edge?). 
It would be good to double check if another module is already configured to that port.

gravitee-gamma/gravitee-gamma-module-portals/src/main/ui/shared/gamma-modules-sdk.ts
jourdiw, line 1: I'm wondering if there are more interfaces that you can use from the SDK for these contracts 🤔

jourdiw (approved):

gravitee-gamma/gravitee-gamma-module-portals/README.md
kovaren: Good catch, AuthZ is using 3003 already.

I bumped the portals port to 3004 which is the next one available.

gravitee-gamma/gravitee-gamma-module-portals/src/main/ui/shared/gamma-modules-sdk.ts
kovaren: Fixed, reused the existing types.

.circleci/ci/src/jobs/backend/job-build-backend.ts
```diff
@@ -94,6 +94,6 @@ export class BuildBackendJob {
         ],
       }),
     ];
-    return new Job(jobName, OpenJdkNodeExecutor.create('large'), steps);
+    return new Job(jobName, OpenJdkNodeExecutor.create('xlarge'), steps);
```
Okhelifi, line 97: Did you encounter any ressource issue or something like this to have to increse the size of the executor ?

.gitignore
```diff
@@ -125,5 +125,6 @@ CLAUDE.md
 !/gravitee-gamma/gravitee-gamma-definition-model/CLAUDE.md
 !/gravitee-gamma/gravitee-gamma-module-apim/CLAUDE.md
 !/gravitee-gamma/gravitee-gamma-module-platform/CLAUDE.md
+!/gravitee-gamma/gravitee-gamma-module-portals/CLAUDE.md
```
Okhelifi, line 128: This line whitelists a CLAUDE.md the PR does not add. The generation step (gbuddy setup, referenced at the top of the existing AGENTS.md files) has not been run, so the module ships without AGENTS.md or CLAUDE.md, while both apim and platform have them. The root AGENTS.md (line 197) does not list the new directory either, so it is missing from the set of paths whose module rules must be read before working there.

The sources are already declared correctly (.ai/manifest.yaml, .ai/rules/apim-java.md, .ai/rules/gamma-clean-architecture.md), so only the generated half is missing. Running the generator should produce the two module files and update the root list.

Worth fixing but is it something you plan to do later maybe ?

gravitee-gamma/gravitee-gamma-module-portals/src/main/ui/index.html
```diff
@@ -0,0 +1,29 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<!doctype html>
+<html lang="en">
```
Okhelifi, line 19: Should this module plan for internationalization? Asking because the legacy gravitee-apim-portal-webui shipped `@ngx-translate` with four locales, so a developer portal is the one surface in APIM that historically was translated.

For the record, nothing here diverges from current practice: no gamma module has i18n, the control-plane-webui host has none either, and gravitee-apim-portal-webui-next dropped it. So this is not a gap in the scaffold, it is a product decision that would have to be taken at the host level (module labels, app switcher and home cards are hardcoded there too) rather than module by module.

Okhelifi (commented):
Quick architecture question here. How do we plan to handle the "public portal use case" ? I mean on our current portal users can register access to public APIs or even access to ressources without being logged in.

Today the module loading mechanism only runs for authenticated users: useGammaModules clears the module list when there is no user, and every route except /login sits behind ProtectedRoute. So this shell can't serve anonymous visitors at all.

Is  it a separate topic the team will handle later ?

.circleci/ci/src/jobs/backend/job-build-backend.ts
kovaren: Yep, the backend build failed on CI because it went out of memory. I increased the executor size just to confirm the cause.

That being said, it might not be the best solution all things considered. Another option is to lower the number of parallel maven builds.

WDYT?

kovaren:
> Quick architecture question here. How do we plan to handle the "public portal use case" ? I mean on our current portal users can register access to public APIs or even access to ressources without being logged in.
> 
> Today the module loading mechanism only runs for authenticated users: useGammaModules clears the module list when there is no user, and every route except /login sits behind ProtectedRoute. So this shell can't serve anonymous visitors at all.
> 
> Is it a separate topic the team will handle later ?

The purpose of this new module is not to house the developer portal itself, but rather to handle its management side, the equivalent of the portal editor + portal settings in Classic Console.

The actual developer portal will be served by a separate app, which will handle public, unauthenticated access.

gravitee-gamma/gravitee-gamma-module-portals/src/main/ui/index.html
kovaren: Nope, the internationalization will be handled by a standalone portal app. This module only covers the management side of things. Answer slightly related to https://github.com/gravitee-io/gravitee-api-management/pull/19426#issuecomment-5439424113

.circleci/ci/src/jobs/backend/job-build-backend.ts
Okhelifi: if it blocks this PR I would say we can increase it and check if we can decrease it in a dedicated PR by changing the number of parallel maven builds like you said. 👍🏼

gravitee-gamma/gravitee-gamma-module-portals/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "gravitee-gamma-module-portals",
+    "private": true,
+    "dependencies": {
+        "@gravitee/gamma-modules-sdk": "1.2.0",
+        "@gravitee/graphene-core": "3.12.0",
```
Okhelifi: Can you just bump to 3.14.0 ( latest ) please 😅

Okhelifi (approved):

gravitee-gamma/gravitee-gamma-module-portals/package.json
kovaren: Done!

.gitignore
kovaren: As discussed together, I added AGENTS.md and CLAUDE.md but left out the unrelated changes from gbuddy.

PR #19428, opened 2026-08-27
feat(gamma-apim): support TCP Proxy in the scratch API creation wizard

ankita-gupta21:
## Issue

https://gravitee.atlassian.net/browse/APIM-14670

## Summary

Adds **TCP Proxy** as a second selectable API type in Gamma's "Start from scratch" wizard, alongside the existing HTTP proxy flow — bringing TCP proxy creation to Gamma for the first time (previously only possible via legacy Console).

## What changed

- **API Details step** — API type now chosen via HTTP Proxy / TCP Proxy cards, gated on the entrypoint plugin's deployment status
- **Configure Proxy step** — TCP swaps HTTP path/vhost fields for TCP-specific ones: gateway hosts, backend target host/port, optional TLS
- **Host validation** — reuses existing hostname-regex validation and the async host-uniqueness-check service (previously only used by the duplicate-API dialog)
- **Secure step** — Keyless-only for TCP (no auth-type selector), matching classic console's rule that TCP listeners can't carry HTTP-level auth
- **Review & Deploy** — builds a TCP listener + `tcp-proxy` endpoint group matching the plugin's config schema
- **Detail pages** — Entrypoints, Endpoints, and the endpoint form all gained TCP-aware variants (TCP hosts, `host:port` display, hidden HTTP-only options like health checks)
- **Sidebar nav** — locks/hides HTTP-only sections (CORS, Response Templates, Policy Studio, API Score) for TCP APIs
- **Access control** — new `CreateApiGate` guards `apis/new*` routes behind `environment-api-c` permission
- **Reused, not rebuilt** — extends the existing wizard/step/store scaffolding rather than a parallel TCP flow

## New files

- `useVerifyTcpHosts.ts` — async TCP host uniqueness validation hook
- `TcpHostsCard.tsx` — TCP entrypoints display card
- `CreateApiGate.tsx` — permission gate for create-API routes
- Specs for the above, plus expanded coverage on `ApiDetailSidebarNav`, `ApiEndpointsPage`, `ApiEntrypointsPage`, `EndpointGroupList`, `CreatePlanDropdown`

https://github.com/user-attachments/assets/4e073d98-0bb4-4f34-b8fa-d4e4cc8d1872

[tcp_proxy_test_scenarios.md](https://github.com/user-attachments/files/31776539/tcp_proxy_test_scenarios.md)

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/app/AppRoutes.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/app/AppRoutes.tsx
@@ -42,6 +42,7 @@ import { ApiProductPlansPage } from '../features/api-products/pages/detail/plans
 import { ApiProductUserPermissionsPage } from '../features/api-products/pages/detail/user-permissions/ApiProductUserPermissionsPage';
 import { ApiDetailIndexRedirect, ApiDetailLayout } from '../features/apis/components/detail/ApiDetailLayout';
 import { ApisPage } from '../features/apis/pages/ApisPage';
+import { CreateApiGate } from '../features/apis/pages/CreateApiGate';
 import { CreateApiProxyPage } from '../features/apis/pages/CreateApiProxyPage';
 import { AlertFormPage } from '../features/apis/pages/detail/alerts/AlertFormPage';
 import { ApiAlertsPage } from '../features/apis/pages/detail/alerts/ApiAlertsPage';
@@ -204,7 +205,7 @@ export function AppRoutes() {
                         <Route path="quick-start" element={<QuickStartPage />} />
                         <Route path="apis">
                             <Route index element={<ApisPage />} />
-                            <Route path="new">
+                            <Route path="new" element={<CreateApiGate />}>
                                 <Route index element={<CreateApiProxyPage />} />
                                 <Route path="scratch" element={<ScratchWizardPage />} />
                                 <Route path="template/:id" element={<TemplateWizardPage />} />
--- a/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/detail/ApiDetailLayout.spec.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/detail/ApiDetailLayout.spec.tsx
@@ -106,6 +106,7 @@ jest.mock('@gravitee/graphene-core/icons', () => new Proxy({}, { get: () => () =
 jest.mock('./ApiDetailSidebarNav', () => ({
     API_PROXY_NAV_GROUPS: [],
     ApiDetailSidebarNav: () => <div />,
+    withTcpRestrictions: (groups: unknown[]) => groups,
 }));
 
 import { ApiDetailIndexRedirect, ApiDetailLayout } from './ApiDetailLayout';
--- a/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/detail/ApiDetailLayout.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/detail/ApiDetailLayout.tsx
@@ -40,13 +40,14 @@ import { useMutation, useQueryClient } from '@tanstack/react-query';
 import { useId, useState } from 'react';
 import { Navigate, Outlet, useParams } from 'react-router-dom';
 
-import { API_PROXY_NAV_GROUPS, ApiDetailSidebarNav } from './ApiDetailSidebarNav';
+import { API_PROXY_NAV_GROUPS, ApiDetailSidebarNav, withTcpRestrictions } from './ApiDetailSidebarNav';
 import { useDetailBasePath } from '../../../../shared/hooks/useDetailBasePath';
 import { ApiDetailContext } from '../../context/ApiDetailContext';
 import { useApiDetail } from '../../hooks/useApiDetail';
 import { useApiPermissions } from '../../hooks/useApiPermissions';
 import { deployApi } from '../../services/apis';
 import type { ApiDetailDto } from '../../types';
+import { hasTcpListeners } from '../../utils/apiHttpProxy';
 import { apiDetailKeys } from '../../utils/queryKeys';
 
 /** Classic console caps the deployment label at 32 characters. */
@@ -232,7 +233,7 @@ function ApiInfoHeader({ api, isLoading }: { api: ApiDetailDto | null; isLoading
             <div className="flex flex-wrap items-center gap-1">
                 {api.type ? (
                     <Badge variant="secondary" className="text-xs px-1.5 py-0 h-5">
-                        {api.type === 'PROXY' ? 'HTTP Proxy' : 'Event-driven'}
+                        {api.type === 'PROXY' ? (hasTcpListeners(api) ? 'TCP Proxy' : 'HTTP Proxy') : 'Event-driven'}
                     </Badge>
                 ) : null}
                 {api.apiVersion ? (
@@ -270,6 +271,7 @@ export function ApiDetailLayout() {
     });
 
     const showDeployBanner = !isError && api?.deploymentState === 'NEED_REDEPLOY' && canDeploy;
+    const navGroups = withTcpRestrictions(API_PROXY_NAV_GROUPS, hasTcpListeners(api));
 
     useLayoutConfig(
         {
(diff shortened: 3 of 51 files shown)


aqibmohammadkhan (commented):
Correctness
Blank extra TCP host row silently blocks the wizard — apiCreationValidation.ts#L49 validateTcpHosts() runs validateDuplicateHost over every row in form.tcpHosts, but EntrypointsStep.tsx#L182 only shows a required * on row 0. Click "Add host", leave the new row blank, and you get an unattributed 'Host is required.' error blocking Next — worse, useVerifyTcpHosts.ts#L43 also bails on this condition, so the async uniqueness check silently never fires. Either mark every row required in the UI, or only validate non-empty rows.

"Coming soon" sidebar rows are keyboard/screen-reader inert — ApiDetailSidebarNav.tsx#L182 (ComingSoonRow) These used to be filtered out of the DOM entirely; now they render as a plain <div> with no tabIndex, role, or aria-disabled, and the tooltip only opens on hover. Keyboard/screen-reader users tabbing through the nav land on unlabeled, inert rows with no way to learn why the item is disabled. Needs at least tabIndex={0} + aria-disabled + a focus-triggered tooltip (or title).

TCP plugin-deployment gate is UI-only — DetailsStep.tsx#L63 (isPluginUnavailable) No step validator or the submit path re-checks deployed. While listEntrypointPlugins is still loading or errors, entrypointPlugins?.find(...)?.deployed === false evaluates to false (i.e. "available"), so TCP is selectable and submittable on an environment where the plugin isn't actually deployed — relying entirely on a backend rejection with no earlier guardrail.

Protocol switch clears errors but not stale field values — apiCreationStore.tsx#L135-L141 Switching HTTP→TCP→HTTP leaves the old HTTP field values (targetUrl, contextPath, etc.) sitting in form — harmless today since apiProxyMapper.ts branches cleanly on protocol, but a latent trap for any future direct form read or draft-save feature. Low severity, worth a defensive reset.

Conventions / test coverage
Unrelated refactor bundled into the feature diff — apiCreationStore.tsx#L258 Swapping <ApiCreationContext.Provider value={{state, dispatch}}> for a memoized useMemo + bare <ApiCreationContext value={value}> isn't needed for TCP support. Per this repo's AGENTS.md "Surgical changes" rule ("don't refactor... unrelated code"), consider splitting this out.

TcpHostsCard's duplicate-host check has no test — TcpHostsCard.tsx#L23 (findRowError) The equivalent wizard-side logic is unit-tested (apiCreationValidation.spec.ts), but this near-duplicate detail-page implementation isn't covered by any spec — a regression here (e.g. a trim/case bug) would ship silently.

Cleanup (non-blocking)
useVerifyTcpHosts near-duplicates useVerifyContextPath — useVerifyTcpHosts.ts#L30. Same debounce/verify skeleton copy-pasted rather than factored into a shared hook.
apiProxyMapper.ts branches on protocol 7 separate times — one if (protocol === 'TCP') per function (buildPreviewGatewayUrl, buildListener, buildEndpointGroups, ...) rather than a per-protocol dispatch table. A third proxy type will mean touching all 7.
CreateApiGate introduces a new permission-check idiom — CreateApiGate.tsx#L34 is the only route-level PermissionGate usage in the module; the other 19 call sites use the useHasPermission hook. Worth converging on one pattern.
isTcp re-derived independently in 6+ files — DetailsStep, EntrypointsStep, SecurityStep, ReviewDeployStep, ApiDetailLayout (via hasTcpListeners), ApiEndpointsPage (via type === 'tcp-proxy') each compute "is this TCP" their own way. A single shared predicate would reduce drift risk.

ankita-gupta21:
> Correctness Blank extra TCP host row silently blocks the wizard — apiCreationValidation.ts#L49 validateTcpHosts() runs validateDuplicateHost over every row in form.tcpHosts, but EntrypointsStep.tsx#L182 only shows a required * on row 0. Click "Add host", leave the new row blank, and you get an unattributed 'Host is required.' error blocking Next — worse, useVerifyTcpHosts.ts#L43 also bails on this condition, so the async uniqueness check silently never fires. Either mark every row required in the UI, or only validate non-empty rows.
> 
> "Coming soon" sidebar rows are keyboard/screen-reader inert — ApiDetailSidebarNav.tsx#L182 (ComingSoonRow) These used to be filtered out of the DOM entirely; now they render as a plain
> 
> with no tabIndex, role, or aria-disabled, and the tooltip only opens on hover. Keyboard/screen-reader users tabbing through the nav land on unlabeled, inert rows with no way to learn why the item is disabled. Needs at least tabIndex={0} + aria-disabled + a focus-triggered tooltip (or title).
> TCP plugin-deployment gate is UI-only — DetailsStep.tsx#L63 (isPluginUnavailable) No step validator or the submit path re-checks deployed. While listEntrypointPlugins is still loading or errors, entrypointPlugins?.find(...)?.deployed === false evaluates to false (i.e. "available"), so TCP is selectable and submittable on an environment where the plugin isn't actually deployed — relying entirely on a backend rejection with no earlier guardrail.
> 
> Protocol switch clears errors but not stale field values — apiCreationStore.tsx#L135-L141 Switching HTTP→TCP→HTTP leaves the old HTTP field values (targetUrl, contextPath, etc.) sitting in form — harmless today since apiProxyMapper.ts branches cleanly on protocol, but a latent trap for any future direct form read or draft-save feature. Low severity, worth a defensive reset.
> 
> Conventions / test coverage Unrelated refactor bundled into the feature diff — apiCreationStore.tsx#L258 Swapping <ApiCreationContext.Provider value={{state, dispatch}}> for a memoized useMemo + bare isn't needed for TCP support. Per this repo's AGENTS.md "Surgical changes" rule ("don't refactor... unrelated code"), consider splitting this out.
> 
> TcpHostsCard's duplicate-host check has no test — TcpHostsCard.tsx#L23 (findRowError) The equivalent wizard-side logic is unit-tested (apiCreationValidation.spec.ts), but this near-duplicate detail-page implementation isn't covered by any spec — a regression here (e.g. a trim/case bug) would ship silently.
> 
> Cleanup (non-blocking) useVerifyTcpHosts near-duplicates useVerifyContextPath — useVerifyTcpHosts.ts#L30. Same debounce/verify skeleton copy-pasted rather than factored into a shared hook. apiProxyMapper.ts branches on protocol 7 separate times — one if (protocol === 'TCP') per function (buildPreviewGatewayUrl, buildListener, buildEndpointGroups, ...) rather than a per-protocol dispatch table. A third proxy type will mean touching all 7. CreateApiGate introduces a new permission-check idiom — CreateApiGate.tsx#L34 is the only route-level PermissionGate usage in the module; the other 19 call sites use the useHasPermission hook. Worth converging on one pattern. isTcp re-derived independently in 6+ files — DetailsStep, EntrypointsStep, SecurityStep, ReviewDeployStep, ApiDetailLayout (via hasTcpListeners), ApiEndpointsPage (via type === 'tcp-proxy') each compute "is this TCP" their own way. A single shared predicate would reduce drift risk.

Addressed all 7 items from this review:

1. Nested ternary in API type badge → extracted via `hasTcpListeners()`:
   https://github.com/gravitee-io/gravitee-api-management/blob/e9bbca693066cbb55244eef510a23072c98983c2/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/detail/ApiDetailLayout.tsx#L236

2. Combined TCP/OAuth2 guard in `buildApiResources` → single early-return condition (kept as one line, reasons are independently checkable):
   https://github.com/gravitee-io/gravitee-api-management/blob/e9bbca693066cbb55244eef510a23072c98983c2/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/utils/apiProxyMapper.ts#L198

3. `TcpEntrypointsFields` size → extracted from the wizard step as a named component:
   https://github.com/gravitee-io/gravitee-api-management/blob/e9bbca693066cbb55244eef510a23072c98983c2/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/steps/EntrypointsStep.tsx#L161

4. `DetailsStep` plugin-loading + form logic → `isPluginUnavailable` isolated as a helper:
   https://github.com/gravitee-io/gravitee-api-management/blob/e9bbca693066cbb55244eef510a23072c98983c2/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/steps/DetailsStep.tsx#L69

5. `apiCreationReducer` protocol-reset complexity → `UPDATE_FORM` case now delegates to helpers:
   https://github.com/gravitee-io/gravitee-api-management/blob/e9bbca693066cbb55244eef510a23072c98983c2/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/store/apiCreationStore.tsx#L128

6. `DetailsStep` overall size (query + availability + form) → same fix as #4, `useQuery` isolated from render logic:
   https://github.com/gravitee-io/gravitee-api-management/blob/e9bbca693066cbb55244eef510a23072c98983c2/gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/components/steps/DetailsStep.tsx#L57

7. `CreateApiGate` ternary → converted to early return, per the suggested diff:
   https://github.com/gravitee-io/gravitee-api-management/blob/cdc09ada6d.../gravitee-gamma/gravitee-gamma-module-apim/src/main/ui/features/apis/pages/CreateApiGate.tsx#L39-L45

aqibmohammadkhan (approved):

PR #19430, opened 2026-08-27
test: cover client abort during streaming and on http/2

wbabyte:
## Issue

https://gravitee.atlassian.net/browse/P1ACT-466

## Description

Post-P1 action for P1ACT-464 (Marex 18000), part 2. Adds the two client-abort cases the existing coverage misses, extending `ClientAbortPropagationV4IntegrationTest` from 2 to 6 tests.

**1. Abort during response streaming.** The APIM-14605 fix guards the window between pool acquisition and the response head via `doOnDispose` in `HttpConnector.connect()`. From the response head onward cancellation is owned by a different branch — `doOnCancel` on the response chunks in `getEndpointResponseChunks()` — which had no slot-release assertion.

**2. Abort on an HTTP/2 endpoint.** All three abort points, against an h2c endpoint where a pool slot is one stream of a multiplexed connection.

The two existing test bodies moved unchanged into three shared scenario helpers so each abort point runs against both endpoint protocols without triplicating the setup. Assertions remain on observable slot availability (probe latency + backend request count).

New API definition `api-client-abort-single-connection-http2.json` caps **both** `maxConcurrentConnections: 1` and `http2MultiplexingLimit: 1` — with the default multiplexing limit the spare streams on the connection hide a leaked one and the test gives a false pass.

### Mutation evidence

Each test fails on the branch it guards, and on the probe-latency assertion:

| Test | fix removed (`doOnDispose`) | `doOnCancel` reset removed | master |
|---|---|---|---|
| queued HTTP/1.1 | FAIL | pass | pass |
| in-flight HTTP/1.1 | FAIL | pass | pass |
| in-flight HTTP/2 | FAIL | pass | pass |
| streaming HTTP/1.1 | pass | **FAIL** | pass |
| streaming HTTP/2 | pass | **FAIL** | pass |
| queued HTTP/2 | FAIL | FAIL | **FAIL — see below** |

### Gap found: HTTP/2 queued abort leaks the stream's multiplexing permit

On an HTTP/2 endpoint, a client abort **while the request is queued** for a pool slot leaves the endpoint unusable. The APIM-14605 release path does run — the stream granted after the abort is reset and the connector logs `Downstream request has been disposed, releasing upstream request to [...]` — but the multiplexing permit it consumed is never returned.

Measured with `maxConcurrentConnections: 1`, `http2MultiplexingLimit: 1`, `readTimeout: 8000`, `idleTimeout: 30000`:

- three consecutive probes → 504 after ~8s each, backend request count frozen at 1 (they never reach the backend);
- after a 35s idle wait → 200 in 16ms, i.e. recovery only once `idleTimeout` evicts the poisoned connection.

It is specifically the **reset of a not-yet-sent H2 stream** that leaks: HTTP/1.1 releases the same slot correctly, and an HTTP/2 abort after the request has been sent (in-flight, streaming) does too. A control run without the abort confirms the configuration is sound — the queued request waits ~2.9s and gets a 200.

Per the ticket's acceptance criteria the gap is reported rather than the assertion relaxed: that test is `@Disabled` with the full characterisation in its javadoc, ready to enable once the permit is released. A Jira bug for it is not filed yet.

### Note on the streaming stub

The streamed response uses `Content-Type: text/event-stream`. The gateway only streams a body downstream for the content types `RequestUtils.hasStreamingContentType` recognises and aggregates every other response, releasing it in one piece once the backend is done — which would leave no streaming window to abort in. Unrelated observation, not addressed here: `application/octet-stream` is listed in that method but did not stream in practice, whereas `text/event-stream` did.

### Verification

`ClientAbortPropagationV4IntegrationTest` — 6 tests, 5 pass, 1 skipped, 8.2s, with the prettier format gate enabled. Run against the working-tree engine (`-Pengine-snapshot`); without that profile the distribution assembles the pinned 4.12.14 engine instead.

Test-only change: no production code, no compatibility-sensitive surface touched.

### DIFF
--- a/gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/http/ClientAbortPropagationV4IntegrationTest.java
+++ b/gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/http/ClientAbortPropagationV4IntegrationTest.java
@@ -28,6 +28,8 @@ import io.gravitee.apim.gateway.tests.sdk.annotations.GatewayTest;
 import io.gravitee.apim.gateway.tests.sdk.connector.EndpointBuilder;
 import io.gravitee.apim.gateway.tests.sdk.connector.EntrypointBuilder;
 import io.gravitee.apim.gateway.tests.sdk.parameters.GatewayDynamicConfig;
+import io.gravitee.common.http.MediaType;
+import io.gravitee.gateway.api.http.HttpHeaderNames;
 import io.gravitee.plugin.endpoint.EndpointConnectorPlugin;
 import io.gravitee.plugin.endpoint.http.proxy.HttpProxyEndpointConnectorFactory;
 import io.gravitee.plugin.entrypoint.EntrypointConnectorPlugin;
@@ -36,10 +38,12 @@ import io.vertx.core.http.HttpMethod;
 import io.vertx.rxjava3.core.http.HttpClient;
 import io.vertx.rxjava3.core.http.HttpClientRequest;
 import io.vertx.rxjava3.core.http.HttpClientResponse;
+import java.io.InputStream;
 import java.net.Socket;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.DisplayNameGeneration;
 import org.junit.jupiter.api.DisplayNameGenerator;
 import org.junit.jupiter.api.Test;
@@ -48,19 +52,39 @@ import org.junit.jupiter.api.Test;
  * End-to-end coverage that a client abort releases the upstream connection-pool slot held (or awaited) by the aborted
  * request instead of burning it until the endpoint read timeout fires.
  *
- * The API under test uses a single-connection pool so one leaked slot is directly observable as latency on the next
+ * The APIs under test use a single-connection pool so one leaked slot is directly observable as latency on the next
  * request. Aborts are raw TCP socket closes, matching what the gateway sees when a real caller gives up.
  *
  * Without abort propagation, an aborted-while-queued request is still granted the connection once it frees up and
  * holds it idle until the read timeout (8s here) — the mechanism behind pool-collapse incidents where abandoned
  * requests shed no load and the collapse outlives the backend recovery.
+ *
+ * Each scenario runs against an HTTP/1.1 and an HTTP/2 endpoint. They are not the same test twice: an HTTP/1.1 slot is
+ * a whole connection released by closing it, while an HTTP/2 slot is one stream of a multiplexed connection released
+ * by an RST_STREAM, so the release path differs even though the observable contract does not. The HTTP/2 endpoint
+ * additionally caps multiplexing at one stream, without which the spare streams of the same connection would hide
+ * a leaked one.
+ *
+ * The three abort points cover the three owners of cancellation in the connector: while queued for the pool and while
+ * in flight before the response head, both guarded by {@code doOnDispose} on the connect chain, and during response
+ * streaming, guarded by {@code doOnCancel} on the response chunks.
+ *
+ * Five of the six combinations hold; the sixth is disabled and documents a gap on HTTP/2 — see
+ * should_release_pool_slot_when_queued_request_is_aborted_on_an_http2_endpoint.
  */
 @GatewayTest
 @DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
 class ClientAbortPropagationV4IntegrationTest extends AbstractGatewayTest {
 
+    private static final String HTTP_1_1_API = "/apis/v4/http/clientabort/api-client-abort-single-connection.json";
+    private static final String HTTP_2_API = "/apis/v4/http/clientabort/api-client-abort-single-connection-http2.json";
+
     private static final String BACKEND_PATH = "/endpoint";
     private static final long PROBE_LATENCY_BUDGET_MS = 2500;
+
+    /** Long enough that a slot held for the whole streamed response blows the probe budget several times over. */
+    private static final int STREAMED_BODY_DURATION_MS = 6000;
+    private static final int STREAMED_BODY_CHUNKS = 20;
 
     @Override
     public void configureEntrypoints(Map<String, EntrypointConnectorPlugin<?, ?>> entrypoints) {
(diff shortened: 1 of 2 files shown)


gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/http/ClientAbortPropagationV4IntegrationTest.java
```diff
@@ -73,9 +97,67 @@ public void configureEndpoints(Map<String, EndpointConnectorPlugin<?, ?>> endpoi
...
+     * release path does run for an abort that happens while the request is queued — the stream granted after the
+     * abort is reset, and the connector logs it — but the multiplexing permit it consumed is never returned. The
+     * pooled connection is then permanently short one stream: with the single stream this endpoint allows, every
+     * later request answers 504 at the 8s readTimeout without ever reaching the backend, and the endpoint only
+     * recovers when the 30s idleTimeout evicts the connection. HTTP/1.1 releases the very same slot correctly, and
+     * an HTTP/2 abort after the request has been sent does too — it is the reset of a not-yet-sent stream that
+     * leaks. Enable once the permit is released.
+     */
+    @Test
+    @Disabled("Gap in APIM-14605 on HTTP/2 endpoints: an abort while queued leaks the stream's multiplexing permit")
```
paulatulis, line 117: What this test documents is a characterised production defect, not only a gap in coverage — and right now it is tracked nowhere but this javadoc.

The behaviour described is that a single client abort while queued poisons the pooled connection until `idleTimeout` (30s here), with every request in between answering 504 at `readTimeout` without ever reaching the backend. That is the same failure shape as the pool collapse this suite exists for. The PR description confirms no Jira bug is filed for it.

The consequence is that the most valuable output of this PR has no owner and is discoverable only by someone reading this file. A ticket referenced from the `@Disabled` reason would make it findable from both directions.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/http/ClientAbortPropagationV4IntegrationTest.java
```diff
@@ -133,11 +254,45 @@ private void stubSlowThenFastBackend(int slowDelayMillis) {
...
+    }
+
+    /**
+     * Same abort, but held until the response head plus a first body byte have come back, so the exchange is provably
+     * past the response head and streaming its body. Sleeping a guessed duration instead would silently degrade into
+     * the in-flight case whenever the machine is slow, and pass for the wrong reason.
+     */
+    private void abortClientRequestOnceResponseIsStreaming(int gatewayPort) throws Exception {
+        try (Socket socket = new Socket("localhost", gatewayPort)) {
+            socket.setSoTimeout(5000);
```
paulatulis, line 270: The protection against this test silently degrading into the *in-flight* / *already-completed* case rests on an undocumented arithmetic relationship.

If the response ever stops streaming (content type changed, a `Content-Length` appearing on the stub, `hasStreamingContentType` changing), the gateway aggregates and the response head only reaches this socket after the full dribble. Today that fails loudly — but only because `setSoTimeout(5000)` happens to be less than `STREAMED_BODY_DURATION_MS` (6000).

Drop that duration (an obvious "speed the suite up" edit) below the socket timeout and the same degradation becomes a silent pass instead: the abort lands after the response is complete, the slot is already free, the probe is fast, and `wiremock.verify(2, ...)` still holds. Nothing in the file records that the two values must stay in that order.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/http/ClientAbortPropagationV4IntegrationTest.java
```diff
@@ -124,6 +218,33 @@ private void stubSlowThenFastBackend(int slowDelayMillis) {
...
+        stubFastBackend();
+    }
+
+    /**
+     * First call answers its head immediately then dribbles the body over {@link #STREAMED_BODY_DURATION_MS}, which is
+     * what puts the exchange in the response-streaming phase long enough to be aborted there.
+     * <p>
+     * The content type is what makes the response stream at all: the gateway only streams a body downstream for the
+     * content types {@code RequestUtils.hasStreamingContentType} recognises, and aggregates every other response,
+     * releasing it in one piece once the backend is done — which would leave no streaming window to abort in.
```
paulatulis, line 230: This precondition is only half the rule, and the missing half explains the open `application/octet-stream` question in the PR description.

`RequestUtils.hasStreamingContentType` returns `false` whenever a `Content-Length` header is present, before it inspects the content type at all (`RequestUtils.java:99-109`). So what makes this stub stream is chunked framing — which `withChunkedDribbleDelay` forces — *together with* a recognised content type, not the content type on its own.

That also resolves the observation in the description: `application/octet-stream` is in the recognised set (`RequestUtils.java:105`); a response carrying it would not stream if it also carried a `Content-Length`.

As written, the javadoc would lead a future maintainer to believe a fixed-length `text/event-stream` body is still enough, which would remove the streaming window this test depends on.

gravitee-apim-distribution/gravitee-apim-distribution-integration-tests/src/test/java/io/gravitee/apim/integration/tests/http/ClientAbortPropagationV4IntegrationTest.java
```diff
@@ -48,20 +52,40 @@
...
+ * a whole connection released by closing it, while an HTTP/2 slot is one stream of a multiplexed connection released
+ * by an RST_STREAM, so the release path differs even though the observable contract does not. The HTTP/2 endpoint
+ * additionally caps multiplexing at one stream, without which the spare streams of the same connection would hide
+ * a leaked one.
+ *
+ * The three abort points cover the three owners of cancellation in the connector: while queued for the pool and while
+ * in flight before the response head, both guarded by {@code doOnDispose} on the connect chain, and during response
+ * streaming, guarded by {@code doOnCancel} on the response chunks.
+ *
+ * Five of the six combinations hold; the sixth is disabled and documents a gap on HTTP/2 — see
```
paulatulis, line 72: This count goes stale as soon as the disabled test is enabled or a seventh combination is added, and the gap it refers to is already stated in two other places in this file (the method javadoc and the `@Disabled` reason).

paulatulis (approved):
None of the findings are blocking.

PR #19431, opened 2026-08-27
fix(api-import): store the fetched spec, not the url, when importing from a url

wbabyte:
## Issue

https://gravitee.atlassian.net/browse/GMA-989

## Problem

Importing an API from an OpenAPI **URL** stores the URL string itself as the imported specification, instead of the document fetched from it.

`OAIDomainServiceImpl.convert()` fetches and parses the URL via `toOAIDescriptor(payload)`, but then passes the raw `payload` on to the two places that persist it. Both end up holding `https://example.com/openapi.json` where an OpenAPI document is expected:

- `addOAIDocumentation` → the generated `SWAGGER` documentation page's `content` (no `PageSource` is attached, so nothing resolves it later)
- `addOASValidationPolicy` → the `content-provider-inline-resource` backing the OAS validation policy

Anything that later parses that content fails. Reported via the AI Management "create a tool from an API" flow, which surfaces the swagger-parser message ``attribute  is not of type `object` `` — swagger-parser reads a bare URL as a valid YAML *scalar* rather than a mapping. The OAS validation policy variant is quieter: the policy validates against a URL string at runtime rather than failing loudly.

## Fix

Resolve the payload once and hand the resolved document to both consumers. An inline payload is kept verbatim, so the existing path is byte-for-byte unchanged; only URL imports change behaviour. Resolution is skipped entirely when neither consumer is going to store anything.

The document is serialized from the descriptor already parsed in `convert()`, so there is no extra network fetch.

## Verification

Reproduced and verified end-to-end against a local 4.12.x instance with the AI Management module:

| Import method | Stored spec content | MCP tool generation |
| --- | --- | --- |
| By URL, before | 48 chars — the URL string | 0 tools, `invalidSpec` |
| By URL, after | the fetched document | **19 tools**, no errors |
| Inline, after | byte-for-byte identical to input | unchanged |

Also verified: Swagger 2.0 URL imports (converted to OAS 3 by the parser) now generate tools; a URL import with both flags off does no serialization; and recursive-schema specs serialize without blowing up (swagger-parser keeps cycles as `$ref`).

Three tests added in `OAIDomainServiceImplTest.RemoteUrlPayload`, using the WireMock convention already used in this module. The two fix-specific tests fail without the production change; the third is a regression guard for the inline path and passes either way.

## Reviewer note — worth a look

For URL imports the parse runs with `resolveFully`, so the stored document has its `$ref`s inlined: **18KB → 102KB** for the Petstore spec (5.6x). Storage is `nclob` on both columns so nothing breaks, but the OAS-validation resource lives in the API definition that ships to the gateway, so this does grow the deploy payload for APIs imported by URL with that option on. It buys correctness — that policy was validating against a URL before — but if the inflation is unwelcome, the alternative is fetching the URL's raw bytes and storing the original text verbatim, at the cost of one extra HTTP GET on import. Happy to switch.

Relatedly, a v2 document imported by URL is now stored as the converted OAS 3 document rather than the URL.

Ref: GMA-989, GMA-951

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImpl.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImpl.java
@@ -97,9 +97,12 @@ public class OAIDomainServiceImpl implements OAIDomainService {
             var importWithEndpointGroupsSharedConfiguration = addEndpointGroupSharedConfiguration(importDefinition);
             var importWithGroups = replaceGroupNamesWithIds(environmentId, importWithEndpointGroupsSharedConfiguration);
             var importWithTags = replaceTagsNamesWithIds(organizationId, importWithGroups);
-            var importWithDocumentation = addOAIDocumentation(withDocumentation, payload, importWithTags);
+            // Both consumers below store their argument as an OpenAPI document, so a URL payload has to be
+            // resolved first. Skip the work when neither of them is going to keep it.
+            var specContent = (withDocumentation || withOASValidationPolicy) ? resolveSpecContent(payload, descriptor) : payload;
+            var importWithDocumentation = addOAIDocumentation(withDocumentation, specContent, importWithTags);
             boolean deferResponseValidation = shouldDeferResponseValidation(importSwaggerDescriptor);
-            return addOASValidationPolicy(withOASValidationPolicy, payload, importWithDocumentation, deferResponseValidation);
+            return addOASValidationPolicy(withOASValidationPolicy, specContent, importWithDocumentation, deferResponseValidation);
         }
 
         return null;
@@ -215,7 +218,26 @@ public class OAIDomainServiceImpl implements OAIDomainService {
             .build();
     }
 
-    private ImportDefinition addOAIDocumentation(boolean withDocumentation, String payload, ImportDefinition importWithTags) {
+    /**
+     * Returns the OpenAPI document to persist for the imported API.
+     *
+     * <p>For an inline import the payload already is the document and is kept verbatim, preserving the
+     * user's original formatting. For a URL import the payload is only a locator, so the document parsed
+     * from it is serialized instead. Note that the parse above runs with {@code resolveFully}, so a
+     * URL-imported document is stored with its {@code $ref}s inlined.
+     */
+    private String resolveSpecContent(String payload, OAIDescriptor descriptor) {
+        if (!UrlSanitizerUtils.isUrl(payload)) {
+            return payload;
+        }
+        try {
+            return descriptor.toJson();
+        } catch (JsonProcessingException e) {
+            throw new TechnicalManagementException("Error while serializing the OpenAPI specification fetched from " + payload, e);
+        }
+    }
+
+    private ImportDefinition addOAIDocumentation(boolean withDocumentation, String specContent, ImportDefinition importWithTags) {
         if (!withDocumentation) {
             return importWithTags;
         }
@@ -223,7 +245,7 @@ public class OAIDomainServiceImpl implements OAIDomainService {
             .name(DEFAULT_IMPORT_PAGE_NAME)
             .type(io.gravitee.apim.core.documentation.model.Page.Type.SWAGGER)
             .homepage(false)
-            .content(payload)
+            .content(specContent)
             .referenceType(io.gravitee.apim.core.documentation.model.Page.ReferenceType.API)
             .published(true)
             .visibility(Page.Visibility.PUBLIC)
@@ -234,7 +256,7 @@ public class OAIDomainServiceImpl implements OAIDomainService {
 
     private ImportDefinition addOASValidationPolicy(
         boolean withOASValidationPolicy,
-        String payload,
+        String specContent,
         ImportDefinition importDefinition,
         boolean deferResponseValidation
     ) {
@@ -251,7 +273,7 @@ public class OAIDomainServiceImpl implements OAIDomainService {
             Resource resource = Resource.builder()
                 .name("OpenAPI Specification")
                 .type("content-provider-inline-resource")
-                .configuration(new ObjectMapper().writeValueAsString(new LinkedHashMap<>(Map.of("content", payload))))
+                .configuration(new ObjectMapper().writeValueAsString(new LinkedHashMap<>(Map.of("content", specContent))))
                 .build();
             importDefinition.getApiExport().setResources(List.of(resource));
 
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImplTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImplTest.java
@@ -15,15 +15,21 @@
  */
 package io.gravitee.apim.infra.domain_service.api;
 
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import com.github.tomakehurst.wiremock.junit5.WireMockRuntimeInfo;
+import com.github.tomakehurst.wiremock.junit5.WireMockTest;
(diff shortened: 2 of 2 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImpl.java
```diff
@@ -215,15 +218,34 @@ private ImportDefinition replaceTagsNamesWithIds(String organizationId, ImportDe
...
+    private ImportDefinition addOAIDocumentation(boolean withDocumentation, String specContent, ImportDefinition importWithTags) {
         if (!withDocumentation) {
             return importWithTags;
         }
         var page = io.gravitee.apim.core.documentation.model.Page.builder()
             .name(DEFAULT_IMPORT_PAGE_NAME)
             .type(io.gravitee.apim.core.documentation.model.Page.Type.SWAGGER)
             .homepage(false)
-            .content(payload)
+            .content(specContent)
```
paulatulis, line 248: This snapshots the fetched document, where the v2 import path does the opposite for the same situation. `PageServiceImpl.createOrUpdateSwaggerPage` stores content only when the descriptor type is `INLINE`; for a URL it attaches an `http-fetcher` `PageSource` and leaves the content empty. That is also the page shape the AI Management module's `resolveContent()` guard (GMA-158) was written against.

So after this change, the same OpenAPI URL imported through v2 yields a page that tracks the remote spec, and through v4 a frozen snapshot with its `$ref`s inlined. `io.gravitee.apim.core.documentation.model.Page` carries a `source` field and the console renders `http-fetcher` sources, so the linked variant is reachable here too.

I am not arguing snapshot is the wrong call — it is simpler, works when the remote is unreachable, and cannot shift underneath an API that is already deployed. But the divergence is deliberate enough to deserve a sentence in the PR description saying why, since the Jira offered attaching a `PageSource` as the other candidate fix.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImpl.java
```diff
@@ -251,7 +273,7 @@ private ImportDefinition addOASValidationPolicy(
             Resource resource = Resource.builder()
                 .name("OpenAPI Specification")
                 .type("content-provider-inline-resource")
-                .configuration(new ObjectMapper().writeValueAsString(new LinkedHashMap<>(Map.of("content", payload))))
+                .configuration(new ObjectMapper().writeValueAsString(new LinkedHashMap<>(Map.of("content", specContent))))
```
paulatulis, line 276: On your reviewer note about the 18KB to 102KB inflation — I would split the answer rather than take or leave it wholesale, because the two consumers pay very different prices for it.

The documentation page is `nclob` read by the console. 102KB there is unremarkable.

This one is not equivalent. The resource lives in the API definition, so the inflated document is not stored once: it is replicated into every gateway sync event, every API export, and every CRD for the lifetime of the API. That is the half where preserving `$ref`s earns its keep, and it is also the half where the extra HTTP GET you offered as the alternative is cheapest — it happens once, at import.

So my suggestion is to keep the snapshot as-is for the page and only reconsider the serialization for this resource, rather than changing both or neither.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImpl.java
```diff
@@ -215,15 +218,34 @@ private ImportDefinition replaceTagsNamesWithIds(String organizationId, ImportDe
...
+     * URL-imported document is stored with its {@code $ref}s inlined.
+     */
+    private String resolveSpecContent(String payload, OAIDescriptor descriptor) {
+        if (!UrlSanitizerUtils.isUrl(payload)) {
+            return payload;
+        }
+        try {
+            return descriptor.toJson();
+        } catch (JsonProcessingException e) {
+            throw new TechnicalManagementException("Error while serializing the OpenAPI specification fetched from " + payload, e);
```
paulatulis, line 236: This surfaces a bad remote document as a 500. Every other path in this class that rejects a remote document throws `SwaggerDescriptorException`, which maps to a 400 — `should_throw_clean_exception_when_remote_url_is_unreachable` a few lines up in the test file asserts precisely that contract, down to not leaking internals into the message.

The way in is a self-referential schema that survives `resolveFully` as a direct object cycle rather than as a `$ref`; Jackson then fails with `Direct self-reference leading to cycle`, which is a `JsonProcessingException` and lands here. You mention recursive-schema specs serialize fine on swagger-parser 2.1.45, so this may not be reachable in practice today — but it is reachable by construction, and it is a remote document deciding the status code.

Noting the counter-argument: `addOASValidationPolicy` below already wraps `JsonProcessingException` this way. The difference is that it serializes a two-entry map that cannot realistically fail, whereas this one serializes a document fetched from a URL the caller supplied.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImplTest.java
```diff
@@ -202,4 +208,106 @@ void should_not_defer_response_validation_for_non_wsdl_format() {
...
+            var url = stubSpecUrl(wm);
+            var descriptor = ImportSwaggerDescriptorEntity.builder().payload(url).build();
+
+            var result = service.convert(ORGANIZATION_ID, ENVIRONMENT_ID, descriptor, true, false);
+
+            assertThat(result).isNotNull();
+            assertThat(result.getPages()).hasSize(1);
+            var page = result.getPages().getFirst();
+            assertThat(page.getType()).isEqualTo(Page.Type.SWAGGER);
+            assertThat(page.getContent()).isNotEqualTo(url).contains("\"openapi\"").contains("listPets");
```
paulatulis, line 286: The defect being fixed is that the stored content is not a parseable OpenAPI document, but these assertions only establish that it is not the URL and that the string `listPets` appears somewhere in it. A truncated document, a half-serialized one, or one that lost its `paths` would satisfy both — as would, in principle, a document that still carried the URL in a `servers` entry.

Round-tripping the stored content back through `OpenAPIParser` and asserting the operation survives would test the property the bug was actually about. Same applies to the resource-configuration assertion in the test below.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImplTest.java
```diff
@@ -202,4 +208,106 @@ void should_not_defer_response_validation_for_non_wsdl_format() {
...
+
+            assertThat(result).isNotNull();
+            assertThat(result.getApiExport().getResources()).hasSize(1);
+            var configuration = result.getApiExport().getResources().getFirst().getConfiguration();
+            assertThat(configuration).isNotNull();
+            assertThat(configuration.toString()).doesNotContain(url).contains("listPets");
+        }
+
+        @Test
+        void should_keep_an_inline_payload_verbatim(WireMockRuntimeInfo wm) {
```
paulatulis, line 304: `wm` is unused here — this test stubs nothing and takes the inline payload directly.

```suggestion
        void should_keep_an_inline_payload_verbatim() {
```

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/domain_service/api/OAIDomainServiceImpl.java
```diff
@@ -97,9 +97,12 @@ public ImportDefinition convert(
             var importWithEndpointGroupsSharedConfiguration = addEndpointGroupSharedConfiguration(importDefinition);
             var importWithGroups = replaceGroupNamesWithIds(environmentId, importWithEndpointGroupsSharedConfiguration);
             var importWithTags = replaceTagsNamesWithIds(organizationId, importWithGroups);
-            var importWithDocumentation = addOAIDocumentation(withDocumentation, payload, importWithTags);
+            // Both consumers below store their argument as an OpenAPI document, so a URL payload has to be
+            // resolved first. Skip the work when neither of them is going to keep it.
+            var specContent = (withDocumentation || withOASValidationPolicy) ? resolveSpecContent(payload, descriptor) : payload;
```
paulatulis, line 102: Worth confirming the rollout story: this corrects new imports, but APIs already imported from a URL keep the URL in their `SWAGGER` page content and in their `content-provider-inline-resource`, and nothing here migrates them. They stay broken after an upgrade, and for the OAS-validation variant that means a policy that has been quietly validating against a URL string carries on doing so.

Re-import is probably the right remedy rather than an upgrader, but since the symptom is silent in the policy case, users will not know they need to. A line in the release note naming both artifacts would close that gap.

paulatulis (approved):
None are blockers.

PR #19435, opened 2026-08-27
feat(gateway): expose the Edge server TLS configuration in gravitee.yml and the Helm chart

ytvnr:
## Issue

https://gravitee.atlassian.net/browse/EDGE-107

## Description

EDGE-107 moves the Edge Reactor listener onto the node server stack, so it now accepts the full `edge.*` server surface — TLS, keystore, truststore, CRL, hot reload. Neither `gravitee.yml` nor the Helm chart exposed any of it: both documented `edge.server.port` alone, which is precisely the property the reactor now **deprecates** in favour of `edge.port`. Anyone reading either file configured the wrong key and got a startup warning.

This exposes the real surface in both places:

- **`gravitee.yml`** — commented `edge:` block with `port`, `host`, `secured` and `ssl.{clientAuth,keystore,truststore}`, aligned with the `http:`, `tcp:` and `kafka:` blocks around it. `server.port` is kept, documented as deprecated.
- **`helm/values.yaml`** — same surface under the commented `gateway.edge` block, with `secret://kubernetes/...` examples matching the `bridge` convention.
- **`helm/templates/gateway/gateway-configmap.yaml`** — renders `edge.port`, `edge.host`, `edge.secured` and `edge.ssl.*`, and still renders `edge.server.port` when set, so existing values files keep working.

Two things the comments make explicit, because they are counter-intuitive:

- The Edge truststore holds the **fleet CA**, not the enrolled leaf certificates — contrary to the rest of APIM. The Edge server is private to its plugin and is not registered in the node `ServerManager`, so per-subscription trust material never reaches it. TLS validates the chain and the mTLS policy still does the exact thumbprint match against the subscription.
- With `clientAuth: none` on a secured listener, no client certificate is ever requested, so an mTLS plan can never match. Calls then fall through to the keyless plan and are served **unauthenticated** — or get a 401 if no keyless plan remains. Both outcomes defeat the plan.

## Additional context

Follow-up to #19429. Verified with `helm lint` and `helm template` on both value shapes — canonical `gateway.edge.port` + TLS, and the legacy `gateway.edge.server.port` alone.


helm/templates/gateway/gateway-configmap.yaml
```diff
@@ -345,12 +345,25 @@ data:
     {{- if hasKey .Values.gateway "edge" }}
     {{- if .Values.gateway.edge.enabled }}
     edge:
+      {{- if hasKey .Values.gateway.edge "port" }}
+      port: {{ .Values.gateway.edge.port }}
+      {{- end }}
+      {{- if hasKey .Values.gateway.edge "host" }}
+      host: {{ .Values.gateway.edge.host }}
+      {{- end }}
+      {{- if .Values.gateway.edge.secured }}
```
Okhelifi: I remember that we had issues with the boolean and how we handle it in the helm chart. Some times we use the hasKey function is it something we need here as there is no default value in the Values.yml ?

helm/values.yaml
```diff
@@ -1460,10 +1460,37 @@ gateway:
...
+#    secured: false
+#    ssl:
+#      # Supports none, request, required.
+#      # With `none` no client certificate is ever requested: an mTLS plan on the Edge API can never match, so
+#      # calls fall through to the keyless plan and are served unauthenticated - or get a 401 if no keyless plan
+#      # remains. Use `request` on dev/test, `required` on production.
+#      clientAuth: none
+#      keystore:
+#        type: pem # Supports jks, pem, pkcs12, self-signed
+#        path: /opt/graviteeio-gateway/security/edge-keystore.jks
```
Okhelifi: The keystore example says type: pem but points at a .jks with a password here and the gravitee.yml I think maybe we can change it wdyt ?

Okhelifi (commented):
When updating the helm chart I think you should update manually the chart.yml et the changelog.yml they are published in the artifacthub and not generated automagically https://artifacthub.io/packages/helm/graviteeio/apim

Okhelifi (approved):
Pre-approved 😉

helm/templates/gateway/gateway-configmap.yaml
ytvnr: Good catch, fixed in c26c55e.

You're right that this is the `hasKey` case. The chart already draws the line: a plain `if` where the boolean has a default in `values.yaml` (`gateway.ssl.enabled`), `hasKey` where it does not (`keystore.watch`, `crl.crl.watch`). `gateway.edge.secured` has no default, so it belongs to the second group and I had it in the first.

It now renders the actual value instead of a hardcoded `secured: true`, so an explicit `secured: false` is emitted rather than silently omitted. Added a unit test for that case.

helm/values.yaml
ytvnr: You're right, that example was incoherent — `type: pem` with a `.jks` path and a password mixes two formats that don't go together.

Aligned on the pattern already used by the `http` and `bridge` blocks: `type: jks` with `path` + `password`, and a commented `certificates:` entry for the `pem` case. Fixed in both `values.yaml` and the distribution `gravitee.yml` (c26c55e).

ytvnr:
Thanks, I wasn't aware of that — done in c26c55e.

Added an entry under `### 4.13.0` in `helm/CHANGELOG.md` and one to the `artifacthub.io/changes` annotation in `helm/Chart.yaml`, with a link back to this PR. I left `version` / `appVersion` at `4.13.0`, since recent chart changes accumulate entries under the current version rather than bumping it — tell me if that's wrong.

PR #19437, opened 2026-08-27
fix(portal): default null visibility to PUBLIC when mapping legacy nav items

dev-marek:
**Issue**

https://gravitee.atlassian.net/browse/PORTAL-159

**Purpose**

Fix a NullPointerException that aborts `EnvironmentsDefaultPortalNavigationItemsUpgrader` when it reads a portal navigation item with a null `visibility`.

**Security-sensitive**

This change touches `visibility`, which gates portal exposure (`PortalNavigationItemViewerContext.shouldNotShow`) — i.e. whether an item is reachable by unauthenticated portal viewers.

**Summary of changes**

- `PortalNavigationItemAdapter` now maps `visibility` through a new `repositoryVisibilityToDomain` helper that defaults a null/missing value to `PUBLIC` (with a warning log identifying the item), instead of passing it straight into the `@Nonnull`-annotated domain model.
- Applied to all five `*FromRepository` mapping methods (Folder, Page, Link, Api, ApiProduct).
- Added a Mongo backfill upgrader (`PortalNavigationItemDefaultVisibilityMongoUpgrader`) that sets null/missing `visibility` to `PUBLIC` directly in MongoDB, following the same pattern as the existing `rootId`/`segment`/`reference` backfills. This closes a divergence where the DB-filtered `searchItems` path excluded null-visibility items for anonymous portal viewers while the Java-side `filterForViewer` path (which relies on the adapter's default) included them.
- Added a companion `PortalNavigationItemVisibilityIndexUpgrader` (the backfill query needs an index on `visibility` to run under Mongo's `notablescan` guard, same as `rootId`/`segment`).
- Added `should_default_null_visibility_to_public_for_every_type` and `should_map_private_visibility_to_private_for_every_type` tests, covering all five navigation item subtypes for both the null-default path and the existing PRIVATE mapping.

**Out of scope**

The same `@Nonnull`-from-nullable-Mongo issue on other fields (`order`, `area`, `title`) — these have always been present/non-null on every persisted item, unlike `visibility`, which was added to the schema between full releases, so this failure mode doesn't apply to them today.

**Testing**

Added unit tests reproducing the exact reported production error and covering the PRIVATE branch. Verified RED → GREEN manually: temporarily removed the `@Mapping` override for one type, confirmed the test throws the exact production `NullPointerException` (`visibility is marked non-null but is null`), then restored the fix and confirmed all 33 tests in `PortalNavigationItemAdapterTest` pass. Also added `PortalNavigationItemDefaultVisibilityMongoUpgraderTest`, covering missing/null/untouched/idempotent backfill cases against a real MongoDB (via `AbstractManagementRepositoryTest`).

**Additional context**

Production error this fixes:
```
io.gravitee.node.api.upgrader.UpgraderException: java.lang.NullPointerException: visibility is marked non-null but is null
	at io.gravitee.node.api.upgrader.Upgrader.wrapException(Upgrader.java:68)
```

No UI/endpoint involved — background upgrader/mapper fix, so no screenshot/Postman recording applies.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapter.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapter.java
@@ -71,7 +71,7 @@ public interface PortalNavigationItemAdapter {
 
     @Mapping(target = "url", expression = "java(parseUrl(portalNavigationItem.getConfiguration()))")
     @Mapping(target = "rootId", source = "rootId", qualifiedByName = "repositoryRootIdToDomain")
-    @Mapping(target = "visibility", source = "visibility", qualifiedByName = "repositoryVisibilityToDomain")
+    @Mapping(target = "visibility", expression = "java(repositoryVisibilityToDomain(portalNavigationItem))")
     @Mapping(target = "reference", expression = "java(referenceFromRepository(portalNavigationItem))")
     @Mapping(
         target = "automationMetadata",
@@ -82,7 +82,7 @@ public interface PortalNavigationItemAdapter {
     );
 
     @Mapping(target = "rootId", source = "rootId", qualifiedByName = "repositoryRootIdToDomain")
-    @Mapping(target = "visibility", source = "visibility", qualifiedByName = "repositoryVisibilityToDomain")
+    @Mapping(target = "visibility", expression = "java(repositoryVisibilityToDomain(portalNavigationItem))")
     @Mapping(target = "reference", expression = "java(referenceFromRepository(portalNavigationItem))")
     @Mapping(
         target = "automationMetadata",
@@ -93,7 +93,7 @@ public interface PortalNavigationItemAdapter {
     );
 
     @Mapping(target = "rootId", source = "rootId", qualifiedByName = "repositoryRootIdToDomain")
-    @Mapping(target = "visibility", source = "visibility", qualifiedByName = "repositoryVisibilityToDomain")
+    @Mapping(target = "visibility", expression = "java(repositoryVisibilityToDomain(portalNavigationItem))")
     @Mapping(target = "reference", expression = "java(referenceFromRepository(portalNavigationItem))")
     @Mapping(
         target = "automationMetadata",
@@ -105,7 +105,7 @@ public interface PortalNavigationItemAdapter {
 
     @Mapping(target = "portalPageContentId", expression = "java(parsePortalPageContentId(portalNavigationItem.getConfiguration()))")
     @Mapping(target = "rootId", source = "rootId", qualifiedByName = "repositoryRootIdToDomain")
-    @Mapping(target = "visibility", source = "visibility", qualifiedByName = "repositoryVisibilityToDomain")
+    @Mapping(target = "visibility", expression = "java(repositoryVisibilityToDomain(portalNavigationItem))")
     @Mapping(target = "source", expression = "java(sourceFromRepository(portalNavigationItem))")
     @Mapping(target = "reference", expression = "java(referenceFromRepository(portalNavigationItem))")
     @Mapping(
@@ -356,22 +356,25 @@ public interface PortalNavigationItemAdapter {
         return PortalNavigationItemId.zero();
     }
 
-    @Named("repositoryVisibilityToDomain")
     default PortalVisibility repositoryVisibilityToDomain(
-        io.gravitee.repository.management.model.PortalNavigationItem.Visibility visibility
+        io.gravitee.repository.management.model.PortalNavigationItem portalNavigationItem
     ) {
+        var visibility = portalNavigationItem.getVisibility();
         if (visibility != null) {
             return switch (visibility) {
                 case PUBLIC -> PortalVisibility.PUBLIC;
                 case PRIVATE -> PortalVisibility.PRIVATE;
             };
         }
-        log.warn("Portal navigation item has null visibility; defaulting to PUBLIC.");
+        log.warn(
+            "Portal navigation item [{}] has null visibility; defaulting to PUBLIC. This is unexpected after visibility backfill.",
+            portalNavigationItem.getId()
+        );
         return PortalVisibility.PUBLIC;
     }
 
     @Mapping(target = "rootId", source = "rootId", qualifiedByName = "repositoryRootIdToDomain")
-    @Mapping(target = "visibility", source = "visibility", qualifiedByName = "repositoryVisibilityToDomain")
+    @Mapping(target = "visibility", expression = "java(repositoryVisibilityToDomain(portalNavigationItem))")
     @Mapping(target = "source", expression = "java(sourceFromRepository(portalNavigationItem))")
     @Mapping(target = "reference", expression = "java(referenceFromRepository(portalNavigationItem))")
     @Mapping(
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapter.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapter.java
@@ -71,6 +71,7 @@ public interface PortalNavigationItemAdapter {
 
     @Mapping(target = "url", expression = "java(parseUrl(portalNavigationItem.getConfiguration()))")
     @Mapping(target = "rootId", source = "rootId", qualifiedByName = "repositoryRootIdToDomain")
+    @Mapping(target = "visibility", source = "visibility", qualifiedByName = "repositoryVisibilityToDomain")
     @Mapping(target = "reference", expression = "java(referenceFromRepository(portalNavigationItem))")
     @Mapping(
         target = "automationMetadata",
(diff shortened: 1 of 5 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapterTest.java
```diff
@@ -191,6 +191,43 @@ void should_map_blank_or_empty_rootId_to_zero() {
             assertThat(((PortalNavigationFolder) entity).getRootId()).isEqualTo(PortalNavigationItemId.zero());
         }
 
+        /**
+         * Items persisted before visibility was mandatory (e.g. MongoDB, which has no NOT NULL
+         * constraint) can still have a null visibility. Defaulting to PUBLIC preserves the pre-existing
+         * behaviour where all items were visible.
+         */
+        @Test
+        void should_default_null_visibility_to_public_for_every_type() {
```
thomasballivet, line 200: The hand-written switch replaces MapStruct's generated enum mapping, but no test covers the `PRIVATE` branch — every fixture in `PortalNavigationItemsRepositoryFixtures` is `PUBLIC`. A wrong `PRIVATE -> PortalVisibility.PRIVATE` would still pass all 32 tests, and the resulting bug would leak private items to anonymous portal viewers. Could you add a `PRIVATE → PRIVATE` assertion alongside the null one?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapter.java
```diff
@@ -352,7 +356,22 @@ default PortalNavigationItemId repositoryRootIdToDomain(String rootId) {
         return PortalNavigationItemId.zero();
     }
 
+    @Named("repositoryVisibilityToDomain")
```
thomasballivet: Without a data backfill, the two listing paths now disagree for anonymous viewers: `searchItems` filters in the DB (`MongoPortalNavigationItemRepository:145`) and excludes null-visibility docs, while `filterForViewer` (`ListPortalNavigationItemsUseCase:246`) maps them to `PUBLIC` and includes them — same item, visible in one listing, invisible in the other. We already have `PortalNavigationItemDefaultRootIdMongoUpgrader` / `…DefaultSegmentMongoUpgrader` / `…DefaultReferenceMongoUpgrader` for exactly this; a fourth one is ~15 lines and removes the divergence. Was skipping it a deliberate call?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapter.java
```diff
@@ -352,7 +356,22 @@ default PortalNavigationItemId repositoryRootIdToDomain(String rootId) {
...
+    default PortalVisibility repositoryVisibilityToDomain(
+        io.gravitee.repository.management.model.PortalNavigationItem.Visibility visibility
+    ) {
+        if (visibility != null) {
+            return switch (visibility) {
+                case PUBLIC -> PortalVisibility.PUBLIC;
+                case PRIVATE -> PortalVisibility.PRIVATE;
+            };
+        }
+        log.warn("Portal navigation item has null visibility; defaulting to PUBLIC.");
```
thomasballivet: This warns on every mapping, so once per affected item per portal request, forever — since nothing ever fixes the stored value. The `rootId` warn just above can afford `warn` because the backfill makes it genuinely exceptional ("This is unexpected after rootId backfill"); here it isn't. Either add the backfill or drop this to `debug`.

thomasballivet: Message has no item id, so it can't actually be acted on — passing the whole item through an `expression` mapping would let us log which one, wdyt?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapter.java
```diff
@@ -71,6 +71,7 @@ default io.gravitee.repository.management.model.PortalNavigationItem.Area mapAre
 
     @Mapping(target = "url", expression = "java(parseUrl(portalNavigationItem.getConfiguration()))")
     @Mapping(target = "rootId", source = "rootId", qualifiedByName = "repositoryRootIdToDomain")
+    @Mapping(target = "visibility", source = "visibility", qualifiedByName = "repositoryVisibilityToDomain")
```
thomasballivet: MapStruct usually auto-selects a matching source→target method even when it's `@Named`, which would make these five `qualifiedByName` lines redundant — did removing only the annotations really reproduce the NPE, or did the helper go too? wdyt about dropping them if they're not needed?

thomasballivet (commented):
Nice, focused fix and the root cause analysis in the description matches what I see in the code — JDBC has `visibility NOT NULL` since `v4_10_0/02_add_portal_navigation_items_table.yml`, so only Mongo documents can carry a null. A few comments inline.

Two general points:

- Same class of bug is still open on the other `@Nonnull` domain fields fed from nullable Mongo columns — `order`, `area`, `title` (`segment`/`rootId` are backfilled, `published` is a primitive) — worth a follow-up ticket rather than widening this PR, wdyt?
- This touches `visibility`, which gates portal exposure (`PortalNavigationItemViewerContext:69`), so per AGENTS.md it should be flagged as security-sensitive in the PR description — wdyt about adding a line?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapterTest.java
dev-marek: Added `should_map_private_visibility_to_private_for_every_type`, mirroring the null-visibility test — asserts PRIVATE → PRIVATE for all five subtypes.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapter.java
dev-marek: Not deliberate — added `PortalNavigationItemDefaultVisibilityMongoUpgrader`, mirroring `...DefaultRootIdMongoUpgrader`/`...DefaultSegmentMongoUpgrader`/`...DefaultReferenceMongoUpgrader`. It backfills null/missing `visibility` to `PUBLIC` directly in Mongo, so `searchItems`'s DB-side filter and `filterForViewer`'s Java-side default now agree — there's no null-visibility document left after startup for the two paths to disagree over. Also added `PortalNavigationItemVisibilityIndexUpgrader`, since the backfill query needs an index to run under `notablescan` — caught this because the new upgrader's own test failed without it.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapter.java
dev-marek: Addressed by the backfill above — once it runs, this warn becomes the same "unexpected after backfill" case as the existing `rootId` warn, so I kept it at `warn` rather than downgrading to `debug`.

dev-marek: Fixed — `repositoryVisibilityToDomain` now takes the whole `PortalNavigationItem` (matching `referenceFromRepository`'s pattern: whole item passed via `expression` instead of `qualifiedByName` on just the field), so the warn now reads `Portal navigation item [{}] has null visibility; defaulting to PUBLIC. This is unexpected after visibility backfill.`

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/infra/adapter/PortalNavigationItemAdapter.java
dev-marek: Verified empirically: removing just the `@Mapping` overrides (keeping the helper as dead code) does reproduce the exact production NPE — `visibility is marked non-null but is null` — immediately at construction time, without touching the helper method. So the original RED repro was accurate as described.

On necessity: this codebase's MapStruct config doesn't auto-select `@Named`/qualified methods by type. Evidence: `id` and `rootId` both map `String → PortalNavigationItemId` with identical signatures, yet `id`'s unqualified mapping never resolves to `repositoryRootIdToDomain` — only the plain `mapPortalNavigationItemId(String)` gets picked implicitly. So explicit wiring is required here, not just stylistic — kept it (now via `expression`, to get the item id into the log).

dev-marek:
Re: the order/area/title follow-up in the review — `visibility` was added to the schema between full releases, so legacy documents predating it can genuinely carry a null value today on dev/master/qa — a live migration gap, not a hypothetical one. `order`, `area`, and `title` have always been present and non-null on every persisted item, so the same failure mode doesn't apply to them — I don't think a follow-up ticket is warranted here.

Also pushed 3 commits addressing the inline comments (PRIVATE test coverage, item id in the warn log, and a Mongo backfill + index upgrader for visibility) and flagged the PR description as security-sensitive per AGENTS.md.

thomasballivet (approved):

PR #19439, opened 2026-08-27
fix(gamma): improve env level filter in sidenav

arpzilla-dev:
## Issue
https://gravitee.atlassian.net/browse/FOUND-245

## Description

Align Gamma Platform nav and pasted URLs with Classic permission gating.
Organization (and Team Users) stays hidden unless the user has `organization-settings-r` or `-u`. Remaining items then require their own `-r`. Users with no visible menus land on an educational no-access page (ask an admin for an environment or organization role) instead of a forbidden screen. Empty context sidebar is hidden. Login/index prefers Applications when `environment-application-r` is present.
This is security-sensitive (authorization / nav gating). Roles UI, licenses, and Cloud multi-tenant are out of this PR.

## Additional context

Verified in multi tenant setup with 3 envs [DEV, TEST, DEFAULT] 

- admin scenario [excess to everywhere]: 

https://github.com/user-attachments/assets/84f55258-4660-48e0-a2bc-522da879925e

- User "envagent@local.test" with access to "DEV" env, & not [TEST/DEFAULT]
[dont have any platform related access]

https://github.com/user-attachments/assets/1af0e842-475b-4bab-8e2a-af16d1589d23

- User "envsplit@local.test" with ENVIRONMENT:USER on DEV & ENVIRONMENT:ADMIN on Test

https://github.com/user-attachments/assets/00c99a9e-5cd7-41fb-83b4-ab013837bd31

### DIFF
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { permissionService, useHasFeature, useHasPermission } from '@gravitee/gamma-modules-sdk';
+import { permissionService, useEnvironment, useHasFeature, useHasPermission } from '@gravitee/gamma-modules-sdk';
 import { useModuleRouting } from '@gravitee/gamma-modules-sdk/routing';
 import {
     buildLinearBreadcrumbs,
@@ -30,7 +30,7 @@ import {
     useLayoutConfig,
 } from '@gravitee/graphene-core';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { createContext, type ReactElement, useCallback, useContext, useMemo, useState } from 'react';
+import { createContext, type ReactElement, useCallback, useContext, useEffect, useMemo, useState } from 'react';
 import { Navigate, Outlet, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
 
 import { PlatformToaster } from './PlatformToaster';
@@ -52,6 +52,7 @@ import { AlertFormPage } from '../features/alerts/pages/AlertFormPage';
 import { AlertsActivityPage } from '../features/alerts/pages/AlertsActivityPage';
 import { ALERT_ENGINE_FEATURE } from '../features/alerts/utils/alertPermissions';
 import { ApplicationDetailIndexRedirect, ApplicationDetailLayout } from '../features/applications/components/detail';
+import { APIM_AUDIT_TRAIL_FEATURE } from '../features/audit-logs/license/auditTrailLicense';
 import { useEnvironmentDictionaries } from '../features/dictionaries/hooks/useEnvironmentDictionaries';
 import { GatewayInstanceDetailLayout } from '../features/gateway-instances/components/GatewayInstanceDetailLayout';
 import { useEnvironmentMetadata } from '../features/metadata/hooks/useEnvironmentMetadata';
@@ -98,6 +99,7 @@ import {
     useEnvironmentPermissions,
     useEnvironmentPermissionsReady,
 } from '../shared/hooks/useEnvironmentPermissions';
+import { resetDeniedNavItemsForEnvironment, useDeniedNavItemKeys } from '../shared/nav/deniedNavItems';
 import { isForbiddenApiError } from '../shared/utils/apiErrors';
 
 const queryClient = new QueryClient({
@@ -109,8 +111,8 @@ const queryClient = new QueryClient({
 
 const APPLICATION_DETAIL_TABS = flattenApplicationDetailNavItems(APPLICATION_NAV_GROUPS);
 const EMPTY_NAV_GROUPS: NavGroup[] = [];
-const NO_LOCKED_NAV_ITEMS: readonly string[] = [];
-const ALERT_ENGINE_LOCKED_NAV_ITEMS: readonly string[] = ['alerts'];
+const ALERT_ENGINE_NAV_ITEMS: readonly string[] = ['alerts'];
+const AUDIT_TRAIL_NAV_ITEMS: readonly string[] = ['organization-audit', 'environment-audit'];
 
 function resolveRequiredPermissions(permission?: string, anyOf?: readonly string[]): readonly string[] {
     if (anyOf) {
@@ -181,6 +183,12 @@ function NavPermissionGuard({
     children,
 }: Readonly<{ itemKey: string; unauthorizedTo?: string; children: ReactElement }>) {
     const { anyOf, alsoAnyOf } = pageGuardForNavItem(itemKey);
+    const deniedNavItemKeys = useDeniedNavItemKeys();
+    // A live 403 outranks the permission map, which can still grant the item from a scope this module
+    // does not own. Without this the page renders, 403s, and redirects on every visit.
+    if (deniedNavItemKeys.has(itemKey)) {
+        return <UnauthorizedRedirect />;
+    }
     return (
         <PermissionPageGuard anyOf={anyOf} alsoAnyOf={alsoAnyOf} unauthorizedTo={unauthorizedTo}>
             {children}
@@ -255,6 +263,12 @@ function usePlatformNavContext(): PlatformNavContextValue {
 function ModuleLayout() {
     useEnvironmentPermissions();
     const permissionVersion = usePermissionServiceSnapshot();
+    const env = useEnvironment();
+    const deniedNavItemKeys = useDeniedNavItemKeys();
+
+    useEffect(() => {
+        resetDeniedNavItemsForEnvironment(env?.id);
+    }, [env?.id]);
 
     const permissionsReady = useEnvironmentPermissionsReady();
     const [contextExpanded, setContextExpanded] = useState(true);
@@ -273,6 +287,14 @@ function ModuleLayout() {
 
     const { activeNavKey, navigateToKey } = useModuleRouting(PLATFORM_ROUTE_CONFIG);
     const hasAlertEngine = useHasFeature(ALERT_ENGINE_FEATURE);
+    const hasAuditTrail = useHasFeature(APIM_AUDIT_TRAIL_FEATURE);
+
+    // Unlicensed pages redirect away or open an upsell dialog, so landing on one bounces the user
+    // straight back out. Alerts redirects; the audit pages show the dialog and cannot be dismissed.
+    const lockedItemKeys = useMemo(
+        () => [...(hasAlertEngine ? [] : ALERT_ENGINE_NAV_ITEMS), ...(hasAuditTrail ? [] : AUDIT_TRAIL_NAV_ITEMS)],
+        [hasAlertEngine, hasAuditTrail],
+    );
 
     // permissionService is an external store, so re-reading it has to be keyed on its version:
     // a 403 patch or a host reload changes the answers without changing any React state below.
@@ -284,9 +306,10 @@ function ModuleLayout() {
             has: hasPermission,
             metadataForbidden,
             dictionariesForbidden,
-            lockedItemKeys: hasAlertEngine ? NO_LOCKED_NAV_ITEMS : ALERT_ENGINE_LOCKED_NAV_ITEMS,
+            lockedItemKeys,
+            deniedItemKeys: deniedNavItemKeys,
         }),
-        [dictionariesForbidden, hasAlertEngine, hasPermission, metadataForbidden, permissionsReady],
+        [deniedNavItemKeys, dictionariesForbidden, hasPermission, lockedItemKeys, metadataForbidden, permissionsReady],
     );
 
     const visibleNavSections = useMemo(
--- a/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
+++ b/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
@@ -13,12 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { useHasFeature, useHasPermission } from '@gravitee/gamma-modules-sdk';
+import { permissionService, useHasFeature, useHasPermission } from '@gravitee/gamma-modules-sdk';
 import { useModuleRouting } from '@gravitee/gamma-modules-sdk/routing';
 import {
     buildLinearBreadcrumbs,
     ContextSidebar,
     ContextToggleButton,
+    type NavGroup,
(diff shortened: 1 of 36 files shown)


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useEnvironmentPermissions.ts
```diff
@@ -20,42 +20,44 @@ import { useEffect } from 'react';
...
-    }, [permissions]);
+    }, [isFetchedAfterMount, permissions]);
 }
 
 /**
- * True once environment-scoped permissions are loaded into {@link permissionService}.
+ * True once environment-scoped permissions for this mount have been fetched.
  * Shares the query with {@link useEnvironmentPermissions} (deduped by React Query).
  */
 export function useEnvironmentPermissionsReady(): boolean {
```
aqibmohammadkhan, line 107: useEnvironmentPermissionsReady() now gates on isFetchedAfterMount, a per-observer flag - every fresh mount of a component that calls it forces a new network refetch and reports not-ready even though permissions are already cached and unchanged.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useEnvironmentPermissions.spec.tsx
aqibmohammadkhan, line 1: The new spec only covers the single-observer, cross-login-cache scenario the fix was aimed at; it never exercises a second, later-mounted useEnvironmentPermissionsReady() caller (e.g. a route guard remounting on navigation), which is exactly the case that regresses (see the correctness finding on useEnvironmentPermissions.ts).

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
```diff
@@ -114,104 +128,41 @@ function resolveRequiredPermissions(permission?: string, anyOf?: readonly string
...
-}: Readonly<{ permission?: string; anyOf?: readonly string[]; unauthorizedTo?: string; children: ReactElement }>) {
+}: Readonly<{
+    permission?: string;
+    anyOf?: readonly string[];
+    alsoAnyOf?: readonly string[];
+    unauthorizedTo?: string;
+    children: ReactElement;
+}>) {
     const required = resolveRequiredPermissions(permission, anyOf);
+    const extra = alsoAnyOf ?? required;
```
aqibmohammadkhan: PermissionPageGuard calls useHasPermission twice with the identical anyOf array whenever alsoAnyOf is omitted (the common case — most routes only pass permission/anyOf), since `extra = alsoAnyOf ?? required` falls back to the same array.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useEnvironmentPermissions.ts
arpzilla-dev: Agreed — that was a real bug. `isFetchedAfterMount` is per-observer, so a remounted route guard looked unready and refetched even with a valid cache.

Ready is `isSuccess` again. Previous-login cache is still dropped once per layout instance, so the user-switch leak stays closed.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useEnvironmentPermissions.spec.tsx
arpzilla-dev: Added `stays ready for a later observer without refetching`. After the layout fetch completes, a second `useEnvironmentPermissionsReady()` (as a remounted route guard would) is immediately true and does not call the API again.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
arpzilla-dev: Fixed
and Thanks. When `alsoAnyOf` is omitted we no longer pass the same `anyOf` list into the extra check — it is an empty list and ignored. The second hook still has to run because hooks cannot be conditional.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/app/AppRoutes.tsx
```diff
@@ -274,67 +227,114 @@ function ModuleLayout() {
...
-                    canReadAlerts,
-                    canReadTenants,
-                    canAccessPlatformPolicies,
-                    canReadOrgAudit,
-                    canReadEnvAudit,
-                    canReadOrgSettings,
-                    canReadIdentityProviders,
-                }),
-            ),
+    const hasPermission = useCallback(
```
arpzilla-dev: Out of scope for this PR. Visibility and gating already live in `navVisibility.ts`; `ModuleLayout` only maps SDK hooks into that model. Splitting the layout further would not change the env-level filter behavior.

@AikidoSec ignore: Visibility policy is already extracted to navVisibility.ts; further ModuleLayout split is out of scope for this change.

aqibmohammadkhan (commented):
Some issues:

1. Access Management is env-scoped, but hidden without org settings

access-management is in ORGANIZATION_SETTINGS_GATED_ITEMS, so it needs organization-settings-r or -u and environment-am_configuration-r.

The API is ENVIRONMENT_AM_CONFIGURATION ([AmConnectionResource](https://github.com/gravitee-io/gravitee-api-management/blob/f7d1d9ea18ffb8e878e07703f8a50faf8be3d55c/gravitee-gamma/gravitee-gamma-module-platform/src/main/java/io/gravitee/gamma/module/platform/rest/resource/am/AmConnectionResource.java)). ENVIRONMENT:ADMIN and GAMMA_IDENTITY_ADMIN get that permission; GAMMA_IDENTITY_ADMIN is defined specifically for AM/agent identity and does not get org settings.

After this PR, an env-only AM admin (including GAMMA_IDENTITY_ADMIN) cannot open the page, and a pasted /access-management URL is rejected. That is encoded as intended in navVisibility.spec.ts and AppRoutes.spec.tsx, but it does not match the permission the backend enforces.

Classic has no equivalent item, so “org section ⇒ org-settings gate” is a Gamma choice, not Classic parity. For FOUND-245 env-level roles this looks like a lockout, not a tightening.

Fix: Keep the item in Organization if you want, but drop it from ORGANIZATION_SETTINGS_GATED_ITEMS. Guard on environment-am_configuration-r only (same as the API). Add a fixture test for GAMMA_IDENTITY_ADMIN / env ADMIN + org USER.

2. organization-settings-u without -r opens Management, CORS, and SMTP

Classic:

Footer Organization requires organization-settings-r only ([gio-side-nav.component.ts](https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts) ~357).
Console Settings route is anyOf: ['organization-settings-r'] only ([organization-settings-routing.module.ts](https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-apim-console-webui/src/organization/configuration/organization-settings-routing.module.ts) ~57). Inner org nav uses r or u, then Settings still requires -r.
Gamma: those three items have empty NAV_ITEM_PERMISSIONS, so the outer gate (r or u) is enough. A custom role with only -u can open and edit them (canEdit is -u).

The AppRoutes test named “without organization-settings-r” does not actually model “has -u, lacks -r”. It does !anyOf.includes('organization-settings-r'), which fails the whole [r, u] check because -r is in the list.

Fix: Put ['organization-settings-r'] on management/cors/smtp (match Classic Settings). Keep r | u only as the outer Organization/Users gate. Add a spec: -u only does not show those three; -r does.

3. Two writers to permissionService environment scope

Host [loadEnvironmentPermissions](https://github.com/gravitee-io/gravitee-api-management/blob/master/gravitee-gamma/gravitee-gamma-control-plane-webui/src/features/permissions/loadEnvironmentPermissions.ts) and the module hook both load('environment', …). The 403 patch in useForbiddenResourceRedirect only updates the module React Query cache; the hook then load()s that patched list.

A later host refetch (login with same env id, env switch) can write the stale backend grant back into permissionService. Metadata/dictionaries still have the extra *Forbidden probe; other items can show again and re-arm the allow → 403 → redirect loop this PR is trying to close.

This is not introduced from scratch here, but this PR’s security story depends on the patched cache winning.

Fix: One owner for environment grants (host or module), or have the 403 patch also permissionService.load / clear so a host refetch cannot restore stripped prefixes without a matching patch.

4. FOUND-245 multi-tenant is only manual

Ticket: org, env, and multi-tenant. Specs cover static role sets, not env switch (USER on DEV, ADMIN on TEST). Query key is env id, and host clears environment on env change, so it should work — there is no automated check that nav/guards follow the new env and do not keep the previous env’s grants.

Worth one hook or AppRoutes test: env id changes → new fetch → load('environment', …) with the new map.

Minor
hasPermission switch in ModuleLayout is a second, lossy policy map. navVisibility is keyed by permission string; the switch collapses anyOf groups (organization-policies-c|d|u all return canAccessPlatformPolicies) and default: false. Pure tests never hit this. A new NAV_ITEM_PERMISSIONS entry that is not in the switch is silently hidden. Prefer permissionService.hasAnyOf (or a Record filled from the same hooks) so there is one mapping.

modulePathFor uses the last path segment that matches a route key. /applications/alerts (application id alerts) would treat alerts as the leaf. Unlikely with UUIDs; brittle for human ids.

Env permissions query error never becomes ready. isSuccess stays false → guards return null → blank Platform, not no-access. Fail-closed, poor UX. Treat error as ready-with-empty if you want the educational page.

bindCacheResetToLogin never unsubscribes. Fine with the module QueryClient singleton; it will accumulate listeners if that client is ever recreated (tests).

PlatformNoAccessPage is a large new surface (~145 lines) for FOUND-245. Fine if product asked for it; copy will drift as role catalogs change.

arpzilla-dev:
Addressed the review on https://github.com/gravitee-io/gravitee-api-management/pull/19439#pullrequestreview-5045001785 (working tree, not pushed):

1. **Access Management** — dropped from `ORGANIZATION_SETTINGS_GATED_ITEMS`. Guard is `environment-am_configuration-r` only. Added `GAMMA_IDENTITY_ADMIN` + org USER fixture; env ADMIN + org USER keeps AM. Pasted `/access-management` works without org settings.

2. **Settings `-u` without `-r`** — Management / CORS / SMTP now require `organization-settings-r` (Classic Settings). Outer Organization/Users gate stays `r | u`. Specs cover `-u` only (hidden) and `-r` (shown).

3. **Two writers** — 403 patch also `permissionService.load`s the stripped list. While the module is mounted, a host refetch that restores stripped `environment-*` grants is overwritten from the React Query cache. Query no longer refetches on remount/focus.

4. **Env switch** — hook spec: env id change → new fetch → `load('environment', new map)`.

Minors: `ModuleLayout` `hasPermission` now uses `permissionService.hasAnyOf`. Fetch error is ready (no blank Platform). Left `modulePathFor` leaf matching and `bindCacheResetToLogin` unsubscribe as-is.

aqibmohammadkhan (commented):
1. Query-key collision silently defeats this PR's core security mechanism. Master added gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/hooks/useEnvironmentPermissions.ts using the same query key environmentPermissionKeys.detail(envId) but with staleTime: 0, refetchOnMount: 'always'. The PR's query on that key is deliberately staleTime: Infinity, refetchOnMount: false — its doc comment says "Remounts must keep a live 403 patch; refetching would restore a stale grant and re-arm the allow → 403 → redirect loop." Master calls the new hook from AppRoutes.tsx and five Shared Policy Group pages/components. After merge, mounting any SPG page forces a refetch that overwrites the 403-stripped patch — re-arming precisely the loop this PR exists to close. useEnvironmentPermissions.ts auto-merges cleanly, so nothing surfaces this.

2. The two branches moved in opposite architectural directions. Master's AppRoutes.tsx never references permissionService (0 occurrences) and deliberately routed SPG gating to the module-local live query, commenting that the SDK's useHasPermission "isn't wired to this module's own permission fetch." This PR routes everything back through permissionService/useHasPermission. Resolving the AppRoutes.tsx conflict by taking the PR side silently reverts master's fix.

3. user-groups → groups rename mismatches on auto-merge. Master renamed the route key (commit 3fc75a3). Post-merge, routes.ts and navigation.ts auto-merge to 'groups', but navVisibility.ts — a PR-only file, so no conflict — keeps NAV_ITEM_PERMISSIONS['user-groups'], and AppRoutes.tsx ends up with <Route path="groups"> wrapped in itemKey="user-groups". Effect: isNavItemVisible('groups') returns false for everyone, so Groups vanishes from the sidebar, and a groups-only user lands on the no-access page. The spec's key-parity test should turn CI red, but the conflict markers won't point at it.

4. The fail-open unguarded default (unmapped item key ⇒ no permission check) — with a route-key rename in flight it's exactly the condition that turns it live. The unexplained narrowing on Tenants/Entrypoints (dropping environment-tenant-r, environment-entrypoint-r, organization-entrypoint-r) remains unaddressed.

the permission-read strategy has to be picked deliberately,

arpzilla-dev:
Thanks @aqibmohammadkhan — you were right that a silent merge with master would have undone this PR. Rebased onto current master and resolved the collisions deliberately rather than taking either side.

**Permission-read strategy:** the React Query cache is the source of truth for environment grants while the module is mounted. A live 403 patches that cache in place and must not be overwritten by a remount refetch. `useHasEnvironmentPermission` now reads the **same** shared query (`staleTime: Infinity`, `refetchOnMount: false`). Master’s second observer on the same key with `refetchOnMount: 'always'` is gone — that is exactly the collision that would restore a stale grant and re-arm the allow → 403 → redirect loop. Shared Policy Groups still gate through `useHasEnvironmentPermission` (master’s live-403 path). Sidebar / org items still go through `navVisibility` + `permissionService`, which this PR keeps projected from that cache.

**`user-groups` → `groups`:** `navVisibility`, routes, and guards all use `groups`. A key-parity spec fails if that drifts again. Unmapped page-guard keys fail closed (`anyOf: []` is deny).

**Tenants / Entrypoints:** Classic org nav is the oracle here — `organization-tenant-r` / `organization-tag-r` after the org-settings outer gate (`organization-navigation.service.ts`). We do not also accept `environment-tenant-r` or `environment-entrypoint-r` for those items.

Also kept master’s Alert Engine lock/redirect, and fixed a fail-closed bootstrap loop on `/platform` (`?? []` as a `useLayoutConfig` dep).

arpzilla-dev:
working after recent changes
image
image

aqibmohammadkhan (approved):

arpzilla-dev:
working after all latest changes
image

PR #19465, opened 2026-08-28
fix: improve subscription filters layout

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/PORTAL-134

## Summary

Fix the subscription filters layout by removing the fixed maximum width from the filters container.

The filters now use the available page width while preserving the existing responsive wrapping behavior.

## Changes

- Remove the `50rem` maximum width from the subscription filters container.
- Keep the existing responsive layout and filter behavior unchanged.

Zrzut ekranu 2026-08-28 o 09 02 41


dev-marek (approved):
lgtm

PR #19466, opened 2026-08-28
Feat/ext 180 api referenced links automation - wrap up

MateuszJozwiak-Incubly:
## Issue

## Description

This branch is a follow-up on feat(automation): api-referenced links automation - openApi and resources: it fixes two small but real correctness gaps that surfaced once the API-link endpoints were exercised end to end, and closes out the test-coverage gaps raised in review on PR [#19285](https://github.com/gravitee-io/gravitee-api-management/pull/19285).

### API-link responses now reflect what was actually persisted

`ApiLinksResource#createOrUpdate` previously always built its response by echoing back the **submitted spec** — so a successful apply would report the raw input, not the sanitized/normalized value that was actually written to the navigation tree (e.g. a trimmed name).

Now, whenever the use case returns a persisted link, the response is built from **that** link instead:

- A successful real apply reports the persisted state — sanitized name, resolved id, everything the store actually holds.
- A dry run (which never persists) and a failed real apply (which persists nothing) both still fall back to echoing the submitted spec, since there's nothing persisted to report.

### Portal documentation & link apply now fails loudly instead of silently skipping

`CreateOrUpdatePortalDocumentationUseCase` and `CreateOrUpdatePortalLinkUseCase` used to **silently skip** navigation-tree materialization whenever the target environment had no default portal — the apply would still return a `200`/success `Output` with no errors, while nothing was actually created or updated. A caller had no way to tell "materialized" from "silently no-opped."

Both use cases now reject the apply outright when there's no portal to attach to:

- **Documentation**: throws `ValidationDomainException("no portal exists in this environment to attach the documentation to")`, same as its other validation failures.
- **Link**: adds `Validator.Error.severe("no portal exists in this environment to attach the link to")` to the errors list, following the same accumulate-and-return pattern it already used for its other validation rules.

This is a **behavioral change** for existing callers on environments without a default portal: an apply that used to silently no-op now fails the request.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/ApiLinkResourceTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/ApiLinkResourceTest.java
@@ -17,31 +17,21 @@ package io.gravitee.apim.rest.api.automation.resource;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.when;
 
-import io.gravitee.apim.core.portal.model.PortalArea;
 import io.gravitee.apim.core.portal_page.exception.PortalLinkNotFoundException;
-import io.gravitee.apim.core.portal_page.model.AutomationMetadata;
-import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
-import io.gravitee.apim.core.portal_page.model.PortalNavigationLink;
-import io.gravitee.apim.core.portal_page.model.PortalVisibility;
 import io.gravitee.apim.core.portal_page.use_case.CreateOrUpdateApiLinkUseCase;
 import io.gravitee.apim.core.portal_page.use_case.DeleteApiLinkUseCase;
 import io.gravitee.apim.core.portal_page.use_case.GetApiLinkUseCase;
 import io.gravitee.apim.core.portal_page.use_case.ValidateApiLinkUseCase;
-import io.gravitee.apim.core.validation.Validator;
 import io.gravitee.apim.rest.api.automation.model.PortalLinkState;
 import io.gravitee.apim.rest.api.automation.resource.base.AbstractResourceTest;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.client.Entity;
 import jakarta.ws.rs.core.MediaType;
 import java.util.List;
-import java.util.Optional;
-import org.assertj.core.api.SoftAssertions;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.DisplayNameGeneration;
 import org.junit.jupiter.api.DisplayNameGenerator;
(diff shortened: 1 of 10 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/use_case/CreateOrUpdatePortalDocumentationUseCase.java
```diff
@@ -119,11 +123,7 @@ public Output execute(Input input) {
         }
 
         var area = sanitized.area() != null ? sanitized.area() : PortalArea.TOP_NAVBAR;
-
-        // Skip nav-tree materialization for non-default portals — app is not ready for that.
-        if (portalAutomationScopeEnforcer.isDefaultPortal(input.auditInfo(), sanitized.portalId())) {
```
konrad-krol-incubly, line 124: I think this gating was dead anyway, because it was relying on initial mechanism we considered of having "default-portal" populated per env -> however this concept was dropped - GKO constructs portal name as: namespace + hrid so it only matched if someone named portal as: "portal" in "default" namespace.

Right now: automation.portal.allowMultiplePortalPerEnv prevents of having multiple portal per env.
This is only set to true for GKO tests -> it materializes anyway.

It shouldn't be part of current PR (its gap anyway), but we need to decide if allowMultiplePortalPerEnv=true
- should skip materialization for every nth portal where N>1 -> how to detect that?
- allow to mix this content as long as APIM is not multiportal ready -> its GKO tests anyway for now

konrad-krol-incubly:
lgtm, just one remark that requires consultation and potential follow-up (as existing gap, not introduced by this PR)

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/ApiLinkResourceTest.java
```diff
@@ -237,21 +268,22 @@ void should_return_404_when_the_link_does_not_exist() {
...
-            .id(PortalNavigationItemId.of("11111111-1111-1111-1111-111111111111"))
+            .id(PortalNavigationItemId.random())
             .organizationId(ORGANIZATION)
             .environmentId(ENVIRONMENT)
-            .title("External Docs")
+            .title(name)
             .segment(LINK_HRID)
             .area(PortalArea.TOP_NAVBAR)
-            .order(3)
+            .order(0)
```
dev-marek: Nit: this fixture hardcodes `.order(0)`, and `Get > should_return_the_link` asserts `state.getOrder()).isEqualTo(0)`. Since order defaults to 0, that assertion can't catch a mapper regression that drops the order field entirely (it'd still read 0). The pre-PR fixture used `.order(3)` specifically so this assertion had teeth — consider giving `aPersistedLink` a distinct, non-default order.

dev-marek (commented):
## Review

**Verdict:** Approve — with two non-blocking improvements to consider.

Traced the control flow in both `CreateOrUpdatePortalDocumentationUseCase` and `CreateOrUpdatePortalLinkUseCase`: the new portal-existence check runs before any persistence in both cases, the accumulate-and-return error pattern is preserved, and the "at most one severe error surfaces" invariant still holds. Test coverage for the new behavior (reject-and-persist-nothing, response built from persisted link) is solid and matches the PR description.

Two things worth a look, neither blocking:

1. **`PortalLinksResource#createOrUpdate`** (not touched by this PR, so no inline anchor) still unconditionally echoes the *submitted* spec via `PortalLinkMapper.toPortalLinkState(spec, ...)`, never `output.link()` — even though this PR's own change to `CreateOrUpdatePortalLinkUseCase` now returns the persisted `link`, and `PortalLinkMapper` already has the matching `toPortalLinkState(PortalNavigationLink, ...)` overload sitting unused. This is the exact same "reports raw input instead of what was persisted" bug this PR fixes for `ApiLinksResource`, left in place on the parallel portal-link surface. Worth confirming whether that's a deliberate follow-up or a miss.
2. See inline comment on the test fixture.

Residual risk: none beyond the above — the change is small and self-contained.

dev-marek (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/resource/ApiLinkResourceTest.java
MateuszJozwiak-Incubly: Fixed the test fixture, and also the PortalLinksResource.createOrUpdate as it was a miss, not deliberately left like that.

PR #19482, opened 2026-08-28
fix(rest): map ConflictDomainException to 409 and keep technical code s on validation errors

araffass:
## Issue

https://gravitee.atlassian.net/browse/AIAM-375

## Description

GammaModuleApplication registers a mapper for NotFoundDomainException, ValidationDomainException, NotAllowedDomainException and TechnicalDomainException, but none for ConflictDomainException — and no such mapper existed in the v1 provider package at all. Every conflict therefore fell through to ThrowableMapper and surfaced as a 500.

That is already visible today, independently of the AI Management work: LlmProxyAlreadyExistsException (an LLM proxy created or duplicated with an entity id already taken) answers 500 instead of 409. The console hides it by checking the entity id beforehand, so it only shows up on a race or a direct API call.

Two changes:
- a ConflictDomainExceptionMapper in the v1 provider package (409, id carried as a parameter, mirroring NotFoundDomainExceptionMapper), registered in GammaModuleApplication and GraviteeManagementApplication — the classic app had the same gap;
- ValidationDomainExceptionMapper now passes the exception's technicalCode and parameters through convert(...). They were silently dropped, so a client could not tell one 400 from another; the response gains the fields only when they are set (empty parameters stay absent).

Three unit tests cover the new mapper (with and without an id) and the technical code propagation. No behaviour change for any 400 that carries neither.

## Additional context

Needed by AIAM-375 (gravitee-gamma-module-aim), where a taken provider slug and a provider still consumed by a workspace both answer 409, and the failure taxonomy of the connection test rides on technical codes (provider.credential.rejected vs provider.service.unreachable). Verified on a local stack: 500 before, 409 with the id parameter after; technical codes now reach the client.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/provider/DomainExceptionMappersTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/provider/DomainExceptionMappersTest.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.management.rest.provider;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.gravitee.apim.core.exception.ConflictDomainException;
+import io.gravitee.apim.core.exception.ValidationDomainException;
+import io.gravitee.rest.api.management.rest.model.ErrorEntity;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public class DomainExceptionMappersTest {
+
+    @Test
+    public void conflict_domain_exception_maps_to_409_with_its_id() {
+        var response = new ConflictDomainExceptionMapper().toResponse(new ConflictDomainException("Slug already used", "my-slug"));
+
+        assertThat(response.getStatus()).isEqualTo(409);
+        var body = (ErrorEntity) response.getEntity();
+        assertThat(body.getMessage()).isEqualTo("Slug already used");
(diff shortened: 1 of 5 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-management/gravitee-apim-rest-api-management-rest/src/test/java/io/gravitee/rest/api/management/rest/provider/DomainExceptionMappersTest.java
```diff
@@ -0,0 +1,57 @@
...
+    @Test
+    public void validation_domain_exception_keeps_its_technical_code() {
+        var response = new ValidationDomainExceptionMapper().toResponse(
+            new ValidationDomainException("Credential rejected", "provider.credential.rejected")
+        );
+
+        assertThat(response.getStatus()).isEqualTo(400);
+        var body = (ErrorEntity) response.getEntity();
+        assertThat(body.getTechnicalCode()).isEqualTo("provider.credential.rejected");
+        assertThat(body.getParameters()).isNull();
```
paulatulis, line 55: The mapper change adds two pass-throughs — `technicalCode` and `parameters` — but only the first is covered here.

This test builds `new ValidationDomainException("Credential rejected", "provider.credential.rejected")`, and that constructor leaves `parameters` as the field-initialised empty `HashMap`. So the assertion below is `isNull()`, and the `exception.getParameters().isEmpty() ? null : exception.getParameters()` branch that actually forwards a populated map is never executed by the suite.

The consequence is that a later refactor of that ternary — inverting the guard, or dropping the argument back to `null` — leaves these tests green while every 400 silently loses its parameters again, which is the regression this PR exists to fix.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/GammaModuleApplication.java
```diff
@@ -82,6 +83,7 @@ public GammaModuleApplication() {
         register(BadRequestExceptionMapper.class);
         register(EnumParamConverterProvider.class);
         register(ValidationDomainExceptionMapper.class);
+        register(ConflictDomainExceptionMapper.class);
```
paulatulis, line 86: This registration falsifies the javadoc on `DashboardVersionConflictExceptionMapper`, which currently reads:

> The platform registers no `ConflictDomainException` mapper at all in `GammaModuleApplication` — management-rest ships none — so without this a refused save would surface through `ThrowableMapper` as a 500.

That is one of the two stated reasons the dashboard-specific mapper exists, and it is no longer true.

The behaviour is fine — JAX-RS resolves to the nearest superclass mapper, `DashboardVersionConflictExceptionMapper` is registered four lines below for the more specific type, and `ObservabilityDashboardsResourceTest` still gets 412 through real Jersey. Only the justification is stale. The risk is that someone re-reading that class later is told it guards against a 500 that can no longer happen, removes it, and quietly turns 412 into 409 for conditional dashboard saves.

paulatulis (approved):
Two minor findings. No blockers.

PR #19483, opened 2026-08-28
test(rest-api): cover the login-to-DCR claim join and id_token precedence

paulatulis:
## Issue

https://gravitee.atlassian.net/browse/APIM-13957
## Description

Tests only — no production code is changed.

Two gaps in the APIM-13949 claims-to-DCR coverage, both found while QA'ing the epic against a real OIDC provider.

**The login-to-DCR join was untested.** Each half had tests, but both stubbed the seam between them: the injection tests hand `ClientRegistrationService#register` a literal claim map, so nothing asserted that a real login actually produces the map the DCR path later reads. That seam is `UserService#findIdpClaims`, which
`ApplicationServiceImpl` calls to feed the registration. A change to what login persists, or to how the claims are read back, would leave both halves green and break the feature.

`PersistedClaimsToDcrInjectionTest` drives a federated login with the shared OAuth2 fixtures, reads the claims back through `findIdpClaims` with nothing stubbed in between, and asserts on the request body WireMock actually received. It also pins the property that matters for the whitelist: a claim login refused to keep stays uninjectable even when the registration provider maps it.

The user repository there is a small in-memory store that copies on both read and write. Production mutates the user instance before writing it, so a store handing back the same instance would stay green even with the repository write deleted — the mistake the APIM-14840 follow-up had to correct in the purge tests. Copying forces the claims through a real round trip.

**`id_token` precedence over `access_token` was untested.** The lookup order is id_token, then access_token, then userinfo. The existing tests cover id_token-over-userinfo and the access_token source on its own, but nothing asserted the first beats the second. The shared JWT fixtures carry no claim whose value *differs* between the two tokens — `sub`, `gis`, `aud` and `iss` are identical in both — and only a differing value can tell them apart, so the test builds two unsigned JWTs inline rather than editing fixtures other tests assert on.

## Additional context

Both tests were verified by mutating the production code and confirming they fail, rather than by observing them pass:

| Mutation | Result |
|---|---|
| Remove `userRepository.update(user)` in `persistWhitelistedClaims` | join fails — claims never reach the store |
| `findIdpClaims` drops the claims it reads | join fails on the claim set |
| Whitelist widened to admit a disallowed claim | join fails on the claim set |
| A disallowed claim substituted for an allowed one, claim set size unchanged | join fails on the DCR request body |
| `extractWhitelistedClaims` source order swapped | precedence test fails |

All mutations were reverted; `git status` confirms only the two test files changed. `prettier:write` has been run on the module.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/PersistedClaimsToDcrInjectionTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/PersistedClaimsToDcrInjectionTest.java
@@ -1,0 +1,295 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.service.impl;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.equalTo;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.matchingJsonPath;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.postRequestedFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.eq;
+import static org.mockito.Mockito.when;
+import static org.mockito.internal.util.collections.Sets.newSet;
+
+import com.github.tomakehurst.wiremock.WireMockServer;
+import com.github.tomakehurst.wiremock.core.Options;
+import io.gravitee.repository.management.api.ClientRegistrationProviderRepository;
+import io.gravitee.repository.management.api.UserRepository;
+import io.gravitee.repository.management.model.ClientRegistrationProvider;
+import io.gravitee.repository.management.model.User;
+import io.gravitee.rest.api.model.EnvironmentEntity;
+import io.gravitee.rest.api.model.NewApplicationEntity;
+import io.gravitee.rest.api.model.OrganizationEntity;
+import io.gravitee.rest.api.model.application.ApplicationSettings;
+import io.gravitee.rest.api.model.application.OAuthClientSettings;
+import io.gravitee.rest.api.model.configuration.identity.SocialIdentityProviderEntity;
+import io.gravitee.rest.api.service.AuditService;
+import io.gravitee.rest.api.service.EnvironmentService;
+import io.gravitee.rest.api.service.MembershipService;
+import io.gravitee.rest.api.service.OrganizationService;
+import io.gravitee.rest.api.service.RoleService;
+import io.gravitee.rest.api.service.common.ExecutionContext;
+import io.gravitee.rest.api.service.common.GraviteeContext;
+import io.gravitee.rest.api.service.converter.UserConverter;
+import io.gravitee.rest.api.service.impl.configuration.application.registration.ClientRegistrationServiceImpl;
+import io.gravitee.rest.api.service.impl.configuration.application.registration.client.register.ClientRegistrationResponse;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.Charset;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import org.apache.commons.io.IOUtils;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.Spy;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+
+/**
+ * Joins the two halves of the claims-to-DCR feature: a federated login captures whitelisted IdP claims onto the user,
+ * and a later dynamic client registration injects those claims into the request body.
+ *
+ * <p>Each half already has its own tests, but both stub the seam between them — the injection tests hand
+ * {@code ClientRegistrationService#register} a literal claim map, so nothing asserts that a real login actually
+ * produces the map the DCR path later reads. That seam is {@code UserService#findIdpClaims}, which
+ * {@code ApplicationServiceImpl} calls to feed the registration. A change to what login persists, or to how the claims
+ * are read back, would leave both halves green and break the feature.
+ *
+ * <p>The user repository here is a small in-memory store rather than a plain stub, and it copies on both read and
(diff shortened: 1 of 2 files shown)


carlos-andres-osorio:
> **gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/configuration/application/registration/ClientRegistrationServiceImpl.java:392**

The APIM-13957 story's seeded acceptance criterion "DCR request is created; injected claims override existing values in request body" looks unreachable with the current DCR request model. `resolveClaimInjections()` only ever produces injections for extension (non-standard) field paths — `isStandardFieldPath(fieldPath)` at this line filters out every field that could already carry a value on `ClientRegistrationRequest`. That class is `@JsonIgnoreProperties(ignoreUnknown = true)` with no `@JsonAnySetter`, so no extension key can ever be pre-populated into it or its serialized body. `injectClaims()` does implement an "override an existing value" code path, but no reachable input can ever place a pre-existing value at an extension field for it to override.

Could you confirm whether the AC needs updating (dropping the "override" scenario as untestable), or whether `ClientRegistrationRequest` should gain an extension/additional-properties map so a reachable override scenario exists?

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
```diff
@@ -2118,6 +2120,41 @@ public void shouldPersistClaimFromAccessTokenSource() throws IOException, Techni
         assertEquals("foobar", user.getIdpClaims().get("custom_access_token"));
     }
 
+    @Test
+    public void should_prefer_the_id_token_value_when_a_claim_is_in_both_tokens() throws IOException, TechnicalException {
```
carlos-andres-osorio, line 2124: `should_prefer_the_id_token_value_when_a_claim_is_in_both_tokens` proves id_token beats access_token, but the precedence ladder's other rung — when no id_token is supplied, the access_token value must win over a differing userinfo value — has no test anywhere in the suite. `shouldPersistClaimFromAccessTokenSource` whitelists `custom_access_token`, a key absent from both the id_token and userinfo fixtures, so it can't distinguish access_token-over-userinfo precedence: it would pass identically if userinfo were checked before access_token. An implementation whose lookup order were id_token → userinfo → access_token would pass every test in the file, including this new one.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/PersistedClaimsToDcrInjectionTest.java
```diff
@@ -0,0 +1,278 @@
...
+
+        NewApplicationEntity application = givenAProviderMapping(Map.of("service_id", "metadata.organization"));
+
+        login();
+
+        Map<String, String> claimsForRegistration = userService.findIdpClaims(EXECUTION_CONTEXT, USER_ID);
+
+        clientRegistrationService.register(GraviteeContext.getExecutionContext(), application, claimsForRegistration);
+
+        wireMockServer.verify(
```
carlos-andres-osorio, line 242: `should_inject_nothing_when_login_captured_no_claims` has only one assertion: a purely negative WireMock check that passes whenever zero requests carry the `metadata.organization` JSON path. That's true both when the SUT correctly omits the unwhitelisted claim from a real registration POST, and when `register()` makes no POST at all. The test never confirms a registration request actually happened, nor does it assert anything about `claimsForRegistration` or the return value of `register(...)`. Contrast with the sibling test `should_inject_the_claims_a_real_login_captured_into_the_dcr_request`, which pairs its positive assertion with the negative check, so it actually proves a request landed.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/PersistedClaimsToDcrInjectionTest.java
```diff
@@ -0,0 +1,278 @@
...
+            postRequestedFor(urlEqualTo("/registrationEp"))
+                .withRequestBody(matchingJsonPath("$.metadata.organization", equalTo("585252525")))
+                .withRequestBody(matchingJsonPath("$.metadata.job", equalTo("API_FRIENDLY_USER")))
+        );
+        // The whitelist filter survives the join: mapping a claim the login never stored injects nothing.
+        wireMockServer.verify(0, postRequestedFor(urlEqualTo("/registrationEp")).withRequestBody(matchingJsonPath("$.metadata.email")));
+    }
+
+    @Test
+    public void should_inject_nothing_when_login_captured_no_claims() throws Exception {
```
carlos-andres-osorio: `should_inject_nothing_when_login_captured_no_claims` isn't load-bearing: it passes with the `login()` call deleted. The user seeded in `setup()` never has `idpClaims` set, so it starts `null`. `persistWhitelistedClaims`'s write-skip guard only calls `userRepository.update` when the newly computed claims differ from what's already stored — with an empty whitelist the newly computed claims are also `null`, so `login()` never writes to the user record here at all. `findIdpClaims` then reads back the same pre-existing `null` idpClaims whether or not `login()` ran, and the downstream injection resolver treats a `null` claims map as producing zero injections, so the `verify(0, ...)` at line 225 holds unconditionally — nothing in the test can fail if the empty-whitelist purge on the login path regresses.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/UserServiceTest.java
paulatulis: You are right, and the reordering you describe is exactly what I checked. Added `should_prefer_the_access_token_value_over_userinfo_when_there_is_no_id_token`, which passes no id_token and gives userinfo and the access_token the same claim with different values.

Verified by mutating the source order to `id_token -> userinfo -> access_token`: the new test fails with `expected: <from-access-token> but was: <from-userinfo>`, and every other test in the file still passes — which is the situation you predicted.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/PersistedClaimsToDcrInjectionTest.java
paulatulis: Addressed. The test now asserts `register(...)` returned non-null and adds `wireMockServer.verify(postRequestedFor(urlEqualTo("/registrationEp")))`, so the negative check can no longer be satisfied by no request having been made at all.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/PersistedClaimsToDcrInjectionTest.java
paulatulis: Confirmed — I reproduced it by deleting the `login()` call and the test still passed. Your analysis of why is exactly right: the seeded user had `idpClaims` null, an empty whitelist computes null, the write guard sees null on both sides and skips the write, so `findIdpClaims` read back the same null either way.

Worth being straight about how this got here: I mutation-tested the other test in this file four ways and then described the file as mutation-verified. I never ran one against this test.

Fixed by seeding the user with a claim from an earlier login, so an empty whitelist has something to purge. Renamed to `should_inject_nothing_after_login_purges_the_claims_it_had_stored`. It now fails two ways — with `login()` removed, and with the APIM-14840 purge regressed (early `return` restored on the null/empty whitelist branch) — both giving `expected: <null> but was: <{service_id=585252525}>`.

paulatulis:
On the unreachable "override" AC — I reached the same conclusion independently while QA'ing the epic, down to the same mechanism, so consider it confirmed from two directions. I added a comment to the ticket. https://gravitee.atlassian.net/browse/APIM-13957?focusedCommentId=103600

To add one detail to your analysis: the disjointness holds on both entry points, not just registration. `ClientRegistrationServiceImpl.convert(NewApplicationEntity)` (create) and `convert(ClientRegistrationRequest, UpdateApplicationEntity)` (update) both funnel `additionalClientMetadata` through `mapper.convertValue(metadata, ClientRegistrationRequest.class)`. Since that class is `@JsonIgnoreProperties(ignoreUnknown = true)` with no `@JsonAnySetter`, extension keys are dropped at conversion — so additional client metadata can only ever set standard fields, while injection can only ever write extension fields. The two mechanisms cannot target the same field by construction.

On your question of which way to resolve it: that is a product call rather than one for this PR, and it is already raised on APIM-13957 along with the story's title, which says "integration tests" that this work deliberately does not deliver. My recommendation there is to drop the AC rather than widen `ClientRegistrationRequest` — adding an extension map to make the scenario reachable would create a second way to place arbitrary keys into a DCR body purely so an AC can be tested, and the standard-field blocklist exists to keep that surface narrow.

Not blocking this PR either way, since nothing here depends on the outcome. Happy to follow whichever the ticket lands on.

carlos-andres-osorio (approved):

PR #19497, opened 2026-08-31
fix(console): keep sidebar navigation working when an environment has no hrids

SamHardi:
## Issue

https://gravitee.atlassian.net/browse/APIM-14748

## Description

Environments that were carried through upgrades from before `hrids` existed have no `hrids` value in the `environments` collection. `GioSideNavComponent` interpolated the raw array into the group base paths:

```ts
routerBasePath: `/${this.currentEnv.hrids}/observability`
```

With `hrids` undefined that yields `/undefined/observability`; with an empty array, `//observability`. `GioMenuItemsComponent#onHeaderClick` then calls `router.navigate([routerBasePath])`, `EnvironmentGuard` cannot resolve the environment segment, falls back to the first environment, and `{ path: '', pathMatch: 'full', redirectTo: 'home' }` lands the user on the Dashboard. Direct URL access worked because the URL carries a resolvable environment segment.

This is not a data/migration gap: "`hrids` is optional, fall back to the id" is already the contract everywhere else — `EnvironmentGuard` and `EnvironmentServiceImpl.findByOrgAndIdOrHrid` on the backend both accept either. The side nav was the only place that violated it. Fixing it in the console repairs every affected installation without a database migration.

Rather than re-deriving a name for the environment, the fix uses the one the URL already carries: the `envHrid` route param, which the component received and dropped. That keeps the group header, its children (relative links) and the address bar in agreement by construction — an environment may hold several hrids, and the guard accepts the id as well, so anything re-derived from `currentEnv` can name the environment differently from the URL the user is on. Caught in review: with `hrids: ['prod', 'production']`, landing on `/production/home` and clicking Observability used to navigate to `/prod/observability/overview`, switching the alias under the user.

## Changes

- `GioSideNavComponent` captures the `envHrid` route param and builds the three group base paths (Kafka, Observability, Analytics) and the menu-search router links from it.
- `gio-side-nav.component.spec.ts`: regression tests on the base paths for `hrids: undefined`, `[]` and `['my-env']`.
- `gio-side-nav.navigation.spec.ts` (new): drives the whole navigation chain — real `Router`, real `EnvironmentGuard`, a real click on the rendered menu group — and asserts the environment segment and group the user ends up on.
- `GioSideNavHarness` gained `clickGroup(title)`, used by the navigation spec.

### Manual verification on a running stack

**Environment.** MongoDB + Management API from `docker/quick-setup/mongodb` (Management API on `:8083`), console served from this branch on `:4000` proxying to it, signed in as `admin`. The legacy state was reproduced on the real database:

```
db.environments.updateOne({_id:"DEFAULT"}, {$unset:{hrids:""}})
```

After restarting the Management API, `GET /management/organizations/DEFAULT/environments` returned the environment with **no `hrids` field at all**:

```json
[ { "id": "DEFAULT", "name": "Default environment", "description": "Default environment", "organizationId": "DEFAULT" } ]
```

**Scenario.** Start on the Dashboard (`#!/default/home/overview`), click a sidebar group header, read the resulting URL. Each group was tested against the unmodified `master` component and then against this branch, on the same running stack.

| Group clicked | Expected | Observed before fix | Observed after fix |
|---|---|---|---|
| Observability | `#!/…/observability/overview` | `#!/DEFAULT/home/overview` ❌ | `#!/DEFAULT/observability/overview` ✅ |
| Analytics | `#!/…/analytics/dashboard` | `#!/DEFAULT/home/overview` ❌ | `#!/DEFAULT/analytics/dashboard?from=…&to=…` ✅ |
| Kafka | `#!/…/clusters/kafka-standalone` | `#!/DEFAULT/home/overview` ❌ | `#!/DEFAULT/clusters/kafka-standalone` ✅ |

All three landed on `#!/DEFAULT/home/overview` before the fix — byte for byte the URL the ticket reports. No browser console error accompanied the redirect, matching the report.

**No regression with `hrids` present.** After restoring `hrids: ["default"]` and restarting the Management API, from the Dashboard, clicking Observability navigates to `#!/default/observability/overview` — still the hrid, not the id.

### Automated

`gio-side-nav.navigation.spec.ts` reproduces the same scenario in-process (real `Router`, real `EnvironmentGuard`, real click on the rendered group, HTTP boundary stubbed with the payload above), so the regression stays covered in CI:

| Scenario | Before fix | After fix |
|---|---|---|
| No `hrids` → Observability / Analytics / Kafka | `/DEFAULT/home` ❌ | the group's own path ✅ |
| `hrids: ['default']` → Observability | `/default/observability` ✅ | ✅ |

The navigation tests assert the environment segment and the group (`/DEFAULT/observability`), not the leaf each group's module redirects to — that redirect belongs to those modules, and asserting it here against a local copy of the route tree would go stale silently.

`gio-side-nav.component.spec.ts` covers the base paths the side nav builds:

| URL segment | `currentEnv.hrids` | Expected base path | Before fix | After fix |
|---|---|---|---|---|
| `DEFAULT` | `undefined` | `/DEFAULT/observability` | `/undefined/observability` ❌ | ✅ |
| `DEFAULT` | `[]` | `/DEFAULT/observability` | `//observability` ❌ | ✅ |
| `my-env` | `['my-env']` | `/my-env/observability` | ✅ | ✅ |

**Suite result:** `nx test console` on `gio-side-nav` — **29/29 passing**. `tsc --noEmit` and `eslint` clean on the changed files.

## Additional context

- Same symptom as APIM-14285 (fixed in 4.11.10), different cause: that one was a build-order race, this one is the missing-hrids fallback.
- Backport labels `apply-on-4-12-x` and `apply-on-4-11-x` are set. `4.12.x` carries the same four `currentEnv.hrids` sites as `master`, so it should apply as is. `4.11.x` has three: the Kafka entry exists there as a **flat** item (`displayName: 'Kafka Clusters'`, relative `routerLink: './clusters'`, no `routerBasePath`), which is why it escapes the bug — it is not absent.
- Noticed in passing and deliberately left out of scope: `EnvironmentServiceImpl.findByOrgAndIdOrHrid` calls `env.getHrids().contains(...)` with no null check, on the same population of legacy environments. Worth its own ticket.

## Reviewer notes

- `gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts` is the whole fix; everything else is tests.
- **Compatibility-sensitive:** this changes navigation behaviour and touches route construction.
- The manual run above was done on `master` + this branch, and is worth a quick repeat on `4.11.x` once the backport PR opens.
- **The `4.11.x` backport will not apply cleanly.** Everything the new specs lean on is present there (`ApimFeature.APIM_CLUSTER` / `APIM_API_PRODUCTS` / `ALERT_ENGINE` / `APIM_AUDIT_TRAIL`, `getFeatureInfoData`, `EnvironmentGuard.initEnvConfigAndLoadPermissions`, `observability.module.ts`, `cluster-routing.module.ts`) **except** the Kafka strand and the harness:
  - the Kafka case in `gio-side-nav.navigation.spec.ts` and the `clusters` branch of its route tree have no counterpart — the flat item has no group header to click;
  - `basePathOf('Kafka')` in `gio-side-nav.component.spec.ts` likewise;
  - `gio-side-nav.component.harness.ts` **does not exist on `4.11.x`** — the whole file has to come along, since the navigation spec now goes through `clickGroup()`.
  
  So four files need a hand, not a clean cherry-pick.

### DIFF
--- a/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
+++ b/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
@@ -37,6 +37,7 @@ import { GioPermissionService } from '../../shared/components/gio-permission/gio
 import { CONSTANTS_TESTING, GioTestingModule } from '../../shared/testing';
 import { Constants } from '../../entities/Constants';
 import { EnvironmentSettingsService } from '../../services-ngx/environment-settings.service';
+import { Environment } from '../../entities/environment/environment';
 
 const PORTAL_SETTINGS_PERMISSIONS_SET = new Set(PORTAL_SETTINGS_PERMISSIONS);
 
@@ -114,7 +115,7 @@ describe('GioSideNavComponent', () => {
           },
         },
         { provide: 'LicenseConfiguration', useValue: LICENSE_CONFIGURATION_TESTING },
-        { provide: ActivatedRoute, useValue: { params: of({ envId: 'DEFAULT' }) } },
+        { provide: ActivatedRoute, useValue: { params: of({ envHrid: 'DEFAULT' }) } },
         { provide: GioMenuSearchService, useValue: menuSearchService },
       ],
     })
@@ -355,6 +356,81 @@ describe('GioSideNavComponent', () => {
       expect(kafka?.routerBasePath).toBe(`/${envHrid}/clusters`);
 
       ctrl.verify();
+    });
+  });
+
+  describe('environment segment in menu paths (APIM-14748)', () => {
+    const buildMenuForEnvironment = async (environment: Environment, envHrid: string) => {
+      await TestBed.configureTestingModule({
+        declarations: [GioSideNavComponent],
+        imports: [NoopAnimationsModule, GioTestingModule, GioSideNavModule, MatIconTestingModule],
+        providers: [
+          { provide: GioPermissionService, useValue: { hasAnyMatching: () => true } },
+          {
+            provide: Constants,
+            useFactory: () => ({
+              ...CONSTANTS_TESTING,
+              org: {
+                ...CONSTANTS_TESTING.org,
+                settings: { ...CONSTANTS_TESTING.org.settings, licenseExpirationNotification: { enabled: true } },
+                currentEnv: environment,
+                environments: [environment],
+              },
+            }),
+          },
+          {
+            provide: EnvironmentSettingsService,
+            useValue: { get: () => of({ apiScore: { enabled: true }, portalNext: { access: { enabled: true } } }) },
+          },
+          { provide: 'LicenseConfiguration', useValue: LICENSE_CONFIGURATION_TESTING },
+          { provide: ActivatedRoute, useValue: { params: of({ envHrid }) } },
+          { provide: GioMenuSearchService, useValue: menuSearchService },
+        ],
+      })
+        .overrideProvider(InteractivityChecker, {
+          useValue: { isFocusable: () => true, isTabbable: () => true },
+        })
+        .compileComponents();
+
+      const fix = TestBed.createComponent(GioSideNavComponent);
+      httpTestingController = TestBed.inject(HttpTestingController);
+
+      fix.detectChanges();
+      httpTestingController
+        .expectOne(`${LICENSE_CONFIGURATION_TESTING.resourceURL}`)
+        .flush({ tier: '', features: [], packs: [], expiresAt: new Date() });
+      fix.detectChanges();
+
+      return fix;
+    };
+
+    // The guard resolves the segment to either an hrid or the environment id, so the menu must mirror
+    // whichever one the URL carries rather than re-deriving one from the environment.
+    it.each([
+      ['no hrids', undefined, 'DEFAULT'],
+      ['empty hrids', [], 'DEFAULT'],
+      ['a single hrid', ['my-env'], 'my-env'],
+    ])('should build group base paths from the URL segment when the environment has %s', async (_label, hrids, envHrid) => {
+      const fix = await buildMenuForEnvironment({ id: 'DEFAULT', name: 'default', hrids, organizationId: 'org' }, envHrid);
+
+      const basePathOf = (displayName: string) =>
+        fix.componentInstance.mainMenuItems.find(i => i.displayName === displayName)?.routerBasePath;
+
+      expect(basePathOf('Observability')).toBe(`/${envHrid}/observability`);
+      expect(basePathOf('Analytics')).toBe(`/${envHrid}/analytics`);
+      expect(basePathOf('Kafka')).toBe(`/${envHrid}/clusters`);
+    });
+
+    it('should build menu search router links from the URL segment', async () => {
+      const addMenuSearchItems = jest.spyOn(menuSearchService, 'addMenuSearchItems');
+      addMenuSearchItems.mockClear();
+
+      await buildMenuForEnvironment({ id: 'DEFAULT', name: 'default', hrids: undefined, organizationId: 'org' }, 'DEFAULT');
+
+      const searchItems: { name: string; routerLink: string }[] = addMenuSearchItems.mock.calls[0][0];
+      expect(searchItems.find(i => i.name === 'Dashboard')?.routerLink).toBe('/DEFAULT/home');
+
+      addMenuSearchItems.mockRestore();
     });
   });
 
--- a/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
+++ b/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
@@ -73,6 +73,7 @@ export class GioSideNavComponent implements OnInit, OnDestroy {
   public environments: SelectorItem[] = [];
 
   public currentEnv: Environment;
+  private envHrid: string;
   public licenseExpirationDate$: Observable<Date>;
   public licenseExpirationNotificationEnabled = true;
 
@@ -91,7 +92,12 @@ export class GioSideNavComponent implements OnInit, OnDestroy {
       .pipe(
         map(p => p.envHrid),
         distinctUntilChanged(),
-        switchMap(_ => {
+        switchMap(envHrid => {
+          // The segment the URL actually carries. Building absolute paths from it keeps them in
+          // agreement with router.url by construction: hrids are optional, an environment can hold
+          // several, and the guard accepts the id as well -- so anything re-derived from currentEnv
+          // can name the environment differently from the address bar the user is on.
+          this.envHrid = envHrid;
           this.environments = this.constants.org.environments.map(env => ({ value: env.id, displayValue: env.name }));
           this.currentEnv = this.constants.org.currentEnv;
 
@@ -201,7 +207,7 @@ export class GioSideNavComponent implements OnInit, OnDestroy {
       permissions: ['environment-cluster-r'],
       licenseOptions: clusterLicenseOptions,
       iconRight$: clusterIconRight$,
-      routerBasePath: `/${this.currentEnv.hrids}/clusters`,
+      routerBasePath: `/${this.envHrid}/clusters`,
       items: [
         {
           displayName: 'Standalone',
@@ -236,7 +242,7 @@ export class GioSideNavComponent implements OnInit, OnDestroy {
       displayName: 'Observability',
       category: 'Observability',
       permissions: ['environment-platform-r'],
-      routerBasePath: `/${this.currentEnv.hrids}/observability`,
+      routerBasePath: `/${this.envHrid}/observability`,
       items: [
         {
           displayName: 'Overview',
@@ -261,7 +267,7 @@ export class GioSideNavComponent implements OnInit, OnDestroy {
       displayName: 'Analytics',
       category: 'Analytics',
       permissions: ['environment-platform-r'],
-      routerBasePath: `/${this.currentEnv.hrids}/analytics`,
+      routerBasePath: `/${this.envHrid}/analytics`,
       iconRight$: of('gio:info'),
       iconRightTooltip: 'This interface supports API V2 only. For API V4, switch to the new Observability interface.',
       items: [
@@ -382,7 +388,7 @@ export class GioSideNavComponent implements OnInit, OnDestroy {
       .map(item => {
         return {
           name: item.displayName,
-          routerLink: `/${this.currentEnv.hrids}/${cleanRouterLink(item.routerLink)}`,
+          routerLink: `/${this.envHrid}/${cleanRouterLink(item.routerLink)}`,
           category: item.category,
           groupIds: [SIDE_NAV_GROUP_ID],
         };
--- a/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
+++ b/gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
@@ -1,0 +1,253 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Component } from '@angular/core';
+import { ComponentFixture, TestBed, fakeAsync, tick } from '@angular/core/testing';
+import { HttpTestingController, provideHttpClientTesting } from '@angular/common/http/testing';
+import { provideHttpClient } from '@angular/common/http';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import { MatIconTestingModule } from '@angular/material/icon/testing';
+import { InteractivityChecker } from '@angular/cdk/a11y';
+import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
+import { Router, RouterModule, Routes } from '@angular/router';
+import { GioMenuSearchService, LICENSE_CONFIGURATION_TESTING } from '@gravitee/ui-particles-angular';
+import { of } from 'rxjs';
+
+import { GioSideNavModule } from './gio-side-nav.module';
+import { GioSideNavHarness } from './gio-side-nav.component.harness';
+
+import { Constants } from '../../entities/Constants';
+import { Environment } from '../../entities/environment/environment';
+import { EnvironmentGuard } from '../../management/environment.guard';
+import { SettingsNavigationService } from '../../management/settings/settings-navigation/settings-navigation.service';
+import { EnvironmentSettingsService } from '../../services-ngx/environment-settings.service';
+import { GioPermissionService } from '../../shared/components/gio-permission/gio-permission.service';
+import { CONSTANTS_TESTING } from '../../shared/testing';
+import { ApimFeature, getFeatureInfoData } from '../../shared/components/gio-license/gio-license-data';
+
+@Component({ template: '', standalone: false })
+class PageComponent {}
+
+// Stands in for ManagementComponent: instantiated by the router only once EnvironmentGuard has resolved.
+@Component({ template: '<gio-side-nav></gio-side-nav><router-outlet></router-outlet>', standalone: false })
+class ManagementShellComponent {}
+
+// Stands in for the application root: only the root outlet, so nothing renders before the first navigation.
+@Component({ template: '<router-outlet></router-outlet>', standalone: false })
+class RootComponent {}
+
+// Mirrors the license configuration built at bootstrap in index.ts, so the gioLicense directive can
+// resolve the APIM features the side nav declares.
+const LICENSE_CONFIGURATION = { ...LICENSE_CONFIGURATION_TESTING, featureInfoData: getFeatureInfoData(undefined) };
+
+// Mirrors the real route tree: management-routing.module.ts plus the default redirects declared by
+// observability.module.ts, env-analytics-legacy.module.ts and cluster-routing.module.ts.
+const routes: Routes = [
+  {
+    path: ':envHrid',
+    component: ManagementShellComponent,
+    canActivate: [EnvironmentGuard.initEnvConfigAndLoadPermissions],
+    children: [
+      { path: 'home', component: PageComponent },
+      {
+        path: 'observability',
+        children: [
+          { path: 'overview', component: PageComponent },
+          { path: '', pathMatch: 'full', redirectTo: 'overview' },
+        ],
+      },
+      {
+        path: 'analytics',
+        children: [
+          { path: 'dashboard', component: PageComponent },
+          { path: '', pathMatch: 'full', redirectTo: 'dashboard' },
+        ],
+      },
+      {
+        path: 'clusters',
+        children: [
+          { path: 'kafka-standalone', component: PageComponent },
+          { path: '', pathMatch: 'full', redirectTo: 'kafka-standalone' },
+        ],
+      },
+      { path: '', pathMatch: 'full', redirectTo: 'home' },
+    ],
+  },
+];
+
+/**
+ * Drives the real navigation chain a user triggers by clicking a sidebar group:
+ * GioSideNavComponent builds routerBasePath -> GioMenuItemsComponent#onHeaderClick calls
+ * router.navigate -> EnvironmentGuard resolves the :envHrid segment.
+ *
+ * Only the HTTP boundary is stubbed; the environments payload is the one a Management API returns
+ * for an environment that predates hrids.
+ */
+describe('GioSideNavComponent navigation (APIM-14748)', () => {
+  let fixture: ComponentFixture<RootComponent>;
+  let router: Router;
+  let httpTestingController: HttpTestingController;
+
+  const MAX_SETTLE_ROUNDS = 20;
+
+  const DEFAULT_ENV_WITHOUT_HRIDS: Environment = {
+    id: 'DEFAULT',
+    name: 'Default environment',
+    organizationId: 'DEFAULT',
+  };
+
+  const flushPendingEnvironmentCalls = (environment: Environment): number => {
+    const requests = httpTestingController.match(req => req.url.endsWith('/environments'));
+    requests.forEach(request => request.flush([environment]));
+    return requests.length;
+  };
+
+  /**
+   * An enterprise license, so the license directive lets every group through to the router instead of
+   * intercepting the click to offer an upgrade.
+   */
+  const flushLicense = (): number => {
+    const requests = httpTestingController.match(req => req.url === LICENSE_CONFIGURATION_TESTING.resourceURL);
+    requests.forEach(request =>
+      request.flush({
+        tier: 'universe',
+        features: [ApimFeature.APIM_CLUSTER, ApimFeature.APIM_AUDIT_TRAIL, ApimFeature.APIM_API_PRODUCTS, ApimFeature.ALERT_ENGINE],
+        packs: [],
+        expiresAt: new Date(),
+      }),
+    );
+    return requests.length;
+  };
+
+  /**
+   * Answers requests until the router is idle and nothing is left to answer. Each guard hop fetches
+   * the environment list again, so the number of rounds is not knowable here; looping on quiescence
+   * rather than a fixed count keeps assertions off half-finished navigations, and the throw makes a
+   * chain that grows an extra hop fail with its own message instead of a puzzling URL mismatch.
+   */
+  const settleNavigation = (environment: Environment) => {
+    for (let round = 0; round < MAX_SETTLE_ROUNDS; round++) {
+      tick();
+      fixture.detectChanges();
+
+      const answered = flushPendingEnvironmentCalls(environment) + flushLicense();
+      if (answered === 0 && !router.getCurrentNavigation()) {
+        return;
+      }
+    }
+    throw new Error(`Navigation did not settle within ${MAX_SETTLE_ROUNDS} rounds (router.url=${router.url})`);
+  };
+
+  const setup = (environment: Environment) => {
+    TestBed.configureTestingModule({
+      declarations: [RootComponent, ManagementShellComponent, PageComponent],
+      imports: [NoopAnimationsModule, MatIconTestingModule, GioSideNavModule, RouterModule.forRoot(routes)],
+      providers: [
+        provideHttpClient(),
+        provideHttpClientTesting(),
+        { provide: Constants, useValue: { ...CONSTANTS_TESTING, org: { ...CONSTANTS_TESTING.org, environments: [environment] } } },
+        { provide: GioPermissionService, useValue: { hasAnyMatching: () => true, loadEnvironmentPermissions: () => of(undefined) } },
+        {
+          provide: EnvironmentSettingsService,
+          useValue: {
+            load: () => of(undefined),
+            get: () => of({ apiScore: { enabled: true }, portalNext: { access: { enabled: false } } }),
+          },
+        },
+        { provide: SettingsNavigationService, useValue: { getSettingsNavigationSearchItems: () => [] } },
+        { provide: GioMenuSearchService, useValue: new GioMenuSearchService() },
+        { provide: 'LicenseConfiguration', useValue: LICENSE_CONFIGURATION },
+      ],
+    }).overrideProvider(InteractivityChecker, { useValue: { isFocusable: () => true, isTabbable: () => true } });
+
+    router = TestBed.inject(Router);
+    httpTestingController = TestBed.inject(HttpTestingController);
+    fixture = TestBed.createComponent(RootComponent);
+    fixture.detectChanges();
+  };
+
+  /** The environment segment plus the group, dropping whatever the group's own module redirects to. */
+  const groupPathOf = (url: string): string => url.split('?')[0].split('/').slice(0, 3).join('/');
+
+  const landOnUrl = (url: string, environment: Environment) => {
+    let navigationError: unknown;
+    router.navigateByUrl(url).catch(error => (navigationError = error));
+    settleNavigation(environment);
+
+    // A guard blowing up on the way in would otherwise resurface as a puzzling assertion on the
+    // post-click URL, several lines later.
+    if (navigationError) {
+      throw navigationError;
+    }
+    expect(router.url).toBe(url);
+  };
+
+  const clickSidebarGroup = (title: string, environment: Environment) => {
+    // The harness is async while the test drives a virtual clock, so the promise chain is started
+    // and then flushed by tick() rather than awaited.
+    let clicked = false;
+    TestbedHarnessEnvironment.harnessForFixture(fixture, GioSideNavHarness)
+      .then(sideNav => sideNav.clickGroup(title))
+      .then(() => (clicked = true));
+    tick();
+    expect(clicked).toBe(true);
+
+    settleNavigation(environment);
+  };
+
+  afterEach(() => {
+    httpTestingController.verify({ ignoreCancelled: true });
+  });
+
+  it('should navigate to Observability when the environment has no hrids', fakeAsync(() => {
+    setup(DEFAULT_ENV_WITHOUT_HRIDS);
+
+    landOnUrl('/DEFAULT/home', DEFAULT_ENV_WITHOUT_HRIDS);
+    clickSidebarGroup('Observability', DEFAULT_ENV_WITHOUT_HRIDS);
+
+    expect(groupPathOf(router.url)).toBe('/DEFAULT/observability');
+  }));
+
+  it('should navigate to Analytics when the environment has no hrids', fakeAsync(() => {
+    setup(DEFAULT_ENV_WITHOUT_HRIDS);
+
+    landOnUrl('/DEFAULT/home', DEFAULT_ENV_WITHOUT_HRIDS);
+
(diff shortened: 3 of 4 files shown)


gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
```diff
@@ -120,6 +120,12 @@ export class GioSideNavComponent implements OnInit, OnDestroy {
     this.unsubscribe$.complete();
   }
 
+  // Environments created before hrids existed have none; URLs then fall back to the environment id,
+  // which the environment guard resolves just like an hrid.
+  private get currentEnvHrid(): string {
+    return this.currentEnv.hrids?.length > 0 ? this.currentEnv.hrids[0] : this.currentEnv.id;
```
phiz71: **MUST** — the fallback rebuilds the environment segment from `currentEnv`, but the component already receives the exact segment the URL carries: `activatedRoute.params.envHrid`, read on line 92 and then dropped by `switchMap(_ => ...)`. Deriving the base path from that param makes it agree with `router.url` by construction.

Two consequences, both measured on this branch using the new `gio-side-nav.navigation.spec.ts` harness:

- `hrids: ['prod', 'production']`, land on `/production/home` (a valid hrid — `EnvironmentGuard` accepts any entry of the list), click Observability -> **`/prod/observability/overview`**. The environment alias in the URL switches under the user.
- `DEFAULT` with no hrids, land on `/default/home` (the guard lowercases the param, so this resolves), click Observability -> **`/DEFAULT/observability/overview`**. The segment flips case mid-session.

In both cases the child items stay relative (`./observability/overview`) and keep the URL's own segment, so the group header and its children navigate to different prefixes. And `GioMenuItemsComponent` computes `isRouteActive` as `router.url.startsWith(routerBasePath)` — case-sensitive — so a base path whose case or alias differs from the URL never marks the group active.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
```diff
@@ -120,6 +120,12 @@ export class GioSideNavComponent implements OnInit, OnDestroy {
     this.unsubscribe$.complete();
   }
 
+  // Environments created before hrids existed have none; URLs then fall back to the environment id,
+  // which the environment guard resolves just like an hrid.
+  private get currentEnvHrid(): string {
```
phiz71: **COULD** — third live spelling of "first hrid, else id" in the console:

- `EnvironmentService.getFirstHridOrElseId` — `src/services/environment.service.ts:68`, still registered in `app.module.ajs.ts:522`
- `EnvironmentGuard` — `src/management/environment.guard.ts:57`: `get(environments[0], 'hrids[0]') ?? environments[0].id`
- this getter

They disagree on a null/undefined environment: this one throws where the other two return `undefined`. `src/util/` is where that kind of helper lives in this app.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
```diff
@@ -358,6 +359,83 @@ describe('GioSideNavComponent', () => {
...
+          { provide: GioMenuSearchService, useValue: menuSearchService },
+        ],
+      })
+        .overrideProvider(InteractivityChecker, {
+          useValue: { isFocusable: () => true, isTabbable: () => true },
+        })
+        .compileComponents();
+
+      const fix = TestBed.createComponent(GioSideNavComponent);
+      const ctrl = TestBed.inject(HttpTestingController);
```
phiz71: **SHOULD** — the helper binds the controller to a local `ctrl` and never assigns the describe-scoped `httpTestingController` that the root `afterEach` (line 138) calls `verify()` on.

Measured:

```
jest --runTestsByPath .../gio-side-nav.component.spec.ts -t "environment without hrids regression"
```

All three new tests fail with `TypeError: Cannot read properties of undefined (reading 'verify')` at line 138, because no earlier test ran to populate the variable. In a full-file run the `afterEach` verifies a controller belonging to an already-destroyed TestBed, so it protects nothing here.

The APIM-14285 block above (line 335) carries the same trap and fails the same way in isolation, so the pattern is pre-existing — but this diff triples the number of tests it affects. Assigning `httpTestingController` in the helper and dropping the local `ctrl.verify()` calls fixes both.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
```diff
@@ -358,6 +359,83 @@ describe('GioSideNavComponent', () => {
     });
   });
 
+  describe('environment without hrids regression (APIM-14748)', () => {
+    const buildMenuForEnvironment = async (environment: Environment) => {
```
phiz71: **SHOULD** — third copy of the same ~45-line TestBed setup in this file: `init()` (line 61) and the inline block of the APIM-14285 describe (line 299) already build the same providers.

They have already drifted: `init()` still provides `{ params: of({ envId: 'DEFAULT' }) }` while the component reads `envHrid`, so that route param has been dead for a while without anyone noticing. One helper parameterised by environment would also give the `hrids: ['my-env']` case for free, instead of it needing the separate APIM-14285 block.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
```diff
@@ -358,6 +359,83 @@ describe('GioSideNavComponent', () => {
...
+          },
+          {
+            provide: Constants,
+            useFactory: () => {
+              const constants = CONSTANTS_TESTING;
+              constants.org.settings = {
+                ...constants.org.settings,
+                licenseExpirationNotification: { enabled: true },
+              };
+              constants.org.currentEnv = environment;
```
phiz71: **COULD** — `const constants = CONSTANTS_TESTING` takes the module-level singleton by reference, so `org.currentEnv` and `org.environments` are written on the shared object with no restore. After this describe, `CONSTANTS_TESTING.org.currentEnv` stays `{ id: 'DEFAULT', name: 'default', hrids: undefined, organizationId: 'org' }` for every later describe in the file (`target attribute on menu item anchors`, line 438) — `init()` never resets `currentEnv`.

Inert today only because the leaked value shares the id and has no hrids. Add the `hrids: ['my-env']` row and later describes start building `/my-env/...` base paths. Same pattern on line 311, so the real fix is the shared helper above deep-cloning `CONSTANTS_TESTING`.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
```diff
@@ -358,6 +359,83 @@ describe('GioSideNavComponent', () => {
...
+      const ctrl = TestBed.inject(HttpTestingController);
+
+      fix.detectChanges();
+      ctrl.expectOne(`${LICENSE_CONFIGURATION_TESTING.resourceURL}`).flush({ tier: '', features: [], packs: [], expiresAt: new Date() });
+      fix.detectChanges();
+
+      return { fix, ctrl };
+    };
+
+    it.each([
```
phiz71, line 409: **COULD** — the PR description's table for this file lists three rows (`undefined`, `[]`, and `['my-env']` -> `/my-env/observability`), but `it.each` only carries the first two. The behaviour is covered elsewhere — the APIM-14285 block line 295, and the fourth navigation test — so no coverage is actually missing; the table just cannot be checked against the diff as written. Either add the row or drop it from the description.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+@Component({ template: '<router-outlet></router-outlet>', standalone: false })
+class RootComponent {}
+
+// Mirrors the license configuration built at bootstrap in index.ts, so the gioLicense directive can
+// resolve the APIM features the side nav declares.
+const LICENSE_CONFIGURATION = { ...LICENSE_CONFIGURATION_TESTING, featureInfoData: getFeatureInfoData(undefined) };
+
+// Mirrors the real route tree: management-routing.module.ts plus the default redirects declared by
+// observability.module.ts, env-analytics-legacy.module.ts and cluster-routing.module.ts.
+const routes: Routes = [
```
phiz71, line 57: **SHOULD** — this route tree is a hand-copy of `management-routing.module.ts` plus the default redirects of `observability.module.ts`, the legacy analytics module and `cluster-routing.module.ts`, so the end-to-end test asserts against a private copy rather than against the real configuration.

Change `observability.module.ts`'s `{ path: '', redirectTo: 'overview' }` or `cluster-routing.module.ts`'s `redirectTo: 'kafka-standalone'` and the console navigates somewhere else while this test stays green against the stale copy — the exact class of regression it exists to catch. Importing the real route arrays, or asserting only the `/DEFAULT/observability` prefix rather than the resolved leaf, removes the drift.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+    organizationId: 'DEFAULT',
+  };
+
+  const flushPendingEnvironmentCalls = (environment: Environment) => {
+    httpTestingController.match(req => req.url.endsWith('/environments')).forEach(req => req.flush([environment]));
+  };
+
+  /** Lets the router settle: each hop re-runs the guard, which fetches the environment list again. */
+  const settleNavigation = (environment: Environment) => {
+    for (let i = 0; i < 5; i++) {
```
phiz71: **COULD** — a fixed `for (let i = 0; i < 5; i++)` is a magic constant for "how many router hops this test tolerates". Each guard re-run issues another `/environments` call, and five happens to cover today's redirect depth. Add one hop — a new parent route, a second redirect, a guard that re-fetches — and the last navigation never settles; the failure then surfaces as a URL mismatch on the assertion instead of pointing at the exhausted loop. `while (httpTestingController.match(...).length) { ... }` followed by `flush()` says the same thing and cannot run out.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+          features: [ApimFeature.APIM_CLUSTER, ApimFeature.APIM_AUDIT_TRAIL, ApimFeature.APIM_API_PRODUCTS, ApimFeature.ALERT_ENGINE],
+          packs: [],
+          expiresAt: new Date(),
+        }),
+      );
+  };
+
+  /** Reproduces the reported entry point: the user is on the Dashboard after login. */
+  const landOnUrl = (url: string, environment: Environment) => {
+    router.navigateByUrl(url).catch(() => undefined);
```
phiz71: **COULD** — `.catch(() => undefined)` drops any navigation error raised on the way to the starting URL. If `EnvironmentGuard` throws — `No environment found!` on a wrong flush payload, or the `TypeError` it raises on `currentEnvironment.id` when nothing matches — the test carries on silently. Only the first test asserts the landing URL; the Analytics and Kafka tests never check it, so a broken landing shows up later as a confusing assertion on the post-click URL. Asserting `router.url` after every `landOnUrl`, and letting the rejection propagate, keeps the cause visible.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+    settleNavigation(environment);
+    flushLicense();
+    settleNavigation(environment);
+  };
+
+  const clickSidebarGroup = (title: string, environment: Environment) => {
+    const header = Array.from(fixture.nativeElement.querySelectorAll('mat-expansion-panel-header')).find((el: HTMLElement) =>
```
phiz71: **COULD** — `GioSideNavHarness` sits next to the component, and `AGENTS.md` asks for harnesses or stable user-facing selectors where one is available. This selector is tied to the Material internals that `gio-menu-items` happens to render today — the component also has a reduced mode that renders the group through a CDK overlay, with no `mat-expansion-panel-header` at all — and `textContent.includes(title)` matches the first group whose label merely contains the string. A `clickGroup(title)` locator on the harness survives both.

phiz71 (commented):
Reviewed on a local checkout of the branch, running the suites rather than reading them.

**What I measured**

- `jest` on both spec files: **28/28 green**.
- Red counter-check: putting `${this.currentEnv.hrids}` back in place of the new accessor makes **all 6 new tests fail**, with exactly the reported symptoms (`/undefined/observability`, `//observability`, the three groups landing back on `/DEFAULT/home`). The regression tests hold.
- `tsc --noEmit -p tsconfig.typecheck.json`: clean.

The fix is at the right depth. "hrids optional, fall back to the id" is already the contract in `EnvironmentGuard` and on the backend, and the side nav was the only place breaking it, so no migration is needed. One `MUST` below, on where that segment is read from; everything else is test robustness.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
phiz71: Fixed by the spread — the factory now builds its own object instead of writing on the module-level singleton, so nothing leaks into the later describes. Resolving.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
phiz71: Fixed, and better than what I asked for: looping on quiescence with a named ceiling and a throw that reports `router.url` turns an exhausted budget into its own message. Measured on this head: landing settles in 3 rounds, the click in 1, so `MAX_SETTLE_ROUNDS = 20` leaves a comfortable margin. Resolving — one remark on the timer choice goes in its own thread.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+   * Answers requests until the router is idle and nothing is left to answer. Each guard hop fetches
+   * the environment list again, so the number of rounds is not knowable here; looping on quiescence
+   * rather than a fixed count keeps assertions off half-finished navigations, and the throw makes a
+   * chain that grows an extra hop fail with its own message instead of a puzzling URL mismatch.
+   */
+  const settleNavigation = async (environment: Environment) => {
+    for (let round = 0; round < MAX_SETTLE_ROUNDS; round++) {
+      const flushed = flushPendingEnvironmentCalls(environment);
+      flushLicense();
+      await new Promise(resolve => setTimeout(resolve, 0));
```
phiz71: **COULD** — the quiescence loop is the right answer, but it was paid for by leaving `fakeAsync`, and the two are not exclusive. `await new Promise(resolve => setTimeout(resolve, 0))` is a real-clock sleep, which is what `AGENTS.md` asks tests to avoid ("Keep async tests deterministic. Await async work explicitly; avoid sleeps, wall-clock assumptions"); it also gives up what `fakeAsync` was buying — a virtual clock, and the end-of-test check that no timer is left pending.

Keeping `fakeAsync` and swapping this line for `tick()` gives both. Measured on this head: `settleNavigation` stays synchronous, `landOnUrl` and `clickSidebarGroup` lose their `await`, the four tests wrap back in `fakeAsync` — **4/4 green**, and the red counter-check still fails the three no-hrids tests with the same messages. So the detection power is identical and the clock goes back to being virtual.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+
+  /**
+   * Answers requests until the router is idle and nothing is left to answer. Each guard hop fetches
+   * the environment list again, so the number of rounds is not knowable here; looping on quiescence
+   * rather than a fixed count keeps assertions off half-finished navigations, and the throw makes a
+   * chain that grows an extra hop fail with its own message instead of a puzzling URL mismatch.
+   */
+  const settleNavigation = async (environment: Environment) => {
+    for (let round = 0; round < MAX_SETTLE_ROUNDS; round++) {
+      const flushed = flushPendingEnvironmentCalls(environment);
```
phiz71: **Nit, readability only — the mechanism is sound.** `flushed` is read *before* the await while the exit test is evaluated *after* it, so the condition on the next line mixes two instants: "nothing was pending when I looked" and "no navigation is running now". A request that leaves during the await is not counted, and what actually saves the loop is the second half of the condition — `!router.getCurrentNavigation()` is still truthy while the navigation waits on that response, so the next round flushes it.

It holds, and the `afterEach` `verify()` would catch a leftover request loudly rather than silently. Re-reading the count after the await, or naming it `flushedBeforeWait`, would just make the loop say what it does.

phiz71 (commented):
Second pass on the update. Both COULDs I raised on the test helpers are fixed and I have resolved those threads; the `MUST` and the three `SHOULD` are untouched, so they stand as written.

**Re-measured on this head (`47419269`)**

- `jest` on both spec files: **28/28 green**.
- Red counter-check rerun: with `${this.currentEnv.hrids}` back in place of the accessor, **all 6 tests fail** with the same symptoms as before. Dropping `fakeAsync` was the one thing that could have blunted the end-to-end test, and it did not.
- `tsc --noEmit -p tsconfig.typecheck.json`: clean.
- `jest -t "environment without hrids regression"` still fails all three tests with `TypeError: Cannot read properties of undefined (reading 'verify')` — the `httpTestingController` thread is unchanged.

Two remarks below, both on the new `settleNavigation`.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
SamHardi: Fixed in b92d2a9a52 (commit reworded to `fix(console): build sidebar group paths from the environment segment in the URL`). The base paths and the menu-search links now come from the `envHrid` route param, captured in the `switchMap` that previously dropped it.

I reproduced your multi-hrid case on the harness before changing anything — `hrids: ['prod','production']`, land on `/production/home`, click Observability -> `/prod/observability/overview`. That alias switch is the argument that decided it, and it holds regardless of casing.

One correction on a sub-claim, because it sent me down a wrong path earlier: "a base path whose case or alias differs from the URL never marks the group active" is true, but a matching one does not mark it active either. With `hrids: ['default']` and the URL exactly matching the base path, no panel carries `active` on a direct load. `GioMenuItemsComponent` builds `isRouteActive` as a `toSignal` over `router.events` with `startWith(null)`, evaluated at construction before Angular sets the `routerBasePath` input — so it reads `undefined`, yields false, and only recomputes on the next `NavigationEnd`. A sidebar created after navigation ends never sees one. That is a ui-particles bug independent of this PR; I have it noted to raise there separately.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.ts
SamHardi: Fixed in b92d2a9a52 — by deletion rather than extraction. Deriving the segment from the route param removed the need for "first hrid, else id" here at all, so the third copy is gone and no new util was needed. The two remaining copies (`getFirstHridOrElseId` and the guard) are untouched: consolidating them means touching the legacy AngularJS service and the guard, which is beyond this fix. Worth its own cleanup if you want it.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
SamHardi: Fixed in b92d2a9a52. The helper now assigns the describe-scoped `httpTestingController` and the local `ctrl.verify()` calls are gone, so the root `afterEach` verifies the live controller.

Verified the way you measured it: `jest --runTestsByPath .../gio-side-nav.component.spec.ts -t "environment segment in menu paths"` — 4 passed, 21 skipped, no `TypeError`. I had hit that same error early on and written it off as pre-existing noise; you were right that it protects nothing as written.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
SamHardi: Partly fixed in b92d2a9a52, and I want to be explicit about what I left.

Done: `init()` now provides `{ params: of({ envHrid: 'DEFAULT' }) }`. That dead param stopped being harmless with this change — once the base paths come from the route param, `envId` would have built `/undefined/...` in every test using `init()`. Good catch. My own block is also down to one parameterised helper.

Not done: folding the APIM-14285 block into the same helper. It exists to pin a build-order race (currentEnv set before `environmentSettingsService.get` resolves), which needs its own `Subject`-driven settings provider and an assertion between the two emissions — a shared helper would have to grow a hole for that, and rewriting someone else's regression test is a good way to quietly lose what it pins. The `hrids: ['my-env']` case you wanted from it is now covered directly by the third `it.each` row. Happy to do the full consolidation as a follow-up if you would rather have one helper.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
SamHardi: Fixed in b92d2a9a52 — the row is now in `it.each` rather than only in the description. The three rows are `no hrids` / `empty hrids` / `a single hrid`, and since the segment now comes from the URL the parameter that varies is the route param, with hrids along for the ride: `['my-env']` with segment `my-env` gives `/my-env/observability`. The PR description table is updated to match.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
SamHardi: Fixed in b92d2a9a52 by taking your second option. The tests now assert the environment segment and the group — `/DEFAULT/observability` — through a `groupPathOf` helper that drops the leaf, so what each module redirects to internally is no longer this test's business. Changing `observability.module.ts`'s default no longer leaves a stale copy silently green, because the copy no longer carries that assertion.

I did not import the real route arrays: they are all `loadChildren` behind lazy modules, so pulling them in drags most of the console into a test whose subject is one component's base paths. The route tree that remains is the shape (`:envHrid` + guard + the three groups), which is what the navigation chain actually exercises.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
SamHardi: Fixed in b92d2a9a52. `landOnUrl` now captures the rejection and rethrows it after the loop, then asserts `router.url` — so every test checks its landing, not just the first, and a guard blowing up on the way in reports itself instead of surfacing as a confusing post-click mismatch.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
SamHardi: Fixed in b92d2a9a52. `GioSideNavHarness` gained `clickGroup(title)`, and the spec goes through it. Exact title match rather than `includes`, as you pointed out.

Two things worth flagging. It collides with your `fakeAsync` remark on the other thread: harness methods are async, the test drives a virtual clock, so the spec starts the promise chain and flushes it with `tick()` rather than awaiting — noted in a comment at the call site. And I implemented the expanded-mode path only: it throws a named error if no group matches, but does not yet reach the CDK overlay of reduced mode, since no test exercises that. The locator is at least now the place to add it.

Say the word if you would rather have the reduced-mode branch in now.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
SamHardi: Fixed in b92d2a9a52, and your measurement was right — including on a point I had got wrong.

I had left `fakeAsync` because I suspected `fakeAsync(async () => ...)` with `compileComponents()` awaited inside the fake zone. The real answer was to remove the `await`, not the `fakeAsync`: `setup()` is now synchronous (the components are inline-compiled under Jest, so `compileComponents()` was doing nothing), `settleNavigation` uses `tick()`, and the four tests wrap back in `fakeAsync` — 4/4 green. No real-clock sleep, virtual clock and end-of-test timer check both back.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
SamHardi: Fixed in b92d2a9a52. The loop now ticks first and counts after, so the count and the exit test are read at the same instant:

```ts
tick();
fixture.detectChanges();
const answered = flushPendingEnvironmentCalls(environment) + flushLicense();
if (answered === 0 && !router.getCurrentNavigation()) {
  return;
}
```

`flushLicense` returns a count too, so the license request is part of the same signal rather than a special case handled elsewhere.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.component.spec.ts
phiz71: Agreed, and your reasoning is the right one — a shared helper that has to grow a hole for the Subject-driven settings provider stops being shared, and rewriting a regression test you did not write is how its subject quietly goes missing. Leaving the APIM-14285 block alone is the correct call; no follow-up needed from my side.

The two things I actually cared about are both in: `init()` no longer provides the dead `envId` param — which, as you say, stopped being harmless the moment the paths started coming from the route param — and your own block is down to one parameterised helper. Resolving.

phiz71 (commented):
Third pass. Every thread is now answered and resolved — the `MUST` included, and the one place we disagreed (folding the APIM-14285 block) you argued out of, correctly.

**Re-verified on `b92d2a9a` rather than taken on trust**

- The two scenarios that justified the `MUST` are gone: `hrids: ['prod','production']` landing on `/production/home` now clicks through to `/production/observability/overview` (was `/prod/...`), and `DEFAULT` with no hrids landing on `/default/home` reaches `/default/observability/overview` (was `/DEFAULT/...`).
- The risk the new approach introduces — everything now rides on the route param — is covered: `app-routing.module.ts` sets `paramsInheritanceStrategy: 'always'`, and `gio-side-nav` is rendered by `ManagementComponent`, mounted on `path: ''` under `:envHrid`. The param cannot be missing, so this cannot regress into `/undefined/`.
- 29/29 green; red counter-check still fails 6; `-t "environment segment in menu paths"` 4 green with no `TypeError`; `tsc --noEmit` clean.

And thank you for the `isRouteActive` correction — you are right, and my wording implied the opposite. A *matching* base path does not mark the group active either, for the reason you give: the `toSignal` is built in the constructor, before Angular sets the `routerBasePath` input, so it reads `undefined` and only recomputes on the next `NavigationEnd`. My own probe had shown `active: false` on both the matching and the mismatched URL and I failed to draw the conclusion from it. Worth raising in ui-particles as you say.

**One thing left, and it has no line to anchor on: the manual test report now describes the previous implementation.**

The scenario states the starting point as `#!/default/home/overview` — lowercase — while the table reports landing on `#!/DEFAULT/observability/overview` — uppercase. That case flip is the signature of the `currentEnv.id` fallback; it is exactly what the `MUST` was about. With the code as it stands the run would end on `#!/default/observability/overview`. So the table contradicts its own scenario, which means the stack run was not repeated after the rework.

That report is the only evidence in this PR that the fix behaves on a real stack, and as written it attests to a mechanism that no longer exists. Either repeat the run and regenerate the table, or drop it and say the automated coverage carries it. I have checked the routing configuration statically and the end-to-end spec drives a real `Router` through a real `:envHrid` tree, so I do not think the behaviour is in doubt — the report just no longer matches the code.

gravitee-apim-console-webui/src/components/gio-side-nav/gio-side-nav.navigation.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+    setup(DEFAULT_ENV_WITHOUT_HRIDS);
+
+    landOnUrl('/DEFAULT/home', DEFAULT_ENV_WITHOUT_HRIDS);
+
+    clickSidebarGroup('Analytics', DEFAULT_ENV_WITHOUT_HRIDS);
+
+    expect(groupPathOf(router.url)).toBe('/DEFAULT/analytics');
+  }));
+
+  it('should navigate to Kafka when the environment has no hrids', fakeAsync(() => {
```
phiz71, line 233: **Backport note, not a defect here — worth knowing before Mergify opens the `4.11.x` PR.**

I checked the two support branches. `4.12.x` carries the same four `currentEnv.hrids` sites as `master`, so that backport should apply as is. `4.11.x` has three, and the description's "the same edit minus the Kafka group" is off on two counts:

- The Kafka group *does* exist on `4.11.x`. It is `displayName: 'Kafka Clusters'` with a **relative** `routerLink: './clusters'` — a flat item with no `routerBasePath`. It escapes the bug precisely because it is relative, not because it is absent.
- It is not one hunk to drop. This test file's Kafka case (here, plus the `clusters` branch of the route tree above) and `basePathOf('Kafka')` in `gio-side-nav.component.spec.ts` have no counterpart there either, so all three files need a hand.

Everything else the new spec leans on is present on `4.11.x`: `ApimFeature.APIM_CLUSTER` / `APIM_API_PRODUCTS` / `ALERT_ENGINE` / `APIM_AUDIT_TRAIL`, `getFeatureInfoData`, `EnvironmentGuard.initEnvConfigAndLoadPermissions`, `observability.module.ts` and `cluster-routing.module.ts`. So the Kafka strand is the whole of the work.

SamHardi: Both corrections verified against the branches, and both were my error — the description is fixed.

- `git show origin/4.11.x:.../gio-side-nav.component.ts` line 183: `routerLink: ./clusters`, `displayName: 'Kafka Clusters'`, no `routerBasePath`. Flat and relative, exactly as you said — it escapes the bug rather than being absent. I had written "Kafka/clusters does not exist there", which is simply wrong.
- Site counts: `grep -c "currentEnv.hrids"` gives 4 on `4.12.x` (same as `master`) and 3 on `4.11.x`. So "the same edit minus the Kafka group" understated the shape of the work as well as getting the reason wrong.

I also confirmed the dependency list you checked — `ApimFeature.APIM_CLUSTER` / `APIM_API_PRODUCTS` / `ALERT_ENGINE` / `APIM_AUDIT_TRAIL`, `getFeatureInfoData`, `EnvironmentGuard.initEnvConfigAndLoadPermissions`, `observability.module.ts`, `cluster-routing.module.ts` all present on `4.11.x`.

One addition to your list, found while checking: **`gio-side-nav.component.harness.ts` does not exist on `4.11.x` at all** — `git ls-tree origin/4.11.x -- .../gio-side-nav/` returns five files, no harness. Since the navigation spec now clicks through `clickGroup()`, the harness file has to come along or the spec falls back to a DOM query there. That makes it four files needing a hand, not three.

Leaving this open — it is a note for whoever picks up the Mergify PR, not something to close from here.

phiz71 (approved):
Approving. The fix is right, and it is right at the depth it should be: the sidebar now names the environment with the segment the URL already carries, instead of re-deriving a name that could differ from it.

**Verified on `b92d2a9a`, not taken on trust**

- The two scenarios that justified the `MUST` are gone: `hrids: ['prod','production']` landing on `/production/home` clicks through to `/production/observability/overview` (was `/prod/...`), and `DEFAULT` with no hrids landing on `/default/home` reaches `/default/observability/overview` (was `/DEFAULT/...`).
- The risk the rework introduces — everything now rides on the route param — is covered: `app-routing.module.ts` sets `paramsInheritanceStrategy: 'always'` and `gio-side-nav` is rendered by `ManagementComponent`, mounted on `path: ''` under `:envHrid`. The param cannot go missing, so this cannot regress into `/undefined/`.
- 29/29 green. Red counter-check: restoring `${this.currentEnv.hrids}` fails all 6 new tests with the reported symptoms, so the regression stays pinned. `-t "environment segment in menu paths"` runs clean in isolation. `tsc --noEmit` clean.

Every thread is answered, and the two you pushed back on you were right about — leaving the APIM-14285 block alone, and the `isRouteActive` correction I had wrong. You also caught the missing harness on `4.11.x`, which I had in front of me and walked past.

**Two things left that do not block this PR**

*The manual test report still describes the previous implementation.* The scenario starts on `#!/default/home/overview` — lowercase — while the table lands on `#!/DEFAULT/observability/overview` — uppercase. That case flip is the signature of the `currentEnv.id` fallback the rework removed; the run would now end on `#!/default/...`. Worth regenerating or dropping, but it is the description, not the code.

*The SonarCloud gate is red, and it is a false negative caused by my own review comment.* `new_coverage` is 20% against a 40% threshold, on 5 new lines to cover in total:

| file | new lines | to cover | uncovered |
|---|---|---|---|
| `gio-side-nav.component.harness.ts` | 15 | 4 | **4** |
| `gio-side-nav.component.ts` | 11 | 1 | 0 |
| both spec files | 331 | 0 | 0 |

`clickGroup()` is exercised by the navigation spec, but `jest.config.js` excludes `!src/**/*.harness.ts` from `collectCoverageFrom`, so the harness never reaches `lcov.info` — while `sonar-project.properties` only excludes `*.stories.ts` and `*.fixture.ts` from coverage. Sonar therefore sees a source file with no coverage data and counts its executable lines as uncovered. Asking for that harness method is what put those lines there; without it the gate would be green.

Sonar is not a required check here (only `ci/circleci: Validate workflow status` is, and it passes), so this is not in your way. The mismatch is structural rather than yours — any PR adding a method to a `.harness.ts` without much other new code will hit it — and one line settles it either way: add `**/*.harness.ts` to `sonar.coverage.exclusions`, or drop the exclusion on the jest side so harnesses are measured. Your call whether that rides along here or gets its own ticket.

PR #19500, opened 2026-08-31
fix(automation): proper API item naming from GKO CRD

MateuszJozwiak-Incubly:
## Issue

https://gravitee.atlassian.net/browse/EXT-181

## Description

When an API was published to the nextgen dev portal via a GKO `PortalListing` CRD, the resulting navigation entry's title showed the CRD's `apiHrid` (e.g. `my-api-name`) instead of the API's real display name (e.g. `Echo API Declarative`).

`NavigationItemEntryMaterializer` set the navigation item's title directly from `entry.apiHrid()` on both create and update, with no way to resolve the referenced API's actual name. The console path (`PortalNavigationItemDomainService`) already did this correctly, looking up the `Api` aggregate via `ApiCrudService.get(...)` and using `api.getName()` — including throwing `ApiNotFoundException` when the API doesn't exist.

The fix brings the `PortalListing` automation path in line with console on both counts:

1. **Correct title.** `NavigationItemEntryMaterializer.upsert()` now resolves the title via `apiCrudService.get(apiId).getName()` — the same call console makes — applied on both the create and update branches. Segment/slug derivation is unchanged, still based on `apiHrid`, so URLs stay stable if the API is later renamed.

2. **Consistent failure on a missing API.** Rather than silently falling back to the hrid when the referenced API doesn't exist, `upsert()` now throws `ApiNotFoundException`, matching console's behavior exactly.

3. **Dry-run parity.** `ValidatePortalListingDomainService` — the single validator already shared by both `CreateOrUpdatePortalListingUseCase` (real apply) and `ValidatePortalListingUseCase` (dry-run) — now checks `apiCrudService.existsById(...)` per listing entry and reports a severe `Validator.Error` when an entry's API can't be found. Because both use cases already consume this one validator (real apply already throws `ValidationDomainException` on any severe error; dry-run already returns `output.errors()`), the Automation API's `?dryRun=true` path gets the same "API not found" signal as the real apply, with no changes needed to either use case or the REST resource.

The portal-reference tolerance (a `PortalListing` may still reference a portal that doesn't exist yet) is untouched — this change is scoped to the API reference only.

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/NavigationItemEntryMaterializer.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/NavigationItemEntryMaterializer.java
@@ -17,7 +17,6 @@ package io.gravitee.apim.core.portal_listing.domain_service;
 
 import io.gravitee.apim.core.DomainService;
 import io.gravitee.apim.core.api.crud_service.ApiCrudService;
-import io.gravitee.apim.core.api.model.Api;
 import io.gravitee.apim.core.audit.model.AuditInfo;
 import io.gravitee.apim.core.portal.domain_service.navigation.PortalNavigationValidator;
 import io.gravitee.apim.core.portal.exception.PathConflictException;
@@ -55,7 +54,7 @@ class NavigationItemEntryMaterializer {
     PortalNavigationApi upsert(AuditInfo auditInfo, PortalId portalId, String apiId, PortalListingApiEntry entry) {
         var navApiId = rowId(auditInfo, portalId, apiId);
         var parent = resolveParent(auditInfo, portalId.toString(), entry.location());
-        var title = apiCrudService.findById(apiId).map(Api::getName).orElse(entry.apiHrid());
+        var title = apiCrudService.get(apiId).getName();
 
         var existing = navigationItemsQueryService.findByIdAndEnvironmentId(auditInfo.environmentId(), navApiId);
         if (existing instanceof PortalNavigationApi navApi) {
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/NavigationItemEntryMaterializer.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/NavigationItemEntryMaterializer.java
@@ -16,6 +16,8 @@
 package io.gravitee.apim.core.portal_listing.domain_service;
 
 import io.gravitee.apim.core.DomainService;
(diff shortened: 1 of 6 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/NavigationItemEntryMaterializer.java
```diff
@@ -45,20 +47,22 @@ class NavigationItemEntryMaterializer {
     private final PortalNavigationItemCrudService navigationItemCrudService;
     private final PortalNavigationItemsQueryService navigationItemsQueryService;
     private final ApiDocumentationSyncDomainService apiDocumentationSyncDomainService;
+    private final ApiCrudService apiCrudService;
 
     PortalNavigationApi upsert(AuditInfo auditInfo, PortalId portalId, String apiId, PortalListingApiEntry entry) {
         var navApiId = rowId(auditInfo, portalId, apiId);
         var parent = resolveParent(auditInfo, portalId.toString(), entry.location());
+        var title = apiCrudService.findById(apiId).map(Api::getName).orElse(entry.apiHrid());
```
konrad-krol-incubly: I would say we should treat this behavior the same way as webconsole does -> ApiNotFoundException instead of fallback.

MateuszJozwiak-Incubly: Changed to throw the exception and also added this for Validation

konrad-krol-incubly:
lgtm, just this comment is out of date: "Falls back to apiHrid when the referenced API hasn't synced yet". Please replace this description with info that API existence is now required - otherwise its not possible to determine title for nav item.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/ValidatePortalListingDomainService.java
```diff
@@ -47,7 +50,11 @@ public Result<Input> validateAndSanitize(Input input) {
         errors.addAll(portalAutomationScopeEnforcer.validate(input.auditInfo(), input.portalId(), "portalHrid"));
         List<PortalListingApiEntry> apis = input.apis() == null ? List.of() : input.apis();
         for (int i = 0; i < apis.size(); i++) {
-            errors.addAll(NavigationPathValidator.validate(apis.get(i).location(), "apis[" + i + "].location"));
+            var entry = apis.get(i);
+            errors.addAll(NavigationPathValidator.validate(entry.location(), "apis[" + i + "].location"));
+            if (!apiCrudService.existsById(entry.apiId(input.auditInfo()))) {
```
konrad-krol-incubly: Please consider optimization here, api crud service exposes `findByIds` that can be used very easily and result can be collected to map.

PR #19502, opened 2026-08-31
feat(portal-next): resolve navigation visibility with caller-then-par…

konrad-krol-incubly:
## Issue

## Description

Removes hardcoded `PortalVisibility.PUBLIC` from every automation-api related sync service that materializes portal navigation items. Producers now inherit the parent's visibility (fallback: PUBLIC at root), so a PUBLIC child under a PRIVATE parent can no longer be materialized silently.

Existing `ParentRule` is reused to validate parent-child visibility constraints.

Requires follow-up PR to expose visibility property on automation-api

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/NavigationFolderMapper.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/NavigationFolderMapper.java
@@ -18,7 +18,7 @@ package io.gravitee.apim.core.portal.domain_service.navigation;
 import io.gravitee.apim.core.portal.domain_service.navigation.actions.FolderActions;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationFolder;
 import io.gravitee.apim.core.portal_page.model.PortalNavigationItemContainer;
-import io.gravitee.apim.core.portal_page.model.PortalNavigationItemId;
+import io.gravitee.apim.core.portal_page.model.PortalVisibility;
 import jakarta.annotation.Nullable;
 import java.util.Objects;
 
@@ -29,13 +29,14 @@ public final class NavigationFolderMapper {
     public static boolean matches(
         PortalNavigationFolder existing,
         FolderActions.DesiredFolder desired,
-        @Nullable PortalNavigationItemId parentId
+        @Nullable PortalNavigationItemContainer parent
     ) {
         return (
             Objects.equals(existing.getTitle(), desired.title()) &&
             Objects.equals(existing.getSegment(), desired.segment().value()) &&
             existing.getOrder() == desired.order() &&
-            Objects.equals(existing.getParentId(), parentId)
+            Objects.equals(existing.getParentId(), parent == null ? null : parent.getId()) &&
+            Objects.equals(existing.getVisibility(), PortalVisibility.inheritedFrom(parent))
         );
     }
 
@@ -43,6 +44,7 @@ public final class NavigationFolderMapper {
         target.setTitle(source.title());
         target.setSegment(source.segment().value());
         target.setOrder(source.order());
+        target.setVisibility(PortalVisibility.inheritedFrom(parent));
         if (parent == null) {
             target.markAsRoot();
         } else {
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationItemValidatorService.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationItemValidatorService.java
@@ -130,10 +130,28 @@ public class PortalNavigationItemValidatorService implements PortalNavigationVal
             .stream()
             .filter(item -> item.getId() != null)
             .collect(Collectors.toMap(CreatePortalNavigationItem::getId, Function.identity(), (first, ignored) -> first));
-        List<PendingSegmentClaim> pendingSegmentClaims = collectPendingSegmentClaims(creates, updates);
 
-        CreateValidationContext createCtx = new CreateValidationContext(navigationItems, itemsById, pendingItemsById, pendingSegmentClaims);
-        UpdateValidationContext updateCtx = new UpdateValidationContext(navigationItems, itemsById, pendingItemsById, pendingSegmentClaims);
+        List<PendingUpdate> augmentedUpdates = withDescendantsOfPendingContainers(creates, updates, environmentId);
+        List<PendingSegmentClaim> pendingSegmentClaims = collectPendingSegmentClaims(creates, augmentedUpdates);
+        Map<PortalNavigationItemId, PendingUpdate> pendingUpdatesByExistingId = updates
+            .stream()
+            .filter(u -> u.existing() != null && u.existing().getId() != null)
+            .collect(Collectors.toMap(u -> u.existing().getId(), Function.identity(), (first, ignored) -> first));
+
+        CreateValidationContext createCtx = new CreateValidationContext(
+            navigationItems,
+            itemsById,
+            pendingItemsById,
+            pendingUpdatesByExistingId,
+            pendingSegmentClaims
+        );
+        UpdateValidationContext updateCtx = new UpdateValidationContext(
+            navigationItems,
+            itemsById,
+            pendingItemsById,
+            pendingUpdatesByExistingId,
+            pendingSegmentClaims
+        );
 
         for (BulkCreatePortalNavigationItemValidationRule rule : bulkCreateRules) {
             rule.validate(creates, environmentId, createCtx);
@@ -141,9 +159,67 @@ public class PortalNavigationItemValidatorService implements PortalNavigationVal
         for (CreatePortalNavigationItem item : creates) {
             applyValidationRules(item, createCtx, environmentId);
         }
-        for (PendingUpdate pending : updates) {
+        for (PendingUpdate pending : augmentedUpdates) {
             applyValidationRules(pending, updateCtx);
         }
+    }
+
+    // Include pre-existing descendants of pending container creates and of pending container updates that change visibility.
+    private List<PendingUpdate> withDescendantsOfPendingContainers(
+        List<CreatePortalNavigationItem> creates,
+        List<PendingUpdate> updates,
+        String environmentId
+    ) {
+        var descendantsOfCreates = creates
+            .stream()
+            .filter(PortalNavigationItemValidatorService::isPendingContainer)
+            .flatMap(container -> descendantsOf(container.getId(), environmentId).stream());
+        var descendantsOfUpdates = updates
+            .stream()
+            .filter(PortalNavigationItemValidatorService::isContainerVisibilityChange)
+            .flatMap(update -> descendantsOf(update.existing().getId(), environmentId).stream());
+        var descendantUpdates = Stream.concat(descendantsOfCreates, descendantsOfUpdates).map(
+            PortalNavigationItemValidatorService::asUnchangedUpdate
+        );
+        return Stream.concat(updates.stream(), descendantUpdates).toList();
+    }
+
+    private static boolean isPendingContainer(CreatePortalNavigationItem item) {
+        return item.getId() != null && item.getType().isContainer();
+    }
+
+    private static boolean isContainerVisibilityChange(PendingUpdate update) {
+        var existing = update.existing();
+        var toUpdate = update.toUpdate();
+        return (
+            existing != null &&
+            existing.getId() != null &&
+            existing.getType().isContainer() &&
+            toUpdate.getVisibility() != null &&
+            toUpdate.getVisibility() != existing.getVisibility()
+        );
+    }
+
+    private List<PortalNavigationItem> descendantsOf(PortalNavigationItemId containerId, String environmentId) {
+        return navigationItemsQueryService.findByParentIdAndEnvironmentId(environmentId, containerId);
+    }
+
+    // Wraps a persisted descendant as a no-op update so the update rules re-run against the pending parent.
+    private static PendingUpdate asUnchangedUpdate(PortalNavigationItem descendant) {
+        return new PendingUpdate(unchangedUpdateFor(descendant), descendant);
+    }
+
+    private static UpdatePortalNavigationItem unchangedUpdateFor(PortalNavigationItem item) {
+        return UpdatePortalNavigationItem.builder()
+            .type(item.getType())
+            .title(item.getTitle())
+            .segment(item.getSegment())
+            .order(item.getOrder())
+            .parentId(item.getParentId())
+            .visibility(item.getVisibility())
+            .published(item.getPublished())
+            .source(item.getSource())
+            .build();
     }
 
     @Override
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItemContainer.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItemContainer.java
@@ -26,6 +26,8 @@ public interface PortalNavigationItemContainer {
 
     PortalNavigationItemId getRootId();
 
+    PortalVisibility getVisibility();
+
     /**
      * Placeholder parent used by materializers when the actual folder hasn't been persisted yet.
      * Returns {@code null} when the given id is {@code null}.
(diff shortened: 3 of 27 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItemContainer.java
```diff
@@ -42,6 +44,11 @@ public PortalNavigationItemId getId() {
             public PortalNavigationItemId getRootId() {
                 return id;
             }
+
+            @Override
+            public PortalVisibility getVisibility() {
+                return PortalVisibility.PUBLIC;
```
dev-marek, line 50: Blocker: this hardcodes `PUBLIC` for the phantom placeholder. All four materializers (`NavigationItemEntryMaterializer`, `ApiDocumentationSyncDomainService`, `PortalDocumentationSyncDomainService`, `PortalLinkSyncDomainService`) fall back to `phantom()` whenever the parent folder hasn't been persisted yet — `ParentRule.java`'s own comment calls this an expected automation scenario ("automation may declare a parent that hasn't materialized yet"). When that happens, `PortalVisibility.inheritedFrom(phantom)` returns `PUBLIC` regardless of what the real, not-yet-created parent's actual inherited visibility will be (e.g. a PRIVATE grandparent). That reproduces, one hop removed, the "PUBLIC child silently materialized under a PRIVATE ancestor" bug this PR sets out to fix.

Suggestion: `phantom()` needs to resolve the eventual parent's real inherited visibility (e.g. walk up the intended hierarchy) rather than defaulting to PUBLIC, or the materializers need to re-resolve visibility once the real parent is persisted.

dev-marek (changes requested):
## Review

Scope checked: diff of `origin/refactor/ext-179-consolidate-nav-validation...HEAD` (matches this PR's file list); read all 13 changed files plus call sites of `PortalNavigationItem.from()`, `NavigationFolderMapper`, and the REST v2 OpenAPI navigation-item DTOs; compiled `gravitee-apim-rest-api-service` and ran the full test classes for every touched production file (84/84 green).

The refactor to `PortalVisibility.resolve()`/`inheritedFrom()` is clean, and the `CreatePortalNavigationItem` side (relying on `PortalNavigationItem.from()` to resolve visibility from the real parent) is correctly wired for every materializer's *create* path. But the fix has two gaps that reproduce the exact bug it's meant to close — a silently-materialized item whose visibility doesn't match its lineage — and the only new test is an isolated unit test of the static helper methods, so nothing in the diff would catch either gap.

### Findings

**Blocker — `PortalNavigationItemContainer.java:48-51`** (inline comment below)

**Blocker — test coverage, no test exercises the new inheritance wiring against a non-default (PRIVATE) parent at any actual production call site.** Only the isolated `PortalVisibilityTest` covers `PRIVATE`; all pre-existing tests for `NavigationSyncPlanExecutorTest`, `PortalNavigationSyncDomainServiceTest`, `ApiDocumentationSyncDomainServiceTest`, `PortalDocumentationSyncDomainServiceTest`, `PortalLinkSyncDomainServiceTest`, and `ParentRuleTest` still only assert the default/PUBLIC case end-to-end. I ran all of them: 84/84 pass even with the phantom-PUBLIC bug present — that's exactly why it wasn't caught. Please add a case per materializer where a child is placed under a PRIVATE folder (or a phantom-not-yet-materialized folder chained to a PRIVATE ancestor) and assert the persisted item's visibility.

**Improvement (widened-by-change) — `NavigationFolderMapper.java:42-51` / `NavigationSyncPlanExecutor.java:129-137`.** Unlike the API/PAGE/LINK materializers touched in this PR (which recompute `visibility = PortalVisibility.inheritedFrom(parent)` on every upsert), `NavigationFolderMapper.apply()` never sets `visibility`, and `NavigationSyncPlanExecutor.applyUpdate` short-circuits via `NavigationFolderMapper.matches()` (title/segment/order/parentId only, no visibility) before even calling `apply()`. So once a FOLDER is created, its visibility is frozen: if an ancestor's visibility later changes, existing folders (both top-level via `PortalNavigationSyncDomainService` and API-doc subtree via `ApiFolderSubtreeReconciler`) never get re-resolved on subsequent syncs. This was inert before (folder visibility was always hardcoded PUBLIC), but now that inheritance is load-bearing, it's a real staleness gap specific to the FOLDER type.

### Residual risk (not a code defect)

The PR description says "Requires follow-up PR to expose visibility property on REST API," but `visibility` is already a **required** field on `BaseCreatePortalNavigationItem`/`BaseUpdatePortalNavigationItem` in `openapi-environments.yaml`, and was already present on the base branch before this PR's commit. Worth updating the description so reviewers/readers don't misjudge current REST exposure — the actual gap this PR closes is scoped to the automation-materializer paths, not REST exposure. I confirmed the `ParentRule` PUBLIC-under-PRIVATE guard is unchanged and still correctly blocks explicit REST-driven creates/updates.

konrad-krol-incubly:
Regarding following point I would like to push back, as it still requires follow up PR:

> ### Residual risk (not a code defect)
> The PR description says "Requires follow-up PR to expose visibility property on REST API," but `visibility` is already a **required** field on `BaseCreatePortalNavigationItem`/`BaseUpdatePortalNavigationItem` in `openapi-environments.yaml`, and was already present on the base branch before this PR's commit. Worth updating the description so reviewers/readers don't misjudge current REST exposure — the actual gap this PR closes is scoped to the automation-materializer paths, not REST exposure. I confirmed the `ParentRule` PUBLIC-under-PRIVATE guard is unchanged and still correctly blocks explicit REST-driven creates/updates.

automation open-api.yaml don't reference the existing Visibility schema, so GKO users can't declare visibility.

I'll rephrase the description to be more precise that it's about automation api.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/model/PortalNavigationItemContainer.java
konrad-krol-incubly: Fixed - didn't touch phantom's PUBLIC default (kept as the transient assumption), but the batch validator now feeds pre-existing DB descendants of pending container creates as no-op updates (withDescendantsOfPendingContainers), and ParentRule's update path (mirrored from the create-side in PR #19381) cross-checks them against the pending parent.

I think this is the easiest way to re-validate these descendants without introducing extra path just for this single case. Any change like that goes naturally through full ParentRule validation.

konrad-krol-incubly:
> **Blocker — test coverage, no test exercises the new inheritance wiring against a non-default (PRIVATE) parent at any actual production call site.** Only the isolated `PortalVisibilityTest` covers `PRIVATE`; all pre-existing tests for `NavigationSyncPlanExecutorTest`, `PortalNavigationSyncDomainServiceTest`, `ApiDocumentationSyncDomainServiceTest`, `PortalDocumentationSyncDomainServiceTest`, `PortalLinkSyncDomainServiceTest`, and `ParentRuleTest` still only assert the default/PUBLIC case end-to-end. I ran all of them: 84/84 pass even with the phantom-PUBLIC bug present — that's exactly why it wasn't caught. Please add a case per materializer where a child is placed under a PRIVATE folder (or a phantom-not-yet-materialized folder chained to a PRIVATE ancestor) and assert the persisted item's visibility.

Added: PRIVATE-parent inheritance tests per materializer (PortalLinkSyncDomainServiceTest, PortalDocumentationSyncDomainServiceTest, ApiDocumentationSyncDomainServiceTest), plus CreatePortalNavigationItemValidatorServiceTest.PhantomThenPrivateParent 

> Improvement (widened-by-change) — NavigationFolderMapper.java:42-51 / NavigationSyncPlanExecutor.java:129-137. Unlike the API/PAGE/LINK materializers touched in this PR (which recompute visibility = PortalVisibility.inheritedFrom(parent) on every upsert), NavigationFolderMapper.apply() never sets visibility, and NavigationSyncPlanExecutor.applyUpdate short-circuits via NavigationFolderMapper.matches() (title/segment/order/parentId only, no visibility) before even calling apply(). So once a FOLDER is created, its visibility is frozen: if an ancestor's visibility later changes, existing folders (both top-level via PortalNavigationSyncDomainService and API-doc subtree via ApiFolderSubtreeReconciler) never get re-resolved on subsequent syncs. This was inert before (folder visibility was always hardcoded PUBLIC), but now that inheritance is load-bearing, it's a real staleness gap specific to the FOLDER type.

Fixed: NavigationFolderMapper.matches() now compares visibility via inheritedFrom(parent), and apply() sets it - automation-managed folders re-derive on every sync, so ancestor-visibility changes propagate on the next apply.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/NavigationFolderMapper.java
```diff
@@ -29,20 +29,22 @@ private NavigationFolderMapper() {}
...
+            Objects.equals(existing.getParentId(), parent == null ? null : parent.getId()) &&
+            Objects.equals(existing.getVisibility(), PortalVisibility.inheritedFrom(parent))
         );
     }
 
     public static void apply(FolderActions.DesiredFolder source, PortalNavigationItemContainer parent, PortalNavigationFolder target) {
         target.setTitle(source.title());
         target.setSegment(source.segment().value());
         target.setOrder(source.order());
+        target.setVisibility(PortalVisibility.inheritedFrom(parent));
```
dev-marek, line 47: Blocker — untested behavior change. `matches()`/`apply()` now recompute a folder's visibility from its parent on every sync, which is exactly the fix for my earlier 'widened-by-change' note. But nothing exercises it: there's no `NavigationFolderMapperTest`, and `NavigationSyncPlanExecutorTest` / `PortalNavigationSyncDomainServiceTest` still only use `PortalVisibility.PUBLIC` in every folder fixture (checked both files — zero PRIVATE folder cases). Please add a case where an existing folder's persisted visibility differs from `inheritedFrom(parent)` and assert the next sync corrects it — right now this could regress silently.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationItemValidatorService.java
```diff
@@ -141,11 +143,51 @@ public void validate(List<CreatePortalNavigationItem> creates, List<PendingUpdat
...
+        var descendantUpdates = creates
+            .stream()
+            .filter(PortalNavigationItemValidatorService::isPendingContainer)
+            .flatMap(container -> descendantsOf(container, environmentId).stream())
+            .map(PortalNavigationItemValidatorService::asUnchangedUpdate);
+        return Stream.concat(updates.stream(), descendantUpdates).toList();
+    }
+
+    private static boolean isPendingContainer(CreatePortalNavigationItem item) {
+        return item.getId() != null && item.getType().isContainer();
```
dev-marek, line 188: Improvement — this only scans `creates` (via `isPendingContainer`), not `updates`. Both `CreatePortalNavigationItem` and `UpdatePortalNavigationItem` expose a caller-settable `visibility`, so a batch that updates an *existing* FOLDER's visibility from PUBLIC to PRIVATE (with pre-existing descendants outside that batch) isn't re-validated here — same silent-mismatch class as the phantom-create case you just fixed, just on the update-an-existing-container path. Worth extending to scan `updates` for containers whose visibility is changing, or tracking as a documented follow-up.

dev-marek (changes requested):
Re-reviewed after `ff05ad30`. Two of the three original findings are resolved; the third is only partially closed, and I found one more gap in the same area while verifying the fix.

**Resolved**
- Phantom-PUBLIC placeholder (`PortalNavigationItemContainer.java:50`): closed via a different route than I suggested. `withDescendantsOfPendingContainers` now re-feeds pre-existing descendants of a newly-created container through `ParentRule`, so creating a real PRIVATE folder over pre-existing PUBLIC descendants now throws instead of silently persisting the mismatch. Covered by `CreatePortalNavigationItemValidatorServiceTest.PhantomThenPrivateParent` (3 cases).
- PR description now correctly scopes the follow-up to automation-api rather than REST API.

**Partially resolved — test coverage**
PRIVATE-parent inheritance tests were added for `PortalLinkSyncDomainServiceTest`, `PortalDocumentationSyncDomainServiceTest`, and `ApiDocumentationSyncDomainServiceTest` — good, verified. `NavigationSyncPlanExecutorTest`, `PortalNavigationSyncDomainServiceTest`, and `ParentRuleTest` — the other three I named — still only exercise the default/PUBLIC case. See the inline comment on `NavigationFolderMapper.java` for why this one still blocks.

Separately (unrelated to this diff): left a note on https://github.com/gravitee-io/gravitee-api-management/pull/19381#discussion_r3902460485 — a non-blocking `SegmentConflictRule` finding from #19381 that's still open. Flagging here since it's the same validator area and you're the author on both.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal/domain_service/navigation/NavigationFolderMapper.java
konrad-krol-incubly: Added tests exercising re-derivation on next sync, at both the executor and sync-service levels.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationItemValidatorService.java
konrad-krol-incubly: Fixed inline — augmentation now also emits descendants for updates that change a container's visibility, and `ParentRule` update-side uses the pending update's new visibility for the parent-side check. Test coverage added.

konrad-krol-incubly:
> Partially resolved — test coverage
PRIVATE-parent inheritance tests were added for PortalLinkSyncDomainServiceTest, PortalDocumentationSyncDomainServiceTest, and ApiDocumentationSyncDomainServiceTest — good, verified. NavigationSyncPlanExecutorTest, PortalNavigationSyncDomainServiceTest, and ParentRuleTest — the other three I named — still only exercise the default/PUBLIC case. See the inline comment on NavigationFolderMapper.java for why this one still blocks.

Added PRIVATE-parent coverage to the three test classes you named.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationItemValidatorService.java
```diff
@@ -130,22 +130,98 @@ public void validate(List<CreatePortalNavigationItem> creates, List<PendingUpdat
...
+    // Include pre-existing descendants of pending container creates and of pending container updates that change visibility.
+    private List<PendingUpdate> withDescendantsOfPendingContainers(
+        List<CreatePortalNavigationItem> creates,
+        List<PendingUpdate> updates,
+        String environmentId
+    ) {
+        var descendantsOfCreates = creates
+            .stream()
+            .filter(PortalNavigationItemValidatorService::isPendingContainer)
+            .flatMap(container -> descendantsOf(container.getId(), environmentId).stream());
```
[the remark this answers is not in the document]
konrad-krol-incubly: This matches the per-item pattern used across the codebase — `UniqueItemIdRule`, `HomepageUniquenessRule`, `SegmentConflictRule`, and `ParentRule` all query per item during validation. Automation batch sizes here are small (folders + sync-materialized rows for a single listing / portal), so the overhead is negligible in practice. Not planning to change it in this PR; a batch-load path would be worth it if we ever see this hit at scale.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalNavigationItemValidatorService.java
```diff
@@ -130,22 +130,98 @@ public void validate(List<CreatePortalNavigationItem> creates, List<PendingUpdat
...
+    private static boolean isContainerVisibilityChange(PendingUpdate update) {
+        var existing = update.existing();
+        var toUpdate = update.toUpdate();
+        return (
+            existing != null &&
+            existing.getId() != null &&
+            existing.getType().isContainer() &&
+            toUpdate.getVisibility() != null &&
+            toUpdate.getVisibility() != existing.getVisibility()
+        );
```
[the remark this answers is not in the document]
konrad-krol-incubly: Left as-is. The predicate reads as a single "all must hold" statement, and two of the guard forms the suggestion would produce (`existing != null`, `existing.getId() != null`) are defensive against situations that don't happen in practice — every other code path assumes `PendingUpdate.existing()` is non-null and persisted items have ids. Refactoring for structure while keeping dead defensive checks would trade one form of noise for another. If it becomes hard to read later, will tighten then.

dev-marek (approved):

PR #19506, opened 2026-08-31
feat(logs): let the Kafka client library version be filtered

ThibaudAV:
Closes the gap the library filter left open on [ESM-2](https://gravitee.atlassian.net/browse/ESM-2).

## The gap

#19425 answers *"who is on librdkafka"*. It cannot answer *"who is on librdkafka 1.8"* — every version collapses into the same bucket.

Until now the version existed in two places, neither of which lets you act on it:

| Where | What you can do |
|---|---|
| Log detail | read it, one connection at a time |
| `Client library` table column | see it, not select on it |

During an incident — the case where a specific version is the suspect — reading connections one by one is not an option.

## On my own earlier reasoning

I closed this door when the library filter was added, writing that *"comparing versions is a range over an unordered keyword"*. That is true of a **comparison**: `< 2.0` means nothing on a keyword. It does not justify withholding **equality**, which is exactly as meaningful on a version as it is on a library name. The argument was sound for a case that was not the one at hand.

## LOGS-only, and why that is the point

This is the one place it differs from its sibling, which is `LOGS + ANALYTICS`.

Several clients share version numbers, so a version **detached from its library identifies nothing** — grouping by it alone produces buckets labelled `1.8.2` with no way to know of what. It narrows a library rather than standing on its own.

That is also why neither an analytics dimension nor a composite `by: [name, version]` facet is the answer: the lib's own `QUERYING.md` says bar charts flatten multi-dimension groupings into composite categories and to prefer a single dimension. Combined with the library filter, two conditions answer the question directly and nothing needs to be charted.

## Scope

The field is already indexed and already aggregated, so this is the same wiring as the library filter — no gateway change, no migration, and it applies to **history** as much as to new traffic.

## Tests

`repository-elasticsearch` **888** · `rest-api-service` **9552** · `management-v2-rest` **1856** · `gamma-rest-api` **394**, plus a full backend build.

Two new cases pin the translation at both ends — the Gamma adapter and the Elasticsearch clause. Verified the ES one by removing the clause registration: the filter would then be accepted and silently dropped from the query, which is the failure mode this chain has produced twice before.

Three snapshots moved on purpose: the catalog gains an entry (59 → 60), the LOGS+NATIVE list gains a name, and `AnalyticsQueryValidator` withholds one more filter from analytics.

[ESM-2]: https://gravitee.atlassian.net/browse/ESM-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/main/java/io/gravitee/repository/elasticsearch/v4/log/adapter/connection/SearchMetricsQueryAdapter.java
```diff
@@ -292,6 +295,21 @@ private static void addNativeClientSoftwareNamesFilter(MetricsQuery.Filter filte
         }
     }
 
+    /**
```
callaertanthony, line 297: The javadoc on `addNativeClientSoftwareNamesFilter` just above (lines 285-289) still states:

> KIP-511 client library name. The version is deliberately not filterable: comparing versions is a range over an unordered string, and the fleet question ("who is still on librdkafka?") is about the library.

The method added right below now makes the version filterable, so that comment is false as of this change. Worth updating in the same commit — a reader who trusts it will remove this filter as an oversight. The distinction the PR description draws (equality on a version is meaningful; ordering is not) is exactly what belongs there instead.

gravitee-gamma/gravitee-gamma-rest-api/src/main/java/io/gravitee/gamma/rest/core/observability/filter/model/StaticFilters.java
```diff
@@ -215,6 +215,21 @@ public enum StaticFilters {
         Defs.LOGS_ANALYTICS,
         Set.of(ApiType.NATIVE)
     ),
+    /**
```
callaertanthony, line 218: Same stale rationale as in `SearchMetricsQueryAdapter`: the javadoc on `NATIVE_CLIENT_SOFTWARE_NAME` immediately above (lines 197-198) reads

> the version is not a filter of its own because comparing versions is a range over an unordered keyword

and it now sits directly above the constant that makes the version a filter of its own. Please update it so the two blocks do not contradict each other.

callaertanthony (commented):
Two stale comments carry the old rationale that this PR reverses. Everything else looks good to me: the wiring matches `NATIVE_CLIENT_SOFTWARE_NAME` everywhere except the analytics-engine facet path (correctly, given the LOGS-only scope), and `ObservabilityLogsFilterCoverageTest` / `AnalyticsVocabularyMappingTest` already cover the new catalog entry end-to-end without needing a new test.

gravitee-gamma/gravitee-gamma-rest-api/src/test/java/io/gravitee/gamma/rest/core/observability/filter/SpiFilterRegistryTest.java
```diff
@@ -257,6 +257,7 @@ void should_apply_both_axes_to_host_and_contributor_filters() {
                 "FAILURE_ORIGIN",
                 "NATIVE_CLIENT_ID",
                 "NATIVE_CLIENT_SOFTWARE_NAME",
+                "NATIVE_CLIENT_SOFTWARE_VERSION",
```
callaertanthony, line 263: Nit: this pins the version on Gamma's LOGS axis, but nothing pins its absence from the ANALYTICS one. `should_advertise_the_kafka_client_dimensions_on_the_analytics_signal` (line 224) uses `.contains(...)`, so switching this constant to `Defs.LOGS_ANALYTICS` in `StaticFilters` would break no test here — and Gamma would then advertise a filter the core `AnalyticsQueryValidator` rejects with a 400.

That is the mirror image of the divergence the `NATIVE_CLIENT_SOFTWARE_NAME` javadoc explains at length, so it seems worth guarding the same way. One line on the existing test does it:

```java
.doesNotContain("FAILURE_ORIGIN", "NATIVE_CLIENT_SOFTWARE_VERSION");
```

gravitee-apim-repository/gravitee-apim-repository-elasticsearch/src/test/java/io/gravitee/repository/elasticsearch/v4/log/adapter/connection/SearchMetricsQueryAdapterTest.java
```diff
@@ -1069,6 +1071,15 @@ void should_add_kafka_client_software_name_terms_filter_on_additional_metrics()
...
+        void should_add_kafka_client_software_version_terms_filter_on_additional_metrics() {
+            var query = MetricsQuery.builder()
+                .filter(MetricsQuery.Filter.builder().nativeClientSoftwareVersions(Set.of("1.8.2")).build())
+                .build();
+
+            assertThat(hasTermsOn(query, SOFTWARE_VERSION_FIELD)).isTrue();
+        }
+
         @Test
         void should_not_add_native_client_filters_when_null_or_empty() {
```
callaertanthony, line 1084: Nit: this guard still covers only `nativeClientIds` and `nativeClientSoftwareNames`, so the new filter is the one native client filter whose null/empty case is unasserted despite the test name. The code path is the same `CollectionUtils.isEmpty` check, so this is about keeping the test honest to its name rather than a suspected bug — adding `.nativeClientSoftwareVersions(null)` / `(Set.of())` to the two builders and one assertion per query closes it.

callaertanthony (commented):
Two test nits, both optional — neither blocks.

ThibaudAV:
All four taken. The two stale comments were the ones that mattered, and they were mine.

## The stale rationale — twice

Both javadocs still argued that the version is deliberately not filterable, sitting directly above the code that makes it one. That is worse than out of date: as you say, a reader who trusts it removes this filter as an oversight. I wrote the original claim and then contradicted it two files away without going back.

Rewritten in both places on the distinction that actually holds: **ordering** a version keyword is meaningless, **selecting** one is not. The Gamma one now points at the version constant rather than denying its existence.

## The ANALYTICS axis was unguarded — good catch

Right, and it is the mirror image of the divergence the sibling javadoc spends a paragraph explaining. `.contains(...)` could not have caught a flip to `Defs.LOGS_ANALYTICS`, and Gamma would then have advertised a filter the core validator answers with a 400.

Added your one-liner, with the reason on both names — `FAILURE_ORIGIN` is derived rather than stored, `NATIVE_CLIENT_SOFTWARE_VERSION` is stored but not groupable alone. Verified by flipping the constant to `LOGS_ANALYTICS`: the suite fails.

## The null/empty guard

Fair — the test named itself after all the native client filters while covering two of three. Both builders and both assertions extended. Same code path, but a test that does not check what its name claims is the kind of thing that gets trusted later.

`repository-elasticsearch` **888** · `gamma-rest-api` **394**, plus a full backend build. Squashed into the single commit.

callaertanthony (approved):
All four comments addressed on `ba48b75` — threads resolved.

Both javadoc rewrites keep the correct half of the old reasoning rather than just deleting it: "ordering a version string is meaningless on a keyword, but selecting one is not" is the distinction this PR turns on, and it now lives in the code instead of only in the description. `git grep "not filterable|not a filter of its own"` at the head returns nothing, so no stale copy is left behind.

What I verified on the change itself:

- **No missed wiring.** Diffing the set of files that mention `*_SOFTWARE_NAME` against `*_SOFTWARE_VERSION` at the head leaves exactly six: `Facet`, `Filter`, `FilterAdapter`, `NativeApiFieldResolver` (+ its test) and `FacetSpec` — all on the analytics-engine facet path, which is precisely the LOGS-only boundary this PR argues for. Console-webui has zero occurrences of either, so the catalog drives the UI and no front-end change is owed.
- **The values dropdown actually resolves.** `ObservabilityFilterValuesResource` takes `filterName` as a raw `String` path param, so there is no enum barrier; `GetFilterValuesUseCase` routes `KEYWORD` to `FilterValuesQueryServiceImpl.resolveEsFieldName`, where the new case was added. Same shape as the other LOGS-only keyword filters (`ERROR_KEY`, `REQUEST_ID`, `TRANSACTION_ID`) that already sit in that switch.
- **Both `THROW_EXCEPTION` value mappings are load-bearing, not defensive.** The target enums (`FilterName`, repository `Filter.Name`) have no such constant, so MapStruct would fail the build without them — and `AnalyticsVocabularyMappingTest` derives the expected refusal set from the catalog, so the pair cannot drift.
- **`LogsEngineMapper.mapFilterName`** aliases only the three historical spellings and falls through to `valueOf`, and both wire and engine enums carry the new name — the console's filter reaches the engine unaliased.
- **End-to-end coverage came for free.** `ObservabilityLogsFilterCoverageTest` loads the real `analytics-definition.yaml` and asserts every `signals: [LOGS]` entry reaches `SearchLogsFilters`; green CI there is what proves `FilterName`, `FilterContext`, `updateFilterIds` and `ConnectionLogsCrudServiceImpl` are correctly wired.

One thing left deliberately alone, noted only so it is not mistaken for an oversight: no `NATIVE_*` filter in `analytics-definition.yaml` carries an `apis:` restriction, although the mechanism exists and `API_PRODUCT` uses it — so these are offered on non-native APIs, while Gamma's `StaticFilters` does scope them to `ApiType.NATIVE`. That divergence predates this PR across all five native filters; this one follows the local convention, which is the right call for this change.

LGTM.

PR #19508, opened 2026-08-31
feat(gamma): added env notification list, create, delete

arpzilla-dev:
## Issue

https://gravitee.atlassian.net/browse/FOUND-60
https://gravitee.atlassian.net/browse/FOUND-61
https://gravitee.atlassian.net/browse/FOUND-62

## Description

Adds environment notification settings to the Gamma Console (Environment → System & Security), using the existing Management API v1 — no backend changes.

Operators can list configured notifications (Console, Email, Webhook), create GENERIC email/webhook subscriptions, edit them, and delete GENERIC rows. The Console Notification (PORTAL) row can be edited but not deleted. Create/update/delete follow backend permissions (`environment-notification` C/R/U/D), including PORTAL edit with read only.

Detailed report with scenario tested is available here : https://gravitee.atlassian.net/browse/FOUND-251

## Additional context

https://github.com/user-attachments/assets/cb5d53fe-75fe-487f-ac0d-9f4b71ffb50d


gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/environment-notifications/hooks/useEnvironmentNotifications.ts
ankita-gupta21, line 1: **Score: 55/100] Duplicate logic — same PORTAL/GENERIC suffix trick, same PUT call shape, as updateApplicationNotification**

`updateApplicationNotification` ([applicationNotifications.ts#L59-L74](https://github.com/gravitee-io/gravitee-api-management/blob/feat/FOUND-61-environment-notifications/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/services/applicationNotifications.ts#L59-L74)) and `updateEnvironmentNotification` (this file, L54-L63) are identical except for the URL prefix (`/applications/{id}` vs `/configuration`). Same is true of create*/delete*/list* in both files.

Risk: `the config_type === 'PORTAL' ? '/' : /${id} `suffix rule now lives in two files. A future fix to one (e.g. the trailing-slash contract changes) is easy to miss in the other, since they're in unrelated folders.

Suggest extracting a shared factory instead of two hand-maintained copies — see suggested fix below.
New file: `gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/shared/services/notificationSettingsClient.ts`

```
import { apimFetchJsonV1Env } from '../api/apimClient';
import type {
    ApplicationNotificationHook,
    ApplicationNotificationSettings,
    ApplicationNotifier,
} from '../../features/applications/types/applicationNotification';

const JSON_HEADERS = { 'Content-Type': 'application/json' };

function updateSuffix(notification: { config_type: string; id?: string }) {
    return notification.config_type === 'PORTAL' ? '/' : `/${encodeURIComponent(notification.id ?? '')}`;
}

/** `basePath` is the only thing that differs between application-scoped and environment-scoped notifications. */
export function notificationSettingsClient(basePath: string) {
    return {
        list: (envId: string) => apimFetchJsonV1Env<ApplicationNotificationSettings[]>(envId, `${basePath}/notificationsettings`),
        listNotifiers: (envId: string) => apimFetchJsonV1Env<ApplicationNotifier[]>(envId, `${basePath}/notifiers`),
        listHooks: (envId: string, hooksPath: string) => apimFetchJsonV1Env<ApplicationNotificationHook[]>(envId, hooksPath),
        create: <T>(envId: string, notification: T) =>
            apimFetchJsonV1Env<ApplicationNotificationSettings>(envId, `${basePath}/notificationsettings`, {
                method: 'POST',
                headers: JSON_HEADERS,
                body: JSON.stringify(notification),
            }),
        update: (envId: string, notification: ApplicationNotificationSettings) =>
            apimFetchJsonV1Env<ApplicationNotificationSettings>(
                envId,
                `${basePath}/notificationsettings${updateSuffix(notification)}`,
                { method: 'PUT', headers: JSON_HEADERS, body: JSON.stringify(notification) },
            ),
        delete: (envId: string, notificationId: string) =>
            apimFetchJsonV1Env<void>(envId, `${basePath}/notificationsettings/${encodeURIComponent(notificationId)}`, {
                method: 'DELETE',
            }),
    };
}

```

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/environment-notifications/utils/notificationChannel.ts
```diff
@@ -0,0 +1,63 @@
...
+        return 'WEBHOOK';
+    }
+    if (type === 'EMAIL') {
+        return 'EMAIL';
+    }
+    const notifierId = row.notification.notifier ?? '';
+    if (notifierId.includes('webhook')) {
+        return 'WEBHOOK';
+    }
+    return 'EMAIL';
```
ankita-gupta21: if `row.notifier `is momentarily `undefined` (notifiers query still loading, or a notifier was later removed but a notification still references it) and the notifier id doesn't literally contain the substring `"webhook"` — e.g. a custom/plugin notifier id like `"slack-default"` — this silently labels it "Email" in the Channel badge, which is operator-facing and just wrong.

**Sample fix:** default to a neutral/"Unknown" badge instead of guessing:
`return type === 'WEBHOOK' || notifierId.includes('webhook') ? 'WEBHOOK' : type === 'EMAIL' ? 'EMAIL' : 'UNKNOWN';`

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/environment-notifications/components/EnvironmentNotificationsTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+}: Readonly<{
+    rows: ApplicationNotificationRow[];
+    isLoading: boolean;
+    isError: boolean;
+    canUpdate: (row: ApplicationNotificationRow) => boolean;
+    canDelete: boolean;
+    onEdit: (row: ApplicationNotificationRow) => void;
+    onDelete: (row: ApplicationNotificationRow) => void;
+}>) {
+    const hasActionsColumn = rows.some(row => canUpdate(row) || (canDelete && canDeleteNotificationRow(row)));
```
ankita-gupta21: for a Kubernetes-origin (`origin !== 'MANAGEMENT'`) environment notification the caller can otherwise update, the header still renders an "Actions" column, but that row's own cell renders nothing — a column that exists for no row that needs it. Harmless today, but it's a latent inconsistency the next person touching this file will trip on.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/environment-notifications/components/EnvironmentNotificationsTable.tsx
ankita-gupta21, line 1: Reinvents Table/loading-state/empty-state instead of using DataTable

`NotificationsSection.tsx` renders the same ApplicationNotificationRow[] as a table one line down: <DataTable columns={...} data={rows} emptyMessage="No notifications configured." /> ([NotificationsSection.tsx#L155](https://github.com/gravitee-io/gravitee-api-management/blob/feat/FOUND-61-environment-notifications/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/applications/components/notifications/NotificationsSection.tsx#L155)). This file instead imports raw Table/TableRow/TableCell and hand-writes a LoadingRows skeleton generator (L71-93) — and as a result has no empty state at all if rows is [].

DataTable is compatible with a dropdown-menu actions column if you want to keep the three-dot menu instead of the sibling's inline icon buttons — switching doesn't force a UX change, just removes ~40 lines of reimplemented plumbing.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/environment-notifications/hooks/useEnvironmentNotifications.ts
arpzilla-dev: Good catch — the PORTAL `/` vs GENERIC `/{id}` rule was the part that actually needed to live in one place.

I extracted `notificationSettingsUpdatePath()` and wired both the environment and application clients through it. GENERIC updates with no id now throw instead of falling through to the PORTAL path.

I didn’t pull in the full `notificationSettingsClient` factory. Application calls also take `applicationId`, hooks live on different paths (`/configuration/hooks` vs `/applications/hooks`), and the application service also owns metadata. Sharing the suffix rule covers the drift risk without coupling those APIs.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/environment-notifications/utils/notificationChannel.ts
arpzilla-dev: Fixed. Unknown or missing notifiers now resolve to an Unknown badge instead of Email.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/environment-notifications/components/EnvironmentNotificationsTable.tsx
arpzilla-dev: Fixed. The Actions column is shown only when at least one row actually has Edit or Delete (`canUpdate && !isReadonly`, or deletable GENERIC). A table of Kubernetes-origin readonly rows no longer gets an empty Actions column.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/environment-notifications/components/EnvironmentNotificationsTable.tsx
arpzilla-dev: Switched to DataTable, same as application notifications. Empty state is now “No notifications configured.” Kept the three-dot menu.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/EnvironmentNotificationSettingsPage.tsx
```diff
@@ -0,0 +1,184 @@
...
+            const created = await createNotificationMutation.mutateAsync({
+                name: payload.name,
+                notifier: payload.notifier,
+                referenceType: 'ENVIRONMENT',
+                referenceId: envId,
+                config_type: 'GENERIC',
+                hooks: payload.hooks,
+            });
+
+            const needsFollowUpUpdate = Boolean(payload.config) || Boolean(payload.useSystemProxy);
```
arpzilla-dev: @AikidoSec ignore: EditNotificationSheet always sends useSystemProxy as a boolean (default false). POST already omits it, so the server default is false. Follow-up PUT is only needed for a config URL or useSystemProxy: true. Treating !== undefined as “provided” would issue a second PUT on every create. Same pattern as application notification create.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/EnvironmentNotificationSettingsPage.tsx
```diff
@@ -0,0 +1,184 @@
...
+    useDeleteEnvironmentNotification,
+    useEnvironmentNotifications,
+    useUpdateEnvironmentNotification,
+} from '../features/environment-notifications/hooks/useEnvironmentNotifications';
+import { buildNewEnvironmentNotificationRow } from '../features/environment-notifications/utils/buildNewEnvironmentNotificationRow';
+import { useForbiddenResourceRedirect } from '../shared/hooks/useForbiddenResourceRedirect';
+import { notify } from '../shared/notify';
+import { isForbiddenApiError } from '../shared/utils/apiErrors';
+
+export function EnvironmentNotificationSettingsPage() {
```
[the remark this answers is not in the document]
arpzilla-dev: @AikidoSec ignore: This page coordinates list/create/edit/delete, permissions, and the sheet/dialog. Splitting it would move the same handlers around without changing behavior. Mutation logic already lives in the hooks; the table is its own component.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/EnvironmentNotificationSettingsPage.tsx
```diff
@@ -0,0 +1,184 @@
...
+                    hooks: payload.hooks,
+                });
+            }
+
+            notify.success('Notification created successfully');
+            setNotificationToEdit(null);
+        } catch (err: unknown) {
+            notify.error(err, 'Failed to create notification.');
+        }
+    }
```
ankita-gupta21, line 118: **Where:** [pages/EnvironmentNotificationSettingsPage.tsx#L74-L103](https://github.com/gravitee-io/gravitee-api-management/blob/feat/FOUND-61-environment-notifications/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/EnvironmentNotificationSettingsPage.tsx#L74-L103) vs. [pages/ApplicationNotificationSettingsPage.tsx#L91-L120](https://github.com/gravitee-io/gravitee-api-management/blob/feat/FOUND-61-environment-notifications/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/ApplicationNotificationSettingsPage.tsx#L91-L120).

low risk today. If a third notification-settings screen shows up later (API-scoped notifications, say), you'd be looking at a third copy of this exact block — that's the point where it stops being "acceptable duplication" and starts being a real maintenance tax.

Sample fix — `features/shared/hooks/useNotificationCrudHandlers.ts (new):`

```
export function useNotificationCrudHandlers({
    createMutation, updateMutation, deleteMutation, onEditDone, onDeleteDone,
}: { /* ...mutation objects... */ }) {
    async function handleCreate(basePayload: object, payload: NotificationSheetCreatePayload) { /* same body */ }
    function handleUpdate(notification: UpdateApplicationNotification) { /* same body */ }
    function handleDeleteConfirm(target: ApplicationNotificationRow | null) { /* same body */ }
    return { handleCreate, handleUpdate, handleDeleteConfirm };
}
```

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/environment-notifications/hooks/useEnvironmentNotificationPermissions.ts
```diff
@@ -0,0 +1,32 @@
...
+ * `canUpdatePortal` is intentionally derived from `-r`, not `-u`: Classic's
+ * `NotificationConfigsResource#updatePortalNotificationSettings` only requires
+ * `ENVIRONMENT_NOTIFICATION` READ, since the Console/Portal row is the caller's own
+ * notification preference, not a shared GENERIC config. `canUpdateGeneric` requires `-u`,
+ * matching `updateGenericNotificationSettings`.
+ */
+export function useEnvironmentNotificationPermissions() {
+    const canCreate = useHasPermission({ anyOf: ['environment-notification-c'] });
+    const canUpdateGeneric = useHasPermission({ anyOf: ['environment-notification-u'] });
+    const canUpdatePortal = useHasPermission({ anyOf: ['environment-notification-r'] });
```
ankita-gupta21, line 28: `canUpdatePortal` derived from `-r `alone is backend-accurate but is a real behavior change vs. Classic Console, which gates the same button behind `-u` for both rows — confirm this permission-model widening for the Console/Portal row is an intended fix, not incidental.

Just need some clarification here

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/EnvironmentNotificationSettingsPage.tsx
```diff
@@ -0,0 +1,184 @@
...
+                    useSystemProxy: payload.useSystemProxy ?? created.useSystemProxy,
+                    hooks: payload.hooks,
+                });
+            }
+
+            notify.success('Notification created successfully');
+            setNotificationToEdit(null);
+        } catch (err: unknown) {
+            notify.error(err, 'Failed to create notification.');
+        }
```
ankita-gupta21: **### OPEN FOR PERSPECTIVE:**

What might go bad, concretely: user fills the create sheet with a webhook URL and clicks "Add notification." The POST (line 79) succeeds — a GENERIC notification now exists server-side, without the URL, since CreateEnvironmentNotification has no config field. The follow-up PUT (line 90) that would attach the URL then fails (network blip, validation error). The catch at line 100 fires notify.error(err, 'Failed to create notification.') — the user reads that as "nothing happened" and, reasonably, clicks "Add notification" again. They now have two GENERIC notifications with the same name, one silently missing its config, and no toast ever told them the first one exists.

Sample fix (same file, handleCreateNotification):
```
async function handleCreateNotification(payload: NotificationSheetCreatePayload) {
    if (!envId) return;
    let created: ApplicationNotificationSettings | undefined;
    try {
        created = await createNotificationMutation.mutateAsync({ /* ...unchanged... */ });
    } catch (err: unknown) {
        notify.error(err, 'Failed to create notification.');
        return;
    }
    const needsFollowUpUpdate = Boolean(payload.config) || Boolean(payload.useSystemProxy);
    if (needsFollowUpUpdate) {
        try {
            await updateNotificationMutation.mutateAsync({ ...created, config: payload.config ?? created.config, useSystemProxy: payload.useSystemProxy ?? created.useSystemProxy, hooks: payload.hooks });
        } catch (err: unknown) {
            notify.error(err, `"${created.name}" was created, but saving its configuration failed — edit it to finish setup.`);
            setNotificationToEdit(null); // sheet was for creating; the row now exists, let the table refresh show it
            return;
        }
    }
    notify.success('Notification created successfully');
    setNotificationToEdit(null);
}
```
This isn't something I'd ask you to fix in this PR alone, since it's copied from working code elsewhere — but if you do fix it here, the applications page has the identical bug and should get the same fix in the same change.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/EnvironmentNotificationSettingsPage.tsx
arpzilla-dev: Agreed it’s the same shape, but the payloads differ (referenceType / referenceId) and we only have two screens. I’d rather keep the handlers on each page until a third notification-settings surface shows up, instead of a shared hook that has to special-case env vs application.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/features/environment-notifications/hooks/useEnvironmentNotificationPermissions.ts
arpzilla-dev: Intended. Classic UI gates every row’s Edit on `environment-notification-u`, but the Management API only requires READ for the PORTAL/Console row (`updatePortalNotificationSettings`). GENERIC still needs `-u`. Gamma follows the API so a user with read (and not update) can edit their own Console Notification. Happy to tighten it to `-u` if we want Classic UI parity instead of backend parity.

gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/EnvironmentNotificationSettingsPage.tsx
arpzilla-dev: Fixed on both pages. POST failure still toasts “Failed to create notification.” If POST succeeds and the follow-up config PUT fails, the toast now says the row was created and needs a manual edit to finish setup, and the create sheet closes so the table can show the new row.

ankita-gupta21:
1. [Nitpick] useCallback still wraps handlers with no memoized consumer
Where: [EnvironmentNotificationSettingsPage.tsx#L66-L69](https://github.com/gravitee-io/gravitee-api-management/blob/feat/FOUND-61-environment-notifications/gravitee-gamma/gravitee-gamma-module-platform/src/main/ui/pages/EnvironmentNotificationSettingsPage.tsx#L66-L69) (canEditRow) and #L71-L76 (handleCreateNotificationClick) — unchanged from last review.

arpzilla-dev:
Removed both `useCallback` wrappers. `canEditRow` and `handleCreateNotificationClick` only feed the page render / the Add button, so the memo did not buy anything.

Also merged `master` — notification settings now goes through the shared `NAV_ITEM_PERMISSIONS` / `NavPermissionGuard` path.

ankita-gupta21 (approved):

PR #19511, opened 2026-09-01
fix(gateway): restore v3 template context before propagating the terminal event

thomasballivet:
## Issue

https://gravitee.atlassian.net/browse/APIM-14749

## Description

When a v4 API chains two v3 policies that write the same EL variable (reported with two
`dynamic-routing` policies, which both publish their regex captures under `group`), the second
policy intermittently evaluates its URL template with the first policy's captures, producing a
corrupted upstream path (`/v1//v1/hello`) and a backend 404.

All v3 policies of a request share one `TemplateContextAdapter`, and `PolicyAdapter` restored its
backed-up variables in `doFinally` — which runs *after* the terminal event reaches the downstream.
Since the worker offload (PEN-88, present from 4.10.x), the next policy is by then already running on
another thread, so the late `restore()` races with it. The fix restores in `doOnTerminate` (before the
terminal event propagates), keeping `doOnDispose` for cancellation — same pattern as
`InvokerAdapter` (728a6315cf). `restore()` keeps the at-most-once property `doFinally` gave it, in `PolicyAdapter` and `InvokerAdapter`
alike. Covered by five new `PolicyAdapterTest` tests, red before the fix.

## Additional context

Gateway only; `dynamic-routing` is a victim of the shared context, not the cause. Validated on a
real gateway under load (interleaved rounds, 60 000 requests per variant): without the fix
4 corrupted paths; with the fix, or with the offload disabled (`gateway.policy.v3.timeoutMs=0`,
which isolates the cause), 0.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-policy/src/main/java/io/gravitee/gateway/reactive/policy/adapter/policy/PolicyAdapter.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-policy/src/main/java/io/gravitee/gateway/reactive/policy/adapter/policy/PolicyAdapter.java
@@ -25,6 +25,7 @@ import io.gravitee.gateway.reactive.policy.adapter.context.ExecutionContextAdapt
 import io.reactivex.rxjava3.core.Completable;
 import io.reactivex.rxjava3.core.Maybe;
 import io.reactivex.rxjava3.core.Scheduler;
+import io.reactivex.rxjava3.functions.Action;
 import io.reactivex.rxjava3.schedulers.Schedulers;
 import java.util.concurrent.atomic.AtomicBoolean;
 
@@ -103,7 +104,23 @@ public class PolicyAdapter implements HttpPolicy {
             completable = completable.andThen(this.policyStream(adaptedCtx, phase));
         }
 
-        return completable.doFinally(adaptedCtx::restore);
+        // APIM-14749: restore must land BEFORE the terminal event reaches the downstream, so this is deliberately not a
+        // doFinally. Every v3 policy of the chain shares this ExecutionContextAdapter, and doFinally runs after the next
+        // policy has been subscribed - and, since the PEN-88 worker offload, is already running on another thread -
+        // so a late restore would replay the previous policy's backup over the variables of the next one.
+        // doOnTerminate runs before the terminal event is propagated; doOnDispose covers the cancelled request
+        // (client abort, timeout). Unlike doFinally, the two legs do not guard each other, hence the at-most-once
+        // guard: a dispose arriving after termination must not restore a second time, for the same reason as above.
+        // Consequence, shared with InvokerAdapter: an exception thrown by restore() fails the request instead of
+        // being reported to RxJavaPlugins.onError as it was under doFinally.
+        final AtomicBoolean restored = new AtomicBoolean();
+        final Action restoreOnce = () -> {
+            if (restored.compareAndSet(false, true)) {
+                adaptedCtx.restore();
+            }
+        };
+
+        return completable.doOnTerminate(restoreOnce).doOnDispose(restoreOnce);
     }
 
     private Completable policyExecute(ExecutionContextAdapter adaptedCtx) {
--- a/gravitee-apim-gateway/gravitee-apim-gateway-policy/src/test/java/io/gravitee/gateway/reactive/policy/adapter/policy/PolicyAdapterTest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-policy/src/test/java/io/gravitee/gateway/reactive/policy/adapter/policy/PolicyAdapterTest.java
@@ -17,13 +17,18 @@ package io.gravitee.gateway.reactive.policy.adapter.policy;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.argThat;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import io.gravitee.el.TemplateContext;
+import io.gravitee.el.TemplateEngine;
 import io.gravitee.gateway.api.ExecutionContext;
 import io.gravitee.gateway.api.Request;
 import io.gravitee.gateway.api.Response;
@@ -39,13 +44,20 @@ import io.gravitee.gateway.reactive.api.context.http.HttpPlainRequest;
 import io.gravitee.gateway.reactive.api.context.http.HttpPlainResponse;
 import io.gravitee.gateway.reactive.policy.adapter.context.ExecutionContextAdapter;
 import io.reactivex.rxjava3.core.Maybe;
+import io.reactivex.rxjava3.disposables.Disposable;
 import io.reactivex.rxjava3.observers.TestObserver;
 import io.reactivex.rxjava3.schedulers.Schedulers;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import org.junit.jupiter.api.Test;
 import org.mockito.ArgumentCaptor;
+import org.mockito.invocation.InvocationOnMock;
 
 /**
  * @author Jeoffrey HAEYAERT (jeoffrey.haeyaert at graviteesource.com)
@@ -54,6 +66,16 @@ import org.mockito.ArgumentCaptor;
 class PolicyAdapterTest {
 
     protected static final String MOCK_EXCEPTION_MESSAGE = "Mock exception";
+
+    /**
+     * Best-effort head start given to the next policy before the previous one replays its backup, to reproduce the
+     * production timing where the restore runs on the event loop while the next policy already runs on a worker thread.
+     * With the fix applied this wait always times out: the restore lands before the next policy is even subscribed.
+     */
+    private static final long RESTORE_HEAD_START_MS = 200L;
+
+    /** How long the next policy waits for the restore of the previous one to land before evaluating its template. */
+    private static final long RESTORE_AWAIT_TIMEOUT_MS = 5_000L;
 
     @Test
     public void shouldCompleteInErrorWhenOnMessageRequest() {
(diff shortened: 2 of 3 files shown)


gravitee-apim-gateway/gravitee-apim-gateway-policy/src/main/java/io/gravitee/gateway/reactive/policy/adapter/policy/PolicyAdapter.java
```diff
@@ -103,7 +103,7 @@ private Completable execute(final HttpPlainExecutionContext ctx, final Execution
             completable = completable.andThen(this.policyStream(adaptedCtx, phase));
         }
 
-        return completable.doFinally(adaptedCtx::restore);
+        return completable.doOnTerminate(adaptedCtx::restore).doOnDispose(adaptedCtx::restore);
```
phiz71: `doFinally` guaranteed the action ran at most once — `CompletableDoFinally.runFinally()` guards it with `compareAndSet(0, 1)`. Two separate peek operators do not: `CompletablePeek.dispose()` runs `onDispose.run()` unconditionally, with no check that `onComplete`/`onError` already fired. A dispose arriving after termination therefore calls `restore()` a second time.

I looked for a live path and did not find one: `FlowableConcatMapCompletable` swaps the inner disposable with `DisposableHelper.replace`, so a terminated policy is not re-disposed once the next one has subscribed, and in the one window that stays open the next policy has not started and `backupVariables` is still empty. Latent gap, not a bug today. But if it ever opens, the consequence is exactly APIM-14749 — a second restore replaying the *next* policy's backup — and closing it costs three lines:

```java
final AtomicBoolean restored = new AtomicBoolean();
final Action restoreOnce = () -> {
    if (restored.compareAndSet(false, true)) {
        adaptedCtx.restore();
    }
};

return completable.doOnTerminate(restoreOnce).doOnDispose(restoreOnce);
```

Same applies to `InvokerAdapter` if you want the two to stay symmetric.

A smaller change is hiding in the same swap: under `doFinally` an exception out of `restore()` went to `RxJavaPlugins.onError` and the terminal event still went through; under `doOnTerminate` it is caught by `CompletablePeek.onComplete`'s try and turned into `downstream.onError(e)`. A `ConcurrentModificationException` out of the `HashMap` walk now fails the request instead of being logged. Defensible either way — worth it being a decision rather than a side effect.

phiz71 (commented):
I read the change against the machinery around it — `TemplateContextAdapter`, `HttpPolicyChain`, `HookHelper` — and against the RxJava 3.1.12 operator sources. The diagnosis holds and the fix is the right one: `CompletablePeek.onComplete()` runs `onTerminate` before `downstream.onComplete()`, so the restore lands before the next policy of the chain is subscribed, and it lines up with the `InvokerAdapter` precedent (728a6315cf). I went looking for a live regression and did not find one, so this is mergeable from my side. Notes below: one blocker on the commit message, one on the labels, and some cheap hardening.

Leaving this as a comment rather than an approval.

## Must

**The Mergify labels are short of two branches.** PEN-88 travelled further than the description says: `workerScheduler` and the `policyTimeoutMs > 0` gate (default 10 s, so the offload is on by default) are present on `4.11.x` (701f44a0a8) and `4.10.x` (a9938f6f77), not only from 4.12.0. Both branches are exposed to the same corruption, so alongside the `apply-on-master` already set:

- `apply-on-4-11-x`
- `apply-on-4-10-x`

`4.9.x` is fine — its `HttpPolicyFactory` still calls `new PolicyAdapter(v3Policy)` with no scheduler. The "Since the worker offload (PEN-88, 4.12.0)" sentence in the description is worth adjusting too.

**The commit message has no JIRA footer.** Add, after a blank line at the very end:

    https://gravitee.atlassian.net/browse/APIM-14749

## Should

Four cheap ones, all inline: a shared guard so `restore()` keeps the at-most-once property `doFinally` gave it, a line of comment explaining why this is not `doFinally`, a test on the `doOnDispose` leg, and an assertion on the latch that gives the race test its meaning.

## Could — follow-up, out of scope here

Two pre-existing hazards this fix does not reach, both on the dispose leg where the ordering argument does not apply:

- `TemplateContextAdapter.backupVariables` is a plain `HashMap` and `restore()` flushes it globally. On a request timeout (`AbstractApiReactor.timeout`) or a client abort, the dispose fires on the disposing thread while the v3 policy is still running on the `Schedulers.io()` worker — v3 policies are not interruptible — so `forEach`/`clear` races with `putIfAbsent`.
- `ExecutionContextAdapter.adaptedTemplateEngine` and `TemplateEngineAdapter.adaptedTemplateContext` are non-volatile lazily-initialised fields. The terminate leg gets its happens-before edge from `observeOn`; the dispose leg gets none, so a restore from the timer thread can read `null` and skip silently.

The structural answer to both is a backup scoped per `PolicyAdapter.execute()` instead of one shared by every v3 policy of the request — it would make APIM-14749 impossible rather than timing-dependent.

gravitee-apim-gateway/gravitee-apim-gateway-policy/src/main/java/io/gravitee/gateway/reactive/policy/adapter/policy/PolicyAdapter.java
phiz71: Worth a line of comment here. `doOnTerminate` + `doOnDispose` reads as a verbose spelling of `doFinally` to anyone who does not already know the terminal event has to reach the downstream *after* the restore, and this line is now the only thing standing between the gateway and APIM-14749. The next person tidying this file has every reason to collapse it back, and no test would stop them.

That is exactly how the bug got here in the first place: `InvokerAdapter` received this pattern in 728a6315cf, `PolicyAdapter` did not, and neither file said why it mattered. This file already carries that kind of note for PEN-88 (lines 41 and 119-123); something along the lines of "restore must land before the terminal event — the next v3 policy of the chain shares this `ExecutionContextAdapter`" would be enough.

gravitee-apim-gateway/gravitee-apim-gateway-policy/src/test/java/io/gravitee/gateway/reactive/policy/adapter/policy/PolicyAdapterTest.java
```diff
@@ -460,6 +473,144 @@ void pen88_shouldNotBlockEventLoop_whenWorkerSchedulerConfigured() throws Except
         releasePolicy.countDown();
     }
 
+    @Test
```
phiz71, line 485: These three cover the terminate leg. Nothing covers the dispose leg the description keeps on purpose: drop `.doOnDispose(adaptedCtx::restore)` from `PolicyAdapter` and all three stay green, along with the rest of the suite — while every cancelled request (client abort, request timeout) leaves the v3 policy's overridden `request`/`response` and capture groups in the v4 template context for the remainder of that request.

A fourth test pins it: subscribe to `cut.onRequest(ctx)` with a policy that never completes, dispose the returned `Disposable`, then `verify(adaptedExecutionContext).restore()`. It is also the natural home for a `times(1)` once the at-most-once guard is in.

gravitee-apim-gateway/gravitee-apim-gateway-policy/src/test/java/io/gravitee/gateway/reactive/policy/adapter/policy/PolicyAdapterTest.java
```diff
@@ -460,6 +473,144 @@ void pen88_shouldNotBlockEventLoop_whenWorkerSchedulerConfigured() throws Except
...
+        when(ctx.getInternalAttribute(InternalContextAttributes.ATTR_INTERNAL_ADAPTED_CONTEXT)).thenAnswer(invocation ->
+            cachedAdaptedContext.get()
+        );
+        final ExecutionContextAdapter adaptedContext = spy(ExecutionContextAdapter.create(ctx));
+        cachedAdaptedContext.set(adaptedContext);
+
+        final CountDownLatch secondPolicyOverrodeGroup = new CountDownLatch(1);
+        final CountDownLatch previousPolicyRestored = new CountDownLatch(1);
+        doAnswer(invocation -> {
+            secondPolicyOverrodeGroup.await(RESTORE_HEAD_START_MS, TimeUnit.MILLISECONDS);
```
phiz71, line 621: Worth a comment saying what these 200 ms are for, because this is the only thing that makes the test red against the unfixed code and it cannot be asserted from inside the test: with the fix applied this await *always* times out — the restore now runs before the second policy is even subscribed — so an `isTrue()` here would break the green path.

Which makes the head start best-effort by construction. If the second policy is not scheduled onto a `Schedulers.io()` thread within 200 ms — first io-pool thread creation on a loaded agent, say — `restore()` finds `backupVariables` empty, does nothing, and the test passes against the bug it exists to catch, silently.

Nothing necessarily to change; a note saying "best-effort head start, the await below is what carries the test" would stop a future reader from trimming it, and the `isTrue()` suggested on line 592 is what keeps a vacuous pass out.

gravitee-apim-gateway/gravitee-apim-gateway-policy/src/test/java/io/gravitee/gateway/reactive/policy/adapter/policy/PolicyAdapterTest.java
```diff
@@ -460,6 +473,144 @@ void pen88_shouldNotBlockEventLoop_whenWorkerSchedulerConfigured() throws Except
...
+
+        final AtomicReference<Object> groupSeenByTheSecondPolicy = new AtomicReference<>();
+        final Policy secondPolicy = mock(Policy.class);
+        when(secondPolicy.isRunnable()).thenReturn(true);
+        doAnswer(invocation -> {
+            final TemplateContext policyContext = templateContextOf(invocation);
+            policyContext.setVariable(variableName, secondGroup);
+            secondPolicyOverrodeGroup.countDown();
+            // Evaluate the template only once a restore of the previous policy could have landed.
+            previousPolicyRestored.await(RESTORE_AWAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
phiz71: `CountDownLatch.await` returns a boolean and it is dropped. If this times out — meaning the restore never ran at all, a real regression — the test carries on, reads `"hello"` because nothing clobbered it, and passes vacuously.

```java
assertThat(previousPolicyRestored.await(RESTORE_AWAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS))
    .as("the previous policy must have restored before the template is evaluated")
    .isTrue();
```

This is the assertion that gives the test its meaning: it proves the restore did land before the lookup, so reading the right capture group is a positive result rather than the absence of an event.

thomasballivet:
Thanks for the thorough review.

Addressed in the latest push:

- **Commit message**: JIRA footer added.
- **Labels**: `apply-on-4-11-x` and `apply-on-4-10-x` added, and the description no longer dates the offload to 4.12.0.
- **At-most-once guard**: `restore()` now goes through a shared `restoreOnce` action in `PolicyAdapter`, and `InvokerAdapter` gets the same guard to stay symmetric.
- **Comment**: the `doOnTerminate` / `doOnDispose` pair is now explained in place, including the deliberate choice that an exception out of `restore()` fails the request rather than going to `RxJavaPlugins.onError`.
- **Dispose leg**: two tests added — `should_restore_context_when_disposed_before_completion` (`times(1)`) and `should_restore_context_only_once_when_disposed_after_completion` for the guard.
- **Race test**: the latch result is now asserted (`isTrue()`), so a restore that never lands fails the test instead of passing vacuously; the 200 ms head start is documented as best-effort.

The two "Could" items (non-thread-safe `backupVariables` on the dispose leg, non-volatile lazy fields) are left out as suggested — a backup scoped per `execute()` is the right follow-up and deserves its own ticket.

phiz71 (approved):
All addressed — and extending the at-most-once guard to `InvokerAdapter` was the right call, it closes the same window there.

Two leftovers, both take-them-or-leave-them:

- The description still reads "Since the worker offload (PEN-88, 4.12.0)" and "three new `PolicyAdapterTest` tests" (five now). Only worth a second because Mergify copies `{{ body }}` verbatim into the three backport PRs, two of which that sentence contradicts.
- The dispose leg still flushes `TemplateContextAdapter.backupVariables` — a plain `HashMap` — from the disposing thread, and `ExecutionContextAdapter.adaptedTemplateEngine` is non-volatile, so that leg has no happens-before edge. Both predate this PR and neither belongs in it; worth a ticket eventually, no more than that.

PR #19512, opened 2026-09-01
fix: add deterministic tie-breaker to user search pagination

ankita-gupta21:
Issue : https://gravitee.atlassian.net/browse/APIM-15027

## What
Adds a deterministic tie-breaker to `GET /organizations/{orgId}/users?q=...` pagination, fixing non-deterministic, node-dependent search results.

## Why
`UserDocumentSearcher` paged Lucene results by relevance score with no secondary sort. Documents that tie on score fall back to Lucene's internal document order, which is local to each index. In a clustered deployment every Management API instance builds its own index from the same data, so ties resolve differently per node. Two identical requests hitting different nodes could return different sets of users for "page 1" — walking all pages then duplicated some users and silently dropped others, while `total_elements` stayed correct throughout.

Regression of #11311 / #10744 (both previously closed as fixed).

## Fix
- `UserDocumentTransformer`: index a new `id_sorted` doc-values field.
- `UserDocumentSearcher`: after relevance (or the browsing sort), always add `id_sorted` ascending as a final tie-breaker, so tied documents come back in the same order regardless of node.

### Dev testing evidence:
**Before Fix:** the 20 identical "give me page 1" requests split cleanly into two different groups of 10 — one group of ids whenever port 8083 happened to answer, a different group whenever port 8086 answered. Only 6 of the 10 ids are shared between the two groups — meaning if you asked "who is on page 1?" and got answer A, then asked again a moment later and landed on the other node, four of the ten names you'd get back would be completely different people. Across a full sweep of all 4 pages, 40 rows always came back (the total count was never wrong), but only 32 unique users were represented — 8 users were returned on more than one page, which mathematically means 8 different users were pushed off every page and effectively vanished from the client's point of view. This happened identically on all 3 repeated sweeps.
Screenshot 2026-09-01 at 12 54 58 PM

**After Fix:** all 20 "give me page 1" requests — half answered by 8083, half by 8086 — now collapse into a single group of 20 identical answers. It no longer matters which node answers: page 1 is page 1. Every full sweep returns exactly 40 rows and exactly 40 distinct users, with zero duplicates and zero missing, on all 3 repeated sweeps.
Screenshot 2026-09-01 at 12 55 20 PM

### Bottom line
The fix makes pagination deterministic and node-independent: for a given query, page, and page size, the same set of users is always returned, regardless of which Management API instance answers the request.

Full [setup](https://gravitee.atlassian.net/browse/APIM-15027?focusedCommentId=103704) + [output](https://gravitee.atlassian.net/browse/APIM-15027?focusedCommentId=103705): see [APIM-15027](https://gravitee.atlassian.net/browse/APIM-15027)

[APIM-15027]: https://gravitee.atlassian.net/browse/APIM-15027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcher.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcher.java
@@ -19,6 +19,7 @@ import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDo
 import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer.FIELD_CUSTOM_SPLIT;
 import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer.FIELD_DISPLAYNAME;
 import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer.FIELD_EMAIL;
+import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer.FIELD_ID_SORTED;
 import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer.FIELD_LASTNAME_FIRSTNAME;
 import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer.FIELD_LASTNAME_FIRSTNAME_SORTED;
 import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer.FIELD_REFERENCE;
@@ -48,8 +49,12 @@ import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.search.Sort;
+import org.apache.lucene.search.SortField;
 import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.TopFieldCollectorManager;
+import org.apache.lucene.search.TopFieldDocs;
 import org.apache.lucene.search.WildcardQuery;
 import org.springframework.stereotype.Component;
 
@@ -136,21 +141,35 @@ public class UserDocumentSearcher extends AbstractDocumentSearcher {
             final Set<String> results = new LinkedHashSet<>();
 
             if (pageable != null) {
-                if (sort != null) {
-                    topDocs = searcher.search(query, pageable.getPageNumber() * pageable.getPageSize(), convert(sort));
-                } else {
-                    topDocs = searcher.search(query, Integer.MAX_VALUE);
-                }
+                Sort effectiveSort = withIdTieBreak(sort != null ? convert(sort) : null);
+                // page/size are client-controlled (page has no upper bound -- only size is capped
+                // at 200), so page * size can overflow int and, worse, TopFieldCollectorManager
+                // -- unlike IndexSearcher.search(query, n, sort) -- does not clamp numHits to the
+                // index size: it eagerly allocates a priority queue of that size per search slice.
+                // An unbounded `to` therefore risks an OutOfMemoryError on a small index rather
+                // than the empty/short page IndexSearcher would have returned. Clamp in long
+                // arithmetic to the index size (and at least 1, matching IndexSearcher's own
+                // handling of an empty index) before it reaches the collector.
+                long window = (long) pageable.getPageNumber() * pageable.getPageSize();
+                int to = (int) Math.max(1, Math.min(window, searcher.getIndexReader().maxDoc()));
+                // Plain searcher.search(query, to, sort) lets Lucene stop early (block-max WAND)
+                // once it's confident of the top `to` hits, which turns totalHits into a lower
+                // bound past ~1000 matches instead of an exact count -- silently breaking
+                // total_elements for large orgs. Disabling the early-termination threshold here
+                // keeps the count exact while still only asking for a bounded window of hits.
+                var collectorManager = new TopFieldCollectorManager(effectiveSort, to, null, Integer.MAX_VALUE);
+                TopFieldDocs allMatchingDocs = searcher.search(query, collectorManager);
+                long offset = (long) (pageable.getPageNumber() - 1) * pageable.getPageSize();
+                int from = (int) Math.min(offset, allMatchingDocs.scoreDocs.length);
+                int end = Math.min(to, allMatchingDocs.scoreDocs.length);
 
-                collectedDocs = Arrays.stream(topDocs.scoreDocs)
-                    .skip((long) (pageable.getPageNumber() - 1) * pageable.getPageSize())
-                    .limit(pageable.getPageSize())
-                    .collect(LinkedHashSet::new, Set::add, Set::addAll);
+                topDocs = allMatchingDocs;
+                collectedDocs = Arrays.stream(allMatchingDocs.scoreDocs, from, end).collect(LinkedHashSet::new, Set::add, Set::addAll);
             } else if (sort != null) {
-                topDocs = searcher.search(query, Integer.MAX_VALUE, convert(sort));
+                topDocs = searcher.search(query, Integer.MAX_VALUE, withIdTieBreak(convert(sort)));
                 collectedDocs = Arrays.stream(topDocs.scoreDocs).collect(LinkedHashSet::new, Set::add, Set::addAll);
             } else {
-                topDocs = searcher.search(query, Integer.MAX_VALUE);
+                topDocs = searcher.search(query, Integer.MAX_VALUE, withIdTieBreak(null));
                 collectedDocs = Arrays.stream(topDocs.scoreDocs).collect(LinkedHashSet::new, Set::add, Set::addAll);
             }
 
@@ -164,6 +183,27 @@ public class UserDocumentSearcher extends AbstractDocumentSearcher {
         } catch (IOException ioe) {
             throw new TechnicalException("An error occurs while getting documents from search result", ioe);
         }
+    }
+
+    /**
+     * Appends a deterministic tie-breaker (the user's own id) to a base sort, or builds one from
+     * relevance score alone when no explicit sort was requested. See APIM-15027.
+     *
+     * Always ties on FIELD_ID_SORTED, the doc-values field UserDocumentTransformer indexes for
+     * this purpose -- deriving the field name from the caller's fieldReference instead (e.g.
+     * FIELD_REFERENCE_ID via the inherited searchReference()) would silently produce a field with
+     * no doc values, degrading the tie-break to a no-op instead of failing loudly.
+     */
+    private Sort withIdTieBreak(Sort baseSort) {
+        SortField idTieBreak = new SortField(FIELD_ID_SORTED, SortField.Type.STRING, false);
+        idTieBreak.setMissingValue(SortField.STRING_LAST);
+        if (baseSort == null) {
+            return new Sort(SortField.FIELD_SCORE, idTieBreak);
+        }
+        SortField[] existing = baseSort.getSort();
+        SortField[] combined = Arrays.copyOf(existing, existing.length + 1);
+        combined[existing.length] = idTieBreak;
+        return new Sort(combined);
     }
 
     private boolean isUserIdFormat(io.gravitee.rest.api.service.search.query.Query<?> query) {
--- a/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcherPaginationStabilityTest.java
+++ b/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcherPaginationStabilityTest.java
@@ -1,0 +1,268 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.rest.api.service.impl.search.lucene.searcher;
+
+import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer.FIELD_ID;
+import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer.FIELD_LASTNAME_FIRSTNAME;
+import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer.FIELD_TYPE;
+import static io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer.FIELD_TYPE_VALUE;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.gravitee.rest.api.model.UserEntity;
+import io.gravitee.rest.api.model.common.PageableImpl;
+import io.gravitee.rest.api.model.common.Sortable;
+import io.gravitee.rest.api.model.common.SortableImpl;
+import io.gravitee.rest.api.service.impl.search.SearchResult;
+import io.gravitee.rest.api.service.impl.search.lucene.transformer.UserDocumentTransformer;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import org.apache.lucene.analysis.standard.StandardAnalyzer;
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.index.Term;
+import org.apache.lucene.search.BooleanClause;
+import org.apache.lucene.search.BooleanQuery;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.TermQuery;
+import org.apache.lucene.search.WildcardQuery;
+import org.apache.lucene.store.ByteBuffersDirectory;
+import org.junit.jupiter.api.Test;
+
+/**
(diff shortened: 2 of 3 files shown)


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcherPaginationStabilityTest.java
aqibmohammadkhan, line 1: All three new regression tests call search(query, null, pageable, FIELD_ID) — none pass a non-null Sortable, so the merge-with-existing-sort branch of withIdTieBreak (appending the tie-break to a caller-supplied Sort via Arrays.copyOf) is never exercised.

aqibmohammadkhan (commented):
the fix is scoped to UserDocumentSearcher only — the shared AbstractDocumentSearcher base class (used by other searchers) still lacks this tie-break, so it's a targeted bug fix rather than a systemic one.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcherPaginationStabilityTest.java
ankita-gupta21: Good catch — added should_break_ties_by_id_when_an_explicit_sort_is_provided, which passes a non-null Sortable (mirroring the browsing/no-query path) and forces a tie on that sort key so withIdTieBreak's merge-onto-existing-sort branch actually gets exercised.

ankita-gupta21:
> the fix is scoped to UserDocumentSearcher only — the shared AbstractDocumentSearcher base class (used by other searchers) still lacks this tie-break, so it's a targeted bug fix rather than a systemic one.

Here's a shorter version:

---

Hi @aqibmohammadkhan ,
you're right that the scope is intentionally narrow, and I want to explain the reasoning so we can decide together if that's the right call.
Quick walk thorugh behind scoping this fix to `UserDocumentSearcher`:

**1. It's the exact path the customer hit.** Tracing their request (`?q=svc-ci-`): `UsersResource` → `UserServiceImpl.search()` → `UserDocumentSearcher.search()` — that last method is the one file I changed. Nothing else in the request path is involved.

**2. I reproduced and proved it end-to-end.** I ran two local Management API instances (ports 8083/8086), same database but separate search indexes each — mirroring how nodes work in production. Seeded 40 test users, hit the search endpoint alternating between both ports:
- **Before:** page 1 differed depending on which port answered; full sweeps returned `total=40 distinct=32` — 8 duplicated, 8 missing.
- **After:** identical page 1 either way; every sweep returned `total=40 distinct=40`.

**3. It's backed by an automated test.** `UserDocumentSearcherPaginationStabilityTest.java` recreates the same idea (two indexes, same users, different insert order) so it's verified in CI going forward, not just my manual run.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcher.java
```diff
@@ -136,21 +139,19 @@ protected SearchResult search(org.apache.lucene.search.Query query, Sortable sor
...
-                    .limit(pageable.getPageSize())
-                    .collect(LinkedHashSet::new, Set::add, Set::addAll);
+                Sort effectiveSort = withIdTieBreak(sort != null ? convert(sort) : null, fieldReference);
+                int to = pageable.getPageNumber() * pageable.getPageSize();
+                TopFieldDocs allMatchingDocs = searcher.search(query, to, effectiveSort);
+                int from = Math.min((pageable.getPageNumber() - 1) * pageable.getPageSize(), allMatchingDocs.scoreDocs.length);
+                int end = Math.min(to, allMatchingDocs.scoreDocs.length);
+
+                topDocs = allMatchingDocs;
+                collectedDocs = Arrays.stream(allMatchingDocs.scoreDocs, from, end).collect(LinkedHashSet::new, Set::add, Set::addAll);
```
aqibmohammadkhan: The old no-sort path was searcher.search(query, Integer.MAX_VALUE). IndexSearcher clamps numHits to maxDoc, so the priority queue effectively never filled, setMinCompetitiveScore was never called, and every match was counted — totalHits was exact.

Now numHits is page * size (10 on page 1) and effectiveSort is relevance-first (new Sort(SortField.FIELD_SCORE, idTieBreak)). The queue fills after 10 hits, and once the 1000-hit threshold is crossed Lucene enables block-max WAND skipping — non-competitive matches are never visited, totalHits.relation flips to GREATER_THAN_OR_EQUAL_TO, and .value() becomes a lower bound. That value flows through SearchResult.getHits() into Page.totalElements at UserServiceImpl.java:1389, i.e. total_elements in the response.

Measured on a 60,000-doc index with a SHOULD-clause query (the shape the user query actually builds):

```
OLD  search(q, MAX_VALUE)              -> totalHits=60000        (exact)
NEW  page=1 search(q, 10, SCORE+id)    -> totalHits=18270+       (lower bound)
NEW  page=2 search(q, 20, SCORE+id)    -> totalHits=23285+
TRUE count                             -> 60000
```

the client is told there are 18,270 users, computes ~914 pages instead of 3,000, and the tail becomes unreachable. That's the same "users silently vanish" class of bug this PR sets out to fix, just relocated to large orgs — and the description specifically calls out that total_elements stayed correct throughout, which stops being true after this change.

It lands on exactly the ?q= path this PR targets: UserServiceImpl.java:1378 deliberately passes no sort, so that's the relevance-first branch. I checked the browsing branch (sort != null, UserServiceImpl.java:1370) separately — it's unaffected. It was already flagged GREATER_THAN_OR_EQUAL_TO before this PR, but the value stayed exact both before and after, because a doc-values-first sort doesn't enable score skipping.

Suggested fix — keep the bounded window, disable the early-termination threshold:

```
var manager = new TopFieldCollectorManager(effectiveSort, to, null, Integer.MAX_VALUE);
TopFieldDocs pageDocs = searcher.search(query, manager);
```
Verified against the same index: totalHits=60000 exact, scoreDocs.length=10. You keep the memory win over the old Integer.MAX_VALUE version and still get an accurate count.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcher.java
```diff
@@ -166,6 +167,21 @@ protected SearchResult search(org.apache.lucene.search.Query query, Sortable sor
         }
     }
 
+    /**
+     * Appends a deterministic tie-breaker (the document's own id) to a base sort, or builds one
+     * from relevance score alone when no explicit sort was requested. See APIM-15027.
+     */
+    private Sort withIdTieBreak(Sort baseSort, String fieldReference) {
+        SortField idTieBreak = new SortField(fieldReference + "_sorted", SortField.Type.STRING, false);
```
aqibmohammadkhan: this derives the tie-break field name by string-concatenating the fieldReference argument. AbstractDocumentSearcher.searchReference() (line 84) passes FIELD_REFERENCE_ID, which would produce reference_id_sorted — a field with no doc values, so the tie-break silently degrades to a no-op instead of failing loudly. It's inherited and public on UserDocumentSearcher, so it's reachable, just not currently called for users.

Two things follow: use the constant you already added, and set a missing value the way convert() does at AbstractDocumentSearcher.java:128-133:

```
SortField idTieBreak = new SortField(FIELD_ID_SORTED, SortField.Type.STRING, false);
idTieBreak.setMissingValue(SortField.STRING_LAST);
```
As it stands, UserDocumentTransformer.FIELD_ID_SORTED is declared but never referenced — the searcher rebuilds the same string by hand.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcherPaginationStabilityTest.java
aqibmohammadkhan, line 1: the four new tests are well built and do genuinely fail without the fix. But they all use 40 users, comfortably below Lucene's 1000-hit threshold, so none of them can observe issue 1. A test that indexes >1000 matching users and asserts getHits() equals the true count would pin the behaviour that regressed here.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcher.java
ankita-gupta21: Good catch — that's a real regression. Fixed in:

[UserDocumentSearcher.java#L151-L152](https://github.com/gravitee-io/gravitee-api-management/blob/APIM-15027-org-users-search-q-stability/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcher.java#L151-L152)

**What changed:** replaced the plain searcher.search(query, to, effectiveSort) call with your suggested TopFieldCollectorManager(effectiveSort, to, null, Integer.MAX_VALUE), passing Integer.MAX_VALUE as the total-hits threshold to disable early termination.

**Why:** as you found, a small bounded numHits combined with a score-first sort lets Lucene's block-max WAND optimization skip non-competitive documents once it's confident of the top N — which is exactly the condition my original change introduced by switching away from Integer.MAX_VALUE. That silently turned total_elements into a lower bound for orgs with 1000+ matching users, reintroducing the same "records vanish" symptom this PR set out to fix, just triggered by index size instead of clustering. Disabling the threshold keeps the exact count while still avoiding the unbounded Integer.MAX_VALUE request.

Added a regression test for this too

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcher.java
ankita-gupta21: Fixed in:

[UserDocumentSearcher.java#L187-L189](https://github.com/gravitee-io/gravitee-api-management/blob/APIM-15027-org-users-search-q-stability/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcher.java#L187-L189)

**What changed:** withIdTieBreak no longer takes a fieldReference argument or derives the sort field name via string concatenation — it now always sorts on the FIELD_ID_SORTED constant directly (imported at [L22](https://github.com/gravitee-io/gravitee-api-management/blob/APIM-15027-org-users-search-q-stability/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcher.java#L22)), and sets idTieBreak.setMissingValue(SortField.STRING_LAST) to match the convention convert() already uses.

**Why:** you're right that deriving the field name from whatever fieldReference the caller passes (e.g. `FIELD_REFERENCE_ID` via the inherited `searchReference()`) would silently build a field name with no doc values, and Lucene doesn't error on that — it just treats it as "no tie-break value" for every document, so the fix would quietly do nothing instead of failing loudly. Tying it to the one field UserDocumentTransformer actually indexes for this purpose removes that failure mode entirely, and reuses the constant instead of re-deriving the same string by hand.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcherPaginationStabilityTest.java
ankita-gupta21: Added:

[UserDocumentSearcherPaginationStabilityTest.java#L211](https://github.com/gravitee-io/gravitee-api-management/blob/APIM-15027-org-users-search-q-stability/gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcherPaginationStabilityTest.java#L211) — `should_report_the_exact_total_when_more_than_1000_users_match`

**What it does:** indexes 2,500 matching users, requests a small bounded first page, and asserts `getHits() `equals the exact count (2,500), not a lower bound.

**Why:** exactly as you noted, the existing tests all use 40 users — well under the ~1000-hit threshold where Lucene's early termination kicks in — so none of them could have caught the totalHits regression above. This test fails against the pre-fix `searcher.search(query, to, effectiveSort)` version and passes with the `TopFieldCollectorManager` fix, pinning the exact behavior that regressed.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcher.java
```diff
@@ -136,21 +141,25 @@ protected SearchResult search(org.apache.lucene.search.Query query, Sortable sor
...
-                    topDocs = searcher.search(query, pageable.getPageNumber() * pageable.getPageSize(), convert(sort));
-                } else {
-                    topDocs = searcher.search(query, Integer.MAX_VALUE);
-                }
-
-                collectedDocs = Arrays.stream(topDocs.scoreDocs)
-                    .skip((long) (pageable.getPageNumber() - 1) * pageable.getPageSize())
-                    .limit(pageable.getPageSize())
-                    .collect(LinkedHashSet::new, Set::add, Set::addAll);
+                Sort effectiveSort = withIdTieBreak(sort != null ? convert(sort) : null);
```
aqibmohammadkhan, line 144: `IndexSearcher.search(query, n, sort)` internally does `numHits = Math.min(numHits, Math.max(1, reader.maxDoc()))` before building the collector. Constructing `TopFieldCollectorManager` directly skips that clamp, and `TopFieldCollector` allocates its priority queue eagerly at numHits - per slice, since `newCollector()` is called once per search slice.

`to` is fully client-controlled: size is capped at 200, but page is only `@Min(1)` with no upper bound (management/rest/model/Pageable.java). So `?page=500000&size=200` asks for a 100,000,000-entry queue on an index that may hold a few thousand documents:

```
IndexSearcher.search(q, 100000000, sort)                  -> OK, docs=2500  (clamped to maxDoc)
new TopFieldCollectorManager(sort, 100000000, null, MAX)  -> OutOfMemoryError: Java heap space
```

That's a single authenticated GET taking down the Management API node, where before the change the same request returned an empty page. It also subsumes the int-overflow point from my last round — page * size overflows negative for very large page, and one clamp in long arithmetic fixes both:

```
long window = (long) pageable.getPageNumber() * pageable.getPageSize();
int to = (int) Math.max(1, Math.min(window, searcher.getIndexReader().maxDoc()));
var collectorManager = new TopFieldCollectorManager(effectiveSort, to, null, Integer.MAX_VALUE);
TopFieldDocs pageDocs = searcher.search(query, collectorManager);

long offset = (long) (pageable.getPageNumber() - 1) * pageable.getPageSize();
int from = (int) Math.min(offset, pageDocs.scoreDocs.length);
```

Verified: clamped, the same deep-page request returns cleanly with an exact hit count. Math.max(1, ...) mirrors what IndexSearcher does for an empty index, where maxDoc() is 0 and numHits of 0 would throw IllegalArgumentException: numHits must be > 0.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcherPaginationStabilityTest.java
```diff
@@ -0,0 +1,238 @@
...
+        List<String> allIdsAcrossAllPages = walkAllPages(nodeA, query, browsingSort, USER_COUNT);
+        assertThat(allIdsAcrossAllPages).hasSize(USER_COUNT);
+        assertThat(allIdsAcrossAllPages).doesNotHaveDuplicates();
+        assertThat(allIdsAcrossAllPages).containsExactlyInAnyOrderElementsOf(
+            users.stream().map(UserEntity::getId).collect(Collectors.toList())
+        );
+    }
+
+    @Test
+    void should_report_the_exact_total_when_more_than_1000_users_match() throws Exception {
```
aqibmohammadkhan, line 223: This passes a buggy code.

The intent is exactly right, but the test can't fail. It uses matchAllUsersQuery() — a bare TermQuery on type:user — so all 2500 documents score identically. Block-max WAND prunes non-competitive documents; when every score is equal there is nothing non-competitive to skip, so the count comes back exact even without the fix:

```
PRE-FIX  search(q, 10, sort)   -> hits=2500  => assertThat(hits).isEqualTo(2500) PASSES
POST-FIX manager(threshold=MAX) -> hits=2500  => PASSES
```
Green CI on this commit is consistent with that — the test would not have caught the regression it's meant to pin.

It needs a query whose scores vary, which is also closer to the real thing (the actual user query is a BooleanQuery of SHOULD wildcard clauses, where scores vary by construction). Giving a small subset of the users an extra matching token is enough:

`String lastname = (i % 250 == 0) ? "svc-ci-local-admin-" + i : "svc-ci-local-" + i;`

then querying MUST type:user + SHOULD lastname_firstname:local + SHOULD lastname_firstname:admin. With that shape at the same 2500 users:

```
PRE-FIX  -> hits=2372  (test FAILS, catches the bug)
POST-FIX -> hits=2500  (test PASSES)
```
Worth also asserting firstPage.getHits() is exact on a deep page (new PageableImpl(200, PAGE_SIZE)), which would have caught issue 1 above.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcher.java
ankita-gupta21: done

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/rest/api/service/impl/search/lucene/searcher/UserDocumentSearcherPaginationStabilityTest.java
ankita-gupta21: done

aqibmohammadkhan (approved):

PR #19516, opened 2026-09-01
fix(console): allow read-only access to endpoint group details

jaroslawlakomy-grav:
## Issue

https://gravitee.atlassian.net/browse/APIM-15026

## Description

API members with `api-definition-r` could see endpoint groups but could not open their details from the Endpoints page because the group action incorrectly required `api-definition-u`.

The endpoint group details component also used the read permission to determine editability, which could expose editable controls to read-only users.

## Changes

- Allow users with `api-definition-r` to see and use the **View endpoint group** action.
- Keep **Edit**, add, delete, and reorder actions restricted to users with update permission.
- Determine endpoint group read-only mode using `api-definition-u`.
- Keep Kubernetes-origin API definitions read-only.
- Prevent an enabled health-check configuration from being re-enabled in read-only mode.
- Add regression coverage for:
  - the endpoint group **View** action and navigation;
  - hidden mutation actions for read-only users;
  - disabled general and health-check configuration fields.

## Expected behavior

- A reviewer with `api-definition-r` sees **View**, can open endpoint group details, and cannot modify the configuration.
- An API owner/editor with `api-definition-u` continues to see **Edit** and can update the endpoint group.

### DIFF
--- a/gravitee-apim-console-webui/src/management/api/endpoints-v4/endpoint-group/api-endpoint-group.component.spec.ts
+++ b/gravitee-apim-console-webui/src/management/api/endpoints-v4/endpoint-group/api-endpoint-group.component.spec.ts
@@ -64,6 +64,21 @@ function expectApiSchemaGetRequests(api: ApiV4, fixture: ComponentFixture<any>, 
   fixture.detectChanges();
 }
 
+function expectSharedConfigurationSchemaGet(
+  api: ApiV4,
+  fixture: ComponentFixture<ApiEndpointGroupComponent>,
+  httpTestingController: HttpTestingController,
+  sharedConfigurationSchema: unknown,
+): void {
+  httpTestingController
+    .expectOne({
+      url: `${CONSTANTS_TESTING.org.v2BaseURL}/plugins/endpoints/${api.endpointGroups[0].type}/shared-configuration-schema`,
+      method: 'GET',
+    })
+    .flush(sharedConfigurationSchema);
+  fixture.detectChanges();
+}
+
 /**
  * Expect that a single PUT request has been made which matches the specified URL
  *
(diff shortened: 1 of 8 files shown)


gravitee-apim-console-webui/src/management/api/endpoints-v4/endpoint-groups/standard/api-endpoint-groups-standard.component.html
```diff
@@ -65,7 +65,7 @@
               <mat-icon svgIcon="gio:arrow-down"></mat-icon>
             </button>
             <button
-              *gioPermission="{ anyOf: ['api-definition-u'] }"
+              *gioPermission="{ anyOf: ['api-definition-r'] }"
```
thomasballivet, line 68: Same gate is still `api-definition-u` in the LLM variant (`endpoint-groups/llm/api-endpoint-groups-llm.component.html`, lines 28 and 57), while the target routes already allow `api-definition-r` and `ApiLlmProviderComponent` already derives its read-only state from `api-definition-u`. So a read-only member on an LLM_PROXY API still gets no action button at all — same symptom, up to you if it worth fixing here too of if it is out of scope.

gravitee-apim-console-webui/src/management/api/endpoints-v4/endpoint-group/api-endpoint-group.component.spec.ts
```diff
@@ -138,6 +142,37 @@ describe('ApiEndpointGroupComponent', () => {
...
+            },
+          }),
+        ],
+      });
+      await initComponent(api, { apiId: API_ID, groupIndex: 0 }, ['api-definition-r']);
+
+      await componentHarness.clickGeneralTab();
+      expectApiSchemaGetRequests(api, fixture, httpTestingController);
+      expect(await componentHarness.isEndpointGroupNameInputDisabled()).toEqual(true);
+      expect(await componentHarness.isEndpointGroupLoadBalancerSelectorDisabled()).toEqual(true);
```
thomasballivet, line 193: Could we also assert on the Configuration tab (`groupConfiguration`)? It's the biggest editable surface and the only control whose disabled state relies solely on the constructor flag — exactly the shape of the health-check `startWith` bug this PR fixes. Without it, a future refactor re-enabling it would keep CI green. wdyt?

thomasballivet (approved):

gravitee-apim-console-webui/src/management/api/endpoints-v4/endpoint-groups/standard/api-endpoint-groups-standard.component.html
jaroslawlakomy-grav: Thanks for spotting this. LLM_PROXY uses a separate list component and was not part of the reported APIM-15026 scenario. To keep this bugfix focused, I’d prefer to leave the LLM variant out of scope and track it separately if needed.

gravitee-apim-console-webui/src/management/api/endpoints-v4/endpoint-group/api-endpoint-group.component.spec.ts
jaroslawlakomy-grav: Agreed. I’ll extend the harness and read-only test to render a shared configuration schema and assert that its input remains disabled.

PR #19519, opened 2026-09-01
Stop shipping the authz PDP service to the Management API

michel-barret:
## Issue

## Description

The exclusion targeted com.graviteesource.service, while the dependency is declared under com.graviteesource.gamma, so it never matched and the PDP service plugin ended up in the Management API plugins directory.

### DIFF
--- a/gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-rest-api/src/main/assembly/plugin-assembly-distribution.xml
+++ b/gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-rest-api/src/main/assembly/plugin-assembly-distribution.xml
@@ -186,7 +186,7 @@
                 <exclude>io.gravitee.reporter:*:zip</exclude>
                 <exclude>io.gravitee.tracer:*:zip</exclude>
                 <exclude>com.graviteesource.service:gravitee-service-secrets:zip</exclude>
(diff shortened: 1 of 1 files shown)


gravitee-apim-distribution/gravitee-apim-distribution-standalone/gravitee-apim-distribution-standalone-rest-api/src/main/assembly/plugin-assembly-distribution.xml
```diff
@@ -186,7 +186,7 @@
                 <exclude>io.gravitee.reporter:*:zip</exclude>
                 <exclude>io.gravitee.tracer:*:zip</exclude>
                 <exclude>com.graviteesource.service:gravitee-service-secrets:zip</exclude>
-                <exclude>com.graviteesource.service:gravitee-service-authz-pdp:zip</exclude>
+                <exclude>com.graviteesource.gamma:gravitee-service-authz-pdp:zip</exclude>
```
phiz71, line 189: actually, the service should follow the convention.
groupId: "com.graviteesource.service"

Do you think we could challenge this and change the group ID of the plugin ?
But I'm open to discussion, maybe this is justified.

michel-barret: I’m fully agree, if we don’t have other compatibility concern go :)
I will check where the pdp is declared in deps

PR #19529, opened 2026-09-01
fix(idp): scope the identity provider enabled flag back to portal login

callaertanthony:
https://gravitee.atlassian.net/browse/APIM-15029

## Problem

A previous fix (commit `65b72f3ae4` and its siblings) made `IdentityProviderEntity.enabled` gate
every activation target instead of only `ENVIRONMENT`. Before that, `enabled` decided whether an
identity provider appeared on the **Portal** login page; Console (`ORGANIZATION`) login ignored it.

The Console toggle for that field reads *"Allow portal authentication to use this identity
provider"*. Organizations that took the label at its word — turning it off to keep an identity
provider off the Portal while keeping Console SSO — lost Console SSO the moment they upgraded. Where
local login was also disabled, the organization was locked out of its own Console with no
in-product way back in.

Affected releases: 4.9.33, 4.10.29, 4.11.26, 4.12.18.

## What this changes

**1. `enabled` is portal-scoped again** (`SocialIdentityProviderImpl`)

`findAll` and `findById` filter on `enabled` only for the `ENVIRONMENT` target. Console login no
longer consults the flag.

**2. Deactivation is the Console off switch** — already true, now pinned by a test

Both methods already reject an identity provider with no activation row for the requested target.
That is the supported way to revoke Console access, it is what the **Deactivate** toggle on the
identity provider list does, and it is what Cockpit uses (`DisableEnvironmentCommandHandler`,
`DeleteOrganizationCommandHandler` remove activation rows, never touch `enabled`). A new test
covers it so the real off switch cannot regress unnoticed.

## No data migration, deliberately

An earlier revision of this PR carried a Mongo upgrader and a Liquibase changeset setting
`enabled = true` for identity providers activated on the organization and on no environment. It was
dropped, for two reasons:

- It rescues nobody. The migration only runs on a build that already contains the fix above, and
  that fix alone restores Console SSO for every affected installation.
- It removes a protection. `enabled = true` is inert only while the provider has no environment
  activation. The moment an administrator later activates it on an environment, the Portal button
  appears immediately — whereas today `enabled = false` would still have held it back.

Zero data writes on a security-sensitive fix shipping to five branches is the safer trade.

## Breaking change

**A "disabled" identity provider is accepted on Console login again.** If you relied on
`enabled = false` to block Console SSO on 4.9.33 / 4.10.29 / 4.11.26 / 4.12.18, that no longer
blocks it.

To actually turn an identity provider off for the Console, **deactivate it**: Organization settings
→ Authentication → the identity provider list → the *Activated / Deactivated* toggle. This removes
the `ORGANIZATION` activation and is enforced on every Console authentication path. The `enabled`
flag governs Portal visibility only, as its label states.

A known-issue note for the four already-published patch releases is tracked separately.

## Security-sensitive

This change touches authentication and identity provider activation. It intentionally reverses the
Console-side tightening shipped in the earlier fix; the reporting path for that issue should be
directed at the Deactivate toggle described above.

## Tests

`SocialIdentityProviderImplTest` — 8 cases, green on a clean build of this branch
(4.12.19-SNAPSHOT).

The two Console cases were written first and watched fail against the old code
(`expected:<[disabled-idp, enabled-idp]> but was:<[enabled-idp]>`, and
`IdentityProviderNotFoundException`), then made green. The Portal cases are unchanged and still
assert that `enabled = false` hides the provider from the `ENVIRONMENT` target.

**Gap:** no test asserts the reported symptom end to end — the SSO button on the Console login
page. `gravitee-apim-e2e/ui-test/.../ui-login.spec.ts` covers username/password only, and the
committed management SDK has no identity provider client, so that test needs an SDK regeneration
plus fixtures. Tracked separately rather than added to a hotfix.

https://gravitee.atlassian.net/browse/APIM-15029

### DIFF
--- a/gravitee-apim-e2e/api-test/src/authentication/mapi-v1/console-social-identity-providers.spec.ts
+++ b/gravitee-apim-e2e/api-test/src/authentication/mapi-v1/console-social-identity-providers.spec.ts
@@ -18,15 +18,32 @@ import { faker } from '@faker-js/faker';
 
 import { ANONYMOUS, forManagement, forManagementAsAdminUser } from '@gravitee/utils/configuration';
 import { ConfigurationApi } from '../../../../lib/management-webclient-sdk/src/lib/apis/ConfigurationApi';
+import { DefaultApi } from '../../../../lib/management-webclient-sdk/src/lib/apis/DefaultApi';
 import { OrganizationApi } from '../../../../lib/management-webclient-sdk/src/lib/apis/OrganizationApi';
 import { PortalApi } from '../../../../lib/management-webclient-sdk/src/lib/apis/PortalApi';
+import type { IdentityProviderActivationEntity } from '../../../../lib/management-webclient-sdk/src/lib/models';
 import { created, noContent, succeed } from '@lib/jest-utils';
 
 const orgId = 'DEFAULT';
 
 const configurationApiAsAdmin = new ConfigurationApi(forManagementAsAdminUser());
+const defaultApiAsAdmin = new DefaultApi(forManagementAsAdminUser());
 const organizationApiAsAdmin = new OrganizationApi(forManagementAsAdminUser());
 const portalApiAsAnonymous = new PortalApi(forManagement(ANONYMOUS));
+
+/**
+ * PUT /organizations/{orgId}/identities replaces the whole activation set — updateTargetIdp
+ * removes every activation absent from the body — so the original set is read up front, extended,
+ * and put back verbatim on teardown rather than cleared.
+ */
+async function setOrganizationActivations(identityProviderIds: string[]) {
+  await noContent(
+    organizationApiAsAdmin.updateOrganizationIdentitiesRaw({
+      orgId,
+      identityProviderActivationEntity: identityProviderIds.map((identityProvider) => ({ identityProvider })),
+    }),
+  );
+}
 
 /**
  * The Console login page reads GET /organizations/{orgId}/social-identities unauthenticated to
@@ -36,8 +53,14 @@ const portalApiAsAnonymous = new PortalApi(forManagement(ANONYMOUS));
  */
 describe('Console social identity providers', () => {
   let identityProviderId: string;
+  let originalActivationIds: string[];
 
   beforeAll(async () => {
+    const originalActivations: IdentityProviderActivationEntity[] = await succeed(
+      defaultApiAsAdmin.listIdentityProviderActivationsRaw({ orgId }),
+    );
+    originalActivationIds = originalActivations.map((activation) => activation.identityProvider);
+
     const identityProvider = await created(
       configurationApiAsAdmin.createIdentityProviderRaw({
         orgId,
@@ -51,27 +74,28 @@ describe('Console social identity providers', () => {
     );
     identityProviderId = identityProvider.id;
 
-    await noContent(
-      organizationApiAsAdmin.updateOrganizationIdentitiesRaw({
-        orgId,
-        identityProviderActivationEntity: [{ identityProvider: identityProviderId }],
-      }),
-    );
+    await setOrganizationActivations([...originalActivationIds, identityProviderId]);
   });
 
   afterAll(async () => {
-    await organizationApiAsAdmin.updateOrganizationIdentities({ orgId, identityProviderActivationEntity: [] });
+    if (!identityProviderId) {
+      return;
+    }
+    await setOrganizationActivations(originalActivationIds);
     await configurationApiAsAdmin.deleteIdentityProvider({ orgId, identityProvider: identityProviderId });
   });
 
   it('should list an identity provider that is disabled but activated on the organization', async () => {
+    // getSocialIdentityProviders1 is the org-level /organizations/{orgId}/social-identities. The
+    // deprecated /environments/{envId}/portal/identities variant takes an envId, so a regeneration
+    // that renumbered them would fail to compile here rather than silently retarget the test.
     const socialIdentityProviders = await succeed(portalApiAsAnonymous.getSocialIdentityProviders1Raw({ orgId }));
 
     expect(socialIdentityProviders.map((idp) => idp.id)).toContain(identityProviderId);
   });
 
   it('should stop listing it once it is deactivated on the organization', async () => {
-    await organizationApiAsAdmin.updateOrganizationIdentities({ orgId, identityProviderActivationEntity: [] });
+    await setOrganizationActivations(originalActivationIds);
 
     const socialIdentityProviders = await succeed(portalApiAsAnonymous.getSocialIdentityProviders1Raw({ orgId }));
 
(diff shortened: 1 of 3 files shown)


gravitee-apim-e2e/api-test/src/authentication/mapi-v1/console-social-identity-providers.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+          type: 'GOOGLE',
+          enabled: false,
+          configuration: { clientId: 'a-client-id', clientSecret: 'a-client-secret' },
+        },
+      }),
+    );
+    identityProviderId = identityProvider.id;
+
+    await noContent(
+      organizationApiAsAdmin.updateOrganizationIdentitiesRaw({
```
aqibmohammadkhan: This PUT is replace-all — updateTargetIdp (IdentityProviderActivationServiceImpl:263) deletes every activation not present in the body. So beforeAll silently drops any pre-existing ORGANIZATION activation on DEFAULT, and line 63 then leaves the org with none rather than restoring what was there. The suite runs --runInBand against one shared APIM, so whatever ran before this file loses its activations for the rest of the run.

Could you read the current set with GET /organizations/{orgId}/identities (listIdentityProviderActivations) in beforeAll, append the new IdP to it, and restore the original list in afterAll?

gravitee-apim-e2e/api-test/src/authentication/mapi-v1/console-social-identity-providers.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    );
+  });
+
+  afterAll(async () => {
+    await organizationApiAsAdmin.updateOrganizationIdentities({ orgId, identityProviderActivationEntity: [] });
+    await configurationApiAsAdmin.deleteIdentityProvider({ orgId, identityProvider: identityProviderId });
+  });
+
+  it('should list an identity provider that is disabled but activated on the organization', async () => {
+    const socialIdentityProviders = await succeed(portalApiAsAnonymous.getSocialIdentityProviders1Raw({ orgId }));
```
aqibmohammadkhan, line 92: SocialIdentityProvidersResource is mounted at two paths — /organizations/{orgId}/social-identities (OrganizationResource:181) and the deprecated /organizations/{orgId}/environments/{envId}/portal/identities (PortalResource:104, under EnvironmentResource:203). Both generate an operation named getSocialIdentityProviders, and the generator suffixes one of them 1; which one gets the suffix is ordering-dependent, not something the call site expresses.

Path ordering suggests 1 does land on the org-level path, which is what you want. But if it lands on the portal one, this needs envId (tsc failure) and isn't permitAll in BasicSecurityConfigurerAdapter:297, so the anonymous call 401s. Since no CI job on this branch compiles or runs this file, that's currently unproven either way — worth confirming against a generated SDK before merge, and maybe worth a comment here recording which path 1 is, so a future regeneration that renumbers them doesn't silently retarget the test.

gravitee-apim-e2e/api-test/src/authentication/mapi-v1/console-social-identity-providers.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+      organizationApiAsAdmin.updateOrganizationIdentitiesRaw({
+        orgId,
+        identityProviderActivationEntity: [{ identityProvider: identityProviderId }],
+      }),
+    );
+  });
+
+  afterAll(async () => {
+    await organizationApiAsAdmin.updateOrganizationIdentities({ orgId, identityProviderActivationEntity: [] });
+    await configurationApiAsAdmin.deleteIdentityProvider({ orgId, identityProvider: identityProviderId });
```
aqibmohammadkhan, line 85: If beforeAll fails, identityProviderId is undefined here and this throws a second, less informative error on top of the real one. A if (identityProviderId) guard keeps the actual failure readable.

gravitee-apim-e2e/api-test/src/authentication/mapi-v1/console-social-identity-providers.spec.ts
callaertanthony: Confirmed and fixed in 8acf2eb768. `updateTargetIdp` (IdentityProviderActivationServiceImpl:263-274) iterates the target's existing activations and calls `removeIdpsFromTarget` for every one absent from the body, so `beforeAll` was dropping whatever DEFAULT already had and `afterAll` was leaving it empty for the rest of the run.

Now reads the original set via `DefaultApi.listIdentityProviderActivations` (GET /organizations/{orgId}/identities), appends the new provider, and restores the original list verbatim on teardown. The deactivation case reuses the same helper, so it asserts removal of just this provider rather than clearing the org.

gravitee-apim-e2e/api-test/src/authentication/mapi-v1/console-social-identity-providers.spec.ts
callaertanthony: Checked this against a generated SDK rather than leaving it inferred. Built console-openapi.json, ran `yarn update:sdk:management`, and read the emitted path template: PortalApi.ts:1507 has `getSocialIdentityProviders1Raw` on `/organizations/{orgId}/social-identities`, and `GetSocialIdentityProviders1Request` carries only `orgId`. So `1` is the org-level path, and the anonymous call is against the permitAll endpoint.

On the renumbering risk: it cannot silently retarget to the portal path, because that variant's request type requires `envId` — a regeneration that renumbered them would fail `tsc` here rather than pass against the wrong endpoint. The one benign case is renumbering onto `/v1/organizations/{orgId}/social-identities`, same resource, same target. I have added a comment at the call site recording which path `1` is and that property.

Your point about no CI job compiling this file stands. I ran `tsc --noEmit` locally (clean, and verified tsc actually includes the file by inserting a bogus param and watching it fail), but the spec has not been executed — ports 8082/8083 were occupied on my machine. Flagging it as unproven at runtime until the e2e pipeline runs it.

gravitee-apim-e2e/api-test/src/authentication/mapi-v1/console-social-identity-providers.spec.ts
callaertanthony: Fixed in 8acf2eb768 — `afterAll` now returns early when `identityProviderId` is unset, so a `beforeAll` failure surfaces its own error instead of a null dereference on top of it.

gravitee-apim-e2e/api-test/src/authentication/mapi-v1/console-social-identity-providers.spec.ts
aqibmohammadkhan: A single push as `...-run-e2e` would close it, no?

aqibmohammadkhan (approved):

PR #19535, opened 2026-09-01
fix(gateway): report a client abort on the request span

remibaptistegio:
## Description

When a client disconnects mid-response, the request's root span was recorded as a success. A cancel is
not a failure in Rx terms: it never reaches `doOnError`, so nothing was stashed in
`ATTR_INTERNAL_TRACING_ERROR`, and `doFinally` closed the span on the response code alone — `200` for a
streamed response whose headers went out long before the client left.

`ClientCloseClassifier.decorate` is the one place that already knows a close for what it is. It now
stashes the cause in `ATTR_INTERNAL_TRACING_ERROR` alongside the metrics it decorates, and the
dispatcher's existing span-ending path picks it up unchanged — no new branch, no new helper, the
dispatcher untouched.

Two consequences worth naming:

- **The span carries a real exception.** The `Throwable` overload reaches
  `InstrumenterTracer.endSpan(..., throwable)`, which emits the `exception` event and `error.type` that
  error breakdowns in Tempo/Datadog key off. Recording only a status description would have left a
  client-abort span invisible to them.
- **Both close entry points are covered.** `VertxHttpServerResponse`'s write path does
  `onErrorResumeNext(... -> Completable.complete())`, so the error is swallowed there and never reaches
  the response exception handler the dispatcher registers. That path is the motivating SSE case.

The stash sits inside `decorate`'s existing guards, deliberately: the span is written exactly when and
only when the metrics are, so a trace and its analytics can no longer disagree. A close on a
kept-alive connection between requests still decorates nothing.

**Security-sensitive:** touches tracing and diagnostics. It changes what a span records for an aborted
request; it adds no new data source, and the stashed value is an exception the gateway already handled
and already reported through the metrics.

## Additional context

**Reproduction.** An SSE stream cut off by a gateway restart: `POST /…` with `http.status_code: 200`
and no error status, while the analytics for the same request were correctly tagged
`CLIENT_ABORTED_CHANNEL_CLOSED`.

**Blast radius.** Any streaming v4 API whose client disconnects mid-response — SSE, LLM proxy, agents.
Traces for those requests change from success to failure, which is the point, but it is a visible
change for anyone alerting on span status.

Nothing else changes. A request with no classified client close ends its span exactly as before — in
particular, a keyed policy rejection (401, 403, 429) keeps the `UNSET` status the OpenTelemetry SERVER
semconv gives a 4xx (`HttpSpanStatusExtractor` with the SERVER converter, `code >= 500 || code < 100`).

**Out of scope.** The v3 path is deliberately untouched: it uses `doOnComplete`/`doOnError` with no
`doFinally`, so a cancelled v3 request appears never to end its span at all. Separate defect, separate
fix.

Two adjacent tracing defects found during review, filed for a follow-up PR rather than folded in here:
a plain `InterruptionException` marks the Request-phase span as an error even though nothing failed,
and an `InterruptionFailureException` records no failure key on either the phase span or the root span.

**Tests.** `DefaultHttpRequestDispatcherTest.shouldEndTracingSpanInErrorWhenTheClientAbortedTheResponseStream`
drives a real dispatch: the metrics are tagged through the classifier, the chain then completes normally
as the write-failure path does, and the root span must end with that exception. It fails on the
pre-change code. Two guard tests in `ClientCloseClassifierTest`: the cause is stashed, and an
already-ended request stashes nothing. RED → GREEN confirmed on all three; both gateway modules green
(321 tests), `prettier:check` green.

### DIFF
--- a/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/test/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcherTest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/test/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcherTest.java
@@ -218,6 +218,7 @@ class DefaultHttpRequestDispatcherTest {
         lenient().when(requestTimeoutConfiguration.getRequestTimeout()).thenReturn(0L);
         lenient().when(requestTimeoutConfiguration.getRequestTimeoutGraceDelay()).thenReturn(10L);
 
+        lenient().when(globalComponentProvider.getComponent(Node.class)).thenReturn(node);
         mockConnectionCreationTimestamp();
 
         spyNoopTracer = spy(new NoOpTracer());
@@ -419,6 +420,25 @@ class DefaultHttpRequestDispatcherTest {
                 .assertError(t -> MOCK_ERROR_MESSAGE.equals(t.getMessage()));
             verify(spyNoopTracer).startRootSpanFrom(any(), any());
             verify(spyNoopTracer, timeout(1000)).endWithResponseAndError(any(), any(), any(), any(Throwable.class));
+        }
+
+        @Test
+        void shouldEndTracingSpanInErrorWhenTheClientAbortedTheResponseStream() {
+            // The write failure is swallowed by VertxHttpServerResponse, so the dispatch completes normally and the
+            // span would otherwise be closed on the response code alone - a 200 for a stream that never finished.
+            final IOException clientClose = new IOException("Connection reset by peer");
+            when(apiReactor.handle(any(MutableExecutionContext.class))).thenAnswer(invocation -> {
+                MutableExecutionContext ctx = invocation.getArgument(0);
+                ctx.metrics(Metrics.builder().build());
+                ClientCloseClassifier.decorate(ctx, clientClose);
+                return Completable.complete();
+            });
+
+            cut.dispatch(rxRequest, SERVER_ID).test().assertComplete();
+
+            ArgumentCaptor<Throwable> endedWith = ArgumentCaptor.forClass(Throwable.class);
+            verify(spyNoopTracer, timeout(1000)).endWithResponseAndError(any(), any(), any(), endedWith.capture());
+            assertThat(endedWith.getValue()).isSameAs(clientClose);
         }
     }
 
--- a/gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/reactive/http/vertx/ClientCloseClassifier.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-http/src/main/java/io/gravitee/gateway/reactive/http/vertx/ClientCloseClassifier.java
@@ -16,6 +16,7 @@
 package io.gravitee.gateway.reactive.http.vertx;
 
 import io.gravitee.gateway.reactive.api.ExecutionFailure;
+import io.gravitee.gateway.reactive.api.context.InternalContextAttributes;
 import io.gravitee.gateway.reactive.api.context.base.BaseExecutionContext;
 import io.gravitee.gateway.reactive.api.context.http.HttpBaseExecutionContext;
 import io.gravitee.gateway.reactive.core.context.ComponentScope;
@@ -112,7 +113,8 @@ public final class ClientCloseClassifier {
     }
 
     /**
-     * Decorate the in-flight request {@link Metrics} with a client-close failure, going through
+     * Decorate the in-flight request {@link Metrics} with a client-close failure, and stash the cause for the request
+     * span so the trace and the analytics name the same failure. Goes through
      * {@link DiagnosticReportHelper} — the same path {@code AbstractExecutionContext.interruptWith} uses — so the
      * {@link Diagnostic} written here is <b>component-attributed</b> (the closing {@code ENDPOINT}/{@code SYSTEM}
      * component, via {@link ComponentScope}). No-op when there is no metrics yet, when the request already completed,
@@ -154,6 +156,13 @@ public final class ClientCloseClassifier {
         // Legacy flat fields kept in sync so dashboards reading error-key/error-message still work.
         metrics.setErrorKey(diagnostic.getKey());
         metrics.setErrorMessage(diagnostic.getMessage());
+        if (cause != null) {
+            // A client abort cancels the request chain rather than failing it, so it reaches no doOnError and the
+            // dispatcher would close the root span on the response code alone - a 200 for a stream the client left
+            // mid-flight. Stashing the cause where the dispatcher already looks records it as the exception it was,
+            // with the exception event and error.type an APM breaks errors down by.
+            ctx.putInternalAttribute(InternalContextAttributes.ATTR_INTERNAL_TRACING_ERROR, cause);
+        }
         ctx.withLogger(log).debug("Classified client close reason as {}", key);
     }
 
--- a/gravitee-apim-gateway/gravitee-apim-gateway-http/src/test/java/io/gravitee/gateway/reactive/http/vertx/ClientCloseClassifierTest.java
+++ b/gravitee-apim-gateway/gravitee-apim-gateway-http/src/test/java/io/gravitee/gateway/reactive/http/vertx/ClientCloseClassifierTest.java
@@ -18,10 +18,14 @@ package io.gravitee.gateway.reactive.http.vertx;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertTimeoutPreemptively;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import io.gravitee.gateway.reactive.api.context.InternalContextAttributes;
 import io.gravitee.gateway.reactive.api.context.http.HttpBaseExecutionContext;
 import io.gravitee.gateway.reactive.http.vertx.ClientCloseClassifier.ClientCloseReason;
 import io.gravitee.reporter.api.v4.metric.Diagnostic;
@@ -213,6 +217,33 @@ class ClientCloseClassifierTest {
 
         assertThat(metrics.getErrorKey()).isNull();
         assertThat(metrics.getFailure().getKey()).isEqualTo("GATEWAY_OAUTH2_ACCESS_DENIED");
+    }
+
+    @Test
+    void should_stash_the_close_cause_for_the_request_span() {
+        // A client abort cancels the chain instead of failing it, so it reaches no doOnError: the dispatcher ends the
+        // root span from this attribute, or reports a success for a stream that never finished.
+        HttpBaseExecutionContext ctx = mockContext();
+        when(ctx.metrics()).thenReturn(Metrics.builder().build());
+        IOException cause = new IOException("Connection reset by peer");
+
+        ClientCloseClassifier.decorate(ctx, cause);
+
+        verify(ctx).putInternalAttribute(InternalContextAttributes.ATTR_INTERNAL_TRACING_ERROR, cause);
+    }
+
+    @Test
+    void should_not_stash_a_close_cause_for_an_already_completed_request() {
+        // Same guard as the metrics: a close on a kept-alive connection between requests must not turn the span of
+        // the request that already completed on it into a failure.
+        HttpBaseExecutionContext ctx = mockContext();
+        Metrics metrics = Metrics.builder().build();
+        metrics.setRequestEnded(true);
+        when(ctx.metrics()).thenReturn(metrics);
+
+        ClientCloseClassifier.decorate(ctx, new IOException("Connection reset by peer"));
+
+        verify(ctx, never()).putInternalAttribute(eq(InternalContextAttributes.ATTR_INTERNAL_TRACING_ERROR), any());
     }
 
     private static HttpBaseExecutionContext mockContext() {


gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -368,6 +354,33 @@ static void classifyResponseStreamError(final MutableExecutionContext ctx, final
...
+     */
+    static void endRootSpan(final MutableExecutionContext ctx, final Span rootSpan, final ObservableHttpServerResponse response) {
+        final Throwable throwable = ctx.getInternalAttribute(ATTR_INTERNAL_TRACING_ERROR);
+        if (throwable != null) {
+            ctx.getTracer().endWithResponseAndError(rootSpan, response, throwable);
+            return;
+        }
+        final Metrics metrics = ctx.metrics();
+        final String errorKey = metrics != null ? metrics.getErrorKey() : null;
+        if (errorKey != null && !errorKey.isBlank()) {
```
phiz71: **This fires on far more than a client abort.**

`AbstractFailureProcessor.processFailure` does `ctx.metrics().setErrorKey(executionFailure.key())` for *every* `interruptWith`, and a policy interruption does not propagate through the outer chain — the failure processor writes the response and the chain completes normally, so nothing is stashed in `ATTR_INTERNAL_TRACING_ERROR` and we land here.

So every keyed policy rejection now ends its root span with `StatusCode.ERROR`: 401 from an API-key or JWT policy, 403, 429 from rate limiting, quota. Before this PR those went through the `Throwable` overload with `null`, and `HttpStatusCodeConverter.SERVER` leaves 4xx `UNSET` — only 5xx became `ERROR`. That is a step change in span error rate driven by ordinary authentication traffic, on every v4 *and* v2-emulated API, not just the streaming ones in the stated scope.

A second one worth looking at: `HttpConnector` sets `STREAM_ENDED_EARLY` when the upstream truncates a body, right under the comment *"WARN, not ERROR: the caller did get a response"*. That span now becomes `ERROR`, which contradicts the judgement made in that file.

Narrowing the condition to the `ClientCloseClassifier` keys (or to `metrics.getFailure()` carrying a 499) would keep the fix and drop the collateral.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -368,6 +354,33 @@ static void classifyResponseStreamError(final MutableExecutionContext ctx, final
...
+     *
+     * <p>A client that goes away cancels the chain rather than failing it, so nothing reaches {@code doOnError} and the
+     * span would be closed on the response code alone — a 200 for a stream that never finished. The abort is already
+     * classified onto the metrics by {@link ClientCloseClassifier}, so the reason is read from there rather than
+     * invented, and a cancelled stream is recorded as the failure it was.</p>
+     *
+     * <p>Package-private and static so the three call sites can be covered directly, as {@code classifyResponseStreamError}
+     * already is.</p>
+     */
+    static void endRootSpan(final MutableExecutionContext ctx, final Span rootSpan, final ObservableHttpServerResponse response) {
```
phiz71: **The abort is already a `Throwable` one frame up — consider stashing it there instead.**

`classifyResponseStreamError` (just above) holds both the ctx and the throwable, and `ClientCloseClassifier.decorate` receives `cause` too. A single

```java
ctx.putInternalAttribute(ATTR_INTERNAL_TRACING_ERROR, throwable);
```

at either point feeds the mechanism the dispatcher already has, and needs no new branch in the span-ending path at all.

It also produces a better span. `OpenTelemetryTracer`'s `String` overload does `span.setStatus(ERROR, message)` and then ends via `endWithResponse`, i.e. `instrumenter.end(..., null)` — so the span gets a status but **no `exception` event and no `error.type` attribute**, which is what OTel emits only on the throwable path. Error breakdowns in Tempo/Datadog key off exactly those, so a client-abort span stays invisible to them today.

Doing it in `decorate` would additionally cover the other close entry point — `VertxHttpServerResponse`'s write-failure path — which goes through the same helper but never reaches the response exception handler registered by `registerClientCloseClassifier`.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -368,6 +354,33 @@ static void classifyResponseStreamError(final MutableExecutionContext ctx, final
...
+        final Throwable throwable = ctx.getInternalAttribute(ATTR_INTERNAL_TRACING_ERROR);
+        if (throwable != null) {
+            ctx.getTracer().endWithResponseAndError(rootSpan, response, throwable);
+            return;
+        }
+        final Metrics metrics = ctx.metrics();
+        final String errorKey = metrics != null ? metrics.getErrorKey() : null;
+        if (errorKey != null && !errorKey.isBlank()) {
+            ctx.getTracer().endWithResponseAndError(rootSpan, response, errorKey);
```
phiz71: **We already have a convention for this, and it is the other way round.**

`AbstractTracingHook.endSpanWithFailure` records a failure on a span as:

```java
span.withAttribute("gravitee.execution-failure.key", failure.key())
    .withAttribute("gravitee.execution-failure.status-code", failure.statusCode())
    ...
tracer.endOnError(span, failure.message());
```

— key as an *attribute*, message as the status description. Here the key becomes the description, and the message is dropped, even though `ClientCloseClassifier` produces one for exactly this purpose (`CLIENT_ABORTED_CHANNEL_CLOSED_MESSAGE`, mirrored into `metrics.errorMessage`).

Net effect: the human-readable reason is lost, and the stable value worth filtering on ends up in the one field that is not indexed as an attribute. `metrics.getErrorMessage()` for the description, key as an attribute, would line this up with the hook.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -368,6 +354,33 @@ static void classifyResponseStreamError(final MutableExecutionContext ctx, final
...
+     * <p>Package-private and static so the three call sites can be covered directly, as {@code classifyResponseStreamError}
+     * already is.</p>
+     */
+    static void endRootSpan(final MutableExecutionContext ctx, final Span rootSpan, final ObservableHttpServerResponse response) {
+        final Throwable throwable = ctx.getInternalAttribute(ATTR_INTERNAL_TRACING_ERROR);
+        if (throwable != null) {
+            ctx.getTracer().endWithResponseAndError(rootSpan, response, throwable);
+            return;
+        }
+        final Metrics metrics = ctx.metrics();
```
phiz71: **Timing: this read happens after the window in which the abort can still be classified.**

On the v4 path this `doFinally` hangs off `postProcessCompletable`, which contains `ReporterProcessor` — and `ReporterProcessor` sets `metrics.setRequestEnded(true)`. `ClientCloseClassifier.decorate` returns early on `metrics.isRequestEnded()`, so a close whose exception handler fires after the reporter has run never writes `errorKey`, we read `null`, and the span is closed on the response code alone: the original symptom.

Minor, but in the same area: that chain is `subscribeOn(Schedulers.computation())` while `decorate` writes on the event loop, and `Metrics.errorKey` is a plain non-volatile field — a write landing after the worker was scheduled is not guaranteed to be visible here.

Works for the motivating case, but the fix is timing-dependent in a way nothing in the code says. Stashing the throwable at classification time (see the comment on the method) sidesteps both.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
```diff
@@ -368,6 +354,33 @@ static void classifyResponseStreamError(final MutableExecutionContext ctx, final
         }
     }
 
+    /**
+     * Ends the request's root span, preferring a stashed failure and falling back to a classified client abort.
```
phiz71: The javadoc promises a narrower contract than the code implements: *"falling back to a classified client abort"* and *"the abort is already classified onto the metrics by `ClientCloseClassifier`, so the reason is read from there"*.

In practice the dominant writer of `metrics.errorKey` on this path is `AbstractFailureProcessor`, with `HttpConnector`, `InvokerAdapter`, `HttpProxyEndpointConnector` and `UnexpectedErrorReporter` also writing it. Someone reading only this will conclude the branch cannot fire outside a client close and will not think to check what it does to policy-rejected requests.

Either narrow the condition to match the doc, or say plainly that any classified failure ends the span in error.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/test/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcherTest.java
```diff
@@ -690,6 +693,54 @@ private void simulateEndHandlerCall(InvocationOnMock i) {
         endHandler.handle(ctx);
     }
 
+    @Nested
```
phiz71: **Nothing covers the scenario the PR fixes.**

The three tests drive the static helper against a fully mocked `MutableExecutionContext` and assert which overload the mocked `Tracer` received — that is the implementation, not the behaviour. No test dispatches a request and *disposes* it, which is the whole point: if the `doFinally` moves, is reordered relative to the post-processor chain, or stops calling `endRootSpan`, all three still pass and the 200-for-an-aborted-stream regression comes back silently.

The method javadoc says it is package-private and static "so the three call sites can be covered directly" — but the call sites are what is left uncovered.

Two guards are also unexercised, and survive being deleted: `metrics != null` and `!errorKey.isBlank()` — no test passes a null `Metrics` or a blank key.

A dispatch-level test that cancels the subscription after the classifier has tagged the metrics would be the one that fails without this change.

phiz71 (commented):
Thanks for the clear write-up — the diagnosis is right and the three call sites were the correct place to unify. The concern is the guard itself: `errorKey != null` is a much wider net than "a client went away", so the blast radius is well beyond the scope stated in the description. Details inline; the non-anchorable points are below.

**Not blocking, but worth a look**

**PR template.** `.github/pull_request_template.md` declares `## Issue`, `## Description` and `## Additional context`; none of the three is present, so the change carries no JIRA reference from the PR, the commit footer, or the branch name (`fix/http-span-client-abort`, where `AGENTS.md` asks for `<type>/<PROJECT-KEY>-<slug>`). `AGENTS.md` also asks for changes touching tracing/diagnostics to be flagged as security-sensitive in the description.

**Commit body.** It repeats the diagnosis almost verbatim from the PR description and the new javadoc. The reasoning is exactly what reviewers want in the PR; the commit body is better kept to what changes, or dropped in favour of the subject alone.

**Verified locally:** module builds, `DefaultHttpRequestDispatcherTest` green (54 tests, including the 3 new ones), `prettier:check` green.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
remibaptistegio: Confirmed, and this is the one that changed the approach. `AbstractFailureProcessor.processFailure` sets the key for every keyed interruption, and `DefaultApiReactor.processThrowable` routes an `InterruptionFailureException` into the error processor chain — which *completes*, so nothing is stashed and the branch fires. Every 401/403/429 would have flipped to `StatusCode.ERROR`.

I checked the extractor to be sure of the baseline: the root span is a SERVER span using `HttpSpanStatusExtractor.create(HttpServerAttributesGetter)`, whose SERVER converter is `code >= 500 || code < 100` — so 4xx really is UNSET today, and only the throwable path (via `SpanStatusExtractor.getDefault()`) would have overridden it.

`STREAM_ENDED_EARLY` was the same exposure. One nuance: the "WARN, not ERROR" comment quoted there is about the log level rather than the span status — but the point stands, that span would have flipped too.

Dropped the errorKey branch entirely rather than narrowing it. See the reply on the method comment.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
remibaptistegio: Taken, and done in `decorate` rather than in `classifyResponseStreamError` — for the reason you gave. I traced the motivating SSE case and it goes through `VertxHttpServerResponse`'s write path, which does `onErrorResumeNext(... -> Completable.complete())`: the error is swallowed there and the dispatch ends normally, so the response exception handler never sees it. Stashing in the dispatcher alone would have left the reported bug unfixed.

Your read of the overloads is exact — I disassembled `OpenTelemetryTracer`: the `String` overload is `span.setStatus(ERROR, msg)` then `endWithResponse` → `endSpan(..., null)`, status only. The `Throwable` overload goes to `InstrumenterTracer.endSpan(..., throwable)`, which is what emits the `exception` event and `error.type`.

Net result: the dispatcher is untouched, `endRootSpan` is gone, and the whole change is one guarded `putInternalAttribute` next to the metrics write. 61 lines instead of 130.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
remibaptistegio: Right on both counts, and moot now — the `String` overload is no longer used, so nothing lands in the status description and nothing is dropped. The cause goes through the `Throwable` path, and the key and message still reach analytics through the `Diagnostic` that `decorate` already writes.

Worth recording that `AbstractTracingHook.endSpanWithFailure` is the convention to follow if we ever do want the failure key on the root span — I looked at that while checking your comment, and `interruptWith` already stashes the `ExecutionFailure` under `ATTR_INTERNAL_EXECUTION_FAILURE`, which survives to the dispatcher's `doFinally`. So key-as-attribute would need no new plumbing. Out of scope here; noted for the follow-up.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
remibaptistegio: Verified the chain: `ReporterProcessor` sets `requestEnded(true)`, it is in `postProcessorChain` (`AbstractPlatformProcessorChainFactory:93`), and the span-ending `doFinally` hangs off that chain — so the read really could land after the window.

The ordering problem is gone, though not by escaping the guard. The stash sits inside `decorate`'s guarded block, deliberately: the span is now written exactly when and only when the metrics are, so the trace and the analytics can no longer disagree. If `decorate` bails because the request already ended, the metrics recorded a success and the span says the same thing — which is the outcome we want for a keep-alive connection dying between requests.

One correction on the visibility half: stashing does not sidestep it. Internal attributes are a plain map read from the same computation-scheduler thread, so a write landing after the worker was scheduled has no more guarantee than the `errorKey` write did. What the `subscribeOn` hand-off gives us is a happens-before for writes made *before* the chain is scheduled, which covers the abort-during-streaming case in both designs. Real but pre-existing, and not something this PR makes better or worse.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/main/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcher.java
remibaptistegio: Fair — the doc described the narrow case while the code took any classified failure. Resolved by narrowing the code instead of the prose: the method is gone, and the only thing that stashes a cause is `ClientCloseClassifier.decorate`, which by construction runs on confirmed client closes only. The javadoc there now says it decorates the metrics *and* stashes the cause for the request span, which is exactly what it does.

gravitee-apim-gateway/gravitee-apim-gateway-reactor/src/test/java/io/gravitee/gateway/reactive/reactor/DefaultHttpRequestDispatcherTest.java
remibaptistegio: Agreed — the three overload-assertion tests are gone. What replaces them:

`DefaultHttpRequestDispatcherTest.shouldEndTracingSpanInErrorWhenTheClientAbortedTheResponseStream` drives a real dispatch: the reactor tags the metrics through `ClientCloseClassifier.decorate` and then completes normally, exactly as the write-failure path does, and the test asserts the root span was ended with that very `IOException`. It fails on the pre-change code with `Expecting actual: null` at the captor, so it holds the behaviour rather than the shape of the call.

Two guard tests in `ClientCloseClassifierTest`: one that the cause is stashed, one that an already-ended request stashes nothing. The `metrics != null` and `isBlank` guards you flagged no longer exist — there is no key-reading branch left to guard.

RED → GREEN confirmed on both, and both modules are green (321 tests).

remibaptistegio:
Picking up the three non-anchorable points — thanks, I had replied only to the inline ones at first and missed these.

**Commit body.** Fair, and cut. It was three paragraphs of diagnosis that the PR already carried. Now a subject plus one sentence on what changed:

> `ClientCloseClassifier` now stashes the close cause in `ATTR_INTERNAL_TRACING_ERROR` alongside the metrics it decorates, so the dispatcher's existing span-ending path records the abort as the exception it was instead of closing the span on the response code alone.

**PR template.** Restructured onto `## Issue` / `## Description` / `## Additional context`, and added the security-sensitive flag for the tracing/diagnostics surface. The `## Issue` section still has no JIRA link — I don't want to invent a key. The code this builds on cites APIM-12769 for the client-close classification, and this extends the same work from analytics to tracing, so that may well be the right ticket, but I've left it as a TODO rather than assert it. Same for the branch name: `fix/http-span-client-abort` misses the project key, and renaming it now would mean closing this PR and opening another, so I'd rather fix the convention on the next one unless you'd prefer otherwise.

**Verified locally.** Note the test counts moved — the three overload-assertion tests are gone, replaced by one dispatch-level test in `DefaultHttpRequestDispatcherTest` and two guard tests in `ClientCloseClassifierTest`. Both gateway modules green (321 tests), `prettier:check` green.

phiz71 (approved):
Much better — and the right level. Moving the stash into `ClientCloseClassifier.decorate` reduces the production change to 11 lines in one file and leaves `DefaultHttpRequestDispatcher` byte-identical to master, which is the strongest thing that can be said about a fix like this. Sitting the stash inside the existing guards is the detail that makes it hold together: the span and the metrics are now written under exactly the same conditions, so a trace and its analytics cannot disagree in either direction.

Ending through the `Throwable` overload is also what makes the span usable — the `exception` event and `error.type` are what an APM actually breaks errors down by, and the earlier status-description approach produced neither.

**Checked locally on `a6e16b62cd`**

- Both touched modules green: gateway-reactor 321 tests, gateway-http 144. `prettier:check` green.
- Mutation check: removing the `putInternalAttribute` block fails `shouldEndTracingSpanInErrorWhenTheClientAbortedTheResponseStream` and `should_stash_the_close_cause_for_the_request_span`, and nothing else — 52 dispatcher tests run, exactly 1 failure. The new tests do catch the regression.
- One thing I went looking for and did not find: `decorate` writes `ATTR_INTERNAL_TRACING_ERROR` unconditionally, so in principle it could overwrite a cause the dispatcher's own `doOnError` had already stashed. It cannot in practice — `DefaultApiReactor` catches everything through `chainWithOnError(t -> handleUnexpectedError(...))` before `response().end()`, so `UnexpectedErrorReporter` has always attributed a key by the time anything reaches the dispatcher's `doOnError`, and that key returns `decorate` early at the `getErrorKey() != null` guard. Worth knowing that the guard is load-bearing for the span too, not just for the metrics.

**Before merge, not blocking review:** `## Issue` still holds a TODO rather than a JIRA link, and there is no ticket in the commit footer either, so the change is untraceable to one from anywhere. (The branch name is missing the project key too, but that ship has sailed for this PR.)

Nit: the description says "both gateway modules green (321 tests)" — 321 is gateway-reactor alone; gateway-http adds 144.

PR #19538, opened 2026-09-01
chore(deps): bump gravitee-node to 9.8.1

josephlomasney:
## Issue

https://gravitee.atlassian.net/browse/CJ-4307

## Description

Picks up gravitee-node 9.8.1, which lets the OTLP exporter read `headers` given as `name`/`value` pairs. Cockpit's OTel reporter emits headers in that shape, and the exporter previously read the pair as two headers literally called `name` and `value`.

Also aligns the hazelcast node plugin pins in `helm/values.yaml` and the federation helm test with the same version.

## Additional context

Delta from the currently pinned 9.8.0 is a single change: gravitee-io/gravitee-node#604.

Labelled `apply-on-4-12-x` so mergify opens the 4.12.x copy.

### DIFF
--- a/helm/tests/api/deployment_federation_test.yaml
+++ b/helm/tests/api/deployment_federation_test.yaml
@@ -37,7 +37,7 @@ tests:
             - command:
                 - sh
                 - -c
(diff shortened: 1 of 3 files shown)


gcusnieux:
@josephlomasney Closing, master and 4.12.x are already on gravitee-node 9.9.0. Reopen if I missed something.

PR #19547, opened 2026-09-01
fix(gateway): let the FIPS image start with an mTLS plan

ThibaudAV:
## Issue

https://gravitee.atlassian.net/browse/ESM-109

## Description

The 4.12.x FIPS gateway image cannot complete its initial sync when any mTLS plan exists. Reported by the FAA team against `4.12.16-FIPS`; a single published mTLS plan anywhere in the environment keeps every gateway from turning ready.

Six commits. Three distinct causes behind the blocker, the chart gap the same customer hit while working around it, their tests, and a quick-setup that reproduces the whole thing:

**1. The FIPS image ships two BouncyCastle distributions.** The FIPS base image puts `/usr/share/java/bouncycastle-fips` on the JVM `--module-path`, but `bin/gravitee` starts with `-cp <bootstrap jar>`, dropping the `CLASSPATH` the image exports. Only `bc-fips` gets resolved that way, so the extension class loader over `lib/ext` takes `org.bouncycastle.asn1.bc` from `bc-fips` while `org.bouncycastle.cms` and `.operator` still come from the non-FIPS 1.84 jars sitting there. Both distributions publish those package names, the classes link across the two, and the first client-certificate subscription blows up with `NoSuchFieldError: BCObjectIdentifiers.sphincsPlus_sha2_128s_r3`.

Note that *deleting* the non-FIPS jars is not the fix — `bcpkix-fips` is on the module path but nothing resolves it, so `org.bouncycastle.cms` simply becomes unreachable. `lib/ext` has to carry the FIPS artifacts instead.

**2. `SubscriptionTrustStoreLoader` hardcoded `PKCS12`.** It was the only store in the codebase not going through `KeyStoreUtils.DEFAULT_KEYSTORE_TYPE`. On the java 21 FIPS base this is survivable (SUN still answers for PKCS12, only serialization fails); on the **java 25 FIPS base that master builds against there is no PKCS12 KeyStore at all** and `getInstance` throws outright, so every mTLS subscription fails to register. JKS is no fallback there either — BC-FIPS answers for it read-only.

**3. A `LinkageError` killed the sync deployer thread.** The per-subscription catch in `SubscriptionDeployer` only covered `Exception`, so the error escaped the runnable and took down `gio.sync-deployer`. Initial sync stopped partway through and the node never turned ready. This one is not FIPS-specific: whatever the cause, one unusable subscription should cost that subscription and nothing more.

## Additional context

Reproduced and verified against the real base images (`graviteeio.azurecr.io/java:{21,25}-chainguard-fips`), by rebuilding the two-level class loader `Bootstrap` sets up over `lib/ext`:

| lib/ext contents | `BCObjectIdentifiers` from | `CMSSignedData` from | Result |
|---|---|---|---|
| non-FIPS 1.84 *(today)* | `bc-fips.jar` | `bcpkix-jdk18on-1.84.jar` | `NoSuchFieldError` — matches the reported stack trace verbatim |
| *(emptied)* | `bc-fips.jar` | not found | `ClassNotFoundException` |
| **`bcpkix-fips` + `bcutil-fips`** | `bc-fips.jar` | `bcpkix-fips.jar` | **links cleanly** |

The Dockerfile change was then built both ways against the real bases: the FIPS image ends up with `bcpkix-fips.jar` + `bcutil-fips.jar` in `lib/ext` and links `CMSSignedData` cleanly, the non-FIPS image keeps its 1.84 jars untouched. The same guard is applied to the rest-api image, and that one is not precautionary: on a full stack (mongo + management API + gateway, licensed), registering a client certificate on an application against the **unpatched** management API FIPS image returns HTTP 500, with `NoSuchFieldError ... sphincsPlus_sha2_128s_r3` thrown from `MtlsSubscriptionSyncDomainServiceImpl.updateActiveMTLSSubscriptions` — the same field as the reported gateway crash. With the patched image the same call returns 201. So an operator on the FIPS images cannot register or rotate an mTLS client certificate at all, independently of the gateway.

On that same stack, with an mTLS plan, an application carrying the customer's certificate and an accepted subscription: the patched gateway registers both subscription certificates in its dynamic trust store and deploys the API, while the published image never gets there — it dies at boot in `KeyStoreUtils.initSelfSigned`, a third path through the same mixed BouncyCastle classes, and logs `Unable to start HTTP server`.

Then end to end, against the **published** `apim-gateway:master-latest-chainguard-fips` (java 25), driving the real production path — `SubscriptionTrustStoreLoader.readSubscriptionCertificate` followed by the loader construction — through the image's own class loader chain, with the customer's actual client certificate:

| image | read the certificate | build the trust store | mTLS subscription |
|---|---|---|---|
| published today | `NoSuchFieldError` — the reported error, verbatim | — | **fails** |
| + BouncyCastle fix only | ok, 1 certificate | `KeyStoreException: PKCS12 not found` | **fails** |
| + both fixes | ok, 1 certificate | ok — `BCFKS` / `BCFIPS` | **registers** |

The middle row is why both commits ship together: fixing the jars alone moves the failure rather than removing it. Both images still boot and complete initial sync with no mTLS subscription present, which is exactly why this went unnoticed — the broken classes link lazily, on the first subscription carrying a client certificate.

Keystore types actually available under `approved_only`, measured with the real client certificate:

| | java 21 FIPS | java 25 FIPS |
|---|---|---|
| PKCS12 | in-memory ok, `store()` fails | **absent** |
| JKS | ok (non-approved crypto) | read-only |
| BCFKS | ok | ok |

**4. The chart could not express a PEM keystore on the HTTP listeners.** Neither `gateway.ssl.keystore` nor `gateway.servers[].ssl.keystore` emitted `certificates[]`, so the only keystore type that is sound under `approved_only` was also the only one that had to be configured outside the chart, through raw indexed environment variables. `gateway.kafka.ssl.keystore` and `gateway.ratelimit.redis.keystore` already emit it in the same template; this aligns the listeners with them.

Rendered with `helm template` against the real chart:

| values | before | after |
|---|---|---|
| defaults | 166 lines | **identical** |
| `gateway.ssl.keystore` JKS, no certificates | — | **identical** |
| `gateway.servers[].ssl.keystore` PEM with 2 certificates | silently dropped | emitted |

No guard clause failing a PEM keystore without certificates, deliberately: deployments already feeding those entries through environment variables would start failing to template.

**Release note needed.** `org.bouncycastle.jce.provider.BouncyCastleProvider` no longer resolves inside the two `-chainguard-fips` images. A constant-pool sweep of every jar in both distributions (184 for the gateway, 305 for the rest-api) found no Gravitee code affected, but four shipped third-party capabilities lose their BouncyCastle backing on the FIPS images only: MSSQL PEM client-certificate authentication and Always Encrypted, the MariaDB `parsec` (Ed25519) authentication plugin, PDFBox encrypted and signed PDF handling, and — present but unreferenced by anything shipped — the `nimbus` BouncyCastle provider singleton and the spring-security `Argon2`/`SCrypt`/`BouncyCastleAes*` encoders. None of them can work on a FIPS image today either, since the class path they need is the broken one, but the change makes the failure explicit rather than intermittent.

**Still open, not in this PR** — lives in another repo, so it cannot ride along: BCFKS is not an accepted keystore/truststore type (`KeyStoreLoader` in gravitee-node offers only JKS / PEM / PKCS12 / SELF-SIGNED / PEM-FOLDER / KUBERNETES-PEM-REGISTRY, and `VertxServerOptions.DEFAULT_STORE_TYPE` is JKS). Under `approved_only` that leaves PEM as the only usable option today, and the table above shows it becomes the *only* one at all on java 25. Worth its own ticket.


gravitee-apim-gateway/docker/Dockerfile.chainguard
```diff
@@ -30,6 +30,36 @@ RUN apk add --no-cache libjemalloc2 \
...
+# NoSuchFieldError on BCObjectIdentifiers.sphincsPlus_sha2_128s_r3, which kills the sync deployer
+# thread as soon as a subscription carries a client certificate, and the node never becomes ready.
+# Swapping lib/ext over to the FIPS artifacts keeps every BouncyCastle class coming from a single
+# distribution. The guard makes this a no-op on the non-FIPS base image, which has no such directory.
+# JAVA_FIPS_CLASSPATH is set by the FIPS base image alone and names where it keeps those jars.
+# BASE_IMAGE decides which variant is being built, so the FIPS branch is entered on that rather
+# than on the variable: a base image that stops exporting it, moves the jars, or renames the
+# non-FIPS ones fails the build here instead of silently shipping the broken mix again.
+ARG BASE_IMAGE
+RUN case "${BASE_IMAGE}" in \
```
phiz71: The swap is gated on the `BASE_IMAGE` *string*. Pin the FIPS base by digest — `--build-arg BASE_IMAGE=graviteeio.azurecr.io/java@sha256:...`, an ordinary supply-chain hardening step — or rename the tag, and `*fips*` stops matching: the whole branch is skipped, none of the guards below ever fire, the build succeeds, and the image ships the mixed BouncyCastle again. That is precisely the silent regression the comment above says this prevents.

`JAVA_FIPS_CLASSPATH` is the actual capability signal, and you already read it on the next line. Entering the branch on it — while keeping a "`BASE_IMAGE` says fips but the variable is missing -> fail" check — gets you the loud failure without depending on how the tag happens to be spelled.

gravitee-apim-gateway/docker/Dockerfile.chainguard
```diff
@@ -30,6 +30,36 @@ RUN apk add --no-cache libjemalloc2 \
...
+# The FIPS base image ships BouncyCastle FIPS under /usr/share/java/bouncycastle-fips and puts that
+# directory on the JVM --module-path, but bin/gravitee starts the JVM with `-cp <bootstrap jar>`,
+# which drops the CLASSPATH the image exports. Only bc-fips ends up resolved that way, so the
+# extension class loader over lib/ext resolves org.bouncycastle.asn1.bc from bc-fips while
+# org.bouncycastle.cms and .operator still come from the non-FIPS jars sitting there. Both
+# distributions publish those package names, so classes link across the two and fail on first use:
+# NoSuchFieldError on BCObjectIdentifiers.sphincsPlus_sha2_128s_r3, which kills the sync deployer
+# thread as soon as a subscription carries a client certificate, and the node never becomes ready.
+# Swapping lib/ext over to the FIPS artifacts keeps every BouncyCastle class coming from a single
+# distribution. The guard makes this a no-op on the non-FIPS base image, which has no such directory.
```
phiz71: This says the no-op comes from the absent `/usr/share/java/bouncycastle-fips` directory, but four lines down the same comment explains that the branch is entered on `BASE_IMAGE` "rather than on the variable" — and that is what the code actually does. Reads like a leftover from an earlier directory test; the body of `ed9aa84` carries the same claim ("The directory test makes this a no-op").

Worth more than tidiness: someone reading this line will believe the swap is self-detecting, and will not realise that a renamed or digest-pinned base disables it. Same text at `gravitee-apim-rest-api/docker/Dockerfile.chainguard:36`.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/SubscriptionDeployer.java
```diff
@@ -83,6 +83,16 @@ public Completable deploy(final SubscriptionDeployable deployable) {
                             }
                             subscriptionService.register(subscription);
                             log.debug("Subscription [{}] deployed for api [{}] ", subscription.getId(), subscription.getApi());
+                        } catch (LinkageError e) {
```
phiz71: `LinkageError` covers the `NoSuchFieldError` you reproduced, but not the whole `Error` family BC-FIPS raises. `org.bouncycastle.crypto.fips.FipsUnapprovedOperationError` extends `FipsOperationError` extends `java.lang.AssertionError` — I checked this against `bc-fips` 2.1.2. Under `approved_only`, a subscription certificate whose key size or signature algorithm is not approved raises that while the trust store is being built.

RxJava's `Exceptions.throwIfFatal` only rethrows `VirtualMachineError`, `ThreadDeath` and `LinkageError`, so an `AssertionError` is routed to `onError` instead: `deploy()` returns an errored `Completable`, the deployable's remaining subscriptions are never registered, and the failure propagates into the synchronizer. Not a dead thread this time, but not "one subscription and nothing more" either — which is the invariant the commit body states.

Catching `Error` and rethrowing `VirtualMachineError` / `ThreadDeath` would match the stated intent, and would keep the "`OutOfMemoryError` and the like stay uncaught on purpose" note true.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductSubscriptionRefresher.java
```diff
@@ -153,6 +153,13 @@ public Completable unregisterByApiProduct(final String apiProductId) {
     private void quietly(final Runnable action, final String itemId, final String type) {
         try {
             action.run();
+        } catch (LinkageError e) {
```
phiz71: Same point as in `SubscriptionDeployer`: `LinkageError` alone lets the BC-FIPS `AssertionError` family (`FipsUnapprovedOperationError`) through this guard. Whatever you settle on over there should apply here too, so the two paths keep the same contract.

gravitee-apim-rest-api/docker/Dockerfile.chainguard
```diff
@@ -24,6 +24,36 @@ ENV GRAVITEEIO_HOME=/opt/graviteeio-management-api
 FROM base AS builder
 COPY --chown=graviteeio:graviteeio ./distribution ${GRAVITEEIO_HOME}/
 USER root
+# The FIPS base image ships BouncyCastle FIPS under /usr/share/java/bouncycastle-fips and puts that
+# directory on the JVM --module-path, but bin/gravitee starts the JVM with `-cp <bootstrap jar>`,
+# which drops the CLASSPATH the image exports. Only bc-fips ends up resolved that way, so the
+# extension class loader over lib/ext resolves org.bouncycastle.asn1.bc from bc-fips while
+# org.bouncycastle.cms and .operator still come from the non-FIPS jars sitting there. Both
+# distributions publish those package names, so classes link across the two and fail on first use:
+# NoSuchFieldError on BCObjectIdentifiers.sphincsPlus_sha2_128s_r3, which kills the sync deployer
```
phiz71: This block is copied verbatim from the gateway Dockerfile, symptom included — but the management API runs no sync deployer and has no readiness tied to one. Per your own testing, its symptom is the HTTP 500 out of `MtlsSubscriptionSyncDomainServiceImpl.updateActiveMTLSSubscriptions`. Worth saying that here instead, so whoever debugs a FIPS management API is not sent looking for machinery this image does not have.

phiz71 (commented):
Reviewed the six commits end to end. The diagnosis holds up on my side: I re-derived the class-loading story from the distribution assembly (`lib/ext` takes every non-Gravitee dependency, and `bcpkix-jdk18on` is the only direct BouncyCastle dependency in the reactor, so the three `rm -f` globs really do cover what is there), confirmed that `KeyStoreUtils.DEFAULT_KEYSTORE_TYPE` is already what `gravitee-node`'s `AbstractKeyStoreLoaderManager` uses for the merged store — so `SubscriptionTrustStoreLoader` was indeed the odd one out — and ran the Helm suite locally (14/14 green, and the new PEM blocks render as expected on both the default listener and `servers[]`).

The release note's blast radius also looks right: outside `gravitee-apim-gateway-tests-sdk` and the integration tests, no Gravitee code references `org.bouncycastle.jce.provider`, and every password encoder in the rest-api is `BCryptPasswordEncoder`, which is pure Java.

Two things I would want addressed before merge, plus a few smaller ones. Inline comments carry the detail; below is what does not fit on a changed line.

## Must

**The undeploy paths keep the `catch (Exception)` that the deploy path just outgrew** — `SubscriptionDeployer#undeployForApi` (L146) and `#undeploySingleSubscription` (L161), and `ApiProductSubscriptionRefresher#unregisterByApiProduct`, which has no `try`/`catch` at all. `subscriptionService.unregister(...)` reaches `SubscriptionTrustStoreLoaderManager#unregisterSubscriptionTrustStoreLoader`, which fires a `KeyStoreEvent.UnloadEvent`; `gravitee-node`'s `AbstractKeyStoreLoaderManager` then rebuilds the merged store (`KeyStore.getInstance(KeyStoreUtils.DEFAULT_KEYSTORE_TYPE)` plus an entry copy, L247-248) — the same mixed-BouncyCastle path. A `LinkageError` raised there escapes the `catch (Exception)`, and because `Completable.defer` runs its supplier inside `subscribeActual` where `throwIfFatal` rethrows it, it takes the thread down exactly as the deploy path did before this fix. Same class of hole, same file, opposite direction.

## Should

**A PUSH subscription is queued for dispatch before `register()` succeeds** — `SubscriptionDeployer#deploy` L73: `dispatchableSubscription.compute(...)` sits inside the same `try`, ahead of `register()`. A PUSH subscription whose registration fails is still drained by `doAfterDeployment` and dispatched, while absent from the subscription cache and with no trust store loader registered. Pre-existing for `Exception`, but the new catch extends it to the `LinkageError` case rather than closing it. Moving the `compute` after a successful `register()` keeps the two in step.

**The guard may be one level too high** — `SubscriptionTrustStoreLoaderManager` already owns the "a bad certificate must not break registration" policy: it catches `MalformedCertificateException`, logs it, and rolls back the loaders it just added (L77-84); `readSubscriptionCertificate` and the loader constructor already wrap `Exception` into that exception. Widening *those* rather than the two deployer loops would cover both current callers and any future one from a single place, and would spare us the same eight-line rationale appearing in two unrelated deployers. Not blocking — the deployer-level catch is also defensible since it protects against anything else `register()` may throw — but worth a deliberate call.

## Nit

Purely personal preference, ignore freely: the commit bodies read as investigation write-ups (diagnosis, options ruled out, measurements). All of it is already in the PR description, which is where I would keep it — the commits themselves tend to age better when they only say what changes.

gravitee-apim-rest-api/docker/Dockerfile.chainguard
```diff
@@ -24,6 +24,36 @@ ENV GRAVITEEIO_HOME=/opt/graviteeio-management-api
...
+# thread as soon as a subscription carries a client certificate, and the node never becomes ready.
+# Swapping lib/ext over to the FIPS artifacts keeps every BouncyCastle class coming from a single
+# distribution. The guard makes this a no-op on the non-FIPS base image, which has no such directory.
+# JAVA_FIPS_CLASSPATH is set by the FIPS base image alone and names where it keeps those jars.
+# BASE_IMAGE decides which variant is being built, so the FIPS branch is entered on that rather
+# than on the variable: a base image that stops exporting it, moves the jars, or renames the
+# non-FIPS ones fails the build here instead of silently shipping the broken mix again.
+ARG BASE_IMAGE
```
phiz71, line 41: These 30 lines are byte-for-byte identical to the gateway Dockerfile's, with nothing keeping them in sync — and they have already drifted semantically (see the comment just above). Any later change to the swap has to land twice, and a miss ships one of the two images with the broken mix while the other is fine.

A small script that both build contexts `COPY` in would remove the sync burden. Fine to leave as is if you would rather not touch the contexts, but it is worth a conscious call rather than an accident.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/test/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductSubscriptionRefresherTest.java
```diff
@@ -120,6 +121,25 @@ void loads_and_deploys_subscriptions() throws TechnicalException {
...
         }
 
+        /**
+         * Registering a client certificate links BouncyCastle classes lazily, so a mismatched provider on the
+         * classpath surfaces as an Error rather than an Exception. This runs inside a Completable.fromRunnable,
+         * which rethrows Errors instead of routing them to onError, so an uncaught one kills the sync thread and
+         * the node never turns ready. An API Product subscription reaches register() through quietly().
+         */
+        @Test
+        void keeps_going_when_a_subscription_fails_with_a_linkage_error() throws TechnicalException {
```
phiz71: The name promises more than the test checks: a single subscription, and `apiKeyRepository.searchAfter` left unstubbed so `loadApiKeys` returns empty. Nothing runs after the failing item, so these assertions would still pass if `quietly()` aborted the whole page on the first failure. The `SubscriptionDeployer` sibling gets this right with two subscriptions.

A second subscription — and ideally an API key — would actually pin the "keeps going" property. `quietly()` is also the guard for `apiKeyService.register`, and no test currently drives that branch with an `Error`.

docker/quick-setup/fips-mtls/setup-mtls-api.sh
```diff
@@ -0,0 +1,65 @@
...
+  "listeners":[{"type":"HTTP","paths":[{"path":"/fips-mtls"}],"entrypoints":[{"type":"http-proxy"}]}],
+  "endpointGroups":[{"name":"default","type":"http-proxy","endpoints":[{"name":"backend","type":"http-proxy",
+    "weight":1,"inheritConfiguration":false,"configuration":{"target":"https://api.gravitee.io/echo"}}]}]
+}' | jsonf id)
+
+echo "2/6 publishing an mTLS plan"
+PLAN_ID=$(curl -s "${AUTH[@]}" -X POST "$MAPI/apis/$API_ID/plans" -d '{
+  "definitionVersion":"V4","name":"mtls-plan","description":"mTLS","security":{"type":"MTLS"},
+  "mode":"STANDARD","validation":"AUTO","flows":[]}' | jsonf id)
+curl -s "${AUTH[@]}" -X POST "$MAPI/apis/$API_ID/plans/$PLAN_ID/_publish" -o /dev/null
```
phiz71: `--fail-with-body` writes the server's explanation to stdout and `-o /dev/null` throws it away. Publish this plan against a stack without an EE licence and the script stops right after `2/6 publishing an mTLS plan` having printed nothing at all — same for steps 4, 5 and 6.

Steps 1-3 lose it a different way: `curl ... | jsonf id` without `set -o pipefail` hides curl's status and feeds the error body into python, which exits 1 — again silently. For a script whose whole purpose is reproducing FIPS failures, every failure is currently mute. Dropping `-o /dev/null` (or sending the body to stderr) plus `set -o pipefail` would cover both.

docker/quick-setup/fips-mtls/setup-mtls-api.sh
```diff
@@ -0,0 +1,65 @@
...
+
+MAPI="http://localhost:8083/management/v2/environments/DEFAULT"
+MAPI_V1="http://localhost:8083/management/organizations/DEFAULT/environments/DEFAULT"
+AUTH=(-u admin:admin -H "Content-Type: application/json" --fail-with-body)
+# Exits non-zero on a missing key: an empty id would otherwise build URLs like /apis//plans
+# and let the script report success while nothing was created.
+jsonf() { python3 -c "import sys,json;v=json.load(sys.stdin).get('$1');sys.exit(1) if not v else print(v)"; }
+
+echo "0/6 waiting for the management API"
+until curl -sf -o /dev/null "http://localhost:8083/management/v2/ui/bootstrap"; do sleep 2; done
```
phiz71: No bound on this loop. If the management API never comes up — missing licence, unreachable Mongo, or the very BouncyCastle failure being reproduced — the script hangs forever on `0/6 waiting for the management API` with no hint of why. A cap (150 attempts, say) plus printing the last curl error before giving up would make the common case debuggable.

gravitee-apim-gateway/gravitee-apim-gateway-security/gravitee-apim-gateway-security-core/src/main/java/io/gravitee/gateway/security/core/SubscriptionTrustStoreLoader.java
```diff
@@ -49,7 +49,9 @@ public SubscriptionTrustStoreLoader(SubscriptionCertificate subscriptionCertific
         try {
             // using hex for fingerprint, as uppercase letters found in base64 are lower cased in KeyStore aliases, cause unregistering issues
             this.id = formatLoaderId(subscriptionCertificate);
-            keystore = KeyStore.getInstance("PKCS12");
+            // Not a hardcoded PKCS12: under a BC-FIPS "approved only" JVM the PKCS12 key derivation has no
+            // provider, so KeyStoreUtils resolves the default to BCFKS there and to PKCS12 everywhere else.
+            keystore = KeyStore.getInstance(KeyStoreUtils.DEFAULT_KEYSTORE_TYPE);
             keystore.load(null, getPassword().toCharArray());
             keystore.setCertificateEntry("cert", subscriptionCertificate.certificate());
         } catch (Exception e) {
```
phiz71, line 57: Not introduced here, but it is the constructor you are touching: `this.id` is assigned as the first statement of the `try`, and this catch formats its message from `this.id()`. If `formatLoaderId` is what throws — `Base64URL.from(fingerprint).decode()` on a fingerprint that is not valid Base64URL, or a null subscription id — the final field is still unset and the message reads `... for loader null`. Definite-assignment analysis misses it because the read goes through the method rather than the field.

Formatting from `subscriptionCertificate` directly would keep the identifier in the one message that actually needs it.

ThibaudAV:
Both MUSTs were right, and the first one is the more serious of the two. Pushed, six commits, all rebased.

**MUST — the `AssertionError` family.** Verified before acting: `javap` on bc-fips 2.1.1 gives `FipsUnapprovedOperationError extends FipsOperationError extends java.lang.AssertionError`, so `catch (LinkageError)` misses it and `throwIfFatal` does not treat it as fatal. The invariant my commit body claimed was therefore false, and this is not a theoretical case for us — a subscription certificate outside the approved set is exactly what a FIPS deployment produces.

**MUST — the undeploy paths.** Also verified rather than taken on trust, because a reviewer had told me these were low-risk on the grounds that unregistering never re-parses a certificate. That is true and beside the point: `AbstractKeyStoreLoaderManager` handles the unload event with `removeKeyStore` then `refreshX509Manager(clone(mainKeyStore))`, and `clone()` builds a keystore and copies entries. You traced one hop further than we did.

Rather than repeat the rationale in two deployers — your SHOULD — both now go through a `SyncIsolation` helper that catches `Throwable` and rethrows `VirtualMachineError`, with the reasoning written once. Five call sites: deploy, both undeploy paths, `quietly`, and `unregisterByApiProduct`, which had no guard at all.

One deviation: I left `ThreadDeath` out of the rethrow. It is deprecated for removal and unraisable since `Thread.stop` went, so naming it buys a deprecation warning and nothing else.

**SHOULD — PUSH queued before `register()`.** Fixed, the `compute` now runs after a successful registration.

**SHOULD — the guard one level too high.** Deliberate call to keep it at the deployer level: it protects against anything `register()`/`unregister()` throws, not only the trust store, and the shared helper removes the duplication that made the alternative attractive.

**The `BASE_IMAGE` string gate.** You were right and I reproduced it. The branch is now entered on `JAVA_FIPS_CLASSPATH` — the capability itself — with `BASE_IMAGE` kept for the reverse check you suggested. Verified against a digest-pinned base, which is the case that silently broke before:

```
--build-arg BASE_IMAGE=graviteeio.azurecr.io/java@sha256:d8ed5d3a…  →  bcpkix-fips.jar bcutil-fips.jar
```

Also fixed: the stale "no such directory" comment (leftover from the first version, as you guessed), and the rest-api block now describes its own symptom — the 500 out of `MtlsSubscriptionSyncDomainServiceImpl` — instead of a sync deployer that image does not run.

**The duplicated 30 lines.** Conscious call, left as is: sharing them means a script both build contexts `COPY` in, and the contexts are assembled by CI. Noted as a follow-up rather than smuggled into this PR.

**The test that promised more than it checked.** Correct, and it is the second time in this PR that a test of mine passed for the wrong reason. Both now run a second item that has to still be processed, the API key branch of `quietly()` is driven, and there is a case per error shape. The negative check, with the guard narrowed back to `LinkageError`:

```
should_ignore_subscription_failing_with_an_assertion_error  FAILURE
  Not completed (latch = 0, values = 0, errors = 1, completions = 0)
```

which is precisely the "errors the Completable instead" mechanism you described.

**`setup-mtls-api.sh`.** All three taken: `set -eo pipefail`, a bounded wait with the last error printed and a pointer to `docker compose logs`, and a `call()` wrapper that keeps the server's explanation instead of sending it to `/dev/null`.

**`SubscriptionTrustStoreLoader` `this.id`.** Fixed — the message is formatted from the subscription now.

**Nit on the commit bodies.** Taken, here and on gravitee-node#606 where you made the same point. They no longer restate the description.

682 tests green in the sync module, 50 in security-core.

phiz71 (commented):
Went through the force-push against `74d2ab3` — same base, so this is a clean amend of the six commits. Ten of my twelve points are addressed, and several of them better than what I suggested: `SyncIsolation` is a nicer home for the guard than widening the two call sites, and inverting the Docker trigger onto `JAVA_FIPS_CLASSPATH` with `BASE_IMAGE` as the reverse check is exactly right. I also checked that the precise-rethrow form in `SyncIsolation` (`catch (Throwable t) { throw t; }` with no `throws` clause) compiles on JDK 21 — it does, `Runnable#run` declares nothing checked so the compiler narrows it to `RuntimeException | Error`.

Resolving those threads. Two things the amend introduced, though, and one of them I would not merge as is.

## Must

**`unregisterByApiProduct` now swallows what used to propagate** — inline comment on `ApiProductSubscriptionRefresher`. Short version: `ApiProductDeployer#undeploy` (L103) awaits that `Completable` inside a `try` that turns any `Exception` into a `SyncException`. The method previously had no guard, so that contract worked. Now the failure is a `warn`, the undeploy carries on and logs success, and the product is marked gone while its subscription certificates are still in the dynamic trust store and its API keys still in the cache. The thread-kill this PR is about is an `Error` problem; on this path only `Error` needs trapping.

## Should

**The systemic signal is gone** — inline comment on `SyncIsolation`. The previous revision logged `LinkageError` at `error` with "Class path is inconsistent" and `Exception` at `warn`, and the commit body argued for that split. All five sites now share one `warn` lambda, so a gateway with the mixed BouncyCastle jars completes its sync, passes readiness, and reports a total mTLS outage as warnings. The symptom that made this bug findable in the first place — the node never turning ready — is gone with nothing in its place.

## Nits

- `SubscriptionDeployerTest`: the new test landed between the old one's javadoc and its method, see inline.
- `ApiProductSubscriptionRefresherTest#keeps_going_when_a_subscription_fails_with_a_linkage_error` now injects an `AssertionError` too; the name only mentions the linkage error.
- `undeploySingleSubscription`'s `AtomicBoolean` would disappear if `SyncIsolation.isolate` returned whether the action completed. Purely cosmetic.

Left the Dockerfile duplication thread open — it is your call, not something the amend was meant to settle.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/ApiProductSubscriptionRefresher.java
```diff
@@ -143,19 +143,20 @@ public Completable unregisterByApiProduct(final String apiProductId) {
         if (apiProductId == null) {
             return Completable.complete();
         }
-        return Completable.fromRunnable(() -> {
-            subscriptionService.unregisterByApiProductId(apiProductId);
-            apiKeyService.unregisterByApiProductId(apiProductId);
-            log.debug("Unregistered subscriptions and API keys of API Product [{}]", apiProductId);
-        });
+        return Completable.fromRunnable(() ->
+            SyncIsolation.isolate(
```
phiz71: This one I would not merge as is: it changes behaviour beyond the Error problem the PR is about.

`ApiProductDeployer#undeploy` (L103) does `subscriptionRefresher.unregisterByApiProduct(apiProductId).blockingAwait()` inside a `try` that wraps any `Exception` into a `SyncException`. That contract worked precisely *because* this method had no guard. With the isolation in place, a failing `subscriptionService.unregisterByApiProductId` is logged at `warn`, the `Completable` completes, `undeploy` moves on to `apiProductManager.unregister`, logs `"API Product [...] undeployed successfully"`, and the synchronizer records the product as gone.

The subscriptions and API keys are still in the runtime caches at that point — so the product's mTLS certificates stay in the dynamic trust store and its API keys keep authenticating, against a product the gateway believes it has dropped. And nothing retries, because the step reported success.

The thread-kill you are fixing is an `Error` problem. Here I would either leave the method unguarded (its caller already has an error path) or trap `Error` only and let `Exception` keep propagating — a variant of `isolate` that rethrows as `SyncException` rather than swallowing would serve this path without losing the guarantee.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/SyncIsolation.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+final class SyncIsolation {
+
+    private SyncIsolation() {}
+
+    /**
+     * @param action the work to run for a single item
+     * @param onFailure receives whatever the action threw, for the caller to log with its own context
+     */
+    static void isolate(final Runnable action, final Consumer<Throwable> onFailure) {
```
phiz71: One thing got lost in the extraction: the previous revision logged a `LinkageError` at `error` with "Class path is inconsistent, cannot deploy subscription [...]" and kept `warn` for an `Exception`, and the commit body defended the split — "the class path is inconsistent, so every subscription is about to fail the same way, that is not the same event as a single unusable subscription". All five call sites now pass a single `warn` lambda.

That matters more now than it did before the fix. With the isolation working, a gateway carrying the mixed BouncyCastle jars completes its initial sync and passes its readiness probe, serving zero mTLS subscriptions, and says so only in WARN lines — a level plenty of production setups filter out. The symptom that made this bug findable at all was the node never turning ready; that is gone, and nothing replaced it.

Branching on `t instanceof Error` — here, or in the lambdas — would keep "the class path is broken, every item is about to fail the same way" distinguishable from "this one item is unusable".

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/test/java/io/gravitee/gateway/services/sync/process/common/deployer/SubscriptionDeployerTest.java
```diff
@@ -129,6 +129,51 @@ void should_ignore_subscription_in_error() {
             verifyNoMoreInteractions(subscriptionService);
         }
 
+        /**
```
phiz71: The new test landed between this javadoc and the method it documented, so there are now two `/** */` blocks in a row: this one describes the linkage-error case but the compiler attaches the *second* one to `should_ignore_subscription_failing_with_an_assertion_error`, and `should_ignore_subscription_failing_with_a_linkage_error` below has lost its documentation entirely.

While you are in there: the two tests are identical apart from the `Throwable` they inject, which is what a `@ParameterizedTest` over `new NoSuchFieldError(...)` and `new AssertionError(...)` is for — one body, both shapes, and the javadoc has one obvious place to live.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/SyncIsolation.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Runs a step whose caller turns a failure into its own error and acts on it — undeploying an API
+     * Product, where the caller must not go on to mark the product gone.
+     *
+     * <p>Only Errors are trapped, and they are translated rather than swallowed: the thread survives, and
+     * the caller still sees the failure in the shape it already handles. Exceptions keep propagating
+     * untouched, so the existing contract is unchanged.
+     */
+    static void translateErrors(final Supplier<String> what, final Runnable action) {
```
phiz71, line 83: `translateErrors` is the only thing in this PR without a test, and it guards the path that has now been wrong twice: first no guard at all, then swallowed, now translated. A third regression here would be silent — the symptom is a product reported as undeployed while its certificates are still trusted, which nothing fails on.

One test in `ApiProductSubscriptionRefresherTest` would pin it: stub `subscriptionService.unregisterByApiProductId` to throw a `NoSuchFieldError`, then assert `cut.unregisterByApiProduct(PRODUCT_1).test().assertError(SyncException.class)` — and, ideally, a second asserting that an `Exception` still comes out unchanged, since that is the contract `ApiProductDeployer` relies on.

The `catch (SyncException e) { throw e; }` you added in `ApiProductDeployer` is uncovered too — `ApiProductDeployerTest` currently stubs the refresher to `Completable.complete()` everywhere, so nothing exercises the rethrow-without-wrapping.

gravitee-apim-gateway/gravitee-apim-gateway-services/gravitee-apim-gateway-services-sync/src/main/java/io/gravitee/gateway/services/sync/process/common/deployer/SubscriptionDeployer.java
```diff
@@ -130,26 +135,29 @@ private Completable distributeIfNeeded(final SubscriptionDeployable deployable)
     }
 
     private Completable undeployForApi(final ApiReactorDeployable apiReactorDeployable) {
```
phiz71, line 137: Not from this PR, and not something I would hold it up for — but the asymmetry is new and worth a conscious call.

This one still swallows and returns `Completable.complete()`, so a failure to unregister *every* subscription of an API leaves them in the cache while the sync records the undeploy as done. That is the same situation you just decided against for `unregisterByApiProduct`, treated the opposite way two methods apart.

The difference in shape is real — `ApiProductDeployer` has an error path to hand the failure back to, this one is called straight from `undeploy()` — so `translateErrors` may not drop in as cleanly. Mostly flagging it so the next reader does not have to work out whether the inconsistency is deliberate.

phiz71 (commented):
Re-read the whole thing on `7a6e264`. All three points from my last pass are fixed, and the `translateErrors` split is a better answer than what I suggested — the caller keeps the failure shape it already handles, and `ApiProductDeployer` rethrowing the `SyncException` instead of re-wrapping is the right detail. I rebuilt the three paths standalone to be sure of the semantics: `isolate` logs a `LinkageError` at error level without calling `onFailure` and returns `false`, an `AssertionError` reaches `onFailure` and returns `false`, and `translateErrors` turns a `NoSuchFieldError` into a `SyncException` with the cause preserved while an `IllegalStateException` passes straight through. `SyncException` being a `RuntimeException`, `blockingAwait()` rethrows it as is, so `ApiProductDeployer#undeploy` really does stop.

**On the `gravitee-node` bump** — I went looking for reasons it might not be safe on a support branch and did not find any:

- 9.9.0 is on Maven Central and both Hazelcast plugin zips answer 200 on `download.gravitee.io`, so neither the build nor the init container will go looking for something that is not published.
- The 9.8.0 → 9.9.0 delta across `node-api`, `node-certificates`, `node-vertx`, `node-container` and `node-management` is only the BCFKS work: the new constant, the four file loaders accepting the type, the provider pre-check, and one new `warn` in `DefaultKeyStoreLoaderFactoryRegistry` when a configured store finds no factory. Nothing touches the JKS/PKCS12/PEM paths, and `node-vertx` has no source change at all.
- 9.8.1 is byte-identical to 9.8.0 in every one of those modules, so jumping over it loses nothing.
- Chart side is consistent: plugin URLs, the federation test fixture, the nine `keystore`/`truststore` type comments and the changelog all moved together, and no `9.8.0` pin is left anywhere. `helm unittest` on the gateway SSL and federation suites is green, 18/18.

Worth a word to whoever owns the 4.12.19 release all the same — a minor dependency bump on a support branch is their call, even when the delta is this narrow.

Mergeable as far as I am concerned. One thing I would still like, and a few that can wait.

## Should

**The fix has no test** — inline on `SyncIsolation`. `translateErrors` and the propagation out of `unregisterByApiProduct` are the only behaviour in this PR without one, on the path that has now been wrong twice running.

## Could

- `undeployForApi` still swallows and reports success — inline. Pre-existing, but the asymmetry is new.
- `refresh` walks its pages outside the isolation: a `LinkageError` from `subscriptionMapper.to` or from the repository would still escape `Completable.fromRunnable` and kill the thread. Nothing on that path links BouncyCastle today, so this is about the invariant rather than a live bug.
- Every call site passes both a `what` supplier and an `onFailure` lambda that restate the same context. Folding the level choice into `SyncIsolation` and keeping only `what` would halve that.
- The two `SubscriptionDeployerTest` methods are still identical apart from the `Throwable` — a `@ParameterizedTest` over the two shapes would do, with one javadoc.

Resolving the three threads from the last round. The Dockerfile duplication one stays open — still your call.

gravitee-apim-rest-api/docker/Dockerfile.chainguard
```diff
@@ -24,6 +24,38 @@ ENV GRAVITEEIO_HOME=/opt/graviteeio-management-api
...
+# it keeps those jars: that is the capability itself, so it still holds when the base is pinned by
+# digest or its tag is renamed. BASE_IMAGE only serves the reverse check — a base that calls itself
+# FIPS while exporting nothing fails the build rather than silently shipping the broken mix again.
+ARG BASE_IMAGE
+RUN if [ -n "${JAVA_FIPS_CLASSPATH}" ]; then \
+        BC_FIPS_DIR=$(dirname "${JAVA_FIPS_CLASSPATH}"); \
+        for jar in bcpkix-fips.jar bcutil-fips.jar; do \
+            [ -f "${BC_FIPS_DIR}/${jar}" ] || { echo "FIPS base image provides no ${BC_FIPS_DIR}/${jar}" >&2; exit 1; }; \
+        done; \
+        ls ${GRAVITEEIO_HOME}/lib/ext/bcprov-jdk18on-*.jar >/dev/null 2>&1 || \
```
phiz71, line 47: This is what is failing the rest-api quality gate. SonarCloud reports eight issues on the PR, all in this block:

- `docker:S6570` (major, x7) — unquoted `${GRAVITEEIO_HOME}` on lines 47, 49, 50, 51, 52 and 53
- `docker:S7020` (minor) — line 45 too long

Together they put `new_maintainability_rating` at E, which fails the gate and leaves the PR `BLOCKED` for the merge queue. The Gateway project passes on identical shell because its new Java dilutes the ratio; here the Dockerfile is the only new code, so the same smell repeated seven times is enough on its own.

No behaviour is at stake — there is no whitespace in `/opt/graviteeio-management-api`, and the globs still need to stay unquoted to expand. Quoting just the variable does both:

```
ls "${GRAVITEEIO_HOME}"/lib/ext/bcprov-jdk18on-*.jar >/dev/null 2>&1 || ...
rm -f "${GRAVITEEIO_HOME}"/lib/ext/bcprov-jdk18on-*.jar ...
cp "${BC_FIPS_DIR}/bcpkix-fips.jar" "${BC_FIPS_DIR}/bcutil-fips.jar" "${GRAVITEEIO_HOME}/lib/ext/"
chmod 644 "${GRAVITEEIO_HOME}/lib/ext/bcpkix-fips.jar" "${GRAVITEEIO_HOME}/lib/ext/bcutil-fips.jar"
```

Worth applying to the gateway Dockerfile in the same move: it passes today only by accident of the ratio, and one Java-free follow-up touching that file would flip it red too. Which is also the practical argument for the duplication thread further down — this is the second time both copies need the same edit.

phiz71 (approved):
Approving — the two fixes from my last round hold up under inspection and the `gravitee-node` bump checks out on every angle I could test. Nothing below changes that.

One thing worth knowing before this goes near the merge queue, since it comes from this PR and not from the branch: **`[Gravitee.io APIM - Rest API] SonarCloud` is red because of the new `RUN` block**, and `mergeStateStatus` is `BLOCKED` as a result. The quality gate fails on `new_maintainability_rating` = E, from eight smells all inside those lines — seven `docker:S6570` (unquoted `${GRAVITEEIO_HOME}`) and one `docker:S7020` (long line). Details inline.

The Gateway Sonar project passes with the same shell in its own Dockerfile: there the block is diluted by all the new Java, while on the rest-api side it is the only new code, so one repeated smell is enough to sink the ratio. Quoting the variable is about ten minutes and does not change behaviour — there is no whitespace in `/opt/graviteeio-management-api` — so this is Sonar being strict rather than a real defect, but it is what is holding the gate.

Everything else is unchanged from my previous review: the missing test around `translateErrors` stays a should, the `undeployForApi` asymmetry and the Dockerfile duplication stay open for you to decide. None of them are worth another round trip in my view — fix the quoting, and this is good to go.

PR #19555, opened 2026-09-02
feat(portal-next): propagate visibility through Automation API surfaces

konrad-krol-incubly:
## Issue

## Description

Final step of the visibility work: exposes the visibility flag on the Automation API so operators can declare PUBLIC/PRIVATE per navigation entry from GKO manifests. Adds an optional visibility field to the four navigation-carrying schemas - `PortalNavigationPath`, `PortalListingApiEntry`, `DocumentationSpec`, `PortalLinkSpec`.

**Backward compatibility**
  
Older GKO clients won't send the field. The wire schema is intentionally left optional; when null reaches the domain boundary, resolution goes through `PortalVisibility.resolve(callerVisibility, parentVisibility) `- caller wins if set, otherwise inherit from parent, root defaults to PUBLIC.

**Note for reviewer**

This PR relocates PortalVisibility from `core.portal_page.model` to `core.portal.model` (same as `PortalArea` / `PortalId`). Otherwise we get cycle, because portal navigation structure needs to reference visibility enum.


gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalLinkSyncDomainService.java
```diff
@@ -98,8 +101,9 @@ private PortalNavigationLink upsert(
             rejectIfSegmentTakenByForeignItem(auditInfo, parent, segment, linkId, location);
         }
 
+        var visibility = PortalVisibility.resolve(callerVisibility, parent == null ? null : parent.getVisibility());
```
MateuszJozwiak-Incubly, line 105: PortalVisibility.resolve() (PortalVisibility.java:25-29) is just a default-picker — caller value, else parent's, else PUBLIC. It never rejects anything. The  
  actual "PUBLIC child under PRIVATE parent is rejected" rule lives only in ParentRule.validateParent(), gated by appliesTo(parentId != null).                             
                                                                                                                                                                           
  CreateOrUpdatePortalLinkUseCase/CreateOrUpdateApiLinkUseCase build their pre-materialize validation objects with title/url only — no parentId, no visibility — so        
  ParentRule never fires for links. PortalLinkSyncDomainService doesn't call ParentRule at all; it resolves the real persisted parent and just calls resolve() at line 104,
  so an explicit PUBLIC is silently persisted under a persisted PRIVATE folder, contradicting open-api.yaml:3365. Same gap on the API-link path and on both dry-run twins  
  (Validate*LinkUseCase), since they call the same validateForConflicts/validateForConflictsForApi.                                                                        
                                                                                                                                                                           
  Fix: add the same check from ParentRule (PUBLIC effective + PRIVATE parent → parentMustBePublic) into validateForConflicts/validateForConflictsForApi, which all four    
  callers already invoke before materializing.

konrad-krol-incubly: Fixed: instead of duplicating the check on the sync path, the link use cases now hand a fully-populated create/update to the shared validator, so ParentRule enforces PUBLIC-under-PRIVATE (and the other parent invariants) for links the same way it does everywhere else.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/use_case/GetApiDocumentationUseCaseTest.java
```diff
@@ -47,11 +48,12 @@ class GetApiDocumentationUseCaseTest {
     private static final String API_ID = "00000000-0000-0000-0000-0000000000a1";
 
     private final PortalPageContentQueryServiceInMemory queryService = new PortalPageContentQueryServiceInMemory();
+    private final PortalNavigationItemsQueryServiceInMemory navigationItemsQueryService = new PortalNavigationItemsQueryServiceInMemory();
     private GetApiDocumentationUseCase useCase;
 
     @BeforeEach
     void setUp() {
-        useCase = new GetApiDocumentationUseCase(queryService);
+        useCase = new GetApiDocumentationUseCase(queryService, navigationItemsQueryService);
```
dev-marek, line 60: **Improvement — test coverage:** `GetApiDocumentationUseCase.Output` gained a `visibility` field in this PR (populated from the persisted nav item), but no test in this file asserts it — the use case is only exercised for `pageContent`. Same gap on `GetPortalDocumentationUseCaseTest.java` (untouched by this PR) for the `Output.visibility()` addition there.

More broadly: the resolve-and-persist behavior for `visibility` is verified with real RED→GREEN tests only for the Link family (`CreateOrUpdateApiLinkUseCaseTest`, `CreateOrUpdatePortalLinkUseCaseTest`, `PortalLinkSyncDomainServiceTest`). Every touched test for Documentation, top-level `PortalNavigationPath`/folders (`NavigationSyncPlannerTest`, `NavigationSyncPlanExecutorTest`, `PortalNavigationSyncDomainServiceTest`), and `PortalListingApiEntry` passes the new visibility parameter as `null` — so the non-null/explicit-visibility path for 3 of the 4 target schemas, plus the whole GET side, isn't covered by a test that would fail without this change.

dev-marek (commented):
Peer review: the write-path logic (resolve-with-inheritance, PUBLIC-under-PRIVATE rejection, backward compatibility) is correct and well-tested for the Link family. Left a few inline notes — mainly a test-coverage gap for the Documentation/PortalNavigationPath/PortalListingApiEntry paths and the GET side, plus two small nits. Nothing blocking; CI is green.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/mapper/ApiDocumentationMapperTest.java
```diff
@@ -102,7 +102,7 @@ void get_state_copies_persisted_entity_fields() {
             meta
         );
 
-        var state = ApiDocumentationMapper.INSTANCE.toDocumentationState(doc, "getting-started", "pets-api");
+        var state = ApiDocumentationMapper.INSTANCE.toDocumentationState(doc, null, "getting-started", "pets-api");
```
dev-marek, line 105: **Nit — test coverage:** the new `toDomainVisibility`/`toWireVisibility` conversions (here and in `PortalDocumentationMapperTest`/`PortalLinkMapperTest`) are only ever exercised with `null`. No test asserts a `PUBLIC`/`PRIVATE` value survives the wire↔domain round trip.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalLinkSyncDomainService.java
```diff
@@ -145,6 +150,12 @@ public void validateForConflicts(AuditInfo auditInfo, String portalId, String li
         );
     }
 
+    @jakarta.annotation.Nullable
```
dev-marek: **Nit — clean code:** `@jakarta.annotation.Nullable` is fully-qualified here and at line 198, but this file has no prior `Nullable` import/usage to justify avoiding one — unlike the `PortalVisibility` dual-package FQN elsewhere in this migration, there's no naming clash. Per repo convention ("explicit imports — no fully-qualified names in bodies"), this should just be `import jakarta.annotation.Nullable;` + `@Nullable`.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/model/PortalVisibilityTest.java
```diff
@@ -17,6 +17,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import io.gravitee.apim.core.portal.model.PortalVisibility;
```
dev-marek: **Nit:** this test class still lives under `io.gravitee.apim.core.portal_page.model` (its old package/directory), even though the class under test moved to `io.gravitee.apim.core.portal.model` in this PR. Worth relocating to `.../src/test/java/io/gravitee/apim/core/portal/model/PortalVisibilityTest.java` to match.

dev-marek (approved):

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/use_case/GetApiDocumentationUseCaseTest.java
konrad-krol-incubly: Fair. Added PUBLIC/PRIVATE assertions in `GetApiDocumentationUseCaseTest`, `GetPortalDocumentationUseCaseTest`, `NavigationSyncPlannerTest` and the listing sync test. Non-null path now covers all four schemas.

gravitee-apim-rest-api/gravitee-apim-rest-api-automation/gravitee-apim-rest-api-automation-rest/src/test/java/io/gravitee/apim/rest/api/automation/mapper/ApiDocumentationMapperTest.java
konrad-krol-incubly: Good catch, added PUBLIC and PRIVATE round-trips in all three mapper tests.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_page/domain_service/PortalLinkSyncDomainService.java
konrad-krol-incubly: Fixed, imported.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/test/java/io/gravitee/apim/core/portal_page/model/PortalVisibilityTest.java
konrad-krol-incubly: Moved.

gravitee-apim-rest-api/gravitee-apim-rest-api-service/src/main/java/io/gravitee/apim/core/portal_listing/domain_service/NavigationItemEntryMaterializer.java
```diff
@@ -52,10 +52,11 @@ class NavigationItemEntryMaterializer {
     PortalNavigationApi upsert(AuditInfo auditInfo, PortalId portalId, String apiId, PortalListingApiEntry entry) {
         var navApiId = rowId(auditInfo, portalId, apiId);
         var parent = resolveParent(auditInfo, portalId.toString(), entry.location());
+        var visibility = PortalVisibility.resolve(entry.visibility(), parent == null ? null : parent.getVisibility());
```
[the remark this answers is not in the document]
konrad-krol-incubly: Leaving it — inlined the call is more readable than a two-line private helper, same result. Will re-extract if a fourth call site shows up.

